PR #36, opened 2018-10-03
Upgrade to rand-0.5

djc:


quinn-proto/src/endpoint.rs
```diff
@@ -213,7 +213,7 @@ impl Endpoint {
                 config,
                 io: VecDeque::new(),
                 // session_ticket_buffer,
-                initial_packet_number: distributions::Range::new(0, 2u64.pow(32) - 1024),
+                initial_packet_number: distributions::Standard,
```
Ralith: Why the semantic change here?

djc: I figured random initial packet numbers are going away soon, anyway, so I felt I could get away with not figuring out how to make it get it right. But if you think it's important for current usage, I can spend a little time on it.

Ralith: Should just be a matter of dropping in [Uniform](https://docs.rs/rand/*/rand/distributions/uniform/struct.Uniform.html).

djc: Ah, don't know how I missed that this afternoon. Should be better now!

PR #45, opened 2018-10-04
Store ConnectionHandles in Connections

djc:
So we don't have to keep passing it around.


djc:
It's still a bit of a layering violation that I'd like to improve on, but it seems better for now.

Issue #32, opened 2018-10-03
Don't set hq as the default ALPN protocol in quinn-proto


djc:
Leave this to higher API levels to decide.

stammw:
Hello,

I'd like to try to solve this one !

If understand well, this ALPN parameter could be set in endpoint building code like so:

```rust
builder
    .logger(log.clone())
    .config(quinn::Config {
        max_remote_bi_streams: 64,
        alpn_protocols: vec!["hq-11"],
        ..Default::default()
}).listen();
```

or

```rust
let mut builder = quinn::Endpoint::new();
builder.set_alpn_protocols(vec!["hq-11"]);
```

These two options would let a possibility for the the user not to give any ALPN value. Is that desirable, or is it better to make this parameter mandatory?

djc:
Hi @stammw, that would be great! I think we would want a method on `EndpointBuilder` that operates similar to `set_certificate()` and `enable_keylog()`: something that mutates the `ClientConfig` contained in the existing `Config`. @Ralith any thoughts?

stammw:
Then TLS's ALPN parameter would be left empty when this method has never been called ?

djc:
Yeah, I think that would be okay for some use cases, but I'll await the other reviewer.

Ralith:
Yes, to the best of my knowledge ALPN is an optional part of the handshake.

djc:
This has been fixed in #47. Thanks, @stammw!

stammw:
Thank you both for all the time spent reviewing and helping !

PR #47, opened 2018-10-05
Let the user select ALPN protocols values #32

stammw:
I think it might do the job ?

It may be interesting to offer features depending on what protocol has been negotiated?

### DIFF
--- a/quinn-proto/src/tests.rs
+++ b/quinn-proto/src/tests.rs
@@ -88,6 +88,7 @@ fn server_config() -> Config {
     };
 
     let mut tls_server_config = crypto::build_server_config();
+    tls_server_config.set_protocols(&["hq-11".into()]);
     tls_server_config
         .set_single_cert(certs, keys[0].clone())
         .unwrap();
@@ -106,6 +107,7 @@ fn client_config() -> Config {
     let anchor_vec = vec![anchor];
 
     let mut tls_client_config = crypto::build_client_config();
+    tls_client_config.set_protocols(&["hq-11".into()]);
     tls_client_config
         .root_store
         .add_server_trust_anchors(&webpki::TLSServerTrustAnchors(&anchor_vec));
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -302,6 +302,14 @@ impl<'a> EndpointBuilder<'a> {
         Ok(self)
     }
 
+    pub fn set_protocols(&mut self, protocols: &[String]) -> &mut Self {
+        {
+            let tls_server_config = Arc::get_mut(&mut self.config.tls_server_config).unwrap();
+            tls_server_config.set_protocols(protocols);
+        }
+        self
+    }
+
     pub fn from_socket(
         self,
         socket: std::net::UdpSocket,
--- a/quinn-proto/src/crypto.rs
+++ b/quinn-proto/src/crypto.rs
@@ -84,14 +84,12 @@ pub fn build_client_config() -> ClientConfig {
         .root_store
         .add_server_trust_anchors(&webpki_roots::TLS_SERVER_ROOTS);
     config.versions = vec![ProtocolVersion::TLSv1_3];
-    config.alpn_protocols = vec![ALPN_PROTOCOL.into()];
     config.key_log = Arc::new(KeyLogFile::new());
     config
 }
 
 pub fn build_server_config() -> ServerConfig {
     let mut config = ServerConfig::new(NoClientAuth::new());
-    config.set_protocols(&[ALPN_PROTOCOL.into()]);
     config.key_log = Arc::new(KeyLogFile::new());
     config
 }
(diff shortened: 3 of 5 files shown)


stammw:
Should I write a test in for code coverage to pass?

quinn-proto/src/tests.rs
```diff
@@ -88,6 +88,7 @@ fn server_config() -> Config {
     };
 
     let mut tls_server_config = crypto::build_server_config();
+    tls_server_config.set_protocols(&["hq-11".into()]);
```
djc: I think we probably still want `"hq-11"` defined somewhere as a constant instead of writing it as a literal all over the place. Maybe a constant named `ALPN_QUIC_HTTP` in `quinn-proto/src/lib.rs`?

djc:
As for coverage, I don't think you have to add a test here. Reduced coverage is largely because we don't have any tests for `quinn` yet.

quinn-proto/src/tests.rs
Ralith: Ideally this constant would live in a separate HTTP-over-QUIC crate, but this'll do for the short term.

stammw:
Should I change the API to using &[u8] even if rustls still uses Strings ?

quinn-proto/src/tests.rs
djc: The usages in `quinn-proto/src/tests.rs` should also use the constant. Can you update your second commit, please?

Ralith:
Yeah (or `Vec<u8>`); it's fine to panic if a `&str`/`String` can't be constructed for now, and hopefully rustls will get fixed.

quinn/src/lib.rs
```diff
@@ -304,10 +304,14 @@ impl<'a> EndpointBuilder<'a> {
         Ok(self)
     }
 
-    pub fn set_protocols(&mut self, protocols: &[String]) -> &mut Self {
+    pub fn set_protocols(&mut self, protocols: &[Vec<u8>]) -> &mut Self {
```
djc: We should probably take `&[&[u8]]` here instead, to prevent having more allocations than necessary.

quinn/src/lib.rs
```diff
@@ -304,10 +304,14 @@ impl<'a> EndpointBuilder<'a> {
...
         {
             let tls_server_config = Arc::get_mut(&mut self.config.tls_server_config).unwrap();
-            tls_server_config.set_protocols(protocols);
+            let protocols_strings = protocols
+                .iter()
+                .map(|p| String::from_utf8(p.to_vec()).unwrap())
+                .collect::<Vec<String>>();
```
djc: Let's replace the `String` with `_` here to only have minimal type annotations.

quinn-proto/src/lib.rs
```diff
@@ -59,7 +58,9 @@ pub use transport_error::Error as TransportError;
 pub const VERSION: u32 = 0xff00_000b;
 
 /// TLS ALPN value for HTTP over QUIC
-pub const ALPN_QUIC_HTTP: &str = "hq-11";
+lazy_static! {
+    pub static ref ALPN_QUIC_HTTP: Vec<u8> = "hq-11".into();
```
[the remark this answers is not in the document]
stammw: Thank you for the advice.

quinn-proto/src/tests.rs
```diff
@@ -88,6 +88,7 @@ fn server_config() -> Config {
     };
 
     let mut tls_server_config = crypto::build_server_config();
+    tls_server_config.set_protocols(&[String::from_utf8(ALPN_QUIC_HTTP.to_vec()).unwrap()]);
```
Ralith: This would be cleaner as `str::from_utf8(ALPN_QUIC_HTTP).unwrap().into()` rather than bouncing through `Vec`

quinn-proto/src/tests.rs
```diff
@@ -106,6 +107,7 @@ fn client_config() -> Config {
     let anchor_vec = vec![anchor];
 
     let mut tls_client_config = crypto::build_client_config();
+    tls_client_config.set_protocols(&["hq-11".into()]);
```
Ralith: Should use the constant.

quinn/src/lib.rs
```diff
@@ -302,6 +304,18 @@ impl<'a> EndpointBuilder<'a> {
         Ok(self)
     }
 
+    pub fn set_protocols(&mut self, protocols: &[&[u8]]) -> &mut Self {
+        {
+            let tls_server_config = Arc::get_mut(&mut self.config.tls_server_config).unwrap();
+            let protocols_strings = protocols
+                .iter()
+                .map(|p| String::from_utf8(p.to_vec()).unwrap())
```
Ralith: `str::from_utf8(ALPN_QUIC_HTTP).unwrap().into()`

PR #50, opened 2018-10-07
Remove ArrayVec in favor of a simple array

### LINKED ISSUE #31: Replace ArrayVec in ConnectionId with a simple array
This is what I had originally in quinn. This would get rid of the extra dependency, and it would let `ConnectionId` implement `Copy`, which would obviate the need for cloning.

twilco:
Closes #31.

### DIFF
--- a/quinn-proto/src/crypto.rs
+++ b/quinn-proto/src/crypto.rs
@@ -473,12 +473,7 @@ mod test {
 
     #[test]
     fn key_derivation() {
-        let id = ConnectionId(
-            [0x83, 0x94, 0xc8, 0xf0, 0x3e, 0x51, 0x57, 0x08]
-                .iter()
-                .cloned()
-                .collect(),
-        );
+        let id = ConnectionId::new(&[0x83, 0x94, 0xc8, 0xf0, 0x3e, 0x51, 0x57, 0x08]);
         let digest = &digest::SHA256;
         let cipher = &aead::AES_128_GCM;
         let hs_secret = handshake_secret(&id);
--- a/quinn-proto/src/packet.rs
+++ b/quinn-proto/src/packet.rs
@@ -1,6 +1,5 @@
 use std::{fmt, io, str};
 
-use arrayvec::ArrayVec;
 use bytes::{BigEndian, Buf, BufMut, ByteOrder, Bytes, BytesMut};
 use rand::Rng;
 use slog;
@@ -227,9 +226,9 @@ impl Packet {
                     ));
                 }
                 buf.copy_to_slice(&mut cid_stage[0..dcil as usize]);
-                let destination_id = ConnectionId::new(cid_stage, dcil as usize);
+                let destination_id = ConnectionId::new(&cid_stage[..dcil as usize]);
                 buf.copy_to_slice(&mut cid_stage[0..scil as usize]);
-                let source_id = ConnectionId::new(cid_stage, scil as usize);
+                let source_id = ConnectionId::new(&cid_stage[..scil as usize]);
                 match version {
                     0 => (
                         buf.position() as usize,
@@ -271,8 +270,8 @@ impl Packet {
                         "destination connection ID longer than packet",
                     ));
                 }
-                buf.copy_to_slice(&mut cid_stage[0..dest_id_len]);
-                let id = ConnectionId::new(cid_stage, dest_id_len);
+                buf.copy_to_slice(&mut cid_stage[..dest_id_len]);
+                let id = ConnectionId::new(&cid_stage[..dest_id_len]);
                 let key_phase = ty & KEY_PHASE_BIT != 0;
                 let number = match ty & 0b11 {
                     0x0 => PacketNumber::U8(buf.get()?),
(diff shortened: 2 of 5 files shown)


twilco:
Hi @djc!  I was able to remove `ArrayVec`, but haven't been able to get the tests to pass after a bit of digging.  Could you help point me in the right direction?  Let's take the `stop_stream` failure as an example:

```
---- tests::stop_stream stdout ----
INFO connecting
DEBG dropping packet from unrecognized connection, side: Client, header: Long { ty: 125, source_id: ConnectionId { len: 18, bytes: [236, 31, 120, 215, 219, 214, 88, 142, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] }, destination_id: ConnectionId { len: 18, bytes: [160, 191, 201, 164, 234, 127, 204, 18, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] }, number: 738246780 }
```
"dropping packet from unrecognized connection" is clearly a bad sign.  Something seems to be up with how `Endpoint`s `connection_ids`, `connection_ids_initial`, and `connection_remotes` are being populated, but I haven't been able to determine what.  Is there anything that jumps out to you in this diff that could cause this?

twilco:
Thanks @Ralith!  I see why my previous implementation was wrong.  That indeed fixed one test - now eight to go.  The ones failing all look similar to that of the one above.

Ralith:
The `Eq`, `PartialEq`, `Ord`, `PartialOrd`, and `Hash` implementations should probably have the same fix applied.

quinn-proto/src/packet.rs
```diff
@@ -271,8 +270,8 @@ impl Packet {
                         "destination connection ID longer than packet",
                     ));
                 }
-                buf.copy_to_slice(&mut cid_stage[0..dest_id_len]);
-                let id = ConnectionId::new(cid_stage, dest_id_len);
+                buf.copy_to_slice(&mut cid_stage[..dest_id_len]);
+                let id = ConnectionId::new(&cid_stage);
```
Ralith: s/`&cid_stage`/`&cid_stage[..dest_id_len]`/

quinn-proto/src/packet.rs
```diff
@@ -227,9 +226,9 @@ impl Packet {
                     ));
                 }
                 buf.copy_to_slice(&mut cid_stage[0..dcil as usize]);
-                let destination_id = ConnectionId::new(cid_stage, dcil as usize);
+                let destination_id = ConnectionId::new(&cid_stage);
                 buf.copy_to_slice(&mut cid_stage[0..scil as usize]);
-                let source_id = ConnectionId::new(cid_stage, scil as usize);
+                let source_id = ConnectionId::new(&cid_stage);
```
Ralith: s/`&cid_stage`/`&cid_stage[..scil as usize]`/

quinn-proto/src/packet.rs
```diff
@@ -227,9 +226,9 @@ impl Packet {
                     ));
                 }
                 buf.copy_to_slice(&mut cid_stage[0..dcil as usize]);
-                let destination_id = ConnectionId::new(cid_stage, dcil as usize);
+                let destination_id = ConnectionId::new(&cid_stage);
```
Ralith: s/`&cid_stage`/`&cid_stage[..dcil as usize]`/

quinn-proto/src/packet.rs
```diff
@@ -309,41 +308,53 @@ impl Packet {
...
+        rng.fill_bytes(&mut rng_bytes);
+        (&mut res.bytes[..len as usize]).clone_from_slice(&rng_bytes[..len as usize]);
+        res
     }
 }
 
 impl fmt::Display for ConnectionId {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        for byte in &self.0 {
+        for byte in &self.bytes {
```
Ralith: This, too, should be bounded by `len`, perhaps by dispatching to the `Deref` impl if that's convenient.

twilco:
And we've got a ✅.  Mind taking another look, @Ralith?

quinn-proto/src/crypto.rs
```diff
@@ -473,12 +473,8 @@ mod test {
...
     #[test]
     fn key_derivation() {
-        let id = ConnectionId(
-            [0x83, 0x94, 0xc8, 0xf0, 0x3e, 0x51, 0x57, 0x08]
-                .iter()
-                .cloned()
-                .collect(),
-        );
+        let data = [0x83, 0x94, 0xc8, 0xf0, 0x3e, 0x51, 0x57, 0x08];
+        let id = ConnectionId::new(&data);
```
Ralith: If `data` isn't needed elsewhere, this should probably just be a oneliner.

quinn-proto/src/endpoint.rs
```diff
@@ -317,7 +317,6 @@ impl Endpoint {
         //
         // Handle packet on existing connection, if any
         //
-
```
Ralith: Accidental deletion?

quinn-proto/src/packet.rs
```diff
@@ -309,41 +308,53 @@ impl Packet {
...
-        let mut x = ConnectionId(data.into());
-        x.0.truncate(len);
-        x
+    pub fn new(bytes: &[u8]) -> Self {
+        debug_assert!(bytes.is_empty() || (bytes.len() > 3 && bytes.len() <= MAX_CID_SIZE));
+        let mut res = Self {
+            len: bytes.len() as u8,
+            bytes: [0; MAX_CID_SIZE],
+        };
+        (&mut res.bytes[..bytes.len()]).clone_from_slice(&bytes);
```
Ralith: Is the `(&mut ...)` trick really necessary?

quinn-proto/src/packet.rs
```diff
@@ -309,41 +308,53 @@ impl Packet {
...
+        rng.fill_bytes(&mut rng_bytes);
+        (&mut res.bytes[..len as usize]).clone_from_slice(&rng_bytes[..len as usize]);
+        res
     }
 }
 
 impl fmt::Display for ConnectionId {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        for byte in &self.0 {
+        for byte in (*(&self)).iter() {
```
Ralith: Could this be written `self.iter()`?

djc:
Also let's squash all these commits before merging?

djc:
This looks great!

twilco:
Thank you both for your help - I learned a lot!

djc:
@twilco would you like to dig in a little more with another issue?

twilco:
@djc, sure, I'd love to.  What did you have in mind?

djc:
I've just written up some notes in #51, have a look to see if that's something you're interested in.

PR #52, opened 2018-10-09
Simplify error handling by returning Result from handle_connected_inner

### LINKED ISSUE #51: Factor out common error handling code
There is a lot of repetition in error handling code in methods of the `Connection` impl. There are multiple ways this could be improved, in increasing order of complexity:

* Factor them out into one or more utility methods
* Create a nice error type that this returned to `Endpoint` so it do the proper handling in one place
* In doing so, make it so that the `Connection` no longer has to know its `ConnectionHandle`

@Ralith does that make sense to you? @twilco this is a little more open-ended, but I think you can ramp it up by taking it one step at a time to gain some familiarity with the issues.

twilco:
Prior to this patch, we had duplicated lots of error handling logic inside `handle_connected_inner`, mostly in the way of event recording.  This patch simplifies this by instead returning a `Result`, which allows the caller to handle either case and create events more succinctly than `handle_connected_inner` could.

Also, this changeset moves the `app_closed` property from `connection::State` to `connection::Connection` itself, as it is much simpler to manage that piece of state on the `Connection` itself. 

Closes #51.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -968,11 +968,10 @@ impl Connection {
                                 Ok(()) => {
                                     self.on_packet_authenticated(ctx, now, number as u64);
                                     trace!(ctx.log, "resending ClientHello"; "remote_id" => %remote_id);
-                                    let local_id = self.local_id.clone();
                                     // Discard transport state
                                     let mut new = Connection::new(
-                                        remote_id.clone(),
-                                        local_id,
+                                        remote_id,
+                                        self.local_id,
                                         remote_id,
                                         remote,
                                         ctx.initial_packet_number.sample(&mut ctx.rng),
@@ -1094,11 +1093,7 @@ impl Connection {
                                     debug!(ctx.log, "remote didn't send transport params");
                                     return Err(TransportError::TLS_HANDSHAKE_FAILED.into());
                                 }
-                                trace!(
-                                    ctx.log,
-                                    "{connection} established",
-                                    connection = id.clone()
-                                );
+                                trace!(ctx.log, "{connection} established", connection = id);
                                 self.handshake_cleanup(&ctx.config);
                                 let mut msgs = Vec::new();
                                 state.tls.write_tls(&mut msgs).unwrap();
@@ -1230,7 +1225,7 @@ impl Connection {
                 }
             }
             State::Established(mut state) => {
-                let id = self.local_id.clone();
+                let id = self.local_id;
                 if let Header::Long { .. } = packet.header {
                     trace!(ctx.log, "discarding unprotected packet"; "connection" => %id);
                     return Ok(State::Established(state));
@@ -1306,12 +1301,12 @@ impl Connection {
         payload: Bytes,
         tls: &mut TlsSession,
     ) -> Result<bool, state::CloseReason> {
-        let cid = self.local_id.clone();
+        let cid = self.local_id;
         for frame in frame::Iter::new(payload) {
             match frame {
                 Frame::Padding => {}
                 _ => {
-                    trace!(ctx.log, "got frame"; "connection" => cid.clone(), "type" => %frame.ty());
+                    trace!(ctx.log, "got frame"; "connection" => cid, "type" => %frame.ty());
                 }
             }
             match frame {
@@ -1610,7 +1605,7 @@ impl Connection {
                 Frame::NewConnectionId { .. } => {
                     if self.remote_id.is_empty() {
                         debug!(ctx.log, "got NEW_CONNECTION_ID for connection {connection} with empty remote ID",
-                               connection=self.local_id.clone());
+                               connection=self.local_id);
                         ctx.events.push_back((
                             self.handle,
                             Event::ConnectionLost {
@@ -1676,8 +1671,8 @@ impl Connection {
                 Header::Long {
                     ty,
                     number: number as u32,
-                    source_id: self.local_id.clone(),
-                    destination_id: self.remote_id.clone(),
+                    source_id: self.local_id,
+                    destination_id: self.remote_id,
                 }.encode(&mut buf);
                 pending = &mut self.handshake_pending;
                 crypto = &self.handshake_crypto;
@@ -1706,7 +1701,7 @@ impl Connection {
                 } else {*/
                 crypto = self.crypto.as_ref().unwrap();
                 Header::Short {
-                    id: self.remote_id.clone(),
+                    id: self.remote_id,
                     number: PacketNumber::new(number, self.largest_acked_packet),
                     key_phase: self.key_phase,
                 }.encode(&mut buf);
@@ -1927,7 +1922,7 @@ impl Connection {
         let number = self.get_tx_number();
         let mut buf = Vec::new();
         Header::Short {
-            id: self.remote_id.clone(),
+            id: self.remote_id,
             number: PacketNumber::new(number, self.largest_acked_packet),
             key_phase: self.key_phase,
         }.encode(&mut buf);
@@ -1956,7 +1951,7 @@ impl Connection {
         let number = self.get_tx_number();
         let mut buf = Vec::new();
         Header::Short {
-            id: self.remote_id.clone(),
+            id: self.remote_id,
             number: PacketNumber::new(number, self.largest_acked_packet),
             key_phase: self.key_phase,
         }.encode(&mut buf);
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -1,5 +1,6 @@
 use std::collections::{hash_map, BTreeMap, VecDeque};
 use std::net::SocketAddrV6;
+use std::sync::Arc;
 use std::{cmp, io, mem};
 
 use bytes::{Buf, Bytes, BytesMut};
@@ -8,7 +9,9 @@ use rand::distributions::Distribution;
 use slog::Logger;
 
 use coding::{BufExt, BufMutExt};
-use crypto::{reset_token_for, ConnectError, Crypto, TLSError, TlsSession, ACK_DELAY_EXPONENT};
+use crypto::{
+    reset_token_for, ClientConfig, ConnectError, Crypto, TLSError, TlsSession, ACK_DELAY_EXPONENT,
+};
 use endpoint::{Config, Context, Event, Io, Timer};
 use packet::{
     set_payload_length, types, ConnectionId, Header, Packet, PacketNumber, AEAD_TAG_SIZE,
@@ -378,12 +381,15 @@ impl Connection {
     }
 
     /// Initiate a connection
-    pub fn connect(&mut self, ctx: &Context, server_name: &str) -> Result<(), ConnectError> {
-        let mut tls = TlsSession::new_client(
-            &ctx.config.tls_client_config,
-            server_name,
-            &TransportParameters::new(&ctx.config),
-        ).unwrap();
+    pub fn connect(
+        &mut self,
+        ctx: &Context,
+        config: &Arc<ClientConfig>,
+        server_name: &str,
+    ) -> Result<(), ConnectError> {
+        let mut tls =
+            TlsSession::new_client(config, server_name, &TransportParameters::new(&ctx.config))
+                .unwrap();
         self.server_name = Some(server_name.into());
         let mut outgoing = Vec::new();
         tls.write_tls(&mut outgoing).unwrap();
@@ -982,16 +988,11 @@ impl Connection {
                                     mem::replace(self, new);
                                     // Send updated ClientHello
                                     let mut outgoing = Vec::new();
-                                    let mut tls = TlsSession::new_client(
-                                        &ctx.config.tls_client_config,
-                                        self.server_name.as_ref().unwrap(),
-                                        &TransportParameters::new(&ctx.config),
-                                    ).unwrap();
                                     state.tls.write_tls(&mut outgoing).unwrap();
                                     self.transmit_handshake(&outgoing);
                                     // Prepare to receive Handshake packets that start stream 0 from offset 0
                                     Ok(State::Handshake(state::Handshake {
-                                        tls,
+                                        tls: state.tls,
                                         clienthello_packet: state.clienthello_packet,
                                         remote_id_set: state.remote_id_set,
                                     }))
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -542,7 +542,7 @@ impl Connection {
                 self.streams.get_mut(&id).unwrap().send_mut().unwrap().state =
                     stream::SendState::ResetRecvd { stop_reason };
                 if stop_reason.is_none() {
-                    self.maybe_cleanup(id);
+                    self.maybe_cleanup(config, id);
                 }
             }
         }
@@ -562,7 +562,7 @@ impl Connection {
                 }
             };
             if recvd {
-                self.maybe_cleanup(frame.id);
+                self.maybe_cleanup(config, frame.id);
                 self.finished_streams.push(frame.id);
             }
         }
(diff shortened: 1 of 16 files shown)


quinn-proto/src/connection.rs
```diff
@@ -1139,10 +1097,10 @@ impl Connection {
                                                 .into(),
                                         },
                                     ));
-                                    return State::handshake_failed(
+                                    return Ok(State::handshake_failed(
```
djc: @Ralith maybe I got this wrong during the TLS overhaul, should we be returning `TRANSPORT_PARAMETER_ERROR` here instead of `TLS_HANDSHAKE_FAILED`?

quinn-proto/src/connection.rs
```diff
@@ -1258,71 +1210,59 @@ impl Connection {
...
                             if version == VERSION {
                                 // Our version is supported, so this packet is spurious
-                                return State::Handshake(state);
+                                return Ok(State::Handshake(state));
                             }
                         }
                         debug!(ctx.log, "remote doesn't support our version");
+                        // TODO: What do we do here?  If we return Err(ConnectionError::VersionMismatch),
+                        // the caller won't be able to get the same `State::Draining` without the
+                        // `state.into()` we have here.  Maybe we leave it for now?
```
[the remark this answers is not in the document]
twilco: Yeah, that's a good idea.  Maybe a new `StatefulConnectionError` that contains the `ConnectionError` and some `State`.  We'd be able to clean up `handle_connected_inner` entirely with something like that.

quinn-proto/src/endpoint.rs
```diff
@@ -595,17 +595,35 @@ impl Endpoint {
         packet: Packet,
     ) {
         trace!(self.ctx.log, "connection got packet"; "connection" => %self.connections[conn.0].local_id, "len" => packet.payload.len());
+        // TODO: Ask quinn owners why we use conn.0 instead of conn.initial_id.  For succinctness?
```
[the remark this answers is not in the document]
twilco: Got it 👍

quinn-proto/src/connection.rs
Ralith: I don't think so. I don't see this case called out specifically in the draft, but `TRANSPORT_PARAMETER_ERROR` describes malformed transport parameters, whereas their absence seems better conveyed by a TLS `missing_extension` alert.

quinn-proto/src/connection.rs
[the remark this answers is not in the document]
Ralith: Perhaps there's a simpler solution? Constructions of `state::Draining` using `into` exist only to help ensure that `app_closed` is propagated between states. This flag is only set by `Connection::close`, and could be pulled out into the larger `Connection` struct at little cost. Its only function is to ensure we don't discard connection state in a way that is unpredictable to the application.

quinn-proto/src/connection.rs
Ralith: Which I now realize probably means that the event should be changed to be consistent.

twilco: Cool, will update event to be `TransportError::TLS_HANDSHAKE_FAILED.into()` then.

quinn-proto/src/connection.rs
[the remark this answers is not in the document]
twilco: Okay, that sounds like it will work.  I'll give it a go 👍

twilco:
Alright guys, I've moved `app_closed` to `Connection` and further simplified `handle_connected_inner`.  Let me know if this is what you had in mind.

quinn-proto/src/connection.rs
```diff
@@ -31,6 +31,10 @@ impl From<ConnectionHandle> for usize {
 }
 
 pub struct Connection {
+    // TODO: Is there a better name for this?  I feel like there is, but I don't know enough about
+    // the way this struct is used to know if the knowledge of "app" makes sense here.
+    // Just `closed`?
+    pub app_closed: bool,
```
Ralith: There's a bunch of different states a connection can be in that qualify as "closed" (see `State::is_closed`). This state is for tracking whether the next layer up has explicitly called `close`, which is used as a signal that it will not make any further calls with that `ConnectionHandle`, and so its resources may be freed as soon as internal logic and the peer no longer needs it.

quinn-proto/src/connection.rs
```diff
@@ -1185,21 +1138,15 @@ impl Connection {
...
                             Err(e) => {
                                 debug!(ctx.log, "handshake failed"; "reason" => %e);
-                                ctx.events.push_back((
-                                    self.handle,
-                                    Event::ConnectionLost {
-                                        reason: TransportError::TLS_HANDSHAKE_FAILED.into(),
-                                    },
-                                ));
-                                State::handshake_failed(TransportError::TLS_HANDSHAKE_FAILED, None)
+                                return Err(TransportError::TLS_HANDSHAKE_FAILED.into());
```
Ralith: s/`return ...;`/`...`/

quinn-proto/src/connection.rs
```diff
@@ -2071,25 +1995,17 @@ impl Connection {
...
-            State::Established(_) => State::Closed(state::Closed {
-                reason,
-                app_closed: true,
             }),
-            State::Closed(x) => State::Closed(state::Closed {
-                app_closed: true,
-                ..x
-            }),
-            State::Draining(_) => State::Draining(state::Draining { app_closed: true }),
+            State::HandshakeFailed(x) => State::HandshakeFailed(state::HandshakeFailed { ..x }),
```
Ralith: s/`SomeStruct { ..x}`/`x`/

quinn-proto/src/connection.rs
```diff
@@ -2071,25 +1995,17 @@ impl Connection {
...
+            State::Established(_) => State::Closed(state::Closed { reason }),
+            State::Closed(x) => State::Closed(state::Closed { ..x }),
```
Ralith: s/`SomeStruct { ..x}`/`x`/

quinn-proto/src/connection.rs
```diff
@@ -2488,6 +2404,17 @@ impl From<ConnectionError> for io::Error {
     }
 }
 
+impl From<state::CloseReason> for ConnectionError {
```
Ralith: Where is this used?

quinn-proto/src/endpoint.rs
```diff
@@ -597,15 +597,51 @@ impl Endpoint {
...
             now,
             remote,
             packet,
             state,
-        );
+        ) {
+            Ok(state) => match state {
+                State::Closed(closed) => {
+                    let copy = closed.clone();
+                    err_to_log = Some(closed.reason.into());
```
Ralith: We don't want to emit a new `ConnectionLost` event every time we receive a packet.

quinn-proto/src/endpoint.rs
```diff
@@ -597,15 +597,51 @@ impl Endpoint {
...
+            Err(conn_err) => match conn_err {
+                ConnectionError::TransportError { error_code } => {
+                    err_to_log = Some(error_code.into());
+                    State::handshake_failed(error_code, None)
+                }
+                ConnectionError::VersionMismatch => {
+                    err_to_log = Some(ConnectionError::VersionMismatch);
+                    // TODO: Just curious - why do we start draining after encountering a version
+                    // mismatch?  What are we draining - a QUIC stream?  What does it mean to drain
+                    // a stream?
```
Ralith: We're draining the whole connection. The `Draining` state is entered when the remote endpoint has closed the connection, and is used to gracefully discard any further packets received on that connection, which can occur due to reordering or packet loss. It completes after 3 times the retransmit time-out period, ending in `Drained`. See [the draft](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#rfc.section.6.13.1) for details.

The `Drained` state is appropriate following a version mismatch because the server has not actually *opened* a connection yet.

quinn-proto/src/connection.rs
```diff
@@ -1341,9 +1265,9 @@ impl Connection {
                         self.drive_tls(ctx, &mut state.tls)?;
                         Ok(x)
                     }) {
-                    Err(e) => State::closed(e),
-                    Ok(true) => State::Draining(state.into()),
-                    Ok(false) => State::Established(state),
+                    Err(e) => Ok(State::closed(e)),
+                    Ok(true) => Ok(State::Draining),
+                    Ok(false) => Ok(State::Established(state)),
                 }
```
Ralith: Could this whole block be rewritten to something like the following?
```rust
let closed = self.process_payload(ctx, now, number, payload.into(), &mut state.tls)?;
self.drive_tls(ctx, &mut state.tls)?;
Ok(if closed { State::Draining } else { State::Established(state) })
```
You'd want to remove the `ConnectionLost` events from `process_payload` if so, letting them be generated automatically like the rest.

quinn-proto/src/endpoint.rs
```diff
@@ -597,15 +597,51 @@ impl Endpoint {
...
+                    let copy = closed.clone();
+                    err_to_log = Some(closed.reason.into());
+                    State::Closed(copy)
+                }
+                _ => state,
+            },
+            Err(conn_err) => match conn_err {
+                ConnectionError::TransportError { error_code } => {
+                    err_to_log = Some(error_code.into());
+                    State::handshake_failed(error_code, None)
```
Ralith: We should only enter `HandshakeFailed` if we were initially in `Handshake`. If the connection was established, we should enter `Closed`.

quinn-proto/src/connection.rs
```diff
@@ -1258,71 +1205,48 @@ impl Connection {
...
                         trace!(ctx.log, "failed to authenticate packet"; "connection" => %id);
-                        return State::Established(state);
+                        return Ok(State::Established(state));
                     }
                     Err(Some(e)) => {
                         warn!(ctx.log, "got illegal packet"; "connection" => %id);
-                        ctx.events
-                            .push_back((self.handle, Event::ConnectionLost { reason: e.into() }));
-                        return State::closed(e);
+                        return Ok(State::closed(e));
```
Ralith: Why aren't we returning an error here? If we did, we could refactor `decrypt_packet` to return `Result<Option<(Vec<u8>, bool)>` and use `?` here for improved concision.

quinn-proto/src/endpoint.rs
```diff
@@ -597,15 +597,51 @@ impl Endpoint {
...
+        ) {
+            Ok(state) => match state {
+                State::Closed(closed) => {
+                    let copy = closed.clone();
+                    err_to_log = Some(closed.reason.into());
+                    State::Closed(copy)
+                }
+                _ => state,
+            },
+            Err(conn_err) => match conn_err {
```
Ralith: Why not unconditionally emit the `ConnectionLost` event regardless of the error type? You'll still need to branch to decide whether to enter `HandshakeFailed`, `Closed`, or `Draining`, but this should be more concise.

quinn-proto/src/connection.rs
djc: Using `and_then` still makes calling `drive_tls()` contingent on whether the result of `process_payload()` is `Ok()`, so you'd still need more pattern matching than what you have shown here.

Ralith: I don't follow. `self.process_payload(...)?` will cause the enclosing function to return immediately if the call doesn't return `Ok()`.

djc: Ah, you're right -- I missed that.

quinn-proto/src/endpoint.rs
twilco: Got it.  By being in this arm (`ConnectionError::TransportError`), do we know that a connection was established?  If so, I'll just change this to `State::Closed`.  Or do we have to check the previous state for `Handshake` from the initial `let state = self.connections[conn.0].state.take().unwrap();` line?

Ralith: > in the previous implementation, every time we logged a ConnectionError::TransportError, it returned a State::handshake_failed.

This is incorrect; for example, examine `Connection::process_payload`.

> By being in this arm (ConnectionError::TransportError), do we know that a connection was established?

Transport errors can arise at any time, so you'll want to inspect the previous state, yeah.

twilco: Alright.  If previous state was `State::Handshake`, we will return `State::handshake_failed(...)`.  Otherwise we'll return `State::closed(...)`.

quinn-proto/src/connection.rs
twilco: Using code you suggested in a previous comment in `Connection::handle_connected_inner -> Result<State, ConnectionError>`:

`let closed = self.process_payload(ctx, now, number, payload.into(), &mut state.tls)?;`

`Connection::process_payload` returns a `Result<bool, state::CloseReason>`, whereas `handle_connected_inner` returns a `Result<State, ConnectionError>`, so we need a way to get convert that `CloseReason` into a `ConnectionError`.

Will this be fine, or is there a better solution that you would suggest?

Ralith: That makes sense to me. I was mainly curious where it was used in your original diff.

twilco:
@Ralith - Addressed some of your review comments in my most recent commit...progress!

I was hoping to get your advice on something I encountered making these changes.  Here is the block of code in question:

```
let prev_state = self.connections[conn.0].state.take().unwrap();
let state = match self.connections[conn.0].handle_connected_inner(
    &mut self.ctx,
    now,
    remote,
    packet,
    prev_state,
) {
    Ok(state) => state,
    Err(conn_err) => match conn_err {
        ConnectionError::TransportError { error_code } => {
            err_to_log = Some(error_code.into());
            match prev_state {
                State::Handshake(_) => {
                    State::handshake_failed(error_code, None)
                },
                _ => {
                    State::closed(error_code)
                }
            }
        }
        ...more unrelated stuff...
```

As we discussed, we check the `prev_state` to see if we're in `Handshake` - if so, we return `State::handshake_failed`.  Otherwise, we return `State::closed`.  The problem is `prev_state` is moved into `handle_connected_inner`, so we can't use it in the `match prev_state` as seen above.

```
error[E0382]: use of moved value: `prev_state`                                                                                                                                                                                                                       
   --> quinn-proto/src/endpoint.rs:602:27                                                                                                                                                                                                                            
    |                                                                                                                                                                                                                                                                
596 |             prev_state,                                                                                                                                                                                                                                        
    |             ---------- value moved here                                                                                                                                                                                                                        
...                                                                                                                                                                                                                                                                  
602 |                     match prev_state {                                                                                                                                                                                                                         
    |                           ^^^^^^^^^^ value used here after move                                                                                                                                                                                                
    |                                                                                                                                                                                                                                                                
    = note: move occurs because `prev_state` has type `connection::State`, which does not implement the `Copy` trait
```

I've explored three possibilities:

1) Deriving clone for `State`.  I don't think this will work because some variants of `State` use `crypto::TlsSession`, which uses `rustls::ClientSession` and friends, which don't implement clone.
2) Giving `handle_connected_inner` a `&State` instead of `State`.  I don't think this works because `state` itself is moved into other things inside of `handle_connected_inner`, so a reference won't do...at least not in a simple way.  We'd need to add a bunch of lifetimes, which just doesn't make sense for the simple thing we're trying to do here.
3) Put `prev_state` into an `Rc`.  Once again, this seems like overkill for this simple thing we're trying to do.

I'm sure there is a simple solution to this problem - maybe just restructuring/ordering things a bit?  Does something jump out to you?

Ralith:
> I'm sure there is a simple solution to this problem

Indeed there is. Before passing `prev_state` to `handle_connected_inner`, inspect it:
`let was_handshake = match prev_state { State::Handshake(_) => true, _ => false };`
This doesn't require consuming `prev_state`.

twilco:
Wow, I way overthought that :tired_face:.  Thanks.

In attempting to squash non-consecutive commits, I managed to botch everything.  Closing this out - please give a final review in the new PR!  #68

Issue #35, opened 2018-10-03
Discuss how TLS `ClientConfig` should be passed around


djc:
@Ralith made the argument that this should be an argument to `connect()` instead of a global configuration setting in `Config`.

I'm actually not quite convinced. We currently change the following things:

* Toggle key log
* Trust anchors
* TLS protocol version
* ALPN protocol

I would argue that it makes sense to set these things on a per-`Endpoint` basis.

Ralith:
A client may want to communicate with different types of servers via different protocols without allocating redundant endpoints, so per-connection ALPN is necessary as a baseline. On review I don't actually see anything else in `ClientConfig` that seems obviously useful (and rustls is mistakenly using `String` for ALPN IDs there too) so maybe it would be better to define our own API just for that.

djc:
And abstract over cloning the underlying `ClientConfig` and mutating the clone? Maybe we should open an issue in rustls to see what they think about changing the API around this? (Something like `ClientSession::with_alpn_protocols(..)` that overrides the `Config`s defaults.)

Ralith:
Another per-connection variable could be certificate verification overrides, which we'll want sooner or later (not just for untrusted connections, but also for unconventional verification schemes like Perspectives).

> Maybe we should open an issue in rustls to see what they think about changing the API around this?

Is there that much room for improvement over duplicate config objects? They're not *that* big, are they?

djc:
> Is there that much room for improvement over duplicate config objects? They're not that big, are they?

The documentation says:

> Making one of these can be expensive, and should be once per process rather than once per connection.

It's not quite clear whether that means cloning would also be affected.

I really don't think we should be optimizing for per-connection certificate verification overrides until a concrete use case comes up.

djc:
I clarified the method name; I think adding CAs (rather than setting one) would mostly cover that?

Ralith:
For P2P, TOFU, or sidechannel-based verification systems, a CA wouldn't accomplish anything. Sometimes you just want to talk to someone that hasn't been vouched for in advance by a central authority.

djc:
CA in this context doesn't mean something in Mozilla's root store, but just the certificate chain root.

Ralith:
[...]

PR #57, opened 2018-10-13
Client configuration API

Ralith:
- Allow configuration of individual clients within an endpoint
- Implement feature-gated config helper for untrusted connections
- Don't enable keylogging by default
- Test rejection of invalid certs

Candidate to address #35.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -1,5 +1,6 @@
 use std::collections::{hash_map, BTreeMap, VecDeque};
 use std::net::SocketAddrV6;
+use std::sync::Arc;
 use std::{cmp, io, mem};
 
 use bytes::{Buf, Bytes, BytesMut};
@@ -8,7 +9,9 @@ use rand::distributions::Distribution;
 use slog::Logger;
 
 use coding::{BufExt, BufMutExt};
-use crypto::{reset_token_for, ConnectError, Crypto, TLSError, TlsSession, ACK_DELAY_EXPONENT};
+use crypto::{
+    reset_token_for, ClientConfig, ConnectError, Crypto, TLSError, TlsSession, ACK_DELAY_EXPONENT,
+};
 use endpoint::{Config, Context, Event, Io, Timer};
 use packet::{
     set_payload_length, types, ConnectionId, Header, Packet, PacketNumber, AEAD_TAG_SIZE,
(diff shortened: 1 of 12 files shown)


quinn-proto/src/connection.rs
```diff
@@ -323,6 +329,7 @@ impl Connection {
             data_recvd: 0,
             local_max_data: config.receive_window as u64,
             server_name: None,
+            tls_config: None,
```
djc: I don't like this very much. We're putting a `Config` state in here, optionally only for the initiators, which is only used in the very short time between the instantiation of the `Connection` and the `connect()`. Seems to me like we might want to create a `new_client()` constructor instead that just does the `connect()` bits straightaway, and then we don't have to store the `ClientConfig`.

quinn-proto/src/connection.rs
```diff
@@ -376,13 +383,22 @@ impl Connection {
...
-            server_name,
-            &TransportParameters::new(&ctx.config),
-        ).unwrap();
+    pub fn connect(
+        &mut self,
+        ctx: &Context,
+        config: &Arc<ClientConfig>,
+        server_name: &str,
+    ) -> Result<(), ConnectError> {
+        assert_eq!(
```
djc: If we're asserting this here, should we have a similar assert for the server side?

quinn-proto/src/tests.rs
```diff
@@ -208,7 +206,10 @@ impl Pair {
 
     fn connect(&mut self) -> (ConnectionHandle, ConnectionHandle) {
         info!(self.log, "connecting");
-        let client_conn = self.client.connect(self.server.addr, "localhost").unwrap();
+        let client_conn = self
```
djc: It still seems to me the repetition that you introduce here points to a worse API. Do you agree that it would be more common for applications using Quinn to have a single `ClientConfig` that can be used across connections versus specific per-connection `ClientConfig`s?

An alternative here would be to have another `connect()` method on `Endpoint` that allows passing in a `ClientConfig` explicitly. I feel that is better because it provides a streamlined, simple API for common use cases and a more complex API for those who need more power/expressivity.

quinn-proto/src/connection.rs
Ralith: In fact, rustls contains a similar assert already, so perhaps this should be removed.

quinn-proto/src/tests.rs
Ralith: I totally agree that this is a niche case, and an API that sets a default config seems reasonable. That would belong in the high-level API crate, though, not -proto, right?

quinn-proto/src/connection.rs
Ralith: The `ClientConfig` is stored for use in resetting state on retry packet receipt. That said, I'm not actually sure resetting TLS state is necessary; tests pass and draft 11 doesn't seem to specifically require it. It won't be needed in draft 15 either, so I'll just strip it out.

Ralith:
Yeah, got a bit carried away with the big blob o' intermingled changes, apologies.

quinn-proto/src/tests.rs
Ralith: Latest revision includes this in the high-level API.

Ralith:
That commit disappeared when I realized the nondeterminism was just that I'd left a test server running in the background. Disabled-by-default (perhaps mediated by an env var?) would make sense, but so long as it's easy to run them in practice I don't think it's urgent.

PR #67, opened 2018-10-15
Readme fixes

djc:
@omac777, would this have helped? ([Rendered.](https://github.com/djc/quinn/blob/readme-fixes/README.md))

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,7 @@ Quinn was created and is maintained by by Dirkjan Ochtman and Benjamin Saunders.
 
 * Simultaneous client/server operation
 * Ordered and unordered reads for improved performance
-* Uses [rustls][rustls] for all TLS operations
-* Uses [*ring*][ring] for AEAD encryption/decryption
+* Uses [rustls][rustls] for all TLS operations and [*ring*][ring] for cryptography
 
 ## Status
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,11 +26,10 @@ Quinn was created and is maintained by by Dirkjan Ochtman and Benjamin Saunders.
 
 ## Features
 
-* Currently targets QUIC draft 11, but with final TLS 1.3 per RFC 8446
```
Ralith: Shouldn't we mention this somewhere? A reasonable person might expect interop we don't currently have.

djc: It's still there a few lines lower down, under Status.

README.md
```diff
@@ -44,6 +43,17 @@ Quinn was created and is maintained by by Dirkjan Ochtman and Benjamin Saunders.
...
 
+## How to start
+
+The example client [currently always verifies][insecure] the server's certificate chain.
+Example certificates are included in the repository for test purposes.
+
+```sh
+$ cd quinn
+$ cargo run --example server -- --cert ../certs/server.chain --key ../certs/server.rsa .
+$ cargo run --example client -- --ca ../certs/ca.der https://localhost:4433
```
Ralith: The client now needs `--features="dangerous_configuration"`

djc: Hmm -- so that doesn't feel like an improvement. Should we remove the insecure option from the example client until the server can actually generate a temporary certificate?

djc:
I think this can be merged now, can you re-review?

Issue #64, opened 2018-10-14
Adopt tokio_timer::DelayQueue


Ralith:
We're currently using an awkward heavyweight hack to schedule timers; see the `quinn` crate's `EndpointInner` struct's `timer` member for details. [`DelayQueue`](https://docs.rs/tokio-timer/0.2.7/tokio_timer/struct.DelayQueue.html) was literally purpose-built for our requirements, so let's put it to work.

twilco:
I'd like to take a go at this one.  I'm busy the next couple days (will be at Rust Belt Rust!), and want to do some initial learning of Tokio as part of the [Tokio docs push](https://github.com/tokio-rs/doc-push), but after that I will have time to work on this.

If someone else comes across this issue and feels motivated to take it before then let me know and we can work something out 👍 .

twilco:
Things have come up and I think I'm going to be busy for a bit, so if there's anyone else who wants to take this go for it.

little-dude:
I'd like to work on that.

Ralith:
Great! Feel free to hit up the gitter if you have any questions.

little-dude:
So far so good, hopefully I'll submit a PR tonight or tomorrow :)

Ralith:
Mooted by #264

PR #68, opened 2018-10-16
Reduce repetitive error handling (take two)

### LINKED ISSUE #51: Factor out common error handling code

twilco:
@djc @Ralith 

I managed to botch the rebase on my previous branch, so starting anew here.

Prior to this patch, we had duplicated lots of error handling logic inside `handle_connected_inner`, mostly in the way of event recording. This patch simplifies this by instead returning a `Result<State, ConnectionError>`, which allows the caller to handle either case and create events more succinctly than `handle_connected_inner` could.

Also, this changeset moves the `app_closed` property from `connection::State` to `connection::Connection` itself, as it is much simpler to manage that piece of state on the `Connection` itself.

Closes #51.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -586,14 +586,61 @@ impl Endpoint {
         let was_closed = self.connections[conn.0].state.as_ref().unwrap().is_closed();
 
         // State transitions
-        let state = self.connections[conn.0].state.take().unwrap();
-        let state = self.connections[conn.0].handle_connected_inner(
+        let prev_state = self.connections[conn.0].state.take().unwrap();
+        let was_handshake = match prev_state {
+            State::Handshake(_) => true,
+            _ => false,
+        };
+        let state = match self.connections[conn.0].handle_connected_inner(
             &mut self.ctx,
             now,
             remote,
             packet,
-            state,
-        );
+            prev_state,
+        ) {
+            Ok(state) => state,
+            Err(conn_err) => {
+                self.ctx.events.push_back((
+                    conn,
+                    Event::ConnectionLost {
+                        reason: conn_err.clone(),
+                    },
+                ));
+
+                match conn_err {
+                    ConnectionError::ApplicationClosed { reason } => {
+                        if was_handshake {
+                            State::handshake_failed(reason, None)
+                        } else {
+                            State::closed(reason)
+                        }
+                    }
+                    ConnectionError::ConnectionClosed { reason } => {
+                        if was_handshake {
+                            State::handshake_failed(reason, None)
+                        } else {
+                            State::closed(reason)
+                        }
+                    }
+                    ConnectionError::Reset => {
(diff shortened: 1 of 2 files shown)


quinn-proto/src/endpoint.rs
```diff
@@ -586,14 +586,57 @@ impl Endpoint {
...
+                            State::closed(reason)
+                        }
+                    },
+                    ConnectionError::TransportError { error_code } => {
+                        if was_handshake {
+                            State::handshake_failed(error_code, None)
+                        } else {
+                            State::closed(error_code)
+                        }
+                    },
```
twilco: @Ralith - Maybe a better way to handle these repetitive `if else`s.  I tried storing the `error_code`/`reason` in an `Option` and then doing a single `if else` below, but that didn't work for me.

If you have a good idea I'd love to hear it - otherwise I think this PR should be good to go.

quinn-proto/src/endpoint.rs
```diff
@@ -586,14 +586,57 @@ impl Endpoint {
...
+                        if was_handshake {
+                            State::handshake_failed(error_code, None)
+                        } else {
+                            State::closed(error_code)
+                        }
+                    }
+                    ConnectionError::VersionMismatch => State::Draining,
+                    _ => {
+                        debug!(self.ctx.log, "unexpected connection error received"; "error" => %conn_err, "initial_conn_id" => %self.connections[conn.0].initial_id);
+                        return;
```
Ralith: This leaves the connection's state empty, which is assumed e.g. on linel 589 to be impossible and will likely lead to a panic down the line. We should probably either panic immediately, or invent an appropriate state for the remaining cases (reset and timeout, so draining, perhaps?). I don't have reason to believe we'll emit either of those from `handle_connection_inner` in the future, so I'd lean towards panic. Probably best to explicitly enumerate the remaining cases either way, since any new additions might need different handling.

twilco: Yeah, good call.

twilco:
@Ralith - Enumerated all `ConnectionError` types.  Please re-review!

djc:
+115/-204, very nice! Thanks for sticking with it! If you want to do more, we have a bunch more issues to chose from. 👍

twilco:
Thank you both once again :).  Sorry it was a bit of a slog.  Learning more on each one!

I have my eyes on #64, but have plans to do some reading up on Tokio first as part of the [Tokio docs push](https://github.com/tokio-rs/doc-push).  I'll leave a comment on the issue itself.

PR #70, opened 2018-10-16
WIP: Implement packet number encryption/decryption

djc:
See also https://github.com/RustCrypto/stream-ciphers/issues/6.

### DIFF
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -19,6 +19,7 @@ maintenance = { status = "experimental" }
 dangerous_configuration = ["rustls/dangerous_configuration"]
 
 [dependencies]
+aes-ctr = "0.2"
 blake2 = "0.8"
 byteorder = "1.1"
 bytes = "0.4.7"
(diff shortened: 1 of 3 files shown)


djc:
@newpavlov the second commit fixes the key length, but it still fails. I also don't see how inferring the length could work in this case, so I find the whole API rather opaque.

newpavlov:
`GenericArray` is an array, i.e. it's allocated on stack, so it can't be both `[u8; 16]` and `[u8; 32]` at the same time. Your code could be written like this:
```rust
    fn decrypt(&self, key: &[u8], sample: &[u8], in_out: &mut [u8]) {
        use self::PacketNumberEncryptionAlgorithm::*
        match *self {
            AesCtr128 => {
                let key = GenericArray::from_slice(key);
                let nonce = GenericArray::from_slice(sample);
                Aes128Ctr::new(key, nonce).apply_keystream(in_out);
            },
            AesCtr256 => {
                let key = GenericArray::from_slice(key);
                let nonce = GenericArray::from_slice(sample);
                Aes256Ctr::new(key, nonce).apply_keystream(in_out);
            },
        }
    }
```
I agree it's not elegant, so I will add a new method `new_var(key: &[u8], nonce: &[u8]) -> Result<Self, InvalidKeyNonceLength>` to the `NewFixStreamCipher` trait.

BTW are you fine with panics if key or nonce will have a wrong length, or is it validated somewhere else?

djc:
I think an `Err` `Result` would probably be better than a panic for this.

newpavlov:
I've published the update. Try to update to `stream-cipher v0.2.2` and change your code to `new_var` method.

PR #71, opened 2018-10-17
WIP: Draft 15

djc:


quinn-proto/src/packet.rs
```diff
@@ -7,21 +7,50 @@ use slog;
...
-        ty: u8,
+        ty: LongType,
         source_id: ConnectionId,
         destination_id: ConnectionId,
         number: u32,
     },
+    Retry {
+        source_id: ConnectionId,
+        destination_id: ConnectionId,
+        orig_dst_cid: ConnectionId,
```
Ralith: Need retry token

quinn-proto/src/packet.rs
```diff
@@ -186,6 +364,8 @@ pub enum HeaderError {
     },
     #[fail(display = "invalid header: {}", _0)]
     InvalidHeader(&'static str),
+    #[fail(display = "invalid long header type: {}", _0)]
```
Ralith: `{:02x}` for consistency with the spec

quinn-proto/src/connection.rs
```diff
@@ -165,7 +165,7 @@ pub struct SentPacket {
     pub time: u64,
     /// 0 iff ack-only
     pub bytes: u16,
-    pub handshake: bool,
+    pub initial: bool,
```
Ralith: This flag is used to implement the draft-11 handshake retransmit strategy, which is obsolete. I'm not sure renaming it makes sense; it doesn't preserve the original meaning (which included `Handshake` packets) and I'm not sure we need anything like it with the new handshake structure.

quinn-proto/src/packet.rs
djc: Yeah, I was going to put that in the payload.

quinn-proto/src/packet.rs
```diff
@@ -7,21 +7,50 @@ use slog;
...
+                header_data,
+                payload,
+            },
+            packet,
+        ))
+    }
+}
+
 #[derive(Debug, Clone)]
 pub enum Header {
```
Ralith: Should we remove `Long` and add `Initial` and `Handshake`?

djc: That might make sense, I'll keep it in mind.

quinn-proto/src/connection.rs
djc: I'll just keep this as `handshake` for now, then.

djc:
So packet number encryption is now implemented and seems to work with the test vector, however Wireshark doesn't seem to like the Initial packet. @Ralith could you do some review and see if I got something wrong?

djc:
Also, how do you feel about merging this in the current state? Our current protocol is already frankensteinian, and it would be nice not to have to worry about rebases when other stuff happens.

quinn-proto/src/crypto.rs
```diff
@@ -453,7 +454,13 @@ impl PacketNumberKey {
                 let nonce = GenericArray::from_slice(sample);
                 Aes128Ctr::new(key, nonce).apply_keystream(in_out)
             }
-            _ => unimplemented!(),
+            ChaCha20(key) => {
+                let counter = io::Cursor::new(&sample[..4]).get_u32_be();
```
Ralith: `byteorder::BE::read_u32(&sample[..4])`

quinn-proto/src/crypto.rs
```diff
@@ -465,7 +472,13 @@ impl PacketNumberKey {
...
```
Ralith: As above

quinn-proto/src/packet.rs
```diff
@@ -52,19 +54,51 @@ impl PartialDecode {
...
             invariant_header,
             mut buf,
         } = self;
         let (payload_len, header, allow_coalesced) = match invariant_header {
             InvariantHeader::Short { first, dst_cid } => {
                 let key_phase = first & KEY_PHASE_BIT != 0;
-                let number = PacketNumber::decode(&mut buf)?;
+                if buf.remaining() < 1 {
+                    return Err(PacketDecodeError::InvalidHeader(
+                        "cannot read packet number",
```
Ralith: "can't read" suggests an unspecified internal error to me; perhaps rephrase as "missing"?

quinn-proto/src/packet.rs
```diff
@@ -169,6 +212,35 @@ impl PartialDecode {
...
+        buf: &mut io::Cursor<BytesMut>,
+        pn_key: &PacketNumberKey,
+        sample_offset: usize,
+    ) -> Result<PacketNumber, PacketDecodeError> {
+        let mut first = [buf.bytes()[0]; 1];
+        let sample = {
+            let mut sample = [0; 16];
+            debug_assert!(pn_key.sample_size() <= 16);
+            sample.copy_from_slice(
+                &buf.get_ref()[sample_offset..sample_offset + pn_key.sample_size()],
```
Ralith: How do we know this isn't out of bounds?

quinn-proto/src/packet.rs
```diff
@@ -225,6 +297,11 @@ impl Header {
                 w.put_slice(token);
                 w.write::<u16>(0); // Placeholder for payload length; see `set_payload_length`
                 number.encode(w);
+                let pn_pos = 9 + dst_cid.len() + src_cid.len() + token.len();
```
Ralith: Does this handle the variable-length encoding of `token.len()`?

quinn-proto/src/packet.rs
```diff
@@ -285,6 +379,64 @@ impl Header {
...
+            }
+            _ => {
+                return;
+            }
+        };
+
+        let sample = {
+            let mut sample = [0; 16];
+            debug_assert!(pn_key.sample_size() <= 16);
+            sample.copy_from_slice(&buf[sample_offset..sample_offset + pn_key.sample_size()]);
```
Ralith: How do we know this isn't out of bounds?

quinn-proto/src/connection.rs
```diff
@@ -1924,15 +1908,19 @@ impl Connection {
                             state.clienthello_packet = Some(number as u32);
                         }
                     }
-                    LongType::Initial
+                    Header::Initial {
+                        src_cid: self.loc_cid,
+                        dst_cid: self.rem_cid,
+                        token: vec![],
```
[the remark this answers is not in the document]
djc: Will do.

quinn-proto/src/packet.rs
```diff
@@ -359,39 +433,63 @@ impl PacketNumber {
...
-            Handshake => 0x7d,
-            ZeroRtt => 0x7c,
+impl From<PacketType> for u8 {
+    fn from(ty: PacketType) -> u8 {
+        use self::{LongType::*, PacketType::*};
+        match ty {
+            Initial => 0xff,
+            Long(Handshake) => 0xfd,
+            Long(ZeroRtt) => 0xfc,
+            Retry => 0xfe,
```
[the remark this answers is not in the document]
djc: Yup, I forgot about fixing that up.

quinn-proto/src/endpoint.rs
```diff
@@ -83,6 +83,9 @@ pub struct Config {
     pub loss_reduction_factor: u16,
 
     pub tls_server_config: Arc<ServerConfig>,
+
+    /// Length of connection IDs for the endpoint
+    pub local_cid_len: usize,
```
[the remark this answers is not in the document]
djc: I'm not so much interested in the CPU performance overhead, but I was thinking the bandwidth overhead might be nice to elide where possible (for example, in many clients). I've added a comment to that effect and the assertion as you suggested.

I think avoiding the maintenance of `self.connection_ids` will clutter up the code and may not pay for itself in performance with any added branches, so I'd prefer to keep it. However, we can check this as a special case in the `new_cid()` method so that it can raise an error.

quinn-proto/src/packet.rs
```diff
@@ -251,8 +251,8 @@ impl InvariantHeader {
                     "destination connection ID longer than packet",
                 ));
             }
-            buf.copy_to_slice(&mut cid_stage[..local_cid_len]);
-            let dst_cid = ConnectionId::new(&cid_stage[..local_cid_len]);
+            let dst_cid = ConnectionId::new(&buf.bytes()[..local_cid_len]);
+            buf.advance(local_cid_len);
```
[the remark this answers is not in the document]
djc: You're right, of course (and I fully admit it's probably a silly micro-optimization).

(On the plus side, I factored it out into a nice associated function.)

quinn-proto/src/endpoint.rs
```diff
@@ -156,6 +156,10 @@ impl Context {
     fn gen_initial_packet_num(&mut self) -> u32 {
         self.initial_packet_number.sample(&mut self.rng) as u32
     }
+
+    fn new_cid(&mut self) -> ConnectionId {
+        ConnectionId::random(&mut self.rng, self.config.local_cid_len)
```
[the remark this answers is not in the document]
djc: Agreed. The good part of having this as a separate function is that we can actually centralize these kinds of checks.

quinn-proto/src/endpoint.rs
```diff
@@ -409,9 +404,10 @@ impl Endpoint {
                 ).saturating_sub(RESET_TOKEN_SIZE),
             );
             buf.reserve_exact(header_len + padding + RESET_TOKEN_SIZE);
+            let number = self.ctx.rng.gen::<u32>() & 0x3fffffff | 0x4000;
```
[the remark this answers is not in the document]
djc: I've named the former one. I'm not sure the latter one is needed, what do you think? Is it a requirement that packet numbers are encoded in their shortest possible encoding?

quinn-proto/src/endpoint.rs
```diff
@@ -153,10 +152,6 @@ pub struct Context {
 }
 
 impl Context {
-    fn gen_initial_packet_num(&mut self) -> u32 {
-        self.initial_packet_number.sample(&mut self.rng) as u32
```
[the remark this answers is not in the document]
djc: Done.

quinn-proto/src/connection.rs
```diff
@@ -1203,13 +1201,14 @@ impl Connection {
                     ctx.incoming_handshakes -= 1;
                 }
                 let n = self.get_tx_number();
+                debug_assert!(n < 64); // should fit in one byte
```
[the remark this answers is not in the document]
djc: `handshake_close()` doesn't have access to `largest_acked_packet`, so it cannot use `PacketNumber::new()`. Since `handshake_close()` is all about the handshake, it seemed reasonable that it can simply default to the single-byte encoding of packet numbers instead. I'll clarify the comment here.

quinn-proto/src/crypto.rs
djc: Hmm, yeah. Can't wait for int_to_from_bytes.

quinn-proto/src/packet.rs
djc: I've changed it to `header ends before packet number`.

quinn-proto/src/packet.rs
djc: I had actually already fixed that, but hadn't pushed it yet.

quinn-proto/src/endpoint.rs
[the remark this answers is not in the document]
Ralith: The bandwidth overhead is what I'm referring to, yeah. The important thing is that we don't try to look up 0-length CIDs in `handle_decode`.

quinn-proto/src/connection.rs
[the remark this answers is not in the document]
Ralith: > handshake_close() doesn't have access to largest_acked_packet, so it cannot use PacketNumber::new().

Perhaps it should take a `PacketNumber`?

> Since handshake_close() is all about the handshake, it seemed reasonable that it can simply default to the single-byte encoding of packet numbers instead.

I'm not sure this is reliable. Handshakes can potentially be large due to e.g. large certificate chains, server transport parameters, or future extensions.

quinn-proto/src/endpoint.rs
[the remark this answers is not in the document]
djc: I don't think the lookup is a problem as long as we restrict the endpoint to only ever giving out the zero-length CID once, which I think my updated patch covers. Take a look.

quinn-proto/src/connection.rs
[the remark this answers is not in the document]
djc: > Perhaps it should take a `PacketNumber`?

That's what I started to do, but then it also needs the raw packet number (as it also doesn't have the state to expand the packet number again) to be able to encrypt the payload.

As I understand it, a normal handshake completes in substantially less than packets per peer, so it seems to me that 6x that ought to provide a safe margin.

quinn-proto/src/packet.rs
djc: I've added a bounds check.

quinn-proto/src/packet.rs
djc: Added check.

djc:
> Lots of great work! I did notice a number of possible issues. This pass was focused on the changes to existing code, cosmetics, and suspicious fragments; I haven't checked the PNE impl against the spec or investigated with wireshark yet.

Thanks, lots of good suggestions. I do still hope you can push a little further on the Wireshark/PNE problem, to make sure my reading of the spec isn't crazy.

Also note I pulled in two different crypto libraries for AES-CTR and ChaCha20, respectively. I hope to improve that soon by using the implementations in *ring* that are going [through review now](https://github.com/briansmith/ring/issues/693).

> Review issues notwithstanding, I'm fine with merging so long as we reasonably expect the result not to be a significant regression in ability to communicate with itself. Frankentsteinitude aside, I'd rather our manpower be spent on keeping up with the IETF than on endless rebases, at least until we hit 1.0.

Yeah, that was my thinking as well. I've kept the test suite passing at every commit in this series, hopefully that is a good approximation of being able to communicate with itself. In this way, we can iterate our way to draft spec compliance without chilling effects on other refactoring steps.

quinn-proto/src/endpoint.rs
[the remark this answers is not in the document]
Ralith: Doesn't that restrict the endpoint to having at most one concurrent connection? §5.1 suggests that we should use the remote address+port to identify connections when zero-length CIDs are in use.

quinn-proto/src/endpoint.rs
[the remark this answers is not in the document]
Ralith: On review, [the spec](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#rfc.section.10.4) now calls for the first byte of the packet to be followed by "at least 20 bytes of random or unpredictable values" (including the random padding bytes such as we already generate) rather than a well-formed header. That seems like the simpler way to proceed.

quinn-proto/src/connection.rs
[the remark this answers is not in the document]
Ralith: Is there a firm guarantee anywhere? I'd rather not generate invalid close packets in weird-but-legal situations.

Ralith:
> I do still hope you can push a little further on the Wireshark/PNE problem, to make sure my reading of the spec isn't crazy.

Yeah, I'll still be going over that in more detail. I do reproduce the decode failure in wireshark.

quinn-proto/src/packet.rs
```diff
@@ -219,6 +382,65 @@ impl Header {
...
+                let (pn_pos, pn_len) = pn.unwrap();
+                (sample_offset, pn_pos, pn_len)
+            }
+            Header::Long {
+                dst_cid, src_cid, ..
+            } => {
+                let sample_offset =
+                    10 + dst_cid.len() + src_cid.len() + varint::size(payload_len).unwrap();
+                let (pn_pos, pn_len) = pn.unwrap();
+                (sample_offset, pn_pos, pn_len)
```
Ralith: Probable missing bounds check

quinn-proto/src/packet.rs
```diff
@@ -85,34 +111,71 @@ impl PartialDecode {
...
+                            token,
+                            number,
+                        },
+                        true,
+                    )
+                }
+                PacketType::Long(ty) => {
+                    let len = buf.get_var()?;
+                    let sample_offset =
+                        10 + dst_cid.len() + src_cid.len() + varint::size(len).unwrap();
```
Ralith: Probable missing bounds check

quinn-proto/src/packet.rs
```diff
@@ -85,34 +111,71 @@ impl PartialDecode {
...
-                        return Err(PacketDecodeError::InvalidHeader(
-                            "payload longer than packet",
-                        ));
-                    }
+                    let sample_offset = 10
+                        + dst_cid.len()
+                        + src_cid.len()
+                        + varint::size(len).unwrap()
+                        + varint::size(token_length as u64).unwrap()
+                        + token.len();
```
Ralith: Probable missing bounds check. Perhaps the check/adjustment should be done inside `get_packet_number`?

quinn-proto/src/packet.rs
```diff
@@ -219,6 +382,65 @@ impl Header {
...
+                ..
+            } => {
+                let sample_offset = 10
+                    + dst_cid.len()
+                    + src_cid.len()
+                    + varint::size(payload_len).unwrap()
+                    + varint::size(token.len() as u64).unwrap()
+                    + token.len();
+                let (pn_pos, pn_len) = pn.unwrap();
+                (sample_offset, pn_pos, pn_len)
```
Ralith: Probable missing bounds check. Perhaps the check/adjustment should be moved after the match?

quinn-proto/src/packet.rs
```diff
@@ -138,6 +212,34 @@ impl PartialDecode {
             if allow_coalesced { Some(bytes) } else { None },
         ))
     }
+
+    fn get_packet_number(
+        buf: &mut io::Cursor<BytesMut>,
+        pn_key: &PacketNumberKey,
+        sample_offset: usize,
+    ) -> Result<PacketNumber, PacketDecodeError> {
+        if buf.get_ref().len() < sample_offset + pn_key.sample_size() {
```
djc: Did you miss this bounds check (which I added in my last revision)? This seems to cover all the cases you comment on.

quinn-proto/src/packet.rs
djc: I think this is covered by the `debug_assert!()`s, below, unless you think they should be non-debug asserts? This is one of those cases where things can only go wrong if we've got bugs in our own code, so I didn't think we'd need to do much more.

quinn-proto/src/packet.rs
djc: Which check/adjustment? The one for the short Short packets? That can really only apply to Short packets, so I don't think it generalizes.

Ralith: Why is it only applicable to short packets? It seems like long packets have exactly the same problem, with the same solution: "If the sample would extend past the end of the packet, the end of the packet is sampled." Do we have a hard lower bound of long packet payload size somewhere?

quinn-proto/src/packet.rs
Ralith: I did miss this, because I was looking for something that adjusts the sample offset and continues rather than bailing out, per the TLS draft. The current state of affairs appears to assume that valid long packets will always have payloads extending at least 20 bytes from the start of the packet number, and it's not clear to me that that holds.

quinn-proto/src/packet.rs
djc: Ah, I think I misunderstood the example code in 5.4. The example doesn't show the adjustment for the long header calculation, but from the text it becomes clear that it is still needed. That'll probably solve the Wireshark issues, then...

quinn-proto/src/packet.rs
djc: Have another look, I think it'll make more sense now.

quinn-proto/src/packet.rs
djc: Hmm, Wireshark appears to be no less confused than before. 😢

quinn-proto/src/endpoint.rs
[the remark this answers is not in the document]
djc: Yup, you're right. I'll tweak this more.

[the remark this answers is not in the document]
djc: Should be better now. Added this:

```diff
diff --git a/quinn-proto/src/endpoint.rs b/quinn-proto/src/endpoint.rs
index bee4054..d85842a 100644
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -333,12 +333,16 @@ impl Endpoint {
         //

         let dst_cid = partial_decode.dst_cid();
-        let conn = self
-            .connection_ids
-            .get(&dst_cid)
-            .or_else(|| self.connection_ids_initial.get(&dst_cid))
-            .or_else(|| self.connection_remotes.get(&remote))
-            .cloned();
+        let conn = {
+            let conn = if self.ctx.config.local_cid_len > 0 {
+                self.connection_ids.get(&dst_cid)
+            } else {
+                None
+            };
+            conn.or_else(|| self.connection_ids_initial.get(&dst_cid))
+                .or_else(|| self.connection_remotes.get(&remote))
+                .cloned()
+        };
         if let Some(conn) = conn {
             return self.connections[conn.0].handle_decode(
                 &mut self.ctx,
@@ -489,7 +493,9 @@ impl Endpoint {
             ));
             conn
         };
-        self.connection_ids.insert(local_id, conn);
+        if self.ctx.config.local_cid_len > 0 {
+            self.connection_ids.insert(local_id, conn);
+        }
         self.connection_remotes.insert(remote, conn);
         conn
     }
@@ -617,8 +623,10 @@ impl Endpoint {
             self.connection_ids_initial
                 .remove(&self.connections[conn.0].init_cid);
         }
-        self.connection_ids
-            .remove(&self.connections[conn.0].loc_cid);
+        if self.ctx.config.local_cid_len > 0 {
+            self.connection_ids
+                .remove(&self.connections[conn.0].loc_cid);
+        }
         self.connection_remotes
             .remove(&self.connections[conn.0].remote);
         self.ctx.dirty_conns.remove(&conn);
```

quinn-proto/src/packet.rs
```diff
@@ -52,19 +54,46 @@ impl PartialDecode {
...
+    ) -> Result<(Packet, Option<BytesMut>), PacketDecodeError> {
         let Self {
             invariant_header,
             mut buf,
         } = self;
         let (payload_len, header, allow_coalesced) = match invariant_header {
             InvariantHeader::Short { first, dst_cid } => {
                 let key_phase = first & KEY_PHASE_BIT != 0;
-                let number = PacketNumber::decode(&mut buf)?;
+                if buf.remaining() < 1 {
```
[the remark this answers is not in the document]
djc: Done.

quinn-proto/src/packet.rs
```diff
@@ -169,6 +207,33 @@ impl PartialDecode {
...
     }
+
+    fn get_packet_number(
+        buf: &mut io::Cursor<BytesMut>,
+        pn_key: &PacketNumberKey,
+        mut sample_offset: usize,
+    ) -> Result<PacketNumber, PacketDecodeError> {
+        let packet_length = buf.get_ref().len();
+        if sample_offset + pn_key.sample_size() > packet_length {
+            sample_offset = packet_length - pn_key.sample_size();
```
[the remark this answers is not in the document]
djc: I think this ought to be enough to stop them:

```diff
diff --git a/quinn-proto/src/packet.rs b/quinn-proto/src/packet.rs
index 05c9d35..2f4150c 100644
--- a/quinn-proto/src/packet.rs
+++ b/quinn-proto/src/packet.rs
@@ -85,7 +85,7 @@ impl PartialDecode {
         let (payload_len, header, allow_coalesced) = match invariant_header {
             InvariantHeader::Short { first, dst_cid } => {
                 let key_phase = first & KEY_PHASE_BIT != 0;
-                if buf.remaining() < 1 {
+                if !buf.has_remaining() {
                     return Err(PacketDecodeError::InvalidHeader(
                         "header ends before packet number",
                     ));
@@ -214,9 +214,20 @@ impl PartialDecode {
         mut sample_offset: usize,
     ) -> Result<PacketNumber, PacketDecodeError> {
         let packet_length = buf.get_ref().len();
+        if packet_length < pn_key.sample_size() {
+            return Err(PacketDecodeError::InvalidHeader(
+                "packet too short to decode packet number",
+            ));
+        }
+
         if sample_offset + pn_key.sample_size() > packet_length {
             sample_offset = packet_length - pn_key.sample_size();
         }
+        if packet_length < sample_offset + pn_key.sample_size() {
+            return Err(PacketDecodeError::InvalidHeader(
+                "packet too short to extract packet number encryption sample",
+            ));
+        }

         let mut first = [buf.bytes()[0]; 1];
         let sample = {
```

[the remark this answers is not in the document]
Ralith: nit: first check could be combined with the subtraction using `packet_length.checked_sub(pn_key.sample_size()).ok_or(PacketDecodeError::...)?`

djc:
Addressed the nit, this is indeed better. Can you approve again? 👍

PR #74, opened 2018-10-18
Implement draft 15 transport parameters

Ralith:
This implements all the new transport parameters and factors out a lot of the error-prone duplication with macros.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -2037,11 +2037,11 @@ impl Connection {
     }
 
     pub fn set_params(&mut self, params: TransportParameters) {
-        self.max_bi_streams = params.initial_max_streams_bidi as u64;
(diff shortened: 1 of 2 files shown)


djc:
The macro business is a little tricky to read/understand. Could you mind adding at least a few comments that explain how things fit together and what problems this is trying to solve?

Another idea that I had (somewhat orthogonal to this PR): can we embed an instance of `TransportParameters` in both `Config` and `Connection`, rather than duplicating a lot of the fields in those types? I think that would make it more obvious what's going on, and it's also another way of reducing repetitive code. (We might chose to remove the `initial_` prefix from some of the field names.)

Ralith:
We do store a `TransportParameters` in `Connection`--that of the peer, specifically. Not all of the parameters make sense to have in a `Config`; for example the server-only block is different based on the client it's being delivered to.

PR #78, opened 2018-10-22
Move TLS out of state enum

Ralith:
Simplifies many signatures.

A lot of things are still a bit awkward about construction, but I think this is a step in the right direction.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -9,9 +9,7 @@ use rand::distributions::Distribution;
 use slog::Logger;
 
 use coding::{BufExt, BufMutExt};
-use crypto::{
-    reset_token_for, ClientConfig, ConnectError, Crypto, TLSError, TlsSession, ACK_DELAY_EXPONENT,
-};
+use crypto::{self, reset_token_for, Crypto, TLSError, TlsSession, ACK_DELAY_EXPONENT};
 use endpoint::{Config, Context, Event, Io, Timer};
 use packet::{
     set_payload_length, types, ConnectionId, Header, Packet, PacketNumber, AEAD_TAG_SIZE,
@@ -34,6 +32,7 @@ impl From<ConnectionHandle> for usize {
 }
 
 pub struct Connection {
+    pub tls: TlsSession,
     pub app_closed: bool,
     /// DCID of Initial packet
     pub initial_id: ConnectionId,
@@ -62,8 +61,7 @@ pub struct Connection {
     pub data_recvd: u64,
     /// Limit on incoming data
     pub local_max_data: u64,
-    /// Server name (for client-side)
-    pub server_name: Option<String>,
+    client_config: Option<ClientConfig>,
 
     //
     // Loss Detection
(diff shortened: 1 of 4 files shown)


quinn-proto/src/connection.rs
```diff
@@ -301,10 +319,38 @@ impl Connection {
         {
             streams.insert(
                 StreamId::new(!side, Directionality::Bi, i as u64),
-                Stream::new_bi(config.stream_receive_window as u64),
+                Stream::new_bi(ctx.config.stream_receive_window as u64),
             );
         }
-        Self {
+        let (tls, tls_client_config) = match conn_config {
```
djc: Hmm, it feels like this chunk of code is more tightly coupled to `Endpoint` than `Connection`, perhaps?

quinn-proto/src/connection.rs
```diff
@@ -267,30 +267,48 @@ impl ::std::iter::FromIterator<Retransmits> for Retransmits {
     }
 }
 
+pub enum ConnectionConfig<'a> {
```
djc: I would prefer to add this code after `Connection` (per the idea that the highest-level API should live on top, and thus be more accessible). What are your thoughts on that?

quinn-proto/src/connection.rs
Ralith: How do you mean? The TLS session is certainly part of the connection.

quinn-proto/src/connection.rs
Ralith: Makes sense to me.

quinn-proto/src/connection.rs
djc: I mean, you have a 40-line chunk of code here that transforms a `ConnectionConfig` into a `TlsSession` + `Option<ClientConfig>`. Shouldn't we do that transformation in `Endpoint::add_connection()` instead of inside `Connection`? In that case I think `Connection::new()` wouldn't need access to `ctx` instead of `config`.

And would it make more sense to save the `server_name` in the `ClientConfig`, instead of wrapping another type around it?

Ralith: > Shouldn't we do that transformation in Endpoint::add_connection() instead of inside Connection? In that case I think Connection::new() wouldn't need access to ctx instead of config.

My thinking here was that future work could get rid of `connect` and `handle_initial` entirely, though it's not certain whether that would work out well. I'll try moving things out into `Endpoint` tomorrow and see how that looks.

> And would it make more sense to save the server_name in the ClientConfig, instead of wrapping another type around it?

The `ClientConfig` here is `rustls`'s; `-proto` doesn't have its own `ClientConfig` at the moment, though I've been thinking it could make sense for one to exist eventually; various parameters taken from the endpoint configuration don't actually need to be global across outgoing connections.

djc: Oh, yes, I forgot that this was rustls's `ClientConfig`. I'll be curious to see your `Endpoint` variant; I do agree with working towards the elimination of `connect()` and `handle_initial()`, but I'm not sure if that invalidates the point that the `Endpoint` should take care of this bit. Anyway, let's see how it turns out.

PR #80, opened 2018-10-30
Manage InnerEndpoint timers with a DelayQueue

little-dude:


little-dude:
This is my attempt at fixing https://github.com/djc/quinn/issues/64

There are also two unrelated clippy fixes but I can remove them from the PR if you prefer.

little-dude:
I hope I'm not changing the behavior here. It would be helpful to have function tests for this kind of things. I have some experience in python and QA, so if that fits your plan, I can probably automate testing a bit. I'd have to read the RFCs first but let me know if you're interested.

Ralith:
Also, it looks like you're running a mismatched version of rustfmt. We're standardized on the current stable release.

little-dude:
> Also, it looks like you're running a mismatched version of rustfmt. We're standardized on the current stable release.

Ah thank you, I was a bit confused. I amended the last commit.

I'll address the rest of your comment tomorrow morning. Thank you for the thorough review, it's much appreciated :)

djc:
Actually, #78 already fixes the formatting problems, so if we merge that and this PR gets rebased on top of that we don't need the extraneous commit anymore.

quinn/src/lib.rs
```diff
@@ -822,41 +807,50 @@ impl Future for Driver {
...
-                        let (send, recv) = oneshot::channel();
-                        *cancel = Some(send);
+                        match timer {
+                            Timer::LossDetection => {
+                                if let Some(mut key) = pending.cancel_loss_detect.as_mut() {
+                                    // NOTE: this panics if the key does not exist
+                                    let _ = endpoint.timers.remove(key);
+                                    *key = endpoint
+                                        .timers
+                                        .insert_at((connection, Timer::LossDetection), instant);
```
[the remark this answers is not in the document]
little-dude: It could, but this logic was wrong. With the fixed version it may not make as much sense:

```rust
                                if let Some(mut key) = pending.cancel_loss_detect.as_ref() {
                                    endpoint.timers.remove(key);
                                }
                                let key = endpoint
                                    .timers
                                    .insert_at((connection, Timer::LossDetection), instant);
                                pending.cancel_loss_detect = Some(key)
```

vs 

```rust
                                if let Some(mut key) = pending.cancel_loss_detect.as_ref() {
                                    endpoint.timers.reset_at(key, instant);
                                } else {
                                    let key = endpoint
                                        .timers
                                        .insert_at((connection, Timer::LossDetection), instant);
                                    pending.cancel_loss_detect = Some(key)
                                }
```

little-dude:
I removed the rustfmt commit, and addressed the comments (except the `reset_at` one).
I'll wait for #78 to be merged and rebase on top of it (and I'll probably squash the last two commits).

quinn/src/lib.rs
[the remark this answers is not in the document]
Ralith: I'm surprised to discover that there's no way to copy a `delay_queue::Key`. I'll work with upstream to get a `Clone` impl added at minimum, but in the meantime this is fine.

quinn/src/lib.rs
```diff
@@ -161,8 +159,8 @@ struct Pending {
     connecting: Option<oneshot::Sender<Option<ConnectionError>>>,
     uni_opening: VecDeque<oneshot::Sender<Result<StreamId, ConnectionError>>>,
     bi_opening: VecDeque<oneshot::Sender<Result<StreamId, ConnectionError>>>,
-    cancel_loss_detect: Option<oneshot::Sender<()>>,
-    cancel_idle: Option<oneshot::Sender<()>>,
+    cancel_loss_detect: Option<TimerKey>,
+    cancel_idle: Option<TimerKey>,
```
Ralith: These names should probably be updated to reflect that they're no longer cancellation handles, e.g. to `loss_detect_timer` and `idle_timer`.

quinn/src/lib.rs
```diff
@@ -879,16 +873,34 @@ impl Future for Driver {
...
+                    // Ready(None) is returned if there's not active timer, and NotReady is
+                    // returned if there are active timers but none expired yet.
                     Ok(Async::Ready(None)) | Ok(Async::NotReady) => {
                         break;
                     }
-                    Err(()) => unreachable!(),
+                    Err(timer_err) => {
+                        if timer_err.is_shutdown() {
+                            // This miht only happen is there's a bug in some low level tokio
+                            // components.
```
Ralith: I'm not sure this comment is strictly correct; it might be possible to produce this behavior by misusing low-level public tokio APIs as well. We should be careful when pointing fingers. Also, typo on "might".

quinn/src/lib.rs
```diff
@@ -879,16 +873,34 @@ impl Future for Driver {
...
+                        if timer_err.is_at_capacity() {
+                            // This occurs if there are too many Delay instances. But the delay
+                            // queue uses a single timer so there's not much we can do here in
+                            // terms of backpressure.
+                            panic!("{}", timer_err);
+                        }
+                        // There are only two types of error in tokio-timer, but when we upgrade,
+                        // there might be more, so let's panic here, in case we forget to update
+                        // this code,
+                        panic!("un-handled error {}", timer_err);
```
Ralith: I don't think there should be a hyphen here.

quinn/src/lib.rs
```diff
@@ -822,41 +807,50 @@ impl Future for Driver {
...
+                                    .timers
+                                    .insert_at((connection, Timer::Idle), instant);
+                                pending.cancel_idle = Some(key);
+                            }
+                            Timer::Close => {
+                                endpoint
+                                    .timers
+                                    .insert_at((connection, Timer::Close), instant);
+                            }
+                        };
```
Ralith: Could this be made more concise and less redundant by refactoring it into two separate match expressions bracketing an `endpoint.timers.insert_at` call?

quinn/src/lib.rs
```diff
@@ -879,16 +873,34 @@ impl Future for Driver {
...
+                        }
+                        if timer_err.is_at_capacity() {
+                            // This occurs if there are too many Delay instances. But the delay
+                            // queue uses a single timer so there's not much we can do here in
+                            // terms of backpressure.
+                            panic!("{}", timer_err);
+                        }
+                        // There are only two types of error in tokio-timer, but when we upgrade,
+                        // there might be more, so let's panic here, in case we forget to update
+                        // this code,
```
Ralith: Misleading comment, since errors can be introduced without a breaking change and therefore without us having upgraded anything.

quinn/src/lib.rs
little-dude: I'm not sure what you mean. I've tried this, but that does not compile due the `Timer::Close` case:

```rust
let timer_key = match timer {
    Timer::LossDetection => {
        if let Some(mut key) = pending.loss_detect_timer.as_ref() {
            endpoint.timers.remove(key);
        }
        &mut pending.loss_detect_timer
    }
    Timer::Idle => {
        if let Some(mut key) = pending.idle_timer.as_ref() {
            endpoint.timers.remove(key);
        }
        &mut pending.idle_timer
    }
    Timer::Close => &mut None,  // cannot return a reference
};
*timer_key = Some(endpoint.timers.insert_at((connection, timer), instant));
```

little-dude: I think I got it. Does 31f61a0c1129cdb8dc7ccba49bc83c12167cc733 do what you described?

Ralith:
Running the example client/server produces a panic at the `reset_at` call with the message `assertion failed: old >= delay.deadline()`

little-dude:
> but I'm also surprised that it's happening in the first place.

One scenario: when timers are polled we may have an expired timer, but we don't reset `loss_detect_timer` or `idle_timer` to `None`, so in the:

```rust
while let Some(io) = endpoint.inner.poll_io(now) {
```

we may use a key for an expired timer.

Ralith:
> we may use a key for an expired timer.

Great catch! That's almost certainly the root cause.

> There's something weird though:

Timer 1 was probably removed due to a `TimerStop` event.

little-dude:
I fixed the "invalid key" panic. But the client still crashes consistently, and the server once in a while.

little-dude:
Thanks for the tip with `slog`. I had implemented my own `now_in_ms()` function that I was calling in each `println!()`... I get the exact same results than you: it crashes when we reset the timer 1ms after inserting it.

quinn/src/lib.rs
```diff
@@ -802,8 +802,9 @@ impl Future for Driver {
                         timer: timer @ Timer::Close,
                         time,
                     } => {
-                        let instant = endpoint.epoch + duration_micros(time);
-                        endpoint.timers.insert_at((connection, timer), instant);
+                        endpoint
+                            .timers
+                            .insert((connection, timer), duration_micros(time));
```
Ralith: This is incorrect; `time` is microseconds since epoch, not from the present.

quinn/src/lib.rs
```diff
@@ -815,17 +816,19 @@ impl Future for Driver {
...
                         if let Some(ref key) = timer_key {
-                            endpoint.timers.reset_at(key, instant);
+                            endpoint.timers.reset(key, duration_micros(time));
                         } else {
-                            *timer_key =
-                                Some(endpoint.timers.insert_at((connection, timer), instant));
+                            *timer_key = Some(
+                                endpoint
+                                    .timers
+                                    .insert((connection, timer), duration_micros(time)),
```
Ralith: As above.

little-dude:
sorry for the confusion I was trying things to see if that got rid of the error. I cleaned things up a little and squashed the commits.

djc:
So what's the status here? How do we move forward?

little-dude:
I need to repro the panic in `tokio-timer`. I haven't had time for the past 2 weeks due to unexpected events IRL. I have to move out by the end of the month so it's unlikely that I'll be able to work on this before December :(

djc:
No worries, good luck wih the move!

djc:
I think this will become blocking soon, do you still have time to work on this in the next week or so? If not, I'll build on your work and finish it off.

Ralith:
I'm having another go with this in the course of experimenting with async/await, and am making progress on narrowing down a root cause: https://github.com/tokio-rs/tokio/issues/849

Ralith:
Testing on the nightly branch suggests that `DelayQueue` itself is now relatively debugged.

Ralith:
Mooted by #264

PR #81, opened 2018-10-31
[WIP] Refactor TransportParameters to separate client/server types

### LINKED ISSUE #30: Implement separate types for client and server TransportParameters
The current API is pretty confusing because it requires passing in a `Side`, but it's not obvious if that is the API user's side or the message originator's side. My original implementation had explicit `ClientTransportParameters` and `ServerTransportParameters` types, which is more obvious, and can also directly implement the `Value`/`Codec` trait.

jansol:
Sorry about the slow progress (and the messy state of the code so far). The month ended up seeing way more action than I like and this task turned out to require a bit more planning than I first expected. Still not quite sure how to best do the parameter parsing for `read()`, since the "common" and server-specific params may be in any order. Maybe something around a mutable TransportParameters that tries to parse & set a single field per call?

(eventually)
Closes #30

### DIFF
--- a/quinn-proto/src/transport_parameters.rs
+++ b/quinn-proto/src/transport_parameters.rs
@@ -8,7 +8,7 @@ use bytes::{Buf, BufMut};
 use coding::{BufExt, BufMutExt};
 use endpoint::Config;
 use packet::ConnectionId;
-use {Side, TransportError, MAX_CID_SIZE, MIN_CID_SIZE, VERSION};
+use {TransportError, MAX_CID_SIZE, MIN_CID_SIZE, VERSION};
 
 // Apply a given macro to a list of all the transport parameters having simple integer types, along with their codes and
 // default values. Using this helps us avoid error-prone duplication of the contained information across decoding,
@@ -41,11 +41,6 @@ macro_rules! make_struct {
             $(pub $name : $ty,)*
 
             pub disable_migration: bool,
-
-            // Server-only
-            pub original_connection_id: Option<ConnectionId>,
-            pub stateless_reset_token: Option<[u8; 16]>,
-            pub preferred_address: Option<PreferredAddress>,
         }
 
         impl Default for TransportParameters {
@@ -55,10 +50,6 @@ macro_rules! make_struct {
                     $($name: $default,)*
 
                     disable_migration: false,
-
-                    original_connection_id: None,
-                    stateless_reset_token: None,
-                    preferred_address: None,
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


djc:
What I was looking for here is simply to have separate Rust types for the "wrapper types". So simply have a `ClientTransportParameters` that has `initial_version` and `parameters` fields, `ServerTransportParameters` that has `negotiated_version`, `supported_versions`, and `parameters` fields, and leave the actual parameter lists in a dedicated `TransportParameters` type.

BTW, skimming your patch so far (which I think will be simpler if you follow the outline above) it looks like you have the `ClientTransportParameters::read()` and `ServerTransportParameters::read()` functions backwards at least compared to what I had in mind -- you have the `ClientTransportParameters::read()` function actually reading `ServerTransportParameters` structure, as if it's the client's transport parameters reader function, whereas what I was aiming for is to have the methods very clearly aligned to the type they're connected with. Does that make sense?

Ralith:
> Maybe something around a mutable TransportParameters that tries to parse & set a single field per call?

That seems like a decent idea, since the type-length-value encoding means you can iterate over fields regardless. Note that the server's `read` function *must* check for the presence of illegally-supplied server transport parameters from clients.

Ralith:
> What I was looking for here is simply to have separate Rust types for the "wrapper types". So simply have a ClientTransportParameters that has initial_version and parameters fields, ServerTransportParameters that has negotiated_version, supported_versions, and parameters fields, and leave the actual parameter lists in a dedicated TransportParameters type.

Isn't that roughly what he's doing? There's the common inner `TransportParameters` type, and the outer types that wrap that and add the endpoint-specific fields.

quinn-proto/src/transport_parameters.rs
```diff
@@ -371,22 +394,29 @@ mod test {
...
+            original_connection_id: None,
+            stateless_reset_token: None,
         };
-        params.write(Side::Server, &mut buf);
+
+        server_params.write(&mut buf);
         assert_eq!(
-            TransportParameters::read(Side::Client, &mut buf.into_buf()).unwrap(),
-            params
+            ClientTransportParameters::read(&mut buf.into_buf()).unwrap().params,
```
Ralith: `ServerTransportParameters::read` is the function that would decode an encoded `ServerTransportParameters` value.

Ralith: This should be `ServerTransportParameters::read`, and compare the whole value.

djc:
@jansol hey, so do you want to finish this off? No pressure, if you don't have time I'll take care of it.

jansol:
Yeah, still on it.

jansol:
Okay now it looks like this is going to rabbithole into some more or less drastic restructuring of `connection.rs` which I'm not sure I have time or energy to do. (That or wrapping the newly split structs into an enum which would defeat the point of splitting them in the first place.) Might be mistaken though, so some feedback at this stage would be welcome either way.

djc:
Why is that? You should definitely not put the `ClientTransportParameters` and `ServerTransportParamaters` types in `Connection` directly, only their nested `TransportParameters` type. We only use the other contained values to check that `VERSION` (the version of the protocol supported by Quinn) is included in the supported versions from the peer.

jansol:
Oh. Yeah in that case it's a lot simpler.

Ralith:
The client does need to store the server's stateless reset token, but that can just be an optional field on `Connection`.

djc:
I've merged all the work from the draft-15 branch now, so it probably makes sense to change this PR to target master (or open a new one).

Issue #76, opened 2018-10-19
Panic when trying to shutdown stream


icefoxen:
My code is actually almost well modularized now, so it's easier to post issue reports.  Huzzah!

This panics with "thread 'main' panicked at 'unknown stream', libcore/option.rs:989:5".  This appears to be happening in the `tokio::io::shutdown(stream)` line.

```rust
    fn receive_message(stream: quinn::Stream) -> impl Future<Item = (), Error = ()> {
        quinn::read_to_end(stream, 1024 * 64)
            .map_err(|e| warn!("failed to read response: {}", e))
            .and_then(move |(stream, req)| {
                let msg: ::std::result::Result<Message, rmp_serde::decode::Error> =
                    rmp_serde::from_slice(&req);
                debug!("Got message: {:?}", msg);
                let to_do_next: Box<dyn Future<Item = quinn::Stream, Error = ()>> = match msg {
                    Ok(Message::Ping { id }) => {
                        info!("Got ping, trying to send pong");
                        let message = Message::Pong { id };
                        let to_send = rmp_serde::to_vec(&message)
                            .expect("Could not serialize message; should never happen!");
                        Box::new(
                            tokio::io::write_all(stream, to_send)
                                .map_err(|e| warn!("Failed to send request: {}", e))
                                .map(|(stream, _vec)| stream),
                        )
                    }
                    Ok(val) => {
                        info!("Got message: {:X?}, not doing anything with it", val);
                        Box::new(future::ok(stream))
                    }
                    Err(e) => {
                        info!("Got unknown message: {:X?}, error {:?}", &req, e);
                        Box::new(future::ok(stream))
                    }
                };
                to_do_next
                    .and_then(|stream| {
                        trace!("Trying to shut down stream");
                        tokio::io::shutdown(stream)
                            .and_then(|v| {
                                trace!("Done!");
                                future::ok(v)
                            }).map_err(|e| warn!("Failed to shut down stream: {}", e))
                    }).map(move |_| info!("request complete"))
            })
    }
```

This appears to correspond to this code in `quinn-proto/src/connection.rs`:

```rust
    pub fn finish(&mut self, id: StreamId) {
        let ss = self
            .streams
            .get_mut(&id)
            .expect("unknown stream")
            .send_mut()
            .expect("recv-only stream");
        ....
    }
```

icefoxen:
[...]

PR #82, opened 2018-11-01
Improve a log message and ConnectionId Debug impl

Ralith:


quinn-proto/src/packet.rs
```diff
@@ -323,6 +323,12 @@ impl ::std::ops::DerefMut for ConnectionId {
     }
 }
 
+impl fmt::Debug for ConnectionId {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
djc: What do you think about using a more compact representation, like I used to have?

https://github.com/djc/quinn/blob/pre-quicr-quinn/src/types.rs#L35

Also, since it's very easy to do, maybe add a unit test for this?

Ralith: That looks a lot like the existing `Display` impl. I believe it's idiomatic for `Debug` impls to produce valid rust syntax, for e.g. easy copy-pasting into a unit test.

djc: Ah, yeah, you're right.

Though in that case I'd be in favor of using `Display` rather than `Debug` in log messages (but that can be a different change -- maybe a first good patch?).

Ralith: Isn't that mostly already being done?

djc: Ah, I suppose you're right. I guess I got confused about the logging in #76.

PR #90, opened 2018-11-16
Clarify some high-level documentation

Ralith:
Fruits of some discussion on obviousness of semantics in IRC.

### DIFF
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -945,11 +945,11 @@ impl Connection {
         let conn = self.0.clone();
         recv.map_err(|_| unreachable!())
             .and_then(|result| result)
-            .map(move |stream| SendStream(Stream::new(conn, stream)))
+            .map(move |stream| SendStream(BiStream::new(conn, stream)))
     }
 
     /// Initiate a new outgoing bidirectional stream.
-    pub fn open_bi(&self) -> impl Future<Item = Stream, Error = ConnectionError> {
+    pub fn open_bi(&self) -> impl Future<Item = BiStream, Error = ConnectionError> {
         let (send, recv) = oneshot::channel();
         {
             let mut endpoint = self.0.endpoint.borrow_mut();
@@ -964,7 +964,7 @@ impl Connection {
         let conn = self.0.clone();
         recv.map_err(|_| unreachable!())
             .and_then(|result| result)
-            .map(move |stream| Stream::new(conn.clone(), stream))
+            .map(move |stream| BiStream::new(conn.clone(), stream))
     }
 
     /// Close the connection immediately.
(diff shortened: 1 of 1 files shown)


quinn/src/lib.rs
```diff
@@ -1074,13 +1074,15 @@ pub trait Read {
     ///
     /// Returns a segment of data and their offset in the stream. Segments may be received in any order and may overlap.
     ///
-    /// Using this function reduces latency improves throughput by avoiding head-of-line blocking within the stream, and
-    /// reduces computational overhead by allowing data to be passed on without any intermediate buffering. Prefer it
-    /// whenever possible.
+    /// Using this function reduces latency and improves throughput by avoiding head-of-line blocking within the stream,
```
djc: You didn't change this much content-wise, but I'll offer a suggestion anyway: I think this comment doesn't do a very good job of highlighting the trade-offs involved. QUIC-as-specified is still largely geared towards in-order delivery (DATAGRAM discussions notwithstanding), and while it can be a viable choice to opt into unordered delivery for some applications, "Prefer it whenever possible." while probably correct feels a little biased to me.

quinn/src/lib.rs
```diff
@@ -1090,29 +1092,42 @@ pub trait Read {
...
+    /// No new data may be written after calling this method. Completes when the peer has acknowledged all sent data,
+    /// retransmitting data as needed.
     fn poll_finish(&mut self) -> Poll<(), ConnectionError>;
 
-    /// Abandon transmitting data on this stream.
+    /// Close the outgoing stream immediately.
     ///
-    /// No new data may be transmitted, and no previously transmitted data will be retransmitted if lost.
+    /// No new data can be written after calling this method. Locally buffered data is dropped, and previously
+    /// transmitted data will be no longer be retransmitted if lost. If `poll_finish` was called previously and all data
```
djc: Nit: "be no longer be".

quinn/src/lib.rs
```diff
@@ -1090,29 +1092,42 @@ pub trait Read {
...
+    /// has already been transmitted at least once, the peer may still receive all written data.
     fn reset(&mut self, error_code: u16);
 }
 
-/// A stream that supports both sending and receiving data
+/// A stream that supports both sending and receiving data.
```
djc: This wording is maybe a little off. Is it a "stream" or a "bundle" of two streams? I'd say it cannot be both.

quinn/src/lib.rs
Ralith: I think it's important to distinguish the delivery of data in the order it was sent from the delivery of information in a manner that preserves the data's position in its stream. QUIC is geared towards doing the latter, but I don't think the former is specifically favored or even all that useful.

quinn/src/lib.rs
Ralith: Yeah, I wasn't sure about this myself. I'm trying to establish a notion of unidirectional streams as a primitive so that the documentation on `Read` and `Write` can easily talk about uniform semantics between unidirectional streams and one half of a bidirectional stream. Do you have specific ideas for a better approach?

quinn/src/lib.rs
djc: Hmm, I don't really agree. In a lot of places, QUIC is being pitched as a TCP replacement, which I think implies reliable in-order delivery to the application layer -- which is also the simplest to use, which is also an important consideration.

All of which to say, I think reliable in-order delivery should be our default public API in a sense, with unordered (and, in the future, unreliable) alternatives for those who really needed.

quinn/src/lib.rs
djc: I think it makes sense to keep "stream" as the unidirectional concept, but then we should perhaps consistently use "bundle" for the combination. Or "pair", "stream pair", which is maybe a little bit less generic.

quinn/src/lib.rs
Ralith: On consideration, the current advice is certainly too strong in light of ordered messaging applications, which are common. I'll go ahead and weaken the language. There are certainly lots of traditional TCP applications (e.g. bulk data transfer) where unordered reads make much more sense, but they're indeed not universal.

quinn/src/lib.rs
Ralith: Think we should go so far as to adjust type names?

djc: Would probably make things easier in the end, yes.

djc: Maybe we should bring it up with the spec folks too?

Ralith: I've attempted something more consistent with the spec; let me know what you think.

PR #92, opened 2018-11-17
Remove Blake2 and use SHA512_256 for reset_token instead

stammw:
Creating this pull request to show what I've already done. I don't quite get the part where:

>  it would probably good to store the SigningKey in the ListenKeys and work from that

But I'll work on this now.

### DIFF
--- a/quinn-proto/src/crypto.rs
+++ b/quinn-proto/src/crypto.rs
@@ -87,9 +87,8 @@ pub const ACK_DELAY_EXPONENT: u8 = 3;
 /// Magic value used to indicate 0-RTT support in NewSessionTicket
 //pub const TLS_MAX_EARLY_DATA: u32 = 0xffff_ffff;
 
-pub fn reset_token_for(key: &[u8], id: &ConnectionId) -> [u8; RESET_TOKEN_SIZE] {
-    let mac = SigningKey::new(&digest::SHA512_256, key);
-    let signature = hmac::sign(&mac, id);
+pub fn reset_token_for(key: &SigningKey, id: &ConnectionId) -> [u8; RESET_TOKEN_SIZE] {
+    let signature = hmac::sign(key, id);
     // TODO: Server ID??
     let mut result = [0; RESET_TOKEN_SIZE];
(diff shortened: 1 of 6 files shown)


quinn-proto/src/crypto.rs
```diff
@@ -92,11 +88,11 @@ pub const ACK_DELAY_EXPONENT: u8 = 3;
...
 
 pub fn reset_token_for(key: &[u8], id: &ConnectionId) -> [u8; RESET_TOKEN_SIZE] {
-    let mut mac = VarBlake2b::new_keyed(key, RESET_TOKEN_SIZE);
-    mac.input(id.as_ref());
+    let mac = SigningKey::new(&digest::SHA512_256, key);
+    let signature = hmac::sign(&mac, id);
     // TODO: Server ID??
     let mut result = [0; RESET_TOKEN_SIZE];
-    mac.variable_result(|res| result.copy_from_slice(res));
+    result.copy_from_slice(&signature.as_ref()[..16]);
```
```suggestion
    result.copy_from_slice(&signature.as_ref()[..RESET_TOKEN_SIZE]);
```

djc:
Looks great to me! @Ralith knows this code better than me, so I'll leave the final review to him.

Thanks again for working on this... if you want to continue working on Quinn, we can definitely come up with more things. 😉

quinn-proto/src/endpoint.rs
```diff
@@ -168,6 +169,7 @@ pub struct ListenKeys {
     ///
     /// Initialize with random bytes.
     pub reset: [u8; 64],
+    pub reset_key: SigningKey,
```
Ralith: I'm not a huge fan of having both of these fields in the struct, especially with no mechanism to prevent a struct from being constructed where their values are unrelated. Perhaps make `reset` private, and we can bug upstream to impl `Clone` for `SigningKey` to make this better in the future?

Ralith: Maybe a better approach would be to get rid of `reset` and modify the tests to generate new `ListenKeys` on demand.

stammw:
@djc Thanks! I'd love to contribute more !

Issue #17, opened 2018-05-24
Cache Sealing and Opening keys in secrets


djc:
The `crypto` module currently has a `PacketKey` type that reconstructs the relevant `SealingKey` and `OpeningKey` for every `seal` or `open` operation. There should be a smarter way to structure the types in the `crypto` module so that a `Secret` has direct access to long-lived `Key` instances.

djc:
@stammw is this something you want to try next? (The description isn't quite valid anymore as some of the type names are different from when I wrote this up, but the gist of it still holds.)

stammw:
Sure ! I'll see what I can come up with !

djc:
Cool. Don't hesitate to ask if you have questions!

stammw:
@djc I believe this issue can be closed.

PR #94, opened 2018-11-22
Retain sealing and opening key

stammw:
Here is a pretty trivial solution for #17. Is it what you expected ?

Similarly to #92, built keys and the data used to build them could be constructed wrongly by using a bad `CryptoContext { ... }`. I propose to move key and `CrytpoState` building logic into a `CryptoContext::new()` method to prevent this from happening.

Everything would be easier with all the key types implementing `Clone`. I'll see what can be done on this side. But I wonder why it is not already the case?

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -1898,7 +1898,8 @@ impl Connection {
         let mut sent = Retransmits::default();
 
         let (number, acks, ack_only, handshake) = {
-            let (number, header, crypto, pending) = if (!established || self.awaiting_handshake)
+            let (number, header, crypto, pending, crypto_level) = if (!established
+                || self.awaiting_handshake)
                 && (!self.handshake_pending.is_empty()
                     || (!self.pending_acks.is_empty() && self.permit_ack_only))
             {
@@ -1936,6 +1937,7 @@ impl Connection {
                     header,
                     &self.handshake_crypto,
                     &mut self.handshake_pending,
+                    CryptoLevel::Initial,
                 )
             } else if established {
                 //|| (self.zero_rtt_crypto.is_some() && self.side == Side::Client) {
@@ -1970,6 +1972,7 @@ impl Connection {
                     header,
                     self.crypto.as_ref().unwrap(),
                     &mut self.pending,
+                    CryptoLevel::OneRtt,
                 )
             } else {
                 return None;
@@ -2157,7 +2160,7 @@ impl Connection {
                     );
                 }
             }
-            if !crypto.is_1rtt() {
+            if crypto_level != CryptoLevel::OneRtt {
                 let pn_len = match header {
                     Header::Initial { number, .. } | Header::Long { number, .. } => number.len(),
                     _ => panic!("invalid header for packet payload length"),
@@ -2166,7 +2169,7 @@ impl Connection {
             }
             crypto.encrypt(number, &mut buf, header_len as usize);
             partial_encode.finish(&mut buf, crypto.pn_encrypt_key(), header_len as usize);
-            (number, acks, ack_only, crypto.is_initial())
+            (number, acks, ack_only, crypto_level == CryptoLevel::Initial)
         };
 
         // If we sent any acks, don't immediately resend them. Setting this even if ack_only is
@@ -2664,6 +2667,12 @@ impl Connection {
         }
         None
     }
+}
+
+#[derive(Eq, PartialEq)]
+enum CryptoLevel {
+    Initial,
+    OneRtt,
 }
 
 /// Extract stream 0 data from an Initial or Retry packet payload
--- a/quinn-proto/src/crypto.rs
+++ b/quinn-proto/src/crypto.rs
@@ -99,10 +99,16 @@ pub fn reset_token_for(key: &SigningKey, id: &ConnectionId) -> [u8; RESET_TOKEN_
     result
 }
 
+pub struct OneRttCrypto {
+    context: CryptoContext,
+    local_secret: Vec<u8>,
+    remote_secret: Vec<u8>,
+}
+
 pub enum Crypto {
     // ZeroRtt(ZeroRttCrypto),
     Initial(CryptoContext),
-    OneRtt(CryptoContext, Vec<u8>, Vec<u8>),
+    OneRtt(OneRttCrypto),
 }
 
 impl Crypto {
@@ -180,8 +186,8 @@ impl Crypto {
         let (local_key, local_iv, local_pn_key) = Self::get_keys(digest, cipher, &local_secret);
         let (remote_key, remote_iv, remote_pn_key) = Self::get_keys(digest, cipher, &remote_secret);
 
-        Crypto::OneRtt(
-            CryptoContext {
+        Crypto::OneRtt(OneRttCrypto {
+            context: CryptoContext {
                 sealing_key: aead::SealingKey::new(cipher, &local_key).unwrap(),
                 local_pn_key,
                 local_iv,
@@ -192,7 +198,7 @@ impl Crypto {
             },
             local_secret,
             remote_secret,
-        )
+        })
     }
 
     pub fn new_1rtt(tls: &TlsSession, side: Side) -> Self {
@@ -236,7 +242,7 @@ impl Crypto {
 
     pub fn is_1rtt(&self) -> bool {
         match *self {
-            Crypto::OneRtt(_, _, _) => true,
+            Crypto::OneRtt(_) => true,
             _ => false,
         }
     }
@@ -266,7 +272,8 @@ impl Crypto {
     fn ctx(&self) -> &CryptoContext {
         use self::Crypto::*;
         match self {
-            Initial(ctx) | OneRtt(ctx, _, _) => ctx,
+            Initial(ctx) => ctx,
+            OneRtt(crypto) => &crypto.context,
         }
     }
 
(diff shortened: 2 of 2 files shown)


quinn-proto/src/crypto.rs
```diff
@@ -270,20 +273,25 @@ impl Crypto {
         self.write_nonce(&state, packet, nonce);
         let payload_len = payload.len();
 
-        let key = aead::OpeningKey::new(cipher, &state.key).unwrap();
         aead::open_in_place(&key, &*nonce, header, 0, payload.as_mut()).map_err(|_| ())?;
         payload.split_off(payload_len - cipher.tag_len());
         Ok(())
     }
 
     pub fn update(&self, side: Side) -> Crypto {
```
stammw: Couldn't `CryptoContext::local` and `CryptoContext::remote` just be swapped in `CryptoContext::update` `CryptoState::update` use the same `cipher` and `digest` [here](https://github.com/djc/quinn/blob/f9a6935b2ffcda1f08e600b5ab4b87668965a5fc/quinn-proto/src/crypto.rs#L384) ?

quinn-proto/src/crypto.rs
```diff
@@ -390,12 +398,26 @@ pub struct ZeroRttCrypto {
...
+    fn clone(&self) -> CryptoContext {
+        Self {
+            sealing_key: aead::SealingKey::new(self.cipher, &self.local.key.clone()).unwrap(),
+            opening_key: aead::OpeningKey::new(self.cipher, &self.remote.key).unwrap(),
+            local: self.local.clone(),
+            remote: self.remote.clone(),
+            digest: self.digest,
+            cipher: self.cipher,
+        }
+    }
```
Ralith: I don't think we actually clone this structure ever, so we might as well remove the impl.

quinn-proto/src/crypto.rs
Ralith: I don't understand this question; can you rephrase?

quinn-proto/src/crypto.rs
```diff
@@ -390,12 +398,26 @@ pub struct ZeroRttCrypto {
     cipher: &'static aead::Algorithm,
 }
 
-#[derive(Clone)]
 pub struct CryptoContext {
     local: CryptoState,
     remote: CryptoState,
     digest: &'static digest::Algorithm,
     cipher: &'static aead::Algorithm,
```
Ralith: *ring* keys store their `&'static aead::Algorithm` internally, so we can remove the `cipher` member.

quinn-proto/src/crypto.rs
```diff
@@ -390,12 +398,26 @@ pub struct ZeroRttCrypto {
...
 }
 
-#[derive(Clone)]
 pub struct CryptoContext {
     local: CryptoState,
     remote: CryptoState,
```
Ralith: As in the previous PR, I'd like to avoid storing the key bytes redundantly. This will probably require some refactoring, perhaps including doing away with `CryptoState`.

quinn-proto/src/crypto.rs
djc: I think you're asking if we can mutate the existing `Crypto` instead of instantiating a new one? In which case, we cannot, because we need to keep the old `Crypto` around for older packets that might still come in.

Ralith:
Another option might be to not keep `CryptoState`s around at all; store the secrets and IVs directly in the `CryptoContext, and construct the suitable AEAD key types. You'd probably want to tweak `CryptoState::update` to replace the `&self` argument with `secret: &[u8]` to make this convenient.

stammw:
I've followed what `ZeroRttCrypto` seemed to work to move retained secrets of `OneRtt` into a new `OneRttCrypto` type. 

The remaining ZeroRtt code as been commented out as I'm not sure of how I should reshape it after removing `CryptoState`. Tell me if you want me to update the commented code.

I think it might be worth to writing a test for `Crypto::update()`. But for the moment it's far from being trivial for me...

quinn-proto/src/crypto.rs
```diff
@@ -204,51 +247,58 @@ impl Crypto {
...
         use self::Crypto::*;
         match self {
-            Initial(ctx) | OneRtt(ctx) => ctx,
+            Initial(ctx) => ctx,
+            OneRtt(crypto) => &crypto.context,
         }
     }
 
     pub fn encrypt(&self, packet: u64, buf: &mut Vec<u8>, header_len: usize) {
         // FIXME: retain crypter
```
Ralith: Stale comment

quinn-proto/src/crypto.rs
```diff
@@ -99,11 +99,16 @@ pub fn reset_token_for(key: &SigningKey, id: &ConnectionId) -> [u8; RESET_TOKEN_
     result
 }
 
-#[derive(Clone)]
+pub struct OneRttCrypto {
+    context: CryptoContext,
+    local_secret: Vec<u8>,
+    remote_secret: Vec<u8>,
+}
```
Ralith: I'm not sure this merits having a separate type, since it's harmless for initial to have these fields and either store its secrets in them for consistency, or just ignore them. @djc, do you have an opinion?

Ralith:
I'd say you should feel free to nuke stale commented 0-RTT related code that the rest of the code has diverged from.

stammw:
Ok great! Can I change the order of definitions a bit? I'm used to:

```
struct S1 { }
pub struct S2 { s1: S1 }

fn f1 { }

impl S1 {}
impl S2 { s1.foo() }
```

(When a symbol is used in the modules, it is always declare before hand.)

stammw:
What about CookieFactory, should go away as well?

quinn-proto/src/crypto.rs
djc: Yeah, I think it should just be a single type, and store the initial secrets in them.

Ralith:
> Can I change the order of definitions a bit?

I prefer that associated items on a type always immediately follow the type definitions, with top-level items close to their use site. Refer to [the Rust style guidelines](https://doc.rust-lang.org/1.0.0/style/style/organization.html#define-types-and-operations-together.).

> What about CookieFactory, should it go away as well?

That's unrelated to your changes, so better to leave it alone. I should probably nuke it in #98 though, totally forgot it was there...

quinn-proto/src/crypto.rs
```diff
@@ -133,25 +149,52 @@ impl Crypto {
...
+            local_pn_key,
+            local_iv,
+            opening_key: aead::OpeningKey::new(cipher, &remote_key).unwrap(),
+            remote_pn_key,
+            remote_iv,
             digest,
         })
     }
 
+    fn generate_1rtt(
```
djc: Can we call this `new_1rtt()` by analogy with `new_initial()`?

quinn-proto/src/crypto.rs
```diff
@@ -102,7 +102,7 @@ pub fn reset_token_for(key: &SigningKey, id: &ConnectionId) -> [u8; RESET_TOKEN_
 pub enum Crypto {
```
djc: Actually, throughout this PR I have been feeling that this enum is not pulling its weight, and we should just use `CryptoContext` as the `Crypto` type. At this point, it seems we have more code working around the variants than using them, and I think usages in `Connection` are disambiguated by their usage sites. @Ralith what do you think?

@stammw I don't want to let your PR spiral out of control if you don't want it to, so do with this comment what you like.

djc:
> Can I change the order of definitions a bit?

What @Ralith said, but also: over my learning Rust, I've come to appreciate that, because the compiler doesn't mandate that dependencies are defined before their dependees, for human readers it's often preferable to have high-level API up first and then fill in the details (dependencies) below that. So in your example, I'd order them like this:

```rust
pub struct S2 { s1: S1 }

impl S2 { s1.foo() }

struct S1 { }

impl S1 {}

fn f1 { }
```

(So, S2 comes before S1 because S2 depends on S1, structs and their impls close together, helper functions at the bottom.)

quinn-proto/src/crypto.rs
Ralith: I'm 100% on board with ditching the enum outright.

stammw:
@djc This is the structure I understood working on the file. I'll fix the codes I added in the wrong places then.

stammw:
@djc This is the structure I understood working on the file. I'll fix the codes I added in the wrong places then.

quinn-proto/src/crypto.rs
stammw: I'll move them to `CryptoContext` and assign them into `new_initial()` as well.

Then removing the `enum` should not be a problem anymore.

quinn-proto/src/crypto.rs
```diff
@@ -200,54 +236,54 @@ impl Crypto {
...
     pub fn is_1rtt(&self) -> bool {
         match *self {
-            Crypto::OneRtt(_) => true,
+            Crypto::OneRtt(_, _, _) => true,
             _ => false,
         }
     }
 
-    pub fn write_nonce(&self, state: &CryptoState, number: u64, out: &mut [u8]) {
+    pub fn write_nonce(&self, iv: &Vec<u8>, number: u64, out: &mut [u8]) {
```
djc: There's almost never a good reason to pass a `&Vec<u8>`; use a `&[u8]` instead (on the caller side, it will coerce automatically).

quinn-proto/src/connection.rs
```diff
@@ -2666,6 +2669,27 @@ impl Connection {
...
+        }
+    }
+
+    fn is_1rtt(&self) -> bool {
+        match *self {
+            CryptoLevel::OneRtt => true,
+            _ => false,
+        }
+    }
+}
```
Ralith: Given that this enum likely won't ever contain data, these methods should be removed in favor of applying `#[derive(Eq, PartialEq)]` to the enum and inspecting values with `==`/`!=`.

stammw: This is what I was looking for, thank you.

stammw:
@Ralith Regarding updating keys, I thought that how they are rebuilt is defined in the RFC. But from what I understand, it is actually arbitrary and implementation-defined, as long as the updated keys are different from the previous ones? 

In this case using this kind of test cas should be sufficient:

```rust
let old = Crypto::new_1rtt();
let new = old.update();
assert(old.keys != new.keys)
```

Am I right? 

Couldn't I use `rand` crate to build them ?

djc:
I've rebased and squashed these changes and submitted them for approval in #102. @stammw thanks for sticking with it!

stammw:
Great! Thank you for the guidance. 

What about the unit test and the update "bug" here ? Should I continue with this?

Else, feel free to point what I can work on next !

djc:
Yeah, continuing with that unit test and fixing the update code would be a great next step!

PR #97, opened 2018-11-24
Clarifications for the streams API

djc:
Hope this doesn't interfere (too much) with any work you've done.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -15,7 +15,7 @@ use packet::{
     AEAD_TAG_SIZE,
 };
 use range_set::RangeSet;
-use stream::{self, Stream};
+use stream::{self, ReadError, Stream};
 use transport_parameters::{self, TransportParameters};
 use {
     frame, Directionality, Frame, Side, StreamId, TransportError, MIN_INITIAL_SIZE, MIN_MTU,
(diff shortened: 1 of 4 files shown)


quinn-proto/src/stream.rs
```diff
@@ -122,6 +122,46 @@ impl Recv {
...
+                RecvState::Recv { .. } => Err(ReadError::Blocked),
+                RecvState::DataRecvd { .. } => {
+                    self.state = RecvState::Closed;
+                    Err(ReadError::Finished)
+                }
+            }
+        }
+    }
+
+    pub fn has_pending_incoming(&self) -> bool {
```
Ralith: This naming is unclear to me. The condition we're actually checking here is whether the stream is in a state where the size (i.e. the final offset) is unknown, so maybe `has_unknown_size`, or invert it into `size_known`? Maybe even `pub fn size(&self) -> Option<u64>`, for use with `is_none()`?

quinn-proto/src/connection.rs
```diff
@@ -2223,6 +2223,14 @@ impl Connection {
...
         data: &[u8],
     ) -> Result<usize, WriteError> {
+        if self.state.as_ref().unwrap().is_closed() {
+            return Err(WriteError::Blocked);
+        }
+        assert!(stream.directionality() == Directionality::Bi || stream.initiator() == self.side);
+        if self.blocked() {
+            self.blocked_streams.insert(stream);
+            return Err(WriteError::Blocked);
+        }
```
Ralith: This partially defeats the debug logging just below, preventing some cases of blocked writes from being traced. What's the motivation?

quinn-proto/src/stream.rs
djc: I renamed it to `receiving_size_known()`, inverting the condition. `size()` would've exposed more details than necessary, so I prefer the simpler variant that returns `bool`. I added `receiving` because it's not *just* about the value of the size, the actual state is important too.

quinn-proto/src/connection.rs
djc: I think code is often easier to follow if error conditions and such are handled first, leading to early returns (or continue/break), then the actual logic follows. I've adjusted the debug logging now, so that it's actually more specific. What do you think?

quinn-proto/src/stream.rs
Ralith: > it's not just about the value of the size, the actual state is important too.

I don't totally follow this. The current impl returns `true` for all states except recv with unknown size, including the non-receiving states like `ResetRecvd` and `Closed`.

quinn-proto/src/connection.rs
Ralith: Looks good! Wouldn't hurt to log the affected stream ID too.

quinn-proto/src/stream.rs
djc: Huh, yeah, that was muddled. I've renamed it to `receiving_unknown_size()` now.

PR #98, opened 2018-11-24
Draft 15 updates

Ralith:
Independent fixes on the path towards updating the handshake state machine. I may add more to this as I pursue that goal.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -36,7 +36,9 @@ pub struct Connection {
     pub side: Side,
     pub handle: ConnectionHandle,
     pub mtu: u16,
+    /// Highest received packet number
     pub rx_packet: u64,
+    /// Time at which the above was received
     pub rx_packet_time: u64,
     pub crypto: Option<Crypto>,
     pub prev_crypto: Option<(u64, Crypto)>,
@@ -2127,7 +2129,7 @@ impl Connection {
                 return Err(None);
             }
         };
-        let number = number.expand(self.rx_packet);
+        let number = number.expand(self.rx_packet + 1);
         if key_phase != self.key_phase {
             if number <= self.rx_packet {
                 // Illegal key update
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -8,7 +8,7 @@ use fnv::{FnvHashMap, FnvHashSet};
 use slog::Logger;
 
 use coding::{BufExt, BufMutExt};
-use crypto::{self, reset_token_for, Crypto, TLSError, TlsSession, ACK_DELAY_EXPONENT};
+use crypto::{self, reset_token_for, Crypto, TlsSession, ACK_DELAY_EXPONENT};
 use endpoint::{Config, Context, Event, Io, Timer};
 use packet::{
     set_payload_length, ConnectionId, Header, LongType, Packet, PacketNumber, PartialDecode,
@@ -766,15 +766,12 @@ impl Connection {
         }
         */
 
-        if let Err(e) = self.tls.process_new_packets() {
-            debug!(self.log, "TLS error {}", e);
-            Err(if let TLSError::AlertReceived(_) = e {
-                TransportError::TLS_FATAL_ALERT_RECEIVED
-            } else {
-                TransportError::PROTOCOL_VIOLATION
-            })
-        } else {
-            Ok(())
+        match self.tls.process_new_packets() {
+            Ok(()) => Ok(()),
+            Err(e) => {
+                debug!(self.log, "TLS error {}", e);
+                Err(TransportError::PROTOCOL_VIOLATION)
+            }
         }
     }
 
@@ -793,8 +790,9 @@ impl Connection {
 
         trace!(self.log, "got initial");
         self.read_tls(&frame);
-        if self.tls.process_new_packets().is_err() {
-            return Err(TransportError::TLS_HANDSHAKE_FAILED);
+        if let Err(e) = self.tls.process_new_packets() {
+            debug!(self.log, "TLS error: {}", e);
+            return Err(TransportError::PROTOCOL_VIOLATION);
         }
         let params = TransportParameters::read(
             Side::Server,
@@ -992,7 +990,6 @@ impl Connection {
                         &self.loc_cid,
                         n as u8,
                         state.reason.clone(),
-                        state.alert.as_ref().map(|x| &x[..]),
                     ),
                 });
                 self.reset_idle_timeout(&ctx.config, now);
@@ -1128,7 +1125,7 @@ impl Connection {
                                     .get_quic_transport_parameters()
                                     .ok_or_else(|| {
                                         debug!(self.log, "remote didn't send transport params");
-                                        ConnectionError::from(TransportError::TLS_HANDSHAKE_FAILED)
+                                        ConnectionError::from(TransportError::PROTOCOL_VIOLATION)
                                     }).and_then(|x| {
                                         TransportParameters::read(
                                             self.side,
@@ -1181,7 +1178,7 @@ impl Connection {
                             }
                             Err(e) => {
                                 debug!(self.log, "handshake failed"; "reason" => %e);
-                                Err(TransportError::TLS_HANDSHAKE_FAILED.into())
+                                Err(TransportError::PROTOCOL_VIOLATION.into())
                             }
                         }
                     }
@@ -1443,14 +1440,14 @@ impl Connection {
                 }
                 Frame::Invalid(ty) => {
                     debug!(self.log, "received malformed frame"; "type" => %ty);
-                    return Err(TransportError::frame(ty));
+                    return Err(TransportError::FRAME_ENCODING_ERROR);
                 }
                 Frame::PathChallenge(x) => {
                     self.pending.path_challenge(number, x);
                 }
                 Frame::PathResponse(_) => {
                     debug!(self.log, "unsolicited PATH_RESPONSE");
-                    return Err(TransportError::UNSOLICITED_PATH_RESPONSE);
+                    return Err(TransportError::PROTOCOL_VIOLATION);
                 }
                 Frame::MaxData(bytes) => {
                     let was_blocked = self.blocked();
@@ -2301,7 +2298,6 @@ pub fn handshake_close<R>(
     local_id: &ConnectionId,
     packet_number: u8,
     reason: R,
-    tls_alert: Option<&[u8]>,
 ) -> Box<[u8]>
 where
     R: Into<state::CloseReason>,
@@ -2321,16 +2317,6 @@ where
     match reason.into() {
         state::CloseReason::Application(ref x) => x.encode(&mut buf, max_len),
         state::CloseReason::Connection(ref x) => x.encode(&mut buf, max_len),
-    }
-    if let Some(data) = tls_alert {
-        if !data.is_empty() {
-            frame::Stream {
-                id: StreamId(0),
-                fin: false,
-                offset: 0,
-                data,
-            }.encode(false, &mut buf);
-        }
     }
     set_payload_length(&mut buf, header_len, number.len());
     crypto.encrypt(packet_number as u64, &mut buf, header_len);
@@ -2375,7 +2361,7 @@ impl Streams {
                 Directionality::Uni => self.max_remote_uni,
             };
             if id.index() >= limit {
-                return Err(TransportError::STREAM_ID_ERROR);
+                return Err(TransportError::STREAM_LIMIT_ERROR);
             }
         }
         Ok(self.streams.get_mut(&id))
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -8,7 +8,7 @@ use fnv::{FnvHashMap, FnvHashSet};
 use slog::Logger;
 
 use coding::{BufExt, BufMutExt};
-use crypto::{self, reset_token_for, Crypto, TlsSession, ACK_DELAY_EXPONENT};
+use crypto::{self, Crypto, TlsSession, ACK_DELAY_EXPONENT};
 use endpoint::{Config, Context, Event, Io, Timer};
 use packet::{
     set_payload_length, ConnectionId, Header, LongType, Packet, PacketNumber, PartialDecode,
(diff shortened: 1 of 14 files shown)


quinn-proto/src/connection.rs
```diff
@@ -806,8 +790,9 @@ impl Connection {
 
         trace!(self.log, "got initial");
         self.read_tls(&frame);
-        if self.tls.process_new_packets().is_err() {
-            return Err(TransportError::TLS_HANDSHAKE_FAILED);
+        if let Err(e) = self.tls.process_new_packets() {
+            debug!(self.log, "TLS error: {}", e);
+            return Err(TransportError::PROTOCOL_VIOLATION);
```
djc: Shouldn't this generate `CRYPTO_ERROR` values?

quinn-proto/src/connection.rs
```diff
@@ -768,24 +768,8 @@ impl Connection {
 
         match self.tls.process_new_packets() {
             Ok(()) => Ok(()),
-            Err(e @ TLSError::AlertReceived(_)) => {
```
djc: Shouldn't this generate `CRYPTO_ERROR` values?

quinn-proto/src/connection.rs
```diff
@@ -1917,12 +1867,22 @@ impl Connection {
...
     fn make_close(&mut self, reason: &state::CloseReason) -> Box<[u8]> {
-        let number = self.get_tx_number();
+        let full_number = self.get_tx_number();
+        let number = PacketNumber::new(full_number, self.largest_acked_packet);
         let mut buf = Vec::new();
-        let header = Header::Short {
-            dst_cid: self.rem_cid,
-            number: PacketNumber::new(number, self.largest_acked_packet),
-            key_phase: self.key_phase,
+        let header = if self.crypto.is_some() {
```
djc: Hmm, seems like it would be clearer/more robust to check for `Established` state, instead of `crypto.is_some()`?

quinn-proto/src/frame.rs
```diff
@@ -35,10 +24,10 @@ impl Type {
 
 impl coding::Codec for Type {
     fn decode<B: Buf>(buf: &mut B) -> coding::Result<Self> {
-        Ok(Type(buf.get()?))
+        Ok(Type(buf.get_var()?))
```
djc: Maybe add some todo about enforcing the minimal encoding of varints here?

quinn-proto/src/packet.rs
```diff
@@ -600,20 +598,35 @@ impl PacketNumber {
...
-            x + d * ((t + d / 2 - x) / d)
+        let nbits = match self {
+            U8(_) => 7,
+            U16(_) => 14,
+            U32(_) => 30,
+        };
+        let win = 1 << nbits;
+        let hwin = win / 2;
+        let mask = win - 1;
+        // The incoming packet number should be greater than expected - hwin and less than or equal to expected + hwin
```
djc: Would you mind wrapping these at 100 columns?

quinn-proto/src/packet.rs
```diff
@@ -820,6 +833,23 @@ mod tests {
         check_pn(PacketNumber::U32(1073741823), &[0xff, 0xff, 0xff, 0xff]);
     }
 
+    #[test]
```
djc: Oops, thanks for catching all this. So do the new tests cover enough of the new behavior, or is there more we could/should test here?

quinn-proto/src/connection.rs
```diff
@@ -57,6 +57,10 @@ pub struct Connection {
     /// Limit on incoming data
     pub local_max_data: u64,
     client_config: Option<ClientConfig>,
+    /// Incoming cryptographic handshake stream
+    crypto_stream: stream::Assembler,
```
djc: So I don't like having the `Assembler` here much. It seems like a layering violation in that `Assembler` should be an implementation detail of `Recv`. Also, don't we actually need more-or-less a full `(Send, Recv)` here? That would also cover the separate `crypto_offset` below. The crypto stream still has to deal with ~all of the issues that another stream would have to deal with, right?

quinn-proto/src/connection.rs
```diff
@@ -234,7 +228,7 @@ impl Connection {
             largest_sent_before_rto: 0,
             time_of_last_sent_retransmittable_packet: 0,
             time_of_last_sent_handshake_packet: 0,
-            largest_sent_packet: 0,
+            largest_sent_packet: !0, // Will wrap around to 0 on first transmit
```
djc: I find this a bit ugly. Could we do `next_packet_number: 0` instead, or do we need the largest sent packet more often than the next packet number?

quinn-proto/src/connection.rs
```diff
@@ -1041,13 +1016,11 @@ impl Connection {
                             }
                             match frame {
                                 Frame::Padding => {}
-                                Frame::Stream(
-                                    frame @ frame::Stream {
-                                        id: StreamId(0), ..
-                                    },
-                                ) => self.read_tls(&frame),
+                                Frame::Crypto(frame) => {
+                                    self.read_tls(&frame);
```
djc: I think these need to be assembled first, right?

quinn-proto/src/connection.rs
```diff
@@ -1101,10 +1074,9 @@ impl Connection {
                                 self.handshake_cleanup(&ctx.config);
                                 let mut msgs = Vec::new();
                                 self.tls.write_tls(&mut msgs).unwrap();
-                                if self.side == Side::Client {
-                                    self.transmit_handshake(&msgs);
-                                } else {
-                                    self.transmit(StreamId(0), msgs.into());
+                                self.transmit_handshake(&msgs);
+                                if self.side == Side::Server {
+                                    self.awaiting_handshake = false;
```
djc: Why do we actually need `awaiting_handshake`? Doesn't that somehow duplicate other bits of state we have in the `Connection` (like, perhaps, the actual `state`)?

quinn-proto/src/connection.rs
```diff
@@ -1252,7 +1224,10 @@ impl Connection {
                         return Err(e.into());
                     }
                 };
-                self.drive_tls(ctx)?;
+                if let Err(e) = self.drive_tls(ctx) {
```
djc: This change shouldn't be necessary now, right?

quinn-proto/src/connection.rs
```diff
@@ -959,60 +946,59 @@ impl Connection {
...
-                            );
-                            mem::replace(self, new);
-                            self.transmit_handshake(&outgoing);
-                            Ok(State::Handshake(state::Handshake {
-                                clienthello_packet: state.clienthello_packet,
-                                rem_cid_set: state.rem_cid_set,
-                                token: Some(packet.payload),
-                            }))
+                        if self.orig_rem_cid.is_some() || orig_dst_cid != self.rem_cid {
+                            // A client MUST accept and process at most one Retry packet for each connection attempt, and
```
djc: Nit: line wrapping, please.

quinn-proto/src/connection.rs
```diff
@@ -1228,14 +1214,17 @@ impl Connection {
...
-                let closed = match self.process_payload(ctx, now, number, packet.payload.into()) {
-                    Ok(closed) => closed,
-                    Err(e) => {
-                        ctx.events
-                            .push_back((self.handle, Event::ConnectionLost { reason: e.into() }));
-                        return Err(e.into());
-                    }
-                };
+                let closed =
+                    match self.process_payload(ctx, now, number.unwrap(), packet.payload.into()) {
```
djc: I think we can just propagate the error here? In fact, I think that was already on master?

quinn-proto/src/endpoint.rs
```diff
@@ -155,36 +168,32 @@ pub struct Context {
...
-    /// Cryptographic key used to ensure integrity of data included in handshake cookies.
-    ///
-    /// Initialize with random bytes.
-    pub cookie: [u8; 64],
-    /// Cryptographic key used to send authenticated connection resets to clients who were
-    /// communicating with a previous instance of tihs endpoint.
-    ///
-    /// Initialize with random bytes.
+    /// Cryptographic key used to authenticate data included in handshake tokens.
+    pub token: TokenKey,
```
djc: Let's not call this `token`, but something more descriptive (`token_key`)?

quinn-proto/src/endpoint.rs
```diff
@@ -155,36 +168,32 @@ pub struct Context {
...
     pub dirty_conns: FnvHashSet<ConnectionHandle>,
     pub readable_conns: FnvHashSet<ConnectionHandle>,
-    pub listen_keys: Option<ListenKeys>,
 }
 
 /// Information that should be preserved between restarts for server endpoints.
 ///
 /// Keeping this around allows better behavior by clients that communicated with a previous
 /// instance of the same endpoint.
 pub struct ListenKeys {
```
djc: I think the design here ends up a little weird. We have `Config::use_stateless_retry` which switches whether we require stateless retries or not and we have `ListenKeys` which is required to be `Some(key)` if you want your endpoint to act as a server but whose key may go unused if `use_stateless_retry` is `false`. Maybe make `stateless_retry_key` an `Option<>` in `Config` which handles both in one go?

djc: And rename `ListenKey` to `StatelessRetryKey` in the process?

quinn-proto/src/connection.rs
Ralith: This is called when the state is already `Closed` (or would be, if it wasn't temporarily removed from the `Option`). Checking `crypto.is_some()` therefore helps ensure we generate a packet that the makes sense to the peer.

quinn-proto/src/frame.rs
Ralith: I went ahead and just added the check. Good catch!

quinn-proto/src/packet.rs
Ralith: I added tests for the bugs I encountered and then some; no gaps jump out at me. Note that the packet number encoding is changing again slightly in draft 17 due to the length tag moving into the flags field.

quinn-proto/src/connection.rs
Ralith: The crypto stream is not subject to finishing, resets, stops, or flow control, and does not benefit from unordered reads. As such, everything but a stream assembler and a send offset (i.e. the vast majority of the `Send` and `Recv` structs) would be vestigal.

quinn-proto/src/connection.rs
Ralith: Not counting logging, `largest_sent_packet` is referenced 3 times in the recovery logic, and helps us stay close to the draft's pseudocode. I agree this is a bit ugly, though; maybe emulate the pseudocode variable with a trivial method?

quinn-proto/src/connection.rs
Ralith: `Connection::read_tls` performs assembly.

quinn-proto/src/connection.rs
Ralith: It's a hack to partially mitigate the brokenness of the draft-11 handshake logic, specifically relating to race conditions around encryption level changes. Will be removed as part of the transition to the new handshake logic.

quinn-proto/src/endpoint.rs
Ralith: If we want to support emitting `NEW_TOKEN` frames, then doing so could reasonably use this key whether or not stateless retries are enabled. Do you have any insight as to how likely we are to want that? If we don't think we'll want it for the forseeable future, I'm inclined to remove `Config::use_stateless_retry` and make the `ListenKeys` field `stateless_retry_key: Option<StatelessRetryKey>`, since its presence is only meaningful to a server.

Ralith:
Added a fix to #101 that I already had mostly written en route to the new handshake logic.

quinn-proto/src/endpoint.rs
Ralith: Maybe we should rename `ListenKeys` to `ListenConfig` or `ServerConfig` and move other server-specific members of `Config` (e.g. `accept_buffer`, `tls_server_config`, `retry_token_lifetime`) there?

quinn-proto/src/connection.rs
djc: Ah, that makes sense then.

quinn-proto/src/connection.rs
djc: Alright, in that case it would be overkill.

quinn-proto/src/connection.rs
djc: Yeah, trivial method sounds like a nice solution.

quinn-proto/src/connection.rs
djc: Ok!

quinn-proto/src/packet.rs
djc: Sounds good.

quinn-proto/src/endpoint.rs
djc: Yeah, doing a `ServerConfig` sounds like the right direction, and we can then use further `Option`s inside that to enable various features.

quinn/src/lib.rs
```diff
@@ -256,29 +258,20 @@ impl<'a> EndpointBuilder<'a> {
...
-    }
-
-    /// Accept incoming connections.
-    pub fn listen(&mut self, config: ServerConfig) -> &mut Self {
-        self.server_config = Some(config);
-        self
+    /// Build an endpoint bound to an unspecified free port.
+    ///
+    /// A good default behavior for clients.
+    pub fn build(self) -> Result<(Endpoint, Driver, Incoming), Error> {
```
djc: So I think this means clients using this API won't work on IPv4, right? That seems a little unfriendly.

Ralith: On Linux, IPv6 sockets are dual-stack by default, which means this should Just Work. Unfortunately it appears that [Windows differs here](https://docs.microsoft.com/en-us/windows/desktop/winsock/dual-stack-sockets). OpenBSD does not support dual-stack mode at all, and I'm not sure what macos does.

Forcing IPv4 also doesn't seem ideal; I think the best compromise would be to find a portable way to request dual-stack mode and let OpenBSD users run two endpoints if they really need to. We could probably support associating multiple sockets with a single endpoint if absolutely necessary, though.

djc: Can you write up a unit test that validates this? We can then at least know whether it works correctly on Linux, I can test on macOS and at some point we could bring up CI for Windows.

(Maybe this should be spun off into a separate issue, don't want to derail this PR too much.)

djc:
LGTM. Let's land this thing!

PR #103, opened 2018-11-29
WIP: clippy cleanup

imp:
Clippy has quite a lot to say about the code base.
This intended to be a gradual cleanup.
Feel free to reject if there are other ideas or policies on clippy, or pickup up commits incrementally if deemed useful.
I'll be adding more work here as time permits.

### DIFF
--- a/quinn-proto/src/stream.rs
+++ b/quinn-proto/src/stream.rs
@@ -238,7 +238,7 @@ impl Recv {
 
     pub fn buffer(&mut self, data: Bytes, offset: u64) {
         // TODO: Dedup
-        if data.len() == 0 {
+        if data.is_empty() {
             return;
         }
         self.buffered.push_back((data, offset));
--- a/quinn-proto/src/stream.rs
+++ b/quinn-proto/src/stream.rs
@@ -408,12 +408,12 @@ mod test {
     fn assemble_ordered() {
         let mut x = Assembler::new();
         assert_matches!(x.next(), None);
-        x.insert(0, (&b"123"[..]).into());
+        x.insert(0, &b"123"[..]);
         assert_matches!(x.next(), Some(ref y) if &y[..] == b"123");
-        x.insert(3, (&b"456"[..]).into());
+        x.insert(3, &b"456"[..]);
         assert_matches!(x.next(), Some(ref y) if &y[..] == b"456");
-        x.insert(6, (&b"789"[..]).into());
-        x.insert(9, (&b"10"[..]).into());
+        x.insert(6, &b"789"[..]);
+        x.insert(9, &b"10"[..]);
         assert_matches!(x.next(), Some(ref y) if &y[..] == b"78910");
         assert_matches!(x.next(), None);
     }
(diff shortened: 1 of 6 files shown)


djc:
@imp thanks for the PR! I still had a branch with some clippy-suggested fixes sitting in my local clone, so I've pushed that out to #104 to prevent duplicate work. If you want, you can cherry-pick my commit into your branch?

imp:
@djc 👍 Sure thing!

djc:
Also, I have run clippy on the code base a few times and cleaned stuff up. However, in doing so I've mostly ignored `clippy::cast_lossless` for now, as I'm not quite convinced of the cost/benefit there. I also feel `clippy::needless_pass_by_value` often has false positives in the sense that we sometimes want to pass by value conceptually even if the callee doesn't technically need ownership. `too_many_arguments` and `cyclomatic_complexity` I've ignored for now as we work out better APIs and merge/split functions to improve the internal design. `option_option` does need work, but definitely needs @Ralith's input for better-defined enums.

I think that covers everything I know about but have not fixed (or postponed) so far.

imp:
Quite agree with your points. Let me see if I can come up with anything simple and non-controversial.

djc:
Perhaps we should also add local `allow()` for the `needless_pass_by_value`, that would be a good addition IMO.

quinn-proto/src/stream.rs
```diff
@@ -91,12 +91,11 @@ impl Send {
             }
             | SendState::ResetRecvd {
                 ref mut stop_reason,
-            } => match stop_reason.take() {
-                Some(error_code) => {
+            } => {
+                if let Some(error_code) = stop_reason.take() {
```
djc: Does `cargo fmt` mandate the surrounding braces here? It seems like it could do without them.

imp: Yep, I was surprised too, but rustfmt insisted extra curly braces are required. Go figure.

imp:
>  However, in doing so I've mostly ignored `clippy::cast_lossless` for now, as I'm not quite convinced of the cost/benefit there.

I personally stopped using ` as ` conversion in my own code in favor of `uXX::from()`.
The latter is essentially the former:
```rust
impl From<u8> for u64 {
    #[inline]
    fn from(smalr: u8) -> u64 {
        small as u64
    }
}
```
However, it is a matter of policy. If ` as ` conversion if preferred though, I'd suggest setting
 `#![allow(clippy::cast_lossless)]` at the top level.
Thoughts?

djc:
Can you fix up my formatting bug in the initial commit? That way CI will pass.

imp:
> Can you fix up my formatting bug in the initial commit? That way CI will pass.

Rebased on latest `master`, which eliminated the offending code.

imp:
Commits ed66828 and 6817993 are not clippy-related.
First a style cleanup, which is admittedly a matter of personal preferences. If this is acceptable I suggest conversion throughout the code base.
Second is a usability improvement, prompted by the follow-up commit.

djc:
I'm not convinced about adding `is_client()` and `is_server()`. It is slightly more concise, but at the expense of adding boilerplatey functions to the enum's impl block.

Collapsing the `if` can be done even without the `is_client()` and `is_server()` change, I assume.

imp:
@djc Correct, collapsing `if` does not require `is_client()` change.

I'd personally prefer concise over the slim impl block, but that's ultimately your call.

djc:
@Ralith do you have an opinion either way on `is_client()`/`is_server()`?

Ralith:
I'm okay with it; I'm generally on board with going to small amounts of extra effort to make downstream code read better.

PR #104, opened 2018-11-29
Fix code style as suggested by clippy

djc:


djc:
This is now part of #103.

PR #106, opened 2018-11-30
WIP: Use the correct label for updating traffic secrets

stammw:
As discussed in #94 : [here](https://github.com/djc/quinn/pull/94#issuecomment-442721019)

This comes along with unit tests for 1-RTT key generation.

I'd like to write another test to be sure Crypto::update() will generate valid keys, using `encrypt` and `decrypt` to assert that. But I feel like I'm going to need writing it in `tests.rs`, as this will need a `TlsSession` to start with. So I'm just opening this to check if I'm going to the right direction ?


stammw:
Thanks !

The test I had in mind is more about functional testing: 

1. generate server and client 1-RTT keys 
2. check they are able to encrypt / decrypt each other packets
3. update keys for both of them
4. check again

Do you think that would be pertinent ?

stammw:
I forgot to mention: values for `ONERTT_LOCAL_SECRET` and `ONERTT_REMOTE_SECRET` are actually from the values built by the example programs.

Ralith:
> functional testing

Sounds like a good idea to me!

Ralith:
Initial investigation confirms that both ends of the failing test are computing the same keys; will keep digging.

Ralith:
I found the problem: the receiving endpoint is decrypting the *packet number* using the previous set of keys, before even checking the key phase. This produces a garbage result, which leads to a garbage IV being computed, which makes the packet undecryptable. We need to check the key phase bit and construct a candidate set of updated keys before decrypting the packet number.

This will likely require some rearranging of code, since packet number decryption currently takes place in `Connection::handle_decode`, whereas the rest of the packet--which must be decrypted using the new keys before we can deem the key update valid--is decrypted in `Connection::decrypt_packet`. `handle_decode` calls `handle_packet` calls `decrypt_packet` and the latter two functions are not called anywhere else, so the impact should be limited. Perhaps merge `handle_decode` with `handle_packet`, then modify `decrypt_packet` to accept an `Option<&Crypto>` for attempting key updates?

djc:
I don't much like the idea of merging `handle_packet()` into `handle_decode()`. I think it could make sense to inline the `decrypt_packet()` logic into the decode handling code, though, perhaps moving parts of it into `PartialDecode::finish()`?

djc:
BTW, really great to have test coverage for this so we found this bug now instead of in a later stage!

quinn-proto/src/connection.rs
```diff
@@ -765,13 +765,32 @@ impl Connection {
...
+                self.key_phase = !self.key_phase;
             } else {
-                &self.crypto.as_ref().unwrap()
-            };
-            partial_decode.finish(crypto.pn_decrypt_key())
+                warn!(self.log, "unable to complete packet decoding with new key");
+                return None;
+            }
+            decoded
+        } else {
```
djc: Can we sort this simple case before the complex one? I think that would make it more readable.

quinn-proto/src/connection.rs
```diff
@@ -2032,46 +2051,33 @@ impl Connection {
...
+            _ => &self.crypto.as_ref().unwrap(),
+        };
+
+        crypto
+            .decrypt(number, &packet.header_data, &mut packet.payload)
+            .map_err(|()| None)?;
+        Ok(Some(number))
+    }
+
+    pub(crate) fn update_keys(&mut self) {
```
djc: This seems to be only used in the tests? If so, it probably should live in the `tests` module.

quinn-proto/src/connection.rs
```diff
@@ -750,12 +750,32 @@ impl Connection {
...
+                ..
+            )) = decoded
+            {
+                let decoded_number = number.expand(self.rx_packet + 1);
+                let old = mem::replace(self.crypto.as_mut().unwrap(), new);
+                self.prev_crypto = Some((decoded_number, old));
+                self.key_phase = !self.key_phase;
+                decoded
+            } else {
+                warn!(self.log, "unable to complete header decoding with new key");
```
Ralith: This should log the exact error (`match` instead of `if`) and be at debug level (I prefer to reserve `warn` for reports of events that are probably peer misbehavior or attack).

quinn-proto/src/connection.rs
```diff
@@ -750,12 +750,32 @@ impl Connection {
...
+                    header: Header::Short { number, .. },
+                    ..
+                },
+                ..
+            )) = decoded
+            {
+                let decoded_number = number.expand(self.rx_packet + 1);
+                let old = mem::replace(self.crypto.as_mut().unwrap(), new);
+                self.prev_crypto = Some((decoded_number, old));
+                self.key_phase = !self.key_phase;
```
Ralith: The key update must not be committed until we have successfully authenticated the packet's payload. Otherwise, connections can be trivially DoSd. One approach would be to pass an optional `&Crypto` to use in preference to the one from `self` down through `handle_packet` to `decrypt_packet`, but there might be a cleaner approach.

quinn-proto/src/connection.rs
```diff
@@ -2035,42 +2055,24 @@ impl Connection {
...
+            Header::Initial { number, .. } | Header::Long { number, .. } if handshake => number,
             _ => {
                 return Err(None);
             }
         };
         let number = number.expand(self.rx_packet + 1);
-        if key_phase != self.key_phase {
-            if number <= self.rx_packet {
-                // Illegal key update
-                return Err(Some(TransportError::PROTOCOL_VIOLATION));
```
Ralith: It looks like this mandatory error check was lost in the refactoring.

quinn-proto/src/connection.rs
```diff
@@ -750,12 +750,32 @@ impl Connection {
         remote: SocketAddrV6,
         partial_decode: PartialDecode,
     ) -> Option<BytesMut> {
-        let crypto = if partial_decode.is_handshake() {
-            &self.handshake_crypto
+        let result = if partial_decode.is_handshake() {
+            partial_decode.finish(self.handshake_crypto.pn_decrypt_key())
+        } else if partial_decode.key_phase() == self.key_phase {
+            partial_decode.finish(self.crypto.as_ref().unwrap().pn_decrypt_key())
```
Ralith: Could a misbehaving peer cause us to panic here by sending a non-handshake packet before `crypto` is populated?

quinn-proto/src/connection.rs
```diff
@@ -745,16 +745,52 @@ impl Connection {
...
+                partial_decode.finish(self.handshake_crypto.pn_decrypt_key()),
+                None,
+            )
+        } else if partial_decode.key_phase() == self.key_phase {
+            if let Some(ref crypto) = self.crypto.as_ref() {
+                (partial_decode.finish(&crypto.pn_decrypt_key()), None)
+            } else {
+                warn!(
+                    self.log,
+                    "recieved a non-hanshake packet before crypto establishment"
```
```suggestion
                    "recieved a non-hanshake packet before handshake completed"
```

quinn-proto/src/connection.rs
```diff
@@ -745,16 +745,52 @@ impl Connection {
...
+                Ok((
+                    Packet {
+                        header: Header::Short { number, .. },
+                        ..
+                    },
+                    ..
+                )) => {
+                    let decoded_number = number.expand(self.rx_packet + 1);
+                    if decoded_number <= self.rx_packet {
+                        warn!(self.log, "recieved a key update packet with invalid number");
```
```suggestion
                        warn!(self.log, "recieved an illegal key update");
```

quinn-proto/src/connection.rs
djc: (Note, you probably also want to replace `hanshake` with `handshake`.)

quinn-proto/src/connection.rs
```diff
@@ -1964,47 +1976,44 @@ impl Connection {
...
-            }
-            let new = self.crypto.as_mut().unwrap().update();
-            new.decrypt(number, &packet.header_data, &mut packet.payload)
-                .map_err(|()| None)?;
 
-            let old = mem::replace(self.crypto.as_mut().unwrap(), new);
+        if crypto_update.is_some() && number <= self.rx_packet {
+            warn!(self.log, "recieved an illegal key update");
+            return Err(Some(TransportError::PROTOCOL_VIOLATION));
+        }
```
djc: Sorry, one more thing: can you add an `else if` here that throws another `PROTOCOL_VIOLATION` if `handshake == true && crypto_update.is_some()`?

stammw: Yes, sure, good catch !

quinn-proto/src/tests.rs
```diff
@@ -719,3 +720,46 @@ fn stream_id_backpressure() {
...
+
+    assert_matches!(pair.server.poll(), Some((conn, Event::StreamReadable { stream, fresh: false })) if conn == server_conn && stream == s);
+    assert_matches!(pair.server.poll(), None);
+    assert_matches!(
+        pair.server.read_unordered(server_conn, s),
+        Ok((ref data, 6)) if data == MSG2
+    );
+}
+
+impl connection::Connection {
```
djc: Ah, sorry, and I mentioned to get rid of the `impl` block here, leaving just the `fn` itself.

stammw:
I guess it can now be merged.

quinn-proto/src/connection.rs
```diff
@@ -1964,47 +1976,45 @@ impl Connection {
...
+        assert!(!(handshake && self.crypto.is_some()));
```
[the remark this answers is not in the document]
stammw: ho, sorry I should have given this more thoughts.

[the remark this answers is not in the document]
stammw: Done

stammw:
I moved update_keys back to `impl Connection` as `Crypto` is now private. I should be useful for #121.

PR #107, opened 2018-12-01
Varint updates

imp:
While making myself acquainted with the code base (varint.rs in this case) I felt like experimenting with the implementation a bit. And then I added unit tests to make sure things are still ok. I believe the net effect is quite good, so I decided to submit it upstream.

Each commit is a separate chunk of changes and should be easy to review.
First three commits are unit tests, and the rest is my attempt to avoid extra computation and copying.

As always feel free to accept, partially accept or reject.

### DIFF
--- a/quinn-proto/src/varint.rs
+++ b/quinn-proto/src/varint.rs
@@ -22,6 +22,52 @@ const FOUR_OCTETS_MAX: u64 = 1_073_741_823;
 const EIGHT_OCTETS_MIN: u64 = FOUR_OCTETS_MAX + 1;
 const EIGHT_OCTETS_MAX: u64 = 4_611_686_018_427_387_903;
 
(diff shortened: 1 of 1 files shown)


imp:
Oh, forgot to mention - it uses features recently stabilized on nightly, but still not available on stable. So meanwhile just for review.

djc:
Thanks for your work. However, I don't think this is the right direction for this code. If anything, I believe we should make `VarInt` a newtype around `u64` that implements `Codec`.

Also, IMO your test macros are a bit too magic and I don't think it makes sense to keep a PR open for more than a few days or so waiting for Rust features to reach stable.

imp:
Completely agree on the new type. Re macros - they are _always_ magic, aren't they?

PR #112, opened 2018-12-08
Abstracted Context

Ralith:
Decouples `Connection` from the implementation details of `Endpoint` and reduces bookkeeping.


djc:
Nice improvements!

PR #113, opened 2018-12-08
Fix travis nightly rustfmt

Ralith:
Let's see if this helps.


Ralith:
On review, we weren't using rustfmt in the nightly build anyway for obvious reasons.

PR #105, opened 2018-11-29
Make the example documentation clearer

psiphi75:
Added comments on updating the certs/openssl.cnf for access across the network.
Changed the path of the certificates folder, since then it will just be copy and paste.
Changed the `.` to a `./` because when I copy-n-pasted I missed the `.` because it's not very visible.

PR #115, opened 2018-12-09
Update example documentation in Readme

psiphi75:
Make it clearer how to run the example across the network.  Supersedes PR #105.


djc:
Very nice, thanks for clarifying this!

PR #117, opened 2018-12-11
Implement explicit congestion notification

Ralith:
Omits the verification step to check for malfunctioning routers due to ambiguities in the current draft. Includes only the state machine logic changes; getting/setting ECN codepoints for real I/O is left for future work.

We could hold off on merging this until verification logic is clarified, but the worst-case scenario once real I/O is wired up is a malfunctioning router causing spurious congestion events to be registered and decreasing performance. If that becomes an issue in practice, we could disable outgoing ECN and still benefit the rest of the ecosystem with the ECN feedback these changes implement, which allow remote endpoints to leverage ECN when sending to us.


Ralith:
Mirja Kühlewind provided some insight on slack, allowing me to implement some reasonable baseline verification logic.

Ralith:
Addressed comments, and restructured the internals to be a bit clearer and closer to the intended behavior.

Ralith:
See https://github.com/quicwg/base-drafts/pull/2113 for ongoing discussion of clarified ECN requirements. I'm fairly confident that the logic I've implemented now is more or less correct (i.e. well-behaved under all conditions) even though it superficially differs from the currently proposed language.

Ralith:
The wording tweak PR was merged following the removal of the wording that superficially differed from this implementation, and I've rebased.

quinn-proto/src/connection.rs
```diff
@@ -118,12 +118,22 @@ pub struct Connection {
...
-    /// The largest packet number sent when QUIC detects a loss. When a larger packet is
-    /// acknowledged, QUIC exits recovery.
-    end_of_recovery: u64,
+    /// The time when QUIC first detects a loss, causing it to enter recovery. When a packet sent
+    /// after this time is acknowledged, QUIC exits recovery.
+    recovery_start_time: u64,
     /// Slow start threshold in bytes. When the congestion window is below ssthresh, the mode is
     /// slow start and the window grows by the number of bytes acknowledged.
     ssthresh: u64,
+    /// Explicit congestion notification (ECN) counters.
```
djc: Nit: this hunk is inconsistent about ending doc comments with a period. My preference is not to do so consistently, since these are often sentence fragments rather than full sentences (so the elaboration part of the comment -- after the first line + newline -- would still use periods).

quinn-proto/src/connection.rs
```diff
@@ -311,24 +325,26 @@ impl Connection {
 
     fn on_ack_received(&mut self, mux: &mut impl Multiplexer, now: u64, ack: frame::Ack) {
         trace!(self.log, "got ack"; "ranges" => ?ack.iter().collect::<Vec<_>>());
+
         let was_blocked = self.blocked();
```
djc: Nit: doesn't make much sense to have this on a line in its own block?

quinn-proto/src/connection.rs
```diff
@@ -337,6 +353,74 @@ impl Connection {
...
+                            debug!(
+                                self.log,
+                                "halting ECN due to verification failure: {error}",
+                                error = e
+                            );
+                            self.sending_ecn = false;
+                        }
+                        Ok(false) => {}
+                        Ok(true) => {
+                            // This should always be `Some(_)` because a new largest ack is by
```
djc: Nit: add `TODO`.

quinn-proto/src/connection.rs
```diff
@@ -337,6 +353,74 @@ impl Connection {
...
+
+        // Explicit congestion notification
+        if self.sending_ecn {
+            if let Some(ecn) = ack.ecn {
+                // We only examine ECN counters from ACKs that we are certain we received in transmit
+                // order, allowing us to compute an increase in ECN counts to compare against the number
+                // of newly acked packets that remains well-defined in the presence of arbitrary packet
+                // reordering.
+                if ack.largest > prev_largest {
+                    match self.detect_ecn(newly_acked.len() as u64, ecn) {
```
djc: Style: maybe extract this `match` from the deeply nested `if`-blocks prevent all the rightward drift?

quinn-proto/src/connection.rs
```diff
@@ -512,22 +597,27 @@ impl Connection {
...
-                self.ssthresh = self.congestion_window;
+            if lost_nonack {
+                self.congestion_event(now, largest_lost_time)
             }
         }
     }
 
-    fn in_recovery(&self, packet: u64) -> bool {
-        packet <= self.end_of_recovery
+    fn congestion_event(&mut self, now: u64, sent_time: u64) {
```
djc: Next time, please put the method extraction in a separate commit from the logic changes? Reviewing both at the same time is painful.

quinn-proto/src/connection.rs
```diff
@@ -118,12 +118,22 @@ pub struct Connection {
...
     congestion_window: u64,
-    /// The largest packet number sent when QUIC detects a loss. When a larger packet is
-    /// acknowledged, QUIC exits recovery.
-    end_of_recovery: u64,
+    /// The time when QUIC first detects a loss, causing it to enter recovery. When a packet sent
+    /// after this time is acknowledged, QUIC exits recovery.
+    recovery_start_time: u64,
     /// Slow start threshold in bytes. When the congestion window is below ssthresh, the mode is
     /// slow start and the window grows by the number of bytes acknowledged.
     ssthresh: u64,
```
djc: Off-topic nit: `ssthresh` as a name is pretty obscure (aside from the effect that the guidelines would probably spell it `ss_thresh`).

quinn-proto/src/connection.rs
```diff
@@ -587,22 +677,25 @@ impl Connection {
...
-            self.pending_acks.pop_min();
+        self.receiving_ecn |= ecn.is_some();
+        if let Some(x) = ecn {
+            self.ecn_counters += x;
         }
-        if packet > self.rx_packet {
-            self.rx_packet = packet;
-            self.rx_packet_time = now;
+
+        if let Some(packet) = packet {
```
djc: I think I prefer the previous style of returning early for `None` `packet`s. Also, this is another entirely non-functional change intermixed in a commit that otherwise very much isn't.

quinn-proto/src/connection.rs
```diff
@@ -311,24 +325,26 @@ impl Connection {
...
-            // Avoid DoS from unreasonably huge ack ranges
-            let packets = self
-                .sent_packets
-                .range(range)
-                .map(|(&n, _)| n)
-                .collect::<Vec<_>>();
-            for packet in packets {
-                self.on_packet_acked(packet);
-            }
+        // Avoid DoS from unreasonably huge ack ranges
```
djc: So is this also just a style change, or are there functional changes, too?

quinn-proto/src/connection.rs
```diff
@@ -1582,13 +1582,13 @@ impl Connection {
         }
 
         // CRYPTO
-        while buf.len() + 16 < max_size {
+        while buf.len() + 17 < max_size {
```
djc: I think this really wants to be a constant (probably in `packet`?). Maybe we can use it in other places, too.

quinn-proto/src/connection.rs
Ralith: This name is taken directly from the recovery draft; I favor doing that whenever possible for ease of cross-referencing.

djc: Ah, in that case it makes sense.

PR #119, opened 2018-12-11
Late reset

djc:


quinn-proto/src/connection.rs
```diff
@@ -2290,6 +2296,9 @@ pub enum ConnectionError {
     /// The peer has become unreachable.
     #[fail(display = "timed out")]
     TimedOut,
+    /// Maybe received a stateless reset.
+    #[fail(display = "maybe stateless reset (authentication failed or duplicate)")]
+    StatelessReset,
```
Ralith: This is exactly redundant to `Reset`. Note that stateless resets are not ambiguous.

quinn-proto/src/connection.rs
```diff
@@ -2298,28 +2307,6 @@ impl From<TransportError> for ConnectionError {
...
-                format!("peer detected an error: {}", reason),
-            ),
-            TransportError { error_code } => {
-                io::Error::new(io::ErrorKind::Other, format!("{}", error_code))
-            }
-            VersionMismatch => io::Error::new(io::ErrorKind::Other, "version mismatch"),
-        }
-    }
-}
-
```
Ralith: This was included as a convenience when interoperating with legacy code written for OS socket APIs. Are we sure we want to make that more difficult?

quinn-proto/src/connection.rs
```diff
@@ -837,6 +827,7 @@ impl Connection {
                     }
                     ConnectionError::TransportError { error_code } => State::closed(error_code),
                     ConnectionError::VersionMismatch => State::Draining,
+                    ConnectionError::StatelessReset => State::Drained,
```
Ralith: Instead of adding a new case, we should just get rid of the panic in the `Reset` case.

quinn-proto/src/connection.rs
```diff
@@ -807,14 +790,21 @@ impl Connection {
                             "discarding possible duplicate packet {packet}",
                             packet = number
                         );
+                        if self.check_for_stateless_reset(mux, &packet) {
+                            Err(ConnectionError::StatelessReset)
+                        } else {
+                            return;
+                        }
+                    } else {
                         return;
```
Ralith: This would cause all packets to be ignored. We should instead fall through to the block that calls `handle_connected_inner`.

quinn-proto/src/connection.rs
```diff
@@ -807,14 +790,21 @@ impl Connection {
...
                     }
+                } else {
+                    if !was_closed {
+                        self.on_packet_authenticated(mux, now, number);
+                    }
+                    let state = self.state.take().unwrap();
+                    self.handle_connected_inner(mux, now, number, packet, state)
```
Ralith: This block should be executed regardless of whether `number.is_some()`.

quinn-proto/src/connection.rs
```diff
@@ -742,7 +742,6 @@ impl Connection {
         &mut self,
         mux: &mut impl Multiplexer,
         now: u64,
-        remote: SocketAddrV6,
```
Ralith: I don't mind removing this for now, but I should note that we'll need it back eventually to support migration.

quinn-proto/src/connection.rs
djc: Removed it and used `Reset` instead.

quinn-proto/src/connection.rs
djc: Kept it and added a comment.

quinn-proto/src/connection.rs

quinn-proto/src/connection.rs
djc: I think the flow is correct now, please check?

quinn-proto/src/connection.rs
djc: Yeah, I just want it to be warning-clean...

quinn-proto/src/connection.rs
```diff
@@ -790,24 +776,41 @@ impl Connection {
...
+                } else {
+                    None
+                };
+
+                if let Some(number) = duplicate {
+                    warn!(
+                        self.log,
+                        "discarding possible duplicate packet {packet}",
+                        packet = number
+                    );
```
Ralith: Let's move this warning into the `else` case just below, since honoring a connection reset does not constitute discarding the packet.

quinn-proto/src/connection.rs
```diff
@@ -790,24 +776,41 @@ impl Connection {
...
-                            "discarding possible duplicate packet {packet}",
-                            packet = number
-                        );
+                        Some(number)
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                };
```
Ralith: `let duplicate = number.and_then(|n| if self.dedup.insert(n) { Some(n) } else { None };` perhaps? I can't wait for try blocks...

quinn-proto/src/connection.rs

quinn-proto/src/connection.rs
djc: Yup, nice one.

djc:
Rebased on top of master.

quinn-proto/src/connection.rs
```diff
@@ -807,24 +793,34 @@ impl Connection {
...
+                }
             }
             Ok(number) => {
-                if let Some(number) = number {
-                    if self.dedup.insert(number) {
+                let duplicate =
+                    number.and_then(|n| if self.dedup.insert(n) { Some(n) } else { None });
+
+                if let Some(number) = duplicate {
+                    if self.check_for_stateless_reset(mux, &packet) {
```
Ralith: Wait, this can't be right. Stateless resets are never encrypted, because the peer cannot possibly have the connection's keys if it's sending one. I guess we need to `check_for_stateless_reset` *before* decrypting, but only inspect its return value afterwards. It's confusing that the draft allows for the possibility that a stateless reset message just coincidentally happens to be decryptable, so maybe I'm missing something here...

djc: I'm not sure why you think it's wrong? We try to decrypt it, which makes sense to me, then when that fails without a transport error we check for the stateless reset.

Arguably the case where the decrypt succeeds but then turns out to be a duplicate packet is very unlikely but technically possible (I think), and we handle that as well.

Ralith: To elaborate on my unease here: a genuine duplicate packet can presumably happen due to network hardware glitchiness on the path between two hosts, while a stateless reset packet that just happens to pass strong cryptographic authentication seems like one of those not-in-the-lifetime-of-the-universe things.

djc:
Here's another take along the lines you suggested, which I think looks nicer anyway.

quinn-proto/src/connection.rs
```diff
@@ -778,51 +778,62 @@ impl Connection {
...
         };
 
+        if stateless_reset && !self.state.is_drained() {
+            if let Err(ConnectionError::Reset) = result {
+                debug!(self.log, "got stateless reset");
+                mux.timer_stop(Timer::LossDetection);
+                mux.timer_stop(Timer::Close);
+                mux.timer_stop(Timer::Idle);
+            }
+        }
```
Ralith: Is checking the value of `stateless_reset` here redundant to the pattern-match? I think we could also move the body of this into the match statement below, retaining the `!is_drained()` guard.

PR #120, opened 2018-12-11
Move Connection getters into Connection type

djc:


quinn-proto/src/endpoint.rs
```diff
@@ -829,55 +829,13 @@ impl Endpoint {
...
-    /// Whether a previous session was successfully resumed by `conn`.
-    pub fn get_session_resumed(&self, _: ConnectionHandle) -> bool {
-        false // TODO: fixme?
-    }
-
     pub fn accept(&mut self) -> Option<ConnectionHandle> {
         self.ctx.incoming.pop_front()
     }
+
+    pub fn connection(&self, handle: ConnectionHandle) -> &Connection {
```
Ralith: I suspect this will produce a private-in-public error, at which point we'll need to either expose `Connection` in all its messy glory or spend some time getting rid of all the unnecessarily exposed internal fields/methods.

djc: Not sure what you mean by this? `Connection` is still private to the `quinn-proto` crate.

PR #122, opened 2018-12-12
More iteration on Connection API

djc:


quinn-proto/src/connection.rs
```diff
@@ -955,10 +948,11 @@ impl Connection {
                             trace!(self.log, "handshake ongoing");
                             self.handshake_cleanup(mux);
                             self.write_tls();
-                            return Ok(State::Handshake(state::Handshake {
+                            self.state = State::Handshake(state::Handshake {
                                 token: None,
                                 ..state
```
Ralith: Instead of cloning `state` above, could we do something like `if let State::Handshake(ref mut x) = self.state { x.token = None; } else { unreachable!(); }` here?

quinn-proto/src/connection.rs
```diff
@@ -1086,26 +1081,28 @@ impl Connection {
...
                         _ => {}
                     }
                 }
-                Ok(State::Closed(state))
+                Ok(())
             }
-            State::Draining => Ok(State::Draining),
-            State::Drained => Ok(State::Drained),
+            State::Draining => Ok(()),
+            State::Drained => Ok(()),
```
Ralith: These could be folded together into `State::Draining | State::Drained => Ok(())`

quinn-proto/src/connection.rs
```diff
@@ -1767,13 +1764,13 @@ impl Connection {
...
-        self.state = Some(match self.state.take().unwrap() {
+        self.state = match &self.state {
             State::Handshake(_) => State::Closed(state::Closed { reason }),
             State::Established => State::Closed(state::Closed { reason }),
-            State::Closed(x) => State::Closed(x),
+            State::Closed(x) => State::Closed(x.clone()),
             State::Draining => State::Draining,
             State::Drained => unreachable!(),
-        });
+        };
```
Ralith: This could be made more concise and clone-free by moving the assignment into the `Handshake`/`Established` cases and removing the no-op cases entirely.

quinn-proto/src/connection.rs
```diff
@@ -2399,13 +2404,6 @@ impl State {
         })
     }
 
-    pub fn is_handshake(&self) -> bool {
-        match *self {
-            State::Handshake(_) => true,
-            _ => false,
-        }
-    }
-
```
Ralith: Unlike the new `Connection::is_handshaking` method, this can be called when a mutable borrow to some other part of the `Connection` is live. For that reason it might be nice to retain, with the public function being a simple wrapper. On the other hand, we can always bring it back if/when we eventually want to do that, so I don't feel strongly.

Ralith:
Seems like all we need to make `state` private is an `is_drained` and moving the heart of the `timeout` implementation into `Connection`.

quinn-proto/src/connection.rs
djc: No, because `ref mut state` overlaps with the `&mut self` which is used all over the place -- this was where I ran into trouble in the first place.

quinn-proto/src/connection.rs
```diff
@@ -813,43 +809,37 @@ impl Connection {
...
-        if !was_closed && state.is_closed() {
+        if !was_closed && self.state.is_closed() {
             self.close_common(mux, now);
         }
 
         // Transmit CONNECTION_CLOSE if necessary
-        if let State::Closed(ref state) = state {
-            self.transmit_close(mux, now, &state.reason);
+        if let State::Closed(ref state) = self.state {
+            self.transmit_close(mux, now, &state.clone().reason);
```
Ralith: If we modify `Connection::close` to update the state *before* calling `transmit_close`, we could in turn get rid of the argument to `transmit_close` and have it read from `self.state` directly, avoiding this clone.

quinn-proto/src/connection.rs
Ralith: Ah, I see. We could add a similar assignment block outside of the if statement to set `rem_cid_set`, but maybe it's not worth the effort.

quinn-proto/src/connection.rs
djc: Yup, good idea.

quinn-proto/src/connection.rs
djc: From the outside it doesn't matter; from the inside I don't think calling `is_handshaking()` will cause more issues? It doesn't take a borrow, just synthesizes a `bool` out of it.

I think I'm angling to remove the final bits of state from `State` variants, too, and just inline them into `Connection` -- though I have a feeling `rem_cid_set` is probably redundant.

quinn-proto/src/connection.rs
djc: Nice.

djc:
Want to have another look?

quinn-proto/src/connection.rs
Ralith: Note that `rem_cid` is always set to *something*; that flag identifies whether it's been set to a value determined by the peer, versus the initial random value.

Ralith: The `&self` borrow would render `is_handshaking` ineligible to called while a mutable borrow of one of `Connection`'s fields is still active. This is not currently needed, but in general I like to limit the scope of method borrows to help keep borrowck at bay; for example, this is the sole function of the `Streams` struct.

djc: Yeah, I know. But I feel like we may be able to derive that from other bits of state that we have in the relevant places.

djc: Alright, I've kept `State::is_handshake()` and made it private.

quinn-proto/src/connection.rs
djc: Let's do that as a follow-up, since it is at least moderately finicky to get right.

PR #123, opened 2018-12-13
Generalize support different crypto levels

djc:


Ralith:
The two handshake-related crypto levels both operate in independent packet number spaces ([§12.3](https://tools.ietf.org/html/draft-ietf-quic-transport-16#section-12.3), which are in turn independent from the application data packet number space, so performing lookup based on packet number isn't appropriate for those, only for key updates within the application data space. Is there a need for us to support having more than two application data keys concurrently? Per @stammw's findings in https://github.com/djc/quinn/issues/121, key updates should be infrequent.

djc:
I'm aware of all that, but I still think this is a good data model and API to support what we need.

We currently have at most three `Crypto` instances, and in order to make incremental changes I think it makes sense to allow 4 for now so we can add the new level independently of making sure we discard them in a timely fashion. We can trivially reduce the deque capacity to 2 once we have code in place to discard `Crypto`s as soon as necessary.

We can also easily add an `assert!()` in `find_crypto()` to enforce that `start` is always 0 for non-application data crypto spaces.

Ralith:
Ah, somehow I missed the `CryptoLevel` field in the deque.

quinn-proto/src/connection.rs
```diff
@@ -744,13 +746,13 @@ impl Connection {
...
         } else if partial_decode.key_phase() != self.key_phase {
-            new_crypto = Some(self.crypto.as_ref().unwrap().update());
+            new_crypto = Some(self.cryptos.back().unwrap().2.update());
             new_crypto.as_ref().unwrap()
         } else {
-            if let Some(ref crypto) = self.crypto {
-                crypto
+            if let Some(crypto) = self.cryptos.back() {
+                &crypto.2
             } else {
```
Ralith: `self.cryptos` is never empty, so this branch will never be executed.

quinn-proto/src/connection.rs
```diff
@@ -2073,6 +2081,28 @@ impl Connection {
...
 
+    fn push_crypto(&mut self, crypto: Crypto, number: u64) {
+        let old_idx = self
+            .cryptos
+            .iter()
+            .position(|(level, _, _)| level == &CryptoLevel::OneRtt)
+            .unwrap();
+        self.cryptos.remove(old_idx);
+        self.cryptos
+            .push_back((CryptoLevel::OneRtt, number, crypto));
```
Ralith: If this would cause three sets of 1-RTT keys to be known, we should drop the oldest as it will never be used (since `key_phase` can only address two sets of keys) and to prevent unbounded memory use wrt. time.

quinn-proto/src/connection.rs
```diff
@@ -2073,6 +2081,28 @@ impl Connection {
...
+        self.key_phase = !self.key_phase;
+    }
+
+    fn find_crypto(&self, level: CryptoLevel, number: u64) -> Option<&Crypto> {
+        self.cryptos.iter().rev().find_map(|(l, s, crypto)| {
```
Ralith: Nice touch scanning from the end!

quinn-proto/src/connection.rs
```diff
@@ -41,8 +41,7 @@ pub struct Connection {
     rx_packet: u64,
     /// Time at which the above was received
     rx_packet_time: u64,
-    crypto: Option<Crypto>,
-    prev_crypto: Option<(u64, Crypto)>,
+    cryptos: VecDeque<(CryptoLevel, u64, Crypto)>,
```
Ralith: Code referencing this would be more readable if its element type was a struct with helpfully-named fields.

quinn-proto/src/connection.rs
djc: Hmm, so here's what I'm wondering about: we create new packet number encryption keys for key phase changes, but if packets arrive out of order (which we can discern only based on the packet number), we will use the incorrect keys to decrypt the packet number. Does that mean packet number protection keys don't get updated alongside the other keys? Or am I missing some other solution to this problem?

quinn-proto/src/connection.rs
djc: Actually, I think this code has the opposite problem from what you suggest. It always discards the oldest `OneRtt` crypto when it pushes back an update, which I think means it will never retain more than one encryption level. (We should probably add a test for that case.)

quinn-proto/src/connection.rs
djc: Yeah, totally fair -- I was thinking about it as I was writing it. Fixed this. Let me know if you think of a better name than `CryptoSpace`.

quinn-proto/src/connection.rs
Ralith: That's been on my mind too since reviewing this. The draft doesn't seem to mention the problem, but my reading is that the PNE key does get updated, and quant and ngtcp2 (for example) both seem to update it. One solution might be: if a packet with a mismatched key phase fails authentication with key set n+1, retry the whole procedure including packet number decryption using key set n-1.

Ralith: Word on slack is that PNE keys will not be affected by key update in draft 17.

quinn-proto/src/connection.rs
Ralith: That's what I get for staying up late doing code review...

djc: Well, you triggered the finding of a bug after all -- not such a bad job.

quinn-proto/src/connection.rs
```diff
@@ -2078,6 +2090,37 @@ impl Connection {
         }
     }
 
+    fn push_crypto(&mut self, crypto: Crypto, number: u64) {
```
Ralith: Could this be renamed to reflect that it's only used during 1RTT key updates? Perhaps rename the current `update_keys` to `initiate_key_update` and use `update_keys` for this?

PR #124, opened 2018-12-14
Clientside stateless resets

Ralith:
This adds support for clients performing stateless resets by transmitting `NEW_CONNECTION_ID` immediately after handshake to establish a reset token, and fixes several bugs encountered along the way.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -1567,7 +1567,7 @@ impl Connection {
                         self.io.retired_cids.push(old);
                     }
                 }
-                Frame::NewConnectionId { .. } => {
+                Frame::NewConnectionId(frame) => {
                     if self.rem_cid.is_empty() {
                         debug!(
                             self.log,
@@ -1575,7 +1575,14 @@ impl Connection {
                         );
                         return Err(TransportError::PROTOCOL_VIOLATION);
                     }
-                    trace!(self.log, "ignoring NEW_CONNECTION_ID (unimplemented)");
+                    if self.params.stateless_reset_token.is_none() {
+                        // We're a server using the initial remote CID for the client, so let's
+                        // switch immediately to enable clientside stateless resets.
+                        debug_assert!(self.side.is_server());
+                        self.update_rem_cid(frame);
+                    } else {
+                        trace!(self.log, "ignoring NEW_CONNECTION_ID (unimplemented)");
+                    }
                 }
                 Frame::NewToken { .. } => {
                     trace!(self.log, "got new token");
@@ -1584,6 +1591,17 @@ impl Connection {
             }
         }
         Ok(false)
+    }
+
+    fn update_rem_cid(&mut self, new: frame::NewConnectionId) {
+        trace!(
+            self.log,
+            "switching to remote CID {sequence}: {connection_id}",
+            sequence = new.sequence,
+            connection_id = new.id
+        );
+        self.rem_cid = new.id;
+        self.params.stateless_reset_token = Some(new.reset_token);
     }
 
     fn next_packet(&mut self, now: u64) -> Option<Box<[u8]>> {
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -21,7 +21,7 @@ use crate::stream::{self, ReadError, Stream, WriteError};
 use crate::transport_parameters::{self, TransportParameters};
 use crate::{
     frame, Directionality, Frame, Side, StreamId, TransportError, MIN_INITIAL_SIZE, MIN_MTU,
-    VERSION,
+    RESET_TOKEN_SIZE, VERSION,
 };
 use rustls::internal::msgs::enums::AlertDescription;
 
@@ -961,7 +961,8 @@ impl Connection {
         let was_closed = self.state.is_closed();
 
         let stateless_reset = self.params.stateless_reset_token.map_or(false, |token| {
-            packet.payload.len() >= 16 && packet.payload[packet.payload.len() - 16..] == token
+            packet.payload.len() >= RESET_TOKEN_SIZE
+                && packet.payload[packet.payload.len() - RESET_TOKEN_SIZE..] == token
         });
 
         let result = match self.decrypt_packet(was_handshake, &mut packet) {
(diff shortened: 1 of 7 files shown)


Ralith:
Managed to factor away the monolithic commit after all.

Ralith:
Rebased. 72d5a80736c10ae944e4a8a186b990731f5e08a7 introduces a shared mutable hash table that provides the synchronization formerly implemented by indirection through `Multiplexer`, in a manner that supports eventual `Send` connection tasks. Synchronization through shared memory is useful here because connections may remove local connection IDs at any time due to remote prompting and may in principle issue new connection IDs at any time, while the dispatch table used by the `Endpoint` must be kept up to date. Overhead is expected to be low as both of these events should be rare.

An alternative option that avoids the `Mutex` would be to have all issuings of new connection IDs initiated by the `Endpoint`, which would be sufficient for the main usecases of client-side stateless resets and of avoiding linkability across migrations. Retiring old CIDs could be done via I/O messages, similar to timer updates.

Ralith:
> Is there value in keeping the other fields outside the `Mutex`?

On review, there's no functional benefit to this. However, I do think it's nice to limit the scope of interior mutability of any kind as narrowly as possible.

Ralith:
> Maybe it's time to bite the bullet and switch to channels for Endpoint <-> Connection communication?

This is onerous if `Connection`s issue new CIDs on their own initiative, because it would involve requesting one and waiting for a new response. I can draft the alternative approach I outlined in a prior comment instead, removing that capability in favor of preallocating (and perhaps eventually dynamically issuing) additional CIDs in the `Endpoint`, with retirement mediated by `Io` messages.

In general, I don't think there will be need to make explicit use of channels inside `Connection` when we can expose methods that consume/yield events, which can be wired up however the caller likes.

djc:
In my mind connections would be fully responsible for creating/issuing/retring CIDs, merely informing their `Endpoint` of (changes to) the current set of allowed CIDs that may reference that `Connection`.

I don't really see how we can avoid channels when we want to run both `Endpoint`s and `Connection`s in separate tasks?

Ralith:
> In my mind connections would be fully responsible for creating/issuing/retring CIDs, merely informing their Endpoint of (changes to) the current set of allowed CIDs that may reference that Connection.

On review, the draft recommends that peers be supplied with at least eight CIDs at all times, so I think this can be much simpler: the `Endpoint` supply every new connection with eight CIDs to issue when established, and whenever `conn.poll_io() == Io::RetireCid(_)`, the `Endpoint` will supply a new CID to replace the retired one. No further coordination should be necessary.

> I don't really see how we can avoid channels when we want to run both Endpoints and Connections in separate tasks?

Channels are necessary for that, but the `Connection` doesn't have to know about them. My thinking is that `Connection` retains its pure, synchronous interface, and may (or may not) be driven by asynchronous communication with an `Endpoint` as needed. This also makes it easy to start testing `Connection` on its own, which seems nice.

djc:
Okay, that sounds pretty good.

Ralith:
Drafted.

quinn-proto/src/endpoint.rs
```diff
@@ -379,12 +379,21 @@ impl Endpoint {
...
+            let remaining = self.connections[conn_id.0].handle_decode(now, ecn, partial_decode);
+            if was_handshake && !self.connections[conn_id.0].is_handshaking() {
+                // Newly established connection
+                if self.connections[conn_id.0].side().is_server() {
+                    self.incoming_handshakes -= 1;
+                    self.ctx.incoming.push_back(conn_id);
+                }
+                if self.config.local_cid_len != 0 {
+                    // endpoints SHOULD provide and maintain at least eight connection IDs
+                    for _ in 0..8 {
```
djc: Should we somehow document what this `8` is?

quinn-proto/src/frame.rs
```diff
@@ -120,42 +120,21 @@ pub enum Frame {
...
-        id: StreamId,
-    },
-    StopSending {
-        id: StreamId,
-        error_code: u16,
-    },
-    RetireConnectionId {
-        sequence: u64,
-    },
+    Blocked { offset: u64 },
```
djc: Unrelated style changes?

quinn-proto/src/frame.rs
```diff
@@ -569,11 +548,11 @@ impl Iter {
                 }
                 let mut reset_token = [0; RESET_TOKEN_SIZE];
                 self.bytes.copy_to_slice(&mut reset_token);
-                Frame::NewConnectionId {
+                Frame::NewConnectionId(NewConnectionId {
```
djc: Separate commit for this type extraction?

quinn-proto/src/endpoint.rs
djc: Also it feels odd for the majority of this hunk to live in `Endpoint`, why is that?

quinn-proto/src/connection.rs
```diff
@@ -2798,7 +2812,13 @@ pub enum Io {
         packet: Box<[u8]>,
     },
     /// Stop or (re)start a timer
-    TimerUpdate { timer: Timer, update: TimerUpdate },
+    TimerUpdate {
```
djc: Unrelated style change.

quinn-proto/src/endpoint.rs
Ralith: Most of the changes in this hunk are semantic no-ops to ensure we can call `self.new_cid()` without an outstanding borrow. That said, I think the placement is justified. We're doing two things here: reporting the establishment of a new incoming connection, and providing new connections with their alternate CIDs. Both of these inherently involve modifying `Endpoint` state, for the incoming connection queue and the connection ID table respectively.

I'll insert a constant for the CID count.

quinn-proto/src/frame.rs
Ralith: rustfmt does this directly in response to the type extraction, for some reason.

quinn-proto/src/connection.rs
Ralith: rustfmt does this directly in response to the added enum variant, for some reason.

Ralith:
I think that covers all the comments.

Ralith:
Rebased.

PR #125, opened 2018-12-15
Test reordering around key updates

Ralith:
This test is expected to fail under draft 16, but should be enabled when we move to draft 17.


djc:
Nice, thanks for getting this done!

Ralith:
`#[ignore]` causes the test not to be ran at all. There's `#[should_panic]` as well; I've switched to that in the spirit of ensuring we don't forget about this.

PR #126, opened 2018-12-15
Send/receive ECN bits on linux

Ralith:
Not sure we want to merge this before libc merges my PR there, but there's lots of esoterica to scrutinize in the meantime.

### DIFF
--- a/quinn/src/platform/fallback.rs
+++ b/quinn/src/platform/fallback.rs
@@ -1,0 +1,24 @@
+use std::{io, net::SocketAddr};
+
+use mio::net::UdpSocket;
+
+use quinn_proto::EcnCodepoint;
+
+impl super::UdpExt for UdpSocket {
+    fn init_ext(&self) -> io::Result<()> {
+        Ok(())
+    }
+
+    fn send_ext(
+        &self,
+        remote: &SocketAddr,
+        _: Option<EcnCodepoint>,
+        msg: &[u8],
+    ) -> io::Result<usize> {
+        self.send_to(msg, remote)
+    }
+
+    fn recv_ext(&self, buf: &mut [u8]) -> io::Result<(usize, SocketAddr, Option<EcnCodepoint>)> {
+        self.recv_from(buf).map(|(x, y)| (x, y, None))
+    }
+}
--- a/quinn/src/platform/linux.rs
+++ b/quinn/src/platform/linux.rs
@@ -1,0 +1,143 @@
+use std::os::unix::io::AsRawFd;
+use std::{
+    io, mem,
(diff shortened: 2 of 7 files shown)


djc:
Can you link to your libc PR from here? Also I think it'd be good to discuss at least with the tokio-udp folks if they want to take this stuff before we land it here?

Ralith:
See https://github.com/rust-lang/libc/pull/1175.

I'm not sure what integration into tokio-udp would look like; ECN is just one of many(?) possible types of ancillary data that can be sent/received alongside packets for niche purposes. A sufficiently generic API seems likely to be deemed out of scope, as it would probably dwarf the other parts of the crate. This also doesn't strike me as a maintenance hazard; udp.rs is the full extent of the code approximately duplicated from tokio-udp, and is trivial.

We may also want to digress further from the beaten bath API in the future. For example, we can use the same interfaces on Linux to get access to hardware timestamps for high-precision timing of packet receipt times for improved congestion control, and we might want to adopt `sendmmsg`/`recvmmsg` for reduced context switching. Using our own UDP binding will make this easier.

That said, I suppose it won't hurt to kick off some discussion and see where it goes.

Ralith:
Opened https://github.com/tokio-rs/tokio/issues/799.

quinn/src/platform/linux.rs
```diff
@@ -0,0 +1,170 @@
...
+    io, mem,
+    net::{SocketAddr, SocketAddrV4, SocketAddrV6},
+    ptr,
+};
+
+use mio::net::UdpSocket;
+
+use quinn_proto::EcnCodepoint;
+
+const CMSG_LEN: usize = 24;
```
djc: Nit: put this at the bottom?

quinn/src/platform/fallback.rs
```diff
@@ -0,0 +1,21 @@
+use mio::net::UdpSocket;
+
+use quinn_proto::EcnCodepoint;
+
+pub fn init(socket: &UdpSocket) -> io::Result<()> {}
```
djc: Feels like this wants to be trait.

quinn/src/platform/fallback.rs
```diff
@@ -0,0 +1,27 @@
+use std::{io, net::SocketAddr};
+
+use mio::net::UdpSocket;
+
+use quinn_proto::EcnCodepoint;
+
+#[allow(dead_code)]
+pub struct Platform;
```
djc: Why do we still need this?

djc:
Approved, though I would be in favor of squashing these commits before merging.

PR #127, opened 2018-12-15
Move core of timeout implementation into Connection

djc:


quinn-proto/src/endpoint.rs
```diff
@@ -690,35 +690,27 @@ impl Endpoint {
...
     pub fn timeout(&mut self, now: u64, conn: ConnectionHandle, timer: Timer) {
+        let mut mux = EndpointMux::new(&mut self.ctx, conn, self.connections[conn.0].side());
+        let closed = self.connections[conn.0].timeout(&mut mux, now, timer);
         match timer {
             Timer::Close => {
                 self.ctx.io.push_back(Io::TimerStop {
                     connection: conn,
                     timer: Timer::Idle,
                 });
                 self.ctx.events.push_back((conn, Event::ConnectionDrained));
```
Ralith: These updates could be moved into `Connection` to be mediated (more concisely) by the mux.

quinn-proto/src/endpoint.rs
```diff
@@ -690,35 +690,27 @@ impl Endpoint {
...
                 }
             }
             Timer::Idle => {
-                let mut mux =
-                    EndpointMux::new(&mut self.ctx, conn, self.connections[conn.0].side());
-                self.connections[conn.0].close_common(&mut mux, now);
-                self.connections[conn.0].state = State::Draining;
                 self.ctx
                     .events
                     .push_back((conn, ConnectionError::TimedOut.into()));
```
Ralith: This too.

quinn-proto/src/endpoint.rs
```diff
@@ -690,35 +690,27 @@ impl Endpoint {
...
                 self.dirty_conns.insert(conn); // Ensure the loss detection timer cancellation
                                                // goes through
```
Ralith: Probably best to do this unconditionally on any timeout; loss detection and close timeout can both incur I/O events, after all.

Ralith: On second thought, a close timeout should *not* cause a new entry in `dirty_conns`, because the only I/O it might incur is queued proactively rather than needing to be polled out, and the connection might cease to exist afterwards.

Ralith: And, in fact, the comment on this specific line is obsolete; we *only* need to mark this connection dirty for the loss detection timer.

Ralith: As of tonight's refactoring, we once again need to mark dirty for potential timer updates.

quinn-proto/src/endpoint.rs
Ralith: On second thought, probably best not to lose track of the `ConnectionDrained` event when we potentially forget the connection below, since it's a useful to the upper layer that this connection's done with.

Actually, come to think of it, `ConnectionDrained` isn't necessary at all, since close timer expiry is always unambiguous. I'll scrap it.

quinn-proto/src/connection.rs
```diff
@@ -580,7 +580,27 @@ impl Connection {
...
     }
 
-    pub fn check_packet_loss(&mut self, now: u64) {
+    pub fn timeout(&mut self, now: u64, timer: Timer) -> bool {
+        match timer {
+            Timer::Close => {
+                self.io.timer_stop(Timer::Idle);
+                if !self.app_closed {
+                    self.state = State::Drained;
+                }
```
Ralith: This shouldn't be conditional; we are certainly in the `Drained` state now. The original code had a branch only because there wasn't much point updating the state one line above forgetting about it, but better to be explicit now that those lines are separated.

djc: Makes sense.

quinn-proto/src/endpoint.rs
```diff
@@ -718,34 +713,14 @@ impl Endpoint {
...
-                self.dirty_conns.insert(conn); // Connection will want to retransmit
-            }
+        if self.connections[conn.0].timeout(now, timer) {
+            self.forget(conn);
+            return;
+        }
+        if let Timer::Idle = timer {
+            self.eventful_conns.insert(conn);
         }
+        self.dirty_conns.insert(conn); // Connection will want to retransmit
```
Ralith: This comment is inaccurate after these changes; best to either remove it or cite the `Connection::poll_io` requirements.

djc: Fixed!

Issue #22, opened 2018-06-02
Implement QUIC-HTTP framing


djc:
Implement HTTP framing, as defined in the spec, here:

https://tools.ietf.org/html/draft-ietf-quic-http-11#section-4

I've started the process with the `SETTINGS` frame, here:

https://github.com/djc/quinn/blob/pre-quicr-quinn/src/http/frame.rs

passchaos:
Any progress?

djc:
Not yet, we've been focusing on updating our spec compliance to the latest draft. Would you like to help out? I'm happy to mentor!

passchaos:
Sorry, I am busy recently. I am happy to do this in the future.

djc:
No problem, I figured it was worth asking. We'll get to it eventually, hopefully somewhat soon.

stammw:
Hey! I'd like to start working on this next !

djc:
@stammw great! The idea here is just to implement the bytes-to-types and types-to-bytes code for the data structures defined in the HTTP spec. Ideally the module would include unit tests that just check whether round-tripping works as expected. I think it would make sense to use the `Codec` trait as is used for example in the `frame` module (which should provide good examples).

Let us know if you have questions!

Ralith:
It might be worth leveraging types from [the http crate](https://crates.io/crates/http), though I'm not sure how much work might be required on their end to play well with HTTP3 concepts.

djc:
I don't think the frames match the HTTP spec itself, so I don't think you'll find much to use from the http crate.

I forgot to mention it, but I agree that a separate crate would make sense. To disambiguate the HTTP protocol version, I think it would be good to call it `quinn-h3`.

stammw:
Hey,
What do you think about how I should handle the QPACK encoded frames (HEADERS, including PUSH_PROMISE ones) ?

1. Return `HttpFrame` variant that contains raw `Bytes`
2. Return `HttpFrame` variant that contains a stubbed QPACK type
3. Do not implement them for the moment

Ralith:
In the interest of being able to run tests, stubbing them out seems most practical. Unless there's some conceivable use for the raw data other than immediately decompressing it?

stammw:
Actually, I think QPACK wouldn't be a data type but a trait. So stubbing and using raw types inside variants might be more or less the same solution. And at the and the encode -> decode -> compare should still be writable even if the `given` part would have to be changed afterward?

[...]

PR #130, opened 2018-12-17
[WIP] HTTP framing implementation

stammw:
Relates to #22.

### DIFF
--- a/quinn-h3/src/frame.rs
+++ b/quinn-h3/src/frame.rs
@@ -15,6 +15,7 @@ pub enum Error {
 #[derive(Debug, PartialEq)]
 pub enum HttpFrame {
     Data(DataFrame),
+    Priority(PriorityFrame),
     Settings(SettingsFrame),
 }
 
@@ -22,6 +23,7 @@ impl HttpFrame {
     pub fn encode<T: BufMut>(&self, buf: &mut T) {
         match self {
             HttpFrame::Data(f) => f.encode(buf),
+            HttpFrame::Priority(f) => f.encode(buf),
             HttpFrame::Settings(f) => f.encode(buf),
         }
     }
@@ -39,6 +41,7 @@ impl HttpFrame {
             Type::DATA => Ok(HttpFrame::Data(DataFrame {
                 payload: payload.collect(),
             })),
+            Type::PRIORITY => Ok(HttpFrame::Priority(PriorityFrame::decode(&mut payload)?)),
             Type::SETTINGS => Ok(HttpFrame::Settings(SettingsFrame::decode(&mut payload)?)),
             _ => Err(Error::UnsupportedFrame),
         }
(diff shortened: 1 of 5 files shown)


quinn-h3/src/frame.rs
```diff
@@ -0,0 +1,272 @@
...
+    fn encode_header<T: BufMut>(&self, buf: &mut T) {
+        println!("len is {}", self.len());
+        varint::write(self.len() as u64, buf).unwrap();
+        buf.put_u8(self.ftype().0);
+    }
+}
+
+#[derive(Debug, PartialEq)]
+pub struct DataFrame {
+    payload: Bytes,
```
stammw: I'm not sure about using `Bytes` here for the interface with the user. Won't it create data copies ? I tried to use `<T> where T: AsRef<[u8]>` without success as I've seen in `quinn-proto::frame::Stream` encoding impl. I believe it's a function-call ergonomics matter as well ?

quinn-proto/src/frame.rs
```diff
@@ -48,6 +48,7 @@ pub trait FrameStruct {
     const SIZE_BOUND: usize;
 }
 
+#[macro_export]
```
Ralith: This macro currently hardcodes special handling of transport protocol STREAM frames in the `Display` impl, which external code probably doesn't want. In lieu of a more abstract macro, I'd recommend not exporting this, and instead including a similar but distinct one in new crate.

quinn-h3/src/frame.rs
Ralith: I think using `Bytes` here is fine. The fancy polymorphic `Stream` type in `-proto` is an implementation detail, and is not actually conferring any benefit at all right now. Buffering and flow control concerns, as well as the difficulty and expense of storing generic types in non-generic data structures, generally dictate that copying occurs anyway.

quinn-h3/src/frame.rs
```diff
@@ -0,0 +1,272 @@
...
+        Ok(Type(buf.get_u8()))
+    }
+    fn encode<B: BufMut>(&self, buf: &mut B) {
+        buf.write(self.0);
+    }
+}
+
+trait FrameHeader {
+    fn len(&self) -> usize;
+    fn ftype(&self) -> Type;
```
Ralith: Could this be
```suggestion
    const TYPE: Type;
```
instead?

quinn-h3/src/frame.rs
```diff
@@ -0,0 +1,272 @@
...
+impl HttpFrame {
+    pub fn encode<T: BufMut>(&self, buf: &mut T) {
+        match self {
+            HttpFrame::Data(f) => f.encode(buf),
+            HttpFrame::Settings(f) => f.encode(buf),
+        }
+    }
+
+    pub fn decode<T: Buf>(buf: &mut T) -> Result<Self, Error> {
+        let len = varint::read(buf).ok_or(Error::Malformed)?;
```
Ralith: `BufExt` defines `fn get_var(&mut self) -> coding::Result<u64>`, which would be more concise and preclude needing to publicize/import `proto::varint`.

quinn-h3/src/frame.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    fn decode<T: Buf>(buf: &mut T) -> Result<SettingsFrame, Error> {
+        let mut settings = SettingsFrame::default();
+        while buf.has_remaining() {
+            if buf.remaining() < 3 {
+                // remains less than id + minimum-size varint
+                return Err(Error::Malformed);
+            }
+            let identifier = buf.get::<SettingId>()?;
+            let value = varint::read(buf).ok_or(Error::Malformed)?;
```
Ralith: As above.

quinn-h3/src/frame.rs
```diff
@@ -0,0 +1,272 @@
...
+        buf.write(self.0);
+    }
+}
+
+trait FrameHeader {
+    fn len(&self) -> usize;
+    fn ftype(&self) -> Type;
+    fn encode_header<T: BufMut>(&self, buf: &mut T) {
+        println!("len is {}", self.len());
+        varint::write(self.len() as u64, buf).unwrap();
```
Ralith: `BufMutExt` defines a `write_var` helper that may be of interest, in this and other cases.

quinn-proto/src/frame.rs
stammw: Yes, this module doesn't actually need to have use the same macro. Fixed.

quinn-h3/src/frame.rs
stammw: Oh, didn't know about associated const. Rust is neat.

stammw:
Almost done !

Still have to implement ignoring some frame fromated as described in `4.2.10. Reserved Frame Types ` and a last check against the specs.

stammw:
Changes imported in #136

PR #134, opened 2018-12-19
Draft 17 recovery logic

Ralith:
Pretty much just went through and transcribed the new pseudocode, with minor adjustments along the same lines as the ones made previously. The logic is a little simpler and makes a little more sense this time around (e.g. no always-false bools).

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -542,7 +542,7 @@ impl Connection {
                 } else {
                     // Congestion avoidance.
                     self.congestion_window +=
-                        self.config.default_mss * info.size as u64 / self.congestion_window;
+                        self.config.max_datagram_size * info.size as u64 / self.congestion_window;
                 }
             }
         }
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -1890,9 +1890,11 @@ impl Connection {
             }
         }
 
+        let mut padded = false;
         if let Header::Initial { .. } = header {
             if buf.len() < MIN_INITIAL_SIZE - AEAD_TAG_SIZE {
                 buf.resize(MIN_INITIAL_SIZE - AEAD_TAG_SIZE, 0);
+                padded = true;
             }
         }
         let pn_len = header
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection.rs
```diff
@@ -1842,7 +1844,7 @@ impl Connection {
                 size: buf.len() as u16,
                 is_crypto_packet: crypto_level == CryptoLevel::Initial && !ack_only,
                 ack_eliciting: !ack_only,
-                in_flight: !ack_only,
+                in_flight: padded || !ack_only,
```
Ralith: This has no effect as-is, but will become more relevant when e.g. bd91d282cf4404090ccd6331b933f8d5686a671e is merged.

quinn-proto/src/connection.rs
```diff
@@ -707,57 +671,49 @@ impl Connection {
...
-    fn set_loss_detection_alarm(&mut self) {
+    fn set_loss_detection_timer(&mut self) {
         if self.bytes_in_flight == 0 {
             self.io.timer_stop(Timer::LossDetection);
             return;
         }
 
-        let mut alarm_duration: u64;
-        if self.awaiting_handshake {
+        let mut timeout: u64;
```
djc: I'm not sure this makes sense? This `timeout`'s liveness in `if self.crypto_in_flight` seems fully disjoint from the other stuff that uses this declaration below, so it would be clearer to just declare this twice?

Ralith: This is another case of me erring on the side of transcribing slightly weird pseudocode from the draft. Think we should diverge for clarity's sake?

djc: Can you point me to the section in the spec?

Ralith: I was thinking of https://tools.ietf.org/html/draft-ietf-quic-recovery-17#section-6.4.7, but on review I see there's no declaration at *all*, so I suppose I might as well scope it sanely.

Ralith:
Rebased.

Ralith:
Rebased. Thanks for all the review work!

PR #20, opened 2018-05-29
QPACK decoder stream

evolix1:
# QPACK encoding

Pull Request: #4 

Current status in functionality:
- [X] Dynamic and Static tables
- [X] Field indexing
- [X] Primitive decoder
- [x] Primitive encoder
- [X] Huffman string decoding
- [x] Huffman string encoding
- [X] Decoder stream
- [ ] Encoder stream
- [ ] Encoding strategies
- [ ] Request and Push stream decoding

Status in development:
- [X] Types are testing separately
- [ ] Testing scenarios that integrate the full pipeline (ex: Frame → Encoder → Decoder → Table)
- [ ] Benchmarks around decoding
- [ ] Benchmarks around encoding

Notable points:
- There are still confusing / misleading points in the draft. Maybe I will have to rework some of the decoding part later.
- I couldn't use `Codec` trait because `Decoder` needs to persists between calls (notable the state of the dynamic table). Instead, you can use `Decoder::feed_stream` as an entry point.

PR #136, opened 2018-12-23
QPACK and naive H3 implementation

stammw:
Continuing the work of @evolix1, from #20, which is blocking #130.

Features:
* [x] Encoder
* [x] Reference tracking
* [x] Track DynamicTable::fields mutation, update DynamicTable::*_map when constructing DynamicTableEncoder instead of rebuilding HashMaps
* [-] N bit handling
* [-] Encode / decode concurreny (needs to have concurrent shared mutable state...)
* [ ] Configuration (may need H/3 impl to start)

Optimization: 
* [ ] Improve Literal argument passing: 
   * ```LiteralWithPostBaseNameRef::new(index, field.value.clone())```: No
   * Look for a solution in https://deterministic.space/elegant-apis-in-rust.html, https://phaazon.net/blog/on-owning-borrowing-pub-interface
* [ ] Header fields lifetimes for Cow
* [ ] Use Reference count for DynamicTable::fields DynamicTable::*_map


quinn-h3/src/qpack/block.rs
```diff
@@ -0,0 +1,365 @@
+use bytes::{Buf, BufMut};
+
+use super::prefix_int;
+use super::prefix_string;
+use super::ParseError;
+
+pub(super) enum HeaderBlocField {
```
Ralith:
```suggestion
pub(super) enum HeaderBlockField {
```

Ralith: Is it really necessary to use `pub(super)` in this module, given that the module is not itself reexported?

quinn-h3/src/qpack/decoder.rs
```diff
@@ -0,0 +1,623 @@
...
+    Duplicate, DynamicTableSizeUpdate, EncoderInstruction, HeaderAck, InsertCountIncrement,
+    InsertWithNameRef, InsertWithoutNameRef, StreamCancel,
+};
+use super::ParseError;
+
+use super::prefix_int;
+use super::prefix_string;
+
+#[derive(Debug, PartialEq)]
+pub enum Error {
```
Ralith: If this error (and the others like it) ever reach a public API, it'd be nice to derive `Display` and `Error` impls for them using err-derive (see e.g. `quinn_proto::ConnectionError` for an example)

quinn-h3/src/qpack/vas.rs
```diff
@@ -0,0 +1,274 @@
+/**
+ * https://tools.ietf.org/html/draft-ietf-quic-qpack-01#section-2.2.1
+ * https://tools.ietf.org/html/draft-ietf-quic-qpack-01#section-2.2.2
+ */
+
+/*
+ *  # Virtualy infinit address space mapper.
```
Ralith:
```suggestion
 *  # Virtually infinite address space mapper.
```

quinn-h3/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "quinn-h3"
+version = "0.0.1"
+authors = ["Benjamin Saunders <ben.e.saunders@gmail.com>", "Dirkjan Ochtman <dirkjan@ochtman.nl>"]
```
djc: You should add yourself here!

stammw:
@djc I did not think of that. I'm not sure the common code surface would be that big. Maybe for the primitive types like `prefix_string`.

djc:
I think that's what I was looking at, but I understand if there's not enough overlap. 👍

stammw:
@djc Regarding the allow(dead_code) usages, do you prefer that I delete them and leave the build yield warnings until the functions are used by the H3 impl?

stammw:
I think we're ready to merge ! 

The interop tool has been updated with only a very simple H3 scenario (with no encoder stream, nor receiving anything from server other than the response). 

Trying to implement a full-featured client, I started to think it was too much code to write in form of a quick implementation. So it would be better to continue writing a solid code into another PR.

interop/src/main.rs
```diff
@@ -176,6 +182,50 @@ fn run(log: Logger, options: Opt) -> Result<()> {
...
+            .and_then(|conn| {
+                assert!(state.lock().unwrap().saw_cert);
+                handshake = true;
+                let conn = conn.connection;
+                let control_stream = conn.open_uni();
+                let control_fut = control_stream
+                    .map_err(|e| format_err!("failed to open control stream: {}", e))
+                    .and_then(move |stream| {
+                        let mut buf = BytesMut::new();
+                        buf.put(0x43u8);
```
djc: What is this for?

Ralith:
I'll try to have a look at the futuresy bits tomorrow per @stammw's request, at least.

interop/src/main.rs
stammw: The control stream is useful to send one's endpoint settings https://quicwg.org/base-drafts/draft-ietf-quic-http.html#rfc.section.3.2.1. This is not mandatory for this code to work but I thought I would leave it as is to validate that the `SETTINGS FRAME` is correctly encoded.

Normally it shouldn't be chained with the request, but executed in parallel. This is something I plan to implement in future HTTP lifecycle module. Moreover the implementation should receive one `SETTINGS FRAME`  from the server and change it's QPACK encoder behavior accordingly. Which will involve shared mutable state.

Ralith: Does it need to involve a magic number hardcoded into the interop tool?

interop/src/main.rs
```diff
@@ -69,7 +72,10 @@ fn run(log: Logger, options: Opt) -> Result<()> {
     tls_config
         .dangerous()
         .set_certificate_verifier(Arc::new(InteropVerifier(state.clone())));
-    tls_config.alpn_protocols = vec![str::from_utf8(quinn::ALPN_QUIC_HTTP).unwrap().into()];
+    tls_config.alpn_protocols = vec![
+        str::from_utf8(quinn::ALPN_QUIC_HTTP).unwrap().into(),
+        str::from_utf8(quinn::ALPN_QUIC_H3).unwrap().into(),
+    ];
```
Ralith: We should have two separate client configs, one for `hq-whatever` and one for `h3-whatever`. Placing both in a single config doesn't mean you can use either, it means you're willing to let the server select either for you, and it's your responsibility to comply with the server's selection. Plus, this way you can enable/disable incoming unidirectional streams for each as appropriate.

Also, we can get rid of the `str::from_utf8(...).unwrap()` now.

interop/src/main.rs
```diff
@@ -176,6 +182,50 @@ fn run(log: Logger, options: Opt) -> Result<()> {
...
+                    .and_then(|req_stream| h3_get(req_stream))
+                    .and_then(move |data| {
+                        println!(
+                            "read {} bytes: \n\n{}\n\n closing",
+                            data.len(),
+                            String::from_utf8_lossy(&data)
+                        );
+                        conn.close(0, b"done").map_err(|_| unreachable!())
+                    });
+                control_fut.and_then(|_| req_fut).map(|_| h3 = true)
```
Ralith: You can make these concurrent by replacing `and_then` with `join`

interop/src/main.rs
```diff
@@ -197,12 +247,74 @@ fn run(log: Logger, options: Opt) -> Result<()> {
...
+    match HttpFrame::decode(&mut cur) {
+        Ok(HttpFrame::Headers(text)) => {
+            let mut resp_block = std::io::Cursor::new(&text.encoded);
+            match qpack::decode_header(table, &mut resp_block) {
+                Ok(_) => (),
+                Err(e) => return Err(format_err!("failed to decode response header {}", e)),
+            }
+        }
+        Ok(f) => {
+            println!("response basd frame type {:?}", f);
```
Ralith:
```suggestion
            println!("response bad frame type {:?}", f);
```

interop/src/main.rs
```diff
@@ -176,6 +182,50 @@ fn run(log: Logger, options: Opt) -> Result<()> {
...
     }
 
+    let mut h3 = false;
+    let result = runtime.block_on(
+        endpoint
+            .connect_with(&client_config, &remote, &options.host)?
+            .map_err(|e| format_err!("failed to connect: {}", e))
+            .and_then(|conn| {
+                assert!(state.lock().unwrap().saw_cert);
+                handshake = true;
```
Ralith: Copy-paste error? These two lines are redundant to their prior instance and should be removed here.

interop/src/main.rs
```diff
@@ -176,6 +182,50 @@ fn run(log: Logger, options: Opt) -> Result<()> {
         }
     }
 
+    let mut h3 = false;
+    let result = runtime.block_on(
+        endpoint
+            .connect_with(&client_config, &remote, &options.host)?
```
Ralith:
```suggestion
            .connect_with(&client_config, &remote, host)?
```
Looks like you missed this in a rebase.

quinn-proto/src/lib.rs
```diff
@@ -54,6 +54,9 @@ pub const VERSION: u32 = 0xff00_0011;
 /// TLS ALPN value for HTTP over QUIC
 pub const ALPN_QUIC_HTTP: &[u8] = b"hq-17";
 
+/// TLS ALPN value for H3
+pub const ALPN_QUIC_H3: &[u8] = b"h3-17";
```
Ralith:
```suggestion
pub const ALPN_QUIC_H3: &[u8] = b"h3-18";
```
As of #225 (whoops, looks like I forgot this above too, feel free to get them both)

interop/src/main.rs
Ralith: re: shared mutable state: it's acceptable to have `Arc<Mutex<State>>` (perhaps hidden in a helper type that the user constructs per-stream?) for cases like this, IMO.

djc: Yeah, I was also mostly responding to the magic number, which seems a bit awkward and hard to understand.

stammw: Ah yes, it's not very clean indeed...

stammw: fixed in 51521a0

interop/src/main.rs
stammw: Thank you, it's actually serially executed on purpose. Not making any request if the control stream fails.

interop/src/main.rs
stammw: fixed in 4ee65b9

interop/src/main.rs
stammw: fixed in 4ee65b9

interop/src/main.rs
stammw: fixed in 4ee65b9

quinn-proto/src/lib.rs
stammw: fixed in 8569932

djc:
🎉

PR #137, opened 2018-12-23
Upgrade orion to 0.11

stammw:


Ralith:
Thanks, this was causing some pain.

PR #139, opened 2018-12-23
Draft 17 transport parameters

Ralith:
Integer parameters were moved to varint encoding, and everything was renumbered.


Ralith:
> what can we do

Be very thankful that it's mostly wrapped up in a macro :D

PR #141, opened 2018-12-26
Spin bit, misc cleanup

Ralith:


Ralith:
Rebased.

djc:
Nice cleanups!

I remember seeing somewhere that implementations should implementations should ignore spin bits some part of the time to provide anonymity to clients who intentionally don't respond to them for privacy reasons. Should we have a follow up issue for that?

Ralith:
> Should we have a follow up issue for that?

I didn't see any such language in the draft, so I'd want such an issue to cite a reference motivating it and providing guidance on the probability to use (or just to present strong arguments for that behavior directly). My reading of the current draft is that implementing the spin bit is mandatory, though admittedly it doesn't seem like a severe interop hazard.

PR #143, opened 2018-12-27
Draft 17 handshake crypto

Ralith:
This introduces the notion of discrete packet number spaces and the associated crypto and retransmit management. Currently non-functional due to, at minimum, bugs in the rustls `QuicExt::get_handshake_secrets` implementation, which I'll continue to investigate tomorrow.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -40,7 +40,7 @@ pub struct Connection {
     loc_cids: HashMap<u64, ConnectionId>,
     rem_cid: ConnectionId,
     rem_cid_seq: u64,
-    pub(crate) remote: SocketAddrV6,
+    remote: SocketAddrV6,
     state: State,
     side: Side,
     mtu: u16,
@@ -748,7 +748,7 @@ impl Connection {
         }
     }
 
-    pub fn reset_idle_timeout(&mut self, now: u64) {
+    fn reset_idle_timeout(&mut self, now: u64) {
         let dt = if self.config.idle_timeout == 0 || self.params.idle_timeout == 0 {
             cmp::max(self.config.idle_timeout, self.params.idle_timeout)
         } else {
@@ -2491,7 +2491,7 @@ impl Streams {
 }
 
 #[derive(Debug, Clone)]
-pub struct Retransmits {
+struct Retransmits {
     max_data: bool,
     max_uni_stream_id: bool,
     max_bi_stream_id: bool,
@@ -2523,7 +2523,7 @@ impl Retransmits {
             && self.retire_cids.is_empty()
     }
 
-    pub fn path_challenge(&mut self, packet: u64, token: u64) {
+    fn path_challenge(&mut self, packet: u64, token: u64) {
         match self.path_response {
             None => {
                 self.path_response = Some((packet, token));
@@ -2698,7 +2698,7 @@ impl State {
     }
 }
 
-pub mod state {
+mod state {
     use super::*;
 
     #[derive(Clone)]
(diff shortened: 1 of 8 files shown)


quinn-proto/src/connection.rs
```diff
@@ -71,6 +73,12 @@ pub struct Connection {
     events: VecDeque<Event>,
     /// Number of local connection IDs that have been issued in NEW_CONNECTION_ID frames.
     cids_issued: u64,
+    /// Packet number spaces: initial, handshake, 1-RTT
```
djc: I'm a little disappointed that my `VecDeque` structure didn't survive this! Is random access to spaces really so important? My intuition was that most access will want highest space anyway, and I liked not having to deal with the `Option`s.

quinn-proto/src/connection.rs
```diff
@@ -113,6 +121,8 @@ pub struct Connection {
     /// count towards bytes_in_flight to ensure congestion control does not impede congestion
     /// feedback.
     bytes_in_flight: u64,
+    /// Number of packets bearing CRYPTO frames in flight
```
djc: Maybe keep this in its former order in this commit and move it in a later commit?

quinn-proto/src/connection.rs
```diff
@@ -630,42 +624,50 @@ impl Connection {
...
-            for packet in lost_packets {
-                let info = self.sent_packets.remove(&packet).unwrap();
-                if !info.in_flight {
-                    continue;
-                }
-                self.bytes_in_flight -= info.size as u64;
-                self.lost_packets += 1;
-                if info.is_crypto_packet {
-                    self.crypto_pending += info.retransmits;
+        let mut lost_nonack = false;
```
djc: (I know this is not new here and don't know if this is literally from the spec, but I dislike the `nonack` name... could use an underscore to disambiguate how to read it.)

quinn-proto/src/connection.rs
```diff
@@ -875,37 +884,96 @@ impl Connection {
...
+                .crypto
+                .push_back(frame::Crypto {
+                    offset,
+                    data: outgoing.into(),
+                });
         }
     }
 
+    /// Switch to stronger cryptography during handshake
+    fn upgrade_crypto(&mut self, space: SpaceId, secrets: rustls::quic::Secrets) {
```
djc: I would prefer to export `rustls` items through the `crypto` module, to maintain a bit more of an abstraction boundary -- the choice of `rustls` seems one of the more controversial aspects of our implementation, so it'd be nice to keep it wrapped up.

(I see `AlertDescription` already snuck in, but let's not make it worse?)

quinn-proto/src/connection.rs
```diff
@@ -1282,14 +1268,68 @@ impl Connection {
         }
     }
 
+    /// Process an Initial or Handshake packet payload
+    fn handle_early(&mut self, now: u64, packet: Packet) -> Result<(), ConnectionError> {
```
djc: Let's call this `handle_early_payload()`? Factoring this out looks really nice.

quinn-proto/src/connection.rs
```diff
@@ -1561,144 +1604,95 @@ impl Connection {
...
-                crypto = self.zero_rtt_crypto.as_ref().unwrap();
-                Header::Long {
-                    ty: types::ZERO_RTT,
-                    number: number as u32,
-                    src_cid: self.loc_cid.clone(),
-                    dst_cid: self.init_cid.clone(),
-                }.encode(&mut buf);
-            } else {*/
-            let header = Header::Short {
+        let header = match space_id {
```
djc: Woo, love this hunk!

quinn-proto/src/connection.rs
Ralith: Your intuition is right in the sense of number of times accessed over the course of a typical connection, but it looks like few accesses are statically guaranteed to always want the highest space per se. The congestion control logic and `next_packet` may both need to use an earlier space if packet loss occurs during the transition, and all that leaves are the unconditional 1-RTT accesses (which look approximately the same regardless), `make_probe`, and `make_close`.

Ralith: I can probably abstract out the unwrapping for accesses via `self.space[_mut](...)`, though, looks like those are always expected to succeed.

PR #145, opened 2018-12-29
Assorted bugfixes

Ralith:
This resolves the issues I've identified so far in the course of interop testing. Some missed draft-17 changes, mostly logic errors.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -120,6 +120,8 @@ pub struct Connection {
     bytes_in_flight: u64,
     /// Number of unacknowledged Initial or Handshake packets bearing CRYPTO frames
     crypto_in_flight: u64,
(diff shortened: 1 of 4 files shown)


quinn-proto/src/connection.rs
```diff
@@ -1914,10 +1914,12 @@ impl Connection {
 
         let space = self.spaces[space_id as usize].as_mut().unwrap();
         let mut padded = false;
-        if let Header::Initial { .. } = header {
-            if buf.len() < MIN_INITIAL_SIZE - AEAD_TAG_SIZE {
-                buf.resize(MIN_INITIAL_SIZE - AEAD_TAG_SIZE, 0);
-                padded = true;
+        if self.side.is_client() {
```
djc: We should probably invert the nesting of these conditions, since the check for initial space is false more often.

Issue #116, opened 2018-12-10
Detect sequential key updates and abort connection on this condition


stammw:
When the KEY_PHASE is flipped twice in two consecutive received packet, the connection must be aborted:

> An endpoint does not always need to send packets when it detects that
> its peer has updated keys.  The next packet that it sends will simply
> use the new keys.  If an endpoint detects a second update before it
> has sent any packets with updated keys, it indicates that its peer
> has updated keys twice without awaiting a reciprocal update.  An
> endpoint MUST treat consecutive key updates as a fatal error and
> abort the connection.

Ralith:
> When the KEY_PHASE is flipped twice in two consecutive received packet

That's not quite right: it is an error for `KEY_PHASE` to be flipped in any packet that is numbered after a prior `KEY_PHASE` flip and received before the local endpoint has sent a packet using the new keys. In other words, an endpoint may not initiate a key update until it has received a packet encrypted using the current keys.

PR #146, opened 2018-12-29
Key discard, bugfixes, cleanup

### LINKED ISSUE #114: Discard old crypto when all packets encrypted with it have been transmitted
As recommended in quic-tls [6.  Key Update](https://tools.ietf.org/html/draft-ietf-quic-tls-16#section-6):

> Keys and their corresponding secrets SHOULD be discarded when an
>  endpoint has received all packets with packet numbers lower than the
>  lowest packet number used for the new key.  An endpoint might discard
>  keys if it determines that the length of the delay to affected
>  packets is excessive.

Ralith:
Fixes #114 and #116


quinn-proto/src/connection.rs
```diff
@@ -1889,10 +1889,13 @@ impl Connection {
         // To ensure that sufficient data is available for sampling, packets are padded so that the
         // combined lengths of the encoded packet number and protected payload is at least 4 bytes
         // longer than the sample required for header protection.
-        if let Some(padding) =
-            (space.header_crypto.sample_size() + 4).checked_sub(buf.len() - header_len + pn_len)
+        let protected_payload_len = (buf.len() + AEAD_TAG_SIZE) - header_len;
+        if let Some(padding_minus_one) =
```
djc: Surely there's some semantic reason for `padding_minus_one`? :grin:

quinn-proto/src/connection.rs
```diff
@@ -691,17 +688,11 @@ impl Connection {
...
-            .then(packet.cmp(&self.rx_packet))
-        {
-            Ordering::Greater | Ordering::Equal => {
-                self.rx_packet = packet;
-                self.rx_packet_time = now;
-                // Update outgoing spin bit, inverting iff we're the client
-                self.spin = self.side.is_client() ^ spin;
-            }
-            _ => {}
+        if (space_id, packet) >= (self.rx_packet_space, self.rx_packet) {
```
djc: Ah, good, I was wondering about that.

quinn-proto/src/connection.rs
Ralith: It's that or have it come out as zero sometimes and execute the body gratuitously, which would be harmless as it was previously written, but now that there's a logging statement I figured I'd make the branch a little more robust.

djc: I just meant the variable name.

Ralith: Whoops, that was actually computing padding *plus* one. Fixed.

PR #153, opened 2018-12-30
Test Quinn API

djc:


quinn/src/lib.rs
```diff
@@ -1599,3 +1599,137 @@ impl<T: Read> Future for ReadToEnd<T> {
...
+                        let conn = conn.connection;
+                        let stream = conn.open_bi();
+                        stream
+                            .map_err(|_| ())
+                            .and_then(move |stream| {
+                                tokio::io::write_all(stream, b"foo".to_vec()).map_err(|_| ())
+                            })
+                            .and_then(|(stream, _)| tokio::io::shutdown(stream).map_err(|_| ()))
+                            .and_then(move |stream| {
+                                tokio::io::read_to_end(stream, Vec::new()).map_err(|_| ())
```
Ralith: Would it be appropriate to use the `read_to_end` defined in this module here?

quinn/src/lib.rs
```diff
@@ -1599,3 +1599,137 @@ impl<T: Read> Future for ReadToEnd<T> {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
Ralith: It'd be nice for this to live in a separate file, I think.

quinn/src/lib.rs
djc: Yup!

quinn/src/lib.rs
djc: Agreed.

djc:
Two reasons to do it now: nice coverage bump, and some automated testing in place before I can dive into multi-threaded endpoints.

PR #155, opened 2018-12-30
Ipv4 fixes

### LINKED ISSUE #154: Fix support for IPv4
It appears that communication between IPv4 endpoints doesn't currently work, while communication between IPv6 endpoints does. Tested with 127.0.0.1 and [::1].

Ralith:
Fix #154

I wasn't sure whether to branch based on the type of the destination address or the type of the local address. I opted for the former since it allows the OS to provide its usual error message when an unsupported address is used.

It's interesting to note that an "Address family not supported by protocol" error currently brings down the whole endpoint. Perhaps send errors should be propagated back to their associated connection instead?

### DIFF
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -58,7 +58,7 @@ mod udp;
 use std::borrow::Cow;
 use std::cell::RefCell;
 use std::collections::{hash_map, VecDeque};
-use std::net::{SocketAddr, SocketAddrV6, ToSocketAddrs};
+use std::net::{SocketAddr, ToSocketAddrs};
 use std::rc::Rc;
 use std::str;
 use std::sync::Arc;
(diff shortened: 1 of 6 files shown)


quinn/src/lib.rs
```diff
@@ -788,8 +785,13 @@ impl Future for Driver {
                         packet,
                         ecn,
                     } => {
+                        let destination = if let Some(v4) = destination.ip().to_ipv4() {
```
djc: Putting this here means we have to do run through this conversion on every send, which seems a little wasteful? As I understand it, the destination will not change very often over time.

Ralith: Right, per connection it doesn't change, just per send call. I suppose we could switch to storing generic `SocketAddr`s in `Connection`, and only use normalized `SocketAddrV6` for hash table keys, it just feels a little silly when that doesn't actually increase expressiveness any.

djc: Maybe, but it seems preferable over converting the address for every send.

djc:
Much better! Probably squash those "Send to..." and "Move SocketAddr..." commits?

Ralith:
Squashed.

PR #157, opened 2018-12-30
Add CI for macOS and Windows

djc:


djc:
macOS seems to work fine (as it should, since I run it locally), but the Windows build doesn't seem to be doing much.

djc:
Apparently Windows builds don't yet work with secrets (such as our code coverage token).

https://travis-ci.community/t/current-known-issues-please-read-this-before-posting-a-new-topic/264

Ralith:
Do we need coverage to run on multiple platforms?

djc:
We don't -- but I haven't figured out a way of disabling the secrets stuff only for one OS.

Ralith:
Aw. Well, just OSX is still a step forwards.

djc:
Limited to just macOS. Can you approve?

PR #166, opened 2019-01-03
Assorted cleanup, panic fix

Ralith:
0-RTT packets weren't being assigned a packet space, which meant no header crypto was selected for them, which lead to a panic in `PartialDecode::finish` which correctly assumed that all standard-form long-header packets have encrypted headers. This corrects the omission and drops 0-RTT packets prior to decryption since rustls isn't giving us 0-RTT keys yet anyway.

### DIFF
--- a/quinn-proto/src/packet.rs
+++ b/quinn-proto/src/packet.rs
@@ -54,12 +54,26 @@ impl PartialDecode {
                 version: VERSION,
                 first,
                 ..
-            } => match LongHeaderType::from_byte(first) {
-                Ok(LongHeaderType::Initial) => Some(SpaceId::Initial),
-                Ok(LongHeaderType::Standard(LongType::Handshake)) => Some(SpaceId::Handshake),
-                _ => None,
+            } => match LongHeaderType::from_byte(first).ok()? {
+                LongHeaderType::Retry => None,
+                LongHeaderType::Initial => Some(SpaceId::Initial),
+                LongHeaderType::Standard(LongType::Handshake) => Some(SpaceId::Handshake),
+                LongHeaderType::Standard(LongType::ZeroRtt) => Some(SpaceId::Data),
             },
             _ => None,
+        }
+    }
+
+    pub fn is_0rtt(&self) -> bool {
+        match self.invariant_header {
+            InvariantHeader::Long {
+                version: VERSION,
+                first,
+                ..
+            } => {
+                LongHeaderType::from_byte(first) == Ok(LongHeaderType::Standard(LongType::ZeroRtt))
+            }
+            _ => false,
         }
     }
 
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -901,6 +901,10 @@ impl Connection {
             );
             return None;
         }
+        if partial_decode.is_0rtt() {
+            debug!(self.log, "dropping 0-RTT packet (currently unimplemented)");
+            return None;
+        }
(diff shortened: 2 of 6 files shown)


quinn-proto/src/packet.rs
```diff
@@ -25,13 +25,58 @@ pub struct PartialDecode {
...
+        bytes: BytesMut,
+        local_cid_len: usize,
+    ) -> Result<(Self, Option<BytesMut>), PacketDecodeError> {
         let mut buf = io::Cursor::new(bytes);
         let invariant_header = InvariantHeader::decode(&mut buf, local_cid_len)?;
-        Ok(Self {
-            invariant_header,
-            buf,
-        })
+        let end = match invariant_header {
```
djc: So clearly all the other parts of this commit are great -- I think you're correct that what I had was not the right place to put that packet splitting.

However, this hunk makes me uncomfortable. I'm thinking this code should be inlined into `InvariantHeader::decode()`, maybe, or there could be a method on `InvariantHeader` that returns the a `Result<usize, PacketDecodeHeader>` with the payload length?

quinn-proto/src/packet.rs
```diff
@@ -138,7 +138,7 @@ impl PartialDecode {
             invariant_header,
             mut buf,
         } = self;
-        let (payload_len, header) = match invariant_header {
+        Ok(match invariant_header {
```
djc: So if I understand correctly, this more or less duplicates some code 5 times in order to special case one of the occurrences? It feels like there might be a better way to do so, maybe by adding another value to the intermediate tuple that is `Some((token_pos, token_len))` for the start and end of the token?

quinn-proto/src/packet.rs
```diff
@@ -54,15 +54,29 @@ impl PartialDecode {
...
-                Ok(LongHeaderType::Initial) => Some(SpaceId::Initial),
-                Ok(LongHeaderType::Standard(LongType::Handshake)) => Some(SpaceId::Handshake),
-                _ => None,
+            } => match LongHeaderType::from_byte(first).ok()? {
+                LongHeaderType::Retry => None,
+                LongHeaderType::Initial => Some(SpaceId::Initial),
+                LongHeaderType::Standard(LongType::Handshake) => Some(SpaceId::Handshake),
+                LongHeaderType::Standard(LongType::ZeroRtt) => Some(SpaceId::Data),
             },
             _ => None,
```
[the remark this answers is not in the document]
Ralith: I think it makes sense for accessors to determine whether a packet satisfies some QUIC 1.0 notion will always return a negative result for non-QUIC 1.0 packets. Additionally, turning these into `Result<bool, UnsupportedVersion>` would just convert every callsite from `if partial.is_foo()` to `if partial.is_foo().unwrap_or(false)`, and I don't think that buys us anything.

quinn-proto/src/packet.rs
Ralith: The length definitely isn't part of the invariant header--we're parsing the data that follows it according to QUIC 1.0 assumptions. I'll try to factor it out into a separate function for clarity, but I don't see a way to reuse the same code elsewhere.

quinn-proto/src/packet.rs
Ralith: The refactoring I did for the above issue makes this a bit cleaner too, I think.

Ralith:
Addressed comments and added some more fixes.

djc:
This is becoming a bit of a grab bag, maybe move the decoalescing and token allocation changes into a separate PR? Then we can merge the rest.

PR #167, opened 2019-01-04
Also allow using ECDSA keys in the example

est31:
ECDSA keys are PKCS#8 formatted and need
the PKCS#8 specific function.

### DIFF
--- a/quinn/examples/server.rs
+++ b/quinn/examples/server.rs
@@ -6,10 +6,11 @@ extern crate slog;
 use std::ascii;
 use std::fmt;
 use std::fs;
+use std::io::{self, Seek, SeekFrom};
 use std::net::SocketAddr;
 use std::path::{self, Path, PathBuf};
 use std::rc::Rc;
-use std::{io, str};
+use std::str;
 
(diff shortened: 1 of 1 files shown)


briansmith:
> ECDSA keys are PKCS#8 formatted and need
> the PKCS#8 specific function.

I've not been tracking this, but I hope that you support the PKCS#8 form of RSA keys too, wherever you would also accept some other form.

Ralith:
It looks like this code just reads all PEM keys supplied, be they PKCS#1 or PKCS#8, which should be fine.

est31:
@briansmith I haven't tried it but I believe that this PR will also add PKCS#8 support for RSA keys.

The main reason why I'm doing this PR is to make keys generated by [rcgen](https://github.com/est31/rcgen) work, which right now can only generate ECDSA keys (RSA keys are blocked on RSA keygen support in ring :p). That's why the title and commit msg talk about ECDSA.

Also, the `rsa_private_keys` function is named a bit badly, it should rather be named `pkcs1_private_keys` or something. But that's a problem of rustls :).

Ralith:
Integrated into #168.

PR #168, opened 2019-01-04
Hide rustls certificate types and parsing methods

Ralith:
This decouples an important part of the API from rustls, and greatly simplifies the task of robustly reading a certificate from disk.

Thanks to @est31 for working out how to wrangle this properly.

### DIFF
--- a/quinn/examples/server.rs
+++ b/quinn/examples/server.rs
@@ -9,11 +9,10 @@ use std::fs;
 use std::net::SocketAddr;
 use std::path::{self, Path, PathBuf};
 use std::rc::Rc;
-use std::{io, str};
+use std::str;
 
-use failure::{err_msg, Error, Fail, ResultExt};
+use failure::{Error, Fail, ResultExt};
 use futures::{Future, Stream};
-use rustls::internal::pemfile;
 use slog::{Drain, Logger};
 use structopt::{self, StructOpt};
 use tokio::runtime::current_thread::Runtime;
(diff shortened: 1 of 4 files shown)


quinn/examples/server.rs
```diff
@@ -97,18 +96,11 @@ fn run(log: Logger, options: Opt) -> Result<()> {
...
-    let cert_chain = {
-        let mut reader = io::BufReader::new(
-            fs::File::open(&options.cert).context("failed to read private key")?,
-        );
-        pemfile::certs(&mut reader).map_err(|_| err_msg("failed to read certificates"))?
-    };
-    server_config.set_certificate(cert_chain, keys[0].clone())?;
+    let key = fs::read(&options.key).context("failed to read private key")?;
+    let key = quinn::PrivateKey::from_pem(&key)?;
+    let cert_chain = fs::read(&options.cert).context("failed to read private key")?;
```
est31: The error is present in the old version as well but this should be changed to "failed to read certificates" or something like that.

est31:
Thanks @Ralith for making this PR! It's a much more general solution than my own PR #167.

Ralith:
> The approach of parsing twice feels a little ugly/fragile, maybe we should have different methods for these, or an enum thing?

The user doesn't typically know or care what type of key is inside their .pem file, and I think it's important to support that gracefully. This is ultimately just TLS file formats being extremely strange.

briansmith:
> > The approach of parsing twice feels a little ugly/fragile, maybe we should have different methods for these, or an enum thing?
> 
> The user doesn't typically know or care what type of key is inside their .pem file, and I think it's important to support that gracefully. This is ultimately just TLS file formats being extremely strange.

They do care, because the private key must match the public key in the certificate that is being used. In general, to be safe, one must always know the public key when loading a private key, and that means knowing which algorithm to use.

Another reason *ring*'s API is the  way it is (and thus Rustls's API is the way it is): In *ring* we avoid supplying any function that references more than one algorithm. This way, if the calling code never references an algorithm, no code for that algorithm will be linked in, without any configuration.

Ralith:
> They do care, because the private key must match the public key in the certificate that is being used. In general, to be safe, one must always know the public key when loading a private key, and that means knowing which algorithm to use.

To be clear, I was referring to PKCS#1 vs PKCS#8, not different algorithms. None of the code in this PR makes any assumptions about or references to cryptographic algorithms.

PR #169, opened 2019-01-04
Fix hasty merge bugs

Ralith:
The one time I get distracted and forget to wait for CI...


djc:
Happens to the best of us. 😊

PR #174, opened 2019-01-04
[WIPCoalescing api

djc:


quinn-proto/src/packet.rs
```diff
@@ -92,20 +82,79 @@ impl PartialDecode {
...
+                src_cid,
+                token,
+                len,
+            } => {
+                let number = Self::decrypt_header(&mut buf, header_crypto.unwrap())?;
+                if len < number.len() as u64 {
+                    return Err(PacketDecodeError::InvalidHeader(
+                        "packet number longer than packet",
+                    ));
+                }
```
Ralith: Let's get rid of this check while we're in here, since it's subsumed by decode_header.

quinn-proto/src/packet.rs
```diff
@@ -92,20 +82,79 @@ impl PartialDecode {
...
+                dst_cid,
+                src_cid,
+                len,
+            } => {
+                let number = Self::decrypt_header(&mut buf, header_crypto.unwrap())?;
+                if len < number.len() as u64 {
+                    return Err(PacketDecodeError::InvalidHeader(
+                        "packet number longer than packet",
+                    ));
+                }
```
Ralith: Redundant check.

quinn-proto/src/packet.rs
```diff
@@ -92,20 +82,79 @@ impl PartialDecode {
...
+                    orig_dst_cid,
+                },
+                false,
+            ),
+            Short { spin, dst_cid, .. } => {
                 if !buf.has_remaining() {
                     return Err(PacketDecodeError::InvalidHeader(
                         "header ends before packet number",
                     ));
                 }
```
Ralith: Redundant check

quinn-proto/src/packet.rs
```diff
@@ -25,9 +25,14 @@ pub struct PartialDecode {
...
 impl PartialDecode {
-    pub fn new(bytes: BytesMut, local_cid_len: usize) -> Result<Self, PacketDecodeError> {
+    pub fn new(bytes: BytesMut, local_cid_len: usize) -> Result<(Self, Option<BytesMut>), PacketDecodeError> {
         let mut buf = io::Cursor::new(bytes);
         let plain_header = PlainHeader::decode(&mut buf, local_cid_len)?;
+        let rest = if let Some(payload_len) = plain_header.payload_len() {
+            
+        } else {
+
+        };
```
Ralith: `plain_header.payload_len().map(|len| ...)`?

quinn-proto/src/packet.rs
```diff
@@ -450,38 +427,65 @@ impl PartialEncode {
...
         match self {
+            Initial { dst_cid, .. } => *dst_cid,
             Long { dst_cid, .. } => *dst_cid,
+            Retry { dst_cid, .. } => *dst_cid,
             Short { dst_cid, .. } => *dst_cid,
+            VersionNegotiate { dst_cid, .. } => *dst_cid,
         }
     }
 
     fn decode<R: Buf>(buf: &mut R, local_cid_len: usize) -> Result<Self, PacketDecodeError> {
```
Ralith: To get zero-copy tokens at this early stage, this function will need to become monomorphic, taking a `(&mut? )BytesMut` and consuming the portion of it that was read so that the `Bytes` can be pulled out. This will require changing the current `HeaderCrypto` API to take `first: &mut u8, pn: &mut [u8]` or similar instead of a single contiguous slice, and the inclusion of a `first` field in `PlainHeader::Initial`/`Long`.

Ralith: Or maybe `HeaderCrypto` should include parsing, so it can take value/shared reference arguments and return flags/pn?

Ralith: To be clear, I don't consider this a blocker, but it'd be neat.

djc: To be honest, all of those options sound like a whole lot of complexity just to avoid a single allocation. I think we should probably do some benchmarks before we do those kinds of optimization.

djc:
@Ralith take another look?

djc:
Hmm, wait -- it looks like the `decode_coalesced` test is not triggering correctly, so decoalescing isn't verified to be working right now.

quinn-proto/src/packet.rs
```diff
@@ -25,10 +25,21 @@ pub struct PartialDecode {
...
+    ) -> Result<(Self, Option<BytesMut>), PacketDecodeError> {
         let mut buf = io::Cursor::new(bytes);
         let plain_header = PlainHeader::decode(&mut buf, local_cid_len)?;
-        Ok(Self { plain_header, buf })
+        let rest = plain_header.payload_len().and_then(|len| {
+            let packet_len = (buf.position() + len) as usize;
+            if buf.get_ref().len() > packet_len {
+                Some(buf.get_mut().split_off(packet_len))
+            } else {
+                None
```
Ralith: I think we should return an error if `packet_len > buf.get_ref().len()`.

djc: Improved it -- though not quite the way you suggested it.

djc:
Now also includes a different take on zero-allocation tokens, which is a bit simpler IMO.

PR #177, opened 2019-01-05
Implement client address migration

Ralith:


Ralith:
This doesn't implement responding to path validation probes from inactive paths along the same path they were sent from, and assumes that packets will always be delivered on the same local address, which isn't necessarily true for sockets bound to the wildcard address. Still, it's a step in the right direction.

djc:
Can you fix the conflict?

Ralith:
Rebased.

PR #181, opened 2019-01-06
Implement 0-RTT data

Ralith:
This *should* work, but it will be difficult to test until rustls supports 0-RTT in servers. Something's weird with picoQUIC interop too; might have introduced an issue with Handshake encryption. Bulk of this should be more or less right, though.

### DIFF
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -28,7 +28,7 @@ lazy_static = "1"
 orion = "0.12"
 rand = "0.6"
 ring = "0.13"
-rustls = { git = "https://github.com/Ralith/rustls.git", rev = "b4b9b1b8e080129c418bddb2e110ee1780b16793", features = ["quic"] }
+rustls = { git = "https://github.com/Ralith/rustls.git", rev = "4e771422a8dbd5459c523a1f1b3be0772e813d07", features = ["quic"] }
 slab = "0.4"
 slog = "2.2"
 webpki = "0.18"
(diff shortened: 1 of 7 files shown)


Ralith:
One major remaining TODO is that a client sending 0-RTT must operate under the transport parameters that were supplied by the server during the session that is being resumed. To accomplish this, I think we need to wrap the session tickets emitted/consumed by rustls, so that an application storing a session ticket by any mechanism is necessarily, transparently storing the associated transport parameters as well. I think the easiest approach to that is silently wrapping the `session_persistence` member of the input `rustls::ClientConfig` in an adapter.

djc:
That's a big commit. How hard would it be to separate the structural changes (introduction of the new space type) from the rest of the 0-RTT implementation.

I'm also a bit wary of the suggested transparency in wrapping session tickets. If we need something like that, I think we should try to make it as explicit and obvious in the API as possible, though hopefully still ergonomic to use.

Ralith:
> How hard would it be to separate the structural changes (introduction of the new space type) from the rest of the 0-RTT implementation.

Done.

> I'm also a bit wary of the suggested transparency in wrapping session tickets. If we need something like that, I think we should try to make it as explicit and obvious in the API as possible, though hopefully still ergonomic to use.

Can you elaborate on the drawbacks you see here? This seems entirely consistent with how tickets are normally used to me; rustls itself, for example, encodes all kinds of extra state into them already.

djc:
> Can you elaborate on the drawbacks you see here? This seems entirely consistent with how tickets are normally used to me; rustls itself, for example, encodes all kinds of extra state into them already.

Hmm, yeah, if you put it like that. I just want to make sure that our changes don't break things for users that fiddle with the session tickets.

Ralith:
Session tickets are expressly opaque in TLS1.3, so anyone fiddling with them deserves what they get.

djc:
Looks okay!

djc:
(Needs rebase.)

Ralith:
Rebased.

PR #184, opened 2019-01-08
Implement dynamic stream ID flow control

### LINKED ISSUE #62: Dynamic stream ID flow control
Currently, the application must specify a maximum number of concurrently operating remotely-initiated streams at endpoint construction time. This is more strict than necessary. A connection's streams can be handled in the same manner as an endpoint's connections are: a small, configurable window of new streams is buffered by the implementation, advancing whenever the application consumes an entry to begin performing I/O on it. This would allow applications to dynamically and implicitly control the amount of stream-level concurrency they want, for example using `FuturesUnordered`.

Ralith:
This makes the implementation significantly more flexible and responsive without requiring any particular effort from the application. I'd be interested to hear any thoughts on enabling incoming streams by default; it seems like doing otherwise is a massive gotcha.

Fixes #62

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -83,6 +83,8 @@ pub struct Connection {
     prev_crypto: Option<PrevCrypto>,
     /// Latest PATH_CHALLENGE token issued to the peer along the current path
     path_challenge: Option<u64>,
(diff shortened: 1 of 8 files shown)


Ralith:
> If we for example had different methods to initialize a Config but no default(), we could force users to make a choice while still offering pretty good ergonomics.

Ooh, I like that idea!

PR #187, opened 2019-01-10
Use rcgen rather than a public cert in tests/examples

Ralith:
This adopts @briansmith's advice for externalizing dangerous configuration, documents its use in the README, and makes use of rcgen to simplify the examples and tests. Thanks to @est31 for making this possible!

Fixes https://github.com/djc/quinn/issues/58.


est31:
@djc I think where @Ralith removed `webpki-roots` from, it actually wasn't needed. It's still present in the quinn crate.

PR #190, opened 2019-01-10
Update to ring 0.14

est31:
Also updates the crates that depend on ring to newer versions


briansmith:
Note that *ring* 0.14 has new APIs specifically for QUIC header protection: https://docs.rs/ring/0.14.0/ring/aead/quic/index.html.

Ralith:
Is there a reason to rush this before the intermediate crates get bumped properly?

est31:
@Ralith I think it's a good idea to open PRs as fast as possible, even with unreleased stuff. Because then you still have a chance to "halt the presses" if there turns out to be a problem with the new version. @briansmith filed all the pr's to @ctz 's repos right away as well. Merging is a different matter, that can also wait until there's been a release.

@briansmith interesting, thanks for the pointer! Will try to have a look at integrating the new code. Hopefully some stuff can be deleted from quinn.

briansmith:
> @briansmith filed all the pr's to @ctz 's repos right away as well.

In my experience, @ctz usually only processes PRs on the weekend, so I did them all at once so it could be done in one weekend instead of 2. That's all.

Ralith:
@briansmith re: header protection helpers: would it make sense to provide an accessor for sample sizes? [The draft](https://quicwg.org/base-drafts/draft-ietf-quic-tls.html#hp-sample) seems to imply that they might be more diverse in the future, so the current implementation takes care not to hardcode it.

briansmith:
> @briansmith re: header protection helpers: would it make sense to provide an accessor for sample sizes? [The draft](https://quicwg.org/base-drafts/draft-ietf-quic-tls.html#hp-sample) seems to imply that they might be more diverse in the future, so the current implementation takes care not to hardcode it.

Yes, I think that makes a lot of sense. Thanks for suggesting that. It would be great if somebody could contribute a PR that adds that as I'm busy traveling for the next several days. 

I'm also thinking that maybe we should add an accessor to get the `ring::aead::Algorithm` from the the `ring::aead::quic::Algorithm`, so that you don't need to keep track of a pair of things.

Ralith:
I'll go ahead and draft it.

est31:
Rebased.

PR #195, opened 2019-01-15
Assorted fixes

Ralith:
Incidentals from the async/await prototype.


Ralith:
> Can we ? instead of expect here?

I'm fine with that, but I think it belongs in a separate commit.

PR #198, opened 2019-01-15
[WIP] Async/await style API

Ralith:
Currently severely broken due to bugs in tokio's `DelayQueue` and in need of considerable testing once that's fixed, but there's no reason for that to get in the way of design/API feedback. I'm particularly interested in thoughts on the 0-RTT API, which notably does not actually involve any futures or async.

### DIFF
--- a/interop/Cargo.toml
+++ b/interop/Cargo.toml
@@ -15,7 +15,7 @@ tokio = "0.1.6"
 tokio-current-thread = "0.1"
 tokio-reactor = "0.1"
 rustls = { version = "0.15", features = ["dangerous_configuration"] }
-failure = "0.1"
+err-ctx = "0.2"
 slog = "2.2"
 futures-preview = { version = "0.3.0-alpha.11", features = ["compat"] }
 webpki = "0.19"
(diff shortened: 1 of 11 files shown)


zmanian:
presumably this is the issue in `DelayQueue` https://github.com/tokio-rs/tokio/issues/854

Ralith:
Yes, that's the thing currently blocking further testing with this.

Ralith:
Next one: https://github.com/tokio-rs/tokio/issues/868

Ralith:
With https://github.com/tokio-rs/tokio/pull/871, this now appears to be usable! All that's left is to wait for stabilization, and in the meantime I encourage people to try out the new API and see if there are any design flaws.

Ralith:
The new approach assumes that restrictions on 0-RTT frames will be relaxed as proposed in https://github.com/quicwg/base-drafts/pull/2355.

djc:
Should we just close this PR and/or convert it to issues with things we can still scavenge at some point? At this point I feel like redoing the work is probably easier than rebasing this.

Ralith:
Yeah, I believe we've successfully extracted all the significant architectural changes here save for the conversion to borrowing futures, which won't be a lot of work to reproduce.

djc:
I guess we might still want the failure -> err-ctx conversion?

Ralith:
The interop tool was completely rewritten with async/await so it's not really a portable change. Something to consider when the time to do that for real comes. Very mechanical, anyway.

PR #200, opened 2019-01-16
Use *ring* for header protection (fixes #196)

### LINKED ISSUE #196: Switch between AES-NI and software implementation at runtime rather than at compile time
Currently, the aes crate switches between aesni and aes-soft using a compile time switch. But most times you want to compile a binary that runs on computers that don't have AES-NI so you won't turn on aesni at compile time. Therefore, I suggest runtime switching which can be acheived by enabling the `nocheck` feature of the aesni crate and checking manually using the `is_x86_feature_detected` macro from std.

Using a fast AES implementation is particularly important for games as here low-latency is very important.

Further reading:

* https://github.com/RustCrypto/block-ciphers/pull/22
* https://github.com/RustCrypto/block-ciphers/issues/25#issuecomment-412787052
* https://doc.rust-lang.org/stable/std/macro.is_x86_feature_detected.html

djc:


djc:
As usual, *ring* doesn't offer access to private key bytes. Should we create a separate stage in our API to check that they are correct? I think we have sufficient coverage through full header encryption/decryption.

briansmith:
> As usual, _ring_ doesn't offer access to private key bytes. Should we create a separate stage in our API to to enable tests for their correctness? I think we have sufficient coverage through full header encryption/decryption.

Check out https://github.com/briansmith/ring/blob/master/tests/quic_tests.rs, which shows how to test these kinds of things.

In this case, your old test was like this:

```rust
assert_eq!(
            client_header_key,	            client_header_key,
            HeaderKey::AesEcb128(hex!("0edd982a6ac527f2eddcbb7348dea5d7"))	            HeaderKey::AesEcb128(hex!("0edd982a6ac527f2eddcbb7348dea5d7"))
```

Instead of testing that the key is equal to some value, test that the result of using the key on a particular input has a particular output.

djc:
We already have tests like that. 👍

PR #203, opened 2019-01-17
Clearer handling of ConnectionHandles

djc:
I opted for using `ch` everywhere except as a field on the `ConnectionInner` type in the quinn crate: since that represents a `Connection`, I felt `handle` made the code easier to read.

Also made it so the type is a little more abstracted.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -3141,15 +3141,6 @@ struct SentPacket {
     retransmits: Retransmits,
 }
 
-#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
-pub struct ConnectionHandle(pub usize);
-
-impl From<ConnectionHandle> for usize {
-    fn from(x: ConnectionHandle) -> usize {
-        x.0
-    }
-}
-
 /// Ensures we can always fit all our ACKs in a single minimum-MTU packet with room to spare
 const MAX_ACK_BLOCKS: usize = 64;
 
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -835,7 +835,9 @@ impl Connection {
         self.set_params(params)?;
         self.write_tls();
         self.init_0rtt();
-        self.handle_coalesced(now, remote, ecn, remaining);
+        if let Some(data) = remaining {
+            self.handle_coalesced(now, remote, ecn, data);
+        }
         Ok(())
     }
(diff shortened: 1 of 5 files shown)


djc:
Sorry, this is a pretty big rebase hazard. Still makes it easier to read though, IMO!

djc:
(Veering into territory of further decoupling of `Endpoint` and `Connection`.)

quinn-proto/src/connection.rs
```diff
@@ -978,6 +982,40 @@ impl Connection {
...
+        remote: SocketAddr,
+        ecn: Option<EcnCodepoint>,
+        mut remaining: Option<BytesMut>,
+    ) {
+        let loc_cid_len = self.loc_cids.iter().next().unwrap().1.len();
+        while let Some(data) = remaining {
+            match PartialDecode::new(data, loc_cid_len) {
+                Ok((partial_decode, rest)) => {
+                    remaining = rest;
+                    self.handle_decode(now, remote, ecn, partial_decode);
```
Ralith: > Receivers SHOULD ignore any subsequent packets with a different Destination Connection ID than the first packet in the datagram.

Something for future work here, I think.

quinn-proto/src/endpoint.rs
```diff
@@ -1031,13 +1031,7 @@ impl slog::Value for Timer {
...
 
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
-pub struct ConnectionHandle(pub usize);
-
-impl From<ConnectionHandle> for usize {
-    fn from(x: ConnectionHandle) -> usize {
-        x.0
-    }
-}
+pub struct ConnectionHandle(usize);
```
Ralith: I'm making pretty heavy use of this in #198 to skip an unnecessary `HashMap` at the higher level, so I'll probably want to restore it there.

quinn-proto/src/connection.rs
```diff
@@ -978,6 +982,40 @@ impl Connection {
...
+    }
+
+    fn handle_coalesced(
+        &mut self,
+        now: u64,
+        remote: SocketAddr,
+        ecn: Option<EcnCodepoint>,
+        mut remaining: Option<BytesMut>,
+    ) {
+        let loc_cid_len = self.loc_cids.iter().next().unwrap().1.len();
```
Ralith: Couldn't this just be `self.config.local_cid_len`, as it was originally?

PR #205, opened 2019-01-19
Event-based communication from Connections to Endpoints

djc:
Please have a look at this early attempt.

(Handling these events in `poll_io()` may be somewhat unnatural, but was nice because it's part of the test driver code already.)

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -69,6 +69,7 @@ pub struct Connection {
     lost_packets: u64,
     io: IoQueue,
     events: VecDeque<Event>,
+    endpoint_events: VecDeque<EndpointEvent>,
     /// Number of local connection IDs that have been issued in NEW_CONNECTION_ID frames.
     cids_issued: u64,
     /// Outgoing spin bit state
@@ -214,6 +215,7 @@ impl Connection {
             lost_packets: 0,
             io: IoQueue::new(),
             events: VecDeque::new(),
+            endpoint_events: VecDeque::new(),
             cids_issued: 0,
             spin: false,
             spaces: [initial_space, PacketSpace::new(), PacketSpace::new()],
@@ -308,6 +310,11 @@ impl Connection {
         }
 
         None
+    }
+
+    /// Return endpoint-facing events
+    pub fn poll_endpoint_events(&mut self) -> Option<EndpointEvent> {
+        self.endpoint_events.pop_front()
     }
 
     fn on_packet_sent(
@@ -1101,7 +1108,6 @@ impl Connection {
             connection = packet.header.dst_cid(),
         );
         let was_closed = self.state.is_closed();
-
         let stateless_reset = self.params.stateless_reset_token.map_or(false, |token| {
             packet.payload.len() >= RESET_TOKEN_SIZE
                 && packet.payload[packet.payload.len() - RESET_TOKEN_SIZE..] == token
@@ -1303,7 +1309,10 @@ impl Connection {
                                 }
                             }
                             self.set_params(params)?;
+                            self.endpoint_events
+                                .push_back(EndpointEvent::NeedIdentifiers);
                         }
+
                         self.events.push_back(Event::Connected);
                         self.state = State::Established;
                         trace!(self.log, "established");
@@ -3216,6 +3225,12 @@ struct SentPacket {
 
 /// Ensures we can always fit all our ACKs in a single minimum-MTU packet with room to spare
 const MAX_ACK_BLOCKS: usize = 64;
+
+/// Events to be sent to the Endpoint
+#[derive(Clone, Debug)]
+pub enum EndpointEvent {
+    NeedIdentifiers,
+}
 
 /// I/O operations to be immediately executed the backend.
 #[derive(Debug)]
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, HashMap, VecDeque};
+use std::collections::{BTreeMap, VecDeque};
 use std::net::SocketAddr;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
@@ -36,7 +36,6 @@ pub struct Connection {
     config: Arc<TransportConfig>,
     rng: OsRng,
     tls: TlsSession,
-    loc_cids: HashMap<u64, ConnectionId>,
     /// The CID we initially chose, for use during the handshake
     handshake_cid: ConnectionId,
     rem_cid: ConnectionId,
@@ -183,8 +182,6 @@ impl Connection {
                 Stream::new_bi(),
             );
         }
-        let mut loc_cids = HashMap::new();
-        loc_cids.insert(0, loc_cid);
         let state = State::Handshake(state::Handshake {
             rem_cid_set: side.is_server(),
             token: None,
@@ -194,7 +191,6 @@ impl Connection {
             endpoint_config,
             rng,
             tls,
-            loc_cids,
             handshake_cid: loc_cid,
             rem_cid,
             rem_handshake_cid: rem_cid,
@@ -290,8 +286,8 @@ impl Connection {
             }
         }
 
-        if let Some(cid) = self.io.retired_cids.pop() {
-            return Some(Io::RetireConnectionId { connection_id: cid });
+        if let Some(seq) = self.io.retired_cids.pop() {
+            return Some(Io::RetireConnectionId(seq));
         }
 
         None
@@ -1457,10 +1453,9 @@ impl Connection {
         Ok(())
     }
 
-    pub fn issue_cid(&mut self, cid: ConnectionId) {
+    pub fn issue_cid(&mut self, sequence: u64, cid: ConnectionId) {
         let token = reset_token_for(&self.endpoint_config.reset_key, &cid);
         self.cids_issued += 1;
-        let sequence = self.cids_issued;
         self.space_mut(SpaceId::Data)
             .pending
             .new_cids
@@ -1469,7 +1464,6 @@ impl Connection {
                 sequence,
                 reset_token: token,
             });
-        self.loc_cids.insert(self.cids_issued, cid);
     }
 
     fn process_payload(
@@ -1776,15 +1770,7 @@ impl Connection {
                             "RETIRE_CONNECTION_ID for unissued sequence number",
                         ));
                     }
-                    if let Some(old) = self.loc_cids.remove(&sequence) {
-                        trace!(
-                            self.log,
-                            "peer retired CID {sequence}: {id}",
-                            sequence = sequence,
-                            id = old
-                        );
-                        self.io.retired_cids.push(old);
-                    }
+                    self.io.retired_cids.push(sequence);
                 }
                 Frame::NewConnectionId(frame) => {
                     trace!(
@@ -2856,11 +2842,6 @@ impl Connection {
         self.side
     }
 
-    /// The `ConnectionId`s defined for this Connection locally.
-    pub fn loc_cids(&self) -> impl Iterator<Item = &ConnectionId> {
-        self.loc_cids.values()
-    }
-
     /// The `ConnectionId` defined for this Connection by the peer.
     pub fn rem_cid(&self) -> ConnectionId {
         self.rem_cid
@@ -3241,8 +3222,8 @@ const MAX_ACK_BLOCKS: usize = 64;
 pub enum Io {
     /// Stop or (re)start a timer
     TimerUpdate(TimerUpdate),
-    /// Stop routing `connection_id` to this `Connection`
-    RetireConnectionId { connection_id: ConnectionId },
+    /// Stop routing connection id for this sequence number to this `Connection`
+    RetireConnectionId(u64),
 }
 
 /// Encoding of I/O operations to emit on upcoming `poll_io` calls
@@ -3257,7 +3238,7 @@ struct IoQueue {
     /// Note that this ordering exactly matches the values of the `Timer` enum for convenient
     /// indexing.
     timers: [Option<TimerSetting>; Timer::COUNT],
-    retired_cids: Vec<ConnectionId>,
+    retired_cids: Vec<u64>,
 }
 
 impl IoQueue {
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -1005,37 +1005,35 @@ impl Connection {
     pub fn handle_event(&mut self, event: ConnectionEvent) {
         use self::ConnectionEvent::*;
         match event {
+            Datagram {
+                now,
+                remote,
+                ecn,
+                first_decode,
+                remaining,
+            } => {
+                if remote != self.remote && self.side.is_client() {
+                    trace!(
+                        self.log,
+                        "discarding packet from unknown server {address}",
+                        address = format!("{}", remote)
+                    );
+                    return;
+                }
+
+                self.total_recvd = self.total_recvd.wrapping_add(first_decode.len() as u64);
+
+                self.handle_decode(now, remote, ecn, first_decode);
+                if let Some(data) = remaining {
+                    self.handle_coalesced(now, remote, ecn, data);
+                }
+            }
             NewIdentifiers(ids) => {
                 ids.into_iter().for_each(|(seq, cid)| {
                     self.issue_cid(seq, cid);
                 });
             }
             Timer(now, timer) => self.timeout(now, timer),
-        }
-    }
-
-    pub fn handle_dgram(
-        &mut self,
-        now: Instant,
-        remote: SocketAddr,
-        ecn: Option<EcnCodepoint>,
-        first_decode: PartialDecode,
-        remaining: Option<BytesMut>,
-    ) {
-        if remote != self.remote && self.side.is_client() {
-            trace!(
-                self.log,
-                "discarding packet from unknown server {address}",
-                address = format!("{}", remote)
-            );
-            return;
-        }
-
-        self.total_recvd = self.total_recvd.wrapping_add(first_decode.len() as u64);
-
-        self.handle_decode(now, remote, ecn, first_decode);
-        if let Some(data) = remaining {
-            self.handle_coalesced(now, remote, ecn, data);
         }
     }
 
@@ -3244,6 +3242,13 @@ const MAX_ACK_BLOCKS: usize = 64;
 
 /// Events to be sent to the Connection
 pub enum ConnectionEvent {
+    Datagram {
+        now: Instant,
+        remote: SocketAddr,
+        ecn: Option<EcnCodepoint>,
+        first_decode: PartialDecode,
+        remaining: Option<BytesMut>,
+    },
     NewIdentifiers(Vec<(u64, ConnectionId)>),
     Timer(Instant, Timer),
 }
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -515,7 +515,7 @@ impl Connection {
         self.space_mut(space).pending_acks.subtract(&info.acks);
     }
 
-    pub fn timeout(&mut self, now: Instant, timer: Timer) {
+    fn timeout(&mut self, now: Instant, timer: Timer) {
         match timer {
             Timer::Close => {
                 self.state = State::Drained;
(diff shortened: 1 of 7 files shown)


quinn-proto/src/endpoint.rs
```diff
@@ -104,6 +104,9 @@ impl Endpoint {
         loop {
             let &ch = self.dirty_conns.iter().next()?;
             loop {
+                while let Some(event) = self.connections[ch].poll_endpoint_events() {
+                    self.handle_event(ch, event);
+                }
```
Ralith: The tokio bindings currently assume that application-facing events are only queued by `timeout` and `handle`; outgoing IO is not expected to expose new information to the application. I see that putting this in `poll` would be awkward wrt. the test cod as it stands; as a temporary measure until the decoupling is complete, maybe we could instead put it in a method that gets called by both `handle` and `timeout`?

quinn-proto/src/connection.rs
```diff
@@ -1413,6 +1428,15 @@ impl Connection {
         Ok(())
     }
 
+    fn send_ready_events(&mut self) {
+        if self.side.is_server() {
+            self.endpoint_events.push_back(EndpointEvent::HandshakeEnded);
+        }
+        if self.config.local_cid_len != 0 && !self.state.is_closed() {
+            self.endpoint_events.push_back(EndpointEvent::IdentifiersNeeded);
```
Ralith: Instead of having a separate case for this, we could rely on `issue_cid` being a harmless no-op on a closed connection and key off `HandshakeEnded` alone, but I'm okay with this approach if you prefer it.

quinn-proto/src/connection.rs
```diff
@@ -3186,6 +3210,12 @@ struct SentPacket {
 /// Ensures we can always fit all our ACKs in a single minimum-MTU packet with room to spare
 const MAX_ACK_BLOCKS: usize = 64;
 
+/// Events to be sent to the Endpoint
+pub enum EndpointEvent {
+    HandshakeEnded,
+    IdentifiersNeeded,
```
Ralith: `Io::RetireConnectionId` should probably be moved here too.

quinn-proto/src/connection.rs
```diff
@@ -3186,6 +3210,12 @@ struct SentPacket {
 /// Ensures we can always fit all our ACKs in a single minimum-MTU packet with room to spare
 const MAX_ACK_BLOCKS: usize = 64;
 
+/// Events to be sent to the Endpoint
+pub enum EndpointEvent {
```
Ralith: Eventually I'd like this to be wrapped inside an opaque struct wherever it occurs in a public API; unlike `Io` and `Event`, external code should never need to see into this enum.

quinn-proto/src/connection.rs
```diff
@@ -1304,7 +1310,13 @@ impl Connection {
                             }
                             self.set_params(params)?;
                         }
+
                         self.events.push_back(Event::Connected);
+                        self.endpoint_events.push_back(EndpointEvent::Connected {
+                            handshake_complete: self.side.is_server(),
+                            identifiers_needed: self.endpoint_config.local_cid_len != 0,
```
Ralith: Couldn't this check be performed in the `Endpoint`?

quinn-proto/src/connection.rs
```diff
@@ -1304,7 +1310,13 @@ impl Connection {
                             }
                             self.set_params(params)?;
                         }
+
                         self.events.push_back(Event::Connected);
+                        self.endpoint_events.push_back(EndpointEvent::Connected {
+                            handshake_complete: self.side.is_server(),
```
Ralith: The handshake is complete regardless of side here, so this naming is a bit confusing.

djc:
I think this more or less completes the `Connection`-to-`Endpoint` side of things.

I'll make the `Endpoint`-to-`Connection` access more coarse-grained in a follow-up PR.

quinn-proto/src/connection.rs
```diff
@@ -3251,6 +3281,11 @@ const MAX_ACK_BLOCKS: usize = 64;
 pub enum EndpointEvent {
     ClientReady,
     Incoming,
+    Closed {
+        init_cid: Option<ConnectionId>,
```
Ralith: `Connection` never reads this value, so I think it would now make more sense for it to be stored in the `Endpoint`'s connection table directly rather than kept inside the `Connection` and passed back in an event.

quinn-proto/src/endpoint.rs
```diff
@@ -109,7 +109,15 @@ impl Endpoint {
...
                 }
+                let done = if let EndpointEvent::Closed { .. } = &event {
+                    true
+                } else {
+                    false
+                };
                 self.handle_event(ch, event);
+                if done {
+                    break;
+                }
```
Ralith: Is this necessary? Are there cases where extra endpoint events might trail after `Closed`?

quinn-proto/src/connection.rs
```diff
@@ -529,13 +529,19 @@ impl Connection {
         match timer {
             Timer::Close => {
                 self.state = State::Drained;
+                if self.app_closed {
```
Ralith: Similarly to `init_cid`, I think `app_closed` is the `Endpoint`'s' concern, and could be removed from `Connection` entirely; instead, the `Connection` oculd notify the `Endpoint` whenever it becomes `Drained`, and the `Endpoint` would dispose of the `Connection` when it has both become `Drained` and been explicitly closed by the user.

quinn-proto/src/connection.rs
```diff
@@ -860,23 +873,38 @@ impl Connection {
...
+            let len = packet.header_data.len() + packet.payload.len();
+            self.on_packet_authenticated(
+                now,
+                SpaceId::Initial,
+                ecn,
+                Some(packet_number),
+                false,
+                len,
+            );
+            self.process_early_payload(now, packet)?;;
```
```suggestion
            self.process_early_payload(now, packet)?;
```

quinn-proto/src/endpoint.rs
```diff
@@ -99,6 +103,18 @@ impl Endpoint {
         None
     }
 
+    pub(crate) fn handle_endpoint_events(&mut self) {
+        while let Some(&ch) = self.endpoint_eventful_conns.iter().next() {
+            while let Some(event) = self.connections[ch].poll_endpoint_events() {
+                if let EndpointEvent::Incoming = &event {
+                    self.incoming.push_back(ch);
+                }
```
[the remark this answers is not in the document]
djc: This is substantially reworked, have another look.

quinn-proto/src/connection.rs
```diff
@@ -876,6 +883,9 @@ impl Connection {
         if let Some(data) = remaining {
             self.handle_coalesced(now, remote, ecn, data);
         }
+        if self.has_1rtt() {
+            self.endpoint_events.push_back(EndpointEvent::Incoming);
+        }
```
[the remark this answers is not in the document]
djc: Yeah, I think that makes sense. Let me know what you think of this take.

This case is special in a sense in that before the completion of `handle_initial()` is actually the only point in time that `Endpoint` will own `Connection` even in the task-per-`Connection` future.

quinn-proto/src/connection.rs

quinn-proto/src/endpoint.rs
djc: With how the loop is setup, there's always at least one `poll_endpoint_events()` for the next loop iteration which we have to prevent, right?

quinn-proto/src/connection.rs

quinn-proto/src/connection.rs

quinn-proto/src/connection.rs
```diff
@@ -867,26 +873,38 @@ impl Connection {
...
+            self.write_tls();
+            self.init_0rtt();
+            if let Some(data) = remaining {
+                self.handle_coalesced(now, remote, ecn, data);
+            }
+            if self.has_1rtt() {
+                self.endpoint_events.push_back(EndpointEvent::Incoming);
+            }
+            Ok(())
+        };
```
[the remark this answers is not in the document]
djc: This is no longer a problem.

quinn-proto/src/connection.rs
```diff
@@ -3243,6 +3260,22 @@ struct SentPacket {
...
+    NewIdentifiers([Option<ConnectionId>; LOC_CID_COUNT]),
+}
+
+/// Events to be sent to the Endpoint
+#[derive(Clone, Debug)]
+pub enum EndpointEvent {
+    ClientReady,
+    Incoming,
+    Closed {
+        loc_cids: Vec<ConnectionId>,
```
Ralith: Could this list be maintained inside `endpoint::ConnectionMeta` rather than being sent back and forth?

quinn-proto/src/connection.rs
```diff
@@ -3243,6 +3260,22 @@ struct SentPacket {
...
 const MAX_ACK_BLOCKS: usize = 64;
 
+/// Events to be sent to the Connection
+pub enum ConnectionEvent {
+    NewIdentifiers([Option<ConnectionId>; LOC_CID_COUNT]),
+}
+
+/// Events to be sent to the Endpoint
+#[derive(Clone, Debug)]
+pub enum EndpointEvent {
```
Ralith: Should we move what is currently `Io::RetireConnectionId` into this enum?

quinn-proto/src/connection.rs
```diff
@@ -1859,7 +1852,7 @@ impl Connection {
         };
         if stream.index() >= *next {
             *next = stream.index() + 1;
-            self.stream_opened = true;
+            self.events.push_back(Event::StreamOpened);
```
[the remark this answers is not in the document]
djc: I came up with a way to remove the `Handshaking` event (which this PR renamed to `Incoming` in a previous iteration) by returning an `Option<ConnectionHandle>` from `Endpoint::handle()`. This was informed by changes not yet included here where that return type will actually change to `Option<(ConnectionHandle, Connection)>`. This (done in 839f616a63221518fe430d8cf10a82b1fdb26bfa) then removed the need for this change, which indeed was done to allow events to be emitted in order.

quinn-proto/src/connection.rs
```diff
@@ -3247,6 +3260,11 @@ struct SentPacket {
 /// Ensures we can always fit all our ACKs in a single minimum-MTU packet with room to spare
 const MAX_ACK_BLOCKS: usize = 64;
 
+/// Events to be sent to the Connection
+pub enum ConnectionEvent {
+    NewIdentifiers([Option<ConnectionId>; LOC_CID_COUNT]),
```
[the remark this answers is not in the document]
djc: Nothing; previously we only used it for `ConnectionId`, and we don't actually use all that much of the `ArrayVec` code, which made me prefer going the literal array route. If we now have more use cases for `ArrayVec` and/or need more of its implementation, we can certainly reconsider that decision (though I'd prefer not to do so as part of this PR -- and I don't think this particular use needs `ArrayVec` all that much).

quinn-proto/src/connection.rs
djc: Good idea, see b82110cb3a33a9db7d83f0512ec4f0fc278dd961.

quinn-proto/src/connection.rs
djc: Yup, see d27cd156790cbac8af8da12f693dba12bd8b5564 and the cleanup in b676a9fd863edd8325f26bfdf5ad7beba3acbfd9.

quinn/src/tests.rs
```diff
@@ -25,7 +25,7 @@ fn echo_v4() {
 }
 
 #[test]
-#[cfg(any(target_os = "linux", target_os = "macos"))] // Dual-stack sockets aren't the default anywhere else.
```
djc: We can ditch this again before we merge it, it's just there to help me test the PR.

djc:
(The dates on the commits are still highly non-linear. Note that the Commits view of the PR does have a list of commits that is topologically sorted rather than chronologically sorted.)

quinn-proto/src/connection.rs
```diff
@@ -515,7 +515,7 @@ impl Connection {
         self.space_mut(space).pending_acks.subtract(&info.acks);
     }
 
-    pub fn timeout(&mut self, now: Instant, timer: Timer) {
+    fn timeout(&mut self, now: Instant, timer: Timer) {
```
Ralith: I don't think there's any benefit to routing timers through the `Endpoint`, so keeping this public would allow a hypothetical decoupled configuration to e.g. run separate timer wheels for separate threads.

quinn-proto/src/endpoint.rs
```diff
@@ -645,11 +645,15 @@ impl Endpoint {
...
-        self.connections[ch].conn.timeout(now, timer);
+        self.connections[ch]
+            .conn
+            .handle_event(ConnectionEvent::Timer(now, timer));
         if self.connections[ch].app_closed {
             self.forget(ch);
             return;
         }
+
+        self.endpoint_eventful_conns.insert(ch);
```
Ralith: This doesn't seem to be needed yet as of this commit.

quinn-proto/src/connection.rs
```diff
@@ -3245,6 +3253,9 @@ pub enum ConnectionEvent {
 /// Events to be sent to the Endpoint
 #[derive(Clone, Debug)]
 pub enum EndpointEvent {
+    Closed {
+        remote: SocketAddr,
```
Ralith: Could this be moved into `ConnectionMeta`?

quinn-proto/src/endpoint.rs
Ralith: If that call will return `None`, why do we need to prevent it?

quinn-proto/src/endpoint.rs
```diff
@@ -772,13 +778,10 @@ impl Endpoint {
...
     /// to call this only when all important communications have been completed.
     pub fn close(&mut self, now: Instant, ch: ConnectionHandle, error_code: u16, reason: Bytes) {
-        if self.connections[ch].conn.is_drained() {
-            self.forget(ch);
-            return;
-        }
         self.connections[ch].conn.close(now, error_code, reason);
         self.connections[ch].app_closed = true;
         self.needs_transmit.insert(ch);
+        self.endpoint_eventful_conns.insert(ch);
```
Ralith: Let's try to keep documentation for times when a connection might have new endpoint events, in the same vein as for the existing `poll_*` functions.

quinn/src/lib.rs
```diff
@@ -546,6 +323,266 @@ impl EndpointInner {
...
+                    continue;
+                }
+                Err(e) => {
+                    return Err(e);
+                }
+            }
+            recvd += 1;
+            if recvd >= IO_LOOP_BOUND {
+                keep_going = true;
+                break;
```
Ralith: If we just did `return Ok(true);` here, we wouldn't need the `keep_going` variable anymore.

quinn/src/lib.rs
```diff
@@ -546,6 +323,266 @@ impl EndpointInner {
             x.notify();
         }
     }
+
+    fn poll_recv(
```
Ralith: `poll_*` naming in Futures code almost universally refers to a function that returns `Poll<_, _>`. Maybe these should be named `drive_*`?

quinn/src/lib.rs
```diff
@@ -546,6 +323,266 @@ impl EndpointInner {
...
+                    break;
+                }
+                Err(e) => {
+                    return Err(e);
+                }
+            }
+            sent += 1;
+            if sent >= IO_LOOP_BOUND {
+                keep_going = true;
+                break;
```
Ralith: Could return directly here.

quinn/src/lib.rs
```diff
@@ -546,6 +323,266 @@ impl EndpointInner {
...
+                }
+            }
+        }
+        Ok(blocked)
+    }
+
+    fn poll_transmit(&mut self, now: Instant) -> Result<bool, io::Error> {
+        let mut sent = 0;
+        let mut keep_going = false;
+        while let Some(x) = self.inner.poll_transmit(now) {
```
Ralith: `while let Some(x) = self.outgoing.take().or_else(|| self.inner.poll_transmit(now))` would render `poll_send` unnecessary; pre-existing issue, though, so not a blocker.

quinn-proto/src/endpoint.rs
djc: Because it indexes into `self.connections[ch]` with a `ConnectionHandle` which is no longer present there?

Ralith: Ah, of course.

quinn-proto/src/endpoint.rs
```diff
@@ -94,6 +98,15 @@ impl Endpoint {
         None
     }
 
+    pub(crate) fn handle_endpoint_events(&mut self) {
+        while let Some(&ch) = self.endpoint_eventful_conns.iter().next() {
```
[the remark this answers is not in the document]
Ralith: Of course, we can't hold a borrow of `endpoint_eventful_conns` across a method call to self, but we could `mem::replace` it with an empty `HashMap`, and then swap them back when we're done to reuse the storage.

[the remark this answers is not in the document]
djc: Seems a little involved for a method that's going away again soon, but okay!

quinn-proto/src/endpoint.rs
```diff
@@ -149,6 +162,26 @@ impl Endpoint {
         }
     }
 
+    pub fn handle_event(&mut self, ch: ConnectionHandle, event: EndpointEvent) {
+        match event {
+            EndpointEvent::NeedIdentifiers => {
+                if self.config.local_cid_len != 0 {
+                    /// connection IDs
```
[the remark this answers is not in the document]
djc: Fixed.

quinn-proto/src/lib.rs
```diff
@@ -212,6 +212,7 @@ pub struct Transmit {
 // Useful internal constants
 //
 
+const LOC_CID_COUNT: usize = 8;
```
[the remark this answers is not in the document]
djc: Yup.

quinn-proto/src/connection.rs
djc: Let's keep it private while it can be, and then we can always make it public again if a better way comes along?

quinn-proto/src/connection.rs
```diff
@@ -3226,6 +3239,11 @@ struct SentPacket {
 /// Ensures we can always fit all our ACKs in a single minimum-MTU packet with room to spare
 const MAX_ACK_BLOCKS: usize = 64;
 
+/// Events to be sent to the Connection
+pub enum ConnectionEvent {
+    NewIdentifiers([Option<(u64, ConnectionId)>; LOC_CID_COUNT]),
```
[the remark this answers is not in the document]
djc: Yeah, just a little premature allocation prevention. Will fix.

quinn-proto/src/connection.rs
djc: I don't think that makes sense, `Connection` actually has a number of mechanics around the `remote`.

quinn-proto/src/endpoint.rs
djc: The reason I did not spend much time on it is that all the connection-tracking bits in `Endpoint` (so `dirty_timers`, `needs_transmit`, `endpoint_eventful_conns`, `eventful_conns`) are going away pretty soon, so spending much more attention on this didn't seem worth it.

quinn/src/lib.rs
djc: Okay.

quinn/src/lib.rs
djc: Fixed.

quinn/src/lib.rs
djc: Yup.

quinn-proto/src/connection.rs
Ralith: It makes sense for `Connection` to *know about* the `remote`, but it seems a little odd for `Endpoint` not to retain a copy, when it's using it as a key in its internal tables. Not a big deal, though.

quinn-proto/src/endpoint.rs
Ralith: Fair enough!

quinn-proto/src/endpoint.rs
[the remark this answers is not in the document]
Ralith: Ah, didn't realize that was coming so soon! I'll be glad to be rid of them, iterating over `Hash*` is never a good idea.

quinn/src/tests.rs
Ralith: Your call, pending Apple's response to that bug report.

Issue #210, opened 2019-01-22
setting idle_timeout to very large numbers yields in immediate timeout


est31:
When I'm setting `idle_timeout` to numbers like `1<<62 - 1` or `1 << 61`, I'm getting an idle timeout right after connecting. "small" values like 2000 work great.

est31:
In debug mode, instead of an idle timeout right away you are getting an "attempt to multiply with overflow". This is what I'm getting for various values in debug mode:

* `1 << 20` works
* `1 << 26` works
* `1 << 27` doesn't work "timer shutdown"
* `1 << 40` doesn't work "timer shutdown"
* `1 << 50` doesn't work "attempt to multiply with overflow"
* `1 << 60` doesn't work "attempt to multiply with overflow"

Ralith:
#212 should address the overflow; the other cases look like tokio behaving weird, which will need to be investigated separately, although I'm not sure there's a genuine usecase for setting your idle timeout longer than 4 years once #212 also fixes disabling the timeout outright.

Ralith:
#213 addressed the "value too large" panic by sanity-checking during varint construction, as well.

djc:
So should we close this now?

Ralith:
I haven't investigated the "timer shutdown" issue yet; I suspect it's an issue with tokio

PR #211, opened 2019-01-22
Add some abstractions around the use of rustls/TLS

djc:


quinn-proto/src/connection.rs
```diff
@@ -834,20 +831,20 @@ impl Connection {
...
+                        params.expect("rustls didn't supply transport parameters with ticket"),
+                    )
+                    .map_err(Into::into)
+                })
+                .unwrap_or_else(|e| {
+                    error!(
+                        self.log,
+                        "session ticket has malformed transport parameters: {}", e
+                    )
+                })
```
Ralith: Accidentally removed `return`? We need to bail out of the function here if the transport parameters are missing or malformed, because sending 0-RTT in that case is forbidden.

quinn-proto/src/connection.rs
```diff
@@ -1231,17 +1228,14 @@ impl Connection {
...
                                     debug!(self.log, "remote didn't send transport params");
-                                    ConnectionError::from(TransportError::PROTOCOL_VIOLATION)
-                                })
-                                .and_then(|x| {
-                                    TransportParameters::read(self.side, &mut io::Cursor::new(x))
-                                        .map_err(Into::into)
-                                })?;
+                                    Err(ConnectionError::from(TransportError::PROTOCOL_VIOLATION))
+                                }
+                                Err(e) => Err(e.into()),
```
Ralith: If we're using `?` below, we can get rid of the `into` and `from` calls above.

quinn-proto/src/crypto.rs
```diff
@@ -20,6 +20,7 @@ use webpki::DNSNameRef;
 
 use crate::coding::{BufExt, BufMutExt};
 use crate::packet::{ConnectionId, PacketNumber, LONG_HEADER_FORM};
+use crate::transport_error::Error as TransportError;
```
Ralith: More convenient to import the already-aliased version of this from the crate root.

quinn-proto/src/crypto.rs
```diff
@@ -79,6 +78,24 @@ impl TlsSession {
             },
         }
     }
+
+    pub fn write_handshake(&mut self, buf: &mut Vec<u8>) -> Result<Option<Crypto>, TLSError> {
+        match self.write_hs(buf) {
+            Some(secrets) => match self.get_negotiated_ciphersuite() {
+                None => Ok(None),
```
Ralith: We shouldn't ever be throwing away secrets rustls provides us; this should probably be unwrapped instead.

quinn-proto/src/endpoint.rs
```diff
@@ -55,18 +54,14 @@ pub struct Endpoint {
 }
 
 impl Endpoint {
-    pub fn new(
-        log: Logger,
-        config: Config,
-        server_config: Option<ServerConfig>,
-    ) -> Result<Self, EndpointError> {
+    pub fn new(log: Logger, config: Config, server_config: Option<ServerConfig>) -> Self {
```
Ralith: #210 just inspired me to put some `Config` sanity-checking in here, though I'm not sure if we should use panics or `Result` to report failures there; I imagine `Config`s will usually be hardcoded.

quinn-proto/src/crypto.rs
```diff
@@ -101,6 +125,11 @@ impl CryptoSession for TlsSession {
 }
 
 pub trait CryptoSession {
+    fn alert(&self) -> Option<AlertDescription>;
```
Ralith: We could replace `AlertDescription` with `u8` here and elsewhere, if desired. I only ever included it for the sake of `TransportError`'s `Debug` impl, but that re-parses it from a byte anyway.

quinn-proto/src/crypto.rs
```diff
@@ -135,6 +121,32 @@ pub trait CryptoSession {
     fn write_handshake(&mut self, buf: &mut Vec<u8>) -> Result<Option<Crypto>, TLSError>;
 }
 
+impl CryptoClientConfig for Arc<ClientConfig> {
```
Ralith: Why for `Arc<ClientConfig>` instead of for `ClientConfig`?

quinn-proto/src/crypto.rs
```diff
@@ -147,6 +140,18 @@ pub trait CryptoClientConfig {
     ) -> Result<Self::Session, ConnectError>;
 }
 
+impl CryptoServerConfig for Arc<ServerConfig> {
```
Ralith: Why for `Arc<ServerConfig>` instead of for `ServerConfig`?

quinn-proto/src/connection.rs

quinn-proto/src/connection.rs
djc: Yup, nice.

quinn-proto/src/crypto.rs
djc: 👍

quinn-proto/src/crypto.rs
djc: I changed it to `panic!()`, which seems warranted for the case that rustls gives us secrets but not a cipher suite.

quinn-proto/src/crypto.rs
djc: I'd like to leave it for now -- this effort isn't done yet anyway and this is clearly a TLS detail seeping through, so it will need changes one way or another down the line.

quinn-proto/src/crypto.rs
djc: With just `ClientConfig` the compiler doesn't accept it. Here too, I'd prefer to leave it slightly ugly for now until the rest of the picture about this abstraction becomes more clear (for example, in this case I think we might want to pull in the `ClientConfig` defined in the quinn crate and define the `impl` on that).

quinn-proto/src/crypto.rs
djc: See `ClientConfig`.

quinn-proto/src/endpoint.rs
djc: Removed this commit from the branch.

quinn-proto/src/crypto.rs
Ralith: Could still be useful for swapping in other TLS impls if someone wants to do that, but we can worry about it when such a requirement actually arises.

djc: My goal is to make this API abstract enough that it can cover non-TLS crypto protocols like Noise, too, so I'd prefer not to expose this TLSicism.

quinn-proto/src/crypto.rs
```diff
@@ -78,6 +77,24 @@ impl TlsSession {
             },
         }
     }
+
+    pub fn write_handshake(&mut self, buf: &mut Vec<u8>) -> Result<Option<Crypto>, TLSError> {
```
Ralith: This error case seems to neither be generated nor handled. I guess the idea is to future proof the type signature in case we ever end up with a handshake that doesn't validate incoming data up front?

quinn-proto/src/crypto.rs
```diff
@@ -59,6 +60,29 @@ impl TlsSession {
...
+
+    fn early_secret(&self) -> Option<&[u8]> {
+        self.get_early_secret()
+    }
+
+    fn is_handshaking(&self) -> bool {
+        match self {
+            TlsSession::Client(session) => session.is_handshaking(),
+            TlsSession::Server(session) => session.is_handshaking(),
+        }
```
Ralith: Why does this explicitly destructure while the other wrappers go through the `Deref` impl?

quinn-proto/src/crypto.rs
Ralith: It's worth noting that QUIC reserves exactly one `u8` worth of transport error codes "for carrying error codes specific to the cryptographic handshake that is used", so while `fn alert(&self)` might not be the precise abstraction we want, whatever we do land on should probably have `u8` error codes all the same.

quinn-proto/src/crypto.rs
Ralith: You could probably make rustc happy with UFCS, but yeah, not important for short-lived code.

quinn-proto/src/crypto.rs
djc: Oh, good catch. Removed that and simplified the code some more in the process.

quinn-proto/src/connection.rs
```diff
@@ -830,10 +830,7 @@ impl Connection {
         if self.state.is_closed() {
             return Ok(());
         }
-        let params = TransportParameters::read(
-            Side::Server,
-            &mut io::Cursor::new(self.tls.get_quic_transport_parameters().unwrap()),
-        )?;
+        let params = self.tls.transport_parameters()?.unwrap();
```
Ralith: rustls doesn't currently report an error if the transport paramters extension is missing, so this should probably be PROTOCOL_VIOLATION (if not some TLS alert) rather than unwrap; compare how the client handles this case. Pre-existing bug, but might as well fix while we're in here.

Ralith: In fact I'll just get this myself in a followup.

PR #212, opened 2019-01-23
Assorted cleanup

Ralith:


Ralith:
The final commit is theoretically compatibility-breaking and might be best off in a separate branch along with the protocol version number bump until interop testing advances, but I don't think anyone's actually using that transport parameter yet anyway.

PR #216, opened 2019-01-26
Config refactor

Ralith:


djc:
Looks okay, but I'm a little confused about what this is trying to achieve/why this is an improvement?

Ralith:
The main practical motivation is that different application-layer protocols might have different permissible streams and flow control requirements, and it struck me as unfortunate to require a lowest common denominator configuration in that case. For example, a HTTP/3 client supports incoming unidirectional streams, while a HTTP/0.9 does not, and a protocol that uses streams primarily as disposable message-passing primitives will have very different optimal flow control settings than one that uses them primarily for bulk data transfer.

I also think it's a little more natural to have `Endpoint` configuration cleanly isolated from `Connection` configuration, and I don't see any benefit to requiring all connections to use the same values for the latter.

Ralith:
This also has the interesting benefit of isolating cryptographic keys from transport configuration, so an application protocol library (such as quinn-h3) can define a `const` `TransportConfig` with suitable defaults for the convenience of downstream code.

djc:
Okay, and one more question about the name: I think I understand where TransportConfig is coming from, but in the end it still changes the behavior of Connection. Should it be called Connection Config instead?

Ralith:
I actually started with that, but I worried that it might be confusing to have both `ConnectionConfig` and `ClientConfig` as per-`Connection` configuration. Following #211 beginning to move things towards `CryptoClientConfig`, I imagined instead us eventually moving towards `struct ClientConfig { crypto: CryptoClientConfig, transport: TransportConfig }` as something a bit clearer, though I'm open to other ideas.

PR #217, opened 2019-01-28
Assorted fixes

Ralith:
The rebinding interop test is a bit lazy in that it only runs on endpoints that implement key update; that's fixable, but the current state is an easy step in the right direction.


Ralith:
[Wireshark expects a different format now](https://github.com/quicwg/base-drafts/wiki/Tools#wireshark):
> The TLS key log file follows the TLS 1.3 labels (with QUIC_ prepended). Every line follows the format <label> <ClientRandom> <TrafficSecret> where <label> is one of: QUIC_CLIENT_EARLY_TRAFFIC_SECRET, QUIC_CLIENT_HANDSHAKE_TRAFFIC_SECRET, QUIC_SERVER_HANDSHAKE_TRAFFIC_SECRET, QUIC_CLIENT_TRAFFIC_SECRET_0, QUIC_SERVER_TRAFFIC_SECRET_0. Example: https://github.com/ngtcp2/ngtcp2/pull/84

djc:
Cool! If you want to redo this PR without that commit we can just land it, in the mean time?

PR #218, opened 2019-01-28
More abstraction of crypto protocol implementation

djc:


djc:
The logging for errors is a bit worse, but I couldn't think of a better way. Any ideas?

quinn-proto/src/crypto.rs
```diff
@@ -64,8 +64,14 @@ impl CryptoSession for TlsSession {
         }
     }
 
-    fn read_handshake(&mut self, buf: &[u8]) -> Result<(), TLSError> {
-        self.read_hs(buf)
+    fn read_handshake(&mut self, buf: &[u8]) -> Result<(), TransportError> {
+        self.read_hs(buf).map_err(|_| {
+            if let Some(code) = self.error_code() {
```
Ralith: I think we can remove `error_code` from the trait entirely with this change.

djc: Right! Fixed.

djc:
Ready for another round of review, I think.

quinn-proto/src/crypto.rs
```diff
@@ -64,8 +59,14 @@ impl CryptoSession for TlsSession {
         }
     }
 
-    fn read_handshake(&mut self, buf: &[u8]) -> Result<(), TLSError> {
-        self.read_hs(buf)
+    fn read_handshake(&mut self, buf: &[u8]) -> Result<(), TransportError> {
+        self.read_hs(buf).map_err(|e| {
+            if let Some(alert) = self.get_alert() {
+                TransportError::crypto(alert.get_u8(), format!("TLS error: {}", e))
```
```suggestion
                TransportError::crypto(alert.get_u8(), e.to_string())
```
TLS-ness is implied by the error code.

quinn-proto/src/connection.rs
```diff
@@ -906,12 +908,8 @@ impl Connection {
             }
             trace!(self.log, "read {} TLS bytes", n);
             if let Err(e) = self.tls.read_handshake(&buf[..n]) {
-                debug!(self.log, "TLS error: {}", e);
-                return Err(if let Some(alert) = self.tls.alert() {
-                    TransportError::crypto(alert)
-                } else {
-                    TransportError::PROTOCOL_VIOLATION("TLS error")
-                });
+                debug!(self.log, "reading handshake bytes resulted in {}", e);
```
Ralith: `TransportError`s are already logged in `handle_packet`, and we're now encoding all information into them, so this is redundant. Should be able to do away with the `if let` entirely.

PR #219, opened 2019-01-30
Congestion control fix

Ralith:
I think this accounts for the bulk of the strangely pessimal performance I've been observing correlated with early packet loss: the connection was getting stuck in congestion recovery mode, because packet numbers tend to be much smaller than microsecond timestamps.


Ralith:
It jumped out at me completely by chance when I started work on extracting the congestion controller into a trait in hopes of isolating the problem. Lucky break!

PR #220, opened 2019-01-31
Use Instant and Duration for timing

djc:
Thought about this after that slowdown bug -- it seems pretty okay? Not sure why you went with `u64` in the first place, so curious about your thinking there.

This compiles and passes the -proto tests, but fails the high-level echo tests for now.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -597,9 +597,11 @@ impl Connection {
     fn detect_lost_packets(&mut self, now: Instant) {
         self.loss_time = None;
         let mut lost_packets = Vec::<u64>::new();
-        let rtt = cmp::max(self.rtt.latest, self.rtt.smoothed);
-        let loss_delay =
-            Duration::from_micros(rtt + ((rtt * self.config.time_threshold as u64) >> 16));
+        let mut rtt = self.rtt.latest;
+        if let Some(smoothed) = self.rtt.smoothed {
+            rtt = cmp::max(rtt, smoothed);
+        }
+        let loss_delay = rtt + ((rtt * self.config.time_threshold as u32) / 65536);
         let lost_send_time = now - loss_delay;
 
         let mut lost_ack_eliciting = false;
@@ -668,14 +670,12 @@ impl Connection {
     fn set_loss_detection_timer(&mut self) {
         if self.in_flight.crypto != 0 || (self.state.is_handshake() && self.side.is_client()) {
             // Handshake retransmission alarm.
-            let timeout = Duration::from_micros(if self.rtt.smoothed == 0 {
-                2 * self.config.initial_rtt
+            let timeout = if let Some(smoothed) = self.rtt.smoothed {
+                2 * smoothed
             } else {
-                2 * self.rtt.smoothed
-            });
-            let timeout = Duration::from_micros(
-                micros_from(cmp::max(timeout, TIMER_GRANULARITY)) * 2u64.pow(self.crypto_count),
-            );
+                2 * Duration::from_micros(self.config.initial_rtt)
+            };
+            let timeout = cmp::max(timeout, TIMER_GRANULARITY) * 2u32.pow(self.crypto_count);
             self.io.timer_start(
                 Timer::LossDetection,
                 self.time_of_last_sent_crypto_packet + timeout,
@@ -705,8 +705,11 @@ impl Connection {
 
     /// Probe Timeout
     fn pto(&self) -> Duration {
-        let computed =
-            Duration::from_micros(self.rtt.smoothed + 4 * self.rtt.var) + self.max_ack_delay();
+        let rtt = self
+            .rtt
+            .smoothed
+            .unwrap_or(Duration::from_micros(self.config.initial_rtt));
+        let computed = rtt + 4 * self.rtt.var + self.max_ack_delay();
         cmp::max(computed, TIMER_GRANULARITY)
     }
 
@@ -3497,30 +3500,27 @@ impl InFlight {
 }
 
 struct RttEstimator {
-    /// The most recent RTT measurement made when receiving an ack for a previously unacked packet.
-    /// μs
-    latest: u64,
-    /// The smoothed RTT of the connection, computed as described in RFC6298. μs
-    smoothed: u64,
+    /// The most recent RTT measurement made when receiving an ack for a previously unacked packet
+    latest: Duration,
+    /// The smoothed RTT of the connection, computed as described in RFC6298
+    smoothed: Option<Duration>,
     /// The RTT variance, computed as described in RFC6298
-    var: u64,
+    var: Duration,
     /// The minimum RTT seen in the connection, ignoring ack delay.
-    min: u64,
+    min: Duration,
 }
 
 impl RttEstimator {
     fn new() -> Self {
         Self {
-            latest: 0,
-            smoothed: 0,
-            var: 0,
-            min: u64::max_value(),
+            latest: Duration::new(0, 0),
+            smoothed: None,
+            var: Duration::new(0, 0),
+            min: Duration::new(u64::max_value(), 0),
         }
     }
 
     fn update(&mut self, ack_delay: Duration, rtt: Duration) {
-        let ack_delay = micros_from(ack_delay);
-        let rtt = micros_from(rtt);
         self.latest = rtt;
         // min_rtt ignores ack delay.
         self.min = cmp::min(self.min, self.latest);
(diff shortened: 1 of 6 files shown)


djc:
(Also spent quite a bit of time making good use of `Option` combinators, but wouldn't be surprised if there are still ways to simplify -- it somehow doesn't come naturally to me and then sometimes I don't have the patience to figure it out.)

Ralith:
I'm hesitant to switch over wholesale to std time types because the draft works exclusively in terms of microseconds, making this simultaneous a divergence from the draft, an increase in the size of our time representations, and a reduction in useful range/increase in likelihood of overflows. Perhaps we could switch to newtypes of `u64` instead?

djc:
So you're saying:

* Divergence from the draft, which exclusively talks about microseconds: I don't see much divergence here, and thus don't understand what downsides you see here? `Instant` and `Duration` can easily represent microsecond timestamps and intervals.

* Increased memory use for time representation sizes: this is indeed an adverse effect, but it's not that clear how much it would matter in practice. From some testing, it appears that `Connection` is about 12k bytes before these changes. We have 4 `Instant`s in `Connection` itself, plus one per `PacketSpace`, one per `SentPacket`, one per `TimerSetting`, so if we have 30 or so `Instant`s in memory that's still less than 1% more memory.

* Reduction in useful range (increasing likelihood of overflows): this seems to be true on macOS and iOS, but for all other platforms it looks like we actually have more range this way? And even on macOS, where we would AFAICT use 64 bits to represent nanoseconds, useful range by a quick calculation is in the range of hundreds of years, so it still doesn't seem like a problem?

On the other hand, using these makes the code instantly clear for new contributors and provides type safety without having to implement newtypes with a bunch of operations resulting in boilerplate.

`Instant` size on platforms (`Duration` seems to be 12 bytes everywhere):

* macOS/iOS, CloudABI: `u64` -> 8 bytes
* Non-macOS Unix: `Timespec` -> 12 bytes
* Windows, WASM, SGX: `Duration` -> 12 bytes

Ralith:
> Divergence from the draft, which exclusively talks about microseconds: I don't see much divergence here, and thus don't understand what downsides you see here?

I guess the only substantial case of this is the replacement of sentinel values with `Option`, which is probably worth it for clarity.

> Increased memory use for time representation sizes: this is indeed an adverse effect, but it's not that clear how much it would matter in practice

Right, this is really just a cosmetic concern.

> Reduction in useful range (increasing likelihood of overflows): this seems to be true on macOS and iOS, but for all other platforms it looks like we actually have more range this way?

Whoops, misread the docs. This is probably fine then.

> On the other hand, using these makes the code instantly clear for new contributors and provides type safety without having to implement newtypes with a bunch of operations resulting in boilerplate.

Fair enough. I'll review in depth and see if I can find any semantic changes responsible for the test failures.

quinn-proto/src/connection.rs
```diff
@@ -630,11 +642,13 @@ impl Connection {
             }
         }
         if lost_ack_eliciting {
-            self.congestion_event(now, largest_lost_time)
+            if let Some(lost_time) = largest_lost_time {
+                self.congestion_event(now, lost_time);
+            }
```
Ralith: `largest_lost_time` should be unwrapped rather than guarded here, because it is always set if lost_ack_eliciting is set (although the reverse does not hold).

quinn-proto/src/connection.rs
```diff
@@ -651,19 +665,21 @@ impl Connection {
...
         if self.in_flight.crypto != 0 || (self.state.is_handshake() && self.side.is_client()) {
             // Handshake retransmission alarm.
-            let timeout = if self.rtt.smoothed == 0 {
+            let timeout = Duration::from_micros(if self.rtt.smoothed == 0 {
                 2 * self.config.initial_rtt
             } else {
                 2 * self.rtt.smoothed
-            };
-            let timeout = cmp::max(timeout, TIMER_GRANULARITY) * 2u64.pow(self.crypto_count);
+            });
```
Ralith: Shouldn't rtt statistics be converted to `Duration`s too?

quinn-proto/src/connection.rs
```diff
@@ -651,19 +665,21 @@ impl Connection {
...
+            let timeout = Duration::from_micros(
+                micros_from(cmp::max(timeout, TIMER_GRANULARITY)) * 2u64.pow(self.crypto_count),
+            );
```
Ralith: It would be nicer to use `Duration`'s impl of `Mul<u32>` here; maybe extract `MAX_PTO_EXPONENT` to module scope and reuse it while we're at it.

quinn-proto/src/connection.rs
```diff
@@ -545,18 +552,18 @@ impl Connection {
         false
     }
 
-    fn set_key_discard_timer(&mut self, now: u64) {
+    fn set_key_discard_timer(&mut self, now: Instant) {
         let time = if self.spaces[SpaceId::Handshake as usize].crypto.is_some() {
             now + self.pto() * 3
         } else if let Some(ref time) = self.prev_crypto.as_ref().and_then(|x| x.update_ack_time) {
-            time + self.pto() * 3
+            *time + self.pto() * 3
```
Ralith: Style nit: get rid of the `ref` above rather than dereffing?

quinn-proto/src/connection.rs
```diff
@@ -3300,7 +3314,7 @@ impl IoQueue {
 #[derive(Debug, Copy, Clone)]
 pub enum TimerSetting {
     /// Set the timer to expire at an a certain point in time, in absolute microseconds
```
```suggestion
     /// Set the timer to expire at an a certain point in time
```

quinn-proto/src/tests.rs
```diff
@@ -192,9 +190,11 @@ impl Pair {
                 socket.send_to(&x.packet, x.destination).unwrap();
             }
             if self.server.addr == x.destination {
-                self.server
-                    .inbound
-                    .push_back((self.time + self.latency, x.ecn, x.packet));
+                self.server.inbound.push_back((
+                    self.time + Duration::from_micros(self.latency),
```
Ralith: Perhaps `self.latency` should be a `Duration`?

quinn-proto/src/tests.rs
```diff
@@ -923,15 +925,20 @@ fn idle_timeout() {
...
+            if let Some(time) = min_opt(pair.client.next_wakeup(), pair.server.next_wakeup()) {
+                pair.time = time;
+            } else {
+                advanced = false;
+            }
         }
         pair.client.inbound.clear(); // Simulate total S->C packet loss
     }
-    assert!(pair.time != u64::max_value());
+    assert!(advanced);
```
Ralith: Maybe assert that the elapsed time is >= the configured idle timeout instead?

quinn-proto/src/connection.rs
```diff
@@ -604,20 +612,24 @@ impl Connection {
...
             // OnPacketsLost
             if let Some(largest_lost) = lost_packets.last().cloned() {
                 let old_bytes_in_flight = self.in_flight.bytes;
-                largest_lost_time = cmp::max(
-                    largest_lost_time,
-                    space.sent_packets[&largest_lost].time_sent,
-                );
+                largest_lost_time =
+                    Some(largest_lost_time.map_or(
+                        space.sent_packets[&largest_lost].time_sent,
```
Ralith: This lookup might be expensive, so it would be best to extract it into a `let` binding rather than perform it twice when `largest_lost_time` is `Some(_)`.

quinn-proto/src/connection.rs

quinn-proto/src/connection.rs

quinn-proto/src/tests.rs
djc: Now asserting that the `next` coming out of the `min_opt(next wakeups)` is `Some`.

quinn-proto/src/connection.rs

quinn-proto/src/connection.rs

quinn-proto/src/tests.rs
djc: Yup, good call.

quinn-proto/src/connection.rs
djc: Yeah, I guess so. Have a look?

quinn-proto/src/connection.rs

djc:
Haven't fixed the test failure yet, but this should fix all your comments and also splits it into multiple commits, making things easier to follow.

quinn-proto/src/connection.rs
```diff
@@ -3538,3 +3537,5 @@ struct PathResponse {
 fn micros_from(x: Duration) -> u64 {
     x.as_secs() * 1000 * 1000 + x.subsec_micros() as u64
 }
+
+const MAX_PTO_EXPONENT: u32 = 24; // 2^24μs = ~17 seconds
```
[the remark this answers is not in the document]
djc: Given that all this is pretty much unrelated to the goal of this PR, would you mind taking care of this in a separate PR?

quinn-proto/src/connection.rs
```diff
@@ -668,14 +667,12 @@ impl Connection {
     fn set_loss_detection_timer(&mut self) {
         if self.in_flight.crypto != 0 || (self.state.is_handshake() && self.side.is_client()) {
             // Handshake retransmission alarm.
-            let timeout = Duration::from_micros(if self.rtt.smoothed == 0 {
-                2 * self.config.initial_rtt
+            let timeout = if self.rtt.smoothed == Duration::new(0, 0) {
```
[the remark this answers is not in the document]
djc: Makes sense.

quinn-proto/src/connection.rs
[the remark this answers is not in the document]
Ralith: It came to mind because switching to `u32` for the exponential factor in this PR makes the exponentiation considerably more likely to overflow. Admittedly it would still take long time by the wall clock, so it can wait if you prefer.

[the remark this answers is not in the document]
djc: We can do it soon, but I haven't quite grokked what you want here, so it feels inefficient for me to figure that out and do it when you can probably just crank it out. :)

PR #221, opened 2019-02-01
Keepalive

Ralith:
This takes a simple, low-impact approach of overloading the idle timer to issue keepalives evenly distributed over the negotiated idle timeout. Other retransmit strategies (such as imitating a PTO) could be pursued, but it's not clear to me that what stands to be gained from a more complex approach, and knowing exactly how many are sent gives you an easily-quantified degree of robustness against random packet loss.


quinn-proto/src/connection.rs
```diff
@@ -684,9 +684,7 @@ impl Connection {
 
         // Calculate PTO duration
         const MAX_PTO_EXPONENT: u32 = 24; // 2^24μs = ~17 seconds
-        let timeout = self.rtt.smoothed + 4 * self.rtt.var + self.max_ack_delay();
-        let timeout = cmp::max(timeout, TIMER_GRANULARITY)
-            * 2u64.pow(cmp::min(self.pto_count, MAX_PTO_EXPONENT));
+        let timeout = self.pto() * 2u64.pow(cmp::min(self.pto_count, MAX_PTO_EXPONENT));
```
djc: Noticed this too in my other PR. 👍

quinn-proto/src/endpoint.rs
```diff
@@ -861,6 +862,15 @@ pub struct TransportConfig {
...
     pub persistent_congestion_threshold: u32,
+    /// Number of keep-alive packets to send before a connection times out.
+    ///
+    /// Keep-alive packets prevent an inactive but otherwise healthy connection from timing
+    /// out. They are distributed evenly over the idle timeout, and are only sent when the
+    /// connection is otherwise idle. Larger numbers will be more robust to packet loss.
+    ///
+    /// 0 to disable, which is the default. Only one side of any given connection should have
+    /// keep-alives enabled.
+    pub keepalives: u32,
```
djc: I think the more natural identifier spelling would be `keep_alives`?

quinn-proto/src/endpoint.rs
```diff
@@ -861,6 +862,15 @@ pub struct TransportConfig {
     pub loss_reduction_factor: u16,
     /// Number of consecutive PTOs after which network is considered to be experiencing persistent congestion.
     pub persistent_congestion_threshold: u32,
+    /// Number of keep-alive packets to send before a connection times out.
```
djc: My intuition is that this is usually configured as a period rather than a frequency. Why does it work this way? Can we make it easier to use?

quinn-proto/src/endpoint.rs
```diff
@@ -861,6 +862,15 @@ pub struct TransportConfig {
...
+    ///
+    /// Keep-alive packets prevent an inactive but otherwise healthy connection from timing
+    /// out. They are distributed evenly over the idle timeout, and are only sent when the
+    /// connection is otherwise idle. Larger numbers will be more robust to packet loss.
+    ///
+    /// 0 to disable, which is the default. Only one side of any given connection should have
```
djc: Not sure I agree that `0` is a good default here -- doesn't TCP usually keep your connection alive for a while by default? Also, what happens in case both connections set this, this seems to imply some adverse effect? (Or is it saying that only one side needs to set it for the connections to stay alive?)

Ralith: I believe TCP's default behavior is no keep-alives and no idle timeout, i.e. connections only time out due to unacked data. I don't have a strong feeling about what defaults are correct; it strikes me as application-specific. Perhaps we should disable the idle timeout by default for consistency with TCP?

Ralith:
Switched to interval-based using a distinct timer to ensure correct behavior when data is being sent but not acknowledged by the peer.

quinn-proto/src/endpoint.rs
Ralith: Actually, that's not right either--disabling the idle timeout would be even more liberal than TCP, since QUIC does not terminate the connection due to excess retransmits.

Ralith:
On further consideration, I think PTOs for keep-alives that aren't immediately acknowledged are appropriate, and keep-alives at regular intervals are appropriate even when there is other ack-eliciting data in flight, so long as that data hasn't been (re)transmitted since one keep-alive interval in the past, because the traditional retransmit schedule can easily extend past the idle timeout, leaving a period in which the link might be restored without the connection being recovered in lieu of a regular keep-alive.

In other words, the current state of this is probably good, though I'd love a sanity-check.

Ralith:
Maybe there's still a case for a more aggressive keep-alive transmit schedule taking effect whenever a keep-alive has been lost and nothing else has been received, but I'm not sure.

quinn-proto/src/endpoint.rs
Ralith: Discussion on slack suggests that issuing keepalives at intervals of <=20 seconds might be useful for preserving NAT bindings (see https://dl.acm.org/citation.cfm?id=1879174). That would mostly only be useful for preserving connections to servers which have connection IDs disabled, but those might not be too rare, if people are trying to minimize overhead.

quinn-proto/src/endpoint.rs
```diff
@@ -861,6 +861,14 @@ pub struct TransportConfig {
     pub loss_reduction_factor: u16,
     /// Number of consecutive PTOs after which network is considered to be experiencing persistent congestion.
     pub persistent_congestion_threshold: u32,
+    /// Number of seconds of inactivity before sending a keep-alive packet
+    ///
+    /// Keep-alive packets prevent an inactive but otherwise healthy connection from timing out.
+    ///
+    /// 0 to disable, which is the default. Only one side of any given connection needs keep-alive
+    /// enabled for the connection to be preserved. Must be set lower than the idle_timeout of both
```
djc: Should we emit a warning if this happens?

quinn-proto/src/connection.rs
```diff
@@ -711,6 +716,7 @@ impl Connection {
     ) {
         self.remote_validated |= self.state.is_handshake() && space_id == SpaceId::Handshake;
         self.total_recvd = self.total_recvd.wrapping_add(size as u64);
+        self.reset_keep_alive(now);
```
djc: Just for my understanding, how do the triggers for `reset_keep_alive` and `reset_idle_timeout` different?

quinn/src/lib.rs
```diff
@@ -554,7 +554,7 @@ impl Pending {
             connecting,
             uni_opening: VecDeque::new(),
             bi_opening: VecDeque::new(),
-            cancel_timers: [None, None, None, None, None],
+            cancel_timers: [None, None, None, None, None, None],
```
djc: Can this now be `[None; Timer::COUNT]` or some such?

quinn-proto/src/connection.rs
Ralith: `reset_keep_alive` is triggered on packet receive or *any* packet send; `reset_idle_timeout` is triggered on packet receive or on the first send following a receive, but not on any following sends. In other words, `rest_idle_timeout` is not triggered by automatic retransmissions, or even long outgoing flights of new data on a dead link.

quinn-proto/src/endpoint.rs
Ralith: Via the logger? Good idea.

quinn-proto/src/connection.rs
djc: Makes sense, thanks!

PR #222, opened 2019-02-02
Expand ECN support

Ralith:
libc's CMSG macros got built out to many more platforms. There may remain some obscure platforms where this will still fail to build, but we can account for them as they show up in practice.

Note that some bugs remain in the new libc bindings, so we may want to defer merging this until https://github.com/rust-lang/libc/pull/1235 shows up in a release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ before_script:
 script:
 - cargo clean
 - cargo build --all-features
-- cargo test --all-features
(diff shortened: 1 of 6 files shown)


Ralith:
I think this is ready for re-review and merge. If the macos inconsistencies are representative, there'll probably be additional tweaks necessary for other platforms, but only minor ones, and their necessity will be immediately obvious when socket configuration fails.

PR #235, opened 2019-02-14
More cmsg cleanup

Ralith:
No semantic changes, just additional insurance against UB.

### DIFF
--- a/quinn/src/platform/cmsg.rs
+++ b/quinn/src/platform/cmsg.rs
@@ -66,7 +66,10 @@ impl<'a> Drop for Encoder<'a> {
 
 pub unsafe fn decode<T: Copy>(cmsg: &libc::cmsghdr) -> T {
     assert!(mem::align_of::<T>() <= mem::align_of::<libc::cmsghdr>());
-    debug_assert_eq!(cmsg.cmsg_len, libc::CMSG_LEN(mem::size_of::<T>() as _) as _);
+    debug_assert_eq!(
+        cmsg.cmsg_len as usize,
+        libc::CMSG_LEN(mem::size_of::<T>() as _) as usize
+    );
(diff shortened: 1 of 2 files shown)


quinn/src/platform/cmsg.rs
```diff
@@ -34,8 +42,10 @@ impl<'a> Encoder<'a> {
         self.len += space;
         self.cmsg = unsafe { libc::CMSG_NXTHDR(self.hdr, cmsg).as_mut() };
     }
+}
 
-    pub fn finish(self) {
+impl<'a> Drop for Encoder<'a> {
```
djc: Do you really feel this is better than what you had before? It feels less idiomatic to me.

Ralith: I was concerned that the prior approach was easy to forget to do. With the new approach, borrowck forces you to ensure the `Encoder` is dropped before the buffer is used. I'm open to other ideas, though.

djc: I understand the concern. I think the way I would write it is to have `finish()` return an opaque newtype around the buffer which is what's actually passed to other code. Would that make sense here?

Ralith: The "other code" in this case is libc, so there's not an opportunity for a richer interface there.

Ralith: Hmm, maybe `Encoder` could take ownership of the buffer and only release it on `finish`.

Ralith: No, that won't work, it needs a stable address to be referenced in the `msghdr`.

Ralith:
Switched to a compromise approach which allows explicit finishing of the encoding operation while preserving the static guarantee.

djc:
Looks okay to me once CI passes.

Ralith:
Should be ready.

PR #237, opened 2019-02-18
Fix version negotiation on outgoing connections

Ralith:
We were trying to decrypt version negotiation packets, which aren't encrypted in the first place, leading to the packets being dropped and the connection timing out, taking longer than necessary and giving a less helpful error when connecting to a server that does not support the version we implement.


Ralith:
Tweaked per comments.

PR #239, opened 2019-02-23
Reduce send/recv starvation when CPU bound

Ralith:
When the CPU is much slower than the link (for example, due to Quinn being built in debug mode and running on an Atom), large requests could cause the `Driver` future to do nothing but send packets until the (peer-controlled, often large) flow control window is filled. Note that the congestion control window will tend to grow without bound in this case, because outgoing packet loss is unlikely. In testing, this lead to `recv` not being called for several seconds at a time, resulting in incoming packet loss.

### DIFF
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -253,6 +253,12 @@ impl Endpoint {
 /// A future that drives IO on an endpoint.
 pub struct Driver(Rc<RefCell<EndpointInner>>);
 
+/// Maximum number of send/recv calls to make before moving on to other processing
+///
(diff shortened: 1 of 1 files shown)


djc:
The concept looks right, but my intuition is that the constant should be substantially lower (like 10-20?). If the CPU is a limiting factor, it feels like prioritizing latency a bit more over throughput (and maybe thus prevent more retransmit) seems like the right thing to do. How did you arrive at 100?

Ralith:
Completely arbitrary. Switched to 10, nightly benchmark shows no difference either way.

PR #241, opened 2019-02-27
Refactor reads

djc:


quinn-proto/src/stream.rs
```diff
@@ -276,6 +276,9 @@ pub enum ReadError {
     /// The data on this stream has been fully delivered and no more will be transmitted.
     #[error(display = "finished")]
     Finished,
+    /// Invalid stream: stream doesn't exist (right now)
+    #[error(display = "invalid")]
+    InvalidStream,
```
Ralith: Would `Unknown` be a little clearer?

quinn-proto/src/stream.rs
```diff
@@ -2,8 +2,81 @@ use std::collections::VecDeque;
 
 use bytes::Bytes;
 use err_derive::Error;
+use fnv::FnvHashMap;
 
 use crate::range_set::RangeSet;
+use crate::{Directionality, Side, StreamId, TransportError};
+
+pub(crate) struct Streams {
```
Ralith: We don't reexport this module, so `(crate)` isn't strictly necessary here and below, though I have no objection if you'd prefer to keep it for documentation reasons.

Ralith:
I guess travis is unhappy due to a change in rustfmt? That's a pain...

PR #246, opened 2019-03-07
More examples

povilasb:
Client/Server examples are great. They demonstrate that the project works, I can wireshark the traffic, see the quinn API, etc. But they are not so great for newbies when I want as less context as possible to see how to use quinn.

I was playing around with the library and made a couple of examples to get a better understanding. Seems like a valuable asset to share :)

### DIFF
--- a/quinn/examples/common/mod.rs
+++ b/quinn/examples/common/mod.rs
@@ -1,0 +1,78 @@
+//! Commonly used code in most examples.
+
+use quinn::{
+    Certificate, CertificateChain, ClientConfig, ClientConfigBuilder, Endpoint, EndpointDriver,
+    Incoming, PrivateKey, ServerConfig, ServerConfigBuilder, TransportConfig,
+};
+use std::error::Error;
+use std::net::ToSocketAddrs;
+use std::sync::Arc;
+
+/// Constructs a QUIC endpoint configured for use a client only.
+///
+/// ## Args
+///
+/// - server_certs: list of trusted certificates.
+#[allow(unused)]
+pub fn make_client_endpoint<A: ToSocketAddrs>(
+    bind_addr: A,
+    server_certs: &[&[u8]],
+) -> Result<(Endpoint, EndpointDriver), Box<Error>> {
+    let client_cfg = configure_client(server_certs)?;
+    let mut endpoint_builder = Endpoint::new();
+    endpoint_builder.default_client_config(client_cfg);
+    let (driver, endpoint, _incoming) = endpoint_builder.bind(bind_addr)?;
+    Ok((endpoint, driver))
+}
+
+/// Constructs a QUIC endpoint configured to listen for incoming connections on a certain address
+/// and port.
+///
+/// ## Returns
+///
+/// - UDP socket driver
+/// - a sream of incoming QUIC connections
+/// - server certificate serialized into DER format
+pub fn make_server_endpoint<A: ToSocketAddrs>(
+    bind_addr: A,
+) -> Result<(EndpointDriver, Incoming, Vec<u8>), Box<Error>> {
+    let (server_config, server_cert) = configure_server()?;
+    let mut endpoint_builder = Endpoint::new();
+    endpoint_builder.listen(server_config);
+    let (driver, _endpoint, incoming) = endpoint_builder.bind(bind_addr)?;
+    Ok((driver, incoming, server_cert))
+}
+
+/// Builds default quinn client config and trusts given certificates.
+///
+/// ## Args
+///
+/// - server_certs: a list of trusted certificates in DER format.
+fn configure_client(server_certs: &[&[u8]]) -> Result<ClientConfig, Box<Error>> {
+    let mut cfg_builder = ClientConfigBuilder::default();
+    for cert in server_certs {
+        cfg_builder.add_certificate_authority(Certificate::from_der(&cert)?)?;
+    }
+    Ok(cfg_builder.build())
+}
+
+/// Returns default server configuration along with its certificate.
+fn configure_server() -> Result<(ServerConfig, Vec<u8>), Box<Error>> {
+    let cert = rcgen::generate_simple_self_signed(vec!["localhost".into()]);
+    let cert_der = cert.serialize_der();
+    let priv_key = cert.serialize_private_key_der();
+    let priv_key = PrivateKey::from_der(&priv_key)?;
+
+    let server_config = ServerConfig {
+        transport: Arc::new(TransportConfig {
+            stream_window_uni: 0,
+            ..Default::default()
+        }),
+        ..Default::default()
+    };
+    let mut cfg_builder = ServerConfigBuilder::new(server_config);
+    let cert = Certificate::from_der(&cert_der)?;
+    cfg_builder.certificate(CertificateChain::from_certs(vec![cert]), priv_key)?;
+
+    Ok((cfg_builder.build(), cert_der))
+}
--- a/quinn/examples/connection.rs
+++ b/quinn/examples/connection.rs
@@ -1,0 +1,68 @@
+//! This example intends to use the smallest amout of code to make a simple QUIC connection.
+//!
+//! The server issues it's own certificate and passes it to the client to trust.
+//!
+//! Run:
+//! ```text
+//! $ cargo run --example connection
+//! ```
+//!
+//! This example will make a QUIC connection on localhost, and you should see output like:
+//! ```text
+//! [server] incoming connection: id=3680c7d3b3ebd250 addr=127.0.0.1:50469
+//! [client] connected: id=61a2df1548935aeb, addr=127.0.0.1:5000
+//! ```
+
+use futures::{Future, Stream};
+use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
+use tokio::runtime::current_thread::{self, Runtime};
+
+mod common;
+use common::{make_client_endpoint, make_server_endpoint};
+
+const SERVER_PORT: u16 = 5000;
+
+fn main() -> Result<(), Box<std::error::Error>> {
+    let mut runtime = Runtime::new()?;
+
+    let (driver, incoming, server_cert) = make_server_endpoint(("0.0.0.0", SERVER_PORT))?;
+    // drive UDP socket
+    runtime.spawn(driver.map_err(|e| panic!("IO error: {}", e)));
+    let handle_incoming_conns = incoming
+        .take(1)
+        .for_each(move |(conn_driver, conn, _incoming)| {
+            current_thread::spawn(conn_driver.map_err(|_| ()));
+            println!(
+                "[server] incoming connection: id={} addr={}",
+                conn.remote_id(),
+                conn.remote_address()
+            );
+            Ok(())
+        });
+    runtime.spawn(handle_incoming_conns);
+
+    let (endpoint, driver) = make_client_endpoint("0.0.0.0:0", &[&server_cert])?;
+    // drive UDP socket
+    runtime.spawn(driver.map_err(|e| panic!("IO error: {}", e)));
+
+    let server_addr = SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), SERVER_PORT));
+    let connect = endpoint
+        .connect(&server_addr, "localhost")?
+        .map_err(|e| panic!("Failed to connect: {}", e))
+        .and_then(|(conn_driver, conn, _)| {
+            current_thread::spawn(conn_driver.map_err(|_| ()));
+            println!(
+                "[client] connected: id={}, addr={}",
+                conn.remote_id(),
+                conn.remote_address()
+            );
+            Ok(())
+        });
+    runtime.spawn(connect);
+
+    // We don't need it anymore and dropping the endpoint will make it's driver finish eventually.
+    drop(endpoint);
+
+    runtime.run()?;
+    Ok(())
+}
--- a/quinn/examples/insecure_connection.rs
+++ b/quinn/examples/insecure_connection.rs
@@ -1,0 +1,109 @@
+//! Demonstrates how to make a QUIC connection that ignores the server certificate.
+//!
+//! Run:
+//! ```text
+//! $ cargo run --example insecure_connection --features="rustls/dangerous_configuration"
+//! ```
+
+use futures::{Future, Stream};
+use std::error::Error;
+use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4, ToSocketAddrs};
+use std::sync::Arc;
+use tokio::runtime::current_thread::{self, Runtime};
+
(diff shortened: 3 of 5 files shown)


quinn/examples/single_socket.rs
```diff
@@ -0,0 +1,144 @@
+//! QUIC does not care about the local endpoints - connection is not associated with source IP and
```
djc: I think this could be potentially confusing. We have the `Endpoint` type which should correspond to exactly one UDP address (IP + port), otherwise things will go wrong. So QUIC does care about local endpoints -- the key point you're trying to make, I think, is that we can run an arbitrary number of connections over a single endpoint. (Also this documentation does not discuss that you can have "server" and "client" connections on the same endpoint, but that might be worth pointing out as well?)

povilasb: Removed the misleading info, but I'm not sure for 
> Also this documentation does not discuss that you can have "server" and "client" connections on the same endpoint

You mean pointing out that "server" and "client" both can have the same endpoint (`source_ip:port`)? If so, maybe a separate example for this case?

Although, I don't know the real world use case for this and hence such example seems synthetic to me :/

djc: For example in a peer-to-peer scenario, a particular process might be acting as the "server" for one peer, while it acts as the "client" for another peer.

djc:
TBH I do have a concern that these three examples between each other duplicate quite a lot of code even though the variations among them are conceptually pretty minor. Maybe it would be nicer to demonstrate some other usage types within the same file?

quinn/examples/single_socket.rs
povilasb: good point, will put this in the docs at the very least :)

povilasb:
> TBH I do have a concern that these three examples between each other duplicate quite a lot of code even though the variations among them are conceptually pretty minor. Maybe it would be nicer to demonstrate some other usage types within the same file?

Makes sense. What about extracting common code to a separate module thouh? :thinking:

djc:
Separate module works for me, I don't know whether/how the Cargo examples structure facilitates that.

povilasb:
I just pushed the changes, see if that looks better :)
I like it this way: now there is a `common.rs` module that exposes commonly used building blocks and then each example composes them in a specific usable way.

djc:
Looks pretty good to me. @Ralith, do you want to take a look?

quinn/examples/insecure_connection.rs
```diff
@@ -0,0 +1,106 @@
+//! Demonstrates how to make a QUIC connection that ignores the server certificate.
+//!
+//! Run:
+//! ```
+//! $ cargo run --example insecure_connection --features="dangerous_configuration"
```
Ralith: How does this work? We don't expose such a feature ourselves.

quinn/examples/common.rs
```diff
@@ -0,0 +1,79 @@
+//! Commonly used code in most examples.
+
+use failure::Error;
```
Ralith: Could uses of this here and in other examples be replaced with `Box<std::error::Error>`? failure seems to be on the way out, so I'd prefer not to introduce new uses of it.

quinn/examples/connection.rs
```diff
@@ -0,0 +1,62 @@
...
+
+const SERVER_PORT: u16 = 5000;
+
+fn main() -> Result<(), Error> {
+    let mut runtime = Runtime::new()?;
+
+    let (driver, incoming, server_cert) =
+        make_server_endpoint(&ipv4_addr(0, 0, 0, 0, SERVER_PORT))?;
+    // drive UDP socket
+    runtime.spawn(driver.map_err(|e| eprintln!("IO error: {}", e)));
```
Ralith: Because a driver failure is fatal to the associated endpoint, I think it would be better to panic here (e.g. `panic!("IO error: {}", e);`). We don't want people following these examples and getting confused when a poorly configured server hangs doing nothing because they missed a single line on stderr.

quinn/examples/common.rs
```diff
@@ -0,0 +1,79 @@
...
+    let mut endpoint_builder = Endpoint::new();
+    endpoint_builder.listen(server_config);
+    let (_endpoint, driver, incoming) = endpoint_builder.bind(bind_addr)?;
+    Ok((driver, incoming, server_cert))
+}
+
+/// Convenience function to build IPv4 address with port.
+pub fn ipv4_addr(a: u8, b: u8, c: u8, d: u8, port: u16) -> SocketAddr {
+    SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(a, b, c, d), port))
+}
```
Ralith: Would it make sense to make this unnecessary by modifying functions to take an impl of [ToSocketAddrs](https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html) instead of a concrete `SocketAddr`? That's what `bind` takes, allowing traditional socket address syntax to be supplied as strings.

quinn/examples/insecure_connection.rs
povilasb: This feature is exposed by `rustls`. Since TLS is first class citizen of QUIC, I figured it's useful to have this example as well.

quinn/examples/common.rs
povilasb: > failure seems to be on the way out
You mean you're switching to smth else for error handling?

quinn/examples/common.rs
povilasb: I would prefer to use [net-literals](https://crates.io/crates/net-literals) crate. Just didn't want to introduce any new dependencies without confirmation. Would you be okay with it as a dev dependency?

povilasb: Actually, let me see what I can do with the `ToSocketAddrs` trait - it seems pretty neat, thx for bringing this up :)

djc: Yeah, I would say using net-literals for this is somewhat idiosyncratic, which is not a great property for example code.

quinn/examples/common.rs
Ralith: We stopped using failure internally some time ago.

quinn/examples/insecure_connection.rs
Ralith: Right, but how does `--features` know you're talking about rustls? Quinn itself has no such feature name.

povilasb: I believe cargo defines features for a whole project so they get passed to all libraries and their dependencies.

povilasb:
I addressed all your comments and pushed the changes.

quinn/examples/common/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+
+use quinn::{
+    Certificate, CertificateChain, ClientConfig, ClientConfigBuilder, Driver, Endpoint, Incoming,
+    PrivateKey, ServerConfig, ServerConfigBuilder, TransportConfig,
+};
+use std::error::Error;
+use std::net::ToSocketAddrs;
+use std::sync::Arc;
+
+/// Returns QUIC endpoint builder configured to build a client endpoint.
```
Ralith: This does not return a builder.

```suggestion
/// Constructs a QUIC endpoint configured for use as a client only.
```

quinn/examples/common/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+///
+/// ## Args
+///
+/// - server_certs: list of trusted certificates.
+#[allow(unused)]
```
Ralith: Why is this `allow` needed?

quinn/examples/common/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+    server_certs: &Vec<Vec<u8>>,
+) -> Result<(Endpoint, Driver), Box<Error>> {
+    let client_cfg = configure_client(server_certs)?;
+    let mut endpoint_builder = Endpoint::new();
+    endpoint_builder.default_client_config(client_cfg);
+    let (endpoint, driver, _incoming) = endpoint_builder.bind(bind_addr)?;
+    Ok((endpoint, driver))
+}
+
+/// Configures server to listen on given address and port.
```
Ralith: This does not modify a configuration, but rather constructs an Endpoint with a particular sort of configuration.
```suggestion
/// Constructs a QUIC endpoint configured to listen for incoming connections on a certain address and port.
```

quinn/examples/insecure_connection.rs
```diff
@@ -0,0 +1,107 @@
...
+mod common;
+use common::make_server_endpoint;
+
+const SERVER_PORT: u16 = 5000;
+
+fn main() -> Result<(), Box<Error>> {
+    let mut runtime = Runtime::new()?;
+
+    // server and client are running on the same thread asynchronously
+    run_server(&mut runtime, ("0.0.0.0", SERVER_PORT))?;
```
Ralith: I didn't know about this `ToSocketAddrs` impl; neat!

quinn/examples/single_socket.rs
```diff
@@ -0,0 +1,98 @@
+//! You can have multiple QUIC connections over a single UDP socket. This is especially
+//! useful, if you are building a peer-to-peer system where you potentially need to communicate with
+//! thousands of peers or if you have a
+//! [hole punched](https://en.wikipedia.org/wiki/UDP_hole_punching) UDP socket.
+//! In addition, both QUIC server and client can operate on the same UDP socket.
```
Ralith:
```suggestion
//! In addition, QUIC servers and clients can both operate on the same UDP socket.
```

quinn/examples/connection.rs
```diff
@@ -0,0 +1,62 @@
+//! This example intends to use the smallest amout of code to make a simple QUIC connection.
+//!
+//! The server issues it's own certificate and passes it to the client to trust.
+//!
+//! Run:
+//! ```
+//! $ cargo run --example connection
+//! ```
+//!
+//! This example will make a QUIC connection on localhost and you should see smth like on stdout:
```
Ralith:
```suggestion
//! This example will make a QUIC connection on localhost, and you should see output like:
```

quinn/examples/insecure_connection.rs
Ralith: Are you sure? It doesn't seem to work here:
```
> cargo run --example insecure_connection --features="dangerous_configuration"
error: Package `quinn v0.2.0 (/home/ralith/src/quinn/quinn)` does not have these features: `dangerous_configuration`
```

Additionally, this breaks `cargo test` because that doesn't know about the [required feature](https://doc.rust-lang.org/cargo/reference/manifest.html#the-required-features-field-optional).

quinn/examples/connection.rs
```diff
@@ -0,0 +1,62 @@
+//! This example intends to use the smallest amout of code to make a simple QUIC connection.
+//!
+//! The server issues it's own certificate and passes it to the client to trust.
+//!
+//! Run:
+//! ```
+//! $ cargo run --example connection
+//! ```
```
Ralith: I think you need to annotate all these code blocks (probably as text) so rustdoc doesn't try to compile and run them as doctests. Might not be an issue for examples in practice, but good style to be explicit all the same.

quinn/examples/common/mod.rs
povilasb: `insecure_connection` example does not use this function and when it includes `common` module the compiler complains about unused function.

quinn/examples/connection.rs
povilasb: Makes sense, will change it. Although  `cargo doc` builds successfully.

povilasb:
Regarding `cargo run --example insecure_connection`, I'm actually not that sure anymore. I tried running without the feature flag and it works for me. To be honest I'm a bit lost because those functions should only be available when `dangerous_configuration` feature is on: https://github.com/ctz/rustls/blob/v/0.15.0/src/client/mod.rs#L221

@Ralith does `cargo run --example insecure_connection` work for you?

Ralith:
It looks like it coincidentally works when you invoke `cargo` from the workspace root, rather than the `quinn` subdir, because the `interop` workspace crate incidentally enables the required feature. I'm not a huge fan of relying on that, or of the resulting breakage of `cargo test` from inside the `quinn` subdir.

povilasb:
> It looks like it coincidentally works when you invoke `cargo` from the workspace root, rather than the `quinn` subdir, because the `interop` workspace crate incidentally enables the required feature. I'm not a huge fan of relying on that, or of the resulting breakage of `cargo test` from inside the `quinn` subdir.

good catch, thanks, will fix this

povilasb:
Pushed the changes ;)

quinn/examples/single_socket.rs
povilasb: `insecure_connection` example does not use this function and when it includes `common` module the compiler complains about unused function.

quinn/examples/common/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+        });
+    current_thread::spawn(task);
+}
+
+/// Builds default quinn client config and trusts given certificates.
+///
+/// ## Args
+///
+/// - server_certs: a list of trusted certificates in DER format.
+fn configure_client(server_certs: &Vec<Vec<u8>>) -> Result<ClientConfig, Box<Error>> {
```
Ralith:
```suggestion
fn configure_client(server_certs: &[&[u8]]) -> Result<ClientConfig, Box<Error>> {
```

Unless there's a specific reason this needs to know about the container type? Seems like we're only passing in statically sized sets of certs anyway.

quinn/examples/common/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+
+/// Constructs a QUIC endpoint configured for use a client only.
+///
+/// ## Args
+///
+/// - server_certs: list of trusted certificates.
+#[allow(unused)]
+pub fn make_client_endpoint<A: ToSocketAddrs>(
+    bind_addr: A,
+    server_certs: &Vec<Vec<u8>>,
```
Ralith:
```suggestion
    server_certs: &[&[u8]],
```

quinn/examples/connection.rs
```diff
@@ -0,0 +1,63 @@
...
+//!
+//! The server issues it's own certificate and passes it to the client to trust.
+//!
+//! Run:
+//! ```text
+//! $ cargo run --example connection
+//! ```
+//!
+//! This example will make a QUIC connection on localhost, and you should see output like:
+//! ```
```
Ralith: Missing `text` annotation

quinn/examples/single_socket.rs
```diff
@@ -0,0 +1,99 @@
...
+//!
+//! This example demonstrate how to make multiple outgoing connections on a single UDP socket.
+//!
+//! Run:
+//! ```text
+//! $ cargo run --example single_socket
+//! ```
+//!
+//! The expected output should be something like:
+//! ```
```
Ralith: Missing `text` annotation

povilasb:
@Ralith I addressed all your comments ;)

PR #249, opened 2019-03-19
API for obtaining our local endpoint

ustulation:
This is especially useful if we were binding to port 0 intending the OS to allocate us a random port. Then we would like to know what port was actually allocated. Besides it's a useful API anyway as it's natural to want to know this detail from a bound socket for completeness.

### DIFF
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -171,6 +171,11 @@ impl Endpoint {
         Ok(())
     }
 
+    /// Get the local endpoint the underlying socket is bound to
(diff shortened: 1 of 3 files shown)


quinn/src/lib.rs
```diff
@@ -171,6 +171,11 @@ impl Endpoint {
         Ok(())
     }
 
+    /// Get the local endpoint the underlying socket is bound to
```
Ralith:
```suggestion
    /// Get the local `SocketAddr` the underlying socket is bound to
```

Issue #252, opened 2019-03-20
Relation between Connection and NewStream/RecvStream/WriteStream


ustulation:
Further to the [bug report here](https://github.com/djc/quinn/issues/250), I can also see that children streams spawned via `IncomingStreams` stream are left untouched by `Connection::close()`. What this means is that for e.g., a `read_to_end` on a `RecvStream` will remain potentially unresolved once the peer has managed to successfully open it and send some stuff on it and then we decide to close the connection to it. Even if we fixed the linked bug and resolve the `IncomingStreams` to completeness, all the children streams continue to exist unresolved until the peer takes some action on them. They can continue to successfully do `write_all` on them on only once they call `shutdown` etc. will they make us resolve `our read_to_end`.

To manually patch this I'm currently keeping extra state such as a `tokio::sync::watch::Sender` which broadcasts to all the children streams to resolve them via stream/future selector (`select/select2`).

 This should also be fixed in quinn ?

Ralith:
This should be handled by `Pending::fail`, but we may have a bug where don't call that for locally-closed connections.

ustulation:
I'm happy to close this now as i can see all stuff regarding and spawning from this are resolved (at-least in my tests), or the admins can close it later if they think they need to work on this more somewhere.

Issue #251, opened 2019-03-20
Dropping the connection does not close it


ustulation:
I can't seem to find a `Drop` impl for [EndpointInner](https://github.com/djc/quinn/blob/0a967f66e2a603a6a7849358fa5aeae18ada9bad/quinn/src/lib.rs#L528) so the dropping the [Connection](https://github.com/djc/quinn/blob/0a967f66e2a603a6a7849358fa5aeae18ada9bad/quinn/src/lib.rs#L659) does not actually do anything like [the doc](https://docs.rs/quinn/0.2.0/quinn/struct.Connection.html) says ([in code](https://github.com/djc/quinn/blob/0a967f66e2a603a6a7849358fa5aeae18ada9bad/quinn/src/lib.rs#L654).

I don't know if that's ^^^ where the `Drop` trait should be as I haven't gone deep into the source code but the end result is:

So what happens is even after I have dropped the `Connection` obtained via listening for connections, the remote can still successfully open a stream (tested with uni-directional stream) to us and continue sending stuff which we actually get a trigger for. This is a deviation from the behaviour which I see if I call close() expliclitly in which case the remote peer gets `ConnectionAborted` error and rightly so, when it tries to open a stream for writing something to us after we have closed the connection.

djc:
There does seem to be a `Drop` impl for `ConnectionInner`?

https://github.com/djc/quinn/blob/0a967f66e2a603a6a7849358fa5aeae18ada9bad/quinn/src/lib.rs#L799

ustulation:
Yes that true. Then is it [this bit here](https://github.com/djc/quinn/blob/0a967f66e2a603a6a7849358fa5aeae18ada9bad/quinn/src/lib.rs#L805) that's causing the problem by not actually calling the [actual close](https://github.com/djc/quinn/blob/0a967f66e2a603a6a7849358fa5aeae18ada9bad/quinn/src/lib.rs#L809) ? I'm just hazarding a guess as I've not looked much into quinn src code and there's difference in behaviour between a drop and an explicit close() so there's something somewhere not quite right.
The other reason could be that I have clones in my code preventing the dtor to be run :P - though i don't think so but ofc always a possibility.

Ralith:
Thanks for investigating! That particular line is probably not the issue, as the "drained" state is only entered after the connection has already been fully closed; sending another close frame wouldn't accomplish anything.

ustulation:
[...]

Issue #250, opened 2019-03-20
Relation between Connection and IncomingStreams


ustulation:
On a successful connection in either direction (i.e., either actively via connect or passively via listen) I get two things eventually - one `Connection` and one `IncomingStreams`. As a user i expect these to be related at some level since they denote _a_ connection to the peer. However it seems that closing the connection does nothing for the `IncomingStreams`.

Details:

0. I get a `Connection` and `IncomingStreams` from a peer while I'm listening
1. Something happens and I decide I don't want this peer any more
2. I get the `Connection` object and do `Connection::close(...)`.
3. Nothing happens to the `IncomingStreams` - it continue to stay indefinitely with tokio event loop - I would have expected it to resolve to completion/failure at this point and destroy itself.
4. Remote peer still tries to send something to us on their connection to us (Which is still alive for them) by trying to open a new stream to us. The peer keeps getting "ConnectionAborted - closed by remote peer" error which is fine and expected.

So the `IncomingStreams` stream (and possibly any other stream obtained from it which the remote hasn't shutdown/closed/destroyed yet - though I haven't checked this part) uselessly remains unresolved with tokio. I have to now keep extra knowledge about closing these streams when I close the `Connection` to the peer.

Wouldn't the better/expected design be that when I close `Connection` (or drop/destroy it) all the related stuff resolve into an error (or anything, but resolve) to gracefully collect all resources ?

djc:
Yup, that seems like an oversight we should fix. Do you want to try a PR? We probably want a `dropped` flag in `ConnectionInner` similar to the one in `EndpointInner`; if `IncomingStreams` is polled and `dropped` is true, it should probably return `Ready(None)`.

ustulation:
Is the `IncomingStreams` definitely notified when the drop flag is set ? Otherwise could be quite a while between the drop flag being set and when the stream is polled (which will probably be when the remote tries opening one to us making us be at the mercy of the remote peer). If that's true, maybe have IncomingStreams pass through a `select` (some stream or future selector) first which selects  `IncomingStreams` or `Blah` (whatever we name it) with `Blah` being a type that gets notified when something fires on drop and then resolves to `Ready(None)` or `Err` or whatever. E.g, `Blah` could be a oneshot or mpsc rx or something custom atop it or whatever makes sense

djc:
When the `dropped` flag is set, we should likely also notify the `pending.incoming_streams_reader`.

[...]

PR #253, opened 2019-03-21
[WIP] Represent connection closing as instant to the application layer

Ralith:
Possible fix for #250, #251, #252. Theoretically complete, but I need to go over this again when I'm more rested to make sure I haven't broken any of the tokio layer's state sync.

### DIFF
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -713,7 +713,10 @@ impl Connection {
         let endpoint = &mut *self.0.endpoint.borrow_mut();
 
         let pending = endpoint.pending.get_mut(&self.0.handle).unwrap();
-        assert!(!pending.closing, "a connection can only be closed once");
+        if pending.closing {
+            debug!(endpoint.log, "connection closed redundantly");
+            return;
+        }
         pending.closing = true;
         pending.fail(ConnectionError::ApplicationClosed {
             reason: quinn::ApplicationClose {
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -438,9 +438,6 @@ impl Future for Driver {
                         if timer == quinn::Timer::Close {
                             // Connection drained
                             if let hash_map::Entry::Occupied(mut p) = endpoint.pending.entry(ch) {
-                                if let Some(x) = p.get_mut().closing.take() {
-                                    let _ = x.send(());
-                                }
                                 if p.get().dropped {
                                     p.remove();
                                 } else {
@@ -558,7 +555,7 @@ struct Pending {
     incoming_streams_reader: Option<Task>,
     finishing: FnvHashMap<StreamId, oneshot::Sender<Option<ConnectionError>>>,
     error: Option<ConnectionError>,
-    closing: Option<oneshot::Sender<()>>,
+    closing: bool,
     dropped: bool,
     drained: bool,
 }
@@ -575,7 +572,7 @@ impl Pending {
             incoming_streams_reader: None,
             finishing: FnvHashMap::default(),
             error: None,
-            closing: None,
+            closing: false,
             dropped: false,
             drained: false,
         }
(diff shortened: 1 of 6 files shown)


quinn/src/lib.rs
```diff
@@ -712,28 +709,22 @@ impl Connection {
     /// # Panics
     /// - If called more than once on handles to the same connection
     // FIXME: Infallible
-    pub fn close(&self, error_code: u16, reason: &[u8]) -> impl Future<Item = (), Error = ()> {
-        let (send, recv) = oneshot::channel();
-        {
-            let endpoint = &mut *self.0.endpoint.borrow_mut();
+    pub fn close(&self, error_code: u16, reason: &[u8]) {
```
djc: Should this consume `self`?

quinn/src/lib.rs
```diff
@@ -712,28 +709,22 @@ impl Connection {
...
+        let endpoint = &mut *self.0.endpoint.borrow_mut();
 
-            let pending = endpoint.pending.get_mut(&self.0.handle).unwrap();
-            assert!(
-                pending.closing.is_none(),
-                "a connection can only be closed once"
-            );
-            pending.closing = Some(send);
+        let pending = endpoint.pending.get_mut(&self.0.handle).unwrap();
+        assert!(!pending.closing, "a connection can only be closed once");
```
djc: Why is this assert necessary? Can't we just early return out if `closing` is `true` already? Might make for a friendlier API.

quinn/src/lib.rs
```diff
@@ -805,13 +796,20 @@ impl Drop for ConnectionInner {
...
                     .close(Instant::now(), self.handle, 0, (&[][..]).into());
                 if let Some(x) = endpoint.driver.as_ref() {
                     x.notify();
                 }
+                pending
+                    .get_mut()
+                    .fail(ConnectionError::TransportError(quinn::TransportError {
+                        code: quinn::TransportErrorCode::INTERNAL_ERROR,
+                        frame: None,
+                        reason: "driver future was dropped".to_string(),
```
djc: I find this error a bit confusing. In my mind the `driver` concept is associated with a particular thing (in our case only the `Driver`), where in this case all references to the `ConnectionInner` have been dropped. I think these are held by `Connection` (which I wouldn't really count as a driver) and `IncomingStreams` (same).

Maybe something like `all references to the connection were dropped`?

Also, what's our take on a `Connection` being alive without `IncomingStreams` (should be okay I think) or an `IncomingStreams` being alive without the `Connection` (maybe more iffy?)?

quinn/src/lib.rs
Ralith: It actually did in the past. Because stable futures make it so difficult to get anything done without taking ownership, `Connection` became `Clone`, with the consequence that consuming `self` doesn't accomplish much. We could get rid of that impl and encourage people to use `Rc` manually, but then closing a connection will require `Rc::try_unwrap`, which makes me worry about ergonomics.

quinn/src/lib.rs
Ralith: > I find this error a bit confusing.

Whoops, certainly did not intend to copy that error message verbatim. Good catch!

> Also, what's our take on a Connection being alive without IncomingStreams (should be okay I think) or an IncomingStreams being alive without the Connection (maybe more iffy?)?

I think these are both reasonable cases. For the `Connection`-only case, not all application protocols will involve streams in both directions being continually opened, and for `IncomingStreams`-only, an application layer might not have any use for the various accessors once the connection gets going in earnest, so forcing them to keep the `Connection` around would be unnatural.

quinn/src/lib.rs
Ralith: I tend waffle about friendliness vs. making unintended behavior obvious to its author, but I think you've sold me okay on "transport protocol impls shouldn't take out your process."

djc: I do also want to make unintended behavior obvious to the author (we could return a `Result` here), but if there is actually no ill effect then it we may not have to signal it.

quinn/src/lib.rs
djc: I would just argue that even if consuming `self` is somewhat pointless (because it's trivial and cheap to `clone()`), there might still be a benefit in terms of documenting the expected behavior and communicating how the system is supposed to work.

djc: (In this case, the user couldn't just call `close()` on it and continue running as normal, they'd have to have ownership of the `Endpoint`, detaching it from its owner.)

Ralith:
This appears to have uncovered a bug in timer notifications, likely relating to new connections being created while the old ones are still being shut down.

Ralith:
Due to rebase hazard I'll probably let this be blocked on #255, then reprise if the issue remains after the merge.

djc:
This is obviated by #271.

PR #255, opened 2019-03-23
Return Connections from Endpoints after creation

djc:
This is the big one.

Currently not quite done: we need to figure out how to handle the connection closing sequence. Since `Endpoint::close()` is no more, `app_closed` never gets set, so the `ConnectionMeta` state doesn't get cleaned up (I think this is why `zero_length_cid` fails). The quinn tests hang and there is some unused code having to do with cancelable timers that probably should be wired up somehow.

Even after all that gets fixed, I'd prefer not to wait for this to be perfect before we merge it, since it's pretty invasive -- I'll follow up with improvements soon after. However, I'd like to get this to pass the tests and not introduce any substantial functional regressions. Things I'd prefer to address after merging this PR:

- Use of unbounded channels
- Lack of documentation updates
- Performance regressions (unless extreme)
- Move from `Rc<RefCell<>>` to `Arc<Mutex<>>`

Would like your review!

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -515,7 +515,7 @@ impl Connection {
         self.space_mut(space).pending_acks.subtract(&info.acks);
     }
 
-    fn timeout(&mut self, now: Instant, timer: Timer) {
+    pub(crate) fn timeout(&mut self, now: Instant, timer: Timer) {
         match timer {
             Timer::Close => {
                 self.state = State::Drained;
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -1,13 +1,13 @@
+use std::cmp;
 use std::collections::{HashMap, VecDeque};
 use std::net::SocketAddr;
 use std::ops::{Index, IndexMut};
 use std::sync::Arc;
 use std::time::{Duration, Instant, SystemTime};
-use std::{cmp, mem};
 
-use bytes::{BufMut, Bytes, BytesMut};
+use bytes::{BufMut, BytesMut};
 use err_derive::Error;
-use fnv::{FnvHashMap, FnvHashSet};
+use fnv::FnvHashMap;
 use rand::{rngs::OsRng, Rng, RngCore};
 use ring::digest;
 use ring::hmac::SigningKey;
@@ -17,14 +17,12 @@ use slog::{self, Logger};
 use crate::coding::BufMutExt;
 use crate::connection::{
     initial_close, ClientConfig, Connection, ConnectionError, ConnectionEvent, EndpointEvent,
-    TimerUpdate,
 };
 use crate::crypto::{
     self, reset_token_for, Crypto, CryptoClientConfig, CryptoServerConfig, RingHeaderCrypto,
     TokenKey,
 };
 use crate::packet::{ConnectionId, EcnCodepoint, Header, Packet, PacketDecodeError, PartialDecode};
-use crate::stream::{ReadError, WriteError};
 use crate::transport_parameters::TransportParameters;
 use crate::{
     varint, Directionality, Side, StreamId, Transmit, TransportError, LOC_CID_COUNT, MAX_CID_SIZE,
@@ -43,17 +41,9 @@ pub struct Endpoint {
     connection_ids_initial: FnvHashMap<ConnectionId, ConnectionHandle>,
     connection_ids: FnvHashMap<ConnectionId, ConnectionHandle>,
     connection_remotes: FnvHashMap<SocketAddr, ConnectionHandle>,
-    pub(crate) connections: Slab<ConnectionMeta>,
+    connections: Slab<ConnectionMeta>,
     config: Arc<EndpointConfig>,
     server_config: Option<Arc<ServerConfig>>,
-    /// Connections that might have timer updates to apply perform
-    dirty_timers: FnvHashSet<ConnectionHandle>,
-    /// Connections that might have packets to send
-    needs_transmit: FnvHashSet<ConnectionHandle>,
-    /// Connections that might have endpoint-facing events to report
-    endpoint_eventful_conns: FnvHashSet<ConnectionHandle>,
-    /// Connections that might have application-facing events to report
-    eventful_conns: FnvHashSet<ConnectionHandle>,
     incoming_handshakes: usize,
 }
 
@@ -73,10 +63,6 @@ impl Endpoint {
             connection_ids: FnvHashMap::default(),
             connection_remotes: FnvHashMap::default(),
             connections: Slab::new(),
-            dirty_timers: FnvHashSet::default(),
-            needs_transmit: FnvHashSet::default(),
-            endpoint_eventful_conns: FnvHashSet::default(),
-            eventful_conns: FnvHashSet::default(),
             incoming_handshakes: 0,
             config,
             server_config,
@@ -87,73 +73,21 @@ impl Endpoint {
         self.server_config.is_some()
     }
 
-    /// Get an application-facing event
-    pub fn poll(&mut self) -> Option<(ConnectionHandle, Event)> {
-        self.handle_endpoint_events();
-        while let Some(&ch) = self.eventful_conns.iter().next() {
-            if let Some(e) = self.connections[ch].conn.poll() {
-                return Some((ch, e));
-            }
-            self.eventful_conns.remove(&ch);
-        }
-        None
-    }
-
-    pub(crate) fn handle_endpoint_events(&mut self) {
-        let mut conns = mem::replace(&mut self.endpoint_eventful_conns, FnvHashSet::default());
-        for ch in conns.drain() {
-            while let Some(event) = self.connections[ch].conn.poll_endpoint_events() {
-                let done = if let EndpointEvent::Closed { .. } = &event {
-                    true
-                } else {
-                    false
-                };
-                self.handle_event(ch, event);
-                if done {
-                    break;
-                }
-            }
-        }
-        mem::replace(&mut self.endpoint_eventful_conns, conns);
-    }
-
-    /// Get a pending timer update
-    pub fn poll_timers(&mut self) -> Option<(ConnectionHandle, TimerUpdate)> {
-        loop {
-            let &ch = self.dirty_timers.iter().next()?;
-            loop {
-                if let Some(update) = self.connections[ch].conn.poll_timers() {
-                    return Some((ch, update));
-                } else {
-                    self.dirty_timers.remove(&ch);
-                    break;
-                }
-            }
-        }
-    }
-
     /// Get the next packet to transmit
-    pub fn poll_transmit(&mut self, now: Instant) -> Option<Transmit> {
-        if let Some(x) = self.transmits.pop_front() {
-            return Some(x);
-        }
-        loop {
-            let &ch = self.needs_transmit.iter().next()?;
-            if let Some(transmit) = self.connections[ch].conn.poll_transmit(now) {
-                self.dirty_timers.insert(ch);
-                return Some(transmit);
-            } else {
-                self.needs_transmit.remove(&ch);
-            }
-        }
+    pub fn poll_transmit(&mut self) -> Option<Transmit> {
+        self.transmits.pop_front()
     }
 
-    pub fn handle_event(&mut self, ch: ConnectionHandle, event: EndpointEvent) {
+    pub fn handle_event(
+        &mut self,
+        ch: ConnectionHandle,
+        event: EndpointEvent,
+    ) -> Option<ConnectionEvent> {
         match event {
             EndpointEvent::NeedIdentifiers => {
                 if self.config.local_cid_len != 0 {
                     // We've already issued one CID as part of the normal handshake process.
-                    self.send_new_identifiers(ch, LOC_CID_COUNT - 1);
+                    return Some(self.send_new_identifiers(ch, LOC_CID_COUNT - 1));
                 }
             }
             EndpointEvent::RetireConnectionId(seq) => {
@@ -165,14 +99,10 @@ impl Endpoint {
                         cid = cid,
                     );
                     self.connection_ids.remove(&cid);
-                    self.send_new_identifiers(ch, 1);
+                    return Some(self.send_new_identifiers(ch, 1));
                 }
             }
             EndpointEvent::Closed { remote } => {
-                if !self.connections[ch].app_closed {
-                    return;
-                }
-
                 let init_cid = self.connections[ch].init_cid;
                 if init_cid.len() > 0 {
                     self.connection_ids_initial.remove(&init_cid);
@@ -183,13 +113,10 @@ impl Endpoint {
                 }
 
                 self.connection_remotes.remove(&remote);
-                self.dirty_timers.remove(&ch);
-                self.eventful_conns.remove(&ch);
-                self.endpoint_eventful_conns.remove(&ch);
-                self.needs_transmit.remove(&ch);
                 self.connections.remove(ch.0);
             }
         }
+        None
     }
 
     /// Process an incoming UDP datagram
@@ -199,7 +126,7 @@ impl Endpoint {
         remote: SocketAddr,
         ecn: Option<EcnCodepoint>,
         data: BytesMut,
-    ) -> Option<ConnectionHandle> {
+    ) -> Option<(ConnectionHandle, DatagramEvent)> {
         let datagram_len = data.len();
         let (first_decode, remaining) = match PartialDecode::new(data, self.config.local_cid_len) {
             Ok(x) => x,
@@ -259,19 +186,16 @@ impl Endpoint {
                 .cloned()
         };
         if let Some(ch) = known_ch {
-            let event = ConnectionEvent::Datagram {
-                now,
-                remote,
-                ecn,
-                first_decode,
-                remaining,
-            };
-            self.connections[ch].conn.handle_event(event);
-            self.needs_transmit.insert(ch);
-            self.dirty_timers.insert(ch);
-            self.endpoint_eventful_conns.insert(ch);
-            self.eventful_conns.insert(ch);
-            return None;
+            return Some((
+                ch,
+                DatagramEvent::ConnectionEvent(ConnectionEvent::Datagram {
+                    now,
+                    remote,
+                    ecn,
+                    first_decode,
+                    remaining,
+                }),
+            ));
         }
 
         //
@@ -289,7 +213,7 @@ impl Endpoint {
         }
 
         if first_decode.has_long_header() {
-            if first_decode.is_initial() {
+            return if first_decode.is_initial() {
                 if datagram_len < MIN_INITIAL_SIZE {
                     debug!(
                         self.log,
@@ -302,8 +226,8 @@ impl Endpoint {
                 let crypto = Crypto::new_initial(&dst_cid, Side::Server);
                 let header_crypto = crypto.header_crypto();
                 match first_decode.finish(Some(&header_crypto)) {
-                    Ok(packet) => {
-                        return self.handle_initial(
+                    Ok(packet) => self
+                        .handle_initial(
                             now,
                             remote,
                             ecn,
@@ -311,21 +235,21 @@ impl Endpoint {
                             remaining,
                             &crypto,
                             &header_crypto,
-                        );
-                    }
+                        )
+                        .map(|(ch, conn)| (ch, DatagramEvent::NewConnection(conn))),
                     Err(e) => {
                         trace!(self.log, "unable to decode packet"; "reason" => %e);
+                        None
                     }
                 }
-                return None;
             } else {
                 debug!(
                     self.log,
                     "ignoring non-initial packet for unknown connection {connection}",
                     connection = dst_cid
                 );
-                return None;
-            }
+                None
+            };
         }
 
         //
@@ -390,11 +314,11 @@ impl Endpoint {
         transport_config: Arc<TransportConfig>,
         crypto_config: Arc<crypto::ClientConfig>,
         server_name: &str,
-    ) -> Result<ConnectionHandle, ConnectError> {
+    ) -> Result<(ConnectionHandle, Connection), ConnectError> {
         transport_config.validate(&self.log)?;
         let remote_id = ConnectionId::random(&mut self.rng, MAX_CID_SIZE);
         trace!(self.log, "initial dcid"; "value" => %remote_id);
-        let ch = self.add_connection(
+        let (ch, conn) = self.add_connection(
             remote_id,
             remote_id,
             remote,
@@ -404,11 +328,10 @@ impl Endpoint {
                 server_name: server_name.into(),
             }),
         )?;
-        self.needs_transmit.insert(ch);
-        Ok(ch)
+        Ok((ch, conn))
     }
 
-    fn send_new_identifiers(&mut self, ch: ConnectionHandle, num: usize) {
+    fn send_new_identifiers(&mut self, ch: ConnectionHandle, num: usize) -> ConnectionEvent {
         let mut ids = vec![];
         for _ in 0..num {
             let cid = self.new_cid();
@@ -419,9 +342,7 @@ impl Endpoint {
             meta.loc_cids.insert(seq, cid);
             ids.push((seq, cid));
         }
-        self.connections[ch]
-            .conn
-            .handle_event(ConnectionEvent::NewIdentifiers(ids));
+        ConnectionEvent::NewIdentifiers(ids)
     }
 
     fn new_cid(&mut self) -> ConnectionId {
@@ -441,7 +362,7 @@ impl Endpoint {
         remote: SocketAddr,
         transport_config: Arc<TransportConfig>,
         opts: ConnectionOpts,
-    ) -> Result<ConnectionHandle, ConnectError> {
+    ) -> Result<(ConnectionHandle, Connection), ConnectError> {
         let loc_cid = self.new_cid();
         let params = TransportParameters::new(&transport_config);
         let (tls, client_config) = match opts {
@@ -484,11 +405,9 @@ impl Endpoint {
             remote_validated,
         );
         let id = self.connections.insert(ConnectionMeta {
-            conn,
             init_cid,
             cids_issued: 0,
             loc_cids: HashMap::new(),
-            app_closed: false,
         });
         let ch = ConnectionHandle(id);
 
@@ -496,7 +415,7 @@ impl Endpoint {
             self.connection_ids.insert(loc_cid, ch);
         }
         self.connection_remotes.insert(remote, ch);
-        Ok(ch)
+        Ok((ch, conn))
     }
 
     fn handle_initial(
@@ -508,7 +427,7 @@ impl Endpoint {
         rest: Option<BytesMut>,
         crypto: &Crypto,
         header_crypto: &RingHeaderCrypto,
-    ) -> Option<ConnectionHandle> {
+    ) -> Option<(ConnectionHandle, Connection)> {
         let (src_cid, dst_cid, token, packet_number) = match packet.header {
             Header::Initial {
                 src_cid,
@@ -616,7 +535,7 @@ impl Endpoint {
             }
         }
 
-        let ch = self
+        let (ch, mut conn) = self
             .add_connection(
                 dst_cid,
                 src_cid,
@@ -630,22 +549,16 @@ impl Endpoint {
         if dst_cid.len() != 0 {
             self.connection_ids_initial.insert(dst_cid, ch);
         }
-        match self.connections[ch].conn.handle_initial(
-            now,
-            remote,
-            ecn,
-            packet_number as u64,
-            packet,
-            rest,
-        ) {
+        match conn.handle_initial(now, remote, ecn, packet_number as u64, packet, rest) {
             Ok(()) => {
                 trace!(self.log, "connection incoming; ICID {icid}", icid = dst_cid);
                 self.incoming_handshakes += 1;
-                self.needs_transmit.insert(ch);
-                if self.connections[ch].conn.has_1rtt() {
-                    self.handle_event(ch, EndpointEvent::NeedIdentifiers);
+                if conn.has_1rtt() {
+                    if let Some(event) = self.handle_event(ch, EndpointEvent::NeedIdentifiers) {
+                        conn.handle_event(event);
+                    }
                 }
-                Some(ch)
+                return Some((ch, conn));
             }
             Err(e) => {
                 debug!(self.log, "handshake failed"; "reason" => %e);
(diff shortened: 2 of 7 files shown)


quinn/src/lib.rs
```diff
@@ -312,14 +297,15 @@ struct EndpointInner {
...
-    pending: FnvHashMap<ConnectionHandle, Pending>,
-    // TODO: Replace this with something custom that avoids using oneshots to cancel
-    timers: FuturesUnordered<Timer>,
-    buffered_incoming: VecDeque<ConnectionHandle>,
+    outgoing: VecDeque<quinn::Transmit>,
+    buffered_incoming: VecDeque<(ConnectionHandle, Connection)>,
     incoming: mpsc::Sender<NewConnection>,
     driver: Option<Task>,
     ipv6: bool,
+    connecting: FnvHashMap<ConnectionHandle, ConnectionDriver>,
```
Ralith: Maybe future work, but we should yield incomplete connections to the application immediately, both to save the effort of hiding them and to support 0-RTT transmits in the future.

quinn/src/lib.rs
```diff
@@ -330,50 +316,63 @@ impl EndpointInner {
...
+                                            driver: ConnectionDriver(conn.clone()),
+                                            connection: Connection(conn.clone()),
+                                            incoming: IncomingStreams(conn),
+                                        })
+                                        .unwrap();
+                                    self.inner.accept();
+                                }
+                                _ => {
+                                    self.buffered_incoming.push_back((handle, Connection(conn)));
+                                }
```
Ralith: Pre-existing issue (i.e. not a blocker), but it would probably be a lot nicer to just unconditionally buffer everything here, and let `drive_incoming` worry about the rest. So long as people aren't setting ridiculously huge accept buffer sizes, anyway.

quinn/src/lib.rs
```diff
@@ -330,50 +316,63 @@ impl EndpointInner {
...
                             }
-                            _ => {
-                                self.buffered_incoming.push_back(ch);
+                        }
+                        Some((ch, e)) => {
+                            if let Some(sender) = self.connections.get_mut(&ch) {
+                                sender.unbounded_send(e).unwrap();
+                            } else {
+                                unreachable!();
```
Ralith: Could this just be `self.connections.get_mut(&ch).unwrap()`?

quinn/src/lib.rs
```diff
@@ -330,50 +316,63 @@ impl EndpointInner {
...
+                                _ => {
+                                    self.buffered_incoming.push_back((handle, Connection(conn)));
+                                }
                             }
-                            _ => {
-                                self.buffered_incoming.push_back(ch);
+                        }
+                        Some((ch, e)) => {
+                            if let Some(sender) = self.connections.get_mut(&ch) {
+                                sender.unbounded_send(e).unwrap();
```
Ralith: Future work, but one way to make this bounded would be to construct the bulk of the driver as a `Stream<Item=(ConnectionHandle, ConnectionEvent)>`, then forward that into a multiplexing sender. That would mean a hung connection = a hung endpoint, of course.

quinn/src/lib.rs
```diff
@@ -481,137 +414,91 @@ impl EndpointInner {
 
     fn drive_send(&mut self) -> Result<bool, io::Error> {
         let mut blocked = false;
-        if let Some(ref x) = self.outgoing {
-            match self.socket.poll_send(&x.destination, x.ecn, &x.packet) {
-                Ok(Async::Ready(_)) => {
-                    self.outgoing = None;
-                }
+        while let Some(t) = self.outgoing.pop_front() {
```
Ralith: Should guard this loop against more than `IO_LOOP_BOUND` iterations.

quinn/src/lib.rs
```diff
@@ -481,137 +414,91 @@ impl EndpointInner {
...
-                    break;
-                }
-                Err(e) => {
-                    return Err(e);
-                }
-            }
-            sent += 1;
-            if sent >= IO_LOOP_BOUND {
+            if blocked {
                 return Ok(true);
```
Ralith: We should return true iff `IO_LOOP_BOUND` iterations were completed, and then or that into `keep_going` in the main loop, rather than discarding the value as is currently being done.

quinn/src/lib.rs
```diff
@@ -330,50 +316,63 @@ impl EndpointInner {
...
 
-    /// Wake up a blocked `Driver` task to process I/O
-    fn notify(&self) {
-        if let Some(x) = self.driver.as_ref() {
-            x.notify();
+    fn drive_connecting(&mut self) {
+        for new in self.connecting.values_mut() {
+            new.poll().unwrap();
         }
     }
```
Ralith: Could the driving of an in-progress connection be done by the future returned from `connect`, rendering the oneshot currently used there unnecessary?

quinn/src/lib.rs
```diff
@@ -481,137 +414,91 @@ impl EndpointInner {
...
-                            if p.get().dropped {
-                                p.remove();
-                            } else {
-                                p.get_mut().drained = true;
+                            if let Some(event) = self.inner.handle_event(ch, e) {
+                                if let hash_map::Entry::Occupied(mut entry) =
+                                    self.connections.entry(ch)
+                                {
+                                    entry.get_mut().unbounded_send(event).unwrap();
+                                }
```
Ralith: Could this use `get_mut()` instead of the entry API?

Ralith: What are the circumstances under which we expect to receive an event for an unknown connection? If none, we probably shouldn't silently ignore the inconsistent state.

quinn/src/lib.rs
```diff
@@ -481,137 +414,91 @@ impl EndpointInner {
...
                             }
                         }
+                        Transmit(t) => self.outgoing.push_back(t),
                     }
-                    // Timeout call may have queued sends
                     keep_going = true;
```
Ralith: A comment as to why this is necessary would be helpful.

quinn/src/lib.rs
```diff
@@ -481,137 +414,91 @@ impl EndpointInner {
...
                 }
-                Ok(Async::Ready(Some(None))) => {}
-                Ok(Async::Ready(None)) | Ok(Async::NotReady) => {
-                    break;
+                Ok(Async::Ready(None)) => unreachable!(),
+                Ok(Async::NotReady) => {
+                    return Ok(keep_going);
                 }
-                Err(()) => unreachable!(),
+                Err(_) => unreachable!(),
```
Ralith: I've started to appreciate leaving little notes (`unreachable!()` takes format args) explaining why an expression is unreachable, as a reminder for if the panic actually happens.

quinn/src/lib.rs
```diff
@@ -672,83 +559,101 @@ pub type Incoming = mpsc::Receiver<NewConnection>;
 
 /// A connection initiated by a remote client.
 pub struct NewConnection {
+    /// The driver for the connection
```
Ralith: Worth having a hint that this is a future that should be spawned immediately here.

quinn/src/lib.rs
```diff
@@ -672,83 +559,101 @@ pub type Incoming = mpsc::Receiver<NewConnection>;
...
-        NewConnection {
-            connection: Connection(conn.clone()),
-            incoming: IncomingStreams(conn),
-        }
-    }
-}
-
 /// A connection initiated locally.
 pub struct NewClientConnection {
+    /// The driver for the connection
```
Ralith: Worth having a hint that this is a future that should be spawned immediately here.

quinn/src/lib.rs
```diff
@@ -672,83 +559,101 @@ pub type Incoming = mpsc::Receiver<NewConnection>;
...
+            driver: ConnectionDriver(conn.clone()),
             connection: Connection(conn.clone()),
-            incoming: IncomingStreams(conn.clone()),
+            incoming: IncomingStreams(conn),
         }
     }
 }
 
+/// A future that drives protocol logic for a connection.
+#[derive(Clone)]
```
Ralith: When does this need to be cloned?

quinn/src/lib.rs
```diff
@@ -798,93 +698,217 @@ impl Connection {
...
 struct ConnectionInner {
-    endpoint: Rc<RefCell<EndpointInner>>,
+    log: Logger,
+    inner: quinn::Connection,
+    driver: Option<Task>,
     handle: ConnectionHandle,
+    pending: Pending,
     side: Side,
+    // TODO: Replace this with something custom that avoids using oneshots to cancel
+    timers: FuturesUnordered<Timer>,
```
Ralith: The debugging experience (and performance!) will probably be much improved if this is replaced with `[Option<tokio_timer::Delay>; quinn::Timer::COUNT]` directly; no need for a dynamic set now that it's per-connection.

quinn/src/lib.rs
```diff
@@ -798,93 +698,217 @@ impl Connection {
...
+
+    fn handle_endpoint_events(&mut self) {
+        while let Some(event) = self.inner.poll_endpoint_events() {
+            self.endpoint_events
+                .unbounded_send((self.handle, EndpointEvent::Proto(event)))
+                .unwrap();
+        }
+    }
+
+    fn handle_conn_events(&mut self) -> Result<bool, ()> {
```
Ralith: What's this `bool` needed for?

quinn/src/lib.rs
```diff
@@ -245,6 +245,12 @@ impl Endpoint {
     }
 }
 
+impl Drop for Endpoint {
+    fn drop(&mut self) {
+        self.inner.borrow_mut().dropped = true;
```
Ralith: Is this value ever actually read?

quinn-proto/src/connection.rs
```diff
@@ -3237,13 +3239,15 @@ const MAX_ACK_BLOCKS: usize = 64;
...
         now: Instant,
         remote: SocketAddr,
         ecn: Option<EcnCodepoint>,
         first_decode: PartialDecode,
         remaining: Option<BytesMut>,
     },
+    NewConnection(Connection),
     NewIdentifiers(Vec<(u64, ConnectionId)>),
     Timer(Instant, Timer),
 }
```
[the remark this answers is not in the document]
djc: Yup, done.

quinn-proto/src/endpoint.rs
```diff
@@ -508,7 +428,7 @@ impl Endpoint {
         rest: Option<BytesMut>,
         crypto: &Crypto,
         header_crypto: &RingHeaderCrypto,
-    ) -> Option<ConnectionHandle> {
+    ) -> Option<(ConnectionHandle, ConnectionEvent)> {
```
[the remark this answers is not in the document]
djc: Yes, done.

quinn-proto/src/tests.rs
```diff
@@ -236,11 +252,12 @@ struct TestEndpoint {
     addr: SocketAddr,
     socket: Option<UdpSocket>,
     timers: [Option<Instant>; Timer::COUNT],
-    conn: Option<ConnectionHandle>,
     outbound: VecDeque<Transmit>,
     delayed: VecDeque<Transmit>,
     inbound: VecDeque<(Instant, Option<EcnCodepoint>, Box<[u8]>)>,
     accepted: Option<ConnectionHandle>,
+    connections: FnvHashMap<ConnectionHandle, Connection>,
+    conn_events: FnvHashMap<ConnectionHandle, VecDeque<ConnectionEvent>>,
```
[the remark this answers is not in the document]
djc: I see it more as something that helps test for correctness by requiring the use of the correct `ConnectionHandle`s. I don't think it would simplify all that much to use a single `Option<Connection>` per `TestEndpoint` or some such.

quinn-proto/src/tests.rs
```diff
@@ -288,47 +330,57 @@ impl TestEndpoint {
...
-            };
+
+        for (ch, event) in endpoint_events {
+            if let Some(event) = self.handle_event(ch, event) {
+                self.conn_events
+                    .entry(ch)
+                    .or_insert_with(|| VecDeque::new())
+                    .push_back(event);
+            }
         }
```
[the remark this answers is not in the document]
djc: We don't do anything like that for sends and receives, right? We do just one round of processing for each, and then leave the next round for the next call. Why would the internal event handling be substantially different from that model?

quinn/src/lib.rs
```diff
@@ -138,19 +138,18 @@ impl Endpoint {
...
-            } else {
-                Ok(NewClientConnection::new(Rc::new(conn)))
-            }
-        }))
+        Ok(fut
+            .map_err(|_| unreachable!())
+            .and_then(move |res| match res {
+                Ok(conn) => Ok(conn),
+                Err(e) => Err(e),
+            }))
```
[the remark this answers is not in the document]
djc: Removed this code in favor of an explicit `ConnectingFuture`.

quinn/src/lib.rs
```diff
@@ -312,14 +297,15 @@ struct EndpointInner {
...
-    timers: FuturesUnordered<Timer>,
-    buffered_incoming: VecDeque<ConnectionHandle>,
+    outgoing: VecDeque<quinn::Transmit>,
+    buffered_incoming: VecDeque<(ConnectionHandle, Connection)>,
     incoming: mpsc::Sender<NewConnection>,
     driver: Option<Task>,
     ipv6: bool,
+    connecting: FnvHashMap<ConnectionHandle, ConnectionDriver>,
+    connections: FnvHashMap<ConnectionHandle, mpsc::UnboundedSender<ConnectionEvent>>,
+    sender: mpsc::UnboundedSender<(ConnectionHandle, EndpointEvent)>,
```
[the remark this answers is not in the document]
djc: Done.

quinn/src/lib.rs
djc: Yeah, this was a bit ugly. Have a look at the `ConnectingFuture` approach.

quinn/src/lib.rs
djc: Yes, that is nicer -- done.

quinn/src/lib.rs
djc: Yes, done.

quinn/src/lib.rs
djc: Let's defer that for now.

quinn/src/lib.rs
djc: Yes, look at the new `ConnectingFuture`.

quinn/src/lib.rs
djc: Done.

quinn/src/lib.rs
djc: Yeah, will try to do better on this.

quinn/src/lib.rs
djc: Done.

quinn/src/lib.rs
djc: Doesn't, removed it.

quinn/src/lib.rs
djc: I'd like to defer that until after this commit/PR. 👍

quinn/src/lib.rs
djc: Nope, removed it.

quinn/src/lib.rs
djc: Done (merged `NewClientConnection` and `NewConnection`, since they seem identical now).

quinn-proto/src/endpoint.rs
```diff
@@ -315,9 +240,9 @@ impl Endpoint {
                     }
                     Err(e) => {
                         trace!(self.log, "unable to decode packet"; "reason" => %e);
+                        return None;
                     }
-                }
-                return None;
+                };
```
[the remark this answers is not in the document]
djc: Done.

quinn/src/lib.rs
djc: Done.

quinn/src/lib.rs
djc: Done.

quinn/src/lib.rs
djc: Nothing that I could think of, removed.

quinn/src/lib.rs
djc: I don't think it is necessary? Removed it for now.

quinn/src/lib.rs
djc: Made it noisy!

quinn-proto/src/tests.rs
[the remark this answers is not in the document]
Ralith: The main thing would be not having to thread the `ConnectionHandle` through to `client_conn_mut` and such. Not that significant, though, and we probably really should get some tests for multiple concurrent connections in one endpoint sooner or later.

quinn-proto/src/tests.rs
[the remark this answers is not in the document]
Ralith: `EndpointEvent::NeedIdentifiers` can be emitted in response to a receive and will elicit an immediate `NewIdentifiers`, which turns into an immediate send. In the old tokio code, we take care to process it after receive and before send, and in the new tokio code the `mpsc` streams ensures we'll be woken up as needed. Here there's no wakeups, so we must be careful to ensure there's no work left to do before returning from `drive`. Rather than looping as I originally suggested, it's probably sufficient (for now) to reorder things to recv -> process events -> send, since sending never has side effects other than the scheduling of timers.

quinn/src/lib.rs
Ralith: I was thinking it specifically might simplify diagnosing malfunctions blocking the merge, but I haven't looked closely at those yet.

quinn/src/lib.rs
```diff
@@ -670,41 +453,85 @@ impl Pending {
...
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        let conn = &mut *self.0.borrow_mut();
+        if conn.driver.is_none() {
+            conn.driver = Some(task::current());
+        }
+        let now = Instant::now();
+        loop {
+            let mut keep_going = false;
+            conn.handle_app_events();
+            conn.handle_conn_events().unwrap();
```
Ralith: `handle_app_events` must be called *after* `handle_conn_events` or we will fail to emit timely notifications of events that occur immediately prior to the connection becoming idle.

quinn/src/lib.rs
```diff
@@ -798,93 +618,220 @@ impl Connection {
...
+                        reason: "connection is closing".to_string(),
+                    }));
+            }
+            self.endpoint_events
+                .unbounded_send((self.handle, EndpointEvent::Proto(event)))
+                .unwrap();
+        }
+    }
+
+    fn handle_conn_events(&mut self) -> Result<(), ()> {
```
Ralith:
```suggestion
    fn process_conn_events(&mut self) -> Result<(), ()> {
```

quinn/src/lib.rs
```diff
@@ -798,93 +618,220 @@ impl Connection {
...
+impl ConnectionInner {
+    fn drive_transmit(&mut self, now: Instant) {
+        while let Some(t) = self.inner.poll_transmit(now) {
+            self.endpoint_events
+                .unbounded_send((self.handle, EndpointEvent::Transmit(t)))
+                .unwrap();
+        }
+    }
+
+    fn handle_endpoint_events(&mut self) {
```
Ralith:
```suggestion
    fn emit_endpoint_events(&mut self) {
```

quinn/src/lib.rs
```diff
@@ -798,93 +618,220 @@ impl Connection {
...
+                    return Ok(());
+                }
+                Err(_) => {
+                    return Err(());
+                }
+            }
+        }
+    }
+
+    fn handle_app_events(&mut self) {
```
Ralith:
```suggestion
    fn emit_app_events(&mut self) {
```

quinn/src/lib.rs
```diff
@@ -798,93 +618,220 @@ impl Connection {
...
+                }
+                Ok(Async::Ready(Some(ConnectionEvent::Close(_)))) => {
+                    self.closed = true;
+                }
+                Ok(Async::Ready(None)) => {
+                    return Ok(());
+                }
+                Ok(Async::NotReady) => {
+                    return Ok(());
+                }
```
Ralith: The two cases above can be combined with `|`

quinn/src/lib.rs
```diff
@@ -798,93 +618,220 @@ impl Connection {
...
+                Err(_) => {
+                    return Err(());
```
Ralith:
```suggestion
                    unreachable!("channel receivers never fail");
```

quinn-proto/src/endpoint.rs
```diff
@@ -1167,6 +928,11 @@ impl IndexMut<ConnectionHandle> for Slab<ConnectionMeta> {
     }
 }
 
+pub enum EndpointAppEvent {
```
Ralith: Not a huge fan of the name; this isn't exactly an application-facing enum. Maybe this?
```suggestion
pub enum DatagramEvent {
```
Or something else.

quinn-proto/src/tests.rs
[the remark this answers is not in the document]
Ralith: On further consideration, this has to be:
1. recv
2. process connection events (such as `Datagram`)
3. process endpoint events (such as `NeedIdentifiers`)
4. process connection events (such as `NewIdentifiers`)
5. send

(or a loop) to ensure that we don't end up with a client failing to send new identifiers until there's other activity.

quinn/src/lib.rs
```diff
@@ -330,50 +248,41 @@ impl EndpointInner {
...
-                                    .unwrap();
-                                self.inner.accept();
-                            }
-                            _ => {
-                                self.buffered_incoming.push_back(ch);
-                            }
+                    match self.inner.handle(now, addr, ecn, (&buf[0..n]).into()) {
+                        Some((handle, EndpointAppEvent::NewConnection(conn))) => {
+                            let conn = self.create_connection(handle, conn);
+                            self.buffered_incoming.push_back((handle, conn));
```
Ralith: It's not ideal that a busy application will let connections languish in the incoming queue (either in `buffered_incoming` or in the `incoming` channel itself) where they won't be driven and may time out. This causes peers to waste time on spurious `Initial` retransmits, and may cause spurious timeouts.

We could solve this by doing away with buffering entirely, which isn't ideal for performance and is impossible with our current channel implementation anyway, or by writing a custom channel that tracks which connections have not yet been yielded to the application, allowing us to automatically drive them, perhaps from a `VecDeque` owned by the endpoint driver.

quinn/src/lib.rs
```diff
@@ -1091,27 +1031,23 @@ impl AsyncWrite for BiStream {
...
+        let mut conn = self.conn.borrow_mut();
+        let ours = self.stream.initiator() == conn.side;
         let (send, recv) = match self.stream.directionality() {
             Directionality::Bi => (true, true),
             Directionality::Uni => (ours, !ours),
         };
 
-        let pending = endpoint.pending.get_mut(&self.conn.handle).unwrap();
-        if pending.closing.is_some() || pending.drained {
+        if conn.pending.closing.is_some() {
```
Ralith:
```suggestion
        if conn.pending.closing.is_some() || conn.pending.error.is_some() {
```
This ensures we don't proceed if the connection has already fully closed.

quinn/src/lib.rs
```diff
@@ -798,93 +618,220 @@ impl Connection {
...
+    }
+
+    fn handle_conn_events(&mut self) -> Result<(), ()> {
+        loop {
+            match self.conn_events.poll() {
+                Ok(Async::Ready(Some(ConnectionEvent::Proto(event)))) => {
+                    self.inner.handle_event(event);
+                }
+                Ok(Async::Ready(Some(ConnectionEvent::Close(_)))) => {
+                    self.closed = true;
```
Ralith: We need to call `Pending::fail` here to ensure all downstream futures are notified.

quinn/src/lib.rs
```diff
@@ -798,93 +618,220 @@ impl Connection {
...
+    driver: Option<Task>,
     handle: ConnectionHandle,
+    pending: Pending,
     side: Side,
+    // TODO: Replace this with something custom that avoids using oneshots to cancel
+    timers: FuturesUnordered<Timer>,
+    conn_events: mpsc::UnboundedReceiver<ConnectionEvent>,
+    endpoint_events: mpsc::UnboundedSender<(ConnectionHandle, EndpointEvent)>,
+    connected: bool,
+    closed: bool,
```
Ralith: It's unclear if this is intended to express "closed gracefully" (in which case unbounded futures/streams could end without an error, but must also check pending.error to determine if the connection failed) or "closed for any reason".

quinn/src/lib.rs
```diff
@@ -670,41 +453,85 @@ impl Pending {
...
-            incoming: IncomingStreams(conn.clone()),
+/// A future that drives protocol logic for a connection.
+#[must_use]
+pub struct ConnectionDriver(Rc<RefCell<ConnectionInner>>);
+
+impl Future for ConnectionDriver {
+    type Item = ();
+    type Error = io::Error;
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        let conn = &mut *self.0.borrow_mut();
```
Ralith: Before proceeding, we should check `conn.pending.error` and return the error if one is found.

quinn/src/lib.rs
```diff
@@ -670,41 +453,85 @@ impl Pending {
...
-        Self {
-            connection: Connection(conn.clone()),
-            incoming: IncomingStreams(conn.clone()),
+/// A future that drives protocol logic for a connection.
+#[must_use]
+pub struct ConnectionDriver(Rc<RefCell<ConnectionInner>>);
+
+impl Future for ConnectionDriver {
+    type Item = ();
+    type Error = io::Error;
```
Ralith:
```suggestion
    type Error = ConnectionError;
```

quinn/src/lib.rs
```diff
@@ -670,41 +453,85 @@ impl Pending {
...
-    /// The connection itself.
-    pub connection: Connection,
-    /// The stream of QUIC streams initiated by the client.
-    pub incoming: IncomingStreams,
+/// Connecting future
+pub struct ConnectingFuture(Option<ConnectionDriver>);
+
+impl Future for ConnectingFuture {
+    type Item = NewConnection;
+    type Error = io::Error;
```
Ralith:
```suggestion
    type Error = ConnectionError;
```

quinn/src/lib.rs
```diff
@@ -670,41 +453,85 @@ impl Pending {
...
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        let connected = match &mut self.0 {
+            Some(driver) => {
+                match driver.poll() {
```
Ralith:
```suggestion
                match driver.poll()? {
```

Ralith:
If we want locally closed connections not to produce errors from their associated futures/streams, we should probably rename `fail` to `terminate` or something and move the error assignment out of it.

quinn/src/lib.rs
djc: With `drive_incoming()` running directly after `drive_recv()` and the latter being constrained by `IO_LOOP_BOUND`, can we agree that this can wait until after this PR?

quinn/src/lib.rs
djc: Done.

quinn/src/lib.rs
djc: Done.

quinn/src/lib.rs
djc: Done.

quinn/src/lib.rs
djc: Fixed.

quinn/src/lib.rs
Ralith: The problem arises if the application isn't reading from `incoming` at all, so a connection can remain there indefinitely. That said, it's probably okay to defer since it's a subtle inefficiency that only affects already-overloaded (or severely buggy) systems.

quinn/src/lib.rs
```diff
@@ -176,87 +175,10 @@ impl Endpoint {
...
-            endpoint: self.inner.clone(),
-            handle,
-            side: Side::Client,
-        };
-        Ok((recv, conn))
-    }
 }
 
 /// A future that drives IO on an endpoint.
+#[must_use]
```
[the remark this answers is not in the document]
djc: Added more documentation here.

quinn/src/lib.rs
```diff
@@ -670,41 +453,85 @@ impl Pending {
...
+    }
 }
 
-impl NewClientConnection {
-    fn new(conn: Rc<ConnectionInner>) -> Self {
-        Self {
-            connection: Connection(conn.clone()),
-            incoming: IncomingStreams(conn.clone()),
+/// A future that drives protocol logic for a connection.
+#[must_use]
```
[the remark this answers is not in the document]
djc: Added some docs here.

quinn/src/lib.rs
djc: Right. I think in what I have now it's mostly used for "closed locally". I'll admit to having gotten a bit lost in the mechanics of it all, and I think we should come up with good invariants on what we want the externally visible behavior of the API want to be, and then we can figure out how to implement that.

But at the same time, this now passes all our tests and represents something like a minimal change to do so, so I'm not sure how much further change we should require to be part of this PR.

quinn-proto/src/tests.rs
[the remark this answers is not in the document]
djc: So I've changed the code to push any `ConnectionEvent`s resulting from `Endpoint::handle_event()` directly into the concerned `Connection`, does that assuage your concerns on this point?

The problem with exactly implementing your proposed change is that the overlapping ownership of the `Endpoint` (for `Endpoint::handle_event()`) and the `Connection`s that are part of it, so the code ends up a bit ugly.

Do you have any concrete further worries here before we can get this PR merged?

djc:
Proposed follow-up work gathered from comments so far:

* Evaluate/verify ordering in `TestEndpoint::drive()`
* Evaluate `Stream<Item=(ConnectionHandle, ConnectionEvent)>` approach for `Driver`
* Replace `FuturesUnordered` contraption with simple array
* How to handle buffering for incoming connections (what to do if app doesn't poll?)
* Clean up closing sequence to properly reflect clean/error/remote closes in public API

djc:
So taking those points as input for the next round of work, I think it'd be reasonable to merge this now so we can continue other work in parallel without rebase hazards and debug the new approach rather than the soon-obsolete approach. What do you think?

quinn/src/lib.rs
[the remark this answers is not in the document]
Ralith: To clarify, I meant explanatory text associated with the `must_use` attribute, as documented at the bottom of https://doc.rust-lang.org/reference/attributes.html#must_use. Not a blocker.

quinn-proto/src/tests.rs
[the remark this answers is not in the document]
Ralith: We'll need to drain `poll_transmit` again to finish, which will set new timers (though in practice we should have just transmitted anyway, so they'll be getting set to the same value again). If the tests are passing I guess we can defer, but a divergence (i.e. letting the new connection IDs packet get stuck in the outbound queue until the something calls `drive` again for an unrelated reason) from realistic behavior is a landmine.

Maybe a more convenient way to structure this would be to iterate over the connections twice? Once to handle incoming packets and timeouts and generate endpoint events, then again after the endpoint events are processed to process any resulting connection events, send packets, and set new timers.

Issue #258, opened 2019-03-26
Optimize stream assembly


Ralith:
Pulling data through `stream::Assembler` is the largest cost on the ordered-read path right now. This is a small, self-contained helper type with a fairly simple contract of "consume chunks of data at arbitrary positions and yield an in-order byte stream," and the current implementation is deliberately naive. There's lots of prior art to reference in TCP implementations. To start with, it's certainly a waste of effort to be storing everything in a single contiguous buffer.

Ralith:
As of #258 this is no longer known to be a cost center.

PR #260, opened 2019-03-26
Optimize stream assembler (see #258)

djc:
I accidentally an attempt to optimize the assembler. Would you mind comparing the benchmarks, since I don't have a nightly environment setup?

### DIFF
--- a/quinn-proto/src/stream.rs
+++ b/quinn-proto/src/stream.rs
@@ -1,5 +1,5 @@
-use std::collections::hash_map;
-use std::collections::VecDeque;
+use std::cmp::Ordering;
+use std::collections::{hash_map, BinaryHeap};
 
 use bytes::Bytes;
 use err_derive::Error;
@@ -220,7 +220,6 @@ pub enum WriteError {
 pub struct Recv {
     pub state: RecvState,
     pub recvd: RangeSet,
-    pub buffered: VecDeque<(Bytes, u64)>,
     /// Whether any unordered reads have been performed, making this stream unusable for ordered
     /// reads
     pub unordered: bool,
@@ -235,7 +234,6 @@ impl Recv {
         Self {
             state: RecvState::Recv { size: None },
             recvd: RangeSet::new(),
-            buffered: VecDeque::new(),
             unordered: false,
             assembler: Assembler::new(),
             bytes_read: 0,
@@ -248,14 +246,10 @@ impl Recv {
             "cannot perform ordered reads following unordered reads on a stream"
         );
 
-        for (data, offset) in self.buffered.drain(..) {
-            self.assembler.insert(offset, &data);
-        }
-
-        if !self.assembler.blocked() {
-            let n = self.assembler.read(buf);
-            self.bytes_read += n as u64;
-            Ok(n)
+        let read = self.assembler.read(buf);
+        if read > 0 {
+            self.bytes_read += read as u64;
+            Ok(read)
         } else {
             Err(self.read_blocked())
         }
@@ -263,12 +257,11 @@ impl Recv {
 
     pub fn read_unordered(&mut self) -> Result<(Bytes, u64), ReadError> {
         self.unordered = true;
-        // TODO: Drain rs.assembler to handle ordered-then-unordered reads reliably
 
         // Return data we already have buffered, regardless of state
-        if let Some(x) = self.buffered.pop_front() {
-            self.bytes_read += x.0.len() as u64;
-            Ok(x)
+        if let Some((offset, bytes)) = self.assembler.pop() {
+            self.bytes_read += bytes.len() as u64;
+            Ok((bytes, offset))
         } else {
             Err(self.read_blocked())
         }
@@ -309,12 +302,12 @@ impl Recv {
         self.state == self::RecvState::Closed
     }
 
-    pub fn buffer(&mut self, data: Bytes, offset: u64) {
+    pub fn buffer(&mut self, bytes: Bytes, offset: u64) {
         // TODO: Dedup
-        if data.is_empty() {
+        if bytes.is_empty() {
             return;
         }
-        self.buffered.push_back((data, offset));
+        self.assembler.insert(offset, bytes);
     }
 
     /// Offset after the largest byte received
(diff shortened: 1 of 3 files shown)


djc:
This failure is also weird to me (when all the unit tests for the `Assembler` pass). It appears like something may be double reading from the stream, or intertwining ordered with unordered reads in a away we don't guard against? It also seems like my backtrace symbols are off (line numbers not matching method names and some method names should never appear in there, like `ring::rsa::padding::pss_digest`), so at this point I'm just confused.

quinn-proto/src/stream.rs
```diff
@@ -400,97 +399,98 @@ pub enum RecvState {
...
+        } else if (chunk.offset + chunk.bytes.len() as u64) < self.offset {
+            return true;
+        }
+
+        // Determine `start` and `len` of slice to read from chunk
+        let start = if chunk.offset < self.offset {
+            (self.offset - chunk.offset) as usize
+        } else {
+            0
+        };
```
Ralith: We already returned if `chunk.offset > self.offset`, so this branch isn't needed.

quinn-proto/src/stream.rs
```diff
@@ -400,97 +399,98 @@ pub enum RecvState {
...
+        self.offset += len as u64;
+
+        return if start + len == chunk.bytes.len() {
+            // This chunk has been fully consumed and can be discarded
+            true
+        } else {
+            // Mutate the chunk; per the documentation, this will update the
+            // heap's ordering accordingly if necessary. Don't pop the chunk.
+            chunk.offset = chunk.offset + start as u64 + len as u64;
+            chunk.bytes.split_to(start + len);
```
Ralith: This should be `advance`.

quinn-proto/src/stream.rs
```diff
@@ -400,97 +399,98 @@ pub enum RecvState {
...
+        *read += len;
+        self.offset += len as u64;
+
+        return if start + len == chunk.bytes.len() {
+            // This chunk has been fully consumed and can be discarded
+            true
+        } else {
+            // Mutate the chunk; per the documentation, this will update the
+            // heap's ordering accordingly if necessary. Don't pop the chunk.
+            chunk.offset = chunk.offset + start as u64 + len as u64;
```
Ralith: Where does the heap get reordered?

Ralith:
Looks like the -proto test suite has a number of failures with this too; CI just doesn't get that far.

Ralith:
One correctness issue is that you're bypassing the `buffered` deque entirely, which renders unordered reads (that draw from it) inoperable. Pointing them at `assembler.data` again doesn't quite seem to entirely get things going again, though...

Ralith:
There, that got it working!

Ralith:
Performance results for 128KiB messages:

- `read_unordered`: `[109.77 MiB/s 110.48 MiB/s 111.10 MiB/s]`
- `read` (master): `[63.342 MiB/s 63.662 MiB/s 63.962 MiB/s]`
- `read` (this branch): `[112.79 MiB/s 113.85 MiB/s 114.80 MiB/s]`

Hooray!

Ralith:
Incidentally, I can't seem to show any performance improvement for keeping the `VecDeque` around versus just shoving everything into the BinaryHeap directly, so we should probably just simplify that down.

quinn-proto/src/stream.rs
djc: Presumably on dropping the `PeekMut`. It's documented for `peek_mut()`, I've tried to clarify the comment.

quinn-proto/src/stream.rs
```diff
@@ -500,8 +500,37 @@ impl Assembler {
...
+
+impl Ord for Chunk {
+    // Invert ordering based on offset (max-heap, min offset first),
+    // prioritize longer chunks at the same offset.
+    fn cmp(&self, other: &Chunk) -> Ordering {
+        match self.offset.cmp(&other.offset) {
+            Ordering::Less => Ordering::Greater,
+            Ordering::Greater => Ordering::Less,
+            Ordering::Equal => self.bytes.len().cmp(&other.bytes.len()),
+        }
```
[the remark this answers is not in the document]
djc: 👍

quinn-proto/src/stream.rs
djc: Oh, nice.

quinn-proto/src/stream.rs
djc: Yes, this is still needed. This is the inverse condition, where the next chunk starts before the current assembler offset (but the chunk extends beyond the current offset). For example, the `assemble_overlapping` test case.

djc:
Also removed the interim `VecDeque`. The obvious replacement of `Bytes::from_static()` with `.into()` didn't seem to work out, so after trying and failing some other options I've left it for now.

quinn-proto/src/stream.rs
Ralith: I understand, but the branch isn't necessary. The test case still passes with
```rust
let start = (self.offset - chunk.offset) as usize;
```
because at this point in control flow it is guaranteed that chunk.offset <= self.offset, so there will never be underflow.

djc: Ah, I misunderstood your comment. Fixed!

quinn-proto/src/tests.rs
```diff
@@ -906,11 +906,11 @@ fn key_update_reordered() {
     assert_matches!(pair.server.accept_stream(server_ch), Some(stream) if stream == s);
     assert_matches!(
         pair.server.read_unordered(server_ch, s),
-        Ok((ref data, 1)) if data == MSG2
+        Ok((ref data, 0)) if data == MSG1
     );
     assert_matches!(
         pair.server.read_unordered(server_ch, s),
-        Ok((ref data, 0)) if data == MSG1
+        Ok((ref data, 1)) if data == MSG2
```
Ralith: Since we've lost the (unimportant) ability to process data in exactly the order it's received, this test might want to be modified to use ordered reads, and strengthened by checking that no client packets were lost.

PR #264, opened 2019-03-28
Assorted fixes

Ralith:


djc:
From the actual fixes here, did you identify things that we could add tests for to make sure we don't regress them?

Ralith:
- Verifying cleanup after premature `ConnectionInner` drop would require exposing e.g. `self.connections.len()` from the high-level `Endpoint`, and is then straightforward (just create and immediately drop a connection!)
- Close-on-drop can be tested easily by chaining onto the connection driver
- Timers not getting scheduled is tricky, because reproducing it relies on arranging for a newly scheduled timeout to be the next wakeup when no other wakeup is expected
- Handling of stateless reset of a locally closed connection should admit a unit test
- Inducing a connection error and checking for graceful shutdown should be easy enough if we expose `is_drained` or similar at the high level, then check it after the connection driver completes

PR #265, opened 2019-03-28
Fix builds on musl environment

nbaksalyar:
msghdr struct in musl libc has padding fields which are private (introduced with this PR: https://github.com/rust-lang/libc/pull/1164). This commit fixes the struct initialisation code, replacing it with manual fields assingment.

Tests pass on all platforms now.


Ralith:
Thanks!

PR #266, opened 2019-03-28
Streams abstraction

djc:
Follow through on my plan to build `Streams` into more of an abstraction.

### DIFF
--- a/quinn-proto/src/stream.rs
+++ b/quinn-proto/src/stream.rs
@@ -30,6 +30,30 @@ pub struct Streams {
 }
 
 impl Streams {
+    pub fn open(&mut self, side: Side, direction: Directionality) -> Option<StreamId> {
+        let (id, stream) = match direction {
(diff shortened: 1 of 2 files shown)


quinn-proto/src/stream.rs
```diff
@@ -23,11 +25,114 @@ pub struct Streams {
...
+        } else if self.next_remote_bi > self.next_reported_remote_bi {
+            let x = self.next_reported_remote_bi;
+            self.next_reported_remote_bi = x + 1;
+            Some(StreamId::new(!side, Directionality::Bi, x))
+        } else {
+            None
+        }
+    }
+
+    pub fn reset(&mut self, side: Side) {
```
Ralith:
```suggestion
    pub fn zero_rtt_rejected(&mut self, side: Side) {
```
(or something else) to avoid confusion with the unrelated `RESET_STREAM` operation (and adjust the comment below to match)

PR #270, opened 2019-03-30
Further redistribution of code into modules

djc:
More suggestions for evening out module sizes and strengthening abstractions.


quinn-proto/src/connection.rs
```diff
@@ -2743,15 +2741,11 @@ impl Connection {
         self.state.is_drained()
     }
 
-    pub fn accepted_0rtt(&self) -> bool {
```
Ralith: This is a necessary part of the public API for 0-RTT support: higher-level code needs it to determine whether 0-RTT streams still exist following connection establishment.

quinn-proto/src/connection.rs
```diff
@@ -2731,10 +2733,6 @@ impl Connection {
         });
     }
 
-    pub fn is_handshaking(&self) -> bool {
```
Ralith: Similarly, this is convenient to determine whether 0-RTT acceptance is defined yet. You could reconstruct that information from events, but it's a lot easier to be able to just look it up.

quinn-proto/src/connection.rs
djc: Brought it back!

quinn-proto/src/connection.rs
djc: Brought it back.

PR #271, opened 2019-03-30
Refactor close behavior

Ralith:
Makes the effects of invoking `close` on a connection saner, and fixes a few bugs found along the way.


quinn/src/tests.rs
```diff
@@ -1,13 +1,15 @@
-use super::{read_to_end, ClientConfigBuilder, Endpoint, NewStream, ServerConfigBuilder};
-use futures::{Future, Stream};
-use slog::{Drain, Logger, KV};
 use std::{
```
djc: While you're at it, maybe normalize this to the single-line uses we've got elsewhere, too?

quinn/src/connection.rs
```diff
@@ -111,17 +107,19 @@ impl Future for ConnectionDriver {
...
             }
         }
 
         if !conn.inner.is_drained() {
-            Ok(Async::NotReady)
-        } else if let Some(e) = &conn.pending.error {
-            Err(e.clone())
-        } else {
-            Ok(Async::Ready(()))
+            conn.driver = Some(task::current());
```
djc: Is it necessary to do this even if `driver` is `Some` already? (I don't know how expensive it is.)

quinn/src/connection.rs
```diff
@@ -111,17 +107,19 @@ impl Future for ConnectionDriver {
...
+            return Ok(Async::NotReady);
+        }
+        match conn.pending.error {
```
djc: Shouldn't this go before the `if !conn.inner.is_drained()`, similar to your changes in `IncomingStreams::poll()`?

quinn/src/connection.rs
```diff
@@ -236,9 +220,18 @@ impl FuturesStream for IncomingStreams {
     type Error = ConnectionError;
     fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
         let mut conn = self.0.lock().unwrap();
-        if conn.closed {
+        if conn.inner.is_drained() {
             return Ok(Async::Ready(None));
         }
+        match conn.pending.error {
```
djc: Alternatively, this `poll()` could be a single `if {} else if {} else {}` chain with no explicit `return`s and `if let`s per differently-handled error case, which I think might be slightly more concise and easier to follow?

quinn/src/connection.rs
```diff
@@ -522,7 +494,6 @@ impl Pending {
             incoming_streams_reader: None,
             finishing: FnvHashMap::default(),
             error: None,
-            closing: None,
         }
     }
```
djc: Maybe still rename `fail()` to `terminate()` or something here?

djc:
Looks very nice! Some nits, but this seems like it clarifies the semantics. I wonder if there should be some comment and/or documentation somewhere that outlines the high-level ideas for this (even if only for our own reference)?

quinn/src/connection.rs
```diff
@@ -186,24 +184,10 @@ impl Connection {
...
-            conn.pending.closing = Some(send);
-            conn.inner.close(Instant::now(), error_code, reason.into());
-        }
-        let handle = self.clone();
-        recv.then(move |_| {
-            // Ensure the connection isn't dropped until it's fully drained.
-            let _ = handle;
-            Ok(())
-        })
+    pub fn close(&self, error_code: u16, reason: &[u8]) {
```
djc: Maybe still take `self` here, as discussed in #253?

quinn/src/connection.rs
Ralith: This should cost at most an `Arc::clone`. I'm not certain off the top of my head whether futures are allowed to move between tasks; if they are, this is absolutely necessary.

quinn/src/connection.rs
Ralith: It specifically must not, because `error` is usually set before the connection becomes drained. Everything but the driver should complete immediately at that point, but the driver *must* remain until draining is complete so that it can gracefully absorb incoming packets.

Future work could be to allow move the entire draining state into `Endpoint`, which would allow drivers to be dropped instantly as well, but that's not the current shape of things.

quinn/src/connection.rs
djc: Fair enough, it does seem like the safer option.

quinn/src/connection.rs
djc: Maybe add a comment then to that make that more obvious? Also, why would we want to move the draining state into the `Endpoint`?

quinn/src/connection.rs
Ralith: Good call, that actually made it so much easier to follow that I found a minor bug!

quinn/src/connection.rs
```diff
@@ -90,8 +90,10 @@ impl NewConnection {
 /// `Connection` API object to the `Endpoint` task and the related stream-related interfaces.
 /// It also keeps track of outstanding timeouts for the `Connection`.
 ///
-/// If the connection encounters an error condition, this future will yield an error. It
-/// will terminate (yielding `Ready(())`) if the connection was closed without error.
+/// If the connection encounters an error condition, this future will yield an error. It will
+/// terminate (yielding `Ok(())`) if the connection was closed without error. Unlike other
```
djc: Not sure `yielding Ok(())` makes sense here, since it really yields `Ok(Async::Ready(()))`?

Ralith: `Future::poll` returns `Ok(Async::Ready(()))`, but if you do `driver.then(|x| ...)` the `x` you get is `Ok(())`. The latter is what the user is typically concerned with, and is what `yield` most precisely refers to.

djc: I don't think it's so clear cut (that is, I clearly had/have a different interpretation). Maybe replace `yielding` with `returning` and then use `Ok(Async::Ready(()))` to be fully explicit? I get where you're coming from (and it makes more sense in an async/await world), but it would bother me that the `Ok(())` does not occur in the actual type signature of the old-style `Future::poll()` return type.

Ralith: Note that we're documenting the future itself here, not the `poll` implementation. It'd a weird wart (corrected in futures 0.3, regardless of async/await) that `poll`'s return type has `Result` inlined into it, but it is conventional to refer to the implied `Result` as "the result of the future," as done in e.g. the documentation of [`Future::then`](https://docs.rs/futures/0.1.25/futures/future/trait.Future.html#method.then). Would you be happier with "(resulting in `Ok(())`)"?

djc: Fine to leave it as is, then.

PR #273, opened 2019-03-30
Keep buffered connections alive and shut down endpoint when dead

Ralith:


quinn/src/endpoint.rs
```diff
@@ -228,9 +242,20 @@ impl EndpointInner {
     }
 
     fn drive_incoming(&mut self) {
-        for conn in &mut self.incoming {
-            // It's safe to poll an already dead connection
-            let _ = conn.poll();
+        if let Some(ref mut incoming) = self.incoming {
```
djc: Maybe hoist the `if` out of this method so that the `else` part can go in a different method? As is this seems pretty confusing and conceptually wrong (`drive_incoming` has nothing to do with the `draining` loop).

quinn/src/endpoint.rs
```diff
@@ -191,10 +196,19 @@ impl EndpointInner {
...
-                            self.incoming.push_back(ConnectionDriver(conn));
-                            if let Some(task) = self.blocked_incoming.take() {
-                                task.notify();
+                            let conn = ConnectionDriver(self.create_connection(handle, conn));
+                            if let Some(ref mut incoming) = self.incoming {
+                                incoming.push_back(conn);
+                                if let Some(task) = self.blocked_incoming.take() {
+                                    task.notify();
+                                }
+                            } else {
```
djc: Maybe add a method on `ConnectionDriver` to do the dirty here, so that `ConnectionInner::inner` doesn't have to be `pub(crate)`?

quinn-proto/src/endpoint.rs
```diff
@@ -45,6 +45,8 @@ pub struct Endpoint {
     config: Arc<EndpointConfig>,
     server_config: Option<Arc<ServerConfig>>,
     incoming_handshakes: usize,
+    /// Whether incoming connections should be unconditionally rejected by a server
+    busy: bool,
```
djc: Don't like the name much, maybe `reject_new` or `reject_new_connections` or `draining`?

quinn-proto/src/endpoint.rs
```diff
@@ -585,6 +588,11 @@ impl Endpoint {
     pub fn accept(&mut self) {
         self.incoming_handshakes -= 1;
     }
+
+    /// Whether to unconditionally reject future incoming connections
+    pub fn busy(&mut self, state: bool) {
```
djc: Since we don't need to reenable here by definition, no need to take an argument? Also let's prefix the name with `set_` unless it can mirror the name of the field and that name starts with a verb?

quinn/src/endpoint.rs
Ralith: `draining` is, strictly speaking, still a set of incoming connections, just ones that we've determined the application will no longer receive. Maybe it just needs a better name? It's nice to have the if statement directly enclosing the loop because then we don't need any unwraps or duplicate checks.

djc: Hmm, okay. I can't think of a better name, maybe just add a short explanatory comment to explain what's going on here, why it's doing what it's doing? (Also, too bad `VecDeque::retain()` doesn't give you `&mut T` rather than `&T` -- maybe I'll file a bug about that since I don't see any downside.)

quinn-proto/src/endpoint.rs
Ralith: The high-level API will never reenable, but is that sufficient reason not to expose the capability when it's so trivial to support?

djc: Yes! YAGNI.

quinn/src/endpoint.rs
Ralith: Renamed to `orphaned_incoming`, which strikes me as a suitable improvement.

See [discussion on `Vec::retain`](https://github.com/rust-lang/rust/issues/25477) for why we don't have a `&mut` version; apparently [`drain_filter`](https://github.com/rust-lang/rust/issues/43244) will subsume that wart eventually.

quinn-proto/src/endpoint.rs
Ralith: I mean, it's public API, but sure.

quinn/src/endpoint.rs
```diff
@@ -137,7 +137,13 @@ impl Future for Driver {
                 break;
             }
         }
-        Ok(Async::NotReady)
+        Ok(
+            if endpoint.unreferenced && endpoint.connections.is_empty() {
+                Async::Ready(())
+            } else {
+                Async::NotReady
+            },
```
djc: The comma here seems unidiomatic?

quinn/src/endpoint.rs
```diff
@@ -358,3 +351,54 @@ impl Drop for Incoming {
         }
     }
 }
+
+pub(crate) struct EndpointRef(Arc<Mutex<EndpointInner>>);
+
+impl EndpointRef {
+    pub(crate) fn new(log: Logger, socket: UdpSocket, inner: quinn::Endpoint, ipv6: bool) -> Self {
+        let (sender, events) = mpsc::unbounded();
+        let x = Self(Arc::new(Mutex::new(EndpointInner {
```
djc: Why do we need the intermediate variable here?

quinn/src/endpoint.rs
Ralith: rustfmt insists, unfortunately. Might be worth opening an issue there?

quinn/src/endpoint.rs
Ralith: Debugging leftovers, good catch.

quinn/src/endpoint.rs
djc: Ugh.

quinn/src/endpoint.rs
Ralith: I just realized that this is all algebraically equivalent to having a single container and a boolean flag on the side, which is simpler. Refactored into that form.

quinn/src/endpoint.rs
```diff
@@ -156,7 +156,9 @@ pub(crate) struct EndpointInner {
     inner: quinn::Endpoint,
     outgoing: VecDeque<quinn::Transmit>,
     incoming: VecDeque<ConnectionDriver>,
-    blocked_incoming: Option<Task>,
+    incoming_reader: Option<Task>,
```
djc: Ohh, nice simplification you got there!

quinn/src/endpoint.rs
```diff
@@ -232,12 +227,13 @@ impl EndpointInner {
...
-                self.inner.accept();
-            } else {
-                break;
+        for i in (0..self.incoming.len()).rev() {
+            match self.incoming[i].poll() {
+                Ok(Async::Ready(())) | Err(_) if !self.incoming_live => {
+                    let _ = self.incoming.swap_remove_back(i);
+                }
+                // It's safe to poll an already dead connection
+                _ => {}
```
Ralith: There's a judgement call here on whether we should hide connections that die before the application sees them or not. Any thoughts?

djc: I guess I'd want to know though maybe if we have fine-grained information about the cause of death we could pass through based on that?

Ralith: The `ConnectionError` we have is quite detailed, but what cases would merit discarding them?

I think I favor the current behavior, which always exposes the fate of any connection that had a valid `Initial`, if nothing else because it would make it easier to notice if someone is doing something weird to your server.

djc: Sounds good to me.

PR #276, opened 2019-04-01
Api

djc:


quinn/src/connection.rs
```diff
@@ -54,31 +54,22 @@ impl Future for ConnectingFuture {
...
-
-impl NewConnection {
-    pub(crate) fn new(conn: ConnectionRef) -> Self {
-        Self {
-            driver: ConnectionDriver(conn.clone()),
-            connection: Connection(conn.clone()),
-            incoming: IncomingStreams(conn),
-        }
-    }
+pub type NewConnection = (ConnectionDriver, Connection, IncomingStreams);
```
Ralith: Would the generated documentation be clearer without this type alias?

djc: Yup. Fixed.

PR #280, opened 2019-04-01
Further clarify implicit close semantics (see #250)

djc:


quinn/src/connection.rs
```diff
@@ -118,8 +118,10 @@ impl Future for ConnectionDriver {
 
 /// A QUIC connection.
 ///
-/// If a `Connection` is dropped without being explicitly closed, it will be automatically closed
-/// with an `error_code` of 0 and an empty `reason`.
+/// If a `Connection` and other references to the underlying state (including `IncomingStreams`
```
```suggestion
/// If all references to a connection (including every clone of the `Connection` handle, `IncomingStreams`,
```

quinn/src/connection.rs
```diff
@@ -118,8 +118,10 @@ impl Future for ConnectionDriver {
...
+/// and the various stream types) other than the `ConnectionDriver` have been dropped, the
+/// the connection will be automatically closed with an `error_code` of 0 and an empty
+/// `reason`. You can also trigger this by calling `close()` on the `Connection` explicitly.
```
```suggestion
/// `reason`. The connection can also be closed explicitly with `Connection::close()`.
```

djc: Hmm, I was thinking about that and think that passive voice is often best avoided. I'm also not crazy about `you`, but I do think it can make documentation easier to digest.

djc: (But the other parts we could change, might be better.)

Ralith: I was mostly trying to get rid of "trigger this," which struck me as a bit unclear.

djc:
Combined our sentences.

PR #284, opened 2019-04-02
More tests

povilasb:
Adds an integration test that checks how quinn deals with many connections via a single endpoint.
This test is slow and thus by default is run when "slow_tests" feature is specified. 
Note, that travis run tests with "--all-features".

### DIFF
--- a/quinn/tests/many_connections.rs
+++ b/quinn/tests/many_connections.rs
@@ -36,6 +36,7 @@ impl TestContext {
 }
 
 #[test]
+#[ignore]
 fn connect_n_nodes_to_1_and_send_1mb_data() {
     let mut runtime = unwrap!(Runtime::new());
 
--- a/quinn/tests/many_connections.rs
+++ b/quinn/tests/many_connections.rs
@@ -1,0 +1,195 @@
+use crc::crc32;
+use futures::unsync::oneshot;
+use futures::{Future, Stream};
+use rand::{self, RngCore};
+use std::cell::RefCell;
+use std::rc::Rc;
+use std::sync::Arc;
+use tokio::runtime::current_thread::{self, Runtime};
+use unwrap::unwrap;
+
+struct TestContext {
+    /// Keep track of how many messages we've already received so that we'd know when test is
+    /// finished.
+    received_messages: usize,
+    expected_messages: usize,
+    done_tx: Option<oneshot::Sender<()>>,
+
+    /// Hold connections so that streams wouldn't be closed prematurely
+    connections: Vec<quinn::Connection>,
+}
+
+impl TestContext {
+    /// Constructs test context and returns test completion receiver as well..
+    fn shared() -> (Rc<RefCell<Self>>, oneshot::Receiver<()>) {
+        let (done_tx, done_rx) = oneshot::channel();
+        (
+            Rc::new(RefCell::new(Self {
+                received_messages: 0,
+                expected_messages: 0,
+                connections: Default::default(),
+                done_tx: Some(done_tx),
+            })),
+            done_rx,
+        )
+    }
+}
+
+#[test]
+fn connect_n_nodes_to_1_and_send_1mb_data() {
+    let mut runtime = unwrap!(Runtime::new());
+
+    let (cfg, listener_cert) = configure_listener();
+    let mut ep_builder = quinn::Endpoint::new();
+    ep_builder.listen(cfg);
+    let (driver, endpoint, incoming_conns) = unwrap!(ep_builder.bind(&("127.0.0.1", 0)));
+    runtime.spawn(driver.map_err(|e| panic!("Listener IO error: {}", e)));
+    let listener_addr = unwrap!(endpoint.local_addr());
+
+    let (ctx, done_rx) = TestContext::shared();
+    ctx.borrow_mut().expected_messages = 50;
+
+    let ctx2 = ctx.clone();
+    let read_incoming_data = incoming_conns
+        .map_err(|()| panic!("Listener failed"))
+        .for_each(move |(conn_driver, _conn, incoming)| {
+            current_thread::spawn(conn_driver.map_err(|_| ()));
+
+            let ctx = ctx2.clone();
+            let task = incoming
+                .map_err(move |e| panic!("Incoming streams failed: {}", e))
+                .for_each(move |stream| {
+                    read_from_peer(stream, ctx.clone());
+                    Ok(())
+                })
+                .then(move |_| Ok(()));
+            current_thread::spawn(task);
+
+            Ok(())
+        });
+    runtime.spawn(read_incoming_data);
+
+    let client_cfg = configure_connector(&listener_cert);
+
+    for _ in 0..ctx.borrow_mut().expected_messages {
+        let data = random_data_with_hash(1024 * 1024);
+        let ctx = ctx.clone();
+        let task = unwrap!(endpoint.connect_with(&client_cfg, &listener_addr, "Test"))
+            .map_err(|e| panic!("Connection failed: {}", e))
+            .and_then(move |(conn_driver, conn, _)| {
+                current_thread::spawn(conn_driver.map_err(|_| ()));
+                write_to_peer(&conn, data);
+                ctx.borrow_mut().connections.push(conn);
+                Ok(())
+            });
+        runtime.spawn(task);
+    }
+
+    let _ = unwrap!(runtime.block_on(done_rx));
+}
+
+fn read_from_peer(stream: quinn::NewStream, ctx: Rc<RefCell<TestContext>>) {
+    let stream = match stream {
+        quinn::NewStream::Bi(_bi) => panic!("Unexpected bidirectional stream here"),
+        quinn::NewStream::Uni(uni) => uni,
+    };
+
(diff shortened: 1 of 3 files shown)


quinn/tests/many_connections.rs
```diff
@@ -0,0 +1,197 @@
+#![cfg(feature = "slow_tests")]
```
Ralith: Would [the `#[ignore]` attribute](https://doc.rust-lang.org/book/ch11-02-running-tests.html#ignoring-some-tests-unless-specifically-requested) be a better way to accomplish this? Along with adding `--ignored` to the CI script.

djc: I think features are more appropriate for this, why would you prefer ignore?

Ralith: The documentation example is for this exact use-case, and it ensures that the test is still typechecked and built as part of the normal workflow.

djc: What documentation example do you mean here? I don't think we have any other tests that really exercise this specific use case (including large payloads).

Plus, I would value making the added dependencies optional conditional on the feature flag.

povilasb: Indeed docs do say that:
```
 Rather than listing as arguments all tests you do want to run, you can instead annotate the time-consuming tests using the ignore attribute to exclude them

When you’re at a point where it makes sense to check the results of the ignored tests and you have time to wait for the results, you can run cargo test -- --ignored instead.
```

The only problem I see with `ignored` is that I will have to decorate helper functions as unused.

Ralith: > What documentation example do you mean here?

The example(s) in the documentation for the ignore attribute: https://doc.rust-lang.org/book/ch11-02-running-tests.html#ignoring-some-tests-unless-specifically-requested

> The only problem I see with ignored is that I will have to decorate helper functions as unused.

`#![allow(unused)]` at the top of the module, perhaps?

Ralith: Wait, why would you have to decorate anything as unused? The function is still compiled and referenced, just not invoked at runtime without a flag; that's still a static use.

povilasb: Good point, updated the PR ;)

povilasb:
I think so. I rebased the PR and updated the test so that when message  is transferred via the connection, server closes this connection, then client side is expected to notice connection close and in such case client connection driver is expected to complete as well.

povilasb:
That of course does not test all scenarios of correct background task completion, e.g.

* connection close, should terminate local end connection driver
* connection close, should terminate remote connect driver
* connection close, should terminate incoming streams task, etc.

I'm planning to make a separate PR for those test cases - I have the barebones code.

povilasb:
Btw, this test occasionally just freezes, which indicates another potential bug. I can reproduce this more reliably with http://github.com/povilasb/quinn-tests and I believe this is not a test code issue. Hopefully will have more info later.

quinn/tests/many_connections.rs
```diff
@@ -0,0 +1,203 @@
...
+                ctx.borrow_mut().connections.push(conn);
+                Ok(())
+            });
+        runtime.spawn(task);
+    }
+
+    let wait_for_all_connections = conn_done_rx
+        .take(ctx.borrow().expected_messages as u64)
+        .for_each(|_| Ok(()));
+    unwrap!(runtime.block_on(wait_for_all_connections));
```
Ralith: Could [`Runtime::run`](https://docs.rs/tokio/0.1.18/tokio/runtime/current_thread/struct.Runtime.html#method.run) be used here instead of explicit message-passing?

quinn/tests/many_connections.rs
```diff
@@ -0,0 +1,203 @@
...
+
+/// Builds client configuration. Trusts given node certificate.
+fn configure_connector(node_cert: &[u8]) -> quinn::ClientConfig {
+    let mut peer_cfg_builder = quinn::ClientConfigBuilder::new();
+    let their_cert = unwrap!(quinn::Certificate::from_der(&node_cert));
+    unwrap!(peer_cfg_builder.add_certificate_authority(their_cert));
+    let mut peer_cfg = peer_cfg_builder.build();
+    let transport_config = unwrap!(Arc::get_mut(&mut peer_cfg.transport));
+    transport_config.idle_timeout = 0;
+    transport_config.keep_alive_interval = 10_000;
```
Ralith: Why disable the idle timeout? Leaving it at something reasonable might be useful for converting the test timeout into a failure. Additionally, keep-alive isn't useful if the idle timeout is disabled.

quinn/tests/many_connections.rs
povilasb: It's a neat idea, thanks. Although I will still have to notify endpoint driver that I'm done with the test, cause it is the last spawned task blocking `Runtime::run()` :)

Ralith: Could you accomplish that without explicit messaging by dropping the `Incoming` stream (along with all other `Endpoint` handles) after the expected number of connections have been handled?

Either way, it's nice to invoke `run` here to ensure that there aren't any stale futures left floating around, since the shutdown logic is nontrivial.

povilasb: hm, `Incoming` and `EndpointDriver` are spawned on the event loop, so I don't have handles to them anymore. Besides signaling them with messaging, I don't know how I could drop those tasks :/

Ralith: You know how many connections you're expecting, so you can use `Stream::take` to finish the `Incoming` stream after that number; once the resulting future is finished, the runtime will automatically drop it. `EndpointDriver` will automatically finish (and be dropped) once `Incoming` and all `Endpoint` values have been dropped and all `Connection`s are complete.

quinn/tests/many_connections.rs
```diff
@@ -0,0 +1,200 @@
...
+            .map_err(|e| panic!("Connection failed: {}", e))
+            .and_then(move |(conn_driver, conn, _)| {
+                current_thread::spawn(conn_driver.map_err(|_| ()));
+                write_to_peer(&conn, data);
+                ctx.borrow_mut().connections.push(conn);
+                Ok(())
+            });
+        runtime.spawn(task);
+    }
+    drop(endpoint); // we don't need it anymore, this will make EndpointDriver finish.
```
Ralith:
```suggestion
    drop(endpoint); // we don't need it anymore, this will make EndpointDriver finish after all connections are cleaned up```

File: quinn/tests/many_connections.rs [hunk 5]
```diff
@@ -0,0 +1,200 @@
...
+use unwrap::unwrap;
+
+struct TestContext {
+    /// Keep track of how many messages we've already received so that we'd know when test is
+    /// finished.
+    received_messages: usize,
+    expected_messages: usize,
+
+    /// Hold connections so that streams wouldn't be closed prematurely
+    connections: Vec<quinn::Connection>,
```
Ralith: This shouldn't be necessary; if dropping a `Connection` interferes with its still-active streams or `IncomingStreams`, that's a bug.

quinn/tests/many_connections.rs
```diff
@@ -0,0 +1,200 @@
...
+}
+
+/// Builds client configuration. Trusts given node certificate.
+fn configure_connector(node_cert: &[u8]) -> quinn::ClientConfig {
+    let mut peer_cfg_builder = quinn::ClientConfigBuilder::new();
+    let their_cert = unwrap!(quinn::Certificate::from_der(&node_cert));
+    unwrap!(peer_cfg_builder.add_certificate_authority(their_cert));
+    let mut peer_cfg = peer_cfg_builder.build();
+    let transport_config = unwrap!(Arc::get_mut(&mut peer_cfg.transport));
+    transport_config.idle_timeout = 30_0000;
```
Ralith:
```suggestion
    transport_config.idle_timeout = 30_000;
```
Unless it actually needs 300 seconds.

quinn/tests/many_connections.rs
```diff
@@ -0,0 +1,200 @@
+use crc::crc32;
+use futures::{Future, Stream};
+use rand::{self, RngCore};
+use std::cell::RefCell;
+use std::rc::Rc;
+use std::sync::Arc;
+use tokio::runtime::current_thread::{self, Runtime};
+use unwrap::unwrap;
+
+struct TestContext {
```
Ralith: Do we actually need this struct at all any more?

quinn/tests/many_connections.rs
```diff
@@ -0,0 +1,200 @@
...
+        .map_err(|e| panic!("read_to_end() failed: {}", e))
+        .and_then(move |(_stream, data)| {
+            assert!(hash_correct(&data));
+
+            let mut ctx = ctx.borrow_mut();
+            ctx.received_messages += 1;
+            if ctx.received_messages == ctx.expected_messages {
+                // let done_tx = unwrap!(ctx.test_done_tx.take());
+                // unwrap!(done_tx.send(()));
+            }
```
Ralith: Dead code

povilasb:
Oops, that was a rushed push. Thanks for the thorough review - I addressed all your comments now.

djc:
This was merged as #304.

PR #294, opened 2019-04-04
if let cleanup

newpavlov:
Following internals discussion I've decided to help improve `if let` usage in this project. I think it makes code smaller and more readable.

### DIFF
--- a/quinn-proto/src/range_set.rs
+++ b/quinn-proto/src/range_set.rs
@@ -95,8 +95,8 @@ impl RangeSet {
     }
 
     pub fn remove(&mut self, x: Range<u64>) -> bool {
-        let before = if let Some((start, end)) = self.pred(x.start) {
-            if end > x.start {
+        let before = match self.pred(x.start) {
+            Some((start, end)) if end > x.start => {
                 self.0.remove(&start);
                 if start < x.start {
                     self.0.insert(start, x.start);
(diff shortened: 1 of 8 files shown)


quinn-proto/src/range_set.rs
```diff
@@ -109,12 +109,11 @@ impl RangeSet {
...
                 }
                 true
-            } else {
-                false
-            }
-        } else {
-            false
+            },
+            Some(_) => false,
+            None => false,
```
Ralith: `Some(_) | None => false`?

quinn/src/platform/unix.rs
```diff
@@ -151,20 +151,20 @@ impl super::UdpExt for UdpSocket {
...
+                cmsg_level: libc::IPPROTO_IPV6, cmsg_type: libc::IPV6_TCLASS, ..
+            }) => unsafe { cmsg::decode::<libc::c_int>(cmsg) as u8 },
+            Some(cmsg @ libc::cmsghdr {
+                cmsg_level: libc::IPPROTO_IP, cmsg_type: libc::IP_TOS, ..
+            }) => unsafe { cmsg::decode::<u8>(cmsg) },
+            Some(cmsg @ libc::cmsghdr {
+                cmsg_level: libc::IPPROTO_IP, cmsg_type: libc::IP_RECVTOS, ..
+            }) => unsafe { cmsg::decode::<u8>(cmsg) },
+            Some(_) => 0,
+            None => 0,
```
Ralith: This doesn't save any lines of code even before rustfmt blows it up, and is a lot noisier; I appreciate the flattening, but I'm not sure it's a net win.

newpavlov: Yeah, the old tuple match is indeed better.

PR #297, opened 2019-04-06
Endpoint::close helper, related debugging

### LINKED ISSUE #293: Resolving quinn::Incoming
Is there a way to cleanly resolve `quinn::Incoming` we get from `quinn::EndpointBuilder::bind()` for instance ? I was wondering if it would be good to have a quit/close on `quinn::Endpoint` (something the user potentially holds on to) to exit everything - all `quinn::Connection/IncomingStreams/Incoming/Driver/ConnectionDriver`, all `read_to_end` and `write_all` etc., as at that point the user likely wants a clean exit of the event loop resolving all futures as quickly as possible. Maybe a drop on (last copy of) `Endpoint` could also do it in additional to explicit `close()/quit()/shutdown()`.

Ralith:
Fixes #293.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -364,17 +364,12 @@ impl ConnectionInner {
                 Async::Ready(Some(ConnectionEvent::Close { reason, error_code })) => {
                     self.close(error_code, reason);
                 }
-                Async::Ready(Some(ConnectionEvent::DriverLost)) => {
+                Async::Ready(None) => {
                     return Err(ConnectionError::TransportError(quinn::TransportError {
                         code: quinn::TransportErrorCode::INTERNAL_ERROR,
                         frame: None,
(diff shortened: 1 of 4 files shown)


Ralith:
I'm open to other ideas for the behavior of `Endpoint::connect` when the endpoint driver has been lost (or even when the endpoint has been closed). Right now it's the path of minimal API impact, and maybe it's nice that a new connection behaves similarly to a pre-existing one if either of those events has occurred. Alternatively, it might be more natural to have a `ConnectError` case for when it no longer makes sense to create new connections, but I'm not sure how that would be named/documented.

djc:
It could be named `ConnectError::ClosedEndpoint` or `EndpointDropped` or `NoEndpointDriver` or some combination of these. `ConnectError` feels more appropriate than `ConnectionError` to me for this case, given that we already know it's going to fail, but this PR is an improvement either way.

PR #299, opened 2019-04-07
Assorted cleanup

Ralith:
No significant behavioral change, but lots of quality-of-life improvements. Allowing per-connection loggers to be specified is especially helpful when trying to diagnose problems in code with many concurrent connections as it allows logs to easily be collected separately and only displayed once the relevant subset are identified.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -787,6 +787,11 @@ enum ConnectionOpts {
 /// These arise before any I/O has been performed.
 #[derive(Debug, Error)]
 pub enum ConnectError {
+    /// The endpoint can no longer create new connections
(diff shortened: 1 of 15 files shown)


quinn-proto/src/endpoint.rs
```diff
@@ -321,25 +321,21 @@ impl Endpoint {
     /// Initiate a connection
     pub fn connect(
         &mut self,
-        log: Option<Logger>,
+        config: ClientConfig,
```
Ralith: I'm not totally sure how I feel about taking this by value when we take `ServerConfig` through an `Arc`, but doing that here feels silly because all we do is pull out the fields; it'd be strictly more work.

PR #301, opened 2019-04-09
Assorted fixes

Ralith:
In particular, delaying handshake key discard reduces odds of connection loss under severe packet loss following connection establishment.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -541,7 +541,11 @@ impl Connection {
 
     fn set_key_discard_timer(&mut self, now: Instant) {
         let time = if self.spaces[SpaceId::Handshake as usize].crypto.is_some() {
-            now + self.pto() * 3
+            // Severe packet loss during this period can cause the connection to fail, so allocate
+            // lots of extra time to retry. This makes it marginally easier for an attacker that
+            // gains access to an endpoint's internal state shortly after the handshake completes to
+            // decrypt packets sent previously within the same connection.
+            now + self.pto() * 3 * 100
         } else if let Some(time) = self.prev_crypto.as_ref().and_then(|x| x.update_ack_time) {
             time + self.pto() * 3
         } else {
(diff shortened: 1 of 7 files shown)


quinn-proto/src/shared.rs
```diff
@@ -311,3 +311,31 @@ pub struct ClientOpts {
...
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub struct ResetToken([u8; RESET_TOKEN_SIZE]);
+
+impl From<[u8; RESET_TOKEN_SIZE]> for ResetToken {
+    fn from(x: [u8; RESET_TOKEN_SIZE]) -> Self {
+        Self(x)
+    }
+}
+
+impl std::ops::Deref for ResetToken {
```
djc: Could we `Deref` in a way that we don't need `[..]` in a few places?

quinn-proto/src/connection.rs
```diff
@@ -541,7 +541,9 @@ impl Connection {
 
     fn set_key_discard_timer(&mut self, now: Instant) {
         let time = if self.spaces[SpaceId::Handshake as usize].crypto.is_some() {
-            now + self.pto() * 3
+            // Severe packet loss during this period can cause the connection to fail, so allocate
```
djc: Maybe add something to the comment about the potential downside of what we're trading off here?

quinn-proto/src/shared.rs
Ralith: I tweaked the `extend` call; can't do much for `==` without adding a significantly larger boilerplatey manual `Eq` impl, I think.

quinn-proto/src/connection.rs
Ralith: Elaborated to the best of my understanding; might be worth double-checking on slack or against TLS references, though.

Ralith: The TLS draft [is very relaxed](https://quicwg.org/base-drafts/draft-ietf-quic-tls.html#discarding-unused-keys) about requirements here; handshake key discard does not appear to actually be mandated at all, though I understand a future draft may be more specific about reasonable behavior here.

PR #307, opened 2019-04-12
Initial attempt at backporting 0-RTT API

djc:
So far lacks the interop changes, and thus is completely untested other than for regressions.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -1,12 +1,10 @@
 use std::collections::VecDeque;
+use std::mem;
 use std::net::SocketAddr;
-use std::str;
 use std::sync::{Arc, Mutex};
 use std::time::Instant;
-use std::{io, mem};
 
 use bytes::Bytes;
-use err_derive::Error;
 use fnv::FnvHashMap;
 use futures::sync::{mpsc, oneshot};
 use futures::task::{self, Task};
@@ -14,7 +12,6 @@ use futures::Stream as FuturesStream;
 use futures::{Async, Future, Poll};
 use quinn_proto::{self as quinn, ConnectionHandle, Directionality, StreamId, TimerUpdate};
 use slog::Logger;
-use tokio_io::{AsyncRead, AsyncWrite};
 use tokio_timer::Delay;
 
 pub use crate::quinn::{
@@ -26,6 +23,7 @@ pub use crate::tls::{Certificate, CertificateChain, PrivateKey};
 pub use crate::builders::{
     ClientConfigBuilder, EndpointBuilder, EndpointError, ServerConfigBuilder,
 };
+use crate::streams::{NewStream, RecvStream, SendStream};
 use crate::{ConnectionEvent, EndpointEvent};
 
 /// Connecting future
@@ -286,14 +284,6 @@ impl FuturesStream for IncomingStreams {
     }
 }
 
-/// A stream initiated by a remote peer.
-pub enum NewStream {
-    /// A unidirectional stream.
-    Uni(RecvStream),
-    /// A bidirectional stream.
-    Bi(SendStream, RecvStream),
-}
-
 pub struct ConnectionRef(Arc<Mutex<ConnectionInner>>);
 
 impl ConnectionRef {
@@ -363,21 +353,21 @@ impl std::ops::Deref for ConnectionRef {
 pub struct ConnectionInner {
     log: Logger,
     epoch: Instant,
-    inner: quinn::Connection,
+    pub(crate) inner: quinn::Connection,
     driver: Option<Task>,
     handle: ConnectionHandle,
     connected: bool,
     timers: [Option<Delay>; quinn::Timer::COUNT],
     conn_events: mpsc::UnboundedReceiver<ConnectionEvent>,
     endpoint_events: mpsc::UnboundedSender<(ConnectionHandle, EndpointEvent)>,
-    blocked_writers: FnvHashMap<StreamId, Task>,
-    blocked_readers: FnvHashMap<StreamId, Task>,
+    pub(crate) blocked_writers: FnvHashMap<StreamId, Task>,
+    pub(crate) blocked_readers: FnvHashMap<StreamId, Task>,
     uni_opening: VecDeque<oneshot::Sender<Result<(StreamId, bool), ConnectionError>>>,
     bi_opening: VecDeque<oneshot::Sender<Result<(StreamId, bool), ConnectionError>>>,
     incoming_streams_reader: Option<Task>,
-    finishing: FnvHashMap<StreamId, oneshot::Sender<Option<ConnectionError>>>,
+    pub(crate) finishing: FnvHashMap<StreamId, oneshot::Sender<Option<ConnectionError>>>,
     /// Always set to Some before the connection becomes drained
-    error: Option<ConnectionError>,
+    pub(crate) error: Option<ConnectionError>,
     /// Number of live handles that can be used to initiate or handle I/O; excludes the driver
     ref_count: usize,
 }
@@ -527,7 +517,7 @@ impl ConnectionInner {
     }
 
     /// Wake up a blocked `Driver` task to process I/O
-    fn notify(&self) {
+    pub(crate) fn notify(&self) {
         if let Some(x) = self.driver.as_ref() {
             x.notify();
         }
@@ -567,7 +557,7 @@ impl ConnectionInner {
         self.close(0, Bytes::new());
     }
 
-    fn check_0rtt(&self) -> Result<(), ()> {
+    pub(crate) fn check_0rtt(&self) -> Result<(), ()> {
         match self.inner.is_handshaking() || self.inner.accepted_0rtt() {
             true => Ok(()),
             false => Err(()),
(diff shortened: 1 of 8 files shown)


quinn/src/builders.rs
```diff
@@ -174,6 +174,15 @@ impl ServerConfigBuilder {
...
 
+    /// Whether to accept 0-RTT data from clients
+    ///
+    /// 0-RTT allows application data to be transmitted without waiting for the handshake to
+    /// complete, but might be replayed by attackers.
+    pub fn zero_rtt(&mut self, enabled: bool) -> &mut Self {
+        Arc::make_mut(&mut self.config.crypto).max_early_data_size = enabled as u32 * 0xffffffff;
+        self
+    }
+
```
Ralith: This is redundant as of 94a37e399e4d68df42db1c79a4a93829e4994df5; we always declare support for 0-RTT in TLS, and leave it up to the application to decide whether to actually use it.

quinn/src/connection.rs
```diff
@@ -29,15 +29,57 @@ pub use crate::builders::{
...
+    /// successfully.
+    ///
+    /// Once [`Connection::is_handshaking`] returns false, further transmissions are unaffected by
+    /// these weaknesses.
+    ///
+    /// # Errors
+    ///
+    /// Outgoing connections are only 0-RTT-capable when a cryptographic session ticket cached from
+    /// a previous connection to the same server is available, and includes a 0-RTT key. If no such
+    /// ticket is found, the `Handshake` is returned unmodified.
```
```suggestion
    /// ticket is found, `self` is returned unmodified.
```

quinn/src/connection.rs
```diff
@@ -29,15 +29,57 @@ pub use crate::builders::{
...
+    ///
+    /// For incoming connections, a 0.5-RTT connection will always be successfully constructed.
+    pub fn into_0rtt(mut self) -> Result<(ConnectionDriver, Connection, IncomingStreams), Self> {
+        let can_0rtt = {
+            let conn = (self.0.as_mut().unwrap().0).lock().unwrap();
+            conn.inner.has_0rtt() & conn.inner.is_handshaking()
```
```suggestion
            conn.inner.has_0rtt() && conn.inner.is_handshaking()
```

Also, do we really need to fail this if the handshake has already completed? Seems like succeeding there would be harmless.

quinn/src/connection.rs
```diff
@@ -29,15 +29,57 @@ pub use crate::builders::{
...
+        };
+        match can_0rtt {
```
Ralith: Why not `if`?

quinn/src/connection.rs
```diff
@@ -883,6 +958,11 @@ pub enum WriteError {
     /// Unknown stream
     #[error(display = "unknown stream")]
     UnknownStream,
+    /// This was a 0-RTT stream and the server rejected it.
+    ///
+    /// Can only occur on clients, for streams initiated before the handshake completes.
```
Ralith: Should we have consistent wording on this comment and the `ReadError` variant?

quinn/src/connection.rs
```diff
@@ -566,6 +617,10 @@ impl SendStream {
     pub fn poll_write(&mut self, buf: &[u8]) -> Poll<usize, WriteError> {
         use crate::quinn::WriteError::*;
         let mut conn = self.conn.lock().unwrap();
+        if self.is_0rtt {
+            conn.check_0rtt()
+                .map_err(|()| WriteError::ZeroRttRejected)?;
+        }
```
Ralith: I think we need this on `poll_finish` too.

quinn/src/connection.rs
```diff
@@ -29,15 +29,57 @@ pub use crate::builders::{
...
-impl ConnectingFuture {
+impl Connecting {
     pub(crate) fn new(conn: ConnectionRef) -> Self {
         Self(Some(ConnectionDriver(conn)))
     }
+
+    /// Convert into a 0-RTT or 0.5-RTT connection at the cost of weakened security
+    ///
+    /// Additionally returns a future that will yield the connection's [`IncomingStreams`] when the
+    /// connection is established.
```
Ralith: This part seems to be missing. I'm not wedded to the specific approach--it was just the easy thing to do--but as is there's no way for a 0-RTT application to determine when the handshake has completed. Although, now that I think about it, I'm not actually certain whether that's really necessary...

quinn/src/connection.rs
```diff
@@ -29,15 +29,57 @@ pub use crate::builders::{
...
+    /// For incoming connections, a 0.5-RTT connection will always be successfully constructed.
+    pub fn into_0rtt(mut self) -> Result<(ConnectionDriver, Connection, IncomingStreams), Self> {
+        let can_0rtt = {
+            let conn = (self.0.as_mut().unwrap().0).lock().unwrap();
+            conn.inner.has_0rtt() & conn.inner.is_handshaking()
+        };
+        match can_0rtt {
+            true => {
+                let ConnectionDriver(conn) = self.0.take().unwrap();
+                conn.lock().unwrap().driver.take();
```
Ralith: This is a little odd; since spurious wakeups are permitted, futures typically don't bother going out of their way to prevent the occasional one. Doesn't hurt, though.

quinn/src/builders.rs
djc: Removed.

quinn/src/connection.rs
djc: Yeah, I did remove it somewhat consciously -- it didn't seem necessary, and this approach is simpler and provides a nice symmetry between the `<Connecting as Future>::Item` and the `into_0rtt()` `Ok` value. I'll update the documentation here and let me know if you think it needs to be a `Future` after all.

quinn/src/connection.rs
djc: Hmm, I guess not.

quinn/src/connection.rs
djc: Changed this to `if` in the end -- just experimenting with what's most readable and concise.

quinn/src/connection.rs
djc: I think that was there from when we only set `driver` if it was `None`, so it used to be needed. I removed it now here and in `<Connecting as Future>::poll()`.

quinn/src/connection.rs
djc: Yup, fiddled with the wording a bit more and made it consistent.

quinn/src/connection.rs
Ralith: I'm happy to wait and see if anyone comes up with a reason to want it. Only thing I can think of is a very specific case of status feedback.

quinn/src/connection.rs
djc: Yup, fixed it up.

quinn/src/connection.rs
```diff
@@ -26,18 +23,60 @@ pub use crate::tls::{Certificate, CertificateChain, PrivateKey};
...
+    /// For incoming connections, a 0.5-RTT connection will always be successfully constructed.
+    pub fn into_0rtt(mut self) -> Result<(ConnectionDriver, Connection, IncomingStreams), Self> {
+        if (self.0.as_mut().unwrap().0)
+            .lock()
+            .unwrap()
+            .inner
+            .has_0rtt()
+        {
+            let ConnectionDriver(conn) = self.0.take().unwrap();
+            conn.lock().unwrap().driver.take();
```
Ralith: Didn't you decide to remove this?

quinn/src/streams.rs
```diff
@@ -0,0 +1,444 @@
...
+    Finished,
+    /// The connection was closed.
+    #[error(display = "connection closed: {}", _0)]
+    ConnectionClosed(ConnectionError),
+    /// Unknown stream
+    #[error(display = "unknown stream")]
+    UnknownStream,
+    /// This was a 0-RTT stream and the server rejected it.
+    ///
+    /// Can only occur on clients for 0-RTT streams (opened using `Connecting::into_0rtt()`)..
```
Ralith:
```suggestion
    /// Can only occur on clients for 0-RTT streams (opened using `Connecting::into_0rtt()`).
```

quinn/src/streams.rs
```diff
@@ -0,0 +1,444 @@
...
+    },
+    /// The connection was closed.
+    #[error(display = "connection closed: {}", _0)]
+    ConnectionClosed(ConnectionError),
+    /// Unknown stream
+    #[error(display = "unknown stream")]
+    UnknownStream,
+    /// This was a 0-RTT stream and the server rejected it.
+    ///
+    /// Can only occur on clients for 0-RTT streams (opened using `Connecting::into_0rtt()`)..
```
Ralith:
```suggestion
    /// Can only occur on clients for 0-RTT streams (opened using `Connecting::into_0rtt()`).
```

quinn/src/connection.rs
djc: Yeah, not sure what happened with that, fixed now though!

quinn/src/connection.rs
```diff
@@ -26,18 +23,59 @@ pub use crate::tls::{Certificate, CertificateChain, PrivateKey};
...
+    /// Convert into a 0-RTT or 0.5-RTT connection at the cost of weakened security
+    ///
+    /// Opens up the connection for use before the handshake finishes, allowing the API user to
+    /// send data with 0-RTT encryption if the necessary key material is available. This is useful
+    /// for reducing start-up latency by beginning transmission of application data without waiting
+    /// for the handshake's cryptographic security guarantees to be established.
+    ///
+    /// # Security
+    ///
+    /// On outgoing connections, this enables transmisison of 0-RTT data, which might be vulnerable
```
```suggestion
    /// On outgoing connections, this enables transmission of 0-RTT data, which might be vulnerable
```

PR #308, opened 2019-04-13
Benchmark

djc:


Ralith:
Runs now (though it panics immediately).

djc:
Rebased onto current nightly. Off the top of my head, not yet sure how to deal with the `Runtime` borrowing issues.

quinn/benches/bench.rs
```diff
@@ -0,0 +1,143 @@
...
+    runtime.spawn(server_driver.map_err(|e| panic!("server driver failed: {}", e)));
+    runtime.spawn(client_driver.map_err(|e| panic!("client driver failed: {}", e)));
+
+    let runtime = Rc::new(RefCell::new(runtime));
+    {
+        let nested = runtime.clone();
+        runtime
+            .borrow_mut()
+            .spawn(server_incoming.for_each(move |(driver, _, incoming)| {
+                nested.borrow_mut().spawn(driver.map_err(|_| panic!()));
```
Ralith: Replacing this with `tokio::runtime::current_thread::spawn` should do the trick, I think.

djc:
@Ralith would you mind taking a look? This relatively straightforward translation to futures-0.1 seems to mostly work except that cleanup doesn't work correctly yet and I haven't been able to figure out why exactly.

Ralith:
Will investigate.

Ralith:
The benchmark was failing due to genuine connection ID backpressure, caused by the `ReadAllUnordered` future incorrectly returning `NotReady`. A future that is built out of other poll operations must never return `NotReady` unless one or more of its components returned `NotReady`, because a poll operation that returns `Ready` does not guarantee that the task will be woken again in the future.

djc:
I propose to merge this with the single benchmark for now, which is a good start and can inform further optimizations (and pulling it into other branches to measure impace is a pain). Left the second one commented out so we can enable it separately.

PR #310, opened 2019-04-13
Initial attempt at documenting public quinn-proto API

djc:
This is just an initial kick to make the docs for quinn-proto a bit more friendly (motivated in part by comments about this comparing us to quiche) for folks that need/want the lower-level API.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -32,6 +32,12 @@ use crate::{
     MIN_INITIAL_SIZE, MIN_MTU, RESET_TOKEN_SIZE, TIMER_GRANULARITY,
 };
 
+/// Protocol state and logic for a single QUIC connection
+///
+/// Objects of this type receive `ConnectionEvent`s and emit `EndpointEvents` and application
+/// `Event`s to make progress. To handle timeouts, a `Connection` returns timer updates and
+/// expects timeouts through various methods. A number of simple getter methods are exposed
+/// to allow callers to inspect some of the connection state.
 pub struct Connection {
     log: Logger,
     endpoint_config: Arc<EndpointConfig>,
@@ -987,6 +993,11 @@ impl Connection {
         }
     }
 
+    /// Process `ConnectionEvent`s, including incoming datagrams
+    ///
+    /// Will execute protocol logic upon receipt of a connection event, in turn preparing signals
+    /// (including application `Event`s, `EndpointEvent`s and outgoing datagrams) that should be
+    /// extracted through the relevant methods.
     pub fn handle_event(&mut self, event: ConnectionEvent) {
         use self::ConnectionEvent::*;
         match event {
@@ -2422,6 +2433,9 @@ impl Connection {
         Ok(())
     }
 
+    /// Open a single stream if possible
+    ///
+    /// Returns `None` if the streams in the given direction are currently exhausted.
     pub fn open(&mut self, direction: Directionality) -> Option<StreamId> {
         let id = self.streams.open(self.side, direction)?;
         // TODO: Queue STREAM_ID_BLOCKED if None
@@ -2453,12 +2467,16 @@ impl Connection {
         self.streams.alloc_remote_stream(self.side, ty);
     }
 
+    /// Accept a remotely initiated stream if possible
+    ///
+    /// Returns `None` if there are no new incoming streams for this connection.
     pub fn accept(&mut self) -> Option<StreamId> {
         let id = self.streams.accept(self.side)?;
         self.alloc_remote_stream(id.directionality());
         Some(id)
     }
 
+    /// Finish a send stream, signalling that no more data will be sent
     pub fn finish(&mut self, id: StreamId) {
         let ss = self
             .streams
@@ -2481,6 +2499,13 @@ impl Connection {
         });
     }
 
+    /// Read from the given recv stream, in undefined order
+    ///
+    /// While stream data is usually returned to the application in order, for some applications
+    /// it can make sense to leverage the reduced latency of unordered reads. For ordered reads,
+    /// the sibling `read()` method can be used.
+    ///
+    /// The return value if `Ok` contains the bytes and their offset in the stream.
     pub fn read_unordered(&mut self, id: StreamId) -> Result<(Bytes, u64), ReadError> {
         let (buf, len, more) = self
             .streams
@@ -2490,6 +2515,7 @@ impl Connection {
         Ok((buf, len))
     }
 
+    /// Read from the given recv stream
     pub fn read(&mut self, id: StreamId, buf: &mut [u8]) -> Result<usize, ReadError> {
         let (len, more) = self
             .streams
@@ -2519,6 +2545,7 @@ impl Connection {
         }
     }
 
+    /// Signal to the peer that it should stop sending on the given recv stream
     pub fn stop_sending(&mut self, id: StreamId, error_code: u16) {
         assert!(
             id.directionality() == Directionality::Bi || id.initiator() != self.side,
@@ -2635,6 +2662,7 @@ impl Connection {
         Ok(Some(number))
     }
 
+    #[doc(hidden)]
     pub fn force_key_update(&mut self) {
         let space = self.space(SpaceId::Data);
         let update = space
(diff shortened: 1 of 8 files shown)


quinn-proto/src/connection.rs
```diff
@@ -2452,12 +2466,17 @@ impl Connection {
         self.streams.alloc_remote_stream(self.side, ty);
     }
 
+    /// Accept a remotely initiated stream if possible
+    ///
+    /// Returns `None` if this connection does not currently allow the peer to open streams in
+    /// this direction.
```
Ralith: This is inaccurate; `accept` returns `None` iff there are no new incoming streams.

quinn-proto/src/connection.rs
```diff
@@ -2480,6 +2499,11 @@ impl Connection {
         });
     }
 
+    /// Read from the given recv stream, in undefined order
+    ///
+    /// While stream data is usually returned to the application in order, for some applications
+    /// it can make sense to leverage the reduced latency of unordered reads. For ordered reads,
+    /// the sibling `read()` method can be used.
```
Ralith: Should specify that the u64 is the offset.

quinn-proto/src/connection.rs
```diff
@@ -2726,18 +2756,22 @@ impl Connection {
...
     pub fn is_closed(&self) -> bool {
         self.state.is_closed()
     }
 
+    /// If the connection is drained
     pub fn is_drained(&self) -> bool {
         self.state.is_drained()
     }
 
+    /// If the connection accepted 0-RTT streams
```
```suggestion
    /// If the connection accepted 0-RTT streams after becoming established
```

quinn-proto/src/shared.rs
```diff
@@ -209,8 +220,8 @@ pub enum EndpointEvent {
 /// Mainly useful for identifying this connection's packets on the wire with tools like Wireshark.
 #[derive(Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub struct ConnectionId {
-    pub len: u8,
-    pub bytes: [u8; MAX_CID_SIZE],
+    pub(crate) len: u8,
+    pub(crate) bytes: [u8; MAX_CID_SIZE],
```
Ralith: I don't think we even need `pub(crate)` here.

quinn-proto/src/stream.rs
```diff
@@ -320,14 +320,18 @@ impl Send {
...
 #[derive(Debug, Error, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub enum WriteError {
     /// The peer is not able to accept additional data, or the connection is congested.
     #[error(display = "unable to accept further writes")]
     Blocked,
     /// The peer is no longer accepting data on this stream.
     #[error(display = "stopped by peer: error {}", error_code)]
-    Stopped { error_code: u16 },
+    Stopped {
+        /// Reason for stopping the stream
```
Ralith:
```suggestion
        /// Application-defined reason for stopping the stream
```

quinn-proto/src/stream.rs
```diff
@@ -495,14 +499,18 @@ impl Recv {
...
 #[derive(Debug, Error, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub enum ReadError {
     /// No more data is currently available on this stream.
     #[error(display = "blocked")]
     Blocked,
     /// The peer abandoned transmitting data on this stream.
     #[error(display = "reset by peer: error {}", error_code)]
-    Reset { error_code: u16 },
+    Reset {
+        /// Error code for the stream reset
```
Ralith:
```suggestion
        /// Application-defined error code for the stream reset
```

quinn-proto/src/connection.rs
djc: I don't quite understand what you mean here, can you elaborate/rephrase?

Ralith: This value may only become true after the connection is established. I'm not sure we should guarantee any particular value prior to that, but at the moment it always starts out false.

djc: Tried to improve this, have a look.

quinn-proto/src/connection.rs
```diff
@@ -3161,21 +3208,28 @@ impl RttEstimator {
...
 pub enum Timer {
+    /// When to declare packets lost if not acked
     LossDetection = 0,
+    /// When to close the connection after no activity
     Idle = 1,
     /// When the close timer expires, the connection has been gracefully terminated.
     Close = 2,
+    /// When keys are discarded because they should not be needed anymore
     KeyDiscard = 3,
+    /// When the path to the peer should be validated again
```
```suggestion
    /// When to give up on validating a new path to the peer
```

quinn-proto/src/connection.rs
```diff
@@ -3161,21 +3208,28 @@ impl RttEstimator {
     }
 }
 
+/// Kinds of timeouts needed to run the protocol logic
 #[derive(Debug, Copy, Clone, Ord, PartialOrd, Eq, PartialEq)]
 pub enum Timer {
+    /// When to declare packets lost if not acked
```
```suggestion
    /// When to send an ack-eliciting probe packet or declare unacked packets lost
```

quinn-proto/src/endpoint.rs
```diff
@@ -772,6 +778,7 @@ impl Default for ClientConfig {
     }
 }
 
+/// Internal key for a `Connection` currently open as part of an endpoint
```
Ralith:
```suggestion
/// Internal identifier for a `Connection` currently associated with an endpoint
```
"key" and "open" both have more formal meanings elsewhere in QUIC that may confuse things

quinn-proto/src/shared.rs
```diff
@@ -183,24 +183,35 @@ pub enum ConfigError {
...
     },
+    /// New connection identifiers have been issued for the Connection
     NewIdentifiers(Vec<(u64, ConnectionId)>),
+    /// A timeout has fired for a Connection
     Timer(Instant, Timer),
 }
 
 /// Events to be sent to the Endpoint
 #[derive(Clone, Debug)]
 pub enum EndpointEvent {
```
Ralith: I feel like we could reduce the API surface area by wrapping this in an opaque newtype (and relying on `Connection::is_drained` instead of matching on events in high level code)

quinn-proto/src/stream.rs
```diff
@@ -320,14 +320,18 @@ impl Send {
...
 #[derive(Debug, Error, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub enum WriteError {
     /// The peer is not able to accept additional data, or the connection is congested.
     #[error(display = "unable to accept further writes")]
     Blocked,
     /// The peer is no longer accepting data on this stream.
     #[error(display = "stopped by peer: error {}", error_code)]
-    Stopped { error_code: u16 },
+    Stopped {
+        /// Application-defined Reason for stopping the stream
```
Ralith:
```suggestion
        /// Application-defined reason for stopping the stream
```

quinn-proto/src/stream.rs
```diff
@@ -495,14 +499,18 @@ impl Recv {
...
 #[derive(Debug, Error, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub enum ReadError {
     /// No more data is currently available on this stream.
     #[error(display = "blocked")]
     Blocked,
     /// The peer abandoned transmitting data on this stream.
     #[error(display = "reset by peer: error {}", error_code)]
-    Reset { error_code: u16 },
+    Reset {
+        /// Application-defined error code for the stream reset
```
Ralith:
```suggestion
        /// Application-defined reason for resetting the stream
```
For consistency.

quinn-proto/src/shared.rs
djc: Do you think that would make the API easier to understand? I actually feel like just looking at the different event enums and how they flow between tasks will make it straightforward to explain what's going on here. Maybe we should just `#[doc(hidden)]` the variants here?

Ralith: All of these strike me as fairly obscure details of QUIC from a user's perspective; I'm not sure it's going to mean much to someone who's not familiar with the drained state, stateless resets, or the CID lifecycle.

djc:
Squashed everything and hid the `EndpointEvent` variants. Can you reapprove?

PR #312, opened 2019-04-14
Draft 19 transport parameter encoding

Ralith:
Missed this in the version negotiation changes.


Ralith:
We've erred towards keeping `Codec` for things that have no invalid bitpatterns in the past; I'm not wedded to that idea, but neither do I see a strong advantage to abandoning it when there's not any interesting polymorphism going on here anyway.

PR #315, opened 2019-04-16
Hack around broken macOS ECN ABI

### LINKED ISSUE #245: echo_dualstack fails on my MacBook
Tests on master are failing for me:

```
---- tests::echo_dualstack stdout ----
INFO connecting from [::]:0 to 127.0.0.1:60805
thread 'tests::echo_dualstack' panicked at 'assertion failed: `(left == right)`
  left: `13`,
 right: `16`', quinn/src/platform/cmsg.rs:69:5
```

This is with macOS 10.14.2.

Ralith:
Theoretically fixes #245. @djc, please test on the affected machine. Bonus points if you can double-check that the "ECN not acknowledged by peer" message is *not* displayed in the trace logs of the dual stack test.

I'd also like to cite the specific bug report tracking this issue upstream to make it easier to see whether it's still needed in the future, but I'm not sure where to find that.

### DIFF
--- a/quinn/src/platform/unix.rs
+++ b/quinn/src/platform/unix.rs
@@ -158,7 +158,16 @@ impl super::UdpExt for UdpSocket {
                     cmsg::decode::<u8>(cmsg)
                 },
                 (libc::IPPROTO_IPV6, libc::IPV6_TCLASS) => unsafe {
-                    cmsg::decode::<libc::c_int>(cmsg) as u8
+                    // Temporary hack around broken macos ABI. Remove once upstream fixes it.
+                    // https://bugreport.apple.com/web/?problemID=48761855
+                    if cfg!(target_os = "macos")
+                        && cmsg.cmsg_len as usize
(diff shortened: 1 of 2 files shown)


djc:
The upstream bug report is https://bugreport.apple.com/web/?problemID=48761855 (reported as fixed by an Apple engineer, but not publicly yet). Not sure how much info other people can see, though.

djc:
`echo_dualstack` is passing but still shows the debug output:

```
DEBG ECN not acknowledged by peer, connection: 99fcfe8637498af4, side: Client
DEBG ECN not acknowledged by peer, connection: 63f561425d0bbc96, side: Server
```

Ralith:
Strange; what's the value of the `cmsg::decode::<u8>(cmsg)` call?

djc:
The value of `cmsg::decode::<u8>(cmsg)` is 0. (BTW, the `echo_v4` test shows the same debug output.)

Ralith:
That's strange, but harmless. Only issue would have been if there was some weird garbage in there; I think we're good to go.

PR #317, opened 2019-04-17
Access stream id from other crates for H3

stammw:


Ralith:
Out of curiosity, what specifically does H3 need this for?

stammw:
@Ralith Some messages on [QPACK streams](https://quicwg.org/base-drafts/draft-ietf-quic-qpack.html#rfc.section.4.4.2) are relying on this value, as well as some [H3 control frames](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#frame-goaway).

Plus, it might come handy if some streams / requests need to be indexed with a key.

PR #318, opened 2019-04-17
Add missing documentation comment

djc:


djc:
@stammw not sure what you mean, this is a getter already?

stammw:
I mean a method that returns stream_id value. Augmenting visibility for the inner value will make the user able to modify it, and I'm not sure that's desirable.

stammw:
Oh sorry, I commented the wrong PR...

PR #321, opened 2019-04-23
ReadToEnd future cleanup

Ralith:
You can't do anything with a finished receive stream, so there's no point preserving it even under futures 0.1 idioms.

### DIFF
--- a/interop/src/main.rs
+++ b/interop/src/main.rs
@@ -336,7 +336,7 @@ fn h3_get(
             recv.read_to_end(usize::max_value())
                 .map_err(|e| format_err!("failed to send Request frame: {}", e))
         })
-        .and_then(move |(_, data)| h3_resp(&table, data))
+        .and_then(move |data| h3_resp(&table, data))
 }
 
 fn h3_resp(table: &qpack::DynamicTable, data: Vec<u8>) -> Result<Bytes> {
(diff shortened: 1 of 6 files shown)


Ralith:
Tweaked the optimization to be less dubious. This was motivated by [a flamegraph](https://i.imgur.com/rTECAka.png) @nbaksalyar posted which showed the reallocs taking a huge chunk of time in many_connections, though it's been difficult to show an absolute time improvement, perhaps due to confounding factors in that particular test. Still, this seems like the same thing that bit us in the original `VecDeque<u8>` `Assembler` impl.

djc:
To be sure this version makes sense. I just think we should have a really good case for using unsafe blocks.

djc:
BTW, this still seems to create a return value that potentially starts with a lot of zero bytes. Is that really what we want? If so, we should at least document the behavior IMO.

Ralith:
> BTW, this still seems to create a return value that potentially starts with a lot of zero bytes

I don't think this is true. We only return a buffer if we receive `ReadError::Finished`, which guarantees that every byte of the stream has been read.

djc:
But what if some parts of the stream had been read using ordered reads before starting unordered reads?

Ralith:
Good catch; documented the case. Alternatively, we could set a flag on any read and panic or return an error if the flag's set on the `read_to_end` call.

djc:
Yeah, I think I'd prefer to set a flag such that `read_to_end()` will return an `Err` if any other read has been done on the stream. In my opinion the current design is a little hard to explain, and only allowing `read_to_end()` from offset 0 doesn't seem a very onerous restriction.

Ralith:
How's this?

djc:
Great!

PR #324, opened 2019-04-26
Fix clippy suggestions in interop tool

djc:


interop/src/main.rs
```diff
@@ -296,7 +296,7 @@ fn run(log: Logger, options: Opt) -> Result<()> {
         print!("3");
     }
 
-    println!("");
+    println!();
```
Ralith: Neat!

PR #332, opened 2019-04-28
Connection lifecycle fixes

### LINKED ISSUE #331: Remotely-initiated streams should be accessible on closed connections
In the case where the peer creates a stream, sends less than a flow control window's worth of data, finishes it, and then closes the connection, the high-level interface will currently prevent the local application from reading the stream if the close is processed before the application has a chance to accept the stream, even though the stream was reliably delivered.

Ralith:
Looks like we forgot to update `quinn::Incoming` to yield a stream of handshake futures to represent the fact that its connections are not yet fully established.

Fixes #331

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -93,9 +93,7 @@ impl Future for Connecting {
     }
 }
 
-pub(crate) fn new_connection(
-    conn: ConnectionRef,
-) -> (ConnectionDriver, Connection, IncomingStreams) {
+fn new_connection(conn: ConnectionRef) -> (ConnectionDriver, Connection, IncomingStreams) {
     (
         ConnectionDriver(conn.clone()),
         Connection(conn.clone()),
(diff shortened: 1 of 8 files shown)


quinn/src/connection.rs
```diff
@@ -19,15 +19,18 @@ use tokio_timer::Delay;
...
+/// Be sure to spawn the `ConnectionDriver` when complete.
 pub struct Connecting(Option<ConnectionDriver>);
 
 impl Connecting {
     pub(crate) fn new(conn: ConnectionRef) -> Self {
         Self(Some(ConnectionDriver(conn)))
     }
 
-    /// Convert into a 0-RTT or 0.5-RTT connection at the cost of weakened security
+    /// Convert into a 0-RTT or 0.5-RTT connection at the cost of weakened security. Be sure to
```
djc: Hmm, so I think it should go after the newline here too, and in the third hunk as well. Otherwise the rustdocs will look weird.

Ralith: Current rustdoc doesn't seem to treat the first line of method documentation specially, unless I'm missing something.

djc: Ah, you're right, it doesn't matter for methods!

PR #334, opened 2019-04-28
Make open_{uni,bi} return nominative type for user convenience

stammw:


djc:
No `try_ready!`?

quinn/src/connection.rs
```diff
@@ -277,6 +272,53 @@ impl FuturesStream for IncomingStreams {
     }
 }
 
+/// A future that will resolve into an opened outgoing unidirectionnal stream
```
```suggestion
/// A future that will resolve into an opened outgoing unidirectional stream
```

quinn/src/connection.rs
```diff
@@ -277,6 +272,53 @@ impl FuturesStream for IncomingStreams {
...
+            Ok(Async::Ready(Ok((stream, is_0rtt)))) => Ok(Async::Ready(SendStream::new(
+                self.conn.clone(),
+                stream,
+                is_0rtt,
+            ))),
+        }
+    }
+}
+
+/// A future that will resolve into an opened outgoing bidirectionnal stream
```
```suggestion
/// A future that will resolve into an opened outgoing bidirectional stream
```

quinn/src/connection.rs
```diff
@@ -277,6 +272,53 @@ impl FuturesStream for IncomingStreams {
...
+    conn: ConnectionRef,
+}
+
+impl Future for OpenUni {
+    type Item = SendStream;
+    type Error = ConnectionError;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        match self.recv.poll() {
+            Err(_) => unreachable!(),
```
```suggestion
            Err(_) => unreachable!("oneshot sender won't be dropped while `self.conn` is keeping the `ConnectionInner` alive"),
```
And similar for `OpenBi`.

stammw:
@djc The fact that there is a nested result to forward as the current future's `Error` and that the enclosing `Result` from the channel is considered unreachable didn't seem to play well with `try_ready!`, but I'll try to put some more effort.

PR #339, opened 2019-05-05
[WIP] H3 exploration

stammw:
This PR is only intended to exchange on early design of the `H3` crate.

### DIFF
--- a/quinn-h3/examples/h3.rs
+++ b/quinn-h3/examples/h3.rs
@@ -30,6 +30,7 @@ use quinn_h3::{
     frame::{Error as FrameError, HeadersFrame, HttpFrame},
     Connection, StreamType,
 };
+use quinn_proto::StreamId;
 
 type Result<T> = std::result::Result<T, Error>;
 
@@ -196,9 +197,12 @@ fn handle_connection(
             |ref mut request| {
                 println!("received an exciting request !: {:?}", request.headers);
                 request
-                    .send_response(HeadersFrame {
-                        encoded: b"wow, much req, very encode"[..].into(),
-                    })
+                    .send_response(
+                        vec![
+                            (":status", "200"), // TODO ResponseBuilder
+                        ]
+                        .into_iter(),
+                    )
                     .map_err(|e| println!("send response failed with: {}", e))
             },
         ));
@@ -460,23 +464,24 @@ impl ControlStream {
 
 enum RecvRequestState {
     Receiving(FrameStream<RecvStream>, SendStream),
+    Decoding(HeadersFrame),
     Ready,
 }
 
 struct RecvRequest {
     state: RecvRequestState,
     conn: H3ConnectionRef,
+    stream_id: StreamId,
     streams: Option<(FrameStream<RecvStream>, SendStream)>,
-    headers: Option<HeadersFrame>,
 }
 
 impl RecvRequest {
     fn new(recv: RecvStream, send: SendStream, conn: H3ConnectionRef) -> Self {
         Self {
             conn,
-            state: RecvRequestState::Receiving(FrameStream::with(recv), send),
+            stream_id: send.stream,
             streams: None,
-            headers: None,
+            state: RecvRequestState::Receiving(FrameStream::with(recv), send),
         }
     }
 }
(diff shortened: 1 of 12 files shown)


quinn-h3/examples/h3.rs
```diff
@@ -486,33 +488,46 @@ impl Future for RecvRequest {
...
+                    let frame = try_ready!(frames.poll());
+                    match frame {
+                        None => return Err(format_err!("recieved an empty request")),
+                        Some(HttpFrame::Headers(f)) => {
+                            match mem::replace(&mut self.state, RecvRequestState::Decoding(f)) {
+                                RecvRequestState::Receiving(f, s) => self.streams = Some((f, s)),
+                                _ => unreachable!("Invalid state"),
+                            }
+                        }
+                        Some(_) => return Err(format_err!("first frame is not headers")),
```
djc: I think we should have `{ }` block around return statements.

quinn-h3/src/frame.rs
```diff
@@ -5,15 +5,15 @@ use bytes::{Buf, BufMut, Bytes};
 use quinn_proto::coding::{BufExt, BufMutExt, Codec, UnexpectedEnd};
 use quinn_proto::varint;
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
```
djc: Why do we need the `Clone` derive here (and elsewhere)?

quinn-h3/src/frame.rs
```diff
@@ -296,15 +296,15 @@ impl Default for SettingsFrame {
 }
 
 impl SettingsFrame {
-    fn encode<T: BufMut>(&self, buf: &mut T) {
+    pub fn encode<T: BufMut>(&self, buf: &mut T) {
```
djc: Should this be `pub(crate)` instead?

quinn-h3/src/lib.rs
```diff
@@ -37,4 +39,206 @@ impl StreamType {
     pub fn encode<W: BufMut>(&self, buf: &mut W) {
         buf.write_var(self.0);
     }
+    pub fn decode<T: Buf>(buf: &mut T) -> Result<Self, UnexpectedEnd> {
+        Ok(StreamType(buf.get_var()?))
+    }
+}
+
+#[derive(Debug, Eq, PartialEq)]
```
djc: Probably better to put all the error types in a dedicated module rather than in `lib.rs`.

quinn-h3/src/connection.rs
```diff
@@ -0,0 +1,179 @@
...
+    EncoderStreamError { reason: qpack::DecoderError },
+}
+
+pub enum State {
+    Open,
+    Closing(ErrorCode),
+    Closed(),
+}
+
+pub struct Connection {
```
djc: Let's `derive(Default)` here and use that to implement `new()`?

quinn-h3/examples/h3.rs
```diff
@@ -449,47 +460,93 @@ impl ControlStream {
...
-        //     }
-        // }
+        let new_state = match &mut self.state {
+            RecvRequestState::Receiving(frames, _) => {
+                let frame = try_ready!(frames.poll());
+                match frame {
+                    None => return Err(format_err!("recieved an empty request")),
+                    Some(HttpFrame::Headers(f)) => {
+                        self.headers = Some(f);
+                        RecvRequestState::Ready
```
djc: You can probably do the replacement in here?

quinn-h3/src/frame.rs
stammw: It is

quinn-h3/src/frame.rs
stammw: Maybe it was useful for testing, but smells like a bad copy / paste.

quinn-h3/examples/h3.rs
stammw: Is that for readability ?

quinn-h3/src/frame.rs
djc: Hmm, maybe it's not actually needed. To me `return` always seems like more of a statement, and so it's required to be a in a block. But of course it's actually an expression that returns the never type, so maybe this is just me being stuck in my old ways.

quinn-h3/examples/h3.rs
stammw: Do you mean using `mem::replace` ?

djc: Yes.

quinn-h3/examples/h3.rs
djc: Hmm, maybe it's not actually needed. To me return always seems like more of a statement, and so it's required to be a in a block. But of course it's actually an expression that returns the never type, so maybe this is just me being stuck in my old ways.

djc:
@stammw is this obviated by #340?

stammw:
Yes. I kept this I case some of those comments apply to #340. But it surely has diverged a lot.

stammw:
I checked your comments in the current code, everything is fixed. Closing

PR #342, opened 2019-05-16
Coalesce outgoing packets

djc:
Results in 3 fewer datagrams on the `echo_v4` test.

Currently the `server_hs_retransmit()` test is still broken. @Ralith any suggestions if/how this should be fixed?

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -1957,25 +1957,20 @@ impl Connection {
                     return None;
                 }
             }
-            _ => {
-                let id = SpaceId::iter()
-                    .find(|&x| self.space(x).crypto.is_some() && self.space(x).can_send())
-                    .or_else(|| {
-                        if self.space(SpaceId::Data).crypto.is_some() && self.can_send_1rtt() {
-                            Some(SpaceId::Data)
-                        } else if self.io.probes != 0 {
-                            Some(self.highest_space)
-                        } else if self.zero_rtt_crypto.is_some()
-                            && self.side.is_client()
-                            && (self.space(SpaceId::Data).can_send() || self.can_send_1rtt())
-                        {
-                            Some(SpaceId::Data)
-                        } else {
-                            None
-                        }
-                    })?;
-                (id, false)
-            }
+            _ => (
+                SpaceId::iter()
+                    .filter(|&x| {
+                        (self.space(x).crypto.is_some() && self.space(x).can_send())
+                            || (x == self.highest_space && self.io.probes != 0)
+                            || (x == SpaceId::Data
+                                && ((self.space(x).crypto.is_some() && self.can_send_1rtt())
+                                    || (self.zero_rtt_crypto.is_some()
+                                        && self.side.is_client()
+                                        && (self.space(x).can_send() || self.can_send_1rtt()))))
+                    })
+                    .next()?,
+                false,
+            ),
         };
         let probe = !close && self.io.probes != 0;
         let mut ack_only = self.space(space_id).pending.is_empty();
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -17,8 +17,8 @@ use crate::crypto::{
 };
 use crate::frame::FrameStruct;
 use crate::packet::{
-    set_payload_length, Header, LongType, Packet, PacketNumber, PartialDecode, SpaceId,
-    LONG_RESERVED_BITS, SHORT_RESERVED_BITS,
+    Header, LongType, Packet, PacketNumber, PartialDecode, SpaceId, LONG_RESERVED_BITS,
+    SHORT_RESERVED_BITS,
 };
 use crate::range_set::RangeSet;
 use crate::shared::{
@@ -2042,7 +2042,6 @@ impl Connection {
         };
         let mut buf = Vec::new();
         let partial_encode = header.encode(&mut buf);
-        let header_len = buf.len();
 
         if probe && ack_only && header.is_1rtt() {
             // Nothing ack-eliciting to send, so we need to make something up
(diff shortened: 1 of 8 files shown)


quinn-proto/src/connection.rs
```diff
@@ -1961,173 +1961,194 @@ impl Connection {
...
-        } else {
-            false
-        };
+            let mut padded = if self.side.is_client() && space_id == SpaceId::Initial {
+                // Initial-only packets MUST be padded
+                buf.resize(MIN_INITIAL_SIZE - crypto.packet.tag_len(), 0);
+                true
+            } else {
+                false
+            };
```
Ralith: When sending 0-RTT with an Initial, the padding should be moved to the 0-RTT packet so that 0-RTT data is able to take the place of padding.

Ralith:
The test failure is unrelated to the correctness of this change. `server_hs_retransmit` attempts to test the retransmission of just the handshake portion of the server's initial flight, but with coalesced packets that no longer makes sense. I'll see if I can refactor it into something that still exercises the anti-amplification logic, at least.

quinn-proto/src/connection.rs
```diff
@@ -1923,6 +1923,253 @@ impl Connection {
...
+                            0
+                        },
+                        is_crypto_packet: space_id != SpaceId::Data && !sent.crypto.is_empty(),
+                        ack_eliciting: !ack_only,
+                        retransmits: sent,
+                    },
+                );
+            }
+
+            if !coalesce || buf.capacity() - buf.len() < 40 {
```
Ralith: Can we make `40` not a magic number here?

quinn-proto/src/connection.rs
```diff
@@ -2172,7 +2172,7 @@ impl Connection {
             })
             .packet
             .tag_len();
-        let max_size = self.mtu as usize - tag_len;
+        let max_size = buf.capacity() - tag_len;
```
Ralith: I'm concerned that this makes it easy to inadvertently extend the buffer past `self.mtu`, if e.g. one of the guards in `populate_packet` has a bug. Maybe assert against that in `poll_transmit`?

quinn-proto/src/connection.rs
djc: Sure, what do you want to call it? `MIN_COALESCING_SIZE`? I just picked a value at random.

quinn-proto/src/packet.rs
```diff
@@ -872,16 +892,9 @@ mod tests {
         };
         let encode = header.encode(&mut buf);
         let header_len = buf.len();
-        buf.resize(header_len + 16, 0);
-        set_payload_length(&mut buf, header_len, 1, client_crypto.tag_len());
-        assert_eq!(
-            buf[..],
-            hex!("c0ff0000145006b858ec6f80452b00402100 00000000000000000000000000000000")[..]
-        );
```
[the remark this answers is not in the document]
djc: The new API no longer exposes the steps separately, so if we still want to have this test we have to duplicate the code from `PartialEncode::finish()` or break that code up in smaller pieces that can be called independently. I prefer not to make the functional code "less nice" in order to appease test code.

quinn-proto/src/packet.rs
```diff
@@ -419,23 +432,39 @@ impl Header {
 }
 
 pub struct PartialEncode {
-    pn: Option<usize>,
+    pub header_len: usize,
+    // Packet number position, packet number length, payload length needed
```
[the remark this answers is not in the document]
djc: I suppose we could.

quinn-proto/src/connection.rs
Ralith: `MIN_PACKET_SPACE` maybe?

djc:
There are some more issues to watch for: the spec mentions coalesced packets should follow the order of Initial, 0-RTT, Handshake, 1-RTT. As such, if the client is sending `0-RTT` and `Handshake` packages they could be in the wrong order. For Quinn this probably doesn't matter but maybe other implementations would have trouble with it?

quinn-proto/src/connection.rs

quinn-proto/src/connection.rs

Ralith:
In lieu of a strict requirement by the draft or specific knowledge of an implementation that cannot decrypt 0-RTT packets received following handshake packets, I don't think that's worth going well out of our way for. After all, implementations should hold onto 0-RTT keys for a while even after reaching 1-RTT to improve performance under packet reordering.

quinn-proto/src/connection.rs
Ralith: On further consideration, that's a danger regardless of this change, so not a blocker.

quinn-proto/src/packet.rs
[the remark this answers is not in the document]
Ralith: Fair enough.

quinn-proto/src/connection.rs
Ralith: A similar case can also arise in the client's second flight, when an Initial ACK is followed immediately by a Handshake ACK+CRYPTO and potentially by 1-RTT data.

quinn-proto/src/packet.rs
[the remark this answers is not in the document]
djc: This is actually much nicer!

quinn-proto/src/connection.rs
djc: Much better.

quinn-proto/src/connection.rs
```diff
@@ -1961,173 +1961,197 @@ impl Connection {
...
+            };
 
-        let mut padded = if self.side.is_client() && space_id == SpaceId::Initial {
-            // Initial-only packets MUST be padded
-            buf.resize(MIN_INITIAL_SIZE - crypto.packet.tag_len(), 0);
-            true
-        } else {
-            false
-        };
+            let mut padded = if exact_number == 0 && self.side.is_client() &&
```
Ralith: The client's first flight might be retransmitted and retransmits do not reuse packet numbers. Initial packets may also be sent after the client's first flight but before handshake keys are available (e.g. ACKs), and those too must be padded.

djc: Can you take a look at the current approach? Seems nice, though I haven't fully thought through the interactions with `close` and `offpath_responses` -- not sure how likely these are to occur at this stage?

Ralith: Close and, I think, challenge responses can both be sent during the handshake, but I think the interactions here are correct.

quinn-proto/src/packet.rs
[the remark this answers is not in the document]
Ralith: Would it make sense to restore this test with an unset payload length, just to have specific coverage of the other aspects of header encoding?

djc:
Might make sense, but in some experiments I couldn't get it to do something that seemed useful. Want to take a swing at it?

PR #345, opened 2019-05-18
Make stream read API more Stream-like

djc:
It occurred to me that `ReadError::Finished` is a bit weird because `Finished` isn't really an error. This change fixes that only in the `quinn` crate, where it does seem to achieve a natural alignment with `futures::Stream`-like handling. I looked at extending it to quinn-proto as well, but I'm not sure it's worth it there.

### DIFF
--- a/quinn/src/streams.rs
+++ b/quinn/src/streams.rs
@@ -4,7 +4,7 @@ use std::str;
 use bytes::Bytes;
 use err_derive::Error;
 use futures::sync::oneshot;
-use futures::task;
+use futures::{task, try_ready};
 use futures::{Async, Future, Poll};
 use proto::{ConnectionError, StreamId};
 use tokio_io::{AsyncRead, AsyncWrite};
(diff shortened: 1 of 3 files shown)


quinn/src/streams.rs
```diff
@@ -330,28 +330,22 @@ impl Future for ReadToEnd {
     type Error = ReadError;
```
djc: The fact that we have the `TooLong` error here suggests to me that maybe a generic `ReadError` isn't quite right for `ReadToEnd` and we should maybe have a wrapper with `Read` and `TooLong` variants?

Ralith: I agree that the error case does not belong in the generic `ReadError`; `ReadToEnd` is intended to be a helper that could just as well well exist outside of the crate entirely.

djc: Fixed that up.

djc:
Will leave proto for a separate PR.

djc:
So I can get on board with making it more like `AsyncRead` instead of more like `Stream` (although I prefer the more-typed API, myself), but for unordered reads it's not as obvious what the sentinel value has to be... we could return an empty `Bytes`, but with what offset? And then if we do use `Option` for unordered reads, does it still make sense to use the sentinel value for ordered reads?

PR #346, opened 2019-05-18
Handle stream cleanup after reading inside Streams

djc:
The way this previously redirected through `Connection` didn't make much sense.

### DIFF
--- a/quinn-proto/src/stream.rs
+++ b/quinn-proto/src/stream.rs
@@ -135,13 +135,26 @@ impl Streams {
 
     pub fn read(&mut self, id: StreamId, buf: &mut [u8]) -> Result<(usize, bool), ReadError> {
         let rs = self.get_recv_mut(id).ok_or(ReadError::UnknownStream)?;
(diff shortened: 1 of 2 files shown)


quinn-proto/src/stream.rs
```diff
@@ -135,13 +135,25 @@ impl Streams {
...
+            Ok((buf, len)) => Ok((buf, len, rs.receiving_unknown_size())),
+            Err(e) => Err(self.read_error_cleanup(id, e)),
+        }
+    }
+
+    fn read_error_cleanup(&mut self, id: StreamId, err: ReadError) -> ReadError {
+        if let ReadError::Finished | ReadError::Reset { .. } = err {
+            self.maybe_cleanup(id);
+        }
+        err
```
Ralith: Feels a bit weird to thread the `ReadError` through this, especially now that we're not being cheeky with `map_err` any more.

djc: Fair point, done.

PR #347, opened 2019-05-20
Move around stream direction abstraction layer

djc:


quinn-proto/src/stream.rs
```diff
@@ -199,11 +190,7 @@ impl Streams {
 
     /// Whether a presumed-local stream is or was previously open
```
Ralith: Update this comment while we're in here?

quinn-proto/src/lib.rs
```diff
@@ -131,6 +131,13 @@ pub enum Directionality {
     Uni = 1,
 }
 
+impl Directionality {
+    #[doc(hidden)]
+    pub fn iter() -> impl Iterator<Item = Self> {
```
Ralith: I don't think this needs to be exported at all.

djc: Fixed.

quinn-proto/src/lib.rs
```diff
@@ -131,6 +131,12 @@ pub enum Directionality {
     Uni = 1,
 }
 
+impl Directionality {
+    pub(crate) fn iter() -> impl Iterator<Item = Self> {
```
Ralith: Does this even need to be `pub(crate)`, since it's defined in the root module?

djc: Oh, you're right!

djc:
`Directionality` is kind of long, but I think `Dir` would not be very obvious. I feel having a shorthand for the variable name in context (field names have been left long) strikes the right balance here.

PR #349, opened 2019-05-21
Fix benchmark

djc:


djc:
Apparently not... I'll fix that in a follow-up.

PR #352, opened 2019-05-24
Move some code around

djc:
Fixes new warnings from 1.35.0.


quinn-proto/src/connection.rs
```diff
@@ -3379,6 +3380,8 @@ fn instant_saturating_sub(x: Instant, y: Instant) -> Duration {
     }
 }
 
+/// Value used in ACKs we transmit
+pub const ACK_DELAY_EXPONENT: u8 = 3;
```
```suggestion
const ACK_DELAY_EXPONENT: u8 = 3;
```

djc: Good point!

PR #367, opened 2019-06-07
Assorted clippy + style fixes

djc:
@stammw please take note, there are a few recurring patterns here that you should try to avoid in the future. 👍

### DIFF
--- a/quinn-h3/src/client.rs
+++ b/quinn-h3/src/client.rs
@@ -31,7 +31,7 @@ pub struct ClientBuilder<'a> {
 impl<'a> ClientBuilder<'a> {
     pub fn new(endpoint: EndpointBuilder<'a>) -> Self {
         Self {
-            endpoint: endpoint,
+            endpoint,
             log: None,
             settings: Settings::default(),
         }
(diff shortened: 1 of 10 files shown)


stammw:
I'm back :)

Sure, I'll get more careful on those. Also, I think I should start to use clippy.

Shall I merge this myself? (I don't remember the workflow

djc:
I don't mind acting as a clippy catch-all, but there are some Rust idioms that you should learn.

Just approve it and then click "Rebase and merge". 👍

quinn-h3/src/client.rs
```diff
@@ -57,7 +57,9 @@ impl<'a> ClientBuilder<'a> {
             Client {
                 endpoint,
                 settings: self.settings,
-                log: self.log.unwrap_or(Logger::root(slog::Discard, o!())),
+                log: self
```
stammw: I don't really understand this change.

djc: The idea here is that clippy dislikes if you make a function call in an `unwrap_or()` (or several other methods) because the function call gets evaluated whether it's necessary or not. With `unwrap_or_else()` (generally the `_else()` variants), you pass in a closure that only gets called when it's actually necessary, which should be better for performance reasons.

stammw: Ah, makes much more sense to me now, thanks.

PR #368, opened 2019-06-07
Switch to git http for now to enable HTTP 3 support

djc:


quinn-h3/Cargo.toml
```diff
@@ -33,7 +33,7 @@ bitlab = "0.8.1"
 lazy_static = "1"
 err-derive = "0.1.5"
 slog = "2.1"
-http = "0.1"
+http = { git = "https://github.com/hyperium/http/", branch = "0.2.x" }
```
Ralith: Should this be pinned to a specific rev for reproducibility?

djc: Fair point! Fixed.

PR #371, opened 2019-06-09
Implement Clone and Debug for builder types

djc:
This is required by hyper, and seems fairly reasonable for these types.

### DIFF
--- a/quinn-proto/src/crypto.rs
+++ b/quinn-proto/src/crypto.rs
@@ -12,7 +12,7 @@ use std::str;
 
 use bytes::BytesMut;
 
-use crate::shared::ConnectionId;
+use crate::shared::{ConfigError, ConnectionId};
 use crate::transport_parameters::TransportParameters;
 use crate::{ConnectError, Side, TransportError};
 
@@ -25,6 +25,8 @@ pub mod rustls;
 pub trait Session: Sized {
     /// Type used to hold configuration for client sessions
     type ClientConfig: ClientConfig<Self>;
+    /// Type used to sign various values
+    type HmacKey: HmacKey;
     /// Type used to represent packet protection keys
     type Keys: Keys + Sized;
     /// Type used to hold configuration for server sessions
@@ -123,4 +125,19 @@ pub trait HeaderKeys {
     fn encrypt(&self, pn_offset: usize, packet: &mut [u8]);
     /// The sample size used for this key's algorithm
     fn sample_size(&self) -> usize;
+}
+
+/// A key for signing with HMAC-based algorithms
+pub trait HmacKey: Sized {
+    /// Length of the key input
+    const KEY_LEN: usize;
+    /// Type of the signatures created by `sign()`
+    type Signature: AsRef<[u8]>;
+
+    /// Method for creating a key
+    fn new(key: &[u8]) -> Result<Self, ConfigError>;
+    /// Method for signing a message
+    fn sign(&self, data: &[u8]) -> Self::Signature;
+    /// Method for verifying a message
+    fn verify(&self, data: &[u8], signature: &[u8]) -> Result<(), ()>;
 }
--- a/quinn-proto/src/shared.rs
+++ b/quinn-proto/src/shared.rs
@@ -10,9 +10,10 @@ use ring::{digest, hmac::SigningKey};
 use slog::Logger;
 
 use crate::connection::Timer;
+use crate::crypto::HmacKey;
 use crate::frame::NewConnectionId;
 use crate::packet::PartialDecode;
-use crate::{crypto, varint, MAX_CID_SIZE, MIN_CID_SIZE, RESET_TOKEN_SIZE};
+use crate::{varint, MAX_CID_SIZE, MIN_CID_SIZE, RESET_TOKEN_SIZE};
 
 /// Parameters governing the core QUIC state machine
 ///
@@ -230,7 +231,7 @@ pub struct ServerConfig<S> {
     pub crypto: S,
 
     /// Private key used to authenticate data included in handshake tokens.
-    pub token_key: crypto::ring::TokenKey,
+    pub token_key: SigningKey,
     /// Whether to require clients to prove ownership of an address before committing resources.
     ///
     /// Introduces an additional round-trip to the handshake to make denial of service attacks more difficult.
@@ -264,7 +265,8 @@ where
             transport: Arc::new(TransportConfig::default()),
             crypto: S::default(),
 
-            token_key: crypto::ring::TokenKey::new(&token_value),
+            // Safe unwrap: key length is as necessary (for SigningKey HmacKey impl)
+            token_key: <SigningKey as HmacKey>::new(&token_value).unwrap(),
             use_stateless_retry: false,
             retry_token_lifetime: 15_000_000,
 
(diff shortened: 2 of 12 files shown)


quinn-proto/src/shared.rs
```diff
@@ -192,15 +193,38 @@ pub struct EndpointConfig {
     ///
     /// Must be persisted across restarts to be useful.
     pub reset_key: SigningKey,
+
+    // SigningKey does not implement Clone, but we'd like our builders to be cloneable
+    reset_key_bytes: [u8; 64],
```
Ralith: A private field breaks things (see CI). If we need this, I think we might as well do away with the `SigningKey` field entirely and instead construct it internally to the connection, which I've been wanting anyway so as to reduce the degree to which other crates are part of our interface (a major semver compat hazard).

djc: This is now solved, at least!

quinn-proto/src/shared.rs
```diff
@@ -181,28 +179,39 @@ impl TransportConfig {
...
+/// `default()` constructor and can be modified using `set_reset_key()`. See the latter method
+/// for more information.
 pub struct EndpointConfig {
     /// Length of connection IDs for the endpoint.
     ///
     /// This must be either 0 or between 4 and 18 inclusive. The length of the local connection IDs
     /// constrains the amount of simultaneous connections the endpoint can maintain. The API user is
     /// responsible for making sure that the pool is large enough to cover the intended usage.
     pub local_cid_len: usize,
+    pub(crate) reset_key: [u8; 64],
```
Ralith: Why can't this be public?

quinn-proto/src/shared.rs
```diff
@@ -232,9 +245,7 @@ where
     ///
     /// Must be set to use TLS 1.3 only.
     pub crypto: S::ServerConfig,
-
-    /// Private key used to authenticate data included in handshake tokens.
-    pub token_key: S::HmacKey,
+    pub(crate) token_key: [u8; 64],
```
Ralith: As above.

quinn-proto/src/crypto.rs
```diff
@@ -124,3 +126,16 @@ pub trait HeaderKeys {
...
+    /// Type of the signatures created by `sign()`
+    type Signature: AsRef<[u8]>;
+
+    /// Method for creating a key
+    fn new(key: &[u8]) -> Self;
+    /// Method for signing a message
+    fn sign(&self, data: &[u8]) -> Self::Signature;
+    /// Method for verifying a message
+    fn verify(&self, data: &[u8], signature: &[u8]) -> Result<(), ()>;
+}
```
Ralith: Could this have a `const KEY_SIZE: usize` (bearing in mind that associated constants aren't usable in array sizes yet, so allocations would still have to be dynamic) or similar so we don't have to inflexibly hardcode 64 bytes everywhere?

djc: I have that mostly done, but then tried to figure out the thing about associated constants as array sizes. Will push it later.

quinn-proto/src/shared.rs
djc: So the idea I had was that since this is private key-like material, we should maybe not make it possible to extract the key material (as it works in *ring*). But maybe this isn't so sensitive...

Ralith: I'm not sure how productive that is; the caller can always pull out the data with unsafe code, and there's a high ergonomic cost. The custom `Debug` impls that don't print it are nice, though.

quinn-proto/src/crypto.rs
Ralith: Yeah, you'll probably have to switch over to `Vec` everywhere, which I think is fine.

quinn-proto/src/shared.rs
djc: What part of the ergonomic cost do you identify as high, exactly? I did a version with `new()` that takes a key as a parameter (and fills the other fields from `default()`), but I found this version easier to explain in documentation.

We should maybe talk about the threat model, in the sense of, what happens when these keys inadvertently leak? (I understand that you could always get them out using unsafe code, but that's sort of besides the point IMO -- defense in depth matters, and this makes it harder for the secrets to escape in any case.)

Ralith: Specifically, the loss of the ability to use functional record update syntax, and the inconsistency in how the one field is written to. A large `new` function would be even worse, I agree there. I also think the added insurance against misuse is extremely marginal; users will almost never have a `Config` struct that they did not create, and when creating one they'll necessarily already have access to the raw bytes.

quinn-proto/src/crypto.rs
djc: So what do you propose slices of an incorrect length do? Panic, return an `Err`? I kind of feel like pending array sizes that actually allow using associated consts and pending an actual other implementation of the trait, keeping the fixed size is actually more ergonomic. (Also arguably I feel like this shouldn't prevent us from landing the rest of these changes which clearly make the API more flexible either way.)

Ralith: > So what do you propose slices of an incorrect length do?

I think we can already return `ConfigError` from every entry point that consumes a config struct; this is not the first case of a config option that can have incoherent values, and this case does not strike me as significantly different.

> I feel like this shouldn't prevent us from landing the rest of these changes

A fixed key size excludes the majority of reasonable implementations. I agree that it's not a hard blocker, but it is a significant wart.

quinn-proto/src/shared.rs
djc: Fixed this part.

If you think the leaking is not a big problem, do you really think the custom `Debug` impls are worth it compared to `derive(Debug)`?

Ralith: I don't have a strong opinion and I'm not as steeped in security idioms as I might be, but my knee-jerk reaction is that a `Debug` dump of a (value containing a) config struct is overwhelmingly the most likely channel for an accidental information leak. Perhaps some users even desire the option to stash `Debug` dumps of config into logs.

quinn-proto/src/crypto.rs
djc: Fixed.

quinn-proto/src/shared.rs
djc: Okay, let's keep it like this.

quinn-proto/src/crypto.rs
```diff
@@ -30,7 +30,7 @@ pub trait Session: Sized {
     /// Type used to represent packet protection keys
     type Keys: Keys + Sized;
     /// Type used to hold configuration for server sessions
-    type ServerConfig: ServerConfig<Self>;
+    type ServerConfig: ServerConfig<Self> + Default;
```
Ralith: The `Default` bound should be placed on `impl Default for ServerConfig` where it's specifically needed, rather than made mandatory.

quinn-h3/examples/h3.rs
```diff
@@ -122,13 +122,11 @@ fn server(
...
-        transport: Arc::new(quinn::TransportConfig {
-            stream_window_uni: 513,
-            ..Default::default()
-        }),
+    let mut server_config = quinn::ServerConfig::default();
+    server_config.transport = Arc::new(quinn::TransportConfig {
+        stream_window_uni: 513,
         ..Default::default()
-    };
+    });
```
Ralith: This change is no longer necessary.

quinn-proto/src/crypto.rs
djc: Right -- this pretty much keeps it more similar to what it was before. Done.

quinn-h3/examples/h3.rs
djc: Oops! Fixed.

PR #373, opened 2019-06-09
Refactoring of HMAC key handling

djc:


quinn-proto/src/crypto.rs
```diff
@@ -124,3 +126,16 @@ pub trait HeaderKeys {
...
+    /// Type of the signatures created by `sign()`
+    type Signature: AsRef<[u8]>;
+
+    /// Method for creating a key
+    fn new(key: &[u8]) -> Self;
+    /// Method for signing a message
+    fn sign(&self, data: &[u8]) -> Self::Signature;
+    /// Method for verifying a message
+    fn verify(&self, data: &[u8], signature: &[u8]) -> Result<(), ()>;
+}
```
Ralith: Should this have a `fn key_size() -> usize;` to obviate the need for restrictive assumptions?

djc:
Folding this into #371.

PR #377, opened 2019-06-14
Use tokio-codec to deframe H3 frames

djc:


quinn-h3/src/proto/frame.rs
```diff
@@ -47,7 +48,7 @@ impl HttpFrame {
         let len = buf.get_var()?;
 
         if buf.remaining() < len as usize {
-            return Err(Error::UnexpectedEnd);
+            return Err(Error::Incomplete(2 + len as usize));
```
stammw: Does this number, `2 + len`  represent the complete size of the message (with header)? If it's the case `2` may not be valid as a header size, as both `len` and `ty` are `varint`s. Maybe passing the expected payload size would be sufficient?

stammw: Ah, It's the minimum valid size, as used in `quinn-h3/src/frame.rs`. Nevermind then.

PR #378, opened 2019-06-17
Optional rustls

djc:
This doesn't yet fix the parameter defaults as discussed, but it does show that we can compile most of the quinn-proto crate without rustls or ring (although it will only run a few tests).

### DIFF
--- a/quinn-proto/src/transport_error.rs
+++ b/quinn-proto/src/transport_error.rs
@@ -56,7 +56,8 @@ impl slog::Value for Error {
 pub struct Code(u16);
 
 impl Code {
-    pub(crate) fn crypto(code: u8) -> Self {
+    /// Create QUIC error code from TLS alert code
+    pub fn crypto(code: u8) -> Self {
         Code(0x100 | u16::from(code))
     }
 }
--- a/quinn-proto/src/transport_error.rs
+++ b/quinn-proto/src/transport_error.rs
@@ -17,23 +17,6 @@ pub struct Error {
     pub reason: String,
 }
 
-impl Error {
-    pub(crate) fn new<T>(code: Code, frame: Option<frame::Type>, reason: T) -> Self
-    where
-        T: Into<String>,
-    {
-        Self {
-            code,
-            frame,
-            reason: reason.into(),
-        }
-    }
-
-    pub(crate) fn crypto(code: u8, reason: String) -> Self {
-        Self::new(Code::crypto(code), None, reason)
-    }
-}
-
 impl fmt::Display for Error {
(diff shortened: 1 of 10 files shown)


quinn-proto/src/transport_error.rs
```diff
@@ -117,10 +116,7 @@ macro_rules! errors {
             fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
                 match self.0 {
                     $($val => f.write_str(stringify!($name)),)*
-                    x if x >= 0x100 && x < 0x200 => match AlertDescription::read_bytes(&[self.0 as u8]) {
-                        Some(desc) => write!(f, "Code::crypto({:?})", desc),
-                        None => write!(f, "Code::crypto({:02x})", self.0 as u8),
-                    },
+                    x if x >= 0x100 && x < 0x200 => write!(f, "Code::crypto({:02x})", self.0 as u8),
```
Ralith: I think this will make error messages such as certificate validation failures much less clear. Perhaps the crypto impl should provide a helper function to format its errors with?

djc: I'm not convinced this is an important issue. As far as I can see, the main issue left where there is no good human-readable error is for `ConnectionClose` frames received from the peer; which already can contain a human-readable description, so adding another one feels silly.

Second, if we were to do this I think the solution would get somewhat ugly, since these error codes get created in contexts where an implementation of the generic type is not easily available and it's non-trivial to pass one in for this `Debug` implementation.

So I feel that this is a reasonable trade-off for pluggable crypto.

One reasonable solution could be to use conditional compilation to improve the `Debug` representation only if `rustls` is enabled. We can always add the additional complexity for a fully generic solution once an actual alternative implementation comes along.

What do you think?

Ralith: > I'm not convinced this is an important issue. As far as I can see, the main issue left where there is no good human-readable error is for ConnectionClose frames received from the peer

Ah, I forgot we're already putting reasonable strings in there ourselves, good call. Of course callers interested in providing good user feedback will in any case want to match on the exact alert themselves, but this should be plenty to make development a pleasant experience.

> Second, if we were to do this I think the solution would get somewhat ugly, since these error codes get created in contexts where an implementation of the generic type is not easily available and it's non-trivial to pass one in for this Debug implementation.

We could use static dispatch to an associated function; no instance should be necessary. Of course, propagating a type parameter through to here would be a bit of a pain. Given the reason strings, I think the current code will do fine and we can revisit in the future if necessary.

PR #380, opened 2019-06-20
Move ALPN tokens around

djc:


quinn-h3/src/lib.rs
```diff
@@ -101,3 +101,6 @@ impl From<proto::headers::Error> for Error {
 fn try_take<T>(item: &mut Option<T>, msg: &'static str) -> Result<T, Error> {
     item.take().ok_or_else(|| Error::Internal(msg))
 }
+
+/// TLS ALPN value for H3
+pub const ALPN_QUIC_H3: &[u8] = b"h3-20";
```
Ralith: Should this just be `ALPN`, which can then be referenced as `quinn_h3::ALPN` without importing?

djc: Done.

PR #381, opened 2019-06-23
Zero rtt testing

Ralith:
Trying to reproduce the erratic 0-RTT failures in a controlled environment. No luck yet, but got some cleanup done along the way.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -533,10 +533,9 @@ where
                     // Might have a key update to discard too
                     self.set_key_discard_timer(now);
                 } else if let Some(ref prev) = self.prev_crypto {
-                    if prev
-                        .update_ack_time
-                        .map_or(false, |x| instant_saturating_sub(now, x) >= self.pto() * 3)
-                    {
+                    if prev.end_packet.map_or(false, |(_, recvd)| {
+                        instant_saturating_sub(now, recvd) >= self.pto() * 3
+                    }) {
                         self.prev_crypto = None;
                     } else {
                         self.set_key_discard_timer(now);
@@ -562,8 +561,8 @@ where
             // gains access to an endpoint's internal state shortly after the handshake completes to
             // decrypt packets sent previously within the same connection.
             now + self.pto() * 3 * 100
-        } else if let Some(time) = self.prev_crypto.as_ref().and_then(|x| x.update_ack_time) {
-            time + self.pto() * 3
+        } else if let Some((_, recvd)) = self.prev_crypto.as_ref().and_then(|x| x.end_packet) {
+            recvd + self.pto() * 3
         } else {
             return;
         };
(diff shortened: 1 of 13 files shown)


quinn/src/connection.rs
```diff
@@ -57,15 +57,13 @@ impl Connecting {
     ///
     /// For incoming connections, a 0.5-RTT connection will always be successfully constructed.
     pub fn into_0rtt(mut self) -> Result<(ConnectionDriver, Connection, IncomingStreams), Self> {
-        if (self.0.as_mut().unwrap().0)
-            .lock()
-            .unwrap()
-            .inner
-            .has_0rtt()
-        {
+        let conn = (self.0.as_mut().unwrap().0).lock().unwrap();
```
djc: Why do we have to drop explicitly here? Maybe add a comment?

quinn/src/tests.rs
```diff
@@ -209,6 +209,131 @@ fn read_after_close() {
     runtime.run().unwrap();
 }
 
+#[test]
+fn zero_rtt() {
+    let mut endpoint = Endpoint::builder();
+
+    let log = logger();
```
djc: Maybe we should extract some of this stuff into a helper method? So much boilerplate...

Ralith:
Comments addressed.

Ralith:
This corrects the erratic 0-RTT failures in the interop tool, which turned out to be caused by a minor bug in locally-initiated key updates, which don't currently occur elsewhere.

quinn-proto/src/connection.rs
```diff
@@ -3166,9 +3174,12 @@ where
...
     crypto: K,
-    end_packet: u64,
-    /// Time at which a packet using the following key phase was received
-    update_ack_time: Option<Instant>,
+    /// The first incoming packet to use the next key phase, and the time at which it was received
+    ///
+    /// Incoming packets with a lower number should be decrypted with `crypto`. `None` indicates
+    /// that we have not yet received a packet using newer keys, which is only possible if the
+    /// update was locally initiated.
+    end_packet: Option<(u64, Instant)>,
```
djc: I find the naming of `end_packet` combined with the description `The first incoming packet (...)` quite confusing. Maybe it should be named `first_packet`?

djc:
Good stuff!

quinn-proto/src/connection.rs
Ralith: Reworded the comment to hopefully be clearer. The current name is correct, because it represents the end of the incoming packet number interval for which the previous keys (which are stored in this structure) should be used.

djc: Hmm, it's still confusing to me. `The incoming packet that triggered or acknowledges [sic] (...)` still seems to be about a first packet of the next range. Maybe the difference is in referencing right-inclusive or right-exclusive ranges? In which case I would suggest this ought to be named `last_packet`.

Ralith: > still seems to be about a first packet of the next range

Yes, that's what the exclusive end-point of the previous range is.

> last_packet

The packet in question is part of the next range, not the previous one.

Ralith: Made another attempt at rephrasing.

Issue #314, opened 2019-04-16
Restrict the quinn-proto API


djc:
In #310, while documenting the quinn-proto API, it came up that we would like to hide more API surface from the quinn-proto. Let's discuss a little bit how/what we can do here.

The good news is that we will be able to iterate on the quinn-proto API but that we could probably keep the quinn API stable while doing so.

stammw:
I don't know how much my concern is connected to this subject, but I figured out I needed to access (read) the `StreamId` inner value  for the `h3` implementation.

djc:
Can you submit a PR that makes it `#[doc(hidden)] pub`?

Ralith:
I think we could make `Timer` opaque, which is desirable as timers may be added/removed in the future. Perhaps we could take it a step further and export a `TimerTable<T>` struct that helps hide the number of timers.

djc:
I think there's nothing left to do here for now.

PR #382, opened 2019-06-23
Assorted cleanup

Ralith:
Partially addresses #314.


Ralith:
Masking `Timer` ended up being a fair bit of boilerplate, but the prospect of being able to add new timers without a breaking change is attractive.

djc:
Yeah, I also like that the `Timer` abstraction is contained in a separate module now. Nice work.

PR #384, opened 2019-06-26
[WIP] Receive directly into uninitialized BytesMut memory

Ralith:
Recalling the visibility of `memset` in profiling results, I am attempting here to improve performance by not initializing memory that will immediately be overwritten, and saving a copy from a temporary stack buffer. This works fine, but surprisingly, the impact on the performance of our single benchmark seems to be below the noise floor. I'm not sure we should merge (even trivial) unsafe optimizations without a clear benefit, so this probably needs more work in developing better benchmarks or optimization of the new code.

### DIFF
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -1,11 +1,10 @@
 use std::collections::VecDeque;
-use std::io;
 use std::net::{SocketAddr, SocketAddrV6};
-use std::str;
 use std::sync::{Arc, Mutex};
 use std::time::Instant;
+use std::{io, slice, str};
 
-use bytes::Bytes;
+use bytes::{Bytes, BytesMut};
 use fnv::FnvHashMap;
 use futures::sync::mpsc;
 use futures::task::{self, Task};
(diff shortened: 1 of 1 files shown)


djc:
The changes that I had in mind seem to have made it ~10% slower. 😢

quinn/src/endpoint.rs
```diff
@@ -192,16 +191,28 @@ pub(crate) struct EndpointInner {
...
+    recv_buf: BytesMut,
 }
 
 impl EndpointInner {
     fn drive_recv(&mut self, now: Instant) -> Result<bool, io::Error> {
-        let mut buf = [0; 64 * 1024];
         let mut recvd = 0;
         loop {
-            match self.socket.poll_recv(&mut buf) {
+            const MAX_DATAGRAM_SIZE: usize = 64 * 1024;
```
TimonPost: I would prefer to move constants out of functions.

TimonPost:
Shouldn't this PR be closed? It seems like it is no longer active.

Ralith:
There's nothing wrong with a PR existing for discussion/future reference.

Ralith:
Closing since #548 addressed the worst of this without any unsafety. It'd be nice to get the copies down further, but they don't look like a bottleneck in their current form.

Issue #356, opened 2019-05-29
H3: Streaming mode for receiving body


stammw:
As opposed to the all-at-once mode that keeps a growable buffer internally and returns the entire response body at once to the caller.  

Implement a body streaming mode API, comparable to [h2::RecvStream](https://docs.rs/h2/0.1.20/h2/struct.RecvStream.html). And use a limited fixed-size buffer whose size is configured for the Connection on construction.

PR #387, opened 2019-07-02
H3 client receive body stream

stammw:
Relates to #356

### DIFF
--- a/quinn-h3/examples/h3.rs
+++ b/quinn-h3/examples/h3.rs
@@ -4,10 +4,10 @@ use std::path::PathBuf;
 use std::rc::Rc;
 use std::sync::Arc;
 use std::time::Instant;
-use std::{fmt, fs, io};
+use std::{fmt, fs, io, mem};
 
 use failure::{bail, format_err, Error, Fail, ResultExt};
-use futures::{Future, Stream};
+use futures::{try_ready, Future, Poll, Stream};
 use http::{header::HeaderValue, method::Method, HeaderMap, Request, Response, StatusCode};
 use slog::{info, o, Drain, Logger};
 use structopt::{self, StructOpt};
@@ -16,8 +16,11 @@ use url::Url;
 
 use quinn::ConnectionDriver as QuicDriver;
 use quinn_h3::{
+    self,
+    body::RecvBodyStream,
     client::Builder as ClientBuilder,
     connection::ConnectionDriver,
+    headers::DecodeHeaders,
     server::{Builder as ServerBuilder, IncomingRequest, RequestReady},
 };
 use quinn_proto::crypto::rustls::{Certificate, CertificateChain, PrivateKey};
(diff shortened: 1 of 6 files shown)


quinn-h3/src/body.rs
```diff
@@ -159,28 +164,66 @@ impl Future for RecvBody {
...
+    type Error = Error;
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        match try_ready!(self.recv.poll()) {
+            Some(HttpFrame::Data(d)) => {
+                Ok(Async::Ready(Some(d.payload)))
+            }
+            Some(HttpFrame::Headers(d)) => {
+                self.trailers = Some(d);
+                Ok(Async::Ready(None))
```
Ralith: If `Headers` can ever be received prior to additional `Data`, this could cause a hung future. I imagine that's forbidden by the h3 spec, but what happens if a peer is ill-behaved?

stammw: From what I understand, the Stream will only appear empty to the user. The additional data will be dropped along with `self.recv`.

I'm not sure how to make polling again for another `DataFrame` after one `HeadersFrame` end up with a correct behavior.

Ralith: You're right, I was confused; this correctly terminates the stream, and assuming the sender handles a dropped receiver gracefully all is well.

stammw:
@Ralith Oh, yeah I'll check what's possible to do with `AsyncRead`. 

One problem I had with `Stream` is that it's not possible (or easy) to take back ownership of it after using `.for_each()` combinator. Because `.into_future()`'s future resolves to `()`, and it makes it impossible to call `.trailers()` afterwards.

Ralith:
Yeah, that's awkward. You could expose it via some shared state, or just punt the problem until async/await makes it much less necessary to thread ownership everywhere.

stammw:
I'd rather leave the `.for_each()` problem for later. Right now the solution looks more or less like `h2`.

I've been checking what's necessary to implement `AsyncRead`, say `Read`. And I think the need of a buffer legitimates using another type. So reading the body could be done in three fashions :
- Everything at once, with `RecvResponse::body()`
- Chunk by chunk, with `RecvResponse::body_stream()`
- With a specific buffer size with `RecvResponse::body_reader()`

But I'd like to do this in another PR. Keeping steps small and easy to manage. (And quick, if I find a way to get more time to spend here:) ) 

@djc Do you want me to change the example so the `Process` future doesn't look like it has debug prints in it? I admit those lines are strongly inspired by [h2 client example](https://github.com/hyperium/h2/blob/master/examples/client.rs#L24)...

quinn-h3/src/headers.rs
```diff
@@ -0,0 +1,47 @@
...
+impl Future for DecodeHeaders {
+    type Item = Header;
+    type Error = crate::Error;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        match self.trailers {
+            None => Err(crate::Error::Internal("trailers none")),
+            Some(ref frame) => {
+                let result = {
+                    let conn = &mut self.conn.h3.lock().unwrap().inner;
```
djc: This whole thing can just be folded into a single expression, which then doesn't need the explicit block?

quinn-h3/src/headers.rs
```diff
@@ -0,0 +1,47 @@
...
+use quinn_proto::StreamId;
+
+use crate::{
+    connection::ConnectionRef,
+    proto::{frame::HeadersFrame, headers::Header},
+    Error,
+};
+
+pub struct DecodeHeaders {
+    trailers: Option<HeadersFrame>,
```
djc: Can you add some comments on why `DecodeHeaders` needs a `trailers` field? It doesn't seem intuitive to me.

quinn-h3/examples/h3.rs
```diff
@@ -268,20 +271,62 @@ fn client(
...
     Ok((endpoint_driver, Box::new(fut)))
 }
 
+enum ProcessState {
+    Body(RecvBodyStream),
+    Trailers(DecodeHeaders),
+    Finished,
+}
+
+struct Process {
```
djc: Where does the term `Process` come from? Is it in the spec? If not, what kind of processing does it represent? As a name it feels very generic and therefore not as helpful as it could be.

quinn-h3/examples/h3.rs
```diff
@@ -268,20 +271,62 @@ fn client(
...
                     }
-                    futures::future::ok(())
+                    .map_err(|e| format_err!("send request: {}", e))
                 })
         });
 
     Ok((endpoint_driver, Box::new(fut)))
 }
 
+enum ProcessState {
```
djc: Please move this below `Process`.

quinn-h3/src/body.rs
```diff
@@ -177,3 +177,53 @@ impl Future for RecvBody {
...
+            trailers: None,
+        }
+    }
+
+    pub fn has_trailers(&self) -> bool {
+        self.trailers.is_some()
+    }
+
+    pub fn trailers(self) -> Option<DecodeHeaders> {
+        match self.trailers {
```
djc: Probably use `.map()` here?

quinn-h3/src/headers.rs
stammw: This is a naming mistake... Renaming to `frame`. Thanks.

quinn-h3/src/headers.rs
stammw: Right. Fixed.

quinn-h3/examples/h3.rs
stammw: I got this renamed to `ProcessResponse`.

quinn-h3/examples/h3.rs
stammw: Done

quinn-h3/src/body.rs
stammw: Done

PR #392, opened 2019-07-06
Replace uses of `mem::uninitialized` with `mem::MaybeUninit`

Ralith:
Corrects some UB and fixes deprecation warnings.


Ralith:
Looks like travis has an outdated stable rust somehow, despite fetching it on demand?

Ralith:
Looks like this can go ahead now.

PR #395, opened 2019-07-12
Implement the unreliable datagram extension

### LINKED ISSUE #207: Implement the unreliable datagram extension
[The unreliable datagram extension](https://datatracker.ietf.org/doc/draft-pauly-quic-datagram/?include_text=1) provides an interesting escape hatch for applications requiring unusual semantics. While streams should support everything from bulk data transfer to realtime games effectively, more niche cases like IP tunneling and VoIP might benefit from lower-level primitives. This extension is simple enough to incur little maintenance cost, and the presence of an escape hatch may reassure prospective users of QUIC's suitability to demanding applications.

The interface should consist of new functions to read/write datagrams. The stream read/write error enums should not be reused, as datagrams have different error semantics: writes may be blocked or may fail due to exceeding the MTU, and reads are infallible.

Ralith:
Fixes #207.

~This extension isn't stable yet, and (partly in the interest of avoiding rebase churn in std-futures branch) I haven't yet implemented the high level API, but this is the bulk of the work, and the API is ready to receive feedback.~

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -12,7 +12,7 @@ use slog::Logger;
 
 use crate::coding::BufMutExt;
 use crate::crypto::{self, HeaderKeys, Keys};
-use crate::frame::{Close, FrameStruct};
+use crate::frame::{Close, Datagram, FrameStruct};
 use crate::packet::{
     Header, LongType, Packet, PacketNumber, PartialDecode, SpaceId, LONG_RESERVED_BITS,
     SHORT_RESERVED_BITS,
@@ -152,6 +152,8 @@ where
     streams: Streams,
     /// Surplus remote CIDs for future use on new paths
     rem_cids: Vec<IssuedCid>,
+    /// State of the unreliable datagram extension
+    datagrams: DatagramState,
 }
 
 impl<S> Connection<S>
@@ -255,6 +257,7 @@ where
             total_sent: 0,
 
             streams: Streams::new(side, config.stream_window_uni, config.stream_window_bidi),
+            datagrams: DatagramState::new(),
             config,
             rem_cids: Vec::new(),
             rng,
@@ -351,6 +354,7 @@ where
         if ack.largest >= self.space(space).next_packet_number {
             return Err(TransportError::PROTOCOL_VIOLATION("unsent packet acked"));
         }
+        let was_congested = self.congestion_blocked();
         let was_blocked = self.blocked();
         let new_largest = {
             let space = self.space_mut(space);
@@ -439,6 +443,12 @@ where
             for stream in self.blocked_streams.drain() {
                 self.events.push_back(Event::StreamWritable { stream });
             }
+        }
+        if was_congested
+            && !self.congestion_blocked()
+            && mem::replace(&mut self.datagrams.send_blocked, false)
+        {
+            self.events.push_back(Event::DatagramSendUnblocked);
         }
         Ok(())
     }
@@ -1949,6 +1959,28 @@ where
                     trace!(self.log, "got new token");
                     // TODO: Cache, or perhaps forward to user?
                 }
+                Frame::Datagram(datagram) => {
+                    let window = match self.config.datagram_window {
+                        None => {
+                            return Err(TransportError::PROTOCOL_VIOLATION(
+                                "unexpected DATAGRAM frame",
+                            ));
+                        }
+                        Some(x) => x,
+                    };
+                    if datagram.data.len() > window {
+                        return Err(TransportError::PROTOCOL_VIOLATION("oversized datagram"));
+                    }
+                    if self.datagrams.recv_buffered == 0 {
+                        self.events.push_back(Event::DatagramReceived);
+                    }
+                    while datagram.data.len() + self.datagrams.recv_buffered > window {
+                        debug!(self.log, "dropping stale datagram");
+                        self.recv_datagram();
+                    }
+                    self.datagrams.recv_buffered += datagram.data.len();
+                    self.datagrams.incoming.push_back(datagram);
+                }
             }
         }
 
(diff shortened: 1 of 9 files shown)


quinn-proto/src/connection.rs
```diff
@@ -2847,6 +2888,57 @@ where
...
+    /// Compute the maximum size of datagrams that may be sent, if supported on this connection
+    ///
+    /// This may change over the lifetime of a connection according to variation in the MTU
+    /// estimate, but will never be below the smaller of `MIN_DATAGRAM_MTU` and the limit specified
+    /// by the peer.
+    pub fn datagram_mtu(&self) -> Option<usize> {
+        self.config.datagram_window?;
+        let limit = self.params.max_datagram_frame_size?.into_inner();
+        Some(limit.min(self.max_payload_size() as u64 - Datagram::SIZE_BOUND as u64) as usize)
+    }
```
Ralith: I'm particularly interested in feedback on this function, and the related constant. It strikes me as potentially confusing, but the only other idea I have so far is exposing the components--peer limit and path MTU--separately, which is error-prone in that users might neglect to check both, and doesn't have an obvious functional benefit.

djc: Can you link to relevant part(s) of the draft?

djc: I don't understand what the call to `into_inner()` is doing. As I understand it, `self.params.max_datagram_frame_size?` is `usize`, so what's the `into_inner()` doing here? Also let's inline `max_payload_size()`, since it's a single-use function (and the name doesn't suggest it is only relevant for datagrams). Also, "will never be below ... the limit specified by the peer" doesn't seem very meaningful or helpful since the limit specified by the peer, as I understand it, might be 1? Also, the name here ending in `_mtu` again suggests an MTU unit, whereas it's actually datagram payload size?

quinn-proto/src/connection.rs
```diff
@@ -3378,3 +3475,36 @@ const ACK_DELAY_EXPONENT: u8 = 3;
...
+    TooLarge,
+    /// The peer does not support receiving datagram frames
+    #[error(display = "datagrams not supported by peer")]
+    UnsupportedByPeer,
+    /// Datagram support is disabled locally
+    #[error(display = "datagram support disabled")]
+    Disabled,
+}
+
+/// The largest unreliable datagram size that can be transmitted to a willing peer over any path
```
djc: I found the comment here quite confusing. The comment starts with "the largest" while the constant is named `MIN_`, so I'm already scratching my head. The `_MTU` suffix here seems to suggest that we're dealing with MTUs but I think this actually something like the payload size for a datagram frame that just fits in the QUIC minimum MTU? In any case it is not the minimum, because that is imposed by the transport parameter.

Ralith:
For some reason github won't let me respond to your first comment directly.

> I don't understand what the call to into_inner() is doing. As I understand it, self.params.max_datagram_frame_size? is usize, so what's the into_inner() doing here?

It's actually `pub max_datagram_frame_size: Option<VarInt>`. I'm toying with the idea of adopting this convention more broadly to make it harder to incorrectly encode/decode a value that's always used as varint.

> Also let's inline max_payload_size(), since it's a single-use function (and the name doesn't suggest it is only relevant for datagrams).

It's not specialized to datagrams, strictly, it's a judgement about the room for frames of any kind. Still, no harm in inlining it.

> "will never be below ... the limit specified by the peer" doesn't seem very meaningful or helpful since the limit specified by the peer, as I understand it, might be 1?

My intention was to express that the value usually has a reasonable lower bound, unless the peer requires otherwise. I'll try rewording it, especially since it doesn't make sense without the constant anyway.

Ralith:
Implemented the high-level API and added support for waiting on congestion.

quinn-proto/src/connection.rs
```diff
@@ -111,13 +111,18 @@ where
     first_1rtt_sent: Option<u64>,
     /// Sequence number of the first remote CID that we haven't been asked to retire
     first_unretired_cid: u64,
+    in_datagrams: VecDeque<Datagram>,
+    datagram_bytes: usize,
```
djc: Can you add a comment about what `datagram_bytes` is, please?

quinn-proto/src/connection.rs
```diff
@@ -111,13 +111,18 @@ where
     first_1rtt_sent: Option<u64>,
     /// Sequence number of the first remote CID that we haven't been asked to retire
     first_unretired_cid: u64,
+    in_datagrams: VecDeque<Datagram>,
```
djc: How do you feel about adding all the datagram state into a separate `DatagramState` (to be bikeshedded) type that gets included in the connection? Presumably the borrowck situation would not be worse than it already is.

quinn/src/broadcast.rs
```diff
@@ -0,0 +1,40 @@
+use std::task::{Context, Waker};
```
djc: I feel like adding this needs a bunch more justification to help me understand what problem we're trying to solve here, why that particular threshold of problems is crossed with the addition of datagrams. I also wonder what benchmarking you have done to back up the need for this mechanism?

quinn-proto/src/connection.rs
Ralith: Seems like a reasonable measure to keep `Connection`'s field count from growing too fast.

quinn/src/broadcast.rs
Ralith: This solves a resource-boundedness problem, not a performance one, and it's a problem we have otherwise completely avoided by arranging for futures to otherwise have at most one task waiting on them at a time. The alternative would be to add a `!Clone` `DatagramSender` handle to `NewConnection` with a `fn send_datagram(&mut self, ...)` to ensure that only one sender can exist at a time; that felt a bit like complicating the interface for the implementation's benefit to me.

Ralith: Although, `DatagramSender` *would* let us have a neater API along the lines of

```rust
impl DatagramSender {
    async fn wait_ready(&mut self);
    fn send(&mut self, data: Bytes) -> Result<(), Blocked + ConnectionError>
}
```

PR #402, opened 2019-07-21
H3: body AsyncRead implementation

stammw:
I'm not sure about the RecvBodyReader::buf_{pop,push}() method names.

And about their implementation, I tried to limit the number of copies and allocations by storing the DataFrame(Bytes)'s buffer only. It makes read() returned size possibly very small compared to the passed buffer size. I'd like some feedback on this as I'm not sure about this choice.

The only example has been updated to exercise this new API.

### DIFF
--- a/quinn-h3/examples/h3.rs
+++ b/quinn-h3/examples/h3.rs
@@ -4,10 +4,10 @@ use std::path::PathBuf;
 use std::rc::Rc;
 use std::sync::Arc;
 use std::time::Instant;
-use std::{fmt, fs, io, mem};
+use std::{fmt, fs, io};
 
 use failure::{bail, format_err, Error, Fail, ResultExt};
-use futures::{try_ready, Future, Poll, Stream};
+use futures::{future::Either, Future, Stream};
 use http::{header::HeaderValue, method::Method, HeaderMap, Request, Response, StatusCode};
 use slog::{info, o, Drain, Logger};
 use structopt::{self, StructOpt};
@@ -17,10 +17,8 @@ use url::Url;
 use quinn::ConnectionDriver as QuicDriver;
 use quinn_h3::{
     self,
-    body::RecvBodyStream,
     client::Builder as ClientBuilder,
     connection::ConnectionDriver,
-    headers::DecodeHeaders,
     server::{Builder as ServerBuilder, IncomingRequest, RequestReady},
 };
 use quinn_proto::crypto::rustls::{Certificate, CertificateChain, PrivateKey};
@@ -204,9 +202,10 @@ fn handle_request(request: RequestReady) -> impl Future<Item = (), Error = Error
                 println!("server received trailers: {:?}", trailers);
             }
 
+            let body = "r".repeat(1024 * 800);
             let response = Response::builder()
                 .status(StatusCode::OK)
-                .body("response body")
+                .body(body.as_str())
                 .expect("failed to build response");
 
             let mut trailer = HeaderMap::with_capacity(2);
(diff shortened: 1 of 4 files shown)


quinn-h3/src/frame.rs
```diff
@@ -48,6 +48,7 @@ impl Decoder for FrameDecoder {
             Err(e) => Err(e)?,
             Ok(frame) => {
                 src.advance(pos);
+                self.expected = None;
```
djc: Oops, good catch!

quinn-h3/src/client.rs
```diff
@@ -318,4 +318,8 @@ impl RecvResponse {
     pub fn body_stream(self) -> RecvBodyStream {
         RecvBodyStream::new(self.recv, self.conn, self.stream_id)
     }
+
+    pub fn body_reader(self) -> RecvBodyReader {
```
djc: So can you give me some context on what the difference between `body_stream()` and `body_reader()` is? (Ideally, write the docstring for `body_reader()` -- and maybe also `body_stream()` so we have that documentation, which we're going to need anyway for this public API?)

quinn-h3/src/body.rs
```diff
@@ -221,3 +225,86 @@ impl Stream for RecvBodyStream {
...
+            recv,
+            conn,
+            stream_id,
+            trailers: None,
+            buf: None,
+        }
+    }
+
+    fn buf_pop(&mut self, buf: &mut [u8]) -> usize {
+        match self.buf {
```
djc: This looks like it could benefit from using `map_or()`, but maybe hold on that until I get my context straight and we figure out if `Option<Bytes>` is the best type for this data structure.

quinn-h3/examples/h3.rs
```diff
@@ -270,64 +269,32 @@ fn client(
...
-                    .map_err(|e| format_err!("send request: {}", e))
+                    let buf = Vec::with_capacity(1024 * 10); // 10K
+                    tokio_io::io::read_to_end(response.body_reader(), buf)
+                        .map_err(|e| format_err!("receive response failed: {}", e))
+                        .and_then(|(reader, data)| {
+                            println!("received body len = {}", data.len());
+                            if let Some(decode_trailers) = reader.trailers() {
+                                return Either::A(
+                                    decode_trailers
+                                        .map_err(|e| format_err!("decod trailers failed: {}", e))
```
djc: Typo.

quinn-h3/examples/h3.rs
```diff
@@ -270,64 +269,32 @@ fn client(
...
+                                        .and_then(|trailers| {
+                                            println!(
+                                                "received  trailers: {:?}",
```
djc: Extraneous space.

djc:
Sorry this took me a while to get to!

quinn-h3/src/client.rs
stammw: `RecvBodyReader` is an object that implements `AsyncRead`, whereas `RecvBodyStream` can only poll `HttpFrame::Data` by `HttpFrame::Data`. They are not implemented on the same struct because an `io::Read` implementation needs to have an intermediate buffer, in case the slice passed to `read()` is smaller than the received data frame. 

Maybe some code can be dedpuplicated between those two... actually.

stammw: And maybe `Recv` and `Reader` are redundant in the name.

quinn-h3/src/body.rs
stammw: Ah, missed that. 

`Option<Bytes>` is some only when the last polled payload hasn't been fully consumed (i.e the client buffer is smaller than the payload size). 

What I try to achieve here is to leverage the `Bytes::split_off()` behavior that only change a bunch of indexes from the original `Bytes` and refcounts the produced `Bytes` to keep data into a contiguous allocated buffer.

In the case where user's buffer is larger than the payload. `read()` will only pass this, and will not try to poll for another data frame. I thought it would maybe save some copies / allocations.

quinn-h3/src/client.rs
djc: So what is the use case for polling `HttpFrame::Data` by `HttpFrame::Data` vs the use case for `RecvBodyReader`? I do think `BodyReader` is probably a better name.

quinn-h3/src/body.rs
djc: It sounds to me like `buf_pop()` could actually just be called `read()`?

quinn-h3/src/client.rs
stammw: The intention is to offer a similar API to most `tokio`-aware libs. 

`BodyReader` is here to make it possible to use tokio convenience functions, as `tokio::io::read_to_end()` or `tokio::io::copy()`.

`RecvBodyStream` offers a `Stream` implementation on the data, and is the minimal copy-overhead API. Though it will probably change to be able to yield only parts of big `HttpFrame::Data` in the future.

quinn-h3/src/body.rs
stammw: Yeah, but as it would probably conflict with the `std::io::Read` impl, I'll rename it `buf_read()`. And maybe `buf_put()` would also be a better name for `buf_push()`.

stammw:
I believe everything has been addressed here?

I'd like to keep the public API documentation for later as changes may happen with future work.

Another [PR is ready](https://github.com/djc/quinn/tree/h3-streamed-body-server) to be opened behind this one.

PR #403, opened 2019-07-21
Azure badge + remove Travis config

djc:


djc:
Note that we're not currently running the ignored tests on Azure. See https://github.com/crate-ci/azure-pipelines/issues/41.

PR #405, opened 2019-07-22
Fix 0-RTT with non-default ciphersuite

Ralith:
Must only be merged after https://github.com/ctz/rustls/pull/274 is released, as otherwise 0-RTT cipher suites won't be available and 0-RTT capable clients will panic.

Fixes 0-RTT interop with quant.

### DIFF
--- a/quinn-proto/src/crypto/rustls.rs
+++ b/quinn-proto/src/crypto/rustls.rs
@@ -40,8 +40,17 @@ impl crypto::Session for TlsSession {
     }
 
     fn early_crypto(&self) -> Option<Self::Keys> {
-        self.get_early_secret()
-            .map(|secret| Crypto::new_0rtt(secret))
+        let secret = self.get_early_secret()?;
+        // If an early secret is known, TLS guarantees it's associated with a resumption
+        // ciphersuite,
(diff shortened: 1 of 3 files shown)


quinn-proto/src/crypto/rustls.rs
```diff
@@ -40,8 +40,15 @@ impl crypto::Session for TlsSession {
     }
 
     fn early_crypto(&self) -> Option<Self::Keys> {
-        self.get_early_secret()
-            .map(|secret| Crypto::new_0rtt(secret))
+        let secret = self.get_early_secret()?;
+        let suite = self.get_negotiated_ciphersuite().unwrap();
```
djc: Do we need this `unwrap()`? In what condition could it occur?

Ralith:
Success here should be implied by the successful acquisition of 0-RTT keys,
subject to the rustls bug getting fixed. I'll make that clearer.

quinn-proto/src/crypto/rustls.rs
Ralith: Explanatory comment added.

djc:
These are now included in #414, so we can close this.

PR #417, opened 2019-08-08
Opaque NewConnection struct

Ralith:
Based on discussion in gitter (see also https://github.com/djc/quinn/issues/207#issuecomment-518042632). I'm not sure how I feel about this. Forwards-compatibility is definitely nice. Panicing on drop without the driver having been taken is interesting, but I worry that it will convert harmless behavior into crashes, e.g. in the quinn-h3 client/server changes you can see where I had to carefully avoid that. Acquiring the components of a connection with one-time accessors also seems less obvious than destructuring them out of public fields, no matter how well-documented the accessors are.


Ralith:
Closing in favor of #421.

Issue #363, opened 2019-06-03
H3: Server early response


stammw:
The client should be able to detect a server response before having send all of the request body. As stated in [H3 - 4.1. HTTP Message Exchanges ](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#request-response): 

> A server can send a complete response prior to the client sending an entire request if the response does not depend on any portion of the request that has not been sent and received. When this is true, a server MAY request that the client abort transmission of a request without error by triggering a QUIC STOP_SENDING frame with error code HTTP_EARLY_RESPONSE, sending a complete response, and cleanly closing its stream. Clients MUST NOT discard complete responses as a result of having their request terminated abruptly, though clients can always discard responses at their discretion for other reasons.

stammw:
I'll rewrite this with an access to send struct as soon as headers are decoded then.

stammw:
Closed via #470

Issue #357, opened 2019-05-29
H3: Streaming mode for sending body


stammw:
Implement a streaming API for sending a body, like [h2::SendStream](https://docs.rs/h2/0.1.20/h2/struct.SendStream.html).

PR #419, opened 2019-08-13
H3 streamed body server

stammw:
This PR makes it possible to use the `body.rs` reading API in server code. Reading a request body from the
server code has something particular compared to reading body from the client: it needs to cary a `SendStream`
to the next step. So response can be sent back.

I added a type parameter that receives this `SendStream`, so APIs can use the same implementation underneath,
while having a slightly different chain of steps:

```
Client:
RecvResponse.body_reader() -> BodyReader<()> -> DecodeHeader<()>

Server:
RequestReady.body_reader() -> BodyReader<SendStream> -> DecodeHeader<Sendstream>.into_sender() -> Sender.send(response)
```

I created a trait to extract this `SendStream` and turn it into a `Sender`:

```rust
impl IntoSender for RecvBodyStream<SendStream> {
    fn into_sender(self) -> Sender {
        Sender::new(self.conn, self.stream_id, self.inner)
    }
}
```

The `Sender` type provides a method chaining API to send the response:

```rust
sender
    .with_trailers(trailers)
    .send(response) // -> server::SendResponse Future
```

Concerning the API parts that involve `Future` implementation (`RecvBody` and `DecodeHeaders`), I made
different impl blocks for each type parameter values:

```
impl Future for DecodeHeaders<()> {
    type Item = Header;
    ...
}

impl Future for DecodeHeaders<SendStream> {
    type Item = (Header, Sender);
    ...
}
```

Notice `DecodeHeaders<SendStream>` resolves with a `Sender`.

The idea behind all that is also to offer a Type-level API usage inforcement. So compilation
ganrantees it'll work as expected.

This should be the last PR related to #356.

### DIFF
--- a/quinn-h3/src/body.rs
+++ b/quinn-h3/src/body.rs
@@ -157,14 +157,6 @@ impl ReadToEnd {
             state: RecvBodyState::Receiving(BytesMut::with_capacity(capacity)),
         }
     }
-
-    pub fn into_reader(self) -> BodyReader {
-        BodyReader::new(self.recv, self.conn, self.stream_id)
-    }
-
-    pub fn into_stream(self) -> RecvBodyStream {
-        RecvBodyStream::new(self.recv, self.conn, self.stream_id)
-    }
 }
 
 enum RecvBodyState {
--- a/quinn-h3/src/body.rs
+++ b/quinn-h3/src/body.rs
@@ -131,6 +131,14 @@ impl RecvBody {
 
         self
     }
+
+    pub fn into_reader(self) -> BodyReader {
+        BodyReader::new(self.recv, self.conn, self.stream_id)
+    }
+
+    pub fn into_stream(self) -> RecvBodyStream {
+        RecvBodyStream::new(self.recv, self.conn, self.stream_id)
+    }
 }
 
 const RECV_BODY_MAX_SIZE_DEFAULT: usize = 20 * 1024 * 1024; // 20 MB
--- a/quinn-h3/src/body.rs
+++ b/quinn-h3/src/body.rs
@@ -97,6 +97,34 @@ impl Future for SendBody {
     }
 }
 
+pub struct Receiver {
+    recv: FrameStream,
+    conn: ConnectionRef,
+    stream_id: StreamId,
+}
+
+impl Receiver {
+    pub(crate) fn new(recv: FrameStream, conn: ConnectionRef, stream_id: StreamId) -> Self {
+        Self {
+            conn,
+            stream_id,
+            recv,
+        }
+    }
+
+    pub fn into_future(self) -> RecvBody {
+        RecvBody::new(self.recv, self.conn, self.stream_id)
+    }
+
+    pub fn into_reader(self) -> BodyReader {
+        BodyReader::new(self.recv, self.conn, self.stream_id)
+    }
+
+    pub fn into_stream(self) -> RecvBodyStream {
+        RecvBodyStream::new(self.recv, self.conn, self.stream_id)
+    }
+}
+
 pub struct RecvBody {
     state: RecvBodyState,
     capacity: usize,
--- a/quinn-h3/src/body.rs
+++ b/quinn-h3/src/body.rs
@@ -144,7 +144,7 @@ enum RecvBodyState {
 }
 
 impl Future for RecvBody {
-    type Item = (Bytes, Option<HeaderMap>);
+    type Item = (Option<Bytes>, Option<HeaderMap>);
     type Error = crate::Error;
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
@@ -172,7 +172,10 @@ impl Future for RecvBody {
                     }
                     None => {
                         let body = match mem::replace(&mut self.state, RecvBodyState::Finished) {
-                            RecvBodyState::Receiving(b) => b.into(),
+                            RecvBodyState::Receiving(b) => match b.len() {
+                                0 => None,
+                                _ => Some(b.into()),
+                            },
                             _ => unreachable!(),
                         };
                         return Ok(Async::Ready((body, None)));
@@ -183,7 +186,7 @@ impl Future for RecvBody {
                     let trailer = try_ready!(trailer.poll());
                     self.state = RecvBodyState::Finished;
                     return Ok(Async::Ready((
-                        try_take(&mut self.body, "body absent")?,
+                        Some(try_take(&mut self.body, "body absent")?),
                         Some(trailer.into_fields()),
                     )));
                 }
--- a/quinn-h3/src/body.rs
+++ b/quinn-h3/src/body.rs
@@ -99,6 +99,7 @@ impl Future for SendBody {
 
 pub struct RecvBody {
     state: RecvBodyState,
+    capacity: usize,
     max_size: usize,
     body: Option<Bytes>,
     recv: FrameStream,
@@ -107,29 +108,36 @@ pub struct RecvBody {
 }
 
 impl RecvBody {
-    pub(crate) fn with_capacity(
-        recv: FrameStream,
-        capacity: usize,
-        max_size: usize,
-        conn: ConnectionRef,
-        stream_id: StreamId,
-    ) -> Self {
-        if capacity < 1 {
-            panic!("capacity cannot be 0");
-        }
-
+    pub(crate) fn new(recv: FrameStream, conn: ConnectionRef, stream_id: StreamId) -> Self {
         Self {
-            max_size,
             conn,
             stream_id,
-            state: RecvBodyState::Receiving(BytesMut::with_capacity(capacity)),
-            body: None,
             recv,
+            body: None,
+            max_size: RECV_BODY_MAX_SIZE_DEFAULT,
+            capacity: RECV_BODY_CAPACITY_DEFAULT,
+            state: RecvBodyState::Initial,
         }
+    }
+
+    pub fn with_capacity(mut self, capacity: usize, max_size: usize) -> Self {
+        match &self.state {
+            RecvBodyState::Initial => (),
+            _ => panic!("cannot change capacity once polled"),
+        }
+
+        self.max_size = max_size;
+        self.capacity = capacity;
+
+        self
     }
 }
 
+const RECV_BODY_MAX_SIZE_DEFAULT: usize = 20 * 1024 * 1024; // 20 MB
+const RECV_BODY_CAPACITY_DEFAULT: usize = 1024 * 1024;
+
 enum RecvBodyState {
+    Initial,
     Receiving(BytesMut),
     Decoding(DecodeHeaders),
     Finished,
@@ -142,6 +150,9 @@ impl Future for RecvBody {
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         loop {
             match self.state {
+                RecvBodyState::Initial => {
+                    self.state = RecvBodyState::Receiving(BytesMut::with_capacity(self.capacity))
+                }
                 RecvBodyState::Receiving(ref mut body) => match try_ready!(self.recv.poll()) {
                     Some(HttpFrame::Data(d)) => {
                         if d.payload.len() + body.len() >= self.max_size {
--- a/quinn-h3/src/body.rs
+++ b/quinn-h3/src/body.rs
@@ -1,8 +1,7 @@
 use std::{
-    cmp,
+    cmp, fmt,
     io::{self, ErrorKind},
     mem,
-    fmt,
 };
 
 use bytes::{Bytes, BytesMut};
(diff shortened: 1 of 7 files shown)


Ralith:
It should be possible for a server to send a response before the request body has been fully read. For example, you could imagine a request that translates incoming data and writes it back out with only a fixed-size buffer, or even an unbounded exchange of framed messages within a single ordered bidirectional stream. Does this API permit that?

stammw:
I guess so. 

`BodyReader<SendStream>` also implements `IntoSender`. One user implementation could choose to stop the reading, pass a flag that would trigger a different reponse than if the body was entirely read. Trailers wouldn't be available though, but I don't think this is something that `h3` permits.

Same principle for `RecvBodyStream<SendStream>`.

I'd like to write one examples for each case but it would mean to copy / paste all the setup code. And I didn't figure out how to have helpers common to all examples.

Or maybe I should take the courage to exercise all this in unit tests.

Nevertheless, there is an issue for early response: #363.

Ralith:
> `BodyReader<SendStream>` also implements `IntoSender`

Unless I'm confused, that makes it impossible to begin responding while still reading the body concurrently, which breaks both example use-cases I outlined.

stammw:
Oh, I totally missed that point. This is not what I understood from the spec at least. 

Did you read about this streaming behavior in the draft, or in slack discussion?

I've read https://quicwg.org/base-drafts/draft-ietf-quic-http.html#http-request-lifecycle again and it seems to me that server response start always means the request body is not of interest anymore. As long as it's not a CONNECT method.

The current code will drop RecvStream on response start. Which will result in an error on the client body writing side. I guess I will need to issue a STOP_SENDING on this event.

Ralith:
Maybe I'm missing something, but my reading is that the only requirement is that the client must not discard a response simply because its request was not read in full; I see absolutely no requirement that the server must not respond until the request has been either deemed irrelevant or fully read. That's an application-level concern, so it would be a strange thing to specify.

stammw:
I rewrote everything from master. 

Now, each future resolving to an entry point for receiving a body returns a type that lets the user choose how he wants to treat the body:

```rust
impl Future for SendRequest {
    type Item = (Response<()>, Receiver);
    ...
}
```

and 

```rust
impl Future for RecvRequest {
    type Item = (Request<()>, Receiver, Sender);
    ...
```

Then body can be read in three fashions: 

```rust

impl Receiver {
    pub fn into_future(self) -> RecvBody; // All at once
    pub fn into_reader(self) -> BodyReader; // AsyncRead implementation
    pub fn into_stream(self) -> RecvBodyStream; // futures::Stream implementation (the most "performant")
}
```

Sending response is now totally independent from receiving the body.

Ralith:
That sounds good; thanks for refactoring things! I'll try to take a closer look tonight.

quinn-h3/src/client.rs
```diff
@@ -173,10 +173,24 @@ impl SendRequest {
...
     }
+
+    fn build_response(header: Header) -> Result<Response<()>, Error> {
+        let (status, headers) = header.into_response_parts()?;
+        let mut response = Response::builder();
+        response.status(status);
+        response.version(http::version::Version::HTTP_3);
+        *response
+            .headers_mut()
+            .ok_or_else(|| Error::peer("invalid response"))? = headers;
```
Ralith: When can this error arise?

quinn-h3/src/server.rs
```diff
@@ -154,10 +153,26 @@ impl RecvRequest {
...
+
+    fn build_request(headers: Header) -> Result<Request<()>, Error> {
+        let (method, uri, headers) = headers.into_request_parts()?;
+        let mut request = Request::builder();
+        request.method(method);
+        request.uri(uri);
+        request.version(http::version::Version::HTTP_3);
+        match request.headers_mut() {
+            Some(h) => *h = headers,
+            None => return Err(Error::peer("invalid header")),
```
Ralith: When can this error arise?

stammw:
> would it make sense to place the futures in the generic of those values

Yes, this is what I intended a couple of month ago, but figured out it was not that ergonomic having to call `.into_parts()`. I guess I should try to use those type parameters now that I can see more clearly.

quinn-h3/src/body.rs
```diff
@@ -107,41 +136,59 @@ pub struct RecvBody {
...
+        self
+    }
+
+    pub fn into_reader(self) -> BodyReader {
+        BodyReader::new(self.recv, self.conn, self.stream_id)
+    }
+
+    pub fn into_stream(self) -> RecvBodyStream {
+        RecvBodyStream::new(self.recv, self.conn, self.stream_id)
     }
```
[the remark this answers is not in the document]
stammw: There is no need for these. I missed removing them when writing `Receiver`.

quinn-h3/src/body.rs
```diff
@@ -97,8 +97,37 @@ impl Future for SendBody {
...
+impl Receiver {
+    pub(crate) fn new(recv: FrameStream, conn: ConnectionRef, stream_id: StreamId) -> Self {
+        Self {
+            conn,
+            stream_id,
+            recv,
+        }
+    }
+
+    pub fn into_future(self) -> RecvBody {
```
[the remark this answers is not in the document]
stammw: Yes it's a lot clearer. I also renamed `RecvBody` to `ReadToEnd`, and reused the name for reaming `Receiver` to `RecvBody`.  I think the naming is more natural that way, especially with `Request<RecvBody>` changes.

quinn-h3/src/body.rs
```diff
@@ -107,41 +136,59 @@ pub struct RecvBody {
...
-    type Item = (Bytes, Option<HeaderMap>);
+    type Item = (Option<Bytes>, Option<HeaderMap>);
     type Error = crate::Error;
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         loop {
             match self.state {
+                RecvBodyState::Initial => {
+                    self.state = RecvBodyState::Receiving(BytesMut::with_capacity(self.capacity))
+                }
```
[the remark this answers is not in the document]
stammw: There is no benefit at all. I forgot to remove that when writing `Receiver`. I changed the implementation to ask for those param when the user call `RecvBody::read_to_end()`. Similarly to `quinn::stream::ReadToEnd`.

quinn-h3/src/client.rs
stammw: The [doc](https://docs.rs/http/0.1.18/http/request/struct.Builder.html#method.headers_mut) isn't clear about that, so I preferred not to `panic!` and didn't bother to write a more specific error.

I guess the error cases are: invalid method or uri.  These values are received from the server.

quinn-h3/src/server.rs
stammw: The [doc](https://docs.rs/http/0.1.18/http/request/struct.Builder.html#method.headers_mut) isn't clear about that, so I preferred not to `panic!`, and didn't bother to write a more specific error.

I guess the error could be triggered by an invalid status code (which is still transferred in ascii).  These values are received from the client.

stammw:
I got this rewritten to use `Response<RecvBody>` and `Request<RecvBody>`. It was totally straightforward. This makes me think it's a fairly good API design solution after all.

My problems doing like this surely came from the fact I didn't identified the `RecvBody` struct (former `Receiver`). Another good thing is that I'll maybe achieve some kind of symmetry for the streaming-send API for #357.

Next step will be to implement a partial `HttpFrame::Data` decoding, in case the peer sends large ones.

quinn-h3/src/server.rs
stammw: Maybe I should use ok_or_else though.

stammw:
Thanks for the review!

quinn-h3/src/client.rs
Ralith: When you call `headers_mut` you haven't yet passed in the headers, so they can't possibly be the cause of an error.

stammw: You're right, this API is not meant to be used this way. I got both functions rewritten. They now emit the error on `.body()` call (the builder end), and then set headers in the real instance. Thanks for catching this.

quinn-h3/src/client.rs
```diff
@@ -173,10 +173,30 @@ impl SendRequest {
...
+            .status(status)
+            .version(http::version::Version::HTTP_3)
+            .body(RecvBody::new(
+                try_take(&mut self.recv, "recv is none")?,
+                self.conn.clone(),
+                try_take(&mut self.stream_id, "stream is none")?,
+            ));
+
+        match response {
+            Err(e) => Err(Error::Peer(format!("invalid response: {:?}", e))),
```
Ralith: I'm still not clear on when this can possibly occur. `status` is already a `StatusCode` so parsing that won't fail, setting the HTTP version and body should be infallible, and the headers still aren't being checked.

stammw: All header fields are checked: they are stored in `http` crate types at the time of decoding, every headers pass through [this validation](https://github.com/djc/quinn/blob/h3-streamed-body-server/quinn-h3/src/proto/headers.rs#L168), `http::HeaderName` and `http::HeaderValue` being the implementation I rely on for non-special fields. 

As this is the case for all other members of `Header`, I guess `.unwrap()`ing the `Request` and `Response` builders is not dangerous at all (only if the `quinn_h3::proto::headers` module do not change).

I just amended my last commit.

stammw: `quinn_h3::proto::header::Field` has some minor problems tha I'mgoing to fix by the way.

Ralith: To be clear, by "not being checked" I meant by the calls made in this function, i.e. anything that could possibly have caused that error state to arise. Seems like that's been addressed.

stammw:
@Ralith Is everything OK for merging then? I'd like to keep the "response before reading headers" feature for #363.

@djc You might want to have a look on this as well? (In case you missed that, I don't want to rush anything)

djc:
I would like to, will prioritize it.

quinn-h3/src/body.rs
```diff
@@ -107,29 +108,44 @@ pub struct RecvBody {
...
+            state: RecvBodyState::Initial,
         }
     }
+
+    pub fn with_capacity(
+        mut self,
+        capacity: usize,
+        max_size: usize,
+    ) -> Self {
+        match &self.state {
```
djc: Since we don't need the actual value of `self.state` here, this would be more idiomatic (and slightly more compact) as an `if let RecvBodyState::Initial =`  block.

quinn-h3/src/client.rs
```diff
@@ -306,11 +306,7 @@ impl RecvResponse {
     }
 
     pub fn body(self) -> RecvBody {
-        RecvBody::new(
-            self.recv,
-            self.conn.clone(),
-            self.stream_id,
-        )
+        RecvBody::new(self.recv, self.conn.clone(), self.stream_id)
```
djc: Some of the changes in this commit seem unrelated and due to reformatting, could you please isolate them in a separate commit?

quinn-h3/src/server.rs
```diff
@@ -160,17 +160,14 @@ impl RecvRequest {
...
-
-        Ok(request
+        let mut request = Request::builder()
+            .method(method)
+            .uri(uri)
+            .version(http::version::Version::HTTP_3)
             .body(RecvBody::new(recv, self.conn.clone(), self.stream_id))
-            .map_err(|e| Error::Peer(format!("invalid request: {:?}", e)))?)
+            .unwrap();
+        *request.headers_mut() = headers;
```
djc: Should `Request::builder()` also have a way to set the `headers`?

quinn-h3/src/proto/frame.rs
```diff
@@ -48,7 +49,11 @@ impl HttpFrame {
         let len = buf.get_var()?;
 
         if buf.remaining() < len as usize {
-            return Err(Error::Incomplete(2 + len as usize));
+            if ty == Type::DATA {
+                return Err(Error::IncompleteData);
```
djc: Maybe consider hoisting `return Err()` out of the `if` here?

quinn-h3/src/body.rs
stammw: This has been removed in a later commit.

djc: Ah, never mind then!

quinn-h3/src/server.rs
stammw: `Request::Builder` has .header()

djc: Makes sense!

quinn-h3/src/client.rs
stammw: Done in the previous commit.

quinn-h3/src/proto/frame.rs
stammw: Done.

stammw:
Good to see you appreciate it ! Thanks for the review, hope you will eventually get more time :)

PR #427, opened 2019-09-06
H3 stream body for sending

stammw:
Relates to #357

### DIFF
--- a/quinn-h3/examples/h3.rs
+++ b/quinn-h3/examples/h3.rs
@@ -228,6 +228,7 @@ fn handle_request(
 
             sender
                 .send_response_trailers(response, trailer)
+                .unwrap()
                 .map_err(|e| format_err!("failed to send response: {:?}", e))
         })
 }
--- a/quinn-h3/examples/h3.rs
+++ b/quinn-h3/examples/h3.rs
@@ -199,7 +199,7 @@ fn handle_request(
     tokio::io::read_to_end(body.into_reader(), content)
         .map_err(|e| format_err!("failed to receive response body: {:?}", e))
         .and_then(move |(reader, content)| {
-            println!("server received body: {:?}", content);
+            println!("server received body len: {:?}", content.len());
 
             match reader.trailers() {
                 None => Either::A(futures::future::ok(())),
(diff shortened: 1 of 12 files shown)


quinn-h3/src/server.rs
```diff
@@ -220,25 +222,69 @@ pub struct Sender {
...
-        trailer: HeaderMap,
-    ) -> Result<SendResponse, Error> {
-        SendResponse::with_trailers(
-            response,
-            Some(trailer),
-            self.send,
-            self.stream_id,
-            self.conn,
-        )
+pub struct RespBuilder<T> {
```
djc: Let's just call it `ResponseBuilder`?

quinn-h3/src/client.rs
```diff
@@ -129,6 +120,31 @@ impl Future for Connecting {
     }
 }
 
+pub struct ReqBuilder<T> {
```
djc: Would also rename this `RequestBuilder`.

quinn-h3/examples/h3.rs
```diff
@@ -291,31 +291,50 @@ fn client(
...
-                            }
-                            Either::B(futures::future::ok(()))
-                        })
+                        .and_then(|response| {
+                            println!("received response: {:?}", response);
+
+                            let (_, body) = response.into_parts();
+
+                            let buf = Vec::with_capacity(1024 * 10); // 10K
+                            tokio_io::io::read_to_end(body.into_reader(), buf)
```
djc: This looks like things have gotten a good deal more complex. Should we add some abstractions to make the common cases easier to deal with?

djc:
So I found some mostly-superficial things that should probably be changed. Otherwise things look mostly okay. I have to say it feels like there is more complexity than what I think should be necessary, but unfortunately I'm not deep enough in that I can confidently explain why/what. I hope to spend more time soon on my hyper integration so hopefully that'll give me deeper insight.

quinn-h3/examples/h3.rs
stammw: That's right. I'm not sure of what I can do to get a simpler API. 
Still, this example is the most complex thing you can acheive as a user right now: sending request body and recieving response body in parallel.

A simpler API is still availabe through `ReqBuilder::send()`. 

I'll write a few more examples to show that way of doing things. This might help finding some more ergonomic solutions.

quinn-h3/src/server.rs
stammw: I wasn't sure about this, this `struct` is always used with `http::RequestBuiler`. But using an abreviation to solve namespace "problem" is maybe not that clean.

stammw: Renamed.

quinn-h3/examples/h3.rs
stammw: I just pushed simpler examples, still more complex than when receiving body was in the same future as receiving response header.

stammw:
Awesome to start talking about Hyper integration :)

Thanks for the review, I'll start working on `std-futures` now !

PR #429, opened 2019-09-08
Per-directionality accept

Ralith:
The order in which streams are opened is preserved only between streams of the same directionality, and backpressure similarly exists only within streams of the same directionality. As a result, if an application wants to process a certain stream identified by opening sequence (e.g. the first unidirectional stream, perhaps containing important session parameters), it must accept and buffer an arbitrarily large number of streams of the other type.

To correct this, we need to allow applications to be notified of and accept streams of a specific directionality, while allowing others to remain pending.

The incidental renaming improves concision substantially, and I don't think the increase in ambiguity is significant.

### DIFF
--- a/quinn-h3/src/client.rs
+++ b/quinn-h3/src/client.rs
@@ -7,7 +7,6 @@ use futures::{ready, stream::Stream, Future, Poll};
 use http::{request, HeaderMap, Request, Response};
 use quinn::{Endpoint, OpenBi};
 use quinn_proto::StreamId;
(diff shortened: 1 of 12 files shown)


Ralith:
In the interest of avoiding churn in the std-futures branch, I haven't made the corresponding high-level API updates, and this shouldn't be merged until that's sorted out. Nonetheless, review would be appreciated.

djc:
LGTM so far!

Ralith:
Ready for final review.

PR #443, opened 2019-09-20
Update initial salt and QUIC version to draft 23

djc:
This passes the tests at least, and it gets VHDR with quicly.


Ralith:
Maybe `quinn_h3::ALPN` too? It's a bit dated already, though, and it's not like we have much interop there to worry about.

djc:
Fixed up `ALPN_QUIC_HTTP`! I'll leave H3 as is for now.

PR #446, opened 2019-09-24
Draft 23 handshake retransmit, data-bearing probes

Ralith:
Modifies our probe strategy to retransmit in-flight data rather than
sending a PING frame. This gives us a chance of accelerated loss
recovery and allows adoption of the new draft's simplified handshake
retransmit logic.

Took me a while to decide how to do this, but in the end we've got (theoretically) improved robustness *and* a net code reduction!


djc:
Looks great!

Ralith:
Substantially changed, please re-review. Still pretty happy with it, though!

PR #455, opened 2019-10-02
Removed 'Initializer' import and use as it seems to not exist anymore

NULLx76:
This was causing compilation issues for me which are now fixed.

### DIFF
--- a/quinn/src/streams.rs
+++ b/quinn/src/streams.rs
@@ -5,7 +5,7 @@ use std::str;
 use bytes::Bytes;
 use err_derive::Error;
 use futures::channel::oneshot;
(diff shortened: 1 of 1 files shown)


jonay2000:
seems to also suffer from https://github.com/rust-lang/rust/issues/64964

NULLx76:
The compiler bug seems to be fixed with the release of `nightly-2019-10-02` so the code now compiles correctly now. The CI can now be ran again if necessary.

djc:
I'm rerunning CI now: https://dev.azure.com/dochtman/Projects/_build/results?buildId=358&view=results

PR #456, opened 2019-10-02
Fixed doc for Session::is_handshaking

jean-airoldie:
Moreover, shouldn't a future we provided for the user to wait for the connection to be done with handshaking? This would be useful for a protocol that transmit idempotent metadata to the peer at the connection start, but then needs a proper session once that is done.

### DIFF
--- a/quinn-proto/src/crypto.rs
+++ b/quinn-proto/src/crypto.rs
@@ -52,7 +52,7 @@ pub trait Session: Sized {
     /// If the 0-RTT-encrypted data has been accepted by the peer
     fn early_data_accepted(&self) -> Option<bool>;
 
-    /// Returns true when the handshake has completed
+    /// Returns `true` until the connection is fully established.
     fn is_handshaking(&self) -> bool;
 
     /// Read bytes of handshake data
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -44,9 +44,6 @@ impl Connecting {
     /// intercepted by a man-in-the-middle. If this occurs, the handshake will not complete
     /// successfully.
     ///
-    /// Once [`Connection::is_handshaking`] returns false, further transmissions are unaffected by
-    /// these weaknesses.
-    ///
     /// # Errors
     ///
     /// Outgoing connections are only 0-RTT-capable when a cryptographic session ticket cached from


quinn-proto/src/crypto.rs
```diff
@@ -52,7 +52,7 @@ pub trait Session: Sized {
     /// If the 0-RTT-encrypted data has been accepted by the peer
     fn early_data_accepted(&self) -> Option<bool>;
 
-    /// Returns true when the handshake has completed
+    /// Returns `false` when the handshake has completed
```
```suggestion
    /// Returns `true` until the connection is fully established
```
Clearer to focus on the positive side, I think.

quinn/src/connection.rs
```diff
@@ -44,7 +44,7 @@ impl Connecting {
     /// intercepted by a man-in-the-middle. If this occurs, the handshake will not complete
     /// successfully.
     ///
-    /// Once [`Connection::is_handshaking`] returns false, further transmissions are unaffected by
+    /// Once [`Session::is_handshaking`] returns false, further transmissions are unaffected by
```
Ralith: This isn't part of the public API of the this higher-level crate, so I don't think we should mention it here.

jean-airoldie: By that do you mean remove the mention altogether? Or do you mean that eventually a `Connection::is_handshaking` method will be added?

jean-airoldie: I assume its the second so I'm gonna leave it be.

Ralith: I mean this paragraph should be deleted. Its place should ultimately be taken by documentation for the hypothetical future that completes when the connection is established and indicates 0-RTT acceptance or rejection, but that doesn't need to be handled in this PR.

Ralith: I suppose there's no strong reason we couldn't have an accessor too, if it's useful to someone.

jean-airoldie:
Btw should the project compile on stable? My stable build fails because of `tokio-sync`.

jean-airoldie:
If its meant for nightly, I think a `rust-toolchain` file should be added.

Ralith:
Master should currently compile on beta; we're targeting 1.39 so we can release 0.5.0 ASAP when async/await is in stable.

quinn/src/connection.rs
jean-airoldie: Can't say I would use such a method. I guess I'll remove this part of the doc, while I'm at it.

PR #468, opened 2019-10-10
Fix stream ID leak when stream opening futures are dropped

Ralith:
By deferring allocation of the stream ID until the futures synchronously completes, we guarantee that either no ID is allocated or ownership is successfully taken by the Send/RecvStream struct(s).

Thanks to @stammw for noticing that the semantics here were poorly thought out!

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -2,7 +2,7 @@ use std::collections::{BTreeMap, VecDeque};
 use std::net::SocketAddr;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
-use std::{cmp, io, mem, fmt};
+use std::{cmp, fmt, io, mem};
 
 use bytes::{Bytes, BytesMut};
 use err_derive::Error;
@@ -3176,7 +3176,10 @@ where
     }
 }
 
-impl<S> fmt::Debug for Connection<S> where S: crypto::Session {
+impl<S> fmt::Debug for Connection<S>
+where
+    S: crypto::Session,
+{
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.debug_struct("Connection")
             .field("handshake_cid", &self.handshake_cid)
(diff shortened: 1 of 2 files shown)


Ralith:
This also makes an incidental semantic change: stream IDs are assigned in the order the futures complete, which may not be the order in which `open_*` were called. Application protocols can assign semantics to stream ordering, so this could cause surprises. However, ordering was of course never guaranteed between calls in different tasks, so this change should only be observable if you do something like
```rust
let a = conn.open_uni();
let b = conn.open_uni();
tokio::spawn(async move { frob(a.await); });
tokio::spawn(async move { frob(b.await); });
```
which is pretty gratuitously weird, so I think the risk is acceptable. We could also prevent this entirely by modifying `OpenUni` and `OpenBi` to borrow their `Connection`, which I've been thinking about doing regardless because it saves a clone, is a little more consistent, and helps keep our options open. Thoughts?

Ralith:
On review the changed behavior would also be observable (even with a borrow!) if you did something like
```rust
let a = conn.open_uni();
let b = conn.open_uni();
futures::join!(a, b)
```
which is more reasonable, albeit (at least in this trivial form) unnecessary. I still feel like it's not a big deal, though, since the solution (if you really care about stream sequencing) is simpler code anyway.

Ralith:
Looks like CI is spuriously broken by nightly missing rustfmt.

djc:
> We could also prevent this entirely by modifying OpenUni and OpenBi to borrow their Connection, which I've been thinking about doing regardless because it saves a clone, is a little more consistent, and helps keep our options open. Thoughts?

This sounds good to me. I do think doing the non-surprising thing might save us future bug reports. I'm not fully sure about what you by borrowing the `Connection`, would this be an ergonomic setback?

> Looks like CI is spuriously broken by nightly missing rustfmt.

Yeah, makes me remember why I usually stick to stable. Just a few more weeks...

Ralith:
I don't immediately see a way to have stream IDs allocated in order of `open` calls without a relatively complex and nonzero-cost scheme involving making reservations and revoking them on drop.

PR #469, opened 2019-10-10
Implement FusedFuture and FusedStream

Ralith:
Incomplete, seeking feedback before investing further effort.

[The `futures::select` macro](https://docs.rs/futures-preview/0.3.0-alpha.19/futures/macro.select.html) is designed to be called in a loop, and relies on `FusedFuture` to avoid causing weird behavior by polling already-complete futures and streams. [The `fuse` combinator](https://docs.rs/futures-preview/0.3.0-alpha.19/futures/future/trait.FutureExt.html#method.fuse) can be used to accomplish this for any future (and there's a stream version too), so while it's convenient for types to implement this directly, it's not strictly necessary.

Implementing them ourselves adds a small amount of extra complexity (and branching), and saves callers who want to use `futures::select` a very small amount of boilerplate. I'm not sure if this is a win, even though we do coincidentally save a `clone` for `OpenUni`/`Bi`.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -8,6 +8,8 @@ use bytes::Bytes;
 use err_derive::Error;
 use fnv::FnvHashMap;
 use futures::channel::{mpsc, oneshot};
+use futures::future::FusedFuture;
+use futures::stream::FusedStream;
 use futures::task::{Context, Waker};
 use futures::{Future, FutureExt, Poll, StreamExt};
 use proto::{ConnectionError, ConnectionHandle, ConnectionId, Dir, StreamId, TimerUpdate};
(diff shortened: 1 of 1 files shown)


djc:
So I guess we just panic if someone keeps calling it? What was our previous behavior in this case, exactly? Have you checked the benchmark differences from the extra `Option` machinations?

Ralith:
> So I guess we just panic if someone keeps calling it?

Polling a future after it's completed is explicit unspecified (though not undefined) behavior territory, so that's a relatively polite thing to do.

> What was our previous behavior in this case, exactly?

It looks like it'd be equivalent to obtaining and polling a fresh future, but this is absolutely not something that a sane caller would rely on, and we should feel free to change that behavior at any time.

> Have you checked the benchmark differences from the extra Option machinations?

We don't have a working benchmark that exercises these code paths very well yet, unfortunately. I don't imagine it'd register, though. My concern is mostly about the added complexity. Currently I'm leaning towards staying away from this; lots of things don't implement `FusedFuture`, the `FutureExt::fuse` combinator is handy enough, and ultimately it's the `select` macro's problem that it has such stringent requirements.

djc:
Agreed.

PR #476, opened 2019-10-22
Refactor benchmarks, bugfix

Ralith:
Reworks the existing benchmarks for concision and run the client and server on different threads, and fixes what I'm pretty sure was the issue behind the second benchmark all along: accepting a stream wasn't waking up the connection driver to issue new stream ID credit, which could lead to deadlocks if the receiver manages to ACK an entire window's worth of stream IDs before the application accepts any, and has no other reason to send packets.


Ralith:
CI failure seems to be unrelated, and due to breakage in the upstream scripts:
> error: --no-default-features is not allowed in the root of a virtual workspace

djc:
Awesome!

PR #478, opened 2019-10-26
Replace slog with tracing

Ralith:
No obvious impact on the benchmarks, though they're super noisy so it'd be easy to miss a small one.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -8,7 +8,7 @@ use bytes::{Bytes, BytesMut};
 use err_derive::Error;
 use fnv::FnvHashSet;
 use rand::{rngs::StdRng, Rng, SeedableRng};
-use slog::Logger;
+use tracing::{debug, error, info, trace, trace_span, warn};
 
 use crate::coding::BufMutExt;
 use crate::crypto::{self, HeaderKeys, Keys};
@@ -42,7 +42,6 @@ pub struct Connection<S>
 where
     S: crypto::Session,
 {
-    log: Logger,
     endpoint_config: Arc<EndpointConfig>,
     server_config: Option<Arc<ServerConfig<S>>>,
     config: Arc<TransportConfig>,
@@ -163,7 +162,6 @@ where
     S: crypto::Session,
 {
     pub(crate) fn new(
-        log: Logger,
         endpoint_config: Arc<EndpointConfig>,
         server_config: Option<Arc<ServerConfig<S>>>,
         config: Arc<TransportConfig>,
@@ -193,7 +191,6 @@ where
             .as_ref()
             .map_or(false, |c| c.use_stateless_retry);
         let mut this = Self {
-            log,
             endpoint_config,
             server_config,
             tls,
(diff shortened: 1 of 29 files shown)


Ralith:
CI failure is spurious.

quinn-proto/src/connection.rs
```diff
@@ -928,6 +914,8 @@ where
         packet: Packet,
         remaining: Option<BytesMut>,
     ) -> Result<(), TransportError> {
+        let span = trace_span!("recv initial");
```
djc: Is it possible to chain the `.enter()` call, or does that run into lifetime issues?

quinn-proto/src/connection.rs
```diff
@@ -2494,12 +2425,7 @@ where
             }
             sent.max_stream_data.insert(id);
             let max = rs.bytes_read + self.config.stream_receive_window;
-            trace!(
-                self.log,
-                "MAX_STREAM_DATA: {stream} = {max}",
-                stream = id,
-                max = max
-            );
+            trace!(stream = %id, max = max, "MAX_STREAM_DATA",);
```
djc: Is the trailing comma mandatory here? Should we file a bug about that?

quinn-proto/src/connection.rs
Ralith: Sadly `creates a temporary which is freed while still in use` territory.

quinn-proto/src/connection.rs
Ralith: Nah, just a refactoring artifact. Fixed.

Ralith:
It's driven by thread locals; I couldn't detect any benchmark impact, as mentioned in the PR description. Wouldn't hurt to see what you get, if you're inclined.

Ultimately I doubt we're anywhere near the level where thread local access is performance-relevant, especially given the considerable effort that's been invested in making `tracing` fast on an absolute scale.

djc:
Cool! I was not sure what to expect in terms of performance changes, so I was just curious. From a single run on my macOS laptop:

```
throughput/large streams
                        time:   [1.0734 ms 1.0821 ms 1.0905 ms]
                        thrpt:  [114.63 MiB/s 115.51 MiB/s 116.45 MiB/s]
                 change:
                        time:   [-2.7225% -0.9625% +0.6692%] (p = 0.28 > 0.05)
                        thrpt:  [-0.6648% +0.9718% +2.7987%]
                        No change in performance detected.
throughput/small streams unpipelined
                        time:   [144.31 us 144.90 us 145.59 us]
                        thrpt:  [6.8688 Kelem/s 6.9014 Kelem/s 6.9295 Kelem/s]
                 change:
                        time:   [-6.5942% -5.4392% -4.3805%] (p = 0.00 < 0.05)
                        thrpt:  [+4.5812% +5.7521% +7.0598%]
                        Performance has improved.
throughput/small datagrams
                        time:   [490.89 ns 496.74 ns 502.71 ns]
                        thrpt:  [1.9892 Melem/s 2.0131 Melem/s 2.0371 Melem/s]
                 change:
                        time:   [-12.058% +4.6656% +25.644%] (p = 0.68 > 0.05)
                        thrpt:  [-20.410% -4.4576% +13.712%]
                        No change in performance detected.
```

PR #479, opened 2019-11-02
Remove fnv hash

### LINKED ISSUE #88: Audit uses of `FnvHashMap` for collision attack potential
We currently use `FnvHashMap` in a number of places, presumably because it is faster than the std `HashMap`. However, the reason `FnvHashMap` is not the default, as I understand it, is because it is possible for attackers to generate input data that will cause denial of service through worst-case performance of hash map algorithms.

In `Endpoint`, we currently use `FnvHashMap` for (1) `connection_ids_initial`, (2) `connection_ids`, and (3) `connection_remotes`. (1) seems definitely vulnerable to this style of attack for server endpoints, since clients can randomly pick initial CIDs for their packets, thus triggering collisions. (2) seems safe, since local CIDs are generated by our own code. For (3), I'm not sure how easy it is to spoof IP addresses these days.

Ralith:
This had no visible performance benefit and may have created a HashDoS vulnerability.

Fixes #88.

### DIFF
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -23,7 +23,6 @@ tls-rustls = ["rustls", "webpki", "ring"]
 [dependencies]
 bytes = "0.4.7"
 err-derive = "0.2"
-fnv = "1.0.6"
 lazy_static = "1"
 rand = "0.7"
 ring = { version = "0.16.7", optional = true }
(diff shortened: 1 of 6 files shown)


djc:
Nice!

Ralith:
Should this eventually become performance relevant, [aHash](https://github.com/tkaitchuck/aHash) seems like an interesting DoS-resistant option.

PR #482, opened 2019-11-05
Added access modifiers to config types

### LINKED ISSUE #444: Consider making the config structs opaque
Presently, quinn is configured through a handful of structs with public members. This is simple to work with, but presents a forwards-compatibility hazard, because the representation of parameters cannot be changed without making a breaking change to the interface. For example, we might want to switch all time intervals to use `Duration`. Per-field validation is also awkward to perform.

Instead, each struct could have entirely private fields and a number of setters, which could consume diverse types and even return `Result` where appropriate. If changes are made, old setters could be deprecated rather than removed wholesale.

TimonPost:
Partially fixed #444 

Refactored: `ServerConfig`, `EndpointConfig` and `ClientConfig`.

- Changed `pub` definition into `pub (crate)`. 
- Added set fields. `fn with_xxx() -> Self { }`
- Added get fields for transport config
- Updated code that used those config types

Use can create and modifie the config as follows: 

```rust
let endpoint = EndpointConfig::default()
     .with_reset_key(vec![])
     .with_local_cid_len(10);
....
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ can require a larger UDP buffer than is configured by default in most
 environments. If you observe erratic latency and/or throughput over a stable
 network link, consider increasing the buffer sizes used. For example, you could
 adjust the `SO_SNDBUF` and `SO_RCVBUF` options of the UDP socket to be used
-before passing it in to Quinn. Note that some platforms (e.g. Linux) require
+before passing it into Quinn. Note that some platforms (e.g. Linux) require
 elevated privileges or modified system configuration for a process to increase
 its UDP buffer sizes.
 
@@ -68,7 +68,7 @@ this can be accomplished by using certificates from [Let's Encrypt][letsencrypt]
 for servers, and relying on the default configuration for clients.
 
 For some cases, including peer-to-peer, trust-on-first-use, deliberately
-insecure applications, or any case where servers are not identified by domain
+insecure applications, or any case where servers are not identified by a domain
 name, this isn't practical. Arbitrary certificate validation logic can be
 implemented by enabling the `dangerous_configuration` feature of `rustls` and
 constructing a Quinn `ClientConfig` with an overridden certificate verifier by
@@ -95,7 +95,7 @@ client will automatically find and trust it.
 ## Development
 
 The quinn-proto test suite uses simulated IO for reproducibility and to avoid
-long sleeps in certain timing-sensitive tests. If the `SSLKEYLOGFILE`
+long periods of sleep in certain timing-sensitive tests. If the `SSLKEYLOGFILE`
 environment variable is set, the tests will emit UDP packets for inspection
 using external protocol analyzers like Wireshark, and NSS-compatible key logs
 for the client side of each connection will be written to the path specified in
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -1,4 +1,5 @@
 use std::collections::{HashMap, VecDeque};
+use std::fmt;
 use std::iter;
 use std::net::SocketAddr;
 use std::ops::{Index, IndexMut};
@@ -7,10 +8,9 @@ use std::time::{Duration, Instant, SystemTime};
 
 use bytes::{BufMut, BytesMut};
 use err_derive::Error;
-use fnv::FnvHashMap;
 use rand::{rngs::StdRng, Rng, RngCore, SeedableRng};
 use slab::Slab;
-use slog::{self, Logger};
+use tracing::{debug, trace, warn};
 
 use crate::coding::BufMutExt;
 use crate::connection::{initial_close, Connection};
@@ -37,13 +37,12 @@ pub struct Endpoint<S>
 where
     S: crypto::Session,
 {
-    log: Logger,
     rng: StdRng,
     transmits: VecDeque<Transmit>,
-    connection_ids_initial: FnvHashMap<ConnectionId, ConnectionHandle>,
-    connection_ids: FnvHashMap<ConnectionId, ConnectionHandle>,
+    connection_ids_initial: HashMap<ConnectionId, ConnectionHandle>,
+    connection_ids: HashMap<ConnectionId, ConnectionHandle>,
     /// Identifies connections with zero-length CIDs
-    connection_remotes: FnvHashMap<SocketAddr, ConnectionHandle>,
+    connection_remotes: HashMap<SocketAddr, ConnectionHandle>,
     /// Reset tokens provided by the peer for the CID each connection is currently sending to
     ///
     /// Incoming stateless resets do not have correct CIDs, so we need this to identify the correct
@@ -69,18 +68,16 @@ where
     ///
     /// Returns `Err` if the configuration is invalid.
     pub fn new(
-        log: Logger,
         config: Arc<EndpointConfig>,
         server_config: Option<Arc<ServerConfig<S>>>,
     ) -> Result<Self, ConfigError> {
         config.validate()?;
         Ok(Self {
-            log,
             rng: StdRng::from_entropy(),
             transmits: VecDeque::new(),
-            connection_ids_initial: FnvHashMap::default(),
-            connection_ids: FnvHashMap::default(),
-            connection_remotes: FnvHashMap::default(),
+            connection_ids_initial: HashMap::new(),
+            connection_ids: HashMap::new(),
+            connection_remotes: HashMap::new(),
             connection_reset_tokens: HashMap::new(),
             connections: Slab::new(),
             incoming_handshakes: 0,
@@ -127,17 +124,12 @@ where
                     self.connection_reset_tokens.remove(&old).unwrap();
                 }
                 if self.connection_reset_tokens.insert(token, ch).is_some() {
-                    warn!(self.log, "duplicate reset token");
+                    warn!("duplicate reset token");
                 }
             }
             RetireConnectionId(seq) => {
                 if let Some(cid) = self.connections[ch].loc_cids.remove(&seq) {
-                    trace!(
-                        self.log,
-                        "peer retired CID {sequence}: {cid}",
-                        sequence = seq,
-                        cid = cid,
-                    );
+                    trace!("peer retired CID {}: {}", seq, cid);
                     self.connection_ids.remove(&cid);
                     return Some(self.send_new_identifiers(ch, 1));
                 }
@@ -175,10 +167,10 @@ where
                 destination,
             }) => {
                 if !self.is_server() {
-                    debug!(self.log, "dropping packet with unsupported version");
+                    debug!("dropping packet with unsupported version");
                     return None;
                 }
-                trace!(self.log, "sending version negotiation");
+                trace!("sending version negotiation");
                 // Negotiate versions
                 let mut buf = Vec::<u8>::new();
                 Header::VersionNegotiate {
@@ -197,7 +189,7 @@ where
                 return None;
             }
             Err(e) => {
-                trace!(self.log, "malformed header"; "reason" => %e);
+                trace!("malformed header: {}", e);
                 return None;
             }
         };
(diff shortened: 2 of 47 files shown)


Ralith:
Do we need public accessors for config values? Keeping it write-only would simplify the API, and let the setters be a bit more concise as no `with_` would be needed to distinguish them.

Ralith:
I don't think there's any need to move everything over to getters within quinn-proto, also. We probably want the fields to all be `pub(crate)`.

TimonPost:
> Do we need public getters for config values

"Instead, each struct could have entirely private fields and a number of setters, which could consume diverse types and even return Result where appropriate." - is what you said in the other issue. 

It depends on what the use case is. If the user is going to interact with it I think it can be of use. If the user is not going to interact with it we can just as well make it `pub (crate)` because there will be less code.

>  Keeping it write-only would simplify the API

What do you mean with a write-only api?

> We probably want the fields to all be pub(crate).

If we only use configuration internally we can do this yes. If the user is coming into play we might create those get/set functions and use the same API as the user does. No need to have both the option `pub (crate)` and get/set functions. 


The motivation for tose get/set functions are that the user probably wants to change a few to no config values. It can do this by calling `Config::default().with_xxx()` or just `Config::default()` in most cases. By having `with` in the name, you make the config behave like a builder. That can be built as the user want's to. 

If you don't see any reasons for those get/set functions. Why the statement: _"Instead, each struct could have entirely private fields and a number of setters, which could consume diverse types and even return Result where appropriate."_  Do you mean only have set functions with `pub (crate)` fields in the config?

Ralith:
> is what you said in the other issue.

Yes, *setters*, as opposed to both setters and getters. I think setters are a big win here, but I don't see the case for getters.

> If the user is going to interact with it I think it can be of use. If the user is not going to interact with it we can just as well make it pub (crate) because there will be less code.

I'm not aware of any case for the user to need to extract information from an existing configuration, so indeed less code and churn seems preferable. Setters are good, but if there're no getters, we can give them more concise names while keeping the same builder-style type signature.

README.md
```diff
@@ -94,8 +94,8 @@ client will automatically find and trust it.
 
 ## Development
 
-The quinn-proto test suite uses simulated IO for reproducibility and to avoid
-long sleeps in certain timing-sensitive tests. If the `SSLKEYLOGFILE`
+The Quinn-proto test suite uses simulated IO for reproducibility and to avoid
```
djc: The crate is spelled with a lower-case q.

quinn-proto/src/endpoint.rs
```diff
@@ -351,7 +351,7 @@ where
         remote: SocketAddr,
         server_name: &str,
     ) -> Result<(ConnectionHandle, Connection<S>), ConnectError> {
-        config.transport.validate()?;
+        config.transport_config().validate()?;
```
djc: We decided we didn't want getters, right?

quinn-proto/src/shared.rs
```diff
@@ -257,33 +270,83 @@ pub struct ServerConfig<S>
...
     S: crypto::Session,
 {
-    /// Transport configuration to use for incoming connections
-    pub transport: Arc<TransportConfig>,
-
-    /// TLS configuration used for incoming connections.
-    ///
-    /// Must be set to use TLS 1.3 only.
+    /// Transport configuration that is used for incoming connections.
+    transport_config: Arc<TransportConfig>,
```
djc: I don't really like this renaming; `config.transport` seems perfectly clear to me, whereas `config.transport_config` is unnecessarily repetitive.

quinn-proto/src/shared.rs
```diff
@@ -257,33 +270,83 @@ pub struct ServerConfig<S>
...
+    /// Mutable TLS configuration used for incoming connections.
```
djc: Why did this lose the sentence about being restricted to TLS 1.3?

quinn-proto/src/shared.rs
```diff
@@ -257,33 +270,83 @@ pub struct ServerConfig<S>
...
     pub crypto: S::ServerConfig,
+    /// Private key used for authentication.
```
djc: Why rewrite this comment to include less information?

quinn-proto/src/shared.rs
TimonPost: Because it was a field wrapped in `Arc` I thought it could be useful to add a getter. This to reduce redundant cloning trough out the lib.

djc: Okay, but in that case please also remove all the cloning that's currently there so we can assess the value of adding this getter.

README.md
TimonPost: Of course, I saw Quin also, however, a crate is not capitalized but the name is. Fil fix.

quinn-proto/src/endpoint.rs
TimonPost: Because it was a field wrapped in `Arc` I thought it could be useful to add a getter. This to reduce redundant cloning trough out the lib.

README.md
djc: Yeah, the project we've usually spelled with a capital. But for crates it doesn't make much sense.

quinn-proto/src/endpoint.rs
Ralith: This has actually added additional unnecessary cloning. It's best to keep these things explicit.

quinn-proto/src/shared.rs
Ralith: I think in context the shorter name is reasonable; it's within a `...Config` struct, it has a `...Config` type, and there's a similarly named field right below it. In any case, renaming things would be better done in a separate PR.

TimonPost: Okay.

quinn-proto/src/shared.rs
```diff
@@ -257,33 +273,81 @@ pub struct ServerConfig<S>
 where
     S: crypto::Session,
 {
-    /// Transport configuration to use for incoming connections
+    /// TLS configuration used for incoming connections.
```
djc: Not sure why this change -- `TransportConfig` does not contain TLS configuration.

quinn-proto/src/shared.rs
```diff
@@ -353,13 +417,30 @@ where
 /// Default values should be suitable for most internet applications.
 #[derive(Clone, Debug, Default)]
 pub struct ClientConfig<C> {
-    /// Transport configuration to use
+    /// TLS configuration used for incoming connections.
```
djc: As above.

quinn-proto/src/shared.rs
```diff
@@ -353,13 +417,30 @@ where
...
     pub transport: Arc<TransportConfig>,
-
-    /// Cryptographic configuration to use
+    /// The TLS configuration.
     pub crypto: C,
 }
 
+impl<C> ClientConfig<C> {
+    /// Set the TLS configuration used for incoming connections.
```
djc: As above.

quinn-proto/src/shared.rs
```diff
@@ -240,6 +235,7 @@ impl Default for EndpointConfig {
 }
 
 impl EndpointConfig {
+    /// Validate if the local cid length is within the legal range.
```
Ralith: Logically, this validates the entire configuration; it just happens that there's currently only one field that we check.

quinn-proto/src/shared.rs
```diff
@@ -257,33 +273,81 @@ pub struct ServerConfig<S>
 where
     S: crypto::Session,
 {
-    /// Transport configuration to use for incoming connections
+    /// TLS configuration used for incoming connections.
     pub transport: Arc<TransportConfig>,
+    /// Mutable TLS configuration used for incoming connections.
+    pub crypto: S::ServerConfig,
```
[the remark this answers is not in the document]
TimonPost: They are used by the benchmarks, interop, tests, that do not belong to the crate, and can't reference those fields. Since getters are not allowed we have to expose those fields.

quinn-proto/src/shared.rs
```diff
@@ -257,33 +273,81 @@ pub struct ServerConfig<S>
...
+    }
+
+    /// Set microseconds after a stateless retry token was issued for which it's considered valid.
+    pub fn with_retry_token_lifetime(&mut self, retry_token_lifetime: u64) -> &mut Self {
+        self.retry_token_lifetime = retry_token_lifetime;
+        self
+    }
 
-    /// Maximum number of incoming connections to buffer.
+    /// Set the maximum supported length for the incoming connection buffer.
```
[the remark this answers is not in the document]
TimonPost: I would like to stick to the same naming convention. Which for a set method is 'singular verb' + what the method does.  This is the convention used in most places. Do you have a proposal?

quinn-proto/src/shared.rs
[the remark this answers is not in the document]
Ralith: Can you provide specific examples of cases outside quinn-proto where either of these fields are read from?

quinn-proto/src/shared.rs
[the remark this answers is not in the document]
Ralith: I'm not sure I understand. This comment is discussing the wording of the comment on the line it's attached to.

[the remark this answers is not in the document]
TimonPost: I found: "of incoming connections to buffer." a bit unclear, because 1) to what buffer, do I need to know there is a buffer and does it make sense for me to know. 2) The sentence just correct English, what do you mean with "connections to buffer"

quinn-proto/src/shared.rs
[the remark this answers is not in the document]
TimonPost: quinn\src\builders.rs:132:9
quinn\src\builders.rs:142:9
self.config.crypto.set_protocols(protocols);

quinn-proto/src/shared.rs
```diff
@@ -248,6 +244,26 @@ impl EndpointConfig {
...
         Ok(())
     }
+
+    /// Set the local cid length.
+    ///
+    /// This is the length of connection IDs for the endpoint.
+    /// This must be either 0 or between 4 and 18 inclusive. The length of the local connection IDs
+    /// constrains the amount of simultaneous connections the endpoint can maintain. The API user is
+    /// responsible for making sure that the pool is large enough to cover the intended usage.
+    pub fn with_local_cid_len(&mut self, local_cid_len: usize) -> &mut Self {
```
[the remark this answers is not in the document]
TimonPost: > I think we should omit the with_ prefix here,

If we do it here, we should do it on all methods as well.

>  It's also confusing when combined with mutating setters, since IMO "with" does not connote in-place modification.

Setters are mutable anyway, we won't have immutable setters.  Therefore we just have to keep one naming convention and use for all setters.

quinn-proto/src/shared.rs
TimonPost: This is moved to the setter because we should denote restrictions when setting a value. When copying it at two places we will get duplicates.

quinn-proto/src/shared.rs
```diff
@@ -353,13 +417,30 @@ where
 /// Default values should be suitable for most internet applications.
 #[derive(Clone, Debug, Default)]
 pub struct ClientConfig<C> {
-    /// Transport configuration to use
+    /// TLS configuration used for incoming connections.
     pub transport: Arc<TransportConfig>,
-
-    /// Cryptographic configuration to use
+    /// The TLS configuration.
     pub crypto: C,
```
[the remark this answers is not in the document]
TimonPost:
```
error[E0616]: field `crypto` of struct `quinn_proto::shared::ClientConfig` is private
   --> quinn\src\builders.rs:193:9
    |
193 |         self.config.crypto.add_certificate_authority(cert)?;
    |         ^^^^^^^^^^^^^^^^^^

error[E0616]: field `crypto` of struct `quinn_proto::shared::ClientConfig` is private
   --> quinn\src\builders.rs:201:9
    |
201 |         self.config.crypto.enable_keylog();
    |         ^^^^^^^^^^^^^^^^^^

error[E0616]: field `crypto` of struct `quinn_proto::shared::ClientConfig` is private
   --> quinn\src\builders.rs:213:9
    |
213 |         self.config.crypto.set_protocols(protocols);
    |         ^^^^^^^^^^^^^^^^^^

error[E0616]: field `crypto` of struct `quinn_proto::shared::ClientConfig` is private
   --> quinn\src\builders.rs:226:41
    |
226 |         let crypto = Arc::make_mut(&mut x.crypto)

error[E0616]: field `transport` of struct `quinn_proto::shared::ClientConfig` is private
  --> quinn\src\tests.rs:28:5
   |
28 |     client_config.transport = Arc::new(crate::TransportConfig {
```

quinn-proto/src/shared.rs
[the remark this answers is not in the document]
Ralith: I see. Maybe we can refactor `ServerConfigBuilder` and `ClientConfigBuilder` to render those accesses unnecessary, but it looks like that may not be trivial. In that light I'm content to leave them public for now.

quinn-proto/src/shared.rs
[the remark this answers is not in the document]
Ralith: In the original wording, "buffer" is a verb, meaning roughly "to collect for future processing." While some discussion of what that means in this specific case may be justified, I don't think the changed wording adds any additional information.

quinn-proto/src/frame.rs
```diff
@@ -40,7 +40,7 @@ impl coding::Codec for Type {
     }
 }
 
-pub trait FrameStruct {
+pub(crate) trait FrameStruct {
```
Ralith: I'm not sure I agree with explicitly crate-qualifying all internal item definitions, and in any case, such a change would belong in a separate PR so it can be more easily reviewed.

quinn-proto/src/shared.rs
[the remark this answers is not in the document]
Ralith: > If we do it here, we should do it on all methods as well.

Yes, by "here" I meant "in this PR." Apologies for the unclear phrasing.

> Setters are mutable anyway, we won't have immutable setters.

I expect `with_`-prefixed methods to take `self` by value and produce a new value as a result, rather than updating a pre-existing value in-place. This is one of two common patterns for builders.

TimonPost:
Sorry, PR is kind of a mess now, squashing went wrong and some changes slipt trough that shouldn't be in here, don't want to spend any more time on this PR.

PR #485, opened 2019-11-15
Merged imports with rustfmt

TimonPost:
- all imports merged  with rustfmt

I very much like imports and think it makes it more readable. If you think otherwise, we can forget about this PR and I close it.

### DIFF
--- a/interop/src/main.rs
+++ b/interop/src/main.rs
@@ -1,5 +1,7 @@
-use std::net::{SocketAddr, ToSocketAddrs};
-use std::sync::{Arc, Mutex};
+use std::{
+    net::{SocketAddr, ToSocketAddrs},
+    sync::{Arc, Mutex},
+};
 
 use failure::{format_err, Error};
 use futures::TryFutureExt;
--- a/quinn-h3/examples/h3.rs
+++ b/quinn-h3/examples/h3.rs
@@ -1,8 +1,9 @@
-use std::net::SocketAddr;
-use std::net::ToSocketAddrs;
-use std::path::PathBuf;
-use std::sync::Arc;
-use std::time::Instant;
+use std::{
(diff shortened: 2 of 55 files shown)


djc:
The decreased vertical density makes this not an improvement in my opinion even though the prefix separation is slightly more emphasized. @Ralith what do you think?

Ralith:
I don't have a strong opinion either way, though in a vacuum I'm inclined to defer to rustfmt's defaults.

djc:
I think rustfmt is unopinionated on this topic, it just deals with formatting (although it sorts them) but doesn't do adjustments like these either way AFAICT.

Ralith:
Good point. In that case I'm indifferent; feel free to defer to your taste, @djc.

djc:
Okay, let's skip it. Still happy to have that formatting commit, though!

TimonPost:
So what is it going to be? Merge or close?

djc:
@stammw so I guess you are in favor, how attached to it are you?

TimonPost:
Haha, what do you mean that I am attached to it? I ran rustfmt with merge imports on it was 5 minutes of work, so if you don't see the added value, just close this PR. 

However, I think it makes the reading of the imports way more cleaner because there is less repetition when imorting types from the same module paths. I personally like to this with my libraries.

stammw:
I feel we should choose one way of writing imports and stick to it for consistency. I like the nested syntax because it exposes modules structure and contains less useless repetitions. In my opinion having several lines with the same prefix looks a bit ceremonial.

I was planning to apply merged imports syntax in a future readability + documentation pass.

I'm indeed in favor of this PR.

djc:
Alright, let's merge it then. @TimonPost can you rebase it?

TimonPost:
Rebased

PR #488, opened 2019-11-16
Replace webpki-roots with optional native certificate store use

### LINKED ISSUE #451: Use OS certificate stores by default
Currently we depend on webpki-roots to bake in sane defaults. This is better than nothing, but doesn't expose user-installed certificates or handle live changes. While those can be accomplished by users employing custom certificate validation logic, it would be nice to handle them correctly out of the box.

Blocked on https://github.com/ctz/rustls-native-certs being released

Ralith:
Fixes #451.

### DIFF
--- a/quinn/Cargo.toml
+++ b/quinn/Cargo.toml
@@ -14,6 +14,11 @@ edition = "2018"
 [package.metadata.docs.rs]
 all-features = true
 
+[features]
+default = ["native-certs", "ct-logs"]
+# Trust the contents of the OS certificate store by default
+native-certs = [ "rustls-native-certs" ]
+
 [badges]
 codecov = { repository = "djc/quinn" }
 maintenance = { status = "experimental" }
@@ -21,7 +26,7 @@ azure-devops = { project = "dochtman/Projects", pipeline = "Quinn", build = "1" 
 
 [dependencies]
 bytes = "0.4.7"
-ct-logs = "0.6"
+ct-logs = { version = "0.6", optional = true }
 err-derive = "0.2"
 fnv = "1.0.6"
 futures = { package = "futures-preview", version = "0.3.0-alpha.18" }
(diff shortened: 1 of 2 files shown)


djc:
I'd kind of like to keep webpki-roots as an option, at least? The [trade-offs](https://github.com/ctz/rustls-native-certs#should-i-use-this-or-webpki-roots) as explained in the native-certs README are very real to me.

Ralith:
Is it not sufficient to pull in webpki-roots in downstream code when desired? I'd like to keep the number of build configurations low.

djc:
What's the goal for keeping the number of build configurations low, just a sense of how well we can document this thing? I would say these types of configurations are at least in the same space (mutually exclusive), so it doesn't add a lot to the configuration space.

Personally I'm on the fence about rustls-native-certs being more deserving of being a pre-defined build time config option than webpki-roots -- I do worry that some native cert stores have large amounts of crap in them.

Ralith:
Documentation, testing, ease of use to non-security-experts who won't know what to do when presented with an array of options. If we have two options, what happens if you enable them both?

I think webpki-roots is an inappropriate default simply because it cannot be updated in-place. We don't want people building software with baked-in trust roots if they can't force their users to update on a very regular basis, which is the overwhelmingly common case.

I'm also receptive to the idea of removing defaults entirely. We've found that development/testing certificates are most easily addressed with `rcgen` so far, and the case of a naive user trying to connect to live internet services to test the library seems relatively niche, particularly given the limited deployment of QUIC at present. Maybe higher-level protocol libraries (e.g. `quinn-h3` or even `reqwest`) should be responsible for choosing such defaults, if any.

djc:
Removing defaults doesn't mesh well with the feedback we've gotten from Carl that we should provide an easy API for (especially client-side) raw QUIC usage.

Issue #360, opened 2019-05-29
H3: stateful QPACK encoding


stammw:
Implement `DynamicTable` usage in `quinn-h3::proto::Connection::encode_header` and `quinn-h3::proto::Connection::decode_header`, and drive data transmission for the `EncoderStream` and `DecoderStream`. Conforming to [4.2.1. Encoder and Decoder Streams](https://quicwg.org/base-drafts/draft-ietf-quic-qpack.html#enc-dec-stream-def).

stammw:
closed via #489

PR #489, opened 2019-11-16
H3 stateful qpack

stammw:
Addresses #360


quinn-h3/src/connection.rs
```diff
@@ -236,16 +240,27 @@ impl ConnectionInner {
...
-        match Pin::new(&mut self.send_control).poll(cx) {
-            Poll::Ready(Err(err)) => {
-                self.set_error(ErrorCode::CLOSED_CRITICAL_STREAM, format!("{:?}", err));
-                return Err(err);
+        let types = [
+            PendingStreamType::CONTROL,
+            PendingStreamType::ENCODER,
+            PendingStreamType::DECODER,
+        ];
+        for ty in types.iter() {
```
Ralith: Writing `&ty` here will make the explicit derefs unnecessary.

quinn-h3/src/proto/connection.rs
```diff
@@ -9,15 +9,20 @@ use crate::proto::headers::{self, Header};
 use crate::qpack::{self, DecoderError, DynamicTable, EncoderError, HeaderField};
 use crate::Settings;
 
+#[derive(Clone, Copy)]
+pub enum PendingStreamType {
+    CONTROL = 0,
+    ENCODER = 1,
+    DECODER = 2,
```
Ralith: These should be camel case.

quinn-h3/src/connection.rs
stammw: Ha, nice !

stammw:
Ready for a review! :)

djc:
The "ribase encoder" commit looks a bit anomalous, is that intended to be there? Maybe give it a proper commit message? I'll try to review soon!

stammw:
This one was supposed to be rebased into another. (oops) Done

quinn-h3/src/connection.rs
```diff
@@ -113,155 +181,76 @@ impl ConnectionDriver {
...
+                Poll::Ready(Some(Ok(frame))) => {
+                    match (self.inner.remote_settings().is_some(), self.side, frame) {
+                        (_, _, HttpFrame::Settings(s)) => {
+                            self.inner.set_remote_settings(s);
+                        }
+                        (true, Side::Client, HttpFrame::Goaway(id)) => {
+                            self.inner.leave(StreamId(id));
+                        }
+                        (true, Side::Server, HttpFrame::CancelPush(_)) => {
+                            println!("CANCEL_PUSH frame ignored");
```
djc: Maybe use logging for these instead of `println!()`?

quinn-h3/src/connection.rs
```diff
@@ -16,60 +17,128 @@ use crate::{
...
+impl ConnectionRef {
+    pub fn new(
+        quic: quinn::Connection,
+        side: Side,
+        uni_streams: IncomingUniStreams,
+        bi_streams: IncomingBiStreams,
+        settings: Settings,
+    ) -> Result<Self, ProtoError> {
+        Ok(Self {
+            quic: quic.clone(),
```
djc: Does it really make sense to hold `quic` in the `ConnectionRef` and also the `ConnectionInner`? I'd think that one ought to be enough.

quinn-h3/src/connection.rs
```diff
@@ -168,21 +172,58 @@ impl ConnectionInner {
...
+
+    fn on_uni_resolved(&mut self, new_stream: NewUni) -> Result<(), Error> {
+        match new_stream {
+            NewUni::Control(stream) => match self.recv_control {
+                None => {
+                    self.recv_control = Some(stream);
+                    Ok(())
+                }
+                Some(_) => {
+                    self.set_error(
```
djc: Maybe worth coming up with some helper that prevents having to duplicate the error message twice in all these places? Looks a bit ugly.

quinn-h3/src/connection.rs
```diff
@@ -282,6 +313,42 @@ impl ConnectionInner {
...
+                Poll::Pending => break,
+                Poll::Ready(0) => {
+                    self.set_error(ErrorCode::CLOSED_CRITICAL_STREAM, "encoder closed");
+                    return Err(Error::Peer("encoder stream closed".into()));
+                }
+                Poll::Ready(n) => {
+                    buffer.extend_from_slice(&read_buf[..n]);
+                    let (pos, max_recieved_ref) = {
+                        let mut cur = Cursor::new(&mut buffer);
+                        let max_recieved_ref = self.inner.on_recv_encoder(&mut cur)?;
```
djc: `recieved` -> `received`

quinn-h3/src/connection.rs
```diff
@@ -24,6 +24,7 @@ use crate::{
 };
 
 const RECV_ENCODER_INITIAL_CAPACITY: usize = 20480;
+const RECV_DECODER_INITIAL_CAPACITY: usize = 2048;
```
djc: Please move constants to the bottom of the module (can do it in an additional commit at the end).

quinn-proto/src/varint.rs
```diff
@@ -60,6 +62,18 @@ impl VarInt {
             unreachable!("malformed VarInt");
         }
     }
+
+    /// Length of an encoded value from its first byte
+    pub fn size_encoded(first: u8) -> usize {
```
djc: Let's call this `encoded_size` instead of `size_encoded`, please?

quinn-proto/src/varint.rs
```diff
@@ -60,6 +62,18 @@ impl VarInt {
...
+        let tag = first >> 6;
+        match tag {
+            0b00 => 1,
```
djc: This could be more concisely written as `pow()`. :)

quinn-h3/src/streams.rs
```diff
@@ -83,56 +82,74 @@ impl Future for RecvUni {
...
+            ty,
+            state: SendUniState::New(quic),
+            data: VecDeque::with_capacity(2),
         }
     }
 }
 
-enum SendControlStreamState {
+enum SendUniState {
+    New(quinn::Connection),
```
djc: What is this `New` state for? It seems a bit iffy to me to have this contain a `quic::Connection`.

quinn-h3/src/connection.rs
djc: I saw this got fixed in the other PR anyway, so let's not bother with this.

quinn-h3/src/connection.rs
stammw: I've been bothered by this choice at one moment but after checking again I just figured out something can be done in combination with the `set_error()` helper you proposed.

quinn-proto/src/varint.rs
stammw: haha, oops :sweat_smile:

quinn-h3/src/streams.rs
stammw: This state reflects the moment when one SendStream has not been used. Depending on Peer's settings, it could never get to be used. I was hesitating between:

1. storing an OpenUni and not polling it until there is some data: The problem here is that we rely on the fact that OpenUni does not open a stream until it's polled.

2. having `ConnectionInner::send_uni` be `[ Option<SendUni>, 3 ]`, which would get some parts of the `SendUni` state machine live in `ConnectionInner`, and require a `quinn::Connection` back again in `ConnectionInner`.

I think `1.` would be the best solution ?

quinn-h3/src/connection.rs
```diff
@@ -1,51 +1,63 @@
...
         frame::{HeadersFrame, HttpFrame},
         ErrorCode, StreamType,
     },
     streams::{NewUni, RecvUni, SendUni},
     Error, Settings,
 };
 
+struct DriverError(Error, ErrorCode, String);
+
+macro_rules! driver_err {
```
djc: Why does this need to be a macro?

quinn-h3/src/connection.rs
```diff
@@ -1,51 +1,63 @@
...
+    };
+}
 
 pub struct ConnectionDriver(pub(crate) ConnectionRef);
 
 impl Future for ConnectionDriver {
     type Output = Result<(), Error>;
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
-        let mut conn = self.0.h3.lock().unwrap();
```
djc: Having this change in this commit seems surprising...

quinn-h3/src/connection.rs
stammw: It doesn't have to, I though it would bring call-site brevity.

quinn-h3/src/connection.rs
stammw: I introduced a type `DriverError` to carry an `ErrorCode` along with an `Error`, the former will be sent in the quic connection shutdown, the latter will be returned to the user by the H3 driver.

This is why I introduced an extra function composition.

I could have separated the introduction of this function in a previous commit, along with `ConnectionInner::quic` removal. But it would have added work to struggle with return types in each function, twice.

stammw: Ho, I didn't see your last comment about this. I'll get that sorted out during lunch break.

quinn-h3/src/streams.rs
stammw: Have going for `1.` in e16bfa9  H3: simplify SendUni state machine

djc:
Nice work!

PR #491, opened 2019-11-16
Use sendmmsg to reduce system call overhead

Ralith:
I noticed `sendmsg` was 35% of the flame graph for the large streams benchmark, so I finally had a go at this. It yields a roughly 15-30% improvement in each one of the throughput benchmarks! Oddly, `sendmmsg` is *still* about 35% of the flame graph, and cranking the batch size way up doesn't seem to help any. We'll probably need to dive into libc or the kernel or even investigate [generic segmentation offload](http://vger.kernel.org/lpc_net2018_talks/willemdebruijn-lpc2018-udpgso-paper-DRAFT-1.pdf) or similar to make further headway on this particular bottleneck.

### DIFF
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -252,35 +252,36 @@ impl EndpointInner {
     }
 
     fn drive_send(&mut self, cx: &mut Context) -> Result<bool, io::Error> {
-        let mut sent = 0;
-        while let Some(t) = self
-            .outgoing
-            .pop_front()
-            .or_else(|| self.inner.poll_transmit())
-        {
-            match self
-                .socket
-                .poll_send(cx, &t.destination, t.ecn, &t.contents)
-            {
-                Poll::Ready(Ok(_)) => {}
+        let mut calls = 0;
+        loop {
+            while self.outgoing.len() < crate::udp::BATCH_SIZE {
+                match self.inner.poll_transmit() {
+                    Some(x) => self.outgoing.push_back(x),
+                    None => break,
+                }
+            }
+            if self.outgoing.is_empty() {
+                return Ok(false);
+            }
+            match self.socket.poll_send(cx, self.outgoing.as_slices().0) {
+                Poll::Ready(Ok(n)) => {
+                    self.outgoing.drain(..n);
+                    calls += 1;
+                    if calls == IO_LOOP_BOUND {
+                        return Ok(true);
+                    }
+                }
                 Poll::Pending => {
-                    self.outgoing.push_front(t);
-                    break;
+                    return Ok(false);
                 }
                 Poll::Ready(Err(ref e)) if e.kind() == io::ErrorKind::PermissionDenied => {
-                    self.outgoing.push_front(t);
-                    break;
+                    return Ok(false);
                 }
                 Poll::Ready(Err(e)) => {
                     return Err(e);
                 }
             }
-            sent += 1;
-            if sent == IO_LOOP_BOUND {
-                return Ok(true);
-            }
         }
-        Ok(false)
     }
 
     fn handle_events(&mut self, cx: &mut Context) {
--- a/quinn/src/platform/fallback.rs
+++ b/quinn/src/platform/fallback.rs
@@ -2,20 +2,32 @@ use std::{io, net::SocketAddr};
 
 use mio::net::UdpSocket;
 
-use proto::EcnCodepoint;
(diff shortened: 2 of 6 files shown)


Ralith:
Looks like macOS doesn't expose this; there's [`sendmsg_x`](https://github.com/apple/darwin-xnu/blob/master/bsd/sys/socket.h#L1366-L1396), but the comments describe it as unstable. How do we feel about using that?

Ralith:
Using the fallback for now, since `sendmsg_x` isn't even bound in the libc crate.

Ralith:
Investigation suggests that the remaining `sendmmsg` presence is just how much time it takes to send UDP packets on Linux. Further optimization will have to come from elsewhere, or resort to userspace networking.

quinn/src/endpoint.rs
```diff
@@ -252,35 +252,36 @@ impl EndpointInner {
...
+            while self.outgoing.len() < crate::udp::BATCH_SIZE {
+                match self.inner.poll_transmit() {
+                    Some(x) => self.outgoing.push_back(x),
+                    None => break,
+                }
+            }
+            if self.outgoing.is_empty() {
+                return Ok(false);
+            }
+            match self.socket.poll_send(cx, self.outgoing.as_slices().0) {
```
djc: Are you sure this is correct? What invariant do we have that makes sure all the transmits are in the first slice?

quinn/src/platform/fallback.rs
```diff
@@ -2,20 +2,29 @@ use std::{io, net::SocketAddr};
...
-    ) -> io::Result<usize> {
-        self.send_to(msg, remote)
+    fn send_ext(&self, transmits: &[Transmit]) -> io::Result<usize> {
+        let mut sent = 0;
+        for transmit in transmits {
+            match self.send_to(&transmit.contents, &transmit.destination) {
+                Ok(_) => {
+                    sent += 1;
+                }
+                Err(_) if sent != 0 => {
```
djc: What errors are being swallowed here? Maybe add a comment to clarify why this is okay?

quinn/src/platform/unix.rs
```diff
@@ -71,51 +71,27 @@ impl super::UdpExt for UdpSocket {
...
-            encoder.push(libc::IPPROTO_IPV6, libc::IPV6_TCLASS, ecn);
+    #[cfg(not(target_os = "macos"))]
+    fn send_ext(&self, transmits: &[Transmit]) -> io::Result<usize> {
+        use crate::udp::BATCH_SIZE;
+        let mut msgs: [libc::mmsghdr; BATCH_SIZE] = unsafe { mem::zeroed() };
+        let mut iovecs: [libc::iovec; BATCH_SIZE] = unsafe { mem::zeroed() };
+        let mut cmsgs = [cmsg::Aligned(MaybeUninit::uninit()); BATCH_SIZE];
+        for (hdr, (iov, (ctrl, transmit))) in msgs
+            .iter_mut()
+            .zip(iovecs.iter_mut().zip(cmsgs.iter_mut().zip(transmits)))
```
djc: Isn't there some `flatten()` thing we can use here?

quinn/src/endpoint.rs
Ralith: They don't need to be. `as_slices` returns the slices in order, i.e. `.0` is the first one, and we use `drain` to drop the first `n` that were actually sent. In the common case where the whole batch is sent this never leads to wrapping, but when it does everything still works fine.

quinn/src/platform/unix.rs
Ralith: Not that I know if. `itertools` has `zip4` and such but this doesn't seem bad enough to justify a dep.

quinn/src/platform/fallback.rs
Ralith: Done.

quinn/src/platform/unix.rs
Ralith: On second thought, maybe it would be better to use explicit indexing here?

Ralith: Applied.

Ralith:
`sendmmsg` isn't nearly so prominent in profiles the benchmark introduced in #507 which *should* be pretty similar to the existing "large streams" benchmark, so maybe criterion is confusing things somehow. The performance benefit seems to stand, however, perhaps in part due to fewer calls into the reactor machinery.

PR #495, opened 2019-11-17
Update `futures` and `tokio` dependency.

daxpedda:
This PR is just updating the in the title mentioned dependencies.
Mainly to move away from `futures-preview` and this new version of tokio re-exports it's `Stream` trait, so we don't have to match tokios `futures` dependency version.


Ralith:
Thanks! I take it the objective here is to allow downstream code to use stable futures?

djc:
Won't it be confusing for downstream code if we rely on a different `Stream` trait? I'm not sure how big of a deal with this, given that it should not be much longer before tokio releases 0.2.0 final.

Also we should probably redo the `Pin` and `Context` imports to come directly from `std`?

daxpedda:
> Thanks! I take it the objective here is to allow downstream code to use stable futures?

Actually not sure about stable or not, but at least a more stable version of the `futures` crate.
> Won't it be confusing for downstream code if we rely on a different `Stream` trait? I'm not sure how big of a deal with this, given that it should not be much longer before tokio releases 0.2.0 final.
So the thing is that `tokio` `0.2.0-alpha.5` was using `futures-preview` `0.3.0-alpha.18`, which isn't compatible with `futures` `0.3.1`. So neither we (unless we depend on two different version of `futures`) nor downstream code could use the new version of `futures`.
But `tokio` `0.2.0-alpha.6` is using `futures-preview` `0.3.0-alpha.19` which is compatible with `futures` `0.3.1` and also re-exports its `Stream` trait. So now we and downstream code can use the new version of `futures` and any internal incompatiblities can be resolved by using the re-exported `Stream` trait from `tokio`.

> Also we should probably redo the `Pin` and `Context` imports to come directly from `std`?
I'm on it.

daxpedda:
I assume you meant `Poll` not `Pin`. I also moved to `std` `Waker` and `Future`.

Ralith:
> 0.3.0-alpha.19 which is compatible with futures 0.3.1

Compatible how? It doesn't look like they're using the semver trick.

daxpedda:
> Compatible how? It doesn't look like they're using the semver trick.

Apparently it does not, I was actually checking through the documentation, which apparently generates very wrong version links 😞.

So I double checked now, and the reason it wasn't working is the following:
Internally we had to use the same version of `futures` because we needed the `Stream` trait to use some functions of `tokio`.
But with the newer version `tokio` re-exports it's `Stream` trait, so we can use that and the newest version of `futures`, the user isn't facing `tokio` implementations, but ours, for example the [`Incoming`](https://github.com/djc/quinn/blob/master/quinn/src/endpoint.rs#L348) trait.
So now we can implement the newest `futures` traits while internally using `tokio`'s old re-exported trait.

Sorry for my mistake there, I'm gonna double-check what happened here and probably write an issue report to rust to fix this.

Ralith:
Per your latest comment, is this still specifically useful for you? I'm inclined to wait for the next tokio update so we can move everything over without so much risk for confusion otherwise.

daxpedda:
Moving to `futures` now or when `tokio` updates will not change a lot internally and it will not change anything in user facing code. Because `tokio` will update to the newest `futures` the result should be the same.

> Per your latest comment, is this still specifically useful for you?

I don't really mind, I can use a local copy for now.

I assumed a gain here for `quinn`, updating earlier will lead to less code breaking later on.
All in all this is really a quite minimal change, but gives us a more stable player facing API earlier.

djc:
Let's postpone this until tokio 0.2 is released.

est31:
> Let's postpone this until tokio 0.2 is released.

There's been a release now: https://tokio.rs/blog/2019-11-tokio-0-2/

daxpedda:
I'm working on it right now.
One issue I encountered was the dependency `cfg-if` in `mio` and `tracing`, mio somehow decided to lock it on `=0.1.9`, see https://github.com/tokio-rs/mio/pull/1099.

Considering we can't downgrade `mio` I just downgraded `tracing` to `0.1.9` to match `mio`s `cfg-if` dependency.

est31:
Thanks for pointing to that. I'll file a PR with the fix @cuviper mentioned in that thread.

est31:
Done! https://github.com/tokio-rs/mio/pull/1170

daxpedda:
Thank you!

PR #500, opened 2019-11-18
Improve on some CI issues

djc:
Not quite sure why `local_addr` and `echo_v6` are failing under tarpaulin only...


quinn/benches/bench.rs
```diff
@@ -162,7 +162,7 @@ impl Context {
                         while let Some(_) = stream.read_unordered().await.unwrap() {}
                     }
                 }
-                    .instrument(error_span!("server")),
+                .instrument(error_span!("server")),
```
Ralith: Looks like this makes stable rustfmt sad, so we can't merge as-is.

Ralith:
> Not quite sure why local_addr and echo_v6 are failing under tarpaulin only...

That is very strange; afaict they both bind only the wildcard port, so that error should be impossible if anything else works. Does it happen consistently?

djc:
Ditched the formatting commit, please have another look.

PR #504, opened 2019-11-19
Use anyhow instead of failure in examples

Ralith:
Anyhow was already in our dev-depgraph and is more modern and actively maintained.


djc:
Yeah, sounds good. Needs a rebase now...

Ralith:
Rebased.

PR #505, opened 2019-11-19
Added 'pub (crate)' where possible quinn-proto.

TimonPost:
From the https://github.com/djc/quinn/issues/444 discussion, @Ralith said "I'm not necessarily opposed to it.", which was about improving encapsulation, therefore I decided to create this PR.

**Changes:**
- Added `pub(crate)` where possible `quinn-proto`. 

**Motivation:**
* I always prefer to limit the access and only expose public when needed. 
* All updated types are not used outside the crate and are not re-exported
* Reads cleaner, because the reader can know whether a type is used outside the crate or not
* Programmer does not expose types by accident.

**Prerequisites:**
* No breaking changes.
* Only updated quinn-proto

**Notice:**
* Structs with public fields can only be made `pub (crate)` when the struct it's self can also be `pub(crate)`. Please review and look for changes in encapsulation that do change the fields but not the struct its self.

### DIFF
--- a/quinn-proto/src/assembler.rs
+++ b/quinn-proto/src/assembler.rs
@@ -5,7 +5,7 @@ use bytes::Bytes;
 
 /// Helper to assemble unordered stream frames into an ordered stream
 #[derive(Debug)]
-pub struct Assembler {
+pub(crate) struct Assembler {
     offset: u64,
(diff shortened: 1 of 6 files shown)


Ralith:
Thanks! I definitely see how this can make it easier to make sense of things.

PR #510, opened 2019-11-21
Don't create new connections if we're running low on connection IDs

Ralith:


quinn-proto/src/endpoint.rs
```diff
@@ -353,6 +353,13 @@ where
         remote: SocketAddr,
         server_name: &str,
     ) -> Result<(ConnectionHandle, Connection<S>), ConnectError> {
+        if self.config.local_cid_len <= 4
```
djc: Since this check is moderately complicated and appears twice, maybe factor it out? Also, couldn't we do similar checks for longer `local_cid_len`s?

djc:
Great idea!

quinn-proto/src/endpoint.rs
Ralith: > Since this check is moderately complicated and appears twice, maybe factor it out?

Good thought, done.

> Also, couldn't we do similar checks for longer local_cid_lens?

Those would require more care due to overflow, and would only be useful if someone had more than 3/4 * 1,099,511,627,776 connections. I don't see that happening.

PR #511, opened 2019-11-21
Remove quinn-proto rustls config newtypes

Ralith:
These were not successful at removing rustls from our public API, and made specifying custom cryptographic configurations unintuitive. Removing them reduces the scope of our public API while preserving the friendliness of the high-level quinn API.

### DIFF
--- a/quinn-proto/src/crypto.rs
+++ b/quinn-proto/src/crypto.rs
@@ -20,10 +20,10 @@ use crate::{
 
 /// Cryptography interface based on *ring*
 #[cfg(feature = "ring")]
-pub mod ring;
+pub(crate) mod ring;
 /// TLS interface based on rustls
 #[cfg(feature = "rustls")]
-pub mod rustls;
+pub(crate) mod rustls;
 
 /// A cryptographic session (commonly TLS)
 pub trait Session: Sized {
@@ -87,6 +87,11 @@ pub trait ClientConfig<S>
 where
     S: Session,
 {
+    /// Construct the default configuration
+    fn new() -> Self
+    where
+        Self: Sized;
+
     /// Start a client session with this configuration
     fn start_session(
         &self,
@@ -100,6 +105,11 @@ pub trait ServerConfig<S>
 where
     S: Session,
 {
+    /// Construct the default configuration
+    fn new() -> Self
+    where
+        Self: Sized;
+
     /// Start a server session with this configuration
     fn start_session(&self, params: &TransportParameters) -> S;
 }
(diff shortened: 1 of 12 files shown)


quinn-proto/src/crypto.rs
```diff
@@ -87,6 +87,11 @@ pub trait ClientConfig<S>
 where
     S: Session,
 {
+    /// Construct the default configuration
```
djc: Should we use `Default` for this, instead?

quinn-proto/src/crypto/rustls.rs
```diff
@@ -238,170 +175,23 @@ impl crypto::ClientConfig<TlsSession> for ClientConfig {
...
-    }
-}
-
-impl DerefMut for ServerConfig {
-    fn deref_mut(&mut self) -> &mut Arc<rustls::ServerConfig> {
-        &mut self.0
-    }
-}
-
-/// A single TLS certificate
```
djc: Why did you move these into quinn?

quinn-proto/src/crypto.rs
Ralith: We can't implement `Default` on `Arc<rustls::Client/ServerConfig>`.

quinn-proto/src/crypto/rustls.rs
Ralith: With the newtypes removed from quinn-proto, they're only used in the quinn builders, so they seemed out of place here.

djc: I feel like they should live closer to the rustls implementation bits, so that they could also be swapped out along with them. Do you think that would make sense?

djc: They also seem like things that a quinn-proto user might still find useful if they wanted to bypass the tokio stuff.

Ralith: > I feel like they should live closer to the rustls implementation bits, so that they could also be swapped out along with them.

Other cryptographic protocols (e.g. noise) won't necessarily have analogous abstractions. For the specific case of other TLS implementations, we might prefer to expose a single abstract type that can be consumed by all backends, though that's pretty speculative.

> They also seem like things that a quinn-proto user might still find useful if they wanted to bypass the tokio stuff.

The sole function of these types was to allow certificates/keys to be constructed without leaving the crate and passed into the helper methods that were removed, and the analogous builder methods in quinn which have been retained. While there's nonzero value in providing those helpers to hypothetical quinn-proto users, I think the resulting quinn-proto API was gratuitously weird for both quinn and quinn-proto users who need to access the rustls configuration directly, and maintaining two separate incomplete abstractions over rustls configurations wasn't a great situation.

Meanwhile, the genuine rustls API is perfectly usable. The justification for wrapping it as all is to maximize approachability for a new high-level API user, but at the quinn-proto level I think the calculus is different.

PR #512, opened 2019-11-21
Expose the peer address of a `Connecting`

DemiMarie-parity:


quinn/src/connection.rs
```diff
@@ -106,6 +106,16 @@ impl Future for Connecting {
     }
 }
 
+impl Connecting {
+    /// Get the peer address of a `Connecting`.
+    ///
+    /// This must not be called after `poll` has returned `Ready`, or it will panic.
+    pub fn peer_address(&self) -> SocketAddr {
```
Ralith: `Connection` calls this `remote_address`. I'm not specifically attached to that, but we should be consistent (and probably word their documentation the same, too).

Ralith:
Why was this closed?

DemiMarie-parity:
@Ralith My fault.  I mixed up this PR with #516 and thought it had already been merged.  Oops!

PR #515, opened 2019-11-21
Switch to stable futures

DemiMarie-parity:


djc:
This was just discussed in #495, and we decided to hold off for a while longer.

Also, please keep commits focused -- this seems to contain any number of changes that are unnecessary for the goal of upgrading futures only.

Ralith:
Note that this is *not* necessary for async/await, as illustrated in the examples. The `Future` trait lives in std now so everything's compatible.

DemiMarie-parity:
Tokio does not support futures 0.3.1, so this isn’t useful for now.  Feel free to close it if you so desire.

PR #516, opened 2019-11-21
Fix the hyper upstream repository

### LINKED ISSUE #514: Depends on a nonexistent revision of hyper
This causes `cargo update` to fail.

DemiMarie-parity:
Otherwise `cargo update` fails.

Fixes #514


djc:
Huh, that is weird. Seems like a Cargo bug?

PR #518, opened 2019-11-24
H3 refactor for a simpler public api

stammw:
This PR removes a lot of complexity in the public API, by offering fewer choices for `Request` and `Response` stories. By keeping the `BodyReader` only. 

This is what a full featured client Request looks like now: 

```rust
    let (quic_driver, h3_driver, conn) = client.connect(url)?.await?;

   // spawn drivers

    let request = Request::builder()
        .method(Method::GET)
        .uri(url.path())
        .header("client", "quinn-h3:0.0.1")
        .body(())?;

    let (recv_response, _) = conn.send_request(request).await?;
    let (response, mut recv_body) = recv_response.await?;

    let mut body = Vec::with_capacity(1024);
    recv_body.read_to_end(&mut body).await?;

    if let Some(trailers) = recv_body.trailers().await {
         // ...
    }
    conn.close();
```

This is more or less how `Hyper`'s `h2` is used.

It also includes logging capabilities that were found useful for interop work.


quinn-h3/Cargo.toml
```diff
@@ -35,6 +35,7 @@ string = "0.2"
 tokio = "0.2.0-alpha.5"
 tokio-io = "0.2.0-alpha.5"
 tokio-codec = "0.2.0-alpha.5"
+webpki = "0.21"
```
djc: It doesn't look like this actually gets used?

quinn-h3/src/client.rs
```diff
@@ -86,14 +92,31 @@ pub struct Client {
 }
 
 impl Client {
-    pub fn connect(
+    pub fn connect(&self, url: &Url) -> Result<Connecting, ConnectError> {
```
djc: I'm a bit wary of adding this method, and I feel this doesn't add that much easy of use compared to the previous `connect()` (now `connect_socket()`) method.

stammw:
I'm going to open a discussion issue soon to see what API we want to keep / add / refactor. I've started looking `Hyper` to have more infos on this.

quinn-h3/src/client.rs
stammw: I got those changes removed.

quinn-h3/Cargo.toml
stammw: It's used here: https://github.com/djc/quinn/blob/adcd78aac74d21095cb96a7eed6cee74ce72d854/quinn-h3/src/client.rs#L63

stammw:
I still have to refactor server builder, then I think it will be a good merging point. 

I wonder if we need the same capacity-aware API as `h2`. For the moment I don't see why it would be needed in place of our `AsyncWrite` implementation.

quinn-h3/Cargo.toml
djc: Ah, okay.

djc:
If we're satisfied that this is a good direction, why do you want to hold off on merging this?

stammw:
Just a little more commit and we're good to go. Server builder interface need some trivial work.

djc:
Ah, ok.

stammw:
I should maybe move `BodyReader::poll_read()` into a `Stream` implementation. I'm not sure that we loose any ease of use that way.

stammw:
I think it's ready now.

djc:
Nice!

PR #528, opened 2019-11-27
Update `tokio`, `futures` and `bytes`.

daxpedda:
<details><summary>Old post</summary>I had to put `tracing` to `0.1.9` until https://github.com/tokio-rs/mio/pull/1170 is resolved.

Also https://github.com/carllerche/string/pull/17 and https://github.com/carllerche/string/pull/18 need to pass, but there are probably workarounds for those.
This is also an option: https://github.com/carllerche/string/pull/20.

Examples and tests aren't done yet, ~~I'm still trying to get the `server` <-> `client` example to work~~.

Any help is appreciated!</details>

This updates the following:
-  `tokio`: `0.2.0-alpha.6` -> `0.2.2`
-  `futures`: `0.3.0-alpha.18` -> `0.3.1`
-  `bytes`: `0.4.7` -> `0.5.2`
-  `string`: `0.2` -> `master`
-  `http`: `a3a8fcb213bc456e0b7a42cf0e2bd57afa49851b` -> `43dffa1eb79f6801e5e07f3338fa56191dc454bb`

Tests, examples and benchmarks are minimally changed to keep the PR small.

### DIFF
--- a/bench/Cargo.toml
+++ b/bench/Cargo.toml
@@ -9,7 +9,7 @@ publish = false
 [dependencies]
 quinn = { path = "../quinn" }
 tokio = { version = "0.2.1", features = ["macros"] }
-tracing = "0.1.9"
+tracing = "0.1.10"
 tracing-subscriber = "0.1.5"
 anyhow = "1.0.22"
 rcgen = "0.7"
--- a/bench/Cargo.toml
+++ b/bench/Cargo.toml
@@ -8,7 +8,7 @@ publish = false
 
 [dependencies]
 quinn = { path = "../quinn" }
-tokio = "0.2.1"
+tokio = { version = "0.2.1", features = ["macros"] }
 tracing = "0.1.9"
 tracing-subscriber = "0.1.5"
 anyhow = "1.0.22"
--- a/bench/Cargo.toml
+++ b/bench/Cargo.toml
@@ -8,12 +8,12 @@ publish = false
 
 [dependencies]
 quinn = { path = "../quinn" }
-tokio = { version = "0.2.0-alpha.6", default-features = false, features = ["rt-current-thread"] }
-tracing = "0.1.10"
+tokio = "0.2.1"
+tracing = "0.1.9"
 tracing-subscriber = "0.1.5"
 anyhow = "1.0.22"
 rcgen = "0.7"
-futures = { package = "futures-preview", version = "0.3.0-alpha.18" }
+futures = "0.3.1"
 rustls = "0.16"
 
 [[bin]]
--- a/bench/Cargo.toml
+++ b/bench/Cargo.toml
@@ -8,7 +8,7 @@ publish = false
 
 [dependencies]
 quinn = { path = "../quinn" }
-tokio = { version = "0.2.1", features = ["macros"] }
+tokio = { version = "0.2.2", features = ["macros", "rt-threaded"] }
 tracing = "0.1.10"
 tracing-subscriber = "0.1.5"
 anyhow = "1.0.22"
--- a/bench/Cargo.toml
+++ b/bench/Cargo.toml
@@ -8,7 +8,7 @@ publish = false
 
 [dependencies]
 quinn = { path = "../quinn" }
-tokio = { version = "0.2.2", features = ["macros", "rt-threaded"] }
+tokio = { version = "0.2.2", features = ["rt-core"] }
 tracing = "0.1.10"
 tracing-subscriber = "0.1.5"
 anyhow = "1.0.22"
--- a/bench/src/bulk.rs
+++ b/bench/src/bulk.rs
@@ -3,7 +3,7 @@ use std::time::Instant;
 
 use anyhow::{anyhow, Context, Result};
 use futures::StreamExt;
-use tokio::runtime::current_thread::Runtime;
+use tokio::runtime::Runtime;
 use tracing::trace;
 
 fn main() {
@@ -31,20 +31,19 @@ fn main() {
     drop(endpoint); // Ensure server shuts down when finished
     let thread = std::thread::spawn(|| {
         let mut runtime = Runtime::new().unwrap();
-        runtime.spawn(async {
+        let driver = runtime.spawn(async {
             driver.await.expect("server endpoint driver");
         });
         if let Err(e) = runtime.block_on(server(incoming)) {
             eprintln!("server failed: {:#}", e);
         }
-        runtime.run().expect("server run");
+        runtime.block_on(driver).expect("server run");
     });
 
     let mut runtime = Runtime::new().unwrap();
     if let Err(e) = runtime.block_on(client(server_addr, cert)) {
         eprintln!("client failed: {:#}", e);
     }
-    runtime.run().expect("client run");
 
     thread.join().expect("server thread");
 }
--- a/bench/src/bulk.rs
+++ b/bench/src/bulk.rs
@@ -5,7 +5,7 @@ use anyhow::{anyhow, Context, Result};
 use futures::StreamExt;
 use tracing::trace;
 
-#[tokio::main]
+#[tokio::main(threaded_scheduler)]
 async fn main() {
     tracing::subscriber::set_global_default(
         tracing_subscriber::FmtSubscriber::builder()
@@ -33,13 +33,13 @@ async fn main() {
         let driver = tokio::spawn(async {
             driver.await.expect("server endpoint driver");
         });
-        if let Err(e) = tokio::spawn(server(incoming)).await {
+        if let Err(e) = server(incoming).await {
             eprintln!("server failed: {:#}", e);
         }
         driver.await.expect("server run");
     });
 
-    if let Err(e) = tokio::spawn(client(server_addr, cert)).await {
+    if let Err(e) = client(server_addr, cert).await {
         eprintln!("client failed: {:#}", e);
     }
 
--- a/bench/src/bulk.rs
+++ b/bench/src/bulk.rs
@@ -3,10 +3,10 @@ use std::time::Instant;
 
 use anyhow::{anyhow, Context, Result};
 use futures::StreamExt;
-use tokio::runtime::Runtime;
 use tracing::trace;
 
-fn main() {
+#[tokio::main]
+async fn main() {
     tracing::subscriber::set_global_default(
         tracing_subscriber::FmtSubscriber::builder()
             .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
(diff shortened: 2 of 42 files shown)


djc:
This is awesome work, thanks for doing this. Left some minor nits.

daxpedda:
> This is awesome work, thanks for doing this. Left some minor nits.

Thanks! I still can't get the `server` <-> `client` example to work though 😞.

djc:
> Thanks! I still can't get the server <-> client example to work though 😞.

Do you have some particular problem? You don't even have them compiling yet, or do you? Let me know if I can help.

If you want to push this through before the tracing/cfg-if issue gets resolved, I'd recommend doing the tracing-0.1.9 downgrade into a separate commit and get everything working based on that first.

daxpedda:
> Do you have some particular problem? You don't even have them compiling yet, or do you? Let me know if I can help.

It was compiling on this commit and I found the bug already. So all good.

> If you want to push this through before the tracing/cfg-if issue gets resolved, I'd recommend doing the tracing-0.1.9 downgrade into a separate commit and get everything working based on that first.

Not sure what you mean with a seperate commit?

I'm currently trying to fix all the examples, tests and benchhes, using the new `#[tokio::main]` syntax is a struggle, but sticking with `Runtime` without having `Runtime::run` makes things convoluted. I'm not sure how to do this. I'm really trying to get the most minimal changes going.

djc:
> I'm currently trying to fix all the examples, tests and benchhes, using the new #[tokio::main] syntax is a struggle, but sticking with Runtime without having Runtime::run makes things convoluted. I'm not sure how to do this. I'm really trying to get the most minimal changes going.

I would recommend using the `spawn()` new `JoinHandle` return value with `futures::try_join!()`.

daxpedda:
`futures::try_join!()` only works in async blocks. The problem I am having is that I seem to not be able to use `tokio` without `#[tokio::main]`, I constantly get `no current reactor` otherwise.

And using `#[tokio::main]` is problematic, because there is no way to `await` inside of `thread::spawn` that is used quite a lot in the examples.

I'm afraid unless somebody educates me otherwise I will just start rewriting the tests.

djc:
Don't rewrite the tests/examples just yet, we can probably come up with a good way. I'll see if I can take a look in a bit.

daxpedda:
I got the `bulk`, `interop` binaries and `client`, `server` examples running.
Gonna stop for now until I get some feedback.

Ralith:
mio 0.6.21 has been released, relaxing the dependency constraint.

bench/src/bulk.rs
```diff
@@ -3,10 +3,10 @@ use std::time::Instant;
 
 use anyhow::{anyhow, Context, Result};
 use futures::StreamExt;
-use tokio::runtime::current_thread::Runtime;
 use tracing::trace;
 
-fn main() {
+#[tokio::main(threaded_scheduler)]
```
Ralith: This still needs to be using the single-threaded runtime, with the client and server on distinct threads.

bench/Cargo.toml
```diff
@@ -8,12 +8,12 @@ publish = false
 
 [dependencies]
 quinn = { path = "../quinn" }
-tokio = { version = "0.2.0-alpha.6", default-features = false, features = ["rt-current-thread"] }
+tokio = { version = "0.2.1", features = ["macros"] }
```
Ralith: This is missing the features necessary for `tokio::spawn` to work, at minimum.

daxpedda:
As far as I my research goes is that the problem lies here:
```rust
    pub fn from_std(socket: std::net::UdpSocket) -> io::Result<UdpSocket> {
        let io = mio::net::UdpSocket::from_socket(socket)?;
        io.init_ext()?;
        let io = PollEvented::new(io)?;
        Ok(UdpSocket { io })
    }
```

`PollEvented::new` requires to be executed in an active runtime, meaning that we can't call `EndpointBuilder::bind` outside of a future/task or without having `#[tokio::main]`.

To solutions come to mind:
1.  Not store `PollEvented<UdpSocket>` in `Endpoint` but create it when we need it.
2.  Not rely on `PollEvented` anymore and switch from `mio::net::UdpSocket` to `tokio::io::PollEvented` which already implements `async` on all of it's methods.

I'm currently in favor of 1. and do 2. in a seperate PR, which means that some of it will be reverted again in thhe seperate PR, but thats still simpler.

Ralith:
Thanks for your work so far!

> PollEvented::new requires to be executed in an active runtime, meaning that we can't call EndpointBuilder::bind outside of a future/task or without having #[tokio::main].

There's nothing wrong with using `#[tokio::main]`, in single- or multi-threaded mode as appropriate. Note that it can be used on any function, not just main, so you can e.g. have a synchronous main that kicks off multiple runtimes if necessary. Requiring endpoint construction to take place in the context of a future is fine, IMO, though if necessary I expect [`Runtime::enter`](https://docs.rs/tokio/0.2.1/tokio/runtime/struct.Runtime.html#method.enter) can be used as well.

> Not rely on PollEvented anymore and switch from mio::net::UdpSocket to tokio::io::PollEvented which already implements async on all of it's methods.

I'm having trouble following this, but in tokio 0.2 everything requires an active runtime to register with on construction, so there's no escaping that save for doing initialization lazily, and I'm not sure we should do things differently than tokio wants. We need to implement UDP I/O ourselves to get access to a number of advanced features, so explicit use of `PollEvented` is certainly here to stay regardless.

bench/Cargo.toml
daxpedda: Fixed.

bench/src/bulk.rs
daxpedda: I found this: https://github.com/tokio-rs/tokio/issues/1846.
It seems to be easier to match the performance with using the threaded scheduler, I can still make it single threaded if you wish.

daxpedda:
> Requiring endpoint construction to take place in the context of a future is fine [...]

Alright.
> We need to implement UDP I/O ourselves to get access to a number of advanced features, so explicit use of `PollEvented` is certainly here to stay regardless.

I don't think using `tokio::net::UdpSocket` prevents this. Will give it a quick try.

Have finished fixing all examples, benchmarks and tests and bring them in the same state as they were before this PR, they still require lots of polishing though! I will continue working on that now.

daxpedda:
Just tried using `tokio::net::UdpSocket`, it's possible, but i'm not sure if there are any benefits.
I just don't know enough about `quinn`, or `tokio`, to give any useful opinion here.

daxpedda:
I removed all uses of `#[tokio::main]` now, except in `bulk`. So now the changes are as "minimal" as possible, we can switch to the macro in a different PR.

Two questions remain:
-  What should I do about `bulk`? I can remove `#[tokio::main]` there to, switch to single threaded, or leave it.
-  To make the changes as minimal as possible, I just replaced every instance of `Runtime::run` with a `Runtime::block_on` and gave it a handle of any spawned task. Is there an alternative that we would like to take here?

I'm sorry this has taken me so long, I'm completly new to `tokio`, `quinn` and specifically `async`, it gave me quite the headache to wrap my head around all of those 😄.

quinn/src/tests.rs
```diff
@@ -442,3 +454,19 @@ impl std::io::Write for TestWriter {
...
+    Builder::new()
+        .basic_scheduler()
+        .enable_all()
+        .build()
+        .unwrap()
+}
+
+fn rt_threaded() -> Runtime {
+    Builder::new()
+        .basic_scheduler()
```
Ralith: Copy-paste error?

bench/src/bulk.rs
Ralith: The real goal for this tool is to make it easy to profile quinn itself, and using exactly two threads with no inter-thread communications means fewer variables. I don't expect it to be extremely difficult to move the server's `Endpoint` construction inwards a little, so that seems like an easy win to me.

This could be accomplished by constructing a `std::net::UdpSocket` to get the local address from and passing that into the spawned thread for use with `EndpointBuilder::with_socket`, or by switching the client/server around so the server is constructed on the main thread, its address fetched, and that address passed to the spawned thread for the client to connect to once constructed.

quinn/src/tests.rs
```diff
@@ -129,14 +140,15 @@ fn close_endpoint() {
             }),
     );
     endpoint.close(0u32.into(), &[]);
-    runtime.run().unwrap();
+    runtime.block_on(handle).unwrap();
```
Ralith: It's important that this test, at least, require termination of both spawned tasks, as that's really what it's testing.

quinn/src/tests.rs
```diff
@@ -407,8 +418,9 @@ fn run_echo(client_addr: SocketAddr, server_addr: SocketAddr) {
             assert_eq!(&data[..], b"foo");
             new_conn.connection.close(0u32.into(), b"done");
         });
-    }
-    runtime.run().unwrap();
+        handler
+    };
+    runtime.block_on(handler).unwrap();
```
Ralith: This test should cover clean shutdown by blocking on *all* spawned tasks.

quinn/src/tests.rs
daxpedda: Yes.

quinn/src/tests.rs
daxpedda: Done.

quinn/src/tests.rs
daxpedda: Done.

bench/src/bulk.rs
daxpedda: I solved it differently. As far as I understood the whole performance issue arose because of the runtime splitting between the server and client, so I just moved the runtime that was used to get the socket to the server and made a new runtime for the client.

daxpedda:
> What you've done here is broadly reasonable, but at least some of the tests need to ensure that all spawned tasks are being shut down gracefully; we've had bugs there in the past. I think this could be done with a [`LocalSet`](https://docs.rs/tokio/0.2.2/tokio/task/struct.LocalSet.html) or, perhaps more easily, by `join`ing all spawned handles explicitly.

I hope this is alright. I don't mind making more changes if it is desired, if you prefer `LocalSet` over `join` or if you want me to do it on all tests/examples please say so.

I found a rather simple solution for `bulk`, I hope it is acceptable.

Ralith:
Looks like it needs a rebase.

daxpedda:
Done.

djc:
Thanks again for working on this!

est31:
Wonderful thank you @daxpedda !

@djc @Ralith could we get a new crates.io release with this PR inside?

daxpedda:
> @djc @Ralith could we get a new crates.io release with this PR inside?

Not possible, we have to wait for [`string`](https://github.com/carllerche/string) and [`http`](https://github.com/hyperium/http) to release an updated version first.
You can't release git dependencies to `crates.io`.

djc:
Those are only used in quinn-h3. I think it makes sense to push out releases for quinn and quinn-proto soon.

PR #536, opened 2019-11-29
Renamed single char variables

TimonPost:
Single char variables should not be used except for counters, indexers, and variables like that. Therefore I renamed those to a more meaning full name. Naming variables, even if those are simple, will make code more readable in my opinion, because one does not have to check where the variable comes from to understand the meaning. QUINN has a lot of places were `x`, `n`, `r` was used, sometimes, it is obvious, and sometimes it isn't. I decided to update all of them for the sake of consistency I used this regex `let [a-z] =`, so this does not include other variables than `let ` statements.

If you think, a certain name should be reverted, is not clear enough, is nonsense, please let me know, and I will do the update.

# Changes
- Renamed single-character variable names to have a more descriptive name.
- No changes in logic

### DIFF
--- a/interop/src/main.rs
+++ b/interop/src/main.rs
@@ -310,32 +310,32 @@ fn run(options: Opt) -> Result<()> {
     drop(state); // Ensure the drivers will shut down once idle
     runtime.run().unwrap();
 
-    let r = results.lock().unwrap();
-    if r.handshake {
+    let results = results.lock().unwrap();
+    if results.handshake {
         print!("VH");
     }
-    if r.stream_data {
+    if results.stream_data {
         print!("D");
     }
-    if r.close {
+    if results.close {
         print!("C");
     }
-    if r.resumption {
+    if results.resumption {
         print!("R");
     }
-    if r.zero_rtt {
+    if results.zero_rtt {
         print!("Z");
     }
-    if r.retry {
+    if results.retry {
         print!("S");
     }
-    if r.rebinding {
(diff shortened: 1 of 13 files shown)


Ralith:
Thanks for the PR. Skimming the diff, I see churn but little added clarity. Most cases seem to be values which are consumed within a few lines, hence referencing the definition does not present a problem. If there are specific cases where a non-descriptive name is used and then referenced after a substantial distance, please focus on those.

> Single char variables should not be used except for counters, indexers, and variables like that.

I specifically reject blanket policies like this. Good style is not black-and-white.

I'll leave this open for a day or two in case anyone else wants to weigh in. In the future, it might save effort to inquire whether a subjective style change is desired before writing a patch. Typically we (and, I think, most projects) are more interested in feature work, debugging, testing, etc.

djc:
On the one hand, I totally agree about rejecting blanket policies. On the other, I'm maybe a little bit more sympathetic to the actual contents of this PR (if not as much the tone of the PR description). Trying to place myself in the shoes of someone trying to grok this codebase for the first time, I can see how having the very generic `x` all over the place might be a barrier to entry.

However, apart from the churn this causes and the amount of effort going into what either amounts to a very small improvement to the code base, this will also tend to get into subject territorry fairly easily, which is another problem with low-value changes like this: bikeshedding can end up taking up too much energy compared to the value of the result.

TimonPost:
Oke, I'll close this PR

PR #541, opened 2019-12-03
Upgrade to http 0.2

djc:


stammw:
Nice,  seems to simplify pseudo headers management quite a bit!

Issue #520, opened 2019-11-25
H3: Update to draft-24


stammw:
Here is a list of changes since `daft-20` that relates to our current `h3`
implementation. Non-implemented features are ignored.

## Since draft-ietf-quic-http-22
- [x] Removed priority signaling (#2922,#2924)
- [x] Further changes to error codes (#2662,#2551):
  - Error codes renumbered
  - HTTP_MALFORMED_FRAME replaced by HTTP_FRAME_ERROR, HTTP_ID_ERROR, and others
- [x] Settings identifiers cannot be duplicated (#2979)
  Treat duplicated setting identifier inside a `Settings` frame as `HTTP_SETTINGS_ERROR`
- [x] Clarify that frame types reserved for greasing can occur on any stream, but
  frame types reserved due to HTTP/2 correspondence are prohibited
  (#2997,#2692,#2693)

  Also relates to: Clarify how unknown frame types interact with required frame sequence
  (#2867,#2858)

  - [x] Make sure Reserved and greasing frames are ignored on
    - Control stream
    - Request stream
    - Response stream
  - [x] Trigger stream error `H3_FRAME_UNEXPECTED` when receiving "HTTP/2 where there is no corresponding HTTP/3".

PR #547, opened 2019-12-04
H3 draft 24

stammw:
#520

### DIFF
--- a/quinn-h3/src/body.rs
+++ b/quinn-h3/src/body.rs
@@ -106,25 +106,28 @@ impl BodyReader {
             return Poll::Ready(Some(Ok(data))); // return buffered data in case user called AsyncRead before
         }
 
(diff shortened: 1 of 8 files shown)


quinn-h3/src/body.rs
```diff
@@ -62,6 +62,7 @@ pub struct BodyReader {
     finish_request: bool,
 }
 
+// TODO don't forget to rebase reorder
```
djc: Don't forget!

djc:
And, it should get rebased. Otherwise, looks good to me!

quinn-h3/src/body.rs
stammw: I forgot to put a comment to remind this comment removal.

stammw: There it is !

Issue #524, opened 2019-11-25
Automate interop client running


stammw:
The interop client could be run on all the known implementations at once. 

For example, we could maintain a list of urls in a yaml file, and add a cli flag to `interop/src/main.rs` to make it run tests against each of them, and output a list of results. 

Feedback for new changes would then be a lot faster to obtain. It could also be run by the CI.

djc:
Not sure about running this in CI. Would be nice to have, though.

Can probably just stuff the list of endpoints in a `const` rather than having the YAML, since it's just us changing it.

Ralith:
Hardcoding a list of peers to test sounds like a great idea! I've gotten pretty lazy about clientside interop testing and this would make it trivial to keep up.

Ralith:
This was fixed by #551, right?

stammw:
Yes it was!

PR #551, opened 2019-12-07
Interop client enhancement

stammw:
- [x] Automate interop test to run against all known endpoints at once, when run whitout arguments #524
- [x] Run interop test with HTTP/0.9 / H3 flags to select which tests to execute

If you guys see other subjects to adress, I'd be pleased to implement them.

Do we have to stick with the manual runtime syntax ?
Could it be interesting to run each endpoint interop in parallel ?

### DIFF
--- a/interop/src/main.rs
+++ b/interop/src/main.rs
@@ -1,14 +1,15 @@
 use std::{
     net::{SocketAddr, ToSocketAddrs},
     sync::{Arc, Mutex},
+    time::Duration,
 };
 
-use anyhow::{anyhow, Error, Result};
-use futures::TryFutureExt;
+use anyhow::{anyhow, Result};
+use futures::{future, TryFutureExt};
 use lazy_static::lazy_static;
 use structopt::StructOpt;
-use tokio::{io::AsyncReadExt, runtime::Builder};
-use tracing::{info, warn};
+use tokio::io::AsyncReadExt;
+use tracing::{error, info, warn};
 
 #[derive(StructOpt, Debug)]
 #[structopt(name = "interop")]
@@ -124,7 +125,8 @@ lazy_static! {
     ];
 }
 
-fn main() {
+#[tokio::main]
+async fn main() {
     let opt = Opt::from_args();
     let mut code = 0;
     tracing::subscriber::set_global_default(
@@ -151,7 +153,7 @@ fn main() {
 
     for peer in peers.into_iter() {
         let name = peer.name.clone();
-        if let Err(e) = run(peer, opt.keylog) {
+        if let Err(e) = run(peer, opt.keylog).await {
             eprintln!("ERROR: {}: {}", name, e);
             code = 1;
         }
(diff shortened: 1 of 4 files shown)


djc:
Looks good to me (though I don't know this code very well.

Doing stuff in parallel seems like an obvious win indeed. Definitely don't have to stick with the manual runtime syntax!

(Needs some `cargo fmt` love according to CI.)

interop/src/main.rs
```diff
@@ -1,44 +1,166 @@
...
+    }
+
+    fn hq(mut self) -> Self {
+        self.alpn = Alpn::H3;
+        self
+    }
+}
+
+lazy_static! {
+    static ref PEERS: Vec<Peer> = vec![
```
Ralith: It's not terribly important, but if you modified `Peer` to use `const fn` and store `&str` rather than `String` then this could be `const PEERS: &[Peer<'static>]` instead.

djc:
This is looking great! Looking forward to merging this.

stammw:
Glad you appreciate this!

For the moment, all it's got is:

```
quant:
nghttp2:
mvfst:
picoquic:
msquic:
msquic-hq:
f5:
ATS: VHDCRZS3
quiche-http/0.9:
quiche:
lsquic: VHDCRZSBU3
ngx_quic:
aioquic:
gQuic: VHDCRSB3
```

Most of the empty result will be fixed by getting the interop tool right, hopefully. I guess some endpoints are not configured well for now, and `hq` part of the interop has regressions.

Also, I still have to implement the `d` test, for Dynamic table (right now it runs by default with dynamic QPACK enabled).

djc:
Should probably focus on fixing the `hq` regressions. Let's not block this PR on implementing the `d` test.

stammw:
I'm not sure I'm actually facing regressions. I guess I'll need help on that. Maybe we could merge it if you are both OK with the implementation.

interop/src/main.rs
stammw: I tried a little in this direction, but then thought it wasn't worth to spend time on this. Isn't `const fn` still a nightly thing? Would it fit with constructing a Peer from cli args?

Ralith: `const fn` is stable, it just can't do much yet. Constructing structs is one of the things it can do. For example, the following [works](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e38e803be2d18674cee2857efd924139):
```rust
struct Peer<'a> {
    host: &'a str
}

impl<'a> Peer<'a> {
    const fn new(host: &'a str) -> Self {
        Self { host }
    }
}

const PEERS: &[Peer<'static>] = &[
    Peer::new("foo")
];
```

For CLI-specified peers, you could place the arguments on the stack and end up with a non-static lifetime. If that's not viable, you could instead use `Cow<'static, str>` to handle both cases while still having a 'static struct.

Ralith:
> I'm not sure I'm actually facing regressions

You can test that by running the old interop test on master against a host manually. For example, `cargo run -- quant.eggert.org` yields `VHDCRZSBU`; at a guess there's probably something broken in the hq tests in this PR.

interop/src/main.rs
stammw: Ok, so I tried to write it like that, but I got problems chaining `const fn` functions.
I don't think it worth it to spend time on this for now...

That's been fun to try nevertheless, thanks.

PR #555, opened 2019-12-09
Simplified, synchronous datagram send API

Ralith:
Applications sending datagrams are expected to prioritize freshness, at the cost of reliability if necessary. Blocking datagram sends when the buffer is full is inconsistent with that, particularly if congestion arises due to a low-priority bulk data stream. Instead, we should opt to discard old data in favor of newly sent data. The significant API and implementation simplification (including the capacity to trivially send datagrams from synchronous code!) is a nice bonus.

CC @LaylConway for thoughts, as the current known user of this API. Note that this only changes behavior on congested links.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -2542,9 +2542,6 @@ where
                 self.datagrams.outgoing.push_front(datagram);
                 break;
             }
-            if self.datagrams.outgoing_total >= self.config.datagram_send_buffer_size {
-                self.events.push_back(Event::DatagramSendUnblocked);
-            }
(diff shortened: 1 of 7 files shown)


LaylConway:
I would definitely myself prefer old datagrams being dropped over sending stalling yea.
Especially since a game server is likely to continue generating new datagrams, where the new ones are always more relevant than the old ones. The expectation is already that datagrams may never arrive, if they don't arrive because they're never sent either it doesn't change much for the general case.

Ralith:
Compare also the outcome of a naive UDP transport, wherein an outgoing flood under congested conditions would result in *hopefully* losing old, rather than new, data; here we can at least guarantee that new data takes precedence.

PR #557, opened 2019-12-10
Fix panic in anti-amplification corner case

Ralith:
Hopefully corrects an issue encountered by @stammw in #551. This is a (semi) known issue in the draft 24 RFC, which is expected to be obsoleted by https://github.com/quicwg/base-drafts/pull/3162, but since we're seeing it actually arise in practice this'll keep us working in the spirit of the current draft for now.


stammw:
Thank you for the unblocking !

PR #560, opened 2019-12-13
Fix hang in ReadExact future

Ralith:
A successful read of less than the desired buffer size would result in `Poll::Pending` being returned without a waker having been registered.

This is a fairly typical hand-implemented future bug; mark down another point in favor of switching to anonymous, `!Unpin`, but more obviously correct async/await style implementations.

### DIFF
--- a/quinn/src/streams.rs
+++ b/quinn/src/streams.rs
@@ -547,17 +547,15 @@ impl<'a> Future for ReadExact<'a> {
     type Output = Result<(), ReadExactError>;
     fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
         let this = self.get_mut();
-        let n: usize = ready!(this
-            .stream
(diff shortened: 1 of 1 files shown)


Ralith:
Thanks to @est31 for noticing that something was wrong!

est31:
Thanks @Ralith for finding & fixing the bug!

djc:
Is this worth another maintenance release?

Ralith:
I think that's a good idea, yeah. Sizable footgun.

est31:
@djc indeed I'd love a release with this patch and dd72e10499b78683ac8197bc6085bd72019d635a inside. Because I'm hitting the bug in this PR and I have to workaround missing Eq impls by using if the let trick (but I need != so it's a bit suboptimal to have an empty if let clause followed by else with content).

Ralith:
Folded dd72e10 into #563, we should be able to get that out soon.

PR #568, opened 2019-12-17
Pub transport params

djc:
@Ralith I was trying to figure out how to add doc attributes to your macro contraption (by prepending `#[$doc:meta]` to the pattern lists) but so far didn't succeed. Any clues on this?

Interestingly, `TransportParameters` is already somewhat public (cannot be made `pub(crate)`) since it appears in a public interface, but it cannot be named from external crates as far as I can tell. This makes it available (but not at the top-level, since most implementers shouldn't need it).

Doing this on the maintenance branch first, we can then merge it into master. Going in that direction can be slightly more straightforward than going the other way around.

### DIFF
--- a/quinn-proto/src/lib.rs
+++ b/quinn-proto/src/lib.rs
@@ -25,7 +25,7 @@ mod range_set;
 mod spaces;
 #[cfg(all(test, feature = "rustls"))]
 mod tests;
-mod transport_parameters;
+pub mod transport_parameters;
 mod varint;
 
 pub use varint::{VarInt, VarIntBoundsExceeded};
--- a/quinn-proto/src/transport_parameters.rs
+++ b/quinn-proto/src/transport_parameters.rs
@@ -1,3 +1,11 @@
+//! QUIC connection transport parameters
+//!
+//! The `TransportParameters` type is used to represent the transport parameters
+//! negotiated by peers while establishing a QUIC connection. This process
+//! happens as part of the establishment of the TLS session. As such, the types
+//! contained in this modules should generally only be referred to by custom
+//! implementations of the `crypto::Session` trait.
+
 use std::net::{Ipv4Addr, Ipv6Addr, SocketAddrV4, SocketAddrV6};
 
 use bytes::{buf::ext::BufExt as _, Buf, BufMut};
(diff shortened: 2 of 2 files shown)


quinn-proto/src/transport_parameters.rs
```diff
@@ -39,16 +47,22 @@ macro_rules! apply_params {
 
 macro_rules! make_struct {
     {$($name:ident ($code:expr) = $default:expr,)*} => {
+        /// Transport parameters used to negotiate connection-level preferences between peers
         #[derive(Debug, Copy, Clone, Eq, PartialEq)]
         pub struct TransportParameters {
```
Ralith: Can we make the fields `pub(crate)`? I feel like they're details we don't need to leak.

quinn-proto/src/transport_parameters.rs
```diff
@@ -203,7 +217,7 @@ impl From<UnexpectedEnd> for Error {
 }
 
 impl TransportParameters {
-    pub fn write<W: BufMut>(&self, w: &mut W) {
+    pub(crate) fn write<W: BufMut>(&self, w: &mut W) {
```
Ralith: I think methods to read from `&[u8]`/write to a `Vec<u8>` need to be exposed for the crypto traits to actually be implementable.

quinn-proto/src/transport_parameters.rs
djc: Isn't that the default anyway?

djc: Ah no, never mind this.

quinn-proto/src/transport_parameters.rs
djc: You're right, of course. Fixed.

quinn-proto/src/transport_parameters.rs
```diff
@@ -39,17 +47,23 @@ macro_rules! apply_params {
...
+            /// Maximum size for datagram frames
+            pub(crate) max_datagram_frame_size: Option<VarInt>,
 
             // Server-only
-            pub original_connection_id: Option<ConnectionId>,
-            pub stateless_reset_token: Option<ResetToken>,
-            pub preferred_address: Option<PreferredAddress>,
+            /// The DCID from the first Initial packet; must be included if sent in a Retry packet
+            pub(crate) original_connection_id: Option<ConnectionId>,
+            /// Token used to verify a stateless reset from the client
```
Ralith:
```suggestion
            /// Token used by the client to verify a stateless reset from the server
```

quinn-proto/src/transport_parameters.rs
```diff
@@ -39,17 +47,23 @@ macro_rules! apply_params {
...
+            pub(crate) stateless_reset_token: Option<ResetToken>,
+            /// The server's preferred address for communication after the handshake completion
```
Ralith:
```suggestion
            /// The server's preferred address for communication after handshake completion
```

quinn-proto/src/transport_parameters.rs
```diff
@@ -114,6 +128,9 @@ impl TransportParameters {
     }
 }
 
+/// A server's preferred address
+///
+/// This is communicated as a transport parameter during TLS session establishment.
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub struct PreferredAddress {
```
Ralith:
```suggestion
pub(crate) struct PreferredAddress {
```

quinn-proto/src/transport_parameters.rs
```diff
@@ -176,20 +193,20 @@ impl PreferredAddress {
     }
 }
 
+/// Errors encountered while decoding `TransportParameters`
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Error)]
 pub enum Error {
```
Ralith:
```suggestion
pub(crate) enum Error {
```

quinn-proto/src/lib.rs
```diff
@@ -25,7 +25,7 @@ mod range_set;
 mod spaces;
 #[cfg(all(test, feature = "rustls"))]
 mod tests;
-mod transport_parameters;
+pub mod transport_parameters;
```
Ralith: Since we don't need to expose `PreferredAddress`, I think we can just reexport the type rather than the whole module.

djc: As mentioned in the PR description, I wanted to avoid exporting `TransportParameters` at the top level, since most API users won't be needing it anyway.

quinn-proto/src/transport_parameters.rs
djc: Actually I don't think this is allowed since `Error` appears in the `read()` result type.

quinn-proto/src/lib.rs
Ralith: I think exporting one type is preferable to exporting one module containing one type regardless--but this turned out to be moot anyway.

quinn-proto/src/transport_parameters.rs
Ralith: Whoops, good catch.

djc:
Still needs a solution for macro-generated doc attrs. If you can point me at an example, I can probably figure it out?

Ralith:
Whoops, sorry, slipped my mind. Will have a closer look tomorrow.

quinn-proto/src/transport_parameters.rs
Ralith: This still applies, I think?

Ralith:
Added, though we're not exposing these fields anyway.

Ralith:
Good to merge as soon as the preferred address visibility is corrected.

PR #570, opened 2019-12-19
H3 interop fixes

stammw:
Fixes needed to get better results at interop, along with some tooling and tracing.

### DIFF
--- a/quinn-h3/examples/h3_server.rs
+++ b/quinn-h3/examples/h3_server.rs
@@ -61,14 +61,15 @@ async fn main() -> Result<()> {
     println!("server listening");
     while let Some(connecting) = incoming.next().await {
         println!("server received connection");
-        let connection = connecting
-            .await
-            .map_err(|e| error!("accept failed: {:?}", e));
-
-        if let Ok(connection) = connection {
-            let _ = handle_connection(connection)
-                .await
-                .map_err(|e| error!("handling connection failed: {:?}", e));
+        match connecting.await {
+            Err(e) => error!("accept failed: {:?}", e),
+            Ok(connection) => {
+                let _ = tokio::spawn(async move {
+                    if let Err(e) = handle_connection(connection).await {
+                        error!("handling connection failed: {:?}", e);
+                    }
+                });
+            }
         }
     }
 
@@ -80,7 +81,7 @@ async fn handle_connection(conn: (QuicDriver, ConnectionDriver, IncomingRequest)
 
     tokio::spawn(async move {
         if let Err(e) = h3_driver.await {
-            eprintln!("quic connection driver error: {}", e)
+            eprintln!("h3 connection driver error: {}", e)
         }
     });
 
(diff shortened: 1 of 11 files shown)


djc:
LGTM (pending test fixes).

stammw:
I think this is ready now.

quinn-h3/src/connection.rs
```diff
@@ -103,6 +103,7 @@ pub(crate) struct ConnectionInner {
 
 impl ConnectionInner {
     fn drive(&mut self, cx: &mut Context) -> Result<bool, DriverError> {
+        self.poll_recv_control(cx)?;
         self.poll_incoming_uni(cx)?;
         self.poll_send(cx)?;
         self.poll_recv_control(cx)?;
```
djc: Should we poll this twice?

quinn-h3/examples/h3_server.rs
```diff
@@ -62,12 +63,13 @@ async fn main() -> Result<()> {
         println!("server received connection");
         let connection = connecting
             .await
-            .map_err(|e| anyhow!("accept failed: {:?}", e))
-            .expect("server failed");
+            .map_err(|e| error!("accept failed: {:?}", e));
```
djc: Why do we no longer fail if the connection establishment fails? Perhaps we should at least log/print something in this case?

quinn-h3/src/proto/headers.rs
```diff
@@ -273,13 +273,15 @@ impl Pseudo {
             },
         );
 
+        let len = if authority.is_some() { 1 } else { 0 };
```
djc: Maybe just put the `3 + ` in here?

quinn-h3/src/proto/connection.rs
```diff
@@ -329,7 +329,13 @@ mod tests {
     fn decode_header() {
         let mut header_map = HeaderMap::new();
         header_map.append("hello", HeaderValue::from_static("text/html"));
-        let header = Header::request(Method::GET, Uri::default(), header_map);
+        let header = Header::request(
+            Method::GET,
+            // NOTE: H3 adds a trailing `/`, so the one in following the url is important
+            //       only to make the `Header` comparaison succeed at the end of this test.
```
djc: Typo in `comparaison`

quinn-h3/examples/h3_server.rs
stammw: We want to continue accepting the following incoming connections, so we don't want to fail `main()`. 

I guess there is something logged with the `error!()` call. 

Writing this with `if let` is a bit convoluted. I'll make this a `match`, will read much better.

djc: Ah, if `error!()` logs then we should be fine.

quinn-h3/src/proto/connection.rs
stammw: Fixed.

quinn-h3/src/proto/headers.rs
stammw: oh yeah, thanks

quinn-h3/src/connection.rs
stammw: Got it better in a separate commit. Polling control upon opening. It's important to use the settings (first frame on control stream) as soon as possible as it contains compression enablement parameters.

quinn-h3/examples/h3_server.rs
stammw: Rewrote it using `match`

PR #572, opened 2019-12-20
Switch CI to github actions

Ralith:
It's a pain how many clicks it takes to get any info out of azure pipelines whatsoever, so let's see how this compares.


Ralith:
Ugh, cargo is still garbage at dealing with features in workspaces :(

Ralith:
Coverage claims to be working, but I'm not sure how to get it to produce a report.

djc:
What about Windows and Mac?

PR #576, opened 2019-12-22
Fix remaining lints

djc:


djc:
Okay, that was a slog. Maybe someone can look into that `AtomicBool` thing in `interop`, but I ran out of steam after this, and I'm neither familiar with that code nor with `AtomicBool`.

djc:
(Also fixing these being a slog is a pretty good reason to maintain them through CI. I just hope there's not too much churn.)

PR #582, opened 2020-01-06
Simplified timers and a bugfix

### LINKED ISSUE #574: Expose which timers are idle and loss detection
The documentation for `quinn_proto::generic::Connection` indicates that certain timers have special meaning.  I currently use an ugly hack (hard-coding numbers) to determine which timer is which.  Applications should be able to tell if a timer is the idle or loss-detection timer.

Ralith:
Fixes #574.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -25,7 +25,7 @@ use crate::{
     },
     spaces::{CryptoSpace, PacketSpace, Retransmits, SentPacket},
     streams::{self, FinishError, ReadError, Streams, UnknownStream, WriteError},
-    timer::{Timer, TimerKind, TimerTable},
+    timer::{Timer, TimerTable},
     transport_parameters::{self, TransportParameters},
     Dir, Frame, Side, StreamId, Transmit, TransportError, TransportErrorCode, VarInt,
     MAX_STREAM_COUNT, MIN_INITIAL_SIZE, MIN_MTU, REM_CID_COUNT, RESET_TOKEN_SIZE,
@@ -115,6 +115,7 @@ where
     first_1rtt_sent: Option<u64>,
     /// Sequence number of the first remote CID that we haven't been asked to retire
     first_unretired_cid: u64,
+    timers: TimerTable,
 
     //
     // Queued non-retransmittable 1-RTT data
@@ -236,6 +237,7 @@ where
             idle_timeout: config.idle_timeout,
             first_1rtt_sent: None,
             first_unretired_cid: 0,
+            timers: TimerTable::default(),
 
             path_challenge_pending: false,
             path_response: None,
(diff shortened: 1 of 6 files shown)


quinn-proto/src/connection.rs
```diff
@@ -522,35 +519,46 @@ where
...
-            }
-            TimerKind::PathValidation => {
-                debug!("path validation failed");
-                self.path_challenge = None;
-                self.path_challenge_pending = false;
-                if let Some(prev) = self.prev_path.take() {
-                    self.path = prev;
+    pub fn handle_timeout(&mut self, now: Instant) {
+        for &timer in &Timer::VALUES {
+            match self.timers[timer] {
```
djc: Style nit: maybe an `if let .. else` would be a more readable way of expressing this?

djc: The double match also feels suboptimal, but I can't think of a better way right now.

quinn-proto/src/connection.rs
```diff
@@ -568,8 +576,7 @@ where
                 .expect("update not acknowledged yet")
                 .1
         };
-        self.io
-            .timer_start(TimerKind::KeyDiscard, start + self.pto() * 3);
+        self.timers[Timer::KeyDiscard] = Some(start + self.pto() * 3);
```
djc: Should we make the `timers[foo] = Some(..)` and `timers[foo] = None` a bit more abstract, along the lines of `timers.arm(foo, ..)`/`timers.reset(foo)` (bikeshed names here)?

quinn-proto/src/timer.rs
```diff
@@ -23,58 +19,51 @@ pub(crate) enum TimerKind {
...
+        Timer::KeyDiscard,
+        Timer::PathValidation,
+        Timer::KeepAlive,
     ];
 }
 
 /// A table of data associated with each distinct kind of `Timer`
 #[derive(Debug, Copy, Clone, Default)]
-pub struct TimerTable<T> {
+pub(crate) struct TimerTable<T> {
```
djc: Is the generic `T` here still pulling its weight?

quinn/src/connection.rs
```diff
@@ -660,49 +657,30 @@ impl ConnectionInner {
...
-                    }
-                    Poll::Pending => {}
+        if self.timer.is_none() {
+            self.timer = self
+                .inner
+                .poll_timeout()
+                .map(|x| delay_until(TokioInstant::from_std(x)));
+        }
+        while let Some(ref mut delay) = self.timer {
+            match delay.poll_unpin(cx) {
```
djc: I think `if let` could be nicer here for the early return -- since we don't need the value from `Ready`, we can dedent the `Ready` block contents by two levels.

quinn/src/connection.rs
```diff
@@ -660,49 +657,30 @@ impl ConnectionInner {
...
-                    timer,
-                    update: proto::TimerSetting::Stop,
-                } => {
-                    if self.timers[timer].take().is_some() {
-                        trace!("{:?} timer stop", timer);
+                Poll::Ready(()) => {
+                    self.inner.handle_timeout(now);
+                    // Timeout call may have queued sends
+                    keep_going = true;
+                    match self.inner.poll_timeout() {
```
djc: I think an `if let` might also be better here.

quinn-proto/src/connection.rs
```diff
@@ -122,6 +121,7 @@ where
     //
     path_challenge_pending: bool,
     path_response: Option<PathResponse>,
+    close: bool,
```
djc: Nice commit splitting!

quinn-proto/src/connection.rs
Ralith: This worked out really nicely, good call.

quinn-proto/src/connection.rs
Ralith: Reprised.

quinn-proto/src/timer.rs
Ralith: Definitely not.

Ralith:
Resolved comments, resulting in another 80 lines or so saved!

PR #583, opened 2020-01-06
Opaque config structs

### LINKED ISSUE #444: Consider making the config structs opaque

Ralith:
Fixes #444.

### DIFF
--- a/quinn-proto/src/shared.rs
+++ b/quinn-proto/src/shared.rs
@@ -41,13 +41,13 @@ pub struct TransportConfig {
 
     pub(crate) max_tlps: u32,
     pub(crate) packet_threshold: u32,
-    pub(crate) time_threshold: u16,
+    pub(crate) time_threshold: f32,
     pub(crate) initial_rtt: Duration,
 
     pub(crate) max_datagram_size: u64,
     pub(crate) initial_window: u64,
     pub(crate) minimum_window: u64,
-    pub(crate) loss_reduction_factor: u16,
+    pub(crate) loss_reduction_factor: f32,
     pub(crate) persistent_congestion_threshold: u32,
     pub(crate) keep_alive_interval: Option<Duration>,
     pub(crate) crypto_buffer_size: usize,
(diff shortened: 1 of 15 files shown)


quinn-proto/src/shared.rs
```diff
@@ -281,40 +280,13 @@ impl Default for TransportConfig {
...
-        Ok(())
-    }
-}
-
 /// Global configuration for the endpoint, affecting all connections
 ///
 /// Default values should be suitable for most internet applications.
-#[derive(Clone)]
-pub struct EndpointConfig {
+pub struct EndpointConfig<S>
```
djc: So this is where we lose `Clone` on `EndpointConfig`, which I think I added for hyper compat... I think that's the reason I moved `HmacKey` et al out of here in the first place.

quinn-proto/src/shared.rs
```diff
@@ -44,7 +50,7 @@ pub struct TransportConfig {
     pub(crate) minimum_window: u64,
     pub(crate) loss_reduction_factor: u16,
     pub(crate) persistent_congestion_threshold: u32,
-    pub(crate) keep_alive_interval: u32,
+    pub(crate) keep_alive_interval: Option<Duration>,
```
djc: I like this, although of course it does make `no_std` harder.

quinn-proto/src/shared.rs
Ralith: I solved this by wrapping the `HmacKey`s in `Arc` and adding an explicit `Clone` impl since `derive(Clone)` is pretty dumb about type parameters.

quinn-proto/src/shared.rs
Ralith: Considering the size and number of barriers in the way of `no_std` I'm inclined to not let it get in the way of correctness/ergonomic wins. We can `cfg` in an alternative representation and add a lowest-common-denominator setter (say, `fn keep_alive_interval_ms(&mut self, value: u32)`) if it ends up being needed.

djc: Agreed.

djc:
I'm +1 on this direction!

Ralith:
I think that accounts for all the most obnoxious parts.

Ralith:
I feel like coveralls shouldn't be failing the CI here.

djc:
This needs rebasing.

Ralith:
Rebased.

PR #584, opened 2020-01-06
H3 interop server and assorted fixes

stammw:
I've been working on this `H3` interop server to deploy on h3.stammw.eu.

I'm not sure we'll want to merge this into quinn as it may be out of scope?

The reasons why I think we need to separate h3 interop from the example code are:
- Interop already needs / will need a different set of features. For example transfer tests are based on urls like `https://example.com/100M`
- The example code will be kept simple
- It may be interesting to implement remote logging for people running their interop client against this endpoint. They could filter by IP or CID with a simple web app served via h1/h2.

~For the moment I'm stuck on a bug where requesting more than 1,6M  results in the client missing the tail of a `Data` frame. Receiving the start of a fresh Data frame instead.~

This PR also contains fixes for critical bugs in:
- data writing management
- parsing incomplete frame headers

### DIFF
--- a/quinn-h3/src/frame.rs
+++ b/quinn-h3/src/frame.rs
@@ -183,7 +183,7 @@ impl Error {
         match self {
             Error::Io(_) => ErrorCode::GENERAL_PROTOCOL_ERROR,
             Error::Proto(frame::Error::Settings(_)) => ErrorCode::SETTINGS_ERROR,
-            Error::Proto(frame::Error::UnsupportedFrame) => ErrorCode::FRAME_UNEXPECTED,
+            Error::Proto(frame::Error::UnsupportedFrame(_)) => ErrorCode::FRAME_UNEXPECTED,
             Error::Proto(_) => ErrorCode::FRAME_ERROR,
         }
     }
--- a/quinn-h3/src/frame.rs
+++ b/quinn-h3/src/frame.rs
@@ -145,33 +145,51 @@ impl Future for WriteFrame {
                 WriteFrameState::Header(ref mut send, ref mut h) => {
                     let wrote = ready!(Pin::new(send).poll_write(cx, h))?;
                     h.advance(wrote);
-                    if h.is_empty() {
-                        self.state = match mem::replace(&mut self.state, WriteFrameState::Finished)
-                        {
-                            WriteFrameState::Header(s, _) => {
-                                WriteFrameState::Payload(s, self.payload.take().unwrap())
-                            }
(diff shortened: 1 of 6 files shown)


Ralith:
> I'm not sure we'll want to merge this into quinn as it may be out of scope?

I think it's appropriate for this to live in the repo so long as people are performing interop tests that benefit from targeted support, which may be indefinitely.

I strongly agree with making this separate from the canonical example server, and I look forward to simplifying the example server (perhaps turning it into an echo server) once this is ready and we can switch over to it for formal testing.

djc:
Fully agree with @Ralith's points.

stammw:
Ready for a review.

I'll get this up and running on my box. Then I think we could publish the endpoint on slack ?

stammw:
Ho, rerty and ZeroRTT aren't actually working...

quinn-h3/src/frame.rs
```diff
@@ -158,12 +158,14 @@ impl Future for WriteFrame {
                 WriteFrameState::Payload(ref mut send, ref mut p) => {
                     let wrote = ready!(Pin::new(send).poll_write(cx, p))?;
                     p.advance(wrote);
-                    let send = match mem::replace(&mut self.state, WriteFrameState::Finished) {
-                        WriteFrameState::Payload(s, _) => s,
-                        _ => unreachable!(),
-                    };
-                    self.state = WriteFrameState::Finished;
-                    return Poll::Ready(Ok(send));
+                    if p.is_empty() {
```
djc: Both here and in the preceding block, it might help readability to invert the `is_empty()` condition and do an early return if not empty (this can be a follow-up commit).

djc: Also, how hard would it be to add a test that covers this?

stammw: It is doable I guess, making a trait out of the used methods of `SendStream` in `WriteFrame`, then using `WriteFrame<T: SendStreamTrait>`. So the tests can pass a write-to-buffer implementation. Some parameters for this mocked impl could then be used to define some `SendStream` desired testing behavior. I'm not sure how hard can this be.

There might be simpler solutions?

Testing all edge cases will be a must-have in the future of this crate .

stammw: Do you want those tests included with this PR?

djc: Nah, but maybe write an issue about it so we don't forget if you don't intend to do it nowish?

PR #586, opened 2020-01-07
Spawn driver tasks implicitly

Ralith:
We're already reliant on tokio's thread locals for UDP socket and timer registration, and [`Handle::enter`](https://docs.rs/tokio/0.2.6/tokio/runtime/struct.Handle.html#method.enter) provides adequate mechanism to route spawns to a particular runtime if absolutely needed, so passing driver tasks to the user explicitly wasn't buying us much considering how big a footgun it is.

### DIFF
--- a/bench/src/bulk.rs
+++ b/bench/src/bulk.rs
@@ -25,7 +25,7 @@ fn main() {
     let mut endpoint = quinn::EndpointBuilder::default();
     endpoint.listen(server_config.build());
     let mut runtime = rt();
-    let (driver, endpoint, incoming) = runtime.enter(|| {
+    let (endpoint, incoming) = runtime.enter(|| {
         endpoint
             .bind(&SocketAddr::new(IpAddr::V6(Ipv6Addr::LOCALHOST), 0))
             .unwrap()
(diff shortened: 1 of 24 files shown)


Ralith:
I'm fairly confident that this is the direction I'd like to go. Just need to update more tests and, at least ideally, find a way to test that the implicitly spawned tasks get cleaned up properly.

djc:
Looks okay to me. I'll give it a full review once it's complete.

Ralith:
Rebased onto #589 and #590, plus a couple additional bugfixes. I think this is ready now, modulo the dependency PRs being merged.

Ralith:
Er, right, still need to port h3.

Ralith:
Okay, *now* we're good. This is still missing checks that we're not leaking tasks, but I'm not sure if there's a good way to do that with tokio 0.2 yet.

Ralith:
On consideration, I think I can refactor the whole connection setup path to be a lot more straightforward; going to try to get that in before merging.

Ralith:
Okay, I think this is ready.

Ralith:
Rebased.

djc:
This needs one more rebase...

stammw:
🎉

PR #591, opened 2020-01-09
Compare reset tokens in constant time

### LINKED ISSUE #585: Stateless reset token comparison should be constant-time
Our current implementation is likely vulnerable to a timing attack as discussed in [the current draft](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#name-detecting-a-stateless-reset). The easiest fix is to drop in a constant time comparison. Is there a popular crate for such that's a suitable hard dep for quinn-proto? We went to some effort to make `ring` not strictly required already.

Ralith:
Fixes #585. Opted to inline the constant time eq comparison based on the (CC0) constant_time_eq crate in the spirit of keeping the depgraph down.


djc:
LGTM, just need to silence clippy here.

Ralith:
Tidied.

PR #592, opened 2020-01-09
Remove coverage from github actions

Ralith:
This is mostly just me expressing frustration at all of tonight's PRs being marked red due to incidental coverage churn. Hopefully there's a better way.


djc:
I've tweaked the coveralls settings to not have a minimum coverage and to allow coverage changes of up to -2%. Let's see if subsequent runs are better.

Ralith:
Thanks!

PR #594, opened 2020-01-11
Expose the underlying TLS session

DemiMarie-parity:


djc:
I think the method should be called `crypto_session()`, but otherwise I'm okay with this.

djc:
@zeroqn would this also satisfy your use case?

zeroqn:
Yes. Thanks.

PR #596, opened 2020-01-13
Expose client-supplied server name

### LINKED ISSUE #595: How would one handle SNI with multiple domains?
Hello,

I am expressing big interest in your crate and have vaguely gone through the documentation and examples, but one thing is not very clear for me. Usually, one might want to create a web server which handles requests for multiple domains, each with their own certificate, but I cannot see a way of doing this with `quinn`, or am I wrong?

Based on your server example, you first create the `ServerConfig`: 
```if let (Some(key_path), Some(cert_path)) = (&options.key, &options.cert) {
... server_config.certificate(quinn::CertificateChain::from_certs(vec![cert]), key)?;
    }
```

And then you create the `Endpoint` using that config:
```
let mut endpoint = quinn::Endpoint::builder();
    endpoint.listen(server_config.build());
```

And finally you instantiate the listener and start accepting new connections: 
```
let mut runtime = Builder::new().threaded_scheduler().enable_all().build()?;

    let (endpoint_driver, mut incoming) = {
        let (driver, endpoint, incoming) = runtime.enter(|| endpoint.bind(&options.listen))?;
        info!("listening on {}", endpoint.local_addr()?);
        (driver, incoming)
    };
runtime.spawn(async move {
        while let Some(conn) = incoming.next().await {
...
handle_connection(root.clone(), conn).unwrap_or_else(move |e| {
                    error!("connection failed: {reason}", reason = e.to_string())
                }),... Ok(())
}
```

And then you handle the streams with the `handle_connection fn`:
```
async fn handle_connection(root: Arc<Path>, conn: quinn::Connecting) -> Result<()> {
    let quinn::NewConnection {
        driver,
        connection,
        mut bi_streams,
        ..
    } = conn.await?;
...
```

All of the above has been taken from your `server.rs` example. So, my question is, with multiple domains you can only know the SNI hostname once the connection has been created and socket established; with your crate, how would one go about doing TLS handshake with the appropriate SSL Certificate for that SNI Hostname (domain) once the connection is established, and not beforehand like it is now?

With Rustls and Tokio you first spawn the `TcpListener`, then start accepting connections over `TcpStream` and then you create a `TlsSession` from that stream, and from then you can read the encrypted data in plaintext through that `TlsSession` and write back to it to the encrypted stream. All of this happens once the `TcpSocket` is created and if one would like, they could attach a different SSL Certificate & Key pair per `TcpStream`.

Maybe I've got something wrong or missed something, but how would one accomplish the same thing with your crate? Any assistance/guidelines/examples would be highly appreciated!

Ralith:
Fixes #595.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -325,6 +325,13 @@ impl Connection {
         self.0.lock().unwrap().inner.protocol().map(|x| x.into())
     }
 
+    /// The server name specified by the client
+    ///
+    /// Returns `None` for outgoing connections.
+    pub fn server_name(&self) -> Option<String> {
+        self.0.lock().unwrap().inner.server_name().map(|x| x.into())
+    }
+
     // Update traffic keys spontaneously for testing purposes.
     #[doc(hidden)]
     pub fn force_key_update(&self) {


quinn/src/connection.rs
```diff
@@ -325,6 +325,13 @@ impl Connection {
         self.0.lock().unwrap().inner.protocol().map(|x| x.into())
     }
 
+    /// The server name specified by the client
```
djc: Should we replace these methods with a `crypto_session()` accessor at the quinn layer, as well? I feel like abstracting over this isn't working out so well for us.

Ralith: That's appealing, but because `quinn::Connection` is `Send + Clone`, access to the interior of a `proto::Connection` must always involve a lock, so we can't expose a reference directly, and returning a lock guard is both a bit weird and a substantial footgun as it would block other operations on the connection.

We could do `with_session<T>(&self, f: impl FnOnce(&S) -> T) -> T` but that's also pretty weird. Conversely, even a hypothetical cryptographic protocol that doesn't support ALPN and/or SNI can harmlessly return `None` for them, so this approach seems the most practical. I'm open to other ideas, though.

We *could* still get rid of the accessors on `proto::Connection`, including in the crypto traits, and rely on the already-hardcoded dependency on TLS in the high level API, but we'd need to expose more of `proto::crypto::rustls` for that to be viable, as `get_sni_hostname` is defined only on `rustls::ServerSession`, not `rustls::Session`, and references to the former cannot currently be constructed externally.

djc: Ah, fair point. Not sure all that is more appealing than what we've got today...

PR #597, opened 2020-01-13
Switch back to tarpaulin & codecov

djc:


djc:
Debugging this here: https://github.com/actions-rs/tarpaulin/issues/2

djc:
I don't really remember why codecov vs coveralls. tarpaulin seems to be a bit faster than grcov, at least based on one data point comparing two PRS.

PR #598, opened 2020-01-14
Defragment incoming stream chunks (fixes #431)

### LINKED ISSUE #431: Don't allow small units of application data to keep large packets in memory
Quinn uses zero-copy refcounted `Bytes` for most processing of incoming data. At the extreme, a hostile peer could send numerous stream frames containing a single byte in a packet padded to MTU size. We could combat this by heuristically opting to copy segments of application data into freshly allocated memory when they fall below a certain proportion of the packet, or when a packet consists of a small proportion of application data, or similar.

djc:
Here's an attempt at solving that old memory fragmentation issue. The 32 threshold is just a simple heuristic: in the worst case we accept 32 1-byte chunks from 32 different datagrams, 32k of unnecessary allocations, but hopefully most applications would not get up to 32 chunks in the buffer so that they would not trigger the (relatively slow) defragmentation code.

Let me know what you think!

### DIFF
--- a/quinn-proto/src/assembler.rs
+++ b/quinn-proto/src/assembler.rs
@@ -1,12 +1,13 @@
-use std::{cmp::Ordering, collections::BinaryHeap};
+use std::{cmp::Ordering, collections::BinaryHeap, mem};
 
-use bytes::{Buf, Bytes};
+use bytes::{Buf, Bytes, BytesMut};
 
 /// Helper to assemble unordered stream frames into an ordered stream
 #[derive(Debug)]
 pub(crate) struct Assembler {
     offset: u64,
     data: BinaryHeap<Chunk>,
+    defragmented: usize,
 }
 
 impl Assembler {
@@ -14,6 +15,7 @@ impl Assembler {
         Self {
             offset: 0,
             data: BinaryHeap::new(),
+            defragmented: 0,
         }
     }
 
@@ -22,6 +24,7 @@ impl Assembler {
         loop {
             if self.consume(buf, &mut read) {
                 self.data.pop();
+                self.defragmented = self.defragmented.saturating_sub(1);
             } else {
                 break;
             }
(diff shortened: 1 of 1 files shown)


quinn-proto/src/assembler.rs
```diff
@@ -71,6 +74,32 @@ impl Assembler {
         }
     }
 
+    // Copy the buffered chunk data to new chunks backed by a single buffer to
+    // make sure we're not unnecessarily holding on to many larger allocations.
+    // Merge contiguous chunks in the process of doing so. Reset the `defragmented`
+    // counter to the new number of chunks left in the heap so that we can decide
+    // when to defragment the queue again if necessary.
```
Ralith: Should this skip previously-defragmented chunks? Little or no space is saved by reprocessing them, and it might substantially reduce the amount of work required by a highly fragmented stream with a large buffer and a slow application, since the unread prefix would otherwise be copied on each defragment pass. Should be easy to do by e.g. `take`ing the first `self.defragmented` elements from `old.into_sorted_vec().into_iter()` before proceeding.

quinn-proto/src/assembler.rs
```diff
@@ -89,6 +118,9 @@ impl Assembler {
 
     pub(crate) fn insert(&mut self, offset: u64, bytes: Bytes) {
         self.data.push(Chunk { offset, bytes });
+        if self.data.len() - self.defragmented > 32 {
```
Ralith: Should have a brief comment describing the arbitrariness of this constant.

quinn-proto/src/assembler.rs
djc: I thought about that a little bit, but the problem is you're not at all sure that the already-defragmented chunks are a prefix of the heap. At least, I think they can come in in basically arbitrary order as long as they stay within flow credit, right?

quinn-proto/src/assembler.rs
djc: Fair, will add something.

djc:
> This is a beautiful solution, great work! I was imagining much more complex approaches.

Me too! I was quite happy that I came up with this little thing last night, although I lost some sleep over it. 😄

quinn-proto/src/assembler.rs
Ralith: Ah, good point. We could accomplish it with a flag bit, or by storing defragmented chunks separately, but the complexity tradeoff vs. a constant factor improvement for a weird case is less clear for those.

djc: I've been thinking a bit more about this. I think the current behavior is even desirable: if the malicious peer continues his sending of 1-byte frames, then eventually we are better off if we redefragment all the chunks rather than only defragmenting only once. Also, the extra state would make the performance in the fast/usual case somewhat worse again; so optimizing the bad/uncommon case that way doesn't seem like the right trade-off.

PR #603, opened 2020-01-16
Enable tokio feature required for tokio::spawn

Ralith:


Ralith:
Coverage error is spurious.

PR #612, opened 2020-01-26
Update rustls-native-certs

Ralith:
Improves robustness to certificate stores containing unsupported or invalid data.


Ralith:
Seems like #608 didn't take?

djc:
#608 indeed wasn't quite ready. I'll prepare some follow-up.

PR #614, opened 2020-01-27
Disable codecov status for now

djc:


djc:
This should really do it!

PR #618, opened 2020-01-31
H3 fix user's facing Futures behaviors on driver errors

stammw:
This PR mainly fixes the `IncomingRequest` stream not resolving when a driver error occurs.

It also fixes driver behaviour when an on error: depending on the error type, it should send a `CONNECTION_CLOSE` frame or not. 

Logging the driver stop reason as an error is also not always pertinent. It depends on what  triggered it.

I'm currently writing a test to check the `IncomingRequest` resolving, but once again i'm blocked because I try to create a generic helper that helps to shorten the test itself.

### DIFF
--- a/quinn-h3/Cargo.toml
+++ b/quinn-h3/Cargo.toml
@@ -46,7 +46,7 @@ proptest = "0.9.1"
 rand = "0.7.0"
 rcgen = "0.7"
 structopt = "0.3.0"
-tokio = { version = "0.2.6", features = ["io-util", "macros", "rt-threaded"] }
+tokio = { version = "0.2.6", features = ["io-util", "macros", "rt-threaded", "time"] }
 tracing-subscriber = "0.1.5"
 url = "2"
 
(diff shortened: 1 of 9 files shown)


quinn-h3/Cargo.toml
```diff
@@ -30,8 +30,8 @@ futures = "0.3.1"
 futures-util = { version = "0.3", default-features = false }
 http = "0.2"
 lazy_static = "1"
-quinn-proto = { path = "../quinn-proto", version = "0.5.0" }
-quinn = { path = "../quinn", version = "0.5.0" }
+quinn-proto = { path = "../quinn-proto" }
```
djc: Why do this? We'll need a specific version whenever we start publishing `quinn-h3`.

stammw: Yeah that was not useful at all. Now I remember `0.5.0` matches `0.5.X`.

stammw: I removed the commit

stammw:
Conflict has been resolved.

PR #620, opened 2020-02-01
Re-export public ParseError type

### LINKED ISSUE #616: tls::ParseError
The tls mod is private right now. This gives me an issue trying to convert the errors from tls such as ParseError. I am not sure it's as simple as making tls pub, but would appreciate your thoughts here.

Ralith:
Fixes #616.


djc:
Apparently 1.41.0 rustfmt has some changes. Want to fix those up, too?

PR #622, opened 2020-02-01
Stream finish ack

### LINKED ISSUE #617: If `Connection::finish` is called when there is no pending data, `StreamFinished` is emitted too soon.
`Connection::finish` causes `StreamFinished` to be emitted after all data has been acknowledged, and causes the `fin` flag to be set on the last packet in the stream that has not already been sent.  However, this is not enough to ensure the peer is notified of the shutdown.  If `Connection::finish` is called when there is no data that needs to be sent, it needs to send a packet that has `fin=true` and `len=0`, and wait for it to be acknowledged before emitting `StreamFinished`.  This only happens intermittently in my tests because there is usually unsent data when `Connection::finish` is called.

Protocols that have explicit in-band close indications are not affected, as the peer knows that the stream has been finished by other means.  I believe that HTTP/3 is such a protocol.

Ralith:
Fixes #617. Thanks to @DemiMarie-parity for identifying this bug!

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -508,13 +508,21 @@ where
             };
             ss.bytes_in_flight -= frame.data.len() as u64;
             self.unacked_data -= frame.data.len() as u64;
-            if ss.state == streams::SendState::DataSent && ss.bytes_in_flight == 0 {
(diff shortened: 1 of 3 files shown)


DemiMarie-parity:
Thank you for the fix @Ralith!

PR #624, opened 2020-02-02
Build benches as part of CI checks

stammw:
We probably want to know when they break ?

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -34,6 +34,10 @@ jobs:
         with:
           command: test
           args: --workspace
+      - uses: actions-rs/cargo@v1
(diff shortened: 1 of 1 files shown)


Ralith:
Isn't this covered by `build --all-targets`? We don't want to actually *run* the benchmarks, which is what this PR does.

stammw:
I believe `--no-run` argument inhibits running. 

Yes, `--all-targets` builds benches, I didn't notice that. Thanks.

Sorry for the vapor PR then. I got confused because benches seem to panic on the current code state.

Ralith:
Ah, no worries! If the benchmarks are panicing we should sort that out regardless.

PR #628, opened 2020-02-05
Improvements to Endpoint::accept

### LINKED ISSUE #627: Use a token to ensure that `Endpoint::accept()` is not called too many times
Currently, it is easy to call `Endpoint::accept()` too many times, which leads to problems.  Ideally, this would be prevented at compile time.

Ralith:
Fixes #627. Seeking feedback in particular on the final commit; I'm not sure how I feel about complicating the interface to further defend against downstream bugs which the second commit is able to trivialize for substantially less effort.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -632,7 +632,8 @@ where
     /// consumed, up to a limit of [`ServerConfig.accept_buffer`]. Calling this indicates the
     /// application's acceptance of that connection and releases the slot for reuse.
     pub fn accept(&mut self) {
-        self.incoming_handshakes -= 1;
+        // Don't overflow if a buggy caller invokes this too many times.
+        self.incoming_handshakes = self.incoming_handshakes.saturating_sub(1);
     }
 
     /// Unconditionally reject future incoming connections
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -628,9 +628,9 @@ where
 
     /// Free a handshake slot for reuse
     ///
-    /// Every time an [`Event::Handshaking`] is emitted, a slot is consumed, up to a limit of
-    /// [`ServerConfig.accept_buffer`]. Calling this indicates the application's acceptance of that
-    /// connection and releases the slot for reuse.
+    /// Every time an [`DatagramEvent::NewConnection`] is yielded by `Endpoint::handle`, a slot is
+    /// consumed, up to a limit of [`ServerConfig.accept_buffer`]. Calling this indicates the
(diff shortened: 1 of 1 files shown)


djc:
Yeah, this solution feels like it's too indirect. Why don't we decrement `incoming_handshakes` as soon as we yield the corresponding `DatagramEvent` from `Endpoint::handle()`?

djc:
As in, I think the threat model that we've modeled here includes the application (potentially through `quinn`) not handling `NewConnection` events fast enough, when actually we're supposed to be defending against peers initiating too many connections. If we only try to handle the latter, this becomes much more straightforward.

DemiMarie-parity:
@djc `libp2p-quic` actually relies on `accept` for correctness, so that it can put backpressure on incoming connections.  Any `NewConnection` event needs to be either queued or handled immediately, so that `ConnectionEvent`s for other connections can be handled.  In `libp2p-quic`, `Endpoint::accept` is not called until the application has accepted the connection.

Personally, I believe that the complexity of `AcceptToken` is worth it.  Calling `Endpoint::accept` too many times is likely to cause a remotely-triggerable DoS due to memory consumption.

Ralith:
> Why don't we decrement incoming_handshakes as soon as we yield the corresponding DatagramEvent from Endpoint::handle()?

This would make it impossible to exert backpressure on incoming connections, which is an explicit requirement.

> As in, I think the threat model that we've modeled here includes the application (potentially through quinn) not handling NewConnection events fast enough, when actually we're supposed to be defending against peers initiating too many connections.

This isn't a threat model so much as a feature. A server that only has capacity for a certain number of connections can only gracefully express that through use of `Endpoint::accept`, as @DemiMarie-parity describes.

> Calling Endpoint::accept too many times is likely to cause a remotely-triggerable DoS due to memory consumption.

In current master, excess accept calls will (in a release build) cause `incoming_handshakes` to underflow, producing a value which will certainly exceed the `accept_buffer` size of any reasonable configuration, therefore preventing any further connections from being accepted. That's also a DoS, though.

However, the second commit defends against this hazard with much less API impact than the third commit, simply by never allowing the value to underflow. An application which calls `accept` excessively would then simply end up accepting an at worst proportionally larger number of connections.

djc:
As another alternative solution, should we disband `DatagramEvent` such that `NewConnection` and `ConnectionEvent`s are retrieved from `Endpoint`s separately?

Ralith:
I don't think that would work well, because the connection needs to be driven immediately. Failing to call `fn accept(&mut self) -> Option<NewConnection>` continuously would lead to clients timing out rather than being gracefully rejected as in the current scheme.

I went ahead and stripped off the contentious commit; let's merge the obviously reasonable stuff.

PR #632, opened 2020-02-09
H3 throughput benchmark

stammw:
First H3 benchmarks, along with an optimisation on frame header encoding when sending.

Also includes some tests for request body reception.


djc:
I saw you previously had this commit in here for switching quinn/quinn-proto versions. Did you know you can set both `version` and `path` for a dependency, and it will basically do the right thing? See for example how `quinn` depends on `quinn-proto` (it additionally aliases the crate name, but you can ignore that).

stammw:
Oh I didn't know! Thanks!

Issue #567, opened 2019-12-17
Allow connection to return peer certificates


zeroqn:
Experiment on implement [libp2p-tls](https://github.com/libp2p/specs/blob/master/tls/tls.md).

But it need peer's certificates. I see rustls has ```get_peer_certificates``` func on ```Session``` trait. 
Consider add it to ```crypto::Session``` in quinn-proto and ```Connection``` in quinn.

djc:
I'm not sure it makes sense to do this at the Quinn layer. Couldn't you implement this as a rustls `ClientCertVerifier`?

Ralith:
This could definitely be done in `ClientCertVerifier`, but you'd have to extract the data through a side channel which is a bit of a pain. An accessor at the `Connection` level would be much more convenient, and rustls's method on `Session` is precedent, so I lean in favor.

djc:
It seems to me the fix for that would be to fix the `ClientCertVerifier` API to make this kind of thing easier on that side, rather than exposing specific APIs for the `crypto::Session` API.

I'd really prefer to avoid putting TLS-specific APIs in `Connection` or the crypto APIs, to keep it flexible for very different implementations of crypto protocols.

djc:
Just linking this here for reference: https://github.com/zeroqn/quinn/commit/5f216f1db423b548c760ed39f4994e42804adb54

Ralith:
> It seems to me the fix for that would be to fix the ClientCertVerifier API to make this kind of thing easier on that side.

What would that look like?

> I'd really prefer to avoid putting TLS-specific APIs in Connection or the crypto APIs, to keep it flexible for very different implementations of crypto protocols.

I agree that this is important, and I don't think it's necessarily an obstacle here. I think practically any reasonable cryptographic protocol will have *some* form of strongly identifying information about the peer, which could be defined as an associated type and exposed through a method that doesn't have "certificate" specifically in the name.

zeroqn:
Even if i save client's certificates, i need one more RTT to associate client's identity to connection.
Is it possible to do this without one more RTT?

> I think practically any reasonable cryptographic protocol will have some form of strongly 
> identifying information about the peer, which could be defined as an associated type and 
> exposed through a method that doesn't have "certificate" specifically in the name.

I agree.

Ralith:
> Is it possible to do this without one more RTT?

You need to wait until the handshake's complete regardless from a cryptographic correctness standpoint, I think.

zeroqn:
[...]

PR #638, opened 2020-02-23
QUINN: Allow retrieving the peer's certificate chain from a connection

kim:
So this would be my take on #567. 

It turns out that there is already a wrapper for `rustls::Certificate`, so we're not leaking implementation details. The bigger design concern, iiuc, is that TLS is not the only possible crypto protocol for QUIC -- however, `quinn` already commits to TLS. It is also the case that the SNI hostname (which is quite TLS-specific) is accessible at various places even in `quinn-proto`.

An alternative, but much more intrusive approach I considered first was to define an associated datatype for `quinn_proto::crypto::Session`, which abstracts over the type of authentication data applicable to the concrete session. Something like this:

```rust
pub trait Session: Sized {
    // ...
    type AuthenticationData;

    fn authentication_data(&self) -> Self::AuthenticationData;
}

// For `TlsSession`, we'd define
pub struct TlsAuthenticationData {
    sni_hostname: Option<String>,
    peer_certificates: Vec<rustls::Certificate>,
}

// Finally, we don't want to leak `rustls` at the `quinn` layer, so we need to
// define an isomorphic type there, which we can return from the
// `quinn::Connection`:
pub struct AuthenticationData {
    server_name: Option<String>,
    peer_certificates: Option<Vec<quinn::tls::Certificate>>,
}

// Enforce accessors, so we can later decide to make `quinn::Connection` 
// polymorphic in the crypto protocol without breaking client code.
impl AuthenticationData {
    pub fn server_name(&self) -> Option<&str> { /* ... */ }
    pub fn peer_certificates(&self) -> Option<&[quinn::tls::Certificate> { /* ... */ }
}

impl From<TlsAuthenticationData> for AuthenticationData { /* ... */ }

impl quinn::Connection {
    pub fn authentication_data(&self) -> AuthenticationData {
        self.0.lock().unwrap().inner.crypto_session().authentication_data().into()
    }
}
```

**EDIT**: Obviously, since the generic `Connection` is parametric, we can always reify the actual session type. Thus, this approach seemed much ado for very little gain. Using the `AuthenticationData` struct only at the `quinn` level might still be an option, if you'd be willing to deprecate the `Connection::server_name()` method.

Let me know what you think.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -22,7 +22,7 @@ use tracing::info_span;
 use crate::{
     broadcast::{self, Broadcast},
     streams::{RecvStream, SendStream, WriteError},
-    ConnectionEvent, EndpointEvent, SendDatagramError, VarInt,
+    tls, ConnectionEvent, EndpointEvent, SendDatagramError, VarInt,
 };
 
 /// In-progress connection attempt future
@@ -308,6 +308,26 @@ impl Connection {
     /// Returns `None` for outgoing connections.
     pub fn server_name(&self) -> Option<String> {
         self.0.lock().unwrap().inner.server_name().map(|x| x.into())
+    }
+
+    /// The certificate chain used by the peer to authenticate
+    ///
+    /// For clients, this is the certificate chain of the server. For servers,
+    /// it is the certificate chain of the client, or [`None`] if client
+    /// authentication was not requested.
+    pub fn presented_certs(&self) -> Option<Vec<tls::Certificate>> {
+        self.0
+            .lock()
+            .unwrap()
+            .inner
+            .crypto_session()
+            .get_peer_certificates()
+            .map(|certs| {
+                certs
+                    .into_iter()
+                    .map(|cert| tls::Certificate { inner: cert })
+                    .collect()
+            })
     }
 
     // Update traffic keys spontaneously for testing purposes.
--- a/quinn/src/tls.rs
+++ b/quinn/src/tls.rs
@@ -15,6 +15,11 @@ impl Certificate {
             inner: rustls::Certificate(der.to_vec()),
         })
     }
+
+    /// View the certificate in DER format
+    pub fn as_der(&self) -> &[u8] {
+        &self.inner.0
+    }
 }
 
 /// A chain of signed TLS certificates ending the one to be used by a server


quinn/src/connection.rs
```diff
@@ -310,6 +310,26 @@ impl Connection {
         self.0.lock().unwrap().inner.server_name().map(|x| x.into())
     }
 
+    /// The certificate chain used by the peer to authenticate
+    ///
+    /// For clients, this is the certificate chain of the server. For servers,
+    /// it is the certificate chain of the client, or [`None`] if client
+    /// authentication was not requested.
```
Ralith: Is this always true? Are clients unable to volunteer certificates unprompted?

djc:
I do feel that your proposed solution is too TLS-specific. I think your alternative proposal of having an associated type for peer identity would make sense, though. I'd do it without the accessors for now, though, instead just exposing the fields on `AuthenticationData` directly.

quinn/src/connection.rs
djc: > Are clients unable to volunteer certificates unprompted?

Based on some skimming of the specs, that does appear to be true. There's a nice graphic here: https://comodosslstore.com/blog/what-is-ssl-tls-client-authentication-how-does-it-work.html.

kim: My understanding of [Section 4.6.2](https://tools.ietf.org/html/rfc8446#section-4.6.2) of the TLS 1.3 RFC is that the client indicates that it is _willing_ to provide client authentication post-handshake. The server then sends `CertificateRequest` to make it actually present the cert.

kim: (And if that fails, we shouldn't get here)

kim:
@djc Thanks for the feedback. I'll happily amend.

Just to double-check: if this is an associated type of `crypto::Session`, we would bubble the `rustls::Certificate` all the way up to `quinn`. Since this type is just DER-encoded bytes, should we perhaps define `peer_certificates: Option<Vec<Vec<u8>>>` with the docstring saying that you have to do your own DER-decoding?

djc:
Yeah, it would make sense to me to just bubble up the DER directly rather than using the newtype. But you should probably await feedback from @Ralith before making further changes.

Ralith:
What's the benefit to having a dedicated -proto layer accessor for authentication data when we already expose `crypto_session` wholesale?

I agree with @kim that the `quinn` layer already strongly commits to TLS and that this change therefore doesn't change much in that regard. It might be nice to abstract crypto out there too, but I don't think we have immediate plans to do so, and I don't think this change would make an appreciable difference in the difficulty of doing so in the future.

Passing DER data out directly seems reasonable to me, but it feels wrong to use newtypes around DER in some places and work in terms of documented-to-be-DER bytes directly elsewhere. I think the main useful function of `quinn::tls` is wrapping rustls's PEM parser, which we could do more directly if desired, but perhaps refactoring that all is best left for a separate PR?

Ralith:
I guess the upside to `AuthenticationData` is that it makes the path towards a crypto-agnostic `quinn` much simpler, because that can easily be fetched regardless of its contents?

djc:
At the -proto layer it doesn't help much if any. To me, the quinn layer doesn't commit strongly to TLS except insofar as I haven't made it generic over `crypto::Session`. As you mention, I would prefer we avoid moving towards tighter coupling with TLS at that layer. In part, just because I am personally interested in using Quinn with Noise at some point (and of course, someone has already done ninn based on Quinn). Also, I think our TLS abstraction may provide some competitive advantage compared to other QUIC implementations.

I do see your point about the consistency of newtype-wrapped DER vs documented DER, we could also leave `quinn-proto::Certificate` in place at the high-level `AuthenticationData` level as long as there is some way for callers to get at the underlying data. I also agree we might decide to leave most of that to another PR/design (I wouldn't mind doing the legwork on that).

kim:
> `quinn-proto::Certificate`

Hm, that doesn't exist, the wrapper is at the `quinn` layer. Perhaps it pans out if we'd move it to `quinn-proto`? We can re-export it from `quinn` to avoid API breakage, and there is still the possibility to get at the raw `rustls` session via `crypto_session()` if you're working with `quinn-proto` directly.

djc:
> Hm, that doesn't exist, the wrapper is at the quinn layer.

That was just my mistake, although if I'm not mistaken it may have moved from quinn to quinn-proto, then back again before.

Ralith:
I'm on board with the idea of isolating the crypto-layer-specific parts of quinn along the lines of an `AuthenticationData` value. I wouldn't be opposed to merging this trivial PR in advance of it, though. Particularly since if `AuthenticationData` is defined in quinn-proto it can't use the `Certificate` newtype and that gets tangled up in the aforementioned refactor.

quinn/src/connection.rs
Ralith: That documents the procedure for a server to request client authentication. What happens if the client provides it without a request?

kim: Ya good point. `rustls`‘ state machine rejects invalid transitions:

https://github.com/ctz/rustls/blob/master/rustls/src/server/tls13.rs#L606
https://github.com/ctz/rustls/blob/master/rustls/src/server/tls13.rs#L641
https://github.com/ctz/rustls/blob/master/rustls/src/handshake.rs#L14

djc:
Alright, let's do it!

kim:
Thanks! Let me know if I can help evolving this further (I, too, am interested in being able to use Noise with quinn at some point :)).

Issue #521, opened 2019-11-25
H3: Do not send Settings that have the default value


stammw:
Thus minimize exchanged volume. Exchanging default values is useless.

stammw:
Closed via #643

PR #643, opened 2020-02-26
Refactor H3 settings for validation and defaults management

stammw:
This PR brings to settings:
* A way to validate the settings as the user set them
* This will make it possible to make some `Settings` usage infallible in lower layers, as a `Settings` struct is always valid now.
* Addressing  #521, only protocol's non-default values are sent.
* Differentiate protocol's defaults and crate's recommended values (used as default).

### DIFF
--- a/quinn-h3/src/proto/frame.rs
+++ b/quinn-h3/src/proto/frame.rs
@@ -276,7 +276,6 @@ impl PushPromiseFrame {
 
 #[derive(Debug, PartialEq, Clone)]
 pub struct SettingsFrame {
-    pub num_placeholders: u64,
     pub max_header_list_size: u64,
     pub qpack_max_table_capacity: u64,
     pub qpack_blocked_streams: u64,
@@ -285,7 +284,6 @@ pub struct SettingsFrame {
 impl Default for SettingsFrame {
     fn default() -> SettingsFrame {
         SettingsFrame {
-            num_placeholders: DEFAULT_NUM_PLACE_HOLDER,
             max_header_list_size: DEFAULT_MAX_HEADER_LIST_SIZE,
             qpack_max_table_capacity: DEFAULT_QPACK_MAX_TABLE_CAPACITY,
             qpack_blocked_streams: DEFAULT_QPACK_BLOCKED_STREAMS,
@@ -293,7 +291,6 @@ impl Default for SettingsFrame {
     }
 }
 
-const DEFAULT_NUM_PLACE_HOLDER: u64 = 16;
 const DEFAULT_MAX_HEADER_LIST_SIZE: u64 = 6;
 const DEFAULT_QPACK_MAX_TABLE_CAPACITY: u64 = 4096;
 const DEFAULT_QPACK_BLOCKED_STREAMS: u64 = 129;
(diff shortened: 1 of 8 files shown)


djc:
Looks very good! What you mind squashing the second and third commit together? These being split makes it harder to see how exactly the "new" implementation replaces the old one.

stammw:
Nice :) Done.

quinn-h3/src/proto/frame.rs
```diff
@@ -5,15 +5,20 @@ use quinn_proto::{
...
 #[derive(Debug, PartialEq)]
 pub enum Error {
     Malformed,
     UnsupportedFrame(u64),
     UnexpectedEnd,
     InvalidFrameValue,
+    InvalidSettingId(u64),
+    InvalidSettingValue(SettingId, u64),
+    SettingRepeated(SettingId),
+    SettingsExceded,
```
djc: Typo: exceeded.

quinn-h3/src/proto/settings.rs
```diff
@@ -0,0 +1,344 @@
...
+            qpack_max_blocked_streams: DEFAULT_PROTO_QPACK_MAX_BLOCKED_STREAMS,
+        }
+    }
+}
+
+impl Settings {
+    /// Create settings with quinn-h3's recomended values
+    pub fn new() -> Self {
+        Self {
+            max_header_list_size: DEFAULT_MAX_HEADER_LIST_SIZE,
```
djc: I think we should just inline the values here, not much point keeping them as constants? You can still use `DEFAULT_MAX_HEADER_LIST_SIZE` to make it clear that that remains at the protocol default value.

quinn-h3/src/proto/settings.rs
```diff
@@ -0,0 +1,344 @@
...
+                SettingId::MAX_HEADER_LIST_SIZE => this.set_max_header_list_size(*val)?,
+                SettingId::QPACK_MAX_TABLE_CAPACITY => this.set_qpack_max_table_capacity(*val)?,
+                SettingId::QPACK_MAX_BLOCKED_STREAMS => this.set_qpack_max_blocked_streams(*val)?,
+                x => return Err(Error::InvalidSettingId(x.0)),
+            };
+        }
+        Ok(this)
+    }
+
+    pub(super) fn to_frame(&self) -> SettingsFrame {
```
djc: Is this only for tests? If so, maybe add `#[cfg(test)]`? Otherwise, we should probably not have the `expect()` calls here? I think this would apply to everything marked `pub(super)` here.

stammw: `to_frame()` is called in `proto::connection::with_setiings()`, and `from_frame()` in `connection::poll_recv_control()`

quinn-h3/src/proto/settings.rs
stammw: Got them inlined in a new commit

quinn-h3/src/proto/frame.rs
stammw: Rebased in cfea581d H3: refactor settings in it's own module

quinn-h3/src/proto/settings.rs
djc: So in that case we should probably make `to_frame()` fallible rather than allow it to panic?

stammw: `to_frame()` operates on `Settings` inner-data only, so I made this function panic because if something bad happens, it can only be a programming error in the `impl` or the encapsulation is broken.

stammw: And, if the user gets an error here, it can only be the fault of the inner Settings code. Making the caller handle that is not fair :).

djc: Fair enough!

PR #648, opened 2020-02-29
Fix Stream::finish hang on closed connections

### LINKED ISSUE #645: Hanging in SendStream::finish
[This program](https://github.com/alecmocatta/chanel/blob/1dbac27a0e13cef18501cfa9a9c06382532ac177/src/main.rs) alternately hangs or errors after a dozen or so iterations. If you remove the sleep on line 73, then it works as expected.

The errors seen so far are:
* `ConnectionClosed` on line 53: doesn't prevent progress so is currently ignored;
* `connect: ApplicationClosed(ApplicationClose { error_code: 0, reason: b"" })` on line 154: 2 bytes have been sent on line 148 and `finish` has been called on line 51, so I'd expect lines 154-157 to succeed. May be same as #641, disappears when I remove the error checking here: https://github.com/djc/quinn/blob/7122eab85712b15b598998b324f3e777bed57ae6/quinn/src/connection.rs#L88-L92
* A hang waiting on `SendStream::finish`, with strace showing the process blocked on mio's epoll_wait/kevent call.

These manifest identically for me on Linux and macOS. I hope I'm not doing anything that shouldn't be expected to work, apologies if so.

Ralith:
Fixes #645.

### DIFF
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -54,7 +54,8 @@ mod udp;
 
 pub use proto::{
     crypto, ApplicationClose, ClientConfig, ConnectError, ConnectionClose, ConnectionError,
-    SendDatagramError, ServerConfig, Transmit, TransportConfig, VarInt,
+    SendDatagramError, ServerConfig, Transmit, TransportConfig, TransportError, TransportErrorCode,
+    VarInt,
 };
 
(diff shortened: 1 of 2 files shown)


Ralith:
Forcibly merging as the only failure is a preexisting flaky h3 test.

PR #653, opened 2020-03-02
H3 0-RTT implementation

stammw:


quinn-h3/src/tests/mod.rs
```diff
@@ -109,3 +110,73 @@ async fn client_send_stream_body() {
...
+    let resp = conn.send_request(get!()).await.expect("request").0.await;
+    assert_matches!(resp.map(|_| ()), Ok(_));
+    assert_matches!(zerortt_accepted.await, true);
+    conn.close();
+
+    assert!(timeout_join(server_handle).await.is_ok());
+}
+
+#[tokio::test(threaded_scheduler)]
+async fn zero_rtt_client_fordbids_non_idempotent() {
```
djc:
```suggestion
async fn zero_rtt_client_forbids_non_idempotent() {
```

quinn-h3/src/tests/mod.rs
```diff
@@ -109,3 +110,73 @@ async fn client_send_stream_body() {
...
+    let mut helper = Helper::new();
+    let (_, incoming) = helper.make_server();
+    tokio::spawn(async move { serve_n_0rtt(incoming, 1).await });
+
+    let (conn, _) = helper.make_0rtt().await;
+    assert!(conn.send_request(post!()).await.is_err());
+}
+
+#[tokio::test(threaded_scheduler)]
+async fn zero_rtt_server_fordbids_non_idempotent() {
```
djc:
```suggestion
async fn zero_rtt_server_forbids_non_idempotent() {
```

quinn-h3/src/connection.rs
```diff
@@ -19,10 +19,7 @@ use crate::{
     frame::{self, FrameStream},
     proto::{
         self,
-        connection::{
-            Connection, DecodeResult, Error as ConnectionError,
-            PendingStreamType,
-        },
+        connection::{Connection, DecodeResult, Error as ConnectionError, PendingStreamType},
```
djc: Can we fold this into the earlier commit that introduced it?

quinn/src/builders.rs
```diff
@@ -222,6 +222,12 @@ impl ClientConfigBuilder {
         self
     }
 
+    /// Enable 0-RTT, by configuring TLS to accept early data.
```
Ralith:
```suggestion
    /// Enable 0-RTT
```
TLS's own terminologoy is a distraction here.

quinn/src/connection.rs
```diff
@@ -746,7 +747,9 @@ impl ConnectionInner {
     }
 
     pub(crate) fn check_0rtt(&self) -> Result<(), ()> {
-        if self.inner.is_handshaking() || self.inner.accepted_0rtt() {
+        if self.inner.is_handshaking()
+            || (self.inner.accepted_0rtt() || self.inner.side().is_server())
```
```suggestion
            || self.inner.accepted_0rtt() || self.inner.side().is_server()
```

stammw:
I've addressed all your comments rebasing rebasing the changes.

The CI is back to being stuck on something though...

djc:
Looks like actually some of your new tests are hanging. That might reflect an actual problem?

stammw:
Yes I think it's precisely what's happening.

I'm trying to reproduce this with tarpaulin in local.

For the moment I couldn't read CI's logs.

djc:
Let me know when you want me to re-review!

stammw:
Turns out that #686 fixed the timeouts, and using `#[tokio::test(threaded_scheduler)]` made 0-RTT impossible to reach with a request (handshake completes before any can be sent). Using `#[tokio::test]` basic scheduler fixes tarpaulin last scheduling surprises.

djc:
It sounds to me like we should make sure the API should still allow sending 0-RTT even if the handshake is very quick to complete. Is that not the case here?

stammw:
Well yes, you always can have a 0-RTT successful request even if the handshake is already completed. The thing here is testing that a non-idempotent requests doesn't make it through the API when the replay safety is not guaranteed. Say in any packet before handshake completes. 

So the tests were failing because such requests where accepted as handshake completion made them valid.

I'm not totally sure about the conditions I use to assert that from `quinn` though.

stammw:
Maybe a test could make that clearer.

stammw:
Added a few tests to explicit everything works. I hope this addresses your doubt ?

quinn-h3/src/connection.rs
djc: Can you still take a look at this?

quinn-h3/src/client.rs
```diff
@@ -136,6 +136,10 @@ impl Connection {
         ) = request.into_parts();
         let (send, recv) = self.0.quic.open_bi().await?;
 
+        if recv.is_0rtt() && !method.is_idempotent() {
+            return Err(Error::internal("non idempotent method tried on 0RTT"));
```
djc:
```suggestion
            return Err(Error::internal("non-idempotent method tried on 0RTT"));
```

quinn-h3/src/server.rs
```diff
@@ -142,6 +144,32 @@ pub struct Connecting {
...
 }
 
+impl Connecting {
+    pub fn into_0rtt(self) -> Result<(IncomingRequest, ZeroRttAccepted), Self> {
+        let Self {
+            connecting,
+            settings,
+        } = self;
+        let (
+            quinn::NewConnection {
```
djc: Style nit: maybe put the destruction of the `NewConnection` in its own statement? The low vertical density makes this complex statement pretty hard to grok.

quinn-h3/src/server.rs
```diff
@@ -227,6 +257,14 @@ impl RecvRequest {
             .version(http::version::Version::HTTP_3)
             .body(())
             .unwrap();
+
+        if self.is_0rtt && !request.method().is_idempotent() {
+            return Err(Error::peer(format!(
+                "Tried an non indempotent method in 0-RTT: {}",
```
djc:
```suggestion
                "Tried an non-idempotent method in 0-RTT: {}",
```

quinn-h3/src/tests/helpers.rs
```diff
@@ -89,6 +113,88 @@ impl Helper {
...
+            .expect("no 0rtt")
+    }
+}
+
+pub struct FakeConnection(client::Connection);
+
+impl FakeConnection {
+    pub async fn post(&mut self) -> FakeRequest {
+        let request = Request::post("https://localhost").body(()).unwrap();
+        let (
```
djc: Same as above, please outline the destruction of `Parts`.

quinn-h3/src/tests/mod.rs
```diff
@@ -137,7 +137,7 @@ async fn serve_n_0rtt(mut incoming: IncomingConnection, n: usize) -> Result<(),
     Ok(())
 }
 
-#[tokio::test(threaded_scheduler)]
+#[tokio::test]
```
djc: Since this commit just fixes the previous one, let's fold these?

quinn-h3/src/tests/helpers.rs
```diff
@@ -143,9 +143,13 @@ impl Helper {
             .map_err(|_| ())
             .expect("no 0rtt")
     }
+
+    pub fn socket(&self) -> SocketAddr {
```
djc:
```suggestion
    pub fn socket_addr(&self) -> SocketAddr {
```

quinn-h3/src/tests/helpers.rs
```diff
@@ -12,6 +12,34 @@ use crate::{
     server::{self, IncomingConnection, Server},
 };
 
+#[macro_export]
```
djc: If the only difference between the zero- and one-argument forms is passing the argument to a nested function, maybe just make these function calls that take an `Option`? Macros seem like more complexity than needed here.

stammw: I admit the YAGNI case here... I've been thinking about the uses-cases for this request construction features. I'd like to be able to pass headers for future tests, similarly with `serde`'s `json!()` macro:

```rust
get!("/path", { "name": "John Doe" })
post!("/path", body)
post!("/path", { "name": "John Doe" },  body)
```

But I was only focused on shortening tests. Do you think that doesn't worth it?

quinn-h3/src/server.rs
stammw: Simplified those and the other occurrence of this nit in a new commit.

quinn-h3/src/server.rs
stammw: Rebased in original commit

quinn-h3/src/tests/helpers.rs
stammw: Rebased in original commit

quinn-h3/src/tests/helpers.rs
djc: I've been looking at stuff like this recently. Maybe it would be useful to just use the [`http::request::Builder`](https://docs.rs/http/0.2.0/http/request/struct.Builder.html) here, and add wrappers if some common cases are still not concise enough? It also has a relatively lightweight way to add headers.

stammw: You are right, I got those macros out and it's fine. Let's write things when we need them.

I got the changes in the 0RTT impl commit. I also squashed the "more tests" commits because it was essentially the same subject (and that made the macros easier to remove).

quinn-h3/src/client.rs
stammw: fixed in original commit

stammw:
Thanks for the review!

PR #659, opened 2020-03-05
Handle overlapping chunks during defragment

### LINKED ISSUE #652: Stream data corruption
Data sent along a stream sometimes differs from the data received. [Test case here](https://github.com/alecmocatta/chanel/blob/corruption/src/main.rs). It takes a while (ballpark ~30k iterations in my case) to reproduce for me on macOS on localhost with no packet unreliability configured.

Comparing the hex dumps from a few different repros, 1167*n (typically 1 but sometimes 2) bytes seem to be duplicated at the beginning, and that same number deleted at some random other offset.

alecmocatta:
Fixes #652

Duplicated various tests with added calls to defragment.

### DIFF
--- a/quinn-proto/src/assembler.rs
+++ b/quinn-proto/src/assembler.rs
@@ -86,12 +86,17 @@ impl Assembler {
         let new = BinaryHeap::with_capacity(self.data.len());
         let old = mem::replace(&mut self.data, new);
         for chunk in old.into_sorted_vec().into_iter().rev() {
-            if offset + (buffer.len() as u64) < chunk.offset {
+            let end = offset + (buffer.len() as u64);
+            if let Some(overlap) = end.checked_sub(chunk.offset) {
+                if let Some(bytes) = chunk.bytes.get(overlap as usize..) {
+                    buffer.extend_from_slice(bytes);
+                }
+            } else {
                 let bytes = buffer.split().freeze();
                 self.data.push(Chunk { offset, bytes });
                 offset = chunk.offset;
+                buffer.extend_from_slice(&chunk.bytes);
             }
-            buffer.extend_from_slice(&chunk.bytes);
         }
 
         let bytes = buffer.split().freeze();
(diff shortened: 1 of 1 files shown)


quinn-proto/src/assembler.rs
```diff
@@ -86,12 +86,17 @@ impl Assembler {
         let new = BinaryHeap::with_capacity(self.data.len());
         let old = mem::replace(&mut self.data, new);
         for chunk in old.into_sorted_vec().into_iter().rev() {
-            if offset + (buffer.len() as u64) < chunk.offset {
+            let end = offset + (buffer.len() as u64);
+            let overlap = end.saturating_sub(chunk.offset);
+            if end - overlap != chunk.offset {
```
Ralith: I find this difficult to follow. Could it be written without comparing values that were derived from eachother? Perhaps `if end < chunk.offset`?

quinn-proto/src/assembler.rs
```diff
@@ -86,12 +86,17 @@ impl Assembler {
         let new = BinaryHeap::with_capacity(self.data.len());
         let old = mem::replace(&mut self.data, new);
         for chunk in old.into_sorted_vec().into_iter().rev() {
-            if offset + (buffer.len() as u64) < chunk.offset {
+            let end = offset + (buffer.len() as u64);
+            let overlap = end.saturating_sub(chunk.offset);
```
djc: We can defer the initial definition of `overlap` until after the `if end < chunk.offset` and inline the cast to `usize`. Let's also introduce an empty line before that definition.

Ralith: I think this whole thing could actually be recast into `if let Some(overlap) = end.saturating_sub(chunk.offset) { /* append */ } else { /* split */ }`, if we want to get really tidy.

djc: That sounds even better!

alecmocatta: I just updated to `checked_sub`. I think it's possibly not quite what you were aiming for as the `extend_from_slice` has to happen depending on a different condition: `overlap < chunk.bytes.len()`. I also updated that condition to avoid the need to redefine `overlap`. Let me know what you think?

quinn-proto/src/assembler.rs
```diff
@@ -86,12 +86,18 @@ impl Assembler {
...
+                overlap
+            } else {
                 let bytes = buffer.split().freeze();
                 self.data.push(Chunk { offset, bytes });
                 offset = chunk.offset;
+                0
+            };
+            if let Some(bytes) = chunk.bytes.get(overlap as usize..) {
+                buffer.extend_from_slice(bytes);
             }
```
Ralith: Perhaps this would be cleaner if `extend_from_slice` was moved into both branches above? Only the first instance would need the checked `get` call, and it'd save the weird dance to get an overlap value out.

alecmocatta: Good idea, updated.

PR #661, opened 2020-03-06
Implement Stream for &Incoming, &IncomingUniStreams etc

alecmocatta:
Calling `.next()` on `Incoming`, `IncomingUniStreams`, `IncomingBiStreams` and `Datagrams` currently requires an `&mut` reference. Their opposing methods `connect`, `open_uni`, `open_bi` and `send_datagram` require only a `&` reference.

This PR resolves that asymmetry, enabling the accepting of connections, streams and datagrams with only a `&` reference. This avoids the need for Mutexes in user code.

### DIFF
--- a/quinn-h3/src/server.rs
+++ b/quinn-h3/src/server.rs
@@ -124,16 +124,24 @@ pub struct IncomingConnection {
     settings: Settings,
 }
 
-impl Stream for IncomingConnection {
+impl Stream for &IncomingConnection {
     type Item = Connecting;
 
-    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
         Poll::Ready(
-            ready!(Pin::new(&mut self.incoming).poll_next(cx)).map(|c| Connecting {
+            ready!(Pin::new(&mut &self.incoming).poll_next(cx)).map(|c| Connecting {
                 connecting: c,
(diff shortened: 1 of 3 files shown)


Ralith:
Also, can you elaborate a bit on the usecase here?

alecmocatta:
Ah, thanks for pointing that out!

In light of that I think the best solution would be not to implement `Stream for &Incoming` (as I don't believe that can use the approach of `ManualResetEvent`?) but instead to provide an explicit `Incoming::next(&self)` method. `ManualResetEvent` is new to me but it looks like it can do the job.

The use case is fairly inconsequential: simply avoiding wrapping `Incoming` etc in redundant Mutexes like [here](https://github.com/alecmocatta/chanel/blob/c1e682a993ff25dc8ddd8dd4d484910c3ec9ea91/src/main.rs#L98).

Ralith:
> The use case is fairly inconsequential: simply avoiding wrapping Incoming etc in redundant Mutexes like here.

It looks like you're only accessing the stream from a single task. Why not just take ownership of it?

djc:
@alecmocatta if you have a moment, it would be nice if you could answer the questions here.

alecmocatta:
> Why not just take ownership of it?

`fn main` in that example is one task, but I'm building an API that will need to be callable from multiple tasks/threads.

Ralith:
Can you sketch the API for us? Accepting messages concurrently from multiple tasks sounds strange; if you're trying to distributed work between threads or similar, you'd likely be better off with an explicit acceptor task feeding into a real load balancer.

djc:
Going to close this for now. Feel free to re-open if you want to pick it up again!

Issue #355, opened 2019-05-29
H3: body buffer size configuration


stammw:
Make it possible for the user to specify body initial buffer size and max size for receiving it. Replacing the static values at `quinn-h3/src/server.rs:296`, and `quinn-h3/src/client.rs:320` with fields settable by the user.

Receiving the body from the user code would then look like: 

```rust
response.set_max_capacity(n);
response.set_capacity(m);
response.body().and_then(...)
```

TimonPost:
The lines pointed in this issue don't exist anymore. Is this issue still valid?

stammw:
The initial concern was about removing `body::ReadToEnd::new()` params
`capacity` and `size_limit` and offer a request / connection / endpoint level
configuration interface to set those.

`body::ReadToEnd` will soon be removed in an effort to simplify the public API,
I started to draft it in #518. I'm currently studying `hyper::h2` API and some
crates I'd like to see `quinn-h3` integrated into, namely `hyper` and `actix-web`.

Still, there is some work to do with Body chunks size management, even if it is
less trivial than I first thought:

Bodies are transported via an arbitrary number of `DataFrame`s, with a header
indicating the size of the payload for each of them. But what if this `Data`
frame size is `1G`? We don't want to grow a buffer until the whole payload is
received, because it would be too easy for a malicious peer to make our peer go
out of memory.

I came up with a quick solution via `proto::frame::PartialData`, which cuts big
`DataFrame`s into smaller ones, then permits `frame::FrameStream` to yield their
payloads to the user. The problem is that the largest size for a chunk is now
determined by the internal buffer size of `tokio_codec::FramedRead`.

If the buffer is too small, the throughput performance will be limited. We also
want to limit it's growth to prevent it from draining the memory, and having a
way to make it grow by user-determined setting would be a nice feature.

For example, a `content-length` header could be present in the response and we
would like to let the user choose how he wants to handle the body.

So now the issue is more about: what should we do to maximize throughput and let
the user influence the memory management via some settings. Keeping in mind that
we want to stay close to what `h2` does so `quinn-h3` integrates nicely with
crates using it.

Definitely not a trivial issue, but it could bring a good grasp on what's this
project problematics are about.

@djc @Ralith Maybe you have some thoughts about all this?

djc:
[...]

PR #662, opened 2020-03-06
H3 adjust read buffer size form decoding

stammw:
Mainly addresses #355.

This PR sets the size of the framed read buffer to the size of one UDP packet payload. Thus reducing the chances of reallocations needed to parse an entire http frame.

It uses a new feature of `tokio-util` made available in `0.3.0`.

Some benchmarks have been added to assert the throughput enhancement.

Also brings connection closure fix and its test.

### DIFF
--- a/bench/Cargo.toml
+++ b/bench/Cargo.toml
@@ -12,7 +12,7 @@ futures = "0.3.1"
 quinn = { path = "../quinn" }
 rcgen = "0.7"
 rustls = "0.17"
-tokio = { version = "0.2.2", features = ["rt-core"] }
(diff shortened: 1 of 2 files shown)


djc:
I think you forgot to add some new files here. :) It also looks like the first two commits flip-flop on the derivation of `Clone` under `test` only or not?

stammw:
I'm not sure for the omitted files. This PR was essentially updating `tokio-util` and using `FramedRead::with_capacity()`.

It was written relatively long ago, so I had to rebase, and fail. Sorry for that.

djc:
Ah, so just ditch the first two commits then maybe?

stammw:
... Done

PR #663, opened 2020-03-06
Implement an `outbound_streams` function

### LINKED ISSUE #654: RFE: Provide a count of open streams
It would be nice to have a count of outbound streams that have not been finished yet.  When this count is zero, the connection can be safely closed.

DemiMarie-parity:
Fixes #654

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -2708,6 +2708,11 @@ where
         Some(id)
     }
 
+    /// The number of streams that may have unacknowledged data.
+    pub fn send_streams(&self) -> usize {
+        self.streams.send_streams()
+    }
+
     /// Finish a send stream, signalling that no more data will be sent
     pub fn finish(&mut self, id: StreamId) -> Result<(), FinishError> {
         let ss = self
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -2598,7 +2598,13 @@ where
     ///
     /// This does not ensure delivery of outstanding data. It is the application's responsibility to
     /// call this only when all important communications have been completed, e.g. by calling
-    /// `finish` on outstanding streams and waiting for the corresponding `StreamFinished` event.
+    /// [`Connection::finish`] on outstanding streams and waiting for the corresponding
+    /// [`StreamFinished`] event.
+    ///
+    /// If [`Connection::send_streams`] returns 0, all outstanding data has been delivered. There
+    /// may still be data from the peer that has not been received.
+    ///
+    /// [`StreamFinished`]: Event::StreamFinished
     pub fn close(&mut self, now: Instant, error_code: VarInt, reason: Bytes) {
         let was_closed = self.state.is_closed();
         if !was_closed {
@@ -2713,7 +2719,9 @@ where
         self.streams.send_streams()
     }
 
-    /// Finish a send stream, signalling that no more data will be sent
+    /// Finish a send stream, signalling that no more data will be sent.
+    ///
+    /// If this fails, no [`Event::StreamFinished`] will be generated.
     pub fn finish(&mut self, id: StreamId) -> Result<(), FinishError> {
         let ss = self
             .streams
--- a/quinn-proto/src/streams.rs
+++ b/quinn-proto/src/streams.rs
@@ -283,13 +283,18 @@ impl Send {
 pub enum WriteError {
     /// The peer is not able to accept additional data, or the connection is congested.
     ///
-    /// If the peer issues additional flow control credit, an `Event::StreamWritable` will be
+    /// If the peer issues additional flow control credit, a [`StreamWritable`] event will be
     /// generated, indicating that retrying the write might succeed.
+    ///
+    /// [`StreamWritable`]: crate::Event::StreamWritable
     #[error(display = "unable to accept further writes")]
     Blocked,
-    /// The peer is no longer accepting data on this stream.
+    /// The peer is no longer accepting data on this stream. No [`StreamFinished`] event will be
+    /// emitted for this stream.
     ///
     /// Carries an application-defined error code.
+    ///
+    /// [`StreamFinished`]: crate::Event::StreamFinished
     #[error(display = "stopped by peer: code {}", 0)]
     Stopped(VarInt),
     /// Unknown stream
@@ -524,7 +529,8 @@ pub(crate) enum RecvState {
 /// Reasons why attempting to finish a stream might fail
 #[derive(Debug, Error, Clone, PartialEq, Eq)]
 pub enum FinishError {
-    /// The peer is no longer accepting data on this stream.
+    /// The peer is no longer accepting data on this stream. No
+    /// [`StreamFinished`](crate::Event::StreamFinished) event will be emitted for this stream.
     ///
     /// Carries an application-defined error code.
     #[error(display = "stopped by peer: code {}", 0)]
--- a/quinn-proto/src/streams.rs
+++ b/quinn-proto/src/streams.rs
@@ -21,6 +21,8 @@ pub(crate) struct Streams {
     pub next_remote: [u64; 2],
     // Next to report to the application, once opened
     next_reported_remote: [u64; 2],
+    // Outbound streams
+    send_streams: usize,
 }
 
 impl Streams {
@@ -33,6 +35,7 @@ impl Streams {
             max_remote: [max_remote_bi, max_remote_uni],
             next_remote: [0, 0],
             next_reported_remote: [0, 0],
+            send_streams: 0,
         };
 
         for dir in Dir::iter() {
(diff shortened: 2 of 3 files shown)


DemiMarie-parity:
I did not implement tests because I am not particularly sure what to write.  Any advice would be greatly appreciated.

djc:
I feel like this state should maybe live inside `Streams`? Otherwise looks okay.

quinn-proto/src/connection.rs
```diff
@@ -2598,7 +2603,10 @@ where
     ///
     /// This does not ensure delivery of outstanding data. It is the application's responsibility to
     /// call this only when all important communications have been completed, e.g. by calling
-    /// `finish` on outstanding streams and waiting for the corresponding `StreamFinished` event.
+    /// [`Connection::finish`] on outstanding streams and waiting for the corresponding
+    /// `StreamFinished` event.
+    ///
+    /// It is safe to call this if [`Connection::outbound_streams`] returns 0.
```
Ralith: This should be reworded as "safe" has a different very specific meaning in rust.

quinn-proto/src/connection.rs
```diff
@@ -154,6 +154,8 @@ where
     rem_cids: Vec<IssuedCid>,
     /// State of the unreliable datagram extension
     datagrams: DatagramState,
+    /// The number of open *outbound* streams.
+    outbound_streams: u64,
```
Ralith: We should use the standard term "send streams" throughout. I also agree that this probably makes more sense as a member of `Streams`.

DemiMarie-parity: Agreed.  The difficult part is knowing when to decrement the number of send streams.  I suspect that there are missing calls to `maybe_cleanup`.

quinn-proto/src/connection.rs
DemiMarie-parity: Done.

djc:
(All of these commits should be squashed before merging.)

DemiMarie-parity:
This may also fix other bugs, in particular a potential memory leak.

djc:
> This may also fix other bugs, in particular a potential memory leak.

Those changes should remain in a separate commit, please.

DemiMarie-parity:
> > This may also fix other bugs, in particular a potential memory leak.
> 
> Those changes should remain in a separate commit, please.

These changes will need to come first, since they are essential for accurate counting of outgoing streams.

djc:
Sure, you can reorder the commits accordingly.

quinn-proto/src/connection.rs
```diff
@@ -910,13 +910,19 @@ where
...
             DataSent { .. } => {
-                self.events.push_back(Event::StreamFinished {
-                    stream: stream_id,
-                    stop_reason,
-                });
                 // No need to hold on to the stop_reason since it's propagated above
-                stream.state = ResetSent { stop_reason: None };
-                self.streams.send_streams -= 1;
+                stream.state = if stopped {
+                    ResetRecvd { stop_reason: None }
```
Ralith: This, and the similar change below, is incorrect. `STOP_SENDING` does not constitute acknowledgement of a reset. See [the Stream States](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#name-stream-states) section of the transport draft for details.

DemiMarie-parity: Fixed.

DemiMarie-parity:
Does the implementation in this PR guarantee that if `Connection::send_streams()` returns 0, all outstanding data has been acknowledged?

quinn-proto/src/connection.rs
```diff
@@ -909,11 +909,13 @@ where
                 return;
             }
             DataSent { .. } => {
-                self.events.push_back(Event::StreamFinished {
-                    stream: stream_id,
-                    stop_reason,
-                });
                 // No need to hold on to the stop_reason since it's propagated above
+                if stopped {
```
djc: This seems to clearly belong in a separate commit.

quinn-proto/src/connection.rs
```diff
@@ -2600,7 +2602,11 @@ where
     ///
     /// This does not ensure delivery of outstanding data. It is the application's responsibility to
     /// call this only when all important communications have been completed, e.g. by calling
-    /// `finish` on outstanding streams and waiting for the corresponding `StreamFinished` event.
+    /// [`Connection::finish`] on outstanding streams and waiting for the corresponding
```
djc: Let's also keep documentation-only improvements in another separate commit.

quinn-proto/src/connection.rs
DemiMarie-parity: Is it correct?

djc: I think @Ralith knows better than I do.

quinn-proto/src/connection.rs
DemiMarie-parity: One reason I changed the documentation is that I could not tell if, should `Connection::write` and/or `Connection::finish` return `WriteError::Stopped`, a `StreamFinished` event will be received.  I assume the answer is “no”, but this should be made clearer.

quinn-proto/src/connection.rs
Ralith: This doesn't impact the protocol state machine, but changes the quinn-proto API behavior such that if you explicitly reset a stream that you're already finishing, no `StreamFinished` event is yielded. This does not affect quinn, as the interface is constructed to statically forbid that.

I can see the case for the change; `StreamFinished` normally indicates that either the stream was successfully delivered in full or the peer stopped it. I'm a bit concerned that in the hypothetical world where you're finishing and resetting a stream concurrently this doesn't give feedback about the status of the finish operation, but as we don't currently permit that in quinn and a hypothetical quinn-proto user won't necessarily need quinn-proto to communicate that for them, I think the change is reasonable as written--though I agree it belongs in an isolated commit.

quinn-proto/src/connection.rs
Ralith: I agree, and this is a good change. It's just easier to review logically independent changes for correctness if they're in separate commits, and easier to bisect changes in the future.

quinn-proto/src/connection.rs
```diff
@@ -915,6 +915,7 @@ where
                 });
                 // No need to hold on to the stop_reason since it's propagated above
                 stream.state = ResetSent { stop_reason: None };
+                self.streams.maybe_cleanup(stream_id)
```
Ralith: This call, and the similar one added below, isn't needed. Cleanup of reset streams occurs when the reset is acknowledged (see `on_packet_acked`).

quinn-proto/src/connection.rs
Ralith: nit: Strictly speaking, the commit message should read "locally reset."

DemiMarie-parity: I am going to revert this.  Guaranteeing that every successful call to `Connnection::finish` results in a `StreamFinished` event (unless the connection is lost) makes API design easier, and avoids difficult-to-debug hangs.

quinn-proto/src/connection.rs
DemiMarie-parity: Fixed.

quinn-proto/src/streams.rs
```diff
@@ -283,13 +283,18 @@ impl Send {
...
-    /// If the peer issues additional flow control credit, an `Event::StreamWritable` will be
+    /// If the peer issues additional flow control credit, a [`StreamWritable`] event will be
     /// generated, indicating that retrying the write might succeed.
+    ///
+    /// [`StreamWritable`]: crate::Event::StreamWritable
     #[error(display = "unable to accept further writes")]
     Blocked,
-    /// The peer is no longer accepting data on this stream.
+    /// The peer is no longer accepting data on this stream. No [`StreamFinished`] event will be
+    /// emitted for this stream.
```
Ralith: Maybe clearer to say "it's not useful to call `finish`", since `StreamFinished` is not normally generated in association with operations that might return a `WriteError`.

quinn-proto/src/connection.rs
```diff
@@ -2598,7 +2598,13 @@ where
     ///
     /// This does not ensure delivery of outstanding data. It is the application's responsibility to
     /// call this only when all important communications have been completed, e.g. by calling
-    /// `finish` on outstanding streams and waiting for the corresponding `StreamFinished` event.
+    /// [`Connection::finish`] on outstanding streams and waiting for the corresponding
+    /// [`StreamFinished`] event.
+    ///
+    /// If [`Connection::send_streams`] returns 0, all outstanding data has been delivered. There
```
```suggestion
    /// If [`Connection::send_streams`] returns 0, all outstanding stream data has been delivered. There
```

e.g. datagrams may not have been delivered.

quinn-proto/src/connection.rs
Ralith: It might be worth adding (in a future PR) a flag in `StreamFinished` which indicates that the finish "completed" due to a local reset, then, since as it stands it's indistinguishable from success.

PR #665, opened 2020-03-08
Update to draft 27

Ralith:
VHDCRZSBU vs picoquic server!


djc:
First pass: mostly looks good. For the retry integrity stuff, placing the interface directly in `Session` feels weird to me since it does not appear to depend on the TLS implementation at all (that is, the impl should live in the `ring` module instead). Maybe make it a separate ZST associated type, since it seems to also not depend on any of the other key state?

quinn-proto/src/connection.rs
```diff
@@ -2357,7 +2357,7 @@ where
         let is_0rtt = space_id == SpaceId::Data && space.crypto.is_none();
 
         // HANDSHAKE_DONE
-        if mem::replace(&mut space.pending.handshake_done, false) {
+        if !is_0rtt && mem::replace(&mut space.pending.handshake_done, false) {
```
djc: Arguably this should be squashed into your "draft 27 handshake" commit.

quinn/Cargo.toml
```diff
@@ -31,7 +31,7 @@ azure-devops = { project = "dochtman/Projects", pipeline = "Quinn", build = "1"
 bytes = "0.5.2"
 err-derive = "0.2.3"
 futures = "0.3.1"
-libc = "0.2.49"
+libc = "0.2.67"
```
djc: But then you don't actually use the new constants from libc?

Ralith: We use `libc::IPV6_MTU_DISCOVER`, which isn't in 0.2.49. The others will come later. Squashed into the relevant commit for clarity.

Ralith:
> For the retry integrity stuff, placing the interface directly in Session feels weird to me since it does not appear to depend on the TLS implementation at all

That's true, but it does depend on the cryptographic primitives used by the TLS implementation, and is in principle a component of QUIC's cryptographic protocol so I don't think having it there is too outlandish.

> Maybe make it a separate ZST associated type, since it seems to also not depend on any of the other key state?

I feel like this amounts to the same thing, but with extra boilerplate. I'll get rid of the `&self` argument for `is_valid_retry` to enforce decoupling, though.

djc:
The current design is carefully setup such that the symmetric encryption primitives are reusable even without reusing the rustls TLS implementation (I think quiche also uses ring for symmetric crypto while using BoringTLS for TLS). It would be a pity if our retry integrity code was not reusable by non-rustls crypto session impls.

Ralith:
We don't currently expose *any* of the `ring`-only types, which means none of them are reusable regardless.

PR #667, opened 2020-03-08
Clean up documentation

DemiMarie-parity:


Ralith:
Thanks! Unfortunately we raced a bit here; this was also covered in #666.

DemiMarie-parity:
You’re welcome!  Your change was better anyway.

PR #668, opened 2020-03-09
Fix spurious error when finishing while peer closes the connection

Ralith:
If the peer acknowledges the stream FIN and the the Finish future isn't polled between then and the connection becoming closed (guaranteed to be the case if the ACK and the CONNECTION_CLOSE are in the same packet, which is common) then the Finish future would yield an error even though the stream is guaranteed to have been finished.

CC @alecmocatta, this should drastically reduce (although cannot, in principle, eliminate) the frequency of a connection close error being reported when `finish`ing a stream that causes the peer to immediately close the connection.

### DIFF
--- a/quinn/src/streams.rs
+++ b/quinn/src/streams.rs
@@ -100,9 +100,6 @@ where
             conn.check_0rtt()
                 .map_err(|()| WriteError::ZeroRttRejected)?;
         }
-        if let Some(ref x) = conn.error {
-            return Poll::Ready(Err(WriteError::ConnectionClosed(x.clone())));
-        }
         if self.finishing.is_none() {
             conn.inner.finish(self.stream).map_err(|e| match e {
                 proto::FinishError::UnknownStream => WriteError::UnknownStream,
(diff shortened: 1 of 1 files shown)


Ralith:
> (1) Do we have other futures where similarly check for errors early, but maybe shouldn't?

Audited:
- Incoming*Streams, Datagrams: these correctly check for errors only after no streams are found
- Open*: these check for errors early, which I think is fine: opening a stream on a dead connection doesn't seem useful.
- poll_read(_unordered): these correctly check for errors only when no more data is available and the stream is not finished
- poll_write: checks for errors only when becoming blocked. I'm not sure this is ideal, but it shouldn't produce incorrect behavior at least. Maybe worth changing so incautious applications don't fill up buffers needlessly.
- send_datagram: doesn't check the connection's state at all. Maybe worth changing as above.

> (2) Is there something we can change at the quinn-proto level to make sure that non-close events from a packet are yielded definitively before close events?

Good idea; will make a separate PR for this. This shouldn't matter to the high-level API as written, but is certainly less confusing.

Ralith:
See #669.

Issue #670, opened 2020-03-09
Intermittent CONNECTION_ID_LIMIT_ERROR with git master


DemiMarie-parity:
On git master, I intermittently run into a CONNECTION_ID_LIMIT_ERROR which causes my tests to fail every few thousand iterations.  I am not sure if this is a regression in my code or in `quinn-proto`.

Ralith:
This error was introduced in 0bfd15a483d06b316119e4d0cc4b8761e934271f and almost certainly indicates that quinn is sending too many `NEW_CONNECTION_ID` frames due to a bug. I'll investigate tonight.

DemiMarie-parity:
Phew!!!!  Glad to know that it is not a regression in my own code before I spend hours trying to track it down.

Ralith:
As always, a trace log may be helpful for diagnosing this.

DemiMarie-parity:
Still seems to happen even after #671.  Sadly, enabling trace logs seems to make the problem go away.

Ralith:
Thanks for confirming!

DemiMarie-parity:
Got a reproduction: https://gist.github.com/DemiMarie-parity/c83746d3f95f94861a446207f58e0752

Ralith:
Thanks. There's logs for a large number of separate connections there. Could you add tracing spans to your code which distingiush them?

DemiMarie-parity:
Here is a gist with spans: https://gist.github.com/DemiMarie-parity/668727454cef081b34ebc876f669428f

Ralith:
Yeah, I see the error, but there's no way to distinguish log messages from all the different connections operating concurrently, or even the two halves of a single connection. I don't see any new spans in the second gist. To add a span, you need to do something like quinn does [in its driver future](https://github.com/djc/quinn/blob/master/quinn/src/connection.rs#L215-L216).

DemiMarie-parity:
I forgot to provide an ID for the span, sorry.  I am not sure how to distinguish the two halves of one connection, though.

Ralith:
@alecmocatta's repro in #650 was able to reproduce this consistently. This revealed a fundamental error in our approach to storing connection IDs that had historically been masked by formerly loose validity requirements in the draft. I've prepared a fix in #674.

PR #671, opened 2020-03-10
Assorted connection ID handling fixes

Ralith:
In particular, 3fc384c5c3d15fde2a4a525846e8d87555625da5 is a possible fix for #670.


Ralith:
Actually almost certainly not a fix for that bug, but still a bugfix regardless.

Issue #300, opened 2019-04-07
Consider reexporting rustls


Ralith:
rustls `ClientConfig` and `ServerConfig` structs are consumed by quinn-proto's public API, but cannot be constructed by a caller without additionally depending upon rustls. This is an ergonomic wart, and could become particularly annoying if a caller needs a different major version of rustls in their crate than ours. Reexporting the crate, and referencing its types directly rather than through the internal `crypto` facade when they appear in public locations, provides convenient access to exactly the version our interfaces need.

This is contrary to the spirit of our internal efforts to avoid being tightly coupled with rustls, but those efforts aren't currently exposed regardless.

Ralith:
We also have a rustls `SigningKey` in the `EndpointConfig`, though that's easier to paper over with a newtype.

djc:
I'd prefer to just push through to start exposing the traits, I just need to explore the best way to go about it. My main open question is if it's possible to close a type over an internal type parameter.

Ralith:
> I'd prefer to just push through to start exposing the traits

I think we'll still want rustls to Just Work by default at least at the high level API, since TLS is standard QUIC and rustls is by far the most convenient impl. This seems difficult to do without publicly referencing its types, even if only by implementing public traits for them and placing them in default type parameters.

> close a type over an internal type parameter

I'm not sure what this is describing; could you provide an example?

djc:
> I think we'll still want rustls to Just Work by default at least at the high level API, since TLS is standard QUIC and rustls is by far the most convenient impl. This seems difficult to do without publicly referencing its types, even if only by implementing public traits for them and placing them in default type parameters.

I wanted to hang this off `ClientConfig` and `ServerConfig` types that contain `dyn` trait impls to kick off `CryptoSession`s somehow. Seems to me like we can make this ergonomic enough while still allowing users to parametrize with different TLS/crypto implementations.

> I'm not sure what this is describing; could you provide an example?

Basically, can we have a `ConnectionRef` type without type parameters that contains a parametrized `Connection` type (potentially if we `Box` the `Connection`, which might make sense anyway).

Ralith:
> Seems to me like we can make this ergonomic enough while still allowing users to parametrize with different TLS/crypto implementations.

I think this is achievable, I just want to make sure that users will get rustls out of the box if they don't go out of their way for something else.

[...]

PR #672, opened 2020-03-10
Provide an iterator for CertificateChain

kim:
Trying to integrate #644 I realise I'm kind of back to square one, because I can't get at the actual certificate :)

### DIFF
--- a/quinn-proto/src/crypto/types.rs
+++ b/quinn-proto/src/crypto/types.rs
@@ -52,6 +52,14 @@ impl CertificateChain {
     pub fn from_certs(certs: impl IntoIterator<Item = Certificate>) -> Self {
         certs.into_iter().collect()
     }
+
+    /// Iterate over the certificates
+    pub fn iter<'a>(&'a self) -> impl Iterator<Item = Certificate> + 'a {
+        self.certs
+            .iter()
+            .cloned()
+            .map(|cert| Certificate { inner: cert })
+    }
 }
 
 impl std::iter::FromIterator<Certificate> for CertificateChain {
--- a/quinn-proto/src/crypto/types.rs
+++ b/quinn-proto/src/crypto/types.rs
@@ -53,12 +53,9 @@ impl CertificateChain {
         certs.into_iter().collect()
     }
 
-    /// Iterate over the certificates
-    pub fn iter<'a>(&'a self) -> impl Iterator<Item = Certificate> + 'a {
-        self.certs
-            .iter()
-            .cloned()
-            .map(|cert| Certificate { inner: cert })
+    /// An iterator over the certificates in the chain
+    pub fn iter(&self) -> impl Iterator<Item = &rustls::Certificate> {
+        self.certs.iter()
     }
 }
 
@@ -76,6 +73,15 @@ impl std::iter::FromIterator<Certificate> for CertificateChain {
 impl From<Vec<rustls::Certificate>> for CertificateChain {
     fn from(certs: Vec<rustls::Certificate>) -> Self {
         Self { certs }
+    }
+}
+
+impl IntoIterator for CertificateChain {
+    type Item = rustls::Certificate;
+    type IntoIter = std::vec::IntoIter<rustls::Certificate>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.certs.into_iter()
     }
 }
 
--- a/quinn-proto/src/crypto/types.rs
+++ b/quinn-proto/src/crypto/types.rs
@@ -85,6 +85,15 @@ impl IntoIterator for CertificateChain {
     }
 }
 
+impl<'a> IntoIterator for &'a CertificateChain {
+    type Item = &'a rustls::Certificate;
+    type IntoIter = std::slice::Iter<'a, rustls::Certificate>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.certs.iter()
+    }
+}
+
 /// The private key of a TLS certificate to be used by a server
 #[derive(Debug, Clone)]
 pub struct PrivateKey {


quinn-proto/src/crypto/types.rs
```diff
@@ -52,6 +52,14 @@ impl CertificateChain {
...
         certs.into_iter().collect()
     }
+
+    /// Iterate over the certificates
+    pub fn iter<'a>(&'a self) -> impl Iterator<Item = Certificate> + 'a {
+        self.certs
+            .iter()
+            .cloned()
+            .map(|cert| Certificate { inner: cert })
+    }
```
Ralith: I'd prefer for this to either be turned into `into_iter` (consuming `self`) or to finagle things until it can become an `impl Iterator<Item = &'a Certificate>`.

djc: Why is that? Implementing `Iterator` does not seem to make sense here.

Ralith: The trait associated with `into_iter` is `IntoIterator`.

djc: Yeah, wasn't sure about what you meant by "finagle things until it can become an `impl Iterator`. So you mean `impl IntoIterator for &CertificateChain`? That sounds like a plan.

kim: I'm not sure I understand how I could turn an `impl Iterator<Item = &'a rustls::Certificate>` into an `impl Iterator<Item = &'a Certificate>`

djc:
Maybe we should also try to upstream some of the more friendly interfaces from our wrappers into rustls? The current solution feels a bit kludgey.

quinn-proto/src/crypto/types.rs
Ralith: I meant `impl IntoIterator for CertificateChain`, or `fn iter(&self) -> impl Iterator<Item = &Certificate>`, as two reasonable alternatives.

Ralith:
AFAIK the main benefit of these interfaces as they stand is allowing people to avoid exposure to rustls's unstable API, so I'm not sure that would b a win.

quinn-proto/src/crypto/types.rs
Ralith: The latter option would involve storing a `Vec<Certificate>` or similar internally and converting for rustls's convenience as necessary. But I think the former is a good choice if it serves your needs.

kim: Ok thanks.

I was hesitant to make swiping changes, as `CertificateChain::certs` is `pub(crate)`. If that’s appreciated, I’ll try to be a bit more audacious next time :)

djc:
> AFAIK the main benefit of these interfaces as they stand is allowing people to avoid exposure to rustls's unstable API, so I'm not sure that would be a win.

Well, my point is whether we can make a similar API stable in rustls. I filed https://github.com/ctz/rustls/issues/348, feel free to comment there if you have any ideas around this.

djc:
@kim, both of your versions implemented iterators that yield owned `Certificate`s. Is there a reason you're staying away from just yielding `&Certificate`s instead? I think that would be preferable here (for one thing, so the `AuthenticationData` can stay intact).

kim:
@djc Because `CertificateChain` stores `rustls::Certificate`s internally, but that field is also `pub(crate)`, I was afraid of having to introduce a lot of conversion boilerplate if I changed that to store quinn's `Certificate` instead (which is the only way I can see implementing `impl Iterator<Item = &Certificate>` can work).

It turns out that only one callsite is affected (or even exists?), so it's not too bad.

I'm not really familiar with the `impl<'a> Move for &'a Ref` idiom, and also find it confusing when a simple `fn iter(&self) -> impl Iterator<Item = &Ref>` would do, so I implemented the latter in 015c498bee74438da0d6d906b52842f1211f290f

djc:
I'm sorry, I meant to distinguish between iterating over references vs iterating over owned types. I think the current changes probably contain too much stuff. Can we just have an `impl IntoIterator<Item = &rustls::Certificate> for &CertificateChain`? Would that solve your use case?

kim:
Sure. You prefer `impl IntoIterator` over just `fn iter()`?

Ralith:
> Well, my point is whether we can make a similar API stable in rustls.

That would still lead to confusing breakage on each rustls update due to crate version mismatches, unless we reexport rustls per #300 and encourage people to use it that way.

> Sure. You prefer impl IntoIterator over just fn iter()?

Idiomatically you'd have both.

kim:
I feel like we're talking past each other. To summarise:

* #644 generalises #638 to provide access to data negotiated during the cryptographic handshake, as

kim:
I feel like we're talking past each other. To summarise:

* #567 requested to provide access to data negotiated during the cryptographic handshake
* #594 implements this, but only on the `quinn-proto` layer
* #638 proposes to expose access to peer certificates on the `quinn` layer as well
* #644 generalises this, the intention is to not only abstract over the cryptographic protocol, but also hide the implementation detail of `rustls`
* Unfortunately, this actually breaks the original motivation, because the `CertificateChain` newtype doesn't provide access to the underlying certificate data. Not being too familiar with the codebase, I missed this during review, sorry about that. This PR is meant to rectify the situation.
* My assumption about the newtypes in `quinn-proto::crypto::types` was that they are there to _hide_ `rustls`. This seems to be wrong, their purpose is to add convenience methods. I do agree with @burdges that this might be better achieved using "Ext" traits.
* Alright, so it's ok to provide an iterator over `rustls::Certificate`s, and not having to wrap it in the newtype again.
* That's cool, but then I don't get how `impl IntoIterator for &CertificateChain` is useful or idiomatic -- it's the same thing as `self.certs.iter()`, only gives the compiler more work. My understanding is that `Into`/`into` is a convention to indicate that the receiver will be consumed. But maybe I'm missing something.

Alright. So if I provide a `impl IntoIterator for CertificateChain`, would you be willing to accept this patch?

kim:
> So if I provide a `impl IntoIterator for CertificateChain`, would you be willing to accept this patch?

Sorry, to clarify: `fn iter(&self) -> impl Iterator<Item = rustls::Certificate>` and `impl IntoIterator for CertificateChain` where `type Item = rustls::Certificate`.

Ralith:
I think me and @djc have somewhat different perspectives here, and that's produced some confusion. I apologize for that.

> My assumption about the newtypes in quinn-proto::crypto::types was that they are there to hide rustls.

This was my intention in adding those newtypes to begin with, though the specific goal was to *allow* callers to be ignorant of rustls, not necessarily to force them to--so exposing the details is acceptable so long as it's not required for  simple cases.

> I don't get how impl IntoIterator for &CertificateChain is useful or idiomatic -- it's the same thing as self.certs.iter(), only gives the compiler more work.

`IntoIterator` on references is the a common idiom because it makes `for x in &xs` work, which is a bit more pleasant than `for x in xs.iter()`. You'll see this pattern in every std collection, e.g. for slices and `Vec`.

> So if I provide a impl IntoIterator for CertificateChain, would you be willing to accept this patch?

I'd be happy to merge either/both of those, though I'd consider an approach which doesn't expose rustls to be better still. An easier solution than wrapping/unwrapping the newtypes might be to fall back on `Vec<u8>` for owning iterators and `&[u8]` for borrowing ones.

kim:
> IntoIterator on references is the a common idiom because it makes for x in &xs work

Ok, thanks. Please forgive my ignorance, I'm coming from a programming language whose main compiler can auto-derive _every_ typeclass through a newtype 🤷‍♂

Ralith:
No worries, always happy to help someone get up to speed!

djc:
Yeah, sorry for the confusion and mixed messages here, and thanks for sticking with it!

PR #674, opened 2020-03-11
Fixed active connection ID window

### LINKED ISSUE #670: Intermittent CONNECTION_ID_LIMIT_ERROR with git master

Ralith:
Fixes #670. This was ~a long-standing bug~ multiple long-standing bugs revealed by the stricter protocol requirements in draft 27.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -21,7 +21,7 @@ use crate::{
     range_set::RangeSet,
     shared::{
         ConnectionEvent, ConnectionEventInner, ConnectionId, EcnCodepoint, EndpointConfig,
-        EndpointEvent, EndpointEventInner, IssuedCid, ServerConfig, TransportConfig,
+        EndpointEvent, EndpointEventInner, ResetToken, ServerConfig, TransportConfig,
     },
     spaces::{CryptoSpace, PacketSpace, Retransmits, SentPacket},
     streams::{self, FinishError, ReadError, Streams, UnknownStream, WriteError},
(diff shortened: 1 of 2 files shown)


Ralith:
On review this still isn't quite right, but it's close. Need to make sure the currently selected CID is always the lower end of the window. Will revise tomorrow.

quinn-proto/src/connection.rs
```diff
@@ -151,7 +153,12 @@ where
 
     streams: Streams,
     /// Surplus remote CIDs for future use on new paths
-    rem_cids: Vec<IssuedCid>,
+    ///
+    /// Sequence numbers start at `self.first_unretired_cid` and are contiguous, including `None`
+    /// slots. Always exactly REM_CID_COUNT elements long. Maintaining a fixed-size window allows us
+    /// to easily discard duplicate and obsolete NEW_CONNECTION_ID frames fast and with constant
+    /// memory.
+    rem_cids: VecDeque<Option<(ConnectionId, ResetToken)>>,
```
djc: I find the design here quite complex. Did you think about storing the sequence numbers in the deque along with the other state instead (it would become `VecDeque<(u64, Option<(ConnectionId, ResetToken)>)>`, I guess)? This should obviate the need for the separate `first_unretired_cid`. It would take a bit of extra memory, but 64 every bytes in the heap probably won't kill us.

(Perhaps `(u64, Option<(ConnectionId, ResetToken)>)` should become a proper struct instead, in which case we can also derive `Default` for it.)

Ralith: We're specifically trying to represent a continuous sliding window. I'm hence not a fan of explicit sequence numbers, because they allow a far larger number of invalid states to be represented. The latest iteration of this is hopefully a bit clearer.

djc:
Yeah, that does look better. I do feel like this would probably be better as an explicit struct with an abstract interface wrapped around it, since it seems like it could hide a lot of the complexity. If we do that, it might also be more feasible to use an array with a cursor pointer rather than a `VecDeque<Option<>>`.

Issue #673, opened 2020-03-10
H3 tests::incoming_request_stream_ends_on_client_closure hanging under coverage


djc:
For example, here: https://github.com/djc/quinn/runs/497269815?check_suite_focus=true

djc:
tests::incoming_request_stream_closed_on_client_drop seems to be similarly intermittent.

stammw:
I think it's caused by the same problem as the one I'm trying to solve on the 0-RTT PR. Seems to me that CONNECTION_CLOSE makes some of the peer's API hang forever under certain conditions.

Maybe we can disable them on coverage or globally ?

Ralith:
Ideally you'd fix the hang (since users will be affected by it as well), but marking known-broken tests as ignored is acceptable in the spirit of keeping other work unobstructed if the issue is subtle.

stammw:
@Ralith Do you think #686 could have an influence on this, by any chance?

I reactivated the faulty tests on master and ran them several times under tarpaulin, which did not reproduce the timeouts.

djc:
Maybe create a PR to reactivate them and rerun the CI jobs a few times (there's a button in the upper right corner of the Actions page)?

stammw:
Done on #687

Ralith:
I haven't investigated the h3 tests, so I could only guess. It's certainly possible, though!

djc:
Let's close this for now, then.

PR #677, opened 2020-03-12
H3: Temporarily disable async tests in coverage

stammw:
As a quick response to #673, let everyone use the CI as async tests are broken and tiemout or fail when run under tarpaulin.

Fixing the test is currently addressed in the mean time.


djc:
I think you want a `!` in there to make the attribute apply to the whole module?

stammw:
Fixed.

PR #678, opened 2020-03-12
Bump

djc:


Ralith:
Let's get #679 in too.

djc:
Can you approve it? I'll hold off publishing until I have your acknowledgement that we're ready.

PR #679, opened 2020-03-12
Account for active CIDs when issuing new ones

Ralith:


Ralith:
Lint failure is unrelated and corrected in #678.

PR #680, opened 2020-03-13
Replace criterion with bencher

Ralith:
I never found criterion's statistics to be correct, and it has a *lot* of dependencies. Results in 32 fewer dev-dependencies.


djc:
LGTM!

PR #681, opened 2020-03-13
Enable hq on interop server

stammw:
So the interop server accepts both protocols.


stammw:
Ah, forgot  to rebase. The coverage is stuck.

stammw:
Here, the coverage now passes.

quinn-h3/src/server.rs
```diff
@@ -164,6 +164,15 @@ impl Future for Connecting {
...
 
+impl From<quinn::Connecting> for Connecting {
+    fn from(connecting: quinn::Connecting) -> Self {
+        Self {
+            connecting,
+            settings: Settings::default(),
+        }
+    }
+}
+
```
Ralith: Should there also be an explicit constructor that allows custom `Settings` to be specified?

interop/src/h3_server.rs
```diff
@@ -55,29 +54,40 @@ async fn main() -> Result<()> {
...
+    let listen = "[::]:4433".parse()?;
+    let (_, mut incoming) = endpoint_builder.bind(&listen)?;
 
     println!("server listening");
     while let Some(connecting) = incoming.next().await {
         tokio::spawn(async move {
-            if let Err(e) = handle_connection(connecting).await {
-                error!("handling connection failed: {:?}", e)
+            let protos = connecting.authentication_data().protocol.unwrap();
+            dbg!(&protos);
```
Ralith: Stray debug print?

stammw: ditched it, thanks.

quinn-h3/src/server.rs
stammw: Made this possible in `Connecting::from_quic()`.

quinn-h3/src/server.rs
```diff
@@ -8,7 +8,7 @@ use std::{
 
 use futures::{ready, Stream};
 use http::{response, Request, Response};
-use quinn::{CertificateChain, EndpointBuilder, PrivateKey, RecvStream, SendStream, self};
+use quinn::{self, CertificateChain, EndpointBuilder, PrivateKey, RecvStream, SendStream};
```
djc: This really belongs in the previous commit.

quinn/src/connection.rs
```diff
@@ -85,6 +85,16 @@ where
             Err(self)
         }
     }
+
+    /// Data conveyed by the peer during the handshake, including cryptographic identity
```
djc: Should this have a documentation caveat about what the potential for this to be incomplete?

interop/src/h3_server.rs
```diff
@@ -55,29 +54,39 @@ async fn main() -> Result<()> {
...
     println!("server listening");
     while let Some(connecting) = incoming.next().await {
         tokio::spawn(async move {
-            if let Err(e) = handle_connection(connecting).await {
-                error!("handling connection failed: {:?}", e)
+            let protos = connecting.authentication_data().protocol.unwrap();
+            println!("server received connection");
+
+            if protos == b"h3-27" {
+                if let Err(e) = handle_connection(connecting).await {
```
djc: Shall we rename `handle_connection()` to `h3_handle_connection()`, for symmetry with `hq_handle_connection()`? (If I were nitpicking I'd also rename them `handle_{h3,hq}_connection()` rather than having the protocol prefix up front.)

quinn/src/connection.rs
stammw: Would `As the handshake is ongoing, the returned data is to be considered potentially incomplete.` do the job? I don't know much about all this, I just copied the inner comment.

quinn-h3/src/server.rs
stammw: rebased in previous commit.

interop/src/h3_server.rs
stammw: Got them renamed. 

I prefer to use prefix as kind of a namespace to highlight which part of the program this belongs.

quinn/src/connection.rs
djc: Maybe something like `Since the handshake is incomplete at this point, the returned data is likely to be incomplete as well.`?

quinn/src/connection.rs
```diff
@@ -344,6 +357,9 @@ where
     }
 
     /// Data conveyed by the peer during the handshake, including cryptographic identity
+    ///
+    /// Since the handshake is incomplete at this point, the returned data is likely to be
```
djc: Should not be there on `Connection`, only for `Connecting`.

stammw: Oops, yes sure.

stammw: Fixed

Ralith:
Looks like coverage is hung for some reason?

djc:
Looks like the same tests that ought to be disabled on master.

PR #684, opened 2020-03-14
Remove obsolete azure pipelines badge

Ralith:


djc:
Yup. Unfortunately https://github.com/rust-lang/crates.io/pull/1838 isn't moving much. :(

PR #686, opened 2020-03-17
Fix busyloop in timer handling

Ralith:
Found this while experimenting.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -229,13 +229,12 @@ where
         let _guard = span.enter();
 
         loop {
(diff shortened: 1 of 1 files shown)


Ralith:
I think so, yeah; my application was hitting it pretty regularly.

djc:
Maybe let's wait until we figure out what's up with #689 (and possibly #688).

DemiMarie-parity:
@djc @Ralith as with all of my issues filed to date, mine only applies to quinn-proto, so it is okay to make a quinn release with this fix.

PR #687, opened 2020-03-17
H3: Test re-enabling h3 async tests after busy-loop fix

stammw:
To see if #686 has fixed them.


stammw:
Seems to pass, I'm not sure Github will let me rerun the coverage once it has passed.

PR #690, opened 2020-03-19
Retry port for the interop server

stammw:
Make the interop server listen on `4434` with stateless retry activated.


stammw:
I added `443` as some implementers use it now.

djc:
Nice work!

PR #691, opened 2020-03-21
H3 expose missing features in the public API

stammw:
Response cancellation and GoAway were not exposed to the public API and the former was unfinished. 

This PR is based on #653, so we should merge this one first.

### DIFF
--- a/quinn-h3/src/lib.rs
+++ b/quinn-h3/src/lib.rs
@@ -28,9 +28,9 @@ pub mod server;
 mod frame;
 mod streams;
 
-use quinn::WriteError;
-
 use err_derive::Error;
+use quinn::{ApplicationClose, ConnectionError, ReadError, WriteError};
+use std::io::ErrorKind;
 
 use proto::ErrorCode;
 
@@ -44,7 +44,9 @@ pub enum Error {
     Proto(proto::connection::Error),
     #[error(display = "QUIC protocol error: {}", _0)]
     Quic(quinn::ConnectionError),
-    #[error(display = "QUIC write error: {}", _0)] // TODO to be refined
+    #[error(display = "QUIC read error: {}", _0)]
+    Read(ReadError),
+    #[error(display = "QUIC write error: {}", _0)]
     Write(WriteError),
     #[error(display = "Internal error: {}", _0)]
     Internal(String),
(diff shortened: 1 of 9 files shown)


quinn-h3/src/qpack/encoder.rs
```diff
@@ -145,6 +145,9 @@ pub fn on_decoder_recv<R: Buf>(table: &mut DynamicTable, read: &mut R) -> Result
     while let Some(instruction) = parse_instruction(read)? {
         match instruction {
             Instruction::Untrack(stream_id) => table.untrack_block(stream_id)?,
+            Instruction::UntrackInfaillible(stream_id) => {
```
djc:
```suggestion
            Instruction::UntrackInfallible(stream_id) => {
```

This is probably a gallicism, you made a similar mistake in some of your earlier commit messages. :)

(BTW, the commit message also has a typo, `streaam` -> `stream`.)

The vocabulary "untrack infallible" seems a little hard to understand to me. Is this what the spec calls it?

quinn-h3/src/lib.rs
```diff
@@ -95,27 +99,52 @@ impl From<quinn::ConnectionError> for Error {
     }
 }
 
+impl From<ErrorCode> for Error {
+    fn from(code: ErrorCode) -> Self {
+        match code {
+            ErrorCode::REQUEST_CANCELLED => Error::RequestCancelled,
+            ErrorCode::REQUEST_REJECTED => Error::RequestRejected,
+            _ => panic!("wut"),
```
djc: Maybe make the `panic!()` message a little more descriptive here? Or otherwise use `unreachable!()` or `unimplemented!()` or some such.

stammw: Oh, sure. (at least this gave me a good laugh)

quinn-h3/src/qpack/encoder.rs
stammw: This is not really defined with a specific terminology in the draft. It only states that stream cancellation instruction on unknown or already deleted stream-related reference block can be ignored. 

Now I figure out this should have another behaviour than `Untrack` nevertheless, at it only removes one block of reference for the stream (there are one for headers and trailers respectively). I'll just rename this to `StreamCancel` or something similar.

stammw:
I rewrote the error management commit to handle more error-path possibilities. There might still be some missing ones, but I feel error management starts to be pretty decent now.

PR #683, opened 2020-03-14
Remote connection ID refactoring

Ralith:

Issue #675, opened 2020-03-11
Test in CI with randomized packet loss/duplication/reordering


Ralith:
e.g. #650 has compellingly illustrated the value here. Ideally we use a fixed seed and run some large number of iterations in CI for reproducibility, but a brute force netem-based solution as in the example is manifestly far better than nothing.

alecmocatta:
A good starting point might be the below in conjunction with `tc qdisc add dev lo root netem delay 5ms 10ms 25% distribution normal loss 5% duplicate 5% reorder 40% 50%`. I believe this would surface all the issues I've observed so far.

```rust
use futures::{future::select_all, lock::Mutex, stream::FuturesUnordered, FutureExt, StreamExt};
use quinn::{
    Certificate, ClientConfig, ClientConfigBuilder, ConnectionError, NewConnection, RecvStream, ServerConfigBuilder, WriteError
};
use rand::{rngs::SmallRng, Rng, SeedableRng};
use std::{
    future::Future, net::{IpAddr, Ipv6Addr, SocketAddr, UdpSocket}, pin::Pin, sync::Arc
};

#[tokio::main]
async fn main() {
    tracing_subscriber::fmt::init();

    let mut rng = SmallRng::seed_from_u64(0);

    let cert = rcgen::generate_simple_self_signed(vec!["localhost".into()]).unwrap();
    let key = quinn::PrivateKey::from_der(&cert.serialize_private_key_der()).unwrap();
    let cert = Certificate::from_der(&cert.serialize_der().unwrap()).unwrap();
    let cert_chain = quinn::CertificateChain::from_certs(vec![cert.clone()]);

    let mut transport = quinn::TransportConfig::default();
    transport.max_idle_timeout(None).unwrap();
    let transport = Arc::new(transport);

    let mut server_config = quinn::ServerConfig::default();
    server_config.transport = transport.clone();
    let mut server_config = ServerConfigBuilder::new(server_config);
    server_config.certificate(cert_chain, key).unwrap();
    let server_config = server_config.build();

    let mut server_endpoint = quinn::Endpoint::builder();
    server_endpoint.listen(server_config);
    let (server_endpoint, incoming) =
        server_endpoint.bind(&SocketAddr::new(IpAddr::V6(Ipv6Addr::LOCALHOST), 0)).unwrap();
    let server_addr = server_endpoint.local_addr().unwrap();
    let incoming = Mutex::new(incoming);

    let (client_endpoint, _) = quinn::Endpoint::builder()
[...]

Issue #15, opened 2018-05-22
Use fuzzing for testing


djc:
Initially, can use this to test decoding routines. There's an example at https://github.com/djc/tokio-imap/tree/master/imap-proto/fuzz that might be useful in getting this off the ground.

spacejam:
Happy to lend a hand on this. Which specific decoding methods are high priority for this?

djc:
Packet decoding by way of `Endpoint::handle()` would be a good starting point.

Ralith:
Naively we could just throw fuzzer data as UDP packets at the implementation, but that wouldn't get us very far because the fuzzer is spectacularly unlikely to make any progress through the cryptographic handshake, leaving the bulk of the implementation untested. A good solution for end-to-end testing might be to add a hook to inject arbitrary data into the plaintext of an arbitrary packet between arbitrary frames, then use that in a fuzzing target to perturb an otherwise valid scripted session like the lifecycle unit test.

Ralith:
Now that we've gracefully encapsulated the stream logic into the `Streams` type, it would make an excellent candidate for targeted fuzzing without any elaborate measures needed. As illustrated in #774, there's definitely enough complexity there to merit it.

stammw:
A fuzzing effort has been made for `VarInt` on [h3](https://github.com/hyperium/h3/pull/16). Would you be interested to import it ?

Ralith:
Certainly! I'm going to go ahead and close this since we've got some fuzzing going now; we should open new issues for specific interfaces of interest.

hmuint32:
i am having compilation issue with: cargo fuzz check in one of my local dev with stable rust, should i open a new issue?
```

92 |     use arbitrary::{Arbitrary, Result, Unstructured};
   |         ^^^^^^^^^ use of undeclared crate or module `arbitrary`

error[E0432]: unresolved import `arbitrary`
  --> /localpath/.cargo/registry/src/index.crates.io-6f17d22bba15001f/quinn-proto-0.10.6/src/lib.rs:92:9
   |
92 |     use arbitrary::{Arbitrary, Result, Unstructured};
   |         ^^^^^^^^^ use of undeclared crate or module `arbitrary`

    Checking zbus v3.10.0
For more information about this error, try `rustc --explain E0432`.
error: could not compile `quinn-proto` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
[...]

PR #692, opened 2020-03-22
Connection ID fixes

### LINKED ISSUE #689: Intermittent timeout with git master
Logs at https://gist.github.com/DemiMarie-parity/c83746d3f95f94861a446207f58e0752

Ralith:
Fixes #689, which turned out to be a regression introduced in #683.

This bug caused unintended `RETIRE_CONNECTION_ID` frames to be sent for active remote CID when a duplicate `NEW_CONNECTION_ID` frame was received containing it. The peer responded to that packet with a `NEW_CONNECTION_ID` to replace the retired CID, triggering a false `CONNECTION_ID_LIMIT_ERROR`. Additionally, future packets received by the peer elicited a stateless reset, as they still bore the retired CID. These stateless resets raced with the `NEW_CONNECTION_ID`, sometimes causing an unexpected connection reset before the `CONNECTION_ID_LIMIT_ERROR` was triggered.

Thanks to @DemiMarie-parity for the brute-force testing that identified this issue! As an issue that only arose when certain packets are spuriously retransmitted, this is another example of the value to be had in #675.

### DIFF
--- a/quinn-proto/src/cid_queue.rs
+++ b/quinn-proto/src/cid_queue.rs
@@ -15,6 +15,9 @@ pub struct CidQueue {
     /// Index at which circular buffer addressing is based
     cursor: usize,
     /// Sequence number of `self.buffer[cursor]`
+    ///
+    /// The CID sequenced immediately prior to this is the active CID, which this data structure is
+    /// not responsible for retiring.
     offset: u64,
 }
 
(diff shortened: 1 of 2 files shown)


DemiMarie-parity:
Thank you for the fix!  Testing it now.  Would it be possible to find such bugs more systematically, such as by fuzzing?

DemiMarie-parity:
Indeed this has passed every test I have thrown at it, although my tests are not as good as something like #675.

Ralith:
#675 would be a good start, for sure. Although I don't think it would trigger spurious retransmit processing bugs in particular very often, we'd certainly like to have fuzzing as well; see discussion in #15. If you're interested, contributions in this area would be very welcome.

PR #695, opened 2020-03-29
Fix initial data limit for remotely initiated bidirectional streams

### LINKED ISSUE #694: Can't send/receive more than `stream_window_bidi` roundtrip messages without the connection timing out
Hi all,

I'm currently trying to write my own QUIC server and client based on quinn v0.6.0, and I can't make more than `stream_window_bidi` roundtrip messages without the connection timing out. I have a minimal reproduction that narrows it down, but I'm unsure what is causing the problem.

Say, for example, that `stream_window_bidi` is set to 10 on the server side. The first 10 bidi streams the client opens on the connection work fine. On the 11th, the client is able to send the data, the server is able to receive the data, but the server hangs trying to write the response to the send stream, and the connection eventually times out. Enabling keepalive just makes the server hang forever. 

I've got a minimal reproduction repo available at https://github.com/SriRamanujam/quinn_minimal_network_repro. `stream_window_bidi` has been set to 10 on the server side. The client simply makes requests in a loop once a second. I've also attached traces for the client and server. 

[client_trace.log](https://github.com/djc/quinn/files/4398109/client_trace.log)
[server_trace.log](https://github.com/djc/quinn/files/4398110/server_trace.log)

/cc @Ralith

Ralith:
Although we were correctly setting the data limit for the send half of initially-available remotely-initiated bidirectional streams, we neglected to apply the limit to streams made available by stream ID flow control updates. As a result, the application would be erroneously prevented from writing to the send half of bidirectional streams opened by the peer after the initial stream_window_bidi streams.

Fixes #694.

### DIFF
--- a/quinn-proto/src/streams.rs
+++ b/quinn-proto/src/streams.rs
@@ -5,7 +5,8 @@ use err_derive::Error;
 use tracing::debug;
 
 use crate::{
-    assembler::Assembler, frame, range_set::RangeSet, Dir, Side, StreamId, TransportError, VarInt,
+    assembler::Assembler, frame, range_set::RangeSet, transport_parameters::TransportParameters,
+    Dir, Side, StreamId, TransportError, VarInt,
 };
 
 pub(crate) struct Streams {
@@ -40,21 +41,21 @@ impl Streams {
 
         for dir in Dir::iter() {
             for i in 0..this.max_remote[dir as usize] {
-                this.insert(true, StreamId::new(!side, dir, i));
+                this.insert(None, true, StreamId::new(!side, dir, i));
             }
         }
 
         this
     }
 
-    pub fn open(&mut self, side: Side, dir: Dir) -> Option<StreamId> {
+    pub fn open(&mut self, params: &TransportParameters, side: Side, dir: Dir) -> Option<StreamId> {
         if self.next[dir as usize] >= self.max[dir as usize] {
             return None;
         }
 
         self.next[dir as usize] += 1;
         let id = StreamId::new(side, dir, self.next[dir as usize] - 1);
-        self.insert(false, id);
+        self.insert(Some(params), false, id);
         self.send_streams += 1;
         Some(id)
     }
@@ -63,10 +64,10 @@ impl Streams {
         self.send_streams
     }
 
-    pub fn alloc_remote_stream(&mut self, side: Side, dir: Dir) {
+    pub fn alloc_remote_stream(&mut self, params: &TransportParameters, side: Side, dir: Dir) {
         self.max_remote[dir as usize] += 1;
         let id = StreamId::new(!side, dir, self.max_remote[dir as usize] - 1);
-        self.insert(true, id);
+        self.insert(Some(params), true, id);
     }
 
     pub fn accept(&mut self, side: Side, dir: Dir) -> Option<StreamId> {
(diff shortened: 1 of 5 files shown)


Ralith:
Added a unit test for this case.

seamlik:
By building against this branch in my project, I can confirm this fix works. Thank you!

quinn-proto/src/streams.rs
```diff
@@ -40,21 +41,21 @@ impl Streams {
...
-                this.insert(true, StreamId::new(!side, dir, i));
+                this.insert(None, true, StreamId::new(!side, dir, i));
             }
         }
 
         this
     }
 
-    pub fn open(&mut self, side: Side, dir: Dir) -> Option<StreamId> {
+    pub fn open(&mut self, params: &TransportParameters, side: Side, dir: Dir) -> Option<StreamId> {
```
djc: Curious why you decided to put `&TransportParameters` first here as well as in `alloc_remote_stream()` and `insert()`. I'd have put it last as less essential for the purpose of the method. Any thoughts?

quinn-proto/src/streams.rs
```diff
@@ -201,10 +202,17 @@ impl Streams {
...
-    fn insert(&mut self, remote: bool, id: StreamId) {
+    fn insert(&mut self, params: Option<&TransportParameters>, remote: bool, id: StreamId) {
         let bi = id.dir() == Dir::Bi;
         if bi || !remote {
-            assert!(self.send.insert(id, Send::new()).is_none());
+            let max_data = params.map_or(0, |params| match id.dir() {
+                Dir::Uni => params.initial_max_stream_data_uni,
+                // Remote/local appear reversed here because the transport parameters are named from
+                // the perspective of the peer.
+                Dir::Bi if remote => params.initial_max_stream_data_bidi_local,
```
djc: You seem to have `initial_max_stream_data_bidi_local` twice here. Probably not intended (especially given the comment)?

djc:
(If you do another take, include version bumps to 0.6.1 for quinn and quinn-proto? I think we can just release these from master, since there are no API changes/removals?)

quinn-proto/src/streams.rs
Ralith: I default to sorting function parameters by frequency of change descending, but I'm not particularly attached to that pattern. Do you have a strong opinion?

quinn-proto/src/streams.rs
Ralith: Whoops, good catch!

Ralith:
Version bump added.

SriRamanujam:
This also fixes the issue for me. Thanks so much for the quick fix!

Ralith:
Thanks for verifying! Detailed, easily-reproducible bug reports like you supplied are what make this sort of response possible.

quinn-proto/src/streams.rs
djc: I guess not that strong. Not sure what frequency of change means in this context, but doesn't the `params` argument change less here than the other ones?

Ralith: First draft of that comment had an error, meant to type "ascending". The habit arises from my Haskell days, where it plays nicely with currying, but that's not relevant in rust. However, it is consistent with the prior form of the function, since `side` doesn't vary within the lifetime of a connection either.

PR #699, opened 2020-04-02
Store outgoing stream data with stream state

### LINKED ISSUE #561: Compact multiple small write calls into fewer STREAM frames
Presently, every application write produces at least one stream frame. This can add significant bandwidth overhead for very small writes. While repeated very small writes are not optimal in any case, we should do better.

One approach would be to maintain a buffer of pending data for each open outgoing stream, from which stream frames can be computed on demand. For efficiency, a list of streams with nonempty buffers should also be maintained. A simple implementation could leave retransmits to the existing logic, while a more advanced one might take responsibility for those as well.

Ralith:
Improves stream frame packing and establishes groundwork for intelligent selection of which stream to transmit for.

Fixes #561.

This is a major internal refactoring; prior drafts broke a bunch of tests and non-obvious further breakage is very possible. In the interest of making a robust release, let's take care not to merge this prior to 0.6.1.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -898,6 +898,17 @@ where
             "only streams supporting outgoing data may be reset"
         );
 
+        // Drain queued data
+        let unacked_data = &mut self.unacked_data;
+        self.streams.pending.retain(|frame| {
+            if frame.id == stream_id {
+                *unacked_data -= frame.data.len() as u64;
+                false
+            } else {
+                true
+            }
+        });
+
         let stop_reason = if stopped { Some(error_code) } else { None };
         let status = self.streams.reset(stream_id, stop_reason);
         let was_conn_blocked = self.blocked_streams.remove(&stream_id);
@@ -2599,14 +2610,6 @@ where
                 Some(x) => x,
                 None => break,
             };
-            if self
-                .streams
-                .send_mut(stream.id)
-                .map_or(true, |s| s.state.was_reset())
-            {
-                self.unacked_data -= stream.data.len() as u64;
-                continue;
-            }
             let len = cmp::min(
                 stream.data.len(),
                 max_size as usize - buf.len() - frame::Stream::SIZE_BOUND,
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -498,28 +498,13 @@ where
         }
 
         for frame in info.stream_frames {
-            let ss = match self.streams.send_mut(frame.id) {
-                Some(x) => x,
-                // ACK for a closed stream is a noop
-                None => continue,
-            };
-            ss.pending.ack(frame.offsets.clone());
             self.unacked_data -= frame.offsets.end - frame.offsets.start;
-            if let streams::SendState::DataSent {
-                ref mut finish_acked,
-            } = ss.state
-            {
-                if frame.fin {
-                    *finish_acked = true;
-                }
-                if *finish_acked && ss.pending.in_flight() == 0 {
-                    ss.state = streams::SendState::DataRecvd;
-                    self.streams.maybe_cleanup(frame.id);
-                    self.events.push_back(Event::StreamFinished {
-                        stream: frame.id,
-                        stop_reason: None,
-                    });
-                }
+            let id = frame.id;
+            if self.streams.ack(frame) {
+                self.events.push_back(Event::StreamFinished {
+                    stream: id,
+                    stop_reason: None,
+                });
             }
         }
     }
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -2589,38 +2589,14 @@ where
 
         // STREAM
         while buf.len() + frame::Stream::SIZE_BOUND < max_size && space_id == SpaceId::Data {
-            let max_data_len = max_size as usize - buf.len() - frame::Stream::SIZE_BOUND;
-            let id = match self.streams.pending.pop() {
+            let meta = match self.streams.poll_transmit(max_size as usize - buf.len()) {
                 Some(x) => x,
                 None => break,
             };
-            let stream = match self.streams.send_mut(id) {
-                Some(s) => s,
-                // Stream was reset with pending data and the reset was acknowledged
-                None => continue,
-            };
-            // Reset streams aren't removed from the pending list and still exist while the peer
-            // hasn't acknowledged the reset, but should not generate STREAM frames, so we need to
-            // check for them explicitly.
-            if stream.is_reset() {
-                continue;
-            }
-            let offsets = stream.pending.poll_transmit(max_data_len);
-            let fin = offsets.end == stream.pending.offset()
-                && mem::replace(&mut stream.fin_pending, false);
-            if stream.is_pending() {
-                self.streams.pending.push(id);
-            }
-            let meta = frame::StreamMeta { id, offsets, fin };
             trace!(id = %meta.id, off = meta.offsets.start, len = meta.offsets.end - meta.offsets.start, fin = meta.fin, "STREAM");
             meta.encode(true, buf);
-            buf.put_slice(
-                self.streams
-                    .send_mut(id)
-                    .unwrap()
-                    .pending
-                    .get(meta.offsets.clone()),
-            );
+            let ss = self.streams.send_mut(meta.id).unwrap();
+            buf.put_slice(ss.pending.get(meta.offsets.clone()));
             stream_frames.push(meta);
         }
 
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -849,22 +849,6 @@ where
         self.timers.set(Timer::KeepAlive, now + interval);
     }
 
-    fn queue_stream_data(&mut self, stream: StreamId, data: &[u8]) -> Result<(), WriteError> {
-        let ss = self
-            .streams
-            .send_mut(stream)
-            .ok_or(WriteError::UnknownStream)?;
-        assert_eq!(ss.state, streams::SendState::Ready);
-        let was_pending = ss.is_pending();
-        ss.pending.write(data);
-        if !was_pending {
-            self.streams.pending.push(stream);
-        }
-        self.data_sent += data.len() as u64;
-        self.unacked_data += data.len() as u64;
-        Ok(())
-    }
-
     /// Abandon transmitting data on a stream
     ///
     /// # Panics
@@ -1733,13 +1717,9 @@ where
                         ));
                     }
                     if let Some(ss) = self.streams.send_mut(id) {
-                        // We only care about budget *increases* for *live* streams
-                        if offset > ss.max_data && ss.state == streams::SendState::Ready {
-                            trace!(stream = %id, old = ss.max_data, new = offset, current_offset = ss.offset(), "stream limit increased");
-                            if ss.offset() == ss.max_data {
-                                self.events.push_back(Event::StreamWritable { stream: id });
-                            }
-                            ss.max_data = offset;
+                        if ss.increase_max_data(offset) {
+                            // Unblocked
+                            self.events.push_back(Event::StreamWritable { stream: id });
                         }
                     } else if id.initiator() == self.side() && self.streams.is_local_unopened(id) {
                         debug!("got MAX_STREAM_DATA on unopened {}", id);
@@ -2906,31 +2886,14 @@ where
             return Err(WriteError::Blocked);
         }
 
-        let budget_res = self
-            .streams
-            .send_mut(stream)
-            .ok_or(WriteError::UnknownStream)?
-            .write_budget();
-
-        let stream_budget = match budget_res {
-            Ok(budget) => budget,
-            Err(e @ WriteError::Stopped { .. }) => {
-                self.streams.maybe_cleanup(stream);
-                return Err(e);
-            }
-            Err(e @ WriteError::Blocked) => {
-                trace!(%stream, "write blocked by flow control");
-                return Err(e);
-            }
-            Err(WriteError::UnknownStream) => unreachable!("not returned here"),
-        };
-
-        let conn_budget = cmp::min(
-            self.max_data - self.data_sent,
-            self.config.send_window - self.unacked_data,
-        );
-        let n = conn_budget.min(stream_budget).min(data.len() as u64) as usize;
-        self.queue_stream_data(stream, &data[0..n])?;
+        // TODO: Send window in terms of actual buffer size, which includes discontinuously acked
+        // data
+        let len = (data.len() as u64)
+            .min(self.max_data - self.data_sent)
+            .min(self.config.send_window - self.unacked_data) as usize;
+        let n = self.streams.write(stream, &data[0..len])?;
+        self.data_sent += n as u64;
+        self.unacked_data += n as u64;
         trace!(%stream, "wrote {} bytes", n);
         Ok(n)
     }
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -7,7 +7,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use bytes::{Bytes, BytesMut};
+use bytes::{BufMut, Bytes, BytesMut};
 use err_derive::Error;
 use rand::{rngs::StdRng, Rng, SeedableRng};
 use tracing::{debug, error, info, trace, trace_span, warn};
@@ -500,10 +500,11 @@ where
         for frame in info.stream_frames {
             let ss = match self.streams.send_mut(frame.id) {
                 Some(x) => x,
+                // ACK for a closed stream is a noop
                 None => continue,
             };
-            ss.bytes_in_flight -= frame.data.len() as u64;
-            self.unacked_data -= frame.data.len() as u64;
+            ss.pending.ack(frame.offsets.clone());
+            self.unacked_data -= frame.offsets.end - frame.offsets.start;
             if let streams::SendState::DataSent {
                 ref mut finish_acked,
             } = ss.state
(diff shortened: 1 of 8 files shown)


quinn-proto/src/connection.rs
```diff
@@ -2605,41 +2553,21 @@ where
         }
 
         // STREAM
-        while buf.len() + frame::Stream::SIZE_BOUND < max_size {
-            let mut stream = match space.pending.stream.pop_front() {
-                Some(x) => x,
-                None => break,
-            };
-            if self
+        while buf.len() + frame::Stream::SIZE_BOUND < max_size && space_id == SpaceId::Data {
```
djc, line 2556: The restriction to `SpaceId::Data` is new, but seems mostly unrelated to this change?

quinn-proto/src/send_buffer.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    pub fn retransmit_all_for_0rtt(&mut self) {
+        debug_assert_eq!(self.offset, self.unacked.len() as u64);
+        self.unsent = 0;
+    }
+
+    /// First stream offset unwritten by the application, i.e. the offset that the next write will
+    /// begin at
+    pub fn offset(&self) -> u64 {
```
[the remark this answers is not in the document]
Ralith: My objective with these getters is to express that these values are only ever written by `SendStream`. In the past we were munging stream state all over the place and this is part of my effort to more strongly encapsulate it. If there was a way to directly modify that data from elsewhere I'd agree that accessors are redundant.

quinn-proto/src/connection.rs
Ralith: This requirement has always existed, but in the past it was accomplished by never writing to `pending.stream` for spaces other than `Data`. With this change, stream buffers are (rightly) no longer isolated per-space, so we need to be more explicit, or else stream frames would illegally arise e.g. in `Initial` packets.

quinn-proto/src/frame.rs
```diff
@@ -414,10 +417,18 @@ impl FrameStruct for Stream {
     const SIZE_BOUND: usize = 1 + 8 + 8 + 8;
 }
 
-impl Stream {
+/// Metadata from a stream frame
```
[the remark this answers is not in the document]
Ralith: This specifically represents the metadata/header portion of a stream frame, rather than the state of a logical stream. Note that it's defined in the `frame` module and we only ever use qualified names to reference it, much like other frame types. I split it out so that `Streams::poll_transmit` does not have to construct a `Bytes`, which would require copying.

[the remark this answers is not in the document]
djc: Makes sense!

Ralith:
Something strange is going on with coverage. Not sure if it's related to this PR or not.

> thread panicked while panicking. aborting.
> [INFO tarpaulin] Failure Failed to get test coverage! Error: Failed to run tests: Error running test - SIGILL raised in 12965

Ralith:
I've successfully split up the commit a decent amount, hopefully easier to review now. In particular, the logic changes to send buffering have been separated from the refactoring to move logic out of connection.rs as much as is practical. I think there's room to move even more stuff into streams, and to break up streams itself, but what's here seems to work and is a good start.

Ralith:
I'd like to hear how this impacts the benchmarks (in particular interop and bulk) for people, as well. I'm having difficulty getting intelligible results out of my system.

stammw:
I'm going to fix a few things up and bench this PR. I could also run the @rmarx's tool that graphs streams distribution on the wire / request completion ordrer.

stammw:
Testing against h3.stammw.eu (that runs this PR branch right now), I get timeouts on big download at `3145691 bytes`. That value was found bisecting the timeout size with:

```sh
cargo run --release --example=h3_client https://h3.stammw.eu/3145690
cargo run --release --example=h3_client https://h3.stammw.eu/3145691
```

Under that value, I get occasional stream resets:

```sh
cargo run --release --example=h3_client https://h3.stammw.eu/3145670
    Finished release [optimized + debuginfo] target(s) in 0.09s
     Running `target/release/examples/h3_client 'https://h3.stammw.eu/3145670'`
Apr 15 13:31:07.301  INFO h3_client: received response: Response { status: 200, version: HTTP/3.0, headers: {"server": "quinn-h3:0.0.1"}, body: () }
Apr 15 13:31:07.431 ERROR quinn_h3::connection: driver error: QUIC protocol error: received a STREAM frame or a RESET_STREAM frame containing a different final size to the one already established
Error: read error: Io(Custom { kind: Other, error: TransportError(Error { code: FINAL_SIZE_ERROR, frame: None, reason: "" }) })
```

Ralith:
> EDIT: Didn't fetch the last changes, now bigger transfers seem slower

Can you compare the absolute time taking for h3 before/after this PR, and/or try the bulk benchmark? Testing locally (h3.stammw.eu was refusing TCP connections), it looks like this PR is similar or slightly faster (before: 60-80-ms, after: 60-70ms), while h2 timing is *wildly* erratic but usually faster regardless (6-50ms).

quinn-proto/src/streams.rs
```diff
@@ -221,36 +228,148 @@ impl Streams {
             assert!(self.recv.insert(id, Recv::new()).is_none());
         }
     }
+
```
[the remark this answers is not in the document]
Ralith: Running this through the rebases turned out to be a huge pain, so I left it in a separate commit. Let me know if you'd strongly prefer this to be baked in.

stammw:
I've been playing this morning and got some stats on 100 samples. Here are average percentage compared to h2 (less is better):

* master: 
** 5M: -19.3598 % 
** 10M: -1.662 %
* round-robin:
** 5M: -20.6651 %
** 10M: -3.0203 %

I got the methodology and samples in [this gist](https://gist.github.com/stammw/1df201db8a69dac99a95e7fda5ae222e).

djc:
When benchmarking this branch at the quinn level, I get a panic (it works on master):

```
thread 'main' panicked at 'missing pending stream', quinn-proto/src/streams.rs:278:26
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "PoisonError { inner: .. }"', /Users/djc/src/quinn/quinn/src/connection.rs:595:26
  15:        0x105d0a8c0 - <quinn::connection::ConnectionRef<S> as core::ops::drop::Drop>::drop::h5f827b87d51ba1ab
                               at /Users/djc/src/quinn/quinn/src/connection.rs:595
```

djc:
quinn-level benchmarks:

```
djc-2019 round-robin quinn $ cargo bench              
test large_streams ... bench:   1,046,551 ns/iter (+/- 193,528) = 125 MB/s
test small_streams ... bench:       1,312 ns/iter (+/- 447)
djc-2019 round-robin quinn $ git checkout master
djc-2019 master quinn $ cargo bench        
test large_streams ... bench:   1,066,135 ns/iter (+/- 330,168) = 122 MB/s
test small_streams ... bench:         614 ns/iter (+/- 435) = 1 MB/s
djc-2019 master quinn $ cargo bench
test large_streams ... bench:   1,054,270 ns/iter (+/- 426,794) = 124 MB/s
test small_streams ... bench:       1,441 ns/iter (+/- 163)
djc-2019 master quinn $ cargo bench
test large_streams ... bench:   1,077,351 ns/iter (+/- 301,750) = 121 MB/s
test small_streams ... bench:       1,520 ns/iter (+/- 309)
djc-2019 master quinn $ git checkout round-robin
djc-2019 round-robin quinn $ cargo bench             
test large_streams ... bench:   1,065,065 ns/iter (+/- 267,801) = 123 MB/s
test small_streams ... bench:       1,287 ns/iter (+/- 204)
djc-2019 round-robin quinn $ cargo bench
test large_streams ... bench:   1,067,245 ns/iter (+/- 227,193) = 122 MB/s
test small_streams ... bench:       1,171 ns/iter (+/- 143)
djc-2019 round-robin quinn $ cargo bench
test large_streams ... bench:   1,042,049 ns/iter (+/- 170,296) = 125 MB/s
test small_streams ... bench:       1,198 ns/iter (+/- 249)
```

So for large streams on round-robin I got [122, 123, 125, 125] (avg 124) and for master [121, 122, 124] (avg 122), suggesting this branch is maybe faster (by 1.6%) but probably within the noise.

quinn-proto/src/connection.rs
```diff
@@ -1392,6 +1374,7 @@ where
                             self.in_flight.remove(&info);
                             self.space_mut(SpaceId::Data).pending += info.retransmits;
                         }
+                        self.streams.retransmit_all_for_0rtt();
```
djc, line 1349: Is this new? I can't find the equivalent code in the old version.

quinn-proto/src/connection.rs
```diff
@@ -1687,12 +1675,9 @@ where
...
                 }
                 Frame::MaxData(bytes) => {
-                    let was_blocked = self.flow_blocked();
-                    self.max_data = cmp::max(bytes, self.max_data);
-                    if was_blocked && !self.flow_blocked() {
-                        for stream in self.blocked_streams.drain() {
-                            self.events.push_back(Event::StreamWritable { stream });
-                        }
+                    self.streams.max_data(bytes);
+                    for stream in self.streams.drain_unblocked() {
```
djc, line 1679: This loop seems to appear in a few places. Can we do it in one place (maybe inside `poll()`?). (Could be a separate commit if you prefer.)

djc:
I've spent quite a bit of time on reviewing this. I honestly don't feel very comfortable with 39e4ca388ccf1793a7773da649e42f6682bafe08 ("Buffer outgoing data") and, to a lesser extent, with 1b929290e923df874031d81af50741948f187eb2 ("Move connection-level flow control").

Follow-up questions: how do you feel about handling other stream-related frames like `MAX_STREAM_DATA`, `STOP_SENDING`, `RESET_STREAM`? I tend to think we should move their logic into `Streams` as much as possible, do you agree? (I could potentially take that on.)

Ralith:
> if you feel okay with them I don't want to require you to split them up more

Those are definitely the riskiest changes here. I'm open to splitting them up further, but it's difficult to come up with axes across which they're divisible. Having said that, though, maybe I can try taking the refactoring that the new buffering logic requires (i.e. `Streams::poll_transmit`, `ack`, `retransmit`, etc) and applying it *before* bringing in the new buffering logic...

quinn-proto/src/streams.rs
```diff
@@ -353,6 +389,41 @@ impl Streams {
...
+    }
+
+    pub fn drain_unblocked(&mut self) -> impl Iterator<Item = StreamId> + '_ {
+        let range = if self.flow_blocked() {
+            debug_assert_eq!(
+                self.data_sent, self.max_data,
+                "violated peer's connection-level flow control limits"
+            );
+            // Empty range because nothing was unblocked
+            0..0
```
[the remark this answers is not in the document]
Ralith: The return type must be a single type. The only way around that would be a boxed trait object, which is too big a hammer here.

quinn-proto/src/streams.rs
```diff
@@ -353,6 +389,41 @@ impl Streams {
...
+        } else {
+            // Discard finished/closed streams and unset connection_blocked flags
+            let send = &mut self.send;
+            self.connection_blocked.retain(|id| match send.get_mut(id) {
+                None => false,
+                Some(stream) => {
+                    stream.connection_blocked = false;
+                    stream.state.is_writable()
```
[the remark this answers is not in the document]
Ralith: I like this approach because it avoids having to allocate new storage for `connection_blocked`. Added more comments to hopefully clarify things.

quinn-proto/src/streams.rs
```diff
@@ -51,6 +64,11 @@ impl Streams {
             next_reported_remote: [0, 0],
             send_streams: 0,
             pending: Vec::new(),
+            connection_blocked: Vec::new(),
```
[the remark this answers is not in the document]
Ralith: `HashSet` iteration can be expensive as it's `O(capacity)` rather than `O(elements)`. This keeps things more predictable.

quinn-proto/src/connection.rs
Ralith: This used to be handled by the operation just above; `info.retransmits` contained sent stream frames from the packet in question, and `pending` contained the list of stream frames to send. With those removed in favor of the per-stream send buffers, we have to be more explicit.

quinn-proto/src/streams.rs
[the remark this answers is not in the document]
djc: Fair enough!

quinn-proto/src/connection.rs
```diff
@@ -866,14 +855,13 @@ where
...
         let status = self.streams.reset(stream_id, stop_reason);
-        let was_conn_blocked = self.blocked_streams.remove(&stream_id);
         if stopped {
             if status == Some(streams::ResetStatus::WasFinishing) {
                 self.events.push_back(Event::StreamFinished {
                     stream: stream_id,
                     stop_reason,
                 });
-            } else if was_conn_blocked || status == Some(streams::ResetStatus::WasBlocked) {
+            } else if status == Some(streams::ResetStatus::WasBlocked) {
```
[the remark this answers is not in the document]
Ralith: Good call; this is a lot more obvious now!

quinn-proto/src/streams.rs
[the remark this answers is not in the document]
djc: Fair enough.

djc:
> Those are definitely the riskiest changes here. I'm open to splitting them up further, but it's difficult to come up with axes across which they're divisible. Having said that, though, maybe I can try taking the refactoring that the new buffering logic requires (i.e. Streams::poll_transmit, ack, retransmit, etc) and applying it before bringing in the new buffering logic...

Yeah, that would probably help. Also you could try moving the state separately from moving any code, keeping fields `pub`, and then move the code separately.

quinn-proto/src/connection.rs
Ralith: Great catch! That makes this more readable *and* more efficient.

Ralith:
I wasn't able to reproduce the panic you mentioned, @djc, but I found a plausible cause: if a stream is reset and the reset is acknowledged while the stream still has data or finish pending, the stream's state can be discarded while it's still in the pending list. I applied the simple fix of discarding such entries in the list when they're encountered.

quinn-proto/src/streams.rs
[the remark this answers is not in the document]
Ralith: Mooted by the refactor into pull-based notification inspired by a separate comment.

djc:
(The panic was from the old version of this branch.)

Ralith:
Alright, I've split things up significantly more, and found and corrected a few regressions in the process, so that was time well spent.

quinn-proto/src/streams.rs
```diff
@@ -255,9 +256,15 @@ impl Streams {
...
     /// Does not cause the actual RESET_STREAM frame to be sent, just updates internal
     /// state.
-    pub fn reset(&mut self, id: StreamId, stop_reason: Option<VarInt>) -> Option<ResetStatus> {
-        let stream = self.send.get_mut(&id)?;
-        stream.reset(stop_reason)
+    pub fn reset(
+        &mut self,
+        id: StreamId,
+        stop_reason: Option<VarInt>,
+    ) -> (u64, Option<ResetStatus>) {
```
djc: Looks like this is flipflopping on the return type within this PR. Is that necessary/useful?

quinn-proto/src/connection.rs
```diff
@@ -2595,37 +2590,49 @@ where
...
+                // Stream was reset with pending data
+                None => continue,
             };
-            frame.encode(true, buf);
-            if !stream.data.is_empty() {
-                stream.offset += len as u64;
-                self.streams.pending.push_front(stream);
-            }
-            stream_frames.push(frame);
+            match stream.state {
```
djc: This seems like a big transformation that I can't fully follow. Where does all this code come from?

quinn-proto/src/streams.rs
```diff
@@ -423,6 +442,16 @@ impl Send {
         }
     }
 
+    /// Returns whether the stream was unblocked
+    pub fn raise_max_data(&mut self, offset: u64) -> bool {
```
djc, line 546: Naming nit: not sold on the use of `raise` here. Can we do `increase` or `bump` or `extend` or something else? It might be my past with Python that's confusing me. I also [looked it up](https://www.merriam-webster.com/dictionary/raise) and found that most highly ranked meanings don't seem to match this usage well.

quinn-proto/src/streams.rs
Ralith: The modified return type is needed to update the send window until send window bookkeeping is moved into `Streams` in turn. I could reorder the changes so that reset isn't moved in until the send window is, but that'd be a moderately painful/error-prone rebase, so my inclination is to leave things as-is unless you feel strongly.

quinn-proto/src/streams.rs
Ralith: Good catch; in particular in software "raise" often refers to error conditions!

quinn-proto/src/streams.rs
djc: Fair enough.

quinn-proto/src/streams.rs
djc: Yeah, that's what I was referring to with my Python experience: in Python, `raise` is used to throw exceptions.

quinn-proto/src/connection.rs
Ralith: Rearranged the send window refactoring and simplified the state checks to make this particular change a bit smaller. Should hopefully be more obvious now; the remaining changes are direct consequences of the move from using prebaked stream frames to composing them based on buffer and finish state on the fly.

djc: Yeah, I think it's better now. Let's merge this thing already!

PR #701, opened 2020-04-09
Accept PEM format in interop server

stammw:
As well as DER.


djc:
LGTM!

stammw:
I could not reproduce the failure with tarpaulin on my machine...

stammw:
I had to make tarpaulin run with `stable` to make it pass.

PR #709, opened 2020-04-11
Interop throughput

stammw:
Introducing a `HTTP/2` endpoint in interop tests to compare throughput with current implementation pair. Feat. `hyper`.

### DIFF
--- a/interop/Cargo.toml
+++ b/interop/Cargo.toml
@@ -11,6 +11,8 @@ anyhow = "1.0.22"
 bytes = "0.5.2"
 futures = "0.3.1"
 http = "0.2"
+hyper = "0.13"
+hyper-rustls = "0.20"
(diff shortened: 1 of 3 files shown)


djc:
Good stuff!

stammw:
The kind of low throughput is gone with last commit's tweaks ! I even get between -30% and -50% compared to h2 :)

Ralith:
Awesome stuff! Should we be adjusting our defaults?

djc:
At least for H3 I think it makes sense to adjust the defaults. Not as sure about base QUIC.

Ralith:
I think the defaults that empirically perform well for H3 in practice are a much better place to start users off than the numbers that I invented on the spot, although we should try to still remain close to the recommendations of the recovery draft where applicable.

PR #710, opened 2020-04-13
Don't block application writes on congestion

Ralith:
With this change, only flow control and the configured send window limit the amount of data we accept from the application. I think this is more consistent with familiar behavior from TCP, and should reduce the amount of stack thrashing when the congestion window is small. CC @stammw regarding performance tests.

### DIFF
--- a/quinn-proto/src/connection.rs
+++ b/quinn-proto/src/connection.rs
@@ -73,7 +73,7 @@ where
     key_phase: bool,
     /// Transport parameters set by the peer
     params: TransportParameters,
-    /// Streams on which writing was blocked on *connection-level* flow or congestion control
+    /// Streams on which writing was blocked on *connection-level* flow control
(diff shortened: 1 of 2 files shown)


stammw:
Nice! I vaguely remember noticing some serious slowdown once some volume is transmitted. But it was some months ago and quinn-h3 was very wacky.

Throughput test still passes with big windows. Removing the settings, It passes in local but not on the remote server. Tell me if you want me to get more comparisons.

Ralith:
So no significant observed effect, then? Good enough for me. On my end, the bulk benchmark seems to go 20-30mbps faster, which is pretty neat, probably due to reduced task switching.

@djc, I had to fix a comment; reapprove?

djc:
😎

stammw:
Awesome! I got a -65% in local compared to h2.

Issue #702, opened 2020-04-09
Implement the 'd' QPACK dynamic table H3 test


stammw:
> One or both endpoints insert entries into dynamic table and subsequenly reference them from header blocks

This will likely need some `spy` features that the interop client would use to check the state of dynamic table. What do you think of adding a feature flag to achieve this ?

djc:
Yeah, I think a feature flag would make sense for this.

Ralith:
We'll need similar stuff in the lower-level crates too, e.g. for #706 and #707. I'd suggest naming the feature something a little more obvious like `interop_test_accessors`, though.

djc:
Done.

PR #711, opened 2020-04-14
Interop d test

stammw:
Adresses #702 

Introduces a feature flag to enable accessors in the private connection state.

### DIFF
--- a/interop/src/main.rs
+++ b/interop/src/main.rs
@@ -13,7 +13,7 @@ use hyper::body::HttpBody as _;
 use lazy_static::lazy_static;
 use quinn_h3::Settings;
 use structopt::StructOpt;
-use tracing::{error, info, info_span, warn};
+use tracing::{debug, error, info, info_span, warn};
 use tracing_futures::Instrument as _;
 
 #[derive(StructOpt, Debug)]
@@ -248,6 +248,7 @@ impl State {
                     .instrument(info_span!("throughput"))
                     .await,
                 None,
+                None,
             ))
         } else {
             let (core, key_update, rebind, retry, throughput) = tokio::join!(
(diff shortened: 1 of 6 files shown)


interop/src/main.rs
```diff
@@ -722,6 +731,28 @@ impl State {
         conn.close();
         Ok(())
     }
+
+    async fn dynamic_encoding(&self) -> Result<()> {
+        dbg!();
```
djc, line 736: This should probably not be merged?

interop/src/server.rs
```diff
@@ -445,3 +448,5 @@ const HOME: &str = r##"
   </body>
 </html>
 "##;
+
+const VERSION: &str = "quinn_h3:0.0.1";
```
djc, line 452: I think it should be spelled as "quinn-h3" in non-code text. What do you think?

interop/src/main.rs
stammw: Probably not, sorry.

interop/src/server.rs
stammw: Yeah kebab case is better here, didn't pay much attention.

quinn-h3/Cargo.toml
```diff
@@ -17,6 +17,9 @@ edition = "2018"
 [package.metadata.docs.rs]
 all-features = true
 
+[features]
+interop_test_accessors = []
```
djc, line 21: Sorry, one more, since we discussed it in the context of the user agent: I think we also use kebab case for feature names in most places (not sure we are consistent about it here). Can we do that here?

stammw: Sure! It is the convention, I just figured out.

PR #712, opened 2020-04-14
Interop server tweaks

Ralith:


interop/src/server.rs
```diff
@@ -46,8 +46,8 @@ struct Opt {
     )]
     cert: PathBuf,
     /// Address to listen on
-    #[structopt(long = "listen", short = "l", default_value = "0.0.0.0:4433")]
-    listen: SocketAddr,
+    #[structopt(long = "listen", short = "l", default_value = "::")]
```
djc, line 49: This flips the default from IPv4 to IPv6. Not sure if that's desirable?

stammw: I was also confused about this. But it does not what it seems to do: "::" also listens on "0.0.0.0" from what I tested.

Ralith: On many (but not all) OSs, IPv6 sockets default to dual-stack, so given that we have to pick one I think it's the reasonable default. At least with this change you can select either!

PR #713, opened 2020-04-14
Replace bitlab dependency with homegrown bit twiddling

djc:
bitlab was a bit of an odd dependency, providing some convenience wrappers for writing bits across byte slices. It additionally pulled in the num crate. Replace it by writing our own version of the bit reading/writing routines, which should amount to much less code and is more optimized for our use case since a number of checks can be elided.

I'm not a very experienced bit twiddler and it took me way too long to get the encode side working. Happily we have pretty good regression tests for this stuff! Any feedback is welcome.

### DIFF
--- a/quinn-h3/src/qpack/prefix_string/decode.rs
+++ b/quinn-h3/src/qpack/prefix_string/decode.rs
@@ -1,6 +1,3 @@
-#![allow(clippy::ptr_arg)] // bitlab doesn't handle `&[u8]` for some reason
-use bitlab::*;
-
 use super::BitWindow;
 
 #[derive(Debug, PartialEq)]
@@ -22,7 +19,7 @@ struct HuffmanDecoder {
 }
 
 impl HuffmanDecoder {
-    fn check_eof(&self, bit_pos: &mut BitWindow, input: &Vec<u8>) -> Result<Option<u32>, Error> {
+    fn check_eof(&self, bit_pos: &mut BitWindow, input: &[u8]) -> Result<Option<u32>, Error> {
         use std::cmp::Ordering;
         match ((bit_pos.byte + 1) as usize).cmp(&input.len()) {
             // Position is out-of-range
@@ -33,9 +30,9 @@ impl HuffmanDecoder {
             Ordering::Equal => {
                 let side = bit_pos.opposite_bit_window();
 
-                let rest = match input.get_u8(side.byte, side.bit, side.count) {
+                let rest = match read_bits(input, side.byte, side.bit, side.count) {
                     Ok(x) => x,
-                    Err(_) => {
+                    Err(()) => {
                         return Err(Error::MissingBits(side));
                     }
                 };
@@ -50,14 +47,14 @@ impl HuffmanDecoder {
         Err(Error::MissingBits(bit_pos.clone()))
     }
 
-    fn fetch_value(&self, bit_pos: &mut BitWindow, input: &Vec<u8>) -> Result<Option<u32>, Error> {
-        match input.get_u32(bit_pos.byte, bit_pos.bit, bit_pos.count) {
-            Ok(value) => Ok(Some(value)),
-            Err(_) => self.check_eof(bit_pos, &input),
+    fn fetch_value(&self, bit_pos: &mut BitWindow, input: &[u8]) -> Result<Option<u32>, Error> {
+        match read_bits(input, bit_pos.byte, bit_pos.bit, bit_pos.count) {
+            Ok(value) => Ok(Some(value as u32)),
+            Err(()) => self.check_eof(bit_pos, &input),
         }
     }
 
-    fn decode_next(&self, bit_pos: &mut BitWindow, input: &Vec<u8>) -> Result<Option<u8>, Error> {
+    fn decode_next(&self, bit_pos: &mut BitWindow, input: &[u8]) -> Result<Option<u8>, Error> {
         bit_pos.forwards(self.lookup);
 
         let value = match self.fetch_value(bit_pos, input) {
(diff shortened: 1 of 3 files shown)


quinn-h3/src/qpack/prefix_string/decode.rs
```diff
@@ -78,6 +75,23 @@ impl HuffmanDecoder {
...
 
+fn extract_bits(src: &[u8], mut byte_offset: u32, mut bit_offset: u32, len: u32) -> Result<u8, ()> {
+    if len == 0 || len > 8 || src.len() as u32 * 8 < byte_offset * 8 + bit_offset + len {
+        return Err(());
+    }
+
+    byte_offset += bit_offset / 8;
+    bit_offset -= (bit_offset / 8) * 8;
+    Ok(if bit_offset + len <= 8 {
+        (u8::from_be(src[byte_offset as usize]) << bit_offset) >> (8 - len)
```
Ralith, line 86: `u8::from_be` is the identity function; endianness only exists between bytes

quinn-h3/src/qpack/prefix_string/decode.rs
```diff
@@ -78,6 +75,23 @@ impl HuffmanDecoder {
     }
 }
 
+fn extract_bits(src: &[u8], mut byte_offset: u32, mut bit_offset: u32, len: u32) -> Result<u8, ()> {
```
Ralith, line 78: As someone who's also not a bit twiddling expert, I think this needs a decent battery of unit tests regardless of my opinion of the implementation.

djc: Both `extract_bits()` and `set_bits()` are [fully covered](https://codecov.io/gh/djc/quinn/commit/4f5f85daeac3d8f12825e9ea31ea5a612897584f) by existing tests. As such, I don't see a whole lot of value in adding more. What do you think?

quinn-h3/src/qpack/prefix_string/decode.rs
djc: Yeah, forgot to cut that out.

quinn-h3/src/qpack/prefix_string/decode.rs
Ralith: That link doesn't work here (and I can't seem to find equivalent information by browsing manually) but I think it's somewhat beside the point. It's useful to have narrowly scoped unit tests even if they're subsumed by larger ones, because it's easier to be sure that the narrowly scoped tests cover all interesting cases (which can be more subtle than just ensuring that every branch is traversed), and if something *is* wrong, the narrow tests are much more helpful for diagnosing an issue, both now and in the future (I'm reluctant to bet that we'll never want to touch a given piece of code again).

djc: Done.

stammw:
Got this tested with QPACK interop test. Everything is fine :+1:

quinn-h3/src/qpack/prefix_string/encode.rs
```diff
@@ -93,6 +75,32 @@ impl HuffmanEncoder {
...
+    debug_assert!(pos.bit < 8);
+    debug_assert!(pos.count <= 8);
+    debug_assert!(pos.count > 0);
+
+    if (pos.bit + pos.count) <= 8 {
+        debug_assert_eq!(out[pos.byte as usize] | PAD_LEFT[pos.bit as usize], 255);
+        let pad_left = out[pos.byte as usize] | PAD_RIGHT[(8 - pos.bit) as usize];
+        let shifted = value << (8 - pos.bit - pos.count) | PAD_LEFT[pos.bit as usize];
+        let pad_right = PAD_RIGHT[(8 - pos.count - pos.bit) as usize];
+        out[pos.byte as usize] = (pad_left & shifted) | pad_right;
```
Ralith, line 88: Is it intended that if bit = 0 and count = 8, `value` is completely ignored? I'm probably missing something here.

djc: Pretty sure it isn't? I think the first part of the test covers this case.

djc: As in this case:

```
pos.count = 8;
set_bits(&mut buf, &pos, 0b1_0101);
assert_eq!(&buf[..1], &[0b1_0101]);
pos.byte += 1;
```

Ralith: Er, sorry, I meant when bit = 0 and count = 0.

Ralith: For example:

```rust
let mut buf = [0; 1];
set_bits(
    &mut buf,
    &BitWindow {
        byte: 0,
        bit: 0,
        count: 1,
    },
    0,
);
```

produces `buf == [255]`, which doesn't seem right, unless I'm wildly misunderstanding the intended semantics (which should probably be commented?)

Ralith:
I'm prepared to believe that these functions are working as intended, but exactly what the intended semantics are is mysterious to me, particularly for `set_bits`. Could we have some comments describing what assumptions they make and why?

djc:
Yeah, I was thinking the same thing. I've renamed these functions to `read_bits()` and `write_bits()` and added a whole bunch of comments. Let me know if that makes things more clear!

quinn-h3/src/qpack/prefix_string/decode.rs
```diff
@@ -78,6 +75,29 @@ impl HuffmanDecoder {
     }
 }
 
+// Read `len` bits from the `src` slice at the specified position
+//
+// Never read more than 8 bits at a time. `bit_offset` may be larger than 8.
```
Ralith, line 80: nit: these might as well be doc comments

quinn-h3/src/qpack/prefix_string/encode.rs
```diff
@@ -93,6 +75,43 @@ impl HuffmanEncoder {
...
 
+// Write bits from `value` to the `out` slice
+//
+// Write the least significant `pos.count` bits from `value` to the position specified by
+// `(pos.byte, pos.bit)`. Writes may span multiple bytes. `out` is expected to be long enough
+// to write these bits; this is ensured by `HuffmanEncoder::ensure_free_space()`, which is
+// always called prior to calling this function.
+//
+// The bits to be written to are expected to be set to 1 when calling this function. Similarly,
+// this function maintains the invariant that unused bits in the output bytes are set to 1.
```
Ralith, line 86: Using 1 for unwritten data rather than 0 seems pretty unconventional. Is there a reason?

Ralith, line 86: nit: doc comment

djc: I think this is described [here](https://tools.ietf.org/html/rfc7541#appendix-B). Added a reference to this in the doc comment for `write_bits()`.

djc: Fixed. (I thought clippy didn't like doc comments for private functions, but that doesn't appear to be an issue.)

quinn-h3/src/qpack/prefix_string/decode.rs
djc: Fixed.

quinn-h3/src/qpack/prefix_string/encode.rs
Ralith: In fact, rustdoc fully supports documenting such things via `--document-private-items`. Handy for exploring a codebase.

PR #717, opened 2020-04-16
H3 fix bench

stammw:
The new bench tool call top-level bench functions multiple times (`throughput_1k` for example). Kill the server at the end of each call fixes the failures this introduced.

### DIFF
--- a/quinn-h3/benches/bench.rs
+++ b/quinn-h3/benches/bench.rs
@@ -1,14 +1,12 @@
 use std::{
-    net::{IpAddr, Ipv6Addr, SocketAddr},
-    ops::RangeFrom,
-    sync::{Arc, Mutex},
+    net::{IpAddr, Ipv6Addr, SocketAddr, UdpSocket},
+    sync::Arc,
     thread,
 };
 
 use bencher::{benchmark_group, benchmark_main, Bencher};
 use futures::{channel::oneshot, StreamExt};
 use http::{Request, Response, StatusCode};
-use lazy_static::lazy_static;
 use tokio::{
     io::AsyncWriteExt as _,
     runtime::{Builder, Runtime},
@@ -18,7 +16,11 @@ use tracing::{error_span, span, Level};
 use tracing_futures::Instrument as _;
 
 use quinn::{ClientConfigBuilder, ServerConfigBuilder};
-use quinn_h3::{self, client, server, BodyWriter};
+use quinn_h3::{
+    self, client,
+    server::{self, IncomingConnection},
+    BodyWriter,
+};
 
 benchmark_group!(
     benches,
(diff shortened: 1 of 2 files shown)


quinn-h3/benches/bench.rs
```diff
@@ -109,29 +118,38 @@ impl Context {
         Self {
             server_config: server::Builder::with_quic_config(server_config),
             client_config: client::Builder::with_quic_config(client_config),
+            stop_server: None,
         }
     }
 
-    pub fn spawn_server(&self) -> (SocketAddr, thread::JoinHandle<()>) {
+    pub fn spawn_server(&mut self) -> (SocketAddr, thread::JoinHandle<()>) {
         // TODO: Let the OS choose a free port for us
```
stammw, line 126: @Ralith I'm not sure how this is possible. In the `quinn` bench, you use the port `0`, but how does the client know to which port it should connect ?

Ralith: The trick is to bind a socket with port 0, then ask the OS what port it chose, then pass that to the client. I think this was covered in gitter earlier this morning.

stammw:
Got this tested with QPACK interop test. Everything is fine :+1:

quinn-h3/benches/bench.rs
```diff
@@ -106,77 +119,41 @@ impl Context {
...
+        if let Some(send) = self.stop_server.take() {
+            send.send(()).expect("stop server");
+        }
+    }
+
     pub fn make_client(&self, server_addr: SocketAddr) -> (client::Connection, Runtime) {
         let mut runtime = rt();
         let my_span = span!(Level::TRACE, "client");
         let _enter = my_span.enter();
+        dbg!(server_addr);
```
djc, line 156: Probably shouldn't leave this in?

quinn-h3/benches/bench.rs
```diff
@@ -106,77 +119,41 @@ impl Context {
...
-                                .expect("body write");
-                            remaining -= size;
-                        }
-                    }
-                }
-                .instrument(error_span!("server")),
-            );
+            let handle = runtime.spawn(async {
+                let incoming_conn = server.with_socket(socket).unwrap();
+                handle_connection(incoming_conn, stop_recv).instrument(error_span!("server"))
```
Ralith, line 138: Missing `.await`. This is why the benchmark is timing out instead of running.

quinn-h3/benches/bench.rs
```diff
@@ -106,77 +119,41 @@ impl Context {
...
-                                .write_all(&data.as_bytes()[..size])
-                                .await
-                                .expect("body write");
-                            remaining -= size;
-                        }
-                    }
-                }
-                .instrument(error_span!("server")),
-            );
+            let handle = runtime.spawn(async {
```
Ralith, line 136: Rather than spawning and then blocking on the spawn handle, it would be simpler to just `runtime.block_on(...)` the future directly.

quinn-h3/benches/bench.rs
stammw: Thanks! Knew it was obvious. I must have missed a warning then :/

quinn-h3/benches/bench.rs
stammw: Sure, thanks

quinn-h3/benches/bench.rs
stammw: Yeah, plus it's not that recommended for a benchmark to print in terminal :)

stammw:
Let's not merge this right away, I'm writing more benches while I am at it.

(If you are ok with adding commits here)

djc:
Not sure why you want to block merging this on adding more? Commits and PRs are cheap. 😊

djc:
Noted your commits are still spawning then blocking on the resulting handle. I think @Ralith  meant you could pass an async expression or block to `block_on()` directly.

stammw:
Ok why not using another PR, I'll just make some changes here that I figured out preparing other benchmarks.

stammw:
Should be Ok now. Thanks.

quinn-h3/benches/bench.rs
```diff
@@ -119,54 +119,8 @@ impl Context {
...
-                                .write_all(&data.as_bytes()[..size])
-                                .await
-                                .expect("body write");
-                            remaining -= size;
-                        }
-                    }
-                }
-                .instrument(error_span!("server")),
-            );
+            let handle = runtime.spawn(handle_connection(server).instrument(error_span!("server")));
```
djc: This is what we meant:

```suggestion
            let handle = runtime.block_on(handle_connection(server).instrument(error_span!("server"))).unwrap();
```

stammw: Yeah, sorry but I got that moved to the first commit.

stammw: @Ralith Is it ok for you ?

djc:
LGTM.

quinn-h3/benches/bench.rs
Ralith: Yep, all sorted. Thanks!

PR #718, opened 2020-04-17
Fix the QPACK offline interop tool qif.

stammw:
The [bin] entry has been commented so it was left behind and couldn't run.

### DIFF
--- a/interop/Cargo.toml
+++ b/interop/Cargo.toml
@@ -33,3 +33,7 @@ path = "src/main.rs"
 [[bin]]
 name = "server"
 path = "src/server.rs"
+
+[[bin]]
+name = "qif"
+path = "src/qif.rs"
--- a/interop/src/qif.rs
+++ b/interop/src/qif.rs
@@ -135,6 +135,8 @@ fn main() -> Result<(), Error> {
     }
 }
 
+struct Block<'a>(std::io::Cursor<&'a [u8]>, u64);
+
 struct BlockIterator<R: Buf> {
     buf: R,
     last_block_len: usize,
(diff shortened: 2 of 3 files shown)


Ralith:
Yeah, I think this would be significantly simplified by just embracing anyhow's dynamically dispatched error type. There's not much value to an error enum that will never, ever be matched on.

stammw:
I left that because I was pretty sure `Error` was also used for logic. I doesn't seem to be the case. 

Here, using an enum as still one goal: reduce call site error message duplication.

I'll rewrite all this.

stammw:
Got rid of `err_derive` and `enum Error`. It's actually much better now :)

I also added a small doc to get started with the tool.

interop/Cargo.toml
```diff
@@ -9,6 +9,7 @@ publish = false
 [dependencies]
 anyhow = "1.0.22"
 bytes = "0.5.2"
+err-derive = "0.2.3"
```
djc, line 12: Not needed anymore?

stammw: oops...

interop/src/qif.rs
```diff
@@ -365,6 +400,7 @@ impl EncodedFile {
             stream_count += 1;
         }
         fs::write(&self.file, buf)?;
+        println!("encoded in {:?}", self.file);
```
djc, line 403: This doesn't seem to belong?

stammw: It actually is. When you run it, you'll want to know what happened. I figured this out writing the doc.

djc: Fair enough!

PR #734, opened 2020-04-25
H3 http body

stammw:


djc:
In trying to integrate this with hyper, I think we need to make `RecvData` and `RecvBody` public, as well?

stammw:
Yes sure! I'll push that soon.

stammw:
In fact `RecvData` is wrapped into `RecvRequest` and `RecvResponse`. Maybe `SendData` should also be completely wrapped (into `SendResponse`). For symmetry with `SendRequest`, but we'll also need this for some control specific to response.

I've got all tests passing into my working copy. I'm currently separating this up into commits. I also have a serious fall in throughput benchmarks I need to solve. I believe it's only the benchmark impl.

Next I'll get the docs a little refresh.

djc:
Let's do the docs update in a separate PR, so that I can rebase my pin-project and other tweaks in parallel with your documentation work?

stammw:
Did you bring more changes? Just 2 failing tests to fix and we're good.

djc:
No, I don't have more changes. Do want to split up my commits a bit more to be a bit cleaner. Thanks for rebasing them into your stack, though!

stammw:
So here we are! It's ready for a review. If you want to edit those commits of yours, feel free ! :)

djc:
@stammw please have another look. I've spent some time improving things:

* Keep `Body` around, making minimal changes to allow it to `impl HttpBody`
* Merged commits so that most commits are properly formatted, pass tests and are clippy clean
* Removed all `Unpin` bounds (the client-side bounds were still there)
* Consistently import `http_body::Body` as `HttpBody`

I'm still not super-happy with the commit history (the client response cancellation still fails for a bunch of these commits until fixed up in the last one, some commits are very large and not very focused) but I'm also not sure how it could be improved much. If you're happy with my updates, let's merge it.

stammw:
Pretty neat ! Thanks for the improvements !

PR #737, opened 2020-04-30
Support the siduck datagram test protocol in interop tools

Ralith:
See https://tools.ietf.org/html/draft-pardue-quic-siduck-00 for details.

### DIFF
--- a/interop/src/main.rs
+++ b/interop/src/main.rs
@@ -7,8 +7,8 @@ use std::{
     time::{Duration, Instant},
 };
 
-use anyhow::{anyhow, Context, Result};
-use futures::future;
+use anyhow::{anyhow, bail, Context, Result};
+use futures::{future, StreamExt};
 use hyper::body::HttpBody as _;
 use lazy_static::lazy_static;
 use quinn_h3::Settings;
@@ -32,9 +32,23 @@ struct Opt {
     hq: bool,
     #[structopt(long)]
     seq: bool,
+    #[structopt(long)]
+    siduck: bool,
     /// Enable key logging
     #[structopt(long = "keylog")]
     keylog: bool,
+}
+
+impl Opt {
+    fn alpn(&self) -> Option<Alpn> {
+        match (self.h3, self.hq, self.siduck) {
+            (false, false, false) => None,
+            (true, false, false) => Some(Alpn::H3),
+            (false, true, false) => Some(Alpn::Hq),
+            (false, false, true) => Some(Alpn::SiDuck),
+            _ => panic!("conflicting protocol options"),
+        }
+    }
 }
 
 #[derive(Clone)]
@@ -42,6 +56,7 @@ enum Alpn {
     Hq,
     H3,
     HqH3,
+    SiDuck,
 }
 
 impl From<&Alpn> for Vec<Vec<u8>> {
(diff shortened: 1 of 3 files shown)


interop/src/server.rs
```diff
@@ -450,3 +451,32 @@ const HOME: &str = r##"
...
+                return Ok(());
+            }
+            Err(e) => {
+                return Err(e.into());
+            }
+            Ok(data) => {
+                if &data[..] == b"quack" {
+                    connection.send_datagram(b"quack-ack"[..].into())?;
+                } else {
+                    connection.close(0x101u32.into(), b"quack quack quack");
```
djc, line 475: Maybe name the magic integer value?

interop/src/main.rs
```diff
@@ -165,9 +169,10 @@ async fn main() {
             port: opt.port.unwrap_or(4433),
             retry_port: opt.retry_port.unwrap_or(4434),
             sequential: opt.seq,
-            alpn: match (opt.h3, opt.hq) {
-                (false, true) => Alpn::Hq,
-                (true, false) => Alpn::H3,
+            alpn: match (opt.h3, opt.hq, opt.siduck) {
+                (false, true, _) => Alpn::Hq,
+                (true, false, _) => Alpn::H3,
+                (_, _, true) => Alpn::Siduck,
```
djc, line 175: I think we should get rid of the `_`s here, except for a fallthrough case that emits an error (or just panics).

interop/src/main.rs
```diff
@@ -180,9 +185,10 @@ async fn main() {
             .collect();
 
         peers.iter_mut().for_each(|mut x| {
-            match (opt.h3, opt.hq) {
-                (false, true) => x.alpn = Alpn::Hq,
-                (true, false) => x.alpn = Alpn::H3,
+            match (opt.h3, opt.hq, opt.siduck) {
+                (false, true, _) => x.alpn = Alpn::Hq,
+                (true, false, _) => x.alpn = Alpn::H3,
+                (_, _, true) => x.alpn = Alpn::Siduck,
```
djc, line 191: This looks suspiciously close to the other one. DRY? (I know it's not new, but I'll invoke the boy scout rule.)

interop/src/main.rs
```diff
@@ -219,10 +225,11 @@ async fn main() {
 async fn run(peer: &Peer, keylog: bool) -> Result<String> {
     let state = State::try_new(&peer, keylog)?;
     let result = match peer.alpn {
-        Alpn::Hq => state.run_hq().instrument(info_span!("hq")).await?,
-        _ => state.run_h3().instrument(info_span!("h3")).await?,
+        Alpn::Hq => state.run_hq().instrument(info_span!("hq")).await?.format(),
+        Alpn::Siduck => state.run_siduck().instrument(info_span!("siduck")).await?,
+        _ => state.run_h3().instrument(info_span!("h3")).await?.format(),
```
djc, line 230: Let's also get rid of the `_` in this case.

Ralith:
Comments applied. Looks like azure pipelines is sticking its nose into our business again for some reason?

djc:
That bit about Azure is weird. I've just deleted the Azure pipeline, so that should take care of business.

PR #740, opened 2020-05-02
Update for rustls quic API changes

Ralith:
Move computation of keys from secrets into rustls, and unbundle local/remote keys to more gracefully handle 0-RTT, which only has keys for one direction.

Draft until https://github.com/ctz/rustls/pull/362 is merged. Based on #742.


quinn-proto/src/crypto/rustls.rs
```diff
@@ -45,16 +43,31 @@ impl crypto::Session for TlsSession {
...
     type HmacKey = hmac::Key;
-    type HeaderKey = aead::quic::HeaderProtectionKey;
     type Key = PacketKey;
+    type HeaderKey = HeaderProtectionKey;
     type ServerConfig = Arc<rustls::ServerConfig>;
 
-    /// Create the initial set of keys given the initial ConnectionId
     fn initial_keys(
-        id: &ConnectionId,
+        conn_id: &ConnectionId,
```
djc, line 51: I think we don't use `conn_id` much because we usually use `cid` (though I guess only in combination with a prefix like `loc_cid`, `dst_cid` or `rem_cid`), maybe it might be nicer to reuse something common here?

quinn-proto/src/crypto/rustls.rs
```diff
@@ -313,3 +297,37 @@ const RETRY_INTEGRITY_KEY: [u8; 16] = [
 const RETRY_INTEGRITY_NONCE: [u8; 12] = [
     0x4d, 0x16, 0x11, 0xd0, 0x55, 0x13, 0xa5, 0x52, 0xc5, 0x87, 0xd5, 0x75,
 ];
+
```
djc, line 300: Let's move this up above `to_vec()` at least?

djc:
All the red in that last commit's diffs looks quite nice!

djc:
So do you want to merge this pointing to your branch for now, or shall we hold off until it gets merged upstream?

Ralith:
We haven't had any feedback from @ctz at all, so I'd rather not commit just yet. The rebase hazard is modest since we split out the incidental refactoring.

PR #742, opened 2020-05-03
Decouple secrets from keys, split keys

Ralith:
Allows more graceful handling of the one-directional 0-RTT keys, and paves the way for secrets to no longer be exposed from rustls.

Motivated by #740. Merging this separately will avoid severe rebase hazards, given that the rustls release cycle is quite slow.

### DIFF
--- a/quinn-proto/src/connection/spaces.rs
+++ b/quinn-proto/src/connection/spaces.rs
@@ -6,7 +6,9 @@ use std::{
 };
 
 use super::assembler::Assembler;
-use crate::{crypto, frame, range_set::RangeSet, shared::IssuedCid, StreamId, VarInt};
+use crate::{
+    crypto, crypto::KeyPair, frame, range_set::RangeSet, shared::IssuedCid, StreamId, VarInt,
+};
 
 pub(crate) struct PacketSpace<S>
 where
@@ -306,8 +308,8 @@ pub struct CryptoSpace<S>
 where
     S: crypto::Session,
 {
-    pub packet: S::Keys,
-    pub header: S::HeaderKeys,
+    pub packet: KeyPair<S::Key>,
+    pub header: KeyPair<S::HeaderKey>,
 }
 
 #[cfg(test)]
--- a/quinn-proto/src/connection/spaces.rs
+++ b/quinn-proto/src/connection/spaces.rs
@@ -308,7 +308,7 @@ pub struct CryptoSpace<S>
 where
     S: crypto::Session,
 {
-    pub packet: KeyPair<S::Key>,
+    pub packet: KeyPair<S::PacketKey>,
     pub header: KeyPair<S::HeaderKey>,
 }
 
--- a/quinn-proto/src/connection/spaces.rs
+++ b/quinn-proto/src/connection/spaces.rs
@@ -7,14 +7,14 @@ use std::{
 
 use super::assembler::Assembler;
 use crate::{
-    crypto, crypto::KeyPair, frame, range_set::RangeSet, shared::IssuedCid, StreamId, VarInt,
+    crypto, crypto::Keys, frame, range_set::RangeSet, shared::IssuedCid, StreamId, VarInt,
 };
 
 pub(crate) struct PacketSpace<S>
 where
     S: crypto::Session,
 {
-    pub(crate) crypto: Option<CryptoSpace<S>>,
+    pub(crate) crypto: Option<Keys<S>>,
     pub(crate) dedup: Dedup,
     /// Highest received packet number
     pub(crate) rx_packet: u64,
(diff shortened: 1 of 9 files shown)


quinn-proto/src/connection/spaces.rs
```diff
@@ -302,24 +302,12 @@ impl Dedup {
...
-where
-    K: crypto::Keys,
-{
-    pub fn new(packet: K) -> Self {
-        Self {
-            header: packet.header_keys(),
-            packet,
-        }
-    }
+    pub packet: crypto::KeyPair<S::Key>,
```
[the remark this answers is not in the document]
djc: (Also maybe consider importing `KeyPair` directly rather than prefixing with `crypto`? I think it was like this for `Session` because that word seems less obviously crypto-related, but that problem doesn't seek as big of an issue for `KeyPair`... and actually maybe it's also not an issue for `Session` in our code base?)

quinn-proto/src/connection/spaces.rs
```diff
@@ -302,24 +302,12 @@ impl Dedup {
     }
 }
 
-pub struct CryptoSpace<K>
+pub struct CryptoSpace<S>
```
djc, line 305: I feel like maybe this should now just be called `Keys` and move into `crypto`?

quinn-proto/src/crypto/ring.rs
```diff
@@ -4,43 +4,23 @@ use ring::{aead, hkdf, hmac};
...
 
 /// Keys for encrypting and decrypting packet payloads
-pub struct Crypto {
-    pub(crate) local_secret: hkdf::Prk,
-    local_iv: Iv,
-    sealing_key: aead::LessSafeKey,
-    pub(crate) remote_secret: hkdf::Prk,
-    remote_iv: Iv,
-    opening_key: aead::LessSafeKey,
+pub struct PacketKey {
```
djc, line 14: (Maybe rename to `PayloadKey`?)

quinn-proto/src/crypto/rustls.rs
```diff
@@ -5,34 +5,38 @@ use std::{
...
 use rustls::{
     self,
-    quic::{ClientQuicExt, ServerQuicExt},
+    quic::{ClientQuicExt, Secrets, ServerQuicExt},
     Session,
 };
 use webpki::DNSNameRef;
 
-use super::ring::Crypto;
+use super::ring::{header_key_from_secret, initial_keys, PacketKey};
```
djc, line 17: I think I would prefer not to important these by name here and just reference them through `ring` to make their provenance more obvious.

quinn-proto/src/crypto/rustls.rs
```diff
@@ -106,33 +118,39 @@ impl crypto::Session for TlsSession {
...
-        ))
-    }
 
-    fn update_keys(&self, keys: &Self::Keys) -> Self::Keys {
-        let (client_secret, server_secret) = match self.side() {
-            Side::Client => (&keys.local_secret, &keys.remote_secret),
-            Side::Server => (&keys.remote_secret, &keys.local_secret),
-        };
+        let aead = suite.get_aead_alg();
+        let result = Some(CryptoSpace {
```
djc, line 129: I feel like we could reduce coupling here by having the `ring` module contain the code to create a `CryptoSpace` from `aead`, `local` and `remote`?

quinn-proto/src/connection/spaces.rs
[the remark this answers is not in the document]
Ralith: I went back and forth on this a few times, but [the TLS draft](https://quicwg.org/base-drafts/draft-ietf-quic-tls.html) consistently uses the term "packet protection key," and I think that wins out.

quinn-proto/src/crypto/ring.rs
Ralith: As above, I think the draft's use of "packet protection key" favors the current form.

quinn-proto/src/crypto/rustls.rs
Ralith: That runs into awkward collisions with the crate name. We could work around those, but I think it's confusing to read as well.

Ralith:
Applied most comments; may not have time to try to split the big commit for a bit yet, though.

quinn-proto/src/connection/spaces.rs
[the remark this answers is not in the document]
djc: Fair enough -- I've instead brought it up in quicdev to see if we should have it renamed in the draft.

quinn-proto/src/crypto/ring.rs
djc: Yup.

quinn-proto/src/crypto/rustls.rs
djc: Okay, let's leave it.

Ralith:
Okay, split things up a bunch.

Ralith:
This layout is motivated by the fact that when 1-RTT keys are updated, we
replace the packet protection keys but not the header protection keys.
There's advantages to the other way too, though, particularly as 0-RTT
calls for the opposite grouping, so my feeling isn't too strong. Thoughts?

djc:
Maybe I'll try it and we'll see what looks better.

PR #748, opened 2020-05-04
Reset interop server initial congestion window to default

Ralith:
Eases testing of congestion window behavior.


djc:
This was changed for H3 perf reasons, right? So we test that some other way?

Ralith:
This was one of multiple changes; I haven't determined how much performance is compromised by rolling it back alone. A significant slowdown is probably worth investigating, because TCP should typically have a similar initial congestion window.

PR #751, opened 2020-05-06
 Signal ALPN/SNI availability during handshake

### LINKED ISSUE #749: authentication_data is unreliable during handshake
When the TLS `ClientHello` is not fully contained within the first packet we receive (e.g. due to reordering or large size), the application may be notified of an incoming connection for which `authentication_data` returns nothing.  Processing 0-RTT with minimum latency requires access to ALPN results, so waiting for the handshake to complete is not an acceptable solution. quinn-proto should therefore expose an event notifying the application of when `authentication_data` has become populated, and quinn should expose that information through a future.

Ralith:
Fixes #749. Will improve our performance on the quic-tracker interop tests.

### DIFF
--- a/quinn-proto/src/crypto/rustls.rs
+++ b/quinn-proto/src/crypto/rustls.rs
@@ -25,6 +25,7 @@ use crate::{
 /// A rustls TLS session
 pub struct TlsSession {
     using_alpn: bool,
+    got_handshake_data: bool,
     inner: SessionKind,
 }
 
@@ -43,7 +44,8 @@ impl TlsSession {
 }
 
 impl crypto::Session for TlsSession {
-    type AuthenticationData = AuthenticationData;
+    type HandshakeData = HandshakeData;
+    type Identity = CertificateChain;
     type ClientConfig = Arc<rustls::ClientConfig>;
     type HmacKey = hmac::Key;
     type PacketKey = PacketKey;
@@ -70,15 +72,21 @@ impl crypto::Session for TlsSession {
         }
     }
 
-    fn authentication_data(&self) -> AuthenticationData {
-        AuthenticationData {
-            peer_certificates: self.get_peer_certificates().map(|v| v.into()),
-            protocol: self.get_alpn_protocol().map(|p| p.into()),
+    fn handshake_data(&self) -> Option<HandshakeData> {
+        if !self.got_handshake_data {
+            return None;
+        }
+        Some(HandshakeData {
+            protocol: self.get_alpn_protocol().map(|x| x.into()),
             server_name: match self.inner {
                 SessionKind::Client(_) => None,
-                SessionKind::Server(ref session) => session.get_sni_hostname().map(|s| s.into()),
+                SessionKind::Server(ref session) => session.get_sni_hostname().map(|x| x.into()),
             },
-        }
+        })
+    }
+
+    fn peer_identity(&self) -> Option<CertificateChain> {
+        self.get_peer_certificates().map(|v| v.into())
     }
 
     fn early_crypto(&self) -> Option<(Self::HeaderKey, Self::PacketKey)> {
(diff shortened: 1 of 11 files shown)


quinn/src/connection.rs
```diff
@@ -38,12 +38,41 @@ where
 
 impl<S> Connecting<S>
 where
-    S: proto::crypto::Session,
+    S: proto::crypto::Session + 'static,
 {
-    pub(crate) fn new(conn: ConnectionRef<S>, connected: oneshot::Receiver<bool>) -> Self {
-        Self {
+    pub(crate) fn new(
```
djc, line 44: I'm fully on board with encapsulating the coupling of the driver and `Connecting`. However, this change appears to closely couple `Connecting` and the `ConnectionRef`, and this seems unnecessary to achieve the former goal (we could still make `ConnectionRef::new()` private to the `connection` module, of course).

quinn-proto/src/crypto/rustls.rs
```diff
@@ -217,16 +219,16 @@ pub struct AuthenticationData {
     /// For clients, this is the certificate chain of the server. For servers, this is the
     /// certificate chain of the client, if client authentication was completed.
     ///
-    /// `None` if this data was requested from the session before this value is available.
-    ///
-    /// If this is `None`, and `Connection::is_handshaking` returns `false`, the connection
-    /// will have already been closed.
+    /// Available while fully connected, if provided by the peer.
```
djc, line 222: So I feel like it's a bit ugly that the future resolving does not have much bearing on the availability of the `peer_certificates`. Should we separate them out after all?

quinn/src/connection.rs
Ralith: I'm a bit concerned about the growth of the `ConnectionRef::new` arguments list, but it's not untenable just yet. Applied.

quinn-proto/src/crypto/rustls.rs
Ralith: I agree that this is a bit ugly. I'm uncertain what a more fine-grained version should look like. The most obvious approach is just split `AuthenticationData` into `EarlyAuthenticationData` and `LateAuthenticationData`, but that's awfully abstract/obscure. We could return to hardcoded interfaces for SNI and ALPN and leave `AuthenticationData` for the certificates; I think this is a bit more appealing, as stubbing out SNI/ALPN when unavailable doesn't seem any worse than having a weird empty `EarlyAuthenticationData`.

djc: I agree that an early/late split isn't particularly appealing. However, thinking about it more, the current approach with the ALPN protocol is also wrong. As I understand, applications are not required to use ALPN, in which case this future doesn't actually ever wake (I think).

"TLS uses Application Layer Protocol Negotiation (ALPN) to select an application protocol. Unless another mechanism is used for agreeing on an application protocol, endpoints MUST use ALPN for this purpose." (-tls, section 8.1)

As such, ideally we should get insight from rustls on how far along the handshake is. I suppose the alternative solution is that we wake the task up when the handshake terminates.

For just certificates, the name `AuthenticationData` feel overly broad. In that case, we might just call it `PeerIdentity` or some such.

Ralith: > applications are not required to use ALPN

Whoops, good catch. I don't think SNI is required either, so we indeed need a new signal from rustls here. Something like the recent proposal in https://github.com/ctz/rustls/issues/89 would trivialize this, but a lower-impact solution would be a new state accessor in the QUIC-specific API.

Ralith: > I suppose the alternative solution is that we wake the task up when the handshake terminates.

That would defeat the goal of allowing ALPN-dependent 0-RTT data to be processed immediately.

> For just certificates, the name AuthenticationData feel overly broad. In that case, we might just call it PeerIdentity or some such.

Yeah, that makes sense.

djc: > That would defeat the goal of allowing ALPN-dependent 0-RTT data to be processed immediately.

Well, you can still optimize for the case where ALPN is provided by the client, right? Which it's easy for any protocol that wants the performance to opt in to, supposedly.

Ralith: Oh, I see what you mean. Yeah, that'd be a decent short-term solution in lieu of a rustls change.

Ralith:
I've rebased and split `AuthenticationData` into separate `HandshakeData` and `Identity` structures, and focused the async notification (which is now reliably delivered) around the former. Note that github is misordering the commits fairly severely.

Issue #723, opened 2020-04-20
H3: Add API to monitor client closing request stream


djc:
In h2, `SendResponse` has a `poll_reset()` method. I think we should have something similar:

https://docs.rs/h2/0.2.4/h2/server/struct.SendResponse.html#method.poll_reset

stammw:
I can work on this if you want ?

djc:
Would be great. There are some other things I'm looking into right now.

djc:
That is, I think we should leverage the `http_body` crate's `Body` trait, which already looks quite similar to what we've got going, and use it in our APIs.

djc:
(Potentially we should constrain our `Body` type to implementing that trait and having some minimal constructors, but not much else.)

stammw:
Yeah sure, HttpBody might also be an for future proxy implementation.

PR #755, opened 2020-05-08
H3 poll reset

stammw:
Addresses #723.

It adds a `StreamEvent::Stopped` variant to carry a `STOP_SENDING` event to the user.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -731,6 +731,11 @@ where
         Ok(())
     }
 
+    /// Check if this stream was stopped, get the reason if it was
+    pub fn stopped(&mut self, id: StreamId) -> Result<Option<VarInt>, UnknownStream> {
+        self.streams.stop_reason(id)
+    }
+
     /// Finish a send stream, signalling that no more data will be sent.
     ///
     /// If this fails, no [`Event::StreamFinished`] will be generated.
--- a/quinn-proto/src/connection/streams.rs
+++ b/quinn-proto/src/connection/streams.rs
@@ -362,6 +362,11 @@ impl Streams {
             return;
         }
 
+        if let Some(error_code) = stop_reason {
+            self.events
+                .push_back(StreamEvent::Stopped { error_code, id });
+        }
+
         // Restore the portion of the send window consumed by the data that we aren't about to
         // send. We leave flow control alone because the peer's responsible for issuing additional
         // credit based on the final offset communicated in the RESET_STREAM frame we send.
@@ -405,6 +410,13 @@ impl Streams {
         // Issue flow control credit for unread data
         self.local_max_data += stream.assembler.limit() - stream.assembler.bytes_read();
         Ok(!stream.is_finished())
+    }
+
+    pub fn stop_reason(&self, id: StreamId) -> Result<Option<VarInt>, UnknownStream> {
+        match self.send.get(&id) {
+            Some(s) => Ok(s.stop_reason()),
+            None => Err(UnknownStream { _private: () }),
+        }
     }
 
     pub fn can_send(&self) -> bool {
@@ -849,6 +861,15 @@ impl Send {
         }
     }
 
+    fn stop_reason(&self) -> Option<VarInt> {
+        match self.state {
+            SendState::ResetSent { stop_reason } | SendState::ResetRecvd { stop_reason } => {
+                stop_reason
+            }
+            _ => None,
+        }
+    }
+
     fn take_stop_reason(&mut self) -> Option<VarInt> {
         match self.state {
             SendState::ResetSent {
@@ -1217,6 +1238,13 @@ pub enum StreamEvent {
         id: StreamId,
         /// Error code supplied by the peer if the stream was stopped
         stop_reason: Option<VarInt>,
+    },
+    /// The peer asked us to stop sending on an outgoing stream
+    Stopped {
+        /// Which stream has been stopped
+        id: StreamId,
+        /// Error code supplied by the peer
+        error_code: VarInt,
     },
     /// At least one new stream of a certain directionality may be opened
     Available {
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -573,6 +573,7 @@ where
             incoming_bi_streams_reader: None,
             datagram_reader: None,
             finishing: HashMap::new(),
+            stopped: HashMap::new(),
             error: None,
             ref_count: 0,
         })))
@@ -638,6 +639,7 @@ where
     incoming_bi_streams_reader: Option<Waker>,
     datagram_reader: Option<Waker>,
     pub(crate) finishing: HashMap<StreamId, oneshot::Sender<Option<WriteError>>>,
+    pub(crate) stopped: HashMap<StreamId, Waker>,
     /// Always set to Some before the connection becomes drained
     pub(crate) error: Option<ConnectionError>,
     /// Number of live handles that can be used to initiate or handle I/O; excludes the driver
@@ -743,6 +745,11 @@ where
                         let _ = finishing.send(stop_reason.map(WriteError::Stopped));
                     }
                 }
+                Stream(StreamEvent::Stopped { id, .. }) => {
+                    if let Some(stopped) = self.stopped.remove(&id) {
+                        stopped.wake();
+                    }
+                }
             }
         }
     }
@@ -815,6 +822,9 @@ where
         }
         if let Some(x) = self.on_connected.take() {
             let _ = x.send(false);
+        }
+        for (_, waker) in self.stopped.drain() {
+            waker.wake();
         }
     }
 
--- a/quinn/src/streams.rs
+++ b/quinn/src/streams.rs
@@ -149,6 +149,30 @@ where
         conn.wake();
     }
 
+    /// Completes if/when the peer stops the stream, yielding the error code
+    pub fn stopped(&mut self) -> Stopped<'_, S> {
+        Stopped { stream: self }
+    }
+
+    #[doc(hidden)]
+    pub fn poll_stopped(&mut self, cx: &mut Context) -> Poll<Result<VarInt, StoppedError>> {
+        let mut conn = self.conn.lock().unwrap();
+
+        if self.is_0rtt {
+            conn.check_0rtt()
+                .map_err(|()| StoppedError::ZeroRttRejected)?;
+        }
+
+        match conn.inner.stopped(self.stream) {
+            Err(_) => Poll::Ready(Err(StoppedError::UnknownStream)),
+            Ok(Some(error_code)) => Poll::Ready(Ok(error_code)),
+            Ok(None) => {
+                conn.stopped.insert(self.stream, cx.waker().clone());
+                Poll::Pending
+            }
+        }
+    }
+
     #[doc(hidden)]
     pub fn id(&self) -> StreamId {
         self.stream
@@ -227,6 +251,25 @@ where
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
         self.get_mut().stream.poll_finish(cx)
+    }
+}
+
+/// Future produced by `SendStream::stopped`
+pub struct Stopped<'a, S>
+where
+    S: proto::crypto::Session,
+{
+    stream: &'a mut SendStream<S>,
+}
+
+impl<S> Future for Stopped<'_, S>
+where
+    S: proto::crypto::Session,
+{
+    type Output = Result<VarInt, StoppedError>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        self.get_mut().stream.poll_stopped(cx)
     }
 }
 
@@ -570,6 +613,22 @@ pub enum WriteError {
     /// Carries an application-defined error code.
     #[error(display = "sending stopped by peer: error {}", 0)]
     Stopped(VarInt),
+    /// The connection was closed.
+    #[error(display = "connection closed: {}", _0)]
+    ConnectionClosed(ConnectionError),
+    /// Unknown stream
+    #[error(display = "unknown stream")]
+    UnknownStream,
+    /// This was a 0-RTT stream and the server rejected it.
+    ///
+    /// Can only occur on clients for 0-RTT streams (opened using `Connecting::into_0rtt()`).
+    #[error(display = "0-RTT rejected")]
+    ZeroRttRejected,
+}
+
+/// Errors that arise while monitoring for a send stream stop from the peer
+#[derive(Debug, Error, Clone, PartialEq, Eq)]
+pub enum StoppedError {
     /// The connection was closed.
     #[error(display = "connection closed: {}", _0)]
     ConnectionClosed(ConnectionError),
--- a/quinn-h3/src/data.rs
+++ b/quinn-h3/src/data.rs
@@ -22,7 +22,7 @@ use crate::{
         ErrorCode,
     },
     streams::Reset,
-    Error,
+    Error, HttpError,
 };
 
 /// Represent data transmission completion for a Request or a Response
@@ -84,6 +84,14 @@ where
     pub fn cancel(&mut self) {
         self.send.reset(ErrorCode::REQUEST_CANCELLED.into());
         self.state = SendDataState::Finished;
+    }
+
+    /// Monitor stop sending signal from the peer
+    ///
+    /// This will return `Ready` when a STOP_SENDING frame from the peer has
+    /// been received for this stream. Else, it will return `Pending` indefinitely.
+    pub fn poll_stopped(&mut self, cx: &mut Context) -> Poll<Result<HttpError, Error>> {
+        Poll::Ready(Ok(ready!(self.send.poll_stopped(cx)?).into()))
     }
 }
 
--- a/quinn-h3/src/lib.rs
+++ b/quinn-h3/src/lib.rs
@@ -109,7 +109,7 @@ mod qpack;
 pub mod qpack;
 
 use err_derive::Error;
-use quinn::{ApplicationClose, ConnectionError, ReadError, WriteError};
+use quinn::{ApplicationClose, ConnectionError, ReadError, StoppedError, VarInt, WriteError};
 use std::{error::Error as StdError, io::ErrorKind};
 
 use proto::ErrorCode;
@@ -135,6 +135,9 @@ pub enum Error {
     /// A `QUIC`-specific write error occurred
     #[error(display = "QUIC write error: {}", _0)]
     Write(WriteError),
+    /// A `QUIC`-specific error occurred while polling for STOP_SENDING
+    #[error(display = "QUIC error while polling for STOP_SENDING: {}", _0)]
+    Stopped(StoppedError),
     /// Programming error within the crate's code
     #[error(display = "Internal error: {}", _0)]
     Internal(String),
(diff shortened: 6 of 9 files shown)


quinn/src/streams.rs
```diff
@@ -149,6 +151,25 @@ where
         conn.wake();
     }
 
+    /// Monitor peer's reset event
+    ///
+    /// This will return the reason of reset when receiving
+    pub fn poll_reset(&mut self, cx: &mut Context) -> Poll<Option<VarInt>> {
```
Ralith, line 157: Resets are already exposed via `ReadError::Reset`.

quinn/src/connection.rs
```diff
@@ -741,6 +743,11 @@ where
                         let _ = finishing.send(stop_reason.map(WriteError::Stopped));
                     }
                 }
+                Stream(StreamEvent::Stopped { id, stop_reason }) => {
```
Ralith, line 746: A stopped stream and a reset stream are different things. A stream is stopped when the receiver tells the sender it's no longer interested (via `STOP_SENDING`). A stream is reset when the sender tells the receiver it's aborted sending on that stream (via `RESET_STREAM`).

quinn-proto/src/connection/streams.rs
```diff
@@ -1201,6 +1206,13 @@ pub enum StreamEvent {
         /// Error code supplied by the peer if the stream was stopped
         stop_reason: Option<VarInt>,
     },
+    /// A STOP_SENDING have been recieved for an outgoing stream
+    Stopped {
```
Ralith, line 1210: This is already exposed via `WriteError::Stopped` and `FinishError::Stopped`. Is there a case those don't cover?

quinn-proto/src/connection/mod.rs
```diff
@@ -745,7 +745,7 @@ where
     /// caller to defer materializing a datagram until one can be sent immediately without redundant
     /// checks
     ///
-    /// Returns `Err` iff a `len`-byte datagram cannot currently be sent
+    /// Returns `Err` if a `len`-byte datagram cannot currently be sent
```
Ralith, line 748: This isn't a typo. "iff" is an abbreviation for ["if and only if"](https://en.wikipedia.org/wiki/If_and_only_if), which refers to mutual implication.

quinn/src/streams.rs
stammw: Here I've chosen to offer the same method name as `h2`. But essentially, this monitors `STOP_SENDING` events. Using QUIC's vocabulary would maybe make more sense, @djc?

quinn/src/connection.rs
stammw: This is why this variant is called Stopped. I don't see your point here?

quinn-proto/src/connection/streams.rs
stammw: This is exactly what thought at first. But Hyper wants to poll for STOP_SENDING before a response as been started. I guess this helps in heavy load conditions, where you want to handle clients abandon the lightest way as possible.

quinn-proto/src/connection/mod.rs
stammw: Ah, good to know then.

stammw:
Regarding the duplication, I understood that `StreamEvents::Finished` is fired once a finishing event as been Acked. And this is not what we want. I want `poll_reset()` to resolve only after receiving a `STOP_SENDING`.  I guess `stop_reason` option of `Finished` can be used to indicate the event of `STOP_SENDING` reception, but I'm not sure when this is fired exactly ? 

Plus, this makes a boolean-to-code-path pattern I'd rather see flattened amongst higher level variants of events. But I'm here to discuss it :)

quinn/src/connection.rs
Ralith: The term `reset` is used in the body of this case.

quinn/src/streams.rs
Ralith: Using "reset" to describe `STOP_SENDING` is specifically incorrect, yeah.

quinn-proto/src/connection/streams.rs
Ralith: What constitutes starting a response?

Ralith:
> I understood that StreamEvents::Finished is fired once a finishing event as been Acked

That's not what's being duplicated. `STOP_SENDING` is presently communicated to the application by an error on a future write attempt or, if a finish is already underway, a `Finished` event containing a stop reason. Neither of these involve a finish being acknowledged.

> Plus, this makes a boolean-to-code-path pattern I'd rather see flattened amongst higher level variants of events.

I agree that the current semantics of `Finished` are ugly due to ~boolean~ `Option` blindness and confusing because a stopped stream is not, in fact, finished. I think introducing `StreamEvent::Stopped` is an improvement *if* the `stop_reason` field is removed from `StreamEvent::Finished`. This will require modifying `streams::Send::reset` to emit `Stopped` rather than `Finished` or `StreamWritable`, and updating code that handles those cases, taking care to wake up both writers and finishers in `quinn` on `Stopped`. I consider these changes justified on ergonomic and code quality grounds alone.

At this point the new path is no longer a duplicate, as the old path will have been removed. A dedicated interface to check for stopped-ness directly will then be a trivial extension. However, it's unclear if hyper (or any other hypothetical caller) needs a synchronous check or a future. A synchronous check is much simpler, and suits the described use case of optimizing code that would otherwise spend resources preparing an unwanted response. If a future is desired, we should probably expose an actual future rather than a `poll` function.

Ralith:
We may also need to reevaluate the lifetime of stream states. Currently we dispose of a formerly writable stopped stream's state when the stop is observed by the application via a write or finish operation. If we're providing direct notifications of the stop, we can no longer trivially guarantee that the application will perform an operation on the stream in the future. However, we dispose of the stream immediately after issuing the notification, then reasonable applications might experience a race condition between processing the notification and performing a stream operation; if the stream operation wins, a spurious "unknown stream" error would result.

Most likely we should document a requirement that streams always be explicitly disposed by calling `finish` or `reset`; this dovetails gracefully with the draft's recommendation that stopped streams be reset, so we might go further and remove the current transport-layer implicit reset on stop. Regardless, we could then refactor state disposal for send streams to occur immediately after issuing a reset or observing acknowledgement of a finish, which is an appealing simplification.

I'll need to think on this further when it's not so late to be confident of correctness, however.

stammw:
Now I remember what pushed me to add `Stopped`: `stop_reason` is taken at some point, probably to signify the `STOP_SENDING` has been treated. I didn't want to interfere with this mechanic (mainly because I didn't put enough effort on understanding it).

quinn-proto/src/connection/streams.rs
stammw: I'm still a little confused. My starting point for this PR is [the `poll_reset()` h2 method](https://docs.rs/h2/0.2.5/h2/struct.SendStream.html#method.poll_reset), which belongs to the equivalent of `SendData` in quinn-h3. It seems intended to be polled before issuing some data to send. I think it should be polled while the web application is processing the request. The request has been passed to the upper layer but no data is ready for sending. At this time, the client can timeout, or react to a new action and cancel loading a resource or whatever (think of someone smashing a `next` link for example).

quinn/src/connection.rs
stammw: Oh! yeah that is an artifact! -_- 
Thanks!

quinn/src/streams.rs
stammw: Is `poll_stopped()` good enough?

quinn-proto/src/connection/streams.rs
Ralith: That sure sounds like a synchronous check, i.e. `fn is_reset(&self) -> bool`, should be sufficient.

quinn/src/streams.rs
```diff
@@ -31,6 +31,7 @@ where
     stream: StreamId,
     is_0rtt: bool,
     finishing: Option<oneshot::Receiver<Option<WriteError>>>,
+    reset_event: Option<oneshot::Receiver<VarInt>>,
```
Ralith, line 34: This is also a misnomer.

quinn-proto/src/connection/streams.rs
stammw: Maybe I don't fully understand the use case then. It seems to me doing that would turn this behavior into active polling. Task waking is still useful here because we want to receive stop events for many requests, while serving many other requests.

quinn/src/streams.rs
stammw: Sorry for the messy PR :/

Ralith: No worries, that's why we have code review.

quinn-proto/src/connection/streams.rs
Ralith: I think I misunderstood your example. You're saying that you'd select on stop and some asynchronous work required to prepare a response?

stammw: Essentially, an HTTP lib receives a request, pass it to the application. At this point a computation is done that can take time (DB access, heavy load etc) and will result into a response.

While the application prepares this, the lib as one role: listen for cancellation events from the peer, and pass it to the application as soon as possible so the expensive computation of preparing a response can be shut down. 

At larger scale, a server with heavy load will want to cleanup cancelled requests as fast as possible to recover resources as they aren't needed anymore. Especially when the clients start to timeout because of the load, or people start smashing buttons, because of the load.

When the response is ready, you can totally pass the `STOP_SENDING` event to the app by throwing errors through any API that is polling for transmit. But until the response is ready for writing, this API stays untouched. 

This is why the App task responsible of the request processing needs to be waken instantly when receiving such a cancellation event.

But again, that only is a humble comprehension of this API needed by hyper.

Ralith: Okay, that seems reasonable; let's expose a future. I don't have a lot of opinions about the structure of HTTP things; my focus is more on the consequences at the transport layer.

quinn-proto/src/connection/streams.rs
```diff
@@ -1178,6 +1183,13 @@ pub enum StreamEvent {
         /// Error code supplied by the peer if the stream was stopped
         stop_reason: Option<VarInt>,
     },
+    /// A STOP_SENDING have been recieved for an outgoing stream
```
djc, line 1186:
```suggestion
    /// A STOP_SENDING have been received for an outgoing stream
```

quinn-h3/src/frame.rs
```diff
@@ -146,6 +146,13 @@ where
             s.reset(err_code.into());
         }
     }
+
```
djc, line 149: I feel like this layer isn't adding much value. Can we bypass it by making some inner stuff `pub(crate)` or something?

quinn-h3/src/headers.rs
```diff
@@ -63,6 +63,10 @@ impl SendHeaders {
     pub fn reset(&mut self, err_code: ErrorCode) {
         self.0.reset(err_code);
     }
+
+    pub fn poll_stopped(&mut self, cx: &mut Context) -> Poll<Option<ErrorCode>> {
```
djc, line 67: Same here, just passing through the method call doesn't seem very useful.

stammw:
I still didn't grasp the reset / stop_sending mechanism for now (didn't spend much time and effort) . @Ralith do you have some more directions on getting stop_sending out after further thinking?

Ralith:
The transport layer needs to be refactored to remove `stop_reason` from `StreamFinished`, and to not implicitly dispose of stopped streams. I can have a go at this myself in the near future if you'd like, though I'm a bit swamped at the moment--in particular, I want to resolve #774 first.

stammw:
All right, so we can merge something close to this PR, then apply this refactoring in another PR ? Or is it necessary for validity?

quinn/src/streams.rs
```diff
@@ -149,6 +151,25 @@ where
         conn.wake();
     }
 
+    /// Monitor peer's STOP_SENDING issue for this stream
+    ///
+    /// This will return the reason of reset when receiving
+    pub fn poll_stopped(&mut self, cx: &mut Context) -> Poll<Option<VarInt>> {
```
Ralith, line 157: Why does this return an `Option`? What does it mean if `None` is returned?

I'd also prefer to expose a future, rather than a `poll_` function.

quinn/src/streams.rs
```diff
@@ -149,6 +151,25 @@ where
         conn.wake();
     }
 
+    /// Monitor peer's STOP_SENDING issue for this stream
+    ///
+    /// This will return the reason of reset when receiving
```
Ralith, line 156:
```suggestion
    /// Completes if/when the peer stops the stream, yielding the error code
```

quinn/src/streams.rs
djc: IIRC in the end this channels into hyper API where it needs to be part of a named future, which I think means we can't only have an async API.

Ralith: We could provide a named future, or both a future and a (`#[doc(hidden)]`?) `poll_` fn as we currently do for `poll_finish`.

stammw: I was concerned about the future non-resolvabity after some events (like closing, errors). But after all this should be handled by the user through other API. And yes, `None` should have been replaced by  an `Result::Err()`.

But I got rid of it.

Also, I implemented a `Stopped` future similar to `Finish`

quinn/src/streams.rs
stammw: Done

quinn/src/streams.rs
```diff
@@ -149,6 +151,27 @@ where
...
+    }
+
+    #[doc(hidden)]
+    pub fn poll_stopped(&mut self, cx: &mut Context) -> Poll<VarInt> {
+        let mut conn = self.conn.lock().unwrap();
+        if self.stopped_event.is_none() {
+            let (send, recv) = oneshot::channel();
+            self.stopped_event = Some(recv);
+            conn.stopped.insert(self.stream, send);
+        }
```
Ralith, line 166: What if the stream was stopped before this function was called for the first time?

quinn-proto/src/connection/streams.rs
```diff
@@ -1218,6 +1223,13 @@ pub enum StreamEvent {
         /// Error code supplied by the peer if the stream was stopped
         stop_reason: Option<VarInt>,
     },
+    /// A STOP_SENDING have been received for an outgoing stream
+    Stopped {
+        /// Which stream has been stopped
+        id: StreamId,
+        /// Error code supplied by the peer
+        stop_reason: VarInt,
```
Ralith, line 1231:
```suggestion
        error_code: VarInt,
```
Let's be consistent with the naming in the spec, and the argument naming for `Stream::stop`.

quinn/src/streams.rs
```diff
@@ -149,6 +151,27 @@ where
...
+
+        self.stopped_event
+            .as_mut()
+            .unwrap()
+            .poll_unpin(cx)
+            .map(|x| x.unwrap())
```
Ralith, line 172: Should we return an error rather than hanging if the connection is lost? Hung futures strike me as a footgun in general, since they produce leaks if you block a task on them. It's hard to imagine why anyone would block a task solely on waiting for a stream to become stopped, since that might never happen and the `&mut` prevents you from doing anything more useful with the stream in the mean time, so I'm not sure, but my inclination is to favor consistency with the other futures (which all yield errors) and minimize potential for surprising behavior.

This should be as simple as checking `conn.error` when polling the oneshot yields `Pending`, as we're doing in `poll_finish`.

quinn/src/streams.rs
stammw: That's a good question! I'll seek a place to store some state with `stop_reason` somewhere.

stammw:
Thanks for the mentoring !

quinn/src/streams.rs
```diff
@@ -149,6 +151,39 @@ where
...
+
+    #[doc(hidden)]
+    pub fn poll_stopped(&mut self, cx: &mut Context) -> Poll<Result<VarInt, WriteError>> {
+        let mut conn = self.conn.lock().unwrap();
+
+        if self.stopped_event.is_none() {
+            match conn.inner.stopped(self.stream) {
+                Ok(Some(error_code)) => return Poll::Ready(Ok(error_code)),
+                Err(_) => return Poll::Ready(Err(WriteError::UnknownStream)),
+                _ => (),
```
Ralith, line 167: Let's fill in these wildcards with explicit patterns in the two cases above so we can benefit from rust's totality checking in the future.

quinn/src/streams.rs
```diff
@@ -149,6 +151,39 @@ where
...
     }
 
+    /// Completes if/when the peer stops the stream, yielding the error code
+    pub fn stopped(&mut self) -> Stopped<'_, S> {
+        Stopped { stream: self }
+    }
+
+    #[doc(hidden)]
+    pub fn poll_stopped(&mut self, cx: &mut Context) -> Poll<Result<VarInt, WriteError>> {
+        let mut conn = self.conn.lock().unwrap();
```
Ralith, line 161: I think this could reasonably be a 0-RTT stream, in which case we should probably handle the possibility of 0-RTT being rejected and emit a `ZeroRttRejected` error, similar to `finish`. That way a client attempting 0-RTT doesn't and up inadvertently waiting for an unrelated stream to be stopped.

quinn/src/streams.rs
```diff
@@ -149,6 +151,39 @@ where
         conn.wake();
     }
 
+    /// Completes if/when the peer stops the stream, yielding the error code
+    pub fn stopped(&mut self) -> Stopped<'_, S> {
+        Stopped { stream: self }
+    }
+
+    #[doc(hidden)]
+    pub fn poll_stopped(&mut self, cx: &mut Context) -> Poll<Result<VarInt, WriteError>> {
```
Ralith, line 160: It feels a little odd to return `WriteError` here due to the `WriteError::Stopped` case; a user might reasonably be confused about how a stopped stream is represented. Maybe it's worth a dedicated error type? Another option might be to return `WriteError` directly--after all, it does have exactly the cases we care about--but that leaves the user without an ergonomic way to distinguish errors from success, which doesn't seem great. Open to discussion.

quinn/src/connection.rs
```diff
@@ -638,6 +639,7 @@ where
     incoming_bi_streams_reader: Option<Waker>,
     datagram_reader: Option<Waker>,
     pub(crate) finishing: HashMap<StreamId, oneshot::Sender<Option<WriteError>>>,
+    pub(crate) stopped: HashMap<StreamId, oneshot::Sender<VarInt>>,
```
Ralith, line 642: To ensure blocked tasks wake up when the connection is lost (e.g. due to timeout), we need to drain this in `Connection::terminate`, similar to `blocked_readers`/`writers`.

quinn/src/streams.rs
```diff
@@ -149,6 +151,39 @@ where
...
+        let mut conn = self.conn.lock().unwrap();
+
+        if self.stopped_event.is_none() {
+            match conn.inner.stopped(self.stream) {
+                Ok(Some(error_code)) => return Poll::Ready(Ok(error_code)),
+                Err(_) => return Poll::Ready(Err(WriteError::UnknownStream)),
+                _ => (),
+            }
+
+            let (send, recv) = oneshot::channel();
```
Ralith, line 170: Since the error code is now persisted in the inner `Connection`, we could simplify this a bit by replacing the oneshot table with a `Waker` table and just calling  `conn.inner.stopped(self.stream)` each poll rather than only once.

quinn/src/streams.rs
stammw: I agree with you but I chose not to add another type for this (at the moment). I think returning "XXXError" type not in an `Err` variant is weird, unless the function call is explicit about error management. 

I guess I prefer the custom type (`WriteError`, but without `Stopped` variant).

stammw:
@Ralith I addressed all your comments in new commits. Let me know if you want to see them squashed.

quinn-proto/src/connection/streams.rs
```diff
@@ -1218,6 +1239,13 @@ pub enum StreamEvent {
         /// Error code supplied by the peer if the stream was stopped
         stop_reason: Option<VarInt>,
     },
+    /// A STOP_SENDING have been received for an outgoing stream
```
Ralith, line 1242:
```suggestion
    /// The peer asked us to stop sending on an outgoing stream
```

stammw: Thanks!

stammw:
Looks like I need an approval, even if I dismissed your review. (Got your requested change into the original commit).

djc:
Also +1 for some squashing. I can approve/merge if Ralith is asleep by then.

stammw:
Squashed everything into two commits

djc:
Good stuff!

PR #756, opened 2020-05-08
Allocate exactly the necessary receive buffer size

Ralith:
Reduces memory use a bit, a benefit which will be compounded dramatically when we switch to receiving multiple packets per syscall, via `recvmmsg` or other means.

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -38,7 +38,7 @@ pub struct TransportConfig {
     pub(crate) time_threshold: f32,
     pub(crate) initial_rtt: Duration,
 
-    pub(crate) max_datagram_size: u64,
+    pub(crate) max_udp_payload_size: u64,
     pub(crate) initial_window: u64,
     pub(crate) minimum_window: u64,
     pub(crate) loss_reduction_factor: f32,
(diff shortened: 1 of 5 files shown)


quinn-proto/src/config.rs
```diff
@@ -319,6 +319,11 @@ where
         self.max_udp_payload_size = value;
         self
     }
+
+    /// Get the current value of `max_udp_payload_size`
+    pub fn get_max_udp_payload_size(&self) -> u64 {
+        self.max_udp_payload_size
+    }
```
Ralith, line 326: This, and `quinn_proto::Endpoint::config`, feel a little weird because nothing else is exposed, but I don't see any other reasonable way to make `quinn::Endpoint` aware of the configuration, and I don't particularly want to write a bunch of getters nobody is likely to ever use for the rest, so exposing as-needed seems like a decent compromise.

Note that I'm deliberately subverting naming idioms by prefixing the getter and leaving the setter with a concise name. I'm opting to do that because downstream code will overwhemlingly (exclusively?) use the setters, so more concision there makes sense.

djc: Maybe these comments should make it into an actual code comment? I generally dislike the `get_` prefix so would be somewhat inclined to revert it to its guideline state. I guess what we're saying here is that `impl EndpointConfig` is actually a kind of `EndpointConfigBuilder`, but we've folded the builder into the actual thing. Maybe unfolding them would make things slightly neater (at the expense of a little more duplication)?

Ralith: Added some brief discussion to the doc comment. I don't really follow your meaning about a builder; are you suggesting we duplicate the entire `EndpointConfig` at the quinn layer? I guess it's not that large, but the precedent worries me.

I'm not a fan of the `get_` prefix, but I imagine practically nobody will need it. I'm almost tempted to `#[doc(hidden)]` this.

djc: Should we have an `allocate_receive_buffer() -> Vec<u8>` method on `EndpointConfig` instead? I like the idea of applying `#[doc(hidden)]` either way.

Ralith: > Should we have an allocate_receive_buffer() -> Vec<u8> method on EndpointConfig instead?

One of the motivations here is to support more advanced memory management. For example, receiving many packets into a single buffer, or using specially allocated memory for e.g. io_uring shenanigans. I think these will be easier if the actual size is available directly.

PR #758, opened 2020-05-09
Refactor receive stream buffering

### LINKED ISSUE #735: read_unordered can issue redundant flow control credit
Stream assembly only deduplicates buffered data when the number of buffered frames exceeds a critical threshold. Because `read_unordered` pulls frames directly out of the buffer and does not keep track of previously-read data, receiving the same stream frame multiple times can lead to arbitrarily large increases in flow control limits. A malicious or buggy peer could leverage this to perform a memory-exhaustion DoS on an application that uses `read_unordered`.

One approach for a fix is to maintain a `RangeSet` defining which portions of the stream have not yet been read when using unordered reads, just as the assembler's offset defines that for ordered reads. Redundant data could then be discarded for both flow control and application processing purposes.

Ralith:
Fixes #735. Should marginally improve performance for ordered reads due to isolation of the (now somewhat more) complex `RangeSet` operations to the unordered case where they're necessary, leaving simple cursor-based logic for ordered behavior.

### DIFF
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -104,7 +104,12 @@ impl Assembler {
     fn defragment(&mut self) {
         let buffered = self.data.iter().map(|c| c.bytes.len()).sum::<usize>();
         let mut buffer = BytesMut::with_capacity(buffered);
-        let mut offset = self.offset;
+        let mut offset = self
+            .data
+            .peek()
+            .as_ref()
+            .expect("defragment is only called when data is buffered")
+            .offset;
 
         let new = BinaryHeap::with_capacity(self.data.len());
         let old = mem::replace(&mut self.data, new);
@@ -386,5 +391,13 @@ mod test {
         x.defragment();
         assert_eq!(x.pop(), Some((0, Bytes::from_static(b"abcdef"))));
         assert_eq!(x.pop(), Some((9, Bytes::from_static(b"jklmno"))));
+    }
+
+    #[test]
+    fn defrag_with_missing_prefix() {
+        let mut x = Assembler::new();
+        x.insert(3, Bytes::from_static(b"def"));
+        x.defragment();
+        assert_eq!(x.pop(), Some((3, Bytes::from_static(b"def"))));
     }
 }
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -25,11 +25,10 @@ impl Assembler {
         Self::default()
     }
 
-    pub(crate) fn read(&mut self, buf: &mut [u8]) -> usize {
-        assert!(
-            matches!(self.state, State::Ordered),
-            "cannot perform ordered reads following unordered reads on a stream"
-        );
+    pub(crate) fn read(&mut self, buf: &mut [u8]) -> Result<usize, IllegalOrderedRead> {
+        if let State::Unordered { .. } = self.state {
+            return Err(IllegalOrderedRead);
+        }
 
         let mut read = 0;
         loop {
@@ -42,7 +41,7 @@ impl Assembler {
                 break;
             }
         }
-        read
+        Ok(read)
     }
 
     pub(crate) fn read_unordered(&mut self) -> Option<(u64, Bytes)> {
@@ -138,7 +137,7 @@ impl Assembler {
     #[cfg(test)]
     fn next(&mut self, size: usize) -> Option<Box<[u8]>> {
         let mut buf = vec![0; size];
-        let read = self.read(&mut buf);
+        let read = self.read(&mut buf).unwrap();
         buf.resize(read, 0);
         if !buf.is_empty() {
             Some(buf.into())
@@ -261,6 +260,10 @@ impl Default for State {
         State::Ordered
     }
 }
+
+/// Error indicating that an ordered read was performed on a stream after an unordered read
+#[derive(Debug, Copy, Clone)]
+pub struct IllegalOrderedRead;
 
 #[cfg(test)]
 mod test {
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -17,7 +17,7 @@ pub(crate) struct Assembler {
     /// Whether to discard data
     stopped: bool,
     /// First offset we haven't received any data at or after
-    limit: u64,
+    end: u64,
 }
 
 impl Assembler {
@@ -153,7 +153,8 @@ impl Assembler {
     }
 
     pub(crate) fn insert(&mut self, mut offset: u64, mut bytes: Bytes) {
-        self.limit = self.limit.max(offset + bytes.len() as u64);
+        self.end = self.end.max(offset + bytes.len() as u64);
+
         if let State::Unordered { ref mut recvd } = self.state {
             // Discard duplicate data
             for duplicate in recvd.replace(offset..offset + bytes.len() as u64) {
@@ -190,13 +191,13 @@ impl Assembler {
     }
 
     /// Offset after the largest byte received
-    pub(crate) fn limit(&self) -> u64 {
-        self.limit
+    pub(crate) fn end(&self) -> u64 {
+        self.end
     }
 
-    /// Whether all data prior to `self.limit()` has been read
+    /// Whether all data prior to `self.end()` has been read
     pub(crate) fn is_fully_read(&self) -> bool {
-        self.bytes_read == self.limit
+        self.bytes_read == self.end
     }
 
     /// Discard all buffered data
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -2,14 +2,14 @@ use std::{cmp::Ordering, collections::BinaryHeap, mem};
 
 use bytes::{Buf, Bytes, BytesMut};
 
+use crate::range_set::RangeSet;
+
 /// Helper to assemble unordered stream frames into an ordered stream
 #[derive(Debug, Default)]
 pub(crate) struct Assembler {
+    state: State,
     data: BinaryHeap<Chunk>,
     defragmented: usize,
-    /// Whether any unordered reads have been performed, making this assembler unusable for ordered
-    /// reads
-    unordered: bool,
     /// Number of bytes read by the application. When only ordered reads have been used, this is the
     /// length of the contiguous prefix of the stream which has been consumed by the application,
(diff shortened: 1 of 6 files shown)


quinn-proto/src/connection/assembler.rs
```diff
@@ -164,16 +175,26 @@ impl Assembler {
...
-    /// Whether we've received all data below `final_offset`
-    ///
-    /// `final_offset` must be greater than the offset of any received data.
-    pub(crate) fn is_complete_at(&self, final_offset: u64) -> bool {
-        debug_assert!(final_offset >= self.limit());
-        (self.recvd.len() == 1 && self.recvd.iter().next().unwrap() == (0..final_offset))
-            || final_offset == 0
+    /// Whether all data prior to `self.limit()` has been read
+    pub(crate) fn is_fully_read(&self) -> bool {
+        if let State::Unordered { ref recvd } = self.state {
```
djc, line 197: Hmm, maybe write this as a `match` expression instead? Would improve the symmetry and obviate the fall-through case (which I think is not very obvious here).

quinn-proto/src/connection/assembler.rs
```diff
@@ -7,19 +7,14 @@ use crate::range_set::RangeSet;
...
-    /// Whether any unordered reads have been performed, making this assembler unusable for ordered
-    /// reads
-    unordered: bool,
     /// Number of bytes read. Equal to offset when `unordered` is false, may or may not be
     /// otherwise.
     bytes_read: u64,
-    /// Offsets received so far
-    recvd: RangeSet,
+    /// First offset we haven't received any data at or after
+    limit: u64,
```
djc, line 17: I feel like `limit` isn't a great name for this. Any other ideas? `end`, `next`, `unseen`? (Of these options I think I like `next` best.)

quinn-proto/src/connection/assembler.rs
```diff
@@ -64,23 +62,28 @@ impl Assembler {
             None => return false,
         };
 
+        let offset = match self.state {
+            State::Ordered { ref mut offset } => offset,
```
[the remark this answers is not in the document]
Ralith: It's the newer pattern, but I disagree that it's a better one, and I don't think there's a consensus either way (hence no clippy lint). I prefer this approach because the type of `offset` is determined by its immediate context, rather than the type of whatever was passed in to the `match`. This makes the type more obvious, especially for larger matches and matches on variables where it isn't immediately apparent whether they contain a reference.

quinn-proto/src/connection/assembler.rs
```diff
@@ -64,23 +62,28 @@ impl Assembler {
...
+            _ => panic!("cannot perform ordered reads following unordered reads on a stream"),
```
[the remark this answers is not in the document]
Ralith: I'd rate silent misbehavior (e.g. returning `None`) as worse than panicking, but I don't see any reason not to return an error. I'll add that tonight.

quinn-proto/src/connection/assembler.rs
Ralith: Yeah, that's definitely clearer.

quinn-proto/src/connection/assembler.rs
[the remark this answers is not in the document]
djc: Hmm, ok.

djc:
(Also very well-factored commit here. In terms of preventing unnecessary rebase efforts, maybe submit the first four commits separately so we can merge them immediately?)

quinn-proto/src/connection/assembler.rs
Ralith: I went with `end`, which had precedent elsewhere in related code. `next` could be confusing because it strictly has no bearing on what will actually be received next; data can be transmitted in any order so long as flow control limits are respected.

Ralith:
Cherry-picked the simple stuff into https://github.com/djc/quinn/pull/770.

quinn-proto/src/connection/assembler.rs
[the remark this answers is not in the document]
Ralith: Done.

Ralith:
Rebased to account for #776. Will need another after #777 is merged, or vis versa.

Ralith:
Rebased.

Ralith:
Got rid of a bunch of unnecessary complexity regarding `is_fully_read`, and found and fixed a significant bug in the process.

quinn-proto/src/connection/assembler.rs
```diff
@@ -104,7 +104,12 @@ impl Assembler {
     fn defragment(&mut self) {
         let buffered = self.data.iter().map(|c| c.bytes.len()).sum::<usize>();
         let mut buffer = BytesMut::with_capacity(buffered);
-        let mut offset = self.offset;
+        let mut offset = self
```
djc, line 109: Should we have a test for this?

quinn-proto/src/connection/assembler.rs
```diff
@@ -383,4 +421,35 @@ mod test {
...
+        assert_eq!(x.read_unordered(), None);
+    }
+
+    #[test]
+    fn unordered_dedup() {
+        let mut x = Assembler::new();
+        x.insert(3, Bytes::from_static(b"def"));
+        assert_eq!(x.read_unordered(), Some((3, Bytes::from_static(b"def"))));
+        assert_eq!(x.read_unordered(), None);
+        x.insert(0, Bytes::from_static(b"abcdefghi"));
```
djc, line 446: I think large chunks are better than small chunks. If we get a large chunk and we have a small chunk buffered which is a subset of the large chunk, should we keep the larger chunk instead of splitting it up?

quinn-proto/src/connection/assembler.rs
```diff
@@ -383,4 +421,35 @@ mod test {
...
+        x.insert(0, Bytes::from_static(b"abc"));
+        assert_eq!(x.read_unordered(), Some((0, Bytes::from_static(b"abc"))));
+        assert_eq!(x.read_unordered(), None);
+        x.insert(3, Bytes::from_static(b"def"));
+        assert_eq!(x.read_unordered(), Some((3, Bytes::from_static(b"def"))));
+        assert_eq!(x.read_unordered(), None);
+    }
+
+    #[test]
+    fn unordered_dedup() {
```
djc, line 441: I think this is testing "middle" and "tail" duplication but not "head" duplication. Should we have a test for that as well to make sure we cover all the edge cases here?

quinn-proto/src/range_set.rs
```diff
@@ -244,6 +271,43 @@ impl<'a> DoubleEndedIterator for EltIter<'a> {
     }
 }
 
+pub struct InsertDrain<'a> {
+    set: &'a mut RangeSet,
```
djc, line 275: Would be good to have documentation here as to what is the semantic meaning of each of these fields and also probably just the semantics of this type.

quinn-proto/src/range_set.rs
```diff
@@ -138,6 +138,33 @@ impl RangeSet {
         before || after
     }
 
+    /// Add a range to the set, returning the ranges it replaced
+    pub fn insert_drain(&mut self, mut range: Range<u64>) -> InsertDrain<'_> {
+        let pred = if let Some((prev_start, prev_end)) = self.pred(range.start) {
```
djc, line 143: Is there some way to prevent the 3-level cascading `else { None }` here? (Something with `Option::filter()`, I suppose.)

(Also, a little documentation on `succ()` and, in particular `pred()`, might help. I guess I haven't looked at this code in much detail before, but `pred` primarily invoked "predicate" for me rather than `predecessor`.)

Don't love this name. Would it make sense to just call this `insert()`? It looks like there's some duplication between `insert()` and `insert_drain()` currently, so if there's not too much overhead it might make sense to not maintain both.

quinn-proto/src/range_set.rs
```diff
@@ -138,6 +138,33 @@ impl RangeSet {
         before || after
     }
 
+    /// Add a range to the set, returning the ranges it replaced
```
djc, line 141:
```suggestion
    /// Add a range to the set, returning the intersection of current ranges with the new one
```

quinn-proto/src/range_set.rs
Ralith: > Is there some way to prevent the 3-level cascading else { None } here?

We could do this with combinators, but the main block of code here has side effects on the data structure, and I'm leery of hiding that. Still, was able to unnest by one level, which is something.

> documentation

:+1: 

> Don't love this name.

I agree the name isn't great. How's `replace`?

> Would it make sense to just call this insert()?

I'm reluctant because, at least syntatically, we go to a bunch of extra effort to produce an iterator. This did prompt me to reexamine and simplify `insert`, though, which makes keeping it that much more pleasant.

quinn-proto/src/range_set.rs
Ralith: Good idea.

quinn-proto/src/connection/assembler.rs
Ralith: That would require random-accessing into the `BinaryHeap` somehow to replace the shorter pre-existing chunk, or deferring the deduplication until `read_unordered`-time, which increases memory use. Given that `defragment` will keep a lid on fragmentation overall, and what I estimate to be a quite low incidence of cases where perfect efficiency here would help much, I'm inclined to stick with the easy thing.

quinn-proto/src/connection/assembler.rs
Ralith: Added a few lines to exercise that too.

quinn-proto/src/connection/assembler.rs
Ralith: Thinking on this further, I suspect unordered mode should have a `BTreeMap<u64, Bytes>` rather than the `RangeSet` + `BinaryHeap` arrangement. Then we could optimize for efficient memory layout just as easily as we currently preserve `RangeSet`'s invariants. Removing the first item from a `BTreeMap` to read is awkward, but when we're traversing it multiple times per packet anyway...

I'd rather not block this PR on it, though, since this PR is still a direct improvement on the status quo.

quinn-proto/src/range_set.rs
djc: Much better!

quinn-proto/src/connection/assembler.rs
djc: Sounds good.

djc:
Nice work.

Issue #693, opened 2020-03-23
BBR congestion control


Demi-Marie:
Google uses it in production, and it should massively increase throughput on bad networks, while simultaneously fighting bufferbloat.

Ralith:
It'd be cool to support advanced congestion control algorithms. However, congestion control is an active research area, and even BBR is subject to ongoing work; last time I reviewed it there was no remotely complete documentation for the current state of the art. Therefore I'd like to approach this by making congestion control pluggable, probably through a trait object. Some interface design will be necessary, including a review of the field to see what information leading congestion control algorithms require.

djc:
@Ralith if you can outline where the current approach to congestion control is happening I'd like to take a whack at extracting what we have now into separate state/abstraction.

Ralith:
Awesome! `Connection::on_packet_acked` and `Connection::congestion_event` are the heart of the current implementation. See also [the reference pseudocode in the draft](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#name-congestion-control-pseudoco). Communication with the rest of the state machine is through these functions being called, resulting in updates to `self.path.congestion_window`, which is also subject to being reset during migration (but not rebinding).

Abstracting the current implementation out sounds like a good start, and will likely be beneficial in terms of shrinking `connection.rs` regardless. However, I think a significantly richer interface, including e.g. timing information, will be necessary for supporting algorithms like BBR. It makes sense to pursue that as a second pass.

Ralith:
Other algorithms to explore and hopefully support: CUBIC, HyStart

Ralith:
Relevant blog post from quiche: https://blog.cloudflare.com/cubic-and-hystart-support-in-quiche/amp/

We should have a look at their approach to abstracting congestion control.

PR #759, opened 2020-05-10
Pluggable congestion control

Ralith:
As discussed in #693. ~Draft until the abstractness of the interface has been validated by the addition of at least one other controller (probably CUBIC/HyStart++).~

### DIFF
--- a/quinn-proto/src/congestion.rs
+++ b/quinn-proto/src/congestion.rs
@@ -1,0 +1,34 @@
+//! Logic for controlling the rate at which data is sent
+
+use std::time::Instant;
+
+mod new_reno;
+pub use new_reno::{NewReno, NewRenoConfig};
+
+/// Common interface for different congestion controllers
+pub trait Controller: Send {
+    /// Packet deliveries were confirmed
+    ///
+    /// `congestion_blocked` indicates whether the connection was blocked on congestion prior to
+    /// receiving these acknowledgements.
+    fn on_ack(&mut self, now: Instant, sent: Instant, bytes: u64, congestion_blocked: bool);
+
+    /// Packets were deemed lost or marked congested
+    ///
+    /// `in_persistent_congestion` indicates whether all packets sent within the persistent
+    /// congestion threshold period ending when the most recent packet in this batch was sent were
+    /// lost.
+    fn on_congestion_event(&mut self, now: Instant, sent: Instant, is_persistent_congestion: bool);
+
+    /// Number of ack-eliciting bytes that may be in flight
+    fn window(&self) -> u64;
+
+    /// Duplicate the controller's state
+    fn clone_box(&self) -> Box<dyn Controller>;
+}
+
+/// Constructs controllers on demand
+pub trait ControllerFactory {
+    /// Construct a fresh `Controller`
+    fn build(&self, now: Instant) -> Box<dyn Controller>;
+}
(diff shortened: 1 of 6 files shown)


djc:
I think I'm okay with the config `Arc` now. Can you sketch what your explicit trait alternative for the factory thing would look like? I can't quite wrap my head around what the constraints are.

(Otherwise, things look pretty good to me, though I'll do a final pass once we figure out the factory thing.)

Ralith:
The explicit trait approach:

```rust
pub trait CongestionControllerFactory {
    fn new_controller(&self) -> Box<dyn CongestionController>;
}
// ...
impl CongestionControllerFactory for Arc<NewRenoConfig> {
    fn new_controller(&self) -> Box<dyn CongestionController> {
        Box::new(NewReno::new(self.clone(), Instant::now()))
    }
}
```
In use:
```rust
config.congestion_controller_factory(Arc::new(config));
```

djc:
How would the `maybe_rebinding` case look for that? I think it'd need a separate impl? How do you feel about current approach vs explicit trait?

Ralith:
`(self.config.congestion_controller_factory)()` would be replaced with `self.config.congestion_controller_factory.new_controller()`; no other changes. A trait with one method that takes `&self` is always algebraically equivalent to an analogous `Fn`.

Between the two, I'm leaning towards the explicit trait right now, because `config.congestion_controller_factory(Arc::new(config));` seems like considerably less ceremony than `config.congestion_controller_factory(|| Box::new(NewReno::new(config, Instant::now())))`. It's more boilerplate if you want to invoke complicated custom logic, but that seems like an *extremely* niche case.

djc:
Cool, I think I agree.

Ralith:
Updated with an explicit factory trait. I also refactored things to make the `congestion` module public, to ease discoverability of different algorithms we add in the future, and limit growth of clutter in the root namespace.

quinn-proto/src/congestion.rs
```diff
@@ -0,0 +1,27 @@
...
+
+use std::time::Instant;
+
+mod new_reno;
+pub use new_reno::{NewReno, NewRenoConfig};
+
+/// Common interface for different congestion controllers
+pub trait Controller: Send {
+    /// Packet deliveries were confirmed
+    fn on_ack(&mut self, sent: Instant, bytes: u64, congestion_blocked: bool);
```
djc, line 11: Hmm, should we pre-emptively add a `now` argument here? That seems like something that might definitely be needed at least conceptually (also nice symmetry with `on_congestion_event`).

quinn-proto/src/congestion.rs
```diff
@@ -0,0 +1,27 @@
...
+
+    /// Packets were deemed lost or marked congested
+    fn on_congestion_event(&mut self, now: Instant, sent: Instant, persistent: bool);
```
djc, line 14: The meaning of `persistent` is not obvious to me. Different name? Some comments? Switch from `bool` to `enum`?

quinn-proto/src/congestion.rs
Ralith: It's needed by BBR, at least, so I suppose so.

quinn-proto/src/congestion.rs
Ralith: Elaborated a bit and expanded the name to the full term used in the recovery draft.

PR #761, opened 2020-05-10
Endpoint client config reuse

Ralith:
Helps avoid apparently expensive `ClientConfig::default()` calls.

### DIFF
--- a/quinn/src/builders.rs
+++ b/quinn/src/builders.rs
@@ -21,7 +21,7 @@ where
 {
     server_config: Option<ServerConfig<S>>,
     config: EndpointConfig<S>,
-    client_config: ClientConfig<S>,
+    default_client_config: ClientConfig<S>,
 }
 
 #[allow(missing_docs)]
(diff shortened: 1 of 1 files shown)


quinn/src/builders.rs
```diff
@@ -30,12 +30,11 @@ where
     S: proto::crypto::Session + Send + 'static,
 {
     /// Start a builder with a specific initial low-level configuration.
-    pub fn new(config: EndpointConfig<S>) -> Self {
+    pub fn new(config: EndpointConfig<S>, default_client_config: ClientConfig<S>) -> Self {
```
djc, line 33: I don't like this part that much. Could we make the field `None` initially? I suppose then we have to pay for per-`Endpoint` construction of the `ClientConfig` at `build()` time if the user doesn't set one explicitly.

Ralith: I feel like this is consistent with the existing use of `new`, which is a niche alternative to `default()` when you already have an `EndpointConfig` for some reason. Sharing instances of the expensive-to-construct default `ClientConfig` between cloned builders seems desirable as well.

djc: Fair enough!

PR #762, opened 2020-05-10
Mention matrix room in README

Ralith:
Matrix is an open, federated platform that Mozilla recently committed to, with a growing Rust presence. The room in question is bridged to gitter to avoid fragmentation, and provides a richer, more provider-independent alternative infrastructure and adds a number of client options.


Ralith:
Strange error from coverage:
> ##[error]Cannot read property 'find' of undefined

PR #767, opened 2020-05-12
Omit trailing lengths

### LINKED ISSUE #720: Omit length field in trailing stream/datagram frames
We currently include a length field unconditionally, which wastes a few bytes when it's implied by the size of the enclosing packet. Blocked on the refactoring in #699.

Ralith:
Fixes #720.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -2482,19 +2482,37 @@ where
                 Some(x) => x,
                 None => break,
             };
-            if buf.len() + datagram.size(true) > max_size {
+            // This is the last frame, and can therefore omit the length tag, if there are no
+            // following datagrams that would fit, no following stream frames, and no padding will
+            // be needed.
+            let is_last_frame = self
+                .datagrams
+                .outgoing
+                .front()
+                .map_or(true, |next_datagram| {
+                    buf.len() + datagram.size(true) + next_datagram.size(false) > max_size
+                })
+                && !self.streams.has_stream_frames()
+                && buf.len() + datagram.size(false) >= min_size;
+            if buf.len() + datagram.size(!is_last_frame) > max_size {
                 // Future work: we could be more clever about cramming small datagrams into
                 // mostly-full packets when a larger one is queued first
                 self.datagrams.outgoing.push_front(datagram);
                 break;
             }
             self.datagrams.outgoing_total -= datagram.data.len();
-            datagram.encode(true, buf);
+            datagram.encode(!is_last_frame, buf);
+            if is_last_frame {
+                // There might be another datagram that's small enough to fit iff we omit the length
+                // tag, but without a length tag it's illegal to add any further frames, so we must
+                // explicitly ignore any remaining datagrams by breaking early here.
+                break;
+            }
         }
 
         // STREAM
         if space_id == SpaceId::Data {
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -2363,7 +2363,7 @@ where
     fn populate_packet(
         &mut self,
         space_id: SpaceId,
-        minimum_length: usize,
+        min_size: usize,
```
djc, line 2366: Can we move this part into the previous commit?

Ralith: Done.

quinn-proto/src/connection/streams.rs
```diff
@@ -523,8 +534,12 @@ impl Streams {
             }
 
             let meta = frame::StreamMeta { id, offsets, fin };
-            trace!(id = %meta.id, off = meta.offsets.start, len = meta.offsets.end - meta.offsets.start, fin = meta.fin, "STREAM");
-            meta.encode(true, buf);
+            let len = (meta.offsets.end - meta.offsets.start) as usize;
+            trace!(id = %meta.id, off = meta.offsets.start, len, fin = meta.fin, "STREAM");
+            // This is the last frame, and can therefore omit the length tag, if there are no
+            // following stream frames and no padding will be needed.
+            let is_last_frame = self.pending.is_empty() && buf.len() + len >= min_buf_size;
```
djc, line 541: Is this intentionally unnecessarily restrictive? For example, if `self.pending.is_empty()` is `false` but we're filling up to within `frame::Stream::SIZE_BOUND` of the `max_buf_size`, we could also say `is_last_frame`.

Ralith: No, good catch. We should have a similar fix for the datagram case as well.

Ralith: Applied.

Ralith:
I feel like this probably needs tests as well; let's not merge until I have those written.

djc:
Yeah. I do wonder, why were you motivated to implement this? With UDP datagram sizes ~always below 2 ** 14, doesn't this gain us at most 2 bytes per packet?

Ralith:
The motivation struck me when someone commented in the chat that they wanted it, but I agree the complexity seems a bit high. I think the largest benefit is when an endpoint is sending many very small packets simultaneously, e.g. a game or VoIP server broadcasting a small amount of unreliable data to all connected clients. Though even two bytes per packet * 100 clients still isn't very much...

I'm open to the idea that the complexity/compute ultimately isn't justifiable here, especially in lieu of concrete evidence of a use case that substantially benefits.

djc:
In our chat? Might be worth digging into their use case more.

I see the main benefit being if you just have fill up packets with one stream frame. Maybe we could simplify the code for that case only so there's less complexity/run-time costs?

(Even if we don't take the last commit, I think we should probably merge the ones before.)

djc:
Making it explicit: as I understand it the ball is in your court to just close this or to see how far it can be simplified to have good complexity cost/bandwidth saved benefit.

Ralith:
Yep, bearing in mind that I don't plan to focus on this in the near future since we have bigger issues.

Ralith:
I think #998 addressed the worst of this, so closing this as no longer necessary.

PR #775, opened 2020-05-16
H3 ignore unknown frames and uni streams

stammw:
Addresses the two failing tests on [quic tracker](https://quic-tracker.info.ucl.ac.be/grid).

### DIFF
--- a/quinn-h3/src/tests/helpers.rs
+++ b/quinn-h3/src/tests/helpers.rs
@@ -47,6 +47,8 @@ pub struct Helper {
 
 impl Helper {
     pub fn new() -> Self {
+        let _ = tracing_subscriber::fmt().try_init();
+
         let port = PORT_COUNT.fetch_add(1, Ordering::SeqCst);
 
         let Certs { chain, key, cert } = CERTS.clone();
--- a/quinn-h3/src/tests/helpers.rs
+++ b/quinn-h3/src/tests/helpers.rs
@@ -6,7 +6,6 @@ use std::{
 };
 
 use anyhow::{bail, Context, Result};
-use bytes::BufMut;
 use futures::stream::StreamExt;
 use http::{request, Request};
 use quinn::{Certificate, CertificateChain, PrivateKey, SendStream, WriteError};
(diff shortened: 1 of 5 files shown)


quinn-h3/src/tests/helpers.rs
```diff
@@ -47,6 +48,15 @@ pub struct Helper {
 
 impl Helper {
     pub fn new() -> Self {
+        if !TRACING_SUBSCRIBED.fetch_or(true, Ordering::Acquire) {
```
djc, line 51: @Ralith can you review this commit?

quinn-h3/src/tests/helpers.rs
```diff
@@ -1,12 +1,11 @@
 use std::{
     fs, io,
     net::{IpAddr, Ipv6Addr, SocketAddr},
-    sync::atomic::{AtomicU16, Ordering, AtomicBool},
+    sync::atomic::{AtomicBool, AtomicU16, Ordering},
```
djc, line 4: This belongs in the previous commit. (Tip: `git rebase -x "cargo fmt -- --check"`.)

quinn-h3/src/tests/helpers.rs
Ralith, line 51: A simpler approach would be to call [`tracing_subscriber::fmt::try_init`](https://docs.rs/tracing-subscriber/0.2.5/tracing_subscriber/fmt/fn.try_init.html) and ignore its return value.

quinn-h3/src/tests/helpers.rs
stammw: Awesome!

quinn-h3/src/tests/helpers.rs
stammw: I was feeling this should be a thing, there it is. Thanks!

stammw:
Got this deployed on the interop server. Let's wait to see if the tests go green before merging.

Ralith:
Looks like the interop tracker had some sort of issue reaching our interop server over IPv4; strange. I don't see anything suspicious in the logs, save for a reboot this morning--maybe just really bad timing?

stammw:
I was able to reproduce running quic-tracker on my machine. I'll investigate tomorrow, I've been busy lately.

djc:
So can/should we merge this now?

Ralith:
I'd still like to know why we're apparently dropping quic-tracker's packets on the floor all of a sudden.

PR #776, opened 2020-05-20
Flow control debugging

### LINKED ISSUE #774: Flow control error encountered when program is under heavy mixed load
I'm going to be honest, the title is probably not 100% accurate as I'm not entirely sure yet what's going on. This could be a Tokio or a Quinn bug. All I can say for sure is that I have developed a reproducer that makes it happen _some_ of the time. 

[The reproducer is here](https://github.com/SriRamanujam/quinn_minimal_network_repro/tree/transport_error). 

My exact steps to run the reproducer are:

```sh
$ cargo build --release
$ RUST_LOG=quinn=trace target/release/server
(in another terminal) 
$ RUST_LOG=quinn=trace target/release/client
```

Most of the time, the client and server negotiate flow control properly and back off. If that happens, ctrl-c both processes and try again. Eventually you will hit a run where the flow control error happens. In my local testing, if the error happened, it always happened within a minute or two. 

I've also attached trace logs for server and client from a failed run, in case it helps.
[client_log.txt](https://github.com/djc/quinn/files/4630254/client_log.txt)
[server_log.txt](https://github.com/djc/quinn/files/4630257/server_log.txt)

Ralith:
Fixes #774.

### DIFF
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -144,6 +144,9 @@ impl Assembler {
 
     pub(crate) fn insert(&mut self, offset: u64, bytes: Bytes) {
         self.recvd.insert(offset..(offset + bytes.len() as u64));
+        if bytes.is_empty() {
+            return;
+        }
         self.data.push(Chunk { offset, bytes });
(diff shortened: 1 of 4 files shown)


djc:
Great work.

PR #778, opened 2020-05-21
Return UnknownStream from Connection::reset for closed/reset streams

Ralith:
Improves consistency with other stream interfaces. H3 logic was
updated to ignore errors to preserve existing behavior, but future
work might want to insert some unwraps there if static correctness
guarantees are available.


djc:
@stammw just to be sure: note that your review is requested here, as well. :)

Ralith:
Resolved merge conflict, just need a quick sanity-check for the h3 changes.

PR #793, opened 2020-06-04
Fix some warnings and a build failure

djc:


quinn-h3/src/connection.rs
```diff
@@ -267,15 +267,9 @@ impl ConnectionInner {
...
-                let mut pending = x.take().unwrap();
-                match Pin::new(&mut pending).poll(cx) {
-                    Poll::Ready(y) => Some((i, y)),
-                    Poll::Pending => {
-                        std::mem::replace(x, Some(pending));
-                        None
-                    }
-                }
+            .filter_map(|(i, x)| match Pin::new(x.as_mut().unwrap()).poll(cx) {
+                Poll::Ready(y) => Some((i, y)),
```
Ralith, line 271: The previous code left `x` as `None` on `Poll::Ready`. Should this be preserved? If not, should `self.pending_uni` still contain `Option`? Not familiar with this code in general, but that seems suspect.

djc: Oops, you're right. Fixed that up.

Ralith:
Lint issue is unrelated; opened https://github.com/djc/quinn/pull/794 to resolve it.

djc:
I've fixed the new lint. Take a look?

Issue #785, opened 2020-05-28
H3: Update to draft-28


stammw:
- [ ] Draft version in `Alt-Svc` could be the cause of incompatibility of some clients. The draft shows something like `Alt-Svc: h3=":50781"`
- [ ] Check authority based on CN field of the certificate subject or  dNSName, iPAddress in the subjectAltName field. [#3558](https://github.com/quicwg/base-drafts/pull/3558/files)
- [ ] GoAway from client: [#3129](https://github.com/quicwg/base-drafts/pull/3129/files)
- [ ] Make it mandatory for requests to have either `:authority` or `Host` fields [#3475](https://github.com/quicwg/base-drafts/pull/3475/files)

PR #795, opened 2020-06-05
H3: check :authority and Host headers

stammw:
This is part of draft-28 update for #785. Adressing draft's [3558](https://github.com/quicwg/base-drafts/pull/3558/files).

Authoritative header values still have to be checked against certificates values. Do you have some recommendations before I start to look for a solution on my own ?  Here is the extract I'll address:

> Once a connection exists to a server endpoint, this connection MAY be reused for requests with multiple different URI authority components. In general, a server is considered authoritative for all URIs with the "https" scheme for which the hostname in the URI is present in the authenticated certificate provided by the server, either as the CN field of the certificate subject or as a dNSName in the subjectAltName field of the certificate; see [RFC6125]. For a host that is an IP address, the client MUST verify that the address appears as an iPAddress in the subjectAltName field of the certificate. If the hostname or address is not present in the certificate, the client MUST NOT consider the server authoritative for origins containing that hostname or address. See Section 5.4 of [SEMANTICS] for more detail on authoritative access.

### DIFF
--- a/quinn-h3/src/client.rs
+++ b/quinn-h3/src/client.rs
@@ -76,7 +76,7 @@ use std::{
 };
 
 use futures::{channel::oneshot, ready, FutureExt};
-use http::{request, Request, Response};
+use http::{request, HeaderMap, Method, Request, Response, Uri};
 use http_body::Body as HttpBody;
 use pin_project::{pin_project, project};
 use quinn::{Certificate, Endpoint, OpenBi, RecvStream};
@@ -540,6 +540,31 @@ where
     }
 }
 
+#[project]
+impl<B> SendRequest<B, B::Data>
+where
+    B: HttpBody + 'static,
+    B::Error: Into<Box<dyn StdError + Send + Sync>> + Send + Sync,
+{
+    fn take(&mut self) -> Result<(Method, Uri, HeaderMap, B), Error> {
+        let (parts, body) = self.request.take().expect("no request").into_parts();
+        let request::Parts {
+            method,
+            uri,
+            headers,
+            ..
+        } = parts;
+
+        match (uri.authority(), headers.get("host")) {
+            (None, None) => Err(Error::Header("Missing authority")),
+            (Some(a), Some(h)) if a.as_str() != h => {
+                Err(Error::Header("Host and :authority are in contradiction"))
+            }
+            _ => Ok((method, uri, headers, body)),
+        }
+    }
+}
+
 impl<B> Future for SendRequest<B, B::Data>
 where
     B: HttpBody + 'static,
@@ -560,13 +585,7 @@ where
                             return Poll::Ready(Err(e.into()));
                         }
                         Ok((send, recv)) => {
-                            let (parts, body) = me.request.take().unwrap().into_parts();
-                            let request::Parts {
-                                method,
-                                uri,
-                                headers,
-                                ..
-                            } = parts;
+                            let (method, uri, headers, body) = me.take()?;
 
                             if recv.is_0rtt() && !method.is_idempotent() {
                                 let err = Error::internal("non-idempotent method tried on 0RTT");
--- a/quinn-h3/src/proto/headers.rs
+++ b/quinn-h3/src/proto/headers.rs
@@ -55,8 +55,12 @@ impl Header {
             uri = uri.scheme(scheme.as_str().as_bytes());
         }
 
-        if let Some(authority) = self.pseudo.authority {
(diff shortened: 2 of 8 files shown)


quinn-h3/src/client.rs
```diff
@@ -540,6 +544,33 @@ where
...
+            headers,
+            ..
+        } = parts;
+
+        match (uri.authority(), headers.get("host")) {
+            (None, None) => return Err(Error::Header("Missing authority")),
+            (Some(a), Some(h)) if a.as_str() != h => {
+                return Err(Error::Header("Host and :authority are in contradiction"));
+            }
+            _ => (),
```
djc, line 563: You can elide the explicit returns by just having the `Ok((...))` in the `_` case.

quinn-h3/src/client.rs
```diff
@@ -540,6 +544,33 @@ where
     }
 }
 
+#[project]
+impl<B> SendRequest<B, B::Data>
+where
+    B: HttpBody + 'static,
+    B::Error: Into<Box<dyn StdError + Send + Sync>> + Send + Sync,
+{
+    fn take_request(&mut self) -> Result<(Method, Uri, HeaderMap, B), Error> {
```
djc, line 549: Maybe just name this `take()`? The `_request` suffix feels stuttery on a type that's called `SendRequest`.

stammw: Oh yeah! I use to rant on this in my dayjob review!

quinn-h3/src/proto/headers.rs
```diff
@@ -355,3 +362,63 @@ impl Error {
         ))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    impl Header {
+        pub(crate) fn authory_mut(&mut self) -> &mut Option<Authority> {
```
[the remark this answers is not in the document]
stammw: That's better, thanks.

quinn-h3/src/client.rs
stammw: Much better (valid) style indeed!

quinn-h3/src/proto/headers.rs
```diff
@@ -151,6 +155,7 @@ impl TryFrom<Vec<HeaderField>> for Header {
                     pseudo.len += 1;
                 }
                 Field::Authority(a) => {
+                    dbg!(&a);
```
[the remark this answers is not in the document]
stammw: -_-

stammw:
I'm speaking for both sides I guess. This means our current implementation is not secure on authority enforcement then. I'm good with that, I'll open an issue on this.

djc:
Hmm, how is it not secure? I got the impression it is stricter than necessary, but that doesn't mean it's insecure.

stammw:
I think a client should check if the served resource is from a server that has valid certificate for the authority it asks for ? Maybe that's already done through `server_name` in rustls ?

I'm not very informed on all this right now, maybe it's secure enough.

djc:
Yeah, rustls will check that the server name passed in by the client matches one of the subject alternative names from the server's certificate.

stammw:
Would the `rustls` dev you're mentioning look like `conn.quic.authentication_data().has_authority(authority_header)` in `h3` code ?

With `authority_header` being a server_name or an IP, and `has_authority()` iterating over server's issued certificates, in their `CN`, `subjectAltName.dNSNamne`, or `subjectAltName.IPAdress` until some one is matched ?

I opened #796 so I guess we can merge this right away.

djc:
So far, the concept of "authentication data" only exists in Quinn, not in rustls itself. We would need a way to check that a particular authority is valid in the context of a particular server certificate, I suppose. That would be some kind of interplay between rustls and, potentially, webpki.

Ralith:
Alternatively, just fetch the certificate from rustls, extract the alt names, and handle the authority checks fully inside quinn-h3?

djc:
I'm afraid correctly fetching and matching these names is highly security sensitive, so it would be nice if we can benefit from Brian's/Joe's expertise and maintenance on this stuff. (Saying this after having observed some of the audit findings this week - nothing serious, they were in fact very positive about the rustls stack, but still.)

Ralith:
I mean call out directly to [the relevant part of webpki](https://briansmith.org/rustdoc/webpki/struct.EndEntityCert.html#method.verify_is_valid_for_dns_name). We shouldn't need any changes in either library to accomplish this (missing support for IP addresses aside).

djc:
Oh yeah, that API looks good!

PR #802, opened 2020-06-13
H3: ignore any number of unknown settings

stammw:
Addressing a bug that emerged from interop test against gQuic.

### DIFF
--- a/quinn-h3/src/proto/settings.rs
+++ b/quinn-h3/src/proto/settings.rs
@@ -19,6 +19,15 @@ macro_rules! setting_identifiers {
 
 impl SettingId {
     const NONE: SettingId = SettingId(0);
+
+    fn is_supported(self) -> bool {
+        match self {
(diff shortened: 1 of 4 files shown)


quinn-h3/src/proto/settings.rs
```diff
@@ -19,6 +19,15 @@ macro_rules! setting_identifiers {
 
 impl SettingId {
     const NONE: SettingId = SettingId(0);
+
+    fn is_supported(id: SettingId) -> bool {
```
Ralith, line 23:
```suggestion
    fn is_supported(self) -> bool {
```

This would be a bit more convenient to call.

Ralith:
CI failures look unrelated, due to a deprecation in an upstream crate.

stammw:
Yeah, there is a breaking change in pin_project... I'm on it, I'll get this right soon.

stammw:
It should be fine now.

PR #809, opened 2020-06-24
Expose some inner workings of quinn-proto in quinn

SoftwareSheriff:
Specifically, this PR enables the use of `StreamId`s from `SendStream`/`RecvStream`, implements `Hash`, `PartialEq`, and `Eq` for `Connection`, and adds a way to get a static ID for a `Connection` since IP addresses may change during the lifetime of a QUIC connection.

I'm not sure `Connection::index()` is the greatest name for a function that does that last part, but it's consistent with how `StreamId`s work. It might be more discoverable if it were named something like `quic_address()`, but I'll leave it to the maintainers to make that decision.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -409,6 +409,14 @@ where
             .peer_identity()
     }
 
+    /// A stable identifier for this connection
+    ///
+    /// Peer addresses and connection IDs can change, but this value will remain
+    /// fixed for the lifetime of the connection.
+    pub fn stable_id(&self) -> usize {
+        self.0.stable_id()
+    }
+
     // Update traffic keys spontaneously for testing purposes.
     #[doc(hidden)]
     pub fn force_key_update(&self) {
@@ -621,6 +629,10 @@ where
             error: None,
             ref_count: 0,
         })))
+    }
+
+    fn stable_id(&self) -> usize {
+        &*self.0 as *const _ as usize
     }
 }
 
--- a/quinn/src/streams.rs
+++ b/quinn/src/streams.rs
@@ -173,7 +173,7 @@ where
         }
     }
 
-    #[doc(hidden)]
+    /// Get the identity of this stream
     pub fn id(&self) -> StreamId {
         self.stream
     }
@@ -459,6 +459,11 @@ where
     /// level. Because read data is subject to replay attacks.
     pub fn is_0rtt(&self) -> bool {
         self.is_0rtt
+    }
+
+    /// Get the identity of this stream
+    pub fn id(&self) -> StreamId {
+        self.stream
     }
 }
 
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -60,7 +60,7 @@ mod udp;
 
 pub use proto::{
     crypto, ApplicationClose, Certificate, CertificateChain, ConnectError, ConnectionClose,
-    ConnectionError, ParseError, PrivateKey, Transmit, TransportConfig, VarInt,
+    ConnectionError, ParseError, PrivateKey, StreamId, Transmit, TransportConfig, VarInt,
 };
 
 pub use crate::builders::EndpointError;


quinn/src/connection.rs
```diff
@@ -409,6 +410,14 @@ where
             .peer_identity()
     }
 
+    /// The peer's QUIC address
```
Ralith, line 413: I think "QUIC address" is a very misleading term for this. All it is is a stable ID, and it should be documented as such.

quinn/src/streams.rs
```diff
@@ -173,7 +173,7 @@ where
         }
     }
 
-    #[doc(hidden)]
+    /// Get the identity of this stream
     pub fn id(&self) -> StreamId {
```
Ralith, line 177: `StreamId` needs to be reexported.

quinn/src/connection.rs
```diff
@@ -622,6 +654,10 @@ where
             ref_count: 0,
         })))
     }
+
+    fn index(&self) -> usize {
+        self.0.lock().unwrap().index()
+    }
```
Ralith, line 660: Given that this requires taking a lock, perhaps we should consider just returning the address contained in the `Arc`. Can you think of any reason why we specifically need the quinn-proto level ID here?

quinn/src/connection.rs
SoftwareSheriff: Good point; how does "A static identifier for this connection" sound instead?

quinn/src/streams.rs
SoftwareSheriff: As `quinn::StreamId` or somewhere else?

Ralith: Just like that.

quinn/src/connection.rs
Ralith: I prefer "stable" to "static", personally.

quinn/src/connection.rs
SoftwareSheriff: Yeah, I don't think it needs to specifically be the quinn-proto ID, just something that's static. `&*arc as *const _ as usize` would good enough for me.

Ralith: Let's favor the simpler/faster approach, then.

Ralith: Although `index` is a bit silly if it's not even remotely sequential/dense... maybe rename to `stable_id`?

SoftwareSheriff: Yeah, I like this idea

quinn/src/connection.rs
```diff
@@ -409,6 +410,14 @@ where
             .peer_identity()
     }
 
+    /// A stable identifier for this peer
```
Ralith, line 413:
```suggestion
    /// A stable identifier for this connection
```

It's entirely possible to have multiple connections to a single peer.

quinn/src/connection.rs
```diff
@@ -409,6 +410,14 @@ where
...
+    ///
+    /// Since clients may change IP addresses, this method serves as a way to
+    /// get a stable identifier for a particular connection.
```
Ralith, line 416:
```suggestion
    /// Peer addresses and connection IDs can change, but this value will
    /// remain fixed for the lifetime of the connection.
```

quinn/src/connection.rs
SoftwareSheriff: Ah, good point

Benjamin-L:
I think we should either have `Hash`, `Eq`, and `PartialEq`, or just have `stable_id()` and none of those traits. I think the reasoning to get rid of `Hash` also applies to `Eq` and `PartialEq`.

I would also prefer `id` instead of `stable_id`.

SoftwareSheriff:
In agreement with the `stable_id` -> `id` suggestion, I've made the relevant changes.

After some consideration, I'm not sure I personally need `Hash` (or even `PartialEq`/`Eq`), but on the other hand, are there any reasons to omit that stuff? "If you don't need it then we may as well omit it" is a valid reason, too, but I thought I'd ask.

djc:
I'm firmly in the camp that the burden is on the side of including more stuff, not on the side of omitting stuff. Conceptually I don't feel implementing `Hash` for `Connection` makes much sense. `PartialEq` makes more sense to me, but I feel like having `id()` ought to be sufficient for most of the relevant use cases.

djc:
Might wait until we get feedback from @Ralith before you make further changes, but yeah.

Ralith:
Omitting the trait impls in lieu of a compelling use case makes sense to me.

I pretty strongly prefer `stable_id` to `id` for the reason discussed in chat: connection IDs are a specific thing in QUIC and quinn-proto, and the value we return here has nothing to do with them. This risks confusion, especially given that we cite them directly in the doc comment. Naming it `stable_id` emphasizes that it's a distinct notion.

djc:
Fair point! Let's keep it at `stable_id()` and ditch the `Hash`/`PartialEq`/`Eq` impls.

SoftwareSheriff:
Alrighty, the relevant commit has been dropped and `id` has been renamed back to `stable_id`. I also reworded the commit's long description to more accurately reflect the actual implementation details and design decisions.

djc:
Awesome, thanks!

PR #812, opened 2020-06-29
Update to draft 29

### LINKED ISSUE #801: Update to draft 29

Ralith:
Fairly minor changes this time around. The refactored PTO logic is the only significant regression hazard.

Fixes #801.


quinn-proto/src/connection/mod.rs
```diff
@@ -1219,28 +1258,31 @@ where
...
+    fn set_loss_detection_timer(&mut self, now: Instant) {
+        if let Some((loss_time, _)) = self.loss_time_and_space() {
             // Time threshold loss detection.
             self.timers.set(Timer::LossDetection, loss_time);
             return;
         }
 
-        // Don't arm timer if there are no ack-eliciting packets
-        // in flight and the handshake is complete.
+        if self.anti_amplification_blocked() {
```
djc, line 1268: Shouldn't this come before the preceding block?

quinn-proto/src/connection/mod.rs
```diff
@@ -1195,18 +1198,54 @@ where
...
     }
 
+    fn pto_time_and_space(&self, now: Instant) -> Option<(Instant, SpaceId)> {
+        let srtt = self.path.rtt.smoothed.unwrap_or(self.config.initial_rtt);
+        let backoff = 2u32.pow(self.pto_count.min(MAX_BACKOFF_EXPONENT));
+        let mut duration = (srtt + TIMER_GRANULARITY.max(4 * self.path.rtt.var)) * backoff;
+
+        if self.in_flight.is_empty() {
+            debug_assert!(!self.peer_completed_address_validation());
+            let space = if self.highest_space == SpaceId::Handshake {
```
djc, line 1214: Use match here to spend one less line on it?

quinn-proto/src/connection/mod.rs
```diff
@@ -1195,18 +1198,54 @@ where
...
+            return Some((now + duration, space));
+        }
+
+        let mut result = None;
+        for space in SpaceId::iter() {
+            if self.spaces[space].in_flight == 0 {
+                continue;
+            }
+            if space == SpaceId::Data {
+                // Skip ApplicationData until handshake complete.
```
djc, line 1228:
```suggestion
                // Skip ApplicationData until handshake completes.
```

djc:
Thanks for doing the work on this!

quinn-proto/src/connection/mod.rs
Ralith: We want to detect packet loss in a timely fashion even if we can't retransmit. Lost packet detection only happens on LossDetection timeout or when we receive an ACK from the client. If the server's first flight is lost and it then becomes blocked on anti-amplification, the client will only send an anti-deadlock packet that contains no ACKs, and the server might not have new data to send, leading to a deadlock unless time loss detection remains effective.

This is transcribed directly from [the draft's pseudocode](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#name-setting-the-loss-detection-).

Ralith: Nevermind, I was confused about how time threshold loss detection works--it's actually not used at all unless a later packet in a given space was already acked. That makes it rather difficult for it to arise when the peer's address hasn't yet been validated, but I'm not sure it's impossible...

djc: Yeah, I see what I was thinking here doesn't make sense.

PR #816, opened 2020-06-30
Fix documentation in quinn-proto and quinn-h3

alexander-jackson:
Fix some documentation links in `quinn-proto` and `quinn-h3` that `cargo doc` was complaining about.

`quinn-proto` had some links that did not point to the correct locations, so these were updated to point to `enum` variants as well.

`quinn-h3` made references to structs that no longer seem to exist so this was updated too.


djc:
Thanks!

PR #818, opened 2020-07-04
Implement pacing

### LINKED ISSUE #180: Implement pacing
As [discussed in the recovery document](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#pacing).

Demi-Marie:
This PR adds pacing.  Fixes #180.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -33,6 +33,8 @@ use crate::{
 };
 
 mod assembler;
+
+mod pacing;
 mod paths;
 use paths::PathData;
 
@@ -203,6 +205,7 @@ where
                 remote,
                 config.initial_rtt,
                 config.congestion_controller_factory.build(now),
+                now,
             ),
             prev_path: None,
             side,
@@ -384,11 +387,17 @@ where
             if space_id == SpaceId::Data {
                 ack_eliciting |= self.can_send_1rtt();
                 // Tail loss probes must not be blocked by congestion, or a deadlock could arise
-                if ack_eliciting
-                    && self.spaces[space_id].loss_probes == 0
-                    && self.congestion_blocked()
-                {
-                    continue;
+                if ack_eliciting && self.spaces[space_id].loss_probes == 0 {
+                    if self.congestion_blocked() {
+                        continue;
+                    }
+                    let smoothed_rtt = self.path.rtt.conservative();
+                    let window = self.path.congestion.window();
+                    if let Some(delay) = self.path.pacing.delay(smoothed_rtt, self.mtu, window, now)
+                    {
+                        self.timers.set(Timer::Pacing, delay);
+                        continue;
+                    }
                 }
             }
 
@@ -705,6 +714,7 @@ where
                     self.path.challenge = None;
                     self.path.challenge_pending = false;
                 }
+                Timer::Pacing => trace!("pacing timer expired"),
             }
         }
     }
@@ -978,6 +988,7 @@ where
                 self.permit_idle_reset = false;
             }
             self.set_loss_detection_timer(now);
+            self.path.pacing.on_transmit(size);
         }
     }
 
@@ -2422,12 +2433,13 @@ where
         // Note that the congestion window will not grow until validation terminates. Helps mitigate
         // amplification attacks performed by spoofing source addresses.
         let mut new_path = if remote.is_ipv4() && remote.ip() == self.path.remote.ip() {
-            PathData::from_previous(remote, &self.path)
+            PathData::from_previous(remote, &self.path, now)
         } else {
             PathData::new(
                 remote,
                 self.config.initial_rtt,
                 self.config.congestion_controller_factory.build(now),
+                now,
             )
         };
         new_path.challenge = Some(self.rng.gen());
--- a/quinn-proto/src/connection/pacing.rs
+++ b/quinn-proto/src/connection/pacing.rs
@@ -1,0 +1,130 @@
+//! Pacing of packet transmissions.
+
+use std::time::{Duration, Instant};
+
+use tracing::warn;
+
+/// A simple token-bucket pacer. The bucket starts full and has an adjustable capacity. Once the
+/// bucket is empty, further transmission is blocked. The bucket refills at a rate slightly faster
+/// than one congestion window per RTT.
+pub struct Pacer {
+    capacity: u64,
+    tokens: u64,
+    prev: Instant,
+}
+
+impl Pacer {
+    /// Obtains a new [`Pacer`].
+    pub fn new(capacity: u64, now: Instant) -> Self {
+        Self {
+            capacity,
+            tokens: capacity,
+            prev: now,
+        }
+    }
+
+    /// Record that a packet has been transmitted.
+    pub fn on_transmit(&mut self, packet_length: u16) {
+        self.tokens = self.tokens.saturating_sub(packet_length.into())
+    }
+
+    /// Return how long we need to wait before sending a packet.
+    ///
+    /// If we can send a packet right away, this returns `None`. Otherwise, returns `Some(d)`,
+    /// where `d` is the time before this function should be called again.
+    ///
+    /// The 5/4 ratio used here comes from the suggestion that N = 1.25 in the draft IETF RFC for
+    /// QUIC.
+    pub fn delay(
+        &mut self,
+        smoothed_rtt: Duration,
+        mtu: u16,
+        window: u64,
+        now: Instant,
+    ) -> Option<Instant> {
+        debug_assert_ne!(
+            window, 0,
+            "zero-sized congestion control window is nonsense"
+        );
+
+        // if we can already send a packet, there is no need for delay
+        if self.tokens > mtu.into() {
+            return None;
+        }
+
+        // we disable pacing for extremely large windows
+        if window > u32::max_value().into() {
+            return None;
+        }
+
+        let window = window as u32;
+
+        let time_elapsed = now.checked_duration_since(self.prev).unwrap_or_else(|| {
+            warn!("received a timestamp early than a previous recorded time, ignoring");
+            Default::default()
+        });
+
+        if smoothed_rtt.as_nanos() == 0 {
+            return None;
+        }
+
+        let elapsed_rtts = time_elapsed.as_secs_f64() / smoothed_rtt.as_secs_f64();
+        let new_tokens = window as f64 * 1.25 * elapsed_rtts;
+        self.tokens = self
+            .tokens
+            .saturating_add(new_tokens as _)
+            .min(self.capacity);
+
+        // if we can already send a packet, there is no need for delay
+        if self.tokens > mtu.into() {
+            return None;
+        }
+
+        let unscaled_delay = smoothed_rtt
+            .checked_mul(((mtu as u64).max(self.capacity) - self.tokens) as _)
+            .unwrap_or_else(|| Duration::new(u64::max_value(), 999_999_999))
+            / window;
+
+        // divisions come before multiplications to prevent overflow
+        // this is the time at which the pacing window becomes empty
+        Some(self.prev + (unscaled_delay / 5) * 4)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn does_not_panic_on_bad_instant() {
+        let old_instant = Instant::now();
+        let new_instant = old_instant + Duration::from_micros(15);
+        assert!(Pacer::new(1500, new_instant)
+            .delay(Duration::from_micros(0), 0, 1, old_instant)
+            .is_none());
+        assert!(Pacer::new(1500, new_instant)
+            .delay(Duration::from_micros(0), 1600, 1, old_instant)
+            .is_none());
+        assert!(Pacer::new(1500, new_instant)
+            .delay(Duration::from_micros(0), 1500, 3000, old_instant)
+            .is_none());
+    }
+
+    #[test]
+    fn computes_pause_correctly() {
+        let old_instant = Instant::now();
+        let mut pacer = Pacer::new(1500, old_instant + Duration::from_micros(15));
+        assert_eq!(
+            pacer.delay(Duration::from_micros(0), 1600, 1, old_instant),
+            None,
+            "Zero RTT means that we should send immediately"
+        );
+        let computed_delay = pacer.delay(Duration::from_micros(5), 1600, 1, old_instant);
+        assert_eq!(
+            computed_delay,
+            Some(pacer.prev + Duration::from_micros(400)),
+            "Difference between expected and computed delays is {}ns",
+            (computed_delay.unwrap() - pacer.prev).as_nanos()
+        );
+    }
+}
--- a/quinn-proto/src/connection/paths.rs
+++ b/quinn-proto/src/connection/paths.rs
@@ -1,5 +1,6 @@
-use std::{cmp, net::SocketAddr, time::Duration};
+use std::{cmp, net::SocketAddr, time::Duration, time::Instant};
 
+use super::pacing::Pacer;
 use crate::{congestion, TIMER_GRANULARITY};
 
 /// Description of a particular network path
@@ -10,6 +11,8 @@ pub struct PathData {
     pub sending_ecn: bool,
     /// Congestion controller state
     pub congestion: Box<dyn congestion::Controller>,
+    /// Pacing state
+    pub pacing: Pacer,
     pub challenge: Option<u64>,
     pub challenge_pending: bool,
 }
@@ -19,23 +22,27 @@ impl PathData {
         remote: SocketAddr,
         initial_rtt: Duration,
         congestion: Box<dyn congestion::Controller>,
+        now: Instant,
     ) -> Self {
         PathData {
             remote,
             rtt: RttEstimator::new(initial_rtt),
             sending_ecn: true,
+            pacing: Pacer::new(congestion.window(), now),
             congestion,
             challenge: None,
             challenge_pending: false,
         }
     }
 
-    pub fn from_previous(remote: SocketAddr, prev: &PathData) -> Self {
+    pub fn from_previous(remote: SocketAddr, prev: &PathData, now: Instant) -> Self {
+        let congestion = prev.congestion.clone_box();
         PathData {
             remote,
             rtt: prev.rtt,
-            congestion: prev.congestion.clone_box(),
+            pacing: Pacer::new(congestion.window(), now),
             sending_ecn: true,
+            congestion,
             challenge: None,
             challenge_pending: false,
         }
--- a/quinn-proto/src/connection/timer.rs
+++ b/quinn-proto/src/connection/timer.rs
@@ -14,23 +14,26 @@ pub(crate) enum Timer {
     PathValidation = 4,
     /// When to send a `PING` frame to keep the connection alive
     KeepAlive = 5,
+    /// When pacing will allow us to send a packet
+    Pacing = 6,
 }
 
 impl Timer {
-    pub(crate) const VALUES: [Self; 6] = [
+    pub(crate) const VALUES: [Self; 7] = [
         Timer::LossDetection,
         Timer::Idle,
         Timer::Close,
         Timer::KeyDiscard,
         Timer::PathValidation,
         Timer::KeepAlive,
+        Timer::Pacing,
     ];
 }
 
 /// A table of data associated with each distinct kind of `Timer`
 #[derive(Debug, Copy, Clone, Default)]
 pub(crate) struct TimerTable {
-    data: [Option<Instant>; 6],
+    data: [Option<Instant>; 7],
 }
 
 impl TimerTable {


djc:
The `RttEstimator` move conflicts with my work in #804 and seems unrelated, let's skip that for now?

quinn-proto/src/connection/mod.rs
```diff
@@ -35,6 +35,8 @@ use crate::{
 
 mod assembler;
 
+mod pacing;
```
djc, line 38: Looks like this module isn't actually included in this commit?

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,73 @@
+//! Pacing of packet transmissions.
```
djc, line 1: Let's squash this commit into the other one.

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,73 @@
...
+                .try_into()
+                .unwrap_or(u64::max_value());
+
+            let new_tokens = new_tokens.saturating_mul(elapsed);
+            self.tokens = self.tokens.saturating_add(new_tokens).min(self.capacity);
+        } else {
+            now = self.last_instant
+        }
+
+        if window > u32::max_value().into() {
```
djc, line 56: Can this come first, so we don't the work above for no reason?

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        let window = window as u32;
+        if self.tokens > mtu.into() {
+            None
+        } else {
+            let rtt = smoothed_rtt
+                .checked_mul((mtu - self.tokens as u16).into())
+                .unwrap_or_else(|| Duration::new(u64::max_value(), 999_999_999));
+            let new_duration: Duration = rtt / window / 5 * 4;
```
djc, line 68: Presumably we could do without type annotations here (and on the next line), in which case we should. Also, I'd prefer using parentheses in this expression to make the grouping clearer (and invert one of the operators).

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,73 @@
...
+            now = self.last_instant
+        }
+
+        if window > u32::max_value().into() {
+            // we disable pacing for extremely large windows
+            return None;
+        }
+
+        let window = window as u32;
+        if self.tokens > mtu.into() {
```
djc, line 62: I think turning this into an early return would be a bit nicer. Maybe move that directly after the assignment to `self.tokens`?

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,73 @@
...
+    /// If we can send a packet right away, this returns [`None`]. Otherwise, returns `Some(d)`,
+    /// where `d` is the time before this function should be called again.
+    ///
+    /// We panic if the passed-in instant is older than `self.last_instant`.
+    pub fn delay(
+        &mut self,
+        smoothed_rtt: Duration,
+        mtu: u16,
+        window: u64,
+        mut now: Instant,
```
djc, line 38: I'd avoid mutating `now` here and instead semantically naming the calculated value.

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,73 @@
...
+            last_instant: Instant::now(),
+        }
+    }
+
+    /// Return how long we need to wait before sending a packet.
+    ///
+    /// If we can send a packet right away, this returns [`None`]. Otherwise, returns `Some(d)`,
+    /// where `d` is the time before this function should be called again.
+    ///
+    /// We panic if the passed-in instant is older than `self.last_instant`.
```
djc, line 32: Since the `instant` comes from the user (indirectly), I'm not sure panicking is the right approach here. Is there some other way to respond here?

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,73 @@
...
+        } else {
+            let rtt = smoothed_rtt
+                .checked_mul((mtu - self.tokens as u16).into())
+                .unwrap_or_else(|| Duration::new(u64::max_value(), 999_999_999));
+            let new_duration: Duration = rtt / window / 5 * 4;
+            let new_instant: Instant = now + new_duration;
+            Some(new_instant)
+        }
+    }
+}
```
djc, line 73: Probably add some unit tests here to prove that the basic functionality works correctly?

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,73 @@
...
+    pub fn delay(
+        &mut self,
+        smoothed_rtt: Duration,
+        mtu: u16,
+        window: u64,
+        mut now: Instant,
+    ) -> Option<Instant> {
+        if let Some(time_elapsed) = now.checked_duration_since(self.last_instant) {
+            self.last_instant = now;
+            let new_tokens = window.saturating_mul(5) / 4;
```
djc, line 42: The 5 / 4 ratio used twice here could use some comments/documentation on why/how that's chosen.

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,73 @@
...
+        Self {
+            capacity,
+            tokens: capacity,
+            last_instant: Instant::now(),
+        }
+    }
+
+    /// Return how long we need to wait before sending a packet.
+    ///
+    /// If we can send a packet right away, this returns [`None`]. Otherwise, returns `Some(d)`,
```
djc, line 29:
```suggestion
    /// If we can send a packet right away, this returns `None`. Otherwise, returns `Some(d)`,
```

quinn-proto/src/connection/mod.rs
DemiMarie-parity: Sorry about that!

quinn-proto/src/connection/pacing.rs
DemiMarie-parity: I changed the code to ignore such timestamps.

quinn-proto/src/connection/pacing.rs
DemiMarie-parity: Done

quinn-proto/src/connection/pacing.rs
DemiMarie-parity: Done

quinn-proto/src/connection/pacing.rs
DemiMarie-parity: It comes from the current base draft as of this writing, which [suggests N = 1.25](https://github.com/quicwg/base-drafts/blob/5e3723af74412f87907ec50025bd9c7510e8ad27/draft-ietf-quic-recovery.md).

quinn-proto/src/connection/mod.rs
```diff
@@ -2442,6 +2450,7 @@ where
             sending_ecn: self.path.sending_ecn || !maybe_rebinding,
             challenge: Some(self.rng.gen()),
             challenge_pending: true,
+            pacing: pacing::Pacer::new(0),
```
djc, line 2453: Why does this make sense? Why not use the current congestion window from the congestion controller we're setting for the path?

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,103 @@
...
+        window: u64,
+        now: Instant,
+    ) -> Option<Instant> {
+        debug_assert_ne!(
+            window, 0,
+            "zero-sized congestion control window is nonsense"
+        );
+        if let Some(time_elapsed) = now.checked_duration_since(self.last_instant) {
+            self.last_instant = now;
+            let new_tokens = window.saturating_mul(5) / 4;
```
djc, line 47: Please move the definition of `new_tokens` down to where it's actually used (not used here if `smoothed_rtt` is 0) (consider collapsing it with the redefinition below).

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,103 @@
...
+    last_instant: Instant,
+}
+
+impl Pacer {
+    /// Obtains a new [`Pacer`].
+    pub fn new(capacity: u64) -> Self {
+        Self {
+            capacity,
+            tokens: capacity,
+            last_instant: Instant::now(),
```
djc, line 23: Maybe just call this `last` or `prev`? The `_instant` suffix feels redundant given the type.

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,103 @@
...
+            return None;
+        }
+
+        let window = window as u32;
+        let rtt = smoothed_rtt
+            .checked_mul((mtu - self.tokens as u16).into())
+            .unwrap_or_else(|| Duration::new(u64::max_value(), 999_999_999));
+        let new_duration = ((rtt / window) / 5) * 4;
+        let new_instant = self.last_instant + new_duration;
+        Some(new_instant)
```
djc, line 70: The `new_` prefixes here feel a bit redundant since there aren't other variables in the method scope that they're easily confused with. Also, a number of definitions here only have a single usage site, so it feels to me like inlining them into larger expressions would make it easier to follow the code.

DemiMarie-parity:
I just realized that this breaks a basic invariant of `quinn-proto`, by obtaining a timestamp from the OS (via `Instant::now()`).

djc:
Good catch! Should be easy to fix, I think?

DemiMarie-parity:
Indeed it is.  I pushed a fix.  We might want to consider making `quinn-proto` `#![no_std]`, which would catch these problems at compile-time.

Also, there are nowhere near enough tests, but I am not sure what those tests should cover.

DemiMarie-parity:
It is also worth noting that this implementation is purely at the `quinn-proto` level, and so does not support any form of offloading.  As such, it will not achieve the maximum possible performance.

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,106 @@
...
+            .delay(Duration::from_micros(0), 1600, 1, old_instant)
+            .is_some());
+        assert!(Pacer::new(1500, Instant::now())
+            .delay(Duration::from_micros(0), 1500, 3000, old_instant)
+            .is_none());
+    }
+    #[test]
+    fn computes_pause_correctly() {
+        let old_instant = Instant::now();
+        while Instant::now() <= old_instant {}
```
Ralith, line 91: Rather than trying to massage the value of `Instant::now()`, just add a fixed `Duration` to it manually. That's simpler and provides better reproducibility.

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,106 @@
...
+            }
+            let elapsed = (time_elapsed.as_nanos() / smoothed_rtt.as_nanos())
+                .try_into()
+                .unwrap_or(u64::max_value());
+
+            let new_tokens = (window.saturating_mul(5) / 4).saturating_mul(elapsed);
+            self.tokens = self.tokens.saturating_add(new_tokens).min(self.capacity);
+        }
+
+        // we disable pacing for extremely large windows
```
Ralith, line 58: I spent a minute wondering how this comment could possibly describe the `self.tokens >= mtu.into()` part before finally realizing that it isn't meant to. Maybe make that more explicit by describing both cases?

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,110 @@
...
+            if smoothed_rtt.as_nanos() == 0 {
+                return None;
+            }
+            let elapsed = (time_elapsed.as_nanos() / smoothed_rtt.as_nanos())
+                .try_into()
+                .unwrap_or(u64::max_value());
+
+            let new_tokens = (window.saturating_mul(5) / 4).saturating_mul(elapsed);
+            self.tokens = self.tokens.saturating_add(new_tokens).min(self.capacity);
+        }
```
Ralith, line 56: Would it make sense to split this into two separate methods--one that queries the next permitted transmit time (and only takes `&self`), and one that updates the state to account for a transmit taking place? I'm concerned that the behavior of this as written might be sensitive to the rate at which it's called, which has no particular bounds in either direction.

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,110 @@
...
+            .delay(Duration::from_micros(0), 0, 1, old_instant)
+            .is_none());
+        assert!(Pacer::new(1500, new_instant)
+            .delay(Duration::from_micros(0), 1600, 1, old_instant)
+            .is_some());
+        assert!(Pacer::new(1500, new_instant)
+            .delay(Duration::from_micros(0), 1500, 3000, old_instant)
+            .is_none());
+    }
+    #[test]
```
Ralith, line 93: Style nit:
```suggestion
    }

    #[test]
```

quinn-proto/src/connection/pacing.rs
DemiMarie-parity: It probably does.  I will try and see what happens.

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,125 @@
...
+        );
+
+        // if we can already send a packet, there is no need for delay
+        if self.tokens > mtu.into() {
+            return None;
+        }
+
+        // we disable pacing for extremely large windows
+        if window > u32::max_value().into() {
+            debug!("window too large, disabling pacing");
```
Ralith, line 57: I feel like this might get really spammy under certain circumstances, and repeating it probably isn't all that useful. Maybe just omit?

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,125 @@
...
+            .saturating_add(new_tokens as _)
+            .min(self.capacity);
+
+        // if we can already send a packet, there is no need for delay
+        if self.tokens > mtu.into() {
+            return None;
+        }
+
+        let rtt = smoothed_rtt
+            .checked_mul(((mtu as u64).max(self.capacity) - self.tokens).min(65535) as _)
```
Ralith, line 83: We should be clear where this constant is coming from, and why it's used here.

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,125 @@
...
+            .unwrap_or_else(|| Duration::new(u64::max_value(), 999_999_999));
+        Some(self.prev + ((rtt / window as u32) / 5) * 4)
```
Ralith, line 85:
```suggestion
        Some(self.prev + (4 * rtt / (window as u32 * 5))
```
Seems a lot clearer.

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,125 @@
...
+            .tokens
+            .saturating_add(new_tokens as _)
+            .min(self.capacity);
+
+        // if we can already send a packet, there is no need for delay
+        if self.tokens > mtu.into() {
+            return None;
+        }
+
+        let rtt = smoothed_rtt
```
Ralith, line 82: `srtt * mtu` is definitely not an rtt. I'm not sure what it *is* exactly, but we shouldn't call it RTT.

quinn-proto/src/connection/pacing.rs
Ralith, line 85: Is this computing the time until the pacing window is 100% empty? If so, a comment should say so.

DemiMarie-parity: I wrote the code the way I did to prevent any possibility of integer overflow.

quinn-proto/src/connection/pacing.rs
DemiMarie-parity: `time_to_transmit_mtu` perhaps?

quinn-proto/src/connection/pacing.rs
Ralith: That's fair, but is sufficiently non-obvious that a comment is needed--else it's liable to inadvertently disappear in a refactoring.

quinn-proto/src/connection/pacing.rs
Ralith: The time required to transmit an MTU depends mainly on the congestion window, which isn't involved in this specific computation at all. `rtt_times_mtu` would be acceptable, if nothing else.

quinn-proto/src/connection/mod.rs
```diff
@@ -2732,6 +2738,18 @@ where
         self.in_flight.bytes + u64::from(self.mtu) >= self.path.congestion.window()
     }
 
+    /// Whether UDP transmits are currently blocked by pacing
+    fn pacing_blocked(&mut self, now: Instant) -> bool {
+        let smoothed_rtt = self.path.rtt.conservative();
+        let window = self.path.congestion.window();
+        if let Some(delay) = self.path.pacing.delay(smoothed_rtt, self.mtu, window, now) {
+            self.timers.set(Timer::Pacing, delay);
```
djc, line 2746: I don't really like the mixing of query and side effect in this method. Can we just outline the call to `self.timers.set()` into the `poll_transmit()` caller? I think that will make it clearer what's going on. The next question is then whether this method still carries its weight, maybe it should just be inlined into `poll_transmit()` altogether, since it's single-caller anyway.

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,125 @@
...
+    /// If we can send a packet right away, this returns `None`. Otherwise, returns `Some(d)`,
+    /// where `d` is the time before this function should be called again.
+    ///
+    /// The 5/4 ratio used here comes from the suggestion that N = 1.25 in the draft IETF RFC for
+    /// QUIC.
+    pub fn delay(
+        &mut self,
+        smoothed_rtt: Duration,
+        mtu: u16,
+        window: u64,
```
djc, line 42: It doesn't look like `window`'s natural type for this algorithm is `u64` -- it gets compared to `u32::max_value()`, cast to `f64` for the token calculation and then case to `u32` again in the return value. Similarly, it looks like `mtu` gets cast to `u64` at all usage sites. Would be better to cast these once early on in the method or take their argument in the "right" type directly, depending on whether the caller or this method should handle any conversion edge cases.

quinn-proto/src/connection/pacing.rs
djc: `rtt_times_mtu` would be pretty meh IMO. Can't we come up with any name that reflects what this value represents semantically? If not, maybe we should move around some of the arithmetic operations so that any named values are semantically comprehensible?

quinn-proto/src/connection/pacing.rs
DemiMarie-parity: It’s an arbitrary heuristic.

DemiMarie-parity: Its purpose is to prevent excessive latency in corner cases.

quinn-proto/src/connection/pacing.rs
DemiMarie-parity: Most of this method *is* conversion edge cases :smile:.  Another option is to just cast to `f64` at the start of the method and then cast back at the end, which would simplify the code significantly.

quinn-proto/src/connection/pacing.rs
DemiMarie-parity: If I move the division up, it could be `unscaled_delay`.

quinn-proto/src/connection/pacing.rs
DemiMarie-parity: Indeed.

quinn-proto/src/connection/pacing.rs
djc: That'd work for me, too.

quinn-proto/src/connection/pacing.rs
djc: Sounds better to me!

quinn-proto/src/connection/pacing.rs
DemiMarie-parity: I fixed it by doing the calculation up-front.  Using floating point everywhere made me a bit nervous about loss of precision.

quinn-proto/src/connection/mod.rs
DemiMarie-parity: Good idea, thanks!

quinn-proto/src/connection/mod.rs
```diff
@@ -386,7 +389,18 @@ where
                 // Tail loss probes must not be blocked by congestion, or a deadlock could arise
                 if ack_eliciting
                     && self.spaces[space_id].loss_probes == 0
-                    && self.congestion_blocked()
+                    && (self.congestion_blocked() || {
```
djc, line 392: Let's just do this as a separate `if` statement? This is a bit too deeply nested for my taste.

DemiMarie-parity:
@djc @Ralith would you mind re-reviewing?

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,133 @@
+//! Pacing of packet transmissions.
+
+use std::time::{Duration, Instant};
+
+use tracing::warn;
+
+/// A simple token-bucket pacer. The bucket starts full and has an adjustable capacity. Once the
+/// bucket is empty, further transmission is blocked. The bucket refills at a rate slightly faster
+/// than one contestion window per RTT.
```
djc, line 9:
```suggestion
/// than one congestion window per RTT.
```

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,133 @@
...
+            .tokens
+            .saturating_add(new_tokens as _)
+            .min(self.capacity);
+
+        // if we can already send a packet, there is no need for delay
+        if self.tokens > mtu.into() {
+            return None;
+        }
+
+        let unscaled_delay = smoothed_rtt
```
Ralith, line 84: Good name choice!

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,133 @@
...
+            .checked_mul(
+                ((mtu as u64).max(self.capacity) - self.tokens).min(MAX_PACING_DELAY_BYTES) as _,
+            )
+            .unwrap_or_else(|| Duration::new(u64::max_value(), 999_999_999))
+            / window;
+
+        // divisions come before multiplications to prevent overflow
+        // this is the time until the pacing window is empty
```
Ralith, line 92:
```suggestion
        // this is the time at which the pacing window is empty
```
as it's an `Instant`, not a `Duration`.

quinn-proto/src/connection/pacing.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// The 5/4 ratio used here comes from the suggestion that N = 1.25 in the draft IETF RFC for
+    /// QUIC.
+    pub fn delay(
+        &mut self,
+        smoothed_rtt: Duration,
+        mtu: u16,
+        window: u64,
+        now: Instant,
+    ) -> Option<Instant> {
+        const MAX_PACING_DELAY_BYTES: u64 = (1 << 16) - 1;
```
Ralith, line 45: Sorry, but I'm still unclear on why we have this bound, and why this particular value was chosen. Can you elaborate in a comment?

Demi-Marie: The value is actually arbitrary.  What would you suggest @Ralith?  The bound is to ensure that pacing does not delay transmission indefinitely.

quinn-proto/src/connection/paths.rs
```diff
@@ -19,23 +22,27 @@ impl PathData {
...
     }
 
-    pub fn from_previous(remote: SocketAddr, prev: &PathData) -> Self {
+    pub fn from_previous(remote: SocketAddr, prev: &PathData, now: Instant) -> Self {
+        let congestion = prev.congestion.clone_box();
         PathData {
             remote,
             rtt: prev.rtt,
-            congestion: prev.congestion.clone_box(),
+            pacing: Pacer::new(congestion.window(), now),
```
Ralith, line 43: We're required to use the *initial* congestion window for the pacer capacity; this uses the current window. A simple solution would be to extend the `congestion::Controller` trait with a `fn initial_window(&self) -> u64`.

quinn-proto/src/connection/pacing.rs
Ralith: Under what circumstance is that a risk?

Note that `mtu` is limited by the maximum theoretical size of a UDP packet (~2^16) and `self.capacity` is limited to at most the initial congestion window, which is [typically even smaller](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#name-initial-and-minimum-congest). As written, I don't think this limit will ever take effect.

If you're concerned about the "wait until a full burst can be sent" strategy, maybe we should compromise that by allowing a lower threshold (say, at least N packets)? Or alternatively, roll back to waiting for a single packet in lieu of hard performance data to support the optimization.

Demi-Marie: In that case, I will just delete this check.

quinn-proto/src/connection/paths.rs
Ralith: It looks like this still needs to be addressed.

Ralith (changes requested):
Sorry I haven't gotten back to this; work's been sapping most of my energy. Only one issue left, which I don't mind sorting out myself if you prefer. Considering how long this has been kicking around I'll proceed with that this weekend if I haven't heard otherwise.

quinn-proto/src/connection/paths.rs
Demi-Marie: I don’t have write access to the source branch anymore, so I would need to close this PR and open it with a different source branch.  Is that okay?

Ralith:
Sure, that's fine!

Ralith:
Closing in favor of #852.

PR #819, opened 2020-07-04
H3: accept goaway from client

stammw:
Make it possible to gracefully shutdown the connection from the client, as described in draft-29. 
This is not very useful as long as server push is not implemented.

### DIFF
--- a/quinn-h3/src/client.rs
+++ b/quinn-h3/src/client.rs
@@ -484,6 +484,14 @@ impl Connection {
             .close(ErrorCode::NO_ERROR.into(), b"Connection closed");
     }
 
+    /// Close the connection gracefully
+    ///
+    /// All ongoing requests will be completed, all new requests or server push will
+    /// be rejected.
+    pub fn go_away(&mut self) {
+        self.0.h3.lock().unwrap().inner.go_away();
+    }
+
     // Update traffic keys spontaneously for testing purposes.
     #[doc(hidden)]
     pub fn force_key_update(&self) {
--- a/quinn-h3/src/client.rs
+++ b/quinn-h3/src/client.rs
@@ -515,7 +515,6 @@ pub struct SendRequest<B, D> {
     request: Option<Request<B>>,
     #[pin]
     state: SendRequestState<B, D>,
-    open: OpenBi,
     chan: Option<oneshot::Sender<(RecvStream, StreamId)>>,
 }
 
@@ -529,13 +528,21 @@ where
         conn: ConnectionRef,
         request: Request<B>,
     ) -> Self {
+        if conn.h3.lock().unwrap().inner.is_closing() {
+            return Self {
+                conn,
+                chan: None,
+                request: None,
+                state: SendRequestState::Aborted,
+            };
+        }
+
         let open = conn.quic.open_bi();
         Self {
             conn,
-            open,
             chan: Some(open_send),
             request: Some(request),
-            state: SendRequestState::Opening,
+            state: SendRequestState::Opening(open),
         }
     }
 
@@ -584,8 +591,11 @@ where
         let mut me = self.project();
         loop {
             match &mut me.state.as_mut().project() {
-                SendRequestStateProj::Opening => {
-                    match ready!(me.open.poll_unpin(cx)) {
+                SendRequestStateProj::Aborted => {
+                    return Poll::Ready(Err(Error::Aborted));
+                }
+                SendRequestStateProj::Opening(open) => {
+                    match ready!(open.poll_unpin(cx)) {
                         Err(e) => {
                             me.chan.take().unwrap(); // drop it so RecvResponse fails
                             return Poll::Ready(Err(e.into()));
@@ -623,8 +633,9 @@ where
 
 #[pin_project(project = SendRequestStateProj)]
 enum SendRequestState<B, D> {
-    Opening,
+    Opening(OpenBi),
     Sending(#[pin] SendData<B, D>),
+    Aborted,
 }
 
 /// Receive an HTTP/3 response
--- a/quinn-h3/src/client.rs
+++ b/quinn-h3/src/client.rs
@@ -81,7 +81,7 @@ use http_body::Body as HttpBody;
 use pin_project::pin_project;
 use quinn::{Certificate, Endpoint, OpenBi, RecvStream};
 use quinn_proto::{Side, StreamId};
-use tracing::trace;
+use tracing::{debug, trace};
 
 use crate::{
     body::RecvBody,
@@ -486,10 +486,11 @@ impl Connection {
 
     /// Close the connection gracefully
     ///
-    /// All ongoing requests will be completed, all new requests or server push will
-    /// be rejected.
-    pub fn go_away(&mut self) {
-        self.0.h3.lock().unwrap().inner.go_away();
+    /// All ongoing requests will be completed, all new requests will be aborted. And server pushes
+    /// will be rejected if there is more than `allow_pushes` in flight before the server recieves
+    /// the shutdown signal.
+    pub fn go_away(&mut self, allow_pushes: u64) {
+        self.0.h3.lock().unwrap().inner.go_away(allow_pushes);
     }
 
     // Update traffic keys spontaneously for testing purposes.
@@ -590,6 +591,7 @@ where
                             return Poll::Ready(Err(e.into()));
                         }
                         Ok((send, recv)) => {
+                            debug!(stream=%send.id(), "request");
                             let (method, uri, headers, body) = me.take()?;
 
                             if recv.is_0rtt() && !method.is_idempotent() {
--- a/quinn-h3/src/connection.rs
+++ b/quinn-h3/src/connection.rs
@@ -89,7 +89,7 @@ impl ConnectionRef {
                 incoming_bi: bi_streams,
                 incoming_uni: uni_streams,
                 pending_uni: VecDeque::with_capacity(3),
-                inner: Connection::with_settings(settings),
+                inner: Connection::new(side, settings),
                 requests: VecDeque::with_capacity(16),
                 requests_task: None,
                 recv_control: None,
--- a/quinn-h3/src/connection.rs
+++ b/quinn-h3/src/connection.rs
@@ -380,7 +380,7 @@ impl ConnectionInner {
                             trace!("Got Settings: {:#?}", s);
                             self.inner.set_remote_settings(Settings::from_frame(s)?)?;
                         }
-                        (true, Side::Client, HttpFrame::Goaway(id)) => {
+                        (true, _, HttpFrame::Goaway(id)) => {
                             trace!("Got Goaway({:?})", id);
                             self.inner.leave(StreamId(id));
                         }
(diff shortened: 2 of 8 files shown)


quinn-h3/src/connection.rs
```diff
@@ -104,6 +104,8 @@ impl ConnectionRef {
     }
 }
 
+type ConnectionEnd = bool;
```
djc, line 110: I don't see this adding much value, since it doesn't provide type safety. Maybe just add some comments to the `drive()` function instead, or instead make it an actual enum? (I've done the latter in work code and I did feel it made the code clearer -- and it should be basically from a performance perspective.)

stammw: I got the enum named `DriveState::{Closed, Running}`. Thanks!

quinn-h3/src/connection.rs
```diff
@@ -133,7 +141,10 @@ impl ConnectionInner {
 
         self.reset_waker(cx);
 
-        Ok(self.inner.is_closing() && self.inner.requests_in_flight() == 0)
+        if self.inner.is_closing() && self.inner.requests_in_flight() == 0 {
```
djc, line 144: Missed this the first time around, but maybe consider hoisting the `Ok()` out and moving the `DriveState::Running` into an `else` branch?

stammw: This is from the last review addressing work, you couldn't catch it! :)

stammw: It's better like this, thanks.

stammw:
I'll get the failing test pass soon. Seems like it will need some effort to get it right, the goaway behavior is not the best work I've done so far.

stammw:
I pushed i little further than just making the test pass. Now, incoming `StreamId` are checked against the `GoAway(StreamId)` value. It is still not perfect though.

quinn-h3/src/client.rs
```diff
@@ -488,8 +488,8 @@ impl Connection {
     ///
     /// All ongoing requests will be completed, all new requests or server push will
     /// be rejected.
-    pub fn go_away(&mut self) {
-        self.0.h3.lock().unwrap().inner.go_away();
+    pub fn go_away(&mut self, grace: u64) {
```
djc, line 491: What is the `grace` value here? Maybe add a comment about that? Alternatively, I think you've named it `max_id` in other places, which is a bit easier to understand.

quinn-h3/src/proto/connection.rs
```diff
@@ -187,34 +187,36 @@ impl Connection {
         self.pending_streams[ty as usize].reserve(capacity);
     }
 
-    pub fn request_initiated(&mut self, id: StreamId) {
-        if !self.go_away {
-            self.requests_in_flight.insert(id);
-            self.max_id_in_flight = cmp::max(id.0, self.max_id_in_flight);
+    // The remote peer initiated a stream: a request from client or a push from server
+    pub fn remote_stream_initiated(&mut self, id: StreamId) -> Result<()> {
+        match self.go_away {
```
djc, line 192: I think this might read nicer as an `if let` (although it's unfortunate that the `id.0 > max_id` will need a nested `if`) with early return. Maybe have a look and decide for yourself which you prefer?

quinn-h3/src/client.rs
```diff
@@ -528,13 +527,21 @@ where
         conn: ConnectionRef,
         request: Request<B>,
     ) -> Self {
+        if conn.h3.lock().unwrap().inner.is_closing() {
```
djc, line 530: Might be nice to add a test for this? I think that would be easy to do from your existing `GoAway` test?

quinn-h3/src/proto/connection.rs
```diff
@@ -226,8 +226,8 @@ impl Connection {
         );
     }
 
-    pub fn requests_in_flight(&self) -> usize {
-        self.requests_in_flight.len()
+    pub fn shutdown_complete(&self) -> bool {
```
djc, line 229: If there's only one user, is it really useful for this to be a separate method?

quinn-h3/src/client.rs
stammw: I changed this from max_id because now it is the number of potentially in-flight streams to gracefully accept before starting rejecting. Here it could be called `allow_push`, and in `servers.rs` it could be `allowed_request`. Do you think these names would be a better fit?

quinn-h3/src/proto/connection.rs
stammw: The `if let` with a nested `if` is more expressive indeed. I wish we could use a combination in `if let` so we wouldn't be tempted to use `match { Pattern if condition => ... } ` syntactic sugar in these situation.

quinn-h3/src/proto/connection.rs
stammw: The idea here is to move non-async logic into `proto` module. There is still too much in `src/connection.rs` in my opinion.

quinn-h3/src/client.rs
djc: Yeah, those sound good!

quinn-h3/src/proto/connection.rs
djc: I think this is being worked on in the lang team, don't know what the current status is though.

quinn-h3/src/proto/connection.rs
djc: OK.

quinn-h3/src/client.rs
stammw: I got it differentiated.

stammw:
I got a little further again.

I figured out that the `GoAway(id)` wasn't correctly calculated, so I thought I'd rather "Make StreamId::new public (but hidden)" to prevent duplicating the `id` calculation code. I hope it doesn't bring an API design problem ?

I got the `go_away` test updated with 4 cases: 
- request is ongoing while `GoAway` is sent
- request is in-flight while `GoAway` is sent and is gracefully handled
- request is in-flight while `GoAway` is sent and is rejected because it's beyond the grace interval
- request is attempted after `GoAway` has been received by the client, so it is aborted locally

Also, I've added some logging calls to see that happen more easily. I think they will be useful so I left them. I'm just not sure for their levels.

I pushed this as a new commit to help reading what has changed really, I'll rebase it into "H3: Refactor GoAway mechanism to actually use the id" afterwards. (But if you think it doesn't help I can do it right away).

quinn-proto/src/lib.rs
```diff
@@ -176,7 +176,8 @@ impl fmt::Display for StreamId {
 }
 
 impl StreamId {
-    pub(crate) fn new(initiator: Side, dir: Dir, index: u64) -> Self {
+    #[doc(hidden)]
```
djc, line 179: @Ralith you okay with this? Should we attach explicit comments to some of these `#[doc(hidden)]` to clarify to readers that we don't consider these to be part of the public API?

stammw: There are no public functions that accept a `StreamId` in `quinn`, so I guess it does not open any unwanted API usage at this level. This is not the case of `quinn-proto`.

We could add a new type that is allowed to be publicly constructed, but not accepted by the API ?

stammw:
I think I've got all this sorted out.

quinn-h3/src/proto/connection.rs
```diff
@@ -291,6 +324,7 @@ mod tests {
     impl Default for Connection {
         fn default() -> Self {
```
Ralith, line 325: Could this dispatch to `Connection::new` to avoid duplication and risk of divergence?

quinn-proto/src/lib.rs
Ralith: Sorry, missed the ping somehow! If I understand correctly, this is motivated by HTTP/3 passing Stream IDs explicitly at the application protocol layer. Is that correct? If so, I think that's evidence that maybe this should actually be public for the benefit of application protocol implementers in general. There's no specific reason I'm aware of that this has to be private except the general principle of "don't export things without a need." If downstream code constructs nonsense Stream IDs and gets nonsense behavior, that's on them.

stammw: This is motivated by the need of calculating a correct `StreamId` raw value to be used in `GoAway` frames and the shutdown mechanism.

All right, we're all set then ! Thanks !

quinn-h3/src/proto/connection.rs
stammw: Yeah, I hesitated to get rid of that. Will do.

stammw: Rebased into "H3: store side information in proto::connection "

quinn-h3/src/proto/connection.rs
```diff
@@ -189,34 +195,54 @@ impl Connection {
...
-    pub fn go_away(&mut self) {
-        if !self.go_away {
-            self.go_away = true;
-            HttpFrame::Goaway(self.max_id_in_flight)
-                .encode(&mut self.pending_streams[PendingStreamType::Control as usize]);
-        }
+    // A graceful shutdown initiated locally, let `allow_streams` potentially in-flight
+    // incoming streams be processed.
+    pub fn go_away(&mut self, allow_streams: u64) {
+        let dir = if self.side == Side::Server {
```
djc, line 225: This is probably better written as a match (for one thing, it takes one less line).

quinn-h3/src/proto/connection.rs
```diff
@@ -72,15 +72,21 @@ impl Connection {
             BytesMut::with_capacity(2048),
         ];
 
+        let dir = if side == Side::Server {
```
djc, line 75: Better as a match.

quinn-h3/src/proto/connection.rs
```diff
@@ -189,34 +195,54 @@ impl Connection {
...
+                return Err(Error::Aborted);
+            }
         }
+        debug!(id=%id, "accepting");
+        self.requests_in_flight.insert(id);
+        self.max_id_in_flight = if id.index() > self.max_id_in_flight.index() {
+            id
+        } else {
+            self.max_id_in_flight
+        };
```
djc, line 212:
```suggestion
        if id.index() > self.max_id_in_flight.index() {
            self.max_id_in_flight = id;
        };
```

quinn-h3/src/connection.rs
```diff
@@ -141,13 +141,11 @@ impl ConnectionInner {
...
-                DriveState::Closed
-            } else {
-                DriveState::Running
-            },
-        )
+        Ok(if self.inner.shutdown_complete() {
+            DriveState::Closed
+        } else {
+            DriveState::Running
+        })
```
djc, line 148:
```suggestion
        Ok(match self.inner.shutdown_complete() {
            true => DriveState::Closed,
            false => DriveState::Running,
        })
```

quinn-proto/src/lib.rs
djc: I think @Ralith was suggesting to make this properly public, so please remove `#[doc(hidden)]` and add a short docstring.

djc:
👍

PR #823, opened 2020-07-05
Move PathData/RttEstimator to a paths module

djc:
The refactoring parts of #804 so we can get this out of the way.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1273,9 +1273,8 @@ where
     }
 
     fn pto_time_and_space(&self, now: Instant) -> Option<(Instant, SpaceId)> {
-        let srtt = self.path.rtt.get();
         let backoff = 2u32.pow(self.pto_count.min(MAX_BACKOFF_EXPONENT));
-        let mut duration = (srtt + TIMER_GRANULARITY.max(4 * self.path.rtt.var)) * backoff;
+        let mut duration = self.path.rtt.pto_base() * backoff;
 
         if self.in_flight.is_empty() {
             debug_assert!(!self.peer_completed_address_validation());
@@ -1356,9 +1355,7 @@ where
 
     /// Probe Timeout
     fn pto(&self) -> Duration {
-        self.path.rtt.get()
-            + cmp::max(4 * self.path.rtt.var, TIMER_GRANULARITY)
-            + self.max_ack_delay()
+        self.path.rtt.pto_base() + self.max_ack_delay()
     }
 
     fn on_packet_authenticated(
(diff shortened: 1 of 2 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -3165,9 +3165,9 @@ impl RttEstimator {
         }
     }
 
-    /// Get current smoothed RTT estimate
-    fn get(&self) -> Duration {
-        self.smoothed.unwrap_or(self.latest)
+    pub fn get(&self) -> Duration {
+        self.smoothed
+            .map_or(self.latest, |x| cmp::max(x, self.latest))
```
Ralith: Ah, I was confused about where that `get` was coming from in the other issue. I think the version introduced in #822 is more correct, however. The only time I see this `max` in the current draft is in [computing the time threshold for loss detection](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#name-time-threshold), which we already apply separately.

quinn-proto/src/connection/mod.rs
```diff
@@ -3167,6 +3167,10 @@ impl RttEstimator {
             .map_or(self.latest, |x| cmp::max(x, self.latest))
     }
 
+    pub fn max(&self) -> Duration {
+        self.get().max(self.latest)
+    }
+
```
Ralith: Is it worth abstracting this when it's used in only one place? Also, it's confusing that `rtt.max()` is not the maximum observed RTT, whereas `rtt.min` *is* the minimum observed RTT.

quinn-proto/src/connection/mod.rs
djc: Ah, that was an unintentional conflict resolution error. Fixed now.

quinn-proto/src/connection/mod.rs
djc: I think you previously argued in favor of accessors for reasons of keeping fields private? I think the same applies here, plus there's actually calculation going on here, so we're not just abstracting over the field. I'm open to a better name; all I could come up with was something like `max_smoothed_or_latest()` or some such, which seemed wordy and still not that clear.

djc:
(Also, #822 was actually necessary to make the abstractions line up here, so no stepping on toes at all...)

quinn-proto/src/connection/mod.rs
Ralith: Ah, right, since we're moving this into its own module it makes sense.

Maybe name it `conservative` and give it a decent doc comment? Still not self-explanatory, but at least it doesn't imply the wrong thing.

djc: Done.

(Actually I guess the argument for this which is strongest IMO is that providing abstractions local to the data makes it easier to prevent divergent use of the data, of the sort you fixed in #822.)

PR #824, opened 2020-07-05
Don’t try to run code coverage on PRs

DemiMarie-parity:
It won’t work, as those builds don’t have access to the CodeCov API
token.


djc:
Thanks!

PR #826, opened 2020-07-06
Add documentation links to improve navigation

alexander-jackson:
Add documentation links and references to some of the main parts of the API so
that relevant structs, enums and methods can be navigated to more quickly. This
reduces the need for the search bar and thus makes it easier to find related
content wherever it is mentioned.


djc:
Sounds great, thanks!

PR #831, opened 2020-07-26
fuzzing

jafow:
# summary

adds a starting point for fuzzing with `cargo-fuzz` following from #15 

this is runnable as `cargo fuzz run streams` with libfuzzer assuming the subcommand [is installed](https://github.com/rust-fuzz/cargo-fuzz#installation)

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -1,0 +1,35 @@
+[package]
+name = "quinn-fuzz"
+version = "0.0.0"
+authors = ["Automatically generated"]
+publish = false
+edition = "2018"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+arbitrary = { version = "0.4.5", features = ["derive"] }
+libfuzzer-sys = "0.3"
+
+[dependencies.proto]
+features = ["arbitrary"]
+path = "../quinn-proto"
+package = "quinn-proto"
+version = "0.6.1"
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
+
+[[bin]]
+name = "streams"
+path = "fuzz_targets/streams.rs"
+test = false
+doc = false
+
+[[bin]]
+name = "streamid"
+path = "fuzz_targets/streamid.rs"
+test = false
+doc = false
--- a/fuzz/fuzz_targets/streamid.rs
+++ b/fuzz/fuzz_targets/streamid.rs
@@ -1,0 +1,19 @@
+#![no_main]
+use arbitrary::Arbitrary;
+use libfuzzer_sys::fuzz_target;
+
+extern crate proto;
+use proto::{Dir, Side, StreamId};
+
+#[derive(Arbitrary, Debug)]
+struct StreamIdParams {
+    side: Side,
+    dir: Dir,
+    index: u64,
+}
+
+fuzz_target!(|data: StreamIdParams| {
+    let s = StreamId::new(data.side, data.dir, data.index);
+    assert_eq!(s.initiator(), data.side);
+    assert_eq!(s.dir(), data.dir);
+});
--- a/fuzz/fuzz_targets/streams.rs
+++ b/fuzz/fuzz_targets/streams.rs
@@ -1,0 +1,65 @@
+#![no_main]
+
+use arbitrary::{Arbitrary, Unstructured};
+use libfuzzer_sys::fuzz_target;
+
+extern crate proto;
+use proto::fuzzing::{ResetStream, Streams, TransportParameters};
+use proto::{Dir, Side, StreamId, VarInt};
+
+#[derive(Arbitrary, Debug)]
+struct StreamParams {
+    side: Side,
+    max_remote_uni: u16,
+    max_remote_bi: u16,
+    send_window: u16,
+    receive_window: u16,
+    stream_receive_window: u16,
+    dir: Dir,
+    transport_params: TransportParameters
+}
+
+#[derive(Arbitrary, Debug)]
+enum Operation {
+    Open,
+    Accept(Dir),
+    Finish(StreamId),
+    ReceivedStopSending(StreamId, VarInt),
+    ReceivedReset(ResetStream),
+    Reset(StreamId),
+}
+
+fuzz_target!(|input: (StreamParams, Vec<Operation>)| {
+    let (params, operations) = input;
+    let mut stream = Streams::new(
+        params.side,
+        params.max_remote_uni.into(),
+        params.max_remote_bi.into(),
+        params.send_window.into(),
+        params.receive_window.into(),
+        params.stream_receive_window.into(),
+    );
+
+    for operation in operations {
+        match operation {
+            Operation::Open => {
+                stream.open(&params.transport_params, params.dir);
+            }
+            Operation::Accept(dir) => {
+                stream.accept(dir);
+            }
+            Operation::Finish(id) => {
+                stream.finish(id);
+            }
+            Operation::ReceivedStopSending(sid, err_code) => {
+                stream.received_stop_sending(sid, err_code);
+            }
+            Operation::ReceivedReset(rs) => {
+                stream.received_reset(rs);
+            }
+            Operation::Reset(id) => {
+                stream.reset(id);
+            }
+        }
+    }
+});
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -25,6 +25,7 @@ tls-rustls = ["rustls", "webpki", "ring"]
 native-certs = ["rustls-native-certs"]
 
 [dependencies]
+arbitrary = { version = "0.4.5", features = ["derive"], optional = true }
 bytes = "0.5.2"
 ct-logs = { version = "0.7", optional = true }
 err-derive = "0.2.3"
--- a/quinn-proto/src/connection/streams.rs
+++ b/quinn-proto/src/connection/streams.rs
@@ -19,7 +19,7 @@ use crate::{
     Dir, Side, StreamId, TransportError, VarInt, MAX_STREAM_COUNT,
 };
 
-pub(crate) struct Streams {
+pub struct Streams {
     side: Side,
     // Set of streams that are currently open, or could be immediately opened by the peer
     send: HashMap<StreamId, Send>,
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,4 @@
+
+target
+corpus
+artifacts
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -42,7 +42,7 @@ mod spaces;
 use spaces::{PacketSpace, Retransmits, SentPacket};
 
 mod streams;
-use streams::Streams;
+pub use streams::Streams;
 pub use streams::{FinishError, ReadError, StreamEvent, UnknownStream, WriteError};
 
 mod timer;
--- a/quinn-proto/src/frame.rs
+++ b/quinn-proto/src/frame.rs
@@ -13,6 +13,9 @@ use crate::{
     RESET_TOKEN_SIZE,
 };
 
+#[cfg(feature = "arbitrary")]
+use arbitrary::Arbitrary;
+
 #[derive(Copy, Clone, Eq, PartialEq)]
 pub struct Type(u64);
 
@@ -732,6 +735,7 @@ impl<'a> Iterator for AckIter<'a> {
     }
 }
 
+#[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
 #[derive(Debug, Copy, Clone)]
 pub struct ResetStream {
     pub id: StreamId,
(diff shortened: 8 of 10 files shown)


fuzz/fuzz_targets/streams.rs
```diff
@@ -0,0 +1,22 @@
...
+use bytes::{Bytes};
+
+extern crate proto;
+use proto::connection::{Streams};
+use proto::frame::{Stream};
+use proto::{Side, Dir, StreamId};
+
+
+fuzz_target!(|data: &[u8]| {
+    let mut client = Streams::new(Side::Client, 128, 128, 1024 * 1024, 1024 * 1024, 1024 * 1024);
```
jafow, line 13: This is copy pasted straight from the streams test module and may not be the most useful target.

quinn-proto/Cargo.toml
```diff
@@ -25,6 +25,7 @@ tls-rustls = ["rustls", "webpki", "ring"]
 native-certs = ["rustls-native-certs"]
 
 [dependencies]
+arbitrary = { version = "0.4.5", features = ["derive"] }
```
Ralith, line 28: This dependency should be optional. We can then guard the derives directly on whether the dependency is enabled, which might theoretically be convenient for other downstream purposes.

quinn-proto/src/connection/streams.rs
```diff
@@ -19,7 +19,8 @@ use crate::{
     Dir, Side, StreamId, TransportError, VarInt, MAX_STREAM_COUNT,
 };
 
-pub(crate) struct Streams {
+#[doc(hidden)]
```
Ralith, line 22: Is this attribute still needed, given that the reexport is guarded?

quinn-proto/Cargo.toml
jafow: :+1: I've moved as an optional dependency behind the `fuzzing` feature. However, I don't fully understand this: 
> We can then guard the derives directly on whether the dependency is enabled

so I can update or change the dependency if this is suggesting a different approach from what I've added

quinn-proto/src/connection/streams.rs
jafow: nope you're quite right; can't make it _extra_ hidden can we? Removed.

quinn-proto/Cargo.toml
Ralith: I mean use `cfg(feature = "arbitrary")` rather than `cfg(fuzzing)`, and perhaps emit the `fuzzing` feature entirely since it looks like you're using a cfg flag for that (not totally clear on where that gets set)

jafow: Ah good I follow now, thank you. I've omitted the `fuzzing` feature. 

>  it looks like you're using a cfg flag for that (not totally clear on where that gets set)

This cfg gets set by `cargo-fuzz` - more [info is in their guide](https://rust-fuzz.github.io/book/cargo-fuzz/guide.html#cfgfuzzing)

fuzz/fuzz_targets/streams.rs
```diff
@@ -0,0 +1,22 @@
...
+    side: Side,
+    max_remote_uni: u64,
+    max_remote_bi: u64,
+    send_window: u64,
+    receive_window: u64,
+    stream_receive_window: u64
+}
+
+fuzz_target!(|data: StreamsParams| {
+    Streams::new(data.side, data.max_remote_uni, data.max_remote_bi, data.send_window, data.receive_window, data.stream_receive_window);
```
jafow, line 21: Here's just a dead simple target of `Streams`, and next week I can try to excersise some of the stream methods like `reset` etc on the control flow. 

On this target the fuzzer will discover OOM error after a few minutes; not sure how significant that is.  [I've pushed a branch on my fork](https://github.com/jafow/quinn/runs/964637707?check_suite_focus=true) that runs the fuzzer with a github action -- I expect that the CI container keeps a conservative limit on memory.

quinn-proto/Cargo.toml
```diff
@@ -23,8 +23,10 @@ certificate-transparency = ["ct-logs"]
 tls-rustls = ["rustls", "webpki", "ring"]
 # Trust the contents of the OS certificate store by default
 native-certs = ["rustls-native-certs"]
+unstructured-data = ["arbitrary"]
```
Ralith, line 26: What's this feature for?

fuzz/fuzz_targets/streams.rs
Ralith: > next week I can try to excersise some of the stream methods like reset etc on the control flow.

That would be great! I'd be particularly interested to see the various methods called in arbitrary sequence; maybe an `#[derive(Arbitrary)] enum Operation { Open, Reset(...), ... }` or similar?

> On this target the fuzzer will discover OOM error after a few minutes; not sure how significant that is

I'm guessing this happens when `max_remote_uni` or `max_remote_bi` are very large, since we preallocate state proportional to those. We should probably give those a modest maximum value since the number of streams shouldn't change behavior much at the margin. A quick hack to do that would be using `u8` instead of `u64` for the relevant fields in `StreamsParams`.

fuzz/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+
+[dependencies.proto]
+features = ["unstructured-data"]
+path = "../quinn-proto"
+package = "quinn-proto"
+version = "0.6.1"
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
```
Ralith, line 28: Why not make this a first-class member of the workspace?

fuzz/fuzz_targets/streamid.rs
```diff
@@ -0,0 +1,20 @@
...
+
+extern crate proto;
+use proto::{StreamId, Side, Dir};
+
+
+#[derive(Arbitrary, Debug)]
+struct StreamIdParams {
+    side: Side,
+    dir: Dir,
+    initiator: u64
```
Ralith, line 13:
```suggestion
    index: u64
```

quinn-proto/Cargo.toml
jafow: Ah, well this is me still misunderstanding above re: setting arbitrary behind a feature :-) 
https://github.com/djc/quinn/pull/831#discussion_r467640018. I'll take this to the chat to sort out

jafow: I took a try at moving the `arbitrary` as an optional dependency and dropping this feature, if I am understanding above correctly.

fuzz/Cargo.toml
```diff
@@ -23,10 +24,6 @@ path = "../quinn-proto"
 package = "quinn-proto"
 version = "0.6.1"
 
-# Prevent this from interfering with workspaces
```
jafow: initializing the fuzz target generates this `Cargo.toml` from a template, and it includes this line. 
Removed now and set as a workspace similar to `bench`, although in so doing I've made tremendous mess in the CI.

fuzz/Cargo.toml
jafow: This turned out to be quite the unexpected challenge! 

When trying to do `cargo build ---all-workspaces` the build fails from an unresolved import error:
```rust
error[E0432]: unresolved import `proto::fuzzing`
 --> fuzz/fuzz_targets/streams.rs:7:12
  |
7 | use proto::fuzzing::{ResetStream, Streams};
  |            ^^^^^^^ could not find `fuzzing` in `proto`
error: could not compile `quinn-fuzz`.
```

We re-export these in `quinn-proto` behind a `cfg(fuzzing)` so it isn't surprising to see that error. 
So then the problem seemed to be: how to set a `cfg` in a workspace. I have not found a solution for it!

I've tried:
- adding a build script to add the necessary rustc flags, e.g. `println!("cargo:rustc-cfg=fuzzing");` 
- addding a .cargo/config.toml file with rustflags config
- settings features block in fuzz/Cargo.toml 

I suppose another approach would be to change the commands in the github action, but I don't have a clear idea yet how that'd work. 

When building and running the fuzzer I am using tools from [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz), and they pass these flags seen here: https://github.com/rust-fuzz/cargo-fuzz/blob/4b7a6a674b3b6f7509c9568462b10cde5fcf841c/src/project.rs#L159
```rust
        let mut rustflags: String = "--cfg fuzzing \
                                     -Cpasses=sancov \
                                     -Cllvm-args=-sanitizer-coverage-level=4 \
                                     -Cllvm-args=-sanitizer-coverage-trace-compares \
                                     -Cllvm-args=-sanitizer-coverage-inline-8bit-counters \
                                     -Cllvm-args=-sanitizer-coverage-pc-table \
                                     -Clink-dead-code"
```

even adding these verbatim to build script did not change the failure of `cargo build --all-workspaces`. 
I would like to get this to build properly and will continue to ask around for ideas. 
In the meantime I will move fuzzing out of the primary workspace and get the CI back upright.

Ralith: In that case I think it's fine for this to not be a workspace member; I clearly underestimated the significance of the barrier here!

fuzz/fuzz_targets/streams.rs
```diff
@@ -0,0 +1,71 @@
...
+    stream_receive_window: u16,
+    dir: Dir,
+}
+
+#[derive(Arbitrary, Debug)]
+enum Operation<A>
+where
+    A: Arbitrary,
+{
+    Open(Vec<A>),
```
Ralith, line 26: Is there a benefit to this being generic rather than hardcoding `u8`?

fuzz/fuzz_targets/streams.rs
```diff
@@ -0,0 +1,71 @@
...
+    Accept(Dir),
+    Finish(StreamId),
+    ReceivedStopSending((StreamId, VarInt)),
```
Ralith, line 29:
```suggestion
    ReceivedStopSending(StreamId, VarInt),
```
would be a bit more conventional style--unless I'm missing the motivation here?

Ralith (commented):
This looks almost ready to me! Is there anything else you wanted to cover before merging?

quinn-proto/Cargo.toml
Ralith: LGTM!

fuzz/fuzz_targets/streams.rs
jafow: no you're right that's not needed - removing.

fuzz/fuzz_targets/streams.rs
jafow: :+1: no motivation just poor style :-) corrected.

fuzz/fuzz_targets/streams.rs
jafow: No that is no longer useful; removed

fuzz/fuzz_targets/streams.rs
jafow: > This looks almost ready to me! Is there anything else you wanted to cover before merging?

Very good! Thank you for the help and for making `quinn` hospitable to new comers.

jafow:
@Ralith I've moved PR from draft to ready for review -- if I should rebase and squash please let me know, otherwise I will leave with you.

Ralith:
A rebase and squash would be great!

fuzz/fuzz_targets/streams.rs
```diff
@@ -0,0 +1,68 @@
...
+        params.send_window.into(),
+        params.receive_window.into(),
+        params.stream_receive_window.into(),
+    );
+
+    for operation in operations {
+        match operation {
+            Operation::Open(v) => {
+                let mut data = Unstructured::new(&v[..]);
+                if let Ok(tp) = data.arbitrary() {
```
Ralith, line 47: I see that `v: Vec<u8>` here is only used to invent a `TransportParameters`. Could we avoid going through raw bytes here, and get the `TransportParameters` directly?

Also, `Streams` isn't expected to handle dynamically changing `TransportParameters`, so it would probably make more sense to generate it once via `StreamParams` and reference it as needed, rather than coming up with a fresh one on every `Open`.

Ralith (changes requested):

fuzz/fuzz_targets/streams.rs
jafow: yep this makes sense, good to know. 

I've squash and rebased all but the last commit that drops the raw bytes and `TransportParameters` only once rather than each call to open. If this is the right track I will squash it in with the rest.

Ralith (approved):
Great work, thanks!

fuzz/fuzz_targets/streams.rs
```diff
@@ -0,0 +1,66 @@
...
+    send_window: u16,
+    receive_window: u16,
+    stream_receive_window: u16,
+    dir: Dir,
+    transport_params: TransportParameters
+}
+
+#[derive(Arbitrary, Debug)]
+enum Operation
+{
```
Ralith, line 24: Non-blocking style nit:
```suggestion
enum Operation {
```

Ralith (approved):

djc:
Thanks for working on this!

PR #833, opened 2020-07-29
[#783] Encrypt retry token

### LINKED ISSUE #783: Encrypt retry token
When stateless retries are enabled, Quinn embeds some state in the retry token and authenticates it with a HMAC. This information is not secret, so the current approach works, but it would be more in the spirit of QUIC to use strong encryption to prevent other parties from coming to rely on that data, creating a stability hazard.

The stateless nature of retry tokens requires some care if we want to expose absolutely zero information to other parties, as there's no way to bundle an AEAD nonce. One solution would be:

- Generate a unique AEAD key for each retry token by using a HKDF from a fixed master key on a fixed-length sequence of random data.
- Encrypt the token payload using a nonce of 0. This is safe because we never reuse a key.
- Send the concatenation of the random data, the encrypted payload, and the AEAD tag as the token. An outside party cannot distinguish any of this from random data.
- Recover the key for decryption by re-running the HKDF on the random data prefix.

kansi:
**DISCLAIMER**: I am a first time contributor (Rust and QUIC).

I would like some feedback on the direction of this :slightly_smiling_face: 

Resolves #783 , following changes have been introduced,

- `master_key` has been added to `ServerConfig` for aiding AEAD key derivation
- A new trait `AeadKey` has been introduced which provided following API,
  - `from_hkdf` : derive a AEAD key using `master_key` and some `random_bytes`
  - `seal`: seal a message with the given AEAD key
  - `open`: open a message with a given AEAD key

~TODO~
- `[test]` scenario when a garbage retry token is received and the opening operation fails

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -8,7 +8,7 @@ use crate::crypto::types::{Certificate, CertificateChain, PrivateKey};
 use crate::{
     cid_generator::{ConnectionIdGenerator, RandomConnectionIdGenerator},
     congestion,
-    crypto::{self, ClientConfig as _, HmacKey as _, ServerConfig as _},
+    crypto::{self, ClientConfig as _, HandshakeTokenKey as _, HmacKey as _, ServerConfig as _},
     VarInt, VarIntBoundsExceeded,
 };
 
@@ -413,8 +413,9 @@ where
     /// Must be set to use TLS 1.3 only.
     pub crypto: S::ServerConfig,
 
-    /// Private key used to authenticate data included in handshake tokens.
-    pub(crate) token_key: Arc<S::HmacKey>,
+    /// Used to generate one-time AEAD keys to protect handshake tokens
+    pub(crate) token_key: Arc<S::HandshakeTokenKey>,
+
     /// Whether to require clients to prove ownership of an address before committing resources.
     ///
     /// Introduces an additional round-trip to the handshake to make denial of service attacks more difficult.
@@ -438,13 +439,13 @@ impl<S> ServerConfig<S>
 where
     S: crypto::Session,
 {
-    /// Create a default config with a particular `token_key`
-    pub fn new(token_key: S::HmacKey) -> Self {
+    /// Create a default config with a particular `master_key`
+    pub fn new(prk: S::HandshakeTokenKey) -> Self {
         Self {
             transport: Arc::new(TransportConfig::default()),
             crypto: S::ServerConfig::new(),
 
-            token_key: Arc::new(token_key),
+            token_key: Arc::new(prk),
             use_stateless_retry: false,
             retry_token_lifetime: 15_000_000,
 
@@ -455,8 +456,8 @@ where
     }
 
     /// Private key used to authenticate data included in handshake tokens.
-    pub fn token_key(&mut self, value: &[u8]) -> Result<&mut Self, ConfigError> {
-        self.token_key = Arc::new(S::HmacKey::new(value)?);
+    pub fn token_key(&mut self, master_key: &[u8]) -> Result<&mut Self, ConfigError> {
+        self.token_key = Arc::new(S::HandshakeTokenKey::from_secret(&master_key));
         Ok(self)
     }
 
@@ -529,12 +530,10 @@ where
     fn default() -> Self {
         let rng = &mut rand::thread_rng();
 
-        let mut token_key = vec![0; S::HmacKey::KEY_LEN];
-        rng.fill_bytes(&mut token_key);
-        Self::new(
-            S::HmacKey::new(&token_key)
-                .expect("HMAC key rejected random bytes; use ServerConfig::new instead"),
-        )
+        let mut master_key = [0u8; 64];
+        rng.fill_bytes(&mut master_key);
+
+        Self::new(S::HandshakeTokenKey::from_secret(&master_key))
     }
 }
 
--- a/quinn-proto/src/crypto.rs
+++ b/quinn-proto/src/crypto.rs
@@ -38,6 +38,8 @@ pub trait Session: Send + Sized {
     type ClientConfig: ClientConfig<Self>;
     /// Type used to sign various values
     type HmacKey: HmacKey;
+    /// Key used to generate one-time-use handshake token keys
+    type HandshakeTokenKey: HandshakeTokenKey;
     /// Type of keys used to protect packet headers
     type HeaderKey: HeaderKey;
     /// Type used to represent packet protection keys
@@ -210,3 +212,26 @@ pub trait HmacKey: Send + Sized + Sync {
 /// This error occurs if the requested output length is too large.
 #[derive(Debug, PartialEq, Eq)]
 pub struct ExportKeyingMaterialError;
+
+/// A pseudo random key for HKDF
+pub trait HandshakeTokenKey: Send + Sized + Sync {
+    /// AEAD key type
+    type AeadKey: AeadKey;
+
+    /// Derive AEAD using hkdf
+    fn aead_from_hkdf(&self, random_bytes: &[u8]) -> Self::AeadKey;
+    /// Method to build pseudo random key from existing bytes
+    fn from_secret(secret: &[u8]) -> Self;
+}
+
+/// A key for sealing data with AEAD-based algorithms
+pub trait AeadKey {
+    /// Length of AEAD Key
+    const KEY_LEN: usize;
+
+    // fn from_hkdf(master_key: &impl PseudoRandomKey, random_bytes: &[u8]) -> Self;
+    /// Method for sealing message `data`
+    fn seal(&self, data: &mut Vec<u8>, additional_data: &[u8]) -> Result<(), ()>;
+    /// Method for opening a sealed message `data`
+    fn open<'a>(&self, data: &'a mut [u8], additional_data: &[u8]) -> Result<&'a mut [u8], ()>;
+}
--- a/quinn-proto/src/crypto/ring.rs
+++ b/quinn-proto/src/crypto/ring.rs
@@ -1,4 +1,4 @@
-use ring::{aead, hmac};
+use ring::{aead, hkdf, hmac};
 
 use crate::{
     config::ConfigError,
@@ -68,5 +68,41 @@ impl crypto::HmacKey for hmac::Key {
 
     fn verify(&self, data: &[u8], signature: &[u8]) -> Result<(), ()> {
         hmac::verify(self, data, signature).map_err(|_| ())
+    }
+}
+
+impl crypto::HandshakeTokenKey for hkdf::Prk {
+    type AeadKey = ring::aead::LessSafeKey;
+
+    fn aead_from_hkdf(&self, random_bytes: &[u8]) -> Self::AeadKey {
+        let mut key_buffer = [0u8; 32];
+        let info = [random_bytes];
+        let okm = self.expand(&info, hkdf::HKDF_SHA256).unwrap();
+
+        okm.fill(&mut key_buffer).unwrap();
+
+        let key = aead::UnboundKey::new(&aead::AES_256_GCM, &key_buffer).unwrap();
+        Self::AeadKey::new(key)
+    }
+
+    fn from_secret(bytes: &[u8]) -> Self {
+        hkdf::Salt::new(hkdf::HKDF_SHA256, &[]).extract(bytes)
+    }
+}
+
+impl crypto::AeadKey for aead::LessSafeKey {
+    const KEY_LEN: usize = 32;
+
+    fn seal(&self, data: &mut Vec<u8>, additional_data: &[u8]) -> Result<(), ()> {
+        let aad = ring::aead::Aad::from(additional_data);
+        let zero_nonce = ring::aead::Nonce::assume_unique_for_key([0u8; 12]);
+        self.seal_in_place_append_tag(zero_nonce, aad, data)
+            .map_err(|_| ())
+    }
+
+    fn open<'a>(&self, data: &'a mut [u8], additional_data: &[u8]) -> Result<&'a mut [u8], ()> {
+        let aad = ring::aead::Aad::from(additional_data);
+        let zero_nonce = ring::aead::Nonce::assume_unique_for_key([0u8; 12]);
+        self.open_in_place(zero_nonce, aad, data).map_err(|_| ())
     }
 }
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -548,9 +548,13 @@ where
         let (retry_src_cid, orig_dst_cid) = if server_config.use_stateless_retry {
             if token.is_empty() {
                 // First Initial
+                let mut random_bytes = vec![0u8; RetryToken::RANDOM_BYTES_LEN];
+                self.rng.fill_bytes(&mut random_bytes);
+
                 let token = RetryToken {
                     orig_dst_cid: dst_cid,
                     issued: SystemTime::now(),
+                    random_bytes: &random_bytes,
                 }
                 .encode(&*server_config.token_key, &remote, &temp_loc_cid);
                 let mut buf = Vec::new();
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -557,11 +557,13 @@ where
                     random_bytes: &random_bytes,
                 }
                 .encode(&*server_config.token_key, &remote, &temp_loc_cid);
-                let mut buf = Vec::new();
+
                 let header = Header::Retry {
                     src_cid: temp_loc_cid,
                     dst_cid: src_cid,
                 };
+
+                let mut buf = Vec::new();
                 let encode = header.encode(&mut buf);
                 buf.put_slice(&token);
                 buf.extend_from_slice(&S::retry_tag(&dst_cid, &buf));
--- a/quinn-proto/src/token.rs
+++ b/quinn-proto/src/token.rs
@@ -8,7 +8,7 @@ use bytes::BufMut;
 
 use crate::{
     coding::{BufExt, BufMutExt},
-    crypto::HmacKey,
+    crypto::{AeadKey, HandshakeTokenKey, HmacKey},
     shared::ConnectionId,
     RESET_TOKEN_SIZE,
 };
@@ -19,20 +19,44 @@ use crate::{
 // - AEAD nonce is always set to 0
 // in other words, for each ticket, use different key derived from random using HKDF
 
-pub struct RetryToken {
+pub struct RetryToken<'a> {
     /// The destination connection ID set in the very first packet from the client
     pub orig_dst_cid: ConnectionId,
     /// The time at which this token was issued
     pub issued: SystemTime,
+    /// Random bytes for deriving AEAD key
+    pub random_bytes: &'a [u8],
 }
 
-impl RetryToken {
+impl<'a> RetryToken<'a> {
+    const MAX_ADDITIONAL_DATA_SIZE: usize = 39; // max(ipv4, ipv6) + port + retry_src_cid
+    pub const RANDOM_BYTES_LEN: usize = 32;
+
+    fn put_additional_data<'b>(
+        address: &SocketAddr,
+        retry_src_cid: &ConnectionId,
+        additional_data: &'b mut [u8],
+    ) -> &'b [u8] {
+        let mut cursor = &mut additional_data[..];
+        match address.ip() {
+            IpAddr::V4(x) => cursor.put_slice(&x.octets()),
+            IpAddr::V6(x) => cursor.put_slice(&x.octets()),
+        }
+        cursor.write(address.port());
+        retry_src_cid.encode_long(&mut cursor);
+
+        let size = Self::MAX_ADDITIONAL_DATA_SIZE - cursor.len();
+        &additional_data[..size]
+    }
+
     pub fn encode(
         &self,
-        key: &impl HmacKey,
+        key: &impl HandshakeTokenKey,
         address: &SocketAddr,
         retry_src_cid: &ConnectionId,
     ) -> Vec<u8> {
+        let aead_key = key.aead_from_hkdf(self.random_bytes);
+
         let mut buf = Vec::new();
 
         self.orig_dst_cid.encode_long(&mut buf);
@@ -43,46 +67,45 @@ impl RetryToken {
                 .unwrap_or(0),
         );
 
-        let signature_pos = buf.len();
-        match address.ip() {
-            IpAddr::V4(x) => buf.put_slice(&x.octets()),
-            IpAddr::V6(x) => buf.put_slice(&x.octets()),
-        }
-        buf.write(address.port());
-        retry_src_cid.encode_long(&mut buf);
+        let mut additional_data = [0u8; Self::MAX_ADDITIONAL_DATA_SIZE];
+        let additional_data =
+            Self::put_additional_data(address, retry_src_cid, &mut additional_data);
+        aead_key.seal(&mut buf, additional_data).unwrap();
 
-        let signature = key.sign(&buf);
-        // No reason to actually encode the IP in the token, since we always have the remote addr for an incoming packet.
-        buf.truncate(signature_pos);
-        buf.extend_from_slice(signature.as_ref());
-        buf
+        let mut token = Vec::new();
+        token.put_slice(self.random_bytes);
+        token.put_slice(&buf);
+        token
     }
 
     pub fn from_bytes(
-        key: &impl HmacKey,
+        key: &impl HandshakeTokenKey,
         address: &SocketAddr,
         retry_src_cid: &ConnectionId,
-        data: &[u8],
+        raw_token_bytes: &'a [u8],
     ) -> Result<Self, ()> {
+        if raw_token_bytes.len() < Self::RANDOM_BYTES_LEN {
+            // Invalid length
+            return Err(());
+        }
+
+        let random_bytes = &raw_token_bytes[..Self::RANDOM_BYTES_LEN];
+        let aead_key = key.aead_from_hkdf(random_bytes);
+        let mut sealed_token = raw_token_bytes[Self::RANDOM_BYTES_LEN..].to_vec();
+
+        let mut additional_data = [0u8; Self::MAX_ADDITIONAL_DATA_SIZE];
+        let additional_data =
+            Self::put_additional_data(address, retry_src_cid, &mut additional_data);
+        let data = aead_key.open(&mut sealed_token, additional_data)?;
         let mut reader = io::Cursor::new(data);
 
         let orig_dst_cid = ConnectionId::decode_long(&mut reader).ok_or(())?;
         let issued = UNIX_EPOCH + Duration::new(reader.get::<u64>().map_err(|_| ())?, 0);
 
-        let signature_start = reader.position() as usize;
-        let mut buf = Vec::new();
-        buf.put_slice(&data[0..signature_start]);
-        match address.ip() {
-            IpAddr::V4(x) => buf.put_slice(&x.octets()),
-            IpAddr::V6(x) => buf.put_slice(&x.octets()),
-        }
-        buf.write(address.port());
-        retry_src_cid.encode_long(&mut buf);
-
-        key.verify(&buf, &data[signature_start..]).map_err(|_| ())?;
         Ok(Self {
             orig_dst_cid,
             issued,
+            random_bytes,
         })
     }
 }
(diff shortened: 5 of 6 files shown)


quinn-proto/src/config.rs
```diff
@@ -428,12 +430,13 @@ where
     S: crypto::Session,
 {
     /// Create a default config with a particular `token_key`
-    pub fn new(token_key: S::HmacKey) -> Self {
+    pub fn new(token_key: S::HmacKey, master_key: Vec<u8>) -> Self {
```
djc, line 433: I think we'd only want to ask for one key here. Presumably the HMAC key can be derived from the master key if necessary.

kansi:
> Hey, thanks for taking a look at this! That's some pretty involved code for your first Rust project. +1

Thanks! :slightly_smiling_face:  :releaved:
 
> Can you give more context on what the design is here? Which of the proposals as made by @Ralith in the original issue did you follow, and why?

Sure, as I understanding the issue #783 proposes to use AEAD instead of HMAC for token authentication, namely following proposal has been made,


> Generate a unique AEAD key for each retry token by using a HKDF from a fixed master key on a fixed-length sequence of random data.

A `master_key` is stored in the `ServerConfig` and this is later use with some randomly generated bytes, based on whenever a new Initial is received, to generate a new AEAD key.

> Encrypt the token payload using a nonce of 0. This is safe because we never reuse a key.

`ZeroNonceSequence` has been implemented and provided then created the opening and sealing AEAD keys

> Send the concatenation of the random data, the encrypted payload, and the AEAD tag as the token. An outside party cannot distinguish any of this from random data.

The `encode` method in `RetryToken` seals with token with the provided AEAD key and prefixes the random bytes which were used to generate the AEAD key

> Recover the key for decryption by re-running the HKDF on the random data prefix.

`extract_random_bytes` method in `RetryToken` is used to extract the random bytes from the retry token and then used to derive the AEAD key. Later the AEAD key is used in `from_bytes` method to `open` the token and returns `RetryToken`


I chose to place `random_bytes` in `RetryToken` struct as the issues suggest that each RetryToken will have a different AEAD key.

quinn-proto/src/crypto/ring.rs
```diff
@@ -70,3 +70,72 @@ impl crypto::HmacKey for hmac::Key {
...
+    key: &[u8],
+) -> K {
+    let key = aead::UnboundKey::new(algorithm, key).unwrap();
+    K::new(key, ZeroNonceSequence::new())
+}
+
+pub struct Aes256GcmKey {
+    opening_key: aead::OpeningKey<ZeroNonceSequence>,
+    sealing_key: aead::SealingKey<ZeroNonceSequence>,
+}
```
Ralith, line 85: This can be replaced entirely with *ring*'s `LessSafeKey`. That'll avoid duplicating storage and make the `ZeroNonceSequence` helper unnecessary. Further, `open` and `seal` will then be able to take `&self` rather than `&mut self`.

quinn-proto/src/endpoint.rs
```diff
@@ -537,18 +537,29 @@ where
         let (retry_src_cid, orig_dst_cid) = if server_config.use_stateless_retry {
             if token.is_empty() {
                 // First Initial
+                let rng = &mut rand::thread_rng();
```
Ralith, line 540: Use `self.rng` instead.

quinn-proto/src/crypto.rs
```diff
@@ -186,3 +188,16 @@ pub trait HmacKey: Send + Sized + Sync {
...
+pub trait AeadKey {
+    /// Length of AEAD Key
+    const KEY_LEN: usize;
+
+    /// Method for deriving a fixed AEAD key using a `master_key` and `random_bytes`
+    fn derive_fixed(master_key: &[u8], random_bytes: &[u8]) -> Self;
+    /// Method for sealing message `data`
+    fn seal(&mut self, data: &mut Vec<u8>) -> Result<(), ()>;
+    /// Method for opening a sealed message `data`
+    fn open<'a>(&mut self, data: &'a mut Vec<u8>) -> Result<&'a mut [u8], ()>;
```
Ralith, line 202:
```suggestion
    fn open<'a>(&mut self, data: &'a mut [u8]) -> Result<&'a mut [u8], ()>;
```
This doesn't need to be a `Vec`, and using a slice lets us save an alloc and copy.

quinn-proto/src/config.rs
```diff
@@ -404,6 +404,8 @@ where
 
     /// Private key used to authenticate data included in handshake tokens.
     pub(crate) token_key: Arc<S::HmacKey>,
+
+    pub(crate) master_key: Arc<Vec<u8>>,
```
Ralith, line 408: We don't need two separate keys; after these changes `token_key` is unused. Rather than replacing `token_key`, I suggest using it as the master key directly.

quinn-proto/src/crypto.rs
```diff
@@ -186,3 +188,16 @@ pub trait HmacKey: Send + Sized + Sync {
...
     fn verify(&self, data: &[u8], signature: &[u8]) -> Result<(), ()>;
 }
+
+/// A key for sealing data with AEAD-based algorithms
+pub trait AeadKey {
+    /// Length of AEAD Key
+    const KEY_LEN: usize;
+
+    /// Method for deriving a fixed AEAD key using a `master_key` and `random_bytes`
+    fn derive_fixed(master_key: &[u8], random_bytes: &[u8]) -> Self;
```
Ralith, line 198: Let's name this `from_hkdf`, and accept a typed `HmacKey` rather than opaque key bytes. The `HmacKey` type can be obtained from a type parameter on `AeadKey`.

quinn-proto/src/crypto.rs
```diff
@@ -38,6 +38,8 @@ pub trait Session: Send + Sized {
     type ClientConfig: ClientConfig<Self>;
     /// Type used to sign various values
     type HmacKey: HmacKey;
+    /// Type used for sealing messages
```
Ralith, line 41:
```suggestion
    /// Key used for protecting handshake tokens
```

quinn-proto/src/crypto.rs
kansi: I didn't fully understand this comment, could you please elaborate? 
It seems that you are suggesting the following,

```rust
    fn from_hkdf(master_key: HmacKey) -> Self; 
```

and then do an implementation 

```rust
impl crypto::AeadKey for ... {
    type HmacKey = hmac::Key;
    fn from_hkdf(master_key: HmacKey, random_bytes: &[u8]) -> Self {
       // How to derive using HmacKey instead of master_key bytes?
    }
}
```
Is that correct?

As as I understand from the docs of [`hkdf`](https://briansmith.org/rustdoc/ring/hkdf/index.html) module, we need the opaque key bytes so that we can build the output key material `Okm`.

Looking at the `hkdf` module APIs, I am not sure how `Okm` could be built from `hmac::Key`. 

In an earlier comment you said,

> We don't need two separate keys; after these changes token_key is unused. Rather than replacing token_key, I suggest using it as the master key directly.

Are you suggesting to combine `token_key.inner` and `token_key.outer` to build the IKM?

Ralith: > Looking at the hkdf module APIs, I am not sure how Okm could be built from hmac::Key.

Ah, you're right. What we should actually be using here is a(n abstraction over a) `ring::hkdf::Salt`; I was confusing `HmacKey` with that.

> Are you suggesting to combine token_key.inner and token_key.outer to build the IKM?

I'm not sure what `inner` and `outer` here are. My point is only that we only need one persistent key, and `token_key` can be what we call it.

Ralith: Er, rather, the `Prk` is the stable value we should be holding on to.

kansi: > My point is only that we only need one persistent key, and token_key can be what we call it.

> Er, rather, the `Prk` is the stable value we should be holding on to.

Makes sense :+1: , then `token_key`  value would be a `Prk` which will be used to derive `Okm`, and that can be used to drive the `hmac` and `aead` keys :+1:

Ralith: Do we need an actual `hmac` key at all? The AEAD takes care of authentication on its own, and the `Prk` is sufficient for generating an AEAD key.

kansi: > Do we need an actual hmac key at all? 

Thank you for putting this question, I didn't realize that the `token_key` wasn't used for any other place apart for generating message auth codes.

quinn-proto/src/config.rs
```diff
@@ -444,10 +445,10 @@ where
...
 
     /// Private key used to authenticate data included in handshake tokens.
-    pub fn token_key(&mut self, value: &[u8]) -> Result<&mut Self, ConfigError> {
-        self.token_key = Arc::new(S::HmacKey::new(value)?);
-        Ok(self)
-    }
+    // pub fn token_key(&mut self, value: &[u8]) -> Result<&mut Self, ConfigError> {
+    //     self.token_key = Arc::new(S::HmacKey::new(value)?);
+    //     Ok(self)
+    // }
```
kansi, line 451: Is this API used anywhere? The test didn't fail when I commented it out.
Should I update it or is it dead code?

Ralith: It's part of the public interface and should be retained. Downstream code might want to specify it to allow servers behind a load balancer to statelessly accept tickets issued by eachother, or to allow a server to accept its own tokens after a restart.

quinn-proto/src/config.rs
```diff
@@ -518,12 +519,10 @@ where
     fn default() -> Self {
         let rng = &mut rand::thread_rng();
 
-        let mut token_key = vec![0; S::HmacKey::KEY_LEN];
-        rng.fill_bytes(&mut token_key);
-        Self::new(
-            S::HmacKey::new(&token_key)
-                .expect("HMAC key rejected random bytes; use ServerConfig::new instead"),
-        )
+        let mut master_key = vec![0u8; 64];
```
kansi, line 522: I would like to know what would be a better way of doing this such that the size of the vector is not hard coded?

quinn-proto/src/config.rs
```diff
@@ -444,8 +445,8 @@ where
     }
 
     /// Private key used to authenticate data included in handshake tokens.
-    pub fn token_key(&mut self, value: &[u8]) -> Result<&mut Self, ConfigError> {
-        self.token_key = Arc::new(S::HmacKey::new(value)?);
+    pub fn token_key(&mut self, master_key: &[u8]) -> Result<&mut Self, ConfigError> {
```
kansi, line 448: `master_key` here is just constrained to `&[u8]` but actually we expect `64` bytes.

Ralith: Does this actually have to be 64 bytes? [The underlying method's docs](https://docs.rs/ring/0.16.9/ring/hkdf/struct.Salt.html#method.extract) don't seem to indicate any particular input size, even in the cited RFC.

quinn-proto/src/config.rs
Ralith: We could use an associated constant on the trait, or modify the trait interface to take a fixed-size array rather than a slice. However, it's not clear to me that we actually need any particular length (see other comment) and in that case 64 bytes seems like a fine default.

quinn-proto/src/config.rs
kansi: You are right, there is no reason for the `master_key` to be `64` bytes, I am not sure why 64 was stuck in my head.
Seems like this implementation is sufficient then :+1:

quinn-proto/src/config.rs
```diff
@@ -402,8 +402,9 @@ where
     /// Must be set to use TLS 1.3 only.
     pub crypto: S::ServerConfig,
 
-    /// Private key used to authenticate data included in handshake tokens.
-    pub(crate) token_key: Arc<S::HmacKey>,
+    /// Pseudo random key used for generating AEAD key for tokens
```
Ralith, line 405:
```suggestion
    /// Used to generate one-time AEAD keys to protect handshake tokens
```

quinn-proto/src/config.rs
Ralith, line 522:
```suggestion
        let mut master_key = [0u8; 64];
```
Since we can use a hardcoded length here, there's no need for a heap allocation.

quinn-proto/src/crypto.rs
```diff
@@ -186,3 +190,24 @@ pub trait HmacKey: Send + Sized + Sync {
     /// Method for verifying a message
     fn verify(&self, data: &[u8], signature: &[u8]) -> Result<(), ()>;
 }
+
+/// A pseudo random key for HKDF
+pub trait PseudoRandomKey: Send + Sized + Sync {
+    /// Method to build pseudo random key from existing bytes
+    fn from_secret(secret: &[u8]) -> Self;
+    /// Generate key using prk and fill the key_buffer
+    fn to_bytes(&self, random_bytes: &[u8], key_buffer: &mut [u8]);
```
Ralith, line 199:
```suggestion
    fn expand(&self, random_bytes: &[u8], key_buffer: &mut [u8]);
```
Let's use the RFC terminology. `to_bytes` makes me think of e.g. [`str::as_bytes`](https://doc.rust-lang.org/std/primitive.str.html#method.as_bytes), which is a lossless, reversible encoding of the parameter--definitely not what we're going for here!

quinn-proto/src/endpoint.rs
```diff
@@ -537,18 +537,27 @@ where
...
-                }
-                .encode(&*server_config.token_key, &remote, &temp_loc_cid);
-                let mut buf = Vec::new();
+                    random_bytes: &random_bytes,
+                };
+
+                // Derive AeadKey
+                let aead_key = S::AeadKey::from_hkdf(&*server_config.token_key, token.random_bytes);
+
+                let encoded_token = token.encode(&aead_key, &remote, &temp_loc_cid);
```
Ralith, line 552: Can we move the key derivation inside `RetryToken::encode` to help keep things encapsulated?

quinn-proto/src/endpoint.rs
```diff
@@ -560,7 +569,12 @@ where
                 return None;
             }
 
-            match RetryToken::from_bytes(&*server_config.token_key, &remote, &dst_cid, &token) {
+            let random_bytes = RetryToken::extract_random_bytes(&token);
+
+            // Derive AeadKey
+            let aead_key = S::AeadKey::from_hkdf(&*server_config.token_key, random_bytes);
+
+            match RetryToken::from_bytes(&aead_key, &remote, &dst_cid, &token) {
```
Ralith, line 577: And similarly for decoding. This should obviate the need to expose `extract_random_bytes`.

quinn-proto/src/token.rs
```diff
@@ -140,27 +146,78 @@ mod test {
...
-        let mut key = [0; 64];
-        rand::thread_rng().fill_bytes(&mut key);
-        let key = <hmac::Key as HmacKey>::new(&key).unwrap();
+        let mut master_key = [0; 64];
+        rand::thread_rng().fill_bytes(&mut master_key);
+
+        let mut random_bytes = [0; 32];
+        rand::thread_rng().fill_bytes(&mut random_bytes);
+
+        let rng = &mut rand::thread_rng();
```
Ralith, line 162: If we're going to introduce a variable for this, might as well use it in place of the repeated `&mut rand::thread_rng()` below.

quinn-proto/src/token.rs
```diff
@@ -43,46 +51,44 @@ impl RetryToken {
...
         match address.ip() {
             IpAddr::V4(x) => buf.put_slice(&x.octets()),
             IpAddr::V6(x) => buf.put_slice(&x.octets()),
         }
         buf.write(address.port());
         retry_src_cid.encode_long(&mut buf);
 
-        let signature = key.sign(&buf);
-        // No reason to actually encode the IP in the token, since we always have the remote addr for an incoming packet.
-        buf.truncate(signature_pos);
```
Ralith, line 56: We still shouldn't be transmitting the IP/port data, encrypted or not; it's just not needed. The added encryption means we can't use the cheeky virtual trailing data approach; instead, the IP and port should be passed in to the encryption/decryption process as additionally authenticated data (AAD).

quinn-proto/src/token.rs
```diff
@@ -43,46 +51,44 @@ impl RetryToken {
...
-        let mut buf = Vec::new();
-        buf.put_slice(&data[0..signature_start]);
-        match address.ip() {
-            IpAddr::V4(x) => buf.put_slice(&x.octets()),
-            IpAddr::V6(x) => buf.put_slice(&x.octets()),
-        }
-        buf.write(address.port());
-        retry_src_cid.encode_long(&mut buf);
-
-        key.verify(&buf, &data[signature_start..]).map_err(|_| ())?;
```
Ralith, line 82: We must retain the logic to verify source address and retry source CID to defend against spoofing. That should be done with AAD as described in the above comment.

quinn-proto/src/token.rs
```diff
@@ -19,20 +19,26 @@ use crate::{
...
+    pub const RANDOM_BYTES_LEN: usize = 32;
+
     pub fn encode(
         &self,
-        key: &impl HmacKey,
+        key: &impl PseudoRandomKey,
         address: &SocketAddr,
         retry_src_cid: &ConnectionId,
     ) -> Vec<u8> {
+        let aead_key: ring::aead::LessSafeKey = AeadKey::from_hkdf(key, self.random_bytes);
```
kansi, line 40: I would like to know how to do this better i.e. I know that `endpoint.rs` knows the actual type i.e. `S::AeadKey` but I am not sure how to use that here?

Ralith: Would it make sense for `encode` and `from_bytes` to take an explicit `K: AeadKey` type parameter?

djc: I think you'll have to use an associated type in either `PseudoRandomKey` or `AeadKey` to link one to the other. In this case, I think it might make sense to add a method to `PseudoRandomKey` that constructs an `AeadKey` (rather than the other way around), and then `PseudoRandomKey` can also have an associated type for the type of the `AeadKey` it builds.

We don't usually use the `impl Trait` in argument position, instead using named bound parameters, I think it would make sense to do so here, too (so give `RetryToken` a `P` argument?).

quinn-proto/src/crypto/ring.rs
```diff
@@ -70,3 +70,42 @@ impl crypto::HmacKey for hmac::Key {
...
+
+impl crypto::PseudoRandomKey for hkdf::Prk {
+    fn from_secret(bytes: &[u8]) -> Self {
+        let salt = hkdf::Salt::new(hkdf::HKDF_SHA256, &[]);
+        salt.extract(bytes)
+    }
+
+    fn expand(&self, random_bytes: &[u8], mut key_buffer: &mut [u8]) {
+        let info = [random_bytes];
+        let okm = self.expand(&info, hkdf::HKDF_SHA256).unwrap();
```
djc, line 82: I think we try to avoid unwrapping in nested library code like this. Return a `Result` instead (potentially just `Result<(), ()>`), which can be propagated outward until it reaches a level where the caller knows how to deal with it.

quinn-proto/src/crypto/ring.rs
```diff
@@ -70,3 +70,42 @@ impl crypto::HmacKey for hmac::Key {
         hmac::verify(self, data, signature).map_err(|_| ())
     }
 }
+
+impl crypto::PseudoRandomKey for hkdf::Prk {
+    fn from_secret(bytes: &[u8]) -> Self {
+        let salt = hkdf::Salt::new(hkdf::HKDF_SHA256, &[]);
+        salt.extract(bytes)
```
djc, line 77: Style nit:

```suggestion
        hkdf::Salt::new(hkdf::HKDF_SHA256, &[]).extract(bytes)
```

quinn-proto/src/token.rs
```diff
@@ -43,46 +49,55 @@ impl RetryToken {
...
-        let orig_dst_cid = ConnectionId::decode_long(&mut reader).ok_or(())?;
-        let issued = UNIX_EPOCH + Duration::new(reader.get::<u64>().map_err(|_| ())?, 0);
+        let random_bytes = &raw_token_bytes[..Self::RANDOM_BYTES_LEN];
+        let aead_key: ring::aead::LessSafeKey = AeadKey::from_hkdf(key, random_bytes);
+        let mut sealed_token = raw_token_bytes[Self::RANDOM_BYTES_LEN..].to_vec();
 
-        let signature_start = reader.position() as usize;
-        let mut buf = Vec::new();
-        buf.put_slice(&data[0..signature_start]);
+        let mut additional_data = Vec::new();
```
djc, line 83: Since it looks like the maximum length of the additional data is known in advance and is pretty limited, it'd be nice to avoid the heap allocation here and use an array instead.

djc:
(Also, since the commits so far don't seem to be logically distinct, we'd prefer if you squash them into a single one.)

quinn-proto/src/token.rs
Ralith: I'm actually a fan of argument position `impl Trait` for the concision when a type parameter never needs to be explicitly specified/otherwise qualified, though I don't mind much either way.

kansi: @djc  As far as I can understand your comment it seems that you are suggesting the following,

```rust
pub trait PseudoRandomKey: Send + Sized + Sync {
    type AeadKey: AeadKey;
    fn aead_from_hkdf(&self) -> Self::AeadKey;
    ...
}
```

```rust
impl crypto::PseudoRandomKey for hkdf::Prk {
    type AeadKey = ring::aead::LessSafeKey;

    fn aead_from_hkdf(&self) -> Self::AeadKey {
        ...
    }
}
```

then I can just do `let aead_key = key.aead_from_hkdf(self.random_bytes);`

@Ralith  and you suggested the following,

```rust
pub fn encode<K: AeadKey, P: PseudoRandomKey>(
    &self,
    key: &P,
    address: &SocketAddr,
    retry_src_cid: &ConnectionId,
) -> Vec<u8>
```

which can then be invoked from `endpoint.rs` 

```rust
let token = RetryToken {
    orig_dst_cid: dst_cid,
    issued: SystemTime::now(),
    random_bytes: &random_bytes,
}
.encode::<S::AeadKey, S::PseudoRandomKey>(
    &*server_config.token_key,
    &remote,
    &temp_loc_cid,
);
```

based on the discussion it seems that we would prefer the first approach right?

Ralith: With the first approach you'd still need a type parameter, but only one instead of two, and it wouldn't need to be passed explicitly. That seems fine to me.

djc: Yup, sounds good to me.

quinn-proto/src/token.rs
Ralith: Use the `BufMut` impl for `&mut [u8]`:
```rust
let mut additional_data = [0u8; 16 + 2]; // max(ipv4, ipv6) + port
let mut cursor = &mut additional_data[..];
cursor.put_slice(etc);
```

Ralith: Also take care to only feed in the portion of `additional_data` that was actually written, otherwise we might have a collision issue.

kansi: @Ralith This was really helpful, thanks :+1:

kansi:
@djc @Ralith I CI build fail with the following error

```
error: reached the type-length limit while instantiating `<std::vec::IntoIter<Peer> as std...rs:216:58: 220:6 keylog:&bool]]>`
##[error]     |
     = note: consider adding a `#![type_length_limit="1848396"]` attribute to your crate

error: aborting due to previous error
##[error]could not compile `interop`.
```
Seems like it cannot compile `inerop` but I cannot replicate this issue locally :thinking: , any pointers here?

djc:
This is the issue tracked in #839. You should probably just increase the type length limit for the interop crate for now.

Ralith:
A work-around for the build issue was merged in https://github.com/djc/quinn/pull/844, so this should be unblocked with a rebase!

DemiMarie:
I suggest using something like XChaCha20-Poly1305 with a random nonce instead of the current ad-hoc construction.  XChaCha20-Poly1305 has a 192-bit nonce, so the likelihood of reuse is negligible.  The nonce should be randomly generated for the first use, and then incremented for all subsequent encryptions.

Ralith:
If we could easily do that, the usual 96 bit nonce would be plenty. The problem is there's no way to recover the nonce used by such a scheme when validating a client-supplied retry token without defeating the goal of producing a retry token indistinguishable from random data.

I also don't think the risk of a custom construction here is high. In the unlikely event that a third party manages to decrypt a retry token, they don't learn anything useful. The main goal here is just future-proofing: nobody can bake knowledge of the token's structure into a middlebox if we don't expose any structure.

DemiMarie:
@Ralith a 96-bit nonce would NOT be plenty, as the birthday bound for it is too low.  However, if you use a nonce that is chosen randomly for every encryption, then the nonce itself is indistinguishable from random bits and can be made part of the retry token itself.

Ralith:
Randomly generating each use has different dynamics than incrementing each use as you suggested in your prior post, yes. I still don't see the need to rewrite this PR and use a cipher unsupported by *ring* when the current approach works fine and there's little at risk, in any case.

quinn-proto/src/lib.rs
```diff
@@ -235,3 +235,4 @@ const MIN_MTU: u16 = 1232;
 const TIMER_GRANULARITY: Duration = Duration::from_millis(1);
 /// Maximum number of streams that can be uniquely identified by a stream ID
 const MAX_STREAM_COUNT: u64 = 1 << 60;
+const MAX_ADDITIONAL_DATA_SIZE: usize = 39; // max(ipv4, ipv6) + port + retry_src_cid
```
Ralith, line 238: This is specific to the internals of `RetryToken` and hence shouldn't be defined cratewide.

quinn-proto/src/token.rs
```diff
@@ -43,46 +49,57 @@ impl RetryToken {
                 .unwrap_or(0),
         );
 
-        let signature_pos = buf.len();
+        let mut additional_data = Vec::new();
```
Ralith, line 52: We should avoid heap allocation here the same way we do in `from_bytes`.

quinn-proto/src/crypto.rs
```diff
@@ -38,6 +38,10 @@ pub trait Session: Send + Sized {
     type ClientConfig: ClientConfig<Self>;
     /// Type used to sign various values
     type HmacKey: HmacKey;
+    /// Pseudo random key used for generating AEAD key
+    type PseudoRandomKey: PseudoRandomKey;
+    /// Key used for protecting handshake tokens
+    type AeadKey: AeadKey;
```
Ralith, line 44: Is this is redundant to `PseudoRandomKey::AeadKey`?

Ralith (changes requested):
Just a few final tweaks!

quinn-proto/src/token.rs
```diff
@@ -215,7 +219,8 @@ mod test {
         let prk: ring::hkdf::Prk = crypto::PseudoRandomKey::from_secret(&master_key);
 
         let addr = SocketAddr::new(Ipv6Addr::LOCALHOST.into(), 4433);
-        let retry_src_cid = ConnectionId::random(rng, MAX_CID_SIZE);
+        let (retry_src_cid, _) = RandomConnectionIdGenerator::new(MAX_CID_SIZE).generate_cid();
+        // let retry_src_cid = ConnectionId::random(rng, MAX_CID_SIZE);
```
Ralith, line 223: Dead comment?

quinn-proto/src/crypto.rs
```diff
@@ -40,8 +40,6 @@ pub trait Session: Send + Sized {
     type HmacKey: HmacKey;
     /// Pseudo random key used for generating AEAD key
     type PseudoRandomKey: PseudoRandomKey;
```
Ralith: On review I think this would be a lot clearer if the name described the application, since we're not trying to define a general-purpose PRK abstraction here.
```suggestion
    /// Key used to generate one-time-use handshake token keys
    type HandshakeTokenKey: HandshakeTokenKey;
```
and similar for the trait definition, of course.

Ralith (changes requested):
Two more cosmetic issues, then please squash.

Ralith (approved):
Thanks for your work!

quinn-proto/src/token.rs
```diff
@@ -43,46 +50,59 @@ impl RetryToken {
                 .unwrap_or(0),
         );
 
-        let signature_pos = buf.len();
+        let mut additional_data = [0u8; Self::MAX_ADDITIONAL_DATA_SIZE];
```
djc, line 53: Would it be viable to factor out some of the duplication between `encode()` and `from_bytes()` in writing out the `additional_data`?

quinn-proto/src/endpoint.rs
```diff
@@ -548,16 +548,21 @@ where
...
+                self.rng.fill_bytes(&mut random_bytes);
+
                 let token = RetryToken {
                     orig_dst_cid: dst_cid,
                     issued: SystemTime::now(),
+                    random_bytes: &random_bytes,
                 }
                 .encode(&*server_config.token_key, &remote, &temp_loc_cid);
-                let mut buf = Vec::new();
+
```
djc, line 560: Nit: moving the initialization of `buf` around here seems orthogonal to the rest of this change. If it's not too much of hassle, maybe revert it?

djc (commented):
Thanks for the work! One hopefully minor concern and, if you work on that, a tiny nit to go along with it.

quinn-proto/src/token.rs
kansi: Sure :slightly_smiling_face:

kansi:
Linting seems to fail, its mostly suggesting to use `matches!` macro in place of `match something { ... }`

djc:
Feel free to ignore, I'll fix that up in a separate PR.

kansi:
okay, then I guess the PR can approved and merged.

djc (approved):
I took the liberty of rebasing this on top of current main to deal with some new conflicts, and split out my formatting nit into a separate commit. Thanks for working on this!

djc:
Merged despite some new clippy problems which are addressed in #864.

kansi:
@djc @Ralith  Thank you for all the support on this, really appreciate it :slightly_smiling_face:

Ralith:
Thanks for your contribution! It's great to have this checked off, and it's super cool that you were willing to take it on as a learning project!

PR #837, opened 2020-08-23
Reexport ReadUnordered

Imberflur:
I am assuming this was not intentionally missing.


djc:
Yup, that's probably my mistake. Thanks for catching it and fixing it up!

PR #838, opened 2020-08-23
Disable beta in CI due to temporary regression

djc:


djc:
The problem has now landed on stable, workaround in #844.

Issue #841, opened 2020-08-31
impl Debug for TlsSession


imp:
Please derive (or implement) `Debug` for `crypto::rustls::TlsSession`. `TlsSession` is default session type parameter for many other structs, and their `derive(Debug)` doesn't work while `TlsSession` is missing `Debug` impl.

djc:
What "many other structs" do you want to have `Debug` impls on?

imp:
Well, here is the partial list:

1. `ClientConfig`
2. `Connecting`
3. `Connection`
4. `Datagarms`
5. `Endpoint`
6. `EndpointBuilder`
7. `Incoming`
8. `NewConnection`
9. `RecvStream`
10. `SendStream`

imp:
It looks like simply deriving `Debug` on `TlsSession` and `SessionKind` does the trick. Is that ok with you?

PR #843, opened 2020-09-03
Derive Debug for TlsSession and SessionKind

imp:
This enables deriving `Debug` for other structures used in `quinn` that use `TlsSession` as a session type (for example `Incoming`, `Endpoint` etc),  which in turn enables 3rd parties structures to derive `Debug` easily when they use said structures in their code.

Addresses #841

### DIFF
--- a/quinn-proto/src/crypto/rustls.rs
+++ b/quinn-proto/src/crypto/rustls.rs
@@ -23,12 +23,14 @@ use crate::{
 };
 
 /// A rustls TLS session
+#[derive(Debug)]
 pub struct TlsSession {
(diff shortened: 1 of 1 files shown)


djc:
Looks good to me. Can you fix up the commit message? The current one is a little terse; it should probably contain what's become the PR title.

djc:
(Ignore the CI failures, we'll land #844 to fix those.)

imp:
Commit message updated. Let me know if anything else is need to be done.

djc (approved):

djc:
Thanks!

PR #849, opened 2020-09-12
fix iOS setsockopt invalid params

SSebo:
fix  https://github.com/djc/quinn/issues/848

### DIFF
--- a/quinn/src/platform/unix.rs
+++ b/quinn/src/platform/unix.rs
@@ -37,8 +37,10 @@ impl super::UdpExt for UdpSocket {
 
         let addr = self.local_addr()?;
 
-        // macos doesn't support IP_RECVTOS on dual-stack sockets :(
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks!

SSebo:
@djc Thank you for creating `quinn`!

djc:
@SSebo so are you using Quinn for TiKV/TiDB or something else? Just out of curiosity!

SSebo:
> @SSebo so are you using Quinn for TiKV/TiDB or something else? Just out of curiosity!

Oh, I use it for an instant messaging system with thousands of iOS and Android devices.

djc:
Oh, sounds cool! Is the instant messaging system public, or something proprietary?

It seems like it's there just fine: https://github.com/quinn-rs/quinn/blob/main/quinn/src/platform/unix.rs#L42.

SSebo:
Right now the whole system is not public. We will open source the connection management sub system which is powered by quinn by when we think it is stable.

PR #850, opened 2020-09-12
expose export_keying_material from rustls

### LINKED ISSUE #834: request: expose export_keying_material() from rustls::Session in Connection
It would be great if export_keying_material() were plumbed through from rustls::Session and made available for a Connection. The reason is that this would make it more straightforward to bind application-layer cryptographic protocols to the transport layer. My use-case is application-layer client authentication, but there are plenty of others---see [RFC5705](https://tools.ietf.org/html/rfc5705).

If this would be OK, I am happy to open a PR to do this. It should only take a few lines of code (plus tests...).

Thanks for a very cool library. I'm enjoying working with it.

kwantam:
This PR adds support for the export_keying_material method from rustls. Specifically:

- adds a trait, `proto::crypto::ExportKeyingMaterial`, that can be implemented on a `proto::crypto::Session`. I did this rather than implement it as part of `Session` because future crypto impls might not provide this functionality. Of course, can move it to `Session` if that would be better.

- implements `ExportKeyingMaterial` for `proto::crypto::rustls::TlsSession`, which calls the corresponding method on the rustls `Session` object.

- implements `export_keying_material` method on `quinn::Connection<S> where S: ExportKeyingMaterial`, which is just plumbing.

- tests `export_keying_material` in both `proto` and `quinn`.

To test this, you'll need to point to the current main branch of rustls in quinn and quinn-proto. You'll also need to point to a local copy of rustls-native-certs that pulls in the main branch of rustls.

Merging this is blocked on rustls and rustls-native-certs updating.

closes #834

### DIFF
--- a/quinn-proto/src/crypto/rustls.rs
+++ b/quinn-proto/src/crypto/rustls.rs
@@ -45,6 +45,23 @@ impl TlsSession {
     }
 }
 
+impl crypto::ExportKeyingMaterial for TlsSession {
+    type Error = TLSError;
+
+    fn export_keying_material(
+        &self,
+        output: &mut [u8],
+        label: &[u8],
+        context: Option<&[u8]>
+    ) -> Result<(), Self::Error> {
+        let session: &dyn rustls::Session = match &self.inner {
+            SessionKind::Client(s) => s,
+            SessionKind::Server(s) => s,
+        };
+        session.export_keying_material(output, label, context)
+    }
+}
+
 impl crypto::Session for TlsSession {
     type HandshakeData = HandshakeData;
     type Identity = CertificateChain;
--- a/quinn-proto/src/crypto/rustls.rs
+++ b/quinn-proto/src/crypto/rustls.rs
@@ -17,7 +17,7 @@ use webpki::DNSNameRef;
 
 use crate::{
     crypto,
-    crypto::{KeyPair, Keys},
+    crypto::{ExportKeyingMaterialError, KeyPair, Keys},
     transport_parameters::TransportParameters,
     CertificateChain, ConnectError, ConnectionId, Side, TransportError, TransportErrorCode,
 };
@@ -53,7 +53,6 @@ impl crypto::Session for TlsSession {
     type PacketKey = PacketKey;
     type HeaderKey = HeaderProtectionKey;
     type ServerConfig = Arc<rustls::ServerConfig>;
-    type ExportKeyingMaterialError = TLSError;
 
     fn initial_keys(dst_cid: &ConnectionId, side: Side) -> Keys<Self> {
         const INITIAL_SALT: [u8; 20] = [
@@ -226,12 +225,13 @@ impl crypto::Session for TlsSession {
         output: &mut [u8],
         label: &[u8],
         context: &[u8],
-    ) -> Result<(), Self::ExportKeyingMaterialError> {
+    ) -> Result<(), ExportKeyingMaterialError> {
         let session: &dyn rustls::Session = match &self.inner {
             SessionKind::Client(s) => s,
             SessionKind::Server(s) => s,
         };
         session.export_keying_material(output, label, Some(context))
+            .map_err(|_| ExportKeyingMaterialError)
     }
 }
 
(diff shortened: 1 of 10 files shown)


kwantam:
(test failures expected until rustls is updated)

djc:
Let's make it part of `Session` for now, we can split it later if an implementation that doesn't provide keying material comes along. Also, please update rustls to a pinned upstream commit that includes the necessary support (in all crates that depend on it).

kwantam:
OK, pushed the following:

- 6344f23 makes `export_keying_material` part of the `Session` trait.

- 1989122 pins all crates with a transitive dependency on rustls. In some cases this required forking the upstream repo and pinning to a rev of the fork with a correctly updated Cargo.toml. I checked that all of these pins were actually necessary, and unfortunately it appears that they are.

All tests pass on my machine. I will investigate the test failure.

djc:
Looks pretty good to me. I'm a bit on the fence about the associated error type, which seems overly specific, but I haven't thought of a great alternative yet. @Ralith, what do you think of these patches?

(FYI, I think all of these commits can be squashed, since they're largely not independent.)

kwantam:
I suppose the alternative is to conver that error into... a TransportError, maybe?

Happy to squash. Would you like me to do it now, or after @Ralith reviews? Also, should I leave the edits to the Cargo.toml files in a separate commit to make them easier to revert?

djc:
Squashing isn't urgent, I can do it when merging to master, too. I think we can leave the edits to Cargo.toml files, we've done it before for other upstream changes we wanted.

quinn-proto/src/crypto/rustls.rs
```diff
@@ -219,6 +220,19 @@ impl crypto::Session for TlsSession {
         let (aad, tag) = pseudo_packet.split_at_mut(tag_start);
         key.open_in_place(nonce, aead::Aad::from(aad), tag).is_ok()
     }
+
+    fn export_keying_material(
+        &self,
+        output: &mut [u8],
+        label: &[u8],
+        context: Option<&[u8]>,
```
Ralith, line 228:
```suggestion
        context: &[u8],
```

As far as I can tell there's no special meaning to `None` here; [rustls just converts it to an empty slice](https://github.com/ctz/rustls/blob/main/rustls/src/key_schedule.rs#L357). Unless there's some subtlety I'm missing, we should keep our signature simple.

Ralith (changes requested):
Apologies for the delay weighing in here. The error question is interesting. It looks like rustls only generates an error if too much data is requested (see https://github.com/ctz/rustls/blob/main/rustls/src/key_schedule.rs#L362). If we can't think of any other possible errors, I'd prefer to just bake in that semantic (either with an error ZST or just by using `Option`) and avoid the associated type.

quinn-proto/src/crypto/rustls.rs
kwantam: Hmm, good point.

Digging into the RFCs, it turns out that in TLS1.2 and before, 'no context' and 'empty context' were different, but in TLS1.3 they are the same. This explains the rustls type signature, and seems to mean that we're fine getting rid of the Option.

(ref: Section 7.5 of RFC 8446)

kwantam:
Thanks for the review, @ralith! I've addressed your comments other than the one about the error.

As far as I can tell you're right about the error. Would you like me to remove the associated type and create a zero-sized `ExportKeyingMaterialError` type instead?

djc:
Yup, that sounds right.

kwantam:
:+1: done!

djc:
LGTM!

quinn/src/tests.rs
```diff
@@ -98,11 +98,16 @@ fn local_addr() {
 }
 
 #[test]
-fn read_after_close() {
+fn read_after_close_and_export_keying_material() {
```
Ralith, line 101: Should this be two separate tests?

kwantam: Sure, that would work! I'm happy to make the change. I did this to avoid duplicating the setup code, which would otherwise end up being most of the new function.

Would you like me to split it?

Ralith: Yeah; if we end up with duplicate code then it's better to find a way to factor that out than to conflate independent tests.

kwantam: Done. :+1:

Ralith (approved):
CI failure looks spurious.

PR #851, opened 2020-09-17
Support customized Cid generation

liwenjieQu:
Currently Quinn generates cid using random number only. Some use case may require a more sophisticated cid structure. This change mainly add a `ConnectionIdGenerator` trait that allows to use any customized cid generator.

### DIFF
--- a/quinn-proto/src/cid_generator.rs
+++ b/quinn-proto/src/cid_generator.rs
@@ -1,0 +1,50 @@
+use rand::RngCore;
+
+use crate::shared::ConnectionId;
+use crate::MAX_CID_SIZE;
+
+/// Generates connection IDs for incoming connections
+pub trait ConnectionIdGenerator: Send {
+    /// Generates a connection ID for a new connection
+    fn generate_cid(&mut self) -> ConnectionId;
+    /// Performs any validation it needs (e.g. HMAC, etc)
+    fn validate_cid(&mut self, cid: &ConnectionId) -> bool;
+    /// Returns the length of a connection id for cononections created by this generator
+    fn cid_len(&self) -> usize;
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct RandomConnectionIdGenerator {
+    cid_len: usize,
+}
+impl Default for RandomConnectionIdGenerator {
+    fn default() -> Self {
+        Self { cid_len: 8 }
+    }
+}
+impl RandomConnectionIdGenerator {
+    pub fn new(cid_len: usize) -> Self {
+        debug_assert!(cid_len <= MAX_CID_SIZE);
+        Self { cid_len }
+    }
+}
+impl ConnectionIdGenerator for RandomConnectionIdGenerator {
+    fn generate_cid(&mut self) -> ConnectionId {
+        let mut res = ConnectionId {
+            len: self.cid_len as u8,
+            bytes: [0; MAX_CID_SIZE],
+        };
+        rand::thread_rng().fill_bytes(&mut res.bytes[..self.cid_len]);
+        res
+    }
+
+    /// Cid is an array of random bytes. We only verify the length
+    fn validate_cid(&mut self, cid: &ConnectionId) -> bool {
+        cid.len as usize == self.cid_len
+    }
+
+    /// Provide the length of dst_cid in short header packet
+    fn cid_len(&self) -> usize {
+        self.cid_len
+    }
+}
--- a/quinn-proto/src/cid_generator.rs
+++ b/quinn-proto/src/cid_generator.rs
@@ -26,9 +26,7 @@ pub trait ConnectionIdGenerator: Send {
     fn cid_len(&self) -> usize;
 }
 
-/// CID filled with random number/byte
-///
-/// This struct generates random CID with a customized length.
+/// Generates purely random connection IDs of a certain length
 #[derive(Debug, Clone, Copy)]
 pub struct RandomConnectionIdGenerator {
     cid_len: usize,
--- a/quinn-proto/src/cid_generator.rs
+++ b/quinn-proto/src/cid_generator.rs
@@ -12,16 +12,6 @@ pub trait ConnectionIdGenerator: Send {
     /// issuer) to correlate them with other connection IDs for the same
     /// connection.
     fn generate_cid(&mut self) -> ConnectionId;
-    /// Performs any validation if it is needed (e.g. HMAC, etc)
-    ///
-    /// Apply validation check on those CIDs that may still exist in hash table
-    /// but considered invalid by application-layer logic.
-    /// e.g. we may want to limit the amount of time for which a CID is valid
-    /// in order to reduce the number of valid IDs that could be accumulated
-    /// by an attacker.
-    fn validate_cid(&mut self, _cid: &ConnectionId) -> bool {
-        true
-    }
     /// Returns the length of a CID for cononections created by this generator
     fn cid_len(&self) -> usize;
 }
--- a/quinn-proto/src/cid_generator.rs
+++ b/quinn-proto/src/cid_generator.rs
@@ -14,7 +14,7 @@ pub trait ConnectionIdGenerator: Send {
     /// issuer) to correlate them with other connection IDs for the same
     /// connection.
     ///
-    /// Connection IDs will be proactively retired if lifetime is given (under development).
+    /// In the future, connection IDs will be retired after the returned `Duration`, if any.
     fn generate_cid(&mut self) -> (ConnectionId, Option<Duration>);
     /// Returns the length of a CID for cononections created by this generator
     fn cid_len(&self) -> usize;
--- a/quinn-proto/src/cid_generator.rs
+++ b/quinn-proto/src/cid_generator.rs
@@ -5,14 +5,17 @@ use crate::MAX_CID_SIZE;
 
 /// Generates connection IDs for incoming connections
 pub trait ConnectionIdGenerator: Send {
-    /// Generates a connection ID for a new connection
+    /// Generates a new CID
     fn generate_cid(&mut self) -> ConnectionId;
-    /// Performs any validation it needs (e.g. HMAC, etc)
+    /// Performs any validation if it is needed (e.g. HMAC, etc)
     fn validate_cid(&mut self, cid: &ConnectionId) -> bool;
-    /// Returns the length of a connection id for cononections created by this generator
+    /// Returns the length of a CID for cononections created by this generator
     fn cid_len(&self) -> usize;
 }
 
+/// CID filled with random number/byte
+///
+/// This struct generates random CID with a customized length.
 #[derive(Debug, Clone, Copy)]
 pub struct RandomConnectionIdGenerator {
     cid_len: usize,
@@ -23,6 +26,7 @@ impl Default for RandomConnectionIdGenerator {
     }
 }
 impl RandomConnectionIdGenerator {
+    /// Initialize Random CID generator with a fixed CID length (which must be less or equal to MAX_CID_SIZE)
     pub fn new(cid_len: usize) -> Self {
         debug_assert!(cid_len <= MAX_CID_SIZE);
         Self { cid_len }
--- a/quinn-proto/src/cid_generator.rs
+++ b/quinn-proto/src/cid_generator.rs
@@ -6,9 +6,22 @@ use crate::MAX_CID_SIZE;
 /// Generates connection IDs for incoming connections
 pub trait ConnectionIdGenerator: Send {
     /// Generates a new CID
+    ///
+    /// Connection IDs MUST NOT contain any information that can be used by
+    //    an external observer (that is, one that does not cooperate with the
+    //    issuer) to correlate them with other connection IDs for the same
+    //    connection.
     fn generate_cid(&mut self) -> ConnectionId;
     /// Performs any validation if it is needed (e.g. HMAC, etc)
-    fn validate_cid(&mut self, cid: &ConnectionId) -> bool;
+    ///
+    /// Apply validation check on those CIDs that may still exist in hash table
+    ///   but considered invalid by application-layer logic.
+    /// e.g., We may want to limit the amount of time for which a CID is valid
+    ///   in order to reduce the number of valid IDs that could be accumulated
+    ///   by an attacker.
+    fn validate_cid(&mut self, _cid: &ConnectionId) -> bool {
+        true
+    }
     /// Returns the length of a CID for cononections created by this generator
     fn cid_len(&self) -> usize;
 }
@@ -34,17 +47,10 @@ impl RandomConnectionIdGenerator {
 }
 impl ConnectionIdGenerator for RandomConnectionIdGenerator {
     fn generate_cid(&mut self) -> ConnectionId {
-        let mut res = ConnectionId {
-            len: self.cid_len as u8,
-            bytes: [0; MAX_CID_SIZE],
-        };
-        rand::thread_rng().fill_bytes(&mut res.bytes[..self.cid_len]);
-        res
-    }
+        let mut bytes_arr = vec![0; self.cid_len];
+        rand::thread_rng().fill_bytes(&mut bytes_arr);
 
-    /// Cid is an array of random bytes. We only verify the length
-    fn validate_cid(&mut self, cid: &ConnectionId) -> bool {
-        cid.len as usize == self.cid_len
+        ConnectionId::new(&bytes_arr)
     }
 
     /// Provide the length of dst_cid in short header packet
--- a/quinn-proto/src/cid_generator.rs
+++ b/quinn-proto/src/cid_generator.rs
@@ -45,10 +45,10 @@ impl RandomConnectionIdGenerator {
 }
 impl ConnectionIdGenerator for RandomConnectionIdGenerator {
     fn generate_cid(&mut self) -> ConnectionId {
-        let mut bytes_arr = vec![0; self.cid_len];
-        rand::thread_rng().fill_bytes(&mut bytes_arr);
+        let mut bytes_arr = [0; MAX_CID_SIZE];
+        rand::thread_rng().fill_bytes(&mut bytes_arr[..self.cid_len]);
 
-        ConnectionId::new(&bytes_arr)
+        ConnectionId::new(&bytes_arr[..self.cid_len])
     }
 
     /// Provide the length of dst_cid in short header packet
--- a/quinn-proto/src/cid_generator.rs
+++ b/quinn-proto/src/cid_generator.rs
@@ -15,10 +15,10 @@ pub trait ConnectionIdGenerator: Send {
     /// Performs any validation if it is needed (e.g. HMAC, etc)
     ///
     /// Apply validation check on those CIDs that may still exist in hash table
-    ///   but considered invalid by application-layer logic.
-    /// e.g., We may want to limit the amount of time for which a CID is valid
-    ///   in order to reduce the number of valid IDs that could be accumulated
-    ///   by an attacker.
+    /// but considered invalid by application-layer logic.
+    /// e.g. we may want to limit the amount of time for which a CID is valid
+    /// in order to reduce the number of valid IDs that could be accumulated
+    /// by an attacker.
     fn validate_cid(&mut self, _cid: &ConnectionId) -> bool {
         true
     }
--- a/quinn-proto/src/cid_generator.rs
+++ b/quinn-proto/src/cid_generator.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use rand::RngCore;
 
 use crate::shared::ConnectionId;
@@ -5,13 +7,15 @@ use crate::MAX_CID_SIZE;
 
 /// Generates connection IDs for incoming connections
 pub trait ConnectionIdGenerator: Send {
-    /// Generates a new CID
+    /// Generates a new CID with finite lifetime
     ///
     /// Connection IDs MUST NOT contain any information that can be used by
     /// an external observer (that is, one that does not cooperate with the
     /// issuer) to correlate them with other connection IDs for the same
     /// connection.
-    fn generate_cid(&mut self) -> ConnectionId;
+    ///
+    /// Connection IDs will be proactively retired if lifetime is given (under development).
+    fn generate_cid(&mut self) -> (ConnectionId, Option<Duration>);
     /// Returns the length of a CID for cononections created by this generator
     fn cid_len(&self) -> usize;
 }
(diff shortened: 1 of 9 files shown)


quinn-proto/src/config.rs
```diff
@@ -308,25 +312,27 @@ where
...
-    /// 0, at most 3/4 * 2^(local_cid_len * 8) simultaneous connections can be supported.
-    pub fn local_cid_len(&mut self, value: usize) -> Result<&mut Self, ConfigError> {
-        if value > MAX_CID_SIZE {
-            return Err(ConfigError::OutOfBounds);
+    /// EndpointConfig can still call fn new() to use a random cid generator
+    /// new_with_cid_generator() can accept any customized cid generator that
+    /// implements ConnectionIdGenerator trait
+    pub fn new_with_cid_generator<
+        F: Fn() -> Box<dyn ConnectionIdGenerator> + Send + Sync + 'static,
+    >(
```
Ralith, line 327: I'd rather this be a setter than an additional constructor, as setters are more scalable/composable.

Ralith (commented):
Thanks for the PR! I agree that we should expose something like this, e.g. to support load balancers. Out of curiosity, are you doing something with Quinn that requires this in practice?

quinn-proto/src/config.rs
Matthias247: The downside is that one would need a default initialization of the factory before (e.g. with the random generator factory), which would then get replaced. Introducing an `Option` and then having to deal later on with the field not being set seems not very convenient.

A `Builder` for the config could avoid this, but that's a lot bigger of an API change. 
Since the cost for the wasted `Box<RandomIdConnectionGeneratorFactory>` is likely negligible compared to the remaing setup (e.g. loading TLS keys) it's likely find to just have the setter.

quinn-proto/src/shared.rs
```diff
@@ -134,6 +122,52 @@ impl fmt::Display for ConnectionId {
...
+    /// Generates a connection ID for a new connection
+    fn generate_cid(&mut self) -> ConnectionId;
+    /// Performs any validation it needs (e.g. HMAC, etc)
+    fn validate_cid(&mut self, cid: &ConnectionId) -> bool;
+    /// Returns the length of a connection id for cononections created by this generator
+    fn cid_len(&self) -> usize;
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct RandomConnectionIdGenerator {
```
Matthias247, line 136: Should have a doc comment since its public

quinn-proto/src/shared.rs
```diff
@@ -134,6 +122,52 @@ impl fmt::Display for ConnectionId {
...
+    cid_len: usize,
+}
+impl Default for RandomConnectionIdGenerator {
+    fn default() -> Self {
+        Self { cid_len: 8 }
+    }
+}
+impl RandomConnectionIdGenerator {
+    pub fn new(cid_len: usize) -> Self {
```
Matthias247, line 145: doc comment

quinn-proto/src/shared.rs
```diff
@@ -134,6 +122,52 @@ impl fmt::Display for ConnectionId {
     }
 }
 
+/// Generates connection IDs for incoming connections
+pub trait ConnectionIdGenerator: Send {
```
Matthias247, line 126: Is the `shared` module actually `pub`lically exported? Otherwise won't be usable 😀
Btw: It could also live an a seperate `connection_id_generator` module, or just be reexported from the top level module.

quinn-proto/src/config.rs
Ralith: In the (indeed unlikely) event that the cost became an issue, we could even use a ZST for the default impl, which wouldn't so much as malloc.

liwenjieQu: I agree. This only occurs at the configuration phase and thus the waste is not a big problem - Will proceed the change to setter.

quinn-proto/src/shared.rs
liwenjieQu: Will do. This sounds good as the customized cid generator implementation can be done in a different crate and later imported into quinn-proto.

quinn-proto/src/lib.rs
```diff
@@ -59,7 +59,10 @@ pub use crate::transport_error::{Code as TransportErrorCode, Error as TransportE
 
 pub mod congestion;
 
+mod cid_generator;
 mod token;
+pub use crate::cid_generator::ConnectionIdGenerator;
```
Matthias247, line 64: Nit: Might swap this with line 63 and have an empty line after it for consistent style

quinn-proto/src/config.rs
```diff
@@ -308,25 +313,27 @@ where
...
         Self {
-            local_cid_len: 8,
             reset_key: Arc::new(reset_key),
             max_udp_payload_size: MAX_UDP_PAYLOAD_SIZE,
+            connection_id_generator_factory: Arc::new(cid_factory),
         }
     }
 
-    /// Length of connection IDs for the endpoint.
+    /// assign a cid generator factory
```
Ralith, line 325:
```suggestion
    /// Supply a custom connection ID generator factory
    ///
    /// Called once by each `Endpoint` constructed from this configuration to obtain the CID generator which will
    /// be used to generate the CIDs used for incoming packets on all connections involving that  `Endpoint`. A
    /// custom CID generator allows applications to embed information in local connection IDs, e.g. to support
    /// stateless packet-level load balancers.
```

quinn-proto/src/config.rs
```diff
@@ -308,25 +313,27 @@ where
...
-        if value > MAX_CID_SIZE {
-            return Err(ConfigError::OutOfBounds);
-        }
-        self.local_cid_len = value;
-        Ok(self)
+    ///
+    /// EndpointConfig::new() applies a default random CID generator factory.
+    /// This functions accepts any customized CID generator to reset CID generator factory that
+    /// implements ConnectionIdGenerator trait
+    pub fn set_cid_generator<F: Fn() -> Box<dyn ConnectionIdGenerator> + Send + Sync + 'static>(
```
Ralith, line 331:
```suggestion
    pub fn cid_generator<F: Fn() -> Box<dyn ConnectionIdGenerator> + Send + Sync + 'static>(
```
For concision, we omit the `set_` prefix from methods on builders where ~all methods are setters.

quinn-proto/src/endpoint.rs
```diff
@@ -793,7 +807,7 @@ pub enum ConnectError {
     EndpointStopping,
     /// The number of active connections on the local endpoint is at the limit
     ///
-    /// Try a larger `EndpointConfig::local_cid_len`.
+    /// Try a larger cid length.
```
Ralith, line 810:
```suggestion
    /// Try using longer connection IDs.
```

quinn-proto/src/lib.rs
```diff
@@ -59,7 +59,11 @@ pub use crate::transport_error::{Code as TransportErrorCode, Error as TransportE
 
 pub mod congestion;
 
+mod cid_generator;
+pub use crate::cid_generator::ConnectionIdGenerator;
+
 mod token;
+
```
Ralith, line 66: Stray newline.
```suggestion
mod token;
```

quinn-proto/src/shared.rs
```diff
@@ -1,9 +1,7 @@
 use std::{fmt, net::SocketAddr, time::Instant};
 
-use bytes::{Buf, BufMut, BytesMut};
-use rand::Rng;
-
 use crate::{coding::BufExt, packet::PartialDecode, ResetToken, MAX_CID_SIZE};
+use bytes::{Buf, BufMut, BytesMut};
```
Ralith, line 4: `use` statements referring to external items should be in a separate block, preceding those referring to internal ones.

quinn-proto/src/shared.rs
```diff
@@ -61,8 +59,10 @@ pub(crate) enum EndpointEventInner {
 /// Mainly useful for identifying this connection's packets on the wire with tools like Wireshark.
 #[derive(Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub struct ConnectionId {
-    len: u8,
-    bytes: [u8; MAX_CID_SIZE],
+    /// length of CID
+    pub len: u8,
+    /// CID in byte array
+    pub bytes: [u8; MAX_CID_SIZE],
```
Ralith, line 65: Would it be sufficient to instead export `ConnectionId::new` and otherwise rely on the pre-existing `Deref` and `DerefMut` impls?

quinn-proto/src/cid_generator.rs
```diff
@@ -0,0 +1,54 @@
...
+
+use crate::shared::ConnectionId;
+use crate::MAX_CID_SIZE;
+
+/// Generates connection IDs for incoming connections
+pub trait ConnectionIdGenerator: Send {
+    /// Generates a new CID
+    fn generate_cid(&mut self) -> ConnectionId;
+    /// Performs any validation if it is needed (e.g. HMAC, etc)
+    fn validate_cid(&mut self, cid: &ConnectionId) -> bool;
```
Ralith, line 11: It looks like this is only used to skip checking an `Endpoint`'s hash table for a CID, but that hash table should return `None` for invalid CIDs anyway, because we'd never have inserted them, so naively we could do without it. Is the intention to guard against timing attacks when a cryptographic mechanism is in use? If so, that should be noted in the docs so future readers aren't confused about the benefit.

quinn-proto/src/cid_generator.rs
```diff
@@ -0,0 +1,54 @@
+use rand::RngCore;
+
+use crate::shared::ConnectionId;
+use crate::MAX_CID_SIZE;
+
+/// Generates connection IDs for incoming connections
+pub trait ConnectionIdGenerator: Send {
+    /// Generates a new CID
+    fn generate_cid(&mut self) -> ConnectionId;
```
Ralith, line 9: We should have a brief note here describing desirable properties for connection IDs, e.g. that they be indistinguishable from random data by an external viewer.

Ralith (changes requested):
Thanks for the updates! Made a deep pass; mostly cosmetic issues.

quinn-proto/src/shared.rs
liwenjieQu: Should be enough to use `ConnectionId::new`. I am reverting the above changes.

liwenjieQu: Yes. I think it is enough to use `ConnectionId::new`. I am reverting the above changes.

quinn-proto/src/cid_generator.rs
liwenjieQu: I will add some notes from RFC doc on Connection ID properties

quinn-proto/src/cid_generator.rs
liwenjieQu: > It looks like this is only used to skip checking an Endpoint's hash table for a CID, but that hash table should return None for invalid CIDs anyway, because we'd never have inserted them, so naively we could do without it.
Yes I have noticed that hash table will guarantee that non-existing CID will remain invalid. The purpose of validating CID is indeed based on some security concern. 

The major purpose of this API is to limit the amount of time for which a CID is valid in order to reduce the number of valid IDs that could be accumulated by an attacker. I am pushing another PR later (where peer actively sends NEW_CONNECTION_ID frame to rotate the CID). As a result, `fn validate_cid` is used to mark any CID that may still appear in hash table but considered invalid by application layer logic. 

This seems like an optional API. Thus, I am providing a default implementation and more docs.

liwenjieQu: And I think timing attack is a good example on what may happen if an attacker holds a number of valid CIDs for a long time.

quinn-proto/src/cid_generator.rs
```diff
@@ -0,0 +1,60 @@
...
+use crate::MAX_CID_SIZE;
+
+/// Generates connection IDs for incoming connections
+pub trait ConnectionIdGenerator: Send {
+    /// Generates a new CID
+    ///
+    /// Connection IDs MUST NOT contain any information that can be used by
+    //    an external observer (that is, one that does not cooperate with the
+    //    issuer) to correlate them with other connection IDs for the same
+    //    connection.
```
Ralith, line 13: Formatting error:
```suggestion
    /// an external observer (that is, one that does not cooperate with the
    /// issuer) to correlate them with other connection IDs for the same
    /// connection.
```

quinn-proto/src/cid_generator.rs
```diff
@@ -0,0 +1,60 @@
...
+    fn validate_cid(&mut self, _cid: &ConnectionId) -> bool {
+        true
+    }
+    /// Returns the length of a CID for cononections created by this generator
+    fn cid_len(&self) -> usize;
+}
+
+/// CID filled with random number/byte
+///
+/// This struct generates random CID with a customized length.
```
Ralith, line 31:
```suggestion
/// Generates purely random connection IDs of a certain length
```

quinn-proto/src/lib.rs
```diff
@@ -59,6 +59,9 @@ pub use crate::transport_error::{Code as TransportErrorCode, Error as TransportE
 
 pub mod congestion;
 
+mod cid_generator;
+pub use crate::cid_generator::ConnectionIdGenerator;
```
Ralith, line 63: We should reexport `RandomConnectionIdGenerator` to allow easy customization of connection ID length.

quinn-proto/src/cid_generator.rs
```diff
@@ -0,0 +1,60 @@
...
+impl RandomConnectionIdGenerator {
+    /// Initialize Random CID generator with a fixed CID length (which must be less or equal to MAX_CID_SIZE)
+    pub fn new(cid_len: usize) -> Self {
+        debug_assert!(cid_len <= MAX_CID_SIZE);
+        Self { cid_len }
+    }
+}
+impl ConnectionIdGenerator for RandomConnectionIdGenerator {
+    fn generate_cid(&mut self) -> ConnectionId {
+        let mut bytes_arr = vec![0; self.cid_len];
```
Ralith, line 50: We can avoid a heap allocation here:
```suggestion
        let mut bytes_arr = [0; MAX_CID_SIZE];
```

quinn-proto/src/cid_generator.rs
```diff
@@ -0,0 +1,60 @@
...
+    //    issuer) to correlate them with other connection IDs for the same
+    //    connection.
+    fn generate_cid(&mut self) -> ConnectionId;
+    /// Performs any validation if it is needed (e.g. HMAC, etc)
+    ///
+    /// Apply validation check on those CIDs that may still exist in hash table
+    ///   but considered invalid by application-layer logic.
+    /// e.g., We may want to limit the amount of time for which a CID is valid
+    ///   in order to reduce the number of valid IDs that could be accumulated
+    ///   by an attacker.
```
Ralith, line 21: Stray whitespace. Unless I'm missing something about markdown?
```suggestion
    /// but considered invalid by application-layer logic.
    /// e.g. we may want to limit the amount of time for which a CID is valid
    /// in order to reduce the number of valid IDs that could be accumulated
    /// by an attacker.
```

quinn-proto/src/cid_generator.rs
```diff
@@ -0,0 +1,60 @@
...
+    fn validate_cid(&mut self, _cid: &ConnectionId) -> bool {
+        true
+    }
```
Ralith, line 24: Good call on the default implementation! I'd still like to understand the problem this is solving in more detail, though. Quinn already limits the number of connection IDs that may be outstanding per connection via the mechanisms discussed in [Issuing Connection IDs](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#name-issuing-connection-ids), enforced [inside `CidQueue`](https://github.com/djc/quinn/blob/master/quinn-proto/src/cid_queue.rs#L41-L43), currently hardcoded to at most 4. This is communicated to the peer via the `active_connection_id_limit` transport parameter, which helps ensure that well-behaved peers don't encounter issues.

Is there a CID exhaustion attack that this does not cover? If so, should we pursue a mechanism that also cleans them out of the hash table proactively, so as to avoid memory exhaustion?

Ralith (changes requested):

quinn-proto/src/cid_generator.rs
liwenjieQu: Sounds good

quinn-proto/src/cid_generator.rs
liwenjieQu: > Is there a CID exhaustion attack that this does not cover? 

i agree that for well-behaved peers we shouldn't expect issues. quinn has followed RFC to protect each pair of communications.

This `fn validate_cid`  is a precaution to avoid when an attack can somehow eardrop `multiple` pairs of communications and accumulate many CIDs if server doesn't proactively invalidate them. As a result, I will suggest peer (especially server side) can push `NEW_CONNECTION_ID` with updated `retire_prior_to` field  (currently, quinn always set it to 0) to rotate CID. This will somehow reduce the risk of "CID exhaustion attack". I have already finished the code on my local branch.

A well-behaved peer should reply with `RETIRE_CONNECTION_ID` to complete the CID rotation. If not,  `validate_cid` then will kick in to stop the server to keep using the retired ones. 

>If so, should we pursue a mechanism that also cleans them out of the hash table proactively, so as to avoid memory exhaustion?

I think it will be good if we can clean those CIDs that are invalidated. Could be a future improvement. 👍

Ralith: Oh, I see, the real goal here is CID key rotation or similar. That's definitely something we want to support. My concern is that the approach currently taken here will place well-behaved peers at risk of connection loss due to not employing `retire_prior_to`. Additionally, there's no mechanism introduced to issue new CIDs, so a long-lived connection will run out and be lost. Therefore I don't think `validate_cid` is useful as written.

Since it sounds like you have a better approach almost ready to go, can we `validate_cid` here, merge the rest, and address CID rotation in a future PR?

Matthias247: > Since it sounds like you have a better approach almost ready to go, can we omit validate_cid here, merge the rest, and address CID rotation in a future PR?

Wouldn't this break users if the start implementing the new trait, and the additional method has to be added later on?
If it's forseeable that it will be required or even that it could be useful there shouldn't be a lot of harm with adding it - even if it's not called right now. The implementation should be straightforward for most CID formats.

liwenjieQu: > My concern is that the approach currently taken here will place well-behaved peers at risk of connection loss due to not employing retire_prior_to

Totally understood. I double check the description in [RFC](https://www.ietf.org/id/draft-ietf-quic-transport-31.html#name-consuming-and-retiring-conn) where I quote 
> "Upon receipt of an increased Retire Prior To field, the peer MUST stop using the corresponding connection IDs and retire them with RETIRE_CONNECTION_ID frames before adding the newly provided connection ID to the set of active connection IDs."

Since peer `MUST` has a reply, I don't think if a peer who doesn't employ the `retire_prior_to` can be categorized as "well-behaved" (at least it is not RFC-compliant). 

The `validate_cid` function provides a API/approach to apply what is mentioned in RFC where
> "The endpoint SHOULD continue to accept the previously issued connection IDs until they are retired by the peer. If the endpoint can no longer process the indicated connection IDs, it MAY close the connection."

Since it is a `MAY` in RFC, the CID generator creator can thus decide whether/when we want to close the connection based on their own implementation. In the current PR, the false result of `validate_cid` will only ignore the packets from peer, which is not exact equivalent to closing the connection. This could be a future improvement.

With that said, 
> Since it sounds like you have a better approach almost ready to go

I have a PR ready that implements the advance of `retire_prior_to` but the not form of cid generator trait.

Since the current calling of `validate_cid` is not perfect yet. I am willing to rollback this API. Regarding to Matthias's concern, it is also feasible to add it back in the future without breaking compatibility if we provide a default implementation as I have now: users can update their implementation if they want to use this feature.

Ralith: > Since peer MUST has a reply, I don't think if a peer who doesn't employ the retire_prior_to can be categorized as "well-behaved" (at least it is not RFC-compliant).

Right, the problem is that quinn itself does not yet generate `retire_prior_to`, hence if any implementer ever returned `false` from `validate_cid` under current conditions it would necessarily cause a CID that might still be legally used by the peer to be ignored.

> If it's forseeable that it will be required or even that it could be useful there shouldn't be a lot of harm with adding it

Ideally we'd want to tightly couple the local invalidation of CIDs with the issuing of a `retire_prior_to` and (deferred) updates to the Endpoint's hash table so that everything is guaranteed to stay in sync, which will likely require a different interface.

Ralith (approved):
Thanks for your work!

liwenjieQu:
Appreciate your time @Ralith !

liwenjieQu:
Could you squash and merge these commits so I can push a new PR? Thanks

Ralith:
I'd prefer to get a signoff from @djc first. You can make a new PR based on this branch, and when we merge this to mater you should be able to rebase the new PR onto master without issue, and I'll happily review the new parts in the mean time. You could also squash this branch into a single commit yourself first for extra insurance against git getting tangled up.

liwenjieQu:
Just realize that I need to remove 'crate' from `pub(crate)` before new function for ConnectionId, and thus a third-party cid generator trait can implement `generate_cid` on its own

Ralith:
Whoops, good catch.

Ralith (approved):

liwenjieQu:
The error from CI is mysterious. I am using a Mac and all tests pass on my side.

Ralith:
There's a similar issue on #850; I'm guessing it's on github's end and we should ignore it.

liwenjieQu:
Please not merge the PR now. I need to update the CID generator trait for proactive cid rotation https://github.com/liwenjieQu/quinn/pull/1 where the lifetime/duration of a CID needs to be returned. I think it is better to declare in the very beginning to avoid breaking compatibility

liwenjieQu:
Sorry for the inconvenience that the PR has to be reviewed again.

quinn-proto/src/cid_generator.rs
```diff
@@ -1,17 +1,21 @@
...
-    /// Generates a new CID
+    /// Generates a new CID with finite lifetime
     ///
     /// Connection IDs MUST NOT contain any information that can be used by
     /// an external observer (that is, one that does not cooperate with the
     /// issuer) to correlate them with other connection IDs for the same
     /// connection.
-    fn generate_cid(&mut self) -> ConnectionId;
+    ///
+    /// Connection IDs will be proactively retired if lifetime is given (under development).
```
Ralith, line 17:
```suggestion
    /// In the future, connection IDs will be retired after the returned `Duration`, if any.
```

Ralith (changes requested):
Good catch. Minor API breakage isn't the end of the world at present, but might as well avoid it when we can.

Ralith (approved):

djc:
Sorry for the delayed review -- thanks for working on this, it looks pretty good!

quinn-proto/src/endpoint.rs
```diff
@@ -341,7 +345,7 @@ where
         if self.is_full() {
             return Err(ConnectError::TooManyConnections);
         }
-        let remote_id = ConnectionId::random(&mut self.rng, MAX_CID_SIZE);
+        let (remote_id, _) = RandomConnectionIdGenerator::new(MAX_CID_SIZE).generate_cid();
```
infinity0, line 348: @liwenjieQu Is there any reason we hardcode the generator here, and don't use the `local_cid_generator`? The main difference I see between `RandomConnectionIdGenerator` vs `local_cid_generator` is the `cid_len`.

Here we pass `remote_id` to `add_connection`. However inside that function, we pass `self._local_cid_generator` to `Connection::new` which uses its length, instead of the length of `remote_id` (which was generated by `RandomConnectionIdGenerator`).

Should we change this to be `self.local_cid_generator.generate_cid()` for consistency.?

Ralith: Note the name `remote_id`; we're generating a *destination* connection ID here, not a local one. The client uses a random destination connection ID for its initial flight of packets; see https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#name-negotiating-connection-ids for details.

infinity0: @Ralith Thanks for the explanation! That clears it up.

PR #852, opened 2020-09-20
Implement pacing

### LINKED ISSUE #180: Implement pacing

DemiMarie:
This implements pacing as suggested by the latest RFC draft.  It also
attempts to batch packets wherever possible.

Fixes #180.


djc (approved):

Ralith (approved):
Thanks, and apologies again that this took so long to iterate! Benchmarking results look similar, so I think we're in business.

PR #855, opened 2020-09-25
Assorted fixes

Ralith:


djc (approved):
Nice! Does the panic fix come from fuzzing stuff?

Ralith:
Nah, that was an embarrassingly simple manual test by @Matthias247.

PR #860, opened 2020-10-04
Proactive CID rotation

liwenjieQu:
This is a follow-up PR after https://github.com/djc/quinn/pull/851

**Objective**: Implement CID rotation. Peer now can proactively retire previously issued CIDs after a fixed amount of duration, in order to prevent "CID exhaustion" attack.

**Approach**:

1. Modify `NeedIdentifiers` event handling logic to setup a timer in the future when CID needs to be retired.
2. Add a `PushNewCID` timer and corresponding logic that eventually issues `NEW_CONNECTION_ID` frame when timer expires.
[3]. Improve `RETIRE_CONNECTION_ID` processing logic. Add a hash set to track the active cid sequence used by the current connection.
[4]. Fix a hidden bug when processing `NEW_CONNECTION_ID` frame where `retire_prior_to` may retire all CIDs in both `cid_queue` and `rem_cid`, and cause a corrupted `offset` value.

[3] and [4] are resolved in #871


liwenjieQu:
The initial PR was already submitted in my personal repository - https://github.com/liwenjieQu/quinn/pull/1. Some comments from Matthias are not addressed yet which will be resolved here.

quinn-proto/src/cid_queue.rs
```diff
@@ -15,7 +15,7 @@ pub struct CidQueue {
     ///
     /// The CID sequenced immediately prior to this is the active CID, which this data structure is
     /// not responsible for retiring.
-    offset: u64,
+    pub(crate) offset: u64,
```
Ralith, line 18: Rather than exposing this directly to external manipulation, can we encapsulate the operations we're trying to perform on it? It's much easier to reason about the possible states of a data structure if its innards aren't exposed to large amounts of code.

quinn-proto/src/connection/mod.rs
```diff
@@ -77,6 +77,23 @@ where
     /// Exactly one prior to `self.rem_cids.offset` except during processing of certain
     /// NEW_CONNECTION_ID frames.
     rem_cid_seq: u64,
+    rem_cid_resettoken: Option<ResetToken>,
```
Ralith, line 80: We're currently storing the current reset token in `self.peer_params.stateless_reset_token`. Is this redundant to that?

quinn-proto/src/connection/mod.rs
```diff
@@ -2376,25 +2529,51 @@ where
                         .retire_cids
                         .extend(retired);
 
+                    let mut valid_rem_cid = Ok(());
+                    if frame.retire_prior_to > self.rem_cid_seq {
+                        // If our current CID is earlier than the first unretired one we must not
+                        // have adopted the one we just got, so it must be stored in rem_cids, so
+                        // this unwrap is guaranteed to succeed.
```
Ralith, line 2536: What unwrap?

quinn-proto/src/connection/mod.rs
```diff
@@ -2376,25 +2529,51 @@ where
...
+                            Err(InsertError::Retired) => {
+                                trace!("discarding already-retired");
+                                self.spaces[SpaceId::Data]
+                                    .pending
+                                    .retire_cids
+                                    .push(frame.sequence);
+                                continue;
+                            }
+                        },
+                        Err(_) => {
```
Ralith, line 2573: If we ignore the `Err` payload, `valid_rem_cid` should just be an `Option`.

quinn-proto/src/connection/mod.rs
```diff
@@ -2519,17 +2693,29 @@ where
...
+                // Retire any CIDs we had to skip.
+                retire_cids.extend(cid.1);
+
+                // Apply new remote CID
+                self.apply_new_rem_cid(cid.0);
+                Ok(())
+            }
+            None => Err(()),
+        }
+    }
```
Ralith, line 2713: By moving the retiring of the current CID before checking whether we have any others, this change removes the check that prevents us from retiring a remote CID when don't have any others to switch to, potentially rendering the connection unusable. Is that intended?

quinn-proto/src/connection/mod.rs
```diff
@@ -752,8 +797,105 @@ where
                     self.path.challenge_pending = false;
                 }
                 Timer::Pacing => trace!("pacing timer expired"),
+                Timer::PushNewCid => {
+                    let num_pushed_cids = 1;
+                    let mut ids_found = false;
+                    for id in self.ack_retire_prior_to..self.retire_cid_seq {
+                        ids_found = ids_found || self.cids_active_seq.contains(&id);
```
Ralith, line 804:
```suggestion
                        ids_found |= self.cids_active_seq.contains(&id);
```

quinn-proto/src/connection/mod.rs
```diff
@@ -77,6 +77,23 @@ where
...
+    /// Expiration timestamp of local connection IDs
+    cids_timeline: VecDeque<CidTimeStamp>,
+    /// The lifetime of Local connection IDs
+    cids_timeout: Option<Duration>,
+    /// Number of local connection IDs that have been issued in NEW_CONNECTION_ID frames.
+    cids_issued: u64,
+    /// The sequence number of active (not yet rotated/retired by peer) local connection IDs
+    cids_active_seq: HashSet<u64>,
+    /// Sequence number so far has been acked in RETIRE_CONNECTION_ID frame
+    ack_retire_prior_to: u64,
```
Ralith, line 93: It doesn't look like this is updated based on ACKs at all, but rather based on what the peer has retired. Maybe rename and reword comment?

Ralith (changes requested):
Thanks! Made an initial pass, but I still need to dig deeper into the new logic.

> Fix a hidden bug when processing NEW_CONNECTION_ID frame where retire_prior_to may retire all CIDs in both cid_queue and rem_cid, and cause a corrupted offset value.

If I understand correctly, you're saying you found and fixed a pre-existing bug? If so, could you please split that fix into a separate PR for easier independent review, and illustrate the issue with a test?

Ralith:
Another thought: Is it actually useful to have the `Duration` of every single CID be variable? This might be simpler if the lifetime were part of the connection configuration instead, making it constant for the lifetime of the connection (and also simplifying the CID generator interface).

liwenjieQu:
> Thanks! Made an initial pass, but I still need to dig deeper into the new logic.
> 
> > Fix a hidden bug when processing NEW_CONNECTION_ID frame where retire_prior_to may retire all CIDs in both cid_queue and rem_cid, and cause a corrupted offset value.
> 
> If I understand correctly, you're saying you found and fixed a pre-existing bug? If so, could you please split that fix into a separate PR for easier independent review, and illustrate the issue with a test?

Will do. Meanwhile, i probably will leave the fix in this CR as well until the it is merged into mainline. The unit test of this CR cannot work without the fix.

quinn-proto/src/cid_queue.rs
liwenjieQu: Sure. Will encapsulate with some functions on this field

quinn-proto/src/connection/mod.rs
liwenjieQu: Good catch. I noticed before that `self.peer_params.stateless_reset_token` has reset token but i chose to add another field anyway. Will revisit my previous decision and see if i can get rid of it

quinn-proto/src/connection/mod.rs
liwenjieQu: I must mess up the comments :D

quinn-proto/src/connection/mod.rs
liwenjieQu: Will update the corresponding function.

quinn-proto/src/connection/mod.rs
liwenjieQu: Ack

quinn-proto/src/connection/mod.rs
liwenjieQu: I see your point.  This situation was (partially) considered when processing NEW_CONNECTION_ID frame. If we don't have any other CIDs to switch to, `fn update_rem_cid ` return `Err(())` and will be caught by `match valid_rem_cid` where the current CID in the frame will be activated.  However, i do observe now that  `fn update_rem_cid` is used in multiple places (e.g., after migration) which could lead the situation as you mentioned. Working on a fix now.

quinn-proto/src/connection/mod.rs
liwenjieQu: Yes you are right. It is the sequence number prior to which all CIDs are retired and acknowledged.

quinn-proto/src/connection/mod.rs
liwenjieQu: Also, may I ask the design motivation of having a separate `rem_cid` in Connection struct, instead of having all remote CIDs managed by `CidQueue`? This makes CID management kind of complicated when dealing with retirement.

liwenjieQu:
> Another thought: Is it actually useful to have the `Duration` of every single CID be variable? This might be simpler if the lifetime were part of the connection configuration instead, making it constant for the lifetime of the connection (and also simplifying the CID generator interface).

This was also across my mind when i designed the API as it is impossible to assign different lifetime to CIDs in the same connection. But i end up with this format because 1. lifetime is more like an attribute of CID and thus it is more natural to get from a cid generator; 2. we need to consider again how to customize this configuration (like we do for cid generator).

liwenjieQu:
Addressed last comments. Working on a separate PR that includes only code for fixing bug when processing NEW_CONNECTION_ID frame

quinn-proto/src/connection/mod.rs
Ralith: No strong reason. If bundling them together makes things simpler/cleaner I'd be happy to see that happen.

liwenjieQu:
Update PR overview and merge from mainline

quinn-proto/src/connection/mod.rs
```diff
@@ -710,12 +722,33 @@ where
...
+                    });
+                } else {
+                    // Handle initial CID with active_connection_id_limit == 1
+                    if let Some(timestamp) = timeout {
+                        self.reset_cid_timer(Some(CidTimeStamp {
+                            sequence: 0,
+                            timestamp,
+                        }));
+                    }
+                }
```
djc, line 751: I think this is equivalent...

```suggestion
            NewIdentifiers(ids, now) => {
                // Record the timestamp of CID with the largest seq number
                let mut sequence = 0;
                if let Some(last_cid) = ids.last() {
                    self.cids_issued += ids.len() as u64;
                    ids.into_iter().rev().for_each(|frame| {
                        self.spaces[SpaceId::Data].pending.new_cids.push(frame);
                        self.cids_active_seq.insert(frame.sequence);
                    });
                    sequence = last_cid.sequence;
                }
                
                if let Some(timestamp) = self.cids_timeout.and_then(|t| now.checked_add(t)) {
                    self.reset_cid_timer(Some(CidTimeStamp {
                        sequence,
                        timestamp,
                    }));
                }
            }
```

quinn-proto/src/connection/mod.rs
```diff
@@ -2942,12 +3072,12 @@ where
             .saturating_sub(self.in_flight.bytes)
     }
 
-    /// Whether no timers but keepalive and idle are running
+    /// Whether no timers but keepalive, idle and pushnewcid are running
```
djc, line 3075: Looks like we no longer have the idle timer, so let's remove it from this comment?

quinn-proto/src/tests/util.rs
```diff
@@ -51,6 +51,25 @@ impl Pair {
         }
     }
 
+    pub fn new_from_endpoint(client: Endpoint, server: Endpoint) -> Self {
```
djc, line 54: Are there any other tests that could use this?

liwenjieQu: The reason for this change is to use separate configs to initialize client/server endpoints. Sometimes, when both client and server share the same config, the tracing log is pretty messy and becomes difficult for diagnosis. 

Current tests can switch to use this function but since they are all working fine, i don't see the motivation to make such change.

With that said, i will improve the code reuse as there are a lot of overlaps with `fn new`

quinn-proto/src/connection/mod.rs
liwenjieQu: Not sure I got what you mean, but from the latest code in mainline branch, the idle timer is still there.

https://github.com/quinn-rs/quinn/blob/main/quinn-proto/src/connection/timer.rs#L7-L8
and
https://github.com/quinn-rs/quinn/blob/main/quinn-proto/src/connection/mod.rs#L2956

quinn-proto/src/connection/mod.rs
liwenjieQu: Indeed! This is much clearer and idiomatic to Rust. Thx!

quinn-proto/src/tests/util.rs
```diff
@@ -33,22 +33,7 @@ impl Pair {
...
-        );
-        Self {
-            server: TestEndpoint::new(server, server_addr),
-            client: TestEndpoint::new(client, client_addr),
-            time: Instant::now(),
-            latency: Duration::new(0, 0),
-            spins: 0,
-            last_spin: false,
-        }
+        Pair::new_from_endpoint(client, server)
```
djc, line 36: This is great. 👍

quinn-proto/src/connection/mod.rs
djc: I was confused by the initial filter call only referencing `KeepAlive` and `PushNewCid`, but you're right. I think the comment here is not great, but that's a preexisting problem, I file a separate issue to clean it up after this PR lands.

quinn-proto/src/connection/mod.rs
```diff
@@ -760,10 +784,104 @@ where
                     self.path.challenge_pending = false;
                 }
                 Timer::Pacing => trace!("pacing timer expired"),
+                Timer::PushNewCid => {
+                    let num_pushed_cids = 1;
```
Ralith, line 788: Should this be a `const`?

quinn-proto/src/connection/mod.rs
```diff
@@ -760,10 +784,104 @@ where
...
+                        && self.retire_cid_seq - current_seq < num_pushed_cids
+                    {
+                        self.retire_cid_seq = current_seq + num_pushed_cids;
+                    }
+
+                    // In case we already received RETIRE_CONNECTION_ID before
+                    let mut push_new_cid = false;
+                    for id in current_seq..self.retire_cid_seq {
+                        push_new_cid = push_new_cid || self.cids_active_seq.contains(&id);
+                    }
```
Ralith, line 828: Use `.any(..)` as above.

quinn-proto/src/connection/mod.rs
```diff
@@ -760,10 +784,104 @@ where
...
+                    // An endpoint MAY
+                    // send connection IDs that temporarily exceed a peer's limit if the
+                    // NEW_CONNECTION_ID frame also requires the retirement of any excess,
+                    // by including a sufficiently large value in the Retire Prior To field.
+                    // https://tools.ietf.org/html/draft-ietf-quic-transport-29#section-5.1.1
+                    // In this case, `retire prior to` must increase to accommodate new cid
+                    let must_increase_retire_cid =
+                        self.cids_active_seq.len() as u64 >= self.peer_params.issue_cids_limit();
+                    if must_increase_retire_cid
+                        && self.retire_cid_seq - current_seq < num_pushed_cids
```
Ralith, line 819: What's the second clause here doing?

quinn-proto/src/connection/mod.rs
```diff
@@ -760,10 +784,104 @@ where
...
+                                .push_back(EndpointEventInner::NeedIdentifiers(
+                                    now,
+                                    num_pushed_cids,
+                                ));
+                        }
+                    } else {
+                        // We don't need to push new cid as peer has actively retired them already
+                        self.reset_cid_timer(None);
+                    }
+                }
```
Ralith, line 846: Overall I find the logic here difficult to follow. Can you add comments expanding on intentions throughout? Citing the spec is useful for explaining why we need to do something, but the part I'm having trouble with is understanding what we're actually doing here.

quinn-proto/src/connection/mod.rs
```diff
@@ -760,10 +784,104 @@ where
...
+                    if !ids_found {
+                        self.retired_prior_to_seq = self.retire_cid_seq;
+                    }
+
+                    let next_checkpoint = self.cids_timeline.pop_front();
+                    let next_seq = next_checkpoint.unwrap().sequence + 1;
+                    let current_seq = self.retire_cid_seq;
+                    //  Endpoints SHOULD NOT issue updates of the Retire Prior To field
+                    //  before receiving RETIRE_CONNECTION_ID frames that retire all
+                    //  connection IDs indicated by the previous Retire Prior To value.
```
Ralith, line 802: Stray leading whitespace

quinn-proto/src/connection/mod.rs
```diff
@@ -760,10 +784,104 @@ where
...
+                break;
+            }
+        }
+
+        if let Some(new_cid) = id {
+            if let Some(nc) = self.cids_timeline.front_mut() {
+                if new_cid.timestamp == nc.timestamp && new_cid.sequence > nc.sequence {
+                    nc.sequence = new_cid.sequence;
+                } else {
+                    self.cids_timeline.push_back(new_cid);
```
Ralith, line 867: Would a `debug_assert!(new_cid.timestamp > nc.timestamp)` make sense here?

quinn-proto/src/connection/mod.rs
```diff
@@ -760,10 +784,104 @@ where
...
+            if nc.sequence < self.retire_cid_seq {
+                self.cids_timeline.pop_front();
+            } else {
+                break;
+            }
+        }
+
+        if let Some(new_cid) = id {
+            if let Some(nc) = self.cids_timeline.front_mut() {
+                if new_cid.timestamp == nc.timestamp && new_cid.sequence > nc.sequence {
```
Ralith, line 864: Under what circumstances might `new_cid.sequence > nc.sequence` be false?

quinn-proto/src/connection/mod.rs
```diff
@@ -760,10 +784,104 @@ where
...
+                    self.cids_timeline.push_back(new_cid);
+                }
+            } else {
+                self.cids_timeline.push_back(new_cid);
+            }
+        }
+
+        if let Some(next_checkpoint) = self.cids_timeline.front() {
+            trace!(
+                "set up a timer at {:?} for cid {:?}",
```
Ralith, line 876:
```suggestion
                "cid {} will expire at {:?}"
```
If that's indeed the intention here. (note change of argument order)

quinn-proto/src/connection/mod.rs
```diff
@@ -2565,15 +2684,17 @@ where
...
             return;
         }
 
         // Subtract 1 to account for the CID we supplied while handshaking
         let n = self.peer_params.issue_cids_limit() - 1;
         self.endpoint_events
-            .push_back(EndpointEventInner::NeedIdentifiers(n));
+            .push_back(EndpointEventInner::NeedIdentifiers(now, n));
+        self.cids_issued += 1;
+        self.cids_active_seq.insert(0);
```
Ralith, line 2697: Were these two lines mistakenly introduced during the rebase?

Ralith (changes requested):
Thanks! Made a deep pass. Biggest issue is that I don't feel like I understand the core logic being introduced very well.

quinn-proto/src/connection/mod.rs
```diff
@@ -70,8 +70,18 @@ where
     handshake_cid: ConnectionId,
     /// The CID the peer initially chose, for use during the handshake
     rem_handshake_cid: ConnectionId,
+    /// Cid rotation
```
Ralith, line 73: If this comment is intended to describe the entire block of variables that follows, it should not be a doc comment (which will get attached specifically to the next field instead), should be preceded by an empty line, and should be bracketed by empty comments.

quinn-proto/src/connection/mod.rs
```diff
@@ -70,8 +70,18 @@ where
...
+    ///
+    /// Expiration timestamp of local connection IDs
+    cids_timeline: VecDeque<CidTimeStamp>,
```
Ralith, line 76: Please expand on which CIDs this contains (all issued CIDs the peer hasn't retired? the highest sequence number of every batch of CIDs that share an expiration time, and that time? something else?). Consider renaming the field/type for clarity as well.

Ralith (changes requested):

djc:
After some discussion with @Ralith, we'd like to ask that you move as much as possible of the CID logic out of the `Connection` type and into a separate type living in a separate module (for example, `CidState` in `connection/cid_state.rs`), to make it easier to review this in isolation (and to prevent the connection module from growing more, as it is already much larger than we would like it to be). Let me know if you have any questions about this.

liwenjieQu:
Hey Dirkjan and Benjamin, sorry for the late reply. I am currently occupied with something else. Just want to check in and let you know i haven't forgot this :-) Will try what you suggested in the next couple of days. 

Thx.

Ralith:
Much appreciated! No pressure; this is a good feature for us to have but it's not blocking anything.

quinn-proto/src/connection/mod.rs
liwenjieQu: Ack

quinn-proto/src/connection/mod.rs
liwenjieQu: Ack. When I will revisit this field definition now, i do feel it is lack of some explanation/comments. Will update in next revision.

liwenjieQu: To give a brief explanation here,  `cids_timeline` records the cid sequence number (attached with a timestamp that shows when this cid is created) that were issued but not yet retired by peer.

Specifically, since quinn may issue multiple cids in one batch, for each record in `cids_timeline`, it shows the highest cid sequence number (in a batch) and the corresponding timestamp (since those cids in a batch are created at the same time).

liwenjieQu: Correction: the timestamp in this struct is not when cid is created, but created + duration = when cid should be retired

quinn-proto/src/connection/mod.rs
```diff
@@ -70,8 +70,18 @@ where
...
+    /// The lifetime of Local connection IDs
+    cids_timeout: Option<Duration>,
```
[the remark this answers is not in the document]
liwenjieQu: > If, as I suspect, we don't need that level of flexibility, ....

I agree. 1. I don't think current RFC ever discusses to use a different duration to retire each CID or not which gives some flexibility in implementation. 2. A different duration will cause extra efforts to sort when each cid should be retired. 3. So far, the application scenario doesn't require a different duration for each CID.

So.... as you suggested, i can consider how to move it to either `TransportConfig`, or `EndpointConfig`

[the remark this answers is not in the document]
liwenjieQu: > we should remove the Option<Duration> from ConnectionIdGenerator entirely

However, i do want to argue against this point. I think `ConnectionIdGenerator` should own all fields related to CID so that a different lifetime can be applied depending on the implementation of `ConnectionIdGenerator`  trait object. Maybe `TransportConfig`, or `EndpointConfig` can be set by calling an API from `ConnectionIdGenerator`?

quinn-proto/src/connection/mod.rs
```diff
@@ -70,8 +70,18 @@ where
...
+    /// Number of local connection IDs that have been issued in NEW_CONNECTION_ID frames.
+    cids_issued: u64,
     /// The sequence numbers of local connection IDs not yet retired by the peer
     cids_active_seq: HashSet<u64>,
+    /// Sequence number prior to which all CIDs are already retired and acked in RETIRE_CONNECTION_ID frame
+    retired_prior_to_seq: u64,
```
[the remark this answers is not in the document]
liwenjieQu: Ack

quinn-proto/src/connection/mod.rs
```diff
@@ -710,12 +722,24 @@ where
...
+                        self.cids_active_seq.insert(frame.sequence);
+                    });
+                }
+
+                if let Some(timestamp) = self.cids_timeout.and_then(|t| now.checked_add(t)) {
+                    self.reset_cid_timer(Some(CidTimeStamp {
+                        sequence,
+                        timestamp,
+                    }));
+                }
```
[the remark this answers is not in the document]
liwenjieQu: Not exactly. `ids` can be `None` once we set the `peer_params.issue_cids_limit()` to 1 in the initial handshake. If so, `reset_cid_timer` will never be called to trigger cid rotation. 

I am adding more comments to explain such situation.

quinn-proto/src/connection/mod.rs
liwenjieQu: Ack

quinn-proto/src/connection/mod.rs
```diff
@@ -760,10 +784,104 @@ where
                     self.path.challenge_pending = false;
                 }
                 Timer::Pacing => trace!("pacing timer expired"),
+                Timer::PushNewCid => {
+                    let num_pushed_cids = 1;
+                    let mut ids_found = false;
+                    for id in self.retired_prior_to_seq..self.retire_cid_seq {
+                        ids_found |= self.cids_active_seq.contains(&id);
+                    }
```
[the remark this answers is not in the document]
liwenjieQu: Ack

quinn-proto/src/connection/mod.rs
```diff
@@ -760,10 +784,104 @@ where
...
+                    if !ids_found {
+                        self.retired_prior_to_seq = self.retire_cid_seq;
+                    }
+
+                    let next_checkpoint = self.cids_timeline.pop_front();
+                    let next_seq = next_checkpoint.unwrap().sequence + 1;
```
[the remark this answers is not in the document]
liwenjieQu: `Timer::PushNewCid` is set when `if let Some(next_checkpoint) = self.cids_retire_timestamp.front()` so we know there is at least an element in `cids_timeline `

quinn-proto/src/connection/mod.rs
liwenjieQu: Ack

quinn-proto/src/connection/mod.rs
liwenjieQu:
```
let must_increase_retire_cid = self.cids_active_seq.len() as u64 >= self.peer_params.issue_cids_limit();
```
Check if the current number of active cid is already the maximum set by peer. It is possible that PushNewCID timer fires but we haven't received any RETIRE_CONNECTION_ID frame from peer. 

There are two statements in the RFC: 1. Endpoints SHOULD NOT issue updates of the Retire Prior To field before receiving RETIRE_CONNECTION_ID frames that retire all connection IDs indicated by the previous Retire Prior To value. 2. An endpoint MUST NOT provide more connection IDs than the peer's limit.

In the above case, we have to retire more cids since the `PushNewCID` timer fires which causes some contradiction between  the statement 1 and 2. From one side, we cannot advance `retire_prior_to` because of (1) but meanwhile we have to advance this field because of (2) since all unretired cids are considered active. Since MUST overrides SHOULD, and there is another RFC statement `An endpoint MAY send connection IDs that temporarily exceed a peer's limit if the NEW_CONNECTION_ID frame also requires the retirement of any excess`, we use `must_increase_retire_cid ` to indicate whether we end up with this awkward situation and have to increase the value in the field without other choice.

liwenjieQu:
```
 if must_increase_retire_cid && self.retire_cid_seq - current_seq < num_pushed_cids
```

This is a safety check which makes sure the difference between last and current `retire_perior_to` is more than `num_pushed_cids `. It means we won't issue more cids than retired.

liwenjieQu: I have to admit the naming is messy and somehow awful. Working on a new version with better readability.

quinn-proto/src/connection/mod.rs
liwenjieQu: Ack.

quinn-proto/src/connection/mod.rs
liwenjieQu: This was possibly because i don't fully know the details of event handling in quinn. If all `ConnectionEvent` are managed by a VecDeque, we shouldn't need this condition.

quinn-proto/src/connection/mod.rs
liwenjieQu: Yes.

quinn-proto/src/connection/mod.rs
liwenjieQu: Ack

quinn-proto/src/connection/mod.rs
liwenjieQu: Ahh yes. My mistake.

liwenjieQu:
I agree that a significant refactoring will increase the code readability and address many detailed questions that Benjamin raised. Working on a revision now.

quinn-proto/src/connection/mod.rs
liwenjieQu: I think the processing logic is too complicated for a realistic scenario. Will simplify in the next iteration

quinn-proto/src/cid_generator.rs
```diff
@@ -7,17 +7,18 @@ use crate::MAX_CID_SIZE;
...
     /// connection.
-    ///
-    /// In the future, connection IDs will be retired after the returned `Duration`, if any.
-    fn generate_cid(&mut self) -> (ConnectionId, Option<Duration>);
+    fn generate_cid(&mut self) -> ConnectionId;
     /// Returns the length of a CID for cononections created by this generator
     fn cid_len(&self) -> usize;
+    /// Returns the lifetime of generated Connection IDs
+    /// Connection IDs will be retired after the returned `Duration`, if any
+    fn cid_lifetime(&self) -> Option<Duration>;
```
Matthias247, line 21: I'm not sure whether it makes things easier or not. But another idea to express "no lifetime" is to pick a very high or maximum duration (like [`Duration::MAX`](https://doc.rust-lang.org/std/time/struct.Duration.html#associatedconstant.MAX) - which is unfortunately not stable), and use this instead of `Option<Duration>`.

This can avoid having to deal with `Option`s everywhere.
Obviously requires all operations to properly use saturating adds.

quinn-proto/src/connection/mod.rs
```diff
@@ -201,9 +201,15 @@ where
             crypto,
             handshake_cid: loc_cid,
             rem_handshake_cid: rem_cid,
-            retire_cid_seq: 0,
-            cids_active_seq: HashSet::new(),
-            local_cid_len,
+            local_cid_state: CidState {
```
Matthias247, line 204: I would recommend moving this into a `CidState::new()`, since there are quite a lot of initializations that are unrelated to the `Connection`

quinn-proto/src/connection/cid_state.rs
```diff
@@ -0,0 +1,103 @@
...
+        self.retire_timestamp.front().map(|nc| {
+            trace!("CID {} will expire at {:?}", nc.sequence, nc.timestamp);
+            nc.timestamp
+        })
+    }
+
+    /// Track the lifetime of issued cids in `retire_timestamp`
+    pub(crate) fn track_lifetime(&mut self, new_cid_seq: u64, now: Instant) {
+        match self.cid_lifetime {
+            None => {}
```
Matthias247, line 49: Maybe just return early, to avoid everything drifting right:

```rust
let lifetime = match self.cid_lifetime {
    Some(lifetime) => lifetime,
    None => return,
};
```

quinn-proto/src/connection/cid_state.rs
```diff
@@ -0,0 +1,103 @@
...
+            Some(lifetime) => {
+                let expire_timestamp = now.checked_add(lifetime);
```
Matthias247, line 51: Could a `saturating_add` also work and avoid the `if let` condition?

quinn-proto/src/connection/cid_state.rs
```diff
@@ -0,0 +1,103 @@
...
+                        sequence: new_cid_seq,
+                        timestamp: expire_at,
+                    });
+                }
+            }
+        }
+    }
+
+    /// Update local CID state when previously issued CID is retired
+    pub(crate) fn on_cid_retirement_timeout(&mut self) -> bool {
```
Matthias247, line 73: Maybe document what this returns. It's not obvious from a callback what its return value signals. Especially not if the type is a as generic as `bool`

quinn-proto/src/cid_generator.rs
djc: I find it a bit of a mixed bag, as I feel like the `Option` makes the intent more clear than having a sentinel `Duration` value.

djc:
(@Matthias247 thanks for the review, good points!)

quinn-proto/src/cid_generator.rs
liwenjieQu: I agree with Dirkjan on this topic that a `Option` shows the intention clearer.

quinn-proto/src/connection/cid_state.rs
liwenjieQu: However, there is no `saturating_add ` for `Instant` type.

quinn-proto/src/connection/cid_state.rs
liwenjieQu: good point

liwenjieQu:
Hi Benjamin @Ralith, may i ask you to look at the update when you are convenient? I am happy to explain if there is any logic that is still unclear after the recent change.

Ralith:
Sorry this has languished a little; dayjob's been taking most of my energy. I'll make time in the next few days, one way or another!

liwenjieQu:
Much appreciated Benjamin! No pressure.

quinn-proto/src/connection/mod.rs
```diff
@@ -710,12 +700,23 @@ where
...
+                if let Some(last_cid) = ids.last() {
+                    self.local_cid_state.issued += ids.len() as u64;
+                    // Record the timestamp of CID with the largest seq number
+                    let sequence = last_cid.sequence;
+                    ids.into_iter().rev().for_each(|frame| {
+                        self.spaces[SpaceId::Data].pending.new_cids.push(frame);
+                        self.local_cid_state.active_seq.insert(frame.sequence);
+                    });
+                    self.local_cid_state.track_lifetime(sequence, now);
+                }
```
djc, line 714: This is manipulating an awful lot of `self.local_cid_state` stuff. Should we move more of this code into a `CidState` method?

quinn-proto/src/connection/mod.rs
```diff
@@ -760,6 +761,18 @@ where
...
+                    // Update `retire_prior_to` field in NEW_CONNECTION_ID frame
+                    let num_new_cid = self.local_cid_state.on_cid_retirement_timeout().into();
+                    if !self.state.is_closed() {
+                        trace!(
+                            "push a new cid to peer RETIRE_PRIOR_TO field {}",
+                            self.local_cid_state.retire_seq
+                        );
+                        self.endpoint_events
+                            .push_back(EndpointEventInner::NeedIdentifiers(now, num_new_cid));
+                    }
```
djc, line 774: This could also be moved around a little to make `CidState` be more of an abstraction.

quinn-proto/src/connection/mod.rs
```diff
@@ -1496,6 +1509,13 @@ where
         self.timers.set(Timer::KeepAlive, now + interval);
     }
 
+    fn reset_cid_retirement(&mut self) {
+        match self.local_cid_state.next_timeout() {
+            None => {}
+            Some(t) => self.timers.set(Timer::PushNewCid, t),
+        };
```
djc, line 1516:
```suggestion
        if let Some(t) = self.local_cid_state.next_timeout() {
            self.timers.set(Timer::PushNewCid, t);
        }
```

quinn-proto/src/connection/mod.rs
```diff
@@ -2352,12 +2372,12 @@ where
                     self.streams.received_stop_sending(id, error_code);
                 }
                 Frame::RetireConnectionId { sequence } => {
-                    if self.local_cid_len == 0 {
+                    if self.local_cid_state.cid_len == 0 {
```
djc, line 2375: As above, we should push more of this code into a `CidState` method.

quinn-proto/src/connection/mod.rs
```diff
@@ -2965,16 +2988,24 @@ where
...
     #[cfg(test)]
-    pub(crate) fn rotate_local_cid(&mut self, v: u64) {
+    pub(crate) fn rotate_local_cid(&mut self, v: u64, now: Instant) {
         // Cannot retire more CIDs than what have been issued
-        debug_assert!(v <= *self.cids_active_seq.iter().max().unwrap() + 1);
-        let n = v.checked_sub(self.retire_cid_seq).unwrap();
-        self.retire_cid_seq = v;
+        debug_assert!(v <= *self.local_cid_state.active_seq.iter().max().unwrap() + 1);
+        let n = v.checked_sub(self.local_cid_state.retire_seq).unwrap();
+        self.local_cid_state.retire_seq = v;
```
djc, line 3006: And here.

djc (commented):
In the current state of the PR, the `CidState` abstraction seems more leaky than necessary. See below where I point out some places where we could move more code into `CidState` to improve the abstraction here, which I think will make things easier to reason about.

quinn-proto/src/connection/mod.rs
liwenjieQu: Ack

quinn-proto/src/connection/mod.rs
liwenjieQu: Related logic is already moved into `on_cid_retirement_timeout`. I don't see more we can move to increase abstraction here.

djc: I guess I was wondering about the `trace!()` call here, but that might not be practical.

liwenjieQu:
Did some refactoring to move as much local cid related code into `CidState`

quinn-proto/src/cid_state.rs
```diff
@@ -0,0 +1,196 @@
+//! Maintain the state of local connection IDs
```
djc, line 1: This module should be inside quinn-proto's `connection` module, I think.

quinn-proto/src/cid_state.rs
```diff
@@ -0,0 +1,196 @@
...
+use crate::{shared::IssuedCid, TransportError};
+use std::{
+    collections::{HashSet, VecDeque},
+    time::{Duration, Instant},
+};
+use tracing::{debug, trace};
```
djc, line 7:
```suggestion
use std::{
    collections::{HashSet, VecDeque},
    time::{Duration, Instant},
};

use tracing::{debug, trace};

use crate::{shared::IssuedCid, TransportError};
```

quinn-proto/src/cid_state.rs
```diff
@@ -0,0 +1,196 @@
...
+
+/// Data structure that records when issued cids should be retired
+#[derive(Copy, Clone, Eq, PartialEq)]
+struct CidTimeStamp {
+    /// Highest cid sequence number created in a batch
+    sequence: u64,
+    /// Timestamp when cid needs to be retired
+    timestamp: Instant,
+}
```
djc, line 16: Move this to the bottom of the module, please.

quinn-proto/src/cid_state.rs
```diff
@@ -0,0 +1,196 @@
...
+            issued: 1, // One CID is already supplied during handshaking
+            active_seq: HashSet::new(),
+            prev_retire_seq: 0,
+            retire_seq: 0,
+            cid_len,
+            cid_lifetime,
+        };
+        // Add sequence number of CID used in handshaking into tracking set
+        this.active_seq.insert(0);
+        this
```
djc, line 51:
```suggestion
        let mut active_seq = HashSet::new();
        // Add sequence number of CID used in handshaking into tracking set
        active_seq.insert(0);
        CidState {
            retire_timestamp: VecDeque::new(),
            issued: 1, // One CID is already supplied during handshaking
            active_seq,
            prev_retire_seq: 0,
            retire_seq: 0,
            cid_len,
            cid_lifetime,
        }
```

quinn-proto/src/cid_state.rs
```diff
@@ -0,0 +1,196 @@
...
+        // `ids` could be `None` once active_connection_id_limit is set to 1 by peer
+        if let Some(last_cid) = ids.last() {
+            self.issued += ids.len() as u64;
+            // Record the timestamp of CID with the largest seq number
+            let sequence = last_cid.sequence;
+            ids.iter().for_each(|frame| {
+                self.active_seq.insert(frame.sequence);
+            });
+            self.track_lifetime(sequence, now);
+        }
```
djc, line 134:
```suggestion
        // `ids` could be `None` once active_connection_id_limit is set to 1 by peer
        let last_cid = match ids.last() {
            Some(cid) => cid,
            None => return,
        };
        self.issued += ids.len() as u64;
        // Record the timestamp of CID with the largest seq number
        let sequence = last_cid.sequence;
        ids.iter().for_each(|frame| {
             self.active_seq.insert(frame.sequence);
        });
        self.track_lifetime(sequence, now);
```

quinn-proto/src/cid_state.rs
```diff
@@ -0,0 +1,196 @@
...
+        self.active_seq.remove(&sequence);
+        // Consider a scenario where peer A has active remote cid 0,1,2.
+        // Peer B first send a NEW_CONNECTION_ID with cid 3 and retire_prior_to set to 1.
+        // Peer A processes this NEW_CONNECTION_ID frame; update remote cid to 1,2,3
+        // and meanwhile send a RETIRE_CONNECTION_ID to retire cid 0 to peer B.
+        // If peer B doesn't check the cid limit here and send a new cid again, peer A will then face CONNECTION_ID_LIMIT_ERROR
+        let allow_more_cids = limit > self.active_seq.len() as u64;
+
+        Ok(allow_more_cids)
```
djc, line 167:
```suggestion
        Ok(limit > self.active_seq.len() as u64)
```

quinn-proto/src/cid_state.rs
```diff
@@ -0,0 +1,196 @@
...
+        self.cid_len
+    }
+
+    /// The value for `retire_prior_to` field in `NEW_CONNECTION_ID` frame
+    pub(crate) fn retire_prior_to(&self) -> u64 {
+        self.retire_seq
+    }
+
+    #[cfg(test)]
+    pub(crate) fn active_seq(&self) -> (u64, u64) {
```
djc, line 181: Maybe this should return a `Range`? Also iterating the same set twice seems a little ugly/inefficient, maybe just write an old-fashioned for-loop that tracks both minimum and maximum?

djc (commented):
This is getting pretty close, some more style nits!

quinn-proto/src/cid_state.rs
liwenjieQu: Ah i see. Make sense

liwenjieQu: Essentially, we don't know the number in this hashset is continuous or not (so `Range` probably is not a good option here), and we only need the max and min from the set.

quinn-proto/src/connection/mod.rs
```diff
@@ -2565,15 +2559,17 @@ where
...
             return;
         }
 
         // Subtract 1 to account for the CID we supplied while handshaking
         let n = self.peer_params.issue_cids_limit() - 1;
         self.endpoint_events
-            .push_back(EndpointEventInner::NeedIdentifiers(n));
+            .push_back(EndpointEventInner::NeedIdentifiers(now, n));
+        // Track lifetime of cid used in handshaking
+        self.local_cid_state.track_lifetime(0, now);
```
Ralith, line 2572: I don't think the difference between the time the `Connection` is created and the time when CIDs are issued is significant enough to justify doing this here. Instead, let's set up the state for the handshake CID in `CidState::new`, and thereby encapsulate `CidState` a bit more tightly.

quinn-proto/src/connection/cid_state.rs
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        // Check if retirement of all CIDs that reach their lifetime is still needed
+        // If yes (return true), a new CID must be pushed with updated `retire_prior_to` field to remote peer.
+        // If no (return false), it means remote peer has proactively retired those CIDs (for other reasons) before CID lifetime is reached.
+        (current_retire_prior_to..self.retire_seq).any(|seq| self.active_seq.contains(&seq))
+    }
+
+    /// Update cid state when `NewIdentifiers` event is received
+    pub(crate) fn new_cids(&mut self, ids: &[IssuedCid], now: Instant) {
```
Ralith, line 117: nit:
```suggestion
    pub(crate) fn insert(&mut self, ids: &[IssuedCid], now: Instant) {
```

quinn-proto/src/connection/cid_state.rs
```diff
@@ -0,0 +1,205 @@
...
+            }
+            self.retire_timestamp.push_back(CidTimeStamp {
+                sequence: new_cid_seq,
+                timestamp: expire_at,
+            });
+        }
+    }
+
+    /// Update local CID state when previously issued CID is retired
+    /// Return a flag that indicates whether a new CID needs to be pushed that notifies remote peer to respond `RETIRE_CONNECTION_ID`
```
Ralith, line 81:
```suggestion
    /// Update local CID state when previously issued CID is retired
    ///
    /// Return whether a new CID needs to be pushed that notifies remote peer to respond `RETIRE_CONNECTION_ID`
```

quinn-proto/src/connection/cid_state.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        let next_retire_sequence = self
+            .retire_timestamp
+            .pop_front()
+            .map(|seq| seq.sequence + 1);
+        let current_retire_prior_to = self.retire_seq;
+
+        // Advance `retire_cid_seq` if next cid that needs to be retired exists
+        if let Some(next_retire_prior_to) = next_retire_sequence {
+            if !unretired_ids_found && next_retire_prior_to > current_retire_prior_to {
```
Ralith, line 105: Is it ever possible that `next_retire_prior_to < current_retire_prior_to`? If not, `!=` would be clearer here.

quinn-proto/src/connection/cid_state.rs
```diff
@@ -0,0 +1,205 @@
...
+                self.retire_seq = next_retire_prior_to;
+            }
+        }
+
+        // Check if retirement of all CIDs that reach their lifetime is still needed
+        // If yes (return true), a new CID must be pushed with updated `retire_prior_to` field to remote peer.
+        // If no (return false), it means remote peer has proactively retired those CIDs (for other reasons) before CID lifetime is reached.
+        (current_retire_prior_to..self.retire_seq).any(|seq| self.active_seq.contains(&seq))
```
Ralith, line 113: This doesn't appear to depend in any way on the actual timestamps. What prevents this from issuing retire_prior_to for every single active CID, including those which should not yet be retired?

quinn-proto/src/connection/cid_state.rs
```diff
@@ -0,0 +1,205 @@
...
+        let sequence = last_cid.sequence;
+        ids.iter().for_each(|frame| {
+            self.active_seq.insert(frame.sequence);
+        });
+        self.track_lifetime(sequence, now);
+    }
+
+    /// Update CidState when recieve a `RETIRE_CONNECTION_ID` frame
+    /// Return a boolean variable or `TransportError` if CID content violates RFC
+    /// When boolean variable is `true`, a new CID should be issued to peer.
```
Ralith, line 134: Style/grammar nits:
```suggestion
    /// Update CidState for receipt of a `RETIRE_CONNECTION_ID` frame
    ///
    /// Returns whether a new CID can be issued, or an error if the frame was illegal.
```

quinn-proto/src/endpoint.rs
```diff
@@ -111,8 +111,10 @@ where
     ) -> Option<ConnectionEvent> {
         use EndpointEventInner::*;
         match event.0 {
-            NeedIdentifiers(n) => {
-                return Some(self.send_new_identifiers(ch, n));
+            NeedIdentifiers(now, n) => {
+                if self.connections.get(ch.into()).is_some() {
```
Ralith, line 115: Why was this guard introduced?

quinn-proto/src/connection/timer.rs
```diff
@@ -54,7 +57,7 @@ impl TimerTable {
         self.data.iter().filter_map(|&x| x).min()
     }
 
-    pub fn is_expired(&self, timer: Timer, after: Instant) -> bool {
-        self.data[timer as usize].map_or(false, |x| x <= after)
+    pub fn is_expired(&self, timer: Timer, after: Instant, default_none: bool) -> bool {
```
Ralith, line 60: The new argument here is detrimental to readability. Instead, let's remove `#[cfg(test)]` from `pub fn get` above and use that when more specialized behavior is required.

Ralith (changes requested):
Thanks for your patience, and for this continued work! I'm a big fan of the new encapsulation; that makes everything a lot easier to follow.

quinn-proto/src/connection/cid_state.rs
liwenjieQu: Nope. I don't think it is possible for this field to decrease.

quinn-proto/src/connection/mod.rs
liwenjieQu: Agree.

liwenjieQu: I recall why i put it here in the first place. It was because UT uses artificial time and thus we have to pass `now` to this function to track the timeline. Moving this to `new` of `CidState` will make `CidState` also dependent on `now`.

Ralith: Is that a problem? `now` is already available in `Connection::new`, where `CidState` would be constructed.

quinn-proto/src/connection/cid_state.rs
liwenjieQu: I am not sure if i catch your point.

The return of this call determines whether the peer wants to publish a new CID with updated `retire_prior_to` field. This is allowed in RFC described as follows

```
 An endpoint MUST NOT
   provide more connection IDs than the peer's limit.  An endpoint MAY
   send connection IDs that temporarily exceed a peer's limit if the
   NEW_CONNECTION_ID frame also requires the retirement of any excess,
   by including a sufficiently large value in the Retire Prior To field.
```
When there is no active CID to retire (after we update the `retire_prior_to`  in this call), pushing a new CID may violate the above RFC statement. That's why a condition check is added here.

quinn-proto/src/connection/mod.rs
liwenjieQu: Right. I didn't finish the response. No issue to move forward :-)

quinn-proto/src/connection/timer.rs
liwenjieQu: Ack. :-) I remember that i changed between these two ways several times

quinn-proto/src/endpoint.rs
liwenjieQu: I remember the condition check was added there to prevent connection is draining while the PushNewCid timer alarms. I am moving this check into `connection/mod.rs` before pushing the `NeedIdentifiers` event

liwenjieQu: Ah... i think `is_closed` already covers the `Draining`. We should be good to remove this check

quinn-proto/src/connection/cid_state.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        let next_retire_sequence = self
+            .retire_timestamp
+            .pop_front()
+            .map(|seq| seq.sequence + 1);
+        let current_retire_prior_to = self.retire_seq;
+
+        // Advance `retire_cid_seq` if next cid that needs to be retired exists
+        if let Some(next_retire_prior_to) = next_retire_sequence {
+            if !unretired_ids_found && next_retire_prior_to != current_retire_prior_to {
```
Ralith, line 107: Given that `current_retire_prior_to` is the current value of `self.retire_seq`, what do we gain by checking that the next value differs from it? Seems like if they're the same then the assignment should be a harmless no-op.

Ralith (changes requested):

quinn-proto/src/connection/cid_state.rs
Ralith: I think I was confused about when this method was called. On review I don't see anything wrong here.

quinn-proto/src/connection/cid_state.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// Find the next timestamp when previously issued CID should be retired
+    pub(crate) fn next_timeout(&mut self) -> Option<Instant> {
+        self.retire_timestamp.front().map(|nc| {
+            trace!("CID {} will expire at {:?}", nc.sequence, nc.timestamp);
+            nc.timestamp
+        })
+    }
+
+    /// Track the lifetime of issued cids in `retire_timestamp`
+    pub(crate) fn track_lifetime(&mut self, new_cid_seq: u64, now: Instant) {
```
Ralith, line 57: I think this is purely internal now:
```suggestion
    fn track_lifetime(&mut self, new_cid_seq: u64, now: Instant) {
```

quinn-proto/src/connection/cid_state.rs
liwenjieQu: haha you are right :-)

Good find!

Ralith (approved):
This looks great, thanks! I really appreciate your patience through these review cycles.

liwenjieQu:
Fix some conflicts when merge to mainline.

Thank you very much @Ralith @djc  for all the support and comments :D

Happy new year!

djc (approved):
Thanks for persisting with this, going to merge it now!

PR #866, opened 2020-10-12
Readme update

TimonPost:
## Live version
https://github.com/TimonPost/quinn/tree/timon/read-me-update

## Changes
**- Grammer fixes**
**- Remove Usage Notes, will be moved into the book.**
**- Add project logo**
**- Add 'Getting Started' section with suggested links and short instructions.**
**- Update draft version**
**- Transform a list of library-creates into the table.**
**- Combine features and status into one single feature-list**
**- Add Authors and License section**

You can open the logo file with https://www.photopea.com/  which is free without accounts and or licenses. 

## Discussion
- ~I removed the cryptography and buffer section since it will be added to the book.~
- ~The features list and the status list seemed very similar to me, I think they can be combined into one single list, perhaps there are reasons to keep it as the old way?~


djc:
I kind of like the logo, though I'd want to look into copyright/trademarks a bit more. Please don't remove any content from the README until it's added back somewhere else, though (as I mentioned already in the tracking issue).

We should probably update the statement about it being ready "for experimental use". Also, we're at draft 29 now. The `Cargo.toml` doesn't make much sense to collapse to me, since it's so short, but maybe it doesn't make sense to have at all, since it's pretty obvious to any potential user.

Merging status and features somehow probably also makes sense.

Agree that we shouldn't spend much space/time on my talk and or the history of the merge anymore.

TimonPost:
Good point! I will remove the cargo toml, it is easy for just quickly copying it, however, rust users will be familiar with where to find it. 

So that's an important point that you mentioned: "for experimental use", what would you like to see in this place? In what stage is Quinn? Like QUIC is already used by more than half of all connections from the Chrome web browser to Google's servers. Microsoft Edge and Firefox support it.  And I heard it is possible to use it for youtube as well.

TimonPost:
The logo is used over here: https://quicwg.org/, but I can find it on many other blog posts or quic related websites.

Over here are the original resources:

https://github.com/quicwg/wg-materials/tree/master/badge

There is a commit 'Delted LICENSE'

https://github.com/quicwg/wg-materials/commit/9749588991177bf9a6ab24a08dbfa54927c42ef9

So it seems like there is no license on it, hence we can use it until they have. 

I replied:

https://github.com/quicwg/wg-materials/commit/9749588991177bf9a6ab24a08dbfa54927c42ef9#commitcomment-43176318

Ralith:
> So it seems like there is no license on it, hence we can use it until they have.

Unfortunately, copyright law dictates that if there is no license specified, you do *not* have permission to do anything with it.

Ralith:
Maybe it's time to get rid of the fine-grained feature list, also? That dates from when most QUIC implementations were significantly incomplete; these days if you just say "quinn is a QUIC implementation" then the rest is implied.

djc:
You mean the what's currently the Status list? Yeah, I think getting rid of that makes sense. But let's keep what we currently have as Features, though we should add a note that our HTTP/3 support is still experimental.

TimonPost:
Good point @Ralith, I will fiddle around with some other logo-ideas such that we will not have to deal with those copyright issues. So, Removing the status list, keeping the features list? Perhaps we can add '- [ ] HTTP over QUIC' to the features list. If there are points to be added or removed from the features list, please let me know.

djc:
I think we should remove the checkboxes in this case, since they're all checked anyway.

djc:
So reviewing the new version you have so far:

* I don't think we should spend so much vertical space on the logo. That's prime real estate wasted on a lot of whitespace.

* I think we should probably move what's currently called "Features" into the top level introduction, before "Getting Started". This answers the all-important question -- what is this project (and to some extent, what are its values).

* I think overview comes next: this answers the question, how is the underlying design available to users, and what layer would make sense for the reader's use case. I'm not sure the table is adding much value here; I think I personally prefer the bullet point list from the old version, though we could make the crate names bold rather than monospace to make them stand out a bit more. (If we're going to keep the table, I think each description should start with a capital letter and end with a period.)

* After this, the getting started. Probably prioritize the Examples over the Links for now, as it seems more actionable.

* I think we should add @stammw as an author, maybe as "Project collaborator, author of the HTTP/3 implementation"?

* I don't think we need to mention the licensing again in the text. It's present in the badges up front and also linked from the sidebar, and we fit in with the ecosystem standard anyway.

TimonPost:
Thanks for the review, I updated all your feedback points. As for the overview table. I also added a new logo. Tough, please let me know if it suits quinn or you would like to see some other variation. 

image

TimonPost:
Is this good to go now, any opinions on the logo and or content?

djc:
I have some nits about the logo. Mainly the spacing seems pretty uneven, in particular the distance between Q and u is much larger than between the other letters. I also wonder how the space between the Q and the circle relates to other spacing (maybe should be a bit more), and I feel that the whitespace on the inside of the circle (between the circle and the other thing) should increase a little bit. It also looks like the horizontal white line throughout the letters is thicker than the spacing between the letters, which feels wrong.

Also, please add the logo as an SVG. I don't have Photoshop and am unlikely to have access.

(Honestly, the logo doesn't seem all that important to me, so focusing on the content first and leaving the logo for later also seems like a fine option to me.)

I think you didn't really respond to my previous remark about table vs bullet list for the overview.

Let's not have periods after the Links items, and only have a single capital per bullet there. Also, lower-case Owner in the Authors section.

TimonPost:
Thanks for the second review! 

> Logo

Yes to a certain degree the logo is a secondary item. Tough I would really recommend in my view it makes the library look way more professional even if it is just a mental trick it works. I will make the changes as described. 

> I don't have Photoshop and am unlikely to have access.
I use https://www.photopea.com/ which is free and almost identical to photoshop

> I think you didn't really respond to my previous remark about table vs bullet list for the overview.

Ah, correct. I will make the changes as described. 

> Let's not have periods after the Links items, and only have a single capital per bullet there. Also, lower-case Owner in the Authors section.

Perfect, will do 👍

djc:
> I use https://www.photopea.com/ which is free and almost identical to photoshop

It looks like that can also export to SVG, so let's use that and store it as SVG. SVG is scalable and can be used on the web as-is, whereas PSD is a proprietary file format that cannot be used on the web.

> For the lower-case 'Owner' I think we should maintain the English grammar rules and maintain the rules for capitalized titles. Thus instead capitalizing Creator and Collaborator which is currently not correct.

I don't see how it makes sense to use title capitalization for the project roles.

TimonPost:
Cool, I will export it to SVG 👍 

> I don't see how it makes sense to use title capitalization for the project roles.

It makes sense as in that it is according to the English grammar rules which state that (job) titles that should begin capitalized every other word. You can check sites like https://titlecaseconverter.com/, https://capitalizemytitle.com/, out to validate this grammar.

djc:
> It makes sense as in that it is according to the English grammar rules which state that (job) titles that should begin capitalized every other word. You can check sites like https://titlecaseconverter.com/, https://capitalizemytitle.com/, out to validate this grammar.

@Ralith you maybe reluctant to weigh in on this particular bikeshedding argument, do you have an opinion as the only native speaker on our team?

README.md
```diff
@@ -91,27 +94,26 @@ trust-on-first-use, servers that automatically generate self-signed certificates
...
-working directory, with the client fetching `./Cargo.toml`. By default, the
-server generates a self-signed certificate and stores it to disk, where the
-client will automatically find and trust it.
-
-## Development
-
-The quinn-proto test suite uses simulated IO for reproducibility and to avoid
+- All feedback welcome. Feel free to file bugs, requests for documentation and
+  any other feedback to the [issue tracker][issues].
+- The quinn-proto test suite uses simulated IO for reproducibility and to avoid
```
Ralith, line 104: I'd drop the bullet points here in favor of a paragraph break.

Ralith (dismissed):
I agree with @djc that the terms in question do not constitute formal job titles; rather, they're plain adjectives, which should not be capitalized.

README.md
djc: @TimonPost if you process this I think this is ready to merge!

TimonPost: @djc I progressed this in 1fa2a43

djc: On https://github.com/TimonPost/quinn/tree/timon/read-me-update I still see the bullet points in the Contribution section.

TimonPost: o whoops my bad. I changed the wrong bullet points I see

djc: Ah yeah, please change those back. :)

TimonPost: Fixed it 👍

djc (approved):

djc:
Thanks for sticking with it, this is a good change!

PR #871, opened 2020-10-16
Bug Fix when accepting and retiring connection ID

liwenjieQu:
As requested in https://github.com/quinn-rs/quinn/pull/860, this PR extracts code change particularly for fixing existing bugs when processing cid retirement.

Currently in quinn-proto, retire_prior_to field in `NEW_CONNECTION_ID` frame is constantly set to 0 which means the logic for cid retirement is not properly tested and thus a few bugs are found.

1. quinn-proto maintains both `rem_cid` (as active remote cid) and `rem_cids` (in CidQueue as cid candidates). In a corner case, we may need to pick one candidate from `rem_cids` to replace `rem_cid`. However, current code cannot properly handle this case as `update_rem_cid` is always called after `insert` function of CidQueue. Please follow the added unit test to understand the details.

2. Based on RFC, "An endpoint SHOULD supply a new connection ID when the peer retires a connection ID". quinn-proto just issues new cid without any condition check. Meanwhile in RFC, there is also following statement "An endpoint MUST NOT provide more connection IDs than the peer's limit". Since "MUST" overrides "SHOULD", [condition check](https://github.com/quinn-rs/quinn/compare/main...liwenjieQu:retire_cid?expand=1#diff-423ed266e9aaf5fcae0fa8b85ca382f495528c18cd7c890cef1cf438221f2ef6R2363-R2374) is needed to avoid creating NEW_CONNECTION_ID frame endlessly.

### DIFF
--- a/quinn-proto/src/cid_queue.rs
+++ b/quinn-proto/src/cid_queue.rs
@@ -8,29 +8,30 @@ use crate::{shared::IssuedCid, ConnectionId, ResetToken};
 #[derive(Debug)]
 pub struct CidQueue {
     /// Ring buffer indexed by `self.cursor`
-    buffer: [Option<(ConnectionId, ResetToken)>; Self::LEN],
+    buffer: [Option<(ConnectionId, Option<ResetToken>)>; Self::LEN],
     /// Index at which circular buffer addressing is based
     cursor: usize,
     /// Sequence number of `self.buffer[cursor]`
     ///
-    /// The CID sequenced immediately prior to this is the active CID, which this data structure is
-    /// not responsible for retiring.
+    /// The sequence of active (in use) CID. This sequence is also the smallest among CIDs in `buffer`.
     offset: u64,
 }
 
 impl CidQueue {
-    pub const LEN: usize = 4;
+    pub const LEN: usize = 5;
 
-    pub fn new(offset: u64) -> Self {
-        Self {
+    pub fn new(cid: ConnectionId) -> Self {
+        let mut this = Self {
             buffer: [None; Self::LEN],
             cursor: 0,
-            offset,
-        }
+            offset: 0,
+        };
+        this.buffer[this.cursor] = Some((cid, None));
+        this
     }
 
     pub fn insert(&mut self, cid: IssuedCid) -> Result<(), InsertError> {
-        if cid.sequence + 1 == self.offset {
+        if cid.sequence == self.offset && self.buffer[self.cursor].is_some() {
             // This is a duplicate of the active CID.
             return Ok(());
         }
@@ -42,7 +43,7 @@ impl CidQueue {
             return Err(InsertError::ExceedsLimit);
         }
         let index = (self.cursor + index as usize) % Self::LEN;
-        self.buffer[index] = Some((cid.id, cid.reset_token));
+        self.buffer[index] = Some((cid.id, Some(cid.reset_token)));
         Ok(())
     }
 
@@ -63,35 +64,43 @@ impl CidQueue {
         orig_offset..sequence
     }
 
-    /// Returns a new CID if any were available and a possibly-empty range preceding it to retire
-    pub fn next(&mut self) -> Option<(IssuedCid, Range<u64>)> {
-        for i in 0..Self::LEN {
+    /// Switch to next active CID if possible, return
+    /// 1) the corresponding ResetToken and 2) a possibly-empty range preceding it to retire
+    pub fn next(&mut self) -> Option<(ResetToken, Range<u64>)> {
+        for i in 1..Self::LEN {
             let index = (self.cursor + i) % Self::LEN;
-            let (id, reset_token) = match self.buffer[index].take() {
+            let reset_token = match self.buffer[index].as_ref() {
                 None => continue,
-                Some(x) => x,
+                Some((_, token)) => *token,
             };
+            self.buffer[self.cursor] = None;
+
             let orig_offset = self.offset;
-            self.offset += i as u64 + 1;
-            self.cursor = (self.cursor + i + 1) % Self::LEN;
+            self.offset += i as u64;
+            self.cursor = (self.cursor + i) % Self::LEN;
             let sequence = orig_offset + i as u64;
-            let cid = IssuedCid {
-                sequence,
-                id,
-                reset_token,
-            };
-            return Some((cid, orig_offset..sequence));
+            return Some((reset_token.unwrap(), orig_offset..sequence));
         }
         None
     }
 
-    /// Reset offset field and guarantee the smallest sequence of active CID is prior to this value
-    pub fn assign_offset(&mut self, v: u64) {
-        self.offset = v;
+    pub fn update_cid(&mut self, cid: ConnectionId) {
+        debug_assert_eq!(self.offset, 0);
+        self.buffer[self.cursor] = Some((cid, None));
     }
 
-    /// Lowest queued sequence number
-    pub fn offset(&self) -> u64 {
+    /// Return active remote CID itself
+    pub fn rem_cid(&self) -> ConnectionId {
+        self.buffer[self.cursor].unwrap().0
+    }
+
+    /// Check whether self.offset points to a valid CID in CidQueue
+    pub fn validate_rem_cid(&mut self) -> bool {
+        self.buffer[self.cursor].as_ref().is_some()
+    }
+
+    /// Return the sequence number of active remote CID
+    pub fn rem_cid_seq(&self) -> u64 {
         self.offset
     }
 }
@@ -116,63 +125,67 @@ mod tests {
         }
     }
 
-    #[test]
-    fn next_dense() {
-        let mut q = CidQueue::new(0);
-        assert!(q.next().is_none());
-        assert!(q.next().is_none());
-
-        for i in 0..CidQueue::LEN as u64 {
-            q.insert(cid(i)).unwrap();
-        }
-        for i in 0..CidQueue::LEN as u64 {
-            let (cid, retire) = q.next().unwrap();
-            assert_eq!(cid.sequence, i);
-            assert_eq!(retire.end - retire.start, 0);
-        }
-        assert!(q.next().is_none());
+    fn initial_cid() -> ConnectionId {
+        ConnectionId::new(&[0xFF; 8])
     }
 
     #[test]
+    fn next_dense() {
+        let mut q = CidQueue::new(initial_cid());
+        assert!(q.next().is_none());
+        assert!(q.next().is_none());
+
+        for i in 1..CidQueue::LEN as u64 {
+            q.insert(cid(i)).unwrap();
+        }
+        for i in 1..CidQueue::LEN as u64 {
+            let (_, retire) = q.next().unwrap();
+            assert_eq!(q.rem_cid_seq(), i);
+            assert_eq!(retire.end - retire.start, 1);
+        }
+        assert!(q.next().is_none());
+    }
+    #[test]
     fn next_sparse() {
-        let mut q = CidQueue::new(0);
-        let seqs = (0..CidQueue::LEN as u64).filter(|x| x % 2 == 0);
+        let mut q = CidQueue::new(initial_cid());
+        let seqs = (1..CidQueue::LEN as u64).filter(|x| x % 2 == 0);
         for i in seqs.clone() {
             q.insert(cid(i)).unwrap();
         }
         for i in seqs {
-            let (cid, retire) = q.next().unwrap();
+            let (_, retire) = q.next().unwrap();
             dbg!(&retire);
-            assert_eq!(cid.sequence, i);
-            assert_eq!(retire, (cid.sequence.saturating_sub(1))..cid.sequence);
+            assert_eq!(q.rem_cid_seq(), i);
+            assert_eq!(retire, (q.rem_cid_seq().saturating_sub(2))..q.rem_cid_seq());
         }
         assert!(q.next().is_none());
     }
 
     #[test]
     fn wrap() {
-        let mut q = CidQueue::new(0);
+        let mut q = CidQueue::new(initial_cid());
 
-        for i in 0..CidQueue::LEN as u64 {
+        for i in 1..CidQueue::LEN as u64 {
             q.insert(cid(i)).unwrap();
         }
-        for _ in 0..3 {
+        for _ in 1..(CidQueue::LEN as u64 - 1) {
             q.next().unwrap();
         }
         for i in CidQueue::LEN as u64..(CidQueue::LEN as u64 + 3) {
             q.insert(cid(i)).unwrap();
         }
-        for i in 3..(CidQueue::LEN as u64 + 3) {
-            assert_eq!(q.next().unwrap().0.sequence, i);
+        for i in (CidQueue::LEN as u64 - 1)..(CidQueue::LEN as u64 + 3) {
+            q.next().unwrap();
+            assert_eq!(q.rem_cid_seq(), i);
         }
         assert!(q.next().is_none());
     }
 
     #[test]
     fn retire() {
-        let mut q = CidQueue::new(0);
+        let mut q = CidQueue::new(initial_cid());
 
-        for i in 0..CidQueue::LEN as u64 {
+        for i in 1..CidQueue::LEN as u64 {
             q.insert(cid(i)).unwrap();
         }
 
@@ -180,17 +193,12 @@ mod tests {
         let r = q.retire_prior_to(4);
         assert_eq!(r.end - r.start, 0);
 
-        for i in 4..CidQueue::LEN as u64 {
-            let (cid, retire) = q.next().unwrap();
-            assert_eq!(cid.sequence, i);
-            assert_eq!(retire.end - retire.start, 0);
-        }
         assert!(q.next().is_none());
     }
 
     #[test]
     fn insert_limit() {
-        let mut q = CidQueue::new(0);
+        let mut q = CidQueue::new(initial_cid());
         assert_eq!(q.insert(cid(CidQueue::LEN as u64 - 1)), Ok(()));
         assert_eq!(
             q.insert(cid(CidQueue::LEN as u64)),
@@ -200,16 +208,14 @@ mod tests {
 
     #[test]
     fn insert_duplicate() {
-        let mut q = CidQueue::new(0);
+        let mut q = CidQueue::new(initial_cid());
         q.insert(cid(0)).unwrap();
         q.insert(cid(0)).unwrap();
     }
 
     #[test]
     fn insert_retired() {
-        let mut q = CidQueue::new(0);
-        q.insert(cid(0)).unwrap();
-        q.next().unwrap();
+        let mut q = CidQueue::new(initial_cid());
         assert_eq!(q.insert(cid(0)), Ok(()), "reinserting active CID succeeds");
         assert!(q.next().is_none(), "active CID isn't requeued");
         q.insert(cid(1)).unwrap();
@@ -223,8 +229,8 @@ mod tests {
 
     #[test]
     fn retire_then_insert_next() {
-        let mut q = CidQueue::new(0);
-        for i in 0..CidQueue::LEN as u64 {
+        let mut q = CidQueue::new(initial_cid());
+        for i in 1..CidQueue::LEN as u64 {
             q.insert(cid(i)).unwrap();
         }
         q.next().unwrap();
@@ -233,5 +239,13 @@ mod tests {
             q.insert(cid(CidQueue::LEN as u64 + 1)),
             Err(InsertError::ExceedsLimit)
         );
+    }
+
+    #[test]
+    fn always_valid() {
+        let mut q = CidQueue::new(initial_cid());
+        assert!(q.next().is_none());
+        assert_eq!(q.rem_cid(), initial_cid());
+        assert_eq!(q.rem_cid_seq(), 0);
     }
 }
--- a/quinn-proto/src/cid_queue.rs
+++ b/quinn-proto/src/cid_queue.rs
@@ -18,16 +18,15 @@ pub struct CidQueue {
 }
 
 impl CidQueue {
-    pub const LEN: usize = 5;
 
     pub fn new(cid: ConnectionId) -> Self {
-        let mut this = Self {
-            buffer: [None; Self::LEN],
+        let mut buffer = [None; Self::LEN];
+        buffer[0] = Some((cid, None));
+        Self {
+            buffer,
             cursor: 0,
             offset: 0,
-        };
-        this.buffer[this.cursor] = Some((cid, None));
-        this
+        }
     }
 
     pub fn insert(&mut self, cid: IssuedCid) -> Result<(), InsertError> {
@@ -103,6 +102,8 @@ impl CidQueue {
     pub fn active_seq(&self) -> u64 {
         self.offset
     }
+
+    pub const LEN: usize = 5;
 }
 
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
@@ -140,7 +141,7 @@ mod tests {
         }
         for i in 1..CidQueue::LEN as u64 {
             let (_, retire) = q.next().unwrap();
-            assert_eq!(q.rem_cid_seq(), i);
+            assert_eq!(q.active_seq(), i);
             assert_eq!(retire.end - retire.start, 1);
         }
         assert!(q.next().is_none());
@@ -155,8 +156,8 @@ mod tests {
         for i in seqs {
             let (_, retire) = q.next().unwrap();
             dbg!(&retire);
-            assert_eq!(q.rem_cid_seq(), i);
-            assert_eq!(retire, (q.rem_cid_seq().saturating_sub(2))..q.rem_cid_seq());
+            assert_eq!(q.active_seq(), i);
+            assert_eq!(retire, (q.active_seq().saturating_sub(2))..q.active_seq());
         }
         assert!(q.next().is_none());
     }
@@ -176,7 +177,7 @@ mod tests {
         }
         for i in (CidQueue::LEN as u64 - 1)..(CidQueue::LEN as u64 + 3) {
             q.next().unwrap();
-            assert_eq!(q.rem_cid_seq(), i);
+            assert_eq!(q.active_seq(), i);
         }
         assert!(q.next().is_none());
     }
@@ -245,7 +246,7 @@ mod tests {
     fn always_valid() {
         let mut q = CidQueue::new(initial_cid());
         assert!(q.next().is_none());
-        assert_eq!(q.rem_cid(), initial_cid());
-        assert_eq!(q.rem_cid_seq(), 0);
+        assert_eq!(q.active(), initial_cid());
+        assert_eq!(q.active_seq(), 0);
     }
 }
--- a/quinn-proto/src/cid_queue.rs
+++ b/quinn-proto/src/cid_queue.rs
@@ -90,8 +90,8 @@ impl CidQueue {
         self.offset = v;
     }
 
-    /// Read the current offset value
-    pub fn read_offset(&self) -> u64 {
+    /// Lowest queued sequence number
+    pub fn offset(&self) -> u64 {
         self.offset
     }
 }
--- a/quinn-proto/src/cid_queue.rs
+++ b/quinn-proto/src/cid_queue.rs
@@ -196,11 +196,20 @@ mod tests {
             q.insert(cid(i)).unwrap();
         }
 
-        assert_eq!(q.retire_prior_to(4), 0..4);
-        let r = q.retire_prior_to(4);
+        assert_eq!(q.retire_prior_to(2), 0..2);
+        let r = q.retire_prior_to(2);
         assert_eq!(r.end - r.start, 0);
 
+        for i in 2..(CidQueue::LEN as u64 - 1) {
+            let _ = q.next().unwrap();
+            assert_eq!(q.active_seq(), i + 1);
+            let retire = q.retire_prior_to(i + 1);
+            assert_eq!(retire.end - retire.start, 0);
+            assert!(!q.is_active_retired());
+        }
+
         assert!(q.next().is_none());
+        assert!(!q.is_active_retired());
     }
 
     #[test]
--- a/quinn-proto/src/cid_queue.rs
+++ b/quinn-proto/src/cid_queue.rs
@@ -13,7 +13,7 @@ pub struct CidQueue {
     cursor: usize,
     /// Sequence number of `self.buffer[cursor]`
     ///
-    /// The sequence of active (in use) CID. This sequence is also the smallest among CIDs in `buffer`.
+    /// The sequence number of the active CID; must be the smallest among CIDs in `buffer`.
     offset: u64,
 }
 
@@ -90,17 +90,17 @@ impl CidQueue {
     }
 
     /// Return active remote CID itself
-    pub fn rem_cid(&self) -> ConnectionId {
+    pub fn active(&self) -> ConnectionId {
         self.buffer[self.cursor].unwrap().0
     }
 
     /// Check whether self.offset points to a valid CID in CidQueue
-    pub fn validate_rem_cid(&mut self) -> bool {
-        self.buffer[self.cursor].as_ref().is_some()
+    pub fn is_active_retired(&mut self) -> bool {
+        self.buffer[self.cursor].is_none()
     }
 
     /// Return the sequence number of active remote CID
-    pub fn rem_cid_seq(&self) -> u64 {
+    pub fn active_seq(&self) -> u64 {
         self.offset
     }
 }
--- a/quinn-proto/src/cid_queue.rs
+++ b/quinn-proto/src/cid_queue.rs
@@ -84,6 +84,16 @@ impl CidQueue {
         }
         None
     }
+
+    /// Reset offset field and guarantee the smallest sequence of active CID is prior to this value
+    pub fn assign_offset(&mut self, v: u64) {
+        self.offset = v;
+    }
+
+    /// Read the current offset value
+    pub fn read_offset(&self) -> u64 {
+        self.offset
+    }
 }
 
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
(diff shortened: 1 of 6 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -2351,13 +2360,24 @@ where
...
+                        .active_connection_id_limit
+                        .0
+                        .min(LOC_CID_COUNT)
+                        > self.cids_active_seq.len() as u64;
                     self.endpoint_events
-                        .push_back(EndpointEventInner::RetireConnectionId(sequence));
+                        .push_back(EndpointEventInner::RetireConnectionId(
+                            sequence,
+                            allow_more_cid,
+                        ));
```
liwenjieQu, line 2374: Consider a scenario where peer A has active remote cid 0,1,2. Peer B first send a NEW_CONNECTION_ID with cid 3 and retire_prior_to set to 0. Peer A processes this NEW_CONNECTION_ID frame; update remote cid to 1,2,3 and meanwhile send a RETIRE_CONNECTION_ID from to peer B. If peer B doesn't check the cid limit here and send a new cid again, peer A will then face CONNECTION_ID_LIMIT_ERROR

quinn-proto/src/tests/mod.rs
```diff
@@ -1177,6 +1177,39 @@ fn keep_alive() {
...
+    let mut pair = Pair::default();
+    let (client_ch, server_ch) = pair.connect();
+
+    // Server retires current active remote CIDs
+    pair.server_conn_mut(server_ch).mock_retire_cid(1, 1);
+    pair.drive();
+    // Any unexpected behavior may trigger TransportError::CONNECTION_ID_LIMIT_ERROR
+    assert!(!pair.client_conn_mut(client_ch).is_closed());
+    assert!(!pair.server_conn_mut(server_ch).is_closed());
+    assert_matches!(pair.client_conn_mut(client_ch).active_rem_cid_seq(), 1);
```
liwenjieQu, line 1192: Client holds active remote cids 0,1,2,3,4. When server actively retire cid 0, current version will respond with CONNECTION_ID_LIMIT_ERROR which is not true. The reason is because current code run `update_rem_cid ` after `insert`.

quinn-proto/src/tests/mod.rs
```diff
@@ -1177,6 +1177,39 @@ fn keep_alive() {
...
+    // Server retires all valid remote CIDs
+    pair.server_conn_mut(server_ch)
+        .mock_retire_cid(next_retire_prior_to, 1);
+    pair.drive();
+    assert!(!pair.client_conn_mut(client_ch).is_closed());
+    assert!(!pair.server_conn_mut(server_ch).is_closed());
+    assert_matches!(
+        pair.client_conn_mut(client_ch).active_rem_cid_seq(),
+        next_retire_prior_to
+    );
```
liwenjieQu, line 1210: if the bug in RETIRE_CONNECTION_ID is not fixed, the cid rotation can continue endlessly.

quinn-proto/src/connection/mod.rs
djc: Probably good to have this comment in the code!

liwenjieQu: Done

quinn-proto/src/tests/mod.rs
liwenjieQu, line 1192: In this test case, 
1. Client initially holds active remote cids 0,1,2,3,4.
2. Server actively retire cid 0 and send a NEW_CONNECTION_ID frame to client.
3. When client processes this frame [here](https://github.com/quinn-rs/quinn/pull/871/files#diff-423ed266e9aaf5fcae0fa8b85ca382f495528c18cd7c890cef1cf438221f2ef6R2398), the offset is not changed (remains to be 1). This is current since we store 1,2,3,4 in CidQueue while we actually need to retire 0. However, later on when the new cid is inserted [here](https://github.com/quinn-rs/quinn/pull/871/files#diff-423ed266e9aaf5fcae0fa8b85ca382f495528c18cd7c890cef1cf438221f2ef6R2431), since offset is still 1, the index is calculated as 4 [here](https://github.com/quinn-rs/quinn/pull/871/files#diff-e07d31049906f88a397dc54fc600e122da5f719a4a91599e5dd513d321d2077fR41) which triggers ExceedsLimit error. In fact, client simply wants to change cid from 0,1,2,3,4 to 1,2,3,4,5 and there is no violation

liwenjieQu: If i checkout commit `55168f3` and run this unit test. Here is the output
```
Oct 16 10:39:06.204 TRACE client:send{space=Data pn=2}: quinn_proto::connection: NEW_CONNECTION_ID sequence=5 id=a97e4574bfe160fd
Oct 16 10:39:06.204 TRACE client: quinn_proto::connection: sending 62 byte datagram
Oct 16 10:39:06.204 TRACE server: quinn_proto::connection: got Data packet (62 bytes) from [::1]:44433 using id 4ba0ad1102d392b7
Oct 16 10:39:06.204 TRACE server:recv{space=Data pn=2}: quinn_proto::connection: authenticated
Oct 16 10:39:06.204 TRACE server:recv{space=Data pn=2}:frame{ty=NEW_CONNECTION_ID}: quinn_proto::connection: sequence=5 id=a97e4574bfe160fd
Oct 16 10:39:06.204 TRACE server:send{space=Data pn=2}: quinn_proto::connection: ACK
Oct 16 10:39:06.204 TRACE server: quinn_proto::connection: sending 34 byte datagram
Oct 16 10:39:06.205 TRACE client: quinn_proto::connection: got Data packet (34 bytes) from [::1]:4433 using id a3710748a204985b
Oct 16 10:39:06.205 TRACE client:recv{space=Data pn=2}: quinn_proto::connection: authenticated
Oct 16 10:39:06.205 TRACE server:send{space=Data pn=3}: quinn_proto::connection: ACK
Oct 16 10:39:06.205 TRACE server:send{space=Data pn=3}: quinn_proto::connection: NEW_CONNECTION_ID sequence=5 id=c17e354033199672
Oct 16 10:39:06.205 TRACE server: quinn_proto::connection: sending 62 byte datagram
Oct 16 10:39:06.206 TRACE client: quinn_proto::connection: got Data packet (62 bytes) from [::1]:4433 using id a3710748a204985b
Oct 16 10:39:06.206 TRACE client:recv{space=Data pn=3}: quinn_proto::connection: authenticated
Oct 16 10:39:06.206 TRACE client:recv{space=Data pn=3}:frame{ty=NEW_CONNECTION_ID}: quinn_proto::connection: sequence=5 id=c17e354033199672
Oct 16 10:39:06.206 DEBUG client: quinn_proto::connection: closing connection due to transport error: the number of connection IDs provided by the peer exceeds the advertised active_connection_id_limit
Oct 16 10:39:06.206 TRACE client: quinn_proto::connection: connection closed
Oct 16 10:39:06.206 TRACE client:send{space=Data pn=3}: quinn_proto::connection: sending CONNECTION_CLOSE
Oct 16 10:39:06.206 TRACE client: quinn_proto::connection: sending 30 byte datagram
Oct 16 10:39:06.206 TRACE server: quinn_proto::connection: got Data packet (30 bytes) from [::1]:44433 using id 4ba0ad1102d392b7
Oct 16 10:39:06.207 TRACE server:recv{space=Data pn=3}: quinn_proto::connection: authenticated
Oct 16 10:39:06.208 TRACE server: quinn_proto::connection: connection closed
Oct 16 10:39:06.208 TRACE server:send{space=Data pn=4}: quinn_proto::connection: sending CONNECTION_CLOSE
Oct 16 10:39:06.208 TRACE server: quinn_proto::connection: sending 30 byte datagram
Oct 16 10:39:06.208 TRACE client: quinn_proto::connection: got Data packet (30 bytes) from [::1]:4433 using id a3710748a204985b
Oct 16 10:39:06.209 TRACE client:recv{space=Data pn=4}: quinn_proto::connection: draining
Oct 16 10:39:06.209 TRACE quinn_proto::tests::util: advancing to Instant { t: 305032168058444 } for client
Oct 16 10:39:06.209 TRACE client: quinn_proto::connection: timeout timer=Close
Oct 16 10:39:06.209 TRACE server: quinn_proto::connection: timeout timer=Close
```

liwenjieQu: To fix this issue, [this code](https://github.com/quinn-rs/quinn/pull/871/files#diff-423ed266e9aaf5fcae0fa8b85ca382f495528c18cd7c890cef1cf438221f2ef6R2404-R2412) has to move before CidQueue insert to allow we get any replacement of cid from CidQueue and reset the offset value.

if retire_prior_to field invalidates everything (including both rem_cid and rem_cids), the execution flow will just to [this](https://github.com/quinn-rs/quinn/pull/871/files#diff-423ed266e9aaf5fcae0fa8b85ca382f495528c18cd7c890cef1cf438221f2ef6R2446-R2455) where we assign cid in current frame to `rem_cid`

[This](https://github.com/quinn-rs/quinn/pull/871/files#diff-423ed266e9aaf5fcae0fa8b85ca382f495528c18cd7c890cef1cf438221f2ef6R2446-R2455) handles the corner case of scenario where retire_prior_to invalidates everything. 
Consider the following scenario: 
1. Peer A has 0,1,2,3,4 as valid cids. 
2. Peer B wants to provide 2 new cids: cid 5 and 6 and invalidates everything it previously issued. To do that, peer B set retire_prior_to to 5 and send these two NEW_CONNECTION_ID frame.
3. As we cannot control on the other end which frame is processed first, it is possible peer A first deals with new cid 6 and then 5. As a result, [This](https://github.com/quinn-rs/quinn/pull/871/files#diff-423ed266e9aaf5fcae0fa8b85ca382f495528c18cd7c890cef1cf438221f2ef6R2446-R2455) is needed to swap cid and make sure `rem_cid` always has the smallest sequence number.

quinn-proto/src/tests/mod.rs
```diff
@@ -1177,6 +1177,39 @@ fn keep_alive() {
...
+    // Server retires all valid remote CIDs
+    pair.server_conn_mut(server_ch)
+        .mock_retire_cid(_next_retire_prior_to, 1);
+    pair.drive();
+    assert!(!pair.client_conn_mut(client_ch).is_closed());
+    assert!(!pair.server_conn_mut(server_ch).is_closed());
+    assert_matches!(
+        pair.client_conn_mut(client_ch).active_rem_cid_seq(),
+        _next_retire_prior_to
+    );
```
liwenjieQu, line 1210: When i checkout commit () (fix bug in NEW_CONNECTION_ID but not fixing in RETIRE_CONNECTION_ID [here](https://github.com/quinn-rs/quinn/pull/871/files#diff-423ed266e9aaf5fcae0fa8b85ca382f495528c18cd7c890cef1cf438221f2ef6R2363-R2379)), I got the following output

```
Oct 16 10:56:19.798 TRACE server:send{space=Data pn=3}: quinn_proto::connection: NEW_CONNECTION_ID sequence=5 id=eaa580934a8f5476
Oct 16 10:56:19.799 TRACE server: quinn_proto::connection: sending 62 byte datagram
Oct 16 10:56:19.799 TRACE client: quinn_proto::connection: got Data packet (62 bytes) from [::1]:4433 using id f08a8aa4490cf931
Oct 16 10:56:19.799 TRACE client:recv{space=Data pn=3}: quinn_proto::connection: authenticated
Oct 16 10:56:19.799 TRACE client:recv{space=Data pn=3}:frame{ty=NEW_CONNECTION_ID}: quinn_proto::connection: sequence=5 id=eaa580934a8f5476 retire_prior_to=1
Oct 16 10:56:19.799 TRACE client:recv{space=Data pn=3}:frame{ty=NEW_CONNECTION_ID}: quinn_proto::connection: switching to remote CID 1: 2af6dbf3da6f1e1f
Oct 16 10:56:19.799 TRACE client:send{space=Data pn=3}: quinn_proto::connection: ACK
Oct 16 10:56:19.799 TRACE client:send{space=Data pn=3}: quinn_proto::connection: RETIRE_CONNECTION_ID sequence=0
Oct 16 10:56:19.800 TRACE client: quinn_proto::connection: sending 36 byte datagram
Oct 16 10:56:19.800 TRACE server: quinn_proto::connection: got Data packet (36 bytes) from [::1]:44433 using id 2af6dbf3da6f1e1f
Oct 16 10:56:19.801 TRACE server:recv{space=Data pn=3}: quinn_proto::connection: authenticated
Oct 16 10:56:19.801 TRACE server:send{space=Data pn=4}: quinn_proto::connection: ACK
Oct 16 10:56:19.801 TRACE server: quinn_proto::connection: sending 34 byte datagram
Oct 16 10:56:19.801 TRACE server: quinn_proto::endpoint: peer retired CID 0: 782fdbd230ae62b1
Oct 16 10:56:19.801 TRACE client: quinn_proto::connection: got Data packet (34 bytes) from [::1]:4433 using id f08a8aa4490cf931
Oct 16 10:56:19.801 TRACE client:recv{space=Data pn=4}: quinn_proto::connection: authenticated
Oct 16 10:56:19.802 TRACE server:send{space=Data pn=5}: quinn_proto::connection: ACK
Oct 16 10:56:19.802 TRACE server:send{space=Data pn=5}: quinn_proto::connection: NEW_CONNECTION_ID sequence=6 id=67d47302296af768
Oct 16 10:56:19.803 TRACE server: quinn_proto::connection: sending 62 byte datagram
Oct 16 10:56:19.803 TRACE client: quinn_proto::connection: got Data packet (62 bytes) from [::1]:4433 using id f08a8aa4490cf931
Oct 16 10:56:19.803 TRACE client:recv{space=Data pn=5}: quinn_proto::connection: authenticated
Oct 16 10:56:19.804 TRACE client:recv{space=Data pn=5}:frame{ty=NEW_CONNECTION_ID}: quinn_proto::connection: sequence=6 id=67d47302296af768 retire_prior_to=1
Oct 16 10:56:19.804 DEBUG client: quinn_proto::connection: closing connection due to transport error: the number of connection IDs provided by the peer exceeds the advertised active_connection_id_limit
Oct 16 10:56:19.805 TRACE client: quinn_proto::connection: connection closed
Oct 16 10:56:19.805 TRACE client:send{space=Data pn=4}: quinn_proto::connection: sending CONNECTION_CLOSE
Oct 16 10:56:19.805 TRACE client: quinn_proto::connection: sending 30 byte datagram
Oct 16 10:56:19.806 TRACE server: quinn_proto::connection: got Data packet (30 bytes) from [::1]:44433 using id 2af6dbf3da6f1e1f
Oct 16 10:56:19.806 TRACE server:recv{space=Data pn=4}: quinn_proto::connection: authenticated
Oct 16 10:56:19.806 TRACE server: quinn_proto::connection: connection closed
Oct 16 10:56:19.806 TRACE server:send{space=Data pn=6}: quinn_proto::connection: sending CONNECTION_CLOSE
Oct 16 10:56:19.807 TRACE server: quinn_proto::connection: sending 30 byte datagram
Oct 16 10:56:19.807 TRACE client: quinn_proto::connection: got Data packet (30 bytes) from [::1]:4433 using id f08a8aa4490cf931
Oct 16 10:56:19.807 TRACE client:recv{space=Data pn=6}: quinn_proto::connection: draining
Oct 16 10:56:19.808 TRACE quinn_proto::tests::util: advancing to Instant { t: 306065761809108 } for client
Oct 16 10:56:19.808 TRACE client: quinn_proto::connection: timeout timer=Close
Oct 16 10:56:19.808 TRACE server: quinn_proto::connection: timeout timer=Close
thread 'tests::cid_retirement' panicked at 'assertion failed: !pair.client_conn_mut(client_ch).is_closed()', quinn-proto/src/tests/mod.rs:1190:5
```

You can see server continue to send
`Oct 16 10:56:19.802 TRACE server:send{space=Data pn=5}: quinn_proto::connection: NEW_CONNECTION_ID sequence=6 id=67d47302296af768
` to client and then triggers the limitexceed error on client side - matches with what i described previously.

Ralith (changes requested):
Great catches on those bugs, thanks!

quinn-proto/src/connection/mod.rs
```diff
@@ -2557,7 +2617,9 @@ where
             - 1;
         self.endpoint_events
             .push_back(EndpointEventInner::NeedIdentifiers(n));
-        self.cids_issued += n;
+        // Only add one CID we supplied while handshaking to self.cids_issued. Rest will be added when we handle NewIdentifiers event
+        self.cids_issued += 1;
+        self.cids_active_seq.insert(0);
```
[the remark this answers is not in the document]
liwenjieQu: Agree.

quinn-proto/src/connection/mod.rs
```diff
@@ -2943,6 +3005,20 @@ where
         self.path.sending_ecn
     }
 
+    /// Mock updating retire_prior_to field in NEW_CONNECTION_ID frame
+    #[cfg(test)]
+    pub(crate) fn mock_retire_cid(&mut self, v: u64, n: u64) {
```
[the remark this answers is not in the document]
liwenjieQu: I will change to `rotate_rem_cids` as it is `remote` cid that this function actually changes. Otherwise i am OK with the change. Will also add the assert there.

quinn-proto/src/shared.rs
```diff
@@ -52,7 +52,7 @@ pub(crate) enum EndpointEventInner {
     /// The connection needs connection identifiers
     NeedIdentifiers(u64),
     /// Stop routing connection ID for this sequence number to the connection
-    RetireConnectionId(u64),
+    RetireConnectionId(u64, bool),
```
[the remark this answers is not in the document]
liwenjieQu: Will do

quinn-proto/src/connection/mod.rs
```diff
@@ -2351,13 +2360,29 @@ where
...
+                    // Consider a scenario where peer A has active remote cid 0,1,2.
+                    // Peer B first send a NEW_CONNECTION_ID with cid 3 and retire_prior_to set to 1.
+                    // Peer A processes this NEW_CONNECTION_ID frame; update remote cid to 1,2,3
+                    // and meanwhile send a RETIRE_CONNECTION_ID to retire cid 0 to peer B.
+                    // If peer B doesn't check the cid limit here and send a new cid again, peer A will then face CONNECTION_ID_LIMIT_ERROR
+                    let allow_more_cid = self
+                        .peer_params
+                        .active_connection_id_limit
+                        .0
+                        .min(LOC_CID_COUNT)
```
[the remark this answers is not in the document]
liwenjieQu: good call

quinn-proto/src/connection/mod.rs
```diff
@@ -2376,38 +2401,67 @@ where
...
+                    };
+
+                    // Guarantee active remote cid (self.rem_cid) always has the smallest sequence number among valid ones in CidQueue
+                    if valid_rem_cid && self.rem_cids.read_offset() == frame.sequence + 1 {
+                        let swap_cid = new_rem_cid;
+                        new_rem_cid.sequence = self.rem_cid_seq;
+                        new_rem_cid.id = self.rem_cid;
+                        new_rem_cid.reset_token = self.peer_params.stateless_reset_token.unwrap();
+                        self.apply_new_rem_cid(swap_cid);
+                    }
```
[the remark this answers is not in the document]
liwenjieQu: > If we receive NEW_CONNECTION_ID frames out of order, this could lead to an already-used CID being returned to the CID queue. If a migration then occurs and we switch to the queued CID, then that CID will have been used on both sides of the migration, allowing an observed to link the two flows. This weakens privacy.

Good catch. I wasn't thinking about the security requirement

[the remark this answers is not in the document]
liwenjieQu: Agree

quinn-proto/src/connection/mod.rs
```diff
@@ -2376,38 +2401,67 @@ where
...
+                                    .pending
+                                    .retire_cids
+                                    .push(frame.sequence);
+                                continue;
+                            }
                         }
+                    } else {
+                        // There is no valid CIDs in CidQueue and active CID self.rem_cid_seq is before retire_prior_to. We must
+                        // 1. Reset offset in CidQueue
+                        self.rem_cids.assign_offset(frame.retire_prior_to + 1);
```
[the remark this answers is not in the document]
liwenjieQu: Will try to refactor CidQueue first and then we can check whether this block of code is still needed. Will update later.

quinn-proto/src/connection/mod.rs
```diff
@@ -2376,38 +2401,67 @@ where
...
+                        // 2. Retire active remote cid
+                        self.spaces[SpaceId::Data]
+                            .pending
+                            .retire_cids
+                            .push(self.rem_cid_seq);
+                        // 3. Accept current cid in the frame as the active one
+                        self.apply_new_rem_cid(new_rem_cid);
                     }
```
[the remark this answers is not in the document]
liwenjieQu: Right. I wasn't 100% confident with the `CidQueue` logic when i made the original change so I chose a workaround. Will  try to refactor `CidQueue` to include tracking active remote CID

quinn-proto/src/connection/mod.rs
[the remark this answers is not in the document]
Ralith: Whoops, looks like github didn't lose the first draft if this comment after all, sorry for the duplicate :D

quinn-proto/src/connection/mod.rs
```diff
@@ -2943,6 +2994,22 @@ where
...
 
+    /// Instruct the peer to retire the next `n` connection IDs, and issue replacements
+    #[cfg(test)]
+    pub(crate) fn rotate_rem_cid(&mut self, v: u64, n: u64) {
+        // Cannot retire more CIDs than what have been issued
+        debug_assert!(v <= *self.cids_active_seq.iter().max().unwrap() + 1);
+        self.retire_cid_seq = v;
+        self.endpoint_events
+            .push_back(EndpointEventInner::NeedIdentifiers(n));
+    }
```
Ralith, line 3005: Do we actually need to have two separate arguments to this, or would a single quantity of CIDs to both retire and issue be sufficient?

quinn-proto/src/connection/mod.rs
```diff
@@ -2943,6 +2994,22 @@ where
         self.path.sending_ecn
     }
 
+    /// Instruct the peer to retire the next `n` connection IDs, and issue replacements
+    #[cfg(test)]
+    pub(crate) fn rotate_rem_cid(&mut self, v: u64, n: u64) {
```
Ralith, line 2999:
```suggestion
    pub(crate) fn rotate_local_cids(&mut self, v: u64, n: u64) {
```
This is instructing the peer to change the set of CIDs it may use to address the local endpoint. We use `rem_cid` to refer CIDs issued by and used to address the peer.

Ralith (changes requested):
Further review pending the discussed CidQueue refactoring.

quinn-proto/src/connection/mod.rs
liwenjieQu: Since it is a test function, i don't have a strong preference here. The intention was to issue less new CIDs than retired ones, which makes it easy to read the log. Anyway, i am happy to update it.

quinn-proto/src/connection/mod.rs
liwenjieQu: Ack. Will update

quinn-proto/src/connection/mod.rs
```diff
@@ -2401,12 +2413,12 @@ where
...
                         // We're a server using the initial remote CID for the client, so let's
                         // switch immediately to enable clientside stateless resets.
-                        debug_assert_eq!(self.rem_cid_seq, 0);
+                        debug_assert_eq!(self.rem_cids.rem_cid_seq(), 0);
                         self.update_rem_cid().unwrap();
-                    } else if frame.retire_prior_to > self.rem_cid_seq {
-                        // If our current CID is earlier than the first unretired one we must not
-                        // have adopted the one we just got, so it must be stored in rem_cids, so
-                        // this unwrap is guaranteed to succeed.
+                    } else if update_reset_token && !self.rem_cids.validate_rem_cid() {
```
Ralith, line 2418: Would this be sufficient? I can't see how we'd end up with a retired active remote CID otherwise.
```suggestion
                    } else if !self.rem_cids.validate_rem_cid() {
```

quinn-proto/src/transport_parameters.rs
```diff
@@ -175,6 +174,13 @@ impl TransportParameters {
         }
         Ok(())
     }
+
+    /// Calculate the maximum number of CIDs that a peer is willing to issue
```
Ralith, line 178:
```suggestion
    /// Maximum number of CIDs to issue to this peer
    ///
```

Ralith (changes requested):
Thanks, this looks great! Just a few minor issues left and we can merge.

quinn-proto/src/cid_queue.rs
```diff
@@ -63,27 +64,45 @@ impl CidQueue {
         orig_offset..sequence
     }
 
-    /// Returns a new CID if any were available and a possibly-empty range preceding it to retire
-    pub fn next(&mut self) -> Option<(IssuedCid, Range<u64>)> {
-        for i in 0..Self::LEN {
+    /// Switch to next active CID if possible, return
+    /// 1) the corresponding ResetToken and 2) a possibly-empty range preceding it to retire
+    pub fn next(&mut self) -> Option<(ResetToken, Range<u64>)> {
+        for i in 1..Self::LEN {
```
djc, line 70: I feel like this could be written more clearly like this:

```rust
let reset_token = (1..Self::LEN).into_iter().find_map(|i| {
    let index = (self.cursor + i) % Self::LEN;
    self.buffer[index].map(|(_, token)| *token)
})?;
```

quinn-proto/src/endpoint.rs
```diff
@@ -122,11 +122,13 @@ where
                     warn!("duplicate reset token");
                 }
             }
-            RetireConnectionId(seq) => {
+            RetireConnectionId(seq, allow_more_cid) => {
```
djc, line 125: Maybe call this `allow_more_cids`? Reads more natural to me.

djc (commented):
This stuff hurts my head... Would it make sense to add more comments to explain what's going on here? For example, all the 1-based indexing doesn't intuitively make sense to me.

quinn-proto/src/connection/mod.rs
```diff
@@ -2943,6 +2947,21 @@ where
         self.path.sending_ecn
     }
 
+    /// Mock updating retire_prior_to field in NEW_CONNECTION_ID frame
+    #[cfg(test)]
+    pub(crate) fn mock_retire_cid(&mut self, v: u64, n: u64) {
```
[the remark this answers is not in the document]
Ralith: This was addressed in a later commit.

quinn-proto/src/cid_queue.rs
```diff
@@ -106,81 +125,80 @@ mod tests {
...
 
         assert_eq!(q.retire_prior_to(4), 0..4);
         let r = q.retire_prior_to(4);
         assert_eq!(r.end - r.start, 0);
 
-        for i in 4..CidQueue::LEN as u64 {
-            let (cid, retire) = q.next().unwrap();
-            assert_eq!(cid.sequence, i);
-            assert_eq!(retire.end - retire.start, 0);
-        }
```
Ralith, line 177: Does this need to be deleted, or could we just tweak the numbers so it's not dead code and properly tests `next` in this case? I think the use of `4` here dates from when `LEN` was higher.

quinn-proto/src/endpoint.rs
Ralith: Or alternatively, `issue_new_cid`?

Ralith:
> all the 1-based indexing doesn't intuitively make sense to me.

We tend to want to iterate over the set of CIDs we haven't used yet. Calling that out in a comment seems like a good idea.

quinn-proto/src/tests/mod.rs
```diff
@@ -1177,6 +1177,39 @@ fn keep_alive() {
...
+    assert!(!pair.client_conn_mut(client_ch).is_closed());
+    assert!(!pair.server_conn_mut(server_ch).is_closed());
+    assert_matches!(pair.client_conn_mut(client_ch).active_rem_cid_seq(), 1);
+
+    use crate::cid_queue::CidQueue;
+    use crate::LOC_CID_COUNT;
+    let mut active_cid_num = CidQueue::LEN as u64;
+    active_cid_num = active_cid_num.min(LOC_CID_COUNT);
+
+    let _next_retire_prior_to = active_cid_num + 1;
```
Ralith, line 1199:
```suggestion
    let next_retire_prior_to = active_cid_num + 1;
```

Leading underscores should only be used for intentionally dead variables.

djc:
> We tend to want to iterate over the set of CIDs we haven't used yet. Calling that out in a comment seems like a good idea.

Might be even better to create an explicit `fn iter_next() -> impl Iterator<Item = usize>` (or even something that returns the values directly), so that it's (a) named, and the method can have a docstring, and (b) deduplicated from the different places where the same pattern is used.

quinn-proto/src/cid_queue.rs
```diff
@@ -8,29 +8,30 @@ use crate::{shared::IssuedCid, ConnectionId, ResetToken};
...
 }
 
 impl CidQueue {
-    pub const LEN: usize = 4;
+    pub const LEN: usize = 5;
 
-    pub fn new(offset: u64) -> Self {
-        Self {
+    pub fn new(cid: ConnectionId) -> Self {
+        let mut this = Self {
```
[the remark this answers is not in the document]
liwenjieQu: Ack

quinn-proto/src/cid_queue.rs
```diff
@@ -8,29 +8,30 @@ use crate::{shared::IssuedCid, ConnectionId, ResetToken};
...
     ///
-    /// The CID sequenced immediately prior to this is the active CID, which this data structure is
-    /// not responsible for retiring.
+    /// The sequence of active (in use) CID. This sequence is also the smallest among CIDs in `buffer`.
     offset: u64,
 }
 
 impl CidQueue {
-    pub const LEN: usize = 4;
+    pub const LEN: usize = 5;
```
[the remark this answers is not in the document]
liwenjieQu: Ack

quinn-proto/src/connection/mod.rs
liwenjieQu: Right. Checking `update_reset_token ` seems not necessary.

quinn-proto/src/connection/mod.rs
```diff
@@ -2525,9 +2486,7 @@ where
             );
             self.migrate(now, remote);
             // Break linkability, if possible
-            if let Some(cid) = self.retire_rem_cid() {
-                self.apply_new_rem_cid(cid);
-            }
+            self.update_rem_cid().unwrap();
```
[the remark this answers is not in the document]
liwenjieQu: Ack

quinn-proto/src/connection/mod.rs
```diff
@@ -2996,9 +2946,10 @@ where
 
     /// Instruct the peer to retire the next `n` connection IDs, and issue replacements
```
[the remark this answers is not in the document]
liwenjieQu: It is changed from a form with `n` as input parameter. I would rather stay with this format except there is strong reason for changing again. Fixing comment :-)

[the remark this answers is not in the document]
Ralith: No strong reason, that's fine. For context, I've been wondering if exposing something like this as a public API might make for a simpler alternative than the built-in rotation timer, but if we end up going that way we can leave addtional polish to future work.

djc:
(Also needs to rebased onto current main.)

quinn-proto/src/cid_queue.rs
liwenjieQu: Before this change, CidQueue::LEN was `4` and thus these lines were not executed as well (seems to be a legacy issue). When CidQueue::LEN is now increased to `5`, if I don't remove these code, the assert will fail. 

`next` is tested in next line `assert!(q.next().is_none());`. Before the test execute this line, we will have only 1 active cid in the queue (with sequence number 4). Since there is no "next" cid to switch, `next` returns none.

quinn-proto/src/cid_queue.rs
liwenjieQu: Combined with your other comments on `fn iter_next()`, I gave a bit thought on the API format and the result is not exactly same as you suggested here. See commit `720431a`

quinn-proto/src/cid_queue.rs
liwenjieQu: it is a bit confusing with my above comment. Overall what I want to say is `next` is already tested in next line in the unit test function and thus the deleted code block seems redundant.

Ralith: The original intention was to verify that, for `retire_prior_to(n)`, CIDs with sequence numbers greater than `n` were not retired. A previous change broke that by reducing `CidQueue::LEN` to 4. We should restore the original test coverage by e.g. replacing every occurrence of 4 in this test with 2 so that both cases are covered, and adjusting the loop to run the correct number of iterations.

Ralith (changes requested):

quinn-proto/src/cid_queue.rs
```diff
@@ -106,81 +132,80 @@ mod tests {
...
+        let mut q = CidQueue::new(initial_cid());
 
-        for i in 0..CidQueue::LEN as u64 {
+        for i in 1..CidQueue::LEN as u64 {
             q.insert(cid(i)).unwrap();
         }
 
         assert_eq!(q.retire_prior_to(4), 0..4);
         let r = q.retire_prior_to(4);
         assert_eq!(r.end - r.start, 0);
```
Ralith, line 201: Let's add an `assert!(!q.is_active_retired());` here as well.

quinn-proto/src/cid_queue.rs
liwenjieQu: Got it!

Ralith (approved):
LGTM; thanks for your great work diagnosing and resolving this issue!

djc:
+1 from me, thanks for all the hard work!

liwenjieQu:
Thx for all the reviews and comments!

Issue #856, opened 2020-09-29
New release


DemiMarie:
This blocks libp2p-quic.

est31:
:+1: a release with rustls 0.18 would be really great for mimas as well. I think at this point one needs to branch off master because of the export_keyring_material feature.

djc:
I've filed https://github.com/ctz/rustls/issues/407 about getting a rustls release.

dvc94ch:
I think that's done now, are there any other blocking issues?

djc:
Our interop crate depends on hyper-0.13, which relies on tokio-0.2; the hyper-rustls release that supports rustls-0.19 only supports tokio 0.3.

jamadazi:
So there will be no new release until quinn migrates to tokio 0.3?

So, currently, to use quinn in my project, i need to depend on tokio 0.2 and rustls 0.17, correct?

djc:
I think that summarizes the current status, yes. Unless you know you need something that's not part of the 0.6.1 release, I'd probably just use that.

jamadazi:
What is holding up the progress on this front? I looked at #872 and #873 but it is unclear to me what the status is (and seems like there hasn't been any update in over a month).

djc:
I haven't been able to prioritize #873. Additionally, there seems to be some doubt about whether there'll even be a hyper release targeting tokio 0.3, so I'm waiting that out for a little bit (see #hyper on the tokio Discord).

dvc94ch:
You actually don't need to depend on tokio, async-io would be enough (I can open a PR for this on thursday). It also works with all executors. For the hyper dependency in the interop tests, it should be fine to use hyper from git, as it's only for testing, so it doesn't really need to be published on crates.io

jamadazi:
Can I use tokio 0.3 with quinn 0.6.1, though? (by doing what you are describing)

If I want tokio, do I have to stick with 0.2?

dvc94ch:
You can use it with any executor, as opposed to only with Tokio 0.2

jamadazi:
So that includes tokio 0.3?

djc:
I think you're talking past each other.

quinn 0.6.1 does not support tokio 0.3, and never will.

@jamadazi if you're going to stick to quinn-proto, we could potentially cut a release for that separately, since it won't have to rely on tokio at all.

@dvc94ch I'm not convinced async-io is the right solution for us, and a change like that should probably be discussed at a design level first, in a separate issue.

[...]

PR #873, opened 2020-10-17
Tokio 1.0, rustls 0.19, etc

djc:
Left to do:

* [x] `mio::net::UdpSocket::only_v6()` (https://github.com/tokio-rs/mio/issues/1424)
* [x] Figure out test failures


Dessix:
Tokio is [at 1.0.1 now](https://crates.io/crates/tokio), but anyone interested in progress updates on this front should probably look toward #856 for updates.

djc:
Thanks for the cross-reference! Let's just close this, the next release will be on tokio 1.

djc:
Ah, never mind, thought this was another issue. I will resume working on this branch soon, but if anyone wants to help out, that would be great!

djc:
This now has upgrades to Tokio 1.0 (including bytes 1.0) and rustls 0.19 (et al). This mostly compiles except for one method that's gone missing on mio's UdpSocket (see description -- maybe we can get around the macOS compatibility issue there) and the tests fail because I botched some changes required for bytes 1.0.

quinn/src/tests.rs
```diff
@@ -120,7 +128,7 @@ fn read_after_close() {
             .unwrap()
             .await
             .expect("connect");
-        tokio::time::delay_until(Instant::now() + Duration::from_millis(100)).await;
+        tokio::time::sleep_until(Instant::now() + Duration::from_millis(100)).await;
```
Darksonn, line 131: This is equivalent to
```suggestion
        tokio::time::sleep(Duration::from_millis(100)).await;
```

djc:
quinn and quinn-proto now pass the tests, but I have a bunch of test failures in quinn-h3.

quinn/src/udp.rs
```diff
@@ -53,11 +54,11 @@ impl UdpSocket {
...
-        ready!(self.io.poll_read_ready(cx, mio::Ready::readable()))?;
+        let mut guard = ready!(self.io.poll_read_ready(cx))?;
         match self.io.get_ref().recv_ext(bufs, meta) {
             Ok(n) => Poll::Ready(Ok(n)),
             Err(ref e) if e.kind() == io::ErrorKind::WouldBlock => {
-                self.io.clear_read_ready(cx, mio::Ready::readable())?;
+                guard.clear_ready();
                 Poll::Pending
             }
             Err(e) => Poll::Ready(Err(e)),
```
Darksonn, line 64: I recommend that you rewrite this to use the [`try_io`][1] method from the guard together with a loop like in the example on `AsyncFd`, because the `clear_ready` method no longer takes a context, and thus it no longer registers the current waker for wakeup.

[1]: https://docs.rs/tokio/1/tokio/io/unix/struct.AsyncFdReadyGuard.html#method.try_io

djc: Done. This seems quite a bit nicer! `clear_ready()` not taking a context seems like a footgun, maybe needs more docs?

quinn/src/udp.rs
```diff
@@ -53,14 +49,9 @@ impl UdpSocket {
...
-            Ok(n) => Poll::Ready(Ok(n)),
-            Err(ref e) if e.kind() == io::ErrorKind::WouldBlock => {
-                self.io.clear_read_ready(cx, mio::Ready::readable())?;
-                Poll::Pending
-            }
-            Err(e) => Poll::Ready(Err(e)),
+        match ready!(self.io.poll_read_ready(cx))?.try_io(|io| io.get_ref().recv_ext(bufs, meta)) {
+            Ok(res) => Poll::Ready(res),
+            Err(_) => Poll::Pending,
         }
```
Darksonn, line 55: This is also wrong. You should loop when it returns an error, like in the example I linked.

djc:
So far unable to make sense of the failure in quinn-h3 here. This looks suspicious though:

```
Jan 05 22:48:57.680 TRACE drive{id=0}:recv{space=Data pn=3}:frame{ty=STREAM}: quinn_proto::connection::streams: got stream id=client bidirectional stream 0 offset=0 len=25 fin=true
```

It seems like the client decides to close the control stream almost immediately? But doesn't emit any logs about why it might be doing so.

@stammw (or someone else), would love to get some help with this.

quinn/src/udp.rs
djc: Ah, that's surprising -- thanks for noticing -- should have focused more on the example as you suggested.

stammw (dismissed):

Dessix:
Is the quinn-h3 failure still holding this back @djc, or are you waiting on further review?

Ralith:
I just need to find time to review this; it's a high priority.

quinn/src/streams.rs
```diff
@@ -573,26 +582,23 @@ where
...
-
     fn poll_read(
         self: Pin<&mut Self>,
         cx: &mut Context<'_>,
-        buf: &mut [u8],
-    ) -> Poll<io::Result<usize>> {
-        AsyncRead::poll_read(self, cx, buf)
+        buf: &mut ReadBuf<'_>,
+    ) -> Poll<io::Result<()>> {
+        let _ = ready!(RecvStream::poll_read(self.get_mut(), cx, buf))?;
```
Ralith, line 600: Is the `let _` here necessary, given that `RecvStream::poll_read` has a `()` success type?

quinn/src/streams.rs
```diff
@@ -770,7 +781,7 @@ pub enum ReadExactError {
     FinishedEarly,
     /// A read error occurred
     #[error("{0}")]
-    ReadError(#[source] ReadError),
+    ReadError(#[from] ReadError),
```
Ralith, line 784: Should this refactoring be separated out?

quinn/src/tests.rs
Ralith: Doesn't look like anything was done here. Was the decision to leave it for future work as an unrelated refactoring?

quinn-proto/src/frame.rs
```diff
@@ -704,7 +703,7 @@ fn scan_ack_blocks(packet: &[u8], largest: u64, n: usize) -> Option<usize> {
         let block = buf.get_var().ok()?;
         smallest = smallest.checked_sub(block)?;
     }
-    Some(buf.position() as usize)
+    Some((buf.position() - start) as usize)
```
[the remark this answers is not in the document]
djc: Should definitely be split out, have done so now.

It's not quite as straightforward as what you're suggesting because of the `checked_sub()?` calls, which need to propagate `None` in order to keep it easy to read. Could return `Option<()>` or create a new error wrapper that has `UnexpectedEnd` and `InvalidSub` or whatever, but that doesn't seem so nice. So I've otherwise kept it as is.

quinn/src/platform/fallback.rs
```diff
@@ -5,24 +5,25 @@ use std::{
...
 /// platforms.
 #[derive(Debug)]
 pub struct UdpSocket {
-    io: PollEvented<mio::net::UdpSocket>,
+    io: tokio::net::UdpSocket,
 }
 
 impl UdpSocket {
     pub fn from_std(socket: std::net::UdpSocket) -> io::Result<UdpSocket> {
+        socket.set_nonblocking(true);
```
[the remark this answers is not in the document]
djc: I'm not seeing any lint, but fixed this up.

quinn/src/platform/fallback.rs
```diff
@@ -31,15 +32,29 @@ impl UdpSocket {
...
+                    // We need to report that some packets were sent in this case, so we rely on
+                    // errors being either harmlessly transient (in the case of WouldBlock) or
+                    // recurring on the next call.
+                    false => return Poll::Ready(Ok(sent)),
+                    true => return Poll::Ready(Err(e)),
+                },
+                Poll::Pending => match sent == 0 {
+                    false => return Poll::Ready(Ok(sent)),
+                    true => return Poll::Pending,
+                },
```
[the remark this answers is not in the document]
djc: Ah, good idea.

quinn/src/streams.rs
djc: Doesn't appear to be!

quinn/src/streams.rs
djc: Yup, done.

quinn-proto/src/frame.rs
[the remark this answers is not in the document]
Ralith: I was thinking we'd write out `.ok_or(UnexpectedEnd)` for those cases, accepting a slightly more verbose (but readable) implementation for a simpler interface.

Ralith (approved):

quinn-proto/src/frame.rs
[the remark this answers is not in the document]
djc: What I dislike about that is that `UnexpectedEnd` is incorrect for the errors arising from `checked_sub()`...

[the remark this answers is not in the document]
Ralith: Oh, right; this could be `IterErr` as used by the caller, which already has the necessary cases.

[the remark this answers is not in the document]
djc: Nice one, fixed that up.

Ralith (approved):

quinn/src/platform/fallback.rs
```diff
@@ -1,46 +1,77 @@
...
-                    // recurring on the next call.
-                    return Ok(sent);
-                }
-                Err(e) => {
-                    return Err(e);
-                }
+                // We need to report that some packets were sent in this case, so we rely on
+                // errors being either harmlessly transient (in the case of WouldBlock) or
+                // recurring on the next call.
+                Poll::Ready(Err(_)) | Poll::Pending if sent == 0 => return Poll::Ready(Ok(sent)),
```
Matthias247, line 47: I think this needs to be `if sent != 0`

djc: Doh... Thanks for looking into it!

Ralith (approved):

PR #877, opened 2020-10-31
Use vectors instead of boxed slices

Matthias247:
The conversion from Vec<u8> into Box<[u8]> is not always for free.
It will call `Vec::into_boxed_slice`, which will call `shrink_to_fit`.
https://github.com/rust-lang/rust/blob/28f03ac4c08fc7ec62428d0b914e1510ce7ee2cb/library/alloc/src/vec.rs#L689

If the `Vec` isn't fully utilized before, this will cause a reallocation
and a copy of all data. This might currently happen with every
outgoing packet.

This change simply keeps things as `Vec<u8>`, which works just fine since
the IO layer can deal with it.

### DIFF
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -135,12 +135,12 @@ impl Assembler {
     }
 
     #[cfg(test)]
-    fn next(&mut self, size: usize) -> Option<Box<[u8]>> {
+    fn next(&mut self, size: usize) -> Option<Vec<u8>> {
         let mut buf = vec![0; size];
         let read = self.read(&mut buf).unwrap();
         buf.resize(read, 0);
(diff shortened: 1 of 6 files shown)


Matthias247:
It should be called out that benchmark results don't really show any interesting impact.

In some runs it might be 1% faster (190MB/s instead of 188MB/s), but it's so far from consistent that it might just be measuring noise.

I think benchmarks which use more concurrent streams and connections might show a bigger difference, but the current setup isn't allocation or memory-bandwidth limited.

djc:
LGTM, though I'd like @Ralith's thoughts before merging.

djc (approved):

Ralith (approved):
It's not our bottleneck, but I'm always happy to see things get a little more efficient regardless.

Issue #730, opened 2020-04-22
Multithreaded UDP send


Ralith:
We could improve performance in multithreaded multi-connection environments by arranging for outgoing UDP datagrams to not typically pass through the endpoint driver task.

This could be accomplished by moving the `UdpSocket` out of the `Endpoint`'s `Mutex` and sharing it with every `Connection`. Care will be necessary to ensure that connection drivers encountering EWOULDBLOCK on send are all properly woken when the socket becomes writable again.

An alternative approach is to use multiple file descriptors that share the UDP port, either one per connection or one per thread. This complicates receiving packets, but dovetails gracefully with multithreaded receive support as discussed in #636.

djc:
By "sharing it with every `Connection`", do you mean cloning?

Ralith:
I meant sharing the same socket, perhaps via `Arc`. The alternative approach would involve either cloning the socket or constructing new sockets that share the port; I'm not sure what the differences in behavior are there.

PR #878, opened 2020-10-31
Transmit multiple packets at once from connection to endpoint task

Matthias247:
Currently all ougoing packets are sent 1 by 1 from the connection
task to the endpoint task using an async queue. This has a couple of
downsides:
1. It requires additional synchronization for each queue operation.
2. It requires an additional heap allocation due to the nature of
  an unbounded futures channel.
3. It might lead to situations where less than the maximum number
  of possible packets could be transmtited in the endpoint task.
  It gets woken up after the first packet(s) arrive, might send those,
  but miss packets which are shortly queued after this.

I opted to add another event variant to cover for single/multiple
packets, to make sure the single packet version doesn't have an
extra allocation. A more elegant version might however be to
import SmallVec, and use a certain inline list size. I however didn't
want to add additional dependencies.
Yet another option is to just always use `Vec`.  But I thought if
this about efficiency - why not look into really efficient.

Performance gain in the bulk transmission test seems to be in the
1% range. Not that much, but better than nothing.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -744,11 +744,29 @@ where
 {
     fn drive_transmit(&mut self) {
         let now = Instant::now();
+
+        let mut single = None;
+        let mut multiple = Vec::new();
+
         while let Some(t) = self.inner.poll_transmit(now) {
+            if multiple.is_empty() && single.is_none() {
+                single = Some(t);
+            } else {
+                if let Some(single) = single.take() {
+                    multiple.push(single);
+                }
+                multiple.push(t);
+            }
+        }
+
+        if let Some(ev) = match (single.is_some(), !multiple.is_empty()) {
+            (true, false) => Some(EndpointEvent::Transmit(single.unwrap())),
+            (false, true) => Some(EndpointEvent::TransmitMultiple(multiple)),
+            (true, true) => panic!("Invalid state"),
+            (false, false) => None,
+        } {
             // If the endpoint driver is gone, noop.
-            let _ = self
-                .endpoint_events
-                .unbounded_send((self.handle, EndpointEvent::Transmit(t)));
+            let _ = self.endpoint_events.unbounded_send((self.handle, ev));
         }
     }
 
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -358,6 +358,9 @@ where
                         }
                     }
                     Transmit(t) => self.outgoing.push_back(t),
+                    EndpointEvent::TransmitMultiple(t) => {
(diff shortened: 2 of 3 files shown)


Ralith (commented):
I'm not sure how I feel about the complexity for a marignal benefit; are you sure it's above the noise floor? For a more impactful approach, https://github.com/quinn-rs/quinn/issues/730 may be of interest.

Matthias247:
I suspect the impact might be bigger when you actually run multiple connections which all race for the endpoint. But there's at the moment not a good benchmark for it.

The code change is not that big. If you would use smallvec, you can get unify both cases again, and it should be < 10LoC. Therefore I still think it won't hurt.

> For a more impactful approach, #730 may be of interest.

Yes and no. My feeling is that one of the biggest bottlenecks besides crypto is currently the task hopping that Quinn does. In order to get data out, it needs to be transmitted from Stream -> Connection Task -> Endpoint task. On the receiving side it's the same. That shows up when you switch to a threaded tokio runtime and the performance immediately drops to 50% => In this case the task switches are no longer just loops which dequeue some instructions, but are now starting to race for everything.

I think there's a lot of potential for improving this without actually using multiple sockets. E.g. when reading, you can check whether there is actually outgoing data to send before waking the connection task again. Or you might even poll for the to-send packets immediately and forward them to the endpoint, and get one less hop. 

Or for sending, you might just add all ready-to-send connections into a set, wake up the endpoint driver, and let the endpoint driver generate and send all the packets at once. 

But all those are rather big refactorings, and might require different internal APIs  - e.g. to determine whether a connection needs to send outgoing packets without polling for packets.

Using multiple sockets is mostly a different opimization - but I think this one would be more impactful if you run multiple endpoints and each endpoint is bound to a different socket. If you clone the socket for the connection, I fear you might just run into races on a different layer (e.g. the OS and socket layer), which might be even harder to debug because they are platform specific.

djc:
How hard it would be to come up with a better benchmark?

Ralith:
A benchmark demonstrating higher impact on more concurrent workloads would definitely help motivate this.

> big refactorings

Good thoughts. There's also a lot of lower hanging fruit left in general, I think, e.g. `recvmmsg`, GSO, delayed ACK, and reduced flow control update frequency.

> If you clone the socket for the connection, I fear you might just run into races on a different layer (e.g. the OS and socket layer), which might be even harder to debug because they are platform specific.

That's definitely true. I was thinking mainly about server performance, where you have many incoming connections to the same endpoint, but maybe the better approach there would be to support preferred addresses.

Matthias247:
Closing, since GSO (#953) changes this behavior anyway

PR #879, opened 2020-10-31
Update to draft 32

Ralith:
Incorporates all documented changes between draft 29 and the wg master, excluding h3. Nothing major.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Quinn was created and is maintained by Dirkjan Ochtman and Benjamin Saunders.
 
 ## Status
 
-- [x] QUIC draft 27 with TLS 1.3
+- [x] QUIC draft 32 with TLS 1.3
 - [x] Cryptographic handshake
 - [x] Stream data w/ flow control and congestion control
 - [x] Connection close
--- a/quinn-proto/src/lib.rs
+++ b/quinn-proto/src/lib.rs
@@ -18,7 +18,7 @@
 #![allow(clippy::cognitive_complexity)]
 #![allow(clippy::too_many_arguments)]
 
-use std::{fmt, net::SocketAddr, ops, time::Duration};
+use std::{convert::TryInto, fmt, net::SocketAddr, ops, time::Duration};
 
 mod cid_queue;
 #[doc(hidden)]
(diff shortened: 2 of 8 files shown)


interop/src/main.rs
```diff
@@ -65,9 +65,9 @@ impl From<&Alpn> for Vec<Vec<u8>> {
     fn from(alpn: &Alpn) -> Vec<Vec<u8>> {
         match alpn {
             Alpn::H3 => vec![quinn_h3::ALPN.into()],
-            Alpn::Hq => vec![b"hq-29"[..].into()],
+            Alpn::Hq => vec![b"hq-31"[..].into()],
```
djc, line 68: I wonder, should we actually leave this at 29 for a while? From the mailing list it looks like large deployments are sticking to the draft-29 version identifiers even as they deploy improvements from 30 and later.

README.md
```diff
@@ -45,7 +45,7 @@ Quinn was created and is maintained by Dirkjan Ochtman and Benjamin Saunders.
 
 ## Status
 
-- [x] QUIC draft 27 with TLS 1.3
+- [x] QUIC draft 31 with TLS 1.3
```
djc, line 48: Another commit mentions draft 32, but you used 31 first. Which is it? Are they different (enough)?

djc (commented):
Thanks for doing this, LGTM.

quinn-proto/src/lib.rs
```diff
@@ -117,7 +117,7 @@ pub mod fuzzing {
     }
 }
 /// The QUIC protocol version implemented
-const VERSION: u32 = 0xff00_001d;
+const VERSION: u32 = 0xff00_001f;
```
Matthias247, line 120: Please leave it at the old version - maybe with a comment.
Browsers like Safari and Chrome will now starting to use Chrome, but they will all announce Draft 29 version support even if they are 30+.
With this change things would be needlessly imcompatible.

README.md
Ralith: 2bff792 is from after draft 31 was cut, but draft 32 has not been formally released yet.

djc: I have email from the mailing list announcing draft 32, from Oct 21st. How do you determine it hasn't been released?

Ralith:
Removed the global version number bump, and relaxed the compatible version check to support a range rather than a single version.

README.md
Ralith: Whoops, good catch.

interop/src/main.rs
Ralith: Done.

djc (approved):

PR #880, opened 2020-10-31
Send only significant MAX_STREAM_DATA updates

Matthias247:
The current version of Quinn tries to enqueue a MAX_STREAM_DATA
update after every read from the stream. Those could potentially be
tiny. Besides wasting network capacity with sending tiny packets, that
behavior also causes a wakeup on the connection task.

This change triggers MAX_STREAM_DATA frames only to be enqueued
if they are deemed significant enough. In the version here, this means
is bound to 12.5% of the overall window - but this could be changed.

Note that the general connection task wakeup after reads is still there.
A similar change would need to be performed for updating the
connection window, in order to determine whether the connection task
wakeup is still necessary.

I gathered some example metrics for the amount and increment of window
updates in the benchmark:

Before:
Num max stream announcements: 1663, Total window announced: 58500130, Avg window diff: 35177

After:
Num max stream announcements: 301, Total window announced: 58195055, Avg window diff: 193339

==> This sends 5x less window updates

### DIFF
--- a/quinn-proto/src/connection/streams.rs
+++ b/quinn-proto/src/connection/streams.rs
@@ -180,7 +180,8 @@ impl Streams {
         match rs.read(buf) {
             Ok(Some(len)) => {
                 self.local_max_data = self.local_max_data.saturating_add(len as u64);
-                Ok(Some((len, rs.receiving_unknown_size())))
+                let (_, transmit_max_stream_data) = rs.max_stream_data(self.stream_receive_window);
+                Ok(Some((len, transmit_max_stream_data)))
             }
             Ok(None) => {
                 entry.remove_entry();
@@ -206,7 +207,8 @@ impl Streams {
         match rs.read_unordered() {
             Ok(Some((buf, offset))) => {
                 self.local_max_data = self.local_max_data.saturating_add(buf.len() as u64);
-                Ok(Some((buf, offset, rs.receiving_unknown_size())))
+                let (_, transmit_max_stream_data) = rs.max_stream_data(self.stream_receive_window);
+                Ok(Some((buf, offset, transmit_max_stream_data)))
             }
             Ok(None) => {
                 entry.remove_entry();
@@ -505,7 +507,10 @@ impl Streams {
                 continue;
             }
             sent.max_stream_data.insert(id);
-            let max = rs.assembler.bytes_read() + self.stream_receive_window;
+
+            let (max, _) = rs.max_stream_data(self.stream_receive_window);
+            rs.record_sent_max_stream_data(max);
+
             trace!(stream = %id, max = max, "MAX_STREAM_DATA");
             buf.write(frame::Type::MAX_STREAM_DATA);
             buf.write(id);
@@ -944,6 +949,7 @@ pub enum WriteError {
 struct Recv {
     state: RecvState,
     assembler: Assembler,
+    sent_max_stream_data: u64,
 }
 
 impl Recv {
@@ -1031,6 +1037,39 @@ impl Recv {
                     Err(ReadError::Blocked)
                 }
             }
+        }
+    }
+
+    /// Returns the window that should be advertised in a `MAX_STREAM_DATA` frame
+    ///
+    /// The method returns a tuple which consists of the window that should be
+    /// announced, as well as a boolean parameter which indicates if a new
+    /// transmission of the value is recommended. If the boolean value is
+    /// `false` the new window should only be transmitted if a previous transmission
+    /// had failed.
+    fn max_stream_data(&mut self, stream_receive_window: u64) -> (u64, bool) {
+        let max_stream_data = self.assembler.bytes_read() + stream_receive_window;
+
+        // Only announce a window update if it's significant enough
+        // to make it worthwhile sending a MAX_STREAM_DATA frame.
+        // We use here a fraction of the configured stream receive window to make
+        // the decision, and accomodate for streams using bigger windows requring
+        // less updates. A fixed size would also work - but it would need to be
+        // smaller than `stream_receive_window` in order to make sure the stream
+        // does not get stuck.
+        let diff = max_stream_data - self.sent_max_stream_data;
+        let transmit = self.receiving_unknown_size() && diff >= (stream_receive_window / 8);
+        (max_stream_data, transmit)
+    }
+
+    /// Records that a `MAX_STREAM_DATA` announcing a certain window was sent
+    ///
+    /// This will suppress enqueuing further `MAX_STREAM_DATA` frames unless
+    /// either the previous transmission was not acknowledged or the window
+    /// further increased.
+    pub fn record_sent_max_stream_data(&mut self, sent_value: u64) {
+        if sent_value > self.sent_max_stream_data {
+            self.sent_max_stream_data = sent_value;
         }
     }
 


quinn-proto/src/connection/mod.rs
```diff
@@ -814,18 +814,24 @@ where
...
     /// The return value if `Ok` contains the bytes and their offset in the stream.
     pub fn read_unordered(&mut self, id: StreamId) -> Result<Option<(Bytes, u64)>, ReadError> {
-        Ok(self.streams.read_unordered(id)?.map(|(buf, offset, more)| {
-            self.add_read_credits(id, more);
-            (buf, offset)
-        }))
+        Ok(self
+            .streams
+            .read_unordered(id)?
+            .map(|(buf, offset, max_stream_data_update)| {
```
Matthias247, line 820: I just rename the `more`s here for a bit more clarity on what the purpose of this parameter is

djc: Could you please pull that out into a separate commit? I can suggest some git incantations if that helps.

quinn-proto/src/connection/streams.rs
```diff
@@ -1034,6 +1048,27 @@ impl Recv {
         }
     }
 
+    /// Returns the window that should be advertised for this Stream in a
```
djc, line 1051: I think I'd prefer to leave the `announcable` prefixes and infixes out here, that feels a little wordy and doesn't seem to add that much value.

Also in doc comments, probably should keep the first line of the doc comment a bit shorter so that it fits on a line, and our convention is to not end these with a period (and have an empty line between the first one and the second, perhaps?).

djc (commented):
This looks like a nice improvement, thanks for working on it!

quinn-proto/src/connection/streams.rs
Matthias247: I updated to follow the  comment conventions, and removed some of the `announcable` in the body.

I wouldn't like to remove it from the function name however, since the destinction seems important here.

quinn-proto/src/connection/mod.rs
Matthias247: I could. Just felt it helps together with this change, and is a rather tiny rename

djc: I find it harder to review this change if it's mixed in with the rename, so please split it.

quinn-proto/src/connection/streams.rs
djc: Distinction between what and what? I think the `update` serves to make it clear enough.

quinn-proto/src/connection/streams.rs
```diff
@@ -180,7 +180,10 @@ impl Streams {
         match rs.read(buf) {
             Ok(Some(len)) => {
                 self.local_max_data = self.local_max_data.saturating_add(len as u64);
-                Ok(Some((len, rs.receiving_unknown_size())))
+                let announce_window_update = rs
+                    .announcable_window_update(self.stream_receive_window)
```
djc, line 184: Please also remove the `announce` in the variable name here, and below.

quinn-proto/src/connection/mod.rs
Matthias247: Done in https://github.com/quinn-rs/quinn/pull/881

quinn-proto/src/connection/streams.rs
Matthias247: Distinction between:
- The maximum window we could announce at all.
- A window update we could sent, but won't do it due to it not being deemed significant enough

I guess I can also change it by switching the return type of the function to `fn window_update(&self, stream_receive_window: u64) ->(u64, bool)` where the bool indicates whether transmission is wanted. Or call it `max_stream_data`.  Will leave it to you which option is preferred, and make that change

djc: To me it makes sense that the stream gets to decide whether a window update needs to be announced. If the stream decides it needs to be announced, it will return it to the connection, which will just act on the stream's behalf. As such, between the connection and stream there's no necessity to introduce the concept of announceability. (And therefore I prefer the original `Option`-based return type.)

quinn-proto/src/connection/streams.rs
```diff
@@ -1034,6 +1042,39 @@ impl Recv {
...
+    /// If there is no window update necessary at the moment, this returns `None`.
+    ///
+    /// If the `peek` parameter is set to `false`, the method will assume the
+    /// window update will actually be enqueued for transmission by the caller,
+    /// and track the transmission. Therefore a immediate follow-up call to this
+    /// method would return `None`.
+    ///
+    /// If `peek` is set to `false`, no transmission of the outgoing update would
+    /// be recorded.
+    fn window_update(&mut self, stream_receive_window: u64, peek: bool) -> Option<u64> {
```
Ralith, line 1056: I think callsites would be a bit more readable if `peek` was removed in favor of a separate call (or direct assignment) that updates the counter.

quinn-proto/src/connection/streams.rs
```diff
@@ -504,8 +506,13 @@ impl Streams {
             if rs.is_finished() {
                 continue;
             }
+
+            let max = match rs.window_update(self.stream_receive_window, false) {
+                Some(max) => max,
+                None => continue,
+            };
```
Ralith, line 513: If this packet is lost, this change will lead us to fail to retransmit the `MAX_STREAM_DATA` frame, possibly deadlocking the sender. We should instead perform this check before adding the stream to `pending.max_stream_data` and leave frame construction unchanged, ensuring that a loss will cause the latest data to be retransmitted.

Ralith (changes requested):
Thanks, this is definitely needed!

quinn-proto/src/connection/streams.rs
Matthias247: I had the separate assignment initially. But I removed it after djc's comments to make the `Streams` struct more self-contained. A separate method would be possible

quinn-proto/src/connection/streams.rs
Matthias247: Oh, that's not good. I thought Quinn retransmits all lost frames completely, and therefore thought this is good. 

So the idea is: Anytime `pending` is set it grabs the highest possible window from the stream. And the `read` methods make sure pending is only set once whenever the increment is high enough? I think this has a catch, but I will look into it.

Ralith: > I thought Quinn retransmits all lost frames completely

We keep track of what needs to be retransmitted if a packet is deemed lost by recording the retransmittable frames carried by each in-flight packet, and re-queueing those frames when the packet is deemed lost. See `sent.max_stream_data.insert(id)` below. The key is that that data's abstract; we don't store the literal frames, because they might become out of date. Instead, we just rerun this exact code again when it's time to retransmit.

> So the idea is:

Any time a stream appears in `pending.max_stream_data`, yeah, that's the idea exactly.

quinn-proto/src/connection/streams.rs
Ralith: I definitely think a separate method is clearer. @djc, thoughts?

Matthias247: This was now kind of necessary to properly handle the retransmits anyway

quinn-proto/src/connection/streams.rs
Matthias247: I think the new logic should cover it. The transmission will now suppress further updates, but retransmits can still grab the value

quinn-proto/src/connection/streams.rs
```diff
@@ -1034,6 +1040,39 @@ impl Recv {
...
+        let transmit = self.receiving_unknown_size() && diff >= (stream_receive_window / 8);
+        (max_stream_data, transmit)
+    }
+
+    /// Records that a `MAX_STREAM_DATA` announcing a certain window was transmitted
+    ///
+    /// This will suppress enqueuing further `MAX_STREAM_DATA` frames unless
+    /// either the previous transmission was not acknowledged or the window
+    /// further increased.
+    pub fn record_transmitted_max_stream_data(&mut self, transmitted_value: u64) {
```
djc, line 1070: I find the `record_` prefix here pretty superfluous. Otherwise this seems fine.

Ralith (dismissed):
LGTM, thanks! I like how retransmits further defer future updates; wouldn't have occurred to me to structure it that way.

quinn-proto/src/connection/streams.rs
Matthias247: Well it should have some prefix, since it’s not a getter. And I don’t prefer using `set_` for this. Since that one sounds like you can always call it with an arbitrary value.

djc: Maybe replace `transmitted` with `sent`? I just find `record_transmitted_max_stream_data` very wordy.

Matthias247: I updated it to `sent`. 

However I personally don't think anything wordy or verbose in code is bad, as long as it helps readers to understand what is meant.

Ralith (approved):

djc:
Thanks!

PR #881, opened 2020-10-31
Update add_read_credits parameter name to max_stream_data_update

Matthias247:
It wasn't super clear what `more` means here. The purpose of this
parameter is to signal whether a `MAX_STREAM_DATA` frame needs to be
sent. This change tries to improve that.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -814,18 +814,24 @@ where
     ///
     /// The return value if `Ok` contains the bytes and their offset in the stream.
     pub fn read_unordered(&mut self, id: StreamId) -> Result<Option<(Bytes, u64)>, ReadError> {
-        Ok(self.streams.read_unordered(id)?.map(|(buf, offset, more)| {
-            self.add_read_credits(id, more);
-            (buf, offset)
-        }))
+        Ok(self
+            .streams
+            .read_unordered(id)?
+            .map(|(buf, offset, transmit_max_stream_data)| {
+                self.add_read_credits(id, transmit_max_stream_data);
+                (buf, offset)
+            }))
     }
 
     /// Read from the given recv stream
     pub fn read(&mut self, id: StreamId, buf: &mut [u8]) -> Result<Option<usize>, ReadError> {
-        Ok(self.streams.read(id, buf)?.map(|(len, more)| {
-            self.add_read_credits(id, more);
-            len
-        }))
+        Ok(self
+            .streams
+            .read(id, buf)?
+            .map(|(len, transmit_max_stream_data)| {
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
(I meant a separate commit, not a whole separate PR. Unlike the tokio project, we aim for clean atomic commits, ideally rebased as-is rather than squashing commits.)

Matthias247:
> (I meant a separate commit, not a whole separate PR. Unlike the tokio project, we aim for clean atomic commits, ideally rebased as-is rather than squashing commits.)

Ok, I wasn't aware about that. Most people on Github don't like rebases/squashes, because github doesn't have a good interface to compare between revisions of the CR then. Adding commits and squashing before merge makes that part easier

djc:
My philosophy is mostly that it's on the author to present the commit series in a way that's easy to review, which will then also go into the main branch's commit history for later perusal. I mostly don't have a problem with re-reviewing some stuff on updates, but that might just be me.

quinn-proto/src/connection/mod.rs
```diff
@@ -2789,10 +2795,10 @@ where
         self.streams.alloc_remote_stream(&self.peer_params, dir);
     }
 
-    fn add_read_credits(&mut self, id: StreamId, more: bool) {
+    fn add_read_credits(&mut self, id: StreamId, max_stream_data_update: bool) {
```
Ralith, line 2798: I'm not sure this is much clearer. Maybe an imperative form (e.g. `update_max_stream_data`) rather than passive? Or something more abstract, like `more_data_expected`?

Ralith (commented):
Regarding process: While diffing PR revisions is poorly supported, multiple interdependent PRs are also poorly supported, so it's a bit of a pick-your-poison thing if you want to have small atomic changes at all. It's also difficult to deeply review a small atomic change that's spread across multiple fixup commits. With comments as a record of previously identified issues, I find @djc's preferred approach works pretty well for me.

quinn-proto/src/connection/mod.rs
Matthias247: You are right. I updated to the first. What I don't like about `more_data_expected` is that this would also be true if we don't want transmit a `MAX_STREAM_DATA` frame.

We can also add the transmission part, like `transmit_max_stream_data`. or `require_max_stream_data_update`?

Ralith: I think I like `transmit` more than `update`, good call.

Ralith (approved):
Approved for content. Commit message is out of date, though!

Matthias247:
I guess you might be able to edit it while merging? If not I can also push another revision

PR #883, opened 2020-11-01
Batched receive support

Ralith:
Surprisingly, this only seems to give about 1% better throughput when CPU-bound. Still, probably worth having.

### DIFF
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -201,6 +201,11 @@ where
                 break;
             }
         }
+        if !endpoint.incoming.is_empty() && endpoint.incoming_live {
+            if let Some(task) = endpoint.incoming_reader.take() {
+                task.wake();
+            }
+        }
         if endpoint.ref_count == 0 && endpoint.connections.is_empty() {
             Poll::Ready(Ok(()))
         } else {
@@ -275,12 +280,7 @@ where
                         match self.inner.handle(now, meta.addr, meta.ecn, data) {
                             Some((handle, DatagramEvent::NewConnection(conn))) => {
                                 let conn = self.connections.insert(handle, conn);
-                                if self.incoming_live {
-                                    self.incoming.push_back(conn);
-                                    if let Some(task) = self.incoming_reader.take() {
-                                        task.wake();
-                                    }
-                                }
+                                self.incoming.push_back(conn);
                             }
                             Some((handle, DatagramEvent::ConnectionEvent(event))) => {
                                 // Ignoring errors from dropped connections that haven't yet been cleaned up
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -84,7 +84,7 @@ where
             *addr
         };
         let (ch, conn) = endpoint.inner.connect(config, addr, server_name)?;
-        Ok(endpoint.create_connection(ch, conn))
+        Ok(endpoint.connections.insert(ch, conn))
     }
 
     /// Switch to a new UDP socket
@@ -115,8 +115,8 @@ where
     pub fn close(&self, error_code: VarInt, reason: &[u8]) {
         let reason = Bytes::copy_from_slice(reason);
         let mut endpoint = self.inner.lock().unwrap();
-        endpoint.close = Some((error_code, reason.clone()));
-        for sender in endpoint.connections.values() {
+        endpoint.connections.close = Some((error_code, reason.clone()));
+        for sender in endpoint.connections.senders.values() {
             // Ignoring errors from dropped connections
             let _ = sender.unbounded_send(ConnectionEvent::Close {
                 error_code,
@@ -220,7 +220,7 @@ where
         }
         // Drop all outgoing channels, signaling the termination of the endpoint to the associated
         // connections.
-        endpoint.connections.clear();
+        endpoint.connections.senders.clear();
     }
 }
 
@@ -238,14 +238,10 @@ where
     incoming_live: bool,
     driver: Option<Waker>,
     ipv6: bool,
-    connections: HashMap<ConnectionHandle, mpsc::UnboundedSender<ConnectionEvent>>,
-    // Stored to give out clones to new ConnectionInners
-    sender: mpsc::UnboundedSender<(ConnectionHandle, EndpointEvent)>,
+    connections: ConnectionSet,
     events: mpsc::UnboundedReceiver<(ConnectionHandle, EndpointEvent)>,
     /// Number of live handles that can be used to initiate or handle I/O; excludes the driver
     ref_count: usize,
-    /// Set if the endpoint has been manually closed
-    close: Option<(VarInt, Bytes)>,
     driver_lost: bool,
     recv_buf: Box<[u8]>,
     idle: Broadcast,
@@ -266,7 +262,7 @@ where
                     let data = (&self.recv_buf[0..meta[0].len]).into();
                     match self.inner.handle(now, meta[0].addr, meta[0].ecn, data) {
                         Some((handle, DatagramEvent::NewConnection(conn))) => {
-                            let conn = self.create_connection(handle, conn);
+                            let conn = self.connections.insert(handle, conn);
                             if self.incoming_live {
                                 self.incoming.push_back(conn);
                                 if let Some(task) = self.incoming_reader.take() {
@@ -278,6 +274,7 @@ where
                             // Ignoring errors from dropped connections that haven't yet been cleaned up
                             let _ = self
                                 .connections
+                                .senders
                                 .get_mut(&handle)
                                 .unwrap()
                                 .unbounded_send(ConnectionEvent::Proto(event));
@@ -345,7 +342,7 @@ where
                 Poll::Ready(Some((ch, event))) => match event {
                     Proto(e) => {
                         if e.is_drained() {
-                            self.connections.remove(&ch);
+                            self.connections.senders.remove(&ch);
                             if self.connections.is_empty() {
                                 self.idle.wake();
                             }
@@ -354,6 +351,7 @@ where
                             // Ignoring errors from dropped connections that haven't yet been cleaned up
                             let _ = self
                                 .connections
+                                .senders
                                 .get_mut(&ch)
                                 .unwrap()
                                 .unbounded_send(ConnectionEvent::Proto(event));
@@ -368,8 +366,19 @@ where
             }
         }
     }
+}
 
-    fn create_connection(
+#[derive(Debug)]
+struct ConnectionSet {
+    senders: HashMap<ConnectionHandle, mpsc::UnboundedSender<ConnectionEvent>>,
+    /// Stored to give out clones to new ConnectionInners
+    sender: mpsc::UnboundedSender<(ConnectionHandle, EndpointEvent)>,
+    /// Set if the endpoint has been manually closed
+    close: Option<(VarInt, Bytes)>,
+}
+
+impl ConnectionSet {
+    fn insert<S: proto::crypto::Session + 'static>(
         &mut self,
         handle: ConnectionHandle,
         conn: proto::generic::Connection<S>,
@@ -382,8 +391,12 @@ where
             })
             .unwrap();
         }
-        self.connections.insert(handle, send);
+        self.senders.insert(handle, send);
         Connecting::new(handle, conn, self.sender.clone(), recv)
+    }
+
+    fn is_empty(&self) -> bool {
+        self.senders.is_empty()
     }
 }
 
@@ -419,7 +432,7 @@ where
         } else if let Some(conn) = endpoint.incoming.pop_front() {
             endpoint.inner.accept();
             Poll::Ready(Some(conn))
-        } else if endpoint.close.is_some() {
+        } else if endpoint.connections.close.is_some() {
             Poll::Ready(None)
         } else {
             endpoint.incoming_reader = Some(cx.waker().clone());
@@ -454,16 +467,18 @@ where
             socket,
             inner,
             ipv6,
-            sender,
             events,
             outgoing: VecDeque::new(),
             incoming: VecDeque::new(),
             incoming_live: true,
             incoming_reader: None,
             driver: None,
-            connections: HashMap::new(),
+            connections: ConnectionSet {
+                senders: HashMap::new(),
+                sender,
+                close: None,
+            },
             ref_count: 0,
-            close: None,
             driver_lost: false,
             recv_buf: recv_buf.into(),
             idle: Broadcast::new(),
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -20,7 +20,8 @@ use crate::{
     broadcast::{self, Broadcast},
     builders::EndpointBuilder,
     connection::Connecting,
-    udp::{RecvMeta, UdpSocket, BATCH_SIZE},
+    platform::BATCH_SIZE,
+    udp::{RecvMeta, UdpSocket},
     ConnectionEvent, EndpointEvent, VarInt, IO_LOOP_BOUND,
 };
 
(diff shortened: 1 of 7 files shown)


quinn/src/endpoint.rs
```diff
@@ -446,22 +471,25 @@ where
     S: proto::crypto::Session,
 {
     pub(crate) fn new(socket: UdpSocket, inner: proto::generic::Endpoint<S>, ipv6: bool) -> Self {
-        let recv_buf = vec![0; inner.config().get_max_udp_payload_size().min(64 * 1024) as usize];
+        let recv_buf =
```
Matthias247, line 474: Do you maybe need to do anything for better alignment of buffers? I get `max_udp_payload_size` could have any arbitrary value

Ralith: Is there reason to believe buffer alignment matters for `recvmmsg`? We could round up if it's useful to.

Matthias247: We probably would need to read kernel code to know for sure :-) 
It could make it more efficient for it to map/copy things, or it might not matter at all. I would maybe just round up to the next 8byte boundary per chunk.

quinn/src/endpoint.rs
```diff
@@ -253,39 +253,43 @@ where
...
+            *iov = IoSliceMut::new(buf);
+        }
         loop {
-            let mut meta = [RecvMeta::default()];
-            match self
-                .socket
-                .poll_recv(cx, &mut [IoSliceMut::new(&mut self.recv_buf)], &mut meta)
-            {
+            let result = self.socket.poll_recv(cx, &mut iovs, &mut metas);
+            match result {
```
djc, line 266: We can match on the expression directly, right?

quinn/src/endpoint.rs
```diff
@@ -253,39 +253,43 @@ where
...
-                                self.incoming.push_back(conn);
-                                if let Some(task) = self.incoming_reader.take() {
-                                    task.wake();
+                    for (meta, buf) in metas.iter().zip(iovs.iter()).take(msgs) {
+                        match self
+                            .inner
+                            .handle(now, meta.addr, meta.ecn, buf[0..meta.len].into())
+                        {
+                            Some((handle, DatagramEvent::NewConnection(conn))) => {
+                                let conn = self.connections.insert(handle, conn);
```
djc, line 274: So here we're at `loop` / `match` / `for` / `match` (and then we still get two levels of `if`). Is there a clean way to make this less deeply nested, perhaps by just extracting some function from the inside, even it has just a single use?

quinn/src/endpoint.rs
```diff
@@ -253,39 +253,43 @@ where
...
     fn drive_recv(&mut self, cx: &mut Context, now: Instant) -> Result<bool, io::Error> {
         let mut recvd = 0;
+        let mut metas = [RecvMeta::default(); BATCH_SIZE];
+        let mut iovs = [IoSliceMut::new(&mut [])];
+        for (iov, buf) in iovs
+            .iter_mut()
+            .zip(self.recv_buf.chunks_mut(self.recv_buf.len() / BATCH_SIZE))
+        {
+            *iov = IoSliceMut::new(buf);
+        }
```
djc, line 263: This confuses me. Why do you instantiate a one-element `IoSliceMut` array first, then iterate over it and overwrite it?

quinn/src/platform/unix.rs
```diff
@@ -167,17 +167,52 @@ impl super::UdpExt for UdpSocket {
         Ok(sent)
     }
 
+    #[cfg(not(any(target_os = "macos", target_os = "ios")))]
```
djc, line 170: Would be nice to do the factoring out of methods in a separate commit from introducing batching.

djc (commented):
First round of thoughts. Not sure I'm best reviewer, don't have any experience with the mechanics of how this is supposed to work. Do you have any reference materials/introductions/tutorials that you used?

quinn/src/endpoint.rs
Ralith: Right; stray change from a previous iteration. Fixed.

quinn/src/endpoint.rs
Ralith: Good catch, this wasn't intended, and was rendering the whole PR moot. Still no drastic change in perf though, even after manually verifying that multiple packets are being received per batch...

Ralith:
> Do you have any reference materials/introductions/tutorials that you used?

Just [the documentation](https://man7.org/linux/man-pages/man2/recvmmsg.2.html).

quinn/src/endpoint.rs
Ralith: This is complicated by borrowck. `iovs` mutably borrows from `self`, so we can't easily extract a method on `EndpointInner`.

quinn/src/platform/unix.rs
Ralith: Done.

quinn/src/endpoint.rs
```diff
@@ -251,41 +252,51 @@ impl<S> EndpointInner<S>
 where
     S: proto::crypto::Session + 'static,
 {
-    fn drive_recv(&mut self, cx: &mut Context, now: Instant) -> Result<bool, io::Error> {
+    fn drive_recv<'a>(&'a mut self, cx: &mut Context, now: Instant) -> Result<bool, io::Error> {
         let mut recvd = 0;
+        let mut metas = [RecvMeta::default(); BATCH_SIZE];
+        let mut iovs = MaybeUninit::<[IoSliceMut<'a>; BATCH_SIZE]>::uninit();
+        for (i, buf) in self
```
djc, line 259: Maybe turn the for-loop into a `for_each()` combinator call?

djc (commented):
LGTM (one more style nit).

quinn/src/endpoint.rs
```diff
@@ -254,33 +252,48 @@ impl<S> EndpointInner<S>
...
-                        .handle(now, addr, ecn, (&self.recv_buf[0..n]).into())
-                    {
-                        Some((handle, DatagramEvent::NewConnection(conn))) => {
-                            let conn = self.create_connection(handle, conn);
-                            if self.incoming_live {
-                                self.incoming.push_back(conn);
-                                if let Some(task) = self.incoming_reader.take() {
-                                    task.wake();
+            match self.socket.poll_recv(cx, &mut iovs, &mut metas) {
+                Poll::Ready(Ok(msgs)) => {
```
Matthias247, line 271: Later on there is a `recvd += 1;`, which counts the number of socket receive calls. Should this now be adjusted by `msgs` to account for the fact that it can now receive more messages at once?

quinn/src/platform/unix.rs
```diff
@@ -165,21 +167,52 @@ impl super::UdpExt for UdpSocket {
...
+                )
+            };
+            if n == -1 {
+                let e = io::Error::last_os_error();
+                if e.kind() == io::ErrorKind::Interrupted {
+                    continue;
+                }
+                return Err(e);
+            }
+            break n;
```
Matthias247, line 202: Want to add an `assert!(n >= 0)` here, to catch other weird OS return values?

quinn/src/platform/unix.rs
```diff
@@ -257,3 +263,57 @@ fn prepare_msg(
...
+fn prepare_recv(
+    buf: &mut IoSliceMut,
+    name: &mut MaybeUninit<libc::sockaddr_storage>,
+    ctrl: &mut cmsg::Aligned<MaybeUninit<[u8; CMSG_LEN]>>,
+    hdr: &mut libc::msghdr,
+) {
+    hdr.msg_name = name.as_mut_ptr() as _;
+    hdr.msg_namelen = mem::size_of::<libc::sockaddr_storage>() as _;
+    hdr.msg_iov = buf as *mut IoSliceMut as *mut libc::iovec;
+    hdr.msg_iovlen = 1;
```
Matthias247, line 276: Why is this 1? I had expected `buf.len()`

quinn/src/platform/unix.rs
```diff
@@ -257,3 +263,57 @@ fn prepare_msg(
...
+
+fn prepare_recv(
+    buf: &mut IoSliceMut,
+    name: &mut MaybeUninit<libc::sockaddr_storage>,
+    ctrl: &mut cmsg::Aligned<MaybeUninit<[u8; CMSG_LEN]>>,
+    hdr: &mut libc::msghdr,
+) {
+    hdr.msg_name = name.as_mut_ptr() as _;
+    hdr.msg_namelen = mem::size_of::<libc::sockaddr_storage>() as _;
+    hdr.msg_iov = buf as *mut IoSliceMut as *mut libc::iovec;
```
Matthias247, line 275: `buf.as_mut_ptr()`?

quinn/src/platform/unix.rs
Ralith: We're using exactly one iovec per datagram in an interface that supports gathering multiple iovecs.

quinn/src/platform/unix.rs
Ralith: We're passing a pointer to the iovec here; `as_mut_ptr` produces a pointer to the payload data.

quinn/src/platform/unix.rs
Ralith: We're using `== -1` for various syscalls throughout this file. If there's reason to believe that might not be portable for any of them (is there?) we should probably address it for them all in a separate PR.

quinn/src/platform/fallback.rs
```diff
@@ -30,7 +32,13 @@ impl super::UdpExt for UdpSocket {
         Ok(sent)
     }
 
-    fn recv_ext(&self, buf: &mut [u8]) -> io::Result<(usize, SocketAddr, Option<EcnCodepoint>)> {
-        self.recv_from(buf).map(|(x, y)| (x, y, None))
+    fn recv_ext(&self, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) -> io::Result<usize> {
+        let (len, addr) = self.recv_from(&mut bufs[0])?;
```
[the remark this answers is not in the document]
Ralith: We'll eventually be using a more efficient syscall on windows, of course. Do you have an idea for how this might be improved (for macOS as well) in the mean time?

quinn/src/endpoint.rs
Ralith: I'm inclined to avoid dubiously-justified magic for now in the spirit of maintenance.

quinn/src/platform/unix.rs
Matthias247: Oh right. I now remember I had a misunderstanding about this before

quinn/src/platform/unix.rs
Matthias247: It should be portable, and the doc say it's `-1`. However I sometimes feel like not trusting foreign code :)

quinn/src/platform/fallback.rs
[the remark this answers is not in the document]
Matthias247: A simple improvement is setting `BATCH_SIZE` dependent on the platform, and to `1` on windows

quinn/src/endpoint.rs
```diff
@@ -199,6 +201,12 @@ where
                 break;
             }
         }
+        if endpoint.incoming_dirty && endpoint.incoming_live {
```
Matthias247, line 204: Any reason why checking `!incoming.is_empty()` wouldn't work?

Ralith: I was thinking we shouldn't wake the consumer repeatedly if they're not reading, but of course consuming the waker takes care of that already. Good catch.

quinn/src/endpoint.rs
```diff
@@ -199,6 +201,11 @@ where
                 break;
             }
         }
+        if !endpoint.incoming.is_empty() && endpoint.incoming_live {
```
Matthias247, line 204: One more thought: Why is `incoming_live` important here? If it doesn't exist anymore, wouldn't it be also fair to get rid of this `Waker` (by at least doing `take`) in order to remove any potential cross-references?

If deregistering `incoming` would already remove the waker, then this check would also be not important.

Ralith: Good catch. I think this was left over from a long time ago. It's definitely not doing anything useful now; removed.

Matthias247: I've seen https://github.com/quinn-rs/quinn/pull/883/commits/d215e586d5c57370599bcb7076cc150640091fc6

Wouldn't it still be to helpful whether to decide to store something in `incoming` at all?
Without it, that list of pending incoming connections grows indefinitely? Not sure if there is any bound to it.

quinn/src/platform/fallback.rs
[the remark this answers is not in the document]
Ralith: Good call; done.

quinn/src/platform/unix.rs
```diff
@@ -317,3 +315,10 @@ fn decode_recv(
         ecn: EcnCodepoint::from_bits(ecn_bits),
     }
 }
+
+#[cfg(not(any(target_os = "macos", target_os = "ios")))]
+// Chosen somewhat arbitrarily; might benefit from additional tuning.
+pub const BATCH_SIZE: usize = 32;
```
djc, line 321: So I think this means we're allocating a 2 MB buffer by default, right? I feel like 8 or 16 would be good enough pending actual data.

quinn/src/endpoint.rs
```diff
@@ -257,12 +258,19 @@ where
...
         let mut recvd = 0;
         loop {
-            match self.socket.poll_recv(cx, &mut self.recv_buf) {
-                Poll::Ready(Ok((n, addr, ecn))) => {
-                    match self
-                        .inner
-                        .handle(now, addr, ecn, (&self.recv_buf[0..n]).into())
-                    {
+            let mut meta = [RecvMeta::default()];
+            match self
```
djc: I don't love the multi-line match expression here, how do you feel about naming `&mut [IoSliceMut::new(&mut self.recv_buf)]` to split it up?

quinn/src/endpoint.rs
```diff
@@ -257,12 +258,19 @@ where
...
+                .socket
+                .poll_recv(cx, &mut [IoSliceMut::new(&mut self.recv_buf)], &mut meta)
+            {
+                Poll::Ready(Ok(msgs)) => {
+                    debug_assert_eq!(msgs, 1);
+                    match self.inner.handle(
+                        now,
+                        meta[0].addr,
+                        meta[0].ecn,
+                        (&self.recv_buf[0..meta[0].len]).into(),
```
djc: If we name this last argument, is that enough to get us within the line limit?

quinn/src/endpoint.rs
djc: Maybe this runs into borrowck?

Ralith: Works out just fine, in fact!

quinn/src/endpoint.rs
Ralith: Yep!

quinn/src/platform/unix.rs
Ralith: IIRC I did dial this in somewhat for sending, though my benchmarking setup makes things a bit too noisy to make this easy. Instead, I've adjusted the max UDP payload size to something realistic for the internet, which ends up with us allocating less memory than before despite the batching.

Matthias247: Keep in mind that this is for a full endpoint, which can handle N connections. 2MB for 1000 connections (on a server) would e.g. be not a lot, but could even be undersized.  For an applicaton it's a different story - but since it needs the 2MB only once and it's still not a lot compared to socket send and receive buffers it won't matter.

I downloaded the branch, and gathered some stats on large the vectors are. Definitely seen a lot of receives which carried around 30 datagrams, so it isn't unused.

djc: Hmm, fair point!

djc (approved):

PR #884, opened 2020-11-01
Add connection level statistics

Matthias247:
This change adds some structs which track connection-level statistics.
These are rather helpful to determine how the implementation performs,
and which impact certain changes have.

The implementation is not complete. It's just a first step on getting
some more visibility into the transmitting side of the library. More
statistics - e.g. around congestion control, retransmission, timing, etc
could be added.

The stats are currently private, and I just debug printed them so far
where found useful. They could however be public if the API is fine.
Another possibility is to have the stats behind a feature flag if people
are concerned about extra memory usage.

### DIFF
--- a/quinn-proto/src/connection/stats.rs
+++ b/quinn-proto/src/connection/stats.rs
@@ -1,0 +1,71 @@
+/// Statistics about packets transmitted on a connection
+#[derive(Default)]
+#[non_exhaustive]
+pub struct PacketTransmissionStats {
+    /// The amount of packets transmitted on a connection
+    pub packets: u64,
+    /// The total amount of bytes transmitted on a connection
+    pub bytes: u64,
+}
+
+impl std::fmt::Debug for PacketTransmissionStats {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("PacketTransmissionStats")
+            .field("packets", &self.packets)
+            .field("bytes", &self.bytes)
+            .finish()
+    }
+}
+
+/// Statistics about frames transmitted or received on a connection
+#[derive(Default)]
+#[non_exhaustive]
+pub struct FrameStats {
+    pub acks: u64,
+    pub crypto: u64,
+    pub datagram: u64,
+    pub handshake_done: u8,
+    pub max_data: u64,
+    pub max_stream_data: u64,
+    pub max_streams_bidi: u64,
+    pub max_streams_uni: u64,
+    pub new_connection_id: u64,
+    pub path_challenge: u64,
+    pub path_response: u64,
+    pub ping: u64,
+    pub reset_stream: u64,
+    pub retire_connection_id: u64,
+    pub stop_sending: u64,
+    pub stream: u64,
+}
+
+impl std::fmt::Debug for FrameStats {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("FrameStats")
+            .field("ACK", &self.acks)
+            .field("CRYPTO", &self.crypto)
+            .field("DATAGRAM", &self.datagram)
+            .field("HANDSHAKE_DONE", &self.handshake_done)
+            .field("MAX_DATA", &self.max_data)
+            .field("MAX_STREAM_DATA", &self.max_stream_data)
+            .field("MAX_STREAMS_BIDI", &self.max_streams_bidi)
+            .field("MAX_STREAMS_UNI", &self.max_streams_uni)
+            .field("NEW_CONNECTION_ID", &self.new_connection_id)
+            .field("PATH_CHALLENGE", &self.path_challenge)
+            .field("PATH_RESPONSE", &self.path_response)
+            .field("PING", &self.ping)
+            .field("RESET_STREAM", &self.reset_stream)
+            .field("RETIRE_CONNECTION_ID", &self.retire_connection_id)
+            .field("STOP_SENDING", &self.stop_sending)
+            .field("STREAM", &self.stream)
+            .finish()
+    }
+}
+
+/// Connection statistics
+#[derive(Debug, Default)]
+#[non_exhaustive]
+pub struct ConnectionStats {
+    pub packet_tx: PacketTransmissionStats,
+    pub frame_tx: FrameStats,
+}
--- a/quinn-proto/src/connection/streams.rs
+++ b/quinn-proto/src/connection/streams.rs
@@ -14,6 +14,7 @@ use super::{
 };
 use crate::{
     coding::BufMutExt,
+    connection::stats::FrameStats,
     frame::{self, FrameStruct},
     transport_parameters::TransportParameters,
     Dir, Side, StreamId, TransportError, VarInt, MAX_STREAM_COUNT,
@@ -435,6 +436,7 @@ impl Streams {
         buf: &mut Vec<u8>,
         pending: &mut Retransmits,
         sent: &mut Retransmits,
+        stats: &mut FrameStats,
         max_size: usize,
     ) {
         // RESET_STREAM
@@ -455,6 +457,7 @@ impl Streams {
                 final_offset: stream.offset(),
             }
             .encode(buf);
+            stats.reset_stream += 1;
         }
 
         // STOP_SENDING
@@ -473,6 +476,7 @@ impl Streams {
             trace!(stream = %frame.id, "STOP_SENDING");
             frame.encode(buf);
             sent.stop_sending.push(frame);
+            stats.stop_sending += 1;
         }
 
         // MAX_DATA
@@ -487,6 +491,7 @@ impl Streams {
                 sent.max_data = true;
                 buf.write(frame::Type::MAX_DATA);
                 buf.write(max);
+                stats.max_data += 1;
             }
         }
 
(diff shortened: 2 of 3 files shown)


quinn-proto/src/connection/stats.rs
```diff
@@ -0,0 +1,80 @@
...
+            .field("number of bytes", &self.bytes)
+            .field("avg packet size", &self.avg_packet_size())
+            .finish()
+    }
+}
+
+/// Statistics about frames transmitted on a connection
+#[derive(Default)]
+pub struct FrameTransmissionStats {
+    pub acks: u64,
```
Matthias247, line 34: We can likely shrink down most of those to `u32` without danger. Or use `u32` with saturating increments

quinn-proto/src/connection/stats.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub max_data: u64,
+    pub max_streams_uni: u64,
+    pub max_streams_bidi: u64,
+    pub stream: u64,
+}
+
+impl std::fmt::Debug for FrameTransmissionStats {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("FrameTransmissionStats")
+            .field("ACK", &self.acks)
```
Matthias247, line 55: We can also live with the pure derive. I however kind of liked the frame names, and then started to order alphabetically

Matthias247:
This yields for example the following results in the benchmark:

### Sender:

> ConnectionStats { packet_tx: PacketTransmissionStats { number of packets: 177, number of bytes: 140055, avg packet size: 791 }, frame_tx: FrameTransmissionStats { ACK: 175, CRYPTO: 2, DATAGRAM: 0, HANDSHAKE_DONE: false, MAX_DATA: 0, MAX_STREAM_DATA: 0, MAX_STREAMS_UNI: 0, MAX_STREAMS_BIDI: 0, NEW_CONNECTION_ID: 5, PATH_CHALLENGE: 0, PATH_RESPONSE: 0, PING: 0, RESET: 0, RETIRE_CONNECTION_ID: 0, STOP_SENDING: 0, STREAM: 112 } }

### Receiver:

> ConnectionStats { packet_tx: PacketTransmissionStats { number of packets: 229, number of bytes: 8807, avg packet size: 38 }, frame_tx: FrameTransmissionStats { ACK: 226, CRYPTO: 3, DATAGRAM: 
0, HANDSHAKE_DONE: true, MAX_DATA: 0, MAX_STREAM_DATA: 111, MAX_STREAMS_UNI: 1, MAX_STREAMS_BIDI: 0, NEW_CONNECTION_ID: 4, PATH_CHALLENGE: 0, PATH_RESPONSE: 0, PING: 0, RESET: 0, RETIRE_CONNECTION_ID: 1, STOP_SENDING: 0, STREAM: 0 } }

quinn-proto/src/connection/streams.rs
```diff
@@ -435,6 +436,7 @@ impl Streams {
         buf: &mut Vec<u8>,
         pending: &mut Retransmits,
         sent: &mut Retransmits,
+        stats: &mut ConnectionStats,
```
Ralith, line 439: Should this just be `FrameTransmissionStats`?

quinn-proto/src/connection/stats.rs
```diff
@@ -0,0 +1,80 @@
+/// Statistics about packets transmitted on a connection
+#[derive(Default)]
+pub struct PacketTransmissionStats {
```
Ralith, line 3: I'd like to see all these structs marked `#[non_exhaustive]` to reflect the intention that more members will be added in the future.

quinn-proto/src/connection/stats.rs
Ralith: I'd favor sticking with `u64` and just not worrying about it.

quinn-proto/src/connection/stats.rs
Ralith: No objection from me.

quinn-proto/src/connection/stats.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// The amount of packets transmitted on a connection
+    pub packets: u64,
+    /// The total amount of bytes transmitted on a connection
+    pub bytes: u64,
+}
+
+impl PacketTransmissionStats {
+    /// Calculates the average packet size
+    pub fn avg_packet_size(&self) -> u64 {
```
Ralith, line 12: Does this really need to be a helper method?

quinn-proto/src/connection/stats.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+}
+
+impl std::fmt::Debug for PacketTransmissionStats {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("PacketTransmissionStats")
+            .field("number of packets", &self.packets)
+            .field("number of bytes", &self.bytes)
+            .field("avg packet size", &self.avg_packet_size())
+            .finish()
```
Ralith, line 27: I feel like a more conventional, perhaps derived, `Debug` impl would be fine here. Not sure `Debug` impls that produce invalid syntax is a good habit.

Ralith (commented):
I think this is valuable to have and worth exposing in the public API as a diagnostic and tuning tool. Similar statistics are exposed by e.g. popular TCP implementations.

quinn-proto/src/connection/streams.rs
Matthias247: Could be. Although might be more in the future. I have no strong opinion. Since it’s internal it could always be changed

quinn-proto/src/connection/stats.rs
Matthias247: What else do you have in mind? A field that gets updated with every packet?

quinn-proto/src/connection/stats.rs
Matthias247: How is it invalid syntax?

Ralith: In the sense that `number of bytes` is not an identifier.

quinn-proto/src/connection/stats.rs
Ralith: Just leaving it for callers to compute if they want it.

quinn-proto/src/connection/streams.rs
Ralith: Given that this function specifically deals with frames, I'd err on the side of the more concise and narrowly-scoped approach, though it's not too important.

quinn-proto/src/connection/stats.rs
Matthias247: I don’t think Debug really cares about it. It produces arbitrary syntax. And even if we go towards „it’s a regular Json-like syntax“, the whitespace still seems ok. 
Anyway, I can change it to have underscores if people care about it

quinn-proto/src/connection/stats.rs
Matthias247: I’m not in favor of that. I found it immensely helpful to just debug print this struct and get all the information I need without adding any more external code. 

And packet utilization can for example be a more interesting metric than pure overall bytes or number of packets. 

I think there will be more metrics which will fall into the same category. Eg average bps, average packet loss, etc. I wouldn’t want all consumers to have to recalculate those on their end.

quinn-proto/src/connection/stats.rs
Ralith: It doesn't really care; it's mostly a stylistic question, in that it's inconsistent with most other `Debug` impls. I also don't think "number of packets: 42" is actually any clearer than "packets: 42".

quinn-proto/src/connection/stats.rs
Ralith: How do we draw the line between derived statistics that are worth providing out of the box and ones that should be left to downstream?

Matthias247: That can be discussed as part of each CR which wants to add more, right?
I think the main criteria should be: Is this statistic understandable for all users?
I don't think having more is any issue, since derived statistics would always be zero cost, apart from when printing them (which you probably won't do unless you are debugging).

Ralith: My concern is that the set of derived statistics that are plausibly useful to someone is very large, and I don't want to end up with hundreds/thousands of lines of statistical analysis code that we have to maintain in the long run. The cost isn't computational, but in maintenance effort (and to a lesser extent, compile time). That sort of logic really belongs in metrics pipelines that the source data can easily be dumped into, and where dependencies for data structures and algorithms to support analysis are justifiable.

That doesn't mean we necessarily can't have helpers for some overwhelmingly simple and useful stuff, but "anything comprehensible" is too low a bar.

quinn-proto/src/connection/stats.rs
```diff
@@ -0,0 +1,83 @@
...
+            .field("bytes", &self.bytes)
+            .field("avg_packet_size", &self.avg_packet_size())
+            .finish()
+    }
+}
+
+/// Statistics about frames transmitted or received on a connection
+#[derive(Default)]
+#[non_exhaustive]
+pub struct FrameCounter {
```
Matthias247, line 35: I now renamed this, since it seems useful for the sending as well as receiving end

quinn-proto/src/connection/stats.rs
djc: @Matthias247 I think this is the only open question here. I'm with @Ralith on the maintenance aspects. Can you please remove the method?

djc (approved):
Thanks, this will be very useful!

Issue #854, opened 2020-09-24
Fuzz pre-decryption packet decoding


Ralith:
The main entry point of interest here is probably `PartialDecode::new`. This is a particularly important interface to audit as there are absolutely zero barriers to feeding data into it.

Ralith:
This was implemented in #885.

PR #885, opened 2020-11-01
Fuzz packet

jafow:
#854 

This adds initial fuzzing target for packet decoding.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -55,7 +55,17 @@ jobs:
         if: always()
         with:
           command: clippy
-          args: --workspace --all-targets -- -D warnings
+          args: --workspace --exclude fuzz --all-targets -- -D warnings
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+          components: clippy
+      - name: lint fuzz
+        run: |
+          cd fuzz
+          cargo clippy -- -D warnings
 
   audit:
     runs-on: ubuntu-latest
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -31,11 +31,11 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --workspace --all-targets
+          args: --all-targets
       - uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --workspace --exclude fuzz
+          args: --workspace
 
   lint:
     runs-on: ubuntu-latest
@@ -55,7 +55,7 @@ jobs:
         if: always()
         with:
           command: clippy
-          args: --workspace --exclude fuzz --all-targets -- -D warnings
+          args: --all-targets -- -D warnings
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -35,7 +35,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --workspace
+          args: --workspace --exclude fuzz
 
   lint:
     runs-on: ubuntu-latest
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -1,8 +1,11 @@
 [package]
-name = "quinn-fuzz"
-version = "0.0.0"
-authors = ["Automatically generated"]
+name = "fuzz"
+version = "0.1.0"
+authors = [
+    "Benjamin Saunders <ben.e.saunders@gmail.com>", "Dirkjan Ochtman <dirkjan@ochtman.nl>",
+]
 publish = false
+license = "MIT OR Apache-2.0"
 edition = "2018"
 
 [package.metadata]
@@ -17,10 +20,6 @@ features = ["arbitrary"]
 path = "../quinn-proto"
 package = "quinn-proto"
 version = "0.6.1"
-
-# Prevent this from interfering with workspaces
-[workspace]
-members = ["."]
 
 [[bin]]
 name = "streams"
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -33,3 +33,9 @@ name = "streamid"
 path = "fuzz_targets/streamid.rs"
 test = false
 doc = false
+
+[[bin]]
+name = "packet"
+path = "fuzz_targets/packet.rs"
+test = false
+doc = false
--- a/fuzz/fuzz_targets/packet.rs
+++ b/fuzz/fuzz_targets/packet.rs
@@ -5,5 +5,8 @@ extern crate proto;
 use proto::fuzzing::{PacketParams, PartialDecode};
 
 fuzz_target!(|data: PacketParams| {
-    let decode = PartialDecode::new(data.buf, data.cid).unwrap();
+    let len = data.buf.len();
+    if let Ok(decoded) = PartialDecode::new(data.buf, data.local_cid_len) {
+        assert_eq!(len, decoded.0.len());
+    }
 });
--- a/fuzz/fuzz_targets/packet.rs
+++ b/fuzz/fuzz_targets/packet.rs
@@ -1,15 +1,15 @@
 #![no_main]
-use libfuzzer_sys::fuzz_target;
 
-extern crate proto;
+use libfuzzer_sys::fuzz_target;
 use proto::fuzzing::{PacketParams, PartialDecode};
+extern crate proto;
 
 fuzz_target!(|data: PacketParams| {
     let len = data.buf.len();
     if let Ok(decoded) = PartialDecode::new(data.buf, data.local_cid_len) {
         match decoded.1 {
             Some(x) => assert_eq!(len, decoded.0.len() + x.len()),
-            None => assert_eq!(len, decoded.0.len())
+            None => assert_eq!(len, decoded.0.len()),
         }
     }
 });
--- a/fuzz/fuzz_targets/packet.rs
+++ b/fuzz/fuzz_targets/packet.rs
@@ -1,0 +1,9 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+
+extern crate proto;
+use proto::fuzzing::{PacketParams, PartialDecode};
+
+fuzz_target!(|data: PacketParams| {
+    let decode = PartialDecode::new(data.buf, data.cid).unwrap();
+});
--- a/fuzz/fuzz_targets/packet.rs
+++ b/fuzz/fuzz_targets/packet.rs
@@ -7,6 +7,9 @@ use proto::fuzzing::{PacketParams, PartialDecode};
 fuzz_target!(|data: PacketParams| {
     let len = data.buf.len();
     if let Ok(decoded) = PartialDecode::new(data.buf, data.local_cid_len) {
-        assert_eq!(len, decoded.0.len());
+        match decoded.1 {
+            Some(x) => assert_eq!(len, decoded.0.len() + x.len()),
+            None => assert_eq!(len, decoded.0.len())
+        }
     }
 });
--- a/fuzz/fuzz_targets/streams.rs
+++ b/fuzz/fuzz_targets/streams.rs
@@ -49,22 +49,16 @@ fuzz_target!(|input: (StreamParams, Vec<Operation>)| {
                 stream.accept(dir);
             }
             Operation::Finish(id) => {
-                if let Ok(x) = stream.finish(id) {
-                    x
-                }
+                let _ = stream.finish(id);
             }
             Operation::ReceivedStopSending(sid, err_code) => {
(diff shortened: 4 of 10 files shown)


quinn-proto/src/packet.rs
```diff
@@ -90,7 +90,7 @@ impl PartialDecode {
     }
 
     /// Length of QUIC packet being decoded
-    pub(crate) fn len(&self) -> usize {
+    pub fn len(&self) -> usize {
```
jafow, line 93: this is only made public for the placeholder assertion in the test and not necessarily required.

jafow (commented):
Hello! This can be run using `cargo-fuzz` with libfuzzer as `cargo fuzz run packet`. 

Running it as is, the fuzzer will cause a panic on the debug assertion on `bytes.len() <= MAX_CID_SIZE` when the connectionid is made, https://github.com/quinn-rs/quinn/blob/main/quinn-proto/src/shared.rs#L72. I don't understand yet why the `PacketDecodeError` isn't raised before that, as it seems like it should. 
I'd be happy to share a test case to reproduce if it is helpful or of interest. 

I welcome any guidance and input on how best to proceed. For example, we could run assertions on some of the underlying data without having to test the encryption code.

quinn-proto/src/packet.rs
```diff
@@ -19,13 +19,13 @@ use crate::{
 // to inspect the version and packet type (which depends on the version).
 // This information allows us to fully decode and decrypt the packet.
 #[derive(Debug)]
-pub(crate) struct PartialDecode {
+pub struct PartialDecode {
```
djc, line 22: Should make this `#[doc(hidden)]` then (same for the other things in this commit).

quinn-proto/src/lib.rs
```diff
@@ -115,6 +117,25 @@ pub mod fuzzing {
...
+    pub struct PacketParams {
+        pub local_cid_len: usize,
+        pub buf: BytesMut
+    }
+
+    impl Arbitrary for PacketParams {
+        fn arbitrary(u: &mut Unstructured<'_>) -> Result<Self> {
+            let local_cid_len: usize = usize::arbitrary(u)?;
+            let bytes: Vec<u8> = Vec::arbitrary(u)?;
+            let mut buf = BytesMut::with_capacity(bytes.len());
```
djc, line 131: I think you can just use `BytesMut::new()` here, since it'll do the equivalent allocation on `put_slice()` anyway.

quinn-proto/src/lib.rs
```diff
@@ -115,6 +117,25 @@ pub mod fuzzing {
...
+
+    #[derive(Debug)]
+    pub struct PacketParams {
+        pub local_cid_len: usize,
+        pub buf: BytesMut
+    }
+
+    impl Arbitrary for PacketParams {
+        fn arbitrary(u: &mut Unstructured<'_>) -> Result<Self> {
+            let local_cid_len: usize = usize::arbitrary(u)?;
```
djc, line 129: Presumably this doesn't need the type annotation?

djc (commented):
Some style nits. This looks really useful!

quinn-proto/src/packet.rs
Ralith: That's fine since this type isn't reexported under normal conditions anyway.

Ralith:
> Running it as is, the fuzzer will cause a panic on the debug assertion on bytes.len() <= MAX_CID_SIZE when the connectionid is made

This is due to an illegal value being generated for `local_cid_len`. The `Arbitrary` impl should be adjusted to never produce a value greater than `MAX_CID_SIZE` for that field.

quinn-proto/src/packet.rs
Ralith: Though, on review, this isn't (at least currently) a very interesting property to check.

fuzz/fuzz_targets/packet.rs
```diff
@@ -0,0 +1,12 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+
+extern crate proto;
+use proto::fuzzing::{PacketParams, PartialDecode};
+
+fuzz_target!(|data: PacketParams| {
+    let len = data.buf.len();
+    if let Ok(decoded) = PartialDecode::new(data.buf, data.local_cid_len) {
+        assert_eq!(len, decoded.0.len());
```
Ralith, line 10: This only holds if `decoded.1` is `None`. If it's instead `Some(x)`, we should check against `decoded.0.len() + x.len()`.

Ralith (changes requested):

.github/workflows/rust.yml
```diff
@@ -55,7 +55,17 @@ jobs:
         if: always()
         with:
           command: clippy
-          args: --workspace --all-targets -- -D warnings
+          args: --workspace --exclude fuzz --all-targets -- -D warnings
```
jafow, line 58: Some explanation is in order here for this hack or workaround :-)

The goal is to move `fuzz` up as a first-class package in the workspace. The challenge to it is to pass the necessary cfg flag to `fuzz` so that its dependencies are built as expected, but there's just no truly excellent way to do this I've found. 

The tradeoff here is to put some hackery into the CI and risk losing chance to be invited to the secret devops society :sunglasses:, but the benefit I see is we keep the configuration local to the package. 

Another alternative is to set configuration with `RUSTFLAGS=` but it affects everything and seems less desirable. I am open to input on this, and am not tied to this specific approach so tightly :-)

quinn-proto/src/lib.rs
```diff
@@ -17,6 +17,7 @@
 // Fixes welcome:
 #![allow(clippy::cognitive_complexity)]
 #![allow(clippy::too_many_arguments)]
+#![allow(clippy::len_without_is_empty)]
```
jafow, line 20: many lint changes in these later PRs.

clippy warns on this because of we make the packet length method public for fuzz test. I'm unsure to ignore outright. Probably better to either implement the `is_empty()` or make the `.len()` private again and remove from packet fuzz target. Any preference?

.github/workflows/rust.yml
djc: I think a nicer solution here might be to use the `workspace.default-members` configuration to exclude the fuzz crate from default cargo commands?

quinn-proto/src/lib.rs
djc: I think ignoring the lint is right, but let's scope it tightly to the particular places it complains, rather than globally for the entire library?

.github/workflows/rust.yml
```diff
@@ -55,7 +55,17 @@ jobs:
...
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+          components: clippy
+      - name: lint fuzz
+        run: |
+          cd fuzz
+          cargo clippy -- -D warnings
```
Ralith, line 68: Do we need something like this for formatting also?

quinn-proto/src/connection/streams.rs
```diff
@@ -21,6 +21,7 @@ use crate::{
     Dir, Side, StreamId, TransportError, VarInt, MAX_STREAM_COUNT,
 };
 
+#[doc(hidden)]
```
Ralith, line 24: Do we actually need these annotations, if the reexports are guarded by `cfg(fuzzing)`?

fuzz/fuzz_targets/streams.rs
```diff
@@ -49,16 +49,22 @@ fuzz_target!(|input: (StreamParams, Vec<Operation>)| {
...
-                stream.received_reset(rs);
+                if let Ok(x) = stream.received_reset(rs) {
+                    assert!(x & true | true);
+                }
             }
             Operation::Reset(id) => {
-                stream.reset(id);
+                if let Ok(x) = stream.reset(id) {
+                    x
+                }
```
Ralith, line 67: What do the above three changes accomplish? They look like no-ops to me.

Ralith (changes requested):

.github/workflows/rust.yml
jafow: To my surprise no, as it seems that the `--all` option will apply formatting to all packages regardless of their membership in the root workspace `workspace.default-members` list. 

To demonstrate, I've added some format errors into the fuzz package and ran the action on a branch of my fork. Happily it is caught [as expected by the lint check](https://github.com/jafow/quinn/runs/1460865241?check_suite_focus=true).

fuzz/fuzz_targets/streams.rs
jafow: That's correct. This change and those below are lint fixes raised from applying clippy to the fuzz package. These changes fix warning on unused `Result`. I'd opted not to ignore any lints that weren't already ignored in quinn but I'd be happy to revert the changes in this file and instead add an allow for the lint warning.

quinn-proto/src/connection/streams.rs
jafow: I believe that is correct, and these annotations are added only to satisfy clippy. 

Without them it raises warnings on missing docs, i.e.
```bash
error: missing documentation for a struct
  --> quinn-proto/src/connection/streams.rs:25:1
   |
25 | pub struct Streams {
   | ^^^^^^^^^^^^^^^^^^
   |
   = note: `-D missing-docs` implied by `-D warnings`

error: missing documentation for an associated function
   --> quinn-proto/src/connection/streams.rs:78:5
    |
78  | /     pub fn new(
79  | |         side: Side,
80  | |         max_remote_uni: VarInt,
81  | |         max_remote_bi: VarInt,
...   |
117 | |         this
118 | |     }
    | |_____^

etc ... 
```

Perhaps instead we add an ignore for missing docs here?

djc: Yeah, if that's the reason than doing `allow(missing_docs)` makes more sense here.

jafow: very good thanks that sounds fine to me. updated

fuzz/fuzz_targets/streams.rs
Ralith: Rather than blanket suppressing lints, let's explicitly discard results we don't care about with `let _ = ...`. As is, this code is confusing.

Ralith (changes requested):

quinn-proto/src/connection/streams.rs
```diff
@@ -21,6 +21,7 @@ use crate::{
     Dir, Side, StreamId, TransportError, VarInt, MAX_STREAM_COUNT,
 };
 
+#[allow(missing_docs)]
```
Ralith, line 24: Rather than spamming this annotation across every fuzzing-only interface, could we replace the crate-level `#![warn(missing_docs)]` with `#![cfg_attr(not(fuzzing), warn(missing_docs))]` or similar?

quinn-proto/src/frame.rs
```diff
@@ -735,6 +735,7 @@ impl<'a> Iterator for AckIter<'a> {
     }
 }
 
+#[doc(hidden)]
```
Ralith, line 738: Stale attribute, also below.

Ralith (changes requested):

fuzz/fuzz_targets/streams.rs
jafow: yep I agree, sounds good

fuzz/Cargo.toml
```diff
@@ -1,8 +1,11 @@
 [package]
-name = "quinn-fuzz"
-version = "0.0.0"
-authors = ["Automatically generated"]
+name = "fuzz"
+version = "0.1.0"
+authors = [
+    "Benjamin Saunders <ben.e.saunders@gmail.com>", "Dirkjan Ochtman <dirkjan@ochtman.nl>",
```
Ralith, line 5: Feel free to add yourself here!

quinn-proto/src/connection/streams.rs
Ralith, line 24: Stale attribute?
```suggestion
```

quinn-proto/src/connection/streams.rs
```diff
@@ -72,6 +73,7 @@ pub struct Streams {
     stream_receive_window: u64,
 }
 
+#[doc(hidden)]
```
Ralith, line 76:
```suggestion
```

Ralith (dismissed):
Thanks, stale attribs aside LGTM!

quinn-proto/src/connection/streams.rs
jafow: ah oops yes indeed my mistake

fuzz/Cargo.toml
jafow: :-D thank you I appreciate it but I am satisfied just being in the git log for now, if I put my name here then someone in the future may ask me to fix bugs!

Ralith: Haha, always something to be mindful of :D

Ralith (approved):

djc:
Looks great, thanks for working on this!

PR #886, opened 2020-11-02
Recreate and reset timers less often

Matthias247:
The idea behind this change is to minimize the overhead for configuring
timers. This is based on a few ideas:
- Most timers never expire. Some other event will happen before
their expiration.
- However we need an active timer most of the time. It isn't worth it
  to deallocate timers and allocate them later on.

  Based on those facts, we can save overhead of creating and resetting
  timers by keeping old timers running, as long as their timeout is
  lower than the new desired timeout. They typically still won't
  expire. If they do expire, we can lazily check whether the expiration
  is real and adjust the timer again.

  Some stats for this CR show:
  In a setup with 700 required timer changes
  - only 1 timer really expired
  - with this change we have to reset the timer 4-8 times forward
  - we get 20-22 wrong expirations reported, and have to shift the
    timer backwards

  This still means this saves about 670 timer adjustments - which
  means reducing timer adjustments by 95%.
  However there is a catch that a wrong expiration also has a certain
  cost, since it wakes the task up when there is nothing to do.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -655,6 +655,7 @@ where
             on_connected: Some(on_connected),
             connected: false,
             timer: None,
+            timer_deadline: None,
             conn_events,
             endpoint_events,
             blocked_writers: HashMap::new(),
@@ -726,6 +727,7 @@ where
     on_connected: Option<oneshot::Sender<bool>>,
     connected: bool,
     timer: Option<Delay>,
+    timer_deadline: Option<TokioInstant>,
     conn_events: mpsc::UnboundedReceiver<ConnectionEvent>,
     endpoint_events: mpsc::UnboundedSender<(ConnectionHandle, EndpointEvent)>,
     pub(crate) blocked_writers: HashMap<StreamId, Waker>,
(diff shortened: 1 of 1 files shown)


Matthias247:
This falls a bit into the category of "smart optimizations, with questionable benefit". 
Besides stats which show that there is less timer churn there is no overall visible benefit in performance.
I guess that part is still overshadowed by other things.

Therefore take it or leave it.

quinn/src/connection.rs
```diff
@@ -859,37 +861,63 @@ where
...
-                    self.inner.handle_timeout(Instant::now());
-                    self.timer = None;
-                    // A timer expired, so the caller needs to check for new transmits, which might
-                    // cause new timers to be set.
-                    keep_going = true;
+
+        // Check whether we need to (re)set the timer. If so, we must poll again to ensure the
+        // timer is registered with the runtime (and check whether it's already
+        // expired). Otherwise, exit the loop.
+        match self.inner.poll_timeout().map(TokioInstant::from_std) {
```
Matthias247, line 868: Btw: I wasn't sure why the current code checked expirations before checking whether timers are still active.
The other way around seems more logical to me, since it also allows to avoid polling the timer.

djc:
My intuition would be that resetting the timer is pretty cheap whereas waking up a task isn't. So I'm inclined to say we should not merge this. Would be curious what others think, though.

quinn/src/connection.rs
Ralith: IIRC the motivation was that invoking `handle_timeout` may cause new timers to be set, although the outer `keep_going` loop probably makes it moot semantically.

Ralith:
This commit seems to include both refactoring and the behavioral change. My intuition is that the actual behavior change could be just a few-line diff, which might make it more palatable? Or at least easier to keep rebased until we've addressed more fundamental perf issues.

> My intuition would be that resetting the timer is pretty cheap whereas waking up a task isn't.

I think it might not be this clear cut, because we're probably updating timers *way* more often than this change causes us to be spuriously woken by timeouts.

Matthias247:
> This commit seems to include both refactoring and the behavioral change. My intuition is that the actual behavior change could be just a few-line diff, which might make it more palatable?

Which part do you see as pure refactoring? Creating a timer just once, or the swap of the order of checking a timer requirement and polling the timer?

In any case, the change is not that much. Half of the 50 LoC in the CR is adding comments :-) 

>> My intuition would be that resetting the timer is pretty cheap whereas waking up a task isn't.

> I think it might not be this clear cut, because we're probably updating timers way more often than this change causes us to be spuriously woken by timeouts.

Right. And timers with tokio 0.2 are not that cheap. When creating them, it synchronizes on a global list of timers. When polling them, it does the same. https://github.com/tokio-rs/tokio/pull/3080 might address some of this overhead, with the same background idea.

djc:
I'm not resisting this change because of complexity, I think that's pretty manageable. If you both think it's an improvement, I'll happily accept the change.

Ralith (commented):
I was thinking the rearrangement of the code could be isolated from the introduction of `timer_deadline` and the logic to make resets lazy, yeah.

I buy that this could be valuable once more immediate bottlenecks are addressed. However, if https://github.com/tokio-rs/tokio/pull/3080 (or other upstream work) might moot the issue, I'm a bit concerned we'll end up with unnecessary complexity that'll never get removed. Between zero observable benefit at present and the possibility of the hypothetical benefit being erased in the future, my inclination is to revisit the lazy reset portion of this change if/when a benefit can be shown, and then take care to document the conditions for its removal.

quinn/src/connection.rs
```diff
@@ -859,37 +861,63 @@ where
...
-            ) {
-                (Some(timeout), &mut None) => self.timer = Some(delay_until(timeout)),
-                (Some(timeout), &mut Some(ref mut delay)) if delay.deadline() != timeout => {
-                    delay.reset(timeout);
-                }
-                (None, _) => {
-                    self.timer = None;
+            None => {
+                self.timer_deadline = None;
+                return keep_going;
```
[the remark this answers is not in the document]
Matthias247: It could totally be removed. The only place where it is `true` is if the timer **really** elapsed

Matthias247:
How about updating it to
- keep the timer in order to avoid having to recreate it, but drop the lazy reset with the expectation that the tokio update will fix that part.
- change order of checking the timeout and polling the timer, as done here

Ralith:
Sounds good to me!

djc:
+1 from me.

quinn/src/connection.rs
Matthias247: yes, that's handled by the outer loop

quinn/src/connection.rs
```diff
@@ -863,39 +865,49 @@ where
...
-                }
-                _ => break,
+            None => {
+                self.timer_deadline = None;
+                return false;
             }
         }
-        keep_going
+
+        if let Some(_deadline) = self.timer_deadline {
```
djc, line 895: Let's invert the condition here to return early `if self.timer_deadline.is_none()`, and same for `delay.poll_unpin(cx).is_ready()`? That prevents the rightward drift here.

djc (commented):
Looks like a nice improvement, some style nits.

quinn/src/connection.rs
Matthias247: Sounds good!

djc: I meant the second condition, too, not just the first.

Matthias247: You mean

```
if delay.poll_unpin(cx).is_pending() {
    return false;
}
```

Can be done - but won't really save another nesting level here

djc: Yes, that's what I mean. It won't save another nesting level, but I think it will make the code easier to comprehend.

djc (approved):
Thanks!

PR #889, opened 2020-11-05
Add `Connection::rtt` method

### LINKED ISSUE #888: Expose RTT probe value to user?
I know that quinn internally keeps track of the round-trip-time of a connection. Could an API be added to expose this value?

This is very useful information, to know the latency of the connection. Quinn can readily provide this information. By adding such a function, applications can easily know this information without needing application-level logic/protocol for it.

jamadazi:
Give users access to our RTT/latency estimate.

Closes #888

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1011,6 +1011,11 @@ where
         self.path.remote
     }
 
+    /// Current best estimate of this connection's latency (round-trip-time)
+    pub fn rtt(&self) -> Duration {
+        self.path.rtt.conservative()
+    }
+
     fn on_packet_sent(
         &mut self,
         now: Instant,
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1013,7 +1013,7 @@ where
 
     /// Current best estimate of this connection's latency (round-trip-time)
     pub fn rtt(&self) -> Duration {
-        self.path.rtt.conservative()
+        self.path.rtt.get()
     }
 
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -1011,6 +1011,11 @@ where
         self.path.remote
     }
 
+    /// Current best estimate of this connection's latency (round-trip-time)
+    pub fn rtt(&self) -> Duration {
+        self.path.rtt.conservative()
```
djc, line 1016: Why pick `conservative()` here, rather than `get()`? As I understand it, the latter represents our best estimate. (If we decide to stick with `conservative()`, we should probably elaborate on this in the docstring.

djc (commented):
Looks okay to modulo one question. Would like to hear what @Ralith thinks, too.

quinn-proto/src/connection/mod.rs
jamadazi: I wasn't sure about which to use. I didn't think about it much, I just saw that `conservative` was marked `pub` and `get` wasn't, so I assumed that `get` isn't intended to be used directly.

But I think you are right. `get` is the more correct thing to use. I'll update this.

jamadazi:
BTW, should I squash the commits and force-push, or something like that? I don't know what the process is like for editing PRs.

djc:
Don't worry about it, we can squash it on merge.

Ralith (approved):
LGTM. I wonder if we might eventually want to switch to exposing a single aggregate statistics structure to avoid churning the lock excessively, but that's an issue for another time.

Issue #890, opened 2020-11-05
Documentation - Examples Tracking Issue


TimonPost:
This issue describes the various examples that would be useful to have for Quinn. It belongs to a bigger overall documentation project as described here: #861.

# Todo

# Ideas

Various examples will be covered in the book. 

- [ ] Simple Send, Receive example with a unidirectional stream. 
- [ ] Simple Send, Receive example with a bidirectional stream. 
- [ ] Simple Send, Receive example with unreliable transport.
- [ ] Simple example demonstrating how to build a certificate.

TimonPost:
After trying to get an example for my self together. I got two main remarks:

- I have a hard time finding an easy example I can just copy and paste from. Examples like simple sending, receiving, using different stream types. I think it would be useful to add more simplistic examples to this library. What do you think?
- The async code makes it really difficult to see what the examples do. I can not inspect what kind of types are returned. My idea doesn't support this yet. Are there ideas that do support this?

PR #891, opened 2020-11-06
Example Modularisation

TimonPost:
When working on #868 I needed to get a bit familiar with the protocol. While looking through the examples I found it hard to find understand the example, therefore I created this issue #890. I am going to improve all examples in later phases, for now, I wanted to modularize the https examples a bit. Meaning, create smaller functions, and reduce big code chunks doing all kinds of things. With this, I hope to accomplish an easier introduction to the examples.   

## Changes
- Rename client/server to https_client/server for more clearance.
- Wrote README.md with instructions and explanation of the examples. 
- Moved the documentation comments at the top of the files to this README.md. Motivation is that README is MD format, and most editors and in GitHub are able to show this format in regards to rust comments. Besides that, it provides the user with a single entry point to the examples. My experience was that I had to first observe all the different files and look at their behaviors in order to figure out what I was needing. I hope this README can give some guidance here. 
- Modularized the server/client example to make it easier to read and understand. 
- Add some comments to the server/client examples.

I did not change any logic, behavior, mainly only moved the code around.


quinn/Cargo.toml
```diff
@@ -53,11 +53,11 @@ unwrap = "1.2.1"
 url = "2"
 
 [[example]]
-name = "server"
+name = "https_server"
```
Ralith, line 56: nit: I feel like the `s` is unnecessary, since no unencrypted form exists.
```suggestion
name = "http_server"
```

quinn/examples/README.md
```diff
@@ -0,0 +1,83 @@
+## HTTPS File Serving Example
+
+The `https` example shows how to request a file on the `server` from the `client` over HTTPS.
```
Ralith, line 3: We should clarify that this is HTTP/0.9, which is a toy protocol that's much simpler than real HTTP, especially HTTP/3.

quinn/examples/README.md
```diff
@@ -0,0 +1,83 @@
...
+The server listens for any client requesting a file. 
+If the file path is valid and allowed, it returns the contents. 
+
+Open up a terminal and execute:
+
+```text
+$ cargo run --example https_server ./
+```
+
+*additional arguments can be passed, see the `Opt` struct in the example.*
```
Ralith, line 16: I don't think this clarification is necessary.

quinn/examples/README.md
```diff
@@ -0,0 +1,83 @@
...
+$ cargo run --example insecure_connection --features="rustls/dangerous_configuration"
+```
+
+## Single Socket Example
+
+You can have multiple QUIC connections over a single UDP socket. This is especially
+useful, if you are building a peer-to-peer system where you potentially need to communicate with
+thousands of peers or if you have a
+[hole punched](https://en.wikipedia.org/wiki/UDP_hole_punching) UDP socket.
+Besides, QUIC servers and clients can both operate on the same UDP socket.
```
Ralith, line 64:
```suggestion
Additionally, QUIC servers and clients can both operate on the same UDP socket.
```
"Besides" is distractingly informal tone here.

quinn/examples/https_client.rs
```diff
@@ -11,9 +11,10 @@ use structopt::StructOpt;
 use tracing::{error, info};
 use url::Url;
 
+use quinn::{Connection, Endpoint, NewConnection};
+
 mod common;
 
-/// HTTP/0.9 over QUIC client
```
Ralith, line 16: Doc comments on a structopt struct appear in the `--help` output, so this is meaningful and should not be removed.

quinn/examples/https_client.rs
```diff
@@ -52,28 +58,51 @@ fn main() {
...
+    perform_request(url, &mut conn).await;
+
+    // Close connection.
+    conn.close(0u32.into(), b"done");
+
+    Ok(())
+}
+
+/// Initializes the client configuration from the command-line arguments.
+fn initialize_client_configuration(options: &Opt) -> Result<quinn::ClientConfig> {
```
Ralith, line 91: Naming nit:
```suggestion
fn make_client_configuration(options: &Opt) -> Result<quinn::ClientConfig> {
```
and similar for the others. IMO this is more readable.

Ralith (changes requested):
Thanks; overall I think this is a significant improvement. I particularly like the dedicated example README.

The code changes would be much easier to review if the commit that renames the files was different than the commit that refactors them, because then git would notice that the files were moved and display with a more succinct diff.

quinn/examples/connection.rs
```diff
@@ -1,18 +1,3 @@
-//! This example intends to use the smallest amount of code to make a simple QUIC connection.
```
djc, line 1: I'm not a fan of outright deleting this. Can we keep 2-3 lines of description and point to the README?

quinn/examples/http_client.rs
```diff
@@ -37,13 +39,18 @@ struct Opt {
 }
 
 fn main() {
+    // Setup logger with environment arguments.
```
djc, line 42: Please, no periods at the end of single-line comments.

quinn/examples/http_client.rs
```diff
@@ -52,28 +59,51 @@ fn main() {
...
 
 #[tokio::main]
 async fn run(options: Opt) -> Result<()> {
-    let url = options.url;
-    let remote = (url.host_str().unwrap(), url.port().unwrap_or(4433))
-        .to_socket_addrs()?
-        .next()
-        .ok_or_else(|| anyhow!("couldn't resolve to an address"))?;
+    let url = options.url.clone();
+    let mut endpoint = make_endpoint(&options)?;
```
djc, line 69: GitHub tells me this doesn't need to be mutable, is it wrong?

djc:
So overall I'm not really a fan of "modularization" by just breaking stuff up into single-use functions -- maybe unless it can be formatted such that the "next" function is defined immediately after the "previous" function, but that doesn't really work if you're extracting some part in the middle. The reason is that the indirection itself has a cost in making it harder to follow what's going on (although it can sometimes be helpful to make it clear what amount of state from the previous block will be used by the next -- this can of course also be done by using explicit lexical scopes and/or comments).

I would prefer to start with delineating different blocks of code that are currently in the same function by just introducing some comments in between that explain what the next step will be.

quinn/examples/connection.rs
TimonPost: I can do that, will introduce a bit of duplication which is a tradeoff to keep in mind.

TimonPost:
The code can be easier to read 'if your familiar' with it. This is dangerous in not seeing how new people would see the library. When I as a starter check this server/client example out I see the following in `server.rs`:

- Configuring endpoint
- Reading, creating, parsing, writing, configuring the certificate
- Creating endpoint
- Listening for connections 
- With man log statements. 

All in one function `run` of 78 lines. When new, 78 lines of code doing all kinds of stuff can be difficult to comprehend and in general a bad practice. I don't see how splitting it up into small functions doing one thing e.g. create a certificate, setup endpoint etc.. is worse. Next to that, similar functions are found in the `common.rs` and other examples. 

Said that the examples need a major rewrite, for the book I reduced everything to the bare minimum. I hope to introduce those in some time. For now, I'll do anything you need or want whether it is dropping this PR or modifying it a bit to meet @Ralith and @djc in the middle.

quinn/examples/connection.rs
djc: That's fair. It's just, people might "get" to this file in different ways, and it would be bad if someone found it without having seen the README and would then be in worse position than they would be now.

djc:
I'm sorry, but I outright disagree with any "best practices" saying that shorter functions are always better. These things are always trade-offs, and in my mind extracting single-use functions is overvalued (though it can be a good thing!). In this case, I'm not convinced it will make the examples easier to understand, because it will make the reader have to "jump around" trying to understand the flow of what's going on, and I think separating different parts with a comment block laying out what's going on is probably a better option.

On the other hand, code duplication *is* an important reason to introduce abstractions.

(I don't think any of @Ralith's comments contradict mine, but please point it out to me if I'm wrong.)

TimonPost:
I very agree along the lines of 'always best practice', I did not mention 'always' but 'in general' neither 'good practice' but 'bad practice'. Indeed, useless abstractions or splitting up code blocks for the sake of splitting up is not always wishable. @Ralith comments do not contradict, agree, or disagree with yours.  Raith only called it a "Significant Improvement". So I am happy to hear his opinion on this matter and find perhaps away in the middle.  

In the meantime, I will be covering your feedback and see what I can do to reduce the amount of code split ups and inline comments in functions. Do you have any particular functions you would like to see back as they were? In case you fully want to revert all functions back to originally I will close this pull request since I have fresh very minimal examples locally on my branches which I think are able to replace all current examples. 

Also if you don't think this does not add value and is not worth time to review further. I can always close it. I would much more like to have full consensus than to make this PR work by reverting all changes and leave in some.

Ralith:
I absolutely agree that one-use functions can be obfuscatory, and I've indeed seen a lot of example code that requires so much jumping around as to be extremely painful to follow. That said, I do think the structural changes here are largely beneficial. The functions that are introduced are small in number, have self-explanatory names, and have very narrow interfaces, taking 1-3 arguments and using simple return types.

Explicit scopes would be less effective at clearly isolating the *inputs* to each operation. Additionally, I think there's a readability benefit in e.g. strongly isolating the certificate wrangling logic in server.rs, which has little to do with routine Quinn API use, for all that it's worth having an example of. A reader who's trying to understand the basic lifecycle of a server will have an easier time without that large block breaking up the initialization.

Note I haven't completed a thorough code review yet due to the outstanding lint failures and because the commit to rename the files has not yet been factored out. We can probably fold things down a bit to find a happy medium in review; e.g. `make_endpoint` seems unnecessary.

The examples-level README introduced here is also a good idea, and could be merged ahead of review completing for the code changes.

djc:
So let's get the README (leaving short comments in place) merged as a separate PR, and another separate PR to move the files to the new desired locations/names, and then revisit what's left here? That would make it easier to review.

TimonPost:
Opened up #908, I will close this and maybe revisit it.

PR #892, opened 2020-11-07
Packet reception stats

Matthias247:
This adds the same set of stats for received datagrams
that have been merged previously for transmitted datagrams.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -500,8 +500,8 @@ where
         trace!("sending {} byte datagram", buf.len());
         self.total_sent = self.total_sent.wrapping_add(buf.len() as u64);
 
-        self.stats.packet_tx.packets += 1;
-        self.stats.packet_tx.bytes += buf.len() as u64;
+        self.stats.udp_tx.datagrams += 1;
+        self.stats.udp_tx.bytes += buf.len() as u64;
 
         Some(Transmit {
             destination: self.path.remote,
@@ -703,10 +703,13 @@ where
                     return;
                 }
 
+                self.stats.udp_rx.datagrams += 1;
+                self.stats.udp_rx.bytes += first_decode.len() as u64;
                 self.total_recvd = self.total_recvd.wrapping_add(first_decode.len() as u64);
 
                 self.handle_decode(now, remote, ecn, first_decode);
                 if let Some(data) = remaining {
+                    self.stats.udp_rx.bytes += data.len() as u64;
                     self.handle_coalesced(now, remote, ecn, data);
                 }
             }
--- a/quinn-proto/src/connection/stats.rs
+++ b/quinn-proto/src/connection/stats.rs
@@ -1,24 +1,15 @@
-/// Statistics about packets transmitted on a connection
-#[derive(Default)]
+/// Statistics about UDP datagrams transmitted or received on a connection
+#[derive(Default, Debug, Copy, Clone)]
 #[non_exhaustive]
-pub struct PacketTransmissionStats {
-    /// The amount of packets transmitted on a connection
-    pub packets: u64,
-    /// The total amount of bytes transmitted on a connection
+pub struct UdpStats {
+    /// The amount of UDP datagrams observed
+    pub datagrams: u64,
+    /// The total amount of bytes which have been transferred inside UDP datagrams
     pub bytes: u64,
 }
 
-impl std::fmt::Debug for PacketTransmissionStats {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("PacketTransmissionStats")
-            .field("packets", &self.packets)
-            .field("bytes", &self.bytes)
-            .finish()
-    }
-}
-
 /// Statistics about frames transmitted or received on a connection
-#[derive(Default)]
+#[derive(Default, Copy, Clone)]
 #[non_exhaustive]
 pub struct FrameStats {
     pub acks: u64,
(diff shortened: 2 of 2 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -706,10 +706,13 @@ where
                     return;
                 }
 
+                self.stats.packet_rx.packets += 1;
```
Matthias247, line 709: I now actually recognized the name `packet` is a bit ambiguous in the Quic context, due to the possibility of sending multiple Quic "packets" inside one UDP datagrams/packet. What this really counts is UDP datagrams.

But I didn't want to be inconsistent => Either all should be names packets, or get renamed to datagrams (or something else)

quinn-proto/src/connection/stats.rs
```diff
@@ -17,6 +17,25 @@ impl std::fmt::Debug for PacketTransmissionStats {
...
 
+/// Statistics about packets received on a connection
+#[derive(Default)]
+#[non_exhaustive]
+pub struct PacketReceptionStats {
+    /// The amount of packets transmitted on a connection
+    pub packets: u64,
+    /// The total amount of bytes transmitted on a connection
+    pub bytes: u64,
+}
```
Ralith, line 28: Should this and `PacketTransmissionStats` be the same struct, similar to the intended use of `FrameStats`?

quinn-proto/src/connection/stats.rs
```diff
@@ -17,6 +17,25 @@ impl std::fmt::Debug for PacketTransmissionStats {
...
+
+impl std::fmt::Debug for PacketReceptionStats {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("PacketReceptionStats")
+            .field("packets", &self.packets)
+            .field("bytes", &self.bytes)
+            .finish()
+    }
+}
```
Ralith, line 37: Does this differ from what you'd get from `derive`?

quinn-proto/src/connection/mod.rs
Ralith: Hmm, good catch. `datagrams` could also be confused with datagram frames. Would `udp_datagrams` be too verbose?

quinn-proto/src/connection/stats.rs
Matthias247: It could be. I guess a question is how one would deal with stats that might later be different between rx and tx. E.g. congestion/pacing/etc would only apply to tx. They could however be in a different struct.

If they are unified, I guess it would be `PacketStats` or `UdpDatagramStats`?

quinn-proto/src/connection/mod.rs
Matthias247: It's fine to me. But I think @djc might have stronger opinions on verbosity :-) 

quic datagram frames  would be counted as part of `FrameStats`.

quinn-proto/src/connection/stats.rs
Matthias247: It doesn't at at the moment. Guess I'm still looking for my calculated metrics 😥

quinn-proto/src/connection/stats.rs
Ralith: > E.g. congestion/pacing/etc would only apply to tx. They could however be in a different struct.

That sounds good to me.

> If they are unified, I guess it would be PacketStats or UdpDatagramStats?

How about `UdpStats`?

quinn-proto/src/connection/mod.rs
Ralith: > quic datagram frames would be counted as part of FrameStats.

Yeah, so in principle it's not ambiguous, I'm just trying to minimize risk of confusion. I think if we rename the struct to `UdpStats` and `udp_tx` then if we call it `datagrams` it might be sufficiently clear in practice, especially with a good doc comment.

quinn-proto/src/connection/stats.rs
Ralith: I think using a derived `Debug` impl and a separate, handwritten, pretty-print-with-derived-metrics function/method could be a fine approach. Deriving `Debug` is so easy that it's not really any extra complexity, and it makes it more natural to prototype or replace downstream.

Matthias247:
Looks now like this:

### Receiver

```
[quinn-proto/src/connection/mod.rs:171] &self.stats = ConnectionStats {
    udp_tx: UdpStats {
        datagrams: 687,
        bytes: 27076,
    },
    udp_rx: UdpStats {
        datagrams: 8977,
        bytes: 10891397,
    },
    frame_tx: FrameStats {
        ACK: 685,
        CRYPTO: 3,
        DATAGRAM: 0,
        HANDSHAKE_DONE: 1,
        MAX_DATA: 0,
        MAX_STREAM_DATA: 62,
        MAX_STREAMS_BIDI: 0,
        MAX_STREAMS_UNI: 10,
        NEW_CONNECTION_ID: 4,
        PATH_CHALLENGE: 0,
        PATH_RESPONSE: 0,
        PING: 0,
        RESET_STREAM: 0,
        RETIRE_CONNECTION_ID: 1,
        STOP_SENDING: 0,
        STREAM: 0,
    },
}
```

### Transmitter:

```
[quinn-proto/src/connection/mod.rs:171] &self.stats = ConnectionStats {
    udp_tx: UdpStats {
        datagrams: 8978,
        bytes: 10892597,
    },
    udp_rx: UdpStats {
        datagrams: 687,
        bytes: 27076,
    },
    frame_tx: FrameStats {
        ACK: 8978,
        CRYPTO: 2,
        DATAGRAM: 0,
        HANDSHAKE_DONE: 0,
        MAX_DATA: 0,
        MAX_STREAM_DATA: 0,
        MAX_STREAMS_BIDI: 0,
        MAX_STREAMS_UNI: 0,
        NEW_CONNECTION_ID: 5,
        PATH_CHALLENGE: 0,
        PATH_RESPONSE: 0,
        PING: 0,
        RESET_STREAM: 0,
        RETIRE_CONNECTION_ID: 0,
        STOP_SENDING: 0,
        STREAM: 8969,
    },
}
```

quinn-proto/src/connection/mod.rs
Matthias247: I went with that now. Sounds good to me.

Ralith (approved):
Thanks!

djc:
This looks great, thanks!

PR #893, opened 2020-11-08
Send only significant MAX_DATA frames

Matthias247:
This change equals #880, but targets the connection window which is
updated via `MAX_DATA` frames instead of the stream windows.

Similar as #880, frames will only be enqueued if the update is
significant enough.

Quinn had already before an optimization, which triggered MAX_DATA
frames not to be sent when the window was set to the maximum at
the start of the connection [1]. However the frame was still set as
`pending` [2], which triggered the connection logic to create a packet
for it [3]. Since the payload was suppressed, this lead the connection
to either send additional ACK-only packets or create empty packets.

This change fixes this behavior, and thereby impacts the amount
of packets and performance even if the maximum connection flow
control window is specified. Therefore the throughput for large streams
increases on windows from 30MB/s to 56MB/s with this change.

[1] https://github.com/quinn-rs/quinn/blob/bd14aa1dc2162e68c3a90027e3ca4e6404aea94d/quinn-proto/src/connection/streams.rs#L485
[2] https://github.com/quinn-rs/quinn/blob/bd14aa1dc2162e68c3a90027e3ca4e6404aea94d/quinn-proto/src/connection/mod.rs#L2789
[3] https://github.com/quinn-rs/quinn/blob/bd14aa1dc2162e68c3a90027e3ca4e6404aea94d/quinn-proto/src/connection/mod.rs#L2789

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -871,6 +871,9 @@ where
                 .stop_sending
                 .push(frame::StopSending { id, error_code });
         }
+        if self.streams.want_send_max_data() {
+            self.spaces[SpaceId::Data].pending.max_data = true;
+        }
         Ok(())
     }
 
@@ -2253,6 +2256,9 @@ where
                 }
                 Frame::Stream(frame) => {
                     self.streams.received(frame)?;
+                    if self.streams.want_send_max_data() {
+                        self.spaces[SpaceId::Data].pending.max_data = true;
+                    }
                 }
                 Frame::Ack(ack) => {
                     self.on_ack_received(now, SpaceId::Data, ack)?;
@@ -2813,7 +2819,9 @@ where
 
     fn add_read_credits(&mut self, id: StreamId, transmit_max_stream_data: bool) {
         let space = &mut self.spaces[SpaceId::Data];
-        space.pending.max_data = true;
+        if self.streams.want_send_max_data() {
+            space.pending.max_data = true;
+        }
         if transmit_max_stream_data {
             // Only bother issuing stream credit if the peer wants to send more
             space.pending.max_stream_data.insert(id);
--- a/quinn-proto/src/connection/streams.rs
+++ b/quinn-proto/src/connection/streams.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::{hash_map, HashMap, VecDeque},
+    convert::TryFrom,
     mem,
 };
 
@@ -52,8 +53,13 @@ pub struct Streams {
     connection_blocked: Vec<StreamId>,
     /// Connection-level flow control budget dictated by the peer
     max_data: u64,
-    /// Limit on incoming data
+    /// The initial receive window
+    receive_window: u64,
+    /// Limit on incoming data, which is transmitted through `MAX_DATA` frames
     local_max_data: u64,
+    /// The last value of `MAX_DATA` which had been queued for transmission in
+    /// an outgoing `MAX_DATA` frame
+    sent_max_data: VarInt,
     /// Sum of current offsets of all send streams.
     data_sent: u64,
     /// Sum of end offsets of all receive streams. Includes gaps, so it's an upper bound.
@@ -90,7 +96,9 @@ impl Streams {
             events: VecDeque::new(),
             connection_blocked: Vec::new(),
             max_data: 0,
+            receive_window: receive_window.into(),
             local_max_data: receive_window.into(),
+            sent_max_data: receive_window,
             data_sent: 0,
             data_recvd: 0,
             unacked_data: 0,
@@ -340,7 +348,7 @@ impl Streams {
             self.local_max_data = self
                 .local_max_data
                 .saturating_add(final_offset - bytes_read);
-            true
+            self.want_send_max_data()
         } else {
             false
         })
@@ -484,17 +492,18 @@ impl Streams {
         // MAX_DATA
         if pending.max_data && buf.len() + 9 < max_size {
             pending.max_data = false;
-            // We assume that if local_max_data is > VarInt::MAX, it must be due to an
-            // astronomically high configured initial credit, so we don't need to worry about the
-            // sender actually getting blocked somewhere between the initial credit and the current
-            // one.
-            if let Ok(max) = VarInt::from_u64(self.local_max_data) {
-                trace!(value = self.local_max_data, "MAX_DATA");
-                sent.max_data = true;
-                buf.write(frame::Type::MAX_DATA);
-                buf.write(max);
-                stats.max_data += 1;
-            }
+
+            // `local_max_data` can grow bigger than `VarInt`.
+            // For transmission inside QUIC frames we need to clamp it to the
+            // maximum allowed `VarInt` size.
+            let max = VarInt::try_from(self.local_max_data).unwrap_or(VarInt::MAX);
+
+            trace!(value = max.into_inner(), "MAX_DATA");
+            self.record_sent_max_data(max);
+            sent.max_data = true;
+            buf.write(frame::Type::MAX_DATA);
+            buf.write(max);
+            stats.max_data += 1;
         }
 
         // MAX_STREAM_DATA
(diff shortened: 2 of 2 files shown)


quinn-proto/src/connection/streams.rs
```diff
@@ -52,8 +52,13 @@ pub struct Streams {
     connection_blocked: Vec<StreamId>,
     /// Connection-level flow control budget dictated by the peer
     max_data: u64,
-    /// Limit on incoming data
+    /// The initial receive window
+    receive_window: u64,
```
Matthias247, line 56: This is now mainly required to be stored in order to know what is "significant".
In order to avoid it, we could pick a fixed size. Or we could likely go to u32 and save a few bytes if we store the minimum diff that is required to be transmitted. I'm good with any option.

quinn-proto/src/connection/mod.rs
```diff
@@ -2813,7 +2818,9 @@ where
 
     fn add_read_credits(&mut self, id: StreamId, transmit_max_stream_data: bool) {
         let space = &mut self.spaces[SpaceId::Data];
-        space.pending.max_data = true;
+        if self.streams.want_send_max_data() {
```
Matthias247, line 2821: What I'm not that happy with this approach is that it might be easy to miss the check before setting `pending`, and that not all paths which actually increase `local_max_data` might be covered. E.g. `stop` which is referenced higher up increased `max_data`, but did never set `pending`. And I think there is yet another path where we receive a `STREAM` frame for a stopped stream, increase `local_max_data` but don't set pending.

An alternative that I thought about is to encapsulate all the `self.local_max_data = self.local_max_data.saturating_add(buf.len() as u64);` into a `add_local_max_data(&mut self, increment: u64) -> WantTransmit` call, where `WantTransmit` is either a `bool` or a `#[must use]struct WantTransmit(bool)`, return that from every function that touches `local_max_data` and hope that the users then are reminded to touch `pending.max_data`.

But it's a bigger change - therefore I wanted to get some feedback first.

quinn-proto/src/connection/streams.rs
Ralith: This seems fine to me as-is. I wouldn't mind the u32 option but I think this has fewer edge cases.

quinn-proto/src/connection/streams.rs
```diff
@@ -490,6 +497,7 @@ impl Streams {
             // one.
             if let Ok(max) = VarInt::from_u64(self.local_max_data) {
```
Ralith, line 498: Should this be replaced with an `expect` now that we're being smart about setting `pending.max_data`?

quinn-proto/src/connection/mod.rs
Ralith: Good catches re: MAX_DATA not being sent. I confirm the remaining issue you describe on stopped streams. Refactoring to lampshade that sounds good, though I might name the method `add_read_credits` to match `Connection`.

quinn-proto/src/connection/mod.rs
```diff
@@ -2306,7 +2309,7 @@ where
                     self.streams.received_max_streams(dir, count)?;
                 }
                 Frame::ResetStream(frame) => {
-                    if self.streams.received_reset(frame)? {
+                    if self.streams.received_reset(frame)? && self.streams.want_send_max_data() {
```
Ralith, line 2312: I think either this check should be embedded in `received_reset`, or (perhaps preferably for consistency) `received_reset`'s `bool` return should be removed as redundant.

Ralith (changes requested):
Thanks!

quinn-proto/src/connection/mod.rs
Matthias247: lampshade? I will try the refactoring, and push if I like it better :)

quinn-proto/src/connection/streams.rs
Matthias247: I think that would work in this place.

However there's still something not 100% correct:
- `local_max_data` can grow beyond `VarInt::MAX`, but we don't allow `sent_max_data` to grow earlier in. That means we might leave the last `receive_window / 8` bytes unused. I guess we could either address this, or just don't care about it - because nobody will ever send that much data.

quinn-proto/src/connection/mod.rs
Matthias247: Yes, that can definitely be merged. As far as I understand the main intention of the boolean return value at the moment is to signal whether there is something to send - so moving it inside would be ok. The main counterargument is that all other checks at the moment are also outside on the connection level. But with the discussed refactoring that point would go away.

quinn-proto/src/connection/streams.rs
Ralith: We could clamp to `VarInt::MAX` rather than omitting. On review, I don't think `expect` would be appropriate here anyway, since `local_max_data` might increase after `pending.max_data` is set, particularly with retransmits.

quinn-proto/src/connection/mod.rs
Ralith: > lampshade?

Draw attention to.

quinn-proto/src/connection/streams.rs
Matthias247: Right - there exists a race! even without retransmits. Updated to use clamping

quinn-proto/src/connection/streams.rs
```diff
@@ -484,17 +491,18 @@ impl Streams {
...
-                sent.max_data = true;
-                buf.write(frame::Type::MAX_DATA);
-                buf.write(max);
-                stats.max_data += 1;
-            }
+
+            // `local_max_data` can grow bigger than `VarInt`.
+            // For transmission inside Quic frames we need to clamp it to the
+            // maximum allowed `VarInt` size.
+            let max = core::cmp::min(self.local_max_data, VarInt::MAX.into_inner());
```
Ralith, line 498: `let max = VarInt::try_from(self.local_max_data).unwrap_or(VarInt::MAX);` and replace `write_var` with `write` for stronger typing

Matthias247: Makes sense - I actually had an error first due to not changing the write call to `write_var`.

But if we immediately move to `VarInt`, we have to convert it back to `u64` for the `trace!` and `record_sent_max_data` (although I could change the latter to take `VarInt`) 🤷‍♂️

Guess I'm fine with either approach

Ralith: I'd lean towards making the change, including updating `record_sent_max_data`, since indeed we can never send an unrepresentable value. I've been trying to gradually move towards making more internal use of `VarInt` project-wide to reduce opportunity for error.

quinn-proto/src/connection/mod.rs
djc: > An alternative that I thought about is to encapsulate all the self.local_max_data = self.local_max_data.saturating_add(buf.len() as u64); into a add_local_max_data(&mut self, increment: u64) -> WantTransmit call, where WantTransmit is either a bool or a #[must use]struct WantTransmit(bool), return that from every function that touches local_max_data and hope that the users then are reminded to touch pending.max_data.

This definitely sounds good to me!

Matthias247: Will do. But also wouldn't mind get this version merged in between, since half of the changes would be the same - and the next change would then be pure refactoring without functional change.

Ralith (approved):
I'd be happy to see this merged in advance of any refactoring, since it's a clear improvement incremental regardless.

quinn-proto/src/connection/mod.rs
Ralith: > without functional change

We do still need to fix the issue with no updates being sent when receiving frames on a stopped stream.

quinn-proto/src/connection/streams.rs
```diff
@@ -809,6 +818,40 @@ impl Streams {
     fn flow_blocked(&self) -> bool {
         self.data_sent >= self.max_data || self.unacked_data >= self.send_window
     }
+
+    /// Returns whether a `MAX_DATA` frame should be emitted
+    ///
+    /// If `true`, a `MAX_DATA` frame should be enqueued as soon as possible.
+    /// This will be only be the case if the window update would be significant
```
djc, line 825: "be only be"

quinn-proto/src/connection/streams.rs
```diff
@@ -809,6 +818,40 @@ impl Streams {
...
+    /// enough. As soon as a window update with a `MAX_DATA` frame had been
```
djc, line 826: `had` -> `has`

quinn-proto/src/connection/streams.rs
```diff
@@ -809,6 +818,40 @@ impl Streams {
...
+    /// again.
+    pub fn want_send_max_data(&mut self) -> bool {
+        // We never need to send invalid windows
+        if self.local_max_data > VarInt::MAX.into_inner() {
+            return false;
+        }
+
+        // Only announce a window update if it's significant enough
+        // to make it worthwhile sending a MAX_DATA frame.
+        // We use here a fraction of the configured connection receive window to make
```
djc, line 838: "We use a fraction of the configured connection receive window to make the decision,
to accomodate for connections using bigger windows requiring fewer updates."

quinn-proto/src/connection/streams.rs
```diff
@@ -484,17 +492,18 @@ impl Streams {
...
-            if let Ok(max) = VarInt::from_u64(self.local_max_data) {
-                trace!(value = self.local_max_data, "MAX_DATA");
-                sent.max_data = true;
-                buf.write(frame::Type::MAX_DATA);
-                buf.write(max);
-                stats.max_data += 1;
-            }
+
+            // `local_max_data` can grow bigger than `VarInt`.
+            // For transmission inside Quic frames we need to clamp it to the
```
djc, line 497: `Quic` -> `QUIC`

djc (commented):
Just some typos and language nits.

quinn-proto/src/connection/mod.rs
djc: > We do still need to fix the issue with no updates being sent when receiving frames on a stopped stream.

Has this been addressed?

Matthias247: It now is added here too. It just initially wasn't, because it would have been addressed by the refactoring

Ralith (approved):

PR #894, opened 2020-11-09
Send only significant MAX_DATA frames (2)

Matthias247:
This is an update to #893 which uses a newtype to indicate whether an outgoing frame needs to be enqueued.
The update is in the second commit.

#893 could be merged before this - or this can be extended and replace the orginal CR.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -18,7 +18,7 @@ use crate::{
     config::{ServerConfig, TransportConfig},
     crypto::{self, HeaderKey, KeyPair, Keys, PacketKey},
     frame,
-    frame::{Close, Datagram, FrameStruct},
+    frame::{Close, Datagram, FrameStruct, ShouldTransmit},
     is_supported_version,
     packet::{Header, LongType, Packet, PacketNumber, PartialDecode, PartialEncode, SpaceId},
     range_set::RangeSet,
@@ -825,24 +825,18 @@ where
     ///
     /// The return value if `Ok` contains the bytes and their offset in the stream.
     pub fn read_unordered(&mut self, id: StreamId) -> Result<Option<(Bytes, u64)>, ReadError> {
-        Ok(self
-            .streams
-            .read_unordered(id)?
-            .map(|(buf, offset, transmit_max_stream_data)| {
-                self.add_read_credits(id, transmit_max_stream_data);
-                (buf, offset)
-            }))
+        Ok(self.streams.read_unordered(id)?.map(|result| {
+            self.add_read_credits(id, result.max_stream_data, result.max_data);
+            (result.buf, result.offset)
+        }))
     }
 
     /// Read from the given recv stream
     pub fn read(&mut self, id: StreamId, buf: &mut [u8]) -> Result<Option<usize>, ReadError> {
-        Ok(self
-            .streams
-            .read(id, buf)?
-            .map(|(len, transmit_max_stream_data)| {
-                self.add_read_credits(id, transmit_max_stream_data);
-                len
-            }))
+        Ok(self.streams.read(id, buf)?.map(|result| {
+            self.add_read_credits(id, result.max_stream_data, result.max_data);
+            result.len
+        }))
     }
 
     /// Send data on the given stream
@@ -867,14 +861,15 @@ where
             "only streams supporting incoming data may be stopped"
         );
         // Only bother if there's data we haven't received yet
-        if self.streams.stop(id)? {
+        let result = self.streams.stop(id)?;
+        if result.stop_sending.should_transmit() {
             let space = &mut self.spaces[SpaceId::Data];
             space
                 .pending
                 .stop_sending
                 .push(frame::StopSending { id, error_code });
         }
-        if self.streams.want_send_max_data() {
+        if result.max_data.should_transmit() {
             self.spaces[SpaceId::Data].pending.max_data = true;
         }
         Ok(())
@@ -2258,8 +2253,7 @@ where
                     self.read_crypto(SpaceId::Data, &frame)?;
                 }
                 Frame::Stream(frame) => {
-                    self.streams.received(frame)?;
-                    if self.streams.want_send_max_data() {
+                    if self.streams.received(frame)?.should_transmit() {
                         self.spaces[SpaceId::Data].pending.max_data = true;
                     }
                 }
@@ -2315,7 +2309,7 @@ where
                     self.streams.received_max_streams(dir, count)?;
                 }
                 Frame::ResetStream(frame) => {
-                    if self.streams.received_reset(frame)? {
+                    if self.streams.received_reset(frame)?.should_transmit() {
                         self.spaces[SpaceId::Data].pending.max_data = true;
                     }
                 }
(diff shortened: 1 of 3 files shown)


Matthias247:
Based on implementing this, I found out that `#[must_use]` part definitely pays off and is useful. It should likely also be adopted to other frames which need to be enqueued.

quinn-proto/src/connection/streams.rs
```diff
@@ -172,17 +180,17 @@ impl Streams {
         &mut self,
         id: StreamId,
         buf: &mut [u8],
-    ) -> Result<Option<(usize, bool)>, ReadError> {
+    ) -> Result<Option<(usize, bool, WantTransmit)>, ReadError> {
```
Matthias247, line 183: What I don't like here, is that if `MAX_STREAM_DATA` would change to the same thing, we would return `WantTransmit, WantTransmit`.

And thereby I get to what I don't think is perfect yet: The naming. 
Should it be a different newtype per frame-type? `WantTransmitMaxData` + `WantTransmitMaxStreamData`? At least makes it obvious in tuples what is what.

Another idea is making `WantTransmit` generic, and embedded the frame type, like `WantTransmit<MaxData>`. But that could be overengineered.

Another option is to simply use structs for these return values, instead of tuples.

Ralith: 3 items is my threshold for replacing structs with tuples, so I'm on board with that regardless. It might also be nice to disambiguate `WantTransmit`; a phantom type parameter using a frame type would be more pleasant than a bunch of functionally identical handwritten newtypes, for all that it's a bit cheeky. No strong opinion on whether that's really needed though.

Ralith (commented):
Seems generally reasonable. Rebase needed.

djc:
I like the idea of a phantom type parameter for `WantTransmit`.

quinn-proto/src/connection/streams.rs
Matthias247: I now added a dedicated result structs for all those operations

Matthias247:
I tried the `PhantomType` approach, but it didn't work out. The reason is that there don't exist dedicated frame types for each those frames. Most are only enum variants within `frame::Frame`, and can't be used as type parameters. The pure `frame::Type::MAX_DATA` version also didn't work out - I guess it needs const-generics to make it work. Therefore this now has a single `WantTransmit` for everything.

However I updated a few more paths to make use of it.
In a few places I also replaced tuples with result structs to at least give the returned fields some names. I however feel like this has a negative impact regarding `#[must_use]`. I don't think `must_use` fields within structs are checked to be used, as long as the struct itself is used.

quinn-proto/src/connection/streams.rs
```diff
@@ -416,18 +429,23 @@ impl Streams {
 
     /// Cease accepting data on a stream
     ///
-    /// Returns `true` iff the peer should be notified.
-    pub fn stop(&mut self, id: StreamId) -> Result<bool, UnknownStream> {
+    /// Returns structure which indicates whether this requires transmissing
```
djc, line 432: `transmissing` is not a word, I think, and "whether this requires transmitting and additional frames" also seems wrong? Maybe "and" -> "any"?

quinn-proto/src/frame.rs
```diff
@@ -826,6 +826,29 @@ impl Datagram {
     }
 }
 
+/// Indicates whether a frame of type `F` needs to be transmitted
```
djc, line 829: This documentation seems outdated, type `F` is no longer there.

quinn-proto/src/frame.rs
```diff
@@ -826,6 +826,29 @@ impl Datagram {
...
+///
+/// This type is around `bool` which uses the `#[must_use]` attribute in order
```
djc, line 831: This sentence reads a bit odd to me, maybe "This type wraps around `bool` and uses the ..."?

quinn-proto/src/frame.rs
```diff
@@ -826,6 +826,29 @@ impl Datagram {
...
+    /// of type `F` should be transmitted
+    pub fn new(want_transmit: bool) -> Self {
+        Self { want_transmit }
+    }
+
+    /// Returns whether a frame ot type `F` should be transmitted
+    pub fn want_transmit(self) -> bool {
+        self.want_transmit
+    }
+}
```
djc, line 850: These methods don't feel like they carrry their weight at this point (perhaps as an effect of losing the type argument?).

quinn-proto/src/connection/mod.rs
```diff
@@ -2820,12 +2822,17 @@ where
         self.streams.alloc_remote_stream(&self.peer_params, dir);
     }
 
-    fn add_read_credits(&mut self, id: StreamId, transmit_max_stream_data: bool) {
+    fn add_read_credits(
+        &mut self,
+        id: StreamId,
+        transmit_max_stream_data: frame::WantTransmit,
```
djc, line 2828: Let's just import `WantTransmit` from `frame` directly.

Also, I feel like `ShouldTransmit` better matches the semantics than `WantTransmit`, do others feel the same?

djc (commented):
Looks useful, but maybe a bit overcomplicated as is?

Ralith:
> there don't exist dedicated frame types for each those frames

I wouldn't mind having a type for every frame for consistency, but I think that's getting pretty scope-creepy for this PR.

> I don't think must_use fields within structs are checked to be used, as long as the struct itself is used.

This is an issue we've hit before, but I'm pretty sure it's true for tuples as well.

quinn-proto/src/connection/mod.rs
Ralith: I'm weakly in favor.

quinn-proto/src/frame.rs
Ralith: Maybe a one-public-element tuple struct? It's not like this can conceivably grow any internal logic in the future.

Or maybe an `enum`? That's a little more idiomatic, I think, but I'm not sure what good variant names would be.

Matthias247: If you don't have them, you end up with checks which look like `if transmit_max_data.0 {`

I'm personally not a fan of those `.0` accesses. But I agree the duplication between method and type names isn't elegant. If someone has ideas around better names, I'm all open for it. It could be as simple as `fn yes(&self)`, but that looks weird too.

Ralith: Maybe `if max_data.transmit()`?

Matthias247: That looks good. But it's kind of debatable whether to name the field in `ReadResult` `max_data`, since it doesn't contain the `max_data` value. If we get to `result.should_transmit_max_data.should_transmit()` we have the doubling again.`  `result.transmit_max_data.transmit()` is kind of the same

djc: I think `result.max_data.should_transmit()` reads exactly right -- it doesn't seem like a big concern that the `bool`-like value of `result.max_data` is not a `max_data` value, since that should be obvious from the type system.

Matthias247:
Clippy error is unrelated to these changes. I guess it's a 1.48 release artifact

djc:
> Clippy error is unrelated to these changes. I guess it's a 1.48 release artifact

Addressed those in #923.

quinn-proto/src/connection/streams.rs
```diff
@@ -288,21 +295,27 @@ impl Streams {
...
         // Stopped streams become closed instantly on FIN, so check whether we need to clean up
         if rs.is_closed() {
             self.recv.remove(&stream);
         }
-        Ok(())
+
+        // We don't buffer data on stopped streams, so issue flow control credit immediately
+        let want_transmit = self.add_read_credits(new_bytes);
+
+        Ok(want_transmit)
```
djc, line 309: Nit: let's inline the call to `self.add_read_credits()` in `Ok()`?

quinn-proto/src/connection/streams.rs
```diff
@@ -416,18 +426,23 @@ impl Streams {
...
-    pub fn stop(&mut self, id: StreamId) -> Result<bool, UnknownStream> {
+    /// Returns a structure which indicates whether this action
+    /// requires transmitting any frames.
+    pub fn stop(&mut self, id: StreamId) -> Result<StopResult, UnknownStream> {
         let stream = match self.recv.get_mut(&id) {
             Some(s) => s,
             None => return Err(UnknownStream { _private: () }),
         };
         stream.assembler.stop();
+        let transmit_stop_sending = ShouldTransmit::new(!stream.is_finished());
```
djc, line 437: Nit: let's call this `stop_sending` and `transmit_max_data` -> `max_data` so that the initialization in the return value is more streamlined.

djc (commented):
This is looking great, almost done!

djc (approved):

djc:
Thanks!

Issue #896, opened 2020-11-09
Book build fails


djc:
@TimonPost can you take a look at this?

https://github.com/quinn-rs/quinn/runs/1376738836?check_suite_focus=true

PR #898, opened 2020-11-10
Fix book ci

TimonPost:
Fixes: #896

- Add the path to mdbook build
- Move book into sub-directory in docs
- Change the publish dir.
- Manual triggerable CI

@djc, you can manually trigger this CI. I tested it in my own repo and it deployed the build files to `gh-pages`. 
One thing to note is that for me it took about an hour for the book to be online.


.idea/workspace.xml
```diff
@@ -0,0 +1,142 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
djc, line 1: Let's not merge this.

.github/workflows/book.yml
```diff
@@ -1,8 +1,9 @@
-name: Build book
+name: Build Book
```
djc, line 1: Let's keep this as is.

.github/workflows/book.yml
```diff
@@ -1,8 +1,9 @@
...
 
 on:
   push:
     branches: ['main']
+  workflow_dispatch:
```
djc, line 6: What is this for?

docs/book/src/quinn.md
```diff
@@ -0,0 +1 @@
+{{#include ../../../README.md}}
```
djc, line 1: Can you do this as a rename?

TimonPost: What do you mean by that?

.github/workflows/book.yml
TimonPost: Its to enable manual triggering. I needed it for testing and I think it can be useful.

djc: Okay!

.idea/workspace.xml
TimonPost: Somehow my global git ignore doesn't ignore this file

djc: (If you like, we can merge a change to the .gitignore to prevent this.)

docs/book/src/quinn.md
djc: This PR removes and re-adds `quinn.md`. It would be better if Git understood it as a rename.

TimonPost: mmm. I moved it to a subfolder. First, all contents were in `./docs` now in `./docs/book`. It is strange it did not account for this move as a move. Let me check interwebs if there is a workaround to trick git

TimonPost: Ah I see https://stackoverflow.com/questions/433111/how-to-make-git-mark-a-deleted-and-a-new-file-as-a-file-move. 

The reason it does this is that I move and 'change' the file. It needs this change because the README is now one folder higher then before.

TimonPost: I think we should go with it for now. Better now than later.

djc (approved):

Issue #865, opened 2020-10-12
Documentation - MdBook tracking issue


TimonPost:

Ralith:
A couple quick notes:

> easy-rsa for a DIY certificate authority

This should be dropped in favor of rcgen.

> Streams - Application datagrams

We shouldn't describe these as a type of stream, though we should definitely discuss them and compare/contrast to unidirectional streams.

TimonPost:
What would you propose as a topic name instead of streams?

djc:
Maybe something like "Data transfer"?

(Note that your capitalization is pretty inconsistent here. I would prefer that we consistently just use "Single capital letter per title" rather than "Capitals For Every Word".)

TimonPost:
Good point to keep in mind when realizing, those titles are still a draft and are mostly the copy of @Ralith proposal of the other issue. I'lll fix it directly to make it more representative to how the book will look like.

TimonPost:
Locally I also have a branch ready for the Networking Introduction. I will be working on the Quinn section now. When this PR quinn-rs/quinn#868 is merged I will create a pull request for the Network Introduction chapter.

PR #901, opened 2020-11-11
(1) Networking Introduction Chapter, Introduction Page

TimonPost:
This PR is part of a series of PR's regarding the book creation as described in #865. 

## Chapter Motivation

The `Networking Introduction` chapter introduces people who are new in networking to some common networking concepts. Before jumping directly into the meat of QUIC, it can be useful to understand its underlying motivations. 
For those motivations, we have to inspect the flaws of TCP and the nature of UDP 
because QUIC tries to improve the flaws of TCP on top of UDP.

## Page Motivation

The page introduces a person to this chapter and the motivations for this chapter. 

## Remarks

- Summary will be part of all PR's in this series such that if one PR is merged it will be shown in the book, If you have any comments on the summary then I think it would be better to handle those in a separate PR after this series is merged.  Otherwise, I will have to update 4 PR's. The same goes for file names, chapter names. 

## Live 

A live version of this page can be found [here](https://timonpost.github.io/quinn/). This book shows all the work that I have up-to-now. The pages that are visible but don't have a PR yet should not be judged by heavy judgment however you can provide feedback or share ideas in #865.

### DIFF
--- a/docs/book/src/SUMMARY.md
+++ b/docs/book/src/SUMMARY.md
@@ -1,4 +1,8 @@
 # Summary
 
-- [Quinn introduction](quinn.md)
-- [The QUIC protocol](quic.md)
+- [Network Introduction](network-introduction.md)
+    - [Transport Guarantees](network-introduction/transport-guarantees.md)
+    - [Transport Protocols](network-introduction/transport-protocols.md)
+    - [Problems of TCP](network-introduction/tcp-problems.md)
+- [Quinn introduction](quinn.md)	- [Quinn introduction](quinn.md)
+- [The QUIC protocol](quic.md)	- [The QUIC protocol](quic.md)
--- a/docs/book/src/network-introduction.md
+++ b/docs/book/src/network-introduction.md
@@ -9,8 +9,8 @@ However, this does not mean that they are perfect or flawless.
 One of the reasons to build QUIC is to solve those problems.  
 
 Before jumping directly into the meat of QUIC, it can be useful to understand its underlying motivations. 
-For those motivations we have to inspect the flaws of TCP and nature of UDP, 
-because QUIC tries to improve the flaws of TCP ontop of UDP.
+For those motivations, we have to inspect the flaws of TCP and the nature of UDP, 
+because QUIC tries to improve the flaws of TCP on top of UDP.
 
 1. [Transport Guarantees](network-introduction/transport-guarantees.md) (what guarantees for transmission exists)
 2. [Transport Protocols](network-introduction/transport-protocols.md) (analyse of UDP and TCP internals)
--- a/docs/book/src/network-introduction.md
+++ b/docs/book/src/network-introduction.md
@@ -1,0 +1,18 @@
+# Networking Introduction
+
+The Internet is unreliable, it changes every second, cables can be cut, network congestion can occur, packets can become corrupted, etc. 
+As soon as we send a packet, it can take any road to reach its destination. 
+To assure the end-user of certain transmission guarantees, such as the arrival of packets, specific protocols are defined.  
+
+Some of those established protocols are TCP and UDP which are supported by all routers, firewalls, servers, and operating systems. 
+However, this does not mean that they are perfect or flawless.
+One of the reasons to build QUIC is to solve those problems.  
+
+Before jumping directly into the meat of QUIC, it can be useful to understand its underlying motivations. 
+For those motivations we have to inspect the flaws of TCP and nature of UDP, 
+because QUIC tries to improve the flaws of TCP ontop of UDP.
+
+1. [Transport Guarantees](network-introduction/transport-guarantees.md) (what guarantees for transmission exists)
+2. [Transport Protocols](network-introduction/transport-protocols.md) (analyse of UDP and TCP internals)
+3. [TCP Problems](network-introduction/tcp-problems.md) (description of TCP its problems)
+ 


djc:
Who do you think the audience is for this chapter? I feel like aiming for an introduction to networking concepts in general does not make a whole lot of sense, and we should assume a certain familiarity on the part of the reader. (And we can state our assumptions up front for readers to verify that the documentation is the right fit for them.)

While I think it makes sense to have a short section introducing QUIC and quickly comparing QUIC to UDP and TCP, I don't think it makes sense for us to document substantial amounts of text on topics that aren't the specific implementation of QUIC in Quinn (or, perhaps, implementation issues related to implementing QUIC in Rust).

So let's start by figuring out the intended audience, and what we think they should already know before digging into our documentation, vs what we're in the best position to teach.

TimonPost:
Before going into your comments, do you mean the contents of this PR or #902, #903 #904 as well?  Since you mentioned 'chapter' ill be assuming you meant the whole chapter and not this page. For this page, I think we can indeed keep it very short or even remove this whole page all together keeping the other pages. 

> Who do you think the audience is for this chapter?

> I feel like aiming for an introduction to networking concepts, in general, does not make a whole lot of sense, and we should assume a certain familiarity on the part of the reader.

**Audience**

For this chapter, I focus on people who are not familiar with networking or who are new to this area in general. You have a point that we should avoid writing extended introductory material that can be found all over the Internet. Because this chapter is focused on 'inexperienced', I first had a disclaimer, but soon I realized that people with experience would probably skip this chapter anyways regardless of me telling them to skip it. 

_When I was working on laminar I came across so many people asking the same questions over and over again. Like: What is laminar compared to TCP? Why build up your own layer of reliability? Or "TCP is always superior, no reason to build your own protocol".  Also, people asking how it compared to QUIC, I even remember some discussions between the amethyst team and Quinn-team where there were some misconceptions between the two teams on some very core principles described in #902 etc.._ 

Based on that background I thought it would be good, to spend a short amount of the book on this subject. I see lots of confusion on the internet surrounding protocol selection and I think it would make sense to 'shortly' depict the motivations for QUIC its existence. 

**Considerable amount of text**

> And we can state our assumptions upfront for readers to verify that the documentation is the right fit for them.

>  I don't think it makes sense for us to document substantial amounts of text on topics that aren't the specific implementation of QUIC in Quinn

> While I think it makes sense to have a short section introducing QUIC and quickly comparing QUIC to UDP and TCP.

In order to respond to this, I wonder what you think is too general and what texts are too substantial? 
I tried to keep it only to the information that is needed in later stages in the book and keep it very compact. I don't see how it is 'substantial'. Therefore I like to ask you what you think is not needed to be explained?

In my view:

**PR: #902**
- Maybe we can drop: '[Ordering VS Sequencing](https://timonpost.github.io/quinn/network-introduction/transport-guarantees.html#ordering-vs-sequencing)' in since it is not very applicable to Quinn, though the concept of order is applicable but also self-explanatory. 

**PR: #903** 
- Maybe we can drop ['IP - Internet Layer](https://timonpost.github.io/quinn/network-introduction/transport-protocols.html#ip---internet-layer)' and keep the '[TCP/IP and UDP Comparison - Transport layer](https://timonpost.github.io/quinn/network-introduction/transport-protocols.html#tcpip-and-udp-comparison---transport-layer)' because it is a 'very' short compiled comparison. 

**PR: #904 **
- Problems of TCP (#904 ) is not general and should be explained. You spent a lot of time in your talk about it therefore I thought it would be good to mention the same issues here.

djc:
Yeah, this comment was targeted at the entire chapter. In a sense, while I value the idea behind splitting up the review into multiple pieces, I have a feeling this way of splitting it does not make reviewing easier -- in documentation like this, we can't really judge different parts without considering what part of the whole they try to do. So maybe I would advocate merging this chapter into a single PR.

Second, I find the writing style in this section not a great fit for technical documentation. It seems to take a very wide view, saying things like "The Internet is unreliable, it changes every second, cables can be cut, network congestion can occur, packets can become corrupted, etc.", whereas the topic at hand is substantially more narrow (for example, has nothing to do with cables that can be cut). It also seems more opinionated than I feel is warranted, for example, I would never write something like "the flaws of TCP" but rather talk about trade-offs or limitations; I also feel the tone here is more familiar than I am comfortable with ("That's a good question to ask yourself.").

Third, there are still some spelling errors or typos in this short piece of text. I recommend that you find some way to check spelling and ideally also some basic grammar checking so that we don't have to spend review cycles on things that the computers can solve for us.

As for the remaining sections:

- The concept of "sequencing" vs "ordering" is not something that I think is a generally accepted concept -- I'm pretty sure it didn't come up in my university networking class. It would make sense to briefly talk about ordering and reliability there. I don't think RUDP, netcode or laminar are consequential enough to be mentioned in our documentation, though.
- I do think we should not spend much text on IP, other than what services it provides to UDP and TCP.

For the audience, my sense would be that the audience here is software developers who know that UDP and TCP exist, who vaguely know that TCP is reliable and in-order while UDP is not, but not much else.

TimonPost:
Gotcha. Let's drop this introduction page. I can agree it might be too general and is not that necessary. I will reduce it to a single page with the following contents:

- Limitations of TCP
- Comparison UDP and TCP without the whole layers' stuff. 

> Third, there are still some spelling errors or typos in this short piece of text.

I can not find any. I ran this introduction about 10 times through Grammarly, translated it from Dutch to English, and in reverse for about 10 times to check correct spelling. There is no single mistake in here from what I can tell.

PR #902, opened 2020-11-11
(1) Networking Introduction Chapter, Transport Guarantees Page

TimonPost:
This PR is part of a series of PR's regarding the book creation as described in #865. 

## Chapter Motivation

The `Networking Introduction` chapter introduces people who are new in networking to some common networking concepts. Before jumping directly into the meat of QUIC, it can be useful to understand its underlying motivations. 
For those motivations, we have to inspect the flaws of TCP and the nature of UDP 
because QUIC tries to improve the flaws of TCP on top of UDP.

## Page Motivation

The page describes transport guarantees and lists different qualities of those guarantees. Different protocols define different guarantees. Before going into one protocol versus another we should shortly touch upon what those guarantees mean. The reader will understand that different protocols serve different use cases and have their pro's and con's.

## Remarks

- Summary will be part of all PR's in this series such that if one PR is merged it will be shown in the book, If you have any comments on the summary then I think it would be better to handle those in a separate PR after this series is merged.  Otherwise, I will have to update 4 PR's. The same goes for file names, chapter names. 

## Live 

A live version of this page can be found [here](https://timonpost.github.io/quinn/). This book shows all the work that I have up-to-now. The pages that are visible but don't have a PR yet should not be judged by heavy judgment however you can provide feedback or share ideas in #865.

### DIFF
--- a/docs/book/src/SUMMARY.md
+++ b/docs/book/src/SUMMARY.md
@@ -1,4 +1,8 @@
 # Summary
 
-- [Quinn introduction](quinn.md)
-- [The QUIC protocol](quic.md)
+- [Network Introduction](network-introduction.md)
+    - [Transport Guarantees](network-introduction/transport-guarantees.md)
+    - [Transport Protocols](network-introduction/transport-protocols.md)
+    - [Problems of TCP](network-introduction/tcp-problems.md)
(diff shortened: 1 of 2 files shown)


TimonPost:
Closing as discussed in #901.

PR #906, opened 2020-11-12
Add README note about address families

djc:
Just one proposal to prevent issues like #77 (which I think we've seen a few times). Bikeshedding on the most lucent and concise way to explain this welcome. (Also, do we need something like this in other places, maybe the example's docstrings?)

### DIFF
--- a/quinn/examples/client.rs
+++ b/quinn/examples/client.rs
@@ -104,6 +104,7 @@ async fn run(options: Opt) -> Result<()> {
         .map_or_else(|| url.host_str(), |x| Some(&x))
         .ok_or_else(|| anyhow!("no hostname specified"))?;
 
+    eprintln!("connecting to {} at {}", host, remote);
     let new_conn = endpoint
         .connect(&remote, &host)?
         .await
--- a/quinn/examples/server.rs
+++ b/quinn/examples/server.rs
@@ -126,7 +126,7 @@ async fn run(options: Opt) -> Result<()> {
     }
 
     let (endpoint, mut incoming) = endpoint.bind(&options.listen)?;
-    info!("listening on {}", endpoint.local_addr()?);
+    eprintln!("listening on {}", endpoint.local_addr()?);
     drop(endpoint);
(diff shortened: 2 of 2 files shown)


TimonPost:
Can we perhaps put this in the readme of #891 ?

djc:
From #908, @Ralith said:

> I'm not sure about this wording. We should note that it listens on IPv6 by default, and I don't think there's anything the client can easily do to change the address family, since a text name is needed for certificate verification.

Let's discuss this change in this PR, I don't want this to muddy up @TimonPost's work.

So is there something else we can suggest? By "listens on IPv6 by default", do you mean that it listens on IPv6, then falls back on IPv4 if IPv6 works? Can we make the server echo back the IP address it's listening on? I guess that might be tricky for dual-stack sockets, which is where we have this problem in the first place? Might be better to make this more obvious in the command's output rather than trying to explain it in documentation context.

Ralith:
> By "listens on IPv6 by default", do you mean that it listens on IPv6, then falls back on IPv4 if IPv6 works?

I mean the behavior if you don't specify otherwise is to unconditionally listen on IPv6. The listen address is something the user can control much more easily than the address that `localhost` resolves to, so that might be a better thing to focus attention on.

djc:
I really liked this idea:

> Maybe we should log the IP we resolve and try to connect to?

So I implemented it:

```
djc-2019 djc-patch-1 quinn $ cargo run --example server ./
     Running `target/debug/examples/server ./`
listening on [::1]:4433
djc-2019 djc-patch-1 quinn $ cargo run --example client https://localhost:4433/Cargo.toml
   Compiling quinn v0.6.1 (/Users/djc/src/quinn/quinn)
    Finished dev [unoptimized + debuginfo] target(s) in 5.51s
     Running `target/debug/examples/client 'https://localhost:4433/Cargo.toml'`
connecting to localhost at [::1]:4433
```

Ralith (approved):

PR #908, opened 2020-11-12
Example README

TimonPost:
- Created example README in separate PR as discussed in #891 
- Remove main README example section because it is not needed with this new example README. 
-  Add the 'note', @djc opened up a PR for in #906, in this README.
- Update logo link (as discussed in gitter, it doesn't show in a book because of the relative link), just took it with this PR


quinn/examples/README.md
```diff
@@ -0,0 +1,83 @@
+## HTTPS File Serving Example
+
+The `https` example shows how to request a file on the `server` from the `client` over HTTPS. 
+This is HTTP/0.9, which is a toy protocol that's much simpler than real HTTP, especially HTTP/3.
+
+(Note: when providing the example server with a specific address to listen on,
+make sure you match the correct IPv4 vs IPv6 address in the client command.)
```
Ralith, line 7: I'm not sure about this wording. We should note that it listens on IPv6 by default, and I don't think there's anything the client can easily do to change the address family, since a text name is needed for certificate verification.

TimonPost: forward to @djc?

djc:
Okay, I still want to split this into smaller pieces that we can discuss separately.

- One PR or commit fixing the link to the image
- One PR or commit moving the README documentation into examples/README (review comment on this: let's keep the first line of each docstring that's there now, then an empty line, then the pointer to the README).
- One PR or commit changing the top-level README (personally I don't like the idea of removing the example commands from the top-level README in favor of just a link to the examples README, but let's discuss it separately).

TimonPost:
Your wasting time now. I'll seperate them.

PR #910, opened 2020-11-13
Create example README

TimonPost:
Create a new example README, this should provide the user with a single entry point for the examples. As a new user of this library, I found it somewhat hard to get a quick idea of what each example its purpose was and how to navigate them quickly.

## Changes

- Create an example readme with detailed instructions. 
- Change module comments from example files to a redirection to this README.

### DIFF
--- a/quinn/examples/README.md
+++ b/quinn/examples/README.md
@@ -3,9 +3,6 @@
 The `https` example shows how to request a file on the `server` from the `client` over HTTPS. 
 This is HTTP/0.9, which is a toy protocol that's much simpler than real HTTP, especially HTTP/3.
 
-(Note: when providing the example server with a specific address to listen on,
-make sure you match the correct IPv4 vs IPv6 address in the client command.)
-
 1. HTTPS Server (`http_server.rs`)
 
 The server listens for any client requesting a file. 
--- a/quinn/examples/README.md
+++ b/quinn/examples/README.md
@@ -1,0 +1,83 @@
+                                                                            ## HTTPS File Serving Example
+                                                                            
+                                                                            The `https` example shows how to request a file on the `server` from the `client` over HTTPS. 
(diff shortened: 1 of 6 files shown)


quinn/examples/connection.rs
```diff
@@ -1,17 +1,6 @@
-//! This example intends to use the smallest amount of code to make a simple QUIC connection.
+//! Checkout the `README.md` for guidance.
```
djc, line 1: I meant this:

```suggestion
//! This example intends to use the smallest amount of code to make a simple QUIC connection.
//!
//! Checkout the `README.md` for guidance.
```

(And skip the link; same for the other instances here.)

quinn/examples/README.md
```diff
@@ -0,0 +1,83 @@
+## HTTPS File Serving Example
+
+The `https` example shows how to request a file on the `server` from the `client` over HTTPS. 
+This is HTTP/0.9, which is a toy protocol that's much simpler than real HTTP, especially HTTP/3.
+
+(Note: when providing the example server with a specific address to listen on,
+make sure you match the correct IPv4 vs IPv6 address in the client command.)
```
djc, line 7: So @Ralith said in https://github.com/quinn-rs/quinn/pull/908#discussion_r522644091:

> I'm not sure about this wording. We should note that it listens on IPv6 by default, and I don't think there's anything the client can easily do to change the address family, since a text name is needed for certificate verification.

The easy way out for this PR is just to leave this out so we can discuss it separately -- it doesn't really match the goal of merging the example READMEs anyway.

djc (commented):
This looks like a good improvement. I noticed some issues in the contents, but since those already existed before this change, we can leave them to a separate issue/PR.

quinn/examples/README.md
TimonPost: Good idea, removed.

djc (approved):
Also, good work tightening up the grammar in these one-line descriptions!

PR #917, opened 2020-11-14
Networking Introduction Book Chapter

TimonPost:
This PR is part of a series of PR's regarding the book creation as described in #865. 

## Chapter Motivation

The `Networking Introduction` chapter introduces people who are new in networking to some common networking concepts. Before jumping directly into the meat of QUIC, it can be useful to understand its underlying motivations. 
For those motivations, we have to inspect the complications of TCP and the nature of UDP 
because QUIC tries to improve the complications of TCP on top of UDP.

## Remarks

-  @djc I made a GIF of your animation making it able to be included in the page.
- I rewrote this chapter into one single page as discussed in #904. And dropped out several texts that were too obvious for most readers. 
 
## Live Version Viewable (recommendation) 

A live version of this page can be found [here](https://timonpost.github.io/quinn/). This book shows all the work that I have up-to-now. The pages that are visible but don't have a PR yet should not be judged by heavy judgment however you can provide feedback or share ideas in #865.

PR #919, opened 2020-11-17
(2) Quinn Introduction Chapter, Introduction Page

TimonPost:
This PR is part of a series of PR's regarding the book creation as described in #865. Each page of this chapter will have its own PR for reducing PR size.

- Network Introduction (#917)
- QUINN Introduction (#868)
    **- Overview (#919)**
    - Certificate Configuration (#920)
    - Connection Setup (#921)
    - Data Transfer (#922)

## Live Version

A live version of this page can be found [here](https://timonpost.github.io/quinn/). This book shows all the work that I have up-to-now. The pages that are visible but don't have a PR yet should are welcome to have general feedback.

## Chapter Motivation

This chapter will go through the protocol by examples. The chapters are in the required order, first, we look at configuring a certificate which is **required**, then at setting up a connection, and finally at sending data about this connection. In the future, more pages can be added.

After reading this chapter the reader should be able to **send data** over a secured with a real or self-signed **certificate**, or insecure, **connection**.

The target audience is people who are new to the library and want to get started quickly by some examples and a short explanation. It is not meant to describe the entire API. 

## Page Motivation

The page introduces a reader to the coming chapter very shortly.

## Remarks

### DIFF
--- a/docs/book/src/SUMMARY.md
+++ b/docs/book/src/SUMMARY.md
@@ -1,4 +1,5 @@
 # Summary
 
-- [Quinn introduction](quinn.md)
+- [QUINN Introduction](quinn.md)
+    - [Overview](quinn/introduction.md)
 - [The QUIC protocol](quic.md)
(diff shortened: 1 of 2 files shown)


djc:
This seems very short to put on a separate page -- maybe should just skip this and/or merge this into the next page?

TimonPost:
Perfect, I will merge this with the other page and do this for any following chapters. c

PR #925, opened 2020-11-24
Expose cid genearator customization API in endpoint builder

liwenjieQu:
A CID generator interface was implemented in transport layer https://github.com/quinn-rs/quinn/pull/851 but there is still no way to actually enable this feature when HTTP/3 endpoint is built. This PR provides an API in `EndpointBuilder` to allow customization on CID generator.

### DIFF
--- a/quinn/src/builders.rs
+++ b/quinn/src/builders.rs
@@ -105,7 +105,9 @@ where
     }
 
     /// Use a customized cid generator factory in the endpoint
-    pub fn connection_id_generator<F: Fn() -> Box<dyn ConnectionIdGenerator> + Send + Sync + 'static>(
+    pub fn connection_id_generator<
+        F: Fn() -> Box<dyn ConnectionIdGenerator> + Send + Sync + 'static,
+    >(
         &mut self,
         factory: F,
     ) -> &mut Self {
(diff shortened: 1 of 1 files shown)


quinn/src/builders.rs
```diff
@@ -100,6 +103,15 @@ where
         self.default_client_config = config;
         self
     }
+
+    /// Use a customized cid generator factory in the endpoint
+    pub fn use_customized_cid<F: Fn() -> Box<dyn ConnectionIdGenerator> + Send + Sync + 'static>(
```
Ralith, line 108:
```suggestion
    pub fn connection_id_generator<F: Fn() -> Box<dyn ConnectionIdGenerator> + Send + Sync + 'static>(
```
For builder methods, preferred naming is a noun specifying the property being modified.

Ralith (approved):
Good catch, thanks!

quinn/src/builders.rs
Ralith: (fixing review status)

Ralith (changes requested):

Ralith (approved):
Thanks!

liwenjieQu:
Thanks Benjamin!

djc:
Nice, thanks for adding this!

PR #926, opened 2020-11-26
proto: add more high-level API docs to Connection, closes #924

### LINKED ISSUE #924: quinn-proto API clarifications
Hi, I am looking at the quinn-proto API docs. [Endpoint](https://docs.rs/quinn-proto/0.6.1/quinn_proto/generic/struct.Endpoint.html) is quite simple but [Connection](https://docs.rs/quinn-proto/0.6.1/quinn_proto/generic/struct.Connection.html) is more complex, and so I have the following questions for clarification: 

- Can you clarify when `poll_endpoint_events` is supposed to be called? All the other `poll_*` methods have docs describing this.
- Where it says "the application performed some I/O on the connection" can you:
  - Please explicitly list out which of the other functions count as "I/O"? I am guessing anything that might change the state of the state machine, however it would be best to have this mentioned explicitly. For example this might be unclear for `reset`. I am guessing this may change some timeouts, but someone less familiar with networking might not realise this counts as "I/O" especially since the docs says "abandon sending" which sounds like the lack of I/O.
  - Having to call `poll_*` afterwards, this makes sense for commands that might want to write onto the network, since you are asking the state machine to do more things. However for read commands, this is less obvious. I'm consuming things that have already happened, why should anything timeout or transmit in response to this? At best, I am emptying a buffer inside the state, then this might mean that future incoming network packets can now trigger additional things to transmit (i.e. I can now ACK it because there's more space in the buffer), but as a direct response to the read command I'd intuitively expect this not to be the case. Can you clarify if the documentation should be changed to say "performed some write on the connection" instead of "I/O"?
- If `poll_timeout` returns 1000 then later 500, does that cancel the original timer at 1000? Or are both 500 and 1000 set?

infinity0:


quinn-proto/src/connection/mod.rs
```diff
@@ -54,10 +54,43 @@ use timer::{Timer, TimerTable};
...
+/// 3. [`poll_endpoint_events`](self::poll_endpoint_events)
+/// 4. [`poll`](self::poll)
+///
+/// Currently the only actual dependency is from (2) to (1), however additional
+/// dependencies may be added in future, so the above order is recommended.
+///
+/// (A) may be called whenever desired.
+///
+/// Care should be made to ensure that the input events represent monotonically
+/// increasing time. Additionally we recommended that, for any given [`Instant`],
```
djc, line 90: Nit: `recommended` -> `recommend`.

djc (dismissed):
This looks great, thanks! I'll let @Ralith review this in extra detail to make sure that it's fully correct.

quinn-proto/src/connection/mod.rs
```diff
@@ -728,6 +761,9 @@ where
     /// Executes protocol logic, potentially preparing signals (including application `Event`s,
     /// `EndpointEvent`s and outgoing datagrams) that should be extracted through the relevant
     /// methods.
+    ///
+    /// It is most efficient to call this only for each `Instant` output by `poll_timeout`;
+    /// however spurious extra calls will simply no-op and therefore are safe.
```
infinity0, line 766: Note: I should reword this slightly so that you can't misinterpret it to do something like `handle_timeout(poll_timeout())` which would be wrong.

quinn-proto/src/connection/mod.rs
```diff
@@ -54,10 +54,43 @@ use timer::{Timer, TimerTable};
...
+/// `Connection` has roughly 4 types of methods:
+///
+/// - A. Simple getters, taking `&self`
+/// - B. Handlers for incoming events from the network or system, named `handle_*`.
+/// - C. State machine mutators, for incoming commands from the application. For
+///   convenience we refer to this as "performing I/O" below, however as per
+///   the design of this library none of the functions actually perform
+///   system-level I/O. For example, [`read`](self::read) and [`write`](self::write),
+///   but also things like [`reset`](self::reset).
+/// - D. Polling functions for outgoing events or results, named `poll_*`.
```
Ralith, line 71: > outgoing events or results

Could we make this a bit less vague? Maybe "for events or actions the caller must take"?

quinn-proto/src/connection/mod.rs
```diff
@@ -54,10 +54,43 @@ use timer::{Timer, TimerTable};
...
+/// passing it to the application or by making a system-level I/O call. You
+/// should call the polling functions in this order:
+///
+/// 1. [`poll_transmit`](self::poll_transmit)
+/// 2. [`poll_timeout`](self::poll_timeout)
+/// 3. [`poll_endpoint_events`](self::poll_endpoint_events)
+/// 4. [`poll`](self::poll)
+///
+/// Currently the only actual dependency is from (2) to (1), however additional
+/// dependencies may be added in future, so the above order is recommended.
```
Ralith, line 85: Maybe we should hide all these behind a single function that returns a struct...

quinn-proto/src/connection/mod.rs
```diff
@@ -54,10 +54,43 @@ use timer::{Timer, TimerTable};
...
+///
+/// (A) may be called whenever desired.
+///
+/// Care should be made to ensure that the input events represent monotonically
+/// increasing time. Additionally we recommended that, for any given [`Instant`],
+/// you call [`handle_timeout`](self::handle_timeout) before you call
+/// [`handle_event`](self::handle_event) on any [`ConnectionEvent`]s having the same `Instant`.
+///
```
Ralith, line 93: Stray newline?

quinn-proto/src/connection/mod.rs
```diff
@@ -54,10 +54,43 @@ use timer::{Timer, TimerTable};
...
+/// Currently the only actual dependency is from (2) to (1), however additional
+/// dependencies may be added in future, so the above order is recommended.
+///
+/// (A) may be called whenever desired.
+///
+/// Care should be made to ensure that the input events represent monotonically
+/// increasing time. Additionally we recommended that, for any given [`Instant`],
+/// you call [`handle_timeout`](self::handle_timeout) before you call
+/// [`handle_event`](self::handle_event) on any [`ConnectionEvent`]s having the same `Instant`.
```
Ralith, line 92: What's the basis for this recommendation?

quinn-proto/src/connection/mod.rs
Ralith: Maybe s/for each `Instant`/after passing `Instant`s/?

Ralith (commented):
Thanks, this is good! Just a few notes.

quinn-proto/src/connection/mod.rs
djc: Yeah, I was thinking along similar lines.

quinn-proto/src/connection/mod.rs
infinity0: In case the logic for handling `Event(now = t)` ever implicitly depends on the timeout for `t` having already been fired, this ensures that this works consistently. (It would also be fine to say the timeout happens afterwards, as long as it's consistent.) It might not matter now and we could drop this, but I thought it'd be good to guard against future possibilities. It also gives something concrete to follow, in case the logic does need to be changed in the future, and the future developer doesn't know which way to go, but there's already different users assuming it differently.

infinity0:
I pushed some more separate commits so it's easier to review but LMK if you'd prefer them squashed.

quinn-proto/src/connection/mod.rs
infinity0: Done - I also noticed neqo uses `#[must_use]` for their [analogous functions](https://github.com/mozilla/neqo/blob/main/neqo-transport/src/connection/mod.rs) so I added that as well.

infinity0:
ping @Ralith - this is ready for merge, assuming you're satisfied with my replies above

quinn-proto/src/connection/mod.rs
Ralith: My inclination is to not specify requirements we don't have or expect to grow into, especially when it might be annoying for the caller to comply with them (e.g. requiring incoming events to be sorted). When there's no clear order between events, the logic should be robust to handling them in either order.

infinity0: OK, in that case I suggest stating as such in the documentation: i.e. "calling handle_event with events of same instant may be interleaved in any order with a call to handle_timeout at that same instant; however events or timeouts with different instants should not be interleaved", so callers know it's safe, and any future maintainers of quinn-proto don't change this behaviour?

What I've sometimes seen is people merging input blocking queues in a naive way, then this results in items being interleaved unpredictable and inappropriately at the consumer end; the latter "not be interleaved" requirement is supposed to expressly remind callers not to do that.

Ralith: I have no objection to a statement along those lines.

djc (approved):

djc:
Thanks!

PR #942, opened 2020-12-21
Update rand and directories-next

djc:


Ralith (approved):
Why the patch version bump on futures?

Issue #508, opened 2019-11-20
Provide local and remote socket addresses from a `Connecting`


Demi-Marie:
libp2p needs to know the local and remote socket addresses for every incoming connection as soon as the connection is accepted.  Is this a feature that you would be willing to include?  If so, I will submit a PR.

Ralith:
You can get the local address from `Endpoint::local_addr`, or just hold onto whatever you passed to `EndpointBuilder::bind`. We don't have plans to support multiple local addresses, so this should be sufficient, but let us know if it's not.

Exposing the remote address from `Connecting` makes perfect sense to me. I imagine many applications will want this even just for diagnostics. A PR would be very welcome!

Ralith:
Alternatively, since this doesn't make a lot of sense for outgoing connections, perhaps `<Incoming as Stream>::Item` could be made into a struct containing `Connecting` and associated metadata, but my gut feeling is that's less ergonomic for dubious benefit.

DemiMarie-parity:
Libp2p needs to know the destination IP address of the UDP datagram that initiated the connection.  This allows it to publish new addresses at which the local host can be reached.  This address is generally different from what was passed to `EndpointBuilder::bind` if a wildcard address was used for the latter.  On Linux, it can be obtained by setting the `IPV6_RECVPKTINFO` option for IPv6 sockets, or the `IP_PKTINFO` option for IPv4 sockets.

Ralith:
Ah, good point, I forgot about wildcard IPs. Exposing that makes sense, though it'll require a certain amount of piping, and perhaps attention to win32/macos/BSD compatibility.

DemiMarie-parity:
Windows is going to require us to handle sockets ourselves, just as we already do on \*nix.  I don’t have access to a Windows machine, so I am not particularly comfortable writing this code.  I am, however, willing to write an implementation for Linux.

Ralith:
If we're lucky it'll be exactly the same, but obscure cmsg stuff is pretty hit-or-miss portability wise. Linux would be a good start for us to build off of, at least!

DemiMarie-parity:
Here is what I have found:

| OS | Linux | FreeBSD | OpenBSD | NetBSD | macOS | Solaris (and presumably AIX, etc) | Windows |
| -- | ----- | ------- | ------- | ------ | ----- | ----------------- | ------- |
[...]

PR #943, opened 2020-12-21
Allow to lookup the initial local IP address

Matthias247:
When a listener is bound to multiple network interfaces (e.g. `::0`),
it is not obvious which IP the peer used to send a packet. We however
might need this information to send packets back to the peer with the
same source address.

This problem is described in #508.

This change makes the destination IP address which was used to send
the initial packet available in the `Conneting` and `Connection` types.

The information is far available only on Linux due to missing test on
other platforms.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -2,7 +2,7 @@ use std::{
     cmp,
     collections::{BTreeMap, HashSet, VecDeque},
     fmt, io, mem,
-    net::SocketAddr,
+    net::{IpAddr, SocketAddr},
     sync::Arc,
     time::{Duration, Instant},
 };
@@ -111,6 +111,10 @@ where
 
     /// cid length used to decode short packet
     local_cid_len: usize,
+    /// The "real" local IP address which was was used to receive the initial packet.
+    /// This is only populated for the server case, and if known
+    local_ip: Option<IpAddr>,
+
     path: PathData,
     prev_path: Option<PathData>,
     state: State,
@@ -208,6 +212,7 @@ where
         loc_cid: ConnectionId,
         rem_cid: ConnectionId,
         remote: SocketAddr,
+        local_ip: Option<IpAddr>,
         crypto: S,
         now: Instant,
         local_cid_len: usize,
@@ -244,6 +249,7 @@ where
                 config.congestion_controller_factory.build(now),
                 now,
             ),
+            local_ip,
             prev_path: None,
             side,
             state,
@@ -1060,6 +1066,24 @@ where
     /// The latest socket address for this connection's peer
     pub fn remote_address(&self) -> SocketAddr {
         self.path.remote
+    }
+
+    /// The local IP address which was used when the peer established
+    /// the connection
+    ///
+    /// This can be different from the address the endpoint is bound to, in case
+    /// the endpoint is bound to a wildcard address like `0.0.0.0` or `::`.
+    ///
+    /// This will return `None` for clients.
+    ///
+    /// Retrieving the local IP address is currently supported on the following
+    /// platforms:
+    /// - Linux
+    ///
+    /// On all non-supported platforms the local IP address will not be available,
+    /// and the method will return `None`.
+    pub fn local_ip(&self) -> Option<IpAddr> {
+        self.local_ip
     }
 
     /// Current best estimate of this connection's latency (round-trip-time)
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -2,7 +2,7 @@ use std::{
     collections::{HashMap, VecDeque},
     convert::TryFrom,
     fmt, iter,
-    net::SocketAddr,
+    net::{IpAddr, SocketAddr},
     ops::{Index, IndexMut},
     sync::Arc,
     time::{Duration, Instant, SystemTime},
@@ -154,6 +154,7 @@ where
         &mut self,
         now: Instant,
         remote: SocketAddr,
+        local_ip: Option<IpAddr>,
         ecn: Option<EcnCodepoint>,
         data: BytesMut,
     ) -> Option<(ConnectionHandle, DatagramEvent<S>)> {
@@ -273,7 +274,7 @@ where
             let crypto = S::initial_keys(&dst_cid, Side::Server);
             return match first_decode.finish(Some(&crypto.header.remote)) {
                 Ok(packet) => self
-                    .handle_first_packet(now, remote, ecn, packet, remaining, &crypto)
+                    .handle_first_packet(now, remote, local_ip, ecn, packet, remaining, &crypto)
                     .map(|(ch, conn)| (ch, DatagramEvent::NewConnection(conn))),
                 Err(e) => {
                     trace!("unable to decode initial packet: {}", e);
@@ -357,6 +358,7 @@ where
             remote_id,
             remote_id,
             remote,
+            None,
             ConnectionOpts::Client {
                 config,
                 server_name: server_name.into(),
@@ -399,6 +401,7 @@ where
         init_cid: ConnectionId,
         rem_cid: ConnectionId,
         remote: SocketAddr,
+        local_ip: Option<IpAddr>,
         opts: ConnectionOpts<S>,
         now: Instant,
     ) -> Result<(ConnectionHandle, Connection<S>), ConnectError> {
@@ -454,6 +457,7 @@ where
             loc_cid,
             rem_cid,
             remote,
+            local_ip,
             tls,
             now,
             self.local_cid_generator.cid_len(),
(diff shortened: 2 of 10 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -111,6 +111,10 @@ where
 
     /// cid length used to decode short packet
     local_cid_len: usize,
+    /// The "real" local IP address which was was used to receive the initial packet.
+    /// This is only populated for the server case, and if known
+    initial_local_ip: Option<IpAddr>,
```
Ralith, line 116: I'm not sure about `initial` here. Servers can't migrate, and if they did, we'd want to update the source address we send from. Maybe just `local_ip`, and note in the comment that it's the source address for outgoing UDP packets?

quinn-proto/src/endpoint.rs
```diff
@@ -154,6 +154,7 @@ where
         &mut self,
         now: Instant,
         remote: SocketAddr,
+        initial_local_ip: Option<IpAddr>,
```
Ralith, line 157:
```suggestion
        local_ip: Option<IpAddr>,
```
At this layer we have no notion of initial vs. not regardless.

quinn/src/connection.rs
```diff
@@ -127,6 +127,17 @@ where
                     .expect("spurious handshake data ready notification")
             })
     }
+
+    /// The local IP address which was used when the peer established
+    /// the connection
+    ///
+    // This is `None` if either not known or if this is the client side of a connection.
```
Ralith, line 134: Was it intended that the last line here not be a doc comment?

Ralith (changes requested):
Nice work! Just some cosmetic thoughts + the CI issue.

quinn-proto/src/connection/mod.rs
Matthias247: > Servers can't migrate
I don't actually remember what the spec says about this, since it's a bit of a weird use-case.

I called it `local_` since it's currently only updated on creation, and not on every packet. Doing the latter would at least be wrong for the use-case I have, where outgoing packets need to be tagged with this target IP. It would seem to allow the client to move the server to use different paths, if it can "guess" other server IPs. 

That said I also wasn't fully sure about the name - I just couldn't come up with a better one. Initially I just had `destination_ip`, but that is even more confusing since it it's not the address the server is sending to but it's own one.

quinn/src/connection.rs
Matthias247: No. Will fix

quinn-proto/src/connection/mod.rs
Ralith: I don't think we'd ever want to update it from incoming packets regardless; since server migration isn't supported it won't change at all at present, and if it became supported, we'd want to update the address explicitly when we migrate.

Matthias247: So you would just having `fn local_ip()`? And maybe document that this is only populated for servers when the connection is established? I can live with that too.

Ralith: Yeah, sounds sensible.

quinn/src/connection.rs
```diff
@@ -127,6 +127,21 @@ where
                     .expect("spurious handshake data ready notification")
             })
     }
+
+    /// The local IP address which was used when the peer established
+    /// the connection
+    ///
+    /// This can be different from the address the endpoint is bound to, in case
+    /// the endpoint is bound to multiple interfaces.
```
Ralith, line 135: Would this be clearer without sacrificing correctness? I'm not sure it makes sense to be bound to multiple interfaces with a specific address.
```suggestion
    /// the endpoint is bound to a wildcard address like `0.0.0.0` or `::`.
```

quinn/src/connection.rs
```diff
@@ -127,6 +127,21 @@ where
...
+    ///
+    /// This will return `None` for clients, as well for servers if capturing
+    /// the local IP from incoming packets is not supported.
```
Ralith, line 138:
```suggestion
    /// the local IP from incoming packets is not supported on the current platform.
```

Ralith (approved):
LGTM, thanks!

Could we add coverage for this in the `quinn`-level unit tests, perhaps?

quinn/src/connection.rs
Matthias247: Yes, I think this is fine. Updated

Matthias247:
> Could we add coverage for this in the quinn-level unit tests, perhaps?

I added some minimal coverage by reading the field and checking that it's available on Linux, and points to the loopback client. I think this covers the most important part (reading the cmsgs). Everything else is pretty much just plumbing.

Ralith (dismissed):

quinn/src/udp.rs
```diff
@@ -74,6 +74,8 @@ pub struct RecvMeta {
     pub addr: SocketAddr,
     pub len: usize,
     pub ecn: Option<EcnCodepoint>,
+    /// The destination IP address which was encoded in this datagram
+    pub dest_ip: Option<IpAddr>,
```
djc, line 78: Let's use `dst` instead of `dest` here, for consistency with similar naming elsewhere.

quinn/src/platform/unix.rs
```diff
@@ -283,36 +320,50 @@ fn decode_recv(
     len: usize,
 ) -> RecvMeta {
     let name = unsafe { name.assume_init() };
-    let ecn_bits = match unsafe { cmsg::Iter::new(&hdr).next() } {
-        Some(cmsg) => match (cmsg.cmsg_level, cmsg.cmsg_type) {
+    let mut ecn_bits = 0;
+    let mut dest_ip = None;
```
djc, line 324: Let's name this `dst_ip`, as well.

quinn-proto/src/connection/mod.rs
```diff
@@ -1062,6 +1068,18 @@ where
         self.path.remote
     }
 
+    /// The local IP address which was used when the peer established
+    /// the connection
+    ///
+    /// This can be different from the address the endpoint is bound to, in case
+    /// the endpoint is bound to a wildcard address like `0.0.0.0` or `::`.
+    ///
+    /// This will return `None` for clients, as well for servers if capturing
```
djc, line 1077: `as well for` seems a bit odd, add an extra `as`? I also think this docstring would be more useful if it explicitly lists supported platforms.

quinn/src/connection.rs
```diff
@@ -397,6 +412,18 @@ where
...
     }
 
+    /// The local IP address which was used when the peer established
+    /// the connection
+    ///
+    /// This can be different from the address the endpoint is bound to, in case
+    /// the endpoint is bound to a wildcard address like `0.0.0.0` or `::`.
+    ///
+    /// This will return `None` for clients, as well for servers if capturing
+    /// the local IP from incoming packets is not supported on the current platform.
```
djc, line 422: Same as above.

quinn/src/connection.rs
```diff
@@ -127,6 +127,21 @@ where
...
     }
+
+    /// The local IP address which was used when the peer established
+    /// the connection
+    ///
+    /// This can be different from the address the endpoint is bound to, in case
+    /// the endpoint is bound to a wildcard address like `0.0.0.0` or `::`.
+    ///
+    /// This will return `None` for clients, as well for servers if capturing
+    /// the local IP from incoming packets is not supported on the current platform.
```
djc, line 138: Same as above.

quinn/src/tests.rs
```diff
@@ -376,6 +376,14 @@ fn run_echo(client_addr: SocketAddr, server_addr: SocketAddr) {
 
         let handle = runtime.spawn(async move {
             let incoming = server_incoming.next().await.unwrap();
+
+            if cfg!(target_os = "linux") {
+                let local_ip = incoming.local_ip().expect("Local IP must be available");
+                assert!(local_ip.is_loopback());
+            } else {
+                assert_eq!(None, incoming.local_ip());
+            }
```
djc, line 385: Add a note here to update the getter docs in case the platform support falls out of sync?

quinn-proto/src/connection/mod.rs
Ralith: Alternatively, `and` might be a bit more concise.

quinn/src/tests.rs
Matthias247: Here, or rather to the platform files? This might be the place where people realize that more changes are necessary - because CI fails. But I somehow don't think that "update doc comments" is best kept in the tests.  In the platform files - maybe. But if someone works on adding it to windows they won't look at the unix file.

quinn-proto/src/connection/mod.rs
Matthias247: I went for listing the supported platform[s]

Ralith (dismissed):

quinn/src/tests.rs
djc: I mean, it could be both, but as you say, I'm not sure someone adding support for a different platform would review the code for unix. So I think adding it here would definitely be useful, maybe do the platform files as well (I didn't see a straightforward single location where it could be added, but I could easily be wrong about that).

djc (approved):

djc:
Thanks for working through this!

PR #945, opened 2020-12-27
Extend benchmark for concurrent streams

Matthias247:
This extends the "cargo bench" benchmark to test transmission of data
on multiple streams in parallel.

Ideally we would want all transmission to take a similar amount of time
(bandwidth shared fairly), and the overall throughput to stay the same
or increase (but never degrade) when multiple streams are in use.

The benchmark can not show the fairness aspect very well, since it only
prints an overall throughput. But it shows that performance with
multiple streams stays similar.

Output:

```
running 4 tests
test large_data_10_streams  ... bench: 180,761,260 ns/iter (+/- 12,090,679) = 58 MB/s
test large_data_1_stream    ... bench:  18,459,500 ns/iter (+/- 3,804,219) = 56 MB/s
test small_data_100_streams ... bench:   4,600,890 ns/iter (+/- 373,842)
test small_data_1_stream    ... bench:      71,019 ns/iter (+/- 34,298)
```

I also increased the size of the LARGE_DATA payload from 128kB to
1MB, since 128kB didn't show quinn peak performance.

### DIFF
--- a/quinn/benches/bench.rs
+++ b/quinn/benches/bench.rs
@@ -12,46 +12,61 @@ use tracing_futures::Instrument as _;
 
 use quinn::{ClientConfigBuilder, Endpoint, ServerConfigBuilder};
 
-benchmark_group!(benches, large_streams, small_streams);
+benchmark_group!(
+    benches,
+    large_data_1_stream,
+    large_data_10_streams,
+    small_data_1_stream,
+    small_data_100_streams
+);
 benchmark_main!(benches);
 
-fn large_streams(bench: &mut Bencher) {
-    let _ = tracing_subscriber::fmt::try_init();
-
-    let ctx = Context::new();
-    let (addr, thread) = ctx.spawn_server();
-    let (endpoint, client, mut runtime) = ctx.make_client(addr);
-    const DATA: &[u8] = &[0xAB; 128 * 1024];
-    bench.bytes = DATA.len() as u64;
-    bench.iter(|| {
-        runtime.block_on(async {
-            let mut stream = client.open_uni().await.unwrap();
-            stream.write_all(DATA).await.unwrap();
-            stream.finish().await.unwrap();
-        });
-    });
-    drop(client);
-    runtime.block_on(endpoint.wait_idle());
-    thread.join().unwrap();
+fn large_data_1_stream(bench: &mut Bencher) {
+    send_data(bench, LARGE_DATA, 1);
 }
 
-fn small_streams(bench: &mut Bencher) {
+fn large_data_10_streams(bench: &mut Bencher) {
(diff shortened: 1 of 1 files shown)


quinn/benches/bench.rs
```diff
@@ -12,46 +12,65 @@ use tracing_futures::Instrument as _;
...
+    large_data_1_stream,
+    large_data_10_streams,
+    small_data_1_stream,
+    small_data_100_streams
+);
 benchmark_main!(benches);
 
-fn large_streams(bench: &mut Bencher) {
-    let _ = tracing_subscriber::fmt::try_init();
+const LARGE_DATA: &[u8] = &[0xAB; 1024 * 1024];
```
djc, line 24: Let's define these constants at the bottom of the module.

quinn/benches/bench.rs
```diff
@@ -12,46 +12,65 @@ use tracing_futures::Instrument as _;
...
     let ctx = Context::new();
     let (addr, thread) = ctx.spawn_server();
     let (endpoint, client, mut runtime) = ctx.make_client(addr);
-    const DATA: &[u8] = &[0xAB; 1];
-    bench.bytes = 1;
+    let client = Arc::new(client);
+
+    bench.bytes = (data.len() as u64) * (concurrent_streams as u64);
     bench.iter(|| {
+        let mut handles = vec![];
```
djc, line 54: Why not use `FuturesUnordered` here? Also, if no arguments are passed we generally prefer `Vec::new()` over `vec![]` (on the premise that, all other things being equal, we prefer using fewer abstractions).

djc (commented):
LGTM, some minor nits.

quinn/benches/bench.rs
Matthias247: I actually found just spawning all and then waiting in order to be the simpler abstraction. But there is not too much difference. It was just the first thing that came to my mind for this use-case.

It willl actually have an impact on performance if the non single-threaded tokio runtime is used:
- With `FuturesUnordered` everything will still run in on tokio task, and there will be no races on accessing the connection from there.
- With `tokio::spawn` tasks on multiple threads for access on the connection.

I think going for any of the 2 approaches as a baseline and then checking out differences in follow-ups is reasonable.
I can change this version to `FuturesUnordered` if you prefer.

Ralith (approved):
Good stuff!

quinn/benches/bench.rs
Ralith: Leaving it as-is makes sense for reducing the changes needed in the future to perform a comparison between schedulers, I think. I don't expect the overhead reduction from `FuturesUnordered` to be huge here.

djc:
Great, thanks!

PR #946, opened 2020-12-27
Extend bulk benchmark

Matthias247:
This change improves the bulk benchmark:
- The total amount of requests the benchmark performs can be configured
- The amount of concurrent requests on a given connection can be
  configured
- The amount of data to transfer per stream can be configured

E.g.

```
bulk --streams 300 --max_streams 100 --stream_size 10
```

will create 300 streams - with a maximum of 100 active streams at a time,
and send 10MB of data on each stream.

The "short" cli flags are taken form h2load where applicable.

The benchmark now also shows metrics which indicate the performance
of individual streams, which allows to judge fairness. Example output:

```
Overall stats:

Sent 3145728000 bytes on 300 streams in 12.11s (247.81 MiB/s)

Stream metrics:

      │  Throughput   │ Duration
 AVG  │  149.17 MiB/s │  396.00ms
 P0   │    0.83 MiB/s │   39.00ms
 P10  │   51.19 MiB/s │   40.00ms
 P50  │  129.75 MiB/s │   77.00ms
 P90  │  247.00 MiB/s │  195.00ms
 P100 │  254.75 MiB/s │    12.11s
```

### DIFF
--- a/bench/src/bulk.rs
+++ b/bench/src/bulk.rs
@@ -1,5 +1,8 @@
-use std::net::{IpAddr, Ipv6Addr, SocketAddr};
-use std::time::Instant;
+use std::{
+    net::{IpAddr, Ipv6Addr, SocketAddr},
+    sync::Arc,
+    time::{Duration, Instant},
+};
 
 use anyhow::{anyhow, Context, Result};
 use futures::StreamExt;
@@ -87,21 +90,62 @@ async fn client(server_addr: SocketAddr, server_cert: quinn::Certificate) -> Res
         .context("unable to connect")?;
     trace!("connected");
 
+    let start = Instant::now();
+
+    let connection = Arc::new(connection);
+
+    let send_result = send_data_on_stream(connection, 1024).await?;
+
+    println!(
+        "sent {} bytes in {:?}",
+        1024 * send_result.size,
+        start.elapsed()
+    );
+    Ok(())
+}
+
+async fn send_data_on_stream(
+    connection: Arc<quinn::Connection>,
+    stream_size_mb: usize,
+) -> Result<SendResult> {
+    const DATA: &[u8] = &[0xAB; 1024 * 1024];
+
+    let start = Instant::now();
+
     let mut stream = connection
         .open_uni()
         .await
         .context("failed to open stream")?;
-    const DATA: &[u8] = &[0xAB; 1024 * 1024];
-    let start = Instant::now();
-    for _ in 0..1024 {
+
+    for _ in 0..stream_size_mb {
         stream
             .write_all(DATA)
             .await
             .context("failed sending data")?;
     }
+
     stream.finish().await.context("failed finishing stream")?;
-    println!("sent {} bytes in {:?}", 1024 * DATA.len(), start.elapsed());
-    Ok(())
+
+    let duration = start.elapsed();
+    let size = DATA.len() * stream_size_mb;
+    let throughput = throughput_bps(duration, size as u64);
+
+    Ok(SendResult {
+        duration,
+        size,
+        throughput,
+    })
+}
+
+#[derive(Debug)]
+struct SendResult {
+    duration: Duration,
+    size: usize,
+    throughput: f64,
+}
+
+fn throughput_bps(duration: Duration, size: u64) -> f64 {
+    (size as f64) / (duration.as_secs_f64())
 }
 
 fn rt() -> Runtime {
--- a/bench/src/bulk.rs
+++ b/bench/src/bulk.rs
@@ -4,12 +4,15 @@ use std::{
     time::{Duration, Instant},
 };
 
-use anyhow::{anyhow, Context, Result};
+use anyhow::{Context, Result};
 use futures::StreamExt;
+use hdrhistogram::Histogram;
+use structopt::StructOpt;
 use tokio::runtime::{Builder, Runtime};
-use tracing::trace;
+use tracing::{info, trace};
 
 fn main() {
+    let opt = Opt::from_args();
     tracing::subscriber::set_global_default(
         tracing_subscriber::FmtSubscriber::builder()
             .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
@@ -25,8 +28,12 @@ fn main() {
     server_config
         .certificate(quinn::CertificateChain::from_certs(vec![cert.clone()]), key)
         .unwrap();
+
+    let mut server_config = server_config.build();
+    server_config.transport = Arc::new(transport_config());
+
     let mut endpoint = quinn::EndpointBuilder::default();
-    endpoint.listen(server_config.build());
+    endpoint.listen(server_config);
     let mut runtime = rt();
     let (endpoint, incoming) = runtime.enter(|| {
         endpoint
@@ -42,7 +49,7 @@ fn main() {
     });
 
     let mut runtime = rt();
-    if let Err(e) = runtime.block_on(client(server_addr, cert)) {
+    if let Err(e) = runtime.block_on(client(server_addr, cert, opt)) {
         eprintln!("client failed: {:#}", e);
     }
 
@@ -54,27 +61,25 @@ async fn server(mut incoming: quinn::Incoming) -> Result<()> {
     let quinn::NewConnection {
         mut uni_streams, ..
     } = handshake.await.context("handshake failed")?;
-    let mut stream = uni_streams
-        .next()
-        .await
-        .ok_or_else(|| anyhow!("accepting stream failed"))??;
-    trace!("stream established");
-    let start = Instant::now();
-    let mut n = 0;
-    while let Some((data, offset)) = stream.read_unordered().await? {
-        n = n.max(offset + data.len() as u64);
+
+    loop {
+        let mut stream = match uni_streams.next().await {
+            None => return Ok(()),
+            Some(Err(quinn::ConnectionError::ApplicationClosed(_))) => return Ok(()),
+            Some(Err(e)) => return Err(e).context("accepting stream failed"),
+            Some(Ok(stream)) => stream,
+        };
+        trace!("stream established");
+
+        let _: tokio::task::JoinHandle<Result<()>> = tokio::spawn(async move {
+            while stream.read_unordered().await?.is_some() {}
+
+            Ok(())
+        });
     }
-    let dt = start.elapsed();
-    println!(
-        "recvd {} bytes in {:?} ({} MiB/s)",
-        n,
-        dt,
-        n as f32 / (dt.as_secs_f32() * 1024.0 * 1024.0)
-    );
-    Ok(())
 }
 
-async fn client(server_addr: SocketAddr, server_cert: quinn::Certificate) -> Result<()> {
+async fn client(server_addr: SocketAddr, server_cert: quinn::Certificate, opt: Opt) -> Result<()> {
     let (endpoint, _) = quinn::EndpointBuilder::default()
         .bind(&SocketAddr::new(IpAddr::V6(Ipv6Addr::LOCALHOST), 0))
         .unwrap();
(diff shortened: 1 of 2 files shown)


bench/src/bulk.rs
```diff
@@ -51,27 +62,72 @@ async fn server(mut incoming: quinn::Incoming) -> Result<()> {
...
+    loop {
+        let mut stream = match uni_streams.next().await {
+            None => return Ok(()),
+            Some(Err(quinn::ConnectionError::ApplicationClosed(_))) => return Ok(()),
+            Some(Err(e)) => return Err(e).context("accepting stream failed"),
+            Some(Ok(stream)) => stream,
+        };
+        trace!("stream established");
+
+        let _jh: tokio::task::JoinHandle<Result<()>> = tokio::spawn(async move {
```
djc, line 75: Why do we need to name the type of the `JoinHandle` here?

bench/src/bulk.rs
```diff
@@ -51,27 +62,72 @@ async fn server(mut incoming: quinn::Incoming) -> Result<()> {
...
+}
+
+#[derive(Debug)]
+struct SendResult {
+    duration: Duration,
+    size: usize,
+    throughput: f64,
+}
+
+async fn send_data_on_stream(
```
djc, line 93: Can we split out a separate commit that extracts `send_data_on_stream()` out of `client()`?

bench/src/bulk.rs
```diff
@@ -51,27 +62,72 @@ async fn server(mut incoming: quinn::Incoming) -> Result<()> {
...
+            while let Some((data, offset)) = stream.read_unordered().await? {
+                n = n.max(offset + data.len() as u64);
+            }
+
+            Ok(())
+        });
+    }
+}
+
+#[derive(Debug)]
```
djc, line 86: Since `SendResult` is used by `send_data_on_stream()`, let's move the type below the function defn.

bench/src/bulk.rs
```diff
@@ -80,34 +136,143 @@ async fn client(server_addr: SocketAddr, server_cert: quinn::Certificate) -> Res
...
+    let connection = Arc::new(connection);
+    let get_connection = || connection.clone();
+
+    let mut ops = futures::stream::iter((0..config.streams).map(|_| async move {
+        send_data_on_stream(get_connection(), config.stream_size_mb).await
+    }))
+    .buffer_unordered(config.max_streams);
+
+    let mut total_size = 0;
+    let mut duration_hist: Histogram<u64> = Histogram::new(3).unwrap();
```
djc, line 160: Can we remove the need for the type annotations here by adding a `u64` suffix to the `new()` parameter?

bench/src/bulk.rs
```diff
@@ -80,34 +136,143 @@ async fn client(server_addr: SocketAddr, server_cert: quinn::Certificate) -> Res
...
+
 fn rt() -> Runtime {
     Builder::new()
         .basic_scheduler()
         .enable_all()
         .build()
         .unwrap()
 }
+
+fn transport_config() -> quinn::TransportConfig {
```
djc, line 238: I think I'd prefer to keep this inline and just clone the `transport_config`, what do you think?

bench/src/bulk.rs
```diff
@@ -80,34 +136,143 @@ async fn client(server_addr: SocketAddr, server_cert: quinn::Certificate) -> Res
...
+
+    println!("Stream metrics:\n");
+
+    println!(
+        "AVG : Throughput: {:7.2} MiB/s, Duration: {:.2?}",
+        throughput_hist.mean() as f64 / 1024.0 / 1024.0,
+        Duration::from_millis(duration_hist.mean() as u64)
+    );
+    println!(
+        "P0  : Throughput: {:7.2} MiB/s, Duration: {:.2?}",
```
djc, line 192: I find the ordering a bit counterintuitive here, I guess because percentiles are more usually used for latencies, so I expect the highest percentile to be worst and the lowest percentile to be best. You probably have more experience with throughput distribution representations than me, though...

djc (commented):
This looks very useful! Some questions/suggestions about style choices follow...

bench/src/bulk.rs
Matthias247: We have to explicitely write the type so that `?` works. `_jh` can be replaced with `_`. I typically like to still see what gets ignored based on the variable name. But it's obviously shown here by the type parameter. I can remove it.

bench/src/bulk.rs
Matthias247: That's the reason why I have it above: I like to learn about dependencies before the places where they are used. Allows to read from top to bottom :-) 

But I will change it if preferred

bench/src/bulk.rs
Matthias247: We could. But given that half of the changes are still separate (e.g. measuring durations and having a return value which contains metrics), I don't think this is a super worthwhile exercise.

bench/src/bulk.rs
Matthias247: done

bench/src/bulk.rs
Matthias247: It's created 2 times - once for the server and once for the client in 2 different functions. I think when trying to play around with new values it's easiest to modify both in one place. Therefore I moved it out (and actually there wasn't any explicit configuration before).

bench/src/bulk.rs
Matthias247: I don't think there is a right or wrong here. Both approaches work. I actually want to add first-byte latency as another column, and there P0 is the best value and P100 the worst 🤷‍♂️

bench/src/bulk.rs
Ralith: I tend to come down on @djc's side here as well, as it's easy to overflow your working memory with dependencies before you get to the root, and the root can be a useful guide to understand the significance of the dependencies.

bench/src/bulk.rs
Ralith: I was confused about the example output because it seemed to suggest that the lowest-throughput stream had the shortest duration. Not sure how to do better without making it massively less compact, though.

Matthias247: I  agree. If one interprets it as "these are the values of individual streams", it gets a bit confusing. It should be clear that those are not, because we have far more streams, but 🤷‍♂️

One thing that can be changed is to make it a real table, and move the throughput / duration labels to a separate row. Not sure if this is clearer or not however.

Individual stream metrics are found in the the trace:

```
Dec 27 12:00:39.056  INFO bulk: stream finished: Ok(SendResult { duration: 11.809008s, size: 10485760, throughput: 887945.8799587569 })
```

Matthias247: Here's the table format. I think I slightly prefer it:

```
Stream metrics:

      │  Throughput   │ Duration
 AVG  │  155.43 MiB/s │  391.00ms
 P0   │    0.84 MiB/s │   39.00ms
 P10  │   50.56 MiB/s │   40.00ms
 P50  │  131.25 MiB/s │   76.00ms
 P90  │  249.87 MiB/s │  195.00ms
 P100 │  255.25 MiB/s │    11.97s
```

bench/src/bulk.rs
```diff
@@ -80,34 +136,128 @@ async fn client(server_addr: SocketAddr, server_cert: quinn::Certificate) -> Res
...
+
+#[derive(Debug, Clone, Copy)]
+struct BenchConfig {
+    /// The total number of streams which should be created
+    streams: usize,
+    /// The amount of concurrent streams which should be used
+    max_streams: usize,
+    /// The amount of data to transfer on a stream in megabytes
+    stream_size_mb: usize,
+}
```
Ralith, line 253: Can we really not just derive `Clone + Copy` on `Opt`?

bench/src/bulk.rs
```diff
@@ -51,27 +62,72 @@ async fn server(mut incoming: quinn::Incoming) -> Result<()> {
...
+            Some(Err(e)) => return Err(e).context("accepting stream failed"),
+            Some(Ok(stream)) => stream,
+        };
+        trace!("stream established");
+
+        let _: tokio::task::JoinHandle<Result<()>> = tokio::spawn(async move {
+            let mut n = 0u64;
+            while let Some((data, offset)) = stream.read_unordered().await? {
+                n = n.max(offset + data.len() as u64);
+            }
```
Ralith, line 79: What are we keeping track of `n` for?

bench/src/bulk.rs
```diff
@@ -80,34 +136,128 @@ async fn client(server_addr: SocketAddr, server_cert: quinn::Certificate) -> Res
...
-    const DATA: &[u8] = &[0xAB; 1024 * 1024];
     let start = Instant::now();
-    for _ in 0..1024 {
-        stream
-            .write_all(DATA)
-            .await
-            .context("failed sending data")?;
+
+    let connection = Arc::new(connection);
+    let get_connection = || connection.clone();
```
Ralith, line 152: Style nit: this feels really cheesy. Let's just move a clone into the closure, even if that means doing a dance with blocks or naming something `connection2`.

Ralith (commented):
Great improvements overall!

bench/src/bulk.rs
Matthias247: Sure. I tried to separate the CLI options and function arguments. But for such a small program it likely doesn't matter.

bench/src/bulk.rs
Matthias247: leftover from the old version. Will remove

bench/src/bulk.rs
Matthias247: I don't get what is recommended here. The workaround with the closure is required to get access to the connection in each stream map iteration (which happens only when the whole future is awaited). A single clone won't be sufficient.

bench/src/bulk.rs
djc: Sorry, must have missed that -- I did find the two usages but thought they were in the same function.

bench/src/bulk.rs
djc: Yeah, I like the table format, too.

bench/src/bulk.rs
```diff
@@ -80,34 +128,112 @@ async fn client(server_addr: SocketAddr, server_cert: quinn::Certificate) -> Res
...
+
+    let mut ops =
+        futures::stream::iter((0..opt.streams).map(|_| async move {
+            send_data_on_stream(get_connection(), opt.stream_size_mb).await
+        }))
```
djc, line 149: Something like this?

```suggestion
        futures::stream::iter((0..opt.streams).map(|_| {
            let connection = connection.clone();
            async move {
                send_data_on_stream(connection, opt.stream_size_mb).await
            }
        }))
```

bench/src/bulk.rs
djc: Okay. Maybe also just import `tokio::task::JoinHandle` up top so that we don't have the whole path inline here.

bench/src/bulk.rs
djc: I do feel that it makes changes like this substantially easier to review, which I feel is valuable even though splitting the commit (or redoing the changes) can feel like drudgery. So while I understand your reluctance, I would appreciate it.

bench/src/bulk.rs
Matthias247: I don't think that is any improvement (as a reader now the first thing I'm asking myself is what this weird extra block is doing there), but I still applied it.

Frankly speaking: I feel there is a bit of too much bikeshedding going on. This comments all lead to spending effort to move code back and forth without any objective reason why it should be in one particular place, and without any functional changes. If this is highly important to you, please consider writing a styleguide

djc: That's fair, I'll see if I can write a style guide soonish. On the other hand, I'm not sure it would have covered everything we commented on here. In my experience, we routinely we give each other feedback on the clearest way to express something in code/submit something for review, and make changes accordingly, and it generally has been fine.

djc (approved):

PR #948, opened 2020-12-27
Fix state leak when a stopped stream is reset

Ralith:


djc (approved):

djc:
Does this merit a bugfix release on the 0.6 branch?

Ralith:
This has good odds of happening whenever an open receive stream is dropped, so that sounds sensible.

quinn-proto/src/connection/streams.rs
```diff
@@ -350,7 +350,12 @@ impl Streams {
             return Ok(ShouldTransmit::new(false));
         }
         let bytes_read = rs.assembler.bytes_read();
-        self.on_stream_frame(true, id);
+        let stopped = rs.assembler.is_stopped();
+        if stopped {
```
Matthias247, line 354: Is just `stopped` enough to remove it, or should it be "dropped by application". One can call `stop` without dropping the stream. The next interaction would then lead to an unknown stream error?

Ralith: IMO this is sufficient. There's no operation that's meaningful to perform on a stopped stream. The documentation for each `pub fn stop` says: `Once stopped, a stream cannot be read from any further.` Perhaps we should make this clearer.

Issue #929, opened 2020-12-01
Add RecvStream::recv(), a middleground between read() and read_unordered()


seanmonstar:
This is a request to add a `recv` like method to `RecvStream`, which yields `Bytes` but still ordered.

There are currently two main ways of reading data from a `RecvStream`:

- `read(&mut [u8])`: This is an ordered read, but copies the data into a pre-allocated buffer.
- `read_unordered() -> (Bytes, u64)`: As the name implies, this is an unordered read, but it gives a `Bytes` object, which allows the user to receive those bytes without a copy.

In h3, when proving the h3<->quinn bridge, in order to try to reduce copies, we've been using the `read_unordered`. However, that means we end up having an intermediate collection to wait to put the `Bytes` in order, since we also need ordered reads. Since quinn is already doing that, it could be useful to provide a method that yields `Bytes` in an ordered manner. Looking in quinn's source, it'd probably be a simpler `read`, since it wouldn't need to handle partial reads into buffers, and can just yield a `Bytes` once the order is right.

So, the final method signature would be something like `async fn recv(&mut self) -> Result<Option<Bytes>, ReadError>`. (As a side note, if this is accepted, perhaps it might make sense to rename `read_unordered` to `recv_unordered`.)

Ralith:
Good call, we should have this.

Ralith:
This is now implemented as [`RecvStream::read_chunk`] with `ordered = true`.

PR #952, opened 2020-12-29
Read ordered chunks from RecvStream

stammw:
Hello !

I've started trying to implement #929.

Still I'd better be adding some tests before its ready. So I'm setting this PR WIP for a first feedback.

### DIFF
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -59,6 +59,37 @@ impl Assembler {
         Some((n, data))
     }
 
+    // Get the the next ordered chunk
+    pub(crate) fn read_chunk(&mut self) -> Result<Option<Bytes>, IllegalOrderedRead> {
+        if let State::Unordered { .. } = self.state {
+            return Err(IllegalOrderedRead);
+        }
+
+        loop {
+            let chunk = match self.data.peek() {
+                Some(chunk) => chunk,
+                None => return Ok(None),
+            };
+
+            if chunk.offset > self.bytes_read {
+                // Next chunk is after current read index
+                return Ok(None);
+            } else if (chunk.offset + chunk.bytes.len() as u64) <= self.bytes_read {
+                // Next chunk is useless as the read index is beyond its end
+                self.data.pop();
+                continue;
+            }
+
+            // Determine `start` and `len` of the slice of useful data in chunk
+            let start = (self.bytes_read - chunk.offset) as usize;
+            let len = chunk.bytes.len() - start;
+
+            self.bytes_read += len as u64;
+
+            return Ok(self.data.pop().map(|mut x| x.bytes.split_off(start)));
+        }
+    }
+
     // Read as much from the first chunk in the heap as fits in the buffer.
     // Takes the buffer to read into and the amount of bytes that has already
     // been read into it. Returns whether the first chunk has been fully consumed.
@@ -466,5 +497,26 @@ mod test {
         assert_eq!(x.read_unordered(), None);
         x.insert(2, Bytes::from_static(b"cde"));
         assert_eq!(x.read_unordered(), None);
+    }
+
+    #[test]
+    fn chunks_dedup() {
+        let mut x = Assembler::new();
+        x.insert(3, Bytes::from_static(b"def"));
+        assert_eq!(x.read_chunk().unwrap(), None);
+        x.insert(0, Bytes::from_static(b"a"));
+        x.insert(1, Bytes::from_static(b"bcdefghi"));
+        x.insert(0, Bytes::from_static(b"abcd"));
+        assert_eq!(x.read_chunk().unwrap(), Some(Bytes::from_static(b"abcd")));
+        assert_eq!(x.read_chunk().unwrap(), Some(Bytes::from_static(b"efghi")));
+        assert_eq!(x.read_chunk().unwrap(), None);
+        x.insert(8, Bytes::from_static(b"ijkl"));
+        assert_eq!(x.read_chunk().unwrap(), Some(Bytes::from_static(b"jkl")));
+        assert_eq!(x.read_chunk().unwrap(), None);
+        x.insert(12, Bytes::from_static(b"mno"));
+        assert_eq!(x.read_chunk().unwrap(), Some(Bytes::from_static(b"mno")));
+        assert_eq!(x.read_chunk().unwrap(), None);
+        x.insert(2, Bytes::from_static(b"cde"));
+        assert_eq!(x.read_chunk().unwrap(), None);
     }
 }
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -868,14 +868,22 @@ where
     ///
     /// While stream data is typically processed by applications in-order, unordered reads improve
     /// performance when packet loss occurs and data cannot be retransmitted before the flow control
-    /// window is filled. When in-order delivery is required, the sibling `read()` method should be
-    /// used.
+    /// window is filled. When in-order delivery is required, the sibling `read()` or `read_chunk()`
+    /// methods should be used.
     ///
     /// The return value if `Ok` contains the bytes and their offset in the stream.
     pub fn read_unordered(&mut self, id: StreamId) -> Result<Option<(Bytes, u64)>, ReadError> {
         Ok(self.streams.read_unordered(id)?.map(|result| {
             self.add_read_credits(id, result.max_stream_data, result.max_data);
             (result.buf, result.offset)
+        }))
+    }
+
+    /// Read the next ordered chunk from the given recv stream
+    pub fn read_chunk(&mut self, id: StreamId) -> Result<Option<Bytes>, ReadError> {
+        Ok(self.streams.read_chunk(id)?.map(|result| {
+            self.add_read_credits(id, result.max_stream_data, result.max_data);
+            result.buf
         }))
     }
 
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -868,7 +868,7 @@ where
     ///
     /// While stream data is typically processed by applications in-order, unordered reads improve
     /// performance when packet loss occurs and data cannot be retransmitted before the flow control
-    /// window is filled. When in-order delivery is required, the sibling `read()` or `read_chunk()`
+    /// window is filled. When in-order delivery is required, the sibling `read()` or `read_chunk[s]()`
     /// methods should be used.
     ///
     /// The return value if `Ok` contains the bytes and their offset in the stream.
@@ -884,6 +884,18 @@ where
         Ok(self.streams.read_chunk(id)?.map(|result| {
             self.add_read_credits(id, result.max_stream_data, result.max_data);
             result.buf
+        }))
+    }
+
+    /// Read the next ordered chunks from the given recv stream
+    pub fn read_chunks(
+        &mut self,
+        id: StreamId,
+        bufs: &mut [Bytes],
+    ) -> Result<Option<usize>, ReadError> {
+        Ok(self.streams.read_chunks(id, bufs)?.map(|result| {
+            self.add_read_credits(id, result.max_stream_data, result.max_data);
+            result.n_bufs
         }))
     }
 
--- a/quinn-proto/src/connection/streams.rs
+++ b/quinn-proto/src/connection/streams.rs
@@ -213,6 +213,23 @@ impl Streams {
         }))
     }
 
+    pub fn read_chunks(
+        &mut self,
+        id: StreamId,
+        bufs: &mut [Bytes],
+    ) -> Result<Option<ReadChunksResult>, ReadError> {
+        let res = self.try_read(id, |rs: &mut Recv| rs.read_chunks(bufs))?;
+        Ok(res.map(|(n_bufs, max_stream_data)| {
+            let len = bufs[..n_bufs].iter().fold(0, |t, b| t + b.len());
+            let max_data = self.add_read_credits(len as u64);
+            ReadChunksResult {
+                n_bufs,
+                max_stream_data,
+                max_data,
+            }
+        }))
+    }
+
     fn try_read<T, U>(
         &mut self,
         id: StreamId,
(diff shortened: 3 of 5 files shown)


quinn-proto/src/connection/assembler.rs
```diff
@@ -59,6 +59,39 @@ impl Assembler {
...
+
+            if chunk.offset > self.bytes_read {
+                // Next chunk is after current read index
+                return Ok(None);
+            } else if (chunk.offset + chunk.bytes.len() as u64) <= self.bytes_read {
+                // Next chunk is useless as the read index is beyond its end
+                drop(chunk);
+                self.data.pop();
+                continue;
+            }
```
djc: Is this actually necessary? I thought we prevented this scenario in the assembler.

quinn-proto/src/connection/streams.rs
```diff
@@ -236,6 +236,34 @@ impl Streams {
         }
     }
 
+    pub fn read_chunk(&mut self, id: StreamId) -> Result<Option<ReadChunkResult>, ReadError> {
+        let mut entry = match self.recv.entry(id) {
+            hash_map::Entry::Vacant(_) => return Err(ReadError::UnknownStream),
+            hash_map::Entry::Occupied(e) => e,
+        };
+        let rs = entry.get_mut();
+        match rs.read_chunk() {
```
djc, line 245: This is pretty repetitive, can we factor out a bunch of the match here (ideally in a separate commit)? Could apply to `read_unordered()` as well, ideally, but just between `read()` and `read_chunk()` would be good if that makes it hard.

quinn-proto/src/connection/assembler.rs
```diff
@@ -59,6 +59,37 @@ impl Assembler {
...
+            let chunk = match self.data.peek() {
+                Some(chunk) => chunk,
+                None => return Ok(None),
+            };
+
+            if chunk.offset > self.bytes_read {
+                // Next chunk is after current read index
+                return Ok(None);
+            } else if (chunk.offset + chunk.bytes.len() as u64) <= self.bytes_read {
+                // Next chunk is useless as the read index is beyond its end
```
djc, line 78: Does this actually happen? I thought the `Assembler` prevented this kind of thing.

quinn-proto/src/connection/streams.rs
```diff
@@ -1124,6 +1161,14 @@ impl Recv {
         }
     }
 
+    fn read_chunk(&mut self) -> Result<Option<Bytes>, ReadError> {
+        if let Some(bytes) = self.assembler.read_chunk()? {
+            Ok(Some(bytes))
+        } else {
+            self.read_blocked().map(|()| None)
+        }
```
djc, line 1169: A `match` expression would be more readable and more compact, I think?

quinn/src/streams.rs
```diff
@@ -431,6 +436,46 @@ where
...
     }
 
+    /// Read the next segment of data from next valid offset in the stream.
+    ///
+    /// Yields a segment of data consecutive to the last ordered read offset, or `None`
+    /// if the stream was finished. Segments are received in order.
+    ///
+    /// Chunk reads have reduced overhead and higher throughput, and should therefore be
+    /// preferred when applicable.
+    fn poll_read_chunk(&mut self, cx: &mut Context) -> Poll<Result<Option<Bytes>, ReadError>> {
```
djc, line 446: Also pretty repetitive. Can we factor these things?

quinn-proto/src/connection/assembler.rs
stammw: It replicates the logic I found in `consume()`.

stammw: Looks like `insert()` doesn't push already received data, but a consecutive insert containing the same data would leave the precedent useless chunk in (unless defragmentation happens ?).

quinn-proto/src/connection/assembler.rs
stammw: Looks like insert() doesn't push already received data, but a consecutive insert containing the same data would leave the precedent useless chunk in (unless defragmentation happens ?).

djc: Ah yeah, fair enough.

quinn/src/streams.rs
```diff
@@ -347,6 +347,11 @@ where
         }
     }
 
+    /// [`read_chunk()`]: RecvStream::read_chunk
```
Ralith, line 350: This should be documented, including an explicit warning against relying on the chunk boundaries having any particular relationship to the sender's writes.

Ralith (changes requested):
Thanks for working on this!

quinn-proto/src/connection/assembler.rs
Ralith: In fact, under normal (ordered) operation `insert` makes no effort to deduplicate whatsoever unless defragmentation occurs, so this is very necessary. See also the analogous code in `read`.

quinn-proto/src/connection/assembler.rs
Ralith: This case is needed per https://github.com/quinn-rs/quinn/pull/952#discussion_r549865148.

quinn/src/streams.rs
stammw: Would a macro be acceptable here ?

stammw: This one wasn't that hard to refactor with a generic `fn`, actually.

quinn-proto/src/connection/streams.rs
stammw: I got a solution I find a little complex in feb3b63, what do you think about this?

quinn-proto/src/connection/streams.rs
stammw: done

quinn-proto/src/connection/streams.rs
djc: It looks like you can just inline the `IntoReadResult` into the closure passed to `try_read()`? We already have the nested `map()` calls anyway. Or am I missing something?

quinn/src/streams.rs
stammw: I got this a better comment. I'm not sure I grasped everything you'd like it to say?

quinn/src/streams.rs
```diff
@@ -399,36 +370,25 @@ where
...
+    ///
+    /// Beware, the returned chunks bounds won't always match those written by the peer.
+    ///
+    /// Chunk reads have reduced overhead and higher throughput, and should therefore be
+    /// preferred when applicable.
+    pub fn read_chunk(&mut self) -> ReadChunk<'_, S> {
+        ReadChunk { stream: self }
+    }
+
+    /// [`read_chunk()`]: RecvStream::read_chunk
```
Ralith, line 389:
```suggestion
    /// Foundation of [`read_chunk()`]: RecvStream::read_chunk
```

quinn/src/streams.rs
```diff
@@ -399,36 +370,25 @@ where
...
-                self.all_data_read = true;
-                Poll::Ready(Err(ReadError::Reset(error_code)))
-            }
-            Err(UnknownStream) => Poll::Ready(Err(ReadError::UnknownStream)),
-            Err(IllegalOrderedRead) => Poll::Ready(Err(ReadError::IllegalOrderedRead)),
-        }
+        self.poll_read_generic(cx, |conn, stream| conn.inner.read_unordered(stream))
+    }
+
+    /// Read the next segment of data from next valid offset in the stream.
```
Ralith, line 376:
```suggestion
    /// Read the next segment of data
```

quinn/src/streams.rs
```diff
@@ -399,36 +370,25 @@ where
...
+    ///
+    /// Yields a segment of data consecutive to the last ordered read offset, or `None`
+    /// if the stream was finished. Segments are received in order.
+    ///
+    /// Beware, the returned chunks bounds won't always match those written by the peer.
+    ///
+    /// Chunk reads have reduced overhead and higher throughput, and should therefore be
+    /// preferred when applicable.
```
Ralith, line 384: Tweaked wording a bit:
```suggestion
    /// Yields a segment of data beginning immediately after the last data yielded by
    /// `read` or `read_chunk`, or `None` if the stream was finished.
    ///
    /// Slightly more efficient than `read` due to not copying. Chunks boundaries
    /// do not correspond to peer writes, and hence cannot be used as framing.
```

quinn-proto/src/connection/mod.rs
```diff
@@ -868,8 +868,8 @@ where
     ///
     /// While stream data is typically processed by applications in-order, unordered reads improve
     /// performance when packet loss occurs and data cannot be retransmitted before the flow control
-    /// window is filled. When in-order delivery is required, the sibling `read()` method should be
-    /// used.
+    /// window is filled. When in-order delivery is required, the siblings `read()` or `read_chunk()`
+    /// method should be used.
```
Ralith, line 872:
```suggestion
    /// window is filled. When in-order delivery is required, the sibling `read()` or `read_chunk()`
    /// methods should be used.
```

Ralith (changes requested):
Wording tweaks for the comments:

quinn-proto/src/connection/streams.rs
stammw: Ho yeah maybe ! :sweat_smile:
For the moment I get stuck because `self.add_read_credits(len as u64)` needs to mutate `self` to compute `transmit_max_data`, and it is already mutably borrowed by the closure. But I'll try a little harder in a bit.

djc:
(Note that #961 is blocked on this, so would be nice to follow through soon...)

Ralith:
My wording in #961 was unclear--I don't actually consider it blocked, I'm just holding off on adding additional refactoring that would make rebasing *this* PR harder than it otherwise would be.

Matthias247:
While we are at it: I would also appreciate a vectored read API - which can retrieve more than 1 chunk at a time. Because locking a Mutex and potentially doing a task wakeup for reading just 1kB of data (due to Quic packet sizes) is fairly unefficient.

One API I've used in a project for that is:

```
async fn read(&mut self, &mut [Bytes]) -> Result<(usize, bool), Error>
```

Where the caller provides an array (initialized with dummy elements / `Bytes::new()`) - and the implementation fills as many array items as possible and returns the number of chunks read in the first return value. Second return value is for indicating end of stream.

stammw:
I finally got the read methods factorization right (yeah the `IntoResult` thing was weird...). 

I've addesd a test for `assembler::read_chunk()` in the first commit.

All the wording changes have been addressed.

A new commit contains @Matthias247  suggestion. This is still WIP as I'm waiting for your feedback on this public API change, and it should probably be tested in the functional test module.

djc:
Did you see #955?

stammw:
I should probably have paid more attention, sorry... Now the `IntoResult` solution is totally removed here. I'm a little too tired to judge which one is better.

djc:
Maybe @Ralith or @Matthias247 want to take a look which version they prefer.

quinn-proto/src/connection/assembler.rs
```diff
@@ -59,6 +59,37 @@ impl Assembler {
...
+
+        loop {
+            let chunk = match self.data.peek() {
+                Some(chunk) => chunk,
+                None => return Ok(None),
+            };
+
+            if chunk.offset > self.bytes_read {
+                // Next chunk is after current read index
+                return Ok(None);
```
Matthias247, line 76: Isn't `None` end of stream? i'm not sure how this function is used, but it might need to return a blocked status instead.

Matthias247:
> Maybe @Ralith or @Matthias247 want to take a look which version they prefer.

I skimmed both commits, and I have a slight preference for this one since I found the `ReadResult/StreamReadResult/IntoReadResult` types a bit harder to follow.

I think this change currently has a good balance between code reuse and readability.

@stammw Also thanks for incorporating my feedback!

quinn-proto/src/connection/streams.rs
```diff
@@ -1124,6 +1166,35 @@ impl Recv {
         }
     }
 
+    fn read_chunk(&mut self) -> Result<Option<Bytes>, ReadError> {
+        match self.assembler.read_chunk()? {
```
Matthias247, line 1170: You can implement this via `read_chunks:`

```rust
let mut bufs = &mut [Bytes::new()];
match self.read_chunks(bufs) {
    Ok(Some(1)) => Ok(Some(std::mem::replace(&mut bufs[0], Bytes::new()))),
    Ok(Some(_)) => unreachable!(),
    Ok(None) => Ok(None),
    Err(e) => Err(e),
}
```

Doesn't however really a buy a lot here. Might be more interesting if assembler would be doing less.

quinn-proto/src/connection/assembler.rs
stammw: This method works as it's siblings: it returns `None` when there is nothing to *assemble*. Blocked status is managed at the *stream* level in `quinn-proto::connection::streams::read_chunk()`.

See the `chunks_dedup()` test at the end of this file for an example of expected behavior.

quinn-proto/src/connection/streams.rs
stammw: To me there is no real advantage in factorizing in this way. Since it is different layers composition already, and doing so would add a tiny overhead without making the overall code shorter at all.

djc: I think I agree that it doesn't seem to help here.

stammw:
The current `read` methods factorization solution (without `IntoResult`) is actually better IMO.

I've added a test for `read_chunks()`, so this is ready to be merged if you guys like the current state of this PR.

quinn/src/streams.rs
```diff
@@ -399,36 +370,44 @@ where
...
+        self.poll_read_generic(cx, |conn, stream| conn.inner.read_unordered(stream))
+    }
+
+    /// Read the next segment of data
+    ///
+    /// Yields a segment of data beginning immediately after the last data yielded by
+    /// `read` or `read_chunk`, or `None` if the stream was finished.
+    ///
+    /// Slightly more efficient than `read` due to not copying. Chunks boundaries
+    /// do not correspond to peer writes, and hence cannot be used as framing.
```
djc, line 382:
```suggestion
    /// Slightly more efficient than `read` due to not copying. Chunk boundaries
    /// do not correspond to peer writes, and hence cannot be used as framing.
```

quinn/src/streams.rs
```diff
@@ -399,36 +370,44 @@ where
...
+    }
+
+    /// Read the next segments of data
+    ///
+    /// Fills `bufs` with the segments of data beginning immediately after the
+    /// last data yielded by `read` or `read_chunk`, or `None` if the stream was
+    /// finished.
+    ///
+    /// Slightly more efficient than `read` due to not copying. Chunks boundaries
+    /// do not correspond to peer writes, and hence cannot be used as framing.
```
djc, line 399:
```suggestion
    /// Slightly more efficient than `read` due to not copying. Chunk boundaries
    /// do not correspond to peer writes, and hence cannot be used as framing.
```

djc (commented):
LGTM, modulo small doc nits.

stammw:
Should be ok now. 

Happy to contribute to quinn a bit :)

djc (approved):

PR #953, opened 2020-12-29
Generic Send Offload (GSO) Support

Matthias247:
Generic Send Offload (GSO) Support
    
This change implements an initial version of GSO support [1][2][3]
for Linux, which improves the effiency of sending data.

The approach taken in this change is to create a buffer which contains
multiple datagrams in `Connection::poll_transmit`. This was picked
over trying to merge packets in the endpoint task, since packets
seem to need to be padded to a common segment size in order to
make them sendable via GSO.

In order to to this in an efficient fashion, the `poll_transmit`
method was restructred. Instead of selecting spaces upfront, it
will now loop through all possible packet spaces, check if there is
pending data to send and create packets and datagrams out of this.

The last packet which was written to a datagram buffer is not
finalized until it is clear whether follow-up packets need to be
written, since we need to know whether this packet should get padded
to MTU length or not.

This change doesn't enable GSO yet, since it will only produce a
single datagram. I will create a separate change for this.

Performance measurements:

**No GSO:**

```
Sent 1073741824 bytes on 1 streams in 4.31s (237.66 MiB/s)
```

**With GSO (up to 8 packets):**

```
Sent 1073741824 bytes on 1 streams in 3.02s (339.18 MiB/s)
```

[1] http://vger.kernel.org/lpc_net2018_talks/willemdebruijn-lpc2018-udpgso-paper-DRAFT-1.pdf
[2] http://vger.kernel.org/lpc_net2018_talks/willemdebruijn-lpc2018-udpgso-presentation-20181104.pdf
[3] https://lwn.net/Articles/752956/


quinn/src/platform/unix.rs
```diff
@@ -146,7 +146,7 @@ impl super::UdpExt for UdpSocket {
     fn send_ext(&self, transmits: &[Transmit]) -> io::Result<usize> {
         let mut msgs: [libc::mmsghdr; BATCH_SIZE] = unsafe { mem::zeroed() };
         let mut iovecs: [libc::iovec; BATCH_SIZE] = unsafe { mem::zeroed() };
-        let mut cmsgs = [cmsg::Aligned(MaybeUninit::uninit()); BATCH_SIZE];
+        let mut cmsgs = [cmsg::Aligned([0u8; CMSG_LEN]); BATCH_SIZE];
```
Matthias247, line 149: FYI: This can't be uninit because it breaks `CMSG_NXTHDR`. The function will try to access the next cmsg in the buffer at https://github.com/rust-lang/libc/blob/ae65df55fe2d0e1348fed575bfc8c0c3e4744c65/src/unix/linux_like/linux/mod.rs#L2585-L2589

This will yield an undefined result, and can let `CMSG_NEXT` return `None` - which allows for no message except the first one to be written. A different approach is to  manually adjust the cmsg offset during encoding - but I think sticking to the official macros seems safer.

Matthias247: This was btw really painful to debug

quinn-proto/src/connection/mod.rs
```diff
@@ -362,6 +362,20 @@ where
             return None;
         }
 
+        let mut num_datagrams = 0;
+        // TODO: Check whether this can be constant or there are situations where
+        // we have to enforce less packets due to path limitations. Seems like
+        // we are beyond anti-amplification check this might already be good.
```
Ralith, line 368: The existing anti-amplification check assumes that we're sending at most one packet at a time, because it checks that we're at least one MTU away from violating the limit. To preserve it with this change, we could increment `self.total_sent` and re-check `self.anti_amplification_blocked()` after recording each packet into the GSO buffer before beginning the next.

quinn-proto/src/connection/mod.rs
```diff
@@ -362,6 +362,20 @@ where
...
+        //
+        // Also figure out how to make sure this is 1 on platforms which don't
+        // support GSO. Maybe it should be part of `ServerConfig`, and default
+        // to 1 on non-Linux. For Linux, should the kernel version and GSO
+        // support be detected or simply a minimum requirement?
+        // We can also keep setting the default on Linux to a reasonable number,
+        // and let applications overwrite if they run on an older kernel. Or
+        // let Quinn detect the Kernel version.
```
Ralith, line 376: Another option would be to take the maximum number of packets as an argument, though that could be relatively disruptive. That might be especially natural if we refactor to write packets into caller-supplied memory.

One way or another, it'd be nice to autodetect optimal behavior for most users. If UDP GSO support isn't near-universally deployed (isn't it pretty new?) we definitely shouldn't mandate it. There's also the possibility of batching outgoing packets unconditionally at this layer, and breaking it up later if the platform doesn't support it.

quinn-proto/src/connection/mod.rs
```diff
@@ -424,14 +448,23 @@ where
         };
 
         let mut buf = Vec::with_capacity(self.mtu as usize);
+        // Reserving capacity can provide more capacity than we asked for.
+        // However we are not allowed to write more than MTU size. Therefore
+        // the maximum capacity is tracked separately.
```
Ralith, line 453: This sounds like a bug in the existing code. Worth pulling out into a separate PR so we can merge ASAP?

quinn-proto/src/connection/mod.rs
```diff
@@ -500,9 +541,47 @@ where
...
+            // is better understood.
+            if num_datagrams == MAX_DATAGRAMS - 1 || space_id != SpaceId::Data {
+                concat_datagram = false;
+            }
+
+            if concat_datagram {
+                // Pad the packet to make it suitable for sending with GSO
+                // which will always send the maxium PDU.
+                // TODO: Can this `tag_len` be taken from the builder?
+                // This is currently copied from the `finish_packet` path.
```
Ralith, line 563: Stashing `tag_len` in the builder for easy access is fine. Or maybe it would be simpler of we redefined `min_size` to include the tag?

quinn-proto/src/connection/mod.rs
```diff
@@ -2645,7 +2760,13 @@ where
             .push_back(EndpointEventInner::NeedIdentifiers(n));
     }
 
-    fn populate_packet(&mut self, space_id: SpaceId, buf: &mut Vec<u8>) -> SentFrames {
+    fn populate_packet(
+        &mut self,
+        space_id: SpaceId,
+        buf: &mut Vec<u8>,
+        buf_capacity: usize,
+        has_more: &mut bool,
```
Ralith, line 2768: Rather than introducing this parameter, could we rerun the logic `poll_transmit` uses in `SpaceId::iter().filter(...)` to determine whether a space is ready to send as many times as necessary? This should give more precise, consistent results, and cause less churn.

quinn/src/platform/unix.rs
Ralith: Oh geez, nice catch!

Ralith (changes requested):
Awesome work so far! Made an initial high-level pass.

One missing consideration is [pacing](https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#name-pacing). In particular, we limit bursts to the initial congestion window. The current pacer tries to always send packets in such bursts to avoid thrashing, so this code is probably safe by coincidence, but to be robust against changes to the pacer to operate more smoothly and/or introduction of dynamic MTU detection, we should take care to ensure that a GSO batch does not exceed the pacing window.

> GSO is currently only used for data spaces, due to a less understood impact on other packet spaces and packet types.

What concerns do you have about other spaces? I discussed anti-amplification concerns below; due to 0-RTT I actually don't think restricting to Data space is sufficient regardless.

quinn-proto/src/connection/mod.rs
```diff
@@ -532,19 +613,46 @@ where
...
+            space_idx += 1;
         }
 
         if buf.is_empty() {
             return None;
         }
 
+        num_datagrams += 1;
+
         trace!("sending {} byte datagram", buf.len());
```
Ralith, line 652: Wording of this should be tweaked.

Matthias247:
> One missing consideration is pacing. In particular, we limit bursts to the initial congestion window. The current pacer tries to always send packets in such bursts to avoid thrashing, so this code is probably safe by coincidence, but to be robust against changes to the pacer to operate more smoothly and/or introduction of dynamic MTU detection, we should take care to ensure that a GSO batch does not exceed the pacing window.

Yes, I've seen the code for it - but I can't claim I fully understand the interactions with everything else.
I think however that this should be relatively safe, unless one sets an extreme GSO size. It would send a few packets more at once in a burst, but since more tokens are subtracted the next pause will also happen at the same time.  I don't think trying to pact at a granularity of < 10 packets makes sense in userspace. At 250MB/s we are sending about 250k packets/s, which means the there is a latency of 4us for 1 packet (or 40us for 10 packets). The precision in user-space - especially under load - will however likely be more in the 1ms region. Figuring out how to integrate kernel pacing (with SO_TXTIME) and how this interacts with a more granular application timer might be interesting here.

quinn-proto/src/connection/mod.rs
```diff
@@ -520,6 +599,8 @@ where
                     SentPacket {
                         acks: sent.acks,
                         time_sent: now,
+                        // TODO: Does hte extra padding we added have an impact
+                        // on congestion control?
```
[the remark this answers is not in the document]
Matthias247: ok. Seems like there is nothing extra to do here then, and the comment can be removed.

quinn-proto/src/connection/mod.rs
Matthias247: I indeed thought about letting the caller pass a limit. But it seemed like it would just move the setting around.

> There's also the possibility of batching outgoing packets unconditionally at this layer, and breaking it up later if the platform doesn't support it.

Yes. The endpoint layer could then do multiple `sendmsg` calls by selecting the sub-slices. But it gets a bit more tricky if it gets `EAGAIN`  in the middle of such a transmission => Requires additional offset tracking. Or putting everything into a `Bytes` instance and splitting the `Transmit` again. But not producing it in the first place might be easiest.

quinn-proto/src/connection/mod.rs
Matthias247: > Or maybe it would be simpler of we redefined min_size to include the tag?

I think this would just require the tag to be known in another place then.

Storing it in the builder sounds good to me. I just wasn't sure about all the places where it is created, and whether it's safe to store it there.

quinn-proto/src/connection/mod.rs
Ralith: > it seemed like it would just move the setting around.

My thinking is that it would allow the parameter to be internal to the `quinn` layer, rather than exposed in the configuration object as a value that the user should generally not be manually adjusting.

> But it gets a bit more tricky if it gets EAGAIN in the middle of such a transmission

Ah, right, buffering is annoying, especially since it risks sending outdated information compared to composing only exactly the packets that will be sent.

quinn-proto/src/connection/mod.rs
Matthias247: I don't think it's more precise. The false positives (like the reset stream) seem to happen in both cases. Maybe those should all be removed in separate cleanup efforts. I'm not sure whether false wakeups also lead to other unexpected side-effects (like sending more ACKs, which happened with the MAX_DATA frames).

But I agree using the same mechanism would be more streamlined.

quinn-proto/src/connection/mod.rs
Matthias247: what do you recommend?

Matthias247: Is this about the `trace!`?

Matthias247:
> What concerns do you have about other spaces? 

Another thing besides amplification was that other spaces might use different packet types, and the padding rules for all of them are not in my head at the moment. A brief glancy at the spec says that all packets besides short packets have explicit length information, and we can probably just pad all of them with zeros without any side effect. But I felt more comfortable with an initially reduced scope. Still captures the main source of multi-datagram transmissions - which is after the connection was established.

quinn-proto/src/connection/mod.rs
Ralith: Right. Moving it up into the loop would be fine, or adjusting it to something lke "sent {} {} byte datagrams".

quinn-proto/src/connection/mod.rs
Ralith: > I think this would just require the tag to be known in another place then.

It might reduce it to places where we're already fetching it anyway.

Ralith:
On review, I think the current code actually handles pacing just fine: we invoke `Pacer::delay` for each packet (rather than the whole batch), and its state is updated by `Connection::on_packet_sent`, which we also call per-packet.

quinn-proto/src/connection/mod.rs
Ralith: You're correct that there are still false positives due to reset streams (perhaps we could correct these cheaply by keeping a counter of reset streams that were previously pending, and comparing it to the length of the pending list?). However, I think `has_more` introduces false *negatives*, such as ignoring the application datagram queue. These could be fixed, but would represent a duplication of existing logic.

quinn-proto/src/connection/mod.rs
Matthias247: Sounds reasonable. Moved it into #954. Also the "cmsg" encoding bug. It doesn't hurt without this change, since only a single cmsg is encoded. But it also won't hurt to treat it independently from this feature.

Matthias247:
> On review, I think the current code actually handles pacing just fine: we invoke Pacer::delay for each packet (rather than the whole batch), and its state is updated by Connection::on_packet_sent, which we also call per-packet.

Yes. It should also place nice with the congestion controller due to this. All the interactions are within a loop. 

I however wondered if it also raises the 0-length datagram problem: I don't think so -> A new datagram is started, but we will not write any byte to it.  When returning it the total packet size will be an exact multiple of the segment size, so no additional empty datagram will be sent. However the `num_datagrams` calculation seems to be off by 1. Maybe this can all be made nicer.

Ralith:
Would it be sufficient to only increment `num_datagrams` immediately after caling `begin_packet`? At that point we're committed to sending a datagram.

Matthias247:
> Would it be sufficient to only increment `num_datagrams` immediately after caling `begin_packet`? At that point we're committed to sending a datagram.

Yes, this should work. Would however still be good to filter the empty packets.

Ralith:
Right, but that's an existing issue, and not a pressing one given that false positives from the can-send detection stuff should be very rare under normal conditions.

Matthias247:
This is still WIP, and misses some more changes regarding concatenation and respecting pacing and congestion-control before creating new packets.
However the following this are now changed based on feedback, and I wanted to get an idea whether that looks right:
- A common `space_can_send` method is used to determine whether there is more data to send
- The method do determine whether a packet needs to be padded was moved **inside** the packet creation loop
- `tag_len` is now stored in PacketBuilder, and reused on `populate_packet`.

Matthias247:
I now updated the change according to the initial discussion about keeping the last packet around until it is known whether we can append more data. This version also has a bit more of a linear flow, since we only check at a single place whether there is more data to send.

Unfortunately it also got a bit bigger than before.

quinn-proto/src/connection/mod.rs
```diff
@@ -477,78 +585,87 @@ where
...
             coalesce = coalesce && !builder.short_header;
 
-            let sent = if close {
+            sent_frames = if close_in_space {
                 trace!("sending CONNECTION_CLOSE");
                 match self.state {
                     State::Closed(state::Closed { ref reason }) => {
                         if space_id == SpaceId::Data {
-                            reason.encode(&mut builder.buffer, builder.max_size)
+                            reason.encode(&mut &mut buf, builder.max_size)
```
djc, line 609: Do we actually need `&mut &mut` here?

djc (commented):
Some smaller comments. I think the last commit here definitely needs to be split up more, it seems like we can do a bunch of the refactoring in incremental steps?

quinn-proto/src/connection/mod.rs
```diff
@@ -571,23 +689,38 @@ where
             } else {
                 None
             },
-            segment_size: None,
+            segment_size: if num_datagrams == 1 {
+                None
+            } else {
+                Some(self.path.mtu as usize)
+            },
```
djc, line 696: Just write this as a `match num_datagrams {}` expression, please.

quinn-proto/src/connection/mod.rs
Matthias247: We don't. Was a rename artifact - thanks for catching!

quinn-proto/src/connection/mod.rs
```diff
@@ -571,23 +689,38 @@ where
...
     }
 
+    /// Returns true if a space has outgoing data to send
+    fn space_can_send(&self, space_id: SpaceId) -> bool {
+        (self.spaces[space_id].crypto.is_some() && self.spaces[space_id].can_send())
+            || (space_id == SpaceId::Data
+                && ((self.spaces[space_id].crypto.is_some() && self.can_send_1rtt())
+                    || (self.zero_rtt_crypto.is_some()
+                        && self.side.is_client()
+                        && (self.spaces[space_id].can_send() || self.can_send_1rtt()))))
```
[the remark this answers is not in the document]
Matthias247: Well, this is just the unmodified original condition which was moved around. While I think it makes sense to modify it, we could also do it in a separate change

Matthias247:
> I think the last commit here definitely needs to be split up more, it seems like we can do a bunch of the refactoring in incremental steps?

I agree that it could. But since most of the changes are in the `poll_transmit` function this would make it harder to address any review feedback (everything needs to be fixed in intermediate commits as well as the final one). So there is this tradeoff too.

Overall I think the intermediate step is that some fields had been moved in the `PacketBuilder`. But not all of those moves make sense individually, since the responsibility for adding padding had been redefined. And I don't want to figure out for an unnecessary intermediate step.

quinn-proto/src/connection/mod.rs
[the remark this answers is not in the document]
djc: Ah, fair. Yeah, that could go in a separate change then.

djc:
I don't feel confident in my ability to review the last commit in its current state, so I do think there need to be some intermediate steps. Perhaps it makes sense to start moving fields into `PacketBuilder` one by one?

Matthias247:
I split it now into a lot of commits. I think this should now be reviewable

I however would love if we would avoid being overly nitpicky on individual things, because all renames and tiny other changes will now have an impact on all follow-ups. I'm happy to apply further changes later on, but would like to avoid having to change 5 commits for the sake of a style discussion.

quinn-proto/src/connection/mod.rs
```diff
@@ -402,67 +405,166 @@ where
...
+                    } as u64;
+                    debug_assert!(untracked_bytes <= self.path.mtu as u64);
+
+                    let bytes_to_send = u64::from(self.path.mtu) + untracked_bytes;
+
+                    if self.congestion_blocked(bytes_to_send) {
+                        space_idx += 1;
                         congestion_blocked = true;
+                        // TODO:Should those rather be break?
+                        // Maybe not due to loss probes
```
Ralith, line 515: I think this comment is correct as we might have a loss probe in a higher space, so let's rephrase it as an explanation.

quinn-proto/src/connection/mod.rs
```diff
@@ -402,67 +405,166 @@ where
...
                         self.path.mtu,
-                        window,
+                        self.path.congestion.window(),
                         now,
                     ) {
                         self.timers.set(Timer::Pacing, delay);
+                        space_idx += 1;
                         congestion_blocked = true;
+                        // TODO:Should those rather be break?
+                        // Maybe not due to loss probes
```
Ralith, line 532: Loss probes should be subject to pacing, since unlike congestion control, pacing relaxes unilaterally. Therefore, let's turn this into a break.

quinn-proto/src/connection/mod.rs
```diff
@@ -402,67 +405,166 @@ where
...
+            } else {
+                // We can append/coalesce the next packet into the current
+                // datagram.
+                // Finish current packet without adding extra padding
+                if let Some(builder) = builder.take() {
+                    self.finish_and_track_packet(now, builder, sent_frames.take(), &mut buf);
+                }
             }
 
+            debug_assert!(buf_capacity - buf.len() >= MIN_PACKET_SPACE);
```
Ralith, line 566: Should this be `==` (or rather, `debug_assert_eq`)? We seem to be inferring the size limit for the current packet from `buf_capacity` in various places.

quinn-proto/src/connection/mod.rs
```diff
@@ -479,17 +581,23 @@ where
...
+            builder = Some(self.begin_packet(
                 now,
                 space_id,
                 &mut buf,
                 buf_capacity,
                 (num_datagrams - 1) * (self.path.mtu as usize),
                 ack_eliciting,
-            )?;
+            )?);
+            let builder = builder.as_mut().unwrap();
```
Ralith, line 597: Style nit: could we use [`Option::get_or_insert_with`](https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert_with) to avoid the unwrap here?

quinn-proto/src/connection/mod.rs
```diff
@@ -520,6 +628,8 @@ where
                     min_datagram_size,
                 );
                 coalesce = false;
+                // We don't want to send 2 close packets
+                close = false;
```
Ralith, line 632: I suspect this is the right place to unset `self.close`, as well. Preexisting issue, though, no need to address it in this PR.

quinn-proto/src/connection/mod.rs
```diff
@@ -542,15 +652,16 @@ where
                 }
             }
 
+            // Don't increment space_idx.
+            // We stay in the current space and check if there is more data to send.
+        }
+
+        // Finish the last packet. This one might also need to be padded
+        if let Some(mut builder) = builder.take() {
```
Ralith, line 660: nit: Is `take()` needed here, or can we move out of `builder` completely?

quinn-proto/src/connection/mod.rs
```diff
@@ -542,15 +652,16 @@ where
                 }
             }
 
+            // Don't increment space_idx.
+            // We stay in the current space and check if there is more data to send.
+        }
+
+        // Finish the last packet. This one might also need to be padded
```
Ralith, line 659: It doesn't look like we're making a special effort to pad here, and calling out the usual machinery at one specific point seems a little odd.

Ralith (changes requested):
Thanks for breaking this up! This has made it *much* easier to review. `poll_transmit` is a bit of a monolith, but between the detailed comments introduced and the independent changes being split out, this was a pleasure to review.

Overall this looks great. Most issues are cosmetic and non-blocking, though I think pacing of loss probes is required for correctness.

If it's not too difficult, I would like to see the "Indicate whether a packet requires padding in `SentFrames`" and "Use padding information from populate_packet" commits folded together, since the former doesn't stand on its own very well. Not the end of the world for me if that's just too painful a rebase, though.

quinn-proto/src/connection/mod.rs
```diff
@@ -748,6 +731,41 @@ where
         (buffer.len() - encode_start, pad)
     }
 
+    fn finish_and_track_packet(
```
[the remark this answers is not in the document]
Matthias247: It could. But then it would be concerned with more than it is currently (at the moment it just builds a packet and doesn't care about congestion control - that would change). I think it's better the way it is now than bloating up the other method.

quinn-proto/src/connection/mod.rs
```diff
@@ -514,35 +514,18 @@ where
...
-                            0
-                        },
-                        ack_eliciting,
-                        retransmits: sent.retransmits,
-                        stream_frames: sent.stream_frames,
-                    },
-                );
             }
 
+            self.finish_and_track_packet(now, builder, sent_frames.take(), &mut buf);
```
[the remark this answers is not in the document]
Matthias247: Yes. In this commit. But not anymore after the factor to GSO, since the `Option` is kept around outside the send loop for the case a previous packet needs to be finalized.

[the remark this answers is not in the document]
Matthias247: Will be fixed in the next intermediate commit

quinn-proto/src/connection/mod.rs
Matthias247: Right, we can move. We just need ´.take()` for the previous iterations to make the borrow checker happy. Same applies for `sent_frames`

quinn-proto/src/connection/mod.rs
Matthias247: Why should it be? If we are starting a new datagram we have lots of space. But the same branch is also taken in the coalescing case, when the current datagram already contains a packet

quinn-proto/src/connection/mod.rs
Matthias247: makes sense. Done

quinn-proto/src/connection/mod.rs
Ralith: Er, right.

quinn-proto/src/connection/mod.rs
Matthias247: Updated it. But I'm not really 100% convinced if it should. Pacing is some kind of averaging congestion control. And if we claim loss probes are not subject to congestion control it doesn't seem true.

On the other hand I don't think this really matters in practice, since the initial and crypto space will be barely congestion or pacing blocked, and therefore the branches will only be taken for the application data space.

quinn-proto/src/connection/mod.rs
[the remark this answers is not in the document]
Ralith: Fair enough.

quinn-proto/src/connection/mod.rs
Matthias247: Since `begin_packet` is fallible and can't be called inside the closure it would be something along:

```rust
let next_builder = self.begin_packet(
                now,
                space_id,
                &mut buf,
                buf_capacity,
                (num_datagrams - 1) * (self.path.mtu as usize),
                ack_eliciting,
            )?;
            let builder = builder.get_or_insert_with(|| next_builder);
```

Which hides a bit the case that should never happen of "what is if `builder` wasn't None before".

I'm rather indifferent on whether its better.
https://doc.rust-lang.org/std/option/enum.Option.html#method.insert seems to be ideal. But isn't stable yet.

Matthias247: Changed now to

```rust
let builder = builder.get_or_insert(self.begin_packet(
                now,
                space_id,
                &mut buf,
                buf_capacity,
                (num_datagrams - 1) * (self.path.mtu as usize),
                ack_eliciting,
            )?);
```

quinn-proto/src/connection/mod.rs
```diff
@@ -402,67 +405,166 @@ where
...
+                if space_id != self.highest_space {
+                    // We ignore data in this space, since the close message
+                    // has higher priority
+                    space_idx += 1;
+                    continue;
+                }
+                true
+            } else {
+                false
+            };
```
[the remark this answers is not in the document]
Matthias247: Indeed! Changed for that.

And probably we can directly refer to `self.close` there in the future. However I would like to defer that fix for the future.

quinn-proto/src/connection/mod.rs
Matthias247: Right, let's address this later. Not sure if it's sufficient here, because it could get lost on some early returns. We might rather need to keep `self.close` set around until it is really sent.

Matthias247:
Oops, this wasn't intended to be closed. Wrong command on git push.

Matthias247:
>  If it's not too difficult, I would like to see the "Indicate whether a packet requires padding in SentFrames" and "Use padding information from populate_packet" commits folded together, since the former doesn't stand on its own very well.

Seems like we had been lucky and at least moving the commit into one direction and merging worked without conflicts!  Updated

PR #955, opened 2020-12-31
Different way of factoring out read logic

djc:
Here's a proposal for a different way to handle the abstraction (feel free to fold this into your second commit).

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -889,7 +889,7 @@ where
     pub fn read_unordered(&mut self, id: StreamId) -> Result<Option<(Bytes, u64)>, ReadError> {
         Ok(self.streams.read_unordered(id)?.map(|result| {
             self.add_read_credits(id, result.max_stream_data, result.max_data);
-            (result.buf, result.offset)
+            result.result
         }))
     }
 
@@ -897,7 +897,7 @@ where
     pub fn read_chunk(&mut self, id: StreamId) -> Result<Option<Bytes>, ReadError> {
         Ok(self.streams.read_chunk(id)?.map(|result| {
             self.add_read_credits(id, result.max_stream_data, result.max_data);
-            result.buf
+            result.result
         }))
     }
 
@@ -909,7 +909,7 @@ where
     ) -> Result<Option<usize>, ReadError> {
         Ok(self.streams.read_chunks(id, bufs)?.map(|result| {
             self.add_read_credits(id, result.max_stream_data, result.max_data);
-            result.n_bufs
+            result.result.bufs
         }))
     }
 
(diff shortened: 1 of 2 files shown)


quinn-proto/src/connection/assembler.rs
```diff
@@ -59,6 +59,37 @@ impl Assembler {
...
+
+        loop {
+            let chunk = match self.data.peek() {
+                Some(chunk) => chunk,
+                None => return Ok(None),
+            };
+
+            if chunk.offset > self.bytes_read {
+                // Next chunk is after current read index
+                return Ok(None);
```
Matthias247, line 76: Isn't `None` end of stream? i'm not sure how this function is used, but it might need to return a blocked status instead.

Ralith: `streams::Recv::read_chunk` translates `None` into an error if appropriate after calling this.

djc:
I still think this makes for a nice improvement, but happy to drop it if others disagree.

djc:
(Now targeting main.)

quinn-proto/src/connection/mod.rs
```diff
@@ -889,15 +889,15 @@ where
     pub fn read_unordered(&mut self, id: StreamId) -> Result<Option<(Bytes, u64)>, ReadError> {
         Ok(self.streams.read_unordered(id)?.map(|result| {
             self.add_read_credits(id, result.max_stream_data, result.max_data);
-            (result.buf, result.offset)
+            (result.result.0, result.result.1)
```
stammw, line 892: Desctructuring `result` into `bytes` and `offset` would make this line a bit clearer.

quinn-proto/src/connection/streams.rs
```diff
@@ -1156,30 +1087,31 @@ impl Recv {
...
-    fn read_chunk(&mut self) -> Result<Option<Bytes>, ReadError> {
+    fn read_chunk(&mut self) -> StreamReadResult<Bytes> {
         match self.assembler.read_chunk()? {
             Some(bytes) => Ok(Some(bytes)),
             None => self.read_blocked().map(|()| None),
         }
     }
 
-    fn read_chunks(&mut self, chunks: &mut [Bytes]) -> Result<Option<usize>, ReadError> {
+    fn read_chunks(&mut self, chunks: &mut [Bytes]) -> Result<Option<(usize, usize)>, ReadError> {
```
stammw, line 1097: A comment to disambiguate which of the two returned `usize` is the number of chunks would help here.

quinn-proto/src/connection/mod.rs
djc: I ended up changing it to just returning `result.result`, as in the other methods. In that case, just naming the values in it feels a little superfluous, since they immediately get repacked into a tuple again.

quinn-proto/src/connection/streams.rs
djc: Yeah, this was a little opaque. I instead defined a simple struct `ReadChunks`, see what you think.

stammw (approved):

PR #956, opened 2020-12-31
Apply clippy fixes for 1.49

djc:


Ralith (approved):
So far so good, though it looks like there's more.

Ralith (approved):

PR #957, opened 2020-12-31
Make connection stats public

Matthias247:
So far those haven't been consumable by applications.
This makes the struct public, and adds an accessor
for `quinn_proto::Connection` and `quinn::Connection`.


djc (approved):
LGTM!

Ralith (approved):
Thanks!

PR #960, opened 2020-12-31
GSO platform support

Matthias247:
This splits out the platform/socket parts for UDP GSO support
from #953 to reduce the amount of code to review.

This change mainly implements setting the segment size
socket option, and adds a runtime detection mechanism for
GSO support.

### DIFF
--- a/quinn/src/platform/unix.rs
+++ b/quinn/src/platform/unix.rs
@@ -7,10 +7,11 @@ use std::{
     ptr,
 };
 
+use lazy_static::lazy_static;
 use mio::net::UdpSocket;
 use proto::{EcnCodepoint, Transmit};
 
-use super::cmsg;
+use super::{cmsg, UdpCapabilities, UdpExt};
 use crate::udp::RecvMeta;
 
 #[cfg(target_os = "freebsd")]
@@ -18,7 +19,7 @@ type IpTosTy = libc::c_uchar;
 #[cfg(not(target_os = "freebsd"))]
 type IpTosTy = libc::c_int;
 
-impl super::UdpExt for UdpSocket {
+impl UdpExt for UdpSocket {
     fn init_ext(&self) -> io::Result<()> {
         // Safety
         assert_eq!(
@@ -267,6 +268,11 @@ impl super::UdpExt for UdpSocket {
     }
 }
 
+/// Returns the platforms UDP socket capabilities
+pub fn caps() -> UdpCapabilities {
+    *CAPABILITIES
+}
+
 const CMSG_LEN: usize = 64;
 
 fn prepare_msg(
@@ -296,6 +302,16 @@ fn prepare_msg(
     } else {
         encoder.push(libc::IPPROTO_IPV6, libc::IPV6_TCLASS, ecn);
     }
+
+    if let Some(segment_size) = transmit.segment_size {
+        debug_assert!(
+            caps().gso,
+            "Platform must support GSO for setting segment size"
+        );
+
+        gso::set_segment_size(&mut encoder, segment_size as u16);
+    }
+
     encoder.finish();
 }
 
@@ -373,3 +389,56 @@ pub const BATCH_SIZE: usize = 32;
 
 #[cfg(any(target_os = "macos", target_os = "ios"))]
 pub const BATCH_SIZE: usize = 1;
+
+#[cfg(target_os = "linux")]
+mod gso {
+    use super::*;
+
+    /// Checks whether GSO support is available by setting the UDP_SEGMENT
+    /// option on a socket
+    pub fn supports_gso() -> bool {
+        const GSO_SIZE: libc::c_int = 1500;
+
+        let socket = match std::net::UdpSocket::bind("[::]:0") {
+            Ok(socket) => socket,
+            Err(_) => return false,
+        };
+
+        let rc = unsafe {
+            libc::setsockopt(
+                socket.as_raw_fd(),
+                libc::SOL_UDP,
+                libc::UDP_SEGMENT,
+                &GSO_SIZE as *const _ as _,
+                mem::size_of_val(&GSO_SIZE) as _,
+            )
+        };
+
+        rc != -1
+    }
+
+    pub fn set_segment_size(encoder: &mut cmsg::Encoder, segment_size: u16) {
+        encoder.push(libc::SOL_UDP, libc::UDP_SEGMENT, segment_size);
+    }
+}
+
+#[cfg(not(target_os = "linux"))]
+mod gso {
+    use super::*;
+
+    pub fn supports_gso() -> bool {
+        false
+    }
+
+    pub fn set_segment_size(_encoder: &mut cmsg::Encoder, _segment_size: u16) {
+        panic!("Setting a segment size is not supported on current platform");
+    }
+}
+
+lazy_static! {
+    static ref CAPABILITIES: UdpCapabilities = {
+        UdpCapabilities {
+            gso: gso::supports_gso(),
+        }
+    };
+}
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -379,6 +379,7 @@ where
                     destination,
                     contents: buf,
                     ecn: None,
+                    segment_size: None,
                 });
             }
         }
@@ -559,6 +560,7 @@ where
             } else {
                 None
             },
+            segment_size: None,
         })
     }
 
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -189,6 +189,7 @@ where
                         destination: remote,
                         ecn: None,
                         contents: buf,
+                        segment_size: None,
                     });
                     return None;
                 }
@@ -334,6 +335,7 @@ where
             destination: remote,
             ecn: None,
             contents: buf,
+            segment_size: None,
         });
     }
 
@@ -587,6 +589,7 @@ where
                     destination: remote,
                     ecn: None,
                     contents: buf,
+                    segment_size: None,
                 });
                 return None;
             }
@@ -679,6 +682,7 @@ where
             destination,
             ecn: None,
             contents: buf,
+            segment_size: None,
         })
     }
 
--- a/quinn-proto/src/lib.rs
+++ b/quinn-proto/src/lib.rs
@@ -274,6 +274,9 @@ pub struct Transmit {
     pub ecn: Option<EcnCodepoint>,
     /// Contents of the datagram
     pub contents: Vec<u8>,
+    /// The segment size if this transmission contains multiple datagrams.
+    /// This is `None` if the transmit only contains a single datagram
+    pub segment_size: Option<usize>,
 }
 
 //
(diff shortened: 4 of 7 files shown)


quinn/src/platform/unix.rs
```diff
@@ -20,6 +21,8 @@ type IpTosTy = libc::c_int;
 
 impl super::UdpExt for UdpSocket {
     fn init_ext(&self) -> io::Result<()> {
+        Self::caps();
+
```
Ralith, line 25: What's this doing here?

Ralith (commented):
Rather than relying on `lazy_static` to hide state behind the scenes, could we pass a `UdpCapabilities` in to `send_ext` and store it in the `quinn::Endpoint`?

quinn/src/platform/unix.rs
Matthias247: good catch! I added this just for testing to get it to be actually used and get a debug output. Will remove

Matthias247:
> Rather than relying on lazy_static to hide state behind the scenes, could we pass a UdpCapabilities in to send_ext and store it in the quinn::Endpoint?

Well, `send_ext` doesn't really require it, apart from the `debug_assert`. If the packet had been produced with GSO support, we can only drop it (or crash). It could definitely be stored in `quinn::Endpoint` too. However I think even if we do this, it's ok to cache the result globally (which is what lazy-static does here).

quinn/src/platform/unix.rs
```diff
@@ -296,6 +301,26 @@ fn prepare_msg(
...
+    if let Some(segment_size) = transmit.segment_size {
+        debug_assert!(
+            U::caps().gso,
+            "Platform must support GSO for setting segment size"
+        );
+
+        #[cfg(target_os = "linux")]
+        fn set_segment_size(encoder: &mut cmsg::Encoder, segment_size: u16) {
+            encoder.push(libc::SOL_UDP, libc::UDP_SEGMENT, segment_size);
```
Ralith, line 313: Doesn't this need to check whether GSO is actually supported, rather than running unconditionally on linux?

Ralith:
> it's ok to cache the result globally (which is what lazy-static does here).

Yeah, it's definitely not a correctness issue, I just prefer to avoid global state when it's not absolutely necessary. Not a huge deal here since it's immutable.

quinn/src/platform/unix.rs
Matthias247: `sendmsg` will check it and return an error if the cmsg is not supported

Matthias247: and there is the `debug_assert`.  Could be a real assert or an error too. But I think leaving it out in release mode is ok, since this is definitely a config issue.

Ralith: Won't that fail the whole operation?

Ralith: Oh, right, this is all guarded on `transmit.segment_size` to begin with.

quinn/src/platform/unix.rs
```diff
@@ -265,11 +266,15 @@ impl super::UdpExt for UdpSocket {
...
+
+    fn caps() -> super::UdpCapabilities {
+        *CAPABILITIES
+    }
 }
 
 const CMSG_LEN: usize = 64;
 
-fn prepare_msg(
+fn prepare_msg<U: super::UdpExt>(
```
Ralith, line 277: If `caps()` became a free function in `platform` (e.g. `pub use imp::udp_caps;`) we could avoid the generic here.

Ralith: Actually, if we want to commit to the global state, we could even just have a `static CAPABILITIES` exported directly from `platform`.

quinn/src/platform/unix.rs
```diff
@@ -296,6 +301,26 @@ fn prepare_msg(
     } else {
         encoder.push(libc::IPPROTO_IPV6, libc::IPV6_TCLASS, ecn);
     }
+
+    if let Some(segment_size) = transmit.segment_size {
+        debug_assert!(
+            U::caps().gso,
+            "Platform must support GSO for setting segment size"
+        );
```
Ralith, line 309: I/O errors currently manifest as an error-level tracing event followed by the endpoint driver halting, which can be very hard to diagnose if the application doesn't have tracing wired up. I'm not *super* worked up about being maximally helpful if somebody explicitly mangles their configuration, but arranging things such that `segment_size` is statically guaranteed to be unset if `caps().gso` is false would be more robust.

Matthias247: I agree. I noticed that the IO errors are not super visible while implementing the other cmsg stuff.
However I don't see a super good way to have the static guarantees here. The messages are produced by some other code block. And after `poll_transmit` has produced such a message there is not we can do anymore besides failing (in one or the other way).

The main checks that we should look out for is that the max datagram amount which will get passed to `poll_transmit` is only 1 if feature isn't supported. However this should be in another place.

quinn/src/platform/unix.rs
Matthias247: I agree, it's not strictly necessary to be on this trait. I can however see one reason why it might stay there: In case there are multiple IO providers for a single platform (e.g. lets say one also implements raw socket support or XDP support for Linux).  
But maybe that should then be in a different platform module?

Regarding function vs public static ref: Static ref seems easier. Doesn't however allow to e.g. have a `const` impl on platforms which don't require runtime checks. But the difference between those seems totally negligible given it's a 1-time thing. The other difference I can see is that with the static ref this detail leaks out of the module.

Ralith: On consideration, I'm weakly in favor of a function since it's totally opaque that way, bearing in mind that we shouldn't have a `const` impl on any platform because that would constitute a cfg-dependent signature which is a huge landmine.

Multiple dynamically-selectable backends on a single platform are an interesting idea, but aren't something we currently support whatsoever, and I don't think this would have a meaningful impact on the amount of work needed to make that happen.

quinn/src/platform/unix.rs
Ralith: > The main checks that we should look out for is that the max datagram amount which will get passed to poll_transmit is only 1 if feature isn't supported

If we'll have a mechanism like that in place, then this error condition is no longer (theoretically) reachable via user misconfiguration, in which case a debug_assert is a perfectly adequate second line of defense.

quinn/src/platform/unix.rs
Matthias247: I switched things for function

Ralith (dismissed):

quinn/src/platform/unix.rs
```diff
@@ -373,3 +399,40 @@ pub const BATCH_SIZE: usize = 32;
 
 #[cfg(any(target_os = "macos", target_os = "ios"))]
 pub const BATCH_SIZE: usize = 1;
+
+#[cfg(target_os = "linux")]
+lazy_static! {
+    static ref CAPABILITIES: super::UdpCapabilities = {
+        /// Checks whether GSO support is available by setting the UDP_SEGMENT
+        // option on a socket.
```
djc, line 407: I guess you meant to continue the docstring here? Let's remove the period at the end.

quinn/src/platform/unix.rs
```diff
@@ -267,6 +268,11 @@ impl super::UdpExt for UdpSocket {
     }
 }
 
+/// Returns the platforms UDP socket capabilities
+pub fn caps() -> super::UdpCapabilities {
```
djc, line 272: Let's just import `UdpCapabilities` at the module level? All the `super::` prefixes get pretty tedious.

quinn/src/platform/unix.rs
```diff
@@ -373,3 +399,40 @@ pub const BATCH_SIZE: usize = 32;
...
+                )
+            };
+
+            rc != -1
+        }
+
+        super::UdpCapabilities {
+            gso: supports_gso(),
+        }
+    };
```
djc, line 432: Could we do this more compactly, like this?

```rust
static ref CAPABILITIES: UdpCapabilities = UdpCapabilities {
    gso: match std::net::UdpSocket::bind("[::]:0") {
        Err(_) => false,
        Ok(socket) => {
             const GSO_SIZE: libc::c_int = 1500;
             unsafe {
                libc::setsockopt(
                    socket.as_raw_fd(),
                    libc::SOL_UDP,
                    libc::UDP_SEGMENT,
                    &GSO_SIZE as *const _ as _,
                    mem::size_of_val(&GSO_SIZE) as _,
                )
            } != -1
        }
    },
};
```

Personally I'm not really a fan of nested functions.

djc (commented):
Some nits, otherwise this looks good.

quinn/src/platform/unix.rs
djc: (Only relevant if we keep the inline function.)

quinn/src/platform/unix.rs
Matthias247: I don't like this one. If we add more capabilities, then this will get a super long block. Imagine having 3 unrelated capabilities here. And the right drift due to the lack of early returns also doesn't make it nicer.

I'm fine with making it either a top level function or move it into a `mod linux`. The strongest reason it is a nested function at the moment is that it avoids requiring another `cfg(target_os)` block, since it is already guarded by the surrounding one.

quinn/src/platform/unix.rs
Matthias247: sure. Do you prefer relative `super::` imports for that or absolute `crate::` imports?

djc: I think `super` makes sense in this case.

Ralith (approved):

quinn/src/platform/unix.rs
djc: I think a `mod linux` would be better, which should then also contain `set_segment_size()`.

djc: (Sorry for the slow turnaround on this.)

Matthias247: Moved it into `mod linux`. Not sure about `set_segment_size`. We would then also a `mod not_linux`, and be able to dispatch on the same name. I feel like this is more weird than worthwhile.

djc: What if we give the module a name that's more about the contained functionality? Then it would be pretty natural, I think? Like maybe `mod gso` or `mod segment` (with `set_size()` and `is_supported()` in it), with one definition for Linux and one definition for other platforms.

quinn/src/platform/unix.rs
```diff
@@ -373,3 +389,56 @@ pub const BATCH_SIZE: usize = 32;
...
+    use super::*;
+
+    pub fn supports_gso() -> bool {
+        false
+    }
+
+    pub fn set_segment_size(_encoder: &mut cmsg::Encoder, _segment_size: u16) {
+        panic!("Setting a segment size is not supported on current platform");
+    }
+}
```
Ralith, line 436: I like this more than I expected. Might want to generalize it in the future, but that's for then.

Ralith (approved):

quinn/src/platform/unix.rs
Matthias247: I updated for this. Looks reasonable to me.

djc:
Looks great, thanks!

PR #961, opened 2021-01-01
Limit concurrent streams rather than accept queue size

Ralith:
See also #962.

### DIFF
--- a/quinn-h3/benches/helpers.rs
+++ b/quinn-h3/benches/helpers.rs
@@ -44,7 +44,7 @@ impl Bench {
         let cert_chain = quinn::CertificateChain::from_certs(vec![cert.clone()]);
 
         let mut transport = quinn::TransportConfig::default();
-        transport.stream_window_bidi(102_400).unwrap();
+        transport.max_concurrent_bidi_streams(102_400).unwrap();
         let mut server_config = quinn::ServerConfig::default();
         server_config.transport = Arc::new(transport);
         let mut server_config = ServerConfigBuilder::new(server_config);
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -225,8 +225,8 @@ impl Default for TransportConfig {
         const STREAM_RWND: u32 = MAX_STREAM_BANDWIDTH / 1000 * EXPECTED_RTT;
 
         TransportConfig {
-            max_concurrent_bidi_streams: 32u32.into(),
-            max_concurrent_uni_streams: 32u32.into(),
+            max_concurrent_bidi_streams: 100u32.into(),
+            max_concurrent_uni_streams: 100u32.into(),
             max_idle_timeout: Some(Duration::from_millis(10_000)),
             stream_receive_window: STREAM_RWND.into(),
             receive_window: VarInt::MAX,
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -15,20 +15,18 @@ use crate::{
 /// Parameters governing the core QUIC state machine
 ///
 /// Default values should be suitable for most internet applications. Applications protocols which
-/// forbid remotely-initiated streams should set `stream_window_bidi` and `stream_window_uni` to
-/// zero.
+/// forbid remotely-initiated streams should set `max_concurrent_bidi_streams` and
+/// `max_concurrent_uni_streams` to zero.
 ///
 /// In some cases, performance or resource requirements can be improved by tuning these values to
-/// suit a particular application and/or network connection. In particular, window sizes for
-/// streams, stream data, and overall connection data can be tuned for a particular expected round
-/// trip time, link capacity, memory availability, and rate of stream creation. Tuning for higher
-/// bandwidths and latencies increases worst-case memory consumption, but does not impair
+/// suit a particular application and/or network connection. In particular, data window sizes can be
+/// tuned for a particular expected round trip time, link capacity, and memory availability. Tuning
+/// for higher bandwidths and latencies increases worst-case memory consumption, but does not impair
 /// performance at lower bandwidths and latencies. The default configuration is tuned for a 100Mbps
-/// link with a 100ms round trip time, with remote endpoints opening at most 320 new streams per
-/// second.
+/// link with a 100ms round trip time.
 pub struct TransportConfig {
-    pub(crate) stream_window_bidi: VarInt,
-    pub(crate) stream_window_uni: VarInt,
+    pub(crate) max_concurrent_bidi_streams: VarInt,
+    pub(crate) max_concurrent_uni_streams: VarInt,
     pub(crate) max_idle_timeout: Option<Duration>,
     pub(crate) stream_receive_window: VarInt,
     pub(crate) receive_window: VarInt,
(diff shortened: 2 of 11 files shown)


quinn-proto/src/config.rs
```diff
@@ -237,8 +225,8 @@ impl Default for TransportConfig {
         const STREAM_RWND: u32 = MAX_STREAM_BANDWIDTH / 1000 * EXPECTED_RTT;
 
         TransportConfig {
-            stream_window_bidi: 32u32.into(),
-            stream_window_uni: 32u32.into(),
+            max_concurrent_bidi_streams: 1024u32.into(),
```
Matthias247, line 228: 1024 is rather high. The config allows a receive buffer size of 1.25MB per stream -> This allows for a memory usage of more than 1GB on reception, in case the application can't keep up (it hopefully can, since the whole thing is limited by the speed of a single connection).

I nevertheless think something between the old 32 and maybe around 100 would make for a reasonable default config.

quinn-proto/src/connection/mod.rs
```diff
@@ -886,18 +885,34 @@ where
     ///
     /// The return value if `Ok` contains the bytes and their offset in the stream.
     pub fn read_unordered(&mut self, id: StreamId) -> Result<Option<(Bytes, u64)>, ReadError> {
-        Ok(self.streams.read_unordered(id)?.map(|result| {
+        let result = self.streams.read_unordered(id)?.map(|result| {
             self.add_read_credits(id, result.max_stream_data, result.max_data);
             (result.buf, result.offset)
-        }))
+        });
+        if self.streams.take_max_streams_dirty(id.dir()) {
```
Matthias247, line 892: Any issues regarding the early returns of `read` via `?`

I think it might, because there seem occasions where streams are removed on error cases.

Matthias247: And consider using `WantTransmit` for those too.

quinn-proto/src/connection/streams.rs
```diff
@@ -70,6 +70,13 @@ pub struct Streams {
     send_window: u64,
     /// Configured upper bound for how much unacked data the peer can send us per stream
     stream_receive_window: u64,
+    /// Whether the corresponding `max_remote` has increased
+    max_streams_dirty: [bool; 2],
```
Matthias247, line 74: This could also use the same mechanism as `max_data` and thereby be `sent_max_data`

quinn-proto/src/connection/streams.rs
```diff
@@ -194,10 +211,12 @@ impl Streams {
             }
             Ok(None) => {
                 entry.remove_entry();
+                self.stream_freed(id, StreamHalf::Recv);
```
Matthias247, line 214: would be nice to somehow combine this with the entry, so that it's harder to miss some places which remove one thing but not the other.

Matthias247 (commented):
Functionally I think this is 100% the right thing to do! Implementation wise I think it could be nice if it would be harder to miss one some of the updates. However since this wasn't ideal before, it shouldn't block this improvement.

quinn-proto/src/config.rs
Ralith: Good catch, didn't really think that one through.

quinn-proto/src/connection/mod.rs
Ralith: Good catch! `ReadError::Reset` indeed can cause stream state to be discarded. `ReadResult` is only available in successful cases, so reusing the `WantTransmit` pattern would require some shuffling around; I'm open to that but I'd like to hold off until after #952 to avoid forcing it to be painfully rebased.

quinn-proto/src/connection/streams.rs
Ralith: I agree in principle, but the obvious approach would involve a newtyped wrapper around the `hash_map::Entry` that captures `self` and therefore makes borrowck sad. We could address that by encapsulating the hash tables along with the set of things `stream_freed` might transitively touch, but it's not clear to me if there's a natural boundary to be drawn there.

quinn-proto/src/connection/streams.rs
Ralith: Punting on this until after #952 due to the error case as discussed in https://github.com/quinn-rs/quinn/pull/961#discussion_r550723332.

quinn-h3/benches/helpers.rs
```diff
@@ -44,7 +44,7 @@ impl Bench {
         let cert_chain = quinn::CertificateChain::from_certs(vec![cert.clone()]);
 
         let mut transport = quinn::TransportConfig::default();
-        transport.stream_window_bidi(102_400).unwrap();
+        transport.max_concurrent_bidi_streams(102_400).unwrap();
```
Matthias247, line 47: Btw: I noticed that the old setting had kind of a memory problem: All the streams where pre-allocated, and if one used a high window (e.g. in order to let a different mechanism take care of limiting), the system run out of memory.

I guess this isn't resolved yet bis this change (which is totally fair). but also something that might be worthwhile to look into.

quinn-proto/src/connection/mod.rs
```diff
@@ -2575,6 +2583,18 @@ where
             }
         }
 
+        // Issue stream ID credit due to ACKs of outgoing finish/resets and incoming finish/resets
```
Matthias247, line 2586: Are early returns an issue with having it here? I guess likely not, because those are all connection errors and in that case there won't be new streams anyway?

quinn-h3/benches/helpers.rs
Ralith: What different mechanism could be used without exposing an endpoint to that same worst-case memory allocation under certain legal peer behavior?

Matthias247: It's definitely the same worst-case allocation. But similar as with receive and transmit buffers the worst-case rarely happens, and services run under some kind of amortized expectations. 

If 95% of connections use only 10 streams, a nice amount of memory could be saved by not allocating the maximum upfront.

Ralith: The set of unopened but permitted remotely-initiated streams is sparse, so we'd need a `HashSet` or something for that. `size_of::<(Send, Recv)>()` = 248 is also a lot smaller than a typical I/O buffer. The savings could still be significant (16x?) for really huge stream budgets, I suppose, but I'm suspicious of the use case for such a budget.

quinn-proto/src/connection/mod.rs
Ralith: Correct; any early return here is a transport error, and transport errors always immediately terminate the connection, rendering flow control moot.

quinn-h3/benches/helpers.rs
Ralith: I see your point that it might matter more with many connections.

quinn-proto/src/connection/mod.rs
```diff
@@ -887,18 +886,16 @@ where
     ///
     /// The return value if `Ok` contains the bytes and their offset in the stream.
     pub fn read_unordered(&mut self, id: StreamId) -> Result<Option<(Bytes, u64)>, ReadError> {
-        Ok(self.streams.read_unordered(id)?.map(|result| {
-            self.add_read_credits(id, result.max_stream_data, result.max_data);
-            result.result
-        }))
+        let result = self.streams.read_unordered(id);
+        self.post_read(id, result.as_ref().ok().map(|x| x.as_ref()).flatten());
```
djc, line 890: Could we move more of the repetitive stuff into `post_read()` here?

quinn-proto/src/connection/streams.rs
```diff
@@ -884,6 +892,17 @@ impl Streams {
 
     /// Update counters for removal of a stream
     fn stream_freed(&mut self, id: StreamId, half: StreamHalf) {
+        let remote = id.initiator() != self.side;
+        if remote {
```
djc, line 896: Change this to `if id.initiator() != self.side`?

quinn-proto/src/connection/streams.rs
djc: Do you want to do this as part of this PR?

Ralith: Because these frames can become queued due to receiving packets, this will be a bit fiddly to pursue. My inclination is not to block the behavioral fix on that.

quinn-proto/src/connection/mod.rs
Ralith: Good call; done.

quinn-proto/src/connection/streams.rs
```diff
@@ -1417,6 +1427,12 @@ pub struct UnknownStream {
     _private: (),
 }
 
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+enum StreamHalf {
```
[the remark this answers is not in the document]
Ralith: I prefer `Half` because "side" evokes endpoints, and it's confusing to say that we locally own both sides of the same stream.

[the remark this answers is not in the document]
djc: But for a unidirectional stream it's not a half, right? Maybe Part?

quinn-proto/src/connection/streams.rs
```diff
@@ -164,14 +179,16 @@ impl Streams {
         // Revert to initial state for outgoing streams
         for dir in Dir::iter() {
             for i in 0..self.next[dir as usize] {
-                self.send.remove(&StreamId::new(self.side, dir, i)).unwrap();
+                let id = StreamId::new(self.side, dir, i);
+                self.send.remove(&id).unwrap();
```
Matthias247, line 183: Since this also seems to free up streams, does it also have to increment the limit via `stream_freed`?
Or am I understanding something wrong regarding zero-rtt here?

quinn-proto/src/connection/streams.rs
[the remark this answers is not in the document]
Matthias247: Debatable. You could also define it as "there is another Half, but it's already in the final/closed state". This is e.g. how HTTP/2 defines unidirectional streams.

quinn-proto/src/connection/streams.rs
Ralith: 0-RTT streams are always locally-initiated (necessarily, since we haven't received anything from the peer yet) so we never need to issue flow control updates to account for them being open or closed, even if they're bidirectional. Added a comment to make it a bit clearer that skipping the `stream_freed` call is intentional.

quinn-proto/src/connection/streams.rs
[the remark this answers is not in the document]
Ralith: It also evokes "half duplex" for me. I think it's fair to describe a unidirectional stream as half of a bidirectional stream, and indeed that's how our types work out at the `quinn` layer.

djc (approved):

PR #965, opened 2021-01-03
Update link to native Gitter branch

djc:


Ralith (approved):
In the long term I'd kinda prefer to make the proper matrix channel the source of truth, but there's no denying the bridge sucks at that mode of operation right now.

PR #967, opened 2021-01-04
Allow to use the incoming IP address for sending outgoing packets

Matthias247:
This is a follow-up for #943. When a socket is bound to a wildcard
IP address, sending the outgoing IP might use a different source IP
address than the one the packet was received on, since the OS might
not be able to identify the necessary route. This would lead packets
not allowing to reach the client.

This change adds a setting which will set an explicit source address
in all outgoing packets. The source address which will be used is
the local IP address which was used to receive the initial incoming
packet.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -365,6 +365,7 @@ where
                     contents: buf,
                     ecn: None,
                     segment_size: None,
+                    src_ip: self.local_ip,
                 });
             }
         }
@@ -554,6 +555,7 @@ where
                 None
             },
             segment_size: None,
+            src_ip: self.local_ip,
         })
     }
 
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -190,6 +190,7 @@ where
                         ecn: None,
                         contents: buf,
                         segment_size: None,
+                        src_ip: local_ip,
                     });
                     return None;
                 }
@@ -253,7 +254,7 @@ where
 
         if !self.is_server() {
             debug!("packet for unrecognized connection {}", dst_cid);
-            self.stateless_reset(datagram_len, remote, &dst_cid);
+            self.stateless_reset(datagram_len, remote, local_ip, &dst_cid);
             return None;
         }
 
@@ -288,7 +289,7 @@ where
         //
 
         if !dst_cid.is_empty() {
-            self.stateless_reset(datagram_len, remote, &dst_cid);
+            self.stateless_reset(datagram_len, remote, local_ip, &dst_cid);
         } else {
             trace!("dropping unrecognized short packet without ID");
         }
@@ -299,6 +300,7 @@ where
         &mut self,
         inciting_dgram_len: usize,
         remote: SocketAddr,
+        local_ip: Option<IpAddr>,
         dst_cid: &ConnectionId,
     ) {
         /// Minimum amount of padding for the stateless reset to look like a short-header packet
@@ -336,6 +338,7 @@ where
             ecn: None,
             contents: buf,
             segment_size: None,
+            src_ip: local_ip,
         });
     }
 
@@ -535,6 +538,7 @@ where
             debug!("refusing connection");
             self.initial_close(
                 remote,
+                local_ip,
                 crypto,
                 &src_cid,
                 &temp_loc_cid,
@@ -553,6 +557,7 @@ where
             );
             self.initial_close(
                 remote,
+                local_ip,
                 crypto,
                 &src_cid,
                 &temp_loc_cid,
@@ -590,6 +595,7 @@ where
                     ecn: None,
                     contents: buf,
                     segment_size: None,
+                    src_ip: local_ip,
                 });
                 return None;
             }
@@ -608,6 +614,7 @@ where
                     debug!("rejecting invalid stateless retry token");
                     self.initial_close(
                         remote,
+                        local_ip,
                         crypto,
                         &src_cid,
                         &temp_loc_cid,
@@ -645,7 +652,7 @@ where
                 debug!("handshake failed: {}", e);
                 self.handle_event(ch, EndpointEvent(EndpointEventInner::Drained));
                 if let ConnectionError::TransportError(e) = e {
-                    self.initial_close(remote, crypto, &src_cid, &temp_loc_cid, e);
+                    self.initial_close(remote, local_ip, crypto, &src_cid, &temp_loc_cid, e);
                 }
                 None
             }
@@ -655,6 +662,7 @@ where
     fn initial_close(
         &mut self,
         destination: SocketAddr,
+        local_ip: Option<IpAddr>,
         crypto: &Keys<S>,
         remote_id: &ConnectionId,
         local_id: &ConnectionId,
@@ -683,6 +691,7 @@ where
             ecn: None,
             contents: buf,
             segment_size: None,
+            src_ip: local_ip,
         })
     }
 
--- a/quinn-proto/src/lib.rs
+++ b/quinn-proto/src/lib.rs
@@ -18,7 +18,13 @@
 #![allow(clippy::cognitive_complexity)]
 #![allow(clippy::too_many_arguments)]
 
-use std::{convert::TryInto, fmt, net::SocketAddr, ops, time::Duration};
+use std::{
+    convert::TryInto,
+    fmt,
+    net::{IpAddr, SocketAddr},
+    ops,
+    time::Duration,
+};
 
 mod cid_queue;
 #[doc(hidden)]
@@ -277,6 +283,8 @@ pub struct Transmit {
     /// The segment size if this transmission contains multiple datagrams.
     /// This is `None` if the transmit only contains a single datagram
     pub segment_size: Option<usize>,
+    /// Optional source IP address for the datagram
+    pub src_ip: Option<IpAddr>,
 }
 
 //
--- a/quinn/src/platform/unix.rs
+++ b/quinn/src/platform/unix.rs
@@ -2,7 +2,7 @@ use std::{
     io,
     io::IoSliceMut,
     mem::{self, MaybeUninit},
-    net::{IpAddr, SocketAddr, SocketAddrV4, SocketAddrV6},
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6},
     os::unix::io::AsRawFd,
     ptr,
 };
@@ -273,7 +273,7 @@ pub fn caps() -> UdpCapabilities {
     *CAPABILITIES
 }
 
-const CMSG_LEN: usize = 64;
+const CMSG_LEN: usize = 80;
 
 fn prepare_msg(
     transmit: &Transmit,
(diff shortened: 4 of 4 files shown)


quinn-proto/src/config.rs
```diff
@@ -434,6 +434,11 @@ where
     /// Improves behavior for clients that move between different internet connections or suffer NAT
     /// rebinding. Enabled by default.
     pub(crate) migration: bool,
+
+    /// Whether to use the local IP address which was used when receiving the initial
+    /// packet to send out all outgoing packets to a clients instead of the IP
+    /// the the socket is bound to.
+    pub(crate) use_intial_local_ip_on_send: bool,
```
djc, line 441: Are there any compelling reasons not to do this on platforms where it is available? Should we default to true? Also, how about a slightly less verbose name, maybe `send_from_initial_ip`?

(Also you have a consistent typo, `intial` -> `initial`.)

quinn-proto/src/connection/mod.rs
```diff
@@ -717,6 +719,22 @@ where
         pad
     }
 
+    /// Returns the destination IP used in outgoing transmit packet
+    ///
+    /// Returns `None` if no specific IP should be used
+    fn transmit_dst_ip(&self) -> Option<IpAddr> {
```
djc, line 725: I think I'd prefer it like this:

```rust
match self.server_config.as_ref()?.use_initial_local_ip_on_send {
    true => self.local_ip,
    false => None,
}
```

Any thoughts?

quinn-proto/src/endpoint.rs
```diff
@@ -389,6 +392,22 @@ where
         ConnectionEvent(ConnectionEventInner::NewIdentifiers(ids, now))
     }
 
+    /// Returns the destination IP used in outgoing transmit packets.
+    ///
+    /// Returns `None` if no specific IP should be used
+    fn transmit_dst_ip(&self, local_ip: Option<IpAddr>) -> Option<IpAddr> {
+        if self
```
djc, line 399: Same here. The `transmit_` prefix seems superfluous here, since it's pretty much implied by `dst`?

quinn/src/platform/unix.rs
```diff
@@ -290,12 +290,40 @@ fn prepare_msg(
...
+
+    if let Some(ip) = &transmit.dst_ip {
+        if cfg!(target_os = "linux") {
+            match ip {
+                IpAddr::V4(v4) => {
+                    let pktinfo = libc::in_pktinfo {
+                        ipi_ifindex: 0,
+                        ipi_spec_dst: libc::in_addr { s_addr: 0 },
+                        ipi_addr: unsafe {
+                            *(v4 as *const Ipv4Addr as *const () as *const libc::in_addr)
```
djc, line 309: Are we allowed to make this cast? I thought this was being reconsidered.

https://github.com/rust-lang/rust/pull/78802

djc (commented):
Nice work!

quinn-proto/src/config.rs
Matthias247: `send_from_initial_ip` is fine to me. I don't see a huge amount of reasons why it shouldn't be the default. But I tried to avoid breaking people by enabling it unconditionally. We can also set the default to `true`, and allow people to disable it?

quinn-proto/src/connection/mod.rs
Matthias247: Definitely preferable. Will update

quinn-proto/src/config.rs
djc: Do you see any reason? I'd lean towards not making it configurable by default, until someone complains. Easier to add the API later than remove it. (But curious what @Ralith thinks.)

quinn-proto/src/config.rs
```diff
@@ -434,6 +434,11 @@ where
     /// Improves behavior for clients that move between different internet connections or suffer NAT
     /// rebinding. Enabled by default.
     pub(crate) migration: bool,
+
+    /// Whether to use the local IP address which was used when receiving the initial
```
djc, line 438: Nit on docstring: we prefer to have a short first line (no period at the end, since it'll be a sentence fragment), then extended explanation after an empty line. This gets the optimal result in rustdoc (where the initial line is used in lists of items).

quinn-proto/src/endpoint.rs
Matthias247: Yes we can remove it, and I did that. However I then noticed that `dst_ip` is wrong, since this is the source IP address which is used for transmits, instead of the destination. We can't make it `local_ip`, because the function is already taken and doesn't take the setting into account (unless we drop the setting). Any other recommendation?

djc: `src_ip()`, then? Or the full `send_from_ip()`?

quinn-proto/src/config.rs
Matthias247: The only reason I see is "I haven't tested on enough platforms". I know it works fine on my dev machine as well as a test server I deploy too, but can't comment on anything else. I'm good with removing too, and adding more APIs if deemed necessary.

I guess one reason could be "if bound to only one socket, attaching the cmsg to every outgoing packet might be more work". But I doubt that is quantifyable.

quinn/src/platform/unix.rs
Matthias247: It definitely is based on assumptions of the stdlib. However those are already made in a few other places. I think if we are not comfortable with them, all those should be updated.

djc: Fair enough, filed https://github.com/quinn-rs/quinn/issues/968 about this.

quinn-proto/src/config.rs
Ralith: AFAICT this change is purely a bugfix and disabling it would just reintroduce the bug, so I don't see a reason for it to be optional.

quinn/src/platform/unix.rs
Ralith, line 309: Is this the correct field? `man 7 ip` says:

> If IP_PKTINFO is passed to sendmsg(2) and ipi_spec_dst is not zero, then it is used as the local  source address for the routing table lookup and for setting up IP source route options.

I'm confused about the distinction between the two address fields in general.

quinn-proto/src/lib.rs
```diff
@@ -274,6 +280,8 @@ pub struct Transmit {
     pub ecn: Option<EcnCodepoint>,
     /// Contents of the datagram
     pub contents: Vec<u8>,
+    /// Optional destination IP address for the datagram
+    pub dst_ip: Option<IpAddr>,
```
Ralith, line 284: Shouldn't this be `src_ip`?

quinn/src/platform/unix.rs
Matthias247: I was confused too. But testing seemed to indicate it is doing the right thing. But I admit I haven't tested the IPv4 version on the host which is more susceptible to the issue. That one uses the dual-stack/IPv6 option

quinn-proto/src/lib.rs
Matthias247: Yes. See also https://github.com/quinn-rs/quinn/pull/967#discussion_r551551134 about the naming of the helper function

quinn-proto/src/endpoint.rs
Matthias247: Went for `src_ip()`

quinn-proto/src/config.rs
```diff
@@ -434,6 +434,10 @@ where
     /// Improves behavior for clients that move between different internet connections or suffer NAT
     /// rebinding. Enabled by default.
     pub(crate) migration: bool,
+
+    /// Whether to use the initial local IP address for outgoing packets
+    /// instead of the IP the the socket is bound to.
```
djc, line 439:
```suggestion
    /// Whether to use the initial local IP address for outgoing packets
    ///
    /// If set to `true`, the local IP address which was target of the initial
    /// QUIC packet will be used as the source IP address for all outgoing packets.
    /// If set to `false`, the source IP address will be the address the socket is
    /// bound to.
    ///
    /// The setting only has an effect on platforms where
    /// [`quinn_proto::Connection::local_ip()`] returns a local IP address.
    /// (Currently only on Linux.)
```

quinn-proto/src/config.rs
```diff
@@ -492,6 +497,21 @@ where
...
+    /// Whether to use the initial local IP address for outgoing packets.
+    ///
+    /// If set to `true`, the local IP address which was target of the initial
+    /// QUIC packet will be used as the source IP address for all outoing packets.
+    ///
+    /// If set to `false`, the source IP address will be address the socket is
+    /// bound to.
+    ///
+    /// The setting is only having an effect on platforms where
+    /// [`quinn_proto::Connection::local_ip()`] returns a local IP address.
```
djc, line 510:
```suggestion
    /// Whether to use the initial local IP address for outgoing packets
    ///
    /// For more information, see the documentation for [`ServerConfig::send_from_initial_ip`].
```

quinn/src/builders.rs
```diff
@@ -168,6 +168,21 @@ where
...
+    /// Whether to use the initial local IP address for outgoing packets.
+    ///
+    /// If set to `true`, the local IP address which was target of the initial
+    /// QUIC packet will be used as the source IP address for all outoing packets.
+    ///
+    /// If set to `false`, the source IP address will be address the socket is
+    /// bound to.
+    ///
+    /// The setting is only having an effect on platforms where
+    /// [`quinn::Connection::local_ip()`] returns a local IP address.
```
djc, line 181:
```suggestion
    /// Whether to use the initial local IP address for outgoing packets
    ///
    /// For more information, see the documentation for [`ServerConfig::send_from_initial_ip`].
```

quinn-proto/src/config.rs
Matthias247: Ok. Should I remove the switch then?

djc: Yup, let's try it!

quinn-proto/src/connection/mod.rs
```diff
@@ -719,6 +721,13 @@ where
         pad
     }
 
+    /// Returns the source IP used in outgoing packets.
+    ///
+    /// Returns `None` if no specific IP should be used
+    fn src_ip(&self) -> Option<IpAddr> {
```
Matthias247, line 727: With the config gone, this is now not too useful anymore. Just let me know whether to keep or remove it. I think keeping it is also fine

quinn/src/platform/unix.rs
Matthias247: I did a bit more validation here, and tested both approaches. I modified the bulk benchmark to use IPv4 only, and configured the client to manually connect to the eth1 IP of my host (not loopback). This works with setting `ipi_spec_dst`, but times out when setting `ipi_addr`. So for the sending side `ipi_spec_dst` seems like the right field, which is also indicated by the documentation.

Only weird that for reception `ipi_addr` delivers the expected value, and is rather undocumented.

quinn/src/platform/unix.rs
```diff
@@ -273,7 +273,7 @@ pub fn caps() -> UdpCapabilities {
     *CAPABILITIES
 }
 
-const CMSG_LEN: usize = 64;
+const CMSG_LEN: usize = 80;
```
Matthias247, line 276: This was updated to account for GSO + src-ip setting

quinn-proto/src/connection/mod.rs
djc: Let's ditch it.

Matthias247: removed the `src_ip` functions.

djc (approved):

PR #969, opened 2021-01-04
Fix missing flow control checks in RESET_STREAM final offset handling

### LINKED ISSUE #966: streams received_reset panics with large final offsets
Fuzzing against the `Streams` target raised this new panic, flagging here for discussion. 

### example output
```text
target/x86_64-unknown-linux-gnu/release/streams: Running 1 inputs 1 time(s) each.
Running: fuzz/artifacts/streams/minimized-from-06eda8ebc2230c5f8f7f12042c3d1c7684018c48
thread '<unnamed>' panicked at 'attempt to add with overflow', quinn-proto/src/connection/streams.rs:365:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

see attached tarball with a minimized example input to reproduce
[minimized-from-06eda8ebc2230c5f8f7f12042c3d1c7684018c48.tar.gz](https://github.com/quinn-rs/quinn/files/5762262/minimized-from-06eda8ebc2230c5f8f7f12042c3d1c7684018c48.tar.gz)

I expect this could be resolved by setting a limit in the fuzzing input, ie on the size of the `ResetStream` final offset and is a non-issue. Alternately there could be [another validation check added](https://github.com/quinn-rs/quinn/blob/main/quinn-proto/src/connection/streams.rs#L337) but that might be overkill

To run this 
1. install [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz#installation)
2. extract the example input attached
3. do `cargo fuzz run streams /path/to/the/minimized-case`

Ralith:
Plus assorted cleanup. Fixes #966.

### DIFF
--- a/quinn-proto/src/connection/streams.rs
+++ b/quinn-proto/src/connection/streams.rs
@@ -291,7 +291,7 @@ impl Streams {
             self.local_max_data,
             self.stream_receive_window,
         )?;
-        self.data_recvd += new_bytes;
+        self.data_recvd = self.data_recvd.saturating_add(new_bytes);
 
         if !rs.assembler.is_stopped() {
             self.on_stream_frame(true, stream);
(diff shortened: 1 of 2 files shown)


djc (approved):
Nice work, with some pretty great refactoring in there!

djc:
(After #952 lands I think we should split `send` and `recv` out of `streams`...)

Ralith:
Good call; it's gotten pretty big and annoying to navigate.

PR #970, opened 2021-01-04
Apply anti-amplification when peer migrates

Ralith:
Defends against amplification attacks where an existing connection is
hijacked by modifying the source address of a legitimate packet, as
required by the spec. Previously we only defended against similar
attacks involving brand new connections being created by an attacker.


quinn-proto/src/connection/paths.rs
```diff
@@ -45,8 +58,17 @@ impl PathData {
...
             challenge_pending: false,
+            validated: false,
+            total_sent: 0,
+            total_recvd: 0,
         }
     }
+
+    /// Indicates whether we're a server that hasn't validated the peer's address and hasn't
+    /// received enough data from the peer to permit sending `bytes_to_send` additional bytes
+    pub fn anti_amplification_blocked(&self, bytes_to_send: u16) -> bool {
```
Matthias247, line 69: Really worthwhile to stick with `u16` here? Guess it's just moving the cast around.

quinn-proto/src/connection/mod.rs
```diff
@@ -1807,7 +1793,7 @@ where
         ecn: Option<EcnCodepoint>,
         data: BytesMut,
     ) {
-        self.total_recvd = self.total_recvd.wrapping_add(data.len() as u64);
+        self.path.total_recvd = self.path.total_recvd.saturating_add(data.len() as u64);
```
Matthias247, line 1796: Will take a long time to saturate :-) 

But fair to be correct

Matthias247 (approved):

quinn-proto/src/connection/mod.rs
Ralith: Yeah, this is mostly a clarification of intent thing.

quinn-proto/src/connection/paths.rs
Ralith: Given that we'll in the very near future be passing multiple MTUs, I suppose widening this is sensible.

quinn-proto/src/connection/mod.rs
```diff
@@ -1493,7 +1479,7 @@ where
             return;
         }
 
-        if self.anti_amplification_blocked() {
+        if self.path.anti_amplification_blocked(self.mtu) {
```
djc: Hmm, does `mtu` also belong in `self.path`?

djc (dismissed):
Looks good to me, nice fixes! (And more logic `Connection` -> `Path`...)

quinn-proto/src/connection/mod.rs
Ralith: I think so. However, I'm leaving the argument in place because I anticipate it being immediately useful in #953.

djc (approved):

PR #972, opened 2021-01-08
Expose the connections congestion window

Matthias247:
This can be useful for debugging performance issues

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1121,6 +1121,11 @@ where
         self.local_ip
     }
 
+    /// Current best estimate of this connection's congestion window
+    pub fn cwnd(&self) -> u64 {
+        self.path.congestion.window()
+    }
+
     /// Current best estimate of this connection's latency (round-trip-time)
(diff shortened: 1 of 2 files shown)


Matthias247:
One alternative: This (as well as similar stats like `rtt`) details could also be returned via `ConnectionStats`. That allows the caller to get everything at once. That requires to add a new field (`path`? `recovery`? `reliability`?)

But maybe both options can be done.

Ralith (approved):
An aggregate return value is an interesting idea. I'd prefer not to have redundant interfaces, though. If we go that way, we should also remove the individual getters.

djc:
Aggregating this stuff would be good. I wonder if it makes sense conceptually to fold everything into `ConnectionStats` or if we should maybe have two types: `ConnectionStats` for cumulative historic data, and `ConnectionState` for current data (names TBD, of course).

Ralith:
That occurred to me as well, but I'm having trouble seeing concrete value in distinguishing them in the API.

Matthias247:
same thought. But they are so often the same - e.g. total number of transmitted packets is historic data as well as current data, since it's a snapshot view

PR #973, opened 2021-01-08
Add recovery stats as part of ConnectionStats

Matthias247:
This is an alternative version of #972, where we add the new stats
to ConnectionStats instead of exposing additional accessors.

The naming of `recovery` could probably be improved. `path` might
be an option. But where would we add something like packet loss?
That seems more like an overall stat.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -935,7 +935,11 @@ where
 
     /// Returns connection statistics
     pub fn stats(&self) -> ConnectionStats {
-        self.stats
+        let mut stats = self.stats;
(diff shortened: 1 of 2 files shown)


Ralith:
I like `path`. The congestion controller treats packet loss on independent paths as independent, so reporting it that way makes sense to me as well.

Ralith (dismissed):

Matthias247:
> I like `path`. The congestion controller treats packet loss on independent paths as independent, so reporting it that way makes sense to me as well.

Seems good to me too for the current values. We can have something else for values which are persistent between paths too. More fields don't have more cost.

Ralith (approved):

djc (approved):

PR #975, opened 2021-01-11
Congestion controller fixes

Matthias247:
This change fixes 2 issues in the congestion controller:

Extra slow slow start
===

The congestion controller had an issue where it didn't ramp up by
doubling the window in the slow start phase as expected. The reason
here was the usage of the "congestion_blocked" field, which aims to
remove the avoid growing the window when there is no data to send.
The issue with the field was that the first incoming ACK for every batch
would increase the congestion window, which would let the next
call to `congestion_blocked()` return false, and thereby let the
congestion controller ignore all of the following ACKs in a received
batch up the point where the next `poll_transmit()` call is made and
the window is filled again. Since often all ACKs for one round-trip
arrive in a single batch that means only the first ACK had an effect
and the others where ignored -> which lead to increasing the
window by 1 MTU instead of doubling it in the slow start phase.

The new approach introduces a new `app_limited` field which
caches whether the last `poll_transmit` attempt couldn't produce
data because there was no application data available.

Numeric precision issue in congestion avoidance
===

The formula
```
self.window += self.config.max_datagram_size * bytes / self.window;
```
which is specified in the RFC requires floating point arithmetic.
If integer arithmetic is used, the multiplication of the first 2 values
yields about 1.7MB for 1300 byte packets. As soon as the window is
bigger than this value, the window won't change at all due the
division leading to 0.

The new implementation follow this guidance from the quic recovery
specification:

> In congestion avoidance, implementers that use an integer
> representation for congestion_window should be careful with division,
> and can use the alternative approach suggested in Section 2.1 of
> [RFC3465].

### DIFF
--- a/quinn-proto/src/congestion/new_reno.rs
+++ b/quinn-proto/src/congestion/new_reno.rs
@@ -15,6 +15,8 @@ pub struct NewReno {
     /// The time when QUIC first detects a loss, causing it to enter recovery. When a packet sent
     /// after this time is acknowledged, QUIC exits recovery.
     recovery_start_time: Instant,
+    /// Bytes which had been acked by the peer since leaving slow start
+    bytes_acked: u64,
 }
 
 impl NewReno {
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -536,6 +546,12 @@ where
             }
         }
 
+        if buf.is_empty() && !was_congestion_blocked {
+            self.app_limited = true;
+        } else {
+            self.app_limited = false;
+        }
```
djc, line 553: Let's just write this as `self.app_limited = buf.is_empty() && !was_congestion_blocked;`?

djc (commented):
This looks good to me, but it probably also needs a review from @Ralith.

quinn-proto/src/connection/mod.rs
Matthias247: sure. Fixed

quinn-proto/src/congestion/new_reno.rs
```diff
@@ -25,21 +27,37 @@ impl NewReno {
...
-            self.window += self.config.max_datagram_size * bytes / self.window;
+            // This implementation uses the method which does not require
+            // floating point math, which also increases the window by 1 datagram
+            // for every round trip.
+            self.bytes_acked += bytes;
+
+            if self.bytes_acked >= self.window {
+                self.bytes_acked -= self.window;
+                self.window += self.config.max_datagram_size;
+            }
```
Ralith, line 60: I think it'd be helpful to cite the RFC we got this logic from.

quinn-proto/src/connection/mod.rs
```diff
@@ -422,6 +428,8 @@ where
                 || self.path_response.is_some()
         });
 
+        let mut was_congestion_blocked = false;
```
Ralith, line 431: Elsewhere (e.g. `handle_packet`) we use "was_*" to refer to the state prior to running some logic which may change that state, whereas this is referring to the state after we're done. Could we say `is_` here instead? Also, maybe `transport_blocked` (as opposed to application) since pacing can be responsible as well as congestion?

quinn-proto/src/congestion/new_reno.rs
```diff
@@ -25,21 +27,37 @@ impl NewReno {
...
         }
+
         if self.window < self.ssthresh {
             // Slow start
             self.window += bytes;
+
+            if self.window >= self.ssthresh {
+                // Exiting slow start
+                // Initialize `bytes_acked` for congestion avoidance
+                self.bytes_acked = self.window - self.ssthresh
```
Ralith, line 48: What's this logic based on? Also, style nit:
```suggestion
                self.bytes_acked = self.window - self.ssthresh;
```

Ralith (commented):
Thanks, this is much needed!

quinn-proto/src/connection/mod.rs
Matthias247: Fine with removing `was_`. But I think even with pacing `congestion_blocked` is fine, because pacing is a form of congestion control

Ralith (approved):
LGTM, thanks!

PR #977, opened 2021-01-19
Non contiguous Send buffer

Matthias247:
During testing with a multithreaded runtime I discovered that threads
block a fair amount of time on Mutexes, even though everything in the
state machine should be non-blocking.

By inserting some more timing checks I discovered that the `poll_write`
call took more than 10ms and locked the Mutex for this duration. This
happened due to the `BytesMut::extend_from_slice` taking this time.

The reason here is that extending the buffer require a reallocation of
the buffer and a copy of the complete data. This can be a big size
(up the maximum internal buffer size) - even if only a tiny chunk of
data is added.

This change improves on this by using a non-contiguous send buffer.
New data is appended as a new `Bytes` segment, which only requires
an allocation for this particular size. The 10ms blocking problem is
gone with this.

Another benefit of this change is that it can easily enable zero-copy
writes by accepting `Bytes` as a parameter in the user-facing API.
This is however not performed here yet.

```
Jan 18 11:14:00.064  WARN quinn_proto::connection::send_buffer: self.unacked.extend_from_slice(data[..1048576]) took 10.3452ms. Now unacked size: 94924462
Jan 18 11:14:00.064  WARN quinn_proto::connection::streams: Long Send::write. Total: 10.3943ms, Budget check: 0ns, data_len: 1048576, written: 1048576
Jan 18 11:14:00.064  WARN quinn_proto::connection::streams: Long Streams::write. Total: 10.4203ms, Limit check: 100ns, Unsent check: 200ns, Write: 10.4202ms. data_len: 1048576, written: 1048576
Jan 18 11:14:00.064  WARN quinn::streams: Long poll_write time: close_check 100ns, write_time 10.4456ms, end_time 10.4456ms
```

### DIFF
--- a/quinn-proto/src/connection/send_buffer.rs
+++ b/quinn-proto/src/connection/send_buffer.rs
@@ -1,6 +1,6 @@
-use std::ops::Range;
+use std::{collections::VecDeque, ops::Range};
 
-use bytes::{Buf, BytesMut};
+use bytes::{Buf, Bytes};
 
 use crate::range_set::RangeSet;
 
@@ -8,7 +8,9 @@ use crate::range_set::RangeSet;
 #[derive(Default, Debug)]
 pub struct SendBuffer {
     /// Data queued by the application but not yet acknowledged. May or may not have been sent.
-    unacked: BytesMut,
+    unacked_segments: VecDeque<Bytes>,
+    /// Total size of `unacked_segments`
+    unacked_len: usize,
     /// The first offset that hasn't been written by the application, i.e. the offset past the end of `unacked`
     offset: u64,
     /// The first offset that hasn't been sent
@@ -31,21 +33,44 @@ impl SendBuffer {
 
     /// Append application data to the end of the stream
     pub fn write(&mut self, data: &[u8]) {
-        self.unacked.extend_from_slice(data);
+        let buf = Bytes::from(data.to_owned());
+        self.unacked_segments.push_back(buf);
+        self.unacked_len += data.len();
         self.offset += data.len() as u64;
     }
 
     /// Discard a range of acknowledged stream data
     pub fn ack(&mut self, mut range: Range<u64>) {
         // Clamp the range to data which is still tracked
-        let base_offset = self.offset - self.unacked.len() as u64;
+        let base_offset = self.offset - self.unacked_len as u64;
         range.start = base_offset.max(range.start);
         range.end = base_offset.max(range.end);
 
         self.acks.insert(range);
-        while self.acks.min() == Some(self.offset - self.unacked.len() as u64) {
+
+        while self.acks.min() == Some(self.offset - self.unacked_len as u64) {
             let prefix = self.acks.pop_min().unwrap();
-            self.unacked.advance((prefix.end - prefix.start) as usize);
+            let mut to_advance = (prefix.end - prefix.start) as usize;
+
+            self.unacked_len -= to_advance;
+            while to_advance > 0 {
+                let front = self
+                    .unacked_segments
+                    .front_mut()
+                    .expect("Expected buffered data");
+
+                if front.len() <= to_advance {
+                    to_advance -= front.len();
+                    self.unacked_segments.pop_front();
+
+                    if self.unacked_segments.len() * 4 < self.unacked_segments.capacity() {
+                        self.unacked_segments.shrink_to_fit();
+                    }
+                } else {
+                    front.advance(to_advance);
+                    to_advance = 0;
+                }
+            }
         }
     }
 
@@ -69,11 +94,29 @@ impl SendBuffer {
         result
     }
 
+    /// Returns data which is associated with a range
+    ///
+    /// This function can return a subset of the range, if the data is stored
+    /// in noncontiguous fashion in the send buffer. In this case callers
+    /// should call the function again with an incremented start offset to
+    /// retrieve more data.
     pub fn get(&self, offsets: Range<u64>) -> &[u8] {
-        let base_offset = self.offset - self.unacked.len() as u64;
-        let start = (offsets.start - base_offset) as usize;
-        let end = (offsets.end - base_offset) as usize;
-        &self.unacked[start..end]
+        let base_offset = self.offset - self.unacked_len as u64;
+
+        let mut segment_offset = base_offset;
+        for segment in self.unacked_segments.iter() {
+            if offsets.start >= segment_offset
+                && offsets.start < segment_offset + segment.len() as u64
+            {
+                let start = (offsets.start - segment_offset) as usize;
+                let end = (offsets.end - segment_offset) as usize;
+
+                return &segment[start..end.min(segment.len())];
+            }
+            segment_offset += segment.len() as u64;
+        }
+
+        &[]
     }
 
     /// Queue a range of sent but unacknowledged data to be retransmitted
@@ -83,7 +126,7 @@ impl SendBuffer {
     }
 
     pub fn retransmit_all_for_0rtt(&mut self) {
-        debug_assert_eq!(self.offset, self.unacked.len() as u64);
+        debug_assert_eq!(self.offset, self.unacked_len as u64);
         self.unsent = 0;
     }
 
@@ -95,7 +138,7 @@ impl SendBuffer {
 
     /// Whether all sent data has been acknowledged
     pub fn is_fully_acked(&self) -> bool {
-        self.unacked.is_empty()
+        self.unacked_len == 0
     }
 
     /// Whether there's data to send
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/send_buffer.rs
```diff
@@ -31,18 +33,46 @@ impl SendBuffer {
...
 
     /// Discard a range of acknowledged stream data
     ///
     /// Each offset must be acknowledged at most once.
-    pub fn ack(&mut self, range: Range<u64>) {
+    pub fn ack(&mut self, mut range: Range<u64>) {
+        // Clamp the range to data which is still tracked
+        let base_offset = self.offset - self.unacked_len as u64;
+        range.start = base_offset.max(range.start);
+        range.end = base_offset.max(range.end);
```
Ralith, line 49: I believe this is an unrelated change; it should be split off into its own commit (and the doc comment updated, if we're relaxing the constraint).

quinn-proto/src/connection/send_buffer.rs
```diff
@@ -31,18 +33,46 @@ impl SendBuffer {
 
     /// Append application data to the end of the stream
     pub fn write(&mut self, data: &[u8]) {
-        self.unacked.extend_from_slice(data);
+        let buf: Bytes = Bytes::from(data.to_owned());
+        self.unacked_segments.push_back(buf);
+        self.unacked_len += data.len();
```
Ralith, line 38: I'm concerned that this will drastically reduce performance for small writes. Could we allocate chunks with an explicit capacity (say, 4k) instead?

quinn-proto/src/connection/streams.rs
```diff
@@ -609,7 +609,16 @@ impl Streams {
             let meta = frame::StreamMeta { id, offsets, fin };
             trace!(id = %meta.id, off = meta.offsets.start, len = meta.offsets.end - meta.offsets.start, fin = meta.fin, "STREAM");
             meta.encode(true, buf);
-            buf.put_slice(stream.pending.get(meta.offsets.clone()));
+
+            // The range might not be retrievable in a singel `get` if it is
```
Ralith, line 613:
```suggestion
            // The range might not be retrievable in a single `get` if it is
```

quinn-proto/src/connection/send_buffer.rs
Matthias247: I thought about it. Buy I feel like if someone wants to optimize small writes they might rather want to have a buffered writer as part of their application. Doing a "quinn syscall" will never be super efficient, due to Mutex boundaries crossed, waking up another task, etc.

Also this would then need to be done via `BytesMut` and splitting into already committed capacity and free one which might be used for another write. But `BytesMut` is currently broken, [and would in that case trigger an additional allocation](https://github.com/tokio-rs/bytes/issues/401).

Ralith: Writes are async wrt. task wakeups, so it's not clear to me that the overhead is *that* high.

> BytesMut is currently broken, and would in that case trigger an additional allocation.

That's disappointing, but I don't think it applies here, since we'd only be freezing after filling a chunk (presuming we get exactly the capacity we ask for, at least, which should be the case in practice).

Matthias247: > Writes are async wrt. task wakeups, so it's not clear to me that the overhead is that high.

But it goes through the complete stack. Waking up connection (a different task), checking all states there, including manipulating timers, etc. I don't think it's that negligible - but it's currently also not that easy to get numbers.

> That's disappointing, but I don't think it applies here, since we'd only be freezing after filling a chunk (presuming we get exactly the capacity we ask for, at least, which should be the case in practice).

Well, you might need to have things frozen directly after the write if you want write queue is based on `Bytes` and things need to go out as soon as possible. Maybe an alternative is too keep the write queue based on `BytesMut`, but that doesn't play well into offering an interface to push `Bytes`.

Matthias247: I did a small benchmark for the reuse approach with
```
pub fn write(&mut self, mut data: &[u8]) {
        let mut buf = if let Some(reserve) = self.reserve.take() {
            reserve
        } else {
            BytesMut::with_capacity(4*1024)
        };
        buf.extend_from_slice(data);

        let enq_buf = buf.split();
        if buf.capacity() > 10 {
            self.reserve = Some(buf);
        }

        assert_eq!(enq_buf.len(), data.len());
        self.unacked_segments.push_back(enq_buf);
        self.unacked_len += data.len();
        self.offset += data.len() as u64;
    }
```
    
which certainly isn't perfect but should od some reuse. It ended up being slower than the version without.
One thing which might come into play is the other weird part of `BytesMut` that the linked PR fixes, which is that it somehow rememebers about capacities from allocations which have long been split off, and then tries to allocate more than it should later on.

Btw: I think this affects large as well as small writes. Even if the write is initially large - at some point the flow control window will be filled. At that point the peer drains the stream, and from there on there will be continuously only small writes possible.

Anyway - I think this is better than the current approach where the full amount of queued bytes are shifted whenever some data is acked (a 1 MTU ack can cause a 16MB memcpy). Therefore I think merging it the way it is is and in the future looking for additional improvements is fine.

quinn-proto/src/connection/send_buffer.rs
```diff
@@ -66,11 +94,29 @@ impl SendBuffer {
         result
     }
 
+    /// Returns data which is associated with a range
+    ///
+    /// This function can return a subset of the range, if the data is stored
+    /// in non contiguous fashion in the send buffer. In this case callers
+    /// should call the function again with an increment start offset to
+    /// retrieve more data.
```
djc, line 102:
```suggestion
    /// This function can return a subset of the range, if the data is stored
    /// in noncontiguous fashion in the send buffer. In this case callers
    /// should call the function again with an incremented start offset to
    /// retrieve more data.
```

quinn-proto/src/connection/send_buffer.rs
```diff
@@ -104,14 +150,22 @@ impl SendBuffer {
...
     use super::*;
 
+    fn aggregate_unacked(buf: &SendBuffer) -> Vec<u8> {
+        let mut result = Vec::new();
+        for segment in buf.unacked_segments.iter() {
+            result.extend_from_slice(&segment[..]);
+        }
+        result
+    }
+
```
djc, line 168:
```suggestion
```

quinn-proto/src/connection/send_buffer.rs
```diff
@@ -155,9 +255,9 @@ mod tests {
         assert_eq!(buf.poll_transmit(5), 0..5);
         assert_eq!(buf.poll_transmit(2), 5..7);
         buf.ack(5..7);
-        assert_eq!(buf.unacked, MSG);
+        assert_eq!(aggregate_unacked(&buf), MSG);
         buf.ack(0..5);
-        assert_eq!(buf.unacked, MSG[7..]);
+        assert_eq!(aggregate_unacked(&buf), MSG[7..]);
         assert!(buf.acks.is_empty());
     }
```
djc, line 262:
```suggestion
    }

    fn aggregate_unacked(buf: &SendBuffer) -> Vec<u8> {
        let mut result = Vec::new();
        for segment in buf.unacked_segments.iter() {
            result.extend_from_slice(&segment[..]);
        }
        result
    }
```

quinn-proto/src/connection/streams.rs
```diff
@@ -609,7 +609,16 @@ impl Streams {
             let meta = frame::StreamMeta { id, offsets, fin };
             trace!(id = %meta.id, off = meta.offsets.start, len = meta.offsets.end - meta.offsets.start, fin = meta.fin, "STREAM");
             meta.encode(true, buf);
-            buf.put_slice(stream.pending.get(meta.offsets.clone()));
+
+            // The range might not be retrievable in a single `get` if it is
+            // stored in non contiguous fashion. Therefore this loop iterates
+            // until the range is fully copied into the frame.
```
djc, line 615:
```suggestion
            // The range might not be retrievable in a single `get` if it is
            // stored in noncontiguous fashion. Therefore this loop iterates
            // until the range is fully copied into the frame.
```

djc (commented):
Some minor stylistic things. Otherwise, this makes sense to me. Nice find!

quinn-proto/src/connection/send_buffer.rs
Ralith: Ah, good point re: ACKs. Fair enough!

Ralith: > Well, you might need to have things frozen directly after the write if you want write queue is based on Bytes and things need to go out as soon as possible.

It'd still give us an upper bound of a 4K (or whatever) realloc/copy vs. an entire flow control window.

quinn-proto/src/connection/send_buffer.rs
```diff
@@ -31,18 +33,44 @@ impl SendBuffer {
 
     /// Append application data to the end of the stream
     pub fn write(&mut self, data: &[u8]) {
-        self.unacked.extend_from_slice(data);
+        let buf: Bytes = Bytes::from(data.to_owned());
```
djc, line 36:
```suggestion
        let buf = Bytes::from(data.to_owned());
```

djc (approved):

PR #978, opened 2021-01-19
Fix bulk benchmark

Matthias247:
The sending data on the benchmark failed it panicked due
to an unwrap and thereby showed no result. This propagates
the first client-side error instead of panicking - which will also
preserve the statistics.


djc (approved):
Makes sense.

PR #983, opened 2021-01-22
Fix StreamAssembler panic/underflow

### LINKED ISSUE #982: Assembler panic
When trying to use the new `Bytes` based receive APIs from #952 I run into

```
thread panicked at 'attempt to subtract with overflow', quinn/quinn-proto/src/connection/assembler.rs:213:12
stack backtrace:
  14: quinn_proto::connection::assembler::Assembler::insert
             at quinn/quinn-proto/src/connection/assembler.rs:213
  15: quinn_proto::connection::streams::Recv::ingest
             at quinn/quinn-proto/src/connection/streams.rs:1070
  16: quinn_proto::connection::streams::Streams::received
             at quinn/quinn-proto/src/connection/streams.rs:280
  17: quinn_proto::connection::Connection<S>::process_payload
             at cargo-project/quinn/quinn-proto/src/connection/mod.rs:2391
  18: quinn_proto::connection::Connection<S>::process_decrypted_packet
             at cargo-project/quinn/quinn-proto/src/connection/mod.rs:2264
  19: quinn_proto::connection::Connection<S>::handle_packet
             at cargo-project/quinn/quinn-proto/src/connection/mod.rs:1993
  20: quinn_proto::connection::Connection<S>::handle_decode
             at cargo-project/quinn/quinn-proto/src/connection/mod.rs:1896
  21: quinn_proto::connection::Connection<S>::handle_event
             at cargo-project/quinn/quinn-proto/src/connection/mod.rs:765
  22: quinn::connection::ConnectionInner<S>::process_conn_events
             at cargo-project/quinn/quinn/src/connection.rs:798
  23: <quinn::connection::ConnectionDriver<S> as core::future::future::Future>::poll
             at cargo-project/quinn/quinn/src/connection.rs:287
```

Code link: [quinn/quinn-proto/src/connection/assembler.rs:213:12](https://github.com/quinn-rs/quinn/blob/4f8a5ebaf112103087466c4797dcd7113a4aa565/quinn-proto/src/connection/assembler.rs#L213)

I guess the `defragmented` flag doesn't get updated in all places it should.

I can't directly share the test which triggers this, but it mostly performs a 16MB upload using small chunks. This never showed up with the copy-based receive API, but the this might have been more due to different performance attributes than that the issue wasn't there?. The copy based API might actually allow the application to read more at once in case of small chunks and make it less likely for the assembler to kick in.

Matthias247:
The variable `defragmented` wasn't properly updated in the
`read_chunks` method, since it directly manipulated the map instead
of delegating to the `Assembler::pop` method. This caused an underflow
when trying to insert more data into the map.

This change fixes that, and adds a repro test for it.

Fixes #982

### DIFF
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -76,7 +76,7 @@ impl Assembler {
                 return Ok(None);
             } else if (chunk.offset + chunk.bytes.len() as u64) <= self.bytes_read {
                 // Next chunk is useless as the read index is beyond its end
-                self.data.pop();
+                self.pop();
                 continue;
             }
 
@@ -86,7 +86,7 @@ impl Assembler {
 
             self.bytes_read += len as u64;
 
-            return Ok(self.data.pop().map(|mut x| x.bytes.split_off(start)));
+            return Ok(self.pop().map(|mut x| x.1.split_off(start)));
         }
     }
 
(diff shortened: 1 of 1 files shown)


quinn-proto/src/connection/assembler.rs
```diff
@@ -465,6 +465,24 @@ mod test {
...
+    #[test]
+    fn defrag_read_chunk() {
+        let mut x = Assembler::new();
+        x.insert(3, Bytes::from_static(b"def"));
+        x.insert(0, Bytes::from_static(b"abc"));
+        x.insert(7, Bytes::from_static(b"ijk"));
+        x.insert(11, Bytes::from_static(b"mno"));
+        x.defragment();
+        assert_matches!(x.read_chunk(), Ok(Some(ref y)) if &y[..] == b"abcdef");
+        x.insert(5, Bytes::from_static(b"fghijklmno"));
```
Ralith, line 477: Is this supplying an inconsistent byte at offset 7? Looks like line 473 puts `i` there, and now we're saying it's `h`.

Ralith (commented):
Good catch, thanks!

quinn-proto/src/connection/assembler.rs
Matthias247: You are right. Now I had to do all the complicated spelling again

djc (approved):

PR #985, opened 2021-01-25
Set up MSRV in CI

djc:


Ralith (approved):
Should we document specifically that we will not publish releases that require a stable rustc < 6 months old?

PR #986, opened 2021-01-25
Simple reads

djc:


quinn-proto/src/connection/assembler.rs
```diff
@@ -60,13 +45,16 @@ impl Assembler {
     }
 
     // Get the the next ordered chunk
-    pub(crate) fn read_chunk(&mut self) -> Result<Option<Bytes>, IllegalOrderedRead> {
+    pub(crate) fn read(
+        &mut self,
+        max_length: Option<usize>,
```
Ralith, line 50: Is there a benefit to this being `Option` rather than allowing indifferent callers to pass in `usize::MAX`?

quinn-proto/src/connection/assembler.rs
```diff
@@ -76,56 +64,30 @@ impl Assembler {
...
-
-        // If this chunk is either after the current offset or fully before it,
-        // return directly, indicating whether the chunk can be discarded.
-        if chunk.offset > self.bytes_read {
-            return false;
-        } else if (chunk.offset + chunk.bytes.len() as u64) <= self.bytes_read {
-            return true;
-        }
+            if start > 0 {
+                let _ = chunk.bytes.split_to(start);
```
Ralith, line 74:
```suggestion
                chunk.bytes.advance(start);
```

quinn-proto/src/connection/streams/types.rs
```diff
@@ -0,0 +1,528 @@
...
+    ///
+    /// Carries an application-defined error code.
+    ///
+    /// [`StreamEvent::Finished`]: crate::StreamEvent::Finished
+    #[error("stopped by peer: code {}", 0)]
+    Stopped(VarInt),
+    /// The stream has not been opened or was already finished or reset
+    #[error("unknown stream")]
+    UnknownStream,
+}
```
Ralith, line 528: I'd like to see this broken up even further, into separate send/recv modules. I've spent a lot of time lost/confused due to file-scoped searches turning up states for the wrong type of stream.

Ralith (commented):
Nice cleanup!

quinn-proto/src/connection/assembler.rs
djc: Benefits I could think of: maybe more obvious that it's a sentinel value, `None` is shorter to write than `usize::MAX`. But these arguments weren't enough to convince me that the simpler type wasn't more attractive, so changed it as you proposed.

quinn-proto/src/connection/assembler.rs
djc: Yup.

quinn-proto/src/connection/streams/types.rs
djc: 👍 done.

Ralith (approved):

quinn-proto/src/connection/assembler.rs
Ralith: I'm not even sure I'd call it a sentinel value; the beauty of it is that no special logic is needed, so we actually have *fewer* branches.

quinn-proto/src/connection/mod.rs
```diff
@@ -1749,20 +1742,18 @@ where
...
-        loop {
-            let n = space.crypto_stream.read(&mut buf).unwrap();
-            if n == 0 {
-                return Ok(());
-            }
-            trace!("consumed {} CRYPTO bytes", n);
-            if self.crypto.read_handshake(&buf[..n])? {
+        while let Some(buf) = space.crypto_stream.read(usize::MAX).unwrap() {
+            trace!("consumed {} CRYPTO bytes", buf.len());
+            if self.crypto.read_handshake(&buf)? {
```
Matthias247, line 1751: What if the crypto stream doesn't consume all bytes (e.g. because they might only be required in a later phase?). IT seems like those are lost then. Is it guaranteed that this won't happen.

Ralith: `read_handshake` must buffer internally; there's no way for it to express incomplete reads, and this is consistent with [the behavior of rustls](https://docs.rs/rustls/0.19.0/rustls/quic/trait.QuicExt.html#tymethod.read_hs).

PR #987, opened 2021-01-25
Sound addr

### LINKED ISSUE #968: Invalid assumption about SocketAddrV{4,6} layout

est31:
Fixes #968 for the upcoming release.

### DIFF
--- a/quinn/src/platform/unix.rs
+++ b/quinn/src/platform/unix.rs
@@ -2,7 +2,7 @@ use std::{
     io,
     io::IoSliceMut,
     mem::{self, MaybeUninit},
-    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6},
+    net::{IpAddr, SocketAddr, SocketAddrV4, SocketAddrV6},
     os::unix::io::AsRawFd,
     ptr,
     task::{Context, Poll},
@@ -375,8 +375,8 @@ fn prepare_msg(
                 IpAddr::V4(v4) => {
                     let pktinfo = libc::in_pktinfo {
                         ipi_ifindex: 0,
-                        ipi_spec_dst: unsafe {
-                            *(v4 as *const Ipv4Addr as *const () as *const libc::in_addr)
+                        ipi_spec_dst: libc::in_addr {
+                            s_addr: u32::from_ne_bytes(v4.octets()),
                         },
                         ipi_addr: libc::in_addr { s_addr: 0 },
                     };
@@ -385,8 +385,8 @@ fn prepare_msg(
                 IpAddr::V6(v6) => {
                     let pktinfo = libc::in6_pktinfo {
                         ipi6_ifindex: 0,
-                        ipi6_addr: unsafe {
-                            *(v6 as *const Ipv6Addr as *const () as *const libc::in6_addr)
+                        ipi6_addr: libc::in6_addr {
+                            s6_addr: v6.octets(),
                         },
                     };
                     encoder.push(libc::IPPROTO_IPV6, libc::IPV6_PKTINFO, pktinfo);
--- a/quinn/src/platform/unix.rs
+++ b/quinn/src/platform/unix.rs
@@ -198,9 +198,25 @@ fn send(io: &mio::net::UdpSocket, transmits: &[Transmit]) -> io::Result<usize> {
     let mut msgs: [libc::mmsghdr; BATCH_SIZE] = unsafe { mem::zeroed() };
     let mut iovecs: [libc::iovec; BATCH_SIZE] = unsafe { mem::zeroed() };
     let mut cmsgs = [cmsg::Aligned([0u8; CMSG_LEN]); BATCH_SIZE];
+    // This assume_init looks a bit weird because one might think it
+    // assumes the SockAddr data to be initialized, but that call
+    // refers to the whole array, which itself is made up of MaybeUninit
+    // containers. Their presence protects the SockAddr inside from
+    // being assumed as initialized by the assume_init call.
+    // TODO: Replace this with uninit_array once it becomes MSRV-stable
+    let mut addrs: [MaybeUninit<socket2::SockAddr>; BATCH_SIZE] =
+        unsafe { MaybeUninit::uninit().assume_init() };
     for (i, transmit) in transmits.iter().enumerate().take(BATCH_SIZE) {
+        let dst_addr = unsafe {
+            std::ptr::write(
+                addrs[i].as_mut_ptr(),
+                socket2::SockAddr::from(transmit.destination),
+            );
+            &*addrs[i].as_ptr()
+        };
         prepare_msg(
             transmit,
+            dst_addr,
             &mut msgs[i].msg_hdr,
             &mut iovecs[i],
             &mut cmsgs[i],
@@ -233,7 +249,8 @@ fn send(io: &mio::net::UdpSocket, transmits: &[Transmit]) -> io::Result<usize> {
     let mut ctrl = cmsg::Aligned([0u8; CMSG_LEN]);
     let mut sent = 0;
     while sent < transmits.len() {
-        prepare_msg(&transmits[sent], &mut hdr, &mut iov, &mut ctrl);
+        let addr = socket2::SockAddr::from(transmits[sent].destination);
+        prepare_msg(&transmits[sent], &addr, &mut hdr, &mut iov, &mut ctrl);
         let n = unsafe { libc::sendmsg(io.as_raw_fd(), &hdr, 0) };
         if n == -1 {
             let e = io::Error::last_os_error();
@@ -334,6 +351,7 @@ const CMSG_LEN: usize = 80;
 
 fn prepare_msg(
     transmit: &Transmit,
+    dst_addr: &socket2::SockAddr,
     hdr: &mut libc::msghdr,
     iov: &mut libc::iovec,
     ctrl: &mut cmsg::Aligned<[u8; CMSG_LEN]>,
@@ -341,12 +359,15 @@ fn prepare_msg(
     iov.iov_base = transmit.contents.as_ptr() as *const _ as *mut _;
     iov.iov_len = transmit.contents.len();
 
-    let (name, namelen) = match transmit.destination {
-        SocketAddr::V4(ref addr) => (addr as *const _ as _, mem::size_of::<libc::sockaddr_in>()),
-        SocketAddr::V6(ref addr) => (addr as *const _ as _, mem::size_of::<libc::sockaddr_in6>()),
-    };
-    hdr.msg_name = name;
-    hdr.msg_namelen = namelen as _;
+    // SAFETY: Casting the pointer to a mutable one is legal,
+    // as sendmsg is guaranteed to not alter the mutable pointer
+    // as per the POSIX spec. See the section on the sys/socket.h
+    // header for details. The type is only mutable in the first
+    // place because it is reused by recvmsg as well.
+    let name = dst_addr.as_ptr() as *mut libc::c_void;
+    let namelen = dst_addr.len();
+    hdr.msg_name = name as *mut _;
+    hdr.msg_namelen = namelen;
     hdr.msg_iov = iov;
     hdr.msg_iovlen = 1;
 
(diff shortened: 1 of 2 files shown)


quinn/src/platform/unix.rs
```diff
@@ -332,19 +336,55 @@ pub fn caps() -> UdpCapabilities {
 
 const CMSG_LEN: usize = 80;
 
+#[inline]
+fn in_addr(addr: &Ipv4Addr) -> libc::in_addr {
+    libc::in_addr { s_addr: u32::from_ne_bytes(addr.octets()) }
+}
+
+enum CSocketAddr {
```
Matthias247, line 344: Why not use `socket2` instead of reinventing this, if that already provides the guarantee?

And I think just changing the socket type in `Transmit` to `socket2::SocketAddr` might still be an interesting option. The main advantage that std could yield is no-std compatibility. But a quic library only really matters for platforms which have a good network stack - and this case it is super likely that `socket2` is available for them too.

est31: > Why not use socket2 instead of reinventing this

Interesting idea. Apparently it's already used by dependencies of quinn so adding it as dependency wouldn't bloat the cargo graph. However, I can't find a way to get a mutable pointer without violating const correctness. It can only obtain const pointers.

> And I think just changing the socket type in Transmit to socket2::SocketAddr might still be an interesting option.

I've thought about that but it would have involved changes in all the places the Transmit struct is used. I also wanted to make a patch that one can backport safely to older releases without breaking the API.

quinn/src/platform/unix.rs
```diff
@@ -332,19 +336,55 @@ pub fn caps() -> UdpCapabilities {
 
 const CMSG_LEN: usize = 80;
 
+#[inline]
```
Ralith, line 339: I don't think this annotation is justified on internal functions (also below) unless we've shown that the optimizer gets it wrong and it matters.

quinn/src/platform/unix.rs
Ralith: The macos issues found by CI in particular seem like a good argument to not duplicate this functionality, assuming socket2 is small enough that we shouldn't be looking to get rid of it in the first place. I think the mutability issue can be avoided by using [`from_raw_parts`](https://docs.rs/socket2/0.3.19/socket2/struct.SockAddr.html#method.from_raw_parts). A minimum-impact, backportable change does seem like a good place to start.

est31: > I think the mutability issue can be avoided by using from_raw_parts.

That function takes in a const pointer of the C representation and can be used to obtain the Rust representation. What we need is the opposite way: given the Rust representation, how can you obtain a *mutable* pointer to the C representation?

I don't think there is a fix for this without changes of the socket2 crate.

Ralith: > given the Rust representation, how can you obtain a mutable pointer to the C representation?

Why do you need this?

est31: @Ralith check the code. We do `hdr.msg_name = name;`. That hdr is of type [`libc::msghdr`](https://docs.rs/libc/0.2.82/libc/struct.msghdr.html) and its `msg_name` member is of type `*mut c_void`. So `name` has to be a mutable thing for const correctness.

See also https://github.com/rust-lang/rust-clippy/issues/4774#issuecomment-565651216 ... the `*mut` may have been a `*const` previously, but it may *not* originate from a `&T` reference but has to come from a `&mut T`. socket2's `as_ptr` function though takes in a `&self` and not a `&mut self`.

est31: I guess what one *could* do would be to copy from the `*const` pointer to a newly assigned location and then create a mutable pointer for that. It'd be a bit ugly but the code would only live in one location. I can change it if you want.

Matthias247: Hmm, is this technically a bug with `libc` then, or would it need two different versions of `msghdr` (a mutable and non mutable one)?

the iov which is passed is also am mutable pointer, but we are passing it an immutable one. IDK, we know that libc doesn't mutate those things and just requires requires an immutable pointer for sending.

Ralith: Casting `*const` to `*mut` is perfectly fine here because we know no mutation will occur. Pointers are not references.

est31: > we know that libc doesn't mutate those things

Do you have sources for this claim? Or in other words, does the posix spec mandate this or can a random libc implementation just mutate if it wants to?

est31: Hmmm....

> The msghdr structure is used to minimize the number of directly supplied parameters to the recvmsg() and sendmsg() functions. This structure is used as a value- result parameter in the recvmsg() function and value only for the sendmsg() function.

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_socket.h.html

I guess I'll just use socket2 then and just cast away the constness.

quinn/src/platform/unix.rs
```diff
@@ -198,9 +198,18 @@ fn send(io: &mio::net::UdpSocket, transmits: &[Transmit]) -> io::Result<usize> {
     let mut msgs: [libc::mmsghdr; BATCH_SIZE] = unsafe { mem::zeroed() };
     let mut iovecs: [libc::iovec; BATCH_SIZE] = unsafe { mem::zeroed() };
     let mut cmsgs = [cmsg::Aligned([0u8; CMSG_LEN]); BATCH_SIZE];
+    let mut addrs: [MaybeUninit<socket2::SockAddr>; BATCH_SIZE] = unsafe {
+        MaybeUninit::uninit().assume_init()
```
faern, line 202: This defeats the purpose of `MaybeUninit`. You should not call `assume_init` until the value has been properly initialized. Which you do further down. Also I think type inference plays a trick here and the value is not what you expect it to be unless I'm misunderstanding. The type of `addrs` is supposed to be uninitialized, but you assume init it. You probably want:
```rust
let mut addrs: [MaybeUninit<socket2::SockAddr>; BATCH_SIZE] = [MaybeUninit::uninit(); BATCH_SIZE];
```

est31: I took it from the [MaybeUninit docs](https://doc.rust-lang.org/stable/core/mem/union.MaybeUninit.html#initializing-an-array-element-by-element). Basically, the `MaybeUninit::uninit().assume_init()` desugars to `MaybeUninit::<[MaybeUninit<socket2::SockAddr>; BATCH_SIZE]>::uninit().assume_init()`. Doing that it safe according to the docs because the inner MaybeUninit is allowed to be uninitialized.

Your suggestion doesn't compile on stable because SockAddr does not impl Copy and the const expr repeat RFC is not stable (and even if it were, I'd like something that compiles on older MSRVs too as I want to backport this patch).

Yes, the MaybeUninit APIs around arrays are a bit weird. There are a couple of unstable proposals to improve upon this like [`uninit_array`](https://doc.rust-lang.org/stable/core/mem/union.MaybeUninit.html#method.uninit_array) but we can't use those due to their unstable-ness :).

quinn/src/platform/unix.rs
```diff
@@ -198,9 +198,19 @@ fn send(io: &mio::net::UdpSocket, transmits: &[Transmit]) -> io::Result<usize> {
     let mut msgs: [libc::mmsghdr; BATCH_SIZE] = unsafe { mem::zeroed() };
     let mut iovecs: [libc::iovec; BATCH_SIZE] = unsafe { mem::zeroed() };
     let mut cmsgs = [cmsg::Aligned([0u8; CMSG_LEN]); BATCH_SIZE];
+    let mut addrs: [MaybeUninit<socket2::SockAddr>; BATCH_SIZE] =
+        unsafe { MaybeUninit::uninit().assume_init() };
```
djc, line 202: I agree with @faern that this looks pretty surprising. Can you add a comment outlining your take on why this is okay, maybe referring to the unstable features that might help here in the future?

quinn/src/platform/unix.rs
```diff
@@ -340,19 +351,22 @@ fn in_addr(addr: &Ipv4Addr) -> libc::in_addr {
...
     iov.iov_len = transmit.contents.len();
 
-    let (name, namelen) = match transmit.destination {
-        SocketAddr::V4(ref addr) => (addr as *const _ as _, mem::size_of::<libc::sockaddr_in>()),
-        SocketAddr::V6(ref addr) => (addr as *const _ as _, mem::size_of::<libc::sockaddr_in6>()),
-    };
-    hdr.msg_name = name;
-    hdr.msg_namelen = namelen as _;
+    // SAFETY: Casting the pointer to a mutable one is legal,
+    // as sendmsg is guaranteed to not alter the mutable pointer.
```
djc, line 363: Since you looked into it, maybe shortly explain who/what makes this guarantee?

quinn/src/platform/unix.rs
```diff
@@ -340,19 +351,22 @@ fn in_addr(addr: &Ipv4Addr) -> libc::in_addr {
 
 fn prepare_msg(
     transmit: &Transmit,
+    dest_addr: &socket2::SockAddr,
```
djc, line 354: Nit: let's use `dst` prefix here, as we do elsewhere.

djc (commented):
Thanks for working on this! Some nits and notes on extra comments...

quinn/src/platform/unix.rs
faern: Oh darn. Did not realize `MaybeUninit<T>` was not always `Copy`. You are right. But very very strange still. We really need const repeat for sure.

djc (dismissed):
LGTM. @Ralith, do you want to make a final round?

quinn/src/platform/unix.rs
```diff
@@ -198,9 +198,25 @@ fn send(io: &mio::net::UdpSocket, transmits: &[Transmit]) -> io::Result<usize> {
...
+    // TODO: Replace this with uninit_array once it becomes MSRV-stable
+    let mut addrs: [MaybeUninit<socket2::SockAddr>; BATCH_SIZE] =
+        unsafe { MaybeUninit::uninit().assume_init() };
     for (i, transmit) in transmits.iter().enumerate().take(BATCH_SIZE) {
+        let dst_addr = unsafe {
+            std::ptr::write(
+                addrs[i].as_mut_ptr(),
+                socket2::SockAddr::from(transmit.destination),
+            );
+            &mut *(&mut addrs[i] as *mut _ as *mut _)
```
Ralith, line 215: Looks like we only pass this as a shared reference. This can also be a lot more concise.
```suggestion
            &*addrs[i].as_ptr()
```

quinn/src/platform/unix.rs
```diff
@@ -375,8 +396,8 @@ fn prepare_msg(
                 IpAddr::V4(v4) => {
                     let pktinfo = libc::in_pktinfo {
                         ipi_ifindex: 0,
-                        ipi_spec_dst: unsafe {
-                            *(v4 as *const Ipv4Addr as *const () as *const libc::in_addr)
+                        ipi_spec_dst: libc::in_addr {
+                            s_addr: u32::from_ne_bytes(v4.octets()),
```
Ralith, line 400: I don't think `from_ne_bytes` is correct here. `s_addr` should be in network byte order. We should probably have unit tests for this code...

Ralith (dismissed):

quinn/src/platform/unix.rs
est31: Good catch. The old code was a relic from the times when I still passed a mutable reference.

quinn/src/platform/unix.rs
est31: > I don't think from_ne_bytes is correct here. s_addr should be in network byte order. 

`from_ne_bytes` takes the 4 bytes and copies them into the resulting `u32` without changing byte order. As the `octets()` function returns in network order, `from_ne_bytes` is what we need to use. There are tests for this code in fact. Try changing to `from_be_bytes` and some tests will fail :). If you are still not convinced, check the socket2 code, [it does the same](https://docs.rs/socket2/0.3.19/src/socket2/sockaddr.rs.html#206).

djc (approved):

djc:
Thanks for working on this, going to merge this now!

Issue #984, opened 2021-01-25
Reading is prone to spurious driver wakeups


Ralith:
`quinn::RecvStream::poll_read_generic` currently wakes the connection driver on every successful operation in case we need to transmit flow control credit. As of #880 and #894 we no longer issue credit after every read, so we can reduce CPU load by only waking the connection driver if flow control actually does need to be issued.

geieredgar:
Now that #992 has landed in main via #991, we could close this issue.

djc:
Yup!

PR #990, opened 2021-01-26
Wake up the connection driver only if flow control needs to be issued

geieredgar:
This tries to fix issue #984. It makes the `DidRead` type of quinn-proto public and changes `quinn_proto::Connection::read_unordered`, `quinn_proto::Connection::read`, and `quinn_proto::Connection::read_chunks` to return `Result<Option<DidRead<U>>, ReadError>` instead of `Result<Option<U>, ReadError>`.
With that we can check in `quinn::RecvStream::poll_read_generic` if we need to wake up the connection driver.

### DIFF
--- a/quinn-proto/src/connection/streams/recv.rs
+++ b/quinn-proto/src/connection/streams/recv.rs
@@ -248,6 +248,27 @@ pub struct DidRead<T> {
     pub(crate) max_data: ShouldTransmit,
 }
 
+impl<T> DidRead<T> {
+    /// Returns `true` iff the producing `Streams::read` call causes a transmit.
+    pub fn causes_transmit(&self) -> bool {
+        self.max_stream_data.should_transmit() || self.max_data.should_transmit()
+    }
+
+    /// Maps an `DidRead<T>` to `DidRead<U>` by applying a function to the contained result.
+    pub fn map<U, F: FnOnce(T) -> U>(self, f: F) -> DidRead<U> {
+        DidRead {
+            result: f(self.result),
+            max_stream_data: self.max_stream_data,
+            max_data: self.max_data,
+        }
+    }
+
+    /// Returns the contained result, consuming the `self` value.
+    pub fn into_inner(self) -> T {
+        self.result
+    }
+}
+
 pub(super) type StreamReadResult<T> = Result<Option<T>, ReadError>;
 
 pub(crate) trait BytesRead {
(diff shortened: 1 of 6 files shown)


quinn-proto/src/connection/streams/recv.rs
```diff
@@ -248,6 +248,27 @@ pub struct DidRead<T> {
...
+    pub fn map<U, F: FnOnce(T) -> U>(self, f: F) -> DidRead<U> {
+        DidRead {
+            result: f(self.result),
+            max_stream_data: self.max_stream_data,
+            max_data: self.max_data,
+        }
+    }
+
+    /// Returns the contained result, consuming the `self` value.
+    pub fn unwrap(self) -> T {
```
Ralith, line 267: `unwrap` is conventionally used for functions that may panic; let's call this `into_inner` or `into_result` or similar instead.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -248,6 +248,27 @@ pub struct DidRead<T> {
     pub(crate) max_data: ShouldTransmit,
 }
 
+impl<T> DidRead<T> {
+    /// Returns `true` iff flow control credit should be issued.
+    pub fn should_issue_flow_control_credit(&self) -> bool {
```
Ralith, line 253: This is awfully verbose. Maybe `causes_transmit`? External crates (such as `quinn`) don't need to know *why* there's a transmit needed.

Ralith (changes requested):
Thanks for taking a swing at this!

djc:
Thanks you for working on this! I'll have to confess that I've been working on taking a swing at a pretty sweeping redesign of this part of quinn-proto that obviates a lot of this stuff, so we'll have to see how that might fit together.

quinn-proto/src/connection/streams/recv.rs
geieredgar: OK, I renamed it to `causes_transmit` and adjusted the doc comment.

quinn-proto/src/connection/streams/recv.rs
geieredgar: OK, I renamed it to `into_inner`.

Ralith (approved):

geieredgar:
@djc I opened a new pull request (#992) based on your changes in #991 to merge into the `reading` branch.

djc:
Closing this in favor of #991, thanks for the work!

Issue #912, opened 2020-11-13
H3 test go_away_from_client (intermittently?) fails in CI


djc:
Happened here, in testing a completely unrelated PR:

https://github.com/quinn-rs/quinn/pull/909/checks?check_run_id=1396835062

```
---- tests::go_away_from_client stdout ----
thread 'tests::go_away_from_client' panicked at 'assertion failed: `Ok(())` does not match `Err(Error::Http(HttpError::RequestRejected, None))`', quinn-h3/src/tests/mod.rs:295:5
```

I haven't seen this before.

stammw:
We could just remove or ignore those `h3`-related tests ? How do you see the future of this crate ?

djc:
Unfortunately I haven't been following along with hyper's h3 work very closely, so you probably have a better read on the future of this crate than I do. I seem to remember you preferred this crate's API over Sean's proposed API for the hyper efforts?

stammw:
For the moment, Sean's API implementation is progressing quite slowly. It is now able to run requests without QPACK or connection life-cycle feature.

I like `quinn-h3` API but I don't really have any preference. 

I don't know what to do with `quinn-h3`, it can't stay pending forever. Releasing it as an alpha would be nice, but I worry about my available bandwidth to maintain it, if it ever gets popular. 

I think we cannot remove it as it is used in interop, and the efforts to transition from `quinn-h3` to `h3` in interop module don't really worth it.

Ralith:
Without speaking to the status of quinn-h3, I think it's fair to see the (h3-related) aspects of the interop infra as being logically part of the h3 implementation. If there's a need for a hyper-flavored interop tool, it should probably be developed as part of hyper.

djc:
I do get the feeling the interop stuff is being deemphasized upstream in favor of the QUIC network simulator (and we might be falling behind a bit as long as we don't have a QNS setup).

djc:
Another reason not to dump quinn-h3 just yet is that I've found in my recent refactoring that quinn-h3 tests exercise some code paths that our other tests do not, so keeping h3 around does improve code coverage for quinn/quinn-proto for now.

Ralith:
Closing this as we've removed the test in question.

PR #991, opened 2021-01-26
Unified reading

djc:
TODO:

- [x] Update documentation
- [x] Check benchmarks to make sure there are no regressions
- [x] Check if the last commit can be split into smaller parts

### DIFF
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -12,7 +12,7 @@ use crate::range_set::RangeSet;
 #[derive(Debug, Default)]
 pub(crate) struct Assembler {
     state: State,
-    data: BinaryHeap<Chunk>,
+    data: BinaryHeap<Buffer>,
     defragmented: usize,
     /// Number of bytes read by the application. When only ordered reads have been used, this is the
     /// length of the contiguous prefix of the stream which has been consumed by the application,
@@ -113,14 +113,14 @@ impl Assembler {
                 }
             } else {
                 let bytes = buffer.split().freeze();
-                self.data.push(Chunk { offset, bytes });
+                self.data.push(Buffer { offset, bytes });
                 offset = chunk.offset;
                 buffer.extend_from_slice(&chunk.bytes);
             }
         }
 
         let bytes = buffer.split().freeze();
-        self.data.push(Chunk { offset, bytes });
+        self.data.push(Buffer { offset, bytes });
         self.defragmented = self.data.len();
     }
 
@@ -131,7 +131,7 @@ impl Assembler {
             // Discard duplicate data
             for duplicate in recvd.replace(offset..offset + bytes.len() as u64) {
                 if duplicate.start > offset {
-                    self.data.push(Chunk {
+                    self.data.push(Buffer {
                         offset,
                         bytes: bytes.split_to((duplicate.start - offset) as usize),
                     });
@@ -144,7 +144,7 @@ impl Assembler {
         if bytes.is_empty() || self.stopped {
             return;
         }
-        self.data.push(Chunk { offset, bytes });
+        self.data.push(Buffer { offset, bytes });
         // Why 32: on the one hand, we want to defragment rarely, ideally never
         // in non-pathological scenarios. However, a pathological or malicious
         // peer could send us one-byte frames, and since we use reference-counted
@@ -190,15 +190,15 @@ impl Assembler {
 }
 
 #[derive(Debug, Eq)]
-struct Chunk {
+struct Buffer {
     offset: u64,
     bytes: Bytes,
 }
 
-impl Ord for Chunk {
+impl Ord for Buffer {
     // Invert ordering based on offset (max-heap, min offset first),
     // prioritize longer chunks at the same offset.
-    fn cmp(&self, other: &Chunk) -> Ordering {
+    fn cmp(&self, other: &Buffer) -> Ordering {
         self.offset
             .cmp(&other.offset)
             .reverse()
@@ -206,14 +206,14 @@ impl Ord for Chunk {
     }
 }
 
-impl PartialOrd for Chunk {
-    fn partial_cmp(&self, other: &Chunk) -> Option<Ordering> {
+impl PartialOrd for Buffer {
+    fn partial_cmp(&self, other: &Buffer) -> Option<Ordering> {
         Some(self.cmp(other))
     }
 }
 
-impl PartialEq for Chunk {
-    fn eq(&self, other: &Chunk) -> bool {
+impl PartialEq for Buffer {
+    fn eq(&self, other: &Buffer) -> bool {
         (self.offset, self.bytes.len()) == (other.offset, other.bytes.len())
     }
 }
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -70,6 +70,7 @@ impl Assembler {
             } else if (chunk.offset + chunk.bytes.len() as u64) <= self.bytes_read {
                 // Next chunk is useless as the read index is beyond its end
                 PeekMut::pop(chunk);
+                self.defragmented = self.defragmented.saturating_sub(1);
                 continue;
             }
 
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -29,12 +29,17 @@ impl Assembler {
         Self::default()
     }
 
-    pub(crate) fn read_unordered(&mut self) -> Result<Option<(u64, Bytes)>, AssembleError> {
+    // Get the the next ordered chunk
+    pub(crate) fn read(
+        &mut self,
+        max_length: usize,
+        ordered: bool,
+    ) -> Result<Option<(u64, Bytes)>, AssembleError> {
         if self.is_stopped() {
             return Err(AssembleError::UnknownStream);
-        }
-
-        if let State::Ordered = self.state {
+        } else if ordered && !self.state.is_ordered() {
+            return Err(AssembleError::IllegalOrderedRead);
+        } else if !ordered && self.state.is_ordered() {
             // Enter unordered mode
             let mut recvd = RangeSet::new();
             recvd.insert(0..self.bytes_read);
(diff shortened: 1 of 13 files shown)


djc:
Tests pass! Would like a preliminary review on this before I polish this up (also, suggestions on how to split it up more).

djc:
(Failure is #912.)

quinn-proto/src/connection/assembler.rs
```diff
@@ -44,51 +49,41 @@ impl Assembler {
...
+                let offset = chunk.offset;
                 chunk.offset += max_length as u64;
-                chunk.bytes.split_to(max_length)
+                (offset, chunk.bytes.split_to(max_length))
             } else {
                 self.bytes_read += chunk.bytes.len() as u64;
                 self.defragmented = self.defragmented.saturating_sub(1);
-                PeekMut::pop(chunk).bytes
+                let chunk = PeekMut::pop(chunk);
+                (chunk.offset, chunk.bytes)
```
Ralith: We should probably respect `max_length` here too.

quinn-proto/src/connection/mod.rs
```diff
@@ -889,23 +889,23 @@ where
...
 
-    /// Read from the given recv stream, in undefined order
+    /// Read from the given recv stream
     ///
-    /// While stream data is typically processed by applications in-order, unordered reads improve
-    /// performance when packet loss occurs and data cannot be retransmitted before the flow control
-    /// window is filled. When in-order delivery is required, the sibling `read()` or `read_chunk[s]()`
-    /// methods should be used.
+    /// If `Ok`, the return value contains the bytes and their offset in the stream. For ordered
+    /// reads, the offset could be ignored, since it provides no new information.
```
Ralith, line 897: Let's actually define "ordered" here, e.g. as "having an offset exactly equal to the previously returned offset plus the previously returned bytes' length"

quinn-proto/src/connection/assembler.rs
```diff
@@ -189,6 +189,22 @@ impl Assembler {
     }
 }
 
+/// A chunk of data from the receive stream
+#[non_exhaustive]
```
Ralith, line 232: What fields might we add to this in the future?

quinn-proto/src/connection/assembler.rs
```diff
@@ -205,6 +244,13 @@ impl Chunk {
     }
 }
 
+enum NextState<'a> {
+    Value(PeekMut<'a, Buffer>),
+    Blocked,
+    Retry,
+    Empty,
```
Ralith, line 251: Should `Empty` and `Blocked` be combined? I don't see any case where we handle them differently.

quinn-proto/src/connection/assembler.rs
```diff
@@ -49,43 +41,90 @@ impl Assembler {
...
+        }
+    }
 
-                // Determine `start` and `len` of the slice of useful data in chunk
-                let start = (self.bytes_read - chunk.offset) as usize;
-                if start > 0 {
-                    chunk.bytes.advance(start);
-                    chunk.offset += start as u64;
-                }
+    fn next<'a>(
```
Ralith, line 97: Should have a comment noting that this is mostly a workaround for borrowck, and link the relevant issue for future reference.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -226,37 +195,105 @@ impl Recv {
...
+        };
 
-pub(crate) trait BytesRead {
-    fn bytes_read(&self) -> u64;
-}
+        let rs = entry.get_mut();
+        let recv = if !rs.assembler.can_read(ordered)? {
+            match rs.read_blocked() {
+                Ok(()) => {
+                    entry.remove_entry();
```
Ralith, line 225: Let's use `Entry::remove` consistently here and below, since we never need to get the ID from it.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -226,37 +195,105 @@ impl Recv {
...
-    pub bufs: usize,
-    pub read: usize,
+    /// Read the next chunk from the stream
+    ///
+    /// Return the next chunk of data from the stream, if any are available. Will return `None` if
+    /// all available data has been read. `max_length` may be used to restrict the maximum size
+    /// of the returned `Bytes` (but `usize::MAX` is supported, too).
+    ///
+    /// Will keep returning `None` once no further data is available for reading.
+    pub fn next(&mut self, max_length: usize) -> Option<Chunk> {
```
Ralith, line 258: If we moved `max_length` into `self`, we could implement `Iterator` in earnest, and also make it easier for the caller to limit the *total* length rather than the length of an individual chunk (which admittedly doesn't seem important for our current use, but is IMO a nicer API in general).

quinn/src/streams.rs
```diff
@@ -353,42 +353,38 @@ where
         buf: &mut ReadBuf<'_>,
     ) -> Poll<Result<(), ReadError>> {
         self.poll_read_generic(cx, |conn, stream| {
-            conn.inner
-                .read(stream, buf.remaining())
-                .map(|val| val.map(|chunk| buf.put_slice(&chunk)))
+            let mut chunks = conn.inner.read(stream, true)?;
```
Matthias247, line 356: Just a confirmation question: `chunks` is the iterator here, which directly points to the assembler, right? So this won't require a new stream lookup on every read? If that's true, then I'm fine with it. However I would really like to avoid having to go through the whole stack for retrieving 1kB of data. hashmap lookups are fast - but not that fast :-)

quinn/src/streams.rs
```diff
@@ -353,42 +353,38 @@ where
...
+            while dbg!(buf.remaining()) > 0 {
```
Matthias247, line 357: Remove `dbg!` before shipping this. Good idea is also to run the benchmark, and see if there is any regression (there will definitely be with extra output).

quinn/src/streams.rs
Ralith: The current impl moves the stream state from the hash table into the iterator for easy access, then sticks it back when it's done, which avoids O(n) hash lookups. We could probably take a reference instead with sufficient refactoring to satisfy borrowck, but that should probably be left for future work.

quinn-proto/src/connection/mod.rs
```diff
@@ -924,8 +923,16 @@ where
 
     fn post_read<T>(&mut self, id: StreamId, result: &streams::ReadResult<T>) {
         if let Ok(Some(ref result)) = *result {
-            self.add_read_credits(id, result.max_stream_data, result.max_data);
+            let pending = &mut self.spaces[SpaceId::Data].pending;
+            if result.max_data.should_transmit() {
+                pending.max_data = true;
+            }
+            if result.max_stream_data.should_transmit() {
+                // Only bother issuing stream credit if the peer wants to send more
```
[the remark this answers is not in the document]
djc: Yup, removed it.

quinn-proto/src/connection/assembler.rs
djc: What do you mean? This is the `else` branch for `if max_length < chunk_bytes.len()`, so we should already respect it, right?

quinn-proto/src/connection/mod.rs
djc: Revised the comments here. Related question: we used to have some docstrings stating that unordered reads could yield chunks in any order and might yield duplicate data, but that's no longer true, right? As far a I understand it, we always yield chunks in order and we now deduplicate them on insertion when the assembler is in unordered mode because we need to do flow control.

Actually, I wonder if what we do is correct and the most simple thing we could do here. In my understanding, the assembler deduplicates on read when in ordered mode whereas we deduplicate on ingest in unordered mode. Could we deduplicate on read always to simplify the code? The deduplication in unordered mode is currently a little hairy. Or, vice versa, isn't deduplication on read too lazy?

quinn-proto/src/connection/assembler.rs
djc: I couldn't think of any; just being defensive, maybe overly so?

quinn-proto/src/connection/assembler.rs
djc: I'm scrapping this part of the PR to land some of the less controversial parts. I think the `read(max_length) -> Option<Chunk>` API is not expressive enough, and therefore maybe prone to misuse.

quinn-proto/src/connection/assembler.rs
djc: (No longer relevant, removed for now.)

quinn-proto/src/connection/streams/recv.rs
djc: Scrapped.

quinn-proto/src/connection/streams/recv.rs
djc: I did consider this, but passing `max_length` on every iteration seemed more flexible (the caller could decide to change the `max_length` in further iterations) and requires no extra complexity on the part of the assembler.

quinn/src/streams.rs
djc: The refactoring would be tricky since, when the `Chunks` is dropped, it may want to call `Streams::stream_freed()`, which actually manipulates the `recv` hash map.

quinn/src/streams.rs
djc: Yeah, will of course remove that.

quinn-proto/src/connection/streams/recv.rs
Ralith: Fair enough; no strong feeling.

djc:
Benchmarks for this scaled back version, before:

```
test large_data_10_streams  ... bench:  60,531,140 ns/iter (+/- 3,931,156) = 173 MB/s
test large_data_1_stream    ... bench:   6,416,910 ns/iter (+/- 1,427,494) = 163 MB/s
test small_data_100_streams ... bench:   1,402,531 ns/iter (+/- 400,224)
test small_data_1_stream    ... bench:     107,698 ns/iter (+/- 62,181)
```

and after:

```
test large_data_10_streams  ... bench:  60,043,676 ns/iter (+/- 3,826,160) = 174 MB/s
test large_data_1_stream    ... bench:   6,212,213 ns/iter (+/- 1,513,790) = 168 MB/s
test small_data_100_streams ... bench:   1,333,047 ns/iter (+/- 309,778)
test small_data_1_stream    ... bench:      90,338 ns/iter (+/- 32,704)
```

Or, about 1, 3, 5, 17% faster respectively (though all well within the noise). Looks like we're especially faster for smaller data/fewer streams.

quinn-proto/src/connection/mod.rs
Ralith: > we always yield chunks in order

This is only true if we're not missing any chunks. Unordered reads can return out-of-order chunks due to network errors or weird peer behavior, whereas ordered reads just block in those cases.

> Could we deduplicate on read always?

Consider this situation:
1. Peer sends offset 5 on a new stream
2. Application reads offset 5 in unordered mode
3. Peer sends offset 5 again

We must not issue flow control credit for the same offset twice. However, we've already evicted the first copy of that data from the buffer. In ordered mode, this is trivial because we've issued flow control credit for an offset iff it's less than `self.bytes_read`. In unordered mode, the ranges consumed by the application may be sparse, so we need the `RangeSet`.

> Or, vice versa, isn't deduplication on read too lazy?

We must also avoid spending unbounded buffer space on data that we've already issued flow control credit for, so we must deduplicate on insert rather than waiting an unbounded duration for the application to get around to reading. I think you're right that we currently have a bug here: in ordered mode, offsets prior to `self.bytes_read` should be immediately dropped, but currently are not. In unordered mode we're behaving correctly.

quinn-proto/src/connection/assembler.rs
Ralith: Whoops, confused this with the `if ordered` branch above. My bad.

quinn-proto/src/connection/mod.rs
djc: > This is only true if we're not missing any chunks. Unordered reads can return out-of-order chunks due to network errors or weird peer behavior, whereas ordered reads just block in those cases.

Ah, fair.

> In unordered mode, the ranges consumed by the application may be sparse, so we need the RangeSet.
> 
> It is true that we can now guarantee not returning duplicate data, as an incidental benefit of the bookkeeping required to avoid issuing duplicate flow control credit.

Right.

> I think you're right that we currently have a bug here: in ordered mode, offsets prior to self.bytes_read should be immediately dropped, but currently are not. In unordered mode we're behaving correctly.

Good -- that should be a straightforward fix.

quinn-proto/src/connection/assembler.rs
Ralith: My inclination is to reserve `non_exhaustive` for cases where we're pretty sure data will be added, at least so long as we have no reason to expect minor breakage to be disruptive.

quinn-proto/src/connection/mod.rs
```diff
@@ -889,23 +889,30 @@ where
...
     ///
-    /// While stream data is typically processed by applications in-order, unordered reads improve
-    /// performance when packet loss occurs and data cannot be retransmitted before the flow control
-    /// window is filled. When in-order delivery is required, the sibling `read()` or `read_chunk[s]()`
-    /// methods should be used.
+    /// `max_length` limits the maximum size of the returned `Bytes` value; passing `usize::MAX`
+    /// will yield the best performance. Chunk sizes should be no larger than the UDP MTU (in the
+    /// 1500 bytes range, unless IPv6 jumbograms are supported). `ordered` will make sure the
+    /// returned chunk's offset will have an offset exactly equal to the previously returned offset
+    /// plus the previously returned bytes' length.
```
Ralith, line 900: I don't think we should make any guarantees (let alone fuzzy ones) about chunk sizes.

quinn/src/streams.rs
```diff
@@ -11,7 +11,7 @@ use futures::{
     io::{AsyncRead, AsyncWrite},
     ready, FutureExt,
 };
-use proto::{ConnectionError, FinishError, StreamId};
+use proto::{Chunk, ConnectionError, FinishError, StreamId};
```
Ralith, line 14: We need to also reexport this at the crate level.

Ralith (changes requested):

quinn/src/streams.rs
djc: I think that's there already?

djc: Done.

quinn-proto/src/connection/mod.rs
djc: Removed that part of the text.

quinn-proto/src/connection/assembler.rs
djc: Removed it.

Ralith (approved):

PR #992, opened 2021-01-26
Wake up the connection driver only if flow control needs to be issued (based on #991)

geieredgar:
Add a has_pending_retransmits method to
quinn_proto::connection::Connection and use it inside of
quinn::RecvStream::poll_read_generic to decide if we should wake up the
connection driver.


djc:
This is great, thanks! I will likely make some further iterations on what this API looks like, but this is a good start in any case. Sorry for the time spent on your other PR!

PR #993, opened 2021-01-27
quinn: properly await client connection setup in benchmarks

djc:


Ralith (changes requested):
I don't think this is correct. First, `async { x.await }` should always be roughly equivalent to an unadorned `x`. Second, thread-local RAII guards like `runtime.enter()` don't work right in an async block since control flow can exit the block without passing through the `Drop` impl.

djc:
Right, the `runtime.enter()` isn't needed anymore, dropped it. On the other hand, the wrapping in an async block was recommended to me by Alice, and I think it makes sense: we need the runtime to be active while the `connect_with()` call is being evaluated, which it wouldn't be without wrapping it in an `async` block.

Ralith:
That's what the prior `runtime.enter()` call was accomplishing, though. I wouldn't expect these changes to have any effect. Do they fix the benchmarks in practice?

djc:
I suspect it may have needed an explicitly bound guard. Would you prefer that over the async block? (Yes, this fixed the issue I was seeing.)

Ralith (approved):
Oh, duh. No, the async block is probably better because it's less prone to that type of mistake.

Issue #981, opened 2021-01-20
Assembler::defragment causes high CPU load


Matthias247:
Here is a flamegraph of a benchmark I ran yesterday:

![image](https://user-images.githubusercontent.com/6330334/105130497-04641d00-5a9c-11eb-88a9-66837335c75c.png)

This one is a on loopback connection with simulated latency, which leads to rather high congestion windows (if flow control limits are also high).

In this case we can observe the receiving end of the connection drowning in `defragment` calls, which take up the majority of the CPU time.

I think there is a major contributing factor here:
[The assembler kicks in if at least 32 packets are queued](https://github.com/quinn-rs/quinn/blob/main/quinn-proto/src/connection/assembler.rs#L213-L215), which will likely always be the case in case a burst of packets is received. I think this can even happen in a less artificial scenario.

I think it might be worthwhile to rethink this strategy - especially for users which intend to receive lots of data.
One option I can imagine is to skip defragmentation of chunks which make up the majority of the MTU (e.g. > 75%). Now that's a bit tricky at this layer, because it isn't aware of the  MTU. But maybe just hardcoding it to 750 bytes might be good enough (at least as long as we don't imagine MTUs to be growing into jumbo-frame territory).

Another option is to immediately copy data of the datagram buffers into a more efficiently sized receive buffer datastructure. But that would be more interesting if the datagram wouldn't already be `Bytes` allocated earlier on and still be on the stack of the endpoint. This way it might just end up double-copying.

djc:
The defragmentation heuristic kicking in is definitely pretty simplistic (and the number 32 is pretty arbitrary), so I think the right fix here is to tweak that. How about if we take the `payload.len()` in `Connection::process_payload()` and pass it down through `Streams::received()` and `Recv::ingest()` to `Assembler::insert()` (potentially storing it in the `Chunk`)? Assuming the header size is more or less constant, we can then make a decent estimate of wasted memory and trigger defragmentation on some threshold.

Matthias247:
Sounds like a reasonable idea to me. Instead of the actual size just a fraction (0 - 100) could also be sufficient if one wants to save some data (but that might be lost due to alignment reasons anyway).

Most packets from nicely playing peers should be nearly full MTU sized. Therefore defragmentation should be rarely necessary. But I agree that it's good thing to have, in order to protect against bad peer behavior.

djc:
Yes, this code is not there for capable/well-intentioned peers, but for the other kinds. 👍

PR #996, opened 2021-01-28
Add and use over-allocation metric to decide when to defragment

geieredgar:
This pull request implements the suggestion made by @djc here: https://github.com/quinn-rs/quinn/issues/981#issuecomment-763424636

### DIFF
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -12,7 +12,7 @@ use crate::range_set::RangeSet;
 #[derive(Debug, Default)]
 pub(crate) struct Assembler {
     state: State,
-    data: BinaryHeap<Chunk>,
+    data: BinaryHeap<Buffer>,
     defragmented: usize,
     /// Number of bytes read by the application. When only ordered reads have been used, this is the
     /// length of the contiguous prefix of the stream which has been consumed by the application,
@@ -113,14 +113,14 @@ impl Assembler {
                 }
             } else {
                 let bytes = buffer.split().freeze();
-                self.data.push(Chunk { offset, bytes });
+                self.data.push(Buffer { offset, bytes });
                 offset = chunk.offset;
                 buffer.extend_from_slice(&chunk.bytes);
             }
         }
 
         let bytes = buffer.split().freeze();
-        self.data.push(Chunk { offset, bytes });
+        self.data.push(Buffer { offset, bytes });
         self.defragmented = self.data.len();
     }
 
@@ -131,7 +131,7 @@ impl Assembler {
             // Discard duplicate data
             for duplicate in recvd.replace(offset..offset + bytes.len() as u64) {
                 if duplicate.start > offset {
-                    self.data.push(Chunk {
+                    self.data.push(Buffer {
                         offset,
                         bytes: bytes.split_to((duplicate.start - offset) as usize),
                     });
@@ -144,7 +144,7 @@ impl Assembler {
         if bytes.is_empty() || self.stopped {
             return;
         }
-        self.data.push(Chunk { offset, bytes });
+        self.data.push(Buffer { offset, bytes });
         // Why 32: on the one hand, we want to defragment rarely, ideally never
         // in non-pathological scenarios. However, a pathological or malicious
         // peer could send us one-byte frames, and since we use reference-counted
@@ -190,15 +190,15 @@ impl Assembler {
 }
 
 #[derive(Debug, Eq)]
-struct Chunk {
+struct Buffer {
     offset: u64,
     bytes: Bytes,
 }
 
-impl Ord for Chunk {
+impl Ord for Buffer {
     // Invert ordering based on offset (max-heap, min offset first),
     // prioritize longer chunks at the same offset.
-    fn cmp(&self, other: &Chunk) -> Ordering {
+    fn cmp(&self, other: &Buffer) -> Ordering {
         self.offset
             .cmp(&other.offset)
             .reverse()
@@ -206,14 +206,14 @@ impl Ord for Chunk {
     }
 }
 
-impl PartialOrd for Chunk {
-    fn partial_cmp(&self, other: &Chunk) -> Option<Ordering> {
+impl PartialOrd for Buffer {
+    fn partial_cmp(&self, other: &Buffer) -> Option<Ordering> {
         Some(self.cmp(other))
     }
 }
 
-impl PartialEq for Chunk {
-    fn eq(&self, other: &Chunk) -> bool {
+impl PartialEq for Buffer {
+    fn eq(&self, other: &Buffer) -> bool {
         (self.offset, self.bytes.len()) == (other.offset, other.bytes.len())
     }
 }
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -34,7 +34,7 @@ impl Assembler {
         &mut self,
         max_length: usize,
         ordered: bool,
-    ) -> Result<Option<(u64, Bytes)>, AssembleError> {
+    ) -> Result<Option<Chunk>, AssembleError> {
         if self.is_stopped() {
             return Err(AssembleError::UnknownStream);
         } else if ordered && !self.state.is_ordered() {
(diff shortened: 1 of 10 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -2313,6 +2313,7 @@ where
         let is_0rtt = self.spaces[SpaceId::Data].crypto.is_none();
         let mut is_probing_packet = true;
         let mut close = None;
+        let payload_len = payload.len();
```
djc, line 2316: I'd prefer to just squash this commit into the previous one, as this one is not too complex, and it makes it more obvious what's going on.

quinn-proto/src/connection/mod.rs
```diff
@@ -1720,7 +1720,7 @@ where
 
         space
             .crypto_stream
-            .insert(crypto.offset, crypto.data.clone());
+            .insert(crypto.offset, crypto.data.clone(), crypto.data.len());
```
djc, line 1723: I think we should instead pass down the `payload.len()` from `process_payload()` into `read_crypto()` and use that here, does that make sense to you?

quinn-proto/src/connection/assembler.rs
```diff
@@ -152,46 +152,58 @@ impl Assembler {
...
                         bytes: bytes.split_to((duplicate.start - offset) as usize),
+                        over_allocation,
                     });
+                    self.over_allocation += over_allocation;
                     offset = duplicate.start;
                 }
                 bytes.advance((duplicate.end - offset) as usize);
                 offset = duplicate.end;
             }
+            allocation_size = allocation_size.saturating_sub((offset - start) as usize);
```
djc, line 193: I'm not sure about this. As far as I understand it, either `Chunk` could keep the entire allocation alive. It's also already the case that a payload may have multiple `STREAM` frames (either for the same stream or for different streams), so we're generally somewhat overcounting the overallocations, and we just have to take that into account when deciding to defragment (but to that end we can just fudge the defragmentation trigger heuristic).

djc (commented):
Thanks again for working on this! This one is a bit tricky, I left some feedback below.

@Matthias247, would you like to take a look at this?

djc:
(Also please mention #981 in the commit message!)

quinn-proto/src/connection/assembler.rs
```diff
@@ -152,46 +152,58 @@ impl Assembler {
...
         }
         if bytes.is_empty() || self.stopped {
             return;
         }
-        self.data.push(Chunk { offset, bytes });
-        // Why 32: on the one hand, we want to defragment rarely, ideally never
```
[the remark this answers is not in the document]
geieredgar: OK, I adjusted the comment and changed the magic number.

quinn-proto/src/connection/mod.rs
geieredgar: OK, I squashed the two commits into one.

quinn-proto/src/connection/mod.rs
geieredgar: Yeah, makes sense. Fixed it.

quinn-proto/src/connection/assembler.rs
geieredgar: I still would like the over-allocation estimate to be correct and precise if the `allocation_size` argument is also correct and precise. 
So e.g. `insert(1, Buffer of size 1022, 1022)` followed by `insert(0, Buffer of size 1024, 1024)` should increase over-allocation by 1022 and not by 2045.
I agree that we will often over-count the over-allocation, but I don't see that as a reason to make the estimate even worse.

djc: That's not really my point though. My point is this: shrinking the `Bytes` doesn't shrink the size of the allocation, it simply changes the index pointing to the allocation. So in that sense, the deduplication happening here should have, in my understanding, no effect on the amount of overallocation.

geieredgar: You are right that only advancing the `Bytes` should not shrink the `allocation_size`, but pushing a `Buffer` should, because we do not want to account for the same bytes twice. My error was that I thought that I could simply adjust the `allocation_size` at the end of the de-duplication loop, but now I see that this does not work when we are not pushing a new `Buffer` at every iteration step. I therefore moved the `allocation_size` adjustment into the loop, only adjusting it when we are pushing a new `Buffer`.

djc: But shrinking the allocation size is only warranted if we're accounting for the same *allocations*. A different but overlapping `Buffer` ("same bytes twice") is substantially more likely to originate from a different packet, and thus from a different allocation.

geieredgar: I try to clarity my thoughts a bit more:

```rust
if duplicate.start > offset {
    let over_allocation = (duplicate.end - duplicate.start) as usize;
    self.data.push(Buffer {
        offset,
        bytes: bytes.split_to((duplicate.start - offset) as usize),
        over_allocation,
    });
    self.over_allocation += over_allocation;
    allocation_size =
        allocation_size.saturating_sub((duplicate.end - offset) as usize);
    offset = duplicate.start;
}
```
Here we are pushing a new `Buffer`, because we need to close the gap to the next duplicate. This new `Buffer` references the same allocation and I use the size of the duplicate as the over-allocation for the new `Buffer`, because this is the amount of bytes that are unused but are still referenced and kept in memory by the `Buffer`. The reasons I associate a over-allocation with every `Buffer` are:
- It could be the last `Buffer` we push
- If we only keep track of over-allocation in a single `Buffer`, when that `Buffer` is later removed, we will underestimate the over-allocation caused by the remaining Buffers pushed by the `insert` call.

Because I account for the over-allocation in the pushed `Buffer`, I reduce the allocation size by the size of the `Buffer` and its over-allocation.

A possible other sensible approach would be to calculate for every pushed Buffer the over-allocation as `allocation_size.saturating_sub(bytes.len())` and never mutating the `allocation_size`. Then the  calculated over-allocation would grow much larger if we have to fill many small gaps in the de-duplication loop.

Matthias247:
I feel like any flag like `defragmented` or `overallocation` is rather challenging to maintain, since it replicates the state of the individual buffers but also needs to be in sync with those. If it isn't, then we get bugs like #982 .

I would tackle this problem a bit different than what is proposed here: I would purely add a `utilization` field to each of the chunks in the queue. When inserting, I would loop through fields, check utilization values, and if I find some chunks with bad utilization values (e.g. < 50%), I would merge those into a bigger chunk - but actually leave others alone. There might be no need to defragment everything.

One important thing: I would **not** touch the utilization value if a part of the chunk is dequeued by the user. The reason for this is that if the user is already reading data, they will read the next data chunk very soon, and it won't stay around with poor utilization for very long. The danger there only exists for chunks which can't be read by the user yet, because there is a gap in the queue.

With that approach another challenge is how not to iterate the complete list of fragments on each insert. Maybe the old `defragmented` field helps for that. Or keeping it as an offset towards the first gap (non consecutive data), which would be the starting point for investigating defragmentation?

geieredgar:
@Matthias247 I pushed a new commit that would make this approach a bit more similar to yours (although it keeps the `over_allocation` field, but like you said, we would probably need one field like `defragmented` or `over_allocation` to prevent iterating the list of fragments on every insert).

The defragment process first checks if we can reduce `over_allocation` below the specified threshold by just defragmenting low utilized `Buffer` (currently that translates to `over_allocation > bytes.len()`). In that case it only defragments those, otherwise it will only defragment buffers with `over_allocation > 0`, therefore preventing an already defragmented `Buffer` to be defragmented again.

djc:
> I feel like any flag like `defragmented` or `overallocation` is rather challenging to maintain, since it replicates the state of the individual buffers but also needs to be in sync with those. If it isn't, then we get bugs like #982.

That bug was due in part to the duplication of the read APIs, which I'm removing in #991. Yes, keeping these in sync is a bit tricky, but also has important benefits like not having to iterate over the chunks when deciding when to defragment, and after the changes in #991 isn't too hard IMO.

djc:
Uh, sorry for the automation closing this. Can you open a new one targeting main?

geieredgar:
OK, I opened a new PR here: #1000.

PR #998, opened 2021-01-30
Increase datagram utilization

Matthias247:
This change will increase the amount of data that will be written in
STREAM frames. The previous approach was pessimistic and assumed
- 8 bytes are used for Stream ID
- 8 bytes are used for the offset
- 8 bytes are used for the length

This change calculates the available size better on the amount of actual
bytes required to encode the stream ID and length. It also allows to
omit the length field if enough information is available to fully utilize the
datagram.

This allows to store up to 22 byte more payload data per MTU.

### DIFF
--- a/quinn-proto/src/connection/send_buffer.rs
+++ b/quinn-proto/src/connection/send_buffer.rs
@@ -2,7 +2,7 @@ use std::{collections::VecDeque, ops::Range};
 
 use bytes::{Buf, Bytes};
 
-use crate::range_set::RangeSet;
+use crate::{range_set::RangeSet, VarInt};
 
 /// Buffer of outgoing retransmittable stream data
 #[derive(Default, Debug)]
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/send_buffer.rs
```diff
@@ -76,22 +76,41 @@ impl SendBuffer {
 
     /// Compute the next range to transmit on this stream and update state to account for that
     /// transmission
-    pub fn poll_transmit(&mut self, max_len: usize) -> Range<u64> {
+    pub fn poll_transmit(&mut self, mut max_len: usize) -> (Range<u64>, bool) {
```
Matthias247, line 79: Before I fix/update the unit-tests, there's the question of whether people prefer this approach (assuming  `max_len` only covers the pure data size, and adjust it), or to pass the `max_data` here which doesn't includes the full remaining size after a stream ID would have been encoded (16 bytes more), and let this function do the remaining calculations.

I would prefer the latter.

djc: The latter does seem more attractive, although I find it a little hard to visualize.

Ralith: To clarify, you're proposing refactoring such that the value passed in is the amount of space available to be used for both the length tag and the actual data, but already accounting for the space used by the stream ID? That sounds good to me.

Ralith (commented):
Thanks, this is a nice tidy improvement! Always bugged me a bit how we were leaving space on the table here, but my prior attempts always ended up too complicated and fragile.

quinn-proto/src/connection/send_buffer.rs
Matthias247: Exactly. I will update for this. Will be a tiny change

quinn-proto/src/connection/send_buffer.rs
```diff
@@ -2,7 +2,7 @@ use std::{collections::VecDeque, ops::Range};
 
 use bytes::{Buf, Bytes};
 
-use crate::range_set::RangeSet;
+use crate::{range_set::RangeSet, VarInt};
```
djc, line 5:
```suggestion
use crate::range_set::{RangeSet, VarInt};
```

quinn-proto/src/connection/streams.rs
```diff
@@ -642,7 +644,11 @@ impl Streams {
             if stream.is_reset() {
                 continue;
             }
-            let offsets = stream.pending.poll_transmit(max_data_len);
+
+            // Since the Stream is now known, we can account better how many
+            // bytes are required to encode the StreamId.
```
djc, line 649:
```suggestion
            // Now that we know the `StreamId`, we can better account for how many bytes
            // are required to encode it.
```

djc (commented):
LGTM, just some tiny nits.

quinn-proto/src/connection/streams.rs
Matthias247: updated

djc (approved):

PR #999, opened 2021-01-30
More stream reading improvements

djc:


quinn-proto/src/connection/streams/recv.rs
```diff
@@ -257,69 +196,173 @@ impl Recv {
...
+/// Results of reading from a stream
+#[derive(Debug, PartialEq)]
+#[non_exhaustive]
+pub struct ReadResult {
+    /// The data read from the stream
+    pub chunk: Option<Chunk>,
+    /// The status of the stream after reading
+    pub status: ReadStatus,
+    /// Whether the connection is ready to transmit
+    pub should_transmit: ShouldTransmit,
```
Ralith, line 335: Could we yield this once at the end, rather than per call to `next`? I understand that you want to make it hard to skip, but there's no way to enforce that a struct field gets looked at, and it seems strange to check this repeatedly when we can only act on it after we're done due to the borrow.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -257,69 +196,173 @@ impl Recv {
...
+    ///
+    /// Return the next chunk of data from the stream, if any are available. Will return `None` if
+    /// all available data has been read. `max_length` may be used to restrict the maximum size
+    /// of the returned `Bytes` (but `usize::MAX` is supported, too).
+    ///
+    /// Will keep returning `None` once no further data is available for reading.
+    pub fn next(&mut self, max_length: usize) -> ReadResult {
+        if let Some(status) = &self.fused {
+            return ReadResult::from_status(*status);
+        }
```
Ralith, line 244: Why do we need this?

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -257,69 +196,173 @@ impl Recv {
...
+    pub(crate) fn new(
+        id: StreamId,
+        streams: &'a mut Streams,
+        pending: &'a mut Retransmits,
+        ordered: bool,
+    ) -> Result<Self, ReadableError> {
+        let mut rs = match streams.recv.remove(&id) {
+            Some(rs) => rs,
+            None => return Err(ReadableError::UnknownStream),
+        };
```
Ralith, line 220: Are we missing a check for whether the stream is stopped here?

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -257,69 +196,173 @@ impl Recv {
...
+pub enum ReadStatus {
+    /// No more data is currently available on this stream
     ///
     /// If more data on this stream is received from the peer, an `Event::StreamReadable` will be
     /// generated for this stream, indicating that retrying the read might succeed.
-    #[error("blocked")]
     Blocked,
-    /// The peer abandoned transmitting data on this stream.
+    /// The stream has been stopped
+    Closed,
```
Ralith, line 357: This looks redundant to the `UnknownStream` case in `ReadableError`; a stream can never become stopped mid-read. It's also confusingly named.

quinn-proto/src/tests/mod.rs
```diff
@@ -238,14 +238,14 @@ fn finish_stream_simple() {
...
-        Ok(None)
-    );
+
+    let mut chunks = pair.server_conn_mut(server_ch).read(s, false).unwrap();
+    let result = chunks.next(usize::MAX);
+    assert_matches!(result.chunk, Some(chunk) if chunk.bytes == MSG);
+
+    let result = chunks.next(usize::MAX);
+    assert_eq!(result.status, ReadStatus::Finished);
+    assert_eq!(result.chunk, None);
```
Ralith, line 248: Could we just check the `status` of the prior call, rather than calling again?

quinn/src/streams.rs
```diff
@@ -353,9 +353,11 @@ where
...
     ) -> Poll<Result<(), ReadError>> {
         self.poll_read_generic(cx, |conn, stream| {
-            conn.inner
-                .read(stream, buf.remaining(), true)
-                .map(|val| val.map(|chunk| buf.put_slice(&chunk.bytes)))
+            Ok(conn
+                .inner
+                .read(stream, true)?
+                .next(buf.remaining())
+                .map(|chunk| buf.put_slice(&chunk.bytes)))
```
Ralith, line 360: To prevent loss of specific errors, we must check the status, stash any terminal states locally, and generate a suitable error without calling `inner.read` on the next poll. Similarly below.

Ralith (changes requested):
Great work so far!

quinn-proto/src/connection/assembler.rs
```diff
@@ -132,10 +132,20 @@ impl Assembler {
...
                 offset = duplicate.end;
             }
+        } else if offset < self.bytes_read {
+            if (offset + bytes.len() as u64) < self.bytes_read {
+                return;
+            } else {
+                let diff = self.bytes_read - offset;
+                offset += diff;
+                bytes.advance(diff as usize);
+            }
```
Ralith, line 141: We should get rid of the corresponding read-time code, since this makes it redundant.

Ralith:
Looks like there'll be a bit of a rebase hazard with #1000. Shouldn't be disastrous, though.

geieredgar:
I would be fine with waiting until this PR is merged and then rebasing my changes on top of it.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -57,11 +59,20 @@ impl Recv {
         }
 
         self.end = self.end.max(end);
-        self.assembler.insert(frame.offset, frame.data);
+        if !self.stopped {
+            self.assembler.insert(frame.offset, frame.data);
+        } else {
+            self.assembler.set_bytes_read(end);
```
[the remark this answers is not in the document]
djc: That's right, this is a bit of a behavior change. Would you like me to split it out of this commit?

quinn-proto/src/connection/assembler.rs
djc: What code do you think that is, exactly? In my mind the read-time code is more general, since it covers things that are duplicated but still after the current `bytes_read`, so I don't think we can just get rid of it?

quinn-proto/src/connection/assembler.rs
```diff
@@ -25,14 +25,9 @@ impl Assembler {
         Self::default()
     }
 
-    // Get the the next ordered chunk
-    pub(crate) fn read(
-        &mut self,
-        max_length: usize,
-        ordered: bool,
-    ) -> Result<Option<Chunk>, AssembleError> {
+    pub(crate) fn is_readable(&mut self, ordered: bool) -> Result<(), IllegalOrderedRead> {
```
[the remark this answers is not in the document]
djc: Yup, that seems better.

quinn-proto/src/connection/assembler.rs
```diff
@@ -42,17 +37,21 @@ impl Assembler {
             }
             self.state = State::Unordered { recvd };
         }
+        Ok(())
+    }
 
+    // Get the the next ordered chunk
```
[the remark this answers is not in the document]
djc: Fixed.

quinn-proto/src/connection/assembler.rs
Ralith: I was thinking of the `if (chunk.offset + chunk.bytes.len() as u64) <= self.bytes_read` clause, within the ordered case. On review, though, you're correct: if `bytes_read` is 0 and we receive two frames starting at 0, they aren't guaranteed to be deduplicated until read, so that code should be retained.

djc:
Scoped this down again to remove the hard stuff (last commit) -- all the other comments should be addressed, so we can just get the simple stuff merged. @geieredgar thanks for being accomodating, I hope the rebase isn't too painful.

djc:
(I will take review comments for the last commit in consideration for the new PR as well, of course.)

Ralith (approved):

PR #1000, opened 2021-01-30
Add and use over-allocation metric to decide when to defragment

geieredgar:
Previous discussion: #996

### DIFF
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -13,11 +13,15 @@ use crate::range_set::RangeSet;
 pub(crate) struct Assembler {
     state: State,
     data: BinaryHeap<Buffer>,
-    defragmented: usize,
+    /// Total number of buffered bytes, including duplicates in ordered mode.
+    buffered: usize,
+    /// Estimated number of allocated bytes, will never be less than `buffered`.
+    allocated: usize,
     /// Number of bytes read by the application. When only ordered reads have been used, this is the
     /// length of the contiguous prefix of the stream which has been consumed by the application,
     /// aka the stream offset.
     bytes_read: u64,
+    end: u64,
 }
 
 impl Assembler {
@@ -58,8 +62,9 @@ impl Assembler {
                     return None;
                 } else if (chunk.offset + chunk.bytes.len() as u64) <= self.bytes_read {
                     // Next chunk is useless as the read index is beyond its end
+                    self.buffered -= chunk.bytes.len();
+                    self.allocated -= chunk.allocation_size;
                     PeekMut::pop(chunk);
-                    self.defragmented = self.defragmented.saturating_sub(1);
                     continue;
                 }
 
@@ -68,6 +73,7 @@ impl Assembler {
                 if start > 0 {
                     chunk.bytes.advance(start);
                     chunk.offset += start as u64;
+                    self.buffered -= start;
                 }
             }
 
@@ -75,61 +81,87 @@ impl Assembler {
                 self.bytes_read += max_length as u64;
                 let offset = chunk.offset;
                 chunk.offset += max_length as u64;
+                self.buffered -= max_length;
                 Chunk::new(offset, chunk.bytes.split_to(max_length))
             } else {
                 self.bytes_read += chunk.bytes.len() as u64;
-                self.defragmented = self.defragmented.saturating_sub(1);
+                self.buffered -= chunk.bytes.len();
+                self.allocated -= chunk.allocation_size;
                 let chunk = PeekMut::pop(chunk);
                 Chunk::new(chunk.offset, chunk.bytes)
             });
         }
     }
 
-    // Copy the buffered chunk data to new chunks backed by a single buffer to
-    // make sure we're not unnecessarily holding on to many larger allocations.
-    // Merge contiguous chunks in the process of doing so. Reset the `defragmented`
-    // counter to the new number of chunks left in the heap so that we can decide
-    // when to defragment the queue again if necessary.
+    /// Copy fragmented chunk data to new chunks backed by a single buffer
+    ///
+    /// This makes sure we're not unnecessarily holding on to many larger allocations.
+    /// We merge contiguous chunks in the process of doing so.
     fn defragment(&mut self) {
-        let buffered = self.data.iter().map(|c| c.bytes.len()).sum::<usize>();
-        let mut buffer = BytesMut::with_capacity(buffered);
-        let mut offset = self
-            .data
-            .peek()
-            .as_ref()
-            .expect("defragment is only called when data is buffered")
-            .offset;
-
         let new = BinaryHeap::with_capacity(self.data.len());
         let old = mem::replace(&mut self.data, new);
-        for chunk in old.into_sorted_vec().into_iter().rev() {
-            let end = offset + (buffer.len() as u64);
-            if let Some(overlap) = end.checked_sub(chunk.offset) {
-                if let Some(bytes) = chunk.bytes.get(overlap as usize..) {
-                    buffer.extend_from_slice(bytes);
-                }
-            } else {
-                let bytes = buffer.split().freeze();
-                self.data.push(Buffer { offset, bytes });
-                offset = chunk.offset;
-                buffer.extend_from_slice(&chunk.bytes);
+        let mut buffers = old.into_sorted_vec();
+        self.buffered = 0;
+        let mut fragmented_buffered = 0;
+        let mut offset = 0;
+        for chunk in buffers.iter_mut().rev() {
+            chunk.try_mark_defragment(offset);
+            let size = chunk.bytes.len();
+            offset = chunk.offset + size as u64;
+            self.buffered += size;
+            if !chunk.defragmented {
+                fragmented_buffered += size;
             }
         }
-
-        let bytes = buffer.split().freeze();
-        self.data.push(Buffer { offset, bytes });
-        self.defragmented = self.data.len();
+        self.allocated = self.buffered;
+        let mut buffer = BytesMut::with_capacity(fragmented_buffered);
+        let mut offset = 0;
+        for chunk in buffers.into_iter().rev() {
+            if chunk.defragmented {
+                // bytes might be empty after try_mark_defragment
+                if !chunk.bytes.is_empty() {
+                    self.data.push(chunk);
+                }
+                continue;
+            }
+            // Overlap is resolved by try_mark_defragment
+            if chunk.offset != offset + (buffer.len() as u64) {
+                if !buffer.is_empty() {
+                    self.data
+                        .push(Buffer::new_defragmented(offset, buffer.split().freeze()));
+                }
+                offset = chunk.offset;
+            }
+            buffer.extend_from_slice(&chunk.bytes);
+        }
+        if !buffer.is_empty() {
+            self.data
+                .push(Buffer::new_defragmented(offset, buffer.split().freeze()));
+        }
     }
 
-    pub(crate) fn insert(&mut self, mut offset: u64, mut bytes: Bytes) {
+    // Note: If a packet contains many frames from the same stream, the estimated over-allocation
+    // will be much higher because we are counting the same allocation multiple times.
+    pub(crate) fn insert(&mut self, mut offset: u64, mut bytes: Bytes, allocation_size: usize) {
+        debug_assert!(
+            bytes.len() <= allocation_size,
+            "allocation_size less than bytes.len(): {:?} < {:?}",
+            allocation_size,
+            bytes.len()
+        );
+        self.end = self.end.max(offset + bytes.len() as u64);
         if let State::Unordered { ref mut recvd } = self.state {
             // Discard duplicate data
             for duplicate in recvd.replace(offset..offset + bytes.len() as u64) {
                 if duplicate.start > offset {
-                    self.data.push(Buffer {
+                    let buffer = Buffer::new(
                         offset,
-                        bytes: bytes.split_to((duplicate.start - offset) as usize),
-                    });
+                        bytes.split_to((duplicate.start - offset) as usize),
+                        allocation_size,
+                    );
+                    self.buffered += buffer.bytes.len();
+                    self.allocated += buffer.allocation_size;
+                    self.data.push(buffer);
                     offset = duplicate.start;
                 }
                 bytes.advance((duplicate.end - offset) as usize);
@@ -148,16 +180,25 @@ impl Assembler {
         if bytes.is_empty() {
             return;
         }
-
-        self.data.push(Buffer { offset, bytes });
-        // Why 32: on the one hand, we want to defragment rarely, ideally never
+        let buffer = Buffer::new(offset, bytes, allocation_size);
+        self.buffered += buffer.bytes.len();
+        self.allocated += buffer.allocation_size;
+        self.data.push(buffer);
+        // `self.buffered` also counts duplicate bytes, therefore we use
+        // `self.end - self.bytes_read` as an upper bound of buffered unique
+        // bytes. This will cause a defragmentation if the amount of duplicate
+        // bytes exceedes a proportion of the receive window size.
+        let buffered = self.buffered.min((self.end - self.bytes_read) as usize);
+        let over_allocation = self.allocated - buffered;
+        // Rationale: on the one hand, we want to defragment rarely, ideally never
         // in non-pathological scenarios. However, a pathological or malicious
         // peer could send us one-byte frames, and since we use reference-counted
         // buffers in order to prevent copying, this could result in keeping a lot
-        // of memory allocated. In the worst case scenario of 32 1-byte chunks,
-        // each one from a ~1000-byte datagram, using 32 limits us to having a
-        // maximum pathological over-allocation of about 32k bytes.
-        if self.data.len() - self.defragmented > 32 {
+        // of memory allocated. This limits over-allocation in proportion to the
+        // buffered data. The constants are chosen somewhat arbitrarily and try to
+        // balance between defragmentation overhead and over-allocation.
+        let threshold = 32768.max(buffered * 3 / 2);
+        if over_allocation > threshold {
             self.defragment()
         }
     }
@@ -174,7 +215,8 @@ impl Assembler {
     /// Discard all buffered data
     pub(crate) fn clear(&mut self) {
         self.data.clear();
-        self.defragmented = 0;
+        self.buffered = 0;
+        self.allocated = 0;
     }
 }
 
@@ -197,6 +239,54 @@ impl Chunk {
 struct Buffer {
     offset: u64,
     bytes: Bytes,
+    /// Size of the allocation behind `bytes`, if `defragmented == false`.
+    /// Otherwise this will be set to `bytes.len()` by `try_mark_defragment`.
+    /// Will never be less than `bytes.len()`.
+    allocation_size: usize,
+    defragmented: bool,
+}
+
+impl Buffer {
+    /// Constructs a new fragmented Buffer
+    fn new(offset: u64, bytes: Bytes, allocation_size: usize) -> Self {
+        Self {
+            offset,
+            bytes,
+            allocation_size,
+            defragmented: false,
+        }
+    }
+
+    /// Constructs a new defragmented Buffer
+    fn new_defragmented(offset: u64, bytes: Bytes) -> Self {
+        let allocation_size = bytes.len();
+        Self {
+            offset,
+            bytes,
+            allocation_size,
+            defragmented: true,
+        }
+    }
+
+    /// Discards data before `offset` and flags `self` as defragmented if it has good utilization
+    fn try_mark_defragment(&mut self, offset: u64) {
+        let duplicate = offset.saturating_sub(self.offset) as usize;
+        self.offset = self.offset.max(offset);
+        if duplicate >= self.bytes.len() {
+            self.bytes = Bytes::new();
+            self.defragmented = true;
+            self.allocation_size = 0;
+            return;
+        }
+        self.bytes.advance(duplicate);
+        // Make sure that fragmented buffers with high utilization become defragmented and
+        // defragmented buffers remain defragmented
+        self.defragmented = self.defragmented || self.bytes.len() * 6 / 5 >= self.allocation_size;
+        if self.defragmented {
+            // Make sure that defragmented buffers do not contribute to over-allocation
+            self.allocation_size = self.bytes.len();
+        }
+    }
 }
 
 impl Ord for Buffer {
@@ -257,13 +347,13 @@ mod test {
     fn assemble_ordered() {
         let mut x = Assembler::new();
         assert_matches!(next(&mut x, 32), None);
-        x.insert(0, Bytes::from_static(b"123"));
+        x.insert(0, Bytes::from_static(b"123"), 3);
         assert_matches!(next(&mut x, 1), Some(ref y) if &y[..] == b"1");
         assert_matches!(next(&mut x, 3), Some(ref y) if &y[..] == b"23");
-        x.insert(3, Bytes::from_static(b"456"));
+        x.insert(3, Bytes::from_static(b"456"), 3);
         assert_matches!(next(&mut x, 32), Some(ref y) if &y[..] == b"456");
-        x.insert(6, Bytes::from_static(b"789"));
-        x.insert(9, Bytes::from_static(b"10"));
+        x.insert(6, Bytes::from_static(b"789"), 3);
+        x.insert(9, Bytes::from_static(b"10"), 2);
         assert_matches!(next(&mut x, 32), Some(ref y) if &y[..] == b"789");
         assert_matches!(next(&mut x, 32), Some(ref y) if &y[..] == b"10");
         assert_matches!(next(&mut x, 32), None);
@@ -273,9 +363,9 @@ mod test {
     fn assemble_unordered() {
         let mut x = Assembler::new();
         x.ensure_ordering(false).unwrap();
-        x.insert(3, Bytes::from_static(b"456"));
+        x.insert(3, Bytes::from_static(b"456"), 3);
         assert_matches!(next(&mut x, 32), None);
-        x.insert(0, Bytes::from_static(b"123"));
+        x.insert(0, Bytes::from_static(b"123"), 3);
         assert_matches!(next(&mut x, 32), Some(ref y) if &y[..] == b"123");
         assert_matches!(next(&mut x, 32), Some(ref y) if &y[..] == b"456");
         assert_matches!(next(&mut x, 32), None);
@@ -284,8 +374,8 @@ mod test {
     #[test]
     fn assemble_duplicate() {
         let mut x = Assembler::new();
-        x.insert(0, Bytes::from_static(b"123"));
-        x.insert(0, Bytes::from_static(b"123"));
+        x.insert(0, Bytes::from_static(b"123"), 3);
+        x.insert(0, Bytes::from_static(b"123"), 3);
         assert_matches!(next(&mut x, 32), Some(ref y) if &y[..] == b"123");
         assert_matches!(next(&mut x, 32), None);
     }
@@ -293,8 +383,8 @@ mod test {
     #[test]
     fn assemble_duplicate_compact() {
         let mut x = Assembler::new();
-        x.insert(0, Bytes::from_static(b"123"));
-        x.insert(0, Bytes::from_static(b"123"));
+        x.insert(0, Bytes::from_static(b"123"), 3);
+        x.insert(0, Bytes::from_static(b"123"), 3);
         x.defragment();
         assert_matches!(next(&mut x, 32), Some(ref y) if &y[..] == b"123");
         assert_matches!(next(&mut x, 32), None);
@@ -303,8 +393,8 @@ mod test {
     #[test]
     fn assemble_contained() {
         let mut x = Assembler::new();
-        x.insert(0, Bytes::from_static(b"12345"));
-        x.insert(1, Bytes::from_static(b"234"));
+        x.insert(0, Bytes::from_static(b"12345"), 5);
+        x.insert(1, Bytes::from_static(b"234"), 3);
         assert_matches!(next(&mut x, 32), Some(ref y) if &y[..] == b"12345");
         assert_matches!(next(&mut x, 32), None);
     }
@@ -312,8 +402,8 @@ mod test {
     #[test]
     fn assemble_contained_compact() {
         let mut x = Assembler::new();
-        x.insert(0, Bytes::from_static(b"12345"));
-        x.insert(1, Bytes::from_static(b"234"));
+        x.insert(0, Bytes::from_static(b"12345"), 5);
+        x.insert(1, Bytes::from_static(b"234"), 3);
         x.defragment();
         assert_matches!(next(&mut x, 32), Some(ref y) if &y[..] == b"12345");
         assert_matches!(next(&mut x, 32), None);
@@ -322,8 +412,8 @@ mod test {
     #[test]
     fn assemble_contains() {
         let mut x = Assembler::new();
-        x.insert(1, Bytes::from_static(b"234"));
-        x.insert(0, Bytes::from_static(b"12345"));
+        x.insert(1, Bytes::from_static(b"234"), 3);
+        x.insert(0, Bytes::from_static(b"12345"), 5);
         assert_matches!(next(&mut x, 32), Some(ref y) if &y[..] == b"12345");
         assert_matches!(next(&mut x, 32), None);
     }
@@ -331,8 +421,8 @@ mod test {
     #[test]
     fn assemble_contains_compact() {
         let mut x = Assembler::new();
-        x.insert(1, Bytes::from_static(b"234"));
-        x.insert(0, Bytes::from_static(b"12345"));
+        x.insert(1, Bytes::from_static(b"234"), 3);
+        x.insert(0, Bytes::from_static(b"12345"), 5);
         x.defragment();
         assert_matches!(next(&mut x, 32), Some(ref y) if &y[..] == b"12345");
         assert_matches!(next(&mut x, 32), None);
@@ -341,8 +431,8 @@ mod test {
     #[test]
     fn assemble_overlapping() {
         let mut x = Assembler::new();
-        x.insert(0, Bytes::from_static(b"123"));
-        x.insert(1, Bytes::from_static(b"234"));
+        x.insert(0, Bytes::from_static(b"123"), 3);
+        x.insert(1, Bytes::from_static(b"234"), 3);
         assert_matches!(next(&mut x, 32), Some(ref y) if &y[..] == b"123");
         assert_matches!(next(&mut x, 32), Some(ref y) if &y[..] == b"4");
         assert_matches!(next(&mut x, 32), None);
@@ -351,8 +441,8 @@ mod test {
     #[test]
     fn assemble_overlapping_compact() {
         let mut x = Assembler::new();
-        x.insert(0, Bytes::from_static(b"123"));
-        x.insert(1, Bytes::from_static(b"234"));
+        x.insert(0, Bytes::from_static(b"123"), 4);
+        x.insert(1, Bytes::from_static(b"234"), 4);
         x.defragment();
         assert_matches!(next(&mut x, 32), Some(ref y) if &y[..] == b"1234");
         assert_matches!(next(&mut x, 32), None);
@@ -361,10 +451,10 @@ mod test {
     #[test]
     fn assemble_complex() {
         let mut x = Assembler::new();
-        x.insert(0, Bytes::from_static(b"1"));
-        x.insert(2, Bytes::from_static(b"3"));
-        x.insert(4, Bytes::from_static(b"5"));
-        x.insert(0, Bytes::from_static(b"123456"));
+        x.insert(0, Bytes::from_static(b"1"), 1);
+        x.insert(2, Bytes::from_static(b"3"), 1);
+        x.insert(4, Bytes::from_static(b"5"), 1);
+        x.insert(0, Bytes::from_static(b"123456"), 6);
         assert_matches!(next(&mut x, 32), Some(ref y) if &y[..] == b"123456");
         assert_matches!(next(&mut x, 32), None);
     }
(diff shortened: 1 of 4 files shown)


quinn-proto/src/connection/assembler.rs
```diff
@@ -106,35 +118,55 @@ impl Assembler {
...
                 if duplicate.start > offset {
+                    let over_allocation = (duplicate.end - duplicate.start) as usize;
                     self.data.push(Buffer {
                         offset,
                         bytes: bytes.split_to((duplicate.start - offset) as usize),
+                        over_allocation,
                     });
+                    self.over_allocation += over_allocation;
+                    allocation_size =
+                        allocation_size.saturating_sub((duplicate.end - offset) as usize);
```
Ralith, line 169: Our main concern here is ensuring that there is no possible pattern of insert and read calls that can cause us to be keeping substantially more memory allocated than the amount of useful data we're actually storing, which is the figure governed by flow control. I think this pattern fails that test due to allowing O(n^2) memory use for n useful bytes: consider a flow control limit of `n` bytes, and an ordered sequence of inserts covering ranges `n-1..n, n-2..n, n-3..n`, ...

Treating each chunk as if it had a completely independent allocation does risk counting the same allocation multiple times, but I think risking unnecessary defragmentation passes is preferable to risking quadratic memory overhead.

Ralith (dismissed):

Ralith:
Another approach might be to keep a separate table of allocations, and map chunks to allocations. That should allow us to track memory overhead precisely by comparing buffered data to net allocation size, but is substantially more complex, so I'm not sure it'd be worthwhile.

quinn-proto/src/connection/assembler.rs
```diff
@@ -94,7 +94,19 @@ impl Assembler {
...
     // when to defragment the queue again if necessary.
     fn defragment(&mut self) {
-        let buffered = self.data.iter().map(|c| c.bytes.len()).sum::<usize>();
+        let high_utilization_over_allocation = self
+            .data
+            .iter()
+            .filter(|b| b.high_utilization())
+            .map(|b| b.over_allocation)
+            .sum::<usize>();
+        let include_highly_utilized = high_utilization_over_allocation > DEFRAGMENTATION_THRESHOLD;
```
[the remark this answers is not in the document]
geieredgar: OK, I will remove it again. But do you think it would make sense to retain skipping defragmentation for a `Buffer` with `over_allocation == 0`?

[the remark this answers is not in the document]
Ralith: > OK, I will remove it again. 

Sorry if this is a bit churny; I don't mind waiting for @Matthias247 to weigh in on how smart we should be here, since I think he was advocating for that in the first place.

> But do you think it would make sense to retain skipping defragmentation for a Buffer with over_allocation == 0?

The original logic takes care to avoid defragmenting the same data multiples times; it might be worth preserving that somehow, perhaps even by just retaining the `defragmented` counter. For newly inserted data, there will always be some overhead due to headers.

> The defragmentation decision will then look like this:

That looks great! Just making some numbers up based on nothing, maybe 1.5 and 2^12?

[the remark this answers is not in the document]
geieredgar: > The original logic takes care to avoid defragmenting the same data multiples times; it might be worth preserving that somehow, perhaps even by just retaining the `defragmented` counter. For newly inserted data, there will always be some overhead due to headers.

Are you sure that the original logic does that? To me it looks like the `defragmented` was necessary to know how many new inserts were done since the last defragmentation and on defragmentation all data was copied into a single new allocation.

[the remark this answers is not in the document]
Ralith: Oops, you're quite right, I was confused. In any case, it seems like skipping already-defragmented chunks might be relatively simple, so perhaps worth considering.

quinn-proto/src/connection/assembler.rs
geieredgar: The new version now treats each chunk as a completely independent allocation.

quinn-proto/src/connection/assembler.rs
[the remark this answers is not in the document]
geieredgar: The new version will skip already-defragmented chunks.

quinn-proto/src/connection/assembler.rs
```diff
@@ -106,34 +120,55 @@ impl Assembler {
...
-    }
-
-    pub(crate) fn insert(&mut self, mut offset: u64, mut bytes: Bytes) {
+        self.data.push(Buffer {
+            offset,
+            bytes,
+            allocation_size: 0,
+        });
+        self.allocated = 0;
+        self.buffered = 0
```
Ralith, line 150: Shouldn't `buffered` remain unaffected by defragmentation?

geieredgar: I've put two comments at the declaration of the `allocated` and `buffered` fields, stating they now only apply to fragmented chunks.

Maybe I should rename the fields to `fragmented_allocated` and `fragmented_buffered`?

Ralith: What seems strange to me here is that the defrag threshold gets smaller after defragmentation. I'd expect the tolerated amount of over-allocation to be proportional to the amount of buffered data in an absolute sense. Perhaps that could be achieved by setting `allocated = buffered` after defrag?

geieredgar: OK, I pushed a new version in which `allocated` and `buffered` keep track of both fragmented and defragmented `Buffer`. To prevent defragmenting a `Buffer` twice, I added a `fragmented` field to `Buffer`.

geieredgar:
Rebased my changes on current main.

quinn-proto/src/connection/assembler.rs
```diff
@@ -99,32 +107,60 @@ impl Assembler {
         let new = BinaryHeap::with_capacity(self.data.len());
         let old = mem::replace(&mut self.data, new);
         for chunk in old.into_sorted_vec().into_iter().rev() {
-            let end = offset + (buffer.len() as u64);
-            if let Some(overlap) = end.checked_sub(chunk.offset) {
-                if let Some(bytes) = chunk.bytes.get(overlap as usize..) {
-                    buffer.extend_from_slice(bytes);
+            if chunk.size < chunk.allocation_size {
```
djc, line 110: If you invert this condition we can continue early, which should make this easier to follow (and a smaller diff).

quinn-proto/src/connection/assembler.rs
```diff
@@ -145,15 +181,25 @@ impl Assembler {
...
         // buffers in order to prevent copying, this could result in keeping a lot
-        // of memory allocated. In the worst case scenario of 32 1-byte chunks,
-        // each one from a ~1000-byte datagram, using 32 limits us to having a
-        // maximum pathological over-allocation of about 32k bytes.
-        if self.data.len() - self.defragmented > 32 {
+        // of memory allocated. This limits over-allocation in proportion to the
+        // buffered data. The constants are chosen somewhat arbitrarily and try to
+        // balance between defragmentation overhead and over-allocation.
+        let over_allocation = (self.allocated - self.buffered) as f32;
+        let threshold = ((self.buffered as f32) * 1.5f32).max(4096f32);
```
djc, line 201:
```suggestion
        let threshold = ((self.buffered as f32) * 1.5).max(4096.0);
```

It also seems like an overallocation of only 4096 bytes shouldn't be enough to trigger this (given the performance impact we have seen). Should go back up to something more like 32k?

djc (commented):
This implementation looks solid to me! Some minor issues below:

quinn-proto/src/connection/assembler.rs
geieredgar: OK, done.

quinn-proto/src/connection/assembler.rs
geieredgar: Yeah, makes sense. Fixed.

djc:
@Matthias247 would you have a chance to review the current iteration of these changes?

Matthias247:
Will definitely take a look. However it's a bit of a busy week at the moment, so it might be start of next week.

quinn-proto/src/connection/assembler.rs
```diff
@@ -99,32 +107,60 @@ impl Assembler {
...
+                self.data.push(chunk);
+                continue;
+            }
             let end = offset + (buffer.len() as u64);
             if let Some(overlap) = end.checked_sub(chunk.offset) {
                 if let Some(bytes) = chunk.bytes.get(overlap as usize..) {
                     buffer.extend_from_slice(bytes);
                 }
             } else {
                 let bytes = buffer.split().freeze();
```
Matthias247, line 120: With the current BytesMut issues (freeze being non free, and and resizing after split allocating too much capacity - see tokio-rs/bytes#435 ) this doesn't seem an efficient strategy. I think at this point I would recommend allocating a new BytesMut whenever we detect a new fragmented section begins, and allocating only the capacity for the next fragment(s).

quinn-proto/src/connection/assembler.rs
```diff
@@ -99,32 +107,60 @@ impl Assembler {
         let new = BinaryHeap::with_capacity(self.data.len());
         let old = mem::replace(&mut self.data, new);
         for chunk in old.into_sorted_vec().into_iter().rev() {
+            if chunk.size >= chunk.allocation_size {
```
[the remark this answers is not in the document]
Matthias247: I also think this misses to split and append the last `buffer` segment?

Matthias247 (commented):
The buffered/allocated strategy in the chunks, and taking the defragment decision based on it seem good to me. 

However I think defragment migh have a bug and some wrong assumptions (`size >= allocation_size`) which don't make it as efficient as it could be. 

While efficiency can further be improved in follow-ups, the data ordering bug should be checked and fixed.

quinn-proto/src/connection/mod.rs
```diff
@@ -2427,10 +2430,10 @@ where
                     return Err(err);
                 }
                 Frame::Crypto(frame) => {
-                    self.read_crypto(SpaceId::Data, &frame)?;
+                    self.read_crypto(SpaceId::Data, &frame, payload_len)?;
                 }
                 Frame::Stream(frame) => {
-                    if self.streams.received(frame)?.should_transmit() {
+                    if self.streams.received(frame, payload_len)?.should_transmit() {
```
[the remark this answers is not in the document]
djc: Yup, at least we'll need comments to make this explicit on the assembler side.

quinn-proto/src/connection/assembler.rs
```diff
@@ -99,32 +107,60 @@ impl Assembler {
...
+            offset,
+            bytes,
+            size,
+            allocation_size: size,
+        });
+        self.allocated = self.buffered;
+    }
+
+    pub(crate) fn insert(&mut self, mut offset: u64, mut bytes: Bytes, allocation_size: usize) {
+        assert!(
```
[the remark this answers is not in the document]
geieredgar: Done.

quinn-proto/src/connection/assembler.rs
```diff
@@ -145,15 +181,25 @@ impl Assembler {
...
         // buffers in order to prevent copying, this could result in keeping a lot
-        // of memory allocated. In the worst case scenario of 32 1-byte chunks,
-        // each one from a ~1000-byte datagram, using 32 limits us to having a
-        // maximum pathological over-allocation of about 32k bytes.
-        if self.data.len() - self.defragmented > 32 {
+        // of memory allocated. This limits over-allocation in proportion to the
+        // buffered data. The constants are chosen somewhat arbitrarily and try to
+        // balance between defragmentation overhead and over-allocation.
+        let over_allocation = (self.allocated - self.buffered) as f32;
+        let threshold = ((self.buffered as f32) * 1.5).max(32.0 * 1024.0);
```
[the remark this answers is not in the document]
geieredgar: OK, I changed that. Is it bad if I don't cast and simply use `usize` here?

quinn-proto/src/connection/assembler.rs
```diff
@@ -87,8 +90,13 @@ impl Assembler {
     // counter to the new number of chunks left in the heap so that we can decide
     // when to defragment the queue again if necessary.
     fn defragment(&mut self) {
-        let buffered = self.data.iter().map(|c| c.bytes.len()).sum::<usize>();
-        let mut buffer = BytesMut::with_capacity(buffered);
+        let fragmented_buffered = self
+            .data
+            .iter()
+            .filter(|c| c.size < c.allocation_size)
```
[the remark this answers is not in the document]
geieredgar: This filter is intended to exclude buffers from a previous defragmentation. For those `size` will be equal to `allocation_size`.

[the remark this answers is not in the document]
djc: Maybe add a comment to make that more obvious?

quinn-proto/src/connection/assembler.rs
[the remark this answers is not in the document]
geieredgar: It will be `true` for a `Buffer` from a previous defragmentation, because every `Buffer` created inside the defragmentation method has `allocation_size == size`.

The split and append of the last `buffer` segment happens after the loop and is not affected by these changes.

I like your idea of skipping defragmentation for a buffer with a small proportional overhead, but I also want to hear the opinion of @djc and @Ralith on this.

quinn-proto/src/connection/assembler.rs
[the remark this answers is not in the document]
geieredgar: Done.

quinn-proto/src/connection/mod.rs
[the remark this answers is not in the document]
geieredgar: Added a comment to `Assembler::insert`

quinn-proto/src/connection/assembler.rs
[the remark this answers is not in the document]
djc: A smaller threshold like 1.2 seems fine to me.

quinn-proto/src/connection/assembler.rs
geieredgar: Hm, I could be wrong, but I looked into the code of `BytesMut`, and for me it looks like we should be fine, because we call `split` before `freeze` which calls `promote_to_shared` and changes the kind to `KIND_ARC`. Therefore `freeze` should be free, I think. And also `promote_to_shared` does not seem to call `into_boxed_slice` like `freeze` does.

quinn-proto/src/connection/assembler.rs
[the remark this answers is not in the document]
geieredgar: Done. I moved the logic to decide if a `Buffer` is considered to be fragmented into `Buffer::new` and adjust `allocation_size` accordingly, because a `Buffer` that will never be defragmented should not contribute to the over-allocation. Also I would not be allowed to do `self.allocated = self.bufferred` in `defragment`, if I simply skipped over a buffer with `allocation_size > size`.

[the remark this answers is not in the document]
Matthias247: Thanks. That part looks good to me now

quinn-proto/src/connection/assembler.rs
Matthias247: Ok, I think you are right regarding `freeze`. Thanks for double-checking. 
And the bad allocation behavior is partially mitigated by allocating  everything upfront. However we might be over-allocating, since the initial filter doesn't check for duplicated ranges.

quinn-proto/src/connection/assembler.rs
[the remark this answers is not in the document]
Matthias247: I think it's fine. That means the memory here takes up more than 1/4 of the machines total addressable memory, which is not realistic.

quinn-proto/src/connection/assembler.rs
```diff
@@ -99,33 +107,49 @@ impl Assembler {
         let new = BinaryHeap::with_capacity(self.data.len());
         let old = mem::replace(&mut self.data, new);
         for chunk in old.into_sorted_vec().into_iter().rev() {
+            if !chunk.is_fragmented() {
+                self.data.push(chunk);
```
Matthias247, line 111: Thought I added the comment, but apparently I messed something up: I'm not sure whether this handles the case where there is data from the previous fragment in `buffer`, which needs to be written before this. But then I'm also rather confused by the going through chunks in reversed order, and pushing here in forward order. 

Also let's keep in mind that data in `chunk` might overlap with other data, and we need to account for this (trim stuff, or make sure it at least stays in order).

Are all those paths currently covered by tests?

Matthias247:
One more general realization I had: The old mechanism covered the case where a client sent lots of overlapping ranges, and thereby caused excessive memory usage despite having a limited receive buffer. This change as it is won't do that anymore, since all chunks could still have good utilization.

Nevertheless I think the current approach is good too.

Maybe deduplication should happen on `insert`. But that actually means having to scan all fragments for duplicate ranges, and organizing the buffer more in a form of `Vec<(Buffer, Range)>` than a heap. Also not perfect in case there are lots of fragments.

djc:
> One more general realization I had: The old mechanism covered the case where a client sent lots of overlapping ranges, and thereby caused excessive memory usage despite having a limited receive buffer. This change as it is won't do that anymore, since all chunks could still have good utilization.

Hmm, that definitely seems like an issue. I think we should still cover that scenario.

geieredgar:
I added two new commits. The first fixes #1009 by adding a `deduplicate` function that is called when switching from ordered to unordered mode. The second commit adds a `end` field to `Assembler` and uses `min(self.buffered, self.end - self.bytes_read)` instead of only `self.buffered` for the calculation of `over_allocation`and `threshold`. Therefore, if we have a lot of duplicate bytes buffered, `deduplicate` and `defragment` will be called,

There is still one flaw that might be abused by a malicious peer: After calling `deduplicate`, an already defragmented `Buffer`might be defragmented  again, leading to unnecessary allocations. I will add a new commit to fix that problem with a slightly more complex `defragment`, `deduplicate` and `Buffer`.

quinn-proto/src/connection/assembler.rs
geieredgar: Over-allocation via duplicated ranges is now fixed by calling `deduplicate()` at the beginning of `defragment`.

quinn-proto/src/connection/assembler.rs
geieredgar: Overlap can now never occur, because we call `deduplicate()` at the beginning of `defragment()`. The order of iteration is not really important, because `self.data` is a `BinaryHeap` that will keep the `Buffer` we push in order. However, iterating in that way will both enable the optimization of merging multiple `Buffer` into one and improve the performance of `self.data.push`.

quinn-proto/src/connection/assembler.rs
```diff
@@ -193,6 +281,55 @@ impl Chunk {
 struct Buffer {
     offset: u64,
     bytes: Bytes,
+    size: usize,
```
Matthias247, line 284: Is this ever different from `bytes.len()`? Can't exactly find the case. Could we maybe have a more descriptive name, which hints under which conditions that might be different?

quinn-proto/src/connection/assembler.rs
```diff
@@ -218,6 +355,30 @@ impl PartialEq for Buffer {
...
+    Fragmented(NonZeroUsize),
+    Defragmented,
+}
+
+impl BufferState {
+    fn new(size: usize, allocation_size: usize, possibly_duplicate: bool) -> Self {
+        if possibly_duplicate || Self::significant_over_allocation(size, allocation_size) {
+            match NonZeroUsize::new(allocation_size) {
+                Some(value) => Self::Fragmented(value),
+                None => Self::Defragmented,
```
Matthias247, line 369: Can that ever be true? Would it be worthwhile to have an `unreachable!` here, and to make sure adding an empty buffer is a no-op?

quinn-proto/src/connection/assembler.rs
```diff
@@ -144,16 +226,21 @@ impl Assembler {
         if bytes.is_empty() {
```
Matthias247, line 226: We should maybe just move this all the way up

quinn-proto/src/connection/assembler.rs
```diff
@@ -98,34 +108,106 @@ impl Assembler {
...
-        let bytes = buffer.split().freeze();
-        self.data.push(Buffer { offset, bytes });
-        self.defragmented = self.data.len();
+    fn deduplicate(&mut self) -> RangeSet {
+        debug_assert!(self.state.is_ordered());
+        let mut ranges = RangeSet::new();
+        ranges.insert(0..self.bytes_read);
+        let new = BinaryHeap::with_capacity(self.data.len());
+        let old = mem::replace(&mut self.data, new).into_vec();
+        for buffer in old.iter() {
```
Matthias247, line 141: `defragment` reverse iterates. This forward iterates. Why the difference?

quinn-proto/src/connection/assembler.rs
Matthias247: Oh, one thing that I got wrong was to assume that `self.data` is a vector or vecdeque, and that therefore you need to do `.push_front()` instead of `.push()`. However this doesn't apply, since this is a `BinaryHeap`.

quinn-proto/src/connection/assembler.rs
```diff
@@ -98,34 +108,106 @@ impl Assembler {
...
+            } else if !chunk.should_defragment() {
+                self.buffered += chunk.bytes.len();
+                self.data
+                    .push(Buffer::new_defragmented(chunk.offset, chunk.bytes));
+                continue;
+            }
+            if chunk.offset != offset + (buffer.len() as u64) {
+                self.data
+                    .push(Buffer::new_defragmented(offset, buffer.split().freeze()));
                 offset = chunk.offset;
```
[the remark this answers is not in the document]
geieredgar: Correct.

quinn-proto/src/connection/assembler.rs
geieredgar: There is currently only one case in which it will differ: inside `read`, if `max_length < bytes.len()`. We didn't want to update `size` there because it could lead to defragmenting an already defragmented buffer again, but this is no longer the case. 

I am now thinking about replacing the `size: usize` and `state: BufferState` simply with `allocation_size: usize` and `fragmented: bool`. I think this would make things a lot clearer.

quinn-proto/src/connection/assembler.rs
```diff
@@ -98,34 +108,106 @@ impl Assembler {
...
-                }
-            } else {
-                let bytes = buffer.split().freeze();
-                self.data.push(Buffer { offset, bytes });
+            if chunk.is_defragmented() {
+                self.buffered += chunk.size;
+                self.data.push(chunk);
+                continue;
+            } else if !chunk.should_defragment() {
+                self.buffered += chunk.bytes.len();
```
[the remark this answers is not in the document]
geieredgar: Currently, it is not fully identical (because `chunk.size` could differ from `chunk.bytes.len()`), but I will get rid of the `size` field and turn this into one `if` condition.

quinn-proto/src/connection/assembler.rs
geieredgar: Yeah, thought about that too.

quinn-proto/src/connection/assembler.rs
geieredgar: In `defragment`, we want to iterate with ascending `offset` so that we can coalesce multiple `Buffer` into one. This is not required for `deduplicate`. Also, iterating it in the order of the underlying `Vec` makes sure that `self.data.push` is optimal (if we do not find any duplicates).

quinn-proto/src/connection/assembler.rs
```diff
@@ -87,8 +89,16 @@ impl Assembler {
     // counter to the new number of chunks left in the heap so that we can decide
     // when to defragment the queue again if necessary.
     fn defragment(&mut self) {
-        let buffered = self.data.iter().map(|c| c.bytes.len()).sum::<usize>();
-        let mut buffer = BytesMut::with_capacity(buffered);
+        if self.state.is_ordered() {
+            self.deduplicate();
```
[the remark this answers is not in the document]
geieredgar: I think I could do that, but what would you do about #1009? Calling `defragment` when entering unordered mode? Or is what I described there not really an issue?

djc:
@geieredgar please take the opportunity to basically add comments for all of Matthias' questions. Would be good to get the documentation in there since this is obviously a tricky topic and it would be nice to make the explanations easier to access if we ever need to iterate more on this. Plus, it will help make my review easier.

quinn-proto/src/connection/assembler.rs
```diff
@@ -87,8 +89,16 @@ impl Assembler {
     // counter to the new number of chunks left in the heap so that we can decide
```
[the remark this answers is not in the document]
geieredgar: Fixed.

quinn-proto/src/connection/assembler.rs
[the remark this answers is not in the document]
geieredgar: Ok, got rid of `deduplicate` again. Deduplication is now also handled by `defragment`.

quinn-proto/src/connection/assembler.rs
```diff
@@ -87,8 +89,16 @@ impl Assembler {
...
     // when to defragment the queue again if necessary.
     fn defragment(&mut self) {
-        let buffered = self.data.iter().map(|c| c.bytes.len()).sum::<usize>();
-        let mut buffer = BytesMut::with_capacity(buffered);
+        if self.state.is_ordered() {
+            self.deduplicate();
+        }
+        let fragmented_buffered = self
```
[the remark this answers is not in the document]
geieredgar: Yes, I think it is best to do this in a follow-up.

quinn-proto/src/connection/assembler.rs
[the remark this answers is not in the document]
geieredgar: Added a comment as clarification.

quinn-proto/src/connection/assembler.rs
geieredgar: The `size` field does no longer exist.

quinn-proto/src/connection/assembler.rs
geieredgar: This is also now obsolete, because I removed `deduplicate`.

quinn-proto/src/connection/assembler.rs
geieredgar: This got also removed in my latest changes, sorry for that.

quinn-proto/src/connection/assembler.rs
geieredgar: I will do that in a follow-up, otherwise the diff will be harder to review.

quinn-proto/src/connection/assembler.rs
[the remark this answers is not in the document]
geieredgar: Fixed.

geieredgar:
@djc Some of Matthias' questions became obsolete after my latest refactoring, for the others I added comments.

quinn-proto/src/connection/assembler.rs
```diff
@@ -64,68 +67,48 @@ impl Assembler {
...
+    // Note: If a packet contains many frames from the same stream, the estimated over-allocation
+    // will be much higher because we are counting the same allocation multiple times.
+    pub(crate) fn insert(&mut self, mut offset: u64, mut bytes: Bytes, allocation_size: usize) {
+        debug_assert!(
+            bytes.len() <= allocation_size,
+            "allocation_size less than bytes.len(): {:?} < {:?}",
+            allocation_size,
+            bytes.len()
+        );
+        self.end = max(self.end, offset + bytes.len() as u64);
```
Matthias247, line 99: Nix: the style in most other files is `self.end = self.end.max(offset + bytes.len() as u64);`
This avoids the need for the import

quinn-proto/src/connection/assembler.rs
Matthias247: Is it? I really meant **all** the way, as one of the first things in `insert`.  Hmm, or does it need to be here to catch the case where we have nothing after removing duplicate data?

quinn-proto/src/connection/assembler.rs
```diff
@@ -193,6 +228,48 @@ impl Chunk {
...
+        Self {
+            offset,
+            bytes,
+            allocation_size,
+            defragmented: true,
+        }
+    }
+
+    /// Discards data before `offset` and flags `self` as defragmented if it has good utilization
+    fn prepare_defragment(&mut self, offset: u64) {
```
Matthias247, line 258: Nit: I would potentially call this `try_mark_defragmented`, since it changes the state already without anything after it. `prepare` sounds like there would be a follow-up call to get it into the state.

quinn-proto/src/connection/assembler.rs
```diff
@@ -144,20 +127,71 @@ impl Assembler {
         if bytes.is_empty() {
             return;
         }
-
-        self.data.push(Buffer { offset, bytes });
-        // Why 32: on the one hand, we want to defragment rarely, ideally never
+        let buffer = Buffer::new(offset, bytes, allocation_size);
```
Matthias247, line 130: What is the reason for not calling `buffer.prepare_defragment` here, or even executing the logic as part of `Buffer::new()`?

quinn-proto/src/connection/assembler.rs
```diff
@@ -144,20 +127,71 @@ impl Assembler {
...
+        let over_allocation = self.allocated - buffered;
+        let threshold = max(buffered * 3 / 2, 32 * 1024);
+        if over_allocation > threshold {
             self.defragment()
         }
     }
 
+    // Copy the buffered chunk data to new chunks backed by a single buffer to
+    // make sure we're not unnecessarily holding on to many larger allocations.
+    // Merge contiguous chunks in the process of doing so.
```
Matthias247, line 151:
```suggestion
    /// Copy fragmented chunk data to new chunks backed by a single buffer
    ///
    /// This makes sure we're not unnecessarily holding on to many larger allocations.
    /// We merge contiguous chunks in the process of doing so.
```

Matthias247: Just to move it to a doc comment, and keep the normal doc comment style with headline and details.

quinn-proto/src/connection/assembler.rs
Matthias247: Thinking more about it, is it to capture the situation where buffers are fully filling the allocation and are therefore not oversized, but where the overlap between buffers requires the defragmentation?

Matthias247 (commented):
Thanks for the hard work of doing all the changes! This looks a lot simpler now. I just pointed out some style nits and added some questions, but I'm fine with merging as is.

quinn-proto/src/connection/assembler.rs
geieredgar: Oops, for some reason I thought you were talking about moving `insert` as a whole up, my bad. But yeah, it is there because `bytes` could become empty after deduplication.

quinn-proto/src/connection/assembler.rs
geieredgar: Yes, if we would mark a `Buffer` immediately as defragmented, we cannot get rid of it in `defragment`, if e.g. it contains only one byte after discarding duplicate data.

quinn-proto/src/connection/assembler.rs
geieredgar: Done.

quinn-proto/src/connection/assembler.rs
geieredgar: Fixed.

quinn-proto/src/connection/assembler.rs
geieredgar: Done.

quinn-proto/src/connection/assembler.rs
```diff
@@ -1,5 +1,5 @@
 use std::{
-    cmp::Ordering,
+    cmp::{max, min, Ordering},
```
Matthias247, line 2: Are those still required?

Matthias247 (approved):
Looks great to me now!
Thanks for all the changes

quinn-proto/src/connection/assembler.rs
geieredgar: Yes, in the threshold calculation.

geieredgar: I replaced the other uses of `min` and `max` with their method equivalent and got rid of the imports.

quinn-proto/src/connection/assembler.rs
```diff
@@ -197,6 +233,48 @@ impl Chunk {
...
+        self.offset = self.offset.max(offset);
+        if duplicate >= self.bytes.len() {
+            self.bytes = Bytes::new();
+            self.defragmented = true;
+        } else {
+            self.bytes.advance(duplicate);
+        }
+        self.defragmented = self.defragmented || self.bytes.len() * 6 / 5 >= self.allocation_size;
+        if self.defragmented {
+            // Make sure that defragmented buffers do not contribute to over-allocation
```
djc, line 274: This logic seems more complicated than necessary. If we've established that `self.defragmented = true` because the chunk is now empty, we should just reset the allocation size and return immediately.

geieredgar: Fixed.

quinn-proto/src/connection/assembler.rs
```diff
@@ -148,20 +131,72 @@ impl Assembler {
...
+        // balance between defragmentation overhead and over-allocation.
+        let buffered = self.buffered.min((self.end - self.bytes_read) as usize);
+        let over_allocation = self.allocated - buffered;
+        let threshold = 32768.max(buffered * 3 / 2);
+        if over_allocation > threshold {
             self.defragment()
         }
     }
 
+    /// Copy fragmented chunk data to new chunks backed by a single buffer
```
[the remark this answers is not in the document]
geieredgar: Fixed.

quinn-proto/src/connection/assembler.rs
```diff
@@ -148,20 +131,72 @@ impl Assembler {
...
         // peer could send us one-byte frames, and since we use reference-counted
         // buffers in order to prevent copying, this could result in keeping a lot
-        // of memory allocated. In the worst case scenario of 32 1-byte chunks,
-        // each one from a ~1000-byte datagram, using 32 limits us to having a
-        // maximum pathological over-allocation of about 32k bytes.
-        if self.data.len() - self.defragmented > 32 {
+        // of memory allocated. This limits over-allocation in proportion to the
+        // buffered data. The constants are chosen somewhat arbitrarily and try to
+        // balance between defragmentation overhead and over-allocation.
+        let buffered = self.buffered.min((self.end - self.bytes_read) as usize);
```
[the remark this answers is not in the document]
geieredgar: Yes, and that is the reason for the upper bound. Duplicates will increase `self.buffered` and therefore we would (without this limit) not trigger defragmentation unless these duplicates have low utilization.

quinn-proto/src/connection/assembler.rs
```diff
@@ -148,20 +131,72 @@ impl Assembler {
...
+        let over_allocation = self.allocated - buffered;
```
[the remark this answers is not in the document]
geieredgar: This is a invariant that must be maintained. I added comments to the `buffered` and `allocated` field stating this invariant.

quinn-proto/src/connection/assembler.rs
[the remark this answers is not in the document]
djc: Ah, that makes sense. Can you add a comment about this?

quinn-proto/src/connection/assembler.rs
```diff
@@ -197,6 +239,50 @@ impl Chunk {
...
+        let duplicate = offset.saturating_sub(self.offset) as usize;
+        self.offset = self.offset.max(offset);
+        if duplicate >= self.bytes.len() {
+            self.bytes = Bytes::new();
+            self.defragmented = true;
+            self.allocation_size = 0;
+            return;
+        } else {
+            self.bytes.advance(duplicate);
+        }
```
djc, line 279: No `else` needed anymore, nor do we still need the `self.defragmented ||` below.

quinn-proto/src/connection/assembler.rs
```diff
@@ -68,68 +73,95 @@ impl Assembler {
...
+            if !chunk.defragmented {
+                fragmented_buffered += size;
+            }
+        }
+        self.allocated = self.buffered;
+        let mut buffer = BytesMut::with_capacity(fragmented_buffered);
+        let mut offset = 0;
+        for chunk in buffers.into_iter().rev() {
+            if chunk.defragmented {
+                // bytes might be empty after prepare_defragment
```
djc, line 121: `prepare_defragment` no longer exists.

geieredgar: Fixed.

quinn-proto/src/connection/assembler.rs
geieredgar: Got rid of the `else` branch. However, the `self.defragmented ||` is required to make sure that defragmented buffers remain defragmented. I added a comment stating that. If a defragmented buffer could become fragmented again, we might end up copying from a `Bytes` allocated by a previous `defragment()`, which does not get rid of the previous allocation because it might be shared between multiple buffers.

quinn-proto/src/connection/assembler.rs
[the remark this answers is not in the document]
geieredgar: Done.

quinn-proto/src/connection/assembler.rs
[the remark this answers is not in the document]
geieredgar: The only way the invariant might be invalidated is when we pass an `allocation_size` that is smaller than `bytes.len()` to `insert` in release builds, because we currently only use a `debug_assert!` to protect against that. Maybe we should change this back to an `assert!`?

quinn-proto/src/connection/assembler.rs
djc: Ah, you're right.

quinn-proto/src/connection/assembler.rs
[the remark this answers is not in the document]
djc: No, that should be fine.

djc (approved):

djc:
Alright, thanks for seeing this through!

PR #1001, opened 2021-01-30
Implement stream prioritization (fix #165)

### LINKED ISSUE #165: Implement stream prioritization
> A QUIC implementation SHOULD provide ways in which an application can indicate the relative priority of streams

Stream priority could be indicated with a ~`u32`~`i32`, associated with every send-capable stream. A stream's priority could be used as the key for a priority queue taking the place of the current `Retransmits::stream` field, with ~`Connection::next_packet`~ `Connection::poll_transmit` always selecting the pending stream frame with the largest priority to send next.

geieredgar:
I tried to implement stream prioritization in a way that should not have an huge impact for users who do not need this feature.

Instead of a single queue, it uses a vector of (priority, queue) pairs, sorted by descending priority, so that we can iterate over it in `write_stream_frames`. Pushing a new pending stream will take O(log n) time to find the correct queue, where n is the number of already used priority levels.  

Current drawback: The vector of (priority, queue) pairs is never shrunk, which is a problem if a lot of different priority levels are used. Therefore it might make sense to change the priority type to `i8` to limit the maximum size of the vector to 256.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1653,6 +1653,37 @@ where
         Ok(())
     }
 
+    /// Set the priority of a stream
+    ///
+    /// # Panics
+    /// - when applied to a receive stream
+    pub fn set_priority(
+        &mut self,
+        stream_id: StreamId,
+        priority: i32,
+    ) -> Result<(), UnknownStream> {
+        assert!(
+            stream_id.dir() == Dir::Bi || stream_id.initiator() == self.side,
+            "only streams supporting outgoing data may change priority"
+        );
+
+        self.streams.set_priority(stream_id, priority)?;
+        Ok(())
+    }
+
+    /// Get the priority of a stream
+    ///
+    /// # Panics
+    /// - when applied to a receive stream
+    pub fn priority(&mut self, stream_id: StreamId) -> Result<i32, UnknownStream> {
+        assert!(
+            stream_id.dir() == Dir::Bi || stream_id.initiator() == self.side,
+            "only streams supporting outgoing data have a priority"
+        );
+
+        Ok(self.streams.priority(stream_id)?)
+    }
+
     /// Handle the already-decrypted first packet from the client
     ///
     /// Decrypting the first packet in the `Endpoint` allows stateless packet handling to be more
--- a/quinn-proto/src/connection/streams.rs
+++ b/quinn-proto/src/connection/streams.rs
@@ -1,5 +1,6 @@
 use std::{
-    collections::{hash_map, HashMap, VecDeque},
+    cell::RefCell,
+    collections::{binary_heap::PeekMut, hash_map, BinaryHeap, HashMap, VecDeque},
     convert::TryFrom,
     mem,
 };
@@ -50,7 +51,7 @@ pub struct Streams {
     /// permitted to open but which have not yet been opened.
     send_streams: usize,
     /// Streams with outgoing data queued
-    pending: VecDeque<StreamId>,
+    pending: BinaryHeap<PendingLevel>,
 
     events: VecDeque<StreamEvent>,
     /// Streams blocked on connection-level flow control or stream window space
@@ -105,7 +106,7 @@ impl Streams {
             opened: [false, false],
             next_reported_remote: [0, 0],
             send_streams: 0,
-            pending: VecDeque::new(),
+            pending: BinaryHeap::new(),
             events: VecDeque::new(),
             connection_blocked: Vec::new(),
             max_data: 0,
@@ -272,7 +273,7 @@ impl Streams {
         self.unacked_data += len as u64;
         trace!(stream = %id, "wrote {} bytes", len);
         if !was_pending {
-            self.pending.push_back(id);
+            push_pending(&mut self.pending, id, stream.priority);
         }
         Ok(len)
     }
@@ -393,7 +394,7 @@ impl Streams {
         let was_pending = stream.is_pending();
         stream.finish()?;
         if !was_pending {
-            self.pending.push_back(id);
+            push_pending(&mut self.pending, id, stream.priority);
         }
         Ok(())
     }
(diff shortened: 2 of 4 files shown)


djc:
Switching to an 8-bit priority seems sensible to me. I'm wondering whether we should trade memory for performance by maybe having a `[VecDeque; 256]` plus maybe a `BinaryHeap` of priorities that currently have pending data.

quinn-proto/src/connection/mod.rs
```diff
@@ -1653,6 +1653,37 @@ where
         Ok(())
     }
 
+    /// Set the priority of a stream
+    ///
+    /// # Panics
+    /// - when applied to a receive stream
```
Ralith, line 1659: Would it be reasonable to return `UnknownStream` in this case (and below)? We generally prefer not to panic even under API misuse, since the caller might be able to isolate and/or report failures more gracefully even when they're caused by caller bugs. Maybe tweak the comment to "... priority of an outgoing stream" for clarity.

Ralith (commented):
Can you elaborate on why we should use a sorted `Vec` rather than a `BinaryHeap` containing only priorities that have data queued?

I don't think we specifically need billions of priorities, but I'd like to be convinced that strictly limiting the number has a major otherwise-unattainable benefit.

quinn-proto/src/connection/mod.rs
geieredgar: Would that not be inconsistent with `reset`? That function also panics when applied to a receive stream and that's also where I copied the assertion from.

geieredgar:
@Ralith The reason I used a sorted `Vec` instead of a `BinaryHeap` is that iterating over an `BinaryHeap` in a sorted manner is O(log(n) * n) instead of O(n) and also cannot be done easily without mutating it. At least I don't know a way to do that.

quinn-proto/src/connection/mod.rs
Ralith: Good question. I suppose we haven't thought too hard about this in the past since the `quinn` layer enforces sanity at the type level. Given the precedent I'm happy to revisit this if/when we change reset's behavior.

quinn-proto/src/connection/streams.rs
```diff
@@ -628,7 +653,11 @@ impl Streams {
                 stream.fin_pending = false;
             }
             if stream.is_pending() {
-                self.pending.push_back(id);
+                if priority == stream.priority {
+                    self.pending.push_back(queue_index, id);
+                } else {
+                    queue_index = queue_index.min(self.pending.push(id, stream.priority));
+                }
```
Ralith, line 660: If a stream switches from high priority to low priority, would this cause us to skip writing data from streams at intermediate priorities?

geieredgar: No, because we take the min of `queue_index` and the result of `self.pending.push`. If the new priority is lower, the new queue index is guaranteed to be larger.

Ralith:
It's not clear to me that we should expect to iterate over such a heap very much at all, provided that priorities which have no pending data do not appear in it. I'm imagining an encoding loop that looks something like:

```rust
while space_available {
    let level = self.heap.peek_mut();
    let stream = level.pop_front();
    write(stream);
    if stream.pending() { level.push_back(stream); }
    if level.is_empty() { self.heap.pop(); }
}
```

The only non-constant-time operation here is `pop`. For it to be invoked often, we would need to frequently have priority levels with significantly less than one MTU of data, and the absolute cost will only be significant with a large number of priority levels with pending data. I expect typical use will have few concurrently-active priority levels, and that there will not be large numbers of priorities having small amounts of data queued. Conversely, it might be useful to have lots of space for different priorities so they can be spent freely, allowing well-defined relationships between many types of communication even though only a few are likely to be concurrently active.

This is ultimately speculative, and a solution that behaves well in all cases will need to be configurable and/or heuristically adaptive. I'm interested in hearing other perspectives, but I do very much like how a heap approach has stateless performance characteristics.

geieredgar:
Hm, you convinced me. As you said, the number of concurrently-active priority levels with pending data will probably be pretty small and therefore we can afford iterating over the whole `BinaryHeap` to find the right queue when pushing. I will push a new version using `BinaryHeap`.

geieredgar:
I pushed a version using `BinaryHeap`. Unfortunately, I had to resort to using `RefCell`, because we need to mutate the queue while iterating over the `BinaryHeap` in `push_pending`.

Ralith (changes requested):
I think this looks good overall. It does need a unit test in streams.rs verifying that prioritization works as intended.

quinn-proto/src/connection/streams.rs
```diff
@@ -922,6 +954,46 @@ impl Streams {
...
+    let mut queue = VecDeque::new();
+    queue.push_back(id);
+    pending.push(PendingLevel {
+        queue: RefCell::new(queue),
+        priority,
+    });
+}
+
+struct PendingLevel {
+    queue: RefCell<VecDeque<StreamId>>,
```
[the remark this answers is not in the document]
geieredgar: Comment added.

quinn-proto/src/connection/streams.rs
```diff
@@ -922,6 +954,46 @@ impl Streams {
     }
 }
 
+fn push_pending(pending: &mut BinaryHeap<PendingLevel>, id: StreamId, priority: i32) {
+    for level in pending.iter() {
+        if priority == level.priority {
+            level.queue.borrow_mut().push_back(id);
+            return;
+        }
+    }
```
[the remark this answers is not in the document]
geieredgar: I would stick to the simpler `BinaryHeap` until we have shown that this does not scale well enough.

geieredgar:
Added a unit test and a comment about the `RefCell`.

Ralith (approved):
Thanks!

quinn-proto/src/connection/streams.rs
```diff
@@ -423,6 +424,25 @@ impl Streams {
...
+        let stream = match self.send.get_mut(&id) {
+            Some(ss) => ss,
+            None => return Err(UnknownStream { _private: () }),
+        };
+
+        stream.priority = priority;
+        Ok(())
+    }
+
+    pub fn get_priority(&mut self, id: StreamId) -> Result<i32, UnknownStream> {
```
djc, line 437: Should we adhere to the API guidelines here and call these just `priority()` rather than using the `get_` prefix?

Ralith: Good catch; `priority` is definitely more idiomatic.

djc: Otherwise this looks good to me, feel free to merge when this is fixed.

geieredgar: Fixed.

Ralith (approved):

Matthias247:
FWIW, the current proposals for HTTP/3 priority use a scheme with just 8 priorities: https://datatracker.ietf.org/doc/draft-ietf-httpbis-priority/?include_text=1

I think those would be totally good enough to support.

Ralith:
If there's a significant benefit I remain unopposed to narrowing things down a bit, but I'm skeptical that `[VecDeque<StreamId>; N]` would ever be substantially faster than, as a straw example, a sorted `Vec<(i32, VecDeque<StreamId>)>`.

PR #1002, opened 2021-02-01
Make fairness on stream transmissions configurable

geieredgar:
This allows making a send stream unfair, so it will block other streams of the same priority.

This enables users of quinn to implement a prioritization scheme like [this](https://datatracker.ietf.org/doc/draft-ietf-httpbis-priority/) for their application protocol.

### DIFF
--- a/quinn/src/streams.rs
+++ b/quinn/src/streams.rs
@@ -11,7 +11,7 @@ use futures::{
     io::{AsyncRead, AsyncWrite},
     ready, FutureExt,
 };
-use proto::{Chunk, ConnectionError, FinishError, StreamId};
+use proto::{Chunk, ConnectionError, FinishError, Priority, StreamId};
 use thiserror::Error;
 use tokio::io::ReadBuf;
 
@@ -152,19 +152,18 @@ where
 
     /// Set the priority of the send stream
     ///
-    /// Every send stream has an initial priority of 0. Locally buffered data from streams with
-    /// higher priority will be transmitted before data from streams with lower priority. Changing
-    /// the priority of a stream with pending data may only take effect after that data has been
-    /// transmitted. Using many different priority levels per connection may have a negative
-    /// impact on performance.
-    pub fn set_priority(&self, priority: i32) -> Result<(), UnknownStream> {
+    /// Locally buffered data from streams with higher priority will be transmitted before data
+    /// from streams with lower priority. Changing the priority of a stream with pending data may
+    /// only take effect after that data has been transmitted. Using many different priorities per
+    /// connection may have a negative impact on performance.
+    pub fn set_priority(&self, priority: Priority) -> Result<(), UnknownStream> {
         let mut conn = self.conn.lock().unwrap();
         conn.inner.set_priority(self.stream, priority)?;
         Ok(())
     }
 
     /// Get the priority of the send stream
-    pub fn priority(&self) -> Result<i32, UnknownStream> {
+    pub fn priority(&self) -> Result<Priority, UnknownStream> {
         let mut conn = self.conn.lock().unwrap();
         Ok(conn.inner.priority(self.stream)?)
     }
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -52,7 +52,9 @@ pub use stats::ConnectionStats;
 
 mod streams;
 pub use streams::Streams;
-pub use streams::{FinishError, ReadError, ShouldTransmit, StreamEvent, UnknownStream, WriteError};
+pub use streams::{
+    FinishError, Priority, ReadError, ShouldTransmit, StreamEvent, UnknownStream, WriteError,
+};
 
 mod timer;
 use timer::{Timer, TimerTable};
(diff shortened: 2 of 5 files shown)


quinn/src/streams.rs
```diff
@@ -169,6 +169,22 @@ where
         Ok(conn.inner.priority(self.stream)?)
     }
 
+    /// Set the fairness of the send stream
+    ///
+    /// Every send stream is initialy fair. Unfair streams will block other streams with the same
```
djc, line 174: `initialy` -> `initially`

djc (commented):
This looks good to me!

quinn/src/streams.rs
geieredgar: Fixed.

Matthias247:
Is there really any use-case for it? If you want unfairness, you can already set different priorities with the last change. This seems like a duplication of that functionality. And a footgun to get really bad and undeterministic performance.

geieredgar:
Quote from the [Extensible Prioritization Scheme for HTTP](https://datatracker.ietf.org/doc/draft-ietf-httpbis-priority/?include_text=1):
> 3.2.  Incremental
>
> The incremental parameter ("i") takes an sf-boolean as the value that
> indicates if an HTTP response can be processed incrementally, i.e.
> provide some meaningful output as chunks of the response arrive.
>
> The default value of the incremental parameter is false ("0").
>
>  A server might distribute the bandwidth of a connection between
>  incremental responses that share the same urgency, hoping that
>  providing those responses in parallel would be more helpful to the
>  client than delivering the responses one by one.
>
>  If a client makes concurrent requests with the incremental parameter
>  set to false, there is no benefit serving responses in parallel
>  because the client is not going to process those responses
>  incrementally.  Serving non-incremental responses one by one, in the
>  order in which those requests were generated is considered to be the
>  best strategy.

If I understand it correctly, that prioritization scheme would per default prefer blocking streams because the incremental flag is set to false. 

With this feature I would simply map HTTPs (urgency, incremental) to quinns (priority, fair) via:
```rust
let priority = if incremental {
    -urgency * 2 + 1
} else {
    -urgency * 2
};
let fair = incremental;
```

Also I don't think one can implement
> Serving non-incremental responses one by one, in the order in which those requests were generated is considered to be the best strategy.

without this feature easily.

Ralith:
My knee-jerk reaction was also that this is redundant, but I do see how ensuring response ordering without compromising performance is otherwise impossible; you could wait for each stream to finish before writing the next, but that adds an idle round-trip, and even if your streams are larger than the flow control window they'd still get blurred together around the edges. Given that the change in logic is *very* simple, I find I'm not opposed.

To reduce interface bloat, maybe we should have `#[non_exhaustive] struct Priority { pub level: i32, pub fair: bool }` and fold it into the existing priority accessors?

Matthias247:
I don't think this helps as it is to make sure requests after served one after another if the client prefers this. I think it rather leads to a mess of priorization that is hard to understand (as visible in some of the old qvis diagrams before priorization was initially fixed). If the flag is not set to `fair`, it will just lead the last stream continue to write. However as soon as that can't write for some time (e.g. for flow control reasons, missing upstream data, etc) it will jump over to write another stream and hang on to that. This might just give the worst of all worlds, with e.g. most of the  data from one stream being written first, then all data from another stream, and then the remaining data from stream 1.

I think if one wants to make sure a particular stream finishes first, either use a separate priority level, or make sure on application level that only data on one stream will be sent (which is totally possible).

geieredgar:
I changed the implementation to use a BinaryHeap for unfair stream priorities. With that, as soon as new data for a stream with a lower `StreamId` can be transmitted, we switch back to that.

Ralith:
The less-trivial (though still admittedly not hugely complex) implementation is making me hesitant. We only buffer data for at most the flow control window, so the worst-case duration of round-robin behavior if the application is sending data strictly in order on equal-priority streams still has a modest bound. When streams are significantly larger than the flow control window, this should have very little impact, and of course streams that fit into a single packet are unaffected as well.

Do we have specific reason to believe that going to extra effort to serialize transmission in the marginal case when multiple streams are buffered in a single flow control window is justified?

geieredgar:
As for now, I think the extra complexity is unjustified. Therefore I will close this pull request.

Ralith:
Sounds good. Thanks for the discussion!

Matthias247:
> Do we have specific reason to believe that going to extra effort to serialize transmission in the marginal case when multiple streams are buffered in a single flow control window is justified?

It is required in order not to fall back into the old LIFO behavior that was described in https://qlog.edm.uhasselt.be/epiq/files/QUICImplementationDiversity_Marx_final_11jun2020.pdf 

But I think the question whether this new complexity is justified is very valid. As long as there isn't a direct users which can't use a different approach (e..g higher priority) - it probably isn't.

If anyone requests is, this could be revived. Maybe it can be moved behind a feature flag, to avoid the extra indirection for people not too concerned with the use-case.

Ralith:
> It is required in order not to fall back into the old LIFO behavior

With the current behavior we're not LIFO under any circumstances, though--strictly round-robin for buffered stuff, and the rest is up to the application.

Matthias247:
Yes. But the additional change here tried to allow a full FIFO behavior in the case the same priorities had been chosen. This wasn't possible with the more simple change, and required another level of priority lists.

PR #1003, opened 2021-02-01
Use iterator-like API for reading

djc:


quinn-proto/src/connection/streams/recv.rs
```diff
@@ -257,71 +198,179 @@ impl Recv {
...
+        let status = match &chunk {
+            Some(chunk) => {
+                self.read += chunk.bytes.len() as u64;
+                ReadStatus::Readable
+            }
+            None => match rs.state {
+                RecvState::Recv { size } => {
+                    if size == Some(rs.end) && rs.assembler.bytes_read() == rs.end {
+                        self.streams.stream_freed(self.id, StreamHalf::Recv);
+                        self.fused = Some(ReadStatus::Finished);
```
Matthias247, line 262: Maybe assign `fused` after the match block, for the `status != ReadStatus::Readable` case

Matthias247: A more descriptive name than `fused` could also be nice. I know it resembles the iterator stuff. But for someone not familiar with those APIs it just sounds like a weird word.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -257,71 +198,179 @@ impl Recv {
...
+            Some(rs) => rs,
+            None => return ShouldTransmit(false),
+        };
+
+        let (_, max_stream_data) = rs.max_stream_data(self.streams.stream_receive_window);
+        let max_data = self.streams.add_read_credits(self.read);
+        let max_dirty = self.streams.take_max_streams_dirty(self.id.dir());
+        self.pending
+            .post_read(self.id, max_data, max_stream_data, max_dirty);
+        max_data | max_stream_data | ShouldTransmit(max_dirty)
```
Matthias247, line 304: Weren't those 2 separate code path before? One one whether max data or max stream data frames could be converted, and one for max streams. I guess if it's only about signalling that something needs to be transmitted then unifying them is great. But if something makes a distinction what to transmit based on the return value it might not work.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -257,71 +198,179 @@ impl Recv {
...
+                    self.streams.stream_freed(self.id, StreamHalf::Recv);
+                    self.fused = Some(ReadStatus::Reset(error_code));
+                    ReadStatus::Reset(error_code)
+                }
+                RecvState::Closed => unreachable!(),
+            },
+        };
+
+        let mut should_transmit = ShouldTransmit(false);
+        if status != ReadStatus::Readable {
```
Matthias247, line 281: Is this condition right? if we have a large receive buffer which is full, and the user only reads 25% of it, we would need to send an update. But the state would stay at readable.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -257,71 +198,179 @@ impl Recv {
...
+        max_data | max_stream_data | ShouldTransmit(max_dirty)
+    }
 }
 
-impl BytesRead for ReadChunks {
-    fn bytes_read(&self) -> u64 {
-        self.read as u64
+impl<'a> Drop for Chunks<'a> {
+    fn drop(&mut self) {
+        let _ = self.finalize();
```
Matthias247, line 310: who executes the transmit if we are dropping the result?

quinn/src/streams.rs
```diff
@@ -427,8 +442,25 @@ where
...
+                let result = chunks.next(usize::MAX);
+                if let Some(chunk) = result.chunk {
+                    bufs[read] = chunk.bytes;
+                    read += 1;
+                    if read == bufs.len() {
+                        return Ok((read, read > 0, result.status, chunks.finalize()));
+                    }
+                }
+                if result.status != ReadStatus::Readable {
+                    return Ok((read, read > 0, result.status, result.should_transmit));
```
Matthias247, line 460: Are intermediate `.status` and `.should_transmit` results not relevant? I think this would require that the final values always accumulate/contain the previous values.

quinn-proto/src/connection/streams/recv.rs
djc: The `pending.post_read()` call should ensure that both are sent; returning the `ShouldTransmit` to the caller is just about helping the caller decide whether calling `poll_transmit()` is necessary.

quinn-proto/src/connection/streams/recv.rs
djc: I don't think it would be efficient to suggest to the caller that they `poll_transmit()` if there is yet more data to read, right? They should ideally read out all of the available data and then inspect `ShouldTransmit` to see if polling for packets to be sent is necessary.

quinn-proto/src/connection/streams/recv.rs
djc: It has to be initiated by someone. To be fair, I don't think this is any worse than what we have today. And this would only happen if the caller did not read out `next()` until it contained `status != Readable`.

quinn/src/streams.rs
djc: If you look at `Chunks::next()`, `status` will always be `Readable` if `chunk` is not `None`, and `should_transmit` will always be `ShouldTransmit(false)`, by definition. Maybe the definition is wrong, but the handling here seems fine.

quinn-proto/src/connection/streams/recv.rs
Matthias247: There is no need or guarantee that a caller will read all available data. E.g. a HTTP/3 server might just enough chunks until it determines it got the header of a request. As soon as it has the header it might execute some business logic - e.g. connecting to a upstream service, preparing a response, etc. That might take hundreds of ms, and for that time it won't bother reading more buffered data. However this shouldn't block the peer from at least sending some more data.

It's the same as there is no expectation that a user of a TCP socket immediately drains it completely, instead of letting data being buffered in there.

quinn/src/streams.rs
Ralith: > status will always be Readable if chunk is not None

Off the cuff, that sounds like maybe we should actually have a `Result<Chunk, Status>`.

djc: So maybe the return type should be `(Result<Option<Chunk>, ReadError>, ShouldTransmit)`?

Ralith: Right, similar to the current interface, except with the new iterator layer.

quinn-proto/src/connection/streams/recv.rs
djc: I've replaced this with a `State` enum.

quinn-proto/src/connection/streams/recv.rs
djc: I completely rewrote this part to be much more like the current logic.

quinn-proto/src/connection/streams/recv.rs
djc: Good idea, done.

quinn/src/streams.rs
djc: Reverted to pretty much the current return types, I think these changes didn't improve things.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -257,71 +198,179 @@ impl Recv {
...
-    pub read: usize,
+    /// Read the next chunk from the stream
+    ///
+    /// Return the next chunk of data from the stream, if any are available. `max_length` may be
+    /// used to restrict the maximum size of the returned `Bytes` (but `usize::MAX` is supported,
+    /// too).
+    pub fn next(&mut self, max_length: usize) -> ReadResult {
+        if let Some(status) = &self.fused {
+            return ReadResult::from_status(*status);
+        }
```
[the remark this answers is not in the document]
djc: That basically means we'd panic when callers call `next()` again after an error, right? I think I'd prefer fusing (but see the current take on this, which is somewhat different).

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -257,71 +198,179 @@ impl Recv {
...
+                        ReadStatus::Blocked
+                    }
+                }
+                RecvState::ResetRecvd { error_code, .. } => {
+                    rs.state = RecvState::Closed;
+                    self.streams.stream_freed(self.id, StreamHalf::Recv);
+                    self.fused = Some(ReadStatus::Reset(error_code));
+                    ReadStatus::Reset(error_code)
+                }
+                RecvState::Closed => unreachable!(),
```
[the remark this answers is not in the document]
djc: That's no longer the case. It still feels like the state management could be improved here (are `Recv::stopped` and `Recv::state` overlapping?), but I'd rather punt that to a future improvement.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -257,71 +198,179 @@ impl Recv {
...
+
+        let mut should_transmit = ShouldTransmit(false);
+        if status != ReadStatus::Readable {
+            should_transmit = self.finalize();
+        }
+
+        ReadResult {
+            chunk,
+            status,
+            should_transmit,
```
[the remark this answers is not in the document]
djc: I think this is fixed.

Matthias247:
One alternative to this kind of change is implement a `BytesSink` in the spirit of `BytesSource` in #1013, and rebase all read APIs based on that. 

I'm however not sure if those kinds of APIs should be purely internals of quinn-proto, or get full public APIs

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -257,6 +247,89 @@ impl Recv {
...
+    }
+
+    pub(crate) fn next(&mut self, max_length: usize) -> Result<Option<Chunk>, ReadError> {
+        let mut rs = self.recv.take().unwrap();
+        if let Some(chunk) = rs.assembler.read(max_length, self.ordered) {
+            let (_, max_stream_data) = rs.max_stream_data(self.streams.stream_receive_window);
+            let max_data = self.streams.add_read_credits(chunk.bytes.len() as u64);
+            self.pending
+                .post_read(self.id, max_data, max_stream_data, false);
+            self.streams.recv.insert(self.id, rs);
```
Ralith: Within the scope of this specific commit, this doesn't make sense; we should put `rs` back into `self.recv` so futures calls to `next` will succeed.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -312,56 +274,75 @@ impl<'a> Chunks<'a> {
...
+                        Self::done(rs, self.read, self.id, self.streams, self.pending);
+                    let old = mem::replace(
+                        &mut self.state,
+                        ChunksState::Error(ReadError::Blocked, should_transmit),
+                    );
+
+                    let rs = match old {
+                        ChunksState::Readable(rs) => rs,
+                        _ => unreachable!(),
+                    };
```
Ralith, line 290: Could we move out of `self.state` at the start of the function without running afoul of borrowck, and thereby avoid the `unreachable!()` here?

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -381,27 +362,6 @@ pub enum ReadError {
     /// The stream has not been opened or was already stopped, finished, or reset
     #[error("unknown stream")]
     UnknownStream,
-    /// Attempted an ordered read following an unordered read
-    ///
-    /// Performing an unordered read allows discontinuities to arise in the receive buffer of a
-    /// stream which cannot be recovered, making further ordered reads impossible.
-    #[error("ordered read after unordered read")]
-    IllegalOrderedRead,
```
Ralith, line 316: Let's remove this case in the previous commit, where the `ReadableError` case that takes its place is introduced.

quinn/src/streams.rs
```diff
@@ -371,10 +371,23 @@ where
...
+                match chunks.next(buf.remaining()) {
+                    Ok(Some(chunk)) => {
+                        buf.put_slice(&chunk.bytes);
+                        read = true;
+                    }
+                    Ok(None) | Err(_) if read => return Ok(Some(())),
+                    Ok(None) => return Ok(None),
+                    Err(e) => return Err(e),
+                }
+            }
```
Ralith, line 390: This pattern (and below) can lose errors. Consider what happens when both data and an error are waiting. If the application reads all the data, the error will not be reported, but the stream's state (error included) may then be disposed. The next read call may then see `UnknownStream` rather than the correct error.

Rather than fusing errors in `Chunks`, which is short lived, we should do so here at the high level, where they're guaranteed to be preserved as long as we like.

quinn/src/streams.rs
```diff
@@ -428,7 +439,27 @@ where
         cx: &mut Context,
         bufs: &mut [Bytes],
     ) -> Poll<Result<Option<usize>, ReadError>> {
-        self.poll_read_generic(cx, |conn, stream| conn.inner.read_chunks(stream, bufs))
+        self.poll_read_generic(cx, true, |chunks| {
+            let mut read = 0;
+            loop {
+                if read >= bufs.len() {
+                    return match read {
+                        0 => Ok(None),
```
Ralith, line 447: Would this cause us to return `Ok(None)` for a non-finished stream when `bufs.len() == 0`?

Ralith (changes requested):

Ralith:
> One alternative to this kind of change is implement a BytesSink in the spirit of BytesSource in #1013, and rebase all read APIs based on that.

I tend to favor pull-based over push-based APIs, because they give the caller more freedom over e.g. control flow.

quinn-proto/src/connection/streams/recv.rs
djc: There no longer is a previous commit? I think you may have reviewed some out-of-date commits as well as the current ones...

quinn-proto/src/connection/streams/recv.rs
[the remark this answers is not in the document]
djc: Not all error cases will result in `UnknownStream` though (in particular, `Blocked`).

quinn/src/streams.rs
djc: Yeah, I wasn't sure how to handle the case where `bufs.len() == 0`. Maybe that needs a new `ReadError` variant?

quinn-proto/src/connection/streams/recv.rs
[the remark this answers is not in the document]
Ralith: Ah, right. I'm still a bit concerned that this can mask caller bugs though, especially since we're usually going to be the caller.

quinn-proto/src/connection/streams/recv.rs
Ralith: Github is currently displaying two commits for this PR. "The previous commit" refers to 7b3953ae58f6bedefa2908ec6d7a327c036c59bf.

quinn/src/streams.rs
Ralith: What's wrong with `Ok(Some(0))`?

quinn-proto/src/connection/streams/recv.rs
Ralith: On review, I think this comment is confusing because it's being displayed on the "all commits" view; the intended reading is on the second commit's diff, where `ReadError::IllegalOrderedRead` was removed.

quinn/src/streams.rs
djc: Yeah, I guess that is correct.

djc:
I'm not sure if it still makes sense to have the first commit separately and/or if the second commit can be split into more parts (I suspect not). This seems to pass the tests for me locally.

quinn/src/streams.rs
djc: This is done.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -257,6 +247,89 @@ impl Recv {
...
+
+        match rs.state {
+            RecvState::ResetRecvd { error_code, .. } => {
+                rs.state = RecvState::Closed;
+                self.streams.stream_freed(self.id, StreamHalf::Recv);
+                self.pending
+                    .post_read(self.id, ShouldTransmit(false), ShouldTransmit(false), true);
+                Err(ReadError::Reset(error_code))
+            }
+            RecvState::Closed => Err(ReadError::UnknownStream),
```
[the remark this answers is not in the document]
djc: Marked as `unreachable!()`.

quinn-proto/src/connection/streams/recv.rs
djc: Yup, that looks to make more sense.

djc: It does end up being a little ugly, having to restore the state in 3 additional places. See what you think?

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -258,40 +199,152 @@ impl Recv {
...
+            state: ChunksState::Readable(recv),
+            read: 0,
+        })
+    }
+
+    /// Next
+    ///
+    /// Should call finalize() when done calling this.
+    pub fn next(&mut self, max_length: usize) -> Result<Option<Chunk>, ReadError> {
+        let mut rs = match mem::replace(&mut self.state, ChunksState::Finalized) {
```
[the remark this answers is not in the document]
djc: Compiler says no:

```rust
error[E0507]: cannot move out of `self.state.0` which is behind a mutable reference
   --> quinn-proto/src/connection/streams/recv.rs:239:28
    |
239 |         let mut rs = match self.state {
    |                            ^^^^^^^^^^ help: consider borrowing here: `&self.state`
240 |             ChunksState::Readable(rs) => rs,
    |                                   -- data moved here
241 |             ChunksState::Error(e, st) => {
    |                                - ...and here
    |
    = note: move occurs because these variables have types that don't implement the `Copy` trait
```

Unless you know of any workarounds for this?

quinn/src/streams.rs
```diff
@@ -482,43 +547,60 @@ where
...
         cx: &mut Context,
+        ordered: bool,
         mut read_fn: T,
     ) -> Poll<Result<Option<U>, ReadError>>
     where
-        T: FnMut(
-            &mut crate::connection::ConnectionInner<S>,
-            StreamId,
-        ) -> Result<Option<U>, proto::ReadError>,
+        T: FnMut(&mut Chunks) -> (Option<U>, Result<Option<()>, proto::ReadError>),
```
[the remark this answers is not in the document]
djc: This is the resulting change:

```diff
diff --git a/quinn/src/streams.rs b/quinn/src/streams.rs
index 69bd8ffc..c37314bd 100644
--- a/quinn/src/streams.rs
+++ b/quinn/src/streams.rs
@@ -386,7 +386,7 @@ where
                             true => Some(()),
                             false => None,
                         },
-                        Ok(Some(())),
+                        Ok(false),
                     );
                 }
 
@@ -401,7 +401,7 @@ where
                                 true => Some(()),
                                 false => None,
                             },
-                            res.map(|inner| inner.map(|_| ())),
+                            res.map(|inner| inner.is_none()),
                         )
                     }
                 }
@@ -437,8 +437,8 @@ where
         ordered: bool,
     ) -> Poll<Result<Option<Chunk>, ReadError>> {
         self.poll_read_generic(cx, ordered, |chunks| match chunks.next(max_length) {
-            Ok(Some(chunk)) => (Some(chunk), Ok(Some(()))),
-            res => (None, res.map(|inner| inner.map(|_| ()))),
+            Ok(Some(chunk)) => (Some(chunk), Ok(false)),
+            res => (None, res.map(|inner| inner.is_none())),
         })
     }
 
@@ -473,7 +473,7 @@ where
                             0 => None,
                             n => Some(n),
                         },
-                        Ok(Some(())),
+                        Ok(false),
                     );
                 }
 
@@ -488,7 +488,7 @@ where
                                 0 => None,
                                 n => Some(n),
                             },
-                            res.map(|inner| inner.map(|_| ())),
+                            res.map(|inner| inner.is_none()),
                         )
                     }
                 }
@@ -551,7 +559,7 @@ where
         mut read_fn: T,
     ) -> Poll<Result<Option<U>, ReadError>>
     where
-        T: FnMut(&mut Chunks) -> (Option<U>, Result<Option<()>, proto::ReadError>),
+        T: FnMut(&mut Chunks) -> (Option<U>, Result<bool, proto::ReadError>),
     {
         use proto::ReadError::*;
         if self.all_data_read {
@@ -576,8 +584,8 @@ where
         };
 
         match result {
-            Ok(Some(())) => Poll::Ready(Ok(done)),
-            Ok(None) => {
+            Ok(false) => Poll::Ready(Ok(done)),
+            Ok(true) => {
                 self.all_data_read = true;
                 Poll::Ready(Ok(done))
             }
```

I don't think it's a clear win, and I find reasoning about the semantics actually harder because the second element moves away from the established `Option` semantics for stream ending. (Quickly, does `true` mean you can ask the stream for more data or does it mean the stream was done?)

(All the matches are actually about the first element, but that has to actually pack a value to return to its caller.)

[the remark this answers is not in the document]
djc: I guess we could use a struct with named fields for the return value here, but the caller just ends up destructing it immediately (and naming its fields), so it doesn't help that much...

quinn-proto/src/connection/streams/recv.rs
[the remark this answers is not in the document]
djc: Ah, you're right! That's much nicer -- I feel stupid for not having noticed it.

quinn-proto/src/connection/streams/recv.rs
djc: (Continued in https://github.com/quinn-rs/quinn/pull/1003#discussion_r581647348.)

quinn/src/streams.rs
```diff
@@ -428,7 +460,40 @@ where
...
+
+        self.poll_read_generic(cx, true, |chunks| {
+            let mut read = 0;
+            loop {
+                if read >= bufs.len() {
+                    return (
+                        match read {
+                            0 => None,
+                            n => Some(n),
+                        },
```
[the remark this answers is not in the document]
djc: Fair point, fixed and added a comment about this.

quinn/src/streams.rs
```diff
@@ -482,43 +547,60 @@ where
     fn poll_read_generic<T, U>(
```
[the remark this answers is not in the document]
djc: Added a bunch of comments to try and make things more clear. Better?

quinn-proto/src/connection/streams/recv.rs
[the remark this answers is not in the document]
Ralith: Nope, just wasn't sure where rustc's line in the sand was. Thanks for having a go at it.

quinn/src/streams.rs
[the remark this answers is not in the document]
Ralith: Much, thanks!

quinn/src/streams.rs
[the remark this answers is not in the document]
Ralith: One of the things that gave me a lot of pause in the first pass were the nested `map`s, especially as the inner closure will never be invoked since we've separately matched the `Ok(Some(_))` case. We could even replace the inner `map`/`is_none` with a constant, perhaps justified by a comment.

Neither `Option<()>` nor `bool` are particularly readable in and of themselves, even if the former has a correspondence to the ultimate return value. Maybe a special-purpose enum would be a better way to go?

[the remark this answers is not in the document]
djc: Hmm, I like that. What do you think of this approach?

quinn/src/streams.rs
```diff
@@ -479,46 +539,88 @@ where
...
         cx: &mut Context,
+        ordered: bool,
         mut read_fn: T,
     ) -> Poll<Result<Option<U>, ReadError>>
     where
-        T: FnMut(
-            &mut crate::connection::ConnectionInner<S>,
-            StreamId,
-        ) -> Result<Option<U>, proto::ReadError>,
+        T: FnMut(&mut Chunks) -> (Option<U>, Result<ReadStatus, proto::ReadError>),
```
Ralith, line 271: Is `(None, Ok(Readable))` a valid state? Should this be `(Option<U>, Result<(), proto::ReadError>)`?

quinn/src/streams.rs
```diff
@@ -479,46 +539,88 @@ where
...
+                    conn.blocked_readers.insert(self.stream, cx.waker().clone());
+                    Poll::Pending
+                }
+            },
+            Err(Reset(error_code)) => match read {
+                None => {
+                    self.all_data_read = true;
+                    Poll::Ready(Err(ReadError::Reset(error_code)))
+                }
+                done => {
```
Ralith, line 319: `done@Some(_)` for clarity

quinn/src/streams.rs
Ralith: Valid states:
- (Some, Ok(Readable))
- (Some, Ok(Finished))
- (Some, Err)
- (None, Ok(Finished))
- (None, Err)

No, that wouldn't work because `(Some, Ok(Finished))` is valid. How about this?
```
enum ReadResult<U> {
  Nonempty(U, Result<ReadStatus, proto::ReadError>),
  Empty(Result<(), proto::ReadError>),
}
```

(terrible naming, please improve)

Ralith: Better yet, I think this is algebraically equivalent, but much saner:
```rust
enum ReadResult<U> {
    Readable(U),
    Finished(Option<U>),
    Failed(Option<U>, proto::ReadError),
}
```

djc: The latter does not seem like an improvement:

```diff
diff --git a/quinn/src/streams.rs b/quinn/src/streams.rs
index 9e7fcfef..317b973a 100644
--- a/quinn/src/streams.rs
+++ b/quinn/src/streams.rs
@@ -381,13 +381,7 @@ where
             let mut read = false;
             loop {
                 if buf.remaining() == 0 {
-                    return (
-                        match read {
-                            true => Some(()),
-                            false => None,
-                        },
-                        Ok(ReadStatus::Readable),
-                    );
+                    return ReadStatus::Readable(());
                 }
 
                 match chunks.next(buf.remaining()) {
@@ -395,13 +389,19 @@ where
                         buf.put_slice(&chunk.bytes);
                         read = true;
                     }
-                    res => {
-                        return (
+                    Ok(None) => {
+                        return ReadStatus::Finished(match read {
+                            true => Some(()),
+                            false => None,
+                        })
+                    }
+                    Err(e) => {
+                        return ReadStatus::Failed(
                             match read {
                                 true => Some(()),
                                 false => None,
                             },
-                            res.map(|inner| inner.into()),
+                            e,
                         )
                     }
                 }
@@ -554,7 +554,7 @@ where
         mut read_fn: T,
     ) -> Poll<Result<Option<U>, ReadError>>
     where
-        T: FnMut(&mut Chunks) -> (Option<U>, Result<ReadStatus, proto::ReadError>),
+        T: FnMut(&mut Chunks) -> ReadStatus<U>,
     {
         use proto::ReadError::*;
         if self.all_data_read {
@@ -568,26 +568,25 @@ where
 
         // If we stored an error during a previous call, return it now. This can happen if a
         // `read_fn` both wants to return data and also returns an error in its final stream status.
-        let (read, result) = match self.reset.take() {
-            Some(code) => (None, Err(Reset(code))),
+        let status = match self.reset.take() {
+            Some(code) => ReadStatus::Failed(None, Reset(code)),
             None => {
                 let mut chunks = conn.inner.read(self.stream, ordered)?;
-                let (read, result) = read_fn(&mut chunks);
+                let status = read_fn(&mut chunks);
                 if chunks.finalize().should_transmit() {
                     conn.wake();
                 }
-                (read, result)
+                status
             }
         };
 
-        match result {
-            Ok(status) => {
-                if let ReadStatus::Finished = status {
-                    self.all_data_read = true;
-                }
+        match status {
+            ReadStatus::Readable(read) => Poll::Ready(Ok(Some(read))),
+            ReadStatus::Finished(read) => {
+                self.all_data_read = true;
                 Poll::Ready(Ok(read))
             }
-            Err(Blocked) => match read {
+            ReadStatus::Failed(read, Blocked) => match read {
                 Some(val) => Poll::Ready(Ok(Some(val))),
                 None => {
                     if let Some(ref x) = conn.error {
@@ -597,7 +596,7 @@ where
                     Poll::Pending
                 }
             },
-            Err(Reset(error_code)) => match read {
+            ReadStatus::Failed(read, Reset(error_code)) => match read {
                 None => {
                     self.all_data_read = true;
                     Poll::Ready(Err(ReadError::Reset(error_code)))
@@ -611,18 +610,10 @@ where
     }
 }
 
-enum ReadStatus {
-    Readable,
-    Finished,
-}
-
-impl<T> From<Option<T>> for ReadStatus {
-    fn from(val: Option<T>) -> Self {
-        match val {
-            Some(_) => ReadStatus::Readable,
-            None => ReadStatus::Finished,
-        }
-    }
+enum ReadStatus<T> {
+    Readable(T),
+    Finished(Option<T>),
+    Failed(Option<T>, proto::ReadError),
 }
 
 /// Future produced by [`RecvStream::read_to_end()`].
```

djc: Okay, I figured out how to make it work. Should be good.

quinn/src/streams.rs
```diff
@@ -479,46 +518,88 @@ where
...
     }
 
+    /// Handle common logic related to reading out of a receive stream
+    ///
+    /// This takes an `FnMut` closure that takes care of the actual reading process, matching
+    /// the detailed read semantics for the calling function with a particular return type.
+    /// The closure can read from the passed `&mut Chunks` and has to return a result in two parts:
+    ///
+    /// * Amount of data read, in the type expected by the calling read method
+    /// * Status after the final read, with the `Ok` value denoting whether the stream has finished
```
Ralith, line 242: Description of the return value needs to be updated

Ralith (dismissed):

quinn/src/streams.rs
djc: Done.

Ralith (approved):

PR #1007, opened 2021-02-05
Allow to use the vectored ordered read API in benchmark

Matthias247:
This extends the benchmark to make the consumer either utilize the
ordered or unordered read API.
The ordered read API is now set to the default, since it might be
applicable to more applications.

### DIFF
--- a/bench/src/bulk.rs
+++ b/bench/src/bulk.rs
@@ -5,6 +5,7 @@ use std::{
 };
 
 use anyhow::{Context, Result};
+use bytes::Bytes;
 use futures::StreamExt;
 use hdrhistogram::Histogram;
 use structopt::StructOpt;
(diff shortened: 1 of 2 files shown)


Ralith (approved):
Yeah, definitely better to be representative. No sense excluding such an important stage from the benchmark.

bench/src/bulk.rs
```diff
@@ -266,4 +284,7 @@ struct Opt {
     /// Show connection stats the at the end of the benchmark
     #[structopt(long = "stats")]
     stats: bool,
+    /// Whether to use the unordered read API
+    #[structopt(long = "unordered")]
+    read_unordered: bool,
```
djc, line 289: I think having the semantics flipped here (`read_ordered`, with structopt setting `true` as the explicit default) is a little easier to reason about. I would also probably just name it ordered.

djc (commented):
LGTM, modulo nit on the CLI option name.

bench/src/bulk.rs
Matthias247: that doesn't work for boolean flags. `--something` actives them. There is no `--something false`.

djc: Ah, sorry about that.

PR #1010, opened 2021-02-07
Use inline storage for the common case of 1 STREAM frame per packet

Matthias247:
When packets are transmitted, the most common case them is to contain at
most a single STREAM frame. For this common case the current system does
still 2 allocations: One for the `SentPacket` structure which tracks the whole
packet, and one for the stream metadata. This double allocation places
some additional pressure on the allocator. This change changes
`SentPacket::stream_frames`  to make use of vector with small element
optimization, in order to be able to handle the common case of sending
1 stream frame without any further allocation.

Benchmark results
===

Overall the differences don't seem huge. However they become bigger when other
optimizations (like bigger MTUs and size reduction for `SentPacket` are also
enabled).

Without GSO
--

**Before:**

> Sent 4294967296 bytes on 1 streams in 15.37s (266.49 MiB/s)

**After:**

> Sent 4294967296 bytes on 1 streams in 15.33s (267.17 MiB/s)

With GSO
--

**Before:**

> Sent 4294967296 bytes on 1 streams in 11.79s (347.48 MiB/s)

**After:**

> Sent 4294967296 bytes on 1 streams in 11.70s (350.11 MiB/s)

### DIFF
--- a/quinn-proto/src/frame.rs
+++ b/quinn-proto/src/frame.rs
@@ -4,6 +4,7 @@ use std::{
 };
 
 use bytes::{Buf, BufMut, Bytes};
+use tinyvec::TinyVec;
 
 use crate::{
     coding::{self, BufExt, BufMutExt, UnexpectedEnd},
@@ -429,6 +430,19 @@ pub struct StreamMeta {
     pub fin: bool,
 }
 
+// This manual implementation exists because `Default` is not implemented for
+// `Range<u64>` on Rust 1.45. If the MSRV is bumped in the future, this can be replaced
+// with `derive(Default)`.
+impl Default for StreamMeta {
+    fn default() -> Self {
+        StreamMeta {
+            id: StreamId(0),
+            offsets: 0..0,
+            fin: false,
+        }
+    }
+}
+
 impl StreamMeta {
     pub fn encode<W: BufMut>(&self, length: bool, out: &mut W) {
         let mut ty = *STREAM_TYS.start();
@@ -451,6 +465,9 @@ impl StreamMeta {
         }
     }
 }
+
+/// A vector of [`StreamMeta`] with optimization for the single element case
+pub type StreamMetaVec = TinyVec<[StreamMeta; 1]>;
 
 #[derive(Debug, Clone)]
 pub struct Crypto {
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -34,6 +34,7 @@ rustls = { version = "0.19", features = ["quic"], optional = true }
 rustls-native-certs = { version = "0.5", optional = true }
 slab = "0.4"
 thiserror = "1.0.21"
+tinyvec = { version = "1.1", features = ["alloc"] }
 tracing = "0.1.10"
 webpki = { version = "0.21", optional = true }
 
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -8,6 +8,7 @@ use std::{
 };
 
 use bytes::{Bytes, BytesMut};
+use frame::StreamMetaVec;
 use rand::{rngs::StdRng, Rng, SeedableRng};
 use thiserror::Error;
 use tracing::{debug, error, trace, trace_span, warn};
@@ -3465,7 +3466,7 @@ struct ZeroRttCrypto<S: crypto::Session> {
 struct SentFrames {
     retransmits: Retransmits,
     acks: RangeSet,
-    stream_frames: Vec<frame::StreamMeta>,
+    stream_frames: StreamMetaVec,
     padding: bool,
 }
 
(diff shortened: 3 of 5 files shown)


djc:
This mostly looks good, with one caveat: have you looked at any alternatives to the smallvec crate? In the past 3 years, it has had 5 security vulnerabilities reported against it in the advisory DB. Maybe try if something like tinyvec can get the same result?

Ralith:
I'm on board with this change, and I suspect we can make a lot of others like it to start adding up to a larger impact. Tinyvec definitely could be used here and this, if any, is an application where taking care with `unsafe` is justified, though, so we should give that a try.

Matthias247:
First time I heard about the `SmallVec` issues, despite having it used in some other projects. Interesting that those are indeed legitimate bugs, instead of more debatable UB discussions where things work fine in practice.

I'm not terribly concerned by the past issues, since past issues always mean people took a closer look and fixed them. The crate is fairly widely used, and so should have a large level of maturity. Besides that, the set of APIs that are used here are also fairly minimal and unlikely to break.

I looked a bit at the difference with `TinyVec`, and it seems like the main drawbacks of those is that it default initializes all elements, and only supports `Default` types. I think for this particular use-case it won't matter a lot. For other potential use-cases (e.g. optimizing `RangeSet`) the `SmallVec` benefits could however be more important, and sticking with a single dependency might be preferable.

Overall I'm fine with picking any of those for the moment.

djc:
> I'm not terribly concerned by the past issues, since past issues always mean people took a closer look and fixed them. The crate is fairly widely used, and so should have a large level of maturity.

Usually I also think this way, but given this track record:

- Jan 8, 2021
- Jul 19, 2019
- Jun 6, 2019
- Sep 25, 2018
- Jul 19, 2018

I see little proof that people started to look closer, nor that wide usage or a high maturity level have helped here.

djc:
(If we truly need a minimal subset of API, maybe we should NIH it with a little enum?)

Ralith:
> For other potential use-cases (e.g. optimizing RangeSet) the SmallVec benefits could however be more important

Can you elaborate here? `Range<u64>` is easy enough to initialize, after all.

Matthias247:
This was more about a longer Inline Array length potentially making sense there

Ralith:
I'm confused, then. Tinyvec should still provide that benefit, no? Is there any case we have that smallvec should be expected to perform better on?

Matthias247:
My understanding is that TinyVec initialed all elements whereas SmallVec leaves them Uninit until use. So the creation costs are a tiny bit higher

Ralith:
Ah, right. I'd like to see whether the difference registers.

Matthias247:
Updated for TinyVec. We can revisit once the other optimizations have landed.

Matthias247:
Seems like `Default` isn't implemented for `Range` on 1.45.

Not sure when that was added.
Options:
1. Increase minimum supported version
2. Back to SmallVec
3. Add a wrapper around Range

I'm not really a fan of 3. Happy with any of the other versions.

Regarding "NIH": I prefer to avoid it, since we then just have one more place where we discuss about unsafe code. I rather would spend the same time reviewing the utilized code path in SmallVec.

djc:
Other option: implement custom `Default` for `StreamMeta`? Seems like the most attractive option to me.

quinn-proto/src/lib.rs
```diff
@@ -212,7 +212,7 @@ impl fmt::Display for Dir {
 
 /// Identifier for a stream within a particular connection
 #[cfg_attr(feature = "arbitrary", derive(Arbitrary))]
-#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+#[derive(Debug, Default, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
```
djc, line 215: I'd prefer for this not to have a `Default` impl, instead creating it explicitly in the `StreamMeta::default()` impl (ideally with some comment explaining how we make sure this value never gets used).

quinn-proto/src/frame.rs
```diff
@@ -429,6 +430,19 @@ pub struct StreamMeta {
     pub fin: bool,
 }
 
+// This manual implementation exists because `Default` is not implemented for
+// `Range<u64>` on Rust 1.45. If the minimum version is bumped in the future,
+// this can be replaced with `derive(Default)`.
```
djc, line 435: Let's specifically use `MSRV` here so it's easier to find when we bump it.

```suggestion
// `Range<u64>` on Rust 1.45. If the MSRV is bumped in the future, this can be replaced
// with `derive(Default)`.
```

djc (commented):
Some minor nits left. Did you observe a perf difference between smallvec and tinyvec?

quinn-proto/src/lib.rs
Matthias247: Ok. I don't think it hurts, since a Stream ID of 0 is just fine and makes sense too. But I am also fine with removing it.

Matthias247:
> Some minor nits left. Did you observe a perf difference between smallvec and tinyvec?

I didn't really benchmark. I thought it would first by worthwhile to improve a few other code parts, and then do some comparisons which simply swap those libraries and check the overall result.

Ralith (approved):

PR #1012, opened 2021-02-10
Minimize SentPacket size and allocations

Matthias247:
`SentPacket` stores information about data transmitted in each packet.
This includes lists for all possible frame types.

However only a tiny subset of frame types are used in most packets:
Those will only contain ACKs and STREAM frames. The frame types
tracked in Retransmits make up less than 1/100 of all other types in
a common setup. E.g. the benchmark which transmits 1GB of data
shows:

```
frame_tx: FrameStats {
        ACK: 903061,
        CRYPTO: 2,
        NEW_CONNECTION_ID: 5,
        STREAM: 903057,
        // All other types are 0
}
```

We can exploit this property to only allocate tracking state for the
additional frame types if they are actually transmitted. This can be
done by boxing the `Retransmit` type in `SentPacket` and `SentFrames`.

Doing this reduces the sizes of the tracking data structures:

**Before**:

> Retransmits size: 184
> SentPacket size: 256

**After:**

> Retransmits size: 184
> SentPacket size: 80

This means memory consumption for packet tracking should be roughly
halved.

Besides this, the change avoids a mandatory allocation that currently
happens inside `Retransmits::default` for the `crypto: VecDeque`
list => In the best case we are avoiding one allocation per packet, in
the worst case the amount of allocations is the same.

The performance impact is very noticable - even without further
optimizations like GSO:

**Before:**

> Sent 1073741824 bytes on 1 streams in 5.11s (200.28 MiB/s)

**After:**

> Sent 1073741824 bytes on 1 streams in 4.80s (213.54 MiB/s)

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -46,7 +46,7 @@ use paths::PathData;
 mod send_buffer;
 
 mod spaces;
-use spaces::{PacketSpace, Retransmits, SentPacket};
+use spaces::{PacketSpace, Retransmits, SentPacket, ThinRetransmits};
 
 mod stats;
 pub use stats::ConnectionStats;
@@ -1325,8 +1325,10 @@ where
         }
 
         // Update state for confirmed delivery of frames
-        for (id, _) in info.retransmits.reset_stream {
-            self.streams.reset_acked(id);
+        if let Some(retransmits) = info.retransmits.get() {
+            for (id, _) in retransmits.reset_stream.iter() {
+                self.streams.reset_acked(*id);
+            }
         }
 
         for frame in info.stream_frames {
@@ -2786,7 +2788,7 @@ where
         // HANDSHAKE_DONE
         if !is_0rtt && mem::replace(&mut space.pending.handshake_done, false) {
             buf.write(frame::Type::HANDSHAKE_DONE);
-            sent.retransmits.handshake_done = true;
+            sent.retransmits.get_or_create().handshake_done = true;
             // This is just a u8 counter and the frame is typically just sent once
             self.stats.frame_tx.handshake_done =
                 self.stats.frame_tx.handshake_done.saturating_add(1);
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -1324,8 +1324,13 @@ where
         }
 
         // Update state for confirmed delivery of frames
-        for (id, _) in info.retransmits.reset_stream {
-            self.streams.reset_acked(id);
+        match &info.retransmits {
```
djc, line 1327: Can we do something like `with_retransmits()` that covers this case and the one in `ensure_probe_queued()`?

djc (commented):
Looks good to me... would be good if we can cut down on the cruft this introduces (especially the `on_packet_acked()` and `ensure_probe_queued()` calls). I think it might make sense to introduce a newtype for `Option<Box<Retransmits>>` that could have some helper methods, what do you think?

quinn-proto/src/connection/streams.rs
```diff
@@ -602,6 +604,13 @@ impl Streams {
         }
     }
 
+    fn with_retransmits(sent: &mut Option<Box<Retransmits>>, func: impl FnOnce(&mut Retransmits)) {
+        if sent.is_none() {
+            *sent = Some(Box::new(Retransmits::default()));
+        }
+        func(sent.as_mut().unwrap())
+    }
+
```
Ralith, line 613: This might be cleaner as a top-level function.

Ralith (dismissed):

quinn-proto/src/connection/mod.rs
Matthias247: I agree, it's probably cleaner. But then we are at the `Retransmits` and `InnerRetransmits` type, or maybe `Retransmits` and `OptionalRetransmits`, and I thought @Ralith wasn't too much of a fan.

I guess I can live with `struct OptionalRetransmits(Option<Box<Retransmits>>)`

Ralith: I appreciate being able to express that something *isn't* optional. I don't have any objection to having both. `ThinRetransmits` could also be a good name.

Ralith (approved):

djc:
Very nice!

PR #1013, opened 2021-02-10
Add a `Bytes` array based send interface to quinn-proto

Matthias247:
Add a `Bytes` array based send interface to `quinn-proto`

This change modifies quinn-proto to allow callers to submit a list
of one or more owned `Bytes` chunks for transmission instead of
pure byte slices. This will provide a more efficient zero-copy
interface for applications which already make use of owned
`Bytes` buffers.

The internals of quinn-proto have been refactored in order to keep
the ability to pass `&[u8]` buffers and defer the conversion into
`Bytes` as long as possible, in order to avoid unnecessary allocations
if no data can be stored due to flow control.

### DIFF
--- a/quinn-proto/src/bytes_source.rs
+++ b/quinn-proto/src/bytes_source.rs
@@ -1,0 +1,211 @@
+//! Defines types which can be lazily converted into `Bytes` chunks
+
+use bytes::Bytes;
+
+/// A source of one or more buffers which can be converted into `Bytes` buffers on demand
+///
+/// The purpose of this data type is to defer conversion as long as possible,
+/// so that no heap allocation is required in case no data is writable.
+pub trait BytesSource {
+    /// Returns the next chunk from the source of owned chunks.
+    ///
+    /// This method will consume parts of the source.
+    /// Calling it will yield `Bytes` elements up to the configured `limit`.
+    ///
+    /// The method returns a tuple:
+    /// - The first item is the yielded `Bytes` element. The element will be
+    ///   empty if the limit is zero or no more data is available.
+    /// - The second item returns how many complete chunks inside the source had
+    ///   had been consumed. This can be less than 1, if a chunk inside the
+    ///   source had been truncated in order to adhere to the limit. It can also
+    ///   be more than 1, if zero-length chunks had been skipped.
+    fn pop_chunk(&mut self, limit: usize) -> (Bytes, usize);
+}
+
+/// Indicates how many bytes and chunks had been transferred in a write operation
+#[derive(Debug, Default, PartialEq, Eq, Clone, Copy)]
+pub struct Written {
+    /// The amount of bytes which had been written
+    pub bytes: usize,
+    /// The amount of full chunks which had been written
+    ///
+    /// If a chunk was only partially written, it will not be counted by this field.
+    pub chunks: usize,
+}
+
+/// A [`BytesSource`] implementation for `&'a mut [Bytes]`
+///
+/// The type allows to dequeue [`Bytes`] chunks from an array of chunks, up to
+/// a configured limit.
+pub struct BytesArray<'a> {
+    /// The wrapped slice of `Bytes`
+    chunks: &'a mut [Bytes],
+    /// The amount of chunks consumed from this source
+    consumed: usize,
+}
+
+impl<'a> BytesArray<'a> {
+    pub fn from_chunks(chunks: &'a mut [Bytes]) -> Self {
+        Self {
+            chunks,
+            consumed: 0,
+        }
+    }
+}
+
+impl<'a> BytesSource for BytesArray<'a> {
+    fn pop_chunk(&mut self, limit: usize) -> (Bytes, usize) {
+        // The loop exists to skip empty chunks while still marking them as
+        // consumed
+        let mut chunks_consumed = 0;
+
+        while self.consumed < self.chunks.len() {
+            let chunk = &mut self.chunks[self.consumed];
+
+            if chunk.len() <= limit {
+                let chunk = std::mem::take(chunk);
+                self.consumed += 1;
+                chunks_consumed += 1;
+                if chunk.is_empty() {
+                    continue;
+                }
+                return (chunk, chunks_consumed);
+            } else if limit > 0 {
+                let chunk = chunk.split_to(limit);
+                return (chunk, chunks_consumed);
+            } else {
+                break;
+            }
+        }
+
+        (Bytes::new(), chunks_consumed)
+    }
+}
+
+/// A [`BytesSource`] implementation for `&[u8]`
+///
+/// The type allows to dequeue a single [`Bytes`] chunk, which will be lazily
+/// created from a reference. This allows to defer the allocation until it is
+/// known how much data needs to be copied.
+pub struct ByteSlice<'a> {
+    /// The wrapped byte slice
+    data: &'a [u8],
+}
+
+impl<'a> ByteSlice<'a> {
+    pub fn from_slice(data: &'a [u8]) -> Self {
+        Self { data }
+    }
+}
+
+impl<'a> BytesSource for ByteSlice<'a> {
+    fn pop_chunk(&mut self, limit: usize) -> (Bytes, usize) {
+        let limit = limit.min(self.data.len());
+        if limit == 0 {
+            return (Bytes::new(), 0);
+        }
+
+        let chunk = Bytes::from(self.data[..limit].to_owned());
+        self.data = &self.data[chunk.len()..];
+
+        let chunks_consumed = if self.data.is_empty() { 1 } else { 0 };
+        (chunk, chunks_consumed)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn bytes_array() {
+        let full = b"Hello World 123456789 ABCDEFGHJIJKLMNOPQRSTUVWXYZ".to_owned();
+        for limit in 0..full.len() {
+            let mut chunks = [
+                Bytes::from_static(b""),
+                Bytes::from_static(b"Hello "),
+                Bytes::from_static(b"Wo"),
+                Bytes::from_static(b""),
+                Bytes::from_static(b"r"),
+                Bytes::from_static(b"ld"),
+                Bytes::from_static(b""),
+                Bytes::from_static(b" 12345678"),
+                Bytes::from_static(b"9 ABCDE"),
+                Bytes::from_static(b"F"),
+                Bytes::from_static(b"GHJIJKLMNOPQRSTUVWXYZ"),
+            ];
+            let num_chunks = chunks.len();
+            let last_chunk_len = chunks[chunks.len() - 1].len();
+
+            let mut array = BytesArray::from_chunks(&mut chunks);
+
+            let mut buf = Vec::new();
+            let mut chunks_popped = 0;
+            let mut chunks_consumed = 0;
+            let mut remaining = limit;
+            loop {
+                let (chunk, consumed) = array.pop_chunk(remaining);
+                chunks_consumed += consumed;
+
+                if !chunk.is_empty() {
+                    buf.extend_from_slice(&chunk);
+                    remaining -= chunk.len();
+                    chunks_popped += 1;
+                } else {
+                    break;
+                }
+            }
+
+            assert_eq!(&buf[..], &full[..limit]);
+
+            if limit == full.len() {
+                // Full consumption of the last chunk
+                assert_eq!(chunks_consumed, num_chunks);
+                // Since there are empty chunks, we consume more than there are popped
+                assert_eq!(chunks_consumed, chunks_popped + 3);
+            } else if limit > full.len() - last_chunk_len {
+                // Partial consumption of the last chunk
+                assert_eq!(chunks_consumed, num_chunks - 1);
+                assert_eq!(chunks_consumed, chunks_popped + 2);
+            }
+        }
+    }
+
+    #[test]
+    fn byte_slice() {
+        let full = b"Hello World 123456789 ABCDEFGHJIJKLMNOPQRSTUVWXYZ".to_owned();
+        for limit in 0..full.len() {
+            let mut array = ByteSlice::from_slice(&full[..]);
+
+            let mut buf = Vec::new();
+            let mut chunks_popped = 0;
+            let mut chunks_consumed = 0;
+            let mut remaining = limit;
+            loop {
+                let (chunk, consumed) = array.pop_chunk(remaining);
+                chunks_consumed += consumed;
+
+                if !chunk.is_empty() {
+                    buf.extend_from_slice(&chunk);
+                    remaining -= chunk.len();
+                    chunks_popped += 1;
+                } else {
+                    break;
+                }
+            }
+
+            assert_eq!(&buf[..], &full[..limit]);
+            if limit != 0 {
+                assert_eq!(chunks_popped, 1);
+            } else {
+                assert_eq!(chunks_popped, 0);
+            }
+
+            if limit == full.len() {
+                assert_eq!(chunks_consumed, 1);
+            } else {
+                assert_eq!(chunks_consumed, 0);
+            }
+        }
+    }
+}
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -14,6 +14,7 @@ use thiserror::Error;
 use tracing::{debug, error, trace, trace_span, warn};
 
 use crate::{
+    bytes_source::{ByteSlice, BytesArray, BytesSource},
     cid_generator::ConnectionIdGenerator,
     cid_queue::CidQueue,
     coding::BufMutExt,
@@ -29,7 +30,7 @@ use crate::{
         EndpointEventInner, IssuedCid,
     },
     transport_parameters::TransportParameters,
-    Dir, Frame, Side, StreamId, Transmit, TransportError, TransportErrorCode, VarInt,
+    Dir, Frame, Side, StreamId, Transmit, TransportError, TransportErrorCode, VarInt, Written,
     MAX_STREAM_COUNT, MIN_INITIAL_SIZE, RESET_TOKEN_SIZE, TIMER_GRANULARITY,
 };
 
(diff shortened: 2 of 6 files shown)


quinn-proto/src/connection/bytes_source.rs
```diff
@@ -0,0 +1,249 @@
...
+    fn pop_chunk(&mut self) -> Option<Bytes>;
+}
+
+/// A [`BytesSource`] implementation for `&'a mut [Bytes]`.
+///
+/// The type allows to dequeue [`Bytes`] chunks from an array of chunks, up to
+/// a configured limit.
+pub struct BytesArray<'a> {
+    chunks: &'a mut [Bytes],
+    len: usize,
```
djc, line 33: Is `len` in bytes or in chunks? Looks like the former, but would be good to clarify either by changing the name or adding a comment.

quinn-proto/src/connection/bytes_source.rs
```diff
@@ -0,0 +1,249 @@
+//! Defines types which can be lazily converted into `Bytes` chunks
+
+use bytes::Bytes;
+
+/// A source of one or more buffers which can be converted into `Bytes` buffers on demand.
```
djc, line 5: By convention, this trait definition should go at the bottom of the module. Also, please drop the periods from the first sentence of each comment.

quinn-proto/src/connection/bytes_source.rs
```diff
@@ -0,0 +1,249 @@
...
+    // Returns the total amount of bytes to write
+    fn len(&self) -> usize;
+    /// Limits the source to the new size
+    fn limit(&mut self, new_len: usize);
+    /// The amount of bytes consumed from this source
+    fn bytes_consumed(&self) -> usize;
+    /// The amount of chunks consumed from this source
+    fn chunks_consumed(&self) -> usize;
+    /// Whether a chunk was partially consumed from this source
+    fn partial_chunk_consumed(&self) -> bool;
```
djc, line 19: I feel like the consumed interface is very verbose. Could we maybe simplify it to `consumed(&self) -> (usize, Option<usize>)`?

quinn-proto/src/connection/streams/send.rs
```diff
@@ -51,7 +50,7 @@ impl Send {
         }
     }
 
-    pub(super) fn write(&mut self, data: &[u8]) -> Result<usize, WriteError> {
+    pub(super) fn write<S: BytesSource>(&mut self, source: &mut S) -> Result<usize, WriteError> {
```
djc, line 53: Should just this return `()` for `Ok`, since the caller can get the consumption from the `BytesSource` itself?

quinn-proto/src/connection/mod.rs
```diff
@@ -947,6 +950,29 @@ where
...
+    }
+
+    /// Send data on the given stream
+    ///
+    /// Returns the number of bytes successfully written.
+    pub fn write_chunks(
+        &mut self,
+        stream: StreamId,
+        data: &mut [Bytes],
+    ) -> Result<(usize, bool), WriteError> {
```
djc, line 965: Why is it useful to return the `partial_chunk_consumed()` value here? Presumably the caller can just forget about/recycle the consumed chunks, and the partially consumed chunk has been split, such that it's equivalent to an unconsumed chunk, isn't it?

djc (commented):
I like the direction here! A bunch of questions about the details...

djc:
Also, do you have any benchmarking results for what this API enables?

quinn-proto/src/connection/streams/send.rs
Matthias247: Yeah. I just kept it because `Streams` directly consumes the return value. But probably it's better to switch for `()` for avoiding duplicate information.

quinn-proto/src/connection/bytes_source.rs
Matthias247: I don't know. It's nice to understand based on the names what is going on. And it would be `consumed(&self) -> (usize, usize, bool)` - since the API can consume e.g. 1.5 buffers. Then users are more likely to mess up the order of arguments.

quinn-proto/src/connection/bytes_source.rs
Matthias247: It's in bytes. What kind of alternative name would you imagine? `num_bytes`? `bytes_len`?

quinn-proto/src/connection/mod.rs
Matthias247: The caller needs to be aware that this buffer had been mutated. We can't claim it as fully consumed, since the caller needs to submit the leftovers again in another write call. For fully consumed buffers, the caller will just drop/skip them after the call. For the partially consumed one this isn't possible.

An alternative can be to report the number of mutated buffers, including the last one. But this requires the caller to check the buffers and see if one wasn't fully consumed.

Matthias247:
> Also, do you have any benchmarking results for what this API enables?

Not really, since I haven't started working on the `quinn` API yet. Based on the flamegraphs I would say maybe 2% if the application already has data in `Bytes` format - which is true for my use-case.

But there is another benefit that the data copy and buffer allocation no longer happens inside the `Connection` mutex, which could have an additional impact.

quinn-proto/src/connection/bytes_source.rs
djc: I guess `len` is fine, just add a comment to make it explicit.

quinn-proto/src/connection/mod.rs
djc: Sorry, perhaps I'm being stupid -- why does the caller need to be aware that the buffer has been mutated? We could just mark it as unconsumed.

Matthias247: One example that I have in code around it is a timeout wrapper around the stream.

If 0 chunks have been consumed, it is equivalent to `Pending` and means the timer needs to continue to run. If a partial chunk is consumed this means the stream is still active and we have to reset the timer. 

Something like that could probably also be done by comparing the size of the first buffer before and after the write, but I kind of liked the partial signal. I guessed there might also be other use-cases where it's important whether e.g. the 3rd buffer was partially consumed or not, but I don't have a concrete example for it. The code I have for this will retry any non-fully consumed buffers again, so there is no special handling required.

So I guess I would also be fine with only reporting the amount of fully consumed buffers, or by reporting the amount of consumed buffers plus consumed bytes

djc: Yes, I think reporting fully consumed buffers + consumed bytes seems like the right approach here. That should easily satisfy the timeout wrapper use case at least (I agree that comparing the size of the first buffer is a little unwieldy).

quinn-proto/src/connection/bytes_source.rs
djc: I'm also fine with making it a named struct.

Matthias247: Ok. Should this then also be used for the public `quinn-proto` API? any recommendations for name? `BytesWritten` sounds like it contains only bytes. `WriteResult` sounds more like a `Result` type

djc: Yeah, I'd be fine with having that as public API, I think.

Matthias247:
The new clippy warning seems like a rustc 1.50 artifact

quinn-proto/src/connection/bytes_source.rs
djc: The trait definition is still at the top, and the `Written` definition should go beneath it.

quinn-proto/src/bytes_source.rs
```diff
@@ -0,0 +1,231 @@
...
+            let chunk = &mut self.chunks[self.consumed.chunks];
+
+            if chunk.len() <= self.len {
+                let chunk = std::mem::take(chunk);
+                self.len -= chunk.len();
+                self.consumed.chunks += 1;
+                self.consumed.bytes += chunk.len();
+                if chunk.is_empty() {
+                    continue;
+                }
```
djc, line 84:
```suggestion
                let chunk = std::mem::take(chunk);
                self.consumed.chunks += 1;
                if chunk.len() == 0 {
                    continue;
                }
                self.len -= chunk.len();
                self.consumed.bytes += chunk.len();
```

quinn-proto/src/bytes_source.rs
```diff
@@ -0,0 +1,231 @@
...
+                return Some(chunk);
+            } else if self.len > 0 {
+                let chunk = chunk.split_to(self.len);
```
djc, line 87: `split_to()` panics if the argument is greater than the chunk's length, which we don't check here, right?

quinn-proto/src/bytes_source.rs
```diff
@@ -0,0 +1,231 @@
...
+    fn limit(&mut self, new_len: usize) {
+        self.len = self.len.min(new_len);
+    }
+
+    fn consumed(&self) -> Written {
+        self.consumed
+    }
+
+    fn pop_chunk(&mut self) -> Option<Bytes> {
+        if self.len > 0 {
```
djc, line 139: Let's invert this check and return `None` early.

djc (commented):
See comments.

quinn-proto/src/connection/bytes_source.rs
djc: As I wrote in the first comment in this thread, the trait definition should go at the bottom of the module.

quinn-proto/src/bytes_source.rs
djc: Ah, fair enough, sorry!

quinn-proto/src/bytes_source.rs
djc: In this case, this was more for the human reader than for the machines.

djc:
@Ralith any thoughts?

quinn-proto/src/connection/bytes_source.rs
Matthias247: Sorry, I think I misunderstood. I put it beyond `Written` since I thought it would still be more important to most readers than the implementations, which are used only in a tiny set of places.

Matthias247: I can move it down

Matthias247: Actually I'm confused now. Either we move all dependencies down (`Written` + `BytesSource`), and keep the less important internal implementations on the top. Or we keep it as currently is. But I'm not sure if you want
- `Written` (publically exported)
- `BytesArray`
- `ByteSlice`
- `BytesSource`

?

Can you specify which order you intent?

quinn-proto/src/bytes_source.rs
```diff
@@ -0,0 +1,231 @@
...
+
+use bytes::Bytes;
+
+/// A source of one or more buffers which can be converted into `Bytes` buffers on demand
+///
+/// The purpose of this data type is to defer conversion as long as possible,
+/// so that no heap allocation is required in case no data is writable.
+pub trait BytesSource {
+    // Returns the total amount of bytes to write
+    fn len(&self) -> usize;
```
Ralith, line 11: Is this method used anywhere outside of tests? If not, can we get rid of it?

quinn-proto/src/bytes_source.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// Limits the source to the new size
+    fn limit(&mut self, new_len: usize);
```
Ralith, line 13: Rather than an extra method and associated state, could we add a `max_len: usize` to `pop_chunk`? We'd have to pass a limit from `Streams` to `streams::Send`, but that seems inoffsensive.

quinn-proto/src/bytes_source.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// The amount of bytes and chunks consumed from this source
+    fn consumed(&self) -> Written;
```
Ralith, line 15: Is there a benefit to this being tracked by the source? Intuitively it seems like the accumulators could easily be maintained source-independently in the ultimate `fn write` implementation.

Ralith (changes requested):
I think this is a good direction. However, I'm concerned about how complex the new trait is. It's not part of the public interface so it's not the end of the world, but I do think we can simplify it considerably. Concrete suggestions follow.

quinn-proto/src/bytes_source.rs
Matthias247: I guess it can be removed. It kind of comes for free and it seems like for anything bytes-like a length makes sense. But if there are no consumers it might indeed to better to leave it out.

quinn-proto/src/bytes_source.rs
Matthias247: It can be. But then we have to subtract the limit in `Stream::write` based on what was already written. I feel like then actually more code with less test coverage is tasked with keeping track of limits. This way its fairly concentrated in one place.

quinn-proto/src/bytes_source.rs
Matthias247: Kind of the same answer as for limits. It more or less requires all intermediate code to track the same information that the source already tracks. And we can't drop the information from there, since it also requires it to determine the offsets of the next data to write.

quinn-proto/src/bytes_source.rs
```diff
@@ -0,0 +1,231 @@
...
+/// A [`BytesSource`] implementation for `&[u8]`
+///
+/// The type allows to dequeue a single [`Bytes`] chunk, which will be lazily
+/// created from a reference. This allows to defer the allocation until it is
+/// known how much data needs to be copied.
+pub struct ByteSlice<'a> {
+    /// The wrapped byte slice
+    data: &'a [u8],
+    /// The amount of bytes which are still available for consumption
+    len: usize,
```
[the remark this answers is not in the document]
Matthias247: There are 2 reasons why not:
1. The borrow checker hates me (I tried it!)
2. We would need a separate `limit` field in order to return later on whether the full chunk was written, or would need to additionally store the original data length without slicing. And while `Written::chunks` iis not terribly useful for the normal slice based write API since it is returned to nobody - I think it is still nice if everything upholds the API contract. Otherwise it might confuse people later on.

quinn-proto/src/bytes_source.rs
Ralith: I'd think there's even less communication between intermediate layers here than in the `limit` case. We have exactly one location that ever calls `pop_chunk`, and it can very naturally return the `Written` counters back up the stack.

quinn-proto/src/bytes_source.rs
Ralith: It seems like the current approach has to duplicate the logic across both impls, whereas moving it out of the trait would yield a single implementation, and a trait that's simpler to reason about besides.

quinn-proto/src/bytes_source.rs
Matthias247: Here is an attempt: https://github.com/quinn-rs/quinn/compare/main...Matthias247:bytes_write_2?expand=1

Ignore most of the `ByteSource` implementation. It's not updated and broken, just has the updated API.
Most things seem to work fine. However the loop in `Send::write` is broken: It can't properly populate `Written::chunks`, since it doesn't know whether `pop_chunk` returned a full chunk or whether it was trunated due to the limit.

In order to fix this, we we would need to add either a bool to `pop_chunk` which indicates if this consumed a full buffer (which is however somewhat weird, since the internal buffer structures are transparent to the caller), or return a `Option<Written, Bytes>` from `pop_chunk`.

Matthias247: So returning `Option<(Written, Bytes)>` wasn't optimal either, because it didn't allow to indicate we consumed chunks in the source if no `Bytes` elements had been pulled (which can happen in case empty `Bytes` elements are enqueued). It's obviously a borderline thing of invalid input, but I still think we need to handle it correctly.

One option to fix it is to change the return value to `(Option<Bytes>, Written)` or `(Option<Bytes>, usize)`, so that consumed chunks can be reported independent of pulling a chunk. I updated for this in the last version of the linked other variant.

I'm split on whether I like it better than this approach here. Yes - the trait is smaller. However returning a consumed chunks number is weird, since the trait nowhere else mentions multiple chunks.

Ralith: The new branch actually looks pretty good to me in its current state. The simpler interface makes the semantics a lot more obvious, and only mentioning multiple chunks in one place isn't obscure when there's only one method on the trait to begin with. I agree with the decision re: return values, as well, except that I wonder if `Option<Bytes>` could be simplified to `Bytes` by returning an empty `Bytes` when appropriate.

Matthias247: ok. I will then update to it. I had the same thought about `Bytes::new()`. It might be actually be a bit more efficient. But on the other hand the explicit `None` is a good signal to stop iteration. I have no strong thoughts on it.

Matthias247: I now updated for `Bytes::new` since I think it looks a bit more concise on the consumer side.

quinn-proto/src/bytes_source.rs
```diff
@@ -0,0 +1,211 @@
...
+    /// - The second item returns how many complete chunks inside the source had
+    ///   had been consumed. This can be less than 1, if a chunk inside the
+    ///   source had been truncated in order to adhere to the limit. It can also
+    ///   be more than 1, if zero-length chunks had been skipped.
+    fn pop_chunk(&mut self, limit: usize) -> (Bytes, usize);
+}
+
+/// Indicates how many bytes and chunks had been transferred in a write operation
+#[derive(Debug, Default, PartialEq, Eq, Clone, Copy)]
+pub struct Written {
```
Ralith, line 27: Post-refactor, this should probably be defined in `connection::streams::send`, the sole location where it's constructed.

quinn-proto/src/bytes_source.rs
```diff
@@ -0,0 +1,211 @@
...
+        (Bytes::new(), chunks_consumed)
+    }
+}
+
+/// A [`BytesSource`] implementation for `&[u8]`
+///
+/// The type allows to dequeue a single [`Bytes`] chunk, which will be lazily
+/// created from a reference. This allows to defer the allocation until it is
+/// known how much data needs to be copied.
+pub struct ByteSlice<'a> {
```
Ralith, line 90: Cheeky bonus points if you can get rid of the struct and impling on `&[u8]` directly. Would make the tests a bit more concise!

Ralith (approved):

Ralith (approved):

PR #1014, opened 2021-02-11
quinn-proto: deduplicate when entering unordered mode (fix #1009)

### LINKED ISSUE #1009: Previously inserted duplicates remain after switching from ordered to unordered state
While working on #1000, I realized that we are not deduplicating already inserted duplicates when entering unordered mode: https://github.com/quinn-rs/quinn/blob/2a8c93bf0f987b4118fdaf1d0d1489c452810e19/quinn-proto/src/connection/assembler.rs#L32-L38
With that behavior, the following test will fail:
```rust
    #[test]
    fn read_unordered() {
        let mut x = Assembler::new();
        x.insert(0, Bytes::from_static(b"abc"), 3);
        x.insert(0, Bytes::from_static(b"abc"), 3);
        x.ensure_ordering(false).unwrap();
        assert_eq!(
            x.read(3, false),
            Some(Chunk::new(0, Bytes::from_static(b"abc")))
        );
        assert_eq!(x.read(3, false), None);
    }
```

Is this intentional?

geieredgar:


Ralith (approved):
Thanks!

PR #1017, opened 2021-02-14
Allow to specify the cipher suite in bulk test

Matthias247:
rustls defaults to CHACHA20, which is rather slow on desktop hardware.
This change allows to specify a single cipher suite for this test, and
sets the default to AES128.

CHACHA20 (default):
> Sent 4294967296 bytes on 1 streams in 14.86s (275.68 MiB/s)

AES128:
> Sent 4294967296 bytes on 1 streams in 12.21s (335.59 MiB/s)

### DIFF
--- a/bench/src/bulk.rs
+++ b/bench/src/bulk.rs
@@ -1,5 +1,6 @@
 use std::{
     net::{IpAddr, Ipv6Addr, SocketAddr},
+    str::FromStr,
     sync::Arc,
     time::{Duration, Instant},
 };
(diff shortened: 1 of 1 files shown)


bench/src/bulk.rs
```diff
@@ -116,7 +117,14 @@ async fn client(server_addr: SocketAddr, server_cert: quinn::Certificate, opt: O
         .bind(&SocketAddr::new(IpAddr::V6(Ipv6Addr::LOCALHOST), 0))
         .unwrap();
 
-    let mut client_config = quinn::ClientConfigBuilder::default();
+    let mut config = quinn::ClientConfig::default();
+    let crypto_config: &mut rustls::ClientConfig =
+        Arc::get_mut(&mut config.crypto).context("retrieving crypto config failed")?;
```
Ralith, line 122: I'd just unwrap here, since it's trivially infallible.

Ralith (approved):

bench/src/bulk.rs
djc: Agreed. We can also probably do without the type annotation here?

djc (approved):

PR #1019, opened 2021-02-14
Prefer AES cipher suites

Matthias247:
Testing showed that there is a huge performance boost using AES ciphers
due to hardware acceleration. Therefore those should be preferred.

As an example, a benchmark run using CHACHA20 reached a throughput
of 350MB/s, whereas the same configuration using AES128 reached 520MB/s.

While CHACHA20 might have higher performance on devices without
hardware acceleration for AES, this set of devices might now be tiny.

### DIFF
--- a/quinn-proto/src/crypto/rustls.rs
+++ b/quinn-proto/src/crypto/rustls.rs
@@ -276,7 +276,7 @@ pub struct HandshakeData {
 
 impl crypto::ClientConfig<TlsSession> for Arc<rustls::ClientConfig> {
     fn new() -> Self {
-        let mut cfg = rustls::ClientConfig::new();
+        let mut cfg = rustls::ClientConfig::with_ciphersuites(&QUIC_CIPHER_SUITES);
         cfg.versions = vec![rustls::ProtocolVersion::TLSv1_3];
         cfg.enable_early_data = true;
         #[cfg(feature = "native-certs")]
@@ -320,7 +320,10 @@ impl crypto::ClientConfig<TlsSession> for Arc<rustls::ClientConfig> {
 
 impl crypto::ServerConfig<TlsSession> for Arc<rustls::ServerConfig> {
     fn new() -> Self {
-        let mut cfg = rustls::ServerConfig::new(rustls::NoClientAuth::new());
+        let mut cfg = rustls::ServerConfig::with_ciphersuites(
+            rustls::NoClientAuth::new(),
+            &QUIC_CIPHER_SUITES,
+        );
         cfg.versions = vec![rustls::ProtocolVersion::TLSv1_3];
         cfg.max_early_data_size = u32::max_value();
         Arc::new(cfg)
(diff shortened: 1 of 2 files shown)


djc:
The new rustls defaults prefer AES256 over AES128 (see https://github.com/ctz/rustls/commit/7117a805e0104705da50259357d8effa7d599e37):

```
### Why is AES-256 preferred over AES-128?
This is a trade-off between:
1. classical security level: searching a 2^128 key space is as implausible as 2^256.
2. post-quantum security level: the difference is more meaningful, and AES-256 seems like the conservative choice.
3. performance: AES-256 is around 40% slower than AES-128, though hardware acceleration typically narrows this gap.
The choice is frankly quite marginal.
```

Should we align with those choices?

quinn-proto/src/crypto/rustls.rs
```diff
@@ -399,3 +402,14 @@ impl crypto::PacketKey for PacketKey {
         }
     }
 }
+
+/// Cipher suites suitable for QUIC
+///
+/// The list is equivalent to TLS1.3 ciphers.
+/// Cipher suites are ordered by efficiency. On most desktop and mobile hardware
+/// AES ciphers suites will be more efficient due to hardware acceleration.
```
Ralith, line 410: Let's add a note here to remove once rustls is updated; I don't see value in us maintaining our own list long term.

Matthias247:
> The new rustls defaults prefer AES256 over AES128 (see [ctz/rustls@7117a80](https://github.com/ctz/rustls/commit/7117a805e0104705da50259357d8effa7d599e37)):
> 
> ```
> ### Why is AES-256 preferred over AES-128?
> This is a trade-off between:
> 1. classical security level: searching a 2^128 key space is as implausible as 2^256.
> 2. post-quantum security level: the difference is more meaningful, and AES-256 seems like the conservative choice.
> 3. performance: AES-256 is around 40% slower than AES-128, though hardware acceleration typically narrows this gap.
> The choice is frankly quite marginal.
> ```
> 
> Should we align with those choices?

I guess that would be fair. I'm not crypto expert. As far as I understand AES128 is deemed safe (otherwise it wouldn't be part of TLS1.3), so I would probably have chosen this as a default in order to be as efficient as possible. 
But given that even AES256 with hardware acceleration seems a lot better than the ChaCha20 config, I would be ok with switching to that. It is still a better default config, and sers can still switch to something else manually.

For the bulk test I would however still recommend to keep the proposed AES128 default (#1017), so that all other optimizations are more visible and less hidden behind crpyto bottlenecks.

quinn-proto/src/crypto/rustls.rs
```diff
@@ -399,3 +402,17 @@ impl crypto::PacketKey for PacketKey {
...
     }
 }
+
+/// Cipher suites suitable for QUIC
+///
+/// The list is equivalent to TLS1.3 ciphers.
+/// It matches the rustls prefernce list that was introduced with
+/// https://github.com/ctz/rustls/commit/7117a805e0104705da50259357d8effa7d599e37.
+/// This list prefers AES ciphers, which are hardware accelerated on most platforms.
+/// This list can be removed if the rustls dependency is updated ot a new version
```
Ralith, line 412:
```suggestion
/// This list can be removed if the rustls dependency is updated to a new version
```

Ralith (dismissed):

Ralith (approved):

djc:
I agree that sticking with AES128 makes sense for the benchmarks.

PR #1020, opened 2021-02-14
Disable hdrhistogram features

Matthias247:
This causes nom to be pulled in, which currently causes the CI error and build
conflict between `funty` and `bitvec`: https://github.com/myrrlyn/funty/issues/3

We don't need the feature in this scenario anyway.


djc (approved):

djc:
Thanks!

Issue #927, opened 2020-11-26
Feasibility of serialising Endpoint/Connection


infinity0:
I'd like to be able to serialise the state of the quinn-proto state machine, so that it can be suspended and resumed, which would enable nice capabilities like deterministic replay of logs. Is this a feature you'd be interested in supporting?

In principle this is simple when designed for initially, but adding it to an already-existing implementation is often quite challenging. I tried adding `#[derive(Serialize, Deserialize)]` to the existing code and can already see the following hurdles:

- StdRng is not Serialize/Deserialize, but could in theory
- Instant is not Serialize/Deserialize, but could in theory - https://github.com/serde-rs/serde/issues/1375
- Slab is not Serialize/Deserialize, but could in theory
- Arc is not Serialize/Deserialize and probably cannot be in theory - however the way I see it being used in quinn-proto is not actually for its atomic properties (is quinn-proto even thread-safe? don't see it being mentioned in the docs), but rather to have shared references to objects elsewhere, e.g. crypto::Session / ServerConfig etc. It's likely theoretically possible to either:
  a. rearchitect the layout of these structs so that such sharing is not necessary or
  b. (1) add surrogate EndpointFlat/ConnectionFlat objects with derive Serialize/Deserialize, then (2) add custom Serialize/Deserialize implementations for Endpoint/Connection that delegate to EndpointFlat/ConnectionFlat that unlink the sharing before serialisation / relink the sharing after deserialisation
- use of dynamic traits like `dyn ConnectionIdGenerator` and `dyn Controller`, which can be solved either by
  a. adding Serialize/Deserialize as a superclass (but this would be inconvenient for people that don't need it), or
  b. by reifying the types as associated types of some trait (similar to CryptoSession) and implementing Serialize/Deserialize for Endpoint/Connection only when these associated types implement Serialize/Deserialize. (If the controller consists only of static functions, then the state can just be a dummy struct with no fields.)

What do you think? It's obviously quite some work, but it's certainly a finite and bounded amount of work, and the outcome would be really useful in terms of protocol debugging.

infinity0:
I pushed the attempt here in case you want to try it out https://github.com/infinity0/quinn/tree/serde

burdges:
I'd presume the AEAD nonces prevent the worst dangers, but maybe TLS 1.3 remarks about storing 0-RTT states, like avoid doing so unless you know you'll update them.

Ralith:
[...]

Issue #911, opened 2020-11-13
Monomorphize public API


Ralith:
Quinn's public API is complicated by abstraction over the cryptographic backend, which is of niche interest. While the crypto traits themselves have good reasons not to be object-safe, we should be able to encapsulate the surrounding logic into an internal object-safe interface, and thereby erase the type parameter that's spread everywhere. By avoiding a need for downstream monomorphization, this may also significantly improve downstream compile times.

burdges:
I'd think binaries should never have more than one cryptographic backend, which usually favors monomorphization over trait objects.

Ralith:
My expectation is that the cost of a trait object will be a single-digit number of virtual calls per packet, which should be insignificant. We'll of course sanity-check against the bulk throughput benchmark before committing to a new approach.

DemiMarie:
Even a single-digit number of calls per packet can be significant at high data rates.  That said, someone flooding a QUIC implementation with lots of tiny packets is probably either trying to perform a DoS attack or has a buggy implementation.

infinity0:
Do you have a concrete idea for doing this? I don't think this is avoidable if you want users to be able to use different crypto implementations. Fundamentally, the user will need to be able to choose their own crypto key type, and some part of the public API will have to expose the concrete key type that the user chose (rather than just a trait), e.g. to get the peer's public key for a given connection.

What you can do however is have a mirror API that wraps everything in a 1-field struct that hard-codes `TlsSession`, then the monomorphisation will be down when compiling quinn rather than downstream.

Also, one parameter is the maximum you need - everything else can be associated types on (some common trait implemented by) that concrete type. This could include things like `dyn ConnectionIdGenerator` and `dyn Controller` that I pointed out in #927, avoiding `dyn` and making type-safe serialisation easier to implement.

Ralith:
> some part of the public API will have to expose the concrete key type that the user chose

No such API currently exists, unless I'm confused about your meaning. The choice of key can be isolated within the type erasure boundary.

infinity0:
> No such API currently exists, unless I'm confused about your meaning.

The current API has a parameter for the security session as a trait, which has various associated types including the public key type (AuthenticationData). I assumed there was an actual method you could call, it turns out not, but you could easily add one without changing anything else, as long as the concrete session type S is exposed as it is currently.

[...]

PR #1021, opened 2021-02-15
Simplify congestion interface by introducing a new top-level Session trait

infinity0:
**NOTE: This PR is incomplete** - it changes `quinn-proto` but I did not yet update the other crates because I wanted to clear the overall approach with you. If you approve, I will update the other crates accordingly.

This PR gets rid of the `ControllerFactory` interface and the ugly `Box<dyn Controller>`.

To make this work, we introduce a new top-level `endpoint::Session` trait that subclasses `crypto::Session` and defines the Controller type as an associated type. This same technique can be used to drop any other cases of `Box<dyn X>` e.g. `Box<dyn ConnectionIdGenerator>`, which I will do in a follow-up PR if this is accepted.

Dropping `Box<dyn X>` makes the state types easier (i.e. possible) to serialise, and makes progress on #927.

In terms of general engineering using advanced type theory mechanisms, I find that associated types are much easier to work with, when writing moderately-complex code in the long run, than existential (dyn) types. Since we have one of those already on `endpoint`, we can reuse it to remove all cases of `Box<dyn>`.

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -7,7 +7,6 @@ use thiserror::Error;
 use crate::crypto::types::{Certificate, CertificateChain, PrivateKey};
 use crate::{
     cid_generator::{ConnectionIdGenerator, RandomConnectionIdGenerator},
-    congestion,
     crypto::{self, ClientConfig as _, HandshakeTokenKey as _, HmacKey as _, ServerConfig as _},
     VarInt, VarIntBoundsExceeded,
 };
@@ -43,8 +42,6 @@ pub struct TransportConfig {
     pub(crate) allow_spin: bool,
     pub(crate) datagram_receive_buffer_size: Option<usize>,
     pub(crate) datagram_send_buffer_size: usize,
-
-    pub(crate) congestion_controller_factory: Box<dyn congestion::ControllerFactory + Send + Sync>,
 }
 
 impl TransportConfig {
@@ -195,25 +192,6 @@ impl TransportConfig {
         self.datagram_send_buffer_size = value;
         self
     }
-
-    /// How to construct new `congestion::Controller`s
-    ///
-    /// Typically the refcounted configuration of a `congestion::Controller`,
-    /// e.g. a `congestion::NewRenoConfig`.
-    ///
-    /// # Example
-    /// ```
-    /// # use quinn_proto::*; use std::sync::Arc;
-    /// let mut config = TransportConfig::default();
-    /// config.congestion_controller_factory(Arc::new(congestion::NewRenoConfig::default()));
-    /// ```
-    pub fn congestion_controller_factory(
-        &mut self,
-        factory: impl congestion::ControllerFactory + Send + Sync + 'static,
-    ) -> &mut Self {
-        self.congestion_controller_factory = Box::new(factory);
-        self
-    }
 }
 
 impl Default for TransportConfig {
@@ -243,8 +221,6 @@ impl Default for TransportConfig {
             allow_spin: true,
             datagram_receive_buffer_size: Some(STREAM_RWND as usize),
             datagram_send_buffer_size: 1024 * 1024,
-
-            congestion_controller_factory: Box::new(Arc::new(congestion::NewRenoConfig::default())),
         }
     }
 }
@@ -280,7 +256,6 @@ impl fmt::Debug for TransportConfig {
                 &self.datagram_receive_buffer_size,
             )
             .field("datagram_send_buffer_size", &self.datagram_send_buffer_size)
-            .field("congestion_controller_factory", &"[ opaque ]")
             .finish()
     }
 }
--- a/quinn-proto/src/congestion.rs
+++ b/quinn-proto/src/congestion.rs
@@ -6,7 +6,7 @@ mod new_reno;
 pub use new_reno::{NewReno, NewRenoConfig};
 
 /// Common interface for different congestion controllers
-pub trait Controller: Send {
+pub trait Controller: Send + Clone {
     /// Packet deliveries were confirmed
     ///
     /// `app_limited` indicates whether the connection was blocked on outgoing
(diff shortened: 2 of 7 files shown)


infinity0:
For simplicity for now I removed the ability to configure using a non-default `NewRenoConfig`, but we can easily add that back in via an additional `ControllerConfig` type if that is deemed necessary.

Ralith:
> remove all cases of Box<dyn>.

I'm not interested in doing this. Trait objects are the preferred extension mechanism for anything but extremely hot code, as they make separate compilation possible.

infinity0:
> Trait objects are the preferred extension mechanism for anything but extremely hot code, as they make separate compilation possible.

Can you explain this in more detail or link me to a place that explains this in more detail? I can't see why this would be such a big deal in practise. Furthermore do you not think the complexity reduction of this PR, more than makes up for any developer time savings due to "separate compilation".

The PR gets rid of `Box <dyn Controller>` in `PathData` as well - do you not consider this a "hot enough" path - it's accessed I'd guess for nearly every packet?

Ralith:
> Can you explain this in more detail or link me to a place that explains this in more detail? I can't see why this would be such a big deal in practise.

Codegen for generics takes place when they are instantiated, because not enough is known until then. This is one of the major sources of compiletime bloat in Rust.

> Furthermore do you not think the complexity reduction of this PR

I disagree that this PR represents a reduction in complexity. Further, it certainly has no impact at all to users who don't override the congestion controller, who are the overwhelming majority, and introduces an additional barrier to https://github.com/quinn-rs/quinn/issues/911.

> The PR gets rid of `Box<dyn Controller>` in `PathData` as well - do you not consider this a "hot enough" path - it's accessed I'd guess for nearly every packet?

I consider a path to be hot enough to merit compromises elsewhere when it is quantitatively shown to be so. We already do a lot of stuff that's far more expensive than a virtual call per packet (e.g. memory allocation). Optimization contributions are very welcome, but should be justified empirically, especially when they impose compromises elsewhere.

Matthias247:
I agree with @Ralith on this one. Adding generic parameters makes things usually More complex than simple. Mostly because those are contagious and bubble up. If they can be avoided without any perf impact it seems better to do so. 


And for this particular instance I actually think there is a good reason to keep dynamic Dispatch. It could allow to configure the congestion controller at handshake time. And to use c different ones based on SNI

infinity0:
> Codegen for generics takes place when they are instantiated, because not enough is known until then.

I know this, but it's not clear to me how this helps in this case, and how your conclusion follows from this premise. Since Rust is statically-linked, everyone has to compile this stuff anyway. In the typical use-case a user of quinn would only instantiate the parameter once. So the total compilation effort is roughly the same, regardless of whether separate compilation occurs or not. There should only be a substantial increase, if the parameter is instantiated several times with different values.

> I disagree that this PR represents a reduction in complexity.

Why's that? This PR removes 73 lines and adds 37 lines. In fact the true ratio is even higher because some of it is 1-1 renaming crypto::Session to endpoint::Session.

>  Adding generic parameters makes things usually More complex than simple. Mostly because those are contagious and bubble up [..]

This point may be true elsewhere, but the specific point of this PR is a design pattern to avoid this - it does not add a generic parameter (to the public API), it adds an associated trait to an existing generic parameter, and so there is no "contagion" towards the end user.

> I actually think there is a good reason to keep dynamic Dispatch. It could allow to configure the congestion controller at handshake time. And to use different ones based on SNI

You can do this with this PR, by implementing a new Controller type, or changing the existing default one, to perform the dynamic dispatch. (What you are talking about is actually conceptually a different thing from what this PR does - this PR changes the API for the *user* to choose how congestion control works. Dynamic dispatch is orthogonal to this.)

Ralith:
> In the typical use-case a user of quinn would only instantiate the parameter once.

The difference is in the experience of developing software downstream of quinn. Monomorphic code in quinn is compiled once; generic code must be, in large part, recompiled every single time the downstream code is itself recompiled.

> This PR removes 73 lines and adds 37 lines

Lines of code are a poor metric of complexity. The bulk of the code removed by this change is a single setter and its documentation. For a user who needs a specific congestion controller, invoking a setter is much simpler than implementing a trait. This is also a reduction in capability, as it replaces a per-connection configuration value with an endpoint-global type parameter.

> it does not add a generic parameter

As mentioned previously, it introduces further barriers to the removal of generic parameters.

> You can do this with this PR

You cannot, as there is no mechanism to direct the selection of congestion controller according to application preferences without sidechannel hacks. Resolving this would introduce further complexity.

infinity0:
(I understand you're not going to merge this PR, so I don't mean to be pushy & will understand if you choose to leave it, but I'm interested in the technical details of this discussion.)

> Monomorphic code in quinn is compiled once; generic code must be, in large part, recompiled every single time the downstream code is itself recompiled.

This is only an issue if 2 or more different downstream crates depend on quinn in the same program, which I also think is an unlikely scenario. (There's a good case to be made too, to allow more flexible compilation models than per-crate compilation in rustc's future, which would further improve this situation, but I'm not sure what their priorities are.)

> invoking a setter is much simpler than implementing a trait.

Typically, users will not have to implement traits (custom congestion algorithms) themselves, they would pick an already-implemented one, e.g. from a different crate. And this will have to be done regardless whether the choice itself is performed by the user via `Box <dyn>` or choosing what `<S>` to instantiate endpoint with. So you are comparing again, orthogonal things here - the direct comparison is rather between `Box <dyn>` vs reusing `<S>`.

>  [..] as there is no mechanism to direct the selection of congestion controller according to application preferences without sidechannel hacks. 

Similar to my above point, you would have to add these sidechannel hacks even without this PR, to the existing trait `X` in `Box <dyn X>` in order for quinn's own logic to pass the necessary information to the congestion controller. (Alternatively if you don't expose these sidechannel hacks, then the choice of controller must be hardcoded, i.e. not dynamically dispatched.) That is, this is again an orthogonal concern to what I changed in this PR.

infinity0:
>>  it does not add a generic parameter
>
> As mentioned previously, it introduces further barriers to the removal of generic parameters.

Yes, fair enough I think this is the main point of disagreement, which we discussed back in #911.

Ralith:
> I don't mean to be pushy & will understand if you choose to leave it, but I'm interested in the technical details of this discussion.

No worries, happy to provide clarity on my reasoning here!

> This is only an issue if 2 or more different downstream crates depend on quinn in the same program

That's not actually true. Consider a single crate that depends directly on quinn and nothing else. Every time you touch that crate and rebuild it, the generic code in quinn must be monomorphized and codegenned all over again. In theory incremental compilation can reduce this, but in practice it often fails to. Considering that LLVM's codegen phase is often one of the most severe bottlenecks for Rust compilation, this can have a major impact on quality of life.

> So you are comparing orthogonal things here - the direct comparison is rather between Box <dyn> vs reusing <S>.

That's only the case if the user wants to swap out the entire proposed monolithic `Session` trait to something off the shelf. Because the proposed trait is monolithic, any attempt to mix and match will require a new trait implementation, whereas decoupled trait objects can be swapped out individually.

PR #1022, opened 2021-02-17
Support non local ip binding

zz85:
Use socket options IP_FREEBIND to allow quinn binding to
interfaces with IP addresses that are nonlocal.

A use case is binding an ipv6 subnet that will be served by quinn.

### DIFF
--- a/quinn/src/platform/unix.rs
+++ b/quinn/src/platform/unix.rs
@@ -173,6 +173,20 @@ fn init(io: &mio::net::UdpSocket) -> io::Result<()> {
             if rc == -1 {
                 return Err(io::Error::last_os_error());
             }
+
+            let on: libc::c_int = 1;
+            let rc = unsafe {
+                libc::setsockopt(
+                    io.as_raw_fd(),
+                    libc::SOL_IP,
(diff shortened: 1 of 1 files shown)


Ralith:
I think for these type of advanced use cases it makes more sense for the application to configure a socket themselves and attach quinn to it with [`Endpoint::with_socket`](https://docs.rs/quinn/0.6.1/quinn/generic/struct.EndpointBuilder.html#method.with_socket)

zz85:
This could be viewed as an advanced use case or could be viewed as quinn having a set of socket options that work out of the box without surprises when an application uses quinn in such scenarios.

I'm fine either ways but I thought it might be a use case that other quinn users might find useful (just like how `IP_PKTINFO` is used to support binding on multiple ip addresses.

Ralith:
We have specific support for `IP_PKTINFO` because it is required for correctness. There are no correctness issues associated with users supplying a socket that they bound themselves.

zz85:
I was thinking about surprises on systems where `ip_nonlocal_bind` is set and isn't respected due to a kernel bug, I think you're right that common users using freebind may encounter a different behavior. Closing this PR.

PR #1024, opened 2021-02-22
Generic Send Offload (GSO) Support

Matthias247:
Generic Send Offload (GSO) Support
    
This change implements an initial version of GSO support [1][2][3]
for Linux, which improves the effiency of sending data.

The approach taken in this change is to create a buffer which contains
multiple datagrams in `Connection::poll_transmit`. This was picked
over trying to merge packets in the endpoint task, since packets
seem to need to be padded to a common segment size in order to
make them sendable via GSO.

In order to to this in an efficient fashion, the `poll_transmit`
method was restructred. Instead of selecting spaces upfront, it
will now loop through all possible packet spaces, check if there is
pending data to send and create packets and datagrams out of this.

The last packet which was written to a datagram buffer is not
finalized until it is clear whether follow-up packets need to be
written, since we need to know whether this packet should get padded
to MTU length or not.

This change doesn't enable GSO yet, since it will only produce a
single datagram. I will create a separate change for this.

Performance measurements:

**No GSO:**

```
Sent 1073741824 bytes on 1 streams in 4.31s (237.66 MiB/s)
```

**With GSO (up to 8 packets):**

```
Sent 1073741824 bytes on 1 streams in 3.02s (339.18 MiB/s)
```

[1] http://vger.kernel.org/lpc_net2018_talks/willemdebruijn-lpc2018-udpgso-paper-DRAFT-1.pdf
[2] http://vger.kernel.org/lpc_net2018_talks/willemdebruijn-lpc2018-udpgso-presentation-20181104.pdf
[3] https://lwn.net/Articles/752956/


Matthias247:
Same as https://github.com/quinn-rs/quinn/pull/953

Unfortunately I seemed to have closed the old version and the UI doesn''t let me open it again.

quinn-proto/src/connection/mod.rs
```diff
@@ -402,67 +405,160 @@ where
...
-                        window,
+                        self.path.congestion.window(),
                         now,
                     ) {
                         self.timers.set(Timer::Pacing, delay);
                         congestion_blocked = true;
-                        continue;
+                        // Loss probes should be subject to pacing, even though
+                        // they are not congestion controlled.
+                        break;
```
Ralith, line 527: Continuing prior discussion:
> I'm not really 100% convinced if [loss probes should be paced]

We don't block loss probes on congestion control because congestion control can become blocked on receiving ACKs from the peer, and the peer may not be aware that we're in that state. Pacing is guaranteed to relax shortly even if the peer takes no action, so it's safe and seems unlikely to compromise performance much, and improves odds of packets not getting dropped by a full buffers along the path.

quinn-proto/src/connection/mod.rs
```diff
@@ -389,76 +405,160 @@ where
...
-            self.side.is_client() && spaces.first() == Some(&SpaceId::Initial)
-                || self.path.challenge.is_some()
-                || self.path_response.is_some()
-        });
+        let mut buf_capacity = 0;
 
+        let mut coalesce = true;
+        let mut builder: Option<PacketBuilder> = None;
+        let mut sent_frames = None;
+        let mut min_datagram_size = 0;
```
Ralith, line 436: Could we replace this with `let mut pad_datagram = false`? It seems to only have two possible values, and this would let us avoid writing the nonzero case twice.

Ralith (changes requested):

quinn-proto/src/connection/mod.rs
Matthias247: It has 3 possible values if we take the next commit into account:
1. Don't pad
2. Pad to min initial size
3. Pad to MTU size

is it worth it to make it an enum?

Matthias247: And I agree, the calculations being spread out over a variety of places isn't great

Ralith: Where's option 3 arise? I only see two assignments to nonzero values, and both are exactly `min_datagram_size.max(builder.datagram_start + MIN_INITIAL_SIZE - builder.tag_len)`

Ralith: Oh, are you referring to a commit that hasn't been pushed yet? If so I'll take your word for it and leave as-is for now; an enum seems questionable.

Matthias247: Here's a proposal:

Have a
```
enum PadDatagram {
   DontPad,
   PadToMinInitialSize,
   PadToMtuSize
}
```

use here `let mut pad_datagram = PadDatagram::DontPad`

and add a `PacketBuilder::set_padding(&mut self, PadDatagram, mtu)` which gets called instead of manually fiddling with `builder.min_size`

Matthias247: > Oh, are you referring to a commit that hasn't been pushed yet? If so I'll take your word for it and leave as-is for now; an enum seems questionable.

Oh sorry, it's confusing. It is pushed as the last GSO commit. But there it doesn't set `min_datagram_size` and copies it over. We directly set `builder.min_size = builder.datagram_start + (self.path.mtu as usize) - builder.tag_len;` instead.

Ralith: Right, so in that case can we get away with a `bool` here?

Matthias247: Yes, I updated with a `bool` and 2 helper methods for `PacketBuilder`.
I now think the final padding instruction could be passed into `finish_and_track_packet`. But that would require the enum again.

Matthias247: So probably its just fine as is.

Ralith (approved):

djc:
Thanks for doing this work, really cool stuff!

djc:
Some follow-up tweaks in #1026.

PR #1027, opened 2021-02-23
Anti-amplification fixes

Ralith:


quinn-proto/src/connection/mod.rs
```diff
@@ -393,11 +393,6 @@ where
             }
         }
 
-        if self.path.anti_amplification_blocked(self.path.mtu.into()) {
```
Matthias247, line 396: What is the reason for removing it here? I mean it definitely is also checked later on and therefore not necessary. But apart from that, will the short circuit prevent here anything?

quinn-proto/src/connection/mod.rs
```diff
@@ -931,6 +926,9 @@ where
                     return;
                 }
 
+                let was_anti_amplification_blocked =
```
Matthias247, line 929: I'm wondering if this can lead to a similar race as with `congestion_blocked`. This could be set to `true`, even though the application didn't actually try to send more data (actually it will be common during the handshake that this happens since most cert chains exactly will fill this window).

Is it ok for this case to start the timer?

quinn-proto/src/connection/mod.rs
Ralith: Just trying to simplify one of our most complex functions. There is a slight change in behavior in that `ensure_probe_queued` might run earlier, but that should be harmless, and could be addressed by moving it into the loop below in any case if we decide to care.

quinn-proto/src/connection/mod.rs
Ralith: > most cert chains exactly will fill this window

How do you judge that?

> Is it ok for this case to start the timer?

I believe `set_loss_detection_timer` will do the right thing in any state; its job is to inspect the current state and set the timer as appropriate, rather than just to start the timer. For example, we currently call it every time we send a packet or receive an ACK. Additionally, spurious probes aren't the end of the world if we do end up with one.

The logic involved is pretty subtle, though, so I welcome further analysis if you think there could be a weird corner case. We do need to fix the underlying issue, but maybe there's a more principled way? Perhaps we should instead remove the anti-amplification check from `set_loss_detection_timer` and rely on `poll_transmit` to handle that case?

Matthias247: > How do you judge that?

That's just derived from a production sample set of certificates. So far the learning is "3 packets with 1230 byte initial packet" will be not enough for a certain set of cert chains. If the initial packet is sent with 1350 bytes (chrome does this) a few more cert chains barely fit, but others still require 4 packets. I can not share concrete numbers. 
Cert compression is a good tool to get back to 3 packets.

=> Overall there are definitely a lot of handshakes going on where the boundary is exactly between 3 and 4 packets.

djc (approved):

quinn-proto/src/connection/mod.rs
Ralith: Relevant text: https://quicwg.org/base-drafts/draft-ietf-quic-recovery.html#name-on-receiving-a-datagram

> When a server is blocked by anti-amplification limits, receiving a datagram unblocks it, even if none of the packets in the datagram are successfully processed. In such a case, the PTO timer will need to be re-armed.

PR #1028, opened 2021-02-23
Perf protocol implementation

Ralith:
For https://github.com/quicwg/base-drafts/wiki/Performance-Hackathon

### DIFF
--- a/perf/Cargo.toml
+++ b/perf/Cargo.toml
@@ -1,0 +1,28 @@
+[package]
+name = "perf"
+version = "0.1.0"
+authors = ["Benjamin Saunders <ben.e.saunders@gmail.com>"]
+edition = "2018"
+license = "MIT OR Apache-2.0"
+publish = false
+
+[dependencies]
+anyhow = "1.0.22"
+futures = "0.3.8"
+quinn = { path = "../quinn" }
+rcgen = "0.8"
+rustls = { version = "0.19", features = ["dangerous_configuration"] }
+webpki = "0.21"
+structopt = "0.3"
+tokio = { version = "1.0.1", features = ["rt", "macros", "signal", "net", "sync"] }
+tracing = "0.1.10"
+tracing-subscriber = { version = "0.2.5", default-features = false, features = ["env-filter", "fmt", "ansi", "chrono"]}
+bytes = "1"
+
+[[bin]]
+name = "server"
+path = "src/server.rs"
+
+[[bin]]
+name = "client"
+path = "src/client.rs"
--- a/perf/src/client.rs
+++ b/perf/src/client.rs
@@ -1,0 +1,225 @@
+use std::{
+    net::{IpAddr, Ipv6Addr, SocketAddr},
+    sync::Arc,
+};
+
+use anyhow::{Context, Result};
+use bytes::Bytes;
+use futures::StreamExt;
+use structopt::StructOpt;
+use tokio::sync::Semaphore;
+use tracing::{error, info, trace};
+
+/// Connects to a QUIC perf server and maintains a specified pattern of requests until interrupted
+#[derive(StructOpt)]
+#[structopt(name = "client")]
+struct Opt {
+    /// Host to connect to
+    #[structopt(default_value = "localhost:4433")]
+    host: String,
+    /// Override DNS resolution for host
+    #[structopt(long)]
+    ip: Option<IpAddr>,
+    /// Number of unidirectional requests to maintain concurrently
+    #[structopt(long, default_value = "0")]
+    uni_requests: u64,
+    /// Number of bidirectional requests to maintain concurrently
+    #[structopt(long, default_value = "1")]
+    bi_requests: u64,
+    /// Number of bytes to request
+    #[structopt(long, default_value = "1048576")]
+    download_size: u64,
+    /// Number of bytes to transmit, in addition to the request header
+    #[structopt(long, default_value = "1048576")]
+    upload_size: u64,
+    /// Whether to skip certificate validation
+    #[structopt(long)]
+    insecure: bool,
+}
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() {
+    let opt = Opt::from_args();
+
+    tracing_subscriber::fmt::init();
+
+    if let Err(e) = run(opt).await {
+        error!("{:#}", e);
+    }
+}
+
+async fn run(opt: Opt) -> Result<()> {
+    let mut host_parts = opt.host.split(':');
+    let host_name = host_parts.next().unwrap();
+    let host_port = host_parts
+        .next()
+        .map_or(Ok(443), |x| x.parse())
+        .context("parsing port")?;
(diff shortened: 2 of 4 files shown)


perf/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+publish = false
+
+[dependencies]
+anyhow = "1.0.22"
+futures = "0.3.8"
+quinn = { path = "../quinn" }
+rcgen = "0.8"
+rustls = "0.19"
+structopt = "0.3"
+tokio = { version = "1.0.1", features = ["rt-multi-thread", "macros"] }
```
Matthias247, line 16: I think this should use a single-threaded runtime given that `bulk` shows very clearly that multithreaded performs worse

perf/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+    let key = quinn::PrivateKey::from_der(&cert.serialize_private_key_der()).unwrap();
+    let cert = quinn::Certificate::from_der(&cert.serialize_der().unwrap()).unwrap();
+
+    let mut server_config = quinn::ServerConfigBuilder::default();
+    server_config
+        .certificate(quinn::CertificateChain::from_certs(vec![cert.clone()]), key)
+        .unwrap();
+    server_config.protocols(&[b"perf"]);
+
+    let server_config = server_config.build();
```
Matthias247, line 21: I guess this is ok for starting the thing. However I think this should later adjust settings for increasing perf. E.g. flow control limits might be raised. Limits should probably equal to what `bulk` does.

perf/src/main.rs
```diff
@@ -0,0 +1,118 @@
+use std::net::{IpAddr, Ipv6Addr, SocketAddr};
+
+use anyhow::{Context, Result};
+use futures::StreamExt;
+use tracing::{debug, error, info};
+
+#[tokio::main]
+async fn main() {
```
Matthias247, line 8: This always starts the server. I think quinn should also contain a client. Should this be an extra executable? Or maybe make it possible to configure via CLI (`perf --client`, `perf --listen ip`) which peer is used.

perf/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+
+async fn drive_uni(
+    connection: quinn::Connection,
+    mut streams: quinn::IncomingUniStreams,
+) -> Result<()> {
+    while let Some(stream) = streams.next().await {
+        let stream = stream?;
+        let connection = connection.clone();
+        tokio::spawn(async move {
+            if let Err(e) = accept_uni(connection, stream).await {
```
Matthias247, line 64: I would rather call it `handle_uni`. `accept` sounds like it would wait for a stream. However that already had happened. Same for `accept_bi`

perf/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+async fn read_req(mut stream: quinn::RecvStream) -> Result<u64> {
+    let mut buf = [0; 8];
+    stream
+        .read_exact(&mut buf)
+        .await
+        .context("reading request")?;
+    let n = u64::from_be_bytes(buf);
+    debug!("got req for {} bytes", n);
```
Matthias247, line 104: This actually misses draining the remaining request body, which is important for upload tests.

I would recommend to copy&paste https://github.com/quinn-rs/quinn/blob/main/bench/src/bulk.rs#L88-L101.
Optionally a `async fn drain()` which does the same could also be added to `RecvStream`.

perf/Cargo.toml
Ralith: I was thinking about the many-connection case, but sure, might as well play to our strengths.

perf/src/client.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    let (endpoint, _) = endpoint
+        .bind(&SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 0))
+        .context("binding endpoint")?;
+
+    let mut cfg = quinn::ClientConfigBuilder::default();
+    cfg.protocols(&[b"perf"]);
+    let mut cfg = cfg.build();
+    let tls_cfg: &mut rustls::ClientConfig = Arc::get_mut(&mut cfg.crypto).unwrap();
+    tls_cfg
```
Matthias247, line 62: Maybe move this behind an `--insecure` flag? There can definitely be use cases to run this against places with a valid cert. For those use-cases it might also be nice to either
- specify the ip independent from the hostname (similar to how `curl --resolve` works)
- allow to overwrite the SNI

I would recommend the first option. But this is also fine for a follow-up, and i'm happy to add it.

perf/src/client.rs
```diff
@@ -0,0 +1,183 @@
...
+        tokio::spawn(async move {
+            if let Err(e) = request_uni(send, size).await {
+                error!("sending request failed: {:#}", e);
+            }
+            drop(permit);
+        });
+    }
+}
+
+async fn request_uni(mut send: quinn::SendStream, size: u64) -> Result<()> {
```
Matthias247, line 136: This as well as `request_bi` will require another parameter to specify how much data to upload (and an implementation of it).

perf/src/server.rs
```diff
@@ -0,0 +1,160 @@
...
+                error!("request failed: {:#}", e);
+            }
+        });
+    }
+    Ok(())
+}
+
+async fn handle_bi(send: quinn::SendStream, recv: quinn::RecvStream) -> Result<()> {
+    let bytes = read_req(recv).await?;
+    respond(bytes, send).await?;
```
Matthias247, line 131: draining of the request stream is still missing.  What I'm not sure is whether this actually should happen concurrently to sending a request. I implemented it now somewhere else as:

```
let bytes = read_header(&mut recv).await?;

let drain_request_fut = drain_data(recv);
let send_response_fut = send_response(bytes, send);

let ((), ()) = tokio::try_join!(drain_request_fut, send_response_fut)?;

Ok(())
```

perf/src/client.rs
```diff
@@ -0,0 +1,183 @@
...
+        x = drive_bi(connection.clone(), opt.bi_requests, opt.request_size) => x?,
+        _ = tokio::signal::ctrl_c() => {
+            info!("shutting down");
+            connection.close(0u32.into(), b"interrupted");
+        }
+    }
+
+    endpoint.wait_idle().await;
+
+    // TODO: Print stats
```
Matthias247, line 98: Yes, we will need those to make it useful. Probably similar to what exists in `bulk`

perf/src/server.rs
Matthias247: Reading through the update https://github.com/nibanks/quic-perf/pull/2/files , concurrent sending of the response doesn't seem required and the `try_join!` can be dropped.

Ralith:
> Should we open an issue for the QUINN BUG?

Whoops, that was fixed in #1032. Removed comment

> I don't think keeping these commits separate is super useful

Squashed.

djc (approved):

PR #1029, opened 2021-02-24
Assorted fixes

Ralith:
Issues encountered in the course of #1028. I also discovered an interesting corner case where receive streams that never experience a blocked read can be leaked if the peer finishes them before we stop them, but I still need to investigate whether #1003 fixes that already.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -865,13 +865,13 @@ where
         let ack_eliciting = builder.ack_eliciting;
         let exact_number = builder.exact_number;
         let space_id = builder.space;
-        let (size, _) = self.finish_packet(builder, buffer);
+        let (size, padded) = self.finish_packet(builder, buffer);
         let sent = match sent {
             Some(sent) => sent,
             None => return,
         };
 
-        let size = match sent.padding || ack_eliciting {
+        let size = match padded || ack_eliciting {
             true => size as u16,
             false => 0,
         };
@@ -3613,7 +3613,6 @@ struct SentFrames {
     retransmits: ThinRetransmits,
     acks: RangeSet,
     stream_frames: StreamMetaVec,
-    padding: bool,
     requires_padding: bool,
 }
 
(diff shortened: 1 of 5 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -634,14 +632,16 @@ where
                 pad_datagram |= sent.requires_padding;
             }
 
+            pad_datagram |= space_id == SpaceId::Initial && self.side.is_client();
```
Matthias247, line 635: Oh, this is definitely good too. But which case had been missed before that?  All non close frames should already have been handled by setting `requires-padding` in `populate_packet`

quinn-proto/src/connection/mod.rs
```diff
@@ -865,13 +865,13 @@ where
...
         let space_id = builder.space;
-        let (size, _) = self.finish_packet(builder, buffer);
+        let (size, padded) = self.finish_packet(builder, buffer);
         let sent = match sent {
             Some(sent) => sent,
             None => return,
         };
 
-        let size = match sent.padding || ack_eliciting {
+        let size = match padded || ack_eliciting {
```
Matthias247, line 874: Good catch! `sent.padding` doesn't capture the cases where we pad for close frames and some other edge-cases

quinn-proto/src/connection/mod.rs
```diff
@@ -3631,12 +3626,10 @@ struct PacketBuilder {
 }
 
 impl PacketBuilder {
-    fn pad_to_min_initial_size(&mut self) {
-        self.min_size = self.datagram_start + MIN_INITIAL_SIZE - self.tag_len;
-    }
-
-    fn pad_to_mtu_size(&mut self, mtu: u16) {
-        self.min_size = self.datagram_start + (mtu as usize) - self.tag_len;
+    fn pad_to(&mut self, min_size: u16) {
```
Matthias247, line 3629: 👍🏻

quinn-proto/src/connection/mod.rs
Ralith: I think close packets were the only incorrect case.

Matthias247: For which case? The close case should have been handled by the `pad_datagram = true` in line 616 which is removed here. However I'm wondering if some other case is missing and fixed by this, since the unit-tests still don't work with GSO support, even after #1030 is added.

Ralith: Setting `pad_datagram = true` unconditionally above caused *all* close packets to be padded, which is incorrect.

djc (approved):

PR #1030, opened 2021-02-24
Make TestEndpoint GSO compatible

Matthias247:
The TestEndpoint so far tried to send a single big datagram instead of using
proper GSO sending to emit multiple datagrams.

In order to avoid relying on platform GSO support, this change will simply
split up GSO transmissions into multiple Transmits in the endpoint.

### DIFF
--- a/quinn-proto/src/tests/util.rs
+++ b/quinn-proto/src/tests/util.rs
@@ -248,7 +248,7 @@ impl TestEndpoint {
         }
 
         while let Some(x) = self.poll_transmit() {
-            self.outbound.push_back(x);
+            self.outbound.extend(split_transmit(x));
         }
 
         let mut endpoint_events: Vec<(ConnectionHandle, EndpointEvent)> = vec![];
@@ -269,7 +269,7 @@ impl TestEndpoint {
             }
 
             while let Some(x) = conn.poll_transmit(now) {
-                self.outbound.push_back(x);
+                self.outbound.extend(split_transmit(x));
             }
             self.timeout = conn.poll_timeout();
         }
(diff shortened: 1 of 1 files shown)


quinn-proto/src/tests/util.rs
```diff
@@ -385,6 +389,32 @@ pub fn min_opt<T: Ord>(x: Option<T>, y: Option<T>) -> Option<T> {
...
+        let contents = transmit.contents[offset..end].to_vec();
+        transmits.push(Transmit {
+            destination: transmit.destination,
+            ecn: transmit.ecn,
+            contents,
+            segment_size: None,
+            src_ip: transmit.src_ip,
+        });
+
+        offset += segment_size;
```
Ralith, line 412: nit:
```suggestion
        offset = end;
```

quinn-proto/src/tests/util.rs
```diff
@@ -248,7 +248,9 @@ impl TestEndpoint {
         }
 
         while let Some(x) = self.poll_transmit() {
-            self.outbound.push_back(x);
+            for transmit in split_transmit(x) {
+                self.outbound.push_back(transmit);
+            }
```
Ralith, line 253:
```suggestion
            self.outbound.extend(split_transmit(x));
```
and below

Ralith (commented):
Maybe this is a good time to make `MAX_DATAGRAMS` a parameter, so this new code isn't dead?

Matthias247:
> Maybe this is a good time to make `MAX_DATAGRAMS` a parameter, so this new code isn't dead?

Indeed. But I think there might be more to discuss around how this would exactly look like, and I decided to try to land this simple change upfront.

Ralith (approved):

PR #1032, opened 2021-02-24
Debug STOP_SENDING corner cases

Ralith:


quinn-proto/src/connection/streams/recv.rs
```diff
@@ -180,7 +180,7 @@ impl Recv {
     }
 
     /// No more data expected from peer
-    pub(super) fn is_finished(&self) -> bool {
+    pub(super) fn is_terminal(&self) -> bool {
```
djc, line 187: Bikeshedding a little: the `terminal` term isn't that obvious to me. `is_closing()`? `no_more_data_expected()`? `is_receiving()` (with inverted semantics)?

Ralith: Went with the last.

djc (approved):

PR #1034, opened 2021-02-24
Connection refactoring

djc:


quinn-proto/src/connection/mod.rs
```diff
@@ -432,21 +432,15 @@ where
         let mut congestion_blocked = false;
 
         // Iterate over all spaces and find data to send
-        let mut space_idx = 0;
-        let spaces = [SpaceId::Initial, SpaceId::Handshake, SpaceId::Data];
-        while space_idx < spaces.len() {
-            let space_id = spaces[space_idx];
-
+        for space_id in SpaceId::iter() {
```
Matthias247, line 456: This was explicitly changed for GSO, and changing it back won't work. There's a comment at the end of the function why.

djc: Ah, oops. The comment wasn't very discoverable from the point of where the iteration happens, so I replaced my previous commit with a new one that explains this closer to the loop setup (8fa5c32).

Ralith:
There appear to be two commits with the description "quinn-proto: move datagram receive logic into DatagramState" which do different things. Disambiguate? Looks good otherwise.

djc:
Added a single character to disambiguate `receive` logic from `received` logic -- is that enough? I couldn't really come up with a better description for the second commit. Also rebased on current main and tacked on a final commit that just moves all of the `bytes_source` module's contents into `connection::streams::send`.

djc:
I've updated the "extract separate *Stream" commits to add `send_stream()`/`recv_stream()` accessors directly on `Connection` rather than going through `Streams`.

Ralith:
> I couldn't really come up with a better description for the second commit

Maybe "incoming datagram frame handling" instead of "datagram received logic"?

djc:
> Maybe "incoming datagram frame handling" instead of "datagram received logic"?

Done.

Ralith (approved):

PR #1036, opened 2021-03-01
Make owned bytes write APIs available in quinn

Matthias247:
This extends the public quinn API to offer support for writing owned buffers.
Besides the universal `write_chunks` API which supports a variable amount
of buffers the API also offers convenience methods for writing either a
single buffer completely or an arbitrary amount of buffers completely.

### DIFF
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -5,8 +5,9 @@ use std::{
     task::{Context, Poll},
 };
 
+use bytes::Bytes;
 use futures::{channel::oneshot, io::AsyncWrite, ready, FutureExt};
-use proto::{ConnectionError, FinishError, StreamId};
+use proto::{ConnectionError, FinishError, StreamId, Written};
 use thiserror::Error;
 
 use crate::{connection::ConnectionRef, recv_stream::UnknownStream, VarInt};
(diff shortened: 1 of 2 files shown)


quinn/src/send_stream.rs
```diff
@@ -54,7 +55,50 @@ where
...
+    }
+
+    fn poll_write_chunks(
+        &mut self,
+        cx: &mut Context,
+        bufs: &mut [Bytes],
+    ) -> Poll<Result<Written, WriteError>> {
+        let stream_id = self.stream;
+        self.execute_poll(cx, |conn| conn.write_chunks(stream_id, bufs))
+    }
```
Ralith, line 96: Would it be more concise to inline these now-trivial methods into the respective `Future::poll` impls?

quinn/src/send_stream.rs
```diff
@@ -54,7 +55,50 @@ where
...
+    /// Write chunks to the stream
+    ///
+    /// Yields the number of bytes and chunks written on success.
+    /// Congestion and flow control may cause this to be shorter than `buf.len()`,
+    /// indicating that only a prefix of `bufs` was written
+    pub fn write_chunks<'a>(&'a mut self, bufs: &'a mut [Bytes]) -> WriteChunks<'a, S> {
+        WriteChunks { stream: self, bufs }
+    }
+
+    /// Convenience method to write a single chunk to the stream
```
Ralith, line 67:
```suggestion
    /// Convenience method to write a single chunk in its entirety to the stream
```

Ralith (changes requested):

quinn/src/send_stream.rs
Matthias247: It has it's pro's and con's. Of course they are unnecessary, but they might make it easier to read the poll methods since the readers of those now don't have to understand how the closure thingy works. I don't have any strong opinion.

Matthias247: I updated it now, since the connection refactoring which made `proto::SendStream` public made the closure a bit easier to understand.

Ralith (approved):

PR #1038, opened 2021-03-02
Perf client and server improvements

Matthias247:
This change aggregates a couple of improvements for the perf client and server.
Some changes make the client and server more efficient.

Two other commits add stats to the client.

Output is now created in 2s intervals and looks like:

```
Overall stats:
RPS: 0 (2 requests in 14.01s)
Success rate: 100.00%

Stream metrics:

      │ Duration  │ FBL       | Upload Throughput | Download Throughput
──────┼───────────┼───────────┼───────────────────┼────────────────────
 AVG  │     4.96s │    0.00ns │      404.00 MiB/s │        402.62 MiB/s
 P0   │     4.94s │    0.00ns │      400.50 MiB/s │        397.00 MiB/s
 P10  │     4.95s │    0.00ns │      400.75 MiB/s │        397.25 MiB/s
 P50  │     4.95s │    0.00ns │      400.75 MiB/s │        397.25 MiB/s
 P90  │     4.97s │    1.00ms │      407.50 MiB/s │        408.25 MiB/s
 P100 │     4.97s │    1.00ms │      407.50 MiB/s │        408.25 MiB/s
```

### DIFF
--- a/perf/src/client.rs
+++ b/perf/src/client.rs
@@ -84,7 +84,7 @@ async fn run(opt: Opt) -> Result<()> {
 
     let quinn::NewConnection {
         connection,
-        mut uni_streams,
+        uni_streams,
         ..
     } = endpoint
         .connect_with(cfg, &addr, &host_name)?
@@ -93,20 +93,23 @@ async fn run(opt: Opt) -> Result<()> {
 
     info!("established");
 
-    tokio::spawn(async move {
-        while let Some(Ok(stream)) = uni_streams.next().await {
-            tokio::spawn(async move {
-                if let Err(e) = drain_stream(stream).await {
-                    error!("reading response failed: {:#}", e);
-                }
-            });
-        }
-    });
+    let acceptor = UniAcceptor(Arc::new(tokio::sync::Mutex::new(uni_streams)));
 
     let drive_fut = async {
         tokio::try_join!(
-            drive_uni(connection.clone(), opt.uni_requests, opt.upload_size, opt.download_size),
-            drive_bi(connection.clone(), opt.bi_requests, opt.upload_size, opt.download_size)
+            drive_uni(
+                connection.clone(),
+                acceptor,
+                opt.uni_requests,
+                opt.upload_size,
+                opt.download_size
+            ),
+            drive_bi(
+                connection.clone(),
+                opt.bi_requests,
+                opt.upload_size,
+                opt.download_size
+            )
         )
     };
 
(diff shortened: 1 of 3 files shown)


perf/src/client.rs
```diff
@@ -93,23 +112,49 @@ async fn run(opt: Opt) -> Result<()> {
...
-        x = drive_bi(connection.clone(), opt.bi_requests, opt.upload_size, opt.download_size) => x?,
+        _ = drive_fut => {}
+        _ = print_fut => {}
         _ = tokio::signal::ctrl_c() => {
             info!("shutting down");
             connection.close(0u32.into(), b"interrupted");
         }
+        _ = tokio::time::sleep(Duration::from_secs(opt.duration)) => {
+            info!("shutting down");
+            connection.close(0u32.into(), b"interrupted");
```
Ralith, line 156:
```suggestion
            connection.close(0u32.into(), b"done");
```

perf/src/client.rs
```diff
@@ -103,9 +103,15 @@ async fn run(opt: Opt) -> Result<()> {
...
+        tokio::try_join!(
+            drive_uni(connection.clone(), opt.uni_requests, opt.upload_size, opt.download_size),
+            drive_bi(connection.clone(), opt.bi_requests, opt.upload_size, opt.download_size)
+        )
+    };
+
     tokio::select! {
-        x = drive_uni(connection.clone(), opt.uni_requests, opt.upload_size, opt.download_size) => x?,
-        x = drive_bi(connection.clone(), opt.bi_requests, opt.upload_size, opt.download_size) => x?,
+        _ = drive_fut => {}
```
Ralith, line 148: These tasks only terminate in the event of an error, so this shouldn't actually make a difference.

perf/src/server.rs
```diff
@@ -85,10 +85,7 @@ async fn handle(handshake: quinn::Connecting) -> Result<()> {
         ..
     } = handshake.await.context("handshake failed")?;
     debug!("{} connected", connection.remote_address());
-    tokio::select! {
-        x = drive_uni(connection, uni_streams) => x?,
-        x = drive_bi(bi_streams) => x?,
-    }
+    tokio::try_join!(drive_uni(connection, uni_streams), drive_bi(bi_streams))?;
```
Ralith, line 101: As above.

Ralith (approved):

perf/src/client.rs
Matthias247: yes, I changed that before I noticed that they run endlessly. My assumption was that we would do a certain amount of requests on each connection. And in this case it's correct to wait for requests on both uni and bidi streams to finish.

I think it's good to keep this change, since it might also be helpful if a non-endless running mode is added. E.g. measuring the amount of possible handshakes requires to wait for connections to end.

Ralith (approved):

PR #1040, opened 2021-03-02
Reduce stream wakeups

Matthias247:
This contains 2 changes which will report a Stream only as writeable if both the connection as well as stream windows are not empty.

### DIFF
--- a/quinn-proto/src/connection/streams/state.rs
+++ b/quinn-proto/src/connection/streams/state.rs
@@ -50,7 +50,7 @@ pub struct StreamsState {
     /// Streams are only added to this list when a write fails.
     pub(super) connection_blocked: Vec<StreamId>,
     /// Connection-level flow control budget dictated by the peer
-    pub(super) max_data: u64,
+    max_data: u64,
     /// The initial receive window
     receive_window: u64,
     /// Limit on incoming data, which is transmitted through `MAX_DATA` frames
@@ -603,9 +603,18 @@ impl StreamsState {
             ));
         }
 
+        let write_limit = self.write_limit();
         if let Some(ss) = self.send.get_mut(&id) {
             if ss.increase_max_data(offset) {
-                self.events.push_back(StreamEvent::Writable { id });
+                if write_limit > 0 {
+                    self.events.push_back(StreamEvent::Writable { id });
+                } else if !ss.connection_blocked {
+                    // The stream is still blocked on the connection flow control
+                    // window. In order to get unblocked when the window relaxes
+                    // it needs to be in the connection blocked list.
+                    ss.connection_blocked = true;
+                    self.connection_blocked.push(id);
+                }
             }
         } else if id.initiator() == self.side && self.is_local_unopened(id) {
             debug!("got MAX_STREAM_DATA on unopened {}", id);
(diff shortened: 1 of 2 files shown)


Ralith (changes requested):
I think this loses a wakeup under the following sequence of events:

- A write is attempted, and blocks on both stream-level and connection-level flow control. Stream is recorded in the `connection_blocked` list.
- Connection-level flow control relaxes. Stream is removed from `connection_blocked`, but a wakeup is not issued due to stream-level flow control.
- Activity on other streams causes connection-level flow control to become blocked again.
- Stream-level flow control is relaxed, but a wakeup is not issued due to connection-level flow control
- Connection-level flow control is relaxed. At this point, a wakeup should be issued, but the stream is no longer in the `connection_blocked` list so we don't know about it.

I think the fix for this particular case is to ensure that the stream is in `connection_blocked` when stream-level flow control is newly relaxed but connection-level flow control is blocking.

Ralith:
Reduced sequence:

1. Stream is blocked at stream level
2. Stream unblocks at stream level while blocked at connection level
3. Connection level unblocks

We could use better test coverage here, that's for sure.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -50,7 +50,7 @@ pub struct StreamsState {
     /// Streams are only added to this list when a write fails.
     pub(super) connection_blocked: Vec<StreamId>,
     /// Connection-level flow control budget dictated by the peer
-    pub(super) max_data: u64,
+    max_data: u64,
```
djc, line 53: This doesn't appear to be necessary (anymore)?

Ralith: What doesn't, exactly?

Matthias247:
Oh, you are absolutely right. I didn't recognize the side effect that independent streams can have here.

quinn-proto/src/connection/streams/state.rs
Matthias247: `pub(super)` was removed because the variable is no longer accessed from the connection and replaced by the `write_limit` function. Not exactly sure what the comment is about otherwise.

Matthias247:
Updated with the suggested solution. In case you think it's not worth the extra complexity I guess it's also fine to drop it. The chances of being blocked on both windows might not be extremely high with well-configured peers.

Although I kind of like the part `write_limit()` function which clears up a bit of the math that is spread out across modules.

quinn-proto/src/connection/streams/state.rs
djc: Ah, sorry, yes, that is great!

quinn-proto/src/connection/streams/state.rs
```diff
@@ -603,9 +603,20 @@ impl StreamsState {
...
+                    self.events.push_back(StreamEvent::Writable { id });
+                } else {
+                    // The stream is still blocked on the connection flow control
+                    // window. In order to get unblocked when the window relaxes
+                    // it needs to be in the connection blocked list.
+                    if !ss.connection_blocked {
+                        ss.connection_blocked = true;
+                        self.connection_blocked.push(id);
+                    }
+                }
```
djc, line 619:
```suggestion
                if write_limit > 0 {
                    self.events.push_back(StreamEvent::Writable { id });
                } else if !ss.connection_blocked {
                    // The stream is still blocked on the connection flow control
                    // window. In order to get unblocked when the window relaxes
                    // it needs to be in the connection blocked list.
                    ss.connection_blocked = true;
                    self.connection_blocked.push(id);
                }
```

djc (approved):

PR #1041, opened 2021-03-02
retry transmitting close frames

Matthias247:
This changes makes sure we don't lose close frames when the first send attempt
is blocked by congestion control.


Ralith (approved):
Whoops, forgot about this!

Matthias247:
The CI failure is unrelated to this change. There seems to be a flaky test in the h3 area

Ralith:
Flakiness confirmed by rerun. At a guess, the h3 test in question (`client_send_body`) or underlying infra is sensitive to races between reporting successful completion and the the connection becoming closed by the peer.

PR #1042, opened 2021-03-02
Perf: Use owned buffers

Matthias247:
Use zero-copy APIs when sending data in perf application.

Performance difference:
1. First line is before
2. Second line is after (but with only the download write change)

```
      │ Duration  │ FBL       | Upload Throughput | Download Throughput
──────┼───────────┼───────────┼───────────────────┼────────────────────
 AVG  │    5.00ms │    1.00ms │      332.82 MiB/s │        334.20 MiB/s
 AVG  │    5.00ms │    0.00ns │      332.64 MiB/s │        335.96 MiB/s
```

=> Makes less than 1%. Still neat to have

### DIFF
--- a/perf/src/client.rs
+++ b/perf/src/client.rs
@@ -255,13 +255,14 @@ async fn request(
 ) -> Result<()> {
     stats.upload_start = Some(Instant::now());
     send.write_all(&download.to_be_bytes()).await?;
+
     const DATA: [u8; 1024 * 1024] = [42; 1024 * 1024];
     while upload > 0 {
-        let n = send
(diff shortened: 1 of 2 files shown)


Ralith (approved):
Should we replace the `const`s with a single `Bytes` that gets cloned to avoid some more allocs/copying? I guess it's big enough now that the difference might not register.

Matthias247:
> Should we replace the `const`s with a single `Bytes` that gets cloned to avoid some more allocs/copying? I guess it's big enough now that the difference might not register.

`Bytes::from_static` has a vtable which doesn't have any copies or allocations

Ralith:
Oh, sweet, didn't realize!

PR #1043, opened 2021-03-02
Re-export tokio

### LINKED ISSUE #1016: Re-export tokio
Currently, users have to depend on `tokio` as well as this crate in order to use this crate. Re-exporting `tokio` would allow users to just depend on this crate.

nashley:
Fixes #1016

Downstream consumers (unless they explicitly depend on `tokio`) will reference this via `use quinn::tokio::etc::etc;` instead of just `use tokio::etc::etc;`.

Code change from user's perspective is simple and opt-in:
```diff
- use tokio::etc::etc;
+ use quinn::tokio::etc::etc;
- let _ = tokio::etc::etc();
+ let _ = quinn::tokio::etc::etc();
```

Users will no longer need to add `tokio` to the `dependencies` section of their `Cargo.toml`.
If they want to enable tokio-specific flags, they'll need to enable the corresponding `tokio-*` features for `quinn` alongside `quinn`'s normal feature flags.

`Cargo.toml` changes:
```diff
-[dependencies.tokio]
-version = "1.2"
-features = [ "rt-multi-thread", "sync", "time", "macros", "io-std", "io-util" ]
...
[dependencies.quinn]
-features = [ "tls-rustls" ]
+features = [ "tls-rustls", "tokio-rt-multi-thread", "tokio-sync", "tokio-time", "tokio-macros", "tokio-io-std", "tokio-io-util" ]
```

If a library wishes to expose `tokio` as given to it by `quinn` to its own users, it would need to include `pub extern crate tokio;` in its own `src/lib.rs` and expose the `tokio`-specific features given to it by `quinn` (much like this PR).
`Cargo.toml` changes for exposing these features further down the dependency chain:
```diff
[features]
...
+#Re-export tokio features:
+tokio-full = ["quinn/tokio-full"]
+tokio-fs = ["quinn/tokio-fs"]
+tokio-io-util = ["quinn/tokio-io-util"]
+tokio-io-std = ["quinn/tokio-io-std"]
+tokio-macros = ["quinn/tokio-macros"]
+tokio-net = ["quinn/tokio-net"]
+tokio-process = ["quinn/tokio-process"]
+tokio-rt = ["quinn/tokio-rt"]
+tokio-rt-multi-thread = ["quinn/tokio-rt-multi-thread"]
+tokio-signal = ["quinn/tokio-signal"]
+tokio-sync = ["quinn/tokio-sync"]
+tokio-test-util = ["quinn/tokio-test-util"]
+tokio-time = ["quinn/tokio-time"]
```


nashley:
Adopting this will mean needing to update this crate's feature flags whenever `tokio` changes theirs. I don't think that's a terrible ask, but it's not ideal.

djc:
I'm not sure what others think, but personally I was thinking of this more in a take-it-or-leave-it capacity. Downstream crates can use tokio as configured by quinn, but I definitely don't want them to "configure" tokio features through quinn.

Ralith:
Yeah, let's not reexport features.

djc:
@nashley just to make sure, are you aware that Cargo will enforce that crates in the depgraph which rely on semver-compatible versions of a particular other crate will automatically pull in the crate? That is, if your project depends on quinn (which depends on tokio ^1.0.1) and on tokio itself (let's say ^1.2.0 -- and note that the `^`operator is usually implicit!), your project will get a single version of crate, in this case just tokio 1.2.0 (which is of course compatible with 1.0.1).

Which is to say, reading to your downstream issues/threads, I'm not sure this work is all that necessary?

djc:
@nashley it looks like the downstream issue for this has been resolved, do you still think this is useful?

nashley:
Sorry for the confusion regarding the `~` version specification. I was playing around with a plugin that automatically updates cargo dependencies; it was not intentional.

I'm no longer sure that this is useful, so I'll go ahead and close it. Thanks for your feedback and time!

PR #1047, opened 2021-03-03
Remove obsolete safety asserts

Ralith:
These will cause spurious panics once https://github.com/rust-lang/rust/pull/78802 is merged.

CC @est31: this will need to be added to the backport PRs.


djc (approved):

est31:
@Ralith thanks for the notification. I've seen your comment in the other thread, they are removed from the backports already :).

djc (approved):

PR #1048, opened 2021-03-03
Pad initial-space datagrams from servers, too

Ralith:
Quoth the spec: a server MUST expand the payload of all UDP datagrams
carrying ack-eliciting Initial packets to at least the smallest
allowed maximum datagram size of 1200 bytes.

Found in @Matthias247's recent interop testing

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -636,7 +636,9 @@ where
             )?);
             coalesce = coalesce && !builder.short_header;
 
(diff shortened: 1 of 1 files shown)


djc (approved):

Matthias247:
I think this can be slightly relaxed, based on 

> **ack-eliciting** Initial packets

`ack_eliciting` is known here

Matthias247:
Also consider adding a link to https://tools.ietf.org/html/draft-ietf-quic-transport-34#section-14.1

Ralith:
Suggestions applied.

djc (approved):

PR #1049, opened 2021-03-03
perf test improvements

Matthias247:
- Allows to overwrite socket buffer sizes
- Change server logging to use `debug` instead of `trace` logs

### DIFF
--- a/perf/src/server.rs
+++ b/perf/src/server.rs
@@ -27,7 +27,7 @@ async fn main() {
     tracing_subscriber::fmt::init();
 
     if let Err(e) = run(opt).await {
-        error!("{}", e);
+        error!("{:#}", e);
     }
 }
 
--- a/perf/src/server.rs
+++ b/perf/src/server.rs
@@ -4,7 +4,7 @@ use anyhow::{Context, Result};
 use bytes::Bytes;
 use futures::StreamExt;
 use structopt::StructOpt;
-use tracing::{debug, error, info, trace};
+use tracing::{debug, error, info};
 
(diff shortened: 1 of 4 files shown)


perf/src/server.rs
```diff
@@ -75,7 +82,21 @@ async fn run(opt: Opt) -> Result<()> {
...
+        .context("create socket")?;
+    socket
+        .bind(&socket2::SockAddr::from(opt.listen))
+        .context("binding endpoint")?;
+    socket
+        .set_send_buffer_size(opt.send_buffer_size)
+        .context("send buffer size")?;
+    socket
+        .set_recv_buffer_size(opt.recv_buffer_size)
+        .context("receive buffer size")?;
```
Ralith, line 95: There's significant risk of confusion since most distros by default enforce a maximum much lower than the default values set here, which cannot be modified without elevated privileges.

Matthias247: I checked that it won't fail even if the OS max is lower. It will just cap the value.

Manually setting the option makes the testing a bit more resilient on platforms which have a higher `_max` and low `_default` socket buffer settings, which are a couple of production platforms I encountered. So I think its reasonable trying to do it.

If the maximum setting is too low, then this obviously won't help and still impact benchmark results. Maybe there can be another piece of code which reads the buffer size back and prints a warning if too low.

Ralith: Right, my concern is that it's easy to assume the options are doing something when, in the common case, they aren't. A warning would be perfect.

Ralith (approved):

PR #1050, opened 2021-03-03
Final draft updates

Ralith:
Still TODO:
- Audit persistent congestion logic. I'm not sure we're correctly considering only ack-eliciting packets here, or waiting for an RTT sample.

### DIFF
--- a/quinn-proto/src/crypto/rustls.rs
+++ b/quinn-proto/src/crypto/rustls.rs
@@ -188,10 +188,10 @@ impl crypto::Session for TlsSession {
 }
 
 const RETRY_INTEGRITY_KEY: [u8; 16] = [
-    0xcc, 0xce, 0x18, 0x7e, 0xd0, 0x9a, 0x09, 0xd0, 0x57, 0x28, 0x15, 0x5a, 0x6c, 0xb9, 0x6b, 0xe1,
+    0xbe, 0x0c, 0x69, 0x0b, 0x9f, 0x66, 0x57, 0x5a, 0x1d, 0x76, 0x6b, 0x54, 0xe3, 0x68, 0xc8, 0x4e,
 ];
 const RETRY_INTEGRITY_NONCE: [u8; 12] = [
-    0xe5, 0x49, 0x30, 0xf9, 0x7f, 0x21, 0x36, 0xf0, 0x53, 0x0a, 0x8c, 0x1c,
+    0x46, 0x15, 0x99, 0xd3, 0x5d, 0x63, 0x2b, 0xf2, 0x23, 0x98, 0x25, 0xbb,
 ];
 
 impl crypto::HeaderKey for HeaderProtectionKey {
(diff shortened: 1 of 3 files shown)


quinn-proto/src/crypto/rustls.rs
```diff
@@ -56,8 +56,8 @@ impl crypto::Session for TlsSession {
 
     fn initial_keys(dst_cid: &ConnectionId, side: Side) -> Keys<Self> {
         const INITIAL_SALT: [u8; 20] = [
-            0xaf, 0xbf, 0xec, 0x28, 0x99, 0x93, 0xd2, 0x4c, 0x9e, 0x97, 0x86, 0xf1, 0x9c, 0x61,
-            0x11, 0xe0, 0x43, 0x90, 0xa8, 0x99,
+            0x38, 0x76, 0x2c, 0xf7, 0xf5, 0x59, 0x34, 0xb3, 0x4d, 0x17, 0x9a, 0xe6, 0xa4, 0xc8,
```
Matthias247, line 59: Can this support both versions for a while? E.g. msquic and picoquic as a server currently support both accepting v1 and draft connections. As a client they allow to specify the version number to send. 

I think there are also more salts than just this  one.

djc: Maybe, but supporting multiple versions in parallel like that is probably a separate project/PR.

Matthias247: Maybe - but unless this is tackled this will immediately break the project for anyone who needs to interact with a draft version - which are currently e.g. all browsers.

I don't think "quinn was usable, but will stop working for the next couple of month until others have caught up" is a great story.

Ralith:
I'm not opposed to supporting multiple versions, but that may significantly expand the scope of the PR; we've never bothered to straddle breaking changes before.

djc:
I think the time is right for multiple version support, I'd just prefer to decouple it a bit from these updates in terms of when/how we land it. It'd be neat to do it first in a separate PR that adds most of the required mechanics with draft 29-32 support, and then we can add a 34 option in this PR.

Ralith:
Multiple version support is blocked on https://github.com/ctz/rustls/pull/544.

Matthias247:
> It'd be neat to do it first in a separate PR that adds most of the required mechanics with draft 29-32 support, and then we can add a 34 option in this PR.

 Correct me if I'm wrong, but I thought all drafts (incl 34) use the old salts and are already compatible. Only the non draft version (quic v1) uses the new salts and requires extra code to be compatible.

Ralith:
Yeah, we should be able to have 29..=34 trivially.

djc:
Even so, we can still separate the implementation of a mechanism to support different-salt versions from the inclusion of a particular version that actually has a different salt, right?

Matthias247:
Sure! Would be nice to do this before merging the new salts, to avoid breaking draft version users.

Ralith:
Moving non-breaking changes into https://github.com/quinn-rs/quinn/pull/1084; multi-version support can then be implemented in another separate PR, followed by the breaking salt and version number changes.

djc (approved):

PR #1054, opened 2021-03-04
Automatically activate GSO based on platform capabilities

Matthias247:
This set of changes will automatically activate GSO on Linux. In order to do this, the platform is queried for the maximum amount of supported segments per transmit call. This amount is capped at a certain level, and then passed to the `poll_transmit` call.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -393,10 +393,13 @@ where
     /// - the application performed some I/O on the connection
     /// - a call was made to `handle_event`
     /// - a call was made to `handle_timeout`
+    ///
+    /// `max_datagrams` specifies how many datagrams can be returned inside a
+    /// single Transmit using GSO. This must be at least 1.
     #[must_use]
-    pub fn poll_transmit(&mut self, now: Instant) -> Option<Transmit> {
-        // This will become a parameter to `poll_transmit` in an additional update
-        const MAX_DATAGRAMS: usize = 1;
+    pub fn poll_transmit(&mut self, now: Instant, max_datagrams: usize) -> Option<Transmit> {
+        assert!(max_datagrams != 0);
+        let max_datagrams = max_datagrams.min(MAX_TRANSMIT_SEGMENTS);
 
         let mut num_datagrams = 0;
 
@@ -518,7 +521,7 @@ where
                 // We need to send 1 more datagram and extend the buffer for that.
 
                 // Is 1 more datagram allowed?
-                if buf_capacity >= self.path.mtu as usize * MAX_DATAGRAMS {
+                if buf_capacity >= self.path.mtu as usize * max_datagrams {
                     // No more datagrams allowed
                     break;
                 }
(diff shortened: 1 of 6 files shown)


Ralith (approved):

Ralith:
Have we addressed the strange packet assembly recently observed during handshaking?

Matthias247:
> Have we addressed the strange packet assembly recently observed during handshaking?

You mean the padded ACK-only packet? I haven't checked for it so far.

quinn-proto/src/connection/mod.rs
```diff
@@ -393,11 +393,12 @@ where
...
     /// - a call was made to `handle_timeout`
+    ///
+    /// `max_datagrams` specifies how many datagrams can be returned inside a
+    /// single Transmit using GSO. This must be at least 1.
     #[must_use]
-    pub fn poll_transmit(&mut self, now: Instant) -> Option<Transmit> {
-        // This will become a parameter to `poll_transmit` in an additional update
-        const MAX_DATAGRAMS: usize = 1;
-
+    pub fn poll_transmit(&mut self, now: Instant, max_datagrams: usize) -> Option<Transmit> {
```
djc, line 400: This is fine for use through the quinn API, but it seems unergonomic for applications that want to rely on quinn-proto directly. Since it looks like `max_datagrams` is constant over the lifetime of the `Connection`, should we stick in the `Connection` state somehow?

quinn/src/connection.rs
```diff
@@ -1069,3 +1073,10 @@ pub enum SendDatagramError {
     #[error("connection closed: {0}")]
     ConnectionClosed(#[source] ConnectionError),
 }
+
+/// The maximum amount of datagrams that are sent in a single transmit
+///
+/// This can be lower than the maximum platform capabilities, to avoid excessive
+/// memory allocations when calling `poll_transmit()`. Benchmarks have shown
+/// that numbers around 10 are a good compromise.
+const MAX_TRANSMIT_SEGMENTS: usize = 10;
```
djc, line 1082: Similarly, should we cap this in the quinn-proto layer instead of in the quinn layer?

quinn-proto/src/connection/mod.rs
Matthias247: I initially thought about adding it into `TransportConfig`. But the issue here is that `TransportConfig::default` can't set it, because it doesn't have access to platform specific data (defined in `quinn`). We could add another field, and make sure another constructor parameter is set when the quinn connection builds the proto connection. But that might actually require adding the parameter in a lot more places (since constructors delegate to each other).

quinn/src/connection.rs
Matthias247: I don't think it matters a lot, and I'm fine either way. This might be rather temporary anyway. Follow-up changes should either
- look into using bigger pooled buffers to reduce the cost of transmitting more segments
- passing a buffer to `poll_transmit`, which would already provide a hint about the maximum size upfront

quinn-proto/src/connection/mod.rs
djc: It looks like quinn-proto `Connection`s are only constructed in one place in quinn-proto `Endpoint`, and quinn-proto `Endpoint`s are only constructed in one place by the quinn `EndpointBuilder` (module some test code), so it doesn't seem that bad?

quinn/src/connection.rs
djc: If we're capping for the benefit of limiting memory allocations in quinn-proto, we should also cap it there IMO.

quinn-proto/src/connection/mod.rs
Ralith: I don't want this to show up in end user facing APIs like the existing `...Config` structs in quinn. Maybe we could introduce a `BackendConfig` that gets passed to the endpoint at the -proto layer, and then from the endpoint to the connections? That would be semantically fine, though significantly more code than the current approach, which I'm also fine with.

quinn/src/connection.rs
Ralith: Moving the cap to -proto seems reasonable to me.

quinn-proto/src/connection/mod.rs
djc: Okay, let's keep this for now.

quinn/src/connection.rs
Matthias247: I moved the cap

Ralith (approved):

PR #1056, opened 2021-03-04
Perf crate tweaks

djc:


perf/Cargo.toml
```diff
@@ -20,11 +20,3 @@ tokio = { version = "1.0.1", features = ["rt", "macros", "signal", "net", "sync"
 tracing = "0.1.10"
 tracing-subscriber = { version = "0.2.5", default-features = false, features = ["env-filter", "fmt", "ansi", "chrono"]}
 bytes = "1"
-
-[[bin]]
```
Matthias247, line 24: While we are at it: Those also have to be renamed, since they are clashing with the interop binary names (and if you run a full project `cargo build` its arbitrary which ends up in the output folder. I would recommend `perf_server` and `perf_client`

Ralith (approved):

PR #1058, opened 2021-03-08
Perf improvements

Matthias247:
1. Rename executables to avoid name conflicts
2. Disable ipv6 only explicitely, to override the OS defaults and make the perf tools behave the same on all platforms. During perf testing users on windows found it confusing that a server bound to `::0`  can't accept a connection via IPv4.


Ralith (approved):

perf/src/bin/perf_client.rs
```diff
@@ -80,14 +80,7 @@ async fn run(opt: Opt) -> Result<()> {
...
     let socket = bind_socket(
-        SocketAddr::new(
-            if addr.is_ipv4() {
-                IpAddr::V4(Ipv4Addr::UNSPECIFIED)
-            } else {
-                IpAddr::V6(Ipv6Addr::UNSPECIFIED)
-            },
-            0,
-        ),
+        SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 0),
```
Ralith, line 83: I feel like it might be more robust to keep this change, since the peer may not be a quinn perf server, and may therefore not be using a dual stack socket.

Matthias247: How does it matter for the peer? The stack will always create IPv4 packets. Without the option we are just not allowed to create them based off an IPv6 socket.

Matthias247: We can keep the change, but it's then more of a tool to explicitely test IPv4-only endpoints on client side.

Ralith: Oh, I forgot the change affected the client as well. Disregard.

Ralith (approved):

PR #1059, opened 2021-03-08
Use a cheaper hash function

Matthias247:
Rusts default HashMap and HashSet use siphash as a hash method to protect again malicious input. The downside of siphash is that it is slightly more expensive than a more primitive mechanism. The siphash operations currently show up in quinn flamegraphs - even if they only act on 8byte input data like Stream IDs.

The security properties of siphash are not really required in the context of quic, where the hashmaps are used with 2 key types:
- Stream IDs: The valid range for those is predetermined by the maximum concurrent stream setting. The peer can't pick from a bigger range in order to get more hashes into the same bucket. Besides this, the risk of resource exhaustion via purely creating more requests/streams is already much higher than exhaustion of the pure maps.
- Connection IDs: Connection IDs which are inserted into tracking data structures are created locally. The peer has no control over them.

Perf implications: This removes a couple of 0.2% CPU time instances in https://gist.githubusercontent.com/Matthias247/47dc290dde72e2e2f581a114234e8382/raw/c7afb63f0eeae4d3c7f8429d54eb6cdeb47df7f3/8x_gso.svg, which add up to around 1.2%

The impact in the bulk application is hard to measure, but but it might be somewhere in the 1% range.

Some runs without:
```
Sent 10737418240 bytes on 1 streams in 18.98s (539.53 MiB/s)
Sent 10737418240 bytes on 1 streams in 18.92s (541.30 MiB/s)
Sent 10737418240 bytes on 1 streams in 18.80s (544.77 MiB/s)
```

Some runs with the change:
```
Sent 10737418240 bytes on 1 streams in 18.75s (546.18 MiB/s)
Sent 10737418240 bytes on 1 streams in 18.54s (552.32 MiB/s)
```

### DIFF
--- a/quinn-proto/src/connection/spaces.rs
+++ b/quinn-proto/src/connection/spaces.rs
@@ -1,10 +1,12 @@
 use std::{
     cmp,
-    collections::{BTreeMap, HashSet, VecDeque},
+    collections::{BTreeMap, VecDeque},
     mem,
     ops::{Index, IndexMut},
     time::Instant,
 };
+
+use fxhash::FxHashSet;
 
 use super::assembler::Assembler;
 use super::streams::ShouldTransmit;
@@ -235,7 +237,7 @@ pub struct Retransmits {
     pub(crate) max_bi_stream_id: bool,
     pub(crate) reset_stream: Vec<(StreamId, VarInt)>,
     pub(crate) stop_sending: Vec<frame::StopSending>,
-    pub(crate) max_stream_data: HashSet<StreamId>,
+    pub(crate) max_stream_data: FxHashSet<StreamId>,
     pub(crate) crypto: VecDeque<frame::Crypto>,
     pub(crate) new_cids: Vec<IssuedCid>,
     pub(crate) retire_cids: Vec<u64>,
@@ -286,7 +288,7 @@ impl Default for Retransmits {
             max_bi_stream_id: false,
             reset_stream: Vec::new(),
             stop_sending: Vec::new(),
-            max_stream_data: HashSet::new(),
+            max_stream_data: FxHashSet::default(),
             crypto: VecDeque::new(),
             new_cids: Vec::new(),
             retire_cids: Vec::new(),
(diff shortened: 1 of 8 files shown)


quinn-proto/src/connection/spaces.rs
```diff
@@ -1,6 +1,7 @@
+use fxhash::FxHashSet;
```
djc, line 1: Let's maintain the usual import grouping/ordering here.

djc (commented):
Nice! We might even have an old issue about this.

Matthias247:
> Nice! We might even have an old issue about this.

Oh, I wasn't aware. Seems like this had been discussed in https://github.com/quinn-rs/quinn/issues/88 and https://github.com/quinn-rs/quinn/pull/479.

I guess back then there had been less performance implications, since other bottlenecks had been much larger. Now it's starting to get more visible.

djc (approved):

Matthias247:
Based on the comments in #88 I kept the maps in `Endpoint`  which index on peer selected values as standard maps. Those are not in the hot path anyway.

quinn-proto/src/endpoint.rs
```diff
@@ -844,7 +851,7 @@ pub enum ConnectError {
 }
 
 #[derive(Default, Debug)]
-struct ResetTokenTable(HashMap<SocketAddr, HashMap<ResetToken, ConnectionHandle>>);
+struct ResetTokenTable(FxHashMap<SocketAddr, FxHashMap<ResetToken, ConnectionHandle>>);
```
Ralith, line 854: Using a `FxHashMap` here but a regular `HashMap` for `connection_remotes` seems inconsistent. I'm uncertain whether there's a meaningful collision attack risk with socket addrs, since they're only partially controlled, but we should be consistent either way.

djc:
It's great that you added comments to clarify why the standard hashmap is used, but maybe the burden of proof should be on the maps that use fx? (Just keep the other comments because it's still very useful information.)

quinn-proto/src/endpoint.rs
Matthias247: Agreed on both things

Ralith (approved):

Issue #1061, opened 2021-03-10
New default value of 'max_concurrent_uni_streams'


bochaco:
I'm having issues after upgrading from v0.6.1 to v0.7.0 with sending many messages over uni-streams and I don't get to understand what could be happening.

From looking at https://github.com/quinn-rs/quinn/blob/main/quinn-proto/src/config.rs#L229 it seems by default now this is set to 100, which in v0.6.1 it was by default set to infinite, is this correct?

Then, looking at the example  https://github.com/quinn-rs/quinn/blob/main/quinn/examples/server.rs#L67 I'm getting the idea that is setting it to be infinite, is this correct?

In a crate I'm working on, when setting this config to 0 somehow is causing that our "server" side app cannot accept any message, and when default is used it can accept up to 100 but then something is going on that we cannot send more than that even after invoking `finish` in the uni-stream used.
Is there anything I should consider in particular to be able to set this config to "infinite" in an attempt to have it behave the same as in previous v0.6.1?

Ralith:
> it seems by default now this is set to 100, which in v0.6.1 it was by default set to infinite, is this correct?

Basically, yeah. Previously we had a "max streams waiting for accept" limit, but we determined that "max concurrent streams" is more useful, since applications that wish to consume bounded resources (i.e. most reasonable applications) would otherwise need to reinvent that themselves. You could set it to `VarInt::MAX` to effectively disable it for testing purposes or if you're certain clients are well-behaved.

> I'm getting the idea that is setting it to be infinite, is this correct?

No, setting that figure to 0 means you may have at most zero concurrent streams, which will disable accepting that type of stream entirely, i.e. the behavior you observe is intended.

> we cannot send more than [100] even after invoking finish in the uni-stream used.

Does the receiving application read the streams all the way to the end? If so, this may be a bug; do you have a repro?

bochaco:
> Does the receiving application read the streams all the way to the end?

This seems like could be the issue I'm facing, let me research a bit more to confirm, and/or to share some test to reproduce it, but I just gave it a quick try to `stop` the `RecvStream` after reading (since I'm reading with `read_exact` only) and things look much better. I'll get back to you with more info.

bochaco:
> Does the receiving application read the streams all the way to the end?

This is likely the issue I'm facing, I just gave it a quick try and things look better by making sure the `RecvStream` is read till the end. Let me either confirm or get back to you with some snippet to reproduce it if not.

[...]

PR #1062, opened 2021-03-11
Fix stream leak on stop after finish

Ralith:
Fixes https://github.com/quinn-rs/quinn/issues/1061. This will deadlock connections after the maximum concurrent number of streams is reached, so we should probably backport to 0.7.


Ralith:
To clarify a little, the deadlock only arises if the application stops a stream after receiving a finish, instead of reading to the end. This isn't typical, but definitely can occur in realistic conditions as illustrated by the issue.

PR #1064, opened 2021-03-11
Additional backports for point release

Ralith:
These will improve interoperability and forwards-compat of the release.


djc (approved):
Good call!

PR #1065, opened 2021-03-12
Re-export `ConfigError`

daxpedda:
Specifically `TransportConfig::max_concurrent_bidi_streams` and `TransportConfig::max_concurrent_uni_streams` return `quinn_proto::ConfigError`.

This prevents consumers of `quinn` having to bring in `quinn-proto` just to use `ConfigError`, for example in their custom error type.

### DIFF
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -60,8 +60,9 @@ mod recv_stream;
 mod send_stream;
 
 pub use proto::{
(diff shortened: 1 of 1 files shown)


Ralith (approved):
Thanks!

Ralith:
Merged in 3a6cf6bfc38e686af7e09d29e21556927ff31246.

PR #1072, opened 2021-03-19
Create `ClientConfig` lazily in EndpointBuilder 

Matthias247:
Creating a `ClientConfig` can on some platforms take an excessive amount
of time, likely due to rustls scanning the systems cert store.
This slows down all functions which make use of `quinn::Endpoint::builder()`,
even if the client config is known upfront and shared between mutliple connections.

In some tests I noticed an extra connection establishment time of 1s for
creating a client.

This changes will create the `ClientConfig` only if the Endpoint is built
when the `ClientConfig` hasn't been set by the application.

Server side code still suffers from the issue, since `ClientConfig`s there are not
known. However it might be less impacting, since a second of startup time for
a server matters less than for a client.

### DIFF
--- a/quinn/src/builders.rs
+++ b/quinn/src/builders.rs
@@ -27,7 +27,7 @@ where
 {
     server_config: Option<ServerConfig<S>>,
     config: EndpointConfig<S>,
-    default_client_config: ClientConfig<S>,
+    default_client_config: Option<ClientConfig<S>>,
 }
 
 #[allow(missing_docs)]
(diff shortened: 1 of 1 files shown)


Ralith (approved):

djc:
Should we change the EndpointBuilder to construct the default config lazily as needed?

Ralith:
I'd be fine with that.

djc (approved):

PR #1073, opened 2021-03-19
Update to socket2 v0.4

Thomasdezeeuw:


Thomasdezeeuw:
Socket2 requires Rust 1.46, is a bump from 1.45 acceptable?

djc:
1.46 was released on August 26th, so more than 6 months ago; should be fine to bump the MSRV.

Thomasdezeeuw:
I've increased the MSRV to 1.46.

Also note that in the `Default`  impl for `quin-proto::frame::StreamMeta` says it can be derived started at 1.46: https://github.com/quinn-rs/quinn/blob/ab61ef89f534e0c9ce82379c0b954591e5f2d99a/quinn-proto/src/frame.rs#L433-L444

I didn't change it to keep the changed to a minimum.

djc:
Ah yeah, 1.46 has known issues with the type length limit. Can you increase it as per the errors?

Thomasdezeeuw:
Do I add to to `interop/src/main.rs` or to one of the crates? I'm not familiar enough with the code base to make that call.

djc:
Just to interop. One of the other crates might need it, too, but rustc will tell you.

Thomasdezeeuw:
I've added the attribute to all binaries in interop, but it still  doesn't work.

djc:
You'll need to bump the limit some more, according to the new error message. Sorry, this stuff is a bit annoying.

Thomasdezeeuw:
CI is finally green.

djc (approved):
Thanks!

Issue #1074, opened 2021-03-19
Missing retransmits when first ACK from client is lost


Matthias247:
I have a trace which looks like:

```
Mar 19 01:00:27.832 TRACE conn{thread_id=0 conn_id=0}: quinn_proto::endpoint: initial_dcid=35ca09fba7ba35ef696086a61b8249496e06aa6f
Mar 19 01:00:27.832 TRACE conn{thread_id=0 conn_id=0}: quinn_proto::connection: wrote 290 Initial CRYPTO bytes
Mar 19 01:00:27.832 TRACE drive{id=0}:send{space=Initial pn=0}: quinn_proto::connection: CRYPTO: off 0 len 290
Mar 19 01:00:27.832 TRACE drive{id=0}:send{space=Initial pn=0}: quinn_proto::connection: PADDING * 851
Mar 19 01:00:27.832 TRACE drive{id=0}: quinn_proto::connection: sending 1200 bytes in 1 datagrams

Mar 19 01:00:27.879 TRACE drive{id=0}: quinn_proto::connection: got Initial packet (154 bytes) from [::ffff:55.116.87.18]:443 using id ea51594835e6eb89
Mar 19 01:00:27.879 TRACE drive{id=0}:recv{space=Initial pn=0}: quinn_proto::connection: authenticated
Mar 19 01:00:27.879 TRACE drive{id=0}:recv{space=Initial pn=0}: quinn_proto::connection: switching remote CID to 108400d95eb73675946a4a55d819f7e2c896840a
Mar 19 01:00:27.879 DEBUG drive{id=0}:recv{space=Initial pn=0}:frame{ty=ACK}: quinn_proto::connection: ECN not acknowledged by peer
Mar 19 01:00:27.879 TRACE drive{id=0}:recv{space=Initial pn=0}:frame{ty=CRYPTO}: quinn_proto::connection: consumed 90 CRYPTO bytes
Mar 19 01:00:27.879 TRACE drive{id=0}:recv{space=Initial pn=0}: quinn_proto::connection: Handshake keys ready
Mar 19 01:00:27.879 TRACE drive{id=0}: quinn_proto::connection: got Handshake packet (1078 bytes) from [::ffff:55.116.87.18]:443 using id ea51594835e6eb89
Mar 19 01:00:27.879 TRACE drive{id=0}:recv{space=Handshake pn=0}: quinn_proto::connection: authenticated
[...]

PR #1076, opened 2021-03-20
Fix lost client ack deadlock

Ralith:
Fixes https://github.com/quinn-rs/quinn/issues/1074.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1292,7 +1292,10 @@ where
             }
             let last_ack_eliciting = match self.spaces[space].time_of_last_ack_eliciting_packet {
                 Some(time) => time,
-                None => continue,
+                None => {
+                    debug_assert!(!self.peer_completed_address_validation());
+                    now
+                }
             };
             let pto = last_ack_eliciting + duration;
             if result.map_or(true, |(earliest_pto, _)| pto < earliest_pto) {
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1176,16 +1176,13 @@ where
             return;
         }
 
-        let (space, count) = self.pto_time_and_space(now).map_or_else(
-            // Handshake anti-deadlock packet
-            || {
-                debug_assert!(self.side.is_client() && self.highest_space <= SpaceId::Handshake);
-                (self.highest_space, 1)
-            },
-            // PTO
-            |(_, pto_space)| (pto_space, 2),
-        );
-
+        let (_, space) = match self.pto_time_and_space(now) {
+            Some(x) => x,
+            None => {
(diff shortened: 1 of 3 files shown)


quinn/examples/server.rs
```diff
@@ -87,7 +87,9 @@ async fn run(options: Opt) -> Result<()> {
         };
         let cert_chain = fs::read(cert_path).context("failed to read certificate chain")?;
         let cert_chain = if cert_path.extension().map_or(false, |x| x == "der") {
-            quinn::CertificateChain::from_certs(quinn::Certificate::from_der(&cert_chain))
+            quinn::CertificateChain::from_certs(Some(
```
Matthias247, line 90: is that intentional?

Ralith: Yes, see the last commit message; just some cleanup I threw in after being briefly confused while referencing the example.

Matthias247: Ah ok. I wasn't even aware that this is a `Result`. I thought the method returns an `Option`, in which case this piece of code would just have unwrapped and re-wrapped it.

Ralith: Yeah, it was pretty confusing, hence the fix :D

quinn-proto/src/connection/mod.rs
```diff
@@ -1176,23 +1176,29 @@ where
...
 
+        let count = if self.in_flight.ack_eliciting == 0 {
+            // A PTO when we're not expecting any ACKs must be due to handshake anti-amplification
+            // deadlock preventions
+            debug_assert!(!self.peer_completed_address_validation());
+            1
+        } else {
+            // Conventional loss probe
+            2
+        };
```
djc, line 1201: I think this might be easier to follow when written as `match self.in_flight.ack_eliciting {}`?

quinn-proto/src/connection/mod.rs
```diff
@@ -1298,10 +1298,7 @@ where
             }
             let last_ack_eliciting = match self.spaces[space].time_of_last_ack_eliciting_packet {
                 Some(time) => time,
-                None => {
-                    debug_assert!(!self.peer_completed_address_validation());
-                    now
-                }
+                None => continue,
```
djc: If you're going to revert it anyway, should we just drop the first commit?

Ralith: The first commit represents the absolute minimal fix to get correct behavior, while the later fix relies on stronger inferences about the behavior of the state machine. Given how subtle the logic is, I'm inclined to separate those out to make it easier to look back, and perhaps amend, our approach without risking a regression, particularly while https://github.com/quicwg/base-drafts/issues/4846 is still open.

djc: Okay.

djc (approved):

PR #1080, opened 2021-03-20
Lazily initialize Endpoint::default_client_config

### LINKED ISSUE #1079: Don't create `ClientConfig` for server `Endpoint`s
This is more or less a follow-up to https://github.com/quinn-rs/quinn/pull/1072, which isn't addressed there

For server-side endpoints, `ClientConfig` is not required. However it will still be created via ` ClientConfig::default()`, which triggers all the cert loading actions. It would be nice to get rid of that part. However since the `Endpoint` is currently agnostic to being client or server side this seems a bit more tricky.

Ralith:
Pure servers never need this. once_cell is already a transitive dep, and is expected to eventually land in std, so it's free to add.

Fixes #1079.

I realized after writing this that a perhaps more natural fix for pure servers wasting time on certificate loading is to disable the `native-certs` feature, so I'm not sure this is needed. Still, this is a small change, so worth considering. Thoughts?

### DIFF
--- a/quinn/Cargo.toml
+++ b/quinn/Cargo.toml
@@ -34,6 +34,7 @@ futures = "0.3.8"
 fxhash = "0.2.1"
 libc = "0.2.69"
 mio = { version = "0.7.7", features = ["net"] }
+once_cell = "1.7.2"
 proto = { package = "quinn-proto", path = "../quinn-proto", version = "0.7" }
 rustls = { version = "0.19", features = ["quic"], optional = true }
 socket2 = "0.3"
--- a/quinn/src/builders.rs
+++ b/quinn/src/builders.rs
@@ -1,5 +1,6 @@
 use std::{io, net::SocketAddr, sync::Arc};
 
+use once_cell::sync::OnceCell;
 use proto::{
     generic::{ClientConfig, EndpointConfig, ServerConfig},
     ConnectionIdGenerator,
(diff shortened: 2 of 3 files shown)


quinn/src/endpoint.rs
```diff
@@ -37,7 +38,7 @@ where
     S: proto::crypto::Session,
 {
     pub(crate) inner: EndpointRef<S>,
-    pub(crate) default_client_config: ClientConfig<S>,
+    pub(crate) default_client_config: OnceCell<ClientConfig<S>>,
```
Matthias247, line 41: I'm not super sure about the global state here. Should it rather be a `Mutex<Option<ClientConfig>>`?

Ralith: What global state?

Matthias247: Oh, I guess I misinterpreted it and thought `OnceCell`  would be like `lazy_static` and initialize things from a global reference.

quinn/Cargo.toml
```diff
@@ -41,6 +41,7 @@ thiserror = "1.0.21"
 tracing = "0.1.10"
 tokio = { version = "1.0.1", features = ["net", "rt", "rt-multi-thread", "time"] }
 webpki = { version = "0.21", optional = true }
+once_cell = "1.7.2"
```
djc, line 44: Let's keep dependencies ordered?

quinn/src/builders.rs
```diff
@@ -82,7 +83,10 @@ where
         Ok((
             Endpoint {
                 inner: rc.clone(),
-                default_client_config: self.default_client_config.unwrap_or_default(),
+                default_client_config: self
+                    .default_client_config
+                    .map(OnceCell::from)
+                    .unwrap_or_default(),
```
djc, line 89: I find this pretty dense. Maybe add a comment?

Ralith:
One case where this might still be useful is if an application contains both server and client functionality, but doesn't necessarily use both at runtime. Not sure how realistic that might be, though.

djc (approved):

PR #1086, opened 2021-03-26
Logging improvements

Matthias247:
This contains 2 small logging changes which should aid debugging

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -2144,6 +2144,22 @@ where
             };
 
             self.stats.frame_rx.record(&frame);
+            // Crypto, Stream and Datagram frames are special cased in order no pollute
+            // the log with payload data
+            match &frame {
+                Frame::Crypto(f) => {
+                    trace!(offset = f.offset, len = f.data.len(), "got crypto frame");
+                }
+                Frame::Stream(f) => {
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -2144,6 +2143,19 @@ where
...
+            // the log with payload data
+            match &frame {
+                Frame::Crypto(f) => {
+                    trace!(offset = f.offset, len = f.data.len(), "got crypto frame");
+                }
+                Frame::Stream(f) => {
+                    trace!(id = %f.id, offset = f.offset, len = f.data.len(), fin = f.fin, "got stream frame");
+                }
+                f => {
+                    trace!("got frame {:?}", f);
```
Ralith, line 2156:
```suggestion
                    trace!("got frame");
```
The frame type is already recorded in the span.

Ralith: Or are you intentionally including the frame payload?

Matthias247: Yes. For all the `MAX_` frames it is interesting to know what the limits are. The payloads should never be too long apart for CRYPTO and STREAM, which are special cased for this reason

Matthias247: Ok, guess we should special case `DATAGRAM` too

Ralith (dismissed):

Ralith (approved):

PR #1087, opened 2021-03-27
Chunks fixes

Ralith:
This should turn into a point release, since the current behavior will cause long-lived connections to become stuck on connection-level flow control.

### DIFF
--- a/quinn-proto/src/connection/streams/recv.rs
+++ b/quinn-proto/src/connection/streams/recv.rs
@@ -242,9 +242,9 @@ impl<'a> Chunks<'a> {
     pub fn next(&mut self, max_length: usize) -> Result<Option<Chunk>, ReadError> {
         let mut rs = match mem::replace(&mut self.state, ChunksState::Finalized) {
             ChunksState::Readable(rs) => rs,
-            ChunksState::Error(e, st) => {
-                self.state = ChunksState::Error(e.clone(), st);
-                return Err(e);
+            ChunksState::Reset(error_code) => {
+                self.state = ChunksState::Reset(error_code);
+                return Err(ReadError::Reset(error_code));
             }
             ChunksState::Finished => {
                 self.state = ChunksState::Finished;
@@ -271,9 +271,8 @@ impl<'a> Chunks<'a> {
                     max_streams_dirty,
                 );
 
-                let err = ReadError::Reset(error_code);
-                self.state = ChunksState::Error(err.clone(), ShouldTransmit(true));
-                Err(err)
+                self.state = ChunksState::Reset(error_code);
+                Err(ReadError::Reset(error_code))
             }
             RecvState::Recv { size } => {
                 if size == Some(rs.end) && rs.assembler.bytes_read() == rs.end {
@@ -287,10 +286,11 @@ impl<'a> Chunks<'a> {
                     self.state = ChunksState::Finished;
                     Ok(None)
                 } else {
-                    let should_transmit =
-                        Self::done(&mut rs, self.read, self.id, self.streams, self.pending);
-                    self.state = ChunksState::Error(ReadError::Blocked, should_transmit);
-                    self.streams.recv.insert(self.id, rs);
+                    // We don't need a distinct `ChunksState` variant for a blocked stream because
+                    // retrying a read harmlessly re-traces our steps back to returning
+                    // `Err(Blocked)` again. The buffers can't refill and the stream's own state
+                    // can't change so long as this `Chunks` exists.
+                    self.state = ChunksState::Readable(rs);
                     Err(ReadError::Blocked)
                 }
             }
@@ -319,9 +319,9 @@ impl<'a> Chunks<'a> {
                 self.pending.max_data |= max_data.0;
                 max_data
             }
-            ChunksState::Error(_, should_transmit) => {
+            ChunksState::Reset(_) => {
                 debug_assert!(!drop);
-                should_transmit
+                ShouldTransmit(true)
             }
             ChunksState::Finalized => ShouldTransmit(false),
         }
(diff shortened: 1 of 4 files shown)


djc:
Do you want to backport refactoring too? Or is there a subset of commits here that we should backport?

Ralith:
I'm more confident in corner-case coverage of the post-refactoring code, subject to review, so I'm inclined to err on the side of more backports.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -304,40 +285,43 @@ impl<'a> Chunks<'a> {
...
-                self.streams.recv.insert(self.id, rs);
-                should_transmit
-            }
-            ChunksState::Finished => {
-                debug_assert!(!drop);
-                // MAX_DATA may need to be issued, but MAX_STREAM_DATA is pointless
-                let max_data = self.streams.add_read_credits(self.read);
-                self.pending.max_data |= max_data.0;
-                max_data
+        debug_assert!(
```
djc, line 284: This looks great! Do you think it's feasible/not too hard to do this change (centralize post-read logic) in like two or three steps? One step seems to be renaming the `Error` state, another to move logic from `next()` into `finalize_inner()`, another to inline `post_read()`, maybe more? As is, this is a bit hard to review, and given how finicky this has been/is turning out to be, I think it'd be good to be careful about changing it.

djc (commented):
Great work! Not sure why I didn't figure out the state handling simplification or to re-inline the `post_read()` logic when I was working on this.

quinn-proto/src/connection/streams/recv.rs
Ralith: Splitting went better than expected; should be much easier to digest now!

Ralith:
> Not sure why I didn't figure out the state handling simplification or to re-inline the post_read() logic when I was working on this.

I think it helped a lot to be looking at it with fresh eyes, rather than comparing it to the even more tangled state it was in originally.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -287,10 +287,7 @@ impl<'a> Chunks<'a> {
                     self.state = ChunksState::Finished;
                     Ok(None)
                 } else {
-                    let should_transmit =
-                        Self::done(&mut rs, self.read, self.id, self.streams, self.pending);
-                    self.state = ChunksState::Error(ReadError::Blocked, should_transmit);
```
djc, line 292: Since `Reset` becomes the only possible error kind for `ChunksState::Error` here, should we get rid of the `error` field as part of this commit (or even squash the stuff where the variant gets renamed)?

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -287,10 +287,7 @@ impl<'a> Chunks<'a> {
...
-                    self.streams.recv.insert(self.id, rs);
+                    self.state = ChunksState::Readable(rs);
```
djc: It feels a bit counterintuitive that we just set `state` back to `Readable` even though we're returning `Blocked` here. Maybe add a comment with some context?

quinn-proto/src/connection/streams/recv.rs
djc: Yup, much better!

djc (commented):
Sorry for taking a while, the nuances of these changes are still quite hard to grok. Maybe some commit messages that are explicit about any unobvious implicit side effects would help?

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -304,24 +304,29 @@ impl<'a> Chunks<'a> {
             !drop || matches!(state, ChunksState::Finalized),
             "finalize must be called before drop"
         );
-        match state {
+        if matches!(state, ChunksState::Finalized) {
+            // Noop on repeated calls
+            return ShouldTransmit(false);
+        }
```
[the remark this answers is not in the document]
Ralith: Moved this to the next commit.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -304,24 +304,29 @@ impl<'a> Chunks<'a> {
...
+            ChunksState::Finished => false,
+            ChunksState::Reset(_) => true,
+            ChunksState::Finalized => false,
+        };
+
+        // Issue connection-level flow control credit for any data we read regardless of state
+        let max_data = self.streams.add_read_credits(self.read);
+        self.pending.max_data |= max_data.0;
+        should_transmit |= max_data.0;
+        ShouldTransmit(should_transmit)
```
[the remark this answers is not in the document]
Ralith: There actually isn't a semantic change. Consider each state:
- Readable: Credit issued before and after the change.
- Finished: Credit issued before and after the change.
- Reset: Before the change, credit never issued because `add_read_credits` isn't called. After the change, credit never issued because `self.read` is necessarily 0, because we drop all assembler buffers immediately upon receiving a reset, therefore no data can have been read. This is pretty subtle, so I added a `debug_assert` to clarify and extended a test to validate.
- Finalized: Credit never issued before or after the change because `add_read_credits` isn't called.

quinn-proto/src/connection/streams/recv.rs
Ralith: Squashed.

quinn-proto/src/connection/streams/recv.rs
Ralith: Added some discussion.

quinn-proto/src/connection/streams/recv.rs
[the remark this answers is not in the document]
djc: Ah, very cool. Thanks!

djc (approved):

PR #1090, opened 2021-03-29
Use an array-based RangeSet for tracking outgoing ACKs

Matthias247:
This change introduces a second `RangeSet` type which is based on a
linear array/vector instead of a tree. For the purpose of tracking
ACK ranges we usually should not have a big number of disjoint ranges,
since those would only occur with severe fragmentation. Therefore the
main benefit of a tree which is able to find ranges in the middle is
rarely used.

The array-based RangeSet provides 2 benefits:
- There exists an inline representation, which avoids the need of heap
  allocating ACK ranges for `SentFrames` for small ranges.
- Iterating over ranges should usually be faster since there is only
  a single cache-friendly contiguous range.

Performance differences:

With the `BTreeMap` based `RangeSet`:
```
Sent 1073741824 bytes on 1 streams in 1.90s (538.89 MiB/s)
```

With the `TinyVec` based `RangeSet`:
```
Sent 1073741824 bytes on 1 streams in 1.78s (574.73 MiB/s)
```

### DIFF
--- a/quinn-proto/src/range_set.rs
+++ b/quinn-proto/src/range_set.rs
@@ -56,7 +56,7 @@ impl RangeSet {
     }
 
     pub fn insert(&mut self, mut x: Range<u64>) -> bool {
-        if x.end == 0 || x.start == x.end {
+        if x.is_empty() {
             return false;
         }
         if let Some((start, end)) = self.pred(x.start) {
@@ -99,6 +99,10 @@ impl RangeSet {
     }
 
     pub fn remove(&mut self, x: Range<u64>) -> bool {
+        if x.is_empty() {
+            return false;
+        }
+
         let before = match self.pred(x.start) {
             Some((start, end)) if end > x.start => {
                 self.0.remove(&start);
--- a/quinn-proto/src/range_set/array_range_set.rs
+++ b/quinn-proto/src/range_set/array_range_set.rs
@@ -1,0 +1,209 @@
+use std::ops::Range;
+
+use tinyvec::TinyVec;
+
+/// A set of u64 values optimized for long runs and random insert/delete/contains
+///
+/// `ArrayRangeSet` uses an array representation, where each array entry represents
+/// a range.
+///
+/// The array-based RangeSet provides 2 benefits:
+/// - There exists an inline representation, which avoids the need of heap
+///   allocating ACK ranges for SentFrames for small ranges.
+/// - Iterating over ranges should usually be faster since there is only
+///   a single cache-friendly contiguous range.
+///
+/// `ArrayRangeSet` is especially useful for tracking ACK ranges where the amount
+/// of ranges is usually very low (since ACK numbers are in consecutive fashion
+/// unless reordering or packet loss occur).
+#[derive(Debug, Default)]
+pub struct ArrayRangeSet(TinyVec<[Range<u64>; ARRAY_RANGE_SET_INLINE_CAPACITY]>);
+
+/// The capacity of elements directly stored in [`ArrayRangeSet`]
+///
+/// An inline capacity of 2 is chosen to keep `SentFrame` below 128 bytes.
+const ARRAY_RANGE_SET_INLINE_CAPACITY: usize = 2;
+
+impl Clone for ArrayRangeSet {
+    fn clone(&self) -> Self {
+        // tinyvec keeps the heap representation after clones.
+        // We rather prefer the inline representation for clones if possible,
+        // since clones (e.g. for storage in `SentFrames`) are rarely mutated
+        if self.0.is_inline() || self.0.len() > ARRAY_RANGE_SET_INLINE_CAPACITY {
+            return Self(self.0.clone());
+        }
+
+        let mut vec = TinyVec::new();
+        vec.extend_from_slice(self.0.as_slice());
+        Self(vec)
+    }
+}
+
+impl ArrayRangeSet {
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    pub fn iter(&self) -> impl DoubleEndedIterator<Item = Range<u64>> + '_ {
+        self.0.iter().cloned()
+    }
+
+    pub fn elts(&self) -> impl Iterator<Item = u64> + '_ {
+        self.iter().flatten()
+    }
+
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    pub fn contains(&self, x: u64) -> bool {
+        for range in self.0.iter() {
+            if range.start > x {
+                // We only get here if there was no prior range that contained x
+                return false;
+            } else if range.contains(&x) {
+                return true;
+            }
+        }
+        false
+    }
+
+    pub fn subtract(&mut self, other: &ArrayRangeSet) {
+        // TODO: This can potentially be made more efficient, since the we know
+        // individual ranges are not overlapping, and the next range must start
+        // after the last one finished
+        for range in &other.0 {
+            self.remove(range.clone());
+        }
+    }
+
+    pub fn insert_one(&mut self, x: u64) -> bool {
+        self.insert(x..x + 1)
+    }
+
+    pub fn insert(&mut self, x: Range<u64>) -> bool {
+        let mut result = false;
+
+        if x.is_empty() {
+            // Don't try to deal with ranges where x.end <= x.start
+            return false;
+        }
+
+        let mut idx = 0;
+        while idx != self.0.len() {
+            let range = &mut self.0[idx];
+
+            if range.start > x.end {
+                // The range is fully before this range and therefore not extensible.
+                // Add a new range to the left
+                self.0.insert(idx, x);
+                return true;
+            } else if range.start > x.start {
+                // The new range starts before this range but overlaps.
+                // Extend the current range to the left
+                // Note that we don't have to merge a potential left range, since
+                // this case would have been captured by merging the right range
+                // in the previous loop iteration
+                result = true;
+                range.start = x.start;
+            }
+
+            // At this point we have handled all parts of the new range which
+            // are in front of the current range. Now we handle everything from
+            // the start of the current range
+
+            if x.end <= range.end {
+                // Fully contained
+                return result;
+            } else if x.start <= range.end {
+                // Extend the current range to the end of the new range.
+                // Since it's not contained it must be bigger
+                range.end = x.end;
+
+                // Merge all follow-up ranges which overlap
+                while idx != self.0.len() - 1 {
+                    let curr = self.0[idx].clone();
+                    let next = self.0[idx + 1].clone();
+                    if curr.end >= next.start {
+                        self.0[idx].end = next.end.max(curr.end);
+                        self.0.remove(idx + 1);
+                    } else {
+                        break;
+                    }
+                }
+
+                return true;
+            }
+
+            idx += 1;
+        }
+
+        // Insert a range at the end
+        self.0.push(x);
+        true
+    }
+
+    pub fn remove(&mut self, x: Range<u64>) -> bool {
+        let mut result = false;
+
+        if x.is_empty() {
+            // Don't try to deal with ranges where x.end <= x.start
+            return false;
+        }
+
+        let mut idx = 0;
+        while idx != self.0.len() && x.start != x.end {
+            let range = self.0[idx].clone();
+
+            if x.end <= range.start {
+                // The range is fully before this range
+                return result;
+            } else if x.start >= range.end {
+                // The range is fully after this range
+                idx += 1;
+                continue;
+            }
+
+            // The range overlaps with this range
+            result = true;
+
+            let left = range.start..x.start;
+            let right = x.end..range.end;
+            if left.is_empty() && right.is_empty() {
+                self.0.remove(idx);
+            } else if left.is_empty() {
+                self.0[idx] = right;
+                idx += 1;
+            } else if right.is_empty() {
+                self.0[idx] = left;
+                idx += 1;
+            } else {
+                self.0[idx] = right;
+                self.0.insert(idx, left);
+                idx += 2;
+            }
+        }
+
+        result
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    pub fn pop_min(&mut self) -> Option<Range<u64>> {
+        if !self.0.is_empty() {
+            Some(self.0.remove(0))
+        } else {
+            None
+        }
+    }
+
+    pub fn min(&self) -> Option<u64> {
+        self.iter().next().map(|x| x.start)
+    }
+
+    pub fn max(&self) -> Option<u64> {
+        self.iter().rev().next().map(|x| x.end - 1)
+    }
+}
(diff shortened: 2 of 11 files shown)


quinn-proto/src/range_set.rs
```diff
@@ -8,6 +8,292 @@ use std::{
...
+            end: 0,
+        }
+    }
+
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    pub fn contains(&self, x: u64) -> bool {
+        for range in self.0.iter() {
```
djc, line 119: Style nit: I guess this could leverage `Iterator::find_map()`?

quinn-proto/src/range_set.rs
```diff
@@ -8,6 +8,292 @@ use std::{
     },
 };
 
+use tinyvec::TinyVec;
```
djc, line 11: Hmm, maybe this deserves a separate file? Could have `range_set/btree.rs`, `range_set/array.rs`, `range_set/tests.rs` so there's a semi-sensible place to put the tests.

quinn-proto/src/range_set.rs
```diff
@@ -8,6 +8,292 @@ use std::{
...
+    front_idx: usize,
+    back_idx: usize,
+}
+
+impl<'a> Iterator for ArrayRangeSetIter<'a> {
+    type Item = Range<u64>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.front_idx == self.back_idx {
+            None
```
djc, line 37: Style nit: in both the `Iterator` and `DoubleEndedIterator` impls, maybe just `return None;` eagerly in the simple case and remove the `else` in favor of continuing?

quinn-proto/src/range_set.rs
```diff
@@ -8,6 +8,292 @@ use std::{
...
+        // tinyvec keeps the heap representation after clones.
+        // We rather prefer the inline representation for clones if possible,
+        // since clones (e.g. for storage in `SentFrames`) are rarely mutated
+        if self.0.is_heap() && self.0.len() <= ARRAY_RANGE_SET_INLINE_CAPACITY {
+            let mut vec = TinyVec::new();
+            vec.extend_from_slice(self.0.as_slice());
+            Self(vec)
+        } else {
+            Self(self.0.clone())
+        }
```
djc, line 89: Here too, I'd invert the conditional and handle the simple case eagerly.

quinn-proto/src/range_set.rs
```diff
@@ -8,6 +8,292 @@ use std::{
     },
 };
 
+use tinyvec::TinyVec;
+
+/// A set of u64 values optimized for long runs and random insert/delete/contains
+///
+/// `ArrayRangeSet` uses an array representation, where each array entry represents
+/// a range.
```
djc, line 16: Would be nice to include some of the higher-level documentation from the commit message here, that is, why we need a second implementation of `RangeSet` (and maybe also explicitly call out ack ranges as the goal). Maybe we should even change the name to `AckRangeSet`?

djc (commented):
Initial round of feedback. Awesome perf improvement!

quinn-proto/src/range_set.rs
Matthias247: I agree and have reorganized things

Matthias247:
What should be done about the 1.46 issue? Will MSRV be updated? If not this requires a custom `Range` struct which impls `Default`. Can add that one too, but will avoid the effort if MSRV is updated anyway

quinn-proto/src/range_set.rs
djc: Ok.

djc:
1.47 was released on October 8th, so I suppose we could release something depending on 1.47 from Apr 8th. Since that's just a ~week ago, adding this on main should be fine.

quinn-proto/src/range_set/array_range_set.rs
```diff
@@ -0,0 +1,301 @@
...
+///
+/// The array-based RangeSet provides 2 benefits:
+/// - There exists an inline representation, which avoids the need of heap
+///   allocating ACK ranges for SentFrames for small ranges.
+/// - Iterating over ranges should usually be faster since there is only
+///   a single cache-friendly contiguous range.
+///
+/// `ArrayRangeSet` is especially useful for tracking ACK ranges where the amount
+/// of ranges is usually very low (since ACK numbers are in consecutive fashion
+/// unless reordering happens).
```
Ralith, line 18: reordering or packet loss

quinn-proto/src/range_set/array_range_set.rs
```diff
@@ -0,0 +1,301 @@
...
+        }
+    }
+
+    pub fn elts(&self) -> ArrayRangeSetEltIter<'_> {
+        ArrayRangeSetEltIter {
+            inner: self.iter(),
+            start: 0,
+            end: 0,
+        }
+    }
```
Ralith, line 114: Do we need these types to be named? We could save a decent chunk of code with `impl (DoubleEnded)Iterator` and combinators here. Even if we do need named types, newtyping the combinators (`Copied<slice::Iter>`, `Flatten<ArrayRangeSetIter>`) might be more concise than writing out the logic by hand.

quinn-proto/src/range_set/array_range_set.rs
```diff
@@ -0,0 +1,301 @@
...
+                return true;
+            }
+        }
+        false
+    }
+
+    pub fn subtract(&mut self, other: &ArrayRangeSet) {
+        // TODO: This can potentially be made more efficient, since the we know
+        // individual ranges are not overlapping, and the next range must start
+        // where the last one finished
```
Ralith, line 135: s/where/after/

quinn-proto/src/range_set/array_range_set.rs
```diff
@@ -0,0 +1,301 @@
...
+                range.start = x.start;
+                // Modify the range to the start of this range, so that we can
+                // afterwards check for extending to the right
+                #[allow(clippy::almost_swapped)]
+                {
+                    x.start = range.start;
+                }
+                x.end = x.end.max(range.end);
+
+                if x.start >= x.end {
```
Ralith, line 173: Should this be `x.is_empty()`?

quinn-proto/src/range_set/array_range_set.rs
```diff
@@ -0,0 +1,301 @@
...
+                    return result;
+                }
+            }
+
+            // At this point we have handled all parts of the new range which
+            // are in front of the current range. Now we handle everything from
+            // the start of the current range
+
+            if x.start >= range.start && x.end <= range.end {
```
Ralith, line 182: Under what conditions can `x.start >= range.start` be false at this point? If `x.start < range.start` initially, then it looks to me like the case above necessarily clamps `range.start` downward until they're equal.

quinn-proto/src/range_set/array_range_set.rs
```diff
@@ -0,0 +1,301 @@
...
+                    }
+                }
+
+                return true;
+            }
+
+            idx += 1;
+        }
+
+        if x.start == x.end {
```
Ralith, line 208: `is_empty()`?

quinn-proto/src/range_set/array_range_set.rs
```diff
@@ -0,0 +1,301 @@
...
+
+            if x.start >= range.start && x.end <= range.end {
+                // Fully contained
+                return result;
+            } else if x.start <= range.end {
+                // Extend the current range to the end of the new range.
+                // Since it's not contained it must be bigger
+                range.end = x.end;
+
+                // Merge all follow-up ranges which are fully contained
```
Ralith, line 190: s/are fully contained/overlap/

quinn-proto/src/range_set/array_range_set.rs
```diff
@@ -0,0 +1,301 @@
...
+                // Modify the range to the start of this range, so that we can
+                // afterwards check for extending to the right
+                #[allow(clippy::almost_swapped)]
+                {
+                    x.start = range.start;
+                }
+                x.end = x.end.max(range.end);
+
+                if x.start >= x.end {
+                    return result;
```
Ralith, line 174: We assign `result = true` above here; should this be simplified to `return true;` for clarity?

quinn-proto/src/range_set/array_range_set.rs
```diff
@@ -0,0 +1,301 @@
...
+                }
+
+                return true;
+            }
+
+            idx += 1;
+        }
+
+        if x.start == x.end {
+            return result;
```
Ralith, line 209: Should this be `return false;`? The only case where we assign `result = true` is when the new range overlaps the beginning of an existing range, and in that case it looks like we always either judge the remainder wholly contained and return `true`, or extend the current range and return `true`, with control never reaching this point regardless.

quinn-proto/src/range_set/array_range_set.rs
```diff
@@ -0,0 +1,301 @@
...
+        let mut idx = 0;
+        while idx != self.0.len() && x.start != x.end {
+            let range = self.0[idx].clone();
+
+            if x.end <= range.start {
+                // The range is fully before this range
+                return result;
+            } else if x.start >= range.end {
+                // The range is fully after this range
+                idx += 1;
```
Ralith, line 229: If we add `continue;` here, we can move the remaining code out of the `else` case and thereby reduce rightward drift.

quinn-proto/src/range_set/array_range_set.rs
```diff
@@ -0,0 +1,301 @@
...
+                    self.0[idx] = right;
+                    self.0.insert(idx, left);
+                    idx += 2;
+                } else if left_len != 0 {
+                    self.0[idx] = left;
+                    idx += 1;
+                } else {
+                    self.0[idx] = right;
+                    idx += 1;
+                }
```
Ralith, line 275: What's the motivation behind modifying `x` here? I feel like it might be a lot simpler without that. Something like:

```rust
let left = range.start .. x.start;
let right = x.end .. range.end;
if left.is_empty() && right.is_empty() {
    self.0.remove(idx)
} else if left.is_empty() {
    self.0[idx] = right;
    // etc
```

quinn-proto/src/range_set/array_range_set.rs
```diff
@@ -0,0 +1,301 @@
...
+        self.0.is_empty()
+    }
+
+    pub fn pop_min(&mut self) -> Option<Range<u64>> {
+        if !self.0.is_empty() {
+            Some(self.0.remove(0))
+        } else {
+            None
+        }
+    }
```
Ralith, line 292: I wonder if it might be worth reversing the ordering to make this constant-time.

Ralith (changes requested):

quinn-proto/src/range_set/array_range_set.rs
Matthias247: I just found it easier to follow that way, and to figure out step by step what is kept and removed and have comments for everything. I can follow the `remove` method now rather well and have a high amount of confidence in it. The `insert` one is imho uglier.

quinn-proto/src/range_set/array_range_set.rs
Matthias247: I can replace them with `impl Iterator`

quinn-proto/src/range_set/array_range_set.rs
Matthias247: Yes. Doesn't really matter.

quinn-proto/src/range_set/array_range_set.rs
Matthias247: I don't get why the ordering makes any difference here? The length check is constant time anyway, and from there on the compiler will optimize the instruction order it likes best

quinn-proto/src/range_set/array_range_set.rs
Matthias247: I guess that's true. But I don't think its worthwhile describing that this can only be false due to preconditions. It seems easier to just return the variable as is and knowing that it had been set correctly.

quinn-proto/src/range_set/array_range_set.rs
Matthias247: It indeed shouldn't be able to be `false` anymore, since we handled all the associated cases already. Therefore it should be correct to remove that condition. And I think the modification of `range.start` is likely not needed (but also won't hurt)

Matthias247: I actually removed the modifications of `x`. Can't see how they are needed anymore

quinn-proto/src/range_set/array_range_set.rs
Ralith: Fair enough.

quinn-proto/src/range_set/array_range_set.rs
Ralith: The intended upside to using `impl Iterator` is that we can replace the handwritten `Iterator` implementations with combinators, which will be much more concise.

quinn-proto/src/range_set/array_range_set.rs
```diff
@@ -0,0 +1,305 @@
...
+
+                return true;
+            }
+
+            idx += 1;
+        }
+
+        if x.is_empty() {
+            return result;
+        }
```
Ralith, line 208: If `x` is no longer mutable, this branch is redundant to the one at the start of the function.

quinn-proto/src/range_set/array_range_set.rs
Ralith: Reversing the ordering of the ranges stored in the container, I mean. Right now we store from smallest to largest, but that makes `pop_min` linear instead of constant, since larger elements must be moved.

quinn-proto/src/range_set/array_range_set.rs
Ralith: I still feel like the approach I outlined would be much simpler, as it avoids all arithmetic and reduces the number of intermediate values considerably.

quinn-proto/src/range_set/array_range_set.rs
Matthias247: Ah ok. This would also be doable by building a `TinyVecDeque` which allows to remove at the front with zero cost. That's probably easier to follow than reverting the order. 

But since this function is called super rarely (we need to have > 64 ranges), it likely doesn't matter

Ralith (approved):
Thanks!

quinn-proto/src/range_set/array_range_set.rs
```diff
@@ -0,0 +1,274 @@
...
+        ArrayRangeSetEltIter {
+            inner: ArrayRangeSetIter {
+                inner: &self.0,
+                front_idx: 0,
+                back_idx: self.0.len(),
+            },
+            start: 0,
+            end: 0,
+        }
+    }
```
Ralith, line 118: Could the hand-implemented iterators here be replaced with combinators?

Ralith (changes requested):

Ralith (approved):
LGTM modulo lints

Ralith (approved):

djc (approved):

PR #1091, opened 2021-03-29
Fix overflow in `crypto_buffer_size`

daxpedda:
Fix overflowing when doing things like `TransportConfig::crypto_buffer_size(usize::MAX)`.


Ralith (approved):
Thanks! Not exactly a recommended configuration, but it's always good to not have weird internal panics in response to a well-formed configuration.

PR #1093, opened 2021-03-31
crypto: return Option from next_1rtt_keys()

djc:
A crypto session cannot return 1-RTT keys before handshaking is finished.
Currently, rustls contains unwrap() calls to deal with this, but hoisting
out the unwrap to the `Connection` level here seems more sensible.

We can adjust the call into rustls once we update rustls for other reasons.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1591,8 +1591,13 @@ where
         trace!("{:?} keys ready", space);
         if space == SpaceId::Data {
             // Precompute the first key update
-            self.next_crypto = Some(self.crypto.next_1rtt_keys());
+            self.next_crypto = Some(
+                self.crypto
+                    .next_1rtt_keys()
+                    .expect("handshake should be complete"),
+            );
         }
+
         self.spaces[space].crypto = Some(crypto);
         debug_assert!(space as usize > self.highest_space as usize);
         self.highest_space = space;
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -1591,7 +1591,7 @@ where
         trace!("{:?} keys ready", space);
         if space == SpaceId::Data {
             // Precompute the first key update
-            self.next_crypto = Some(self.crypto.next_1rtt_keys());
+            self.next_crypto = self.crypto.next_1rtt_keys();
```
Ralith, line 1594: I'd prefer to do `Some(unwrap())` here, to make clear that the crypto layer returning `None` at this stage would violate our invariants.

quinn-proto/src/connection/mod.rs
```diff
@@ -2788,7 +2788,7 @@ where
         // Generate keys for the key phase after the one we're switching to, store them in
         // `next_crypto`, make the contents of `next_crypto` current, and move the current keys into
         // `prev_crypto`.
-        let new = self.crypto.next_1rtt_keys();
+        let new = self.crypto.next_1rtt_keys().unwrap(); // safe: only called for `Data` packets
```
Ralith, line 2791: I feel like `unwrap()` + explanatory comment should usually be an `expect` with the rationale as a string. Does that seem reasonable you?

Ralith (changes requested):
Seems sensible.

quinn-proto/src/connection/mod.rs
djc: Yup!

Ralith (approved):

PR #1096, opened 2021-03-31
Adds a `poll_connection`.

### LINKED ISSUE #1094: `Connection::poll` order
This can cause an `Event::Connected` to be delivered after an `Event::Stream`. I'm not sure if this is intended behaviour because of 0-rtt, but for my application that doesn't use 0-rtt it means I have to buffer stream events until `Event::Connected` happens. Reversing the order presumably would deliver an `Event::ConnectionLost` before the stream data has finished processing, which is also not ideal. Maybe it would be useful to split `poll` into a  `poll_stream` and `poll_connection` and let the application handle it?

```rust
pub fn poll(&mut self) -> Option<Event> {
    if let Some(event) = self.streams.poll() {
        return Some(Event::Stream(event));
    }
    if let Some(x) = self.events.pop_front() {
        return Some(x);
    }
```

dvc94ch:
Closes #1094

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -331,6 +331,15 @@ where
         self.timers.next_timeout()
     }
 
+    /// Returns application-facing event.
+    ///
+    /// When calling `poll` an `Event::Stream` may be returned before an `Event::Connected`. To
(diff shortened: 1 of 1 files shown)


Ralith:
Thanks for having a go at this! Closing in favor of #1100 which should be more convenient for downstream code.

PR #1098, opened 2021-04-02
Backport flow control fixes

Ralith:
Backports changes from https://github.com/quinn-rs/quinn/pull/1087. We could maybe strip this down to just 
a095410a50e5182aa76e8d8c8acacb5baf1b30b4, but the refactoring makes me more confident in correctness and the more minor fixes along the way shouldn't hurt, and if we have to touch this again it'll be easier without divergence.

I'll prepare a point release once this is merged.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,16 +12,16 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        rust: [stable, beta, 1.45.0]
+        rust: [stable, beta, 1.46.0]
         exclude:
           - os: macos-latest
             rust: beta
           - os: macos-latest
-            rust: 1.45.0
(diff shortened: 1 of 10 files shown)


Ralith:
CI is failing due to style issues fixed in main. Should we backport those too? Seems kind of gratuitous. An alternative would be to disable linting in version branches.

djc:
Actually this also seems to break because of MSRV issues? It's not just the lint job, at least.

Ralith:
Looks like hyper bumped their MSRV to 1.46, and the interop tool is pulling that in. I've added the MSRV bump commit to this backport; we should also consider getting rid of the interop tool since it's not seeing much (any?) use at this point.

djc:
Stupid type limits. If you want to ditch the clippy part of CI on the branch, that sounds good to me.

djc (approved):

PR #1100, opened 2021-04-04
Improve connection event ordering consistency

### LINKED ISSUE #1094: `Connection::poll` order

Ralith:
Ensures that 1-RTT stream events are delivered after `Event::Connected`, without interfering with the placement of connection errors at the end of the event stream.

Fixes #1094. Moots #1096.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -177,6 +177,8 @@ where
     timers: TimerTable,
     /// Number of packets received which could not be authenticated
     authentication_failures: u64,
+    /// Why the connection was lost, if it has been
+    error: Option<ConnectionError>,
 
     //
     // Queued non-retransmittable 1-RTT data
@@ -286,6 +288,7 @@ where
             idle_timeout: config.max_idle_timeout,
             timers: TimerTable::default(),
             authentication_failures: 0,
+            error: None,
 
             path_response: None,
             close: false,
@@ -344,6 +347,10 @@ where
 
         if let Some(x) = self.events.pop_front() {
             return Some(x);
+        }
+
+        if let Some(err) = self.error.take() {
+            return Some(Event::ConnectionLost { reason: err });
         }
 
         None
@@ -1779,7 +1786,7 @@ where
 
         // State transitions for error cases
         if let Err(conn_err) = result {
-            self.events.push_back(conn_err.clone().into());
+            self.error = Some(conn_err.clone());
             self.state = match conn_err {
                 ConnectionError::ApplicationClosed(reason) => State::closed(reason),
                 ConnectionError::ConnectionClosed(reason) => State::closed(reason),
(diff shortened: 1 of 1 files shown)


dvc94ch:
Thanks! this is great. I would have tried implementing this eventually, but at the moment I'm trying to get gso to work, because the perf of quinn using an `async_io::Async<std::net::UdpSocket>` wasn't very impressive. I hope that will help substantially in closing the gap to a plain tcp socket.

djc:
Out of curiosity, what's the performance goal you're trying to hit?

dvc94ch:
The reason I went down the quic rabbit hole this week, was because in the ipfs-embed netsim-embed bitswap benchmark I noticed that an increase in delay completely ruined the throughput benchmarks. It seemed like the window size was responsible for that, but just increasing the tcp buffer size didn't help. Also I thought experimenting with FEC would be useful, but after reading that FEC still requires triggering the exponential congestion control backoff, I'm not so sure about it's utility. I guess there is no way of distinguishing between packet loss due to poor wifi or due to congestion.

The bench takes 1400ms using yamux/noise/tcp but about 7000ms using quic/noise/udp when 0ms of artificial delay is used, so the target would be about 4x faster.

`netsim_embed` doesn't seem to like gso, not sure why. I get i/o error 5 which I don't get when not using a TUN adapter.

Ralith:
Perhaps the tun device doesn't implement GSO? What does `ethtool -k tun0 |grep generic-segmentation-offload` say?

dvc94ch:
> Perhaps the tun device doesn't implement GSO? What does ethtool -k tun0 |grep generic-segmentation-offload say?

that results in a "no device matches name". I guess because it operates at the ip and not the ethernet layer?

there seems to be a `tunsetoffload` ioctl that can be used to enable the `TUN_F_CSUM` option. this solves the error on send, but results in `EAGAIN` on receive. my guess about what is going on is that I need to compute the udp checksums myself, and that invalid checksums are causing the socket to be readable but then not delivering any packets because of invalid checksums. I'll test my theory tomorrow by checking the checksums are valid and if not correcting them.

djc (approved):

dvc94ch:
So it turns out I was using unoptimized crypto :facepalm: It's 3x faster now, in the ballpark of reasonably fast.

Ralith:
Awesome, thanks for the update! Crypto is definitely expected to be a bottleneck; we saw major benefits tweaking rustls to use a hardware-accelerated cipher (AES) in our own benchmarking. GSO could definitely help close the remaining gap,.

dvc94ch:
The numbers were with gso. Gso by itself didn't help as it was limited by the crypto, not the socket.

Ralith:
Ah, glad you got that working!

PR #1103, opened 2021-04-18
Allow read_handshake to return keys.

dvc94ch:
This allows `quinn-noise` to work.

PR #1104, opened 2021-04-23
Allow returning keys before the next frame.

dvc94ch:
Superceeds #1103

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1569,7 +1569,7 @@ where
                     _ => unreachable!("got updated secrets during 1-RTT"),
                 }
             }
-            if outgoing.is_empty() {
+            if outgoing.is_empty() && space == self.highest_space {
                 break;
             }
             let offset = self.spaces[space].crypto_offset;
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1569,8 +1569,13 @@ where
                     _ => unreachable!("got updated secrets during 1-RTT"),
                 }
             }
-            if outgoing.is_empty() && space == self.highest_space {
-                break;
+            if outgoing.is_empty() {
+                if space == self.highest_space {
+                    break;
+                } else {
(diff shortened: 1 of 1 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -1569,7 +1569,7 @@ where
                     _ => unreachable!("got updated secrets during 1-RTT"),
                 }
             }
-            if outgoing.is_empty() {
+            if outgoing.is_empty() && space == self.highest_space {
                 break;
             }
```
Ralith, line 1574:
```suggestion
            if outgoing.is_empty() {
                if space == self.highest_space {
                    break;
                } else {
                    // Keys updated, check for another frame
                    continue;
                }
            }
```

Ralith (changes requested):
Thanks! As written I think this will produce unnecessary zero-length CRYPTO frames, so one minor tweak.

Ralith (approved):

dvc94ch:
One more potential change would be adding an api for specifiying a custom version. Any opinions on what the best way to make sure that quinn-noise uses a different quic version than quic-tls?

Ralith:
We should definitely support that for this case. I think a decent minimal solution would be to add something like `override_version: Option<u32>`, or `Option<RangeInclusive<u32>>`, to `EndpointConfig`. A better solution would make it harder to inadvertently run an endpoint with a standard version but a nonstandard configuration, but nothing immediately occurs to me.

dvc94ch:
But we probably want to support all ietf versions. So maybe we can change the `0xff` version prefix to something else?

Ralith:
I'm not sure I follow. The 0xff prefix in IETF draft versions is part of the version; a non-standard QUIC variant would not be an IETF draft. Quinn doesn't support all IETF draft versions because support for long-obsolete drafts is both difficult and not particularly useful. If you want to integrate breaking post-v1 changes to IETF QUIC into a non-standard QUIC variant, you'd probably do so by allocating an additional non-standard version.

PR #1106, opened 2021-04-23
Allow overriding supported versions in EndpointConfig.

### LINKED ISSUE #719: (Pie in the sky) Supporting Noise as the cryptographic protocol
We have an internal abstraction over the cryptographic protocol used called the [`crypto::Session`](https://github.com/djc/quinn/blob/3ec36129829d4b871438de9aa942f38a5d63c4eb/quinn-proto/src/crypto.rs#L32). (Note that this trait is currently private, so you'd have to make this public to start work. We'd be happy to merge that, but would likely prefer doing so when something materializes that actually uses that.)

Currently we have a [single implementation for rustls](https://github.com/djc/quinn/blob/3ec36129829d4b871438de9aa942f38a5d63c4eb/quinn-proto/src/crypto/rustls.rs#L41). In QUIC v1, only TLS 1.3 is officially supported, so supporting something else is technically outside of the spec and can thus probably only be made to work between two consenting endpoints.

If we wanted to support Noise, that would likely start by writing an implementation of this thread that is based on Noise. I think there are some crates that implement (parts of) Noise -- [snow](https://lib.rs/crates/snow) probably makes for a good start.

There was a [2018 attempt](https://github.com/rot256/ninn) at integrating Noise support into a much earlier version of Quinn. Quinn looks basically nothing like it did at the time, but maybe there are still useful bits to look at there.

If you want to work on this, we're usually available for questions on our Gitter channel -- just commenting in this issue should also work fine.

dvc94ch:
Allows configuring the supported versions in `EndpointConfig`.

Closes #719

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -367,7 +367,7 @@ where
         self.max_udp_payload_size.into()
     }
 
-    /// Override the supported quic versions.
+    /// Override supported QUIC versions
     pub fn supported_versions(
         &mut self,
         supported_versions: Vec<u32>,
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -1,4 +1,6 @@
-use std::{convert::TryInto, fmt, num::TryFromIntError, sync::Arc, time::Duration};
+use std::{
+    convert::TryInto, fmt, num::TryFromIntError, ops::RangeInclusive, sync::Arc, time::Duration,
+};
 
 use rand::RngCore;
 use thiserror::Error;
@@ -9,7 +11,7 @@ use crate::{
     cid_generator::{ConnectionIdGenerator, RandomConnectionIdGenerator},
     congestion,
     crypto::{self, ClientConfig as _, HandshakeTokenKey as _, HmacKey as _, ServerConfig as _},
-    VarInt, VarIntBoundsExceeded,
+    VarInt, VarIntBoundsExceeded, DEFAULT_VERSION,
 };
 
 /// Parameters governing the core QUIC state machine
@@ -299,6 +301,7 @@ where
     /// Create a cid generator for local cid in Endpoint struct
     pub(crate) connection_id_generator_factory:
         Arc<dyn Fn() -> Box<dyn ConnectionIdGenerator> + Send + Sync>,
+    pub(crate) supported_versions: RangeInclusive<u32>,
 }
 
 impl<S> EndpointConfig<S>
@@ -313,6 +316,7 @@ where
             reset_key: Arc::new(reset_key),
             max_udp_payload_size: 1480u32.into(), // Typical internet MTU minus IPv4 and UDP overhead, rounded up to a multiple of 8
             connection_id_generator_factory: Arc::new(cid_factory),
+            supported_versions: DEFAULT_VERSION,
         }
     }
 
@@ -362,6 +366,12 @@ where
     pub fn get_max_udp_payload_size(&self) -> u64 {
         self.max_udp_payload_size.into()
     }
+
+    /// Override the supported quic versions.
+    pub fn supported_versions(&mut self, supported_versions: RangeInclusive<u32>) -> &mut Self {
+        self.supported_versions = supported_versions;
+        self
+    }
 }
 
 impl<S: crypto::Session> fmt::Debug for EndpointConfig<S> {
@@ -370,6 +380,7 @@ impl<S: crypto::Session> fmt::Debug for EndpointConfig<S> {
             .field("reset_key", &"[ elided ]")
             .field("max_udp_payload_size", &self.max_udp_payload_size)
             .field("cid_generator_factory", &"[ elided ]")
+            .field("supported_versions", &self.supported_versions)
             .finish()
     }
 }
@@ -391,6 +402,7 @@ impl<S: crypto::Session> Clone for EndpointConfig<S> {
             reset_key: self.reset_key.clone(),
             max_udp_payload_size: self.max_udp_payload_size,
             connection_id_generator_factory: self.connection_id_generator_factory.clone(),
+            supported_versions: self.supported_versions.clone(),
         }
     }
 }
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -1,6 +1,4 @@
-use std::{
-    convert::TryInto, fmt, num::TryFromIntError, ops::RangeInclusive, sync::Arc, time::Duration,
-};
+use std::{convert::TryInto, fmt, num::TryFromIntError, sync::Arc, time::Duration};
 
 use rand::RngCore;
 use thiserror::Error;
@@ -11,7 +9,7 @@ use crate::{
     cid_generator::{ConnectionIdGenerator, RandomConnectionIdGenerator},
     congestion,
     crypto::{self, ClientConfig as _, HandshakeTokenKey as _, HmacKey as _, ServerConfig as _},
-    VarInt, VarIntBoundsExceeded, DEFAULT_VERSION,
+    VarInt, VarIntBoundsExceeded, DEFAULT_SUPPORTED_VERSIONS,
 };
 
 /// Parameters governing the core QUIC state machine
@@ -301,7 +299,8 @@ where
     /// Create a cid generator for local cid in Endpoint struct
     pub(crate) connection_id_generator_factory:
         Arc<dyn Fn() -> Box<dyn ConnectionIdGenerator> + Send + Sync>,
-    pub(crate) supported_versions: RangeInclusive<u32>,
+    pub(crate) supported_versions: Vec<u32>,
+    pub(crate) initial_version: u32,
 }
 
 impl<S> EndpointConfig<S>
@@ -316,7 +315,8 @@ where
             reset_key: Arc::new(reset_key),
             max_udp_payload_size: 1480u32.into(), // Typical internet MTU minus IPv4 and UDP overhead, rounded up to a multiple of 8
             connection_id_generator_factory: Arc::new(cid_factory),
-            supported_versions: DEFAULT_VERSION,
+            initial_version: *DEFAULT_SUPPORTED_VERSIONS.start(),
+            supported_versions: DEFAULT_SUPPORTED_VERSIONS.into_iter().collect(),
         }
     }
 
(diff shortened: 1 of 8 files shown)


quinn-proto/src/config.rs
```diff
@@ -299,6 +301,7 @@ where
     /// Create a cid generator for local cid in Endpoint struct
     pub(crate) connection_id_generator_factory:
         Arc<dyn Fn() -> Box<dyn ConnectionIdGenerator> + Send + Sync>,
+    pub(crate) supported_versions: RangeInclusive<u32>,
```
djc, line 304: Hmm, not sure a range is expressive enough. Maybe just `Vec<u32>` or `&'static [u32]`?

quinn-proto/src/connection/mod.rs
```diff
@@ -643,6 +656,7 @@ where
                 (num_datagrams - 1) * (self.path.mtu as usize),
                 ack_eliciting,
                 self,
+                *self.supported_versions.start(),
```
djc, line 659: I think we should probably also make the configuration about which is the explicit/initiating version.

quinn-proto/src/connection/mod.rs
```diff
@@ -1643,7 +1657,11 @@ where
         self.path.total_recvd = self.path.total_recvd.saturating_add(data.len() as u64);
         let mut remaining = Some(data);
         while let Some(data) = remaining {
-            match PartialDecode::new(data, self.local_cid_state.cid_len()) {
+            match PartialDecode::new(
+                data,
+                self.local_cid_state.cid_len(),
+                self.supported_versions.clone(),
```
djc, line 1663: Cloning this here doesn't seem great...

quinn-proto/src/lib.rs
```diff
@@ -150,12 +149,7 @@ pub mod fuzzing {
 }
 
 /// The QUIC protocol version implemented
-const VERSION: std::ops::RangeInclusive<u32> = 0xff00_001d..=0xff00_0020;
-
-/// Whether a 4-byte slice represents a supported version number
-fn is_supported_version(x: &[u8]) -> bool {
-    VERSION.contains(&u32::from_be_bytes(x.try_into().unwrap()))
-}
+const DEFAULT_VERSION: std::ops::RangeInclusive<u32> = 0xff00_001d..=0xff00_0020;
```
djc, line 152: This name implies a single version, when it's actually a range. Maybe `DEFAULT_SUPPORTED_VERSIONS`?

djc (commented):
Thanks for working on this! Some things to think about:

quinn-proto/src/config.rs
```diff
@@ -362,6 +366,19 @@ where
...
+        self.supported_versions = supported_versions;
+        self.initial_version = self.supported_versions[0];
+        self
+    }
+
+    /// Override the initial version.
+    pub fn initial_version(&mut self, initial_version: u32) -> &mut Self {
+        self.initial_version = initial_version;
+        self
+    }
```
dvc94ch, line 381: Collapse into a single function to prevent missuse

quinn-proto/src/endpoint.rs
```diff
@@ -166,48 +166,51 @@ where
         data: BytesMut,
     ) -> Option<(ConnectionHandle, DatagramEvent<S>)> {
         let datagram_len = data.len();
-        let (first_decode, remaining) =
```
dvc94ch, line 169: this large diff was caused by rustfmt :/

quinn-proto/src/connection/mod.rs
```diff
@@ -2074,7 +2095,11 @@ where
                 if self.total_authed_packets > 1 {
                     return Ok(());
                 }
-                if packet.payload.chunks(4).any(is_supported_version) {
+                let supported = packet.payload.chunks(4).any(|x| {
+                    self.supported_versions
+                        .contains(&u32::from_be_bytes(x.try_into().unwrap()))
+                });
+                if supported {
```
dvc94ch, line 2102: I can add back the `is_supported_version` utility.

quinn-proto/src/packet.rs
```diff
@@ -124,6 +125,7 @@ impl PartialDecode {
                     src_cid,
                     token,
                     number,
+                    version: Default::default(),
```
dvc94ch, line 128: since the same enum is used for encoding/decoding but the version is only needed for encoding this is a little uggly.

dvc94ch (commented):
@Ralith wdyt?

quinn-proto/src/connection/mod.rs
Ralith: Nah, this is fine.

quinn-proto/src/connection/mod.rs
```diff
@@ -214,6 +214,10 @@ where
     datagrams: DatagramState,
     /// Connection level statistics
     stats: ConnectionStats,
+    /// Supported versions.
+    supported_versions: Vec<u32>,
+    /// Initial version.
+    initial_version: u32,
```
Ralith, line 220: Style nits:
- Sentence fragments shouldn't end in periods.
- Comments should add useful information. If it just repeats the name of the variable, don't waste the space.

Also, I think we can replace both of these fields with
```rust
/// QUIC version used for the connection
version: u32
```
and pass `&[self.version]` in place of `&self.supported_versions`, as it's impossible for the version used by a connection to change.

quinn-proto/src/config.rs
```diff
@@ -362,6 +366,20 @@ where
     pub fn get_max_udp_payload_size(&self) -> u64 {
         self.max_udp_payload_size.into()
     }
+
+    /// Override the supported quic versions.
```
Ralith, line 370:
```suggestion
    /// Override supported QUIC versions
    ///
    /// `initial_version` is used for outgoing connections and version negotiation for incoming connections.
```

quinn-proto/src/packet.rs
Ralith: Rather than making something up here and below, let's pipe the packet's actual version number through `PlainHeader` in each applicable case, here and below. Aside from being cleaner, this may come in handy when we implement version negotiation.

Ralith (changes requested):

quinn-proto/src/connection/mod.rs
dvc94ch: it seems like version negotiation is (or would be) part of the connection as it handles version negotiation packets. So I'm not sure about this change.

dvc94ch: did the change anyway, but I can revert it if you change your mind

quinn-proto/src/packet.rs
Ralith: Looks like you missed this one?

Ralith (changes requested):
We can always add additional state to `Connection` for version negotiation if/when it becomes necessary; QUIC 1 doesn't actually specify a complete version negotiation mechanism at all.

Ralith (approved):
Thanks!

quinn-proto/src/connection/mod.rs
```diff
@@ -2074,7 +2087,11 @@ where
                 if self.total_authed_packets > 1 {
                     return Ok(());
                 }
-                if packet.payload.chunks(4).any(is_supported_version) {
+                let supported = packet
+                    .payload
+                    .chunks(4)
+                    .any(|x| self.version == u32::from_be_bytes(x.try_into().unwrap()));
```
djc, line 2093: Let's try avoiding the unwrap here?

```suggestion
                    .any(|x| match [u32; 4]::try_from(x) {
                        Ok(version) => self.version == version,
                        Err(_) => false,
                    })
```

quinn-proto/src/config.rs
```diff
@@ -313,6 +315,8 @@ where
             reset_key: Arc::new(reset_key),
             max_udp_payload_size: 1480u32.into(), // Typical internet MTU minus IPv4 and UDP overhead, rounded up to a multiple of 8
             connection_id_generator_factory: Arc::new(cid_factory),
+            initial_version: *DEFAULT_SUPPORTED_VERSIONS.start(),
```
djc, line 318: Shouldn't we default to the latest supported version rather than the oldest?

quinn-proto/src/lib.rs
```diff
@@ -149,13 +148,8 @@ pub mod fuzzing {
...
 
-/// The QUIC protocol version implemented
-const VERSION: std::ops::RangeInclusive<u32> = 0xff00_001d..=0xff00_0020;
-
-/// Whether a 4-byte slice represents a supported version number
-fn is_supported_version(x: &[u8]) -> bool {
-    VERSION.contains(&u32::from_be_bytes(x.try_into().unwrap()))
-}
+/// The QUIC protocol version implemented.
+pub const DEFAULT_SUPPORTED_VERSIONS: std::ops::RangeInclusive<u32> = 0xff00_001d..=0xff00_0020;
```
djc, line 152: Here too, I feel like we should just use a `Vec` rather than a contiguous `Range`.

quinn-proto/src/config.rs
Ralith: This is preserving the current behavior, which was chosen to maximize compatibility with other draft implementations. We might want to revisit that soon, but I think it's appropriate to leave alone here.

djc: Fair enough!

quinn-proto/src/connection/mod.rs
Ralith: Or maybe `x.try_into().map_or(false, u32::from_be_bytes)`?

djc: I'm fine with that, too, though it seems a little harder to read to me.

Ralith: Ah, yeah, I guess the `match` is a bit less obscure even once you add the `u32::from_be_bytes`.

quinn-proto/src/lib.rs
dvc94ch: heap allocations are not allowed in constants, and making it static requires a clone, which I don't think adds much advantage.

Ralith: A constant slice wouldn't require a heap allocation.

djc (approved):
Thanks!

PR #1112, opened 2021-05-04
Maximize `CRYPTO` data

Matthias247:
This change increases the amount of `CRYPTO` data sent in packets during handshakes.

Instead of reserving a fixed amount of 17 bytes for frame overhead,
we reserve the exactly required amount of bytes to encode frame type
and offset, and only 2 bytes for the frame length. This leads to an
additional 12-13 bytes of payload data being used.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -2620,10 +2620,23 @@ where
                 Some(x) => x,
                 None => break,
             };
-            let len = cmp::min(
-                frame.data.len(),
-                max_size as usize - buf.len() - frame::Crypto::SIZE_BOUND,
-            );
+
+            // Calculate the maximum amount of crypto data we can store in the buffer.
+            // Since the offset is known, we can reserve the exact size required to encode it.
+            // For length we reserve 2bytes which allows to encode up to 2^14,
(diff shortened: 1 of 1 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -2620,10 +2620,23 @@ where
...
-            );
+
+            // Calculate the maximum amount of crypto data we can store in the buffer.
+            // Since the offset is known, we can reserve the exact size required to encode it.
+            // For length we reserve 2bytes which allows to encode up to 2^14,
+            // which is more than what fits into normally sized QUIC frames.
+            let max_crypto_data_size = max_size
+                - buf.len()
+                - 1 // Frame Type
+                - VarInt::size(unsafe { VarInt::from_u64_unchecked(frame.offset) })
```
Ralith, line 2631: Hmm, we should use `VarInt` more pervasively in frame structures.

Ralith (approved):
Nice!

quinn-proto/src/connection/mod.rs
Matthias247: Agreed. But it would be better as a separate change

PR #1113, opened 2021-05-06
Add generic connection support

LEXUGE:
This PR adds:
1. Nix flake which is used to create a reproducible developing environment (can be cherry-picked out if not appropriate)
2. Generic connection abstraction (i.e. support modify the raw packets before sending, support other connections like ICMP).

Though the connection is defined to be UDP per IETF draft, this PR still adds flexibility to users who want to somehow use other connection types and take the well-built foundation  of the `quinn`, but not do all the works starting from `quinn-proto`.

This is a breaking change, but the change is minimal (all current API callings are working, except the return type of `rebind` has been changed).

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,93 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1614513358,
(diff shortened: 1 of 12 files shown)


djc:
This seems to add a bunch of complexity for what seems like a very niche use case. I think for your use case it might make more sense if you just rely on quinn-proto directly (potentially by forking quinn).

LEXUGE:
Yes, I believe it's better to be a fork. I am closing now

PR #1115, opened 2021-05-08
Use an `ArrayRangeSet` to store new ACKs

Matthias247:
This seems slightly more efficient than extending a `Vec`.
Would probably be even more efficient if there would be an intersection
method on `SentMap` insert of returning packet numbers one by one.

Peak perf before:
```
Sent 1073741824 bytes on 1 streams in 1.77s (577.02 MiB/s)
```

After:
```
Sent 1073741824 bytes on 1 streams in 1.75s (584.25 MiB/s)
```


djc (dismissed):

djc (approved):

Ralith (approved):
Nice!

PR #1117, opened 2021-05-09
Optimize VarInt decoding

Matthias247:
Upfront: This isn't the most elegant change. But given VarInts sit in the
hot path it seems worth it.

The current VarInt decoding performs some duplicate bounds check.
`VarInt::decode` checks for available bytes in order to return an error.
And `bytes::Buf` performs additional bounds checks in various helper methods.

By using `Buf::chunk` and `Buf::advance` directly we can save some of those.
We could improve even more by only support decoding from contiguous buffers,
but that would change the API contract.

In addition it seemed like the `u16` be->le conversion wasn't as free
as it should be. A 2byte LE buffer improves things here.

Overall this reduces CPU time spent in VarInt decoding from about 2% to around 1%.

### DIFF
--- a/quinn-proto/src/varint.rs
+++ b/quinn-proto/src/varint.rs
@@ -150,7 +150,11 @@ impl Codec for VarInt {
             return Err(UnexpectedEnd);
         }
         let mut buf = [0; 8];
-        buf[0] = r.get_u8();
+        // This does not make use of `r.get_u8()` to avoid the additional bounds
+        // check. We already performed this one earlier on.
+        buf[0] = r.chunk()[0];
+        r.advance(1);
+
         let tag = buf[0] >> 6;
         buf[0] &= 0b0011_1111;
         let x = match tag {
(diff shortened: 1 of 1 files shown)


Matthias247:
For anyone who is more curious. Here is a baseline flamegraph. "Matched" is the amount of samples in `VarInt::decode` (2.6%).
Inside the method we can also see 0.68% time spent in `swap_bytes`.

grafik

Here is the changed version, with 1.2% of samples in `VarInt::decode`:

grafik

quinn-proto/src/varint.rs
```diff
@@ -150,7 +150,11 @@ impl Codec for VarInt {
             return Err(UnexpectedEnd);
         }
         let mut buf = [0; 8];
-        buf[0] = r.get_u8();
+        // This does not make use of `r.get_u8()` to avoid the additional bounds
+        // check. We already performed this one earlier on.
+        buf[0] = r.chunk()[0];
+        r.advance(1);
+
```
Ralith, line 157: Does this specific change actually help? I wouldn't expect this to have any difference in bounds checking semantics.

Ralith (commented):
Should we have a microbenchmark for this path?

quinn-proto/src/varint.rs
Matthias247: There's an `assert!` in the `get_u8`, not just a rust array bounds check. I don't think those are removed. And since even without the "le" change this proved to be faster, I guess practical experience shows that it helps.

Ralith (approved):

djc (commented):
Uh, left this sitting around for a long time. @Matthias247 what do you think?

PR #1119, opened 2021-05-13
Improve fairness of endpoint task

Matthias247:
This contains 2 commits which in combination try to prevent the endpoint task from using too many executor resources while keeping performance or event improving on it.

Performance difference various a bit between the machinesI tested. On one there is no change, on another an increase in througput:

**Before:**
```
Sent 10737418240 bytes on 1 streams in 22.50s (455.04 MiB/s)
```

**After:**
```
Sent 10737418240 bytes on 1 streams in 21.92s (467.06 MiB/s)
```

### DIFF
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -155,8 +155,8 @@ enum EndpointEvent {
     Transmit(proto::Transmit),
 }
 
-/// Maximum number of send/recv calls to make before moving on to other processing
+/// Maximum number of datagrams processed in send/recv calls to make before moving on to other processing
 ///
-/// This helps ensure we don't starve anything when the CPU is slower than the link. Value selected
-/// more or less arbitrarily.
-const IO_LOOP_BOUND: usize = 10;
+/// This helps ensure we don't starve anything when the CPU is slower than the link.
+/// Value is selected by picking a low number which didn't degrade throughput in benchmarks.
+const IO_LOOP_BOUND: usize = 160;
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -200,28 +200,33 @@ where
 
     #[allow(unused_mut)] // MSRV
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
-        let endpoint = &mut *self.0.lock().unwrap();
+        let mut endpoint = self.0.lock().unwrap();
         if endpoint.driver.is_none() {
             endpoint.driver = Some(cx.waker().clone());
         }
-        loop {
-            let now = Instant::now();
-            let mut keep_going = false;
-            keep_going |= endpoint.drive_recv(cx, now)?;
-            endpoint.handle_events(cx);
-            keep_going |= endpoint.drive_send(cx)?;
-            if !keep_going {
-                break;
-            }
-        }
+
+        let now = Instant::now();
+        let mut keep_going = false;
+        keep_going |= endpoint.drive_recv(cx, now)?;
+        endpoint.handle_events(cx);
+        keep_going |= endpoint.drive_send(cx)?;
+
         if !endpoint.incoming.is_empty() {
             if let Some(task) = endpoint.incoming_reader.take() {
                 task.wake();
             }
         }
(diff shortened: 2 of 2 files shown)


quinn/src/lib.rs
```diff
@@ -155,8 +155,9 @@ enum EndpointEvent {
...
 
-/// Maximum number of send/recv calls to make before moving on to other processing
+/// Maximum number of datagrams processed in send/recv calls to make before moving on to other processing
 ///
-/// This helps ensure we don't starve anything when the CPU is slower than the link. Value selected
-/// more or less arbitrarily.
-const IO_LOOP_BOUND: usize = 10;
+/// This helps ensure we don't starve anything when the CPU is slower than the link.
+/// Value is selected by picking a low number which didn't degrade throughput in benchmarks.
+/// It is also a multiple of `platform::BATCH_SIZE`
```
Ralith, line 162: I'm not sure what the last sentence here means. The actual value is a constant independent of `platform::BATCH_SIZE`, and there doesn't seem to be a strict need for it to be otherwise. Is it meant to express that the *effective* limit may differ from the constant by rounding up to the next batch size?

Ralith (approved):

quinn/src/lib.rs
Matthias247: I guess I can follow that its not obvious. The idea in having a multiple is to say "we send exactly N batches", and not N.2 batches. I first had `IO_LOOP_BOUND: usize = BATCH_SIZE * 5`, but that is obviously wrong for `BATCH_SIZE == 1`. 

Then I wondered whether I should redefine `BATCH_SIZE` on top of `IO_LOOP_BOUND`, but that is also wrong. So this just ended up being a comment. I guess we can also remove it

Ralith: Since we break after sending `>=` the bound, I think it's fair to say that the effective size is rounded up to the next multiple of the batch size? Leaving it unstated is fine too, I don't think this is a property that specifically requires documentation.

Matthias247: ok. Removed

Ralith (approved):

djc:
Sorry for not reviewing this sooner -- I was on holiday for two weeks. The reason I hadn't approved yet was because I was wondering about `handle_events()`. As I understand it, it is not currently limited by `IO_BOUND_LOOP`. Should we worry about a scenario where there are a lot of events to handle and we get stuck inside `handle_events()` for too long? I guess that's pretty much orthogonal to this PR, but still wanted to mention it.

Matthias247:
> Should we worry about a scenario where there are a lot of events to handle and we get stuck inside handle_events() for too long? I guess that's pretty much orthogonal to this PR, but still wanted to mention it.

Yes, I think it also makes sense to limit those interactions. It however won't be an issue with a singlethreaded runtime since connections can't produce new events while the endpoint task is running, so it's guaranteed to terminate. The IO part was more risky since it gets events from the outside, and a peer sending data fast enough could prevent the endpoint task from ever yielding.

However for better performance with multithreaded runtimes it could make sense to limit the amount of events handled in `handle_events()`. As you said, it's kind of orthogonal.

Ralith:
The expected rate of endpoint events is *much* lower than the expected rate of incoming/outgoing packets, FWIW. They're only raised during connection lifecycle events like handshaking, migration, and loss.

Matthias247:
[`Endpoint::handle_events`](https://github.com/quinn-rs/quinn/blob/fa0ac1f85a70867c3ecb45ba1e778cd29db4d98f/quinn/src/endpoint.rs#L371-L401) does not only handle the endpoint related events. It receives all the outgoing messages from connections and enqueues them for transmission. This could theoretically lead to an unbounded `outgoing`  queue.

Another potential issue I noticed is in [`drive_send`](https://github.com/quinn-rs/quinn/blob/fa0ac1f85a70867c3ecb45ba1e778cd29db4d98f/quinn/src/endpoint.rs#L339-L344)  endpoint related datagrams are only enqueued if not enough other packets are enqueued. That could mean that in a busy endpoint those might never get sent. Probably that condition should just be dropped.

PR #1120, opened 2021-05-14
Avoid allocations for CIDs

Matthias247:
When reading CIDs from packets the current code makes some short-lived
allocations due to the use of `copy_to_bytes`, which allocate for 1.5% of
CPU samples. This change avoids this.

**Before:**
```
Sent 1073741824 bytes on 1 streams in 1.72s (594.74 MiB/s)
```

**After:**
```
Sent 1073741824 bytes on 1 streams in 1.70s (602.78 MiB/s)
```

### DIFF
--- a/quinn-proto/src/shared.rs
+++ b/quinn-proto/src/shared.rs
@@ -75,7 +75,20 @@ impl ConnectionId {
             len: bytes.len() as u8,
             bytes: [0; MAX_CID_SIZE],
         };
-        res.bytes[..bytes.len()].clone_from_slice(&bytes);
+        res.bytes[..bytes.len()].copy_from_slice(bytes);
+        res
+    }
+
+    /// Constructs cid by reading `len` bytes from a `Buf`
+    ///
+    /// Callers need to assure that `buf.remaining() >= len`
+    pub(crate) fn from_buf(buf: &mut impl Buf, len: usize) -> Self {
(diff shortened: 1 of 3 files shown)


quinn-proto/src/shared.rs
```diff
@@ -75,17 +75,31 @@ impl ConnectionId {
...
         res
     }
 
     /// Decode from long header format
     pub(crate) fn decode_long(buf: &mut impl Buf) -> Option<Self> {
         let len = buf.get::<u8>().ok()? as usize;
-        match len > MAX_CID_SIZE || buf.remaining() < len {
-            false => Some(ConnectionId::new(&buf.copy_to_bytes(len))),
-            true => None,
+        if len > MAX_CID_SIZE || buf.remaining() < len {
```
djc, line 98: The style change here seems unrelated?

djc (commented):
Seems nice!

quinn-proto/src/shared.rs
Matthias247: Kind of. There was first more code in the `Some` branch, and I moved it out then because I know you prefer early returns instead of big if/else blocks. Then I added the delegate but kept it. I can change it if you prefer

djc: Makes sense. I do prefer if you change it back.

djc (approved):

djc:
Thanks!

PR #1122, opened 2021-05-14
feat(cca): cubic

FrankSpitulski:
mostly copied from quiche, modified to fit here

still working out the best way to get a ref to the current RTT inside the CCA

Relates to #693

### DIFF
--- a/bench/src/bulk.rs
+++ b/bench/src/bulk.rs
@@ -274,6 +274,7 @@ fn transport_config(opt: &Opt) -> quinn::TransportConfig {
     config
         .max_concurrent_uni_streams(opt.max_streams as u64)
         .unwrap();
+    config.congestion_controller_factory(Arc::new(quinn_proto::congestion::CubicConfig::default()));
     config
 }
 
--- a/bench/src/bulk.rs
+++ b/bench/src/bulk.rs
@@ -274,7 +274,6 @@ fn transport_config(opt: &Opt) -> quinn::TransportConfig {
     config
         .max_concurrent_uni_streams(opt.max_streams as u64)
         .unwrap();
-    config.congestion_controller_factory(Arc::new(quinn_proto::congestion::CubicConfig::default()));
     config
 }
 
--- a/quinn-proto/src/congestion/cubic.rs
+++ b/quinn-proto/src/congestion/cubic.rs
@@ -97,7 +97,7 @@ impl Controller for Cubic {
         sent: Instant,
         bytes: u64,
         app_limited: bool,
-        min_rtt: Duration,
+        rtt: Duration,
     ) {
         if app_limited
             || self
@@ -143,12 +143,12 @@ impl Controller for Cubic {
             // w_cubic(t + rtt)
             let w_cubic = self
                 .cubic_state
-                .w_cubic(t + min_rtt, self.config.max_datagram_size);
+                .w_cubic(t + rtt, self.config.max_datagram_size);
 
             // w_est(t)
             let w_est = self
                 .cubic_state
-                .w_est(t, min_rtt, self.config.max_datagram_size);
+                .w_est(t, rtt, self.config.max_datagram_size);
 
             let mut cubic_cwnd = self.window;
 
--- a/quinn-proto/src/congestion/cubic.rs
+++ b/quinn-proto/src/congestion/cubic.rs
@@ -39,7 +39,7 @@ impl State {
     // K = cbrt(w_max * (1 - beta_cubic) / C) (Eq. 2)
     fn cubic_k(&self, max_datagram_size: u64) -> f64 {
         let w_max = self.w_max / max_datagram_size as f64;
-        libm::cbrt(w_max * (1.0 - BETA_CUBIC) / C)
+        (w_max * (1.0 - BETA_CUBIC) / C).cbrt()
     }
 
     // W_cubic(t) = C * (t - K)^3 - w_max (Eq. 1)
--- a/quinn-proto/src/congestion/cubic.rs
+++ b/quinn-proto/src/congestion/cubic.rs
@@ -1,0 +1,262 @@
+use std::sync::Arc;
+use std::time::{Instant, Duration};
+
+use super::{Controller, ControllerFactory};
+use std::cmp;
+
+
+/// CUBIC Constants.
+///
+/// These are recommended value in RFC8312.
+const BETA_CUBIC: f64 = 0.7;
+
+const C: f64 = 0.4;
+
+/// CUBIC State Variables.
+///
+/// We need to keep those variables across the connection.
+/// k, w_max, w_last_max is described in the RFC.
+#[derive(Debug, Default, Clone)]
+pub struct State {
+    k: f64,
+
+    w_max: f64,
+
+    w_last_max: f64,
+
+    // Used in CUBIC fix (see on_packet_sent())
+    last_sent_time: Option<Instant>,
+
+    // Store cwnd increment during congestion avoidance.
+    cwnd_inc: u64,
+}
+
+/// CUBIC Functions.
+///
+/// Note that these calculations are based on a count of cwnd as bytes,
+/// not packets.
+/// Unit of t (duration) and RTT are based on seconds (f64).
+impl State {
+    // K = cbrt(w_max * (1 - beta_cubic) / C) (Eq. 2)
+    fn cubic_k(&self, max_datagram_size: u64) -> f64 {
+        let w_max = self.w_max / max_datagram_size as f64;
+        libm::cbrt(w_max * (1.0 - BETA_CUBIC) / C)
+    }
+
+    // W_cubic(t) = C * (t - K)^3 - w_max (Eq. 1)
+    fn w_cubic(&self, t: Duration, max_datagram_size: u64) -> f64 {
+        let w_max = self.w_max / max_datagram_size as f64;
+
+        (C * (t.as_secs_f64() - self.k).powi(3) + w_max) *
+            max_datagram_size as f64
+    }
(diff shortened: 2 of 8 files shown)


Ralith:
Thanks for taking a swing at this! I think it'd be fine to pass the current RTT to be made an argument to whatever `congestion::Controller` methods need it. The current interface for `Controller` isn't expected to be comprehensive, just a starting point to grow as needed.

FrankSpitulski:
@Ralith thanks for the suggestion, I've pushed that impl up.

quinn-proto/src/connection/mod.rs
```diff
@@ -1162,9 +1162,13 @@ where
...
             // path, so as to ignore any ACKs from older paths still coming in.
-            self.path
-                .congestion
-                .on_ack(now, info.time_sent, info.size.into(), self.app_limited);
+            self.path.congestion.on_ack(
+                now,
+                info.time_sent,
+                info.size.into(),
+                self.app_limited,
+                self.path.rtt.get(),
```
Matthias247, line 1170: The parameter is called `min_rtt`, but this passes the smoothed RTT. Is this what is supposed to happen?

bench/src/bulk.rs
```diff
@@ -274,6 +274,7 @@ fn transport_config(opt: &Opt) -> quinn::TransportConfig {
     config
         .max_concurrent_uni_streams(opt.max_streams as u64)
         .unwrap();
+    config.congestion_controller_factory(Arc::new(quinn_proto::congestion::CubicConfig::default()));
```
Matthias247, line 277: I doubt this makes a difference in the benchmark without latency injection. Maybe if socket buffers are set rather low, where New Reno will back off with a sawtooth profile once the socket buffer size had been hit. But with a more realistic setting this doesn't happen.

Probably needs some kind of loss or latency injection to really show the benefit.

FrankSpitulski: Right that part was just so I had an easy test that it was working. I can revert this.

quinn-proto/src/connection/mod.rs
FrankSpitulski: From the equation on Wikipedia it seems like it should be neither. I got it from here https://github.com/cloudflare/quiche/blob/567cc5ebc530ed5cf2c14dd4af88a2496bacb290/src/recovery/cubic.rs#L286 which applies to the first equation here https://en.wikipedia.org/wiki/CUBIC_TCP. T is defined as "Time elapsed since the last window reduction". I'm not sure why min_rtt is used in quiche but I'm also not that familiar with CCAs. For what it's worth just observing the cwnd and congestion events behaves like I'd expect a cubic graph to look like from the examples so it can't be too far off either way.

Ralith: I think this is related to quiche's use of [HyStart](https://datatracker.ietf.org/doc/html/draft-ietf-tcpm-hystartplusplus). We should refer to that document for the proper parameters.

FrankSpitulski: When porting it over I assumed false for the hystart flag to simplify things and just focus on cubic, so this impl should not have hystart in it.

Matthias247: > Time elapsed since the last window reduction

That definition of T doesn’t sound related to RTT at all to me? window is reduced when there is a congestion event, so is this the time since the last congestion event? 

Or does reduced also apply for „using“ parts of the congestion window by sending data. But I doubt it, with pacing that time is more or less constant.

FrankSpitulski: so I had a closer look at the equations and code, it looks like the min_rtt should be removed from the w_cubic function since T is already covered in that first arg by the literal t variable. I'm not sure where w_est came from though. the comments say it came from equation 4. still looking.

FrankSpitulski: found the 4th equation. rtt should be there, but not min rtt. https://datatracker.ietf.org/doc/html/rfc8312#section-4.2

Matthias247: I'm not sure why `min_rtt` is added here, but I guess there was a reason. 
Note that there is an update to the Cubic spec which Quiche already seems to follow: 

https://github.com/NTAP/rfc8312bis/pull/3

https://github.com/NTAP/rfc8312bis/blob/main/draft-ietf-tcpm-rfc8312bis.md

Maybe this one is different from Wikipedia?

Matthias247: https://github.com/NTAP/rfc8312bis/pull/3/files#diff-b956a53223cc21d8f0bc2affbb81d6f32e0eb24f3a38f494a53af70e9dd070acR393 in line 393 adds the rtt, but doesn't specify which one. 

And for RTT the spec describes:

> RTT: Smoothed round-trip time in seconds, calculated as described in {{!RFC6298}}.

So it seems like passing the smoothed RTT is correct (even though I don't know whether the smoothing is the same between QUIC reocvery and that spec).

In that case I would remember just renaming the parameter from `min_rtt` to to just `rtt` or `smoothed_rtt`

Matthias247: here's a rendered version of the new spec: https://ntap.github.io/rfc8312bis/draft-ietf-tcpm-rfc8312bis.html#name-introduction

Matthias247:
Some more discussions around Cubic and QUIC (also around `min_rtt`):

https://github.com/NTAP/rfc8312bis/issues/45

https://github.com/quicwg/base-drafts/issues/4833

quinn-proto/src/connection/mod.rs
FrankSpitulski: 👍  rtt it is

quinn-proto/src/congestion/cubic.rs
```diff
@@ -0,0 +1,277 @@
...
+/// CUBIC Functions.
+///
+/// Note that these calculations are based on a count of cwnd as bytes,
+/// not packets.
+/// Unit of t (duration) and RTT are based on seconds (f64).
+impl State {
+    // K = cbrt(w_max * (1 - beta_cubic) / C) (Eq. 2)
+    fn cubic_k(&self, max_datagram_size: u64) -> f64 {
+        let w_max = self.w_max / max_datagram_size as f64;
+        libm::cbrt(w_max * (1.0 - BETA_CUBIC) / C)
```
Ralith, line 42: Why `libm` rather than [`f32::cbrt`](https://doc.rust-lang.org/std/primitive.f32.html#method.cbrt)`?

FrankSpitulski: that's an effect of [inspiration from quiche's implementation](https://github.com/cloudflare/quiche/blob/master/src/recovery/cubic.rs#L125)

FrankSpitulski: looks like it's to support no_std

Ralith: We don't support no_std, so it would be better to avoid the extra dependency.

quinn-proto/src/connection/mod.rs
FrankSpitulski: @Matthias247 good to close this thread?

Matthias247: If the implementation follows the new draft - sure

quinn-proto/src/congestion/cubic.rs
```diff
@@ -0,0 +1,264 @@
...
+    // (t / RTT) (Eq. 4)
+    fn w_est(&self, t: Duration, rtt: Duration, max_datagram_size: u64) -> f64 {
+        let w_max = self.w_max / max_datagram_size as f64;
+        (w_max * BETA_CUBIC
+            + 3.0 * (1.0 - BETA_CUBIC) / (1.0 + BETA_CUBIC) * t.as_secs_f64() / rtt.as_secs_f64())
+            * max_datagram_size as f64
+    }
+}
+
+/// A simple, standard congestion controller
```
Ralith, line 62: We need to give the reader some idea how this differs from `NewReno`.
```suggestion
/// The RFC8312 congestion controller, as widely used for TCP
```

quinn-proto/src/congestion/cubic.rs
```diff
@@ -0,0 +1,264 @@
...
+                    * self.config.max_datagram_size as f64;
+
+                cubic_cwnd += cubic_inc as u64;
+            }
+
+            // Update the increment and increase cwnd by MSS.
+            self.cubic_state.cwnd_inc += cubic_cwnd - self.window;
+
+            // cwnd_inc can be more than 1 MSS in the late stage of max probing.
+            // however QUIC recovery draft 7.4 (Congestion Avoidance) limits
```
Ralith, line 157:
```suggestion
            // however RFC9002 §7.3.3 (Congestion Avoidance) limits
```

I'm not actually sure if the wording in that section should be considered applicable to non-NewReno congestion controllers. Does quiche clamp the window growth similarly?

Ralith (changes requested):
The RFC8312bis draft seems to be continuing to diverge from what we have here, but it's a moving target, so unless you're feeling super motivated to track it closely then I think we can call the underlying math here good.

quinn-proto/src/congestion/cubic.rs
FrankSpitulski: yes the only major difference with the quiche impl is the slow start. I kept the same slow start as quinn's new reno instead of implementing hystart.

Ralith (approved):
Thanks!

djc:
Sorry, must have lost track of this one!

PR #1124, opened 2021-05-22
Add a mean to set the local address in the perf_client

lu-zero:


perf/src/bin/perf_client.rs
```diff
@@ -48,6 +48,9 @@ struct Opt {
     /// Receive buffer size in bytes
     #[structopt(long, default_value = "2097152")]
     recv_buffer_size: usize,
+    /// Specify the local ip
+    #[structopt(long)]
+    local_ip: Option<IpAddr>,
```
Ralith, line 53:
```suggestion
    /// Specify the local address
    #[structopt(long)]
    local_addr: Option<SocketAddr>,
```

perf/src/bin/perf_client.rs
```diff
@@ -79,11 +82,14 @@ async fn run(opt: Opt) -> Result<()> {
...
-        Ipv6Addr::UNSPECIFIED.into()
-    };
+    let bind_addr = opt.local_ip.unwrap_or_else(|| {
+        if addr.is_ipv4() {
+            Ipv4Addr::UNSPECIFIED.into()
+        } else {
+            Ipv6Addr::UNSPECIFIED.into()
+        }
+    });
+    info!("Local addr {:?}", bind_addr);
```
Ralith, line 92:
```suggestion
    info!("local addr {:?}", bind_addr);
```

Ralith (changes requested):
Thanks! Might as well make this a little more flexible, I think.

perf/src/bin/perf_client.rs
lu-zero: I'll update it to use a socketaddr soon :)

djc (approved):

Issue #1126, opened 2021-05-22
Congestion window grows without bound when CPU-bound


Ralith:
If throughput is limited by CPU resources, then every non-app-limited transmit will, once ACKed, cause the congestion window to grow. If CPU resources abruptly increase (e.g. due to reduced contention from other tasks and/or processes) then we risk flooding the link. The congestion controller interface/implementation should be modified such that success with N in-flight bytes is interpreted as license to grow the congestion window by at most a constant multiple of N.

Matthias247:
When implementing https://github.com/quinn-rs/quinn/pull/1127 I noticed that cwnd growth depends heavily on the setting of the new `MAX_TRANSMIT_DATAGRAMS` flag. When configured lower, CWND is more likely to grow to infinitiy since a single poll will never yield all available datagrams, which leads `app_limited` set to `false` and allows CWND to grow. If we increase the value and all datagrams are polled then `app_limited` might get set to `true` which stops CWND from growing. However that also means higher latencies.

Some examples:

MAX_TRANSMIT_DATAGRAMS = 20:

``` 
path: PathStats {
        rtt: 389.489µs,
        cwnd: 536692490,
```

MAX_TRANSMIT_DATAGRAMS = 30:

```
 path: PathStats {
        rtt: 478.052µs,
        cwnd: 2821703,
```

MAX_TRANSMIT_DATAGRAMS = 50:

```
path: PathStats {
        rtt: 646.804µs,
        cwnd: 1481010,
```

MAX_TRANSMIT_DATAGRAMS = 100:

```
path: PathStats {
        rtt: 533.493µs,
        cwnd: 1476074,
```

While this benchmark looks like `30` would be a better value then `20` I would remind that I'm running this on a desktop with extremely high single-threaded performance, and pretty much everything else would have a harder time producing 30 packets and have higher latencies.

Overall I think the way `app_limited` is set at the moment is not optimal, and there needs to be a strategy improvement

Ralith:
What if we maintain a separate `cpu_limited` flag, set it whenever we stop calling `poll_transmit` early, and unset it whenever `poll_transmit` is fully drained?

Matthias247:
Interesting thought. But it kind of relies on whether the application does call `poll_transmit` again as early as possible, and not delays it for other reasons - right? The `quinn` wrapper does, but it would be some kind of implicit contract (similar to the `read/EAGAIN/epoll` ones.

[...]

PR #1127, opened 2021-05-22
Improve fairness between tasks further

Matthias247:
This introduces two further limits which place bounds on task iterations.

This significantly reduces RTT for a loopback connection since the connection
is no longer busy producing packets and not doing any reception work.

Raw throughput can be a bit slower since a lot more ACK packets are now
processed in the loopback test. However RTT is halfed. The difference
in ACK packets (7200 vs 83000) and RTT (800us vs 400us) is very
visible in the benchmark:

**Before:**

```
Overall stats:

Sent 1073741824 bytes on 1 streams in 1.70s (603.88 MiB/s)

Server connection stats:
ConnectionStats {
    frame_tx: FrameStats {
        ACK: 7199,
    },
    frame_rx: FrameStats {
        ACK: 903129,
    },
    path: PathStats {
        rtt: 459.755µs,
        cwnd: 12320,
    },
}

Client connection stats:
ConnectionStats {
    frame_tx: FrameStats {
        ACK: 903129,
    },
    frame_rx: FrameStats {
        ACK: 7199,
    },
    path: PathStats {
        rtt: 769.84µs,
        cwnd: 1094264975,
    },
}
```

**After:**
```
Sent 5389680640 bytes on 1 streams in 8.51s (603.90 MiB/s)

Server connection stats:
ConnectionStats {
    frame_tx: FrameStats {
        ACK: 83555,
    },
    frame_rx: FrameStats {
        ACK: 4563602,
        STREAM: 4562497,
    },
    path: PathStats {
        rtt: 205.582µs,
        cwnd: 12320,
    },
}

Client connection stats:
ConnectionStats {
    frame_tx: FrameStats {
        ACK: 4563602,
        STREAM: 4562497,
    },
    frame_rx: FrameStats {
        ACK: 83555,
    },
    path: PathStats {
        rtt: 393.448µs,
        cwnd: 495821204,
    },
}
```

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -287,25 +287,24 @@ where
         let span = info_span!("drive", id = conn.handle.0);
         let _guard = span.enter();
 
-        loop {
-            let mut keep_going = false;
-            if let Err(e) = conn.process_conn_events(cx) {
-                conn.terminate(e);
-                return Poll::Ready(());
-            }
-            conn.drive_transmit();
-            // If a timer expires, there might be more to transmit. When we transmit something, we
-            // might need to reset a timer. Hence, we must loop until neither happens.
-            keep_going |= conn.drive_timer(cx);
-            conn.forward_endpoint_events();
-            conn.forward_app_events();
-            if !keep_going || conn.inner.is_drained() {
-                break;
-            }
+        if let Err(e) = conn.process_conn_events(cx) {
+            conn.terminate(e);
+            return Poll::Ready(());
         }
+        let mut keep_going = conn.drive_transmit();
+        // If a timer expires, there might be more to transmit. When we transmit something, we
+        // might need to reset a timer. Hence, we must loop until neither happens.
+        keep_going |= conn.drive_timer(cx);
+        conn.forward_endpoint_events();
+        conn.forward_app_events();
 
         if !conn.inner.is_drained() {
-            conn.driver = Some(cx.waker().clone());
+            if keep_going {
+                // If the connection hasn't processed all tasks, schedule it again
+                cx.waker().wake_by_ref();
+            } else {
+                conn.driver = Some(cx.waker().clone());
+            }
             return Poll::Pending;
         }
         if conn.error.is_none() {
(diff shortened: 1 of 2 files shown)


quinn/src/endpoint.rs
```diff
@@ -368,8 +368,9 @@ where
         }
     }
 
-    fn handle_events(&mut self, cx: &mut Context) {
+    fn handle_events(&mut self, cx: &mut Context) -> bool {
         use EndpointEvent::*;
+        let mut num_events = 0;
         loop {
```
Ralith, line 374: This could be a bit more concise as `for _ in 0..IO_LOOP_BOUND`

quinn/src/connection.rs
```diff
@@ -287,25 +287,24 @@ where
...
+        // might need to reset a timer. Hence, we must loop until neither happens.
+        keep_going |= conn.drive_timer(cx);
+        conn.forward_endpoint_events();
+        conn.forward_app_events();
 
         if !conn.inner.is_drained() {
-            conn.driver = Some(cx.waker().clone());
+            if keep_going {
+                // If the connection hasn't processed all tasks, schedule it again
+                cx.waker().wake_by_ref();
```
Ralith, line 304: This means when we get a timer wakeup (e.g. due to keep-alives) we'll have to go through two wakeup cycles to transmit a packet in response. Is that desired? Admittedly that's not the hot path...

quinn/src/connection.rs
```diff
@@ -799,17 +798,31 @@ impl<S> ConnectionInner<S>
...
+                None => 1,
+                Some(s) => (t.contents.len() + s - 1) / s, // round up
+            };
             // If the endpoint driver is gone, noop.
             let _ = self
                 .endpoint_events
                 .unbounded_send((self.handle, EndpointEvent::Transmit(t)));
+
+            if transmits >= MAX_TRANSMIT_DATAGRAMS {
+                // TODO: What isn't ideal here yet is that we don't poll all
```
Ralith, line 818: Missing "if" in "is that we don't"? Maybe cite #1126?

quinn/src/connection.rs
```diff
@@ -1073,3 +1086,9 @@ pub enum SendDatagramError {
     #[error("connection closed: {0}")]
     ConnectionClosed(#[source] ConnectionError),
 }
+
+/// The maximum amount of datagrams which will be produced in a single `drive_transmit` call
+///
+/// This limits the amount of CPU resources required for datagram generation,
```
Ralith, line 1092: s/required for/consumed by/?

quinn/src/connection.rs
Matthias247: linked

quinn/src/connection.rs
Matthias247: Yes, I wondered about that too. But given it's not a regular event I didn't feel it was worth keeping the loop just for that case. I can add it back if others feel different about it. 

However I also thought it would only matter if a timer expired immediately or between the start of `drive` and the timer call. Now I see that this is not true, and that all timer expirations would require 2 iterations since timer expiration is not checked unless the tokio timer asks us to do so. That makes it kind of more interesting to add the loop back (or to move the timer check between processing incoming events and sending outgoing events?)

Ralith: > or to move the timer check between processing incoming events and sending outgoing events

We update the loss detection and pacing timers when sending a packet, so if we move the timers before `poll_transmit` we'll need far more double-loops than the current configuration.

Most timer expirations are rare, but the pacing timer might be firing ~continuously when the network is saturated. Might be worth gathering some metrics to inform this. Though, empirically the impact isn't that large, so we don't have to make a huge deal out of it.

Ralith (approved):

PR #1129, opened 2021-05-23
Disable GSO after encountering EIO on send

### LINKED ISSUE #1128: Fall back to disabling GSO after encountering an I/O error
A report from @lu-zero revealed that Linux can erroneously advertise GSO support on systems where adapter support for GSO support (`ethtool -k <interface> |grep generic-segmentation-offload`) is mixed. This manifests as `EIO` errors from `sendmmsg` when the `UDP_SEGMENT` cmsg is set. For compatibility, when GSO is enabled, we should handle the first `EIO` by disabling GSO support.

Ralith:
Fixes #1128.

I couldn't find a way to isolate the changes to just endpoint.rs or the UDP platform code. I'm not super happy with the global state, but I think getting rid of it would require moving `UdpCapabilities` into memory explicitly shared between the `EndpointInner` and `ConnectionInner` structures, which can be pursued separately.


quinn/src/platform/unix.rs
```diff
@@ -334,7 +338,11 @@ fn recv(
 
 /// Returns the platforms UDP socket capabilities
 pub fn caps() -> UdpCapabilities {
-    *CAPABILITIES
+    let mut caps = *CAPABILITIES;
+    if SAW_IO_ERROR.load(Ordering::Relaxed) {
```
Matthias247, line 342: Should capabilities be updated instead of doing this? The fallback generating an addditional atomic operation is a bit inconvenient. I'm wondering what the cheapest way to get this done is. `CAPABILITIES` being a `static mut`? It doesn't really matter if the access is too slowly propagated as long as the new value gets visible at some time.

quinn/src/endpoint.rs
```diff
@@ -558,3 +577,32 @@ where
         &self.0
     }
 }
+
+#[cfg(target_os = "linux")]
+/// Expand GSO-layout Transmits into single segment Transmits
+fn flatten_gso(transmits: &mut VecDeque<proto::Transmit>) {
```
Matthias247, line 583: Not sure whether it's worth it. The first transmit was already lost at the point, and the connection doesn't know. Before it sets something up for retransmits the capability change will also have propagated, and there is no need for flatten for the retry.

Instead of copying contents here we could also have a cheaper fallback which just does sends multiple messages and just references the payload slices. But that might obviously be even more invasive.

Ralith: The latter was my first idea, but it turns out to be a huge pain because sendmmsg can have partial success whereas GSO can't(?). With this code, we never discard the failed transmit, so it's not lost, though I agree it's kinda ugly. Are you thinking we might as well just pretend we lost the packets in this case and deal with the transient extra latency? We do need to handle the possibility of multiple bad flights, since there could be multiple connections in flight and getting new data to send vs. updating the capabilities is racey.

Matthias247: > The latter was my first idea, but it turns out to be a huge pain because sendmmsg can have partial success whereas GSO can't(?).

I think that's the theory, but I'm not sure whether it matters in practice. It looks to me like all the `sendmsg` APIs will signal success anyway, and packets will get dropped later on a lower layer (or they might even get dropped when success is signaled).

> Are you thinking we might as well just pretend we lost the packets in this case and deal with the transient extra latency? 

Yes, that was the idea. But I kind of agree its ugly that those systems will then encounter an additional handshake delay - which makes QUIC certainly perform less than TCP + TLS1.2.  Some better understanding on how widespread those broken systems are would be interesting to determine whether that's a reasonable path forward

PR #1130, opened 2021-05-24
Prevent sending of ACK-only packets

Matthias247:
The priority-based send state implementation had an issue where if returned
`self.can_send() == true` even if no data was available for sending.
This condition lead the `poll_transmit` method to initiate a transmission which
passed the ACK-only check since it assumed there was stream data to send.
However when trying to write stream frames no actual stream data could be
written.

The reason for this was that the `can_send()` condition only checked for the
length of the binary heap. However the length of this one was only reduced
in `write_stream_frames()` when the level was inspected the next time, and
not after data of one level was actually written.

This change fixes that, and drops the unused levels immediately after data
was written. There is on exception however: If only one level is left, it is
kept around and reused for future transmits to avoid deallocating and
reallocating the transmit queue continuously. However the `can_send` check
was updated to account for the empty level.

I also added some additional debug asserts which cross-checks
if an operation which announced to write more than just ACKs
still ended up only writing ACKs. That should make it easier to
determine similar issues in the future.

**Ack only transmissions in benchmark before change:** 1462

**Ack only transmissions in benchmark after change:** 45


quinn-proto/src/connection/mod.rs
```diff
@@ -757,23 +768,28 @@ where
         })
     }
 
-    /// Returns `true` if a space has outgoing data to send
-    fn space_can_send(&self, space_id: SpaceId) -> bool {
+    /// Returns `if a space has outgoing data to send and what the expected type of data is
+    fn space_can_send(&self, space_id: SpaceId) -> CanSendResult {
         if self.spaces[space_id].crypto.is_some() && self.spaces[space_id].can_send() {
-            return true;
+            return CanSendResult::AckOnly;
```
Ralith, line 774: I don't think this is correct, or at least it's misleading. `Space::can_send` checks lots of non-ACK stuff like stream control frames.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -293,7 +293,10 @@ impl StreamsState {
     }
 
     pub fn can_send(&self) -> bool {
-        !self.pending.is_empty()
+        match self.pending.peek() {
+            Some(head) => head.queue.borrow().len() != 0,
```
Ralith, line 297: Style nit:
```suggestion
            Some(head) => !head.queue.borrow().is_empty(),
```

quinn-proto/src/connection/mod.rs
Matthias247: That's true. Any idea for a better name?

Matthias247: Ping on this one, since it seems to be the main outstanding question.

The main thing that comes to my mind now is `CanSendResult::NonApplicationData` or `CanSendResult::ManagementData` - but even that is not 100% right since path challenge/response are are part of OneRtt/ZeroRtt

Matthias247: Thinking a bit more about it, we can also just change the type to

```
struct CanSendReesult {
   acks: bool,
   other: bool,
}
```

then it's also clear whether ACK-only was indicated or not.

djc: Seems like a pretty decent type, though maybe a name like `FrameTypes` would make more sense for it?

Matthias247: I updated with `SendableFrames`. I don't think it's 100% perfect either, since it actually is sent at least since checks are short-circuiting. If ACKs are indicated, something else can be send too (`can_send()` doesn't look into 1RTT spaces - we can make it do, but that requires more work just to keep the debug assert around). Maybe I should just remove the assert to land the bugfix?

Matthias247:
Lints fail due to rustc 1.53 update. Should be fixed with https://github.com/quinn-rs/quinn/pull/1147

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -318,6 +332,7 @@ fn push_pending(pending: &mut BinaryHeap<PendingLevel>, id: StreamId, priority:
     });
 }
 
+#[derive(Debug)]
```
Ralith, line 335: Was this intentional?

quinn-proto/src/connection/mod.rs
```diff
@@ -757,23 +769,35 @@ where
...
+        if self.spaces[space_id].crypto.is_some() {
+            let can_send = self.spaces[space_id].can_send();
+            if !can_send.is_empty() {
+                return can_send;
+            }
         }
 
         if space_id != SpaceId::Data {
-            return false;
+            return SendableFrames::default();
```
Ralith, line 782: It wasn't immediately obvious to me what "default" meant here. Let's replace uses of it with a more explicit `SendableFrames::empty()` helper, for symmetry with `is_empty`.

Ralith (changes requested):
LGTM aside from a readability issue and the lints.

quinn-proto/src/connection/streams/mod.rs
Matthias247: I likely added it while debugging the problem. Can either keep or remove

Ralith: I weakly favor feeding rustc less code for this sort of thing, but I don't care that much.

Matthias247: I removed it now. If it's required by someone it's trivial to add back

Ralith (approved):

quinn-proto/src/connection/mod.rs
```diff
@@ -703,6 +704,21 @@ where
             }
 
             let sent = self.populate_packet(space_id, &mut buf, buf_capacity - builder.tag_len);
+
+            // ACK-only packets should only be sent when explicitely allowed
+            // If we write them due to another reason there is a bug somewhere
+            // which leads to one component announcing write readiness and not writing any data.
+            // This degrades performance.
+            // The condition is only checked if the full MTU is available, so that lack of
+            // space in the datagram isn't the reaason for just writing ACKs.
```
djc, line 713: Fixed some typos, also compacted the comment a bit.

```suggestion
            // ACK-only packets should only be sent when explicitly allowed. If we write them due
            // to any other reason, there is a bug which leads to one component announcing write
            // readiness while not writing any data. This degrades performance. The condition is
            // only checked if the full MTU is available, so that lack of space in the datagram isn't
            // the reason for just writing ACKs.
```

quinn-proto/src/connection/mod.rs
```diff
@@ -757,23 +773,35 @@ where
         })
     }
 
-    /// Returns `true` if a space has outgoing data to send
-    fn space_can_send(&self, space_id: SpaceId) -> bool {
-        if self.spaces[space_id].crypto.is_some() && self.spaces[space_id].can_send() {
-            return true;
+    /// Returns `if a space has outgoing data to send and what the expected type of data is
```
djc, line 776:
```suggestion
    /// Indicate what types of frames are ready to send for the given space
```

djc (commented):
Thanks for working on this! Some nits:

quinn-proto/src/connection/streams/state.rs
```diff
@@ -473,13 +480,26 @@ impl StreamsState {
...
                 if level.priority == stream.priority {
+                    // Enqueue for the same level
                     level.queue.get_mut().push_back(id);
                 } else {
-                    drop(level);
+                    // Enqueue for a different level. If the current level is empty, drop it
+                    if level.queue.borrow().is_empty() && num_levels != 1 {
+                        // We keep the last level around even in empty form so that
+                        // the next insert doesn't have to reallocate the queue
+                        PeekMut::pop(level);
```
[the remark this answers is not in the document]
Matthias247: Remove the priorities again or don't care about performance :-) 

While I think there might be ways to make this prettier, I'm not super inclined to investigate further in this. This is a rather performance-impacting bug that is open for 30 days, and I would rather like to see it fixed ASAP and spend energy into more impactful things than moving code around.

It's also not exactly duplicated.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -310,6 +310,20 @@ fn push_pending(pending: &mut BinaryHeap<PendingLevel>, id: StreamId, priority:
             return;
         }
     }
+
+    // If there is only a single level and it's empty, repurpose it for the
+    // required priority
+    if pending.len() == 1 {
+        let mut first = pending.peek_mut().unwrap();
```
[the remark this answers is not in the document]
Matthias247: done

quinn-proto/src/connection/streams/state.rs
```diff
@@ -295,7 +295,10 @@ impl StreamsState {
     }
 
     pub fn can_send(&self) -> bool {
-        !self.pending.is_empty()
+        match self.pending.peek() {
```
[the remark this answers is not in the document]
Matthias247: I replaced it with
```
self.pending
            .peek()
            .map(|head| !head.queue.borrow().is_empty())
            .unwrap_or(false)
```

but I actually don't find it better, and I feel I gotten the advice to move to a match in the past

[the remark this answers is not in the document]
djc: Sorry, I was actually thinking of map_or. Does that help?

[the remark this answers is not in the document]
Matthias247: Updated with that. Saves at least I line, but I overall still say those are all the same to me 🤷🏻‍♂️

Ralith (approved):

PR #1132, opened 2021-05-27
Print connection statistics in perf client and server

Matthias247:
This adds the `--conn-stats` flag to both CLI tools, which will print
the connection statistics together with other data every 2s.

### DIFF
--- a/perf/src/bin/perf_server.rs
+++ b/perf/src/bin/perf_server.rs
@@ -1,4 +1,4 @@
-use std::{fs, net::SocketAddr, path::PathBuf, sync::Arc};
+use std::{fs, net::SocketAddr, path::PathBuf, sync::Arc, time::Duration};
 
 use anyhow::{Context, Result};
 use bytes::Bytes;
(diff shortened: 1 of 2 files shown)


perf/src/bin/perf_server.rs
```diff
@@ -100,18 +106,33 @@ async fn run(opt: Opt) -> Result<()> {
...
 
+async fn conn_stats(connection: quinn::Connection, opt: Arc<Opt>) -> Result<()> {
+    loop {
+        tokio::time::sleep(Duration::from_secs(2)).await;
+        {
+            if opt.conn_stats {
+                println!("{:?}\n", connection.stats());
+            }
+        }
+    }
```
Ralith, line 133: Could we move this branch outside of the loop?

Matthias247: Sure. I first thought the function may not end in order not to disconnect the server prematurely - but it's ok since it uses `try_join!` and not `select!`

Ralith (approved):

PR #1133, opened 2021-05-27
Dynamically determine the amount of work that can be performed in an endpoint iteration

Matthias247:
While the hardcoded numbers might work good on some machines, they can be problematic on others.
They also make it harder to distinguish on how much time is spent on send vs receive.

This introduces a dynamic `WorkLimiter` which measures the elapsed time per operation, and allows to configure how much time we intend to spend on send vs receive.

This change integrates it so far for `recv` which is at the moment more problematic, but it can easily be extended for the send path.

Performance differences are negligible on this machine (probably because
`IO_LOOP_BOUND` was set to a number which works for it), but it can improve
things on less known environments.

I instrumented the endpoints receive method to see how much time it spends
on average in `drive_recv`.

**Baseline:**
```
Recv time: AvgTime { total: 3.280880841s, calls: 34559, avg: 94.935µs, min: 3.146µs, max: 312.574µs }
path: PathStats {a
    rtt: 511.656µs,
```

**With this change:**
```
Recv time: AvgTime { total: 3.333642823s, calls: 54627, avg: 61.024µs, min: 2.645µs, max: 319.147µs }
path: PathStats {
    rtt: 446.641µs,
```
Note that 50µs are not reached because a single `recvmmsg` batch takes about 30µs, so this is just rounding up to 2 batches.

**When set to 200µs (for comparison purposes):**
```
Recv time: AvgTime { total: 3.243954076s, calls: 19558, avg: 165.862µs, min: 2.525µs, max: 358.711µs }
path: PathStats {
    rtt: 700.34µs,
}
```

### DIFF
--- a/quinn/src/work_limiter.rs
+++ b/quinn/src/work_limiter.rs
@@ -1,0 +1,234 @@
+use std::time::{Duration, Instant};
+
+/// Limits the amount of time spent on a certain type of work in a cycle
+///
+/// The limiter works dynamically: For a sampled subset of cycles it measures
+/// the time that is approximately required for fulfilling 1 work item, and
+/// calculates the amount of allowed work items per cycle.
+/// The estimates are smoothed over all cycles where the exact duration is measured.
+///
+/// In cycles where no measurement is performed the previously determined work limit
+/// is used.
+///
+/// For the limiter the exact definition of a work item does not matter.
+/// It could for example track the amount of transmitted bytes per cycle,
+/// or the amount of transmitted datagrams per cycle.
+/// It will however work best if the required time to complete a work item is
+/// constant.
+#[derive(Debug)]
+pub struct WorkLimiter {
+    /// Whether to measure the required work time, or to use the previous estimates
+    mode: Mode,
+    /// The current cycle number
+    cycle: u16,
+    /// The time the cycle started - only used in measurement mode
+    start_time: Instant,
+    /// How many work items have been completed in the cycle
+    completed: usize,
+    /// The amount of work items which are allowed for a cycle
+    allowed: usize,
+    /// The desired cycle time
+    desired_cycle_time: Duration,
+    /// The estimated and smoothed time per work item in nanoseconds
+    smoothed_time_per_work_item_nanos: f64,
+    /// Retrieves the current time for unit-test purposes
+    #[cfg(test)]
+    get_time: fn() -> Instant,
+}
+
+impl WorkLimiter {
+    pub fn new(desired_cycle_time: Duration) -> Self {
+        Self {
+            mode: Mode::Measure,
+            cycle: 0,
+            start_time: Instant::now(),
+            completed: 0,
+            allowed: 0,
+            desired_cycle_time,
+            smoothed_time_per_work_item_nanos: 0.0,
+            #[cfg(test)]
+            get_time: std::time::Instant::now,
+        }
+    }
+
+    /// Starts one work cycle
+    pub fn start_cycle(&mut self) {
+        self.completed = 0;
+        if let Mode::Measure = self.mode {
+            self.start_time = self.now();
+        }
+    }
+
+    /// Returns whether more work can be performed inside the `desired_cycle_time`
+    ///
+    /// Requires that previous work was tracked using `record_work`.
+    pub fn allow_work(&mut self) -> bool {
+        match self.mode {
+            Mode::Measure => (self.now() - self.start_time) < self.desired_cycle_time,
+            Mode::HistoricData => self.completed < self.allowed,
+        }
+    }
+
+    /// Records that `work` additional work items have been completed inside the cycle
+    ///
+    /// Must be called between `start_cycle` and `finish_cycle`.
+    pub fn record_work(&mut self, work: usize) {
+        self.completed += work;
+    }
+
+    /// Finishes one work cycle
+    ///
+    /// For cycles where the exact duration is measured this will update the estimates
+    /// for the time per work item and the limit of allowed work items per cycle.
+    /// The estimate is updated using the same exponential averaging (smoothing)
+    /// mechanism which is used for determining QUIC path rtts: The last value is
+    /// weighted by 1/8, and the previous average by 7/8.
+    pub fn finish_cycle(&mut self) {
+        // If no work was done in the cycle drop the measurement, it won't be useful
+        if self.completed == 0 {
+            return;
+        }
+
+        if let Mode::Measure = self.mode {
+            let elapsed = self.now() - self.start_time;
+
+            let time_per_work_item_nanos = (elapsed.as_nanos()) as f64 / self.completed as f64;
+
+            self.smoothed_time_per_work_item_nanos = if self.allowed == 0 {
+                // Initial estimate
+                time_per_work_item_nanos
+            } else {
+                // Smoothed estimate
+                (7.0 * self.smoothed_time_per_work_item_nanos + time_per_work_item_nanos) / 8.0
+            };
+
+            self.allowed = ((self.desired_cycle_time.as_nanos()) as f64
+                / self.smoothed_time_per_work_item_nanos) as usize;
+        }
+
+        self.cycle = self.cycle.wrapping_add(1);
+        self.mode = match self.cycle % SAMPLING_INTERVAL {
+            0 => Mode::Measure,
+            _ => Mode::HistoricData,
+        };
+    }
+
+    #[cfg(not(test))]
+    fn now(&self) -> Instant {
+        Instant::now()
+    }
+
+    #[cfg(test)]
+    fn now(&self) -> Instant {
+        (self.get_time)()
+    }
+}
+
+/// We take a measurement sample once every `SAMPLING_INTERVAL` cycles
+const SAMPLING_INTERVAL: u16 = 256;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum Mode {
+    Measure,
+    HistoricData,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::cell::RefCell;
+
+    #[test]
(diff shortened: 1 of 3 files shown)


quinn/src/work_limiter.rs
```diff
@@ -0,0 +1,235 @@
...
+///
+/// For the limiter the exact definition of a work item does not matter.
+/// It could for example track the amount of transmitted bytes per cycle,
+/// or the amount of transmitted datagrams per cycle.
+/// It will however work best if the required time to complete a work item is
+/// constant.
+#[derive(Debug)]
+pub struct WorkLimiter {
+    /// Whether to measure the required work time, or to use the previous estimates
+    mode: LimiterMode,
```
djc, line 21: Let's make these field names a bit more concise, similar to most of the code. `LimiterMode` can just be called `Mode` (it's private anyway) and we can cut `work_item` from all field names. I think we can also do without the `_nanos` suffix, keeping as documentation only.

quinn/src/work_limiter.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    /// Starts one work cycle
+    pub fn start_cycle(&mut self) {
+        self.completed_work_items = 0;
+        match self.mode {
+            LimiterMode::Measure => {
+                self.start_time = Instant::now();
+            }
+            LimiterMode::HistoricData => {}
+        }
```
djc, line 57: Let's use `if let` here.

quinn/src/work_limiter.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    /// Returns whether more work can be performed inside the `desired_cycle_time`
+    ///
+    /// Requires that previous work was tracked using `record_work`.
+    pub fn allow_work(&mut self) -> bool {
+        match self.mode {
+            LimiterMode::Measure => self.start_time.elapsed() < self.desired_cycle_time,
+            LimiterMode::HistoricData => self.completed_work_items < self.allowed_work_items,
+        }
+    }
```
djc, line 68: I feel like we should maybe leverage the type system more here? For example, `start() -> Cycle`, `Cycle::continue() -> bool`, `Cycle::record()`, and `impl Drop for Cycle`.

quinn/src/work_limiter.rs
```diff
@@ -0,0 +1,235 @@
...
+                (elapsed.as_nanos()) as f64 / self.completed_work_items as f64;
+
+            if self.allowed_work_items == 0 {
+                // Initial estimate
+                self.smoothed_time_per_work_item_nanos = time_per_work_item_nanos;
+            } else {
+                // Smoothed estimate
+                self.smoothed_time_per_work_item_nanos =
+                    (7.0 * self.smoothed_time_per_work_item_nanos + time_per_work_item_nanos) / 8.0;
+            }
```
djc, line 100: Let's hoist the assignment out here.

quinn/src/work_limiter.rs
```diff
@@ -0,0 +1,235 @@
...
+                / self.smoothed_time_per_work_item_nanos)
+                as usize;
+        }
+
+        self.cycle = self.cycle.wrapping_add(1);
+        if self.cycle % SAMPLING_INTERVAL == 0 {
+            self.mode = LimiterMode::Measure;
+        } else {
+            self.mode = LimiterMode::HistoricData;
+        }
```
djc, line 112:
```rust
match self.cycle % SAMPLING_INTERVAL {
    0 => Mode::Measure,
    _ => Mode::HistoricData,
}
```

quinn/src/work_limiter.rs
```diff
@@ -0,0 +1,235 @@
+use std::time::{Duration, Instant};
+
+/// Limits the amount of time spent on a certain type of work in a cycle
+///
+/// The limiter works dynamically: For a sampled subset of cycles it measures the
+/// the time that is approximately required for fulfilling 1 work item, and
```
djc, line 6: `the the`

quinn/src/work_limiter.rs
```diff
@@ -0,0 +1,235 @@
...
+    ///
+    /// Must be called between `start_cycle` and `finish_cycle`.
+    pub fn record_work(&mut self, work: usize) {
+        self.completed_work_items += work;
+    }
+
+    /// Finishes one work cycle
+    ///
+    /// For cycles where the exact duration is measured this will update the estimates
+    /// for the time per work item and the limit of allowed work items per cycle.
```
djc, line 80: Could use some documentation on how smoothing is done.

djc (commented):
The idea here makes sense to me!

Please squash these commits together, since we usually introduce new code together with changes that use them; and please consider shortening the first line of the commit message a bit.

quinn/src/work_limiter.rs
Matthias247: The idea is neat. But I have been bitten so many times before where it didn't work in the end, because another call to `self.something()` was required inside the loop and the borrow-checker didn't allow it. Therefore I didn't try here.

While it seems to work with the current code (nothing inside `drive_send/drive_recv` makes a `self.` call - it might be an annoyance down the road. If this is accepted and preferred I can change it.

The reason this works better with `Mutex/RefCell` is that they are often the only field within a struct, so there is less desire to make additional `self` calls besides locking the mutex.

quinn/src/work_limiter.rs
Matthias247: I made most of them shorter. I kept `_nanos` since this is really something that I find immensly useful when looking at foreign code where the unit isn't clear, and have to point out in reviews constantly.

Also please keep in mind that there is a low downside of having longer names as long as they don't span the full line. Code readability wins in most cases by having them - and code is more often read than written..

Matthias247:
CI failure is on the flaky h3 test. Can someone restart?

Ralith:
Haven't carefully reviewed the implementation yet, but overall this looks reasonable to me. I'd be happier if we had empirical evidence of an environment where the hardcoded values are significantly bad, but that's somewhat mitigated by the good encapsulation here and the infrequent sampling.

Matthias247:
Seems like even though I increased the timer and gave it a lot of grace time the test is still flaky on CI. Maybe the CI macos CI runner is rather overscheduled. I will look into in way to mock time for the test to de-flake it.

Ralith:
Ultimately we want to get rid of these tests since they're a maintenance burden, we just need to investigate making up for any coverage losses that would entail.

Matthias247:
The last failure was about the newly added test, not about the H3 one. I made this one now deterministic (at the cost of some other ugliness in the implementation)

quinn/src/work_limiter.rs
```diff
@@ -0,0 +1,267 @@
...
+    completed: usize,
+    /// The amount of work items which are allowed for a cycle
+    allowed: usize,
+    /// The desired cycle time
+    desired_cycle_time: Duration,
+    /// The estimated and smoothed time per work item in nanoseconds
+    smoothed_time_per_work_item_nanos: f64,
+    /// Retrieves the current time for unit-test purposes
+    #[cfg(test)]
+    get_time: fn() -> Instant,
```
Matthias247, line 36: We can also solve this by either passing a function to retrieve the time to all calls, or by parameterizing `WorkLimiter` with a trait to retrieve the time. I went with that version since its the least invasive from the usage point of view.

quinn/src/work_limiter.rs
djc: Personally I would prefer it; we can tackle any borrowing issues when they arise. But let's see what @Ralith thinks.

Ralith: I don't have a super strong opinion; APIs that are constructively impossible to abuse are nice, but this case doesn't strike me as incredibly error prone to begin with, and keeping the impl simple is nice too.

djc: Okay, let's keep it.

quinn/src/work_limiter.rs
```diff
@@ -0,0 +1,267 @@
...
+        let mut initial_batches = 0;
+        while limiter.allow_work() {
+            limiter.record_work(BATCH_WORK_ITEMS);
+            advance_time(BATCH_TIME);
+            initial_batches += 1;
+        }
+        limiter.finish_cycle();
+
+        assert!(
+            approximates(initial_batches, EXPECTED_INITIAL_BATCHES),
```
Ralith, line 166: Do we still need `approximates` now that we're using mocked time?

Ralith (approved):

quinn/src/work_limiter.rs
Ralith: This seems fine to me. Might be cleaner with a `Box<dyn Fn() -> Instant>` that the test stuffs an `Arc<RefCell<Instant>>` inside, but eh.

quinn/src/work_limiter.rs
Matthias247: I kept it around to accomodate for potential rounding errors, but changed the tolerance for just 10%. Haven't tested whether it would work without it.

Matthias247: works without, so I removed it in favor of `assert_eq`

Matthias247:
I'm not sure what happened to those macos tests:

```
---- tests::echo_dualstack stdout ----
thread 'tests::echo_dualstack' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }', quinn/src/tests.rs:597:56
```

However I feel like those failures are unrelated to these changes

Ralith:
Must be a MacOS kernel change...

Ralith (approved):

Matthias247:
CI failure was most likely caused by an issue in mio 0.7.12 which was picked up by the build: https://github.com/tokio-rs/mio/issues/1497

That version is now yanked, so retrying the run might lead to success

Ralith:
Yep, good find.

PR #1141, opened 2021-06-12
Handle STOP_SENDING only once

Matthias247:
This avoids duplicate emission of various events

### DIFF
--- a/quinn-proto/src/connection/streams/state.rs
+++ b/quinn-proto/src/connection/streams/state.rs
@@ -274,10 +274,12 @@ impl StreamsState {
             Some(ss) => ss,
             None => return,
         };
-        self.events
-            .push_back(StreamEvent::Stopped { id, error_code });
-        stream.stop(error_code);
-        self.on_stream_frame(false, id);
+
+        if stream.try_stop(error_code) {
+            self.events
+                .push_back(StreamEvent::Stopped { id, error_code });
+            self.on_stream_frame(false, id);
+        }
     }
 
     pub fn reset_acked(&mut self, id: StreamId) {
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/streams/state.rs
```diff
@@ -1056,26 +1058,40 @@ mod tests {
...
         };
 
         let reason = 0u32.into();
-        stream.state.received_stop_sending(id, reason);
+        stream.state.received_stop_sending(stream_id, reason);
+        assert!(stream
+            .state
+            .events
+            .iter()
+            .any(|ev| matches!(ev, StreamEvent::Stopped {
```
djc, line 1081: The way this gets formatted makes it hard to follow. Maybe wrap matches!() in braces? Or just use an old-fashioned match expression?

djc (commented):
Looks good! One formatting nit.

quinn-proto/src/connection/streams/state.rs
Ralith: I also don't see any harm in deriving `PartialEq` for `StreamEvent`, which would make this syntatically a lot more straightforward.

Ralith (dismissed):

quinn-proto/src/connection/streams/state.rs
Matthias247: replaced with a match which brings in braces. And yes, probably `PartialEq` could be derived, but since that also changes the public API I didn't want to go for it as part of this change

Matthias247: Now clippy is unhappy:

```
error: match expression looks like `matches!` macro
Error:     --> quinn-proto/src/connection/streams/state.rs:1077:53
     |
1077 |           assert!(stream.state.events.iter().any(|ev| match ev {
     |  _____________________________________________________^
1078 | |             StreamEvent::Stopped { id, error_code }
1079 | |                 if *id == stream_id && *error_code == reason =>
1080 | |                 true,
1081 | |             _ => false,
1082 | |         }));
     | |_________^ help: try this: `matches!(ev, StreamEvent::Stopped { id, error_code } if *id == stream_id && *error_code == reason)`
```

Let me know which approaches you prefer to fix it - adding a clippy ignore or going back to matches and trying to add a brace or adding the derive - everything is fine to me

Ralith: I favor the derive; we'll be rebasing this PR anyway so it doesn't really matter in the end whether it's submitted separately or not. Failing that, I'd prefer falling back to the original version over a clippy ignore.

djc: Deriving `PartialEq` sounds good to me.

Matthias247: done in in extra commit which preceeds the other one

Ralith (approved):

PR #1143, opened 2021-06-12
Account received data for the most recent path

Matthias247:
When path migration happened, the data received was accounted for
the old path, and provided the server less anti-amplification budget
to respond to the client.
By accessing path only after the first packet had been processed and
migration was initiated we make sure the new path gets the budget.


Ralith (approved):
Good fix!

PR #1145, opened 2021-06-15
Fix some h3 test races

Matthias247:
The response body in those tests was never read. Therefore it was possible
for the test to drop the connection before the stream was actually read to
completion.


Ralith (approved):

Ralith:
Hooray for deflaking!

djc:
Nice, thanks!

PR #1146, opened 2021-06-16
Ack improvements

Matthias247:
This change contains a first set of improvements to ACK handling, on the way to sending ACKs more efficiently.

In these changes the number of ACKs is actually not yet reduced, and the previous strategies are kept.
It however introduces some refactorings and prerequesites for adding ACK delay successfully:
1. The ACK state is moved into a separate struct
2. ACK delay is reported to peers, in order not to disturb RTT calculations in case ACK delay is applied
3. Pending ACKs are sent with `CONNECTION_CLOSE` frames

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -678,28 +678,51 @@ where
 
             if close {
                 trace!("sending CONNECTION_CLOSE");
-                match self.state {
-                    State::Closed(state::Closed { ref reason }) => {
-                        if space_id == SpaceId::Data {
-                            reason.encode(&mut buf, builder.max_size)
-                        } else {
-                            frame::ConnectionClose {
-                                error_code: TransportErrorCode::APPLICATION_ERROR,
-                                frame_type: None,
-                                reason: Bytes::new(),
+                // Encode ACKs before the ConnectionClose message, to give the receiver
+                // a better approximate on what data has been processed. This is
+                // especially important with ack delay, since the peer might not
+                // have gotten any other ACK for the data earlier on.
+                if !self.spaces[space_id].pending_acks.ranges().is_empty() {
+                    Self::populate_acks(
+                        self.receiving_ecn,
+                        &mut SentFrames::default(),
+                        &mut self.spaces[space_id],
+                        &mut buf,
+                        &mut self.stats,
+                    );
+                }
+
+                // Since there only 64 ACK frames there will always be enough space
+                // to encode the ConnectionClose frame too. However we still have the
+                // check here to prevent crashes if something changes.
+                debug_assert!(
+                    buf.len() + frame::ConnectionClose::SIZE_BOUND < builder.max_size,
+                    "ACKs should leave space for ConnectionClose"
+                );
+                if buf.len() + frame::ConnectionClose::SIZE_BOUND < builder.max_size {
+                    match self.state {
+                        State::Closed(state::Closed { ref reason }) => {
+                            if space_id == SpaceId::Data {
+                                reason.encode(&mut buf, builder.max_size)
+                            } else {
+                                frame::ConnectionClose {
+                                    error_code: TransportErrorCode::APPLICATION_ERROR,
+                                    frame_type: None,
+                                    reason: Bytes::new(),
+                                }
+                                .encode(&mut buf, builder.max_size)
                             }
-                            .encode(&mut buf, builder.max_size)
                         }
+                        State::Draining => frame::ConnectionClose {
+                            error_code: TransportErrorCode::NO_ERROR,
+                            frame_type: None,
+                            reason: Bytes::new(),
+                        }
+                        .encode(&mut buf, builder.max_size),
+                        _ => unreachable!(
+                            "tried to make a close packet when the connection wasn't closed"
+                        ),
                     }
-                    State::Draining => frame::ConnectionClose {
-                        error_code: TransportErrorCode::NO_ERROR,
-                        frame_type: None,
-                        reason: Bytes::new(),
-                    }
-                    .encode(&mut buf, builder.max_size),
-                    _ => unreachable!(
-                        "tried to make a close packet when the connection wasn't closed"
-                    ),
                 }
                 // Don't send another close packet
                 self.close = false;
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -725,13 +725,9 @@ where
             );
             pad_datagram |= sent.requires_padding;
 
-            // If we sent any acks, don't immediately resend them. Setting this even if ack_only is
-            // false needlessly prevents us from ACKing the next packet if it's ACK-only, but saves
-            // the need for subtler logic to avoid double-transmitting acks all the time.
-            // This reset needs to happen before we check whether more data
-            // is available in this space - because otherwise it would return
-            // `true` purely due to the ACKs.
-            self.spaces[space_id].permit_ack_only &= sent.acks.is_empty();
+            if !sent.acks.is_empty() {
+                self.spaces[space_id].pending_acks.acks_sent();
+            }
 
             // Keep information about the packet around until it gets finalized
             sent_frames = Some(sent);
@@ -1456,9 +1452,6 @@ where
         }
         let space = &mut self.spaces[space_id];
         space.pending_acks.insert_one(packet);
-        if space.pending_acks.len() > MAX_ACK_BLOCKS {
-            space.pending_acks.pop_min();
-        }
         if packet >= space.rx_packet {
             space.rx_packet = packet;
             // Update outgoing spin bit, inverting iff we're the client
@@ -2166,7 +2159,9 @@ where
             match frame {
                 Frame::Ack(_) | Frame::Padding | Frame::Close(Close::Connection(_)) => {}
                 _ => {
-                    self.spaces[packet.header.space()].permit_ack_only = true;
+                    self.spaces[packet.header.space()]
+                        .pending_acks
+                        .ack_eliciting_frame_received();
                 }
             }
             // Process frames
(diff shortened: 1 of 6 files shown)


Ralith:
Are you sure the second commit is correct?

RFC9000 [§13.2.5](https://datatracker.ietf.org/doc/html/rfc9000#section-13.2.5):
> An endpoint measures the delays **intentionally** introduced between the
   time the packet with the largest packet number is received and the
   time an acknowledgment is sent.  The endpoint encodes this
   acknowledgment delay in the ACK Delay field of an ACK frame; see
   Section 19.3
(emphasis added)

quinn-proto/src/connection/mod.rs
```diff
@@ -673,28 +673,49 @@ where
...
+                        &mut buf,
+                        &mut self.stats,
+                    );
+                }
+
+                if buf.len() + frame::ConnectionClose::SIZE_BOUND < builder.max_size {
+                    // TODO: Under very rare conditions the ACKs might take up so
+                    // much space that the actual close reason gets swallowed.
+                    // We could send another packet for it, but for this unlikely
+                    // condition we can also just send nothing and let the peer time out.
```
Ralith, line 695: I'm not sure this is actually possible; we know we're sending at most 64 ACK ranges, and `ConnectionClose` already truncates the reason if necessary. Given the minimum MTU, I think we can rely on there being enough space for the critical parts.

quinn-proto/src/tests/mod.rs
```diff
@@ -930,6 +930,28 @@ fn ack_delay_report() {
...
+    pair.client_conn_mut(client_ch)
+        .close(time, VarInt(42), Bytes::new());
+
+    pair.drive();
+
+    let client_acks_2 = pair.client_conn_mut(client_ch).stats().frame_rx.acks;
+    assert!(
+        client_acks_2 > client_acks,
+        "Connection close should send pending ACKs"
+    );
```
Ralith, line 952: If the client's the one sending the close, shouldn't we check the server's received ACK count?

Matthias247: Oh - yes and no - the server ACKs the clients ping, so we need to check whether the client got it.
But obviously we would need to close the server. Interesting that the test still captured the conditions :)

quinn-proto/src/connection/mod.rs
Matthias247: Oh that's true. I initially had nothing, but then felt bad about potential crashes due to having no bounds check.
Probably fine to just replace the comment with "There will always be enough space"?

quinn-proto/src/tests/mod.rs
Matthias247: fixed that by closing the server

Matthias247:
> Are you sure the second commit is correct?

Hmm, that's true. Reading that again this way it seems like we should not include this kind of delay.

quinn-proto/src/connection/mod.rs
Ralith: Sure, maybe plus a `debug_assert`.

Matthias247:
> > Are you sure the second commit is correct?
> 
> Hmm, that's true. Reading that again this way it seems like we should not include this kind of delay.

So I thought a bit more about it:

The delay could be calculated as the time between a packet was received (`insert_one`), and the time the N-th packet unblocks sending (in `on_ack_frame_received`) or the delay timer elapsing unblocks sending a packet. In that case the the delay that is signaled would however only include the delta between the incoming packets, which might be what the spec wants us to there. For the timer elapses case its slightly different, since it's time "time between packet incoming" (on the **endpoint** task) and the time the timer elapses (on the **connection task**), so it wouldn't necessarily account for the endpoint vs connection task latency. But then again the timer would be started relative to the packet reception time, so it should end up being the same as if another packet was received during that time.

If that is done, I think `ack_delay` might be pretty fixed and repeated in retransmissions - which is probably ok since the receiver should only take it into account for the highest acknowledged PN.

Matthias247:
Updated commit 2) and removed the ack delay calculation for now - it writes a static 0us. However it at least introduces the path for getting the delay from the `PendingAcks` struct and encoding the result.

quinn-proto/src/connection/spaces.rs
```diff
@@ -436,6 +433,72 @@ impl SendableFrames {
...
+}
+
+impl Default for PendingAcks {
+    fn default() -> Self {
+        Self {
+            permit_ack_only: false,
+            pending_acks: ArrayRangeSet::new(),
+        }
+    }
+}
```
djc: This can be derived.

quinn-proto/src/connection/spaces.rs
```diff
@@ -436,6 +433,72 @@ impl SendableFrames {
     }
 }
 
+#[derive(Debug)]
+pub(crate) struct PendingAcks {
+    permit_ack_only: bool,
+    pending_acks: ArrayRangeSet,
```
djc, line 439: Maybe rename this to `ranges` now? Feels a bit stuttery otherwise.

quinn-proto/src/connection/mod.rs
```diff
@@ -2747,6 +2737,40 @@ where
         sent
     }
 
+    /// Write pending ACKs into a buffer
+    ///
+    /// This method assumes ACKs are pending, and should only be called if
+    /// `PendingAcks::can_send()` or `!PendingAcks::ranges().is_empty()` returns `true`.
+    fn populate_acks(
```
djc, line 2767: Maybe move this into `PacketSpace`? Methods without a receiver are a bit of an anti-pattern IMO, and we're always trying to move code out of `Connection` since it's already so large. I understand that means that `PacketSpace` now becomes aware of `SentFrames` and `ConnectionStats`, but that seems fine to me.

quinn-proto/src/connection/spaces.rs
Matthias247: done

quinn-proto/src/connection/spaces.rs
Matthias247: done

quinn-proto/src/connection/mod.rs
Matthias247: Although the module is definitely on the too big side, I think `populate_acks` and `populate_packet` are currently extremely similar in style and touch the same data - therefore I don't think it's better to separate them. Also `spaces.rs` currently mostly contains datastructures, and less of serialization methods.

quinn-proto/src/connection/mod.rs
```diff
@@ -2747,6 +2737,40 @@ where
         sent
     }
 
+    /// Write pending ACKs into a buffer
+    ///
+    /// This method assumes ACKs are pending, and should only be called if
+    /// `PendingAcks::can_send()` or `!PendingAcks::ranges().is_empty()` returns `true`.
```
Ralith, line 2766: Would it be clearer to omit mention of `PendingAcks::can_send()`, since it always implies `!PendingAcks::ranges().is_empty()`?

quinn-proto/src/connection/spaces.rs
```diff
@@ -437,6 +437,8 @@ impl SendableFrames {
 pub(crate) struct PendingAcks {
     permit_ack_only: bool,
     ranges: ArrayRangeSet,
+    latest_incoming: Option<Instant>,
```
Ralith, line 440: Let's leave a comment here explaining the intended future application, since it's otherwise mysteriously unused.

quinn-proto/src/transport_parameters.rs
```diff
@@ -139,7 +139,6 @@ impl TransportParameters {
                     .try_into()
                     .expect("setter guarantees this is in-bounds")
             }),
-            max_ack_delay: 0u32.into(),
```
Ralith, line 142: Is there any reason to change this before we actually start sending nonzero values?

quinn-proto/src/connection/mod.rs
Ralith: Note that the non-receiver-ness here is just a workaround because it's called when a space is borrowed. It could be refactored to support a receiver by passing the space ID instead, and restarting the borrow afterwards in the caller, though in this instance the difference is minor.

quinn-proto/src/connection/mod.rs
Matthias247: Maybe - but it's debatable whether it should rather be the other way around and only `can_send()` should be checked in order to suppress always including all acks. But since that change isn't done here - and a reader of only that file might not see that `can_send` will check `!ranges().is_empy()`, I opted for putting both there. I would however be fine removing `can_send` for now if you think that's preferrable

quinn-proto/src/connection/spaces.rs
Matthias247: done

quinn-proto/src/transport_parameters.rs
Matthias247: It doesn't really make a change for this anymore, and is mostly a leftover from the version where the commit reported actual delay. I decided to keep it here to minimize the size of the next change, and since it doesn't really make a difference on where it is - as long as no delay is reported the situation is the same for the peer anyway.

Plus it was kind of weird that that this field is explicitely set to something different than the default value :-)

Ralith: I figured it might be useful to a peer, but realistically it's probably moot.

quinn-proto/src/connection/mod.rs
Ralith: I think `can_send` isn't actually that strong--we use that to determine if we can send an ACK *only* packet, but we still encode ACKs regardless when sending ACK-eliciting packets. Conversely, `ranges()` is always sufficient to sanely call the method, even if we don't actually want to do that under all circumstances for other reasons.

Matthias247: ok - updated the comment

djc (approved):

PR #1150, opened 2021-06-22
Upgrade to rustls 0.20

djc:


quinn-proto/src/crypto/rustls.rs
```diff
@@ -157,7 +157,7 @@ impl crypto::Session for TlsSession {
     }
 
     fn write_handshake(&mut self, buf: &mut Vec<u8>) -> Option<Keys<Self>> {
-        let keys = self.write_hs(buf)?;
+        let secrets = self.write_hs(buf)?;
```
djc, line 160: @Ralith any suggestions for how to deal with this?

Ralith: Deal with what, exactly? At a glance it looks like the desired keys are available via `Secrets::packet_keys`, but maybe I'm missing something? We'd also want to hold onto the 1-RTT `secrets` for key update purposes.

djc: Mostly, where do we stick `Secrets`? Does it need yet another associated type on the `crypto::Session`?

Ralith: Can we store it inside `struct TlsSession` directly?

djc: Yeah, I think that might work. I forgot we had that, I remembered us having the `SessionKind` enum as the top level thing.

quinn-proto/src/crypto/rustls.rs
```diff
@@ -364,42 +431,24 @@ impl crypto::PacketKey for PacketKey {
...
+            .decrypt_in_place(packet, &*header, payload.as_mut())
+            .map_err(|_| CryptoError)?;
+        let plain_len = plain.len();
+        payload.truncate(plain_len);
         Ok(())
     }
 
     fn tag_len(&self) -> usize {
-        self.key.algorithm().tag_len()
+        self.tag_len()
```
Matthias247, line 443: looks like a recursive call which will stack overflow

quinn-proto/src/crypto/rustls.rs
```diff
@@ -364,42 +431,24 @@ impl crypto::PacketKey for PacketKey {
...
     fn confidentiality_limit(&self) -> u64 {
-        let cipher = self.key.algorithm();
-        if cipher == &aead::AES_128_GCM || cipher == &aead::AES_256_GCM {
-            2u64.pow(23)
-        } else if cipher == &aead::CHACHA20_POLY1305 {
-            u64::MAX
-        } else {
-            panic!("unknown cipher")
-        }
+        self.confidentiality_limit()
```
Matthias247, line 447: looks like a recursive call which will stack overflow

quinn-proto/src/crypto/rustls.rs
```diff
@@ -364,42 +431,24 @@ impl crypto::PacketKey for PacketKey {
...
     fn integrity_limit(&self) -> u64 {
-        let cipher = self.key.algorithm();
-        if cipher == &aead::AES_128_GCM || cipher == &aead::AES_256_GCM {
-            2u64.pow(52)
-        } else if cipher == &aead::CHACHA20_POLY1305 {
-            2u64.pow(36)
-        } else {
-            panic!("unknown cipher")
-        }
+        self.integrity_limit()
```
Matthias247, line 451: looks like a recursive call which will stack overflow

quinn-proto/src/crypto/rustls.rs
```diff
@@ -411,8 +460,8 @@ impl crypto::PacketKey for PacketKey {
 /// This list prefers AES ciphers, which are hardware accelerated on most platforms.
 /// This list can be removed if the rustls dependency is updated to a new version
 /// which contains the linked change.
-static QUIC_CIPHER_SUITES: [&rustls::SupportedCipherSuite; 3] = [
-    &rustls::ciphersuite::TLS13_AES_256_GCM_SHA384,
-    &rustls::ciphersuite::TLS13_AES_128_GCM_SHA256,
-    &rustls::ciphersuite::TLS13_CHACHA20_POLY1305_SHA256,
+static QUIC_CIPHER_SUITES: [rustls::SupportedCipherSuite; 3] = [
```
Matthias247, line 463: see comment above - can also be removed if you like (or being kept in order to keep quinn behavior independent of rustls defaults)

quinn-proto/src/crypto/rustls.rs
djc: It isn't; inherent impls get priority over trait impls.

Cargo.toml
```diff
@@ -7,3 +7,9 @@ debug = true
 
 [profile.release]
 debug = true
+
+[patch."crates-io"]
```
Matthias247, line 11: I think this is fine for making sure CI works. But could you please consider not to merge this before the actual rustls release? For some people the github dependencies don't work, and it's a not a great place to have this non-buildable state for an indefinite time (until rustls actually gets published)

interop/src/main.rs
```diff
@@ -257,19 +257,23 @@ impl State {
...
+        let verifier = InteropVerifier(Arc::new(Mutex::new(false)));
+        let mut tls_config = rustls::ClientConfig::builder()
+            .with_safe_default_cipher_suites()
+            .with_safe_default_kx_groups()
+            .with_protocol_versions(&[&rustls::version::TLS13])
+            .unwrap()
+            .with_custom_certificate_verifier(Arc::new(verifier))
+            .with_no_client_auth();
+
         tls_config.enable_early_data = true;
```
Matthias247, line 276: No setter on the builder for that and alpn?

perf/src/bin/perf_client.rs
```diff
@@ -102,27 +100,20 @@ async fn run(opt: Opt) -> Result<()> {
...
-    tls_config
-        .ciphersuites
-        .push(&rustls::ciphersuite::TLS13_AES_128_GCM_SHA256);
-    tls_config
-        .ciphersuites
-        .push(&rustls::ciphersuite::TLS13_AES_256_GCM_SHA384);
-    tls_config
-        .ciphersuites
-        .push(&rustls::ciphersuite::TLS13_CHACHA20_POLY1305_SHA256);
+    let mut crypto = rustls::ClientConfig::builder()
```
Matthias247, line 103: Looks so much nicer than the `Arc::get_mut` dance

quinn-proto/src/crypto/rustls.rs
Matthias247: Interesting. Didn't even knew one can have the same method name as a direct impl and trait

Cargo.toml
djc: I'm fine with deferring it for now.

interop/src/main.rs
djc: Nope.

quinn-proto/src/crypto/types.rs
```diff
@@ -80,9 +83,11 @@ impl std::iter::FromIterator<Certificate> for CertificateChain {
...
 }
 
-impl From<Vec<rustls::Certificate>> for CertificateChain {
-    fn from(certs: Vec<rustls::Certificate>) -> Self {
-        Self { certs }
+impl From<&[rustls::Certificate]> for CertificateChain {
+    fn from(certs: &[rustls::Certificate]) -> Self {
+        Self {
+            certs: certs.to_vec(),
+        }
```
Ralith, line 90: Why this change? `From<Vec<_>>` is nice in that a user who already has a `Vec` can avoid some work, and others can call `to_vec` as needed.

quinn-proto/src/crypto/rustls.rs
```diff
@@ -391,12 +391,8 @@ impl crypto::PacketKey for PacketKey {
...
 /// Cipher suites suitable for QUIC
 ///
-/// The list is equivalent to TLS1.3 ciphers.
-/// It matches the rustls prefernce list that was introduced with
-/// https://github.com/ctz/rustls/commit/7117a805e0104705da50259357d8effa7d599e37.
-/// This list prefers AES ciphers, which are hardware accelerated on most platforms.
-/// This list can be removed if the rustls dependency is updated to a new version
-/// which contains the linked change.
+/// This includes only the TLS 1.3 cipher suites, so that we make sure no TLS 1.2
+/// crypto code needs to be linked into projects that (only) use quinn.
```
Ralith, line 395: Can't we drop this entirely now that rustls has the fixed priority list?

Ralith (commented):
I like the reduced magic around configuration!

quinn-proto/src/crypto/types.rs
djc: Fair enough, reverted. (I think I changed this to make it easier to feed directly from `TlsSession::peer_certificates()` into `CertificateChain`, but adding a `to_vec()` there seems fine.)

quinn-proto/src/crypto/rustls.rs
djc: Yup, done.

djc:
Test failure is due to https://github.com/hawkw/sharded-slab/issues/60.

PR #1151, opened 2021-06-24
feat(cca): BBR

### LINKED ISSUE #693: BBR congestion control

FrankSpitulski:
implements BBR congestion control based on google's quiche. There are some differences in implementation due to chromium's CCA having a different API, especially around sending vs acking packets.

It's a pretty rough first implementation but from my testing it seems to work well. It is very resistant to packet loss especially in high BDP situations.

Please critique and suggest changes as needed.

I've added additional fields and traits for debugging using stats which can be removed if needed.

Closes #693

### DIFF
--- a/quinn-proto/src/congestion.rs
+++ b/quinn-proto/src/congestion.rs
@@ -1,27 +1,62 @@
 //! Logic for controlling the rate at which data is sent
 
-use std::time::{Duration, Instant};
+use crate::connection::paths::RttEstimator;
+use std::any::Any;
+use std::time::Instant;
 
+mod bbr;
 mod cubic;
 mod new_reno;
 
+pub use bbr::{Bbr, BbrConfig};
 pub use cubic::{Cubic, CubicConfig};
 pub use new_reno::{NewReno, NewRenoConfig};
 
 /// Common interface for different congestion controllers
 pub trait Controller: Send {
+    /// One or more packets were just sent
+    #[allow(unused_variables)]
+    fn on_sent(&mut self, now: Instant, bytes: u64, last_packet_number: u64) {}
+
     /// Packet deliveries were confirmed
     ///
     /// `app_limited` indicates whether the connection was blocked on outgoing
     /// application data prior to receiving these acknowledgements.
-    fn on_ack(&mut self, now: Instant, sent: Instant, bytes: u64, app_limited: bool, rtt: Duration);
+    #[allow(unused_variables)]
+    fn on_ack(
+        &mut self,
+        now: Instant,
+        sent: Instant,
+        bytes: u64,
+        app_limited: bool,
+        rtt: &RttEstimator,
+    ) {
+    }
+
+    /// Packets are acked in batches, all with the same `now` argument. This indicates one of those batches has completed.
+    #[allow(unused_variables)]
+    fn on_end_acks(
+        &mut self,
+        now: Instant,
+        in_flight: u64,
+        app_limited: bool,
+        largest_packet_num_acked: Option<u64>,
+    ) {
+    }
 
     /// Packets were deemed lost or marked congested
     ///
     /// `in_persistent_congestion` indicates whether all packets sent within the persistent
     /// congestion threshold period ending when the most recent packet in this batch was sent were
     /// lost.
-    fn on_congestion_event(&mut self, now: Instant, sent: Instant, is_persistent_congestion: bool);
+    /// `lost_bytes` indicates how many bytes were lost. This value will be 0 for ECN triggers.
+    fn on_congestion_event(
+        &mut self,
+        now: Instant,
+        sent: Instant,
+        is_persistent_congestion: bool,
+        lost_bytes: u64,
+    );
 
     /// Number of ack-eliciting bytes that may be in flight
     fn window(&self) -> u64;
@@ -31,6 +66,9 @@ pub trait Controller: Send {
 
     /// Initial congestion window
     fn initial_window(&self) -> u64;
+
+    /// Returns Self for use in down-casting to extract implementation details
+    fn into_any(self: Box<Self>) -> Box<dyn Any>;
 }
 
 /// Constructs controllers on demand
--- a/quinn-proto/src/congestion/cubic.rs
+++ b/quinn-proto/src/congestion/cubic.rs
@@ -1,7 +1,9 @@
+use std::any::Any;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 use super::{Controller, ControllerFactory};
+use crate::connection::paths::RttEstimator;
 use std::cmp;
 
 /// CUBIC Constants.
@@ -91,7 +93,7 @@ impl Controller for Cubic {
         sent: Instant,
         bytes: u64,
         app_limited: bool,
-        rtt: Duration,
+        rtt: &RttEstimator,
     ) {
         if app_limited
             || self
@@ -127,12 +129,12 @@ impl Controller for Cubic {
             // w_cubic(t + rtt)
             let w_cubic = self
                 .cubic_state
-                .w_cubic(t + rtt, self.config.max_datagram_size);
+                .w_cubic(t + rtt.get(), self.config.max_datagram_size);
 
             // w_est(t)
             let w_est = self
                 .cubic_state
-                .w_est(t, rtt, self.config.max_datagram_size);
+                .w_est(t, rtt.get(), self.config.max_datagram_size);
 
             let mut cubic_cwnd = self.window;
 
@@ -165,6 +167,7 @@ impl Controller for Cubic {
         now: Instant,
         sent: Instant,
         _is_persistent_congestion: bool,
+        _lost_bytes: u64,
     ) {
         if self
             .recovery_start_time
@@ -205,6 +208,10 @@ impl Controller for Cubic {
 
     fn initial_window(&self) -> u64 {
         self.config.initial_window
+    }
+
+    fn into_any(self: Box<Self>) -> Box<dyn Any> {
+        self
     }
 }
 
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -48,7 +48,7 @@ mod pacing;
 mod packet_builder;
 use packet_builder::PacketBuilder;
 
-mod paths;
+pub(crate) mod paths;
 use paths::PathData;
 
 mod send_buffer;
@@ -74,6 +74,7 @@ pub use streams::{
 };
 
 mod timer;
+use crate::congestion::Controller;
 use timer::{Timer, TimerTable};
 
 /// Protocol state and logic for a single QUIC connection
@@ -127,7 +128,6 @@ pub struct Connection {
     /// The "real" local IP address which was was used to receive the initial packet.
     /// This is only populated for the server case, and if known
     local_ip: Option<IpAddr>,
-
     path: PathData,
     prev_path: Option<PathData>,
     state: State,
@@ -757,7 +757,11 @@ impl Connection {
             if pad_datagram {
                 builder.pad_to(MIN_INITIAL_SIZE);
             }
+            let last_packet_number = builder.exact_number;
             builder.finish_and_track(now, self, sent_frames, &mut buf);
+            self.path
+                .congestion
+                .on_sent(now, buf.len() as u64, last_packet_number);
         }
 
         self.app_limited = buf.is_empty() && !congestion_blocked;
@@ -1083,6 +1087,11 @@ impl Connection {
         self.path.rtt.get()
     }
 
+    /// Current state of this connection's congestion controller, for debugging purposes
+    pub fn congestion_state(&self) -> &dyn Controller {
+        self.path.congestion.as_ref()
+    }
+
     fn on_ack_received(
         &mut self,
         now: Instant,
@@ -1131,6 +1140,13 @@ impl Connection {
                 self.on_packet_acked(now, space, info);
             }
         }
+
+        self.path.congestion.on_end_acks(
+            now,
+            self.in_flight.bytes,
+            self.app_limited,
+            self.spaces[space].largest_acked_packet,
+        );
 
         if new_largest && ack_eliciting_acked {
             let ack_delay = if space != SpaceId::Data {
@@ -1196,7 +1212,7 @@ impl Connection {
                 self.stats.path.congestion_events += 1;
                 self.path
                     .congestion
-                    .on_congestion_event(now, largest_sent_time, false);
+                    .on_congestion_event(now, largest_sent_time, false, 0);
             }
         }
     }
@@ -1213,7 +1229,7 @@ impl Connection {
                 info.time_sent,
                 info.size.into(),
                 self.app_limited,
-                self.path.rtt.get(),
+                &self.path.rtt,
             );
         }
 
@@ -1290,6 +1306,7 @@ impl Connection {
         let lost_send_time = now - loss_delay;
         let largest_acked_packet = self.spaces[pn_space].largest_acked_packet.unwrap();
         let packet_threshold = self.config.packet_threshold as u64;
+        let mut size_of_lost_packets = 0u64;
 
         let space = &mut self.spaces[pn_space];
         space.loss_time = None;
@@ -1297,6 +1314,7 @@ impl Connection {
             if info.time_sent <= lost_send_time || largest_acked_packet >= packet + packet_threshold
             {
                 lost_packets.push(packet);
+                size_of_lost_packets += info.size as u64;
             } else {
                 let next_loss_time = info.time_sent + loss_delay;
                 space.loss_time = Some(
@@ -1336,6 +1354,7 @@ impl Connection {
                     now,
                     largest_lost_sent,
                     in_persistent_congestion,
+                    size_of_lost_packets,
                 );
             }
         }
@@ -2945,7 +2964,7 @@ impl Connection {
 
     /// Number of bytes worth of non-ack-only packets that may be sent
     #[cfg(test)]
-    pub(crate) fn congestion_state(&self) -> u64 {
+    pub(crate) fn congestion_window(&self) -> u64 {
         self.path
             .congestion
             .window()
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -27,6 +27,7 @@ use crate::{
     send_stream::{SendStream, WriteError},
     ConnectionEvent, EndpointEvent, VarInt,
 };
+use proto::congestion::Controller;
 
 /// In-progress connection attempt future
 #[derive(Debug)]
@@ -438,6 +439,15 @@ impl Connection {
     /// Returns connection statistics
     pub fn stats(&self) -> ConnectionStats {
         self.0.lock("stats").inner.stats()
+    }
+
+    /// Current state of the congestion control algorithm, for debugging purposes
+    pub fn congestion_state(&self) -> Box<dyn Controller> {
+        self.0
+            .lock("congestion_state")
+            .inner
+            .congestion_state()
+            .clone_box()
     }
 
     /// Parameters negotiated during the handshake
--- a/quinn-proto/src/congestion/bbr/bw_estimation.rs
+++ b/quinn-proto/src/congestion/bbr/bw_estimation.rs
@@ -1,0 +1,121 @@
+use std::fmt::{Debug, Display, Formatter};
+use std::time::{Duration, Instant};
+
+use super::min_max::MinMax;
+
+#[derive(Clone, Debug)]
+pub(crate) struct BandwidthEstimation {
+    total_acked: u64,
+    prev_total_acked: u64,
+    acked_time: Option<Instant>,
+    prev_acked_time: Option<Instant>,
+    total_sent: u64,
+    prev_total_sent: u64,
+    sent_time: Option<Instant>,
+    prev_sent_time: Option<Instant>,
+    max_filter: MinMax,
+    acked_at_last_window: u64,
+}
+
+impl Default for BandwidthEstimation {
+    fn default() -> Self {
+        BandwidthEstimation {
+            total_acked: 0,
+            prev_total_acked: 0,
+            acked_time: None,
+            prev_acked_time: None,
+            total_sent: 0,
+            prev_total_sent: 0,
+            sent_time: None,
+            prev_sent_time: None,
+            max_filter: MinMax::new(10),
+            acked_at_last_window: 0,
+        }
+    }
+}
+
+impl Display for BandwidthEstimation {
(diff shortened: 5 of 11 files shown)


quinn-proto/src/congestion.rs
```diff
@@ -1,27 +1,65 @@
...
+mod bbr_min_max;
 mod cubic;
 mod new_reno;
 
+pub use bbr::{BBRConfig, BBR};
 pub use cubic::{Cubic, CubicConfig};
 pub use new_reno::{NewReno, NewRenoConfig};
 
+/// To help differentiate between packet sizes and packet numbers
+pub type PacketNumber = u64;
```
Matthias247, line 18: typedefs are not that useful since they still allow wrong types being used. Newtypes (like `struct PacketNumber(u64)`) are more useful. But I think if such a change is applied, it should be done in the whole crate.

quinn-proto/src/congestion.rs
```diff
@@ -1,27 +1,65 @@
...
+    fn on_sent(&mut self, _now: Instant, _bytes: u64) {}
+
     /// Packet deliveries were confirmed
     ///
     /// `app_limited` indicates whether the connection was blocked on outgoing
     /// application data prior to receiving these acknowledgements.
-    fn on_ack(&mut self, now: Instant, sent: Instant, bytes: u64, app_limited: bool, rtt: Duration);
+    fn on_ack(
+        &mut self,
+        _now: Instant,
```
Matthias247, line 30: It's a trait. No need to add `_` for unused parameters

quinn-proto/src/congestion.rs
```diff
@@ -1,27 +1,65 @@
...
+        _sent: Instant,
+        _is_persistent_congestion: bool,
+    ) {
+    }
+
+    fn on_loss(&mut self, _now: Instant, _sent: Instant, _bytes: u64) {}
+
+    fn update_last_sent(&mut self, _packet_number: u64) {}
```
Matthias247, line 62: here `_packet_number` is not a `PacketNumber`

quinn-proto/src/congestion.rs
```diff
@@ -1,27 +1,65 @@
...
-    fn on_congestion_event(&mut self, now: Instant, sent: Instant, is_persistent_congestion: bool);
+    fn on_congestion_event(
+        &mut self,
+        _now: Instant,
+        _sent: Instant,
+        _is_persistent_congestion: bool,
+    ) {
+    }
+
+    fn on_loss(&mut self, _now: Instant, _sent: Instant, _bytes: u64) {}
```
Matthias247, line 60: from an interface point of view the difference between `on_congestion_event` and `on_loss` is not super clear, since a congestion event can be loss too.

quinn-proto/src/congestion/bbr.rs
```diff
@@ -0,0 +1,672 @@
...
+use crate::congestion::PacketNumber;
+
+use super::{Controller, ControllerFactory};
+use crate::congestion::bbr_min_max::MinMax;
+use crate::connection::paths::RttEstimator;
+
+// Constants based on TCP defaults.
+// The minimum CWND to ensure delayed acks don't reduce bandwidth measurements.
+// Does not inflate the pacing rate.
+const kDefaultMinimumCongestionWindow: u64 = 4 * MAX_DATAGRAM_SIZE;
```
Matthias247, line 15: Those should be using Rust style, like `const DEFAULT_MINIMUM_CONGESTION_WINDOW`

quinn-proto/src/congestion/bbr.rs
```diff
@@ -0,0 +1,672 @@
...
+    // Allow an extra outstanding byte for each byte acknowledged.
+    CONSERVATION,
+    // Allow two extra outstanding bytes for each byte acknowledged (slow
+    // start).
+    GROWTH,
+}
+
+impl RecoveryState {
+    pub fn in_recovery(&self) -> bool {
+        !matches!(self, RecoveryState::NOT_IN_RECOVERY)
```
Matthias247, line 66: Since this a "simple enum", `self != RecoveryState::NOT_IN_RECOVERY` will be fine. And clippy will probably tell you that the method should take `self` by value since its a small copyable value

quinn-proto/src/congestion/bbr.rs
```diff
@@ -0,0 +1,672 @@
...
+    }
+}
+
+#[derive(Debug, Copy, Clone)]
+struct AckAggregationState {
+    bbr_max_ack_height: MinMax,
```
Matthias247, line 72: no point in prefixing them with `bbr_` since everything in this file is about bbr?

quinn-proto/src/congestion/bbr.rs
```diff
@@ -0,0 +1,672 @@
...
+        &mut self,
+        newly_acked_bytes: u64,
+        now: Instant,
+        round: u64,
+        max_bandwidth: u64,
+    ) -> u64 {
+        // Compute how many bytes are expected to be delivered, assuming max
+        // bandwidth is correct.
+        let expected_bytes_acked = max_bandwidth
+            * (now - self.bbr_aggregation_epoch_start_time.unwrap_or(now)).as_micros() as u64
```
Matthias247, line 88: you probably want `now.saturating_duration_since(...)` for safety reasons and conciseness

quinn-proto/src/congestion/bbr.rs
```diff
@@ -0,0 +1,672 @@
...
+    fn update_ack_aggregation_bytes(
+        &mut self,
+        newly_acked_bytes: u64,
+        now: Instant,
+        round: u64,
+        max_bandwidth: u64,
+    ) -> u64 {
+        // Compute how many bytes are expected to be delivered, assuming max
+        // bandwidth is correct.
+        let expected_bytes_acked = max_bandwidth
```
Matthias247, line 87: if the time between those calls is small, this might often end up being zero. Is this expected?

quinn-proto/src/congestion/bbr_min_max.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    fn fill(&mut self, sample: MinMaxSample) {
+        self.samples.fill(sample);
+    }
+
+    pub fn reset(&mut self) {
+        self.fill(Default::default())
+    }
+
+    /// unused, but still translated from original source for completeness
```
Matthias247, line 64: would need an `#[allow(dead_code)]`.  But keeping unused code around is usually an antipattern

quinn/src/connection.rs
```diff
@@ -458,6 +459,10 @@ where
         self.0.lock("stats").inner.stats()
     }
 
+    pub fn cca(&self) -> Box<dyn Controller> {
```
Matthias247, line 462: `congestion_controller`?  `cca` seems  only obvious for people who work on it. And what's the reason for making this one public - with an expensive copy instead of just returning a reference (which is problematic due to the lock)?

Matthias247 (commented):
Very cool to see BBR is integrateable and to hear it performs good! Definitely interested to see this landing.

I did a short skim on this, but comments are mostly stylistic since I don't have any background on BBR.

Since this is rather complex, I would be interested to see some tests as part of these changes. Ideally also in the form that show the benefits compared ot the other congestion controllers.

quinn-proto/src/congestion.rs
FrankSpitulski: is that something you'd be interested in? I found it very helpful while developing do know which was which even if it was just in the function def. I had confused the two a few times when developing before I added this.

quinn-proto/src/congestion.rs
FrankSpitulski: my bad, cargo fix was complaining so I just let it do its thing.

quinn-proto/src/congestion.rs
FrankSpitulski: that's right, a congestion event is both while a loss is just loss

quinn-proto/src/congestion.rs
FrankSpitulski: fixed

quinn-proto/src/congestion/bbr.rs
FrankSpitulski: will remove- I just kept those from the C impl

quinn-proto/src/congestion/bbr.rs
FrankSpitulski: yes, it's just trying to find the max bandwidth overall so it can swallow 0s no issue. I'd welcome more scrutiny on this of course but that's what the google quiche impl seems to do.

quinn-proto/src/congestion/bbr_min_max.rs
FrankSpitulski: you'd prefer to remove it then? I'm not attached.

quinn/src/connection.rs
FrankSpitulski: this whole section was just for my debugging ease. any changes at all here are fine with me. my only goal was to be able to print out the internal variables once every few seconds while running a transfer.

quinn-proto/src/congestion.rs
djc: Yeah, I think that would be good to have (but it should likely be a separate PR).

quinn-proto/src/congestion/bbr.rs
djc: If we can avoid the need for deriving `PartialEq` (if this is the only use), I'd just as soon use `matches!()` here.

quinn-proto/src/congestion.rs
FrankSpitulski: okay, I'll remove it from this one

quinn-proto/src/congestion.rs
FrankSpitulski: it's still not happy about it. I think it's because of the empty default impl. I'll add an #[allow(unused_variables)].

quinn-proto/src/congestion.rs
FrankSpitulski: fixed

quinn-proto/src/congestion/bbr.rs
FrankSpitulski: fixed

quinn-proto/src/congestion/bbr.rs
FrankSpitulski: fixed, but it doesn't look any more concise to me. can you make sure it's what you intended?

quinn-proto/src/congestion/bbr_min_max.rs
Ralith: I'd prefer to drop it, since otherwise it's at high risk of bitrot. Maybe leave a comment on `update_max` that an inverse version can be defined, and cite a reference for it.

quinn-proto/src/congestion/cubic.rs
```diff
@@ -179,6 +181,7 @@ impl Controller for Cubic {
 
         self.recovery_start_time = Some(now);
 
+        // clippy doesn't like the duplicate line, but it would change behaviour to move it
```
Ralith, line 184: nit: has the false positive been reported to clippy?

quinn-proto/src/congestion/cubic.rs
```diff
@@ -163,6 +164,7 @@ impl Controller for Cubic {
         }
     }
 
+    #[allow(clippy::branches_sharing_code)] // self.cubic_state.w_last_max = self.cubic_state.w_max;
```
Ralith, line 167:
```suggestion
    #[allow(clippy::branches_sharing_code)] // False positive, see discussion inline
```

quinn-proto/src/connection/mod.rs
```diff
@@ -131,8 +131,8 @@ where
     /// The "real" local IP address which was was used to receive the initial packet.
     /// This is only populated for the server case, and if known
     local_ip: Option<IpAddr>,
-
-    path: PathData,
+    /// public so that path data can be exposed to stats
+    pub path: PathData,
```
Ralith, line 135: This shouldn't be public. What does "exposed to stats" mean? I don't see any interfaces that would consume this.

quinn-proto/src/congestion.rs
```diff
@@ -1,27 +1,67 @@
...
 mod new_reno;
 
+pub use bbr::{BBRConfig, BBR};
 pub use cubic::{Cubic, CubicConfig};
 pub use new_reno::{NewReno, NewRenoConfig};
 
 /// Common interface for different congestion controllers
-pub trait Controller: Send {
+pub trait Controller: Send + Debug {
+    /// Packets were sent, called when the send buffer is sent, not when each packet is sent.
```
Ralith, line 19: Would this still be in line with the intended semantics?
```suggestion
    /// One or more packets were just sent
```

quinn-proto/src/congestion/bbr.rs
```diff
@@ -0,0 +1,653 @@
...
+            in_flight,
+        );
+
+        self.bbr_state.prev_in_flight_count = in_flight;
+        self.bbr_state.loss_state.reset();
+    }
+
+    fn on_loss(&mut self, _now: Instant, _sent: Instant, bytes: u64) {
+        self.bbr_state.loss_state.lost_bytes += bytes;
+    }
```
Ralith, line 577: I'm surprised BBR completely ignores ECN. You're certain that's intended?

quinn-proto/src/congestion.rs
```diff
@@ -1,27 +1,67 @@
...
     /// congestion threshold period ending when the most recent packet in this batch was sent were
     /// lost.
-    fn on_congestion_event(&mut self, now: Instant, sent: Instant, is_persistent_congestion: bool);
+    #[allow(unused_variables)]
+    fn on_congestion_event(&mut self, now: Instant, sent: Instant, is_persistent_congestion: bool) {
+    }
+
+    /// Packets were deemed lost. This is only invoked for loss, not congestion signals. `on_congestion_event` is still invoked in the event of loss.
+    #[allow(unused_variables)]
+    fn on_loss(&mut self, now: Instant, sent: Instant, bytes: u64) {}
```
Ralith, line 60: Would it be simpler to add a `lost_bytes: u64` argument to `on_congestion_event`?

quinn-proto/src/congestion.rs
```diff
@@ -1,27 +1,67 @@
...
+
+    /// Updates the last packet number that was sent
+    #[allow(unused_variables)]
+    fn update_last_sent(&mut self, packet_number: u64) {}
```
Ralith, line 64: Could this be folded into `on_sent`?

quinn-proto/src/congestion/bbr.rs
```diff
@@ -0,0 +1,653 @@
...
+        // Include the bytes most recently acknowledged to account for stretch acks.
+        self.aggregation_epoch_bytes += newly_acked_bytes;
+        let diff = self.aggregation_epoch_bytes - expected_bytes_acked;
+        self.max_ack_height.update_max(round, diff);
+        diff
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct State {
```
Ralith, line 101: Does this really need to be `pub`?

quinn-proto/src/congestion/bbr.rs
```diff
@@ -0,0 +1,653 @@
...
+        if self.round_wo_bw_gain >= K_ROUND_TRIPS_WITHOUT_GROWTH_BEFORE_EXITING_STARTUP as u64
+            || (self.recovery_state.in_recovery())
+        {
+            self.is_at_full_bandwidth = true;
+        }
+    }
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct LossState {
```
Ralith, line 420: Does this really need to be `pub`?

quinn-proto/src/congestion/bbr.rs
```diff
@@ -0,0 +1,653 @@
...
+    fn clone_box(&self) -> Box<dyn Controller> {
+        Box::new(self.clone())
+    }
+
+    fn initial_window(&self) -> u64 {
+        self.config.initial_window
+    }
+}
+
+/// Configuration for the `BBR` congestion controller
```
Ralith, line 603:
```suggestion
/// Configuration for the [`BBR`] congestion controller
```

quinn/src/connection.rs
```diff
@@ -458,6 +459,11 @@ where
         self.0.lock("stats").inner.stats()
     }
 
+    /// Current state of the congestion control algorithm
+    pub fn congestion(&self) -> Box<dyn Controller> {
```
Ralith, line 463:
```suggestion
    pub fn congestion_state(&self) -> Box<dyn Controller> {
```
How can a user downcast this? Do we really want to expose it, or would it be better off as something you patch in locally when needed for development?

quinn-proto/src/congestion/bbr.rs
```diff
@@ -0,0 +1,653 @@
...
+    pub fn reset(&mut self) {
+        self.lost_bytes = 0;
+    }
+
+    pub fn has_losses(&self) -> bool {
+        self.lost_bytes != 0
+    }
+}
+
+/// A not so simple congestion controller to overcome high latency and packet loss
```
Ralith, line 434: Let's cite the code this is based on, as well as the current revisions of the papers documenting the algorithm. A brief summary of how it compares to Cubic (e.g. "aims for reduced buffer bloat and improved performance over high bandwidth-delay product networks", if that's accurate) would also be nice.

I'm also not sure any of us are currently able to review the *correctness* of this implementation, and afaik its development is still ongoing upstream, so maybe we should also clearly (i.e. in the first line) designate it as experimental.

Ralith (commented):
Thanks, this looks pretty cool! Reviewed mostly for style and API impact for now, since I don't presently have time to dig deep into correctness, and it's not clear what the best reference would be to derive such a review from either.

quinn-proto/src/congestion/cubic.rs
FrankSpitulski: it has not. can I get a second opinion on whether clippy is right here? I'd rather not cite clippy as having an error to the maintainers only for them to fix my code for me.

quinn/src/connection.rs
FrankSpitulski: I don't think it should just be patched in locally. Hopefully I'm not the only one who will do development on CCAs. As for downcasting, that's not something I've needed to do. I just print out the debug representation.

quinn-proto/src/congestion/bbr.rs
FrankSpitulski: I think BBR is pretty much done, but BBR2 is under development

quinn-proto/src/congestion/bbr_min_max.rs
FrankSpitulski: done

quinn-proto/src/congestion/bbr.rs
FrankSpitulski: yes. BBR ignores it. BBR2 does not.

quinn-proto/src/congestion.rs
FrankSpitulski: yes, I will make the change

quinn-proto/src/congestion.rs
FrankSpitulski: collapsed, but please check how it's invoked in quinn-proto/src/connection/mod.rs to make sure it still makes sense

quinn-proto/src/congestion/bbr.rs
FrankSpitulski: fixed

quinn-proto/src/congestion/bbr.rs
FrankSpitulski: fixed

quinn-proto/src/congestion/bbr.rs
FrankSpitulski: changed description

quinn-proto/src/connection/mod.rs
FrankSpitulski: it's public so that I can do `self.0.lock("congestion").inner.path.congestion.clone_box()` in congestion_state()

FrankSpitulski: I've refactored it to make this not public.

quinn/src/connection.rs
Ralith: If the intended use is just for diagnostics during development, we should at least document that fact.

> I just print out the debug representation.

How does that work? `Controller` doesn't imply `Debug`.

quinn-proto/src/congestion/cubic.rs
```diff
@@ -163,11 +164,13 @@ impl Controller for Cubic {
         }
     }
 
+    #[allow(clippy::branches_sharing_code)] // False positive, see discussion inline
```
Ralith, line 167: Wait, we also have an `#[allow]` on the actual `if` expression. Is it not effective? If so, we should remove it.

quinn-proto/src/congestion/cubic.rs
Ralith: I see there's actually already a report linked--not sure how I missed that.

quinn/src/connection.rs
FrankSpitulski: controller does imply debug since controller implements send + debug

quinn-proto/src/congestion.rs
FrankSpitulski: combined into one

quinn/src/connection.rs
Ralith: Oh, right, that was added in this PR. Feels like a bit of a hack, but I guess it's harmless.

quinn-proto/src/congestion/cubic.rs
FrankSpitulski: I just rebased. Seems like @Matthias247 already covered it. I'll remove all related changes since it isn't related to BBR.

quinn/src/connection.rs
FrankSpitulski: it's definitely a hack- any better way to achieve printing this data out given a connection is  welcome

quinn/src/connection.rs
FrankSpitulski: continued discussion here https://github.com/quinn-rs/quinn/pull/1151#discussion_r667590209

FrankSpitulski:
@Ralith is this still in the pipeline?

Ralith:
Yeah, sorry, just on limited bandwidth at the moment.

djc:
@FrankSpitulski this is ready for review, right? Mind marking it as such?

Ralith (dismissed):
Thanks!

FrankSpitulski:
had to rebase @Ralith

Ralith (approved):

PR #1155, opened 2021-07-06
Remove quinn-h3

djc:
quinn-h3 has been mostly unmaintained since @stammw decided to work with @seanmonstar on https://github.com/hyperium/h3/. Although those efforts seem to have yet to reach the same level of functionality that quinn-h3 has, keeping up with maintenance on quinn-h3 has become a burden that does not make sense when the project has no future.

Most interop peers seem to time out, though we can still get a VHDCRZ from quiche.


Ralith (approved):
I believe automated interop testing has in any case moved on to other formats, e.g. https://github.com/marten-seemann/quic-interop-runner.

moderation:
I was using quinn-h3 for client testing at https://dev.to/moderation/experiments-with-h3-clients-envoy-52gj. Sad to see it go. You should update the `README.md` which still refers to the deleted code

djc:
Feel free to fork it if it's still useful for you!

Issue #1154, opened 2021-07-06
Min MTU


BiagioFesta:
Currently (`3f908a`), *quinn* uses [`const MIN_MTU: u16 = 1232;`](https://github.com/quinn-rs/quinn/blob/3f908a2c8c1ec4585212d776fafe536ea17bf2b4/quinn-proto/src/lib.rs#L313).

Maybe I am missing something, but looking at it, it seems this value is used more as *"datagram size"*.
Indeed, running the quinn server example (over IPv4):

* Client Invocation:
```
cargo run --example client https:/server:1234/README.md
```

* Server log:
```
[...]
Jul 06 02:59:45.746 TRACE drive{id=0}: quinn_proto::connection: sending 1232 bytes in 1 datagrams
[...]
```

The server sends a UDP datagram payload with 1232 bytes.

From my understanding, ([quic-draft-32 #14](https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport-34#section-14)) the maximum UDP datagram payload should be set at `1200` bytes (unless afterwards increased by  PMTUD).

Is there any requirement for this design choice?

djc:
Hmm, I think you're right. I find the wording in the spec a little confusing, since it discusses the maximum datagram size:

> The maximum datagram size is defined as the largest size of UDP payload that can be sent across a network path using a single UDP datagram. QUIC MUST NOT be used if the network path cannot support a maximum datagram size of at least 1200 bytes.

But then also goes on to discuss the minimum IP packet size:

> QUIC assumes a minimum IP packet size of at least 1280 bytes. This is the IPv6 minimum size [IPv6] and is also supported by most modern IPv4 networks. Assuming the minimum IP header size of 40 bytes for IPv6 and 20 bytes for IPv4 and a UDP header size of 8 bytes, this results in a maximum datagram size of 1232 bytes for IPv6 and 1252 bytes for IPv4. Thus, modern IPv4 and all IPv6 network paths are expected to be able to support QUIC.

And goes on to state:

> Any maximum datagram size larger than 1200 bytes can be discovered using Path Maximum Transmission Unit Discovery (PMTUD) (see Section 14.2.1) or Datagram Packetization Layer PMTU Discovery (DPLPMTUD) (see Section 14.3).

So we have three different things:

* IP packet size: minimum of 1280 for IPv6 and "most modern IPv4 networks"
[...]

Issue #14, opened 2018-05-22
Change `Codec::decode()` to return `QuicResult<T>`


djc:
Decodes can obviously fail on invalid input. Currently some of them don't handle this at all; others sometimes panic. The explicit panics should definitely be converted to returning `QuicError`; there should also be some minimal checking that the input is sane (for example, length of input slice).

djc:
Fixed by Emeric in cbaf2b515464c7e5097187c76f37fceaae48aa8f.

PR #1156, opened 2021-07-06
Bounding UDP payload size to 1200 and renaming

BiagioFesta:
As mentioned in #1154, it seems quinn sends UDP datagrams with a payload 1232 bytes by default.

This is easily reproducible with the client.rs/server.rs quinn example, making a GET request and serve a reasonable big file.
```
cargo run --example client https:/server:1234/README.md
```
```
[...]
Jul 06 02:59:45.746 TRACE drive{id=0}: quinn_proto::connection: sending 1232 bytes in 1 datagrams
[...]
```

---

Looking at the QUIC spec, ([quic-draft-32 #14](https://datatracker.ietf.org/doc/html/draft-ietf-quic-transport-34#section-14)) , it seems the protocol should supports networks where the maximum UDP payload is 1200 bytes.
Any larger value should be discovered using PMTUD.

This PR contains two changes:
* Renaming the [`const MIN_MTU: u16 = 1232;`](https://github.com/quinn-rs/quinn/blob/3f908a2c8c1ec4585212d776fafe536ea17bf2b4/quinn-proto/src/lib.rs#L313) into `MAX_INITIAL_UDP_PAYLOAD_SIZE`. This change has no impact from the semantic point of view. The rationale behind this change is because that value seems to be used across the code to compute what in QUIC is defined as "the maximum datagram size" (i.e., the UDP datagram payload).
* Changing this value from `1232` to `1200` and avoid quinn assembles packets (UDP datagrams) with a payload greater than this value.

### DIFF
--- a/quinn-proto/src/lib.rs
+++ b/quinn-proto/src/lib.rs
@@ -310,7 +310,7 @@ const LOC_CID_COUNT: u64 = 8;
 const RESET_TOKEN_SIZE: usize = 16;
 const MAX_CID_SIZE: usize = 20;
 const MIN_INITIAL_SIZE: u16 = 1200;
-const INITIAL_MAX_UDP_PAYLOAD_SIZE: u16 = 1232;
+const INITIAL_MAX_UDP_PAYLOAD_SIZE: u16 = 1200;
 const TIMER_GRANULARITY: Duration = Duration::from_millis(1);
 /// Maximum number of streams that can be uniquely identified by a stream ID
 const MAX_STREAM_COUNT: u64 = 1 << 60;
--- a/quinn-proto/src/lib.rs
+++ b/quinn-proto/src/lib.rs
@@ -310,7 +310,7 @@ const LOC_CID_COUNT: u64 = 8;
 const RESET_TOKEN_SIZE: usize = 16;
 const MAX_CID_SIZE: usize = 20;
 const MIN_INITIAL_SIZE: u16 = 1200;
-const MIN_MTU: u16 = 1232;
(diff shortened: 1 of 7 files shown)


quinn-proto/src/lib.rs
```diff
@@ -310,7 +310,7 @@ const LOC_CID_COUNT: u64 = 8;
 const RESET_TOKEN_SIZE: usize = 16;
 const MAX_CID_SIZE: usize = 20;
 const MIN_INITIAL_SIZE: u16 = 1200;
-const MIN_MTU: u16 = 1232;
+const MAX_INITIAL_UDP_PAYLOAD_SIZE: u16 = 1200;
```
Ralith, line 313:
```suggestion
const INITIAL_MAX_UDP_PAYLOAD_SIZE: u16 = 1200;
```

Ralith (changes requested):
Thanks!

quinn-proto/src/lib.rs
BiagioFesta: Changed

djc:
Just need to fix up the formatting a bit, then I'm happy to merge this.

BiagioFesta:
> Just need to fix up the formatting a bit, then I'm happy to merge this.

I guess a `cargo fmt` should be enough even if I don't see a `rustfmt.toml`

djc:
Would you mind squashing the formatting fixes into the first commit? Should be easy with an interactive rebase.

djc (approved):

PR #1157, opened 2021-07-08
Drop outgoing packets on permission errors

Matthias247:
When packets failed to transmit - e.g. due to being blocked by iptables - the
quinn endpoint kept them around and tried to retransmit them later on. This
usually failed for the same reason, which resulted in an infinitely growing list
of packets to transmit.

To prevent this, this change drops all packets where sending fails for a
a non-fatal reason.

### DIFF
--- a/quinn/src/platform/unix.rs
+++ b/quinn/src/platform/unix.rs
@@ -212,21 +212,28 @@ fn send(io: &mio::net::UdpSocket, transmits: &[Transmit]) -> io::Result<usize> {
             &mut cmsgs[i],
         );
     }
+    let num_transmits = transmits.len().min(BATCH_SIZE);
+
     loop {
-        let n = unsafe {
-            libc::sendmmsg(
-                io.as_raw_fd(),
-                msgs.as_mut_ptr(),
-                transmits.len().min(BATCH_SIZE) as _,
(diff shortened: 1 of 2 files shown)


quinn/src/platform/unix.rs
```diff
@@ -223,10 +223,21 @@ fn send(io: &mio::net::UdpSocket, transmits: &[Transmit]) -> io::Result<usize> {
...
+                    // Retry the transmission
+                    continue;
+                }
+                io::ErrorKind::PermissionDenied => {
+                    // Transmissions can fail with permission errors for example
+                    // due to iptable rules. Those are not fatal errors, since the
+                    // configuration can be dynamically changed.
+                    // In this case we drop the outgoing packets and let higher
+                    // layers retransmit if required.
+                    return Ok(transmits.len().min(BATCH_SIZE) as _);
```
Ralith, line 237: nit: putting this in a temporary and reusing it above would be a bit less error-prone.

Ralith (approved):

quinn/src/platform/unix.rs
Matthias247: done

Ralith (approved):

PR #1163, opened 2021-07-16
send new path challenge after previous path challenge

SSebo:


Ralith:
Thanks for having a go at this! This change causes path validation to be sent unconditionally, which isn't quite right. Instead, we should send an unpadded path validation iff validation is required but padding would violate the anti-amplification limit, as recommended in https://datatracker.ietf.org/doc/html/rfc9000#section-8.2.1, and then follow up with a padded validation to verify MTU.

However, I think your immediate issue will already have been addressed by the relaxed anti-amplification check introduced in #1148.

PR #1164, opened 2021-07-16
Fix example in "Self Signed Certificates" section of book

themaxdavitt:
Was trying out some example code from the book and found a typo, nothing big though.


djc (approved):
Thanks!

PR #1168, opened 2021-07-29
Output json throughput statistics from perf client

WesleyRosenblum:
This change adds a `--json` option to the  perf_client, allowing for throughput statistics to be output in a style partially matching that of iperf3. An `--interval` option is also added, to allow specifying the interval (in seconds) at which the statistics are gathered. This will allow the data gathered by the perf client to be used in a variety of ways, including through tools built for iperf3, such as https://github.com/rkost/iperf3Vis.

Example output (truncated):

```
{
  "start": {
    "timestamp": {
      "timesecs": 1627418034
    }
  },
  "intervals": [
    {
      "streams": [
        {
          "id": 0,
          "start": 0.010186108,
          "end": 1.010887858,
          "seconds": 1.00070175,
          "bytes": 10000000,
          "bits_per_second": 79943899.36861807,
          "sender": false
        },
        {
          "id": 4,
          "start": 0.010186108,
          "end": 1.010887858,
          "seconds": 1.00070175,
          "bytes": 10000000,
          "bits_per_second": 79943899.36861807,
          "sender": false
        },
       ...
      ],
      "sum": {
        "bytes": 153470294,
        "start": 0.010186108,
        "end": 1.010887858,
        "seconds": 1.00070175,
        "bits_per_second": 1226901373.960823,
        "sender": false
      }
    },
    {
      "streams": [
        {
          "id": 60,
          "start": 1.01089626,
          "end": 2.012940925,
          "seconds": 1.002044665,
          "bytes": 6529706,
          "bits_per_second": 52131057.451416105,
          "sender": false
        },
        {
          "id": 64,
          "start": 1.01089626,
          "end": 2.012940925,
          "seconds": 1.002044665,
          "bytes": 10000000,
          "bits_per_second": 79836760.56994925,
          "sender": false
        },
        ...
      ],
      "sum": {
        "bytes": 145551156,
        "start": 1.01089626,
        "end": 2.012940925,
        "seconds": 1.002044665,
        "bits_per_second": 1162033279.2251332,
        "sender": false
      }
    }
  ]
}
```

Example visualization with iperf3Vis:
image

### DIFF
--- a/perf/src/bin/perf_client.rs
+++ b/perf/src/bin/perf_client.rs
@@ -13,6 +13,7 @@ use tracing::{debug, error, info};
 
 use perf::bind_socket;
 use perf::stats::{OpenStreamStats, Stats};
+#[cfg(feature = "json-output")]
 use std::path::PathBuf;
 
 /// Connects to a QUIC perf server and maintains a specified pattern of requests until interrupted
@@ -59,6 +60,7 @@ struct Opt {
     #[structopt(long)]
     conn_stats: bool,
     /// File path to output JSON statistics to. If the file is '-', stdout will be used
+    #[cfg(feature = "json-output")]
     #[structopt(long)]
     json: Option<PathBuf>,
 }
@@ -201,6 +203,7 @@ async fn run(opt: Opt) -> Result<()> {
 
     endpoint.wait_idle().await;
 
+    #[cfg(feature = "json-output")]
     if let Some(path) = opt.json {
         stats.print_json(path.as_path())?;
     }
--- a/perf/src/bin/perf_client.rs
+++ b/perf/src/bin/perf_client.rs
@@ -1,6 +1,6 @@
 use std::{
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
-    sync::{Arc, Mutex},
+    sync::Arc,
     time::{Duration, Instant},
 };
 
@@ -12,7 +12,7 @@ use tokio::sync::Semaphore;
 use tracing::{debug, error, info};
 
 use perf::bind_socket;
-use perf::stats::{RequestStats, Stats};
+use perf::stats::{OpenStreamStats, Stats};
 
 /// Connects to a QUIC perf server and maintains a specified pattern of requests until interrupted
 #[derive(StructOpt)]
@@ -42,6 +42,9 @@ struct Opt {
     /// The time to run in seconds
     #[structopt(long, default_value = "60")]
     duration: u64,
+    /// The interval in seconds at which stats are reported
+    #[structopt(long, default_value = "1")]
+    interval: u64,
     /// Send buffer size in bytes
     #[structopt(long, default_value = "2097152")]
     send_buffer_size: usize,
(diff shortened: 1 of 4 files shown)


perf/src/bin/perf_client.rs
```diff
@@ -54,6 +59,9 @@ struct Opt {
     /// Whether to print connection statistics
     #[structopt(long)]
     conn_stats: bool,
+    /// File path to output JSON statistics to. If the file is '-', stdout will be used
+    #[structopt(long)]
+    json: Option<String>,
```
djc, line 64: How do you feel about making this `PathBuf` rather than `String`?

perf/src/bin/perf_client.rs
```diff
@@ -223,7 +257,7 @@ async fn drain_stream(mut stream: quinn::RecvStream, stats: &mut RequestStats) -
 async fn drive_uni(
     connection: quinn::Connection,
     acceptor: UniAcceptor,
-    stats: Arc<Mutex<Stats>>,
+    stream_stats: Arc<Mutex<Vec<Arc<StreamStats>>>>,
```
djc, line 260: Now that we're passing it around wrapped, I'd prefer that we abstract over the wrapping in a named type. (So some type potentially still named `StreamStats` should contain `Arc<Mutex<_>>` in a field.)

perf/src/bin/perf_client.rs
```diff
@@ -368,142 +391,3 @@ impl rustls::ServerCertVerifier for SkipServerVerification {
         Ok(rustls::ServerCertVerified::assertion())
     }
 }
-
```
djc, line 371: Please keep moving this code into a separate file/module as a separate commit.

djc (commented):
Thanks for working on this! It looks like you had many commits and then squashed them all, while we typically like to at least keep some of the logically consistent commits to review separately. Especially moving the code into a separate module.

I'm not sure whether there are other logically separate changes that could make sense here. In particular, somewhat mechanical refactoring would be good to isolate in separate commits.

Also, please clean up the clippy issues.

WesleyRosenblum:
Thanks! I've restored the original commits and have made the suggested changes.

djc:
Okay, sorry, you were probably right to squash these. I think I'd like to see these commits:

* Move stats types into a separate module
* Introduce new stats types (including `OpenStreamStats`); ideally this would abstract more over its internals then what you currently have (for example, can `stats.on_interval()` just take an `&OpenStreamStats` that it locks internally)
* Add new JSON module + dependencies; here too I wonder if we could have `print_json()` take `Option<Path>` or some such, moving more of that logic into the `json` module

I hope that's not too hard to synthesize from what you currently have.

WesleyRosenblum:
I've cleaned up the commits and made the suggested changes

djc (dismissed):
LGTM, thanks!

djc:
(I'm still going to give @Ralith a change to look this over before we merge this, though.)

WesleyRosenblum:
Hi @Ralith, can this be merged?

Ralith:
Looks reasonable overall. Would it be difficult to make json output support, and the rather heavy serde deps it entails, optional?

WesleyRosenblum:
I've replaced the merge commit with a rebase and made json-output an optional feature

Ralith (approved):
Thanks!

djc:
Thanks for all the work on this!

PR #1170, opened 2021-07-30
Encode token lifetime configuration as a Duration

Ralith:
This is less error-prone and consistent with other configuration values.


djc:
The lint failure is actually related to your changes. Otherwise LGTM!

djc (approved):

PR #1172, opened 2021-08-06
Log and ignore transmit errors

Matthias247:
Some more transient transmit errors have been observed. E.g. for some
peers transmission failed with
```
error: Os { code: 101, kind: Other, message: "Network is unreachable" }
```

That lead the endpoint to shut down and not process any messages for any
other clients.

To prevent this, this changes the behavior to ignore errors inside the endpoint.
In order to not lose all visibility, the change however adds low frequency
logging for the errors.

### DIFF
--- a/quinn/src/platform/unix.rs
+++ b/quinn/src/platform/unix.rs
@@ -6,6 +6,7 @@ use std::{
     os::unix::io::AsRawFd,
     ptr,
     task::{Context, Poll},
+    time::Instant,
 };
 
 use futures::ready;
@@ -13,7 +14,7 @@ use lazy_static::lazy_static;
 use proto::{EcnCodepoint, Transmit};
 use tokio::io::unix::AsyncFd;
 
-use super::{cmsg, RecvMeta, UdpCapabilities};
+use super::{cmsg, log_sendmsg_error, RecvMeta, UdpCapabilities, IO_ERROR_LOG_INTERVAL};
 
 #[cfg(target_os = "freebsd")]
 type IpTosTy = libc::c_uchar;
@@ -27,6 +28,7 @@ type IpTosTy = libc::c_int;
 #[derive(Debug)]
 pub struct UdpSocket {
     io: AsyncFd<mio::net::UdpSocket>,
+    last_send_error: Instant,
 }
 
 impl UdpSocket {
@@ -34,19 +36,22 @@ impl UdpSocket {
         socket.set_nonblocking(true)?;
         let io = mio::net::UdpSocket::from_std(socket);
         init(&io)?;
+        let now = Instant::now();
         Ok(UdpSocket {
             io: AsyncFd::new(io)?,
+            last_send_error: now.checked_sub(2 * IO_ERROR_LOG_INTERVAL).unwrap_or(now),
         })
     }
 
     pub fn poll_send(
-        &self,
+        &mut self,
         cx: &mut Context,
         transmits: &[Transmit],
     ) -> Poll<Result<usize, io::Error>> {
         loop {
+            let last_send_error = &mut self.last_send_error;
             let mut guard = ready!(self.io.poll_write_ready(cx))?;
-            if let Ok(res) = guard.try_io(|io| send(io.get_ref(), transmits)) {
+            if let Ok(res) = guard.try_io(|io| send(io.get_ref(), last_send_error, transmits)) {
                 return Poll::Ready(res);
             }
         }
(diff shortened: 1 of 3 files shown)


quinn/src/platform/unix.rs
```diff
@@ -224,23 +234,38 @@ fn send(io: &mio::net::UdpSocket, transmits: &[Transmit]) -> io::Result<usize> {
...
+                    // Other errors are ignored, since they will ususally be handled
+                    // by higher level retransmits and timeouts.
+                    // - PermissionDenied errors have been observed due to iptable rules.
+                    //   Those are not fatal errors, since the
+                    //   configuration can be dynamically changed.
+                    // - Destination unreachable errors have been observed for other
+                    let now = Instant::now();
+                    if now.saturating_duration_since(*last_send_error) > IO_ERROR_LOG_INTERVAL {
+                        *last_send_error = now;
+                        warn!("sendmmsg error: {:?}, transmits: {:?}", e, transmits);
```
Ralith, line 248: Do we really want to log the *contents* of the packets being sent? That seems very verbose and uninteresting.

quinn/src/platform/unix.rs
```diff
@@ -27,26 +29,30 @@ type IpTosTy = libc::c_int;
...
+        let now = Instant::now();
         Ok(UdpSocket {
             io: AsyncFd::new(io)?,
+            last_send_error: now.checked_sub(2 * IO_ERROR_LOG_INTERVAL).unwrap_or(now),
         })
     }
 
     pub fn poll_send(
-        &self,
+        &mut self,
```
Ralith, line 48: We should keep the type signature of the fallback module in sync with this.

Ralith (commented):
General approach looks sound. I'm concerned that someone might be very confused by the subsampled logging, but then logging isn't really a well-defined interface anyway so it's not a huge issue. Still, might be good to document somewhere.

This will also avoid surprising misbehavior when connecting to an unsupported address family, a common issue on OSs which don't default to dual-stack sockets.

quinn/src/platform/unix.rs
Matthias247: yes, the contents are too verbose. However I thought it was interesting to see the src and destination address as well as if e.g. GSO is used. Could either manually log those fields, or add a `Debug` impl for `Transmit` which does not log the contents.

Matthias247:
> Still, might be good to document somewhere.

Where should that be? I actually think most of the error/warning logs need to be sampled, because otherwise the amount of logging that can be caused by peers is immense (servers can be easily DOSd by forcing them to log).

quinn/src/platform/unix.rs
Matthias247: Oops, I initially read this as also implement the logic in the fallback module. Well, done both now.

quinn/src/platform/unix.rs
Ralith: My knee-jerk thought is to favor manually logging, probably pulled out into a helper function.

Ralith:
> I actually think most of the error/warning logs need to be sampled, because otherwise the amount of logging that can be caused by peers is immense (servers can be easily DOSd by forcing them to log).

Hmm, that's worrying. Should we push for tracing to provide subsampled events out of the box, and adopt them pervasively? Seems like addressing it in this specific case alone is of comparatively limited value, particularly as this condition should be pretty difficult for a peer to induce.

quinn/src/platform/unix.rs
Matthias247: It would need at least 2 helper functions to accommodate for `Transmit` vs `[Transmit]` (mmsg case). Not sure whether that's favorable.

djc:
I'm pretty sure tracing subscribers usually have sampling functionality built-in? I know the opentelemetry ecosystem does, at least.

Matthias247:
> I'm pretty sure tracing subscribers usually have sampling functionality built-in? I know the opentelemetry ecosystem does, at least.

Well, that would only help you if you have such a subscriber. The only application that I've worked with so far used tracing purely for debug logs, which dumps everything in log files or console. And not even sure how the tracing usage of quinn and other libraries would be a good fit for a structured logging system, since it doesn't emit metrics/telemetry but random text. On the projects I've worked so far we've usually used different solutions for metrics - but I'm also not experienced enough with `tracing` to see how to make it work.

Back to the problem: Yes, `tracing` could either directly offer such an API, but it might have the same challenge you have when e.g. implementing a `sampled_warn!`  macro: Where to store the last timestamp? It can't be part of the `span`, because that doesn't know all the events emitted in it. It could be a thread-local, which means you potentially get more samples than anticipated on a multithreaded runtime. Or a global, which synchronization for logging - and would suppress messages even if they come from different runtime/connection instances.

Probably both not the end of the world, and quinn including and using those kind of macros could make sense.

quinn/src/platform/unix.rs
Matthias247: Added the helper function, after I recognized we only need to log the first `Transmit`s content (which will always be the "offending" one). Also moved the debounce check there, which makes things a bit cleaner

djc (approved):

Ralith:
If our logging is dangerously verbose in general, and general tools are available to solve the problem, does it make sense to go out of our way to sample this one specific message?

Matthias247:
> If our logging is dangerously verbose in general, and general tools are available to solve the problem, does it make sense to go out of our way to sample this one specific message?

My take on this is that everything that is in the info/warn/error category,  and can be triggered by sending a single picket, and especially can happen before a connection is established has the potential to flood logs and thereby to slow down the local end of the connection. That should be avoided. But I don't think there are too many of those logs.

Regarding the "general tools" - I'm not familiar with that enough. People (including me) just set a log level on tracing and assume they are good :)

Ralith:
I suppose we can revisit this logic if needed in an audit pass to address other instances.

PR #1175, opened 2021-08-10
rather than importing futures only import the crates we really need. …

xMAC94x:
…This way we can streamline our dependencies, e.g. futures-executor is not used at all

Additionally i updated the dependencies to futures 0.3.11, as 0.3.8 is yanked (same for 0.3.9 and 0.3.10)

### DIFF
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -1,14 +1,14 @@
-use bytes::Bytes;
-use futures_channel::oneshot;
-use futures_core::ready;
-use futures_util::{io::AsyncWrite, FutureExt};
-use proto::{ConnectionError, FinishError, StreamId, Written};
 use std::{
     future::Future,
     io,
     pin::Pin,
     task::{Context, Poll},
 };
+
+use bytes::Bytes;
+use futures_channel::oneshot;
+use futures_util::{io::AsyncWrite, ready, FutureExt};
+use proto::{ConnectionError, FinishError, StreamId, Written};
 use thiserror::Error;
 
 use crate::{connection::ConnectionRef, recv_stream::UnknownStream, VarInt};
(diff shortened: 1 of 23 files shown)


djc (approved):

djc:
Thanks, this looks good!

djc:
Looks like CI is not quite happy. Maybe it would also make sense to import items from core via until, so we need to maintain one less direct import?

xMAC94x:
fixed tests, benches

Ralith (dismissed):
Thanks! Eventually I'd like to drop these entirely as they're kind of heavy; this is a good first step.

quinn/src/send_stream.rs
```diff
@@ -1,13 +1,13 @@
...
+use futures_channel::oneshot;
+use futures_util::{io::AsyncWrite, ready, FutureExt};
+use proto::{ConnectionError, FinishError, StreamId, Written};
 use std::{
     future::Future,
     io,
     pin::Pin,
     task::{Context, Poll},
 };
-
```
Ralith, line 7: This empty line shouldn't have been removed.

Ralith (dismissed):

quinn/src/send_stream.rs
xMAC94x: what use blocks should remain seperate ?
can you modify it exactly how you want it ? 
```
use bytes
use futures
use proto
use std {
 ..
 ..
}
< One line space
use thiserror
< one line space
use crate::...
< one line space
```
what is the reason for the space before thiserror?


```

Ralith: Imports from `std` should be first, followed by an empty line, followed by external crate imports (if any). You've moved a bunch of non-std imports into the std block.

xMAC94x: @Ralith  whoops sorry, i prob deleted a newline and then run cargo fmt.
Fixed it :)

djc (approved):

Issue #1176, opened 2021-08-13
Suggestion: infallible `TransportConfig` APIs


connec:
I'm currently using `quinn` for the transport layer in another library. This has generally been pretty smooth, and is definitely helped overall by the granularity of error types in `quinn` 👍 

However, I'm finding the `TransportConfig` API particularly awkward to work with, since many methods can fail with `ConfigError::OutOfBounds` (e.g. [`quinn::TransportConfig::max_concurrent_bidi_streams`](https://docs.rs/quinn/0.7.2/quinn/struct.TransportConfig.html#method.max_concurrent_bidi_streams)). Since `ConfigError` non-exhaustive I can't easily wrap it without some kind of catch-all variant, or exposing `quinn::ConfigError` from my library.

Since all the fallibility comes from converting to `VarInt`s, would you be open to adding relevant 'lower level' methods to `TransportConfig` that would accept a `VarInt` and be infallible, so that consumers can do their own validation/sanitisation?

Some other options I considered, but I don't think really solve the issue:

- Make `quinn::ConfigError` exhaustive. That makes it easier to hide `quinn::ConfigError`, but doesn't really help with infallible use (e.g. if I want to simply clamp out-of-bounds values, I'll still end up `unwrap`ping a `ConfigError`, and/or dealing with other variants (with `unreachable` etc.)).
- Use a specific `OutOfBounds` error, rather than the `ConfigError` enum. That has the same issue as above, though I'd be more comfortable `unwrap`ping it.

Ralith:
Thanks for the issue! I'm fine with these methods being modified to consume `VarInt` directly, for the reasons you outline. `0u32.into()` isn't too bad for people who are hardcoding these values, so I'm not inclined to bother with multiple variants, though maybe `impl Into<VarInt>` wouldn't be too bad.

connec:
That makes sense, I had also thought about changing the signature to `impl TryInto<VarInt>`, returning the associated error – that way you can do an empty match on `Infallible` (and hopefully one day [`Result::into_ok`](https://doc.rust-lang.org/stable/std/result/enum.Result.html#method.into_ok)). That could be a bit much type gymnastics though.

Also, what do you think about [`max_idle_timeout`](https://docs.rs/quinn/0.7.2/quinn/struct.TransportConfig.html#method.max_idle_timeout)? It's a shame to lose `Duration`, and it's not obvious from the current name that a raw value would be in milliseconds.

Ralith:
[...]

PR #1177, opened 2021-08-16
Infallible `TransportConfig` setters

connec:
Based on the discussion in #1176, I've updated the relevant `TransportConfig` setters to take `VarInt`, so they become infallible. In most cases, this is just a case of applying `.into()`/`.try_into()` to the argument (particularly with the addition of `TryFrom<Duration> for VarInt`).

In that issue one thing suggested was using `impl Into<VarInt>`, in the end I left that out since it doesn't do much for literal arguments, which still need to be ascribed with a relevant type (e.g. `1u8`).

---

- 0f840f01 **Use `VarInt` in `TransportConfig`'s `VarInt` field setters**

  Rather than taking a `u64` and returning a `ConfigError::OutOfBounds`,
  these methods are now infallible by taking a `VarInt` directly.
  
  In many cases, call-sites can be updated simply by adding a `.into()` or
  `.try_into()` to the argument.
  
  BREAKING CHANGE: The `max_concurrent_bidi_streams`,
  `max_concurrent_uni_streams`, `stream_receive_window`, and
  `receive_window` methods on `TransportConfig` now take `VarInt`
  arguments and return `&mut Self`.

- 5713ddaf **Use `VarInt` in `TransportConfig` `max_idle_timeout` field**

  Rather than using `Duration` internally, max idle timoout is now stored
  as a `VarInt`. This avoids some conversions and an internal `unwrap`.

- edf1e786 **Implement `TryFrom<Duration>` for `VarInt`**

  This tries to encode a `Duration`, as milliseconds, into a `VarInt`.

- d707827b **Use `VarInt` in `TransportConfig` `max_idle_timeout` setter**

  This allows the setter to be infallible. The impl of `TryFrom<Duration>`
  for `VarInt` means call-sites can just use `.try_into()`.
  
  BREAKING CHANGE: The `max_idle_timeout` method on `TransportConfig` now
  takes a `VarInt` argument and returns `&mut Self`.

### DIFF
--- a/bench/src/bulk.rs
+++ b/bench/src/bulk.rs
@@ -1,4 +1,5 @@
 use std::{
+    convert::TryInto,
     net::{IpAddr, Ipv6Addr, SocketAddr},
     str::FromStr,
     sync::Arc,
@@ -271,9 +272,7 @@ fn transport_config(opt: &Opt) -> quinn::TransportConfig {
     // High stream windows are chosen because the amount of concurrent streams
     // is configurable as a parameter.
     let mut config = quinn::TransportConfig::default();
-    config
-        .max_concurrent_uni_streams(opt.max_streams as u64)
-        .unwrap();
+    config.max_concurrent_uni_streams(opt.max_streams.try_into().unwrap());
     config
 }
 
--- a/interop/src/main.rs
+++ b/interop/src/main.rs
@@ -2,6 +2,7 @@
 #![type_length_limit = "2121396"]
 
 use std::{
+    convert::TryInto,
     env,
     net::{SocketAddr, ToSocketAddrs},
     sync::{Arc, Mutex},
@@ -276,10 +277,8 @@ impl State {
         }
         let mut transport = quinn::TransportConfig::default();
         transport.send_window(1024 * 1024 * 2);
-        transport.receive_window(1024 * 1024 * 2).unwrap();
-        transport
-            .max_idle_timeout(Some(Duration::from_secs(1)))
-            .unwrap();
+        transport.receive_window((1024_u32 * 1024 * 2).into());
+        transport.max_idle_timeout(Some(Duration::from_secs(1).try_into().unwrap()));
         let client_config = quinn::ClientConfig {
             crypto: Arc::new(tls_config),
             transport: Arc::new(transport),
--- a/interop/src/server.rs
+++ b/interop/src/server.rs
@@ -92,7 +92,7 @@ async fn main() -> Result<()> {
 async fn server(server_config: quinn::ServerConfigBuilder, addr: SocketAddr) -> Result<()> {
     let mut transport = quinn::TransportConfig::default();
     transport.send_window(1024 * 1024 * 3);
-    transport.receive_window(1024 * 1024).unwrap();
+    transport.receive_window((1024_u32 * 1024).into());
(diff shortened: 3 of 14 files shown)


quinn-proto/src/varint.rs
```diff
@@ -120,6 +120,14 @@ impl std::convert::TryFrom<usize> for VarInt {
     }
 }
 
+impl std::convert::TryFrom<Duration> for VarInt {
+    type Error = VarIntBoundsExceeded;
+    /// Succeeds iff `x.as_millis()` < 2^62
+    fn try_from(x: Duration) -> Result<Self, Self::Error> {
+        VarInt::try_from(x.as_millis())
```
djc, line 127: Not sure about how specific this is -- if we ever want a config value that's also naturally expressed in `Duration` but requires μs or ns on the wire, what do we do?

djc (commented):
Mostly looks good to me, though I have one concern...

quinn-proto/src/varint.rs
Ralith: Yeah, this gives me pause as well. It also introduces the odd possibility of passing a Duration to fields that aren't times at all. I'd prefer to go with the newtype approach leave the setter fallible. @Matthias247 raised the reasonable concern of a newtype making the API less obvious, but I'm not convinced that's a serious detriment as it's not a heavily-trafficked bit of API.

djc: Yup, newtype sounds good to me.

connec: Thank you for the review. I'll update this to use a newtype, though might be a day or two before I get to it!

connec:
I've rebased on `origin/main`, and pushed a [new commit](https://github.com/quinn-rs/quinn/pull/1177/commits/476e47b3ddcf3ebf3992736147ff1908dcac58be) replacing the `impl TryFrom<Duration> for VarInt` with an `IdleTimeout` newtype.

I've left the `max_idle_timeout` setter without any built-in conversion. This allows inference to work better for both `into` and `try_into` (e.g. if it was `fn max_idle_timeout(&mut self, value: Option<impl Into<IdleTimeout>>)`, then `config.max_idle_timeout(Some(Duration::from_secs(123).try_into()?))` would not work).

connec:
Copying out an example from a doc-comment, for easy reference:

```rust
use std::{convert::TryInto, time::Duration};
use quinn_proto::{TransportConfig, VarInt, VarIntBoundsExceeded};

fn main() -> Result<(), VarIntBoundsExceeded> {
    let mut config = TransportConfig::default();

    // Set the idle timeout as `VarInt`-encoded milliseconds
    config.max_idle_timeout(Some(VarInt::from_u32(10_000).into()));

    // Set the idle timeout as a `Duration`
    config.max_idle_timeout(Some(Duration::from_secs(10).try_into()?));

    Ok(())
}
```

djc (approved):

djc:
This looks good to me. @Ralith do you still want to take a look?

Ralith (approved):
LGTM, thanks! Squashing the intermediates would be great; otherwise I'll hit the button in a day or two.

connec:
I've squashed all the commits into one, based on the overall change of making `TransportConfig` setters infallible.

Ralith (approved):
Thanks!

PR #1179, opened 2021-08-19
Fix cubic minimum congestion window handling

Matthias247:
`config.minimum_window` is already in bytes - it doesn't have to be multiplied by them..


Ralith (approved):

Ralith:
Great catch!

djc:
Well done 👍

Issue #1178, opened 2021-08-16
extract udp socket handling into a separate crate


dvc94ch:
I kind of did that for libp2p-quic [here](https://github.com/ipfs-rust/udp-socket) but it might make sense to unify the code bases, as it's probably useful to many people using udp sockets.

Ralith:
It's an interesting idea. There's some possible drawbacks:
- In the long run, people with different use cases might end up each using only a small subset of the capabilities of such a library, which makes it less appealing.
- It could make porting to new platforms much more difficult for a given user, if such a port is expected to cover every feature, not just that user's requirements (see e.g. winit's maintenance challenges)

kpp:
I think such a crate should not be general but specialized for use whithin quinn(-proto). 

> make porting to new platforms much more difficult for a given user

Extracting or not, such an issue remains. So far it is handled as a `fallback` module. We can do the same.

dvc94ch:
> In the long run, people with different use cases might end up each using only a small subset of the capabilities of such a library, which makes it less appealing.

I assume the concern here is mainly with the maintainance overhead if people start requesting lots of features quinn doesn't use. But we can say no to such feature requests.

Ralith:
If it's explicitly quinn- (or at least QUIC-) specific, and only split out for the benefit of people working with quinn-proto directly for whatever reason, then I don't see any major issues.

djc:
Sounds okay to me, too.

PR #1180, opened 2021-08-19
Extract udp crate

kpp:
This one is easy - moved code from quinn/platform into quinn-udp subcrate.
I will try to apply the code from https://github.com/ipfs-rust/udp-socket (which is runtime-agnostic due to use of `async-io` instead of `tokio`) in the [next PR](https://github.com/quinn-rs/quinn/pull/1183).

Fix https://github.com/quinn-rs/quinn/issues/1178


quinn-udp/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "quinn-udp"
+version = "0.7.0"
+authors = ["Benjamin Saunders <ben.e.saunders@gmail.com>", "Dirkjan Ochtman <dirkjan@ochtman.nl>"]
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/djc/quinn"
```
djc, line 6: Should use the quinn-rs URL here.

quinn-udp/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+all-features = true
+
+[badges]
+maintenance = { status = "experimental" }
+
+[dependencies]
+futures-util = { version = "0.3.11", features = ["io"] }
+libc = "0.2.69"
+mio = { version = "0.7.7", features = ["net"] }
+proto = { package = "quinn-proto", path = "../quinn-proto", version = "0.7" }
```
djc, line 23: Feels weird for this crate to depend on proto. Why is that? (Hard to see since there are no diffs for the moved files.)

djc (commented):
Looks good, some comments.

quinn-udp/Cargo.toml
kpp: That's because of:

> use proto::{EcnCodepoint, Transmit};

quinn-udp/Cargo.toml
kpp: Done. Also fixed repo url in quinn and quinn-proto.

quinn-udp/Cargo.toml
Ralith: Per discussion in #1178 we're expressly intending for this crate to be used only with quinn-proto, so this seems reasonable to me.

Ralith (dismissed):
Thanks, this generally LGTM. Sorry for the long review turnaround!

I appreciate the logically isolated commits; they make it a lot easier to scan things. Would you mind squashing the first two together so that there's not an unbuildable state between them?

kpp:
Done. I also rebased onto main.

djc (approved):

djc:
Thanks!

PR #1181, opened 2021-08-19
Remove unused `ConnectError::Config` variant

connec:
There has been no implicit conversion to this variant since the switch
to `thiserror` (1bca0f7c), since `#[source]` does not imply `#[from]`.
There has never been any explicit construction of this variant, so it
likely became unused some time before during various configuration
refactors.

BREAKING CHANGE: The `ConnectError::Config` variant is removed.


djc (approved):
Nice catch!

Ralith:
Thanks!

PR #1182, opened 2021-08-19
Return owned memory via BytesMut to reduce downstream copying

tbraun96:
Compiles and runs. Tests return Ok(())

### DIFF
--- a/quinn-proto/src/frame.rs
+++ b/quinn-proto/src/frame.rs
@@ -517,7 +517,8 @@ impl Crypto {
 }
 
 pub struct Iter {
-    bytes: Cursor<BytesMut>,
+    // TODO: ditch io::Cursor after bytes 0.5
+    bytes: io::Cursor<BytesMut>,
     last_ty: Option<Type>,
 }
 
@@ -850,10 +851,10 @@ pub enum Datagram {
 }
 
 impl Datagram {
-    pub(crate) fn assert_incoming(self) -> BytesMut {
+    pub(crate) fn assert_incoming(self) -> Option<BytesMut> {
         match self {
-            Self::Incoming(ret) => ret,
-            _ => panic!("Asserted datagram was incoming type, but was actually outgoing type"),
+            Self::Incoming(ret) => Some(ret),
+            _ => None,
         }
     }
 }
--- a/quinn-proto/src/frame.rs
+++ b/quinn-proto/src/frame.rs
@@ -17,6 +17,8 @@ use crate::{
 
 #[cfg(feature = "arbitrary")]
 use arbitrary::Arbitrary;
+use std::ops::Deref;
+use std::io::Cursor;
 
 #[derive(Copy, Clone, Eq, PartialEq)]
 pub struct Type(u64);
@@ -515,8 +517,7 @@ impl Crypto {
 }
 
 pub struct Iter {
-    // TODO: ditch io::Cursor after bytes 0.5
-    bytes: io::Cursor<Bytes>,
+    bytes: Cursor<BytesMut>,
     last_ty: Option<Type>,
 }
 
@@ -544,7 +545,7 @@ impl From<UnexpectedEnd> for IterErr {
 }
 
 impl Iter {
-    pub fn new(payload: Bytes) -> Self {
+    pub fn new(payload: BytesMut) -> Self {
         Iter {
             bytes: io::Cursor::new(payload),
             last_ty: None,
@@ -633,7 +634,7 @@ impl Iter {
                 Frame::Ack(Ack {
                     delay,
                     largest,
-                    additional: self.bytes.get_ref().slice(start..end),
+                    additional: Bytes::copy_from_slice(&self.bytes.get_ref()[start..end] as &[u8]),
                     ecn: if ty != Type::ACK_ECN {
                         None
                     } else {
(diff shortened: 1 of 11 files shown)


djc:
What is the difference between the  two commits you have here?

tbraun96:
> What is the difference between the two commits you have here?

There was a warning of an unused import That I cleared. Also, just made a new commit. ``Cargo test`` checks out fine on my machine

tbraun96:
Might I also add that in order for this to work, publicly-exposed functions like ``send_datagram`` now accept a ``BytesMut``, so this would naturally be a breaking change.

Ralith:
Thanks for the PR!

> publicly-exposed functions like send_datagram now accept a BytesMut

Interfaces that consume data should always take `Bytes`, because `BytesMut` can be efficiently converted to `Bytes` with `BytesMut::freeze` but the reverse is not true.

tbraun96:
Okay, I adjusted it. Public functions are now normal as desired. Had to make ``Datagram`` an enum to differentiate between outgoing and incoming to ensure zerocopying, otherwise, there were parts in the code where there would be unnecessary copying.

quinn-proto/src/frame.rs
```diff
@@ -515,8 +517,7 @@ impl Crypto {
 }
 
 pub struct Iter {
-    // TODO: ditch io::Cursor after bytes 0.5
```
djc, line 518: You removed this TODO without resolving it -- wondering why?

quinn-proto/src/frame.rs
```diff
@@ -841,9 +842,31 @@ pub const RETIRE_CONNECTION_ID_SIZE_BOUND: usize = 9;
...
-    pub data: Bytes,
+pub enum Datagram {
+    /// Used to designate an outgoing immutable buffer
+    Outgoing(Bytes),
+    /// Used to designate an incoming mutable buffer
+    Incoming(BytesMut),
+}
+
+impl Datagram {
+    pub(crate) fn assert_incoming(self) -> BytesMut {
```
djc, line 853: Let's not panic here. Instead, call the function `incoming()` and have it return `Option<BytesMut>`.

djc (commented):
Here's some things that stood out to me -- not quite a detailed review yet.

quinn-proto/src/frame.rs
tbraun96: Oh, I remember now. I originally tried resolving that todo by using ``BytesMut``, but ran into some issues with the tests failing. So, I reverted it back to using a cursor while forgetting to put the TODO back. I'll update this in the next push

quinn-proto/src/frame.rs
tbraun96: Okay

quinn-proto/src/frame.rs
tbraun96: Okay, it's been reverted. Originally, I tried solving the problem, but the tests failed so I reverted everything except that TODO. How might the new TODO be solved? I could include that in this PR if you want.

quinn-proto/src/frame.rs
```diff
@@ -544,21 +546,23 @@ impl From<UnexpectedEnd> for IterErr {
...
-    fn take_len(&mut self) -> Result<Bytes, UnexpectedEnd> {
+    fn take_len(&mut self) -> Result<BytesMut, UnexpectedEnd> {
         let len = self.bytes.get_var()?;
         if len > self.bytes.remaining() as u64 {
             return Err(UnexpectedEnd);
         }
         let start = self.bytes.position() as usize;
         self.bytes.advance(len as usize);
-        Ok(self.bytes.get_ref().slice(start..(start + len as usize)))
+        Ok(BytesMut::from(
```
Matthias247, line 563: I think this now creates a deep copy of all incoming data, where there hasn't been one before.

Since includes datagram frames, it should effectively negate the whole purpose of this PR?

My guess is that it didn't show up more negative in the benchmark since that one just uses a single stream and has always sufficient data to send, to the application runs into the `take_remaining` case which stays cheap.  But if people try to send smaller datagrams or use small streams it would it get more expensive.

quinn-proto/src/frame.rs
```diff
@@ -633,7 +637,7 @@ impl Iter {
                 Frame::Ack(Ack {
                     delay,
                     largest,
-                    additional: self.bytes.get_ref().slice(start..end),
+                    additional: Bytes::copy_from_slice(&self.bytes.get_ref()[start..end] as &[u8]),
```
Matthias247, line 640: looks like another new deep copy?

I don't think it should do this. If it already has to allocate, it should at least immediately decode the ACKs blocks and have those stored in a list, so that this work goes away later on.

quinn-proto/src/frame.rs
tbraun96: > Since includes datagram frames, it should effectively negate the whole purpose of this PR?

True. I wasn't aware ``slice`` was a cheap operation; I somehow thought it was a copy.

Going through this all, it seems this PR won't work without significantly changing the buffering mechanism. Have any ideas?

djc: It seems like one thing we could do is to buffer `BytesMut`s throughout the (quinn-proto) stack. That way we can split them wherever necessary, I think, without copying. Would that make sense?

Ralith:
Closing as stale and development has moved on considerably, and outstanding feedback suggests a significantly different approach. Feel free to open a new PR if this is still of interest.

PR #1184, opened 2021-08-25
Chore: disable unused default features for various crates

Some-Dood:
This PR disables some unused default features enabled by various crates in the production dependencies of `quinn` and `quinn-proto`. While I was removing default features, I also found an unused import at `quinn/src/platform/fallback.rs`, which I removed accordingly. Overall, there should be no changes in behavior. Thanks!


djc:
Thanks for looking at this! What goal are you trying to achieve here? For webpki, this change doesn't look like it'll have a big impact.

We prefer to have commits that are logically distinct, would you mind squashing the formatting change as well as the in-PR flip-flopping on the default features for proto into a single commit?

Some-Dood:
> What goal are you trying to achieve here?

Hello! Thanks for the quick response. This PR was originally supposed to remove a chunk of the dependency tree (via `cargo tree`) caused by `webpki` and `rustls`.

In one of my projects, I wanted to remove all possible traces of built-in certificates so that I could pin the application to trusting only one specific self-signed certificate. Since the application only connects to one specific host internally, I figured that bindings to external certificates (via `native`, `webpki-roots`, or otherwise) is unnecessary bloat for build times and binary sizes. Moreover, pinning the certificate allows me to have greater control over the allowed connections of the application, which is great from a security standpoint.

However, I would eventually find out that it is not so easy to minimize the dependency tree of `quinn` and `quinn-rs` without breaking some code and refactoring module exports. Hence, I have settled to only removing default features so that I could at least lessen the dependency tree by a little bit.

> We prefer to have commits that are logically distinct, would you mind squashing the formatting change as well as the in-PR flip-flopping on the default features for proto into a single commit?

No problem! Though, I definitely need to review how to do commit squashing. It's been a while since I've last tinkered with Git history. I'll let you know when it's ready!

Some-Dood:
Hello again @djc! As requested, I have rebased the commits. The Git history is definitely much cleaner now. 👌

djc (approved):

djc:
Makes sense. While quinn enables support for native roots by default, this should be easily disabled by disabling its default features in your project.

PR #1187, opened 2021-09-05
Add RecvStream::read_to_end_into

geeklint:
This is similar to RecvStream::read_to_end, but doesn't do any allocation.  It would enable reading a full stream into stack arrays for small messages, or for instance re-use of buffers between calls.

### DIFF
--- a/quinn/src/recv_stream.rs
+++ b/quinn/src/recv_stream.rs
@@ -191,6 +191,25 @@ where
         }
     }
 
+    /// Convenience method to read all remaining data into a buffer
+    ///
+    /// The returned future fails with [`ReadToEndError::TooLong`] if it's longer than
+    /// `buffer.len()` bytes. Uses unordered reads to be more efficient than using `AsyncRead`
+    /// would allow.
+    ///
+    /// If unordered reads have already been made, the resulting buffer may have gaps containing
+    /// arbitrary data.
+    ///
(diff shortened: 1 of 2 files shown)


Ralith:
Thanks for the PR! I'm a bit concerned that this is a very small optimization, adding to our already unreasonably large variety of read methods. Users who really want to avoid copies already have `read_chunk`(s), which is one less copy than `read_to_end_into`. Have you measured any performance impact?

geeklint:
If you feel like the number of read methods is already unreasonably large, that's probably a fine reason not to include it.

My intention wasn't necessarily to avoid the performance impact of copying, I was more interested in avoiding the overhead of invoking the allocator the way read_to_end currently does, and/or the extra memory usage by the current read_to_end.

My benchmark attempts this morning don't necessarily support the former, as I wasn't able to measure a performance difference between the two for the small sizes I was interested in (<1k).

For larger buffers (4M), I did observe it to be slightly faster, and use half the memory of the existing read_to_end.

geeklint:
I would almost suggest I merge the memory saving effect into the existing read_to_end, except that this impl has the drawback of being quadratic in the face of a malicious peer (which is why I imagined it being mostly useful for small messages).

Since I have maybe convinced myself it's not useful for small messages anyway, maybe it's not useful at all.

PR #1188, opened 2021-09-05
Fix inner crate test

geeklint:
There's a test in the quinn crate that uses `futures_util::join!` which is behind a feature gate the quinn itself doesn't enable.  (Other crates in the workspace do enable that feature, so it only errors if you run cargo test from within quinn before running it for the workspace).

This adds a line to dev-dependancies which enables that feature when testing quinn.


Ralith (approved):
Thanks!

PR #1191, opened 2021-09-18
Endpoint - allowing override server configuration

BiagioFesta:
The idea here is to reload the server configuration allowing a new setup for incoming connections without dropping the endpoint and closing existing ones.

The main rationale is the fact that, in some circumstances, on long server sessions, it is needed to reload its TLS configuration. In particular, to have the possibility to renew an expiring TLS certificate keeping the server/service operative.

---

<details>
  <summary>Pseudo Code Usage Idea</summary>

```
let (endpoint, mut incoming) = endpoint.bind(&listen_addr).unwrap();

loop {
  select! {
      Some(incoming) = incoming.next() => {
          // Handle new incoming connection
       }
       new_config = config_reload.next() => {
           let new_server_config = build_new_server_config(&config);
           endpoint.override_server_config(new_server_config);
        }
    }
}

```  
</details>

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -109,6 +109,11 @@ where
         self.transmits.pop_front()
     }
 
+    /// Replace the server configuration, affecting new incoming connections only
+    pub fn set_server_config(&mut self, server_config: Option<Arc<ServerConfig<S>>>) {
+        self.server_config = server_config;
+    }
+
     /// Process `EndpointEvent`s emitted from related `Connection`s
     ///
     /// In turn, processing this event may return a `ConnectionEvent` for the same `Connection`.
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -17,7 +17,11 @@ use futures_channel::mpsc;
 use futures_util::StreamExt;
 use fxhash::FxHashMap;
 use once_cell::sync::OnceCell;
-use proto::{self as proto, generic::ClientConfig, ConnectError, ConnectionHandle, DatagramEvent};
+use proto::{
+    self as proto,
+    generic::{ClientConfig, ServerConfig},
+    ConnectError, ConnectionHandle, DatagramEvent,
+};
 
 use crate::{
     broadcast::{self, Broadcast},
@@ -114,6 +118,17 @@ where
         inner.socket = socket;
         inner.ipv6 = addr.is_ipv6();
         Ok(())
+    }
+
+    /// Replace the server configuration, affecting new incoming connections only
+    ///
+    /// Useful for e.g. refreshing TLS certificates without disrupting existing connections.
+    pub fn set_server_config(&self, server_config: Option<ServerConfig<S>>) {
+        self.inner
+            .lock()
+            .unwrap()
+            .inner
+            .set_server_config(server_config.map(Arc::new))
     }
 
     /// Get the local `SocketAddr` the underlying socket is bound to


quinn-proto/src/endpoint.rs
```diff
@@ -109,6 +109,13 @@ where
         self.transmits.pop_front()
     }
 
+    /// Override the server configuration if any, otherwise does nothing
+    pub fn override_server_config(&mut self, server_config: Arc<ServerConfig<S>>) {
+        if self.is_server() {
+            self.server_config = Some(server_config);
+        }
+    }
```
Ralith, line 117:
```suggestion
    /// Replace the server configuration, affecting new incoming connections only
    pub fn set_server_config(&mut self, server_config: Option<Arc<ServerConfig<S>>>) {
        self.server_config = server_config;
    }
```

quinn/src/endpoint.rs
```diff
@@ -116,6 +120,19 @@ where
         Ok(())
     }
 
+    /// Set the server configuration overriding the existing one.
+    /// If this endpoint has been built without a server configuration, it does nothing.
+    ///
+    /// It allows to adapt some environment changes when this endpoint is acting as a server.
+    /// For instance, changing the TLS configuration without dropping the endpoint.
```
Ralith, line 127:
```suggestion
    /// Useful for e.g. refreshing TLS certificates without disrupting existing connections.
```

quinn/src/endpoint.rs
```diff
@@ -116,6 +120,19 @@ where
...
+    pub fn override_server_config(&self, server_config: ServerConfig<S>) {
```
Ralith, line 128: As above.
```suggestion
    pub fn set_server_config(&self, server_config: Option<ServerConfig<S>>) {
```

Ralith (changes requested):
Thanks for the PR! This will be great to have; I honestly didn't realize how easy it would be.

quinn-proto/src/endpoint.rs
BiagioFesta: Done

quinn/src/endpoint.rs
BiagioFesta: Done

quinn/src/endpoint.rs
BiagioFesta: Done

quinn/src/endpoint.rs
```diff
@@ -116,6 +120,18 @@ where
         Ok(())
     }
 
+    /// Set the server configuration overriding the existing one.
+    /// If this endpoint has been built without a server configuration, it does nothing.
```
Ralith, line 124:
```suggestion
    /// Replace the server configuration, affecting new incoming connections only
```

Ralith (changes requested):

Ralith (approved):
Thanks!

quinn/src/endpoint.rs
```diff
@@ -116,6 +120,17 @@ where
         Ok(())
     }
 
+    /// Replace the server configuration, affecting new incoming connections only
+    ///
+    /// Useful for e.g. refreshing TLS certificates without disrupting existing connections.
```
Matthias247, line 125: You should already be able to do this with the current config, since the cert is not necessarily static but a lookup callback. Same applies for other functions for which "providers" can be hooked in (like the CidGenerator).

I guess it won't hurt to swap the complete config instead of just adding more callbacks. But the question is whether swapping this config dynamically might break anything that doesn't expect the values to change over time.

quinn-proto/src/endpoint.rs
```diff
@@ -109,6 +109,11 @@ where
         self.transmits.pop_front()
     }
 
+    /// Replace the server configuration, affecting new incoming connections only
+    pub fn set_server_config(&mut self, server_config: Option<Arc<ServerConfig<S>>>) {
```
Matthias247, line 113: Why allow `Option`? For servers, this must always be `Some`.

quinn/src/endpoint.rs
Ralith: > You should already be able to do this with the current config

Good point. Setters are a lot more convenient than setting up a sidechannel, though, and it might be useful to adjust some of the other options on the fly non-disruptively as well.

> But the question is whether swapping this config dynamically might break anything

I've audited this and I'm satisfied that there aren't any issues with the current set of config options. We actually don't read the server config in very many places; the overwhelming bulk is in `Endpoint::handle_first_packet`. At worst, swapping out `token_key` might cause spurious failure for in-flight handshakes, but that's trivially recoverable and far less disruptive than restarting the endpoint outright. It's also nice for rotating `token_key` to be enabled in its own right.

What could break things would be changing the config for an existing connection, but existing connections hold onto the `Arc` of the config they were created with, so that's not a risk.

quinn-proto/src/endpoint.rs
Ralith: Reviewing the code, it looks to me like this will allow enabling/disabling server operation at will, which seems somewhere between harmless and useful, and less surprising than silently doing nothing. I guess we could omit the `Option` and let it be possible to enable server mode but not disable it, but why not be symmetrical?

PR #1192, opened 2021-09-19
Use WorkLimiter also for sending data

Matthias247:
This adds time-based yielding to the send loop in the same fashion it
had been previously added ot the receive loop.
In my performance testing this didn't show a noticeable difference - likely
because in the current benchmark the client is the bottleneck. But it should
make things more deterministic.

### DIFF
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -25,7 +25,7 @@ use crate::{
     connection::Connecting,
     platform::{RecvMeta, UdpSocket, BATCH_SIZE},
     work_limiter::WorkLimiter,
-    ConnectionEvent, EndpointEvent, VarInt, IO_LOOP_BOUND, RECV_TIME_BOUND,
+    ConnectionEvent, EndpointEvent, VarInt, IO_LOOP_BOUND, RECV_TIME_BOUND, SEND_TIME_BOUND,
 };
 
 /// A QUIC endpoint.
@@ -269,6 +269,7 @@ where
     driver_lost: bool,
     recv_limiter: WorkLimiter,
     recv_buf: Box<[u8]>,
+    send_limiter: WorkLimiter,
     idle: Broadcast,
 }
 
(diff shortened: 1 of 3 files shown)


Matthias247:
Seems like this uncovered a bug where the sometimes a limit of 0 was calculated if the cycle time in measuring mode was extremely high. That blocked all IO until the next measuring mode. I added one commit to fix that by allowing a minimum of 1 work item.

quinn/src/work_limiter.rs
```diff
@@ -102,8 +102,10 @@ impl WorkLimiter {
                 (7.0 * self.smoothed_time_per_work_item_nanos + time_per_work_item_nanos) / 8.0
             };
 
-            self.allowed = ((self.desired_cycle_time.as_nanos()) as f64
-                / self.smoothed_time_per_work_item_nanos) as usize;
+            // Allow at least 1 work item in order to make progress
+            self.allowed = (((self.desired_cycle_time.as_nanos()) as f64
+                / self.smoothed_time_per_work_item_nanos) as usize)
```
Ralith, line 107: I wonder if there's also a divide-by-zero risk here, if `now` decides to behave funny (which it absolutely does sometimes).

quinn/src/endpoint.rs
```diff
@@ -349,22 +350,28 @@ where
...
-                        return Ok(true);
-                    }
+                    self.send_limiter.record_work(n);
                 }
                 Poll::Pending => {
+                    self.send_limiter.finish_cycle();
                     return Ok(false);
                 }
                 Poll::Ready(Err(e)) => {
+                    self.send_limiter.finish_cycle();
```
Ralith, line 374: All these separate exits make me start leaning more towards @djc's favored `Drop` guard approach for bounding cycle. Might be easy to miss one in a refactor.

Ralith (dismissed):
Seems straightforward enough. Prompted some non-blocking incidental thoughts.

quinn/src/work_limiter.rs
Matthias247: I don't think there really was a risk, because 1ns is really tiny, and I think division by zero is even defined for floating point. But maybe the cast to `usize` not. Anyhow - I now added a protection against that by setting the minimum time to 1ns.

quinn/src/endpoint.rs
Matthias247: Right - I didn't like the duplication too much either. Now I improved it a bit by breaking from the loop and having a common return place. Doesn't protect against someone acccidentally adding a `return` later on, but provides some decent improvement for the amount of changes.

quinn/src/work_limiter.rs
Ralith: > 1ns is really tiny

Right, the issue is that `Instant::now()` isn't necessarily anywhere near that accurate, and moving between cores can do weird things. I believe rust only guarantees monotonicity, and even that only with considerable effort!

Ralith (approved):

djc:
Thanks!

PR #1193, opened 2021-09-20
Various bulk benchmark improvements

Matthias247:
This adds a set of improvements to the bulk benchmark:
- Data transfer in the benchmark is now possible from client to server (existed before) as well as client to server (new). The latter was set as default, because its the more common use-case.
- Data sizes can be specified with units (e.g. `10G`)
- Multiple clients can be run concurrently, which provides information how well endpoints scale in case multiple connections are used (they don't do well - throughput drops 550MB/s from to 350MB/s overall when going from 1 to 2 clients)

The first commit in the set refactors the executable a bit in order to avoid having a single huge file and should make the other changes a bit easier to follow.

**Here's a invocation which gives a summary about the new functionality:**
```
cargo run --release --bin bulk --  --clients 2 --download-size 1G --upload-size 1G

Client 0 stats:
Overall upload stats:

Transferred 1073741824 bytes on 1 streams in 10.54s (97.16 MiB/s)

Stream upload metrics:

      │  Throughput   │ Duration
 AVG  │  222.94 MiB/s │     4.59s
 P0   │  222.88 MiB/s │     4.59s
 P10  │  223.00 MiB/s │     4.60s
 P50  │  223.00 MiB/s │     4.60s
 P90  │  223.00 MiB/s │     4.60s
 P100 │  223.00 MiB/s │     4.60s
Overall download stats:

Transferred 1073741824 bytes on 1 streams in 10.54s (97.16 MiB/s)

Stream download metrics:

      │  Throughput   │ Duration
 AVG  │  172.19 MiB/s │     5.95s
 P0   │  172.12 MiB/s │     5.94s
 P10  │  172.25 MiB/s │     5.95s
 P50  │  172.25 MiB/s │     5.95s
 P90  │  172.25 MiB/s │     5.95s
 P100 │  172.25 MiB/s │     5.95s

Client 1 stats:
Overall upload stats:

Transferred 1073741824 bytes on 1 streams in 10.59s (96.65 MiB/s)

Stream upload metrics:

      │  Throughput   │ Duration
 AVG  │  210.69 MiB/s │     4.86s
 P0   │  210.62 MiB/s │     4.86s
 P10  │  210.75 MiB/s │     4.86s
 P50  │  210.75 MiB/s │     4.86s
 P90  │  210.75 MiB/s │     4.86s
 P100 │  210.75 MiB/s │     4.86s
Overall download stats:

Transferred 1073741824 bytes on 1 streams in 10.59s (96.65 MiB/s)

Stream download metrics:

      │  Throughput   │ Duration
 AVG  │  178.56 MiB/s │     5.73s
 P0   │  178.50 MiB/s │     5.73s
 P10  │  178.62 MiB/s │     5.74s
 P50  │  178.62 MiB/s │     5.74s
 P90  │  178.62 MiB/s │     5.74s
 P100 │  178.62 MiB/s │     5.74s
 ```

### CODE REVIEW

File: bench/src/bin/bulk.rs [hunk 1]
```diff
@@ -155,3 +186,30 @@ async fn handle_client_stream(
...
+    download_stats: Stats,
+}
+
+impl ClientStats {
+    pub fn new() -> Self {
+        Self {
+            upload_stats: Stats::default(),
+            download_stats: Stats::default(),
+        }
+    }
```
djc, line 201: Derive `Default` for `ClientStats` instead?

djc:
Nice work!

bench/src/bin/bulk.rs
Matthias247: makes sense. updated

bench/src/bin/bulk.rs
```diff
@@ -58,8 +59,11 @@ async fn server(mut incoming: quinn::Incoming, opt: Opt) -> Result<()> {
         };
         trace!("stream established");
 
-        let _: tokio::task::JoinHandle<Result<usize>> =
-            tokio::spawn(async move { drain_stream(&mut stream, opt.read_unordered).await });
+        let _: tokio::task::JoinHandle<Result<()>> = tokio::spawn(async move {
+            drain_stream(&mut recv_stream, opt.read_unordered).await?;
+            send_data_on_stream(&mut send_stream, opt.download_size).await?;
```
Ralith: Should these be concurrent? Sequential seems of limited value, since you could get similar data from just running two tests. Alternatively, use unidirectional streams in both directions, with handlers spawned independently?

bench/src/lib.rs
```diff
@@ -160,10 +161,16 @@ pub struct Opt {
     #[structopt(long = "max_streams", short = "m", default_value = "1")]
     pub max_streams: usize,
     /// Number of bytes to transmit from server to client
-    #[structopt(long, default_value = "1073741824")]
+    ///
+    /// This can use SI prefixes for sizes. E.g. 1M will transfer 1MB, 10G
+    /// will transfer 10GB.
```
Ralith, line 169: Should say `1MiB` and `10GiB`, and similar for `upload_size`.

Ralith (changes requested):

bench/src/lib.rs
Matthias247: good call. I actually also don't like the ambiguous units. Updated

bench/src/bin/bulk.rs
Matthias247: I thought about it. But that would require adding either a `join!` which adds overhead due to polling non-ready tasks, or adding more subtasks or futures-unordered. In the end I decided its not worth it since testing both directions at the  same time doesn't seem super useful, and sequential still gives useful values since at least the per-stream metrics only cover the actual transfer time inside the stream.

Ralith: If we use unidirectional streams in each direction, I don't think the actual number of subtasks needs to change, since streams for an unused transfer direction can be dynamically omitted.

That said, I don't see a clear requirement for the capability, so :shrug:, nonblocking.

Ralith (approved):

PR #1194, opened 2021-09-27
Explicit TLS configuration setup

djc:
Extracts the parts of #1150 that don't rely on rustls 0.20 (now released) and the rustls-native-certs update (not yet done). Arguably should have done this sooner to avoid the rebase, but let's get it done now?

### DIFF
--- a/bench/src/lib.rs
+++ b/bench/src/lib.rs
@@ -30,12 +30,8 @@ pub fn server_endpoint(
     key: quinn::PrivateKey,
     opt: &Opt,
 ) -> (SocketAddr, quinn::Incoming) {
-    let mut server_config = quinn::ServerConfigBuilder::default();
-    server_config
-        .certificate(quinn::CertificateChain::from_certs(vec![cert]), key)
-        .unwrap();
-
-    let mut server_config = server_config.build();
+    let cert_chain = quinn::CertificateChain::from_certs(vec![cert]);
+    let mut server_config = quinn::ServerConfig::with_single_cert(cert_chain, key).unwrap();
     server_config.transport = Arc::new(transport_config(opt));
 
     let mut endpoint = quinn::EndpointBuilder::default();
--- a/bench/src/lib.rs
+++ b/bench/src/lib.rs
@@ -61,18 +61,12 @@ pub async fn connect_client(
         .bind(&SocketAddr::new(IpAddr::V6(Ipv6Addr::LOCALHOST), 0))
         .unwrap();
 
-    let mut config = quinn::ClientConfig::default();
-    let crypto_config = Arc::get_mut(&mut config.crypto).unwrap();
+    let mut client_config = quinn::ClientConfig::with_root_certificates(vec![server_cert]).unwrap();
+    client_config.transport = Arc::new(transport_config(&opt));
 
+    let crypto_config = Arc::get_mut(&mut client_config.crypto).unwrap();
     crypto_config.ciphersuites.clear();
     crypto_config.ciphersuites.push(opt.cipher.as_rustls());
-
-    let mut client_config = quinn::ClientConfigBuilder::new(config);
-    client_config
-        .add_certificate_authority(server_cert)
-        .unwrap();
-    let mut client_config = client_config.build();
-    client_config.transport = Arc::new(transport_config(&opt));
 
     let quinn::NewConnection { connection, .. } = endpoint
         .connect_with(client_config, &server_addr, "localhost")
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -8,7 +8,7 @@ use crate::crypto::types::{Certificate, CertificateChain, PrivateKey};
 use crate::{
     cid_generator::{ConnectionIdGenerator, RandomConnectionIdGenerator},
     congestion,
-    crypto::{self, HandshakeTokenKey as _, HmacKey as _, ServerConfig as _},
+    crypto::{self, HandshakeTokenKey as _, HmacKey as _},
     VarInt, VarIntBoundsExceeded, DEFAULT_SUPPORTED_VERSIONS,
 };
 
@@ -469,10 +469,10 @@ where
     S: crypto::Session,
 {
     /// Create a default config with a particular `master_key`
-    pub fn new(prk: S::HandshakeTokenKey) -> Self {
+    pub fn new(crypto: S::ServerConfig, prk: S::HandshakeTokenKey) -> Self {
         Self {
             transport: Arc::new(TransportConfig::default()),
-            crypto: S::ServerConfig::new(),
+            crypto,
 
             token_key: Arc::new(prk),
             use_stateless_retry: false,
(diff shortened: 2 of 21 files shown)


quinn-proto/src/tests/util.rs
```diff
@@ -375,38 +375,29 @@ impl Write for TestWriter {
...
-        .set_single_cert(
-            rustls::internal::pemfile::certs(&mut cert.as_bytes()).unwrap(),
-            rustls::PrivateKey(key.to_vec()),
-        )
-        .unwrap();
-    ServerConfig {
-        crypto,
-        ..Default::default()
-    }
+pub fn server_config(cert_and_key: Option<(Certificate, PrivateKey)>) -> ServerConfig {
```
Ralith, line 378: Rather than adding a new `None` argument in a ton of places, should we have two different helper methods?

quinn/src/endpoint.rs
```diff
@@ -69,13 +68,12 @@ where
...
-        self.connect_with(
-            self.default_client_config
-                .get_or_init(ClientConfig::default)
-                .clone(),
-            addr,
-            server_name,
-        )
+        let config = match &self.default_client_config {
+            Some(config) => config.clone(),
+            None => return Err(ConnectError::NoDefaultClientConfig),
```
Ralith, line 73: Could the introduction of this error, and removal of `OnceCell`, be factored out into a separate commit? Also, can/should the `once_cell` dep be dropped?

Ralith (commented):
Looks good overall; this definitely bites people on a regular basis.

Seems like there's two identically named commits for some reason?

quinn-proto/src/config.rs
```diff
@@ -525,14 +525,38 @@ where
...
+        crypto.set_single_cert(cert_chain.certs, key.inner)?;
+        crypto.max_early_data_size = u32::max_value();
+
+        Ok(Self::with_crypto(Arc::new(crypto)))
+    }
+
+    /// Create a server config with the given [`rustls::ServerConfig`]
+    ///
+    /// Uses a randomized handshake token key.
+    pub fn with_crypto(crypto: Arc<rustls::ServerConfig>) -> Self {
```
Matthias247, line 550: Do you want to validate here that the `ServerConfig` is suitable for QUIC? E.g. check for TLS1.3 only protocol and QUIC cipher suites?

Matthias247 (commented):
I've only skimmed it.

So one of the gotchas with the previous configs was that some `::default()` methods made rustls load TLS root certs from the system, which turned out to be very expensive (100ms+). Sometimes those even had not been used and where overwritten by custom config later on. Is this now better/worse/the-same with these changes?

Ralith:
> Is this now better/worse/the-same with these changes?

I think this is now the same or better. No `ClientConfig` is constructed by default, so servers will never try to load roots, and clients now have more explicit control in the form of calling `with_native_roots` or `with_root_certificates`.

quinn-proto/src/config.rs
```diff
@@ -601,27 +611,60 @@ where
...
+            }
+        };
+
+        Self::new(roots, None)
+    }
+
+    /// Create a client configuration that trusts specified trust anchors
+    ///
+    /// If `ct_logs` is `None`, will use the default Certificate Transparency logs depending on
+    /// whether the `ct-logs` features is enabled or not. Otherwise, will use the specified logs.
```
Ralith, line 635: The feature is named "certificate-transparency". However, this is now convenient enough to pass in explicitly that I feel like maybe we should just drop the ct-logs dep entirely and let downstream users (who can much more easily update the package regularly) deal with it as needed.

djc:
> Seems like there's two identically named commits for some reason?

They're not identically named, one is for `ClientConfig` and one is for `ServerConfig`.

quinn-proto/src/config.rs
djc: Right now `ServerConfig` does not support introspection of cipher suites and protocol versions. There are no QUIC cipher suites AFAICT, there are TLS 1.3 cipher suites and all of these are supported by QUIC (although Rustls does not support the CCM cipher suites). The config builder pattern in Rustls 0.20 makes sure that protocol versions and cipher suites configured in Rustls match; it also returns an error for new QUIC connections if the configuration cannot support QUIC. We can add a check here once https://github.com/rustls/rustls/pull/834 is released.

quinn/src/endpoint.rs
djc: Done. Had some sloppy rebasing going on that nixed the removal of the dependency.

quinn-proto/src/tests/util.rs
djc: Fair enough, done.

quinn-proto/src/config.rs
djc: I think that makes sense, especially now that Rustls leaves certificate transparency implicitly disabled unless explicitly configured.

bench/src/lib.rs
```diff
@@ -61,19 +61,14 @@ pub async fn connect_client(
         .bind(&SocketAddr::new(IpAddr::V6(Ipv6Addr::LOCALHOST), 0))
         .unwrap();
 
-    let mut config = quinn::ClientConfig::default();
-    let crypto_config = Arc::get_mut(&mut config.crypto).unwrap();
+    let mut client_config =
+        quinn::ClientConfig::with_root_certificates(vec![server_cert.clone()]).unwrap();
```
Ralith, line 61: Is the `clone` here really needed? `server_cert` is owned and doesn't seem to be used below.

quinn-proto/src/config.rs
```diff
@@ -601,27 +601,48 @@ where
...
+            }
+        };
+
+        Self::new(roots)
+    }
+
+    /// Create a client configuration that trusts specified trust anchors
+    ///
+    /// If `ct_logs` is `None`, will use the default Certificate Transparency logs depending on
+    /// whether the `ct-logs` features is enabled or not. Otherwise, will use the specified logs.
```
Ralith, line 635: Outdated comment

quinn-proto/src/tests/util.rs
```diff
@@ -392,21 +392,16 @@ pub fn server_config() -> ServerConfig {
...
-    Arc::make_mut(&mut crypto).key_log = Arc::new(KeyLogFile::new());
-    Arc::make_mut(&mut crypto).enable_early_data = true;
-    ClientConfig {
-        transport: Default::default(),
-        crypto,
-    }
+    let mut config = ClientConfig::with_root_certificates(certs).unwrap();
+    Arc::make_mut(&mut config.crypto).key_log = Arc::new(KeyLogFile::new());
+    Arc::make_mut(&mut config.crypto).enable_early_data = true;
+    config
```
Ralith, line 400: Can reduce churn here similarly to the next commit (especially if we retain the name `client_config()` and use a new name for the explicit-certs version)

quinn-proto/src/tests/util.rs
```diff
@@ -375,21 +375,17 @@ impl Write for TestWriter {
...
-        .unwrap();
-    ServerConfig {
-        crypto,
-        ..Default::default()
-    }
+pub fn server_config(cert: Certificate, key: PrivateKey) -> ServerConfig {
+    ServerConfig::with_single_cert(CertificateChain::from_certs(vec![cert]), key).unwrap()
+}
+
+pub fn default_server_config() -> ServerConfig {
```
Ralith: nit: If we named this `server_config`, and the above `server_config_with_cert` or something, it'd reduce churn

Ralith (changes requested):
It's pretty great how much code this saves, while simultaneously simplifying the API.

> They're not identically named

Oops, reading is hard.

bench/src/lib.rs
djc: You're right. I'm surprised clippy didn't make a fuss about this one.

quinn-proto/src/config.rs
djc: Killed it.

quinn-proto/src/tests/util.rs
djc: Done.

quinn-proto/src/tests/util.rs
djc: Done.

Ralith (approved):
Great stuff!

PR #1195, opened 2021-09-28
make example code truly async

dzvon:
In the original case, the code was actually running synchronously. So, fix it.

### DIFF
--- a/quinn/examples/server.rs
+++ b/quinn/examples/server.rs
@@ -133,11 +133,11 @@ async fn run(options: Opt) -> Result<()> {
 
     while let Some(conn) = incoming.next().await {
         info!("connection incoming");
-        tokio::spawn(
+        tokio::spawn(async move {
             handle_connection(root.clone(), conn).unwrap_or_else(move |e| {
                 error!("connection failed: {reason}", reason = e.to_string())
-            }),
-        );
+            })
+        });
     }
 
     Ok(())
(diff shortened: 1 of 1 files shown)


djc (approved):
LGTM, thanks!

djc (approved):

Ralith (changes requested):
This looks like a no-op to me. `async { x.await }` is exactly the same as `x`, aside from inconsequential borrowck stuff. What makes you think it wasn't already async?

dzvon:
I'm sorry, I misunderstood the meaning of the `futures::TryFutureExt::unwrap_or_else` method.

https://docs.rs/futures-util/0.3.17/futures_util/future/trait.TryFutureExt.html#method.unwrap_or_else
```
Unwraps this future's output, producing a future with this future's
[`Ok`](https://docs.rs/futures_core/*/futures_core/future/trait.TryFuture.html#associatedtype.Ok) type as its
[`Output`](https://docs.rs/core/*/core/future/future/trait.Future.html#associatedtype.Output) type.
```

```rust
tokio::spawn(x.unwrap_or_else(|_|{}));
```

In this example, I thought `x` will be `await`ed first, then the future producing by `unwrap_or_else` will be put in `spawn`. Thus the flow isn't async. Sorry, it's all right now. I'll close this PR.

Sorry for waste your time.

Ralith:
No worries, we appreciate the thought!

PR #1199, opened 2021-10-06
Prepare quinn-udp for a beta release

djc:
cc @daxpedda

### DIFF
--- a/quinn-udp/Cargo.toml
+++ b/quinn-udp/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quinn-udp"
-version = "0.7.0"
+version = "0.1.0-beta.1"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/quinn-rs/quinn"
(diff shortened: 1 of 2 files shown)


quinn-udp/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "quinn-udp"
-version = "0.7.0"
+version = "0.8.0-beta.1"
```
djc, line 3: Should this be 0.1.0 (since this is a new crate) or 0.8.0 (to match quinn/quinn-proto)?

daxpedda: I vote for 0.1.0.

daxpedda (approved):

quinn-udp/Cargo.toml
Ralith: I don't feel any particular need to artificially sync versions, especially as this is an impl detail for most users. It also seems unlikely that the breakage cadence will match that of the main public APIs in the long term. So, 0.1.0 seems fine.

PR #1201, opened 2021-10-10
Monomorphize public API

### LINKED ISSUE #911: Monomorphize public API

Ralith:
Makes the assorted `quinn_proto::crypto` traits object-safe and replaces their uses with trait objects.

Fixes #911.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -803,6 +803,7 @@ impl IndexMut<ConnectionHandle> for Slab<ConnectionMeta> {
 }
 
 /// Event resulting from processing a single datagram
+#[allow(clippy::large_enum_variant)] // Not passed around extensively
 pub enum DatagramEvent {
     /// The datagram is redirected to its `Connection`
     ConnectionEvent(ConnectionEvent),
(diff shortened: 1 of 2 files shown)


Matthias247:
> 
> 
> There was no visible impact for the bulk benchmark, though someone with a better benchmarking setup (@Matthias247?) might want to have a look. We should probably have separate benchmarks for handshaking.

Where would you expect to see an impact? In general data transmission benchmarks, due to different inlining properties? Or in handshakes, due to the use of trait objects there.

I think you might be able to test the latter by configuring lots of clients with `0` requests in some of the benchmarks. (not sure at the moment which one supports more than 1 client).

I wouldn't expect anything measureable however. Handshakes are already super expensive due to crypto, and the amount of dynamic calls is tiny.

PR #1217, opened 2021-10-30
Send ping on rebind to ensure peers notice migration

Ralith:
Otherwise, a client that is only receiving will lose its connection on rebind.


djc (approved):

djc:
Nice idea!

Issue #914, opened 2020-11-13
Reduce task-switching overhead on I/O path


Ralith:
In the high-level crate, all I/O on a connection travels across a `mpsc` between connection and endpoint drivers. This has a number of costs:
- I/O buffers must constantly be reallocated
- Inbound data must be copied
- Many round trips are made through the scheduler (both userspace and, in multithreaded environments, kernel)

We could avoid these by synchronously invoking connection driver logic in the endpoint, e.g. by replacing `mpsc::UnboundedSender<ConnectionEvent>` with `Arc<Mutex<Connection>>`. This will allow for I/O to operate directly on a single set of shared buffers without excess copying or allocator traffic, and bypass the scheduler entirely for purely internal I/O such as ACKs and flow control updates.

The obvious downside is loss of parallelism in the `Connection` logic, where the overwhelming majority of the intended computational cost resides. Following discussion with @Matthias247, I suspect intra-`Endpoint` parallelism is a red herring due to the mandatory synchronization involved in sharing a single socket, which exists at the kernel level if nowhere else. QUIC is designed to scale out across multiple endpoints using load balancers, which allows for unlimited horizontal scalability (at least below the application layer). By sacrificing transport-layer parallelism within an endpoint to reduce overhead for single-threaded configurations, we therefore also reduce overhead for arbitrarily parallel deployments, even on a single host, while improving performance for the overwhelmingly more common case of few connections.

Load balancers are a common feature of massive-scale internet service deployments, but may be unfamiliar to smaller-scale service operators accustomed to getting by with a sufficiently powerful single host with failover. For future work, we should explore options for lightweight, self-contained single-host load balancing to support this pattern. QUIC's "preferred address" mechanism is an interesting option, though it requires a cooperative client.

djc:
Do we have some sense of the relative costs of the three things you mentioned? I wonder if we could alleviate the first two by doing smarter buffer management (I am thinking some kind of reference-counted ring buffer segments).

The loss of parallelism in the `Connection` logic does seem like a big downside.

Do we have a sense of how other implementations are solving these issues?

Ralith:
> Do we have some sense of the relative costs of the three things you mentioned?

I understand that @Matthias247's profiling indicates task switching is a major cost center for us, though he should confirm. He mentions 10% time spent in memcpy at 330MB/s, which is significant. Not sure about malloc.

[...]

PR #1219, opened 2021-10-30
Unify connection and endpoint drivers

Ralith:
Fixes https://github.com/quinn-rs/quinn/issues/914. This is a high-risk change, so we probably don't want it in 0.8, but I'm getting a head start on it all the same.

### DIFF
--- a/quinn/Cargo.toml
+++ b/quinn/Cargo.toml
@@ -45,6 +45,7 @@ rustls = { version = "0.20.3", default-features = false, features = ["quic"], op
 thiserror = "1.0.21"
 tracing = "0.1.10"
 tokio = { version = "1.13.0", features = ["sync"] }
+tokio-util = { version = "0.6.9", features = ["time"] }
 udp = { package = "quinn-udp", path = "../quinn-udp", version = "0.3", default-features = false }
 webpki = { version = "0.22", default-features = false, optional = true }
 
--- a/quinn/Cargo.toml
+++ b/quinn/Cargo.toml
@@ -42,10 +42,10 @@ rustc-hash = "1.1"
 pin-project-lite = "0.2"
 proto = { package = "quinn-proto", path = "../quinn-proto", version = "0.9", default-features = false }
 rustls = { version = "0.20.3", default-features = false, features = ["quic"], optional = true }
+slab = "0.4"
 thiserror = "1.0.21"
 tracing = "0.1.10"
 tokio = { version = "1.13.0", features = ["sync"] }
-tokio-util = { version = "0.6.9", features = ["time"] }
 udp = { package = "quinn-udp", path = "../quinn-udp", version = "0.3", default-features = false }
 webpki = { version = "0.22", default-features = false, optional = true }
 
@@ -54,6 +54,7 @@ anyhow = "1.0.22"
 crc = "3"
 bencher = "0.1.5"
 directories-next = "2"
+proptest = "=1.0.0" # Pinned for MSRV
 rand = "0.8"
 rcgen = "0.10.0"
 rustls-pemfile = "1.0.0"
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -21,7 +21,6 @@ use tokio::{
 };
 use tokio_util::time::delay_queue;
 use tracing::debug_span;
-use udp::UdpState;
 
 use crate::{
     mutex::Mutex,
@@ -45,7 +44,6 @@ impl Connecting {
         dirty: mpsc::UnboundedSender<ConnectionHandle>,
         handle: ConnectionHandle,
         conn: proto::Connection,
-        udp_state: Arc<UdpState>,
     ) -> (Connecting, ConnectionRef) {
         let (on_handshake_data_send, on_handshake_data_recv) = oneshot::channel();
         let (on_connected_send, on_connected_recv) = oneshot::channel();
@@ -55,7 +53,6 @@ impl Connecting {
             dirty,
             on_handshake_data_send,
             on_connected_send,
-            udp_state,
         );
 
         (
@@ -698,7 +695,6 @@ impl ConnectionRef {
         dirty: mpsc::UnboundedSender<ConnectionHandle>,
         on_handshake_data: oneshot::Sender<()>,
         on_connected: oneshot::Sender<bool>,
-        udp_state: Arc<UdpState>,
     ) -> Self {
         let _ = dirty.send(handle);
         Self(Arc::new(ConnectionInner {
@@ -719,7 +715,6 @@ impl ConnectionRef {
                 stopped: FxHashMap::default(),
                 error: None,
                 ref_count: 0,
-                udp_state,
             }),
             shared: Shared::default(),
         }))
@@ -799,15 +794,16 @@ pub(crate) struct State {
     pub(crate) error: Option<ConnectionError>,
     /// Number of live handles that can be used to initiate or handle I/O; excludes the driver
     ref_count: usize,
-    udp_state: Arc<UdpState>,
 }
 
 impl State {
-    pub(crate) fn drive_transmit(&mut self, out: &mut VecDeque<proto::Transmit>) -> bool {
+    pub(crate) fn drive_transmit(
+        &mut self,
+        out: &mut VecDeque<proto::Transmit>,
+        max_datagrams: usize,
+    ) -> bool {
         let now = Instant::now();
         let mut transmits = 0;
-
-        let max_datagrams = self.udp_state.max_gso_segments();
 
         while let Some(t) = self.inner.poll_transmit(now, max_datagrams) {
             transmits += match t.segment_size {
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -15,14 +15,11 @@ use pin_project_lite::pin_project;
 use proto::{ConnectionError, ConnectionHandle, ConnectionStats, Dir, StreamEvent, StreamId};
 use rustc_hash::FxHashMap;
 use thiserror::Error;
-use tokio::{
-    sync::{futures::Notified, mpsc, oneshot, Notify},
-    time::Instant as TokioInstant,
-};
-use tokio_util::time::delay_queue;
+use tokio::sync::{futures::Notified, mpsc, oneshot, Notify};
 use tracing::debug_span;
 
 use crate::{
+    delay_queue::Timer,
     mutex::Mutex,
     recv_stream::RecvStream,
     send_stream::{SendStream, WriteError},
@@ -784,8 +781,8 @@ pub(crate) struct State {
     on_handshake_data: Option<oneshot::Sender<()>>,
     on_connected: Option<oneshot::Sender<bool>>,
     connected: bool,
-    pub(crate) timer_handle: Option<delay_queue::Key>,
-    pub(crate) timer_deadline: Option<TokioInstant>,
+    pub(crate) timer_handle: Option<Timer>,
+    pub(crate) timer_deadline: Option<Instant>,
     pub(crate) blocked_writers: FxHashMap<StreamId, Waker>,
     pub(crate) blocked_readers: FxHashMap<StreamId, Waker>,
     pub(crate) finishing: FxHashMap<StreamId, oneshot::Sender<Option<WriteError>>>,
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -1,5 +1,6 @@
 use std::{
     any::Any,
+    collections::VecDeque,
     fmt,
     future::Future,
     net::{IpAddr, SocketAddr},
@@ -9,13 +10,16 @@ use std::{
     time::{Duration, Instant},
 };
 
-use crate::runtime::{AsyncTimer, Runtime};
 use bytes::Bytes;
 use pin_project_lite::pin_project;
 use proto::{ConnectionError, ConnectionHandle, ConnectionStats, Dir, StreamEvent, StreamId};
 use rustc_hash::FxHashMap;
 use thiserror::Error;
-use tokio::sync::{futures::Notified, mpsc, oneshot, Notify};
+use tokio::{
+    sync::{futures::Notified, mpsc, oneshot, Notify},
+    time::Instant as TokioInstant,
+};
+use tokio_util::time::delay_queue;
 use tracing::debug_span;
 use udp::UdpState;
 
@@ -23,7 +27,7 @@ use crate::{
     mutex::Mutex,
     recv_stream::RecvStream,
     send_stream::{SendStream, WriteError},
-    EndpointEvent, VarInt,
+    VarInt,
 };
 use proto::congestion::Controller;
 
@@ -38,25 +42,22 @@ pub struct Connecting {
 
 impl Connecting {
     pub(crate) fn new(
+        dirty: mpsc::UnboundedSender<ConnectionHandle>,
         handle: ConnectionHandle,
         conn: proto::Connection,
-        endpoint_events: mpsc::UnboundedSender<(ConnectionHandle, EndpointEvent)>,
         udp_state: Arc<UdpState>,
-        runtime: Arc<dyn Runtime>,
     ) -> (Connecting, ConnectionRef) {
         let (on_handshake_data_send, on_handshake_data_recv) = oneshot::channel();
         let (on_connected_send, on_connected_recv) = oneshot::channel();
         let conn = ConnectionRef::new(
             handle,
             conn,
-            endpoint_events,
+            dirty,
             on_handshake_data_send,
             on_connected_send,
             udp_state,
-            runtime.clone(),
         );
 
-        runtime.spawn(Box::pin(ConnectionDriver(conn.clone())));
         (
             Connecting {
                 conn: Some(conn.clone()),
@@ -199,53 +200,6 @@ impl Future for ZeroRttAccepted {
     type Output = bool;
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
         Pin::new(&mut self.0).poll(cx).map(|x| x.unwrap_or(false))
-    }
-}
-
-/// A future that drives protocol logic for a connection
-///
-/// This future handles the protocol logic for a single connection, routing events from the
-/// `Connection` API object to the `Endpoint` task and the related stream-related interfaces.
-/// It also keeps track of outstanding timeouts for the `Connection`.
-///
-/// If the connection encounters an error condition, this future will yield an error. It will
-/// terminate (yielding `Ok(())`) if the connection was closed without error. Unlike other
-/// connection-related futures, this waits for the draining period to complete to ensure that
-/// packets still in flight from the peer are handled gracefully.
-#[must_use = "connection drivers must be spawned for their connections to function"]
-#[derive(Debug)]
-struct ConnectionDriver(ConnectionRef);
-
-impl Future for ConnectionDriver {
-    type Output = ();
-
-    #[allow(unused_mut)] // MSRV
-    fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
-        let conn = &mut *self.0.state.lock("poll");
-
-        let span = debug_span!("drive", id = conn.handle.0);
-        let _guard = span.enter();
-
-        let mut keep_going = conn.drive_transmit();
-        // If a timer expires, there might be more to transmit. When we transmit something, we
-        // might need to reset a timer. Hence, we must loop until neither happens.
-        keep_going |= conn.drive_timer(cx);
-        conn.forward_endpoint_events();
-        conn.forward_app_events(&self.0.shared);
-
-        if !conn.inner.is_drained() {
-            if keep_going {
-                // If the connection hasn't processed all tasks, schedule it again
-                cx.waker().wake_by_ref();
-            } else {
-                conn.driver = Some(cx.waker().clone());
-            }
-            return Poll::Pending;
-        }
-        if conn.error.is_none() {
-            unreachable!("drained connections always have an error");
-        }
-        Poll::Ready(())
     }
 }
 
@@ -741,23 +695,24 @@ impl ConnectionRef {
     fn new(
         handle: ConnectionHandle,
         conn: proto::Connection,
-        endpoint_events: mpsc::UnboundedSender<(ConnectionHandle, EndpointEvent)>,
+        dirty: mpsc::UnboundedSender<ConnectionHandle>,
         on_handshake_data: oneshot::Sender<()>,
         on_connected: oneshot::Sender<bool>,
         udp_state: Arc<UdpState>,
-        runtime: Arc<dyn Runtime>,
     ) -> Self {
+        let _ = dirty.send(handle);
         Self(Arc::new(ConnectionInner {
             state: Mutex::new(State {
                 inner: conn,
-                driver: None,
                 handle,
+                span: debug_span!("connection", id = handle.0),
+                is_dirty: true,
+                dirty,
                 on_handshake_data: Some(on_handshake_data),
                 on_connected: Some(on_connected),
                 connected: false,
-                timer: None,
+                timer_handle: None,
                 timer_deadline: None,
-                endpoint_events,
                 blocked_writers: FxHashMap::default(),
                 blocked_readers: FxHashMap::default(),
                 finishing: FxHashMap::default(),
@@ -765,7 +720,6 @@ impl ConnectionRef {
                 error: None,
                 ref_count: 0,
                 udp_state,
-                runtime,
             }),
             shared: Shared::default(),
         }))
@@ -825,14 +779,18 @@ pub(crate) struct Shared {
 
 pub(crate) struct State {
     pub(crate) inner: proto::Connection,
-    driver: Option<Waker>,
     handle: ConnectionHandle,
+    pub(crate) span: tracing::Span,
+    /// Whether `handle` has been sent to `dirty` since the last time this connection was driven by
+    /// the endpoint. Ensures `dirty`'s size stays bounded regardless of activity.
+    pub(crate) is_dirty: bool,
+    /// `handle` is sent here when `wake` is called, prompting the endpoint to drive the connection
+    dirty: mpsc::UnboundedSender<ConnectionHandle>,
     on_handshake_data: Option<oneshot::Sender<()>>,
     on_connected: Option<oneshot::Sender<bool>>,
     connected: bool,
-    timer: Option<Pin<Box<dyn AsyncTimer>>>,
-    timer_deadline: Option<Instant>,
-    endpoint_events: mpsc::UnboundedSender<(ConnectionHandle, EndpointEvent)>,
+    pub(crate) timer_handle: Option<delay_queue::Key>,
+    pub(crate) timer_deadline: Option<TokioInstant>,
     pub(crate) blocked_writers: FxHashMap<StreamId, Waker>,
     pub(crate) blocked_readers: FxHashMap<StreamId, Waker>,
     pub(crate) finishing: FxHashMap<StreamId, oneshot::Sender<Option<WriteError>>>,
@@ -842,11 +800,10 @@ pub(crate) struct State {
     /// Number of live handles that can be used to initiate or handle I/O; excludes the driver
     ref_count: usize,
     udp_state: Arc<UdpState>,
-    runtime: Arc<dyn Runtime>,
 }
 
 impl State {
-    fn drive_transmit(&mut self) -> bool {
+    pub(crate) fn drive_transmit(&mut self, out: &mut VecDeque<proto::Transmit>) -> bool {
         let now = Instant::now();
         let mut transmits = 0;
 
@@ -857,10 +814,7 @@ impl State {
                 None => 1,
                 Some(s) => (t.contents.len() + s - 1) / s, // round up
             };
-            // If the endpoint driver is gone, noop.
-            let _ = self
-                .endpoint_events
-                .send((self.handle, EndpointEvent::Transmit(t)));
+            out.push_back(t);
 
             if transmits >= MAX_TRANSMIT_DATAGRAMS {
                 // TODO: What isn't ideal here yet is that if we don't poll all
@@ -874,16 +828,7 @@ impl State {
         false
     }
 
-    fn forward_endpoint_events(&mut self) {
-        while let Some(event) = self.inner.poll_endpoint_events() {
-            // If the endpoint driver is gone, noop.
-            let _ = self
-                .endpoint_events
-                .send((self.handle, EndpointEvent::Proto(event)));
-        }
-    }
-
-    fn forward_app_events(&mut self, shared: &Shared) {
+    pub(crate) fn forward_app_events(&mut self, shared: &Shared) {
         while let Some(event) = self.inner.poll() {
             use proto::Event::*;
             match event {
@@ -949,61 +894,14 @@ impl State {
         }
     }
 
-    fn drive_timer(&mut self, cx: &mut Context) -> bool {
-        // Check whether we need to (re)set the timer. If so, we must poll again to ensure the
-        // timer is registered with the runtime (and check whether it's already
-        // expired).
-        match self.inner.poll_timeout() {
-            Some(deadline) => {
-                if let Some(delay) = &mut self.timer {
-                    // There is no need to reset the tokio timer if the deadline
-                    // did not change
-                    if self
-                        .timer_deadline
-                        .map(|current_deadline| current_deadline != deadline)
-                        .unwrap_or(true)
-                    {
-                        delay.as_mut().reset(deadline);
-                    }
-                } else {
-                    self.timer = Some(self.runtime.new_timer(deadline));
-                }
-                // Store the actual expiration time of the timer
-                self.timer_deadline = Some(deadline);
-            }
-            None => {
-                self.timer_deadline = None;
-                return false;
-            }
-        }
-
-        if self.timer_deadline.is_none() {
-            return false;
-        }
-
-        let delay = self
-            .timer
-            .as_mut()
-            .expect("timer must exist in this state")
-            .as_mut();
-        if delay.poll(cx).is_pending() {
-            // Since there wasn't a timeout event, there is nothing new
-            // for the connection to do
-            return false;
-        }
-
-        // A timer expired, so the caller needs to check for
-        // new transmits, which might cause new timers to be set.
-        self.inner.handle_timeout(Instant::now());
-        self.timer_deadline = None;
-        true
-    }
-
-    /// Wake up a blocked `Driver` task to process I/O
+    /// Wake up endpoint to process I/O by marking it as "dirty" for the endpoint
     pub(crate) fn wake(&mut self) {
-        if let Some(x) = self.driver.take() {
-            x.wake();
+        if self.is_dirty {
+            return;
         }
+        self.is_dirty = true;
+        // Take no action if the endpoint is gone
+        let _ = self.dirty.send(self.handle);
     }
 
     /// Used to wake up all blocked futures when the connection becomes closed for any reason
@@ -1054,18 +952,6 @@ impl State {
             Ok(())
         } else {
             Err(())
-        }
-    }
-}
-
-impl Drop for State {
-    fn drop(&mut self) {
-        if !self.inner.is_drained() {
-            // Ensure the endpoint can tidy up
-            let _ = self.endpoint_events.send((
-                self.handle,
-                EndpointEvent::Proto(proto::EndpointEvent::drained()),
-            ));
         }
     }
 }
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -23,7 +23,7 @@ use crate::{
     mutex::Mutex,
     recv_stream::RecvStream,
     send_stream::{SendStream, WriteError},
-    ConnectionEvent, EndpointEvent, VarInt,
+    EndpointEvent, VarInt,
 };
 use proto::congestion::Controller;
 
@@ -41,17 +41,15 @@ impl Connecting {
         handle: ConnectionHandle,
         conn: proto::Connection,
         endpoint_events: mpsc::UnboundedSender<(ConnectionHandle, EndpointEvent)>,
-        conn_events: mpsc::UnboundedReceiver<ConnectionEvent>,
         udp_state: Arc<UdpState>,
         runtime: Arc<dyn Runtime>,
-    ) -> Connecting {
+    ) -> (Connecting, ConnectionRef) {
         let (on_handshake_data_send, on_handshake_data_recv) = oneshot::channel();
         let (on_connected_send, on_connected_recv) = oneshot::channel();
         let conn = ConnectionRef::new(
             handle,
             conn,
             endpoint_events,
-            conn_events,
             on_handshake_data_send,
             on_connected_send,
             udp_state,
@@ -59,12 +57,14 @@ impl Connecting {
         );
 
         runtime.spawn(Box::pin(ConnectionDriver(conn.clone())));
-
-        Connecting {
-            conn: Some(conn),
-            connected: on_connected_recv,
-            handshake_data_ready: Some(on_handshake_data_recv),
-        }
+        (
+            Connecting {
+                conn: Some(conn.clone()),
+                connected: on_connected_recv,
+                handshake_data_ready: Some(on_handshake_data_recv),
+            },
+            conn,
+        )
     }
 
     /// Convert into a 0-RTT or 0.5-RTT connection at the cost of weakened security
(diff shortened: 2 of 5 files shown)


Ralith:
This could probably use some more iteration, but is now functionally complete. Next step is careful benchmarking.

djc:
Is this still waiting for benchmarking?

Ralith:
Initial results were inconclusive, with increases in ACK frames sent, presumed due to reduced internal buffering. I've therefore been waiting for @Matthias247 to finish his WIP delayed ACK support which should remove that variable.

DemiMarie:
How much overhead results from the use of atomic operations and mutexes?  I wonder if a better approach would be to make everything `!Send` and `!Sync` and to use non-atomic operations.  Users would be expected to shard their program across cores using `SO_RESUEPORT` or similar.

DemiMarie:
RDMA verbs are thread-safe, but I wonder if their performance is hurt less due to less reference counting.

Ralith:
> How much overhead results from the use of atomic operations and mutexes

Approximately none. CPU time is mostly spent in crypto, syscalls, and memory management, IIRC (regardless of this PR). If you want to profile and experiment yourself, that would be very welcome!

> I wonder if a better approach would be to make everything !Send and !Sync and to use non-atomic operations

Very early versions of quinn took this approach. It didn't seem to help.

> Users would be expected to shard their program across cores using SO_RESUEPORT or similar.

This is appealing at first, but unfortunately routing UDP datagrams to the correct thread is difficult. Further, any genuinely large-scale deployment of QUIC will have a load-balancing front-end that renders such measures redundant, so the set of users who might stand to benefit seems very small.

DemiMarie:
> > Users would be expected to shard their program across cores using SO_RESUEPORT or similar.
> 
> This is appealing at first, but unfortunately routing UDP datagrams to the correct thread is difficult. Further, any genuinely large-scale deployment of QUIC will have a load-balancing front-end that renders such measures redundant, so the set of users who might stand to benefit seems very small.

My understanding is that load balancers can only send the packet to the correct device, and something else has to get the packet to the correct thread.  This can be done using hardware or software RSS or a custom software eBPF program.

Ralith:
That's incorrect, or at best needlessly complicated. A device with multiple independent threads can be seamlessly handled by a load balancer by giving each thread its own address. Reusing the same port only adds complexity.

Ralith:
> That assumes you have unlimited addresses

It assumes you have more addresses available than CPU cores. Recall that UDP ports alone trivially give you ~2^16 addresses.

> Also forwarding of misdirected packets needs to be addressed.

A load balancer routing packets deliberately does not have that problem.

DemiMarie:
> > Also forwarding of misdirected packets needs to be addressed.
> 
> A load balancer routing packets deliberately does not have that problem.

Depends on whether the load balancer can use connection IDs, or is only able to use the 5-tuple.  Lots of switches are only capable of the latter, and hardware load balancers are much more expensive.

Matthias247:
I don't think this discussion adds a lot to this PR, since it's completely orthogonal and might be better off in it's own topic.

But yes, Demi is right that a there can be different types of packet load balancers for horizontal scaling. And they might either just route by 5tuple (in which case connection migration would not necessarily work - since packets will be misrouted after migration), or by connection ID.

DemiMarie:
> I don't think this discussion adds a lot to this PR, since it's completely orthogonal and might be better off in it's own topic.
> 
> But yes, Demi is right that a there can be different types of packet load balancers for horizontal scaling. And they might either just route by 5tuple (in which case connection migration would not necessarily work - since packets will be misrouted after migration), or by connection ID.

Migration is sufficiently rare that one can explicitly send misrouted packets to the correct server.

Ralith:
This has been rebased, but the `DelayQueue` it introduces depends on `tokio::time::Sleep`, so additional work is required to be runtime-independent.

Ralith:
Sure, I can split those out tonight.

WilliamVenner:
Hi, I gave your branch a try while doing some benchmarks of my own. I did not find any significant difference in the transfer speed of a 500 MiB file.

Code I used: https://github.com/WilliamVenner/quictransfer/blob/master/src/main.rs

quinn/src/delay_queue.rs
```diff
@@ -0,0 +1,442 @@
...
+
+/// Stores values to be yielded at specific times in the future
+#[derive(Debug)]
+pub struct DelayQueue<T> {
+    /// Definitions of each active timer
+    timers: Slab<TimerState<T>>,
+
+    /// Hierarchical timer wheel
+    ///
+    /// Timers are bucketed into slots organized into levels. Level n spans 2^(`LOG_2_SLOTS` *
```
Matthias247, line 13: are the backticks here for the sake of having a rustdoc link to the constant? For comment readability the might be better placed around the whole math (before `2^`). Not sure if rustdoc actually does something for private fields and private constants.

quinn/src/delay_queue.rs
```diff
@@ -0,0 +1,442 @@
...
+    ///
+    /// The set of timers in a slot is an intrusive doubly-linked list, allowing constant-time
+    /// insertion and removal.
+    levels: [Level; LEVELS],
+
+    /// Earliest point at which a timer may be pending
+    ///
+    /// Each `LOG_2_SLOTS` bits of this are a cursor into the associated level, in order of
+    /// ascending significance.
+    next_tick: u64,
```
Matthias247, line 31: I assume the queue assumes no specific time unit (like `ms`) and just supports any, as long as the user is consistent using the same?

quinn/src/delay_queue.rs
```diff
@@ -0,0 +1,442 @@
...
+    }
+}
+
+const LEVELS: usize = 11;
+const LOG_2_SLOTS: usize = 6;
+const SLOTS: usize = 1 << LOG_2_SLOTS;
+
+// Future work: add a niche here
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub struct Timer(usize);
```
Matthias247, line 301: maybe add a comment that the `usize` is an index into the slab. When I looked at this before checking how `Timer` is actually used I assumed it would be a timestamp.

quinn/src/delay_queue.rs
```diff
@@ -0,0 +1,442 @@
...
+    #[cfg(test)]
+    pub fn remove(&mut self, timer: Timer) -> T {
+        self.unlink(timer);
+        self.timers.remove(timer.0).value
+    }
+
+    /// Remove all references to `timer`
+    fn unlink(&mut self, timer: Timer) {
+        let (level, slot) = timer_index(self.next_tick, self.timers[timer.0].expiry);
+        let link = self.levels[level].slots[slot].unwrap();
```
Matthias247, line 197: Doesn't this require a `.take()` to set the slot to `None`? Or does it not matter because the slot won't be linked anymore? Maybe for testing add some marker that prevents `Copy` on `Timer` to avoid accidental issues.

quinn/src/delay_queue.rs
```diff
@@ -0,0 +1,442 @@
...
+        self.unlink(timer);
+        self.timers[timer.0].expiry = timeout.max(self.next_tick);
+        self.schedule(timer);
+    }
+
+    /// Cancel `timer`
+    #[cfg(test)]
+    pub fn remove(&mut self, timer: Timer) -> T {
+        self.unlink(timer);
+        self.timers.remove(timer.0).value
```
Matthias247, line 191: Does `Slab::remove()` ever shrink the allocation if the amount of required items gets reduced? Without that a lot of timer memory might hang around if one had a large amount of connections, and then moved to a lower amount. A "real" intrusive list would avoid that. Or one could do a manual shrink if e.g. the utilized capacity gets below a certain percentage..

quinn/src/delay_queue.rs
```diff
@@ -0,0 +1,442 @@
...
+/// Compute the first tick that lies within a slot
+fn slot_start(base: u64, level: usize, slot: usize) -> u64 {
+    let shift = (level * LOG_2_SLOTS) as u64;
+    (base & ((!0 << shift) << LOG_2_SLOTS as u64)) | ((slot as u64) << shift)
+}
+
+/// Compute the level and slot for a certain expiry
+fn timer_index(base: u64, expiry: u64) -> (usize, usize) {
+    let differing_bits = base ^ expiry;
+    let level = (63 - (differing_bits | 1).leading_zeros()) as usize / LOG_2_SLOTS;
```
Matthias247, line 240: Looks quite magic!

quinn/src/delay_queue.rs
```diff
@@ -0,0 +1,442 @@
...
+        debug_assert_eq!(
+            self.timers[timer.0].next, None,
+            "timer isn't already scheduled"
+        );
+        debug_assert_eq!(
+            self.timers[timer.0].prev, None,
+            "timer isn't already scheduled"
+        );
+        let (level, slot) = timer_index(self.next_tick, self.timers[timer.0].expiry);
+        // Insert `timer` at the head of the list in the target slot
```
Matthias247, line 146: Does that mean a freshly added timer will expire before other timers for the same timestamp? I think for fairness reasons it would be good that it expires after those.

quinn/src/delay_queue.rs
```diff
@@ -0,0 +1,442 @@
...
+                }
+            }
+        }
+        None
+    }
+
+    /// Register a timer that will yield `value` at `timeout`
+    pub fn insert(&mut self, timeout: u64, value: T) -> Timer {
+        let timer = Timer(self.timers.insert(TimerState {
+            expiry: timeout.max(self.next_tick),
```
Matthias247, line 171: `next_tick` is the current time - so one can schedule immediate timeouts - right?

Matthias247 (commented):
just looked a bit at the DelayQueue. Looks pretty cool! And I mostly added clarification questions

quinn/src/endpoint.rs
```diff
@@ -183,10 +185,9 @@ impl Endpoint {
             addr
         };
         let (ch, conn) = endpoint.inner.connect(config, addr, server_name)?;
-        let udp_state = endpoint.udp_state.clone();
-        Ok(endpoint
-            .connections
-            .insert(ch, conn, udp_state, self.runtime.clone()))
+        let dirty = endpoint.dirty_send.clone();
```
Matthias247, line 188: would it be worthwhile to do this in the `Connection` context/constructor, which also triggers the other wakeups?   As far as I understand the dirty sender gets forwarded there anyway.

quinn/src/endpoint.rs
```diff
@@ -300,15 +299,98 @@ impl Future for EndpointDriver {
...
+                None => endpoint
+                    .base_timer
+                    .insert(endpoint.runtime.new_timer(deadline)),
+            };
+            if let Poll::Ready(()) = timer.as_mut().poll(cx) {
+                keep_going = true;
+            }
+        }
+
+        let mut dirty = Vec::new();
```
Matthias247, line 340: could be a member of the endpoint to avoid allocating it in every iteration. Same for `drained`, but that one likely matters less since connection disposure is rare.

quinn/src/endpoint.rs
```diff
@@ -300,15 +299,98 @@ impl Future for EndpointDriver {
...
+                    x
+                }
+                None => endpoint
+                    .base_timer
+                    .insert(endpoint.runtime.new_timer(deadline)),
+            };
+            if let Poll::Ready(()) = timer.as_mut().poll(cx) {
+                keep_going = true;
+            }
+        }
```
Matthias247, line 338: worthwhile moving this into a `drive_timers` function?

quinn/src/delay_queue.rs
Ralith: Right. `poll` will advance `next_tick` to either the next timer's expiry or the passed-in `now`, whichever is earlier.

quinn/src/delay_queue.rs
Ralith: They were just to express that it's a constant, but I was ambivalent. Reformatted as advised.

quinn/src/delay_queue.rs
Ralith: Right. Added some discussion to the type's documentation.

quinn/src/delay_queue.rs
Ralith: Done.

quinn/src/delay_queue.rs
Ralith: It does not, because if `link` is `timer` then it is replaced in the slot by its successor, and otherwise it should not be replaced. Tweaked naming and added a comment to clarify.

quinn/src/delay_queue.rs
Ralith: Good thought. Added logic to shrink if the number of timers is less than 1/16th the capacity.

quinn/src/delay_queue.rs
Ralith: Yep. Added some discussion.

quinn/src/delay_queue.rs
Ralith: I'm not sure fairness is the most important concern here. If the system is severely overloaded, LIFO processing allows for some, rather than no, timeouts to be served in a timely manner, which can help reduce cascading failures.

Matthias247: That's true. But it can have tricky side effects: E.g. let's say one only prioritizes new connections if the system is highly loaded. Old connections then would time out - but if they are not handled at all the stack won't be able to finalize them and reclaim their resources - which then again means the system might not be able to get out of the high load situation.

Now this problem might not really occur due to how quinn uses the timer: As far as I see in the endpoint code it will handle all expired timers anyway, just in LIFO order.

Ralith: Hmm, good point, we currently drain the whole expired set regardless. We might want to change that in the future for the same reasons we've done so in the send/receive loops, though.

> if they are not handled at all the stack won't be able to finalize them and reclaim their resources

Also an interesting point; I guess I remain ambivalent. One nice thing about LIFO is it doesn't need an extra index, at least.

quinn/src/endpoint.rs
Ralith: That does seem a little cleaner, especially since it deduplicates the equivalent send for incoming connections.

quinn/src/endpoint.rs
Ralith: Done.

quinn/src/endpoint.rs
Ralith: Good thought! Done.

quinn/src/delay_queue.rs
Ralith: Added a comment to `drive_timers` describing this issue so we'll see it if we make a relevant change, and leaving in LIFO for now for simplicity.

Ralith:
Rebased and addressed outstanding feedback. Note outstanding report of a performance degradation at https://github.com/libp2p/rust-libp2p/pull/2801#issuecomment-1244567213, to be investigated.

Matthias247:
> Rebased and addressed outstanding feedback. Note outstanding report of a performance degradation at https://github.com/libp2p/rust-libp2p/pull/2801#issuecomment-1244567213, to be investigated.

This is about efficiency, not about performance/throughput - right? There's no loss in that setup? I think the evaluation for efficiency should be made on the benchmarks in the quinn repository. Everything else just adds code on top of it which makes it harder to understand what actually causes the degradation.

Ralith:
I understood that post to be reporting decreased throughput when CPU-bound: 2412 Mbps without this branch, and 1913.57 Mbps with it. Certainly a minimal standalone repro would be more actionable.

Matthias247:
Maybe it's because they use async-std in the benchmark which is multithreaded, and then crypto runs in a different thread than IO.  In combination with them picking an order of magnitude worse performing cipher suite (ChaCha20) for an unknown reason, which makes the now combined IO/crypto thread doing more work.

Matthias247:
When running this with uploads from multiple clients, I get a panic in the delay queue:

```
 RUST_BACKTRACE=1 cargo run --release -- -c 24
 --download-size 0 --upload-size 100M --stats

    Finished release [optimized + debuginfo] target(s) in 1.55s
     Running `/mnt/c/Users/matth/Code/rust/quinn/target/release/bulk -c 24 --download-size 0 --upload-size 100M --stats`

thread '<unnamed>' panicked at 'invalid key', quinn/src/delay_queue.rs:84:21
stack backtrace:
   0: std::panicking::begin_panic
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/std/src/panicking.rs:616:12
   1: quinn::delay_queue::DelayQueue<T>::poll
   2: quinn::endpoint::EndpointInner::drive_timers
             at /mnt/c/Users/matth/Code/rust/quinn/quinn/src/endpoint.rs:499:39
   3: <quinn::endpoint::EndpointDriver as core::future::future::Future>::poll
             at /mnt/c/Users/matth/Code/rust/quinn/quinn/src/endpoint.rs:307:23
```

Matthias247:
Debug build panics at a different place:

```
thread '<unnamed>' panicked at 'invalid key', quinn/src/delay_queue.rs:235:13
stack backtrace:
   0: std::panicking::begin_panic
             at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/std/src/panicking.rs:616:12
   1: <slab::Slab<T> as core::ops::index::IndexMut<usize>>::index_mut
             at /home/matthias/.cargo/registry/src/github.com-1ecc6299db9ec823/slab-0.4.7/src/lib.rs:1192:18
   2: quinn::delay_queue::DelayQueue<T>::list_unlink
             at /mnt/c/Users/matth/Code/rust/quinn/quinn/src/delay_queue.rs:235:13
   3: quinn::delay_queue::DelayQueue<T>::unlink
             at /mnt/c/Users/matth/Code/rust/quinn/quinn/src/delay_queue.rs:224:9
   4: quinn::delay_queue::DelayQueue<T>::reset
             at /mnt/c/Users/matth/Code/rust/quinn/quinn/src/delay_queue.rs:189:9
   5: <quinn::endpoint::EndpointDriver as core::future::future::Future>::poll
             at /mnt/c/Users/matth/Code/rust/quinn/quinn/src/endpoint.rs:331:29
```

Matthias247:
Due to the panics on all multi-connection tests, I only got benchark results for a single connection. Looks like an improvement however!

### Main

Throughput: 580 MB/s
RTT: Client 180 us, Server 170 us
CWND Server: 1103736476 (extremely ridiculous)

### Unify Drivers

Throughput: 610MB/s
RTT: Client 240us, Server 440us
CWND Server: 51262370

djc:
So throughput improves 5%, but latency is 33%-160% worse?

Ralith:
I wouldn't read too heavily into μs variations in latency. That's probably either noise or inconsequential constant variation.

Good catch on the multi-connection failures, will investigate.

Ralith:
`reset` was corrupting the intrusive list. Fixed.

Matthias247:
> I wouldn't read too heavily into μs variations in latency.

Agreed. It seems to switch between runs in the 100-500us range, and that's all fine. The multi-millisecond delays that occur in other tests are the more concerning ones, since they are symptoms of excessive buffering.

I tried now again with multiple connections, and it seem to work fine. Did a 24 connection (1 connection per core) upload and download. Results are pretty much equivalent between `main` and `unify-drivers`, which are both around:

#### 24 connections

Throughput: Each Stream 13.37 MB/s. Total: 316 MB/s
RTT: Client 72ms, Server 60ms
CWND Server: 50651475

#### 24 connections upload

Throughput: Each Stream 10.10 MB/s. Total: 242MB/s
RTT: Client: 390us, Server 3.3ms
CWND Client: 304000

Based on that I think merging is fine. And as a follow-up it could be investigated how to avoid the large RTT under heavy server load (I assume it's due to the connections all producing packets for the endpoint, and that one not being able to keep up with sending).

Ralith:
Note that the `bulk` benchmark uses a single thread for the server. If I use a multithreaded runtime for the server (but retain one thread per client) and use cores / 2 clients (so one core is available for each half of a connection) I see a ~15% speedup on `main` (presumably from parallel crypto), and a ~15% *slowdown* on this branch (presumably from inter-thread comms overhead).

On the other hand, in a hypothetical configuration with a totally discrete server endpoint per core, we'd almost certainly see a linear speedup, on top of the baseline efficiency and consistency improvements on this branch. I think that's the case that makes sense to optimize for, since it scales better in large-scale use which must distribute load across machines anyway, and for very small-scale use like game servers where most CPU is being spent elsewhere, but it's not a 100% clean win, especially for naive/mid-scale users that won't have load balancing as a matter of course.

Ralith:
If you discount delay_queue.rs (which I think is mostly fair, since it's strongly encapsulated), this is net -119 LoC.

quinn/src/delay_queue.rs
```diff
@@ -0,0 +1,521 @@
...
+/// Time is expressed as a bare u64 representing an absolute point in time. The caller may use any
+/// consistent unit, e.g. milliseconds, and any consistent definition of time zero. Larger units
+/// limit resolution but make `poll`ing over the same real time interval proportionately faster,
+/// whereas smaller units improve resolution, limit total range, and reduce `poll` performance.
+#[derive(Debug)]
+pub struct DelayQueue<T> {
+    /// Definitions of each active timer
+    timers: Slab<TimerState<T>>,
+
+    /// Hierarchical timer wheel
```
djc, line 16: So I need a bunch more/different documentation here. I've not encountered this data structure before, and I feel the documentation goes straight into a bunch details while skipping some of the big picture of how this data structure is supposed to work. One thing that seems mostly implicit here is the use of bits from the timer values as entries into the levels (which I only accidentally somewhat inferred from the `next_tick` docstring).

The structure with the `timers` field acting as a value map with `levels` acting as a hierarchical lookup table is also not explained very much here.

Is it fair to say there's one `slot` per bit in the `u64`, except organized in levels to make it a little faster to make the next one in case the slots are sparse? Maybe the documentation could show the expected organization in one example case with say two near and one far timer.

Ralith: I've expanded the documentation here substantially and added a diagram. Is it clearer now?

> Is it fair to say there's one slot per bit in the u64, except organized in levels to make it a little faster to make the next one in case the slots are sparse?

No, it is more or less coincidental that the bit width of a time is the same as the number of slots in a level. I'm cribbing tokio's guesstimate of an efficient level size here, and I suspect they're motivated by the use of a bitfield to represent slot occupancy to accelerate scans, which I'm leaving for future work here. The data structure will work for any value of `LOG_2_SLOTS`.

quinn/src/delay_queue.rs
```diff
@@ -0,0 +1,521 @@
...
+            // If we can't advance any further, bail out
+            if self.next_tick >= now {
+                return None;
+            }
+        }
+    }
+
+    /// Find a timer expired by `now` in level 0
+    fn scan_bottom(&mut self, now: u64) -> Option<T> {
+        for slot in &mut self.levels[0].slots[range_in_level(0, self.next_tick..=now)] {
```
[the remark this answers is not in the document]
Ralith: That's cheeky. I like it.

quinn/Cargo.toml
```diff
@@ -61,6 +61,7 @@ tokio = { version = "1.0.1", features = ["rt", "rt-multi-thread", "time", "macro
 tracing-subscriber = { version = "0.3.0", default-features = false, features = ["env-filter", "fmt", "ansi", "time", "local-time"] }
 tracing-futures = { version = "0.2.0", default-features = false, features = ["std-future"] }
 url = "2"
+proptest = "1"
```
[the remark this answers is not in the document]
Ralith: This list is a bit disordered to begin with, but I'll avoid making it worse.

quinn/src/endpoint.rs
```diff
@@ -518,8 +509,7 @@ impl EndpointInner {
 
 #[derive(Debug)]
 struct ConnectionSet {
-    /// Senders for communicating with the endpoint's connections
-    senders: FxHashMap<ConnectionHandle, mpsc::UnboundedSender<ConnectionEvent>>,
+    connections: FxHashMap<ConnectionHandle, ConnectionRef>,
```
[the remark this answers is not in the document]
Ralith: Went with `refs`. Hooray for `lsp-rename`!

quinn/src/endpoint.rs
```diff
@@ -598,33 +628,26 @@ impl Drop for Incoming {
...
+    pub(crate) fn new(socket: Box<dyn AsyncUdpSocket>, inner: proto::Endpoint, ipv6: bool) -> Self {
         let udp_state = Arc::new(UdpState::new());
         let recv_buf = vec![
             0;
             inner.config().get_max_udp_payload_size().min(64 * 1024) as usize
                 * udp_state.gro_segments()
                 * BATCH_SIZE
         ];
-        let (sender, events) = mpsc::unbounded_channel();
+        let (dirty_send, dirty) = mpsc::unbounded_channel();
```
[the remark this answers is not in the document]
Ralith: Renamed and documented.

quinn/src/endpoint.rs
```diff
@@ -299,15 +297,81 @@ impl Future for EndpointDriver {
...
-        let mut endpoint = self.0.lock().unwrap();
+        let mut endpoint = &mut *self.0.lock().unwrap();
         if endpoint.driver.is_none() {
             endpoint.driver = Some(cx.waker().clone());
         }
 
-        let now = Instant::now();
-        let mut keep_going = false;
-        keep_going |= endpoint.drive_recv(cx, now)?;
-        keep_going |= endpoint.handle_events(cx);
```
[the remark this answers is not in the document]
Ralith: Yep. Not sure it's even useful to inline it, so I didn't.

quinn/src/connection.rs
```diff
@@ -751,8 +704,6 @@ impl ConnectionRef {
             closed: Arc::new(Notify::new()),
             error: None,
             ref_count: 0,
-            udp_state,
```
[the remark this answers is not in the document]
Ralith: I've split out the `udp_state` refactor; good call there. Staring at this for a bit I'm not sure there's any other low hanging fruit to decompose. The heart of this commit is folding the responsibilities of `ConnectionDriver` into `EndpointDriver`. This requires restructuring how connections are woken (via the new channel) and switching from timer-per-connection to a `DelayQueue` (since we no longer have a discrete task to wake for each connection). Because timeouts can wake connections, and woken connections can adjust timers, I'm not sure these can be easily decoupled.

quinn/src/connection.rs
```diff
@@ -771,16 +766,17 @@ pub struct ConnectionInner {
...
     /// Number of live handles that can be used to initiate or handle I/O; excludes the driver
     ref_count: usize,
-    udp_state: Arc<UdpState>,
 }
 
 impl ConnectionInner {
-    pub(crate) fn drive_transmit(&mut self, out: &mut VecDeque<proto::Transmit>) -> bool {
+    pub(crate) fn drive_transmit(
+        &mut self,
+        max_datagrams: usize,
```
djc, line 771: I think the argument order should be the other way around here?

quinn/src/delay_queue.rs
```diff
@@ -0,0 +1,584 @@
...
+        debug_assert!(
+            self.levels[level].slots[..slot].iter().all(Option::is_none),
+            "lower slots in this level are empty"
+        );
+
+        // Advance into the slot
+        self.next_tick = slot_start(self.next_tick, level, slot);
+
+        // Drain it into the lower levels, if possible
+        if level > 0 {
```
djc, line 192: Invert condition + early return?

quinn/src/delay_queue.rs
```diff
@@ -0,0 +1,584 @@
...
+        Self {
+            slots: [None; SLOTS],
+        }
+    }
+}
+
+impl fmt::Debug for Level {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut m = f.debug_map();
+        for (i, Timer(t)) in self
```
djc, line 374: How do you feel about naming the scrutinee (maybe just `iter`) to avoid the multi-line scrutinee -- which makes the block header harder to follow?

quinn/src/endpoint.rs
```diff
@@ -637,7 +669,12 @@ impl Drop for Incoming {
 pub(crate) struct EndpointRef(Arc<Mutex<EndpointInner>>);
 
 impl EndpointRef {
-    pub(crate) fn new(socket: Box<dyn AsyncUdpSocket>, inner: proto::Endpoint, ipv6: bool) -> Self {
+    pub(crate) fn new(
+        runtime: Arc<dyn Runtime>,
```
djc, line 673: Should the `runtime` argument come last?

quinn/src/delay_queue.rs
djc: > I've expanded the documentation here substantially and added a diagram. Is it clearer now?

It's much better! One nit with it is that the shift from base-10 to base-2 is a little misleading (and it might be easy to miss the shift towards the end of the docstring). Maybe it would be better to make the diagram base-2 as well, maybe in the `u8` domain or something?

quinn/src/connection.rs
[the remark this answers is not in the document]
djc: So the new channel is what you've called `dirty`, right? That's kind of where I struggle, making sense of what's going dirty and how the mechanics for that could make sense. Do you think the documentation/naming could be improved to clarify how this stuff fits together? Like, `dirty` is a nice short name but it doesn't say much about what that state applies to and/or when it's cleared/how it should it should be cleared.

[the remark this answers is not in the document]
Ralith: > Do you think the documentation/naming could be improved to clarify how this stuff fits together?

I've added additional comments on this subject in several places, including `EndpointInner::dirty_recv`, `ConnectionInner::dirty`, and `Endpoint::drive_connections` (formerly `handle_events`). I do think "dirty" is well precedented for describing the pattern of tracking a subset of components that need to be updated, but better documentation never hurts.

> Sorry to be a drag on this review

To the contrary, your feedback has been plenty timely, especially on the time scale that I've been chipping away at this! Maybe you didn't notice but addressing your last round of comments was spread over a week or so :D

If anything, a reduced-context perspective is helpful, since we do want the architecture to be intelligible to others.

quinn/src/delay_queue.rs
Ralith: Moved that information to the front and filled it out into a better introductory paragraph. I like using base 10 because it's particularly easy to read/visualize, and levels with only 2 slots each might be a bit confusing, so hopefully this makes the compromise more obvious.

quinn/src/delay_queue.rs
Ralith: Sure, that does seem a bit clearer.

quinn/src/delay_queue.rs
Ralith: Good catch!

Ralith:
Since it's neat and might be useful in other projects, I've split delay_queue.rs from this out into a separate crate, [timer-queue](https://crates.io/crates/timer-queue). Should we depend on that rather than having a separate copy? If so, should we move maintenance of it under quinn-rs?

Ralith:
Rebased; friendly bump.

Ralith:
Dropped a superfluous refactoring commit.

djc:
Sorry for the slow response to this -- currently trying to soak up rustls review bandwidth while it's available.

quinn/src/endpoint.rs
```diff
@@ -244,12 +246,9 @@ impl Endpoint {
         let reason = Bytes::copy_from_slice(reason);
         let mut endpoint = self.inner.state.lock().unwrap();
         endpoint.connections.close = Some((error_code, reason.clone()));
-        for sender in endpoint.connections.senders.values() {
-            // Ignoring errors from dropped connections
-            let _ = sender.send(ConnectionEvent::Close {
-                error_code,
-                reason: reason.clone(),
-            });
+        for conn in endpoint.connections.refs.values() {
```
Matthias247, line 248: Any chance this is racy - so that connections that have been slightly after this don't get the close signal? I guess not, because it all goes through the endpoint mutex. And the path to open a connection hopefully checks `endpoints.connection.close`?

quinn/src/endpoint.rs
```diff
@@ -244,12 +246,9 @@ impl Endpoint {
...
+            let mut state = conn.state.lock("close");
```
Matthias247, line 249: Is the endpoint mutex always taken before the connection mutex in case both are used? This will be required to avoid deadlocks.

Matthias247 (dismissed):
I haven't looked at the delay queue again in depth since the last time. But everything else looks totally reasonable to me.

I would be inclined to see this landing :-)

quinn/src/endpoint.rs
Ralith: Yep; `Accept::poll` locks the same mutex, and attempts to connect will go through `ConnectionSet::insert` which is inside the same mutex and immediately `close`s the connection that it adds if `self.close` is set.

quinn/src/endpoint.rs
Ralith: When we're traversing connections via the endpoint, that's statically enforced, since the `ConnectionSet` is inside the `Endpoint` state mutex. Connection state is also locked independently via the public `Connection` API, but that shouldn't overlap with endpoint locks.

Things might get fishy if a caller captures and manipulates an `Endpoint` handle inside an extension trait object fed to that same endpoint, but I think even those will generally be invoked inside the endpoint mutex if any, and hence gracefully panic.

Ralith:
> prioritizing soaking up rustls review bandwidth while it's available.

Makes sense; that's definitely a precious resource! I don't think this is blocking anything immediate; @rrichardson's interest in horizontal scaling as discussed in #914 can be pursued independently at the proto layer.

mxinden:
:wave: [rust-libp2p](https://github.com/libp2p/rust-libp2p) maintainer here,

Is there any way I could help move this pull request forward?

Ways that come to my mind:

- More reviews
- Testing in production environments
- Writing tests
- Providing benchmark results

My motivation:

- We are currently using `quinn-proto` [in `rust-libp2p`](https://github.com/libp2p/rust-libp2p/tree/master/transports/quic).
- We don't use `quinn` directly, but instead our own wrapper around `quinn-proto` due to `quinn` using unbounded channels between the _Endpoint_ and the _Connection_. See https://github.com/quinn-rs/quinn/issues/1353.
- I don't think it makes sense for us (rust-libp2p) to maintain our own wrapper long term. It introduces unnecessary maintenance work and we don't benefit from upstream (`quinn`) improvements and fixes.
- This pull request would remove the unbounded channels, thus enabling us to use `quinn` directly.

On a general note, thank you for all your work! With the release of QUIC support in rust-libp2p via `quinn-proto`:

- We are seeing significant connection establishment improvements in the 95th percentile compared to TCP:

    image
    
    See https://kademlia-exporter.max-inden.de/d/Pfr0Fj6Mk/rust-libp2p for more numbers.

- We can leverage stream creation backpressure at the protocol level.
- We see significant improvements to our hole punching success rate.
- ...

Ralith:
Thanks for sharing those numbers, that's really cool! I think knowing there's demand is helpful already, but thoughtful review of the core [294e5d7](https://github.com/quinn-rs/quinn/pull/1219/commits/294e5d7626af0e38704d2b79b7ccea0f6d1db129) commit that I think @djc is mainly hung up on certainly wouldn't hurt.

DemiMarie:
@mxinden @Ralith This does not remove all unbounded channels.  What would it take to remove the rest of them?

djc:
Is there any evidence that this particular usage of unbounded channels is bad? If you are so worried about this, feel free to submit a PR and we'll consider it.

Ralith:
The remaining "unbounded" channel introduced in this PR, in fact, bounded by the number of concurrently active connections. It could be replaced with various other sorts of queue to make that more obvious (an intrusive list might be cool), but the behavior would be the same.

mxinden:
> @mxinden @Ralith This does not remove all unbounded channels. What would it take to remove the rest of them?

@DemiMarie are you referring to the unbounded channel used by a connection to wake up the endpoint? 

``` rust
    /// `handle` is sent here when `wake` is called, prompting the endpoint to drive the connection
    dirty: mpsc::UnboundedSender<ConnectionHandle>,
```

https://github.com/quinn-rs/quinn/pull/1219/files#diff-60e27999f6922becedd55cb952eb6a67db93ed5d136f2e22451d0ce83a73b89bR779-R780

I don't think these pose any issues. As @Ralith says above, they are bounded by the number of connections. I consider the usage a neat pattern of _many-to-wake-one_, transporting the id along the way.

> Is there any evidence that this particular usage of unbounded channels is bad?

In case I am not missing something and indeed we are talking about the `dirty` unbounded channel, I don't see a downside to the usage.

In case one would want to ban the usage of unbounded channels across the codebase, one could also uses a bounded channel. Given that the sender is cloned for each connection, there is a guaranteed slot per sender, thus always capacity to send, no `.await` needed.

> The channel’s capacity is equal to `buffer + num-senders`.

https://docs.rs/futures/latest/futures/channel/mpsc/fn.channel.html

thomaseizinger:
> In case one would want to ban the usage of unbounded channels across the codebase

FWIW, clippy can help with this using its [`disallowed_methods`](https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_methods) lint:

```toml
disallowed-methods = [
  { path = "futures::channel::mpsc::unbounded", reason = "does not enforce backpressure" },
]
```

mxinden (commented):
Small review with small suggestion and question for better understanding.

djc:
@mxinden thanks for the careful review!

quinn/src/connection.rs
```diff
@@ -745,33 +689,29 @@ impl ConnectionRef {
...
     ) -> Self {
+        let _ = dirty.send(handle);
         Self(Arc::new(ConnectionInner {
             state: Mutex::new(State {
                 inner: conn,
-                driver: None,
                 handle,
+                span: debug_span!("connection", id = handle.0),
+                is_dirty: false,
+                dirty,
```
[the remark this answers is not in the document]
Ralith: Good catch; IIRC this was an incomplete bugfix.

quinn/src/endpoint.rs
```diff
@@ -347,22 +336,32 @@ pub(crate) struct EndpointInner {
...
-    runtime: Arc<dyn Runtime>,
+    /// Connections add themselves to this queue when they need to be driven
+    ///
+    /// Occurs e.g. due to application-layer activity
+    dirty_recv: mpsc::UnboundedReceiver<ConnectionHandle>,
+    /// Passed in to connections to enable the above
+    dirty_send: mpsc::UnboundedSender<ConnectionHandle>,
+    timers: DelayQueue<ConnectionHandle>,
+    /// Temporary buffer for connections dirty before a poll pass
+    dirty_buffer: Vec<ConnectionHandle>,
```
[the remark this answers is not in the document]
Ralith: Correct; this was a premature optimization. Removed for clarity.

quinn/src/endpoint.rs
```diff
@@ -485,74 +518,97 @@ impl State {
...
+    fn drive_connections(&mut self, cx: &mut Context, shared: &Shared) -> bool {
+        let mut keep_going = false;
+
+        keep_going |= self.drive_timers(cx, Instant::now());
+
+        // Buffer the list of initially dirty connections, guaranteeing that the connection
+        // processing loop below takes a predictable amount of time.
+        while let Poll::Ready(Some(conn_handle)) = self.dirty_recv.poll_recv(cx) {
+            self.dirty_buffer.push(conn_handle);
+        }
```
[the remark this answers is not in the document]
Ralith: Yep, exactly. This initial loop is safe because it doesn't reset `is_dirty`.

quinn/src/endpoint.rs
```diff
@@ -485,74 +518,97 @@ impl State {
...
+
+        let max_datagrams = self.udp_state.max_gso_segments();
+        let mut drained = Vec::new();
+        for conn_handle in self.dirty_buffer.drain(..) {
```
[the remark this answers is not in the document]
Ralith: Interesting thought. Favoring the simpler approach for now; if we want to further limit the duration of this loop, we can do so by exiting early while populating `dirty_buffer`.

Ralith:
Tweaked per @mxinden's feedback (thanks!) and rebased.

Arqu:
Hey 👋 

We're using quinn in [iroh](https://github.com/n0-computer/iroh). I've run some quick benchmarks with a branch propped up to use this PR. I've noticed a significant drop in throughput (40-50%) when serving data from one node to 2 or more clients. One to one transfers don't seem significantly impacted yet (most probably because other parts of our code don't saturate the pipe). 

From my observations with the PR the CPU on my test rig seems to peg to around 220% ie 2.2 cores and more clients just keep diluting the throughput between them. The same test when running against `0.9.3` seems to be able to utilize the CPU for as much as it is requested for and scaling ~linearly with clients. Seeing some 700-900% in the 1 to 10 test cases.

Didn't dive too deep into it yet, but my initial guess would be that we have either some mutex locking or polling happening that makes some run loop starve itself. 

Let me know if there's anything specific you would like me to poke at or test out.

Ralith:
Closing as this is quite stale and understood to be a suboptimal direction due to reduced opportunity for work stealing and parallelism.

PR #1222, opened 2021-11-02
Cache dependencies for CI

Ralith:
https://github.com/marketplace/actions/rust-cache

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -32,6 +32,7 @@ jobs:
           profile: minimal
           toolchain: ${{ matrix.rust }}
           override: true
+      - uses: Swatinem/rust-cache@v1
       - uses: actions-rs/cargo@v1
         with:
           command: build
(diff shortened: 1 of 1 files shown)


djc (dismissed):
Seems okay. Have you verified the impact?

djc:
Looking at the action's docs, it seems to discuss using the top-level `Cargo.toml` and `Cargo.lock` in the cache key. We don't have a `Cargo.lock` but we actually only reference dependencies from the crates' `Cargo.toml` files. I wonder if the cache key works correctly in this particular situation?

Ralith:
Comparing the two commits in this PR, it looks like it's consistently saving about a minute in the build stage, and indeed not recompiling most dependencies. I think the cache key will be fine so long as we're usually running CI on the same branch, since it looks like it gets rewritten every run anyway; backports PRs might trash it, but those should be infrequent, and could be mitigated if there's a way we can add the branch to the cache key.

djc (approved):

PR #1223, opened 2021-11-03
Persistent congestion fixes

Ralith:
Minor correctness tweaks that should allow faster recovery and reduce spurious persistent congestion determinations.


djc (approved):
Nice work!

PR #1232, opened 2021-11-06
Backwards-compatibility with late draft versions

Ralith:
Inspired by conversation with @Dessix. Allows the upcoming Quinn release to remain backwards-compatible with 0.7.x.

### DIFF
--- a/quinn-proto/src/crypto.rs
+++ b/quinn-proto/src/crypto.rs
@@ -121,15 +121,6 @@ pub trait ClientConfig: Send + Sync {
 
 /// Server-side configuration for the crypto protocol
 pub trait ServerConfig: Send + Sync {
-    /// Start a server session with this configuration
-    ///
-    /// Never called if `initial_keys` rejected `version`.
-    fn start_session(
-        self: Arc<Self>,
-        version: u32,
-        params: &TransportParameters,
-    ) -> Box<dyn Session>;
-
     /// Create the initial set of keys given the client's initial destination ConnectionId
     fn initial_keys(
         &self,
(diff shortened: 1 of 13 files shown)


Ralith:
This does not include support for version negotiation, as that's still a draft; see https://datatracker.ietf.org/doc/draft-ietf-quic-version-negotiation/.

quinn-proto/src/crypto.rs
```diff
@@ -113,6 +113,7 @@ pub trait ClientConfig: Send + Sync {
     /// Start a client session with this configuration
     fn start_session(
         self: Arc<Self>,
+        version: u32,
```
djc, line 116: Why go with `u32` here rather than an enum like we do in rustls? Would be nice to avoid picking an unsupported version at run-time, while avoiding the fallibility of `interpret_version()`?

djc (commented):
Good stuff! One question about the chosen interface/design.

quinn-proto/src/crypto.rs
Ralith: e.g. quinn-noise is using a distinct range of versions compared to standard QUIC (0xf0f0f2f[0-f], per https://github.com/quicwg/base-drafts/wiki/QUIC-Versions). We could include quinn-noise versions in the enum, but that's at odds with the crypto layer being an open extension interface, requiring each new addition to go through us.

Maybe we should make a bunch of the methods fallible with an `InvalidVersion` error?

djc: Yeah, I would prefer fallibillity to bubble up, though it probably gets a bit ugly. We can only check when we bring the `dyn ClientConfig` or `dyn ServerConfig` together with the `TransportConfig`, so in `Endpoint::new()` for the server (currently infallible) or `Endpoint::connect()` for the client (yields `Result<_, ConnectError>`).

Ralith: Updated. It's actually pretty tidy, if we bake in some ordering assumptions.

> in Endpoint::new() for the server

I don't think this would work well. We only create initial keys once we have an initial dest CID from a client, so there's no robust way to validate the endpoint's supported versions list against the crypto trait. Fortunately, if the user gets this wrong then it's fairly harmless: incompatible incoming connections that pass the supported_versions test are ignored, and attempts to connect generate a `ConnectError`.

djc (approved):

PR #1236, opened 2021-11-10
Avoid unwrapping in Endpoint

djc:
Some followup to #1232.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -624,25 +624,19 @@ impl Endpoint {
 
         let server_config = server_config.clone();
         let loc_cid = self.new_cid();
-        let params = TransportParameters::new(
+        let mut params = TransportParameters::new(
             &server_config.transport,
             &self.config,
             self.local_cid_generator.as_ref(),
             loc_cid,
             Some(&server_config),
         );
-        let server_params = TransportParameters {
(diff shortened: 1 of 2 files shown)


quinn-proto/src/endpoint.rs
```diff
@@ -618,15 +617,7 @@ impl Endpoint {
...
-                    self.initial_close(
-                        version,
-                        remote,
-                        local_ip,
-                        crypto,
-                        &src_cid,
-                        &temp_loc_cid,
-                        e,
-                    );
+                    self.initial_close(version, remote, local_ip, crypto, &src_cid, &loc_cid, e);
```
djc, line 620: In this case we end up with a connection with the `loc_cid` but we also send an initial close with that `loc_cid`. That might not be what we want?

Ralith: The `handle_event` call above cleans up the connection. Perhaps this could be refactored not to add it in the first place, but the existing behavior looks correct to me.

quinn-proto/src/endpoint.rs
```diff
@@ -481,7 +481,7 @@ impl Endpoint {
         }
 
         // Local CID used for stateless packets
```
Ralith, line 483: Outdated comment.

Ralith (changes requested):
Awesome cleanup!

Ralith (approved):

PR #1238, opened 2021-11-14
Bump version numbers for 0.8.0 release

djc:


Ralith (changes requested):
`quinn` needs its `quinn-udp` dependency version number updated too, right?

djc:
Good catch. Cargo [stupidly](https://github.com/rust-lang/cargo/issues/2222) considers pre-releases semver compatible with their eventual final release, so I guess that's why it didn't report on that.

Ralith (approved):

PR #1241, opened 2021-11-15
Adjust link to quinn crate entry point in docs

themaxdavitt:
The link to `Endpoint` in [the documentation for the new release](https://docs.rs/quinn/0.8.0/quinn/) (in the sentence "The entry point of this crate is the `Endpoint`." at the top of the page) points into the no longer existent "generic" directory. This fixes that link.

### DIFF
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -6,7 +6,7 @@
 //! builds on top of quinn-proto, which implements protocol logic independent of any particular
 //! runtime.
 //!
-//! The entry point of this crate is the [`Endpoint`](struct.Endpoint.html).
+//! The entry point of this crate is the [`Endpoint`].
 //!
 //! # About QUIC
 //!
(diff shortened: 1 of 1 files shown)


quinn/src/lib.rs
```diff
@@ -6,7 +6,7 @@
 //! builds on top of quinn-proto, which implements protocol logic independent of any particular
 //! runtime.
 //!
-//! The entry point of this crate is the [`Endpoint`](generic/struct.Endpoint.html).
+//! The entry point of this crate is the [`Endpoint`](struct.Endpoint.html).
```
djc, line 9: This is crate! Presumably now we could do with just `` [`Endpoint`] ``, want to try that?

themaxdavitt: You're right! Just changed it to that.

Ralith (approved):
Thanks!

Issue #1242, opened 2021-11-18
Controller Trait is not externally implementable anymore


BiagioFesta:
With the new release `0.8.0`, the trait [Controller](https://docs.rs/quinn-proto/0.8.0/quinn_proto/congestion/trait.Controller.html#) cannot be implemented (e.g. a custom congestion controller) because the method [on_ack](https://docs.rs/quinn-proto/0.8.0/quinn_proto/congestion/trait.Controller.html#method.on_ack) requires `RttEstimator` which seems to not be pub re-exported.

Or am I missing something? (in that case I apologize for my laziness in not had a deeper check and I am going to immediately close this issue).

The trait and the library design with `dyn` polymorphism seems to encourage the possibility to implement a custom cc.
Considering this was allowed before and I don't find any relevant in the release notes.

Waiting for feedback. If needed, I will be glad to open a small PR to fix this.

djc:
Yes, I think this is just an oversight and we should make `RttEstimator` public.

Ralith:
Thanks for catching this; a PR would be very welcome!

Ralith:
Fixed by https://github.com/quinn-rs/quinn/pull/1243.

PR #1243, opened 2021-11-18
Make pub visibility for RttEstimator used by pub trait quinn::Congestion

BiagioFesta:
Attempt to fix this: #1242 

---

*Feedback on documentation are more than welcome* :smile:


djc (approved):
Looks great, thanks!

PR #1245, opened 2021-11-20
Only send ACKs when they've changed

Ralith:
Saves a great deal of redundancy when transmitting bulk data, and gives the receiver less work to do. If an ACK is lost, we'll necessarily retransmit it, because packets are only deemed lost when an ACK is received from the peer which does not cover them.

Reduces the number of ACKs sent by the server in the bulk download benchmark from e.g. 930818 to 4228.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -2218,6 +2218,7 @@ impl Connection {
     ) -> Result<(), TransportError> {
         debug_assert_ne!(packet.header.space(), SpaceId::Data);
         let payload_len = packet.payload.len();
+        let mut ack_eliciting = false;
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -2227,15 +2228,8 @@ impl Connection {
             self.stats.frame_rx.record(&frame);
 
             let _guard = span.as_ref().map(|x| x.enter());
-            // Check for ack-eliciting frames
```
djc, line 2230: This group of changes could technically be a commit on its own, right?

djc (approved):

quinn-proto/src/connection/mod.rs
Ralith: Isn't it?

PR #1247, opened 2021-11-21
Use config to reject connections rather than extra state

Ralith:
Minor simplification. Note that `ServerConfig` is fairly small and contains no interior mutability, so cloning it is cheap and not a significant change in behavior.


djc (approved):
Makes sense, thanks!

PR #1248, opened 2021-11-25
Add additional metrics for lost packets and bytes

Matthias247:
These are a bit more explicit than just tracking congestion events, since a
congestion event can cover multiple packets.
The total amount of sent packets is now also logged, in order for applications
to calculate the relative amount of packet loss. The amount of packets can
be different from the amount of datagrams in case of coalesced packets.


Matthias247:
audit failure is unrelated to this PR

Ralith (approved):

PR #1250, opened 2021-11-27
WIP: Simple connect function

Ralith:


quinn/src/connect.rs
```diff
@@ -0,0 +1,21 @@
+use std::{
+    io,
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
+};
+
+use crate::{ClientConfig, Connecting, Endpoint};
+
+/// Connect to `server_address`, authenticating it as `server_name`, using a new endpoint
```
djc, line 8: Seems reasonable, though I don't think it should live in its own module. Maybe an associated method for `ClientConfig`?

djc:
@Ralith do you still want to work on this? If not, maybe close it?

Ralith:
In the years since I drafted this I'm still not sure it makes sense. Happy to close for now, trivial to reinvent it later if we want.

PR #1253, opened 2021-11-30
VersionMismatch error now includes the list of supported versions

BiagioFesta:
*Feel free to reject and close this PR without any clarification!
It is just an experiment done in 2 minutes spare time.*

The idea is to add a little bit of context to the `ConnectionError::VersionMismatch` error. That is, including a list of supported versions returned by the server.

I do not know whether this might be useful or not. In theory, it might allow library users to implement some fallback mechanism.

Like that, it would include the "grease version 0x0a1a2a3a/0x0a1a2a4a" which might be omitted.


Ralith:
While I like this for diagnostic purposes...

> In theory, it might allow library users to implement some fallback mechanism.

...unfortunately that would be a major security vulnerability. QUIC isn't lacking a fallback mechanism because it didn't occur to anyone to build one, but rather because unless extreme caution is taken, it enables downgrade attacks. See the version negotiation draft for some discussion of the measures needed to *possibly* mitigate this. The risk that a well-meaning downstream might, given the opportunity, roll their own vulnerable version negotiation is too high.

Maybe we could expose this in a way only visible in the `Display`/`Debug` impls, but I don't know if that's a high enough barrier to prevent abuse.

Ralith:
Closing due to inactivity and unclear motivation; feel free to reopen with further discussion if you have a concrete requirement!

PR #1256, opened 2021-12-10
Fix ALPN misconfiguration in example

Ralith:
Introduced by a copy-paste error in 9daaa15da231f56559d771cf8d674ccdbe98890e


djc:
Oops! Thanks for fixing.

djc (approved):

Issue #1262, opened 2021-12-19
next() method not found in Incoming


euberdeveloper:
Part of my code in the server is:

```rs
let (endpoint, mut incoming) = quinn::Endpoint::server(server_config, listen_address)?;
println!("listening on {}", endpoint.local_addr()?);

while let Some(conn) = incoming.next().await {
        // my code
}
```

I have this compile error when I call `.next()`:

```no method named `next` found for struct `quinn::Incoming` in the current scope```

Is it anything with the version `0.8.0` of Quinn?

euberdeveloper:
Sorry for the mistake, I am new to rust

I just had to use ```use futures_util::stream::stream::StreamExt;``` because the implementation of the trait with `next` for `Incoming` was not on scope

Ralith:
That's correct; no worries! I've actually been thinking of adding an inherent version, to avoid the mandatory (surprisingly heavy) futures_util dep.

wmmc88:
> That's correct; no worries! I've actually been thinking of adding an inherent version, to avoid the mandatory (surprisingly heavy) futures_util dep.

I thought #1263 removes the need for `use futures_util::stream::stream::StreamExt;`, but it seems like its still required?

PR #1263, opened 2021-12-19
Replace futures-util with optional futures-io dep

Ralith:
A trivial amount of additional code lets us save another dep and be a bit more accessible to new users (see e.g. #1262).

### DIFF
--- a/perf/src/bin/perf_client.rs
+++ b/perf/src/bin/perf_client.rs
@@ -6,7 +6,6 @@ use std::{
 
 use anyhow::{Context, Result};
 use bytes::Bytes;
-use futures_util::StreamExt;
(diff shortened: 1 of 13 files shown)


Ralith:
This also improves consistency with tokio primitives, e.g. [`mpsc::Receiver::recv`](https://docs.rs/tokio/1.15.0/tokio/sync/mpsc/struct.Receiver.html#method.recv).

djc:
Should we use futures-io as the optional dep rather than futures-util? Otherwise this looks great!

Ralith:
Good catch, I didn't know that crate existed! Fixed.

djc (approved):

Issue #502, opened 2019-11-18
async_std support


DemiMarie-parity:
`quinn-proto` is extremely low-level and difficult to use.  `quinn` is much nicer to use, but depends on Tokio.  Would it be possible for `quinn` to also support `async_std` via a Cargo feature?

Ralith:
Thanks for the issue! I think the right way to approach this is to make `quinn` abstract over runtime, i.e. the thing providing wake-ups for I/O and timers, as cargo features can't be mutually exclusive. This might also be useful for more pure end-to-end testing. Note that `quinn` implements its own `UdpSocket` to gain access to low-level features like explicit congestion notification flags, and a good solution should avoid duplicating that code.

DemiMarie-parity:
Due to <https://github.com/async-rs/async-std/issues/293> quinn’s `UdpSocket` won’t be usable on `async-std`, so we will need to provide an alternative.  That should be doable, however.

jean-airoldie:
From what i understand you could replace the tokio dependency with smol & futures-io:
* tokio::io::Async{Read,Write} becomes futures_io::Async{Read,Write}
* tokio-rustls becomes async-tls
* PollEvented<mio::net::UdpSocket> becomes smol::Async<std::net::UdpSocket>
* tokio::time becomes either futures-timer or smol::Timer

However you would need to find a non tokio alternative for tokio_util::codec, although it wouldn't be too hard to adapt the code.

I think this should make quinn abstract over runtime.

Ralith:
> From what i understand you could replace the tokio dependency with smol & futures-io:

My understanding is that this amounts to a hidden hard dep on a smol background thread, so we'd just be trading one required runtime for another. Unless I misunderstand how smol works these days? I'd prefer to be genuinely runtime-agnostic as outlined above.

> tokio::io::Async{Read,Write} becomes futures_io::Async{Read,Write}

`quinn` implements both of these traits for streams already; the tokio variants could be feature flagged easily enough.

> However you would need to find a non tokio alternative for tokio_util::codec, although it wouldn't be too hard to adapt the code.

Indeed that's pretty trivial code, which we could easily vendor if need be.

jean-airoldie:
> Unless I misunderstand how smol works these days? I'd prefer to be genuinely runtime-agnostic as outlined above.

Right. The only real benefit is that you can run your quinn sockets in a non-tokio runtime without crashing.

Ralith:
[...]

PR #1264, opened 2021-12-19
Replace Broadcast helper with tokio::sync::Notify 

Ralith:
Drops a chunk of complex and under-audited synchronization code in favor of a similar, better-tested and more efficient primitive from tokio.


Ralith:
One subtler consequence of this is that we're going from returning a named, `'static` future to an anonymous, `self`-borrowing one. I think that's okay, and more forwards-compatible in any case. It does make it more difficult to use with hand-written futures, however.

quinn/src/connection.rs
```diff
@@ -323,10 +325,23 @@ impl Connection {
     /// Streams are cheap and instantaneous to open unless blocked by flow control. As a
     /// consequence, the peer won't be notified that a stream has been opened until the stream is
     /// actually used.
-    pub fn open_uni(&self) -> OpenUni {
-        OpenUni {
-            conn: self.0.clone(),
-            state: broadcast::State::default(),
+    pub async fn open_uni(&self) -> Result<SendStream, ConnectionError> {
```
djc, line 328: These hunks of this commit make it hard to review this because seem to include both changing the approach to futures and the new use of `Notify` over our previous `Broadcast` thing. Do you think it might be feasible to separate two in two steps?

daxpedda:
I was hoping that eventually quinn will become executor independent, especially now that quinn-udp is external. I couldn't find a `Notify` equivalent in futures though. I guess this can also be dealt with later.

See #502.

Ralith:
@daxpedda Not to worry, this would be utterly trivial to replace with [`futures_intrusive::ManualResetEvent`](https://docs.rs/futures-intrusive/0.4.0/futures_intrusive/sync/type.ManualResetEvent.html) or similar should the larger challenges of executor-independence be addressed. The only reason I didn't go with that immediately is to avoid the extra dependency, and due to some lurking [soundness concerns](https://github.com/Matthias247/futures-intrusive/issues/56) in futures-intrusive.

quinn/src/connection.rs
Ralith: Good thought; split out the `async fn` stuff into a separate commit.

Ralith: Felt inspired and split up the refactoring some more as well.

quinn/src/connection.rs
```diff
@@ -324,22 +324,40 @@ impl Connection {
...
+                let mut conn = self.0.lock("open");
+                if let Some(ref e) = conn.error {
+                    return Err(e.clone());
+                }
+                if let Some(id) = conn.inner.streams().open(dir) {
+                    let is_0rtt = conn.inner.side().is_client() && conn.inner.is_handshaking();
+                    return Ok((id, is_0rtt));
+                }
+                opening = conn.stream_opening[dir as usize].clone();
+                opening.notified() // Must be called, but not awaited, while the connection lock is held
```
djc, line 358: After all the refactoring the core of this change is now much clearer but I don't have a good intuition for what we are looping against here and why we need the awkward binding declaration and clone. Maybe add just a sprinkling of comments to communicate the intent here?

Ralith: Added detailed discussion.

djc (approved):

PR #1271, opened 2021-12-28
Remove redundant copying outgoing packets.

TimonPost:
Both `proto::Endpoint.transmits` and and `quinn::Endpoint.outgoing` queue transmit packets, and in `drive send` both queues are fetched and drained while this isn't required (from what I can tell). This PR removes the `outgoing` queue in `quinn::Endpoint` and simplifies it to only use `proto::Endpoint.transmits` for all outgoing packets.

This should remove redundant copying of transmits, and queue allocations. 

_left: before, right: this PR_

image

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -103,7 +103,7 @@ impl Endpoint {
         if self.transmits.len() >= batch {
             &self.transmits.as_slices().0[0..batch-1]
         } else {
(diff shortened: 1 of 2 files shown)


djc:
Probably good to rebase this and make sure it passes the tests first.

TimonPost:
Im closing it temporarily to prevent CI from triggering while I try to fix the tests on unix

PR #1272, opened 2021-12-28
Remove redundant packet copying

TimonPost:
Both `proto::Endpoint.transmits` and and `quinn::Endpoint.outgoing` queue transmit packets, and in `drive send` both queues are fetched and drained while this isn't required (from what I can tell). This PR removes the `outgoing` queue in `quinn::Endpoint` and simplifies it to only use `proto::Endpoint.transmits` for all outgoing packets.

This should remove redundant copying of transmits, and queue allocations. 

_left: before, right: this PR_

image

To make the API consistent, I renamed `poll_transmit` to `deque_transmit`. I could name `queue_transmit` to `push_transmit` to make my function more consistent although I am more satisfied with queue terminology because 1) the transmit storage is a queue 2) poll doesn't really say some dequeue operation is going to happen. 

The Linux CI error also occurs on a fresh clone: run with `cargo test echo_dualstack -- --nocapture` to validate this.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -3,7 +3,7 @@ use std::{
     convert::TryFrom,
     fmt, iter,
     net::{IpAddr, SocketAddr},
-    ops::{Index, IndexMut},
+    ops::{Index, IndexMut, RangeTo},
     sync::Arc,
     time::{Instant, SystemTime},
 };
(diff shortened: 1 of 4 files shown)


quinn-proto/src/endpoint.rs
```diff
@@ -84,10 +84,34 @@ impl Endpoint {
...
+
+    /// Drains a range of transmittable packets
+    pub fn drain_transmits(&mut self, range: RangeTo<usize>) {
+        self.transmits.drain(range);
+    }
+
+    /// Returns a batch of transmittable packets
+    pub fn transmit_batch(&self, batch: usize) -> &[Transmit] {
+        if self.transmits.len() >= batch {
+            &self.transmits.as_slices().0[..batch]
```
Ralith, line 104: This may panic, as there's no guarantee that the first slice of a `VecDeque` contains its full length.

Ralith (commented):
Thanks for the PR! Unfortunately, I don't think this change is justified; `Transmit` is small and cheap to move, and `proto::Endpoint`-originated packets are small in number regardless; most I/O goes directly through connections.

TimonPost:
Feel free to close it if you don't see it fit. Tho even if it is not justified, I don't see why you would want to have two places to store things while it can easaly be one storage.

Ralith:
To avoid significant expansion of the `quinn_proto::Endpoint` API, at least.

PR #1273, opened 2021-12-29
Strip down futures-* dependencies

Ralith:
This drops 6 packages from the workspace-scoped Cargo.lock, and should save a decent amount of buildtime, both locally and downstream.


djc (approved):
Nice!

PR #1283, opened 2022-01-10
Adjust the book with version 0.8.0

thekuwayama:
Hi!
This PR updates the book.

- adjust links that point into the no longer existent "generic" directory
- `quinn::ClientConfigBuilder`, `ServerConfigBuilder` and `EndpointBuilder` have already been removed (#1213), so adjust
- `quinn::Certificate` have already been removed (#1214), so adjust

I referred to [quinn/examples/insecure_connection.rs](https://github.com/quinn-rs/quinn/blob/0.8.0/quinn/examples/insecure_connection.rs), [quinn/examples/client.rs](https://github.com/quinn-rs/quinn/blob/0.8.0/quinn/examples/client.rs) and [quinn/examples/server.rs](https://github.com/quinn-rs/quinn/blob/0.8.0/quinn/examples/server.rs) to adjust the book.

### DIFF
--- a/docs/book/src/quinn/certificate.md
+++ b/docs/book/src/quinn/certificate.md
@@ -129,14 +129,9 @@ This is the only thing you need to do for your server to be secured.
 **Configure Client**
 
 ```rust
-let mut roots = rustls::RootCertStore::empty()
-roots.add(&cert).unwrap();
 let mut crypto = rustls::ClientConfig::builder()
-    .with_safe_default_cipher_suites()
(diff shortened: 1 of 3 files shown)


djc:
This looks great, thanks! A couple of ideas riffing on the same theme:

* Should we fix the links to docs.rs to point to specific versions rather than `latest`?
* Would it be viable to use [skeptic](https://docs.rs/skeptic/latest/skeptic/) to compile-test these code samples?

djc (approved):

Issue #1281, opened 2022-01-07
`Readable` event not queued on first stream open when data is readable. 


TimonPost:
Given the following events: 
- `Event::Readable  // One or more new streams has been opened`
- `Event::Opened // A currently open stream has data or errors waiting to be read`

`Readable` is generated in `StreamState::on_stream_frame`

```rust
    /// Notify the application that new streams were opened or a stream became readable.
    fn on_stream_frame(&mut self, notify_readable: bool, stream: StreamId) {
        if stream.initiator() == self.side {
            // Notifying about the opening of locally-initiated streams would be redundant.
            if notify_readable {
                self.events.push_back(StreamEvent::Readable { id: stream });
            }
            return;
        }
        let next = &mut self.next_remote[stream.dir() as usize];

        if stream.index() >= *next {
            *next = stream.index() + 1;
            self.opened[stream.dir() as usize] = true;
        } else if notify_readable {
            self.events.push_back(StreamEvent::Readable { id: stream });
        }
    }
```
The flag for `opened` streams is also set here. This means that when we process an incoming frame it either marks an opening stream or a readable event. While, as a user, I would expect both an `Opened` and `Readable` event when I receive data 'and' a stream is opened in the same packet. 

**Why should we care**
When creating an FFI, or implementation of `quinn-proto` it can be confusing as one assumes to receive a `Readable` event when data is `Readable`. But instead, data could be ready to be read on `Opened` event as well. 

**Solutions:**

- Add an extra comment that `Opened` could mean `Readable`
- Pushback a `Readable` event when `stream.index() >= *next`. The `StreamState::poll` first processes any opened events, then later it processes other events. So this could be done without impacting the code too much. Tho it might be a false positive. But would a false positive be better than nothing at all?

TimonPost:
Not related to this issue, but a question I had regarding a note in this code:

```
// Notifying about the opening of locally-initiated streams would be redundant.
```

However, `if stream.initiator() == self.side ` checks if `initiated stream == the locally endpoint side`. So why if this is redundant is this code still queuing the `Readable` event? I am probably misinterpreting this comment or code.

Ralith:
[...]

PR #1284, opened 2022-01-11
Clarify meaning of events.

TimonPost:
Related: #1281

### DIFF
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -362,12 +362,12 @@ impl Ord for PendingLevel {
 /// Application events about streams
 #[derive(Debug, PartialEq, Eq)]
 pub enum StreamEvent {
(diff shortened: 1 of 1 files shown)


quinn-proto/src/connection/streams/mod.rs
```diff
@@ -362,12 +362,12 @@ impl Ord for PendingLevel {
 /// Application events about streams
 #[derive(Debug, PartialEq, Eq)]
 pub enum StreamEvent {
-    /// One or more new streams has been opened
+    /// One or more new streams has been opened and might be readable.
```
Ralith, line 365: nit: we avoid trailing commas on the first line of doc comments
```suggestion
    /// One or more new streams has been opened and might be readable
```

Ralith (commented):
Thanks, this is good wording.

djc (approved):

PR #1286, opened 2022-01-22
Implement draft-ietf-quic-bit-grease-02

Ralith:
Draft backwards-compatible anti-ossification measure.

### DIFF
--- a/quinn-proto/src/transport_parameters.rs
+++ b/quinn-proto/src/transport_parameters.rs
@@ -77,6 +77,9 @@ macro_rules! make_struct {
             /// The value that the endpoint included in the Source Connection ID field of the first
             /// Initial packet it sends for the connection
             pub(crate) initial_src_cid: Option<ConnectionId>,
+            /// The endpoint is willing to receive QUIC packets containing any value for the fixed
+            /// bit
+            pub(crate) grease_quic_bit: bool,
 
             // Server-only
             /// The value of the Destination Connection ID field from the first Initial packet sent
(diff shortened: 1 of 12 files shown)


quinn-proto/src/transport_parameters.rs
```diff
@@ -381,6 +392,12 @@ impl TransportParameters {
                     }
                     params.max_datagram_frame_size = Some(r.get().unwrap());
                 }
+                0x2ab2 => {
+                    if len != 0 {
+                        return Err(Error::Malformed);
+                    }
+                    params.grease_quic_bit = true;
```
djc, line 399: Nit: `match`?

djc (dismissed):
Nice, LGTM.

quinn-proto/src/transport_parameters.rs
Ralith: I don't see a way to use `match` here that doesn't add more syntatic noise. Did you have something specific in mind?

djc:
```rust
match len {
    0 => params.grease_quic_bit = true,
    _ => return Err(Error::Malformed),
}
```
?

Ralith: Ah, my brain has trouble putting side-effect-oriented code in value positions like that :D

djc: Statements are just expressions that yield () as far as Rust is concerned. 😎

djc (approved):

PR #1287, opened 2022-01-22
Replace structopt with its successor, clap 3

Ralith:
Only affects examples/tools.

### DIFF
--- a/bench/Cargo.toml
+++ b/bench/Cargo.toml
@@ -12,7 +12,7 @@ hdrhistogram = { version = "7.2", default-features = false }
 quinn = { path = "../quinn" }
 rcgen = "0.9.1"
 rustls = { version = "0.20", default-features = false, features = ["quic"] }
-structopt = "0.3"
+clap = { version = "3.2", features = ["derive"] }
 tokio = { version = "1.0.1", features = ["rt", "sync"] }
 tracing = "0.1.10"
 tracing-subscriber = { version = "0.3.0", default-features = false, features = ["env-filter", "fmt", "ansi", "time", "local-time"] }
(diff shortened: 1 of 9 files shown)


Ralith:
Looks like this is blocked on updating MSRV to 1.54.

djc:
See https://github.com/clap-rs/clap/issues/3267.

bench/Cargo.toml
```diff
@@ -12,7 +12,7 @@ hdrhistogram = { version = "7.2", default-features = false }
 quinn = { path = "../quinn" }
 rcgen = "0.8"
 rustls = { version = "0.20", default-features = false, features = ["quic"] }
-structopt = "0.3"
+clap = { version = "3.0.10", features = ["derive"] }
```
KirillLykov, line 15: Why did you choose `3.0.10` instead of `3.1.x`?

perf/src/bin/perf_client.rs
```diff
@@ -16,54 +16,54 @@ use perf::stats::{OpenStreamStats, Stats};
...
 
 /// Connects to a QUIC perf server and maintains a specified pattern of requests until interrupted
-#[derive(StructOpt)]
-#[structopt(name = "client")]
+#[derive(Parser)]
+#[clap(name = "client")]
 struct Opt {
     /// Host to connect to
-    #[structopt(default_value = "localhost:4433")]
+    #[clap(default_value = "localhost:4433")]
```
KirillLykov, line 23: Looks like the only place we you use string instead of `SocketAddr`, might be good to format in later prs

Ralith: `SocketAddr` cannot represent hostnames.

Ralith:
This should be mergeable now that we've bumped MSRV.

djc (approved):

PR #1288, opened 2022-01-25
Fix document example

sizumita:
- Make send mutable because of fixes rust error.
- Fix typo


sizumita:
and I think

```rust
let (mut send, recv) = connection
        .open_bi()
        .await?;
```

is

```
let (mut send, recv) = connection.connection
        .open_bi()
        .await?;
```

djc (approved):

djc:
Thanks! Would you like to submit another PR for the `connection.connection` issue?

PR #1289, opened 2022-01-26
Rename variable of example

sizumita:
`endpoint.connect` returns NewConnection so I think we should rename from connection to new_connection.
And `let connection = new_connection.connection;` is easy to understand but it's not good because of partial move.
Any ideas?


djc:
I think it would be nice to destruct it? `let NewConnection { connection, .. } = new_connection;`.

sizumita:
That's a good one.

djc (approved):

PR #1290, opened 2022-01-28
chore: Clean up a few Result patterns that fit an std combinator

huitseeker:
Opinions on this refactoring loosely held.
Tool-aided by [comby-rust](https://github.com/huitseeker/comby-rust).

### DIFF
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -254,10 +254,10 @@ impl<'a> SendStream<'a> {
     /// # Panics
     /// - when applied to a receive stream
     pub fn reset(&mut self, error_code: VarInt) -> Result<(), UnknownStream> {
-        let stream = match self.state.send.get_mut(&self.id) {
-            Some(ss) => ss,
-            None => return Err(UnknownStream { _private: () }),
-        };
+        let stream = self
+            .state
+            .send
+            .get_mut(&self.id).ok_or(UnknownStream { _private: () })?;
 
         if matches!(stream.state, SendState::ResetSent) {
             // Redundant reset call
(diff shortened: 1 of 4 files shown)


quinn/examples/server.rs
```diff
@@ -247,7 +247,7 @@ fn process_get(root: &Path, x: &[u8]) -> Result<Vec<u8>> {
         bail!("missing \\r\\n");
     }
     let x = &x[4..x.len() - 2];
-    let end = x.iter().position(|&c| c == b' ').unwrap_or_else(|| x.len());
+    let end = x.iter().position(|&c| c == b' ').unwrap_or(x.len());
```
Ralith, line 250: I think clippy won't like this, since it makes a function call unconditional.

Ralith (commented):
Thanks, overall this looks great!

quinn/examples/server.rs
huitseeker: The following is free of clippy errors:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cae14bfe382e66bfdda86fc25b5ff652

More details here:
https://github.com/rust-lang/rust-clippy/issues/6943

Ralith (approved):

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -254,10 +254,10 @@ impl<'a> SendStream<'a> {
     /// # Panics
     /// - when applied to a receive stream
     pub fn reset(&mut self, error_code: VarInt) -> Result<(), UnknownStream> {
-        let stream = match self.state.send.get_mut(&self.id) {
-            Some(ss) => ss,
-            None => return Err(UnknownStream { _private: () }),
-        };
+        let stream = self
```
djc, line 257: It's not that important, but I don't think I'd count this as an improvement; basically I think my argument would be is that this is the same number of lines but with all the density on a single line. By comparison, the `match` version spreads the density around between "what are we matching" and "what do we do if we can't find the thing we're looking for", and makes it more visually obvious that there are two alternatives.

Ralith: I appreciate the readability of a standard combinator, personally; it's very easy to interpret at a glance, at least if you use a lot of `ok_or`, which I find that I do.

PR #1291, opened 2022-01-29
Minor doc fixes, document private items on ci

koivunej:
Very minor drive-by doc fixes removing the warnings from `cargo doc --document-private-links` but also enables the `#[warn(rustdoc::broken_intra_doc_links)]` to catch these earlier. With the additional `--document-private-items` in workflow the private links are checked as well.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -69,7 +69,7 @@ jobs:
           override: true
           components: clippy
       - name: doc
-        run: cargo doc --no-deps
+        run: cargo doc --no-deps --document-private-items
         env:
           RUSTDOCFLAGS: -Dwarnings
       - name: lint fuzz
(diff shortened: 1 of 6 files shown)


koivunej:
No strong feelings on the `#[deny(rustdoc::broken_intra_doc_links)]` nor the workflow change, happy to drop them if they are not wanted.

djc:
Thanks for working on this!

Can we do -D warn in CI instead of  denying at compile time?

djc (dismissed):

koivunej:
Changed the deny to warn by changing the 4th commit.

koivunej:
> Can we do -D warn in CI instead of denying at compile time?

Initially misunderstood this and changed the `#![deny(rustdoc::...)]` to `#![warn(rustdoc::...)]` but instead you meant to use RUSTDOCFLAGS when doing `cargo doc`. I had missed that RUSTDOCFLAGS="-Dwarnings" was already being set so the only needed thing was the `--document-private-items`.

Ralith (approved):
Thanks!

PR #1294, opened 2022-02-09
Don't update remote client CIDs gratuitously

Ralith:
Handling retire_prior_to might have already moved us off the initial CID.


quinn-proto/src/cid_queue.rs
```diff
@@ -94,9 +96,9 @@ impl CidQueue {
         self.buffer[self.cursor].unwrap().0
     }
 
-    /// Check whether self.offset points to a valid CID in CidQueue
-    pub fn is_active_retired(&mut self) -> bool {
-        self.buffer[self.cursor].is_none()
+    /// Return active reset token
+    pub fn active_reset_token(&self) -> Option<ResetToken> {
+        self.buffer[self.cursor].unwrap().1
```
djc, line 101: Why is this unwrap() safe? Could/should we `map()` instead?

quinn-proto/src/connection/mod.rs
```diff
@@ -2502,16 +2499,23 @@ impl Connection {
...
-                        if self.rem_cids.active_seq() == 0 {
-                            // We're a server still using the initial remote CID for the client, so
-                            // let's switch immediately to enable clientside stateless resets.
-                            let _ = self.update_rem_cid();
-                        }
-                    } else if self.rem_cids.is_active_retired() {
-                        // If our current CID is meant to be retired; or
-                        // active remote CID is invalid (due to packet loss or reordering),
-                        // we must switch to next valid CID
+                    if !retired.is_empty() {
```
djc, line 2502: If the `retired.is_empty()` invariant implies that `active_reset_token()` must be `Some`, can we rely on that more directly?

quinn-proto/src/connection/mod.rs
```diff
@@ -2502,16 +2499,23 @@ impl Connection {
...
+                        self.set_reset_token(
+                            self.rem_cids
+                                .active_reset_token()
+                                .expect("missing reset token for non-initial CID"),
+                        );
+                    }
+
+                    if self.side.is_server() && self.rem_cids.active_seq() == 0 {
+                        // We're a server still using the initial remote CID for the client, so
+                        // let's switch immediately to enable clientside stateless resets.
```
djc, line 2518: Document the invariant for the `unwrap()` below a little more explicitly?

djc (commented):
I'm not very familiar with this logic, so this is a bit of a superficial review.

@Matthias247 would you mind reviewing this?

quinn-proto/src/cid_queue.rs
Ralith: Mooted in refactoring.

quinn-proto/src/connection/mod.rs
Ralith: This was a very good question and led me on an adventure that ended in refactoring the entire thing again, this time better encapsulating the NEW_CONNECTION_ID handling logic in CidQueue by way of merging `insert` and `retire_prior_to`.

quinn-proto/src/connection/mod.rs
Ralith: Mooted by refactoring.

quinn-proto/src/cid_queue.rs
```diff
@@ -31,42 +31,71 @@ impl CidQueue {
...
-        let orig_offset = self.offset;
-        self.offset = sequence;
-        self.cursor = (self.cursor + n) % Self::LEN;
-        orig_offset..sequence
+
+        // Record the new CID
+        let index = ((self.cursor as u64 + index) % Self::LEN as u64) as usize;
+        self.buffer[index] = Some((cid.id, Some(cid.reset_token)));
+
+        if retired_count != 0 {
```
djc, line 68: Invert this for early return?

djc (commented):
(Still going to take a more thorough look at the changes here.)

djc (approved):

quinn-proto/src/cid_queue.rs
Ralith: Good catch! Done.

Ralith:
Tidied things up a bit more, in particular to get rid of the dubious `Result` from `update_rem_cid`.

djc (approved):

djc:
We should probably timeout on additional reviews and just get this merged. It seems unlikely to be worse given the added tests!

quinn-proto/src/cid_queue.rs
```diff
@@ -31,60 +31,83 @@ impl CidQueue {
...
+        &mut self,
+        cid: NewConnectionId,
+    ) -> Result<Option<(Range<u64>, ResetToken)>, InsertError> {
+        // Position of new CID wrt. the current active CID
+        let index = match cid.sequence.checked_sub(self.offset) {
+            None => return Err(InsertError::Retired),
+            Some(x) => x,
+        };
+
+        let retired_count = cid.retire_prior_to.saturating_sub(self.offset);
```
Matthias247, line 48: If the list had gaps, is this still the right number? I guess it doesn't matter since it resets anything to `None` anyway

quinn-proto/src/cid_queue.rs
```diff
@@ -31,60 +31,83 @@ impl CidQueue {
...
+        let (i, (_, token)) = self
+            .iter()
+            .next()
+            .expect("it is impossible to retire a CID without supplying a new one");
+        self.cursor = (self.cursor + i) % Self::LEN;
         let orig_offset = self.offset;
-        self.offset = sequence;
-        self.cursor = (self.cursor + n) % Self::LEN;
-        orig_offset..sequence
+        self.offset = cid.retire_prior_to + i as u64;
```
Matthias247, line 76: why does this one not wrap?

quinn-proto/src/cid_queue.rs
```diff
@@ -31,60 +31,83 @@ impl CidQueue {
...
+
+        // Record the new CID
+        let index = ((self.cursor as u64 + index) % Self::LEN as u64) as usize;
+        self.buffer[index] = Some((cid.id, Some(cid.reset_token)));
+
+        if retired_count == 0 {
+            return Ok(None);
+        }
+
+        // The active CID was retired. Find the first known CID with sequence number of at least
```
Matthias247, line 66: why was it the active one, instead of a lower one? Is the active one always the lowest in the list?

quinn-proto/src/cid_queue.rs
Ralith: CIDs must be issued in order, so semantically the gaps don't matter, and we take special care not to do work for CIDs between initial offset + LEN and the first unretired CID. It's true that `retired_count` is not the number of actually known CIDs that are retired, but that number isn't what we need anywhere anyway.

quinn-proto/src/cid_queue.rs
Ralith: `offset` is absolute; it's the sequence number at which the circular buffer logically starts.

quinn-proto/src/cid_queue.rs
Ralith: Yep; the set of known CIDs is a sliding window into the sequence number space, where the first element in the window is the CID we use.

PR #1295, opened 2022-02-09
Establish 0.8 branch, backport bugfixes

Ralith:


Ralith:
CI failure is due to an upstream MSRV issue, which I think shouldn't block.

Ralith:
I figured we'd try to get https://github.com/rustls/rustls/pull/1006 in as well, since that's probably breaking everyone.

PR #1297, opened 2022-02-10
Upgrade rcgen to 0.9.1

djc:


djc:
(https://github.com/rust-fuzz/arbitrary/issues/99.)

djc:
Yup, I'll fix that up tomorrow.

Ralith:
> Only build fuzz targets against MSRV

Should this read "don't" instead of "only"?

djc:
Err, yes!

djc:
Fixed.

PR #1314, opened 2022-02-23
config: add ServerConfig::with_transport_config() builder method

### LINKED ISSUE #1301: [Feature]: Make configuration fields in `quinn-proto` public
# Some `Arc` Shenanigans
To initialize a server/client endpoint, Quinn first requires configuration (e.g. via `TransportConfig`, `ServerConfig`, etc.). As mentioned in the documentation, the default options are sufficient for most use cases. Otherwise, we have to go through some hoops with setter methods to achieve the behavior we want.

```rust
use quinn::{ServerConfig, TransportConifg, VarInt};
use std::{sync::Arc, time::Duration};

// Configure some special ALPN protocols via `rustls`
let mut server_config = ServerConfig::with_crypto(crypto);

// Create a new transport configuration, and modify its defaults
let mut transport_config = TransportConfig::default();
transport_config
    .keep_alive_interval(Some(Duration::from_secs(30)))
    .max_idle_timeout(Some(VarInt::from(60u32).into()));

// And then... replace the old transport config? This is necessary
// because there's no public constructor for `ServerConfig` that accepts a
// `TransportConfig` at the moment; there is only one for the `crypto` field.
server_config.transport = Arc::new(transport_config);

// Alternatively, we may also modify through the `Arc` directly
// since we're the sole owner of the reference at the moment.
// But, now we're forced to use a nasty `unwrap` in the code...
//
// We may, of course, succumb to the `unsafe` version: `unwrap_unchecked`.
// Although this is technically legal, the Quinn public API documents
// no guarantees about the reference count of the `transport` field. Hence,
// a version bump may unintentionally invoke undefined behavior if we're not careful.
Arc::get_mut(&mut server_config.transport)
    .unwrap()
    .keep_alive_interval(Some(Duration::from_secs(30)))
    .max_idle_timeout(Some(VarInt::from(60u32).into()));

// Finally initialize the endpoint
let (endpoint, incoming) = Endpoint::server(server_config, addr)?;
```

I hope the example above demonstrates my nitpicks with the current API. This is exactly what I faced in one of my projects which required a custom protocol (mostly for network efficiency's sake because JSON-over-HTTP is too verbose).

# My Proposed Solution
Instead, I propose that all fields of the various `*Config` structs be made public. After all, most of their methods are solely transparent setters. We may as well treat the fields to be public.

To accommodate for default values, we simply use the default spread syntax (`..Default::default()`) when initializing from the user code.

```rust
use quinn::{ServerConfig, TransportConifg, VarInt};
use std::{sync::Arc, time::Duration};

// NOTE: Observe that `mut` is no longer necessary.
let server_config = ServerConfig {
    // Configure some special ALPN protocols via `rustls`
    crypto: Arc::new(rustls::server::ServerConfig { ... }),
    transport: Arc::new(TransportConfig {
        keep_alive_interval: Some(Duration::from_secs(30)),
        max_idle_timeout: Some(VarInt::from(60u32).into()),
        // 🎉 Hooray for default values! 🎉
        ..Default::default()
    }),
    // 🎉 Hooray for **more** default values! 🎉
    // ASIDE: `ServerConfig` does not currently implement `Default`
    // as of Quinn 0.8. This should be addressed before proceeding.
    ..Default::default()
};

// Finally initialize the endpoint
let (endpoint, incoming) = Endpoint::server(server_config, addr)?;
```

Here, I demonstrate the fact that we no longer have to jump through the hoops of indirection. In particular, direct initialization allows us to remove the `Arc` shenanigans altogether.

This, to me, is a significantly more ergonomic API than using the original setters. Of course, we don't have to remove the setters from the library. Though, I am not totally against deprecation.

I would love to know your thoughts about this API change. I'd love to send in a PR that addresses this. Just wanted to collect some feedback before committing my time.

djc:
Fixes #1301.

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -562,10 +562,10 @@ impl fmt::Debug for ServerConfig {
 #[non_exhaustive]
 pub struct ClientConfig {
     /// Transport configuration to use
-    pub transport: Arc<TransportConfig>,
+    pub(crate) transport: Arc<TransportConfig>,
 
     /// Cryptographic configuration to use
-    pub crypto: Arc<dyn crypto::ClientConfig>,
+    pub(crate) crypto: Arc<dyn crypto::ClientConfig>,
 
     /// QUIC protocol version to use
     pub(crate) version: u32,
@@ -579,6 +579,12 @@ impl ClientConfig {
             crypto,
             version: 1,
         }
+    }
+
+    /// Set a custom [`TransportConfig`]
+    pub fn transport_config(&mut self, transport: Arc<TransportConfig>) -> &mut Self {
+        self.transport = transport;
+        self
     }
 
     /// Set the QUIC version to use
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -467,6 +467,12 @@ impl ServerConfig {
         }
     }
 
+    /// Set a custom [`TransportConfig`]
+    pub fn transport_config(&mut self, transport: Arc<TransportConfig>) -> &mut Self {
+        self.transport = transport;
+        self
+    }
+
     /// Private key used to authenticate data included in handshake tokens.
     pub fn token_key(&mut self, value: Arc<dyn HandshakeTokenKey>) -> &mut Self {
         self.token_key = value;
--- a/bench/src/lib.rs
+++ b/bench/src/lib.rs
@@ -68,7 +68,7 @@ pub async fn connect_client(
         .with_no_client_auth();
 
     let mut client_config = quinn::ClientConfig::new(Arc::new(crypto));
-    client_config.transport = Arc::new(transport_config(&opt));
+    client_config.transport_config(Arc::new(transport_config(&opt)));
 
     let quinn::NewConnection { connection, .. } = endpoint
         .connect_with(client_config, server_addr, "localhost")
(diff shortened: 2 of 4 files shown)


quinn-proto/src/config.rs
```diff
@@ -575,6 +581,12 @@ impl ClientConfig {
         }
     }
 
+    /// Set a custom [`TransportConfig`]
+    pub fn with_transport_config(mut self, transport: TransportConfig) -> Self {
+        self.transport = Arc::new(transport);
+        self
+    }
```
Ralith, line 588: The `with_` prefix is used below for constructors, rather than setters, which seems a bit confusing. We're also using by-`&mut` builders everywhere else.

djc:
(Failure is unrelated, fixed in #1316.)

Some-Dood:
Hello there! I'm surprised that the `with_transport_config` setters take an `Arc<TransportConfig>` instead of a plain `TransportConfig`. Shouldn't it be the latter instead since `TransportConfig` is a concrete type rather than a `dyn` object (in which case the `Arc` is actually necessary)?

Although it is honest to its internal representation, shouldn't the user code not have to worry about wrapping it in an `Arc` first?

djc:
I would say this is actually useful. For example, if you have multiple endpoints with the same `TransportConfig`, you can now share the same `TransportConfig`. Additionally it can often be useful to have the caller supply the exact type you need, rather than the callee taking care of some conversion (arguably my example is just one specific example of that same principle).

Some-Dood:
> If you have multiple endpoints with the same `TransportConfig`, you can now share the same `TransportConfig`.

Fair point. I'm convinced. 🎉

Ralith (approved):
First commit message seems to be obsolete; actual changes LGTM though.

djc:
Fixed the commit message.

Ralith (approved):

PR #1315, opened 2022-02-25
Allow adjustment of per-connection concurrent stream limits

### LINKED ISSUE #1313: Per-connection stream limits
I'm looking into limiting users by applying a different outstanding stream limit per connection to allow some users to get more resources than others. I don't believe this is possible in quinn as of now, is that correct?

Ralith:
Fixes #1313.

This is useful when multiple application protocols are hosted on the same endpoint, or when application logic such as authentication might want to adjust the limits of pre-existing connections.

### DIFF
--- a/quinn-proto/src/connection/streams/state.rs
+++ b/quinn-proto/src/connection/streams/state.rs
@@ -26,11 +26,20 @@ pub struct StreamsState {
     pub(super) send: FxHashMap<StreamId, Send>,
     pub(super) recv: FxHashMap<StreamId, Recv>,
     pub(super) next: [u64; 2],
-    // Locally initiated
+    /// Maximum number of locally-initiated streams that may be opened over the lifetime of the
+    /// connection so far, per direction
     pub(super) max: [u64; 2],
-    // Maximum that can be remotely initiated
+    /// Maximum number of remotely-initiated streams that may be opened over the lifetime of the
+    /// connection so far, per direction
     max_remote: [u64; 2],
-    // Lowest that hasn't actually been opened
+    /// Number of streams that we've given the peer permission to open
+    allocated_remote_count: [u64; 2],
+    /// Size of the desired stream flow control window. May be smaller than `allocated_remote_count`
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/streams/state.rs
```diff
@@ -782,7 +801,10 @@ impl StreamsState {
                     StreamHalf::Recv => !self.send.contains_key(&id),
                 };
             if fully_free {
-                self.alloc_remote_stream(id.dir());
+                self.remote_count[id.dir() as usize] -= 1;
+                if self.remote_count[id.dir() as usize] < self.max_remote_count[id.dir() as usize] {
```
djc, line 805: Did you consider putting this condition inside `alloc_remote_streams()`? I think that would prevent the repetition here and in `set_max_concurrent()`?

djc (approved):

quinn-proto/src/connection/streams/state.rs
Ralith: Tried this out in a separate commit. It's a semantic change, but I can see how it might be nicer. Thoughts?

djc (dismissed):

quinn-proto/src/connection/streams/state.rs
djc: It does seem to centralize the logic more which seems nice? You decide which you prefer.

Ralith:
On review, the previous implementation had significant bugs:
- `ensure_remote_streams` was not idempotent
- `MAX_STREAMS` frames sent in 0-RTT would not be retransmitted if 0-RTT was rejected

I've adjusted the implementation to resolve those by:
- Counting *allocated* remote streams rather than *concurrently opened* ones
- Testing that we allocate the expected number of streams on `set_max_concurrent` calls
- Explicitly tracking whether flow control was adjusted during 0-RTT

Please re-review.

sakridge:
Any updates on this?

djc (approved):

PR #1316, opened 2022-02-25
Clean up BBR code

djc:
Started by fixing some clippy warnings, cleaned up some other things that stuck out at me.

### DIFF
--- a/quinn-proto/src/congestion/bbr/bw_estimation.rs
+++ b/quinn-proto/src/congestion/bbr/bw_estimation.rs
@@ -84,7 +84,7 @@ impl BandwidthEstimation {
         } else {
             BandwidthEstimation::bw_from_delta(
                 self.total_acked - self.prev_total_acked,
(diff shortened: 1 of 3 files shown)


quinn-proto/src/congestion/bbr/bw_estimation.rs
```diff
@@ -26,7 +26,7 @@ impl Default for BandwidthEstimation {
             prev_acked_time: None,
             total_sent: 0,
             prev_total_sent: 0,
-            sent_time: None,
+            sent_time: Instant::now(),
```
Ralith: Maybe comment to indicate this is a dummy value, rather than something semantically meaningful?

Ralith (approved):

quinn-proto/src/congestion/bbr/bw_estimation.rs
djc: Done.

Ralith (approved):

PR #1332, opened 2022-03-25
Retry token fixes

Ralith:
Resolves some correctness issues detected in interop testing, and makes us a bit better at detecting them ourselves.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1587,6 +1587,15 @@ impl Connection {
         debug_assert!(self.side.is_server());
         let len = packet.header_data.len() + packet.payload.len();
         self.path.total_recvd = len as u64;
+        match self.state {
+            State::Handshake(ref mut state) => match packet.header {
+                Header::Initial { ref token, .. } => {
+                    state.token = Some(token.clone());
+                }
+                _ => unreachable!("first packet must be an Initial packet"),
+            },
+            _ => unreachable!("first packet must be delivered in Handshake state"),
+        }
 
         self.on_packet_authenticated(
             now,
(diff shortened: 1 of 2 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -1908,6 +1917,23 @@ impl Connection {
                     trace!("dropping short packet during handshake");
                     return;
                 } else {
+                    if let (
+                        &Header::Initial { ref token, .. },
+                        &State::Handshake(state::Handshake {
+                            token: ref expected_token,
+                            ..
+                        }),
+                    ) = (&packet.header, &self.state)
```
djc, line 1934: Oof, I often find multi-line `if let`s prety hard to follow. Is there some other way we can format this, maybe using some temporary named variables?

djc (commented):
I'm confused by all this. This seems to replace `Option`s with empty byte sequences all over the place, which doesn't seem like an improvement, especially when we just add comments saying they're always empty for clients.

quinn-proto/src/connection/mod.rs
```diff
@@ -243,7 +243,7 @@ impl Connection {
         };
         let state = State::Handshake(state::Handshake {
             rem_cid_set: side.is_server(),
-            token: None,
+            token: Bytes::new(),
```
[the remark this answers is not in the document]
Ralith: Because `None` is always handled exactly as if it were the empty `Bytes`. It's not a sentinel value, because we don't detect it and thereby special-case behavior: because the normal case is literally an empty token, the empty `Bytes` produces the correct behavior without special effort.

djc (dismissed):

quinn-proto/src/connection/mod.rs
Ralith: Split into two `if let`s and dropped the `state::Handshake` destructuring. Much clearer, I agree.

djc (approved):

PR #1344, opened 2022-04-15
remove direct mio dependency

yu-re-ka:
Unless I overlooked something, we don't use the mio UdpSocket for anything other than the `only_v6`, which can be easily replaced

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -20,24 +20,42 @@ type IpTosTy = libc::c_uchar;
 #[cfg(not(target_os = "freebsd"))]
 type IpTosTy = libc::c_int;
 
+fn only_v6(sock: &std::net::UdpSocket) -> io::Result<bool> {
+    let raw_fd = sock.as_raw_fd();
+    let mut val: libc::c_int = 0;
+    let mut len = mem::size_of::<libc::c_int>() as libc::socklen_t;
+    let res = unsafe {
+        libc::getsockopt(
+            raw_fd,
+            libc::IPPROTO_IPV6,
+            libc::IPV6_V6ONLY,
+            &mut val as *mut _ as *mut _,
+            &mut len,
+        )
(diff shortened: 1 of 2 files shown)


quinn-udp/src/unix.rs
```diff
@@ -20,24 +20,48 @@ type IpTosTy = libc::c_uchar;
...
+        let res = unsafe {
+            libc::getsockopt(
+                raw_fd,
+                libc::IPPROTO_IPV6,
+                libc::IPV6_V6ONLY,
+                &mut val as *mut _ as *mut _,
+                &mut len,
+            )
+        };
+        if res == -1 {
```
djc, line 41: Style nit: I think this would be nicer as a `match` expression:

```rust
match res {
    -1 => Err(io::Error::last_os_err()),
    _ => Ok(val != 0),
}
```

quinn-udp/src/unix.rs
```diff
@@ -20,24 +20,48 @@ type IpTosTy = libc::c_uchar;
 #[cfg(not(target_os = "freebsd"))]
 type IpTosTy = libc::c_int;
 
+trait StdUdpSocketExt {
```
djc, line 23: Given that we have only one caller, the extension trait seems heavy-handed. Are you planning some follow-up that will build on this more?

djc (commented):
The idea seems pretty reasonable, some further questions below:

quinn-udp/src/unix.rs
yu-re-ka: There was no thought behind this really, other than being able to call the function with `socket.only_v6()` syntax

quinn-udp/src/unix.rs
yu-re-ka: will do

djc (approved):

Ralith (approved):
Nice!

PR #1345, opened 2022-04-15
async-std support, second try

yu-re-ka:
Unlike the first try, this does not rely on mutually exclusive feature flags. Both runtime-async-std and runtime-tokio features can be enabled at the same time. The actual runtime used is chosen by the user when creating the endpoint.

This does not touch the synchronization primitives (still uses channels etc. from tokio, even when running under async-std), and thus keeps tokio as hard-coded dependency.

To-Do:

- [x] Tokio socket read/write paths needs to go via the AsyncFdReadyGuard again, otherwise it breaks
- [x] Fix windows / !unix compilation
- [x] Deal with lint errors that are not fixed by `cargo fmt` (" error: very complex type used. Consider factoring parts into `type` definitions")

### DIFF
--- a/quinn-runtime/Cargo.toml
+++ b/quinn-runtime/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "quinn-runtime"
+version = "0.1.0"
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/quinn-rs/quinn"
+description = "Abstractions for writing async-runtime-independent code"
+keywords = ["quic"]
+categories = ["network-programming", "asynchronous"]
+workspace = ".."
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+tokio = { version = "1.0.1", features = ["net", "time"], optional = true }
+async-io = { version = "1.6", optional = true }
+async-std = { version = "1.11", optional = true }
+
+[features]
+runtime-tokio = [ "tokio" ]
+runtime-async-std = [ "async-io", "async-std" ]
--- a/quinn-runtime/src/lib.rs
+++ b/quinn-runtime/src/lib.rs
@@ -1,0 +1,71 @@
+#[cfg(feature = "runtime-tokio")]
+mod tokio_runtime;
+#[cfg(feature = "runtime-tokio")]
+pub use tokio_runtime::*;
+
+#[cfg(feature = "runtime-async-std")]
+mod async_std_runtime;
+#[cfg(feature = "runtime-async-std")]
+pub use async_std_runtime::*;
+
+use std::fmt::Debug;
+use std::future::Future;
+use std::io;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use std::time::Instant;
+
+pub trait AsyncWrappedUdpSocket: Sized + Send + Debug {
+    fn new(t: std::net::UdpSocket) -> io::Result<Self>;
+
+    // On Unix we expect to be able to access the underlying std UdpSocket
+    // to be able to implement more advanced features
+    #[cfg(unix)]
+    fn poll_read<T>(
+        &self,
+        f: impl FnOnce(&std::net::UdpSocket) -> io::Result<T>,
+        cx: &mut Context,
+    ) -> Poll<io::Result<T>>;
+    #[cfg(unix)]
+    fn poll_write<T>(
+        &mut self,
+        f: impl FnOnce(&std::net::UdpSocket) -> io::Result<T>,
+        cx: &mut Context,
+    ) -> Poll<io::Result<T>>;
+    #[cfg(unix)]
+    fn get_ref(&self) -> &std::net::UdpSocket;
+
+    // On Non-Unix platforms we only expect to be able to do basic
+    // send_to / recv_from operations.
+    #[cfg(not(unix))]
+    fn poll_recv_from(
+        &self,
+        cx: &mut Context,
+        buf: &mut [u8],
+    ) -> Poll<io::Result<(usize, std::net::SocketAddr)>>;
+    #[cfg(not(unix))]
+    fn poll_send_to(
+        &mut self,
+        cx: &mut Context,
+        buf: &[u8],
+        target: std::net::SocketAddr,
+    ) -> Poll<io::Result<usize>>;
+    #[cfg(not(unix))]
+    fn local_addr(&self) -> io::Result<std::net::SocketAddr>;
+}
+
+pub trait AsyncTimer: Sized + Send + Debug {
+    fn new(i: Instant) -> Self;
+    fn reset(self: Pin<&mut Self>, i: Instant);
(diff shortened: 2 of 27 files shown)


yu-re-ka:
@Ralith do you think this is an acceptable solution?

yu-re-ka:
CI passes 🎉

quinn-runtime/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+pub use async_std_runtime::*;
+
+use std::fmt::Debug;
+use std::future::Future;
+use std::io;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use std::time::Instant;
+
+pub trait AsyncWrappedUdpSocket: Sized + Send + Debug {
```
Ralith, line 18: Why the cfg gates on these methods?

quinn-runtime/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "quinn-runtime"
```
Ralith, line 2: What's the benefit of introducing a new crate, rather than defining this inside `quinn`?

quinn-runtime/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn local_addr(&self) -> io::Result<std::net::SocketAddr>;
+}
+
+pub trait AsyncTimer: Sized + Send + Debug {
+    fn new(i: Instant) -> Self;
+    fn reset(self: Pin<&mut Self>, i: Instant);
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<()>;
+}
+
+pub trait Runtime: Send + 'static {
```
Ralith, line 63: I believe the API churn and ergonomic cost to this change would be dramatically reduced if this trait were made object-safe (e.g. replace the associated types with methods on `&self` that return trait objects) and handled in a type-erased manner inside `quinn`.

Ralith (changes requested):
Thanks, this looks much closer to what I had in mind! This touches a lot of stuff as written, so I haven't yet gone through it in detail, but here's some initial thoughts that might help trim things down besides.

quinn-runtime/src/lib.rs
yu-re-ka: I don't fully understand this proposal, how would the user specify the runtime with this?

quinn-runtime/Cargo.toml
yu-re-ka: As you noticed yourself after writing the initial comment, we can not define it in `quinn` because `quinn-udp` relies on it. Defining it in `quinn-udp` would be weird because it contains things that have nothing to do with udp.
I could change it to put only the AsyncWrappedUdpSocket inside `quinn-udp` and the rest into `quinn`.
But also this seperate crate helps understand the dependencies structure better. The runtime crate depends on tokio and/or async-std/io. quinn and quinn-udp just depend on the runtime crate. Makes it harder to accidentally introduce direct tokio calls into the rest of quinn-udp again.

quinn-runtime/src/lib.rs
yu-re-ka: As the comments explain, on Unix we expect to be able to do more specific operations in order to support ECN. On non-unix the tokio backend can not provide these methods as there is no AsyncFd on non-unix, and the tokio::net::UdpSocket does not provide a way to get a reference to the underlying std::net::UdpSocket (if there is even one?).

quinn-runtime/src/lib.rs
yu-re-ka: Would that mean storing a `Box<dyn AsyncWrappedUdpSocket>` instead of a `RT::AsyncWrappedUdpSocket`?

yu-re-ka: If I try to make AsyncWrappedUdpSocket object-safe, I also have to change the `impl FnOnce ... ` to an `Box<dyn FnOnce ...>`.
I can create a third PR for this strategy, but it has a runtime performance cost (~15%) and I don't think we want that.

yu-re-ka: #1346

quinn-runtime/src/lib.rs
Ralith: > there is no AsyncFd on non-unix

Perhaps something could be implemented in terms of e.g. https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.ready instead?

> the tokio::net::UdpSocket does not provide a way to get a reference to the underlying std::net::UdpSocket

What if we exposed the underlying file descriptor directly instead?

quinn-runtime/src/lib.rs
Ralith: Thanks! You're correct that dynamic dispatch isn't free, but it also isn't very expensive. We must also weigh the cost of impact on compile time (monomorphization undermines separate compilation) and ergonomics (piping type variables around everywhere is a pain). I do not expect that the difference here is measurable, especially compared to the cost of encryption and I/O, but you can run benchmarks to confirm.

Ralith: > I also have to change the impl FnOnce ... to an Box<dyn FnOnce ...>.

I suspect this can be avoided with careful API design. For example, the use of callbacks could be dropped in favor of returning information sufficient to allow the caller to run appropriate code directly.

quinn-runtime/Cargo.toml
Ralith: > I could change it to put only the AsyncWrappedUdpSocket inside quinn-udp and the rest into quinn.

This sounds broadly reasonable to me. Alternatively, it should be possible to refactor quinn-udp to be entirely runtime-agnostic, without requiring traits at all, just primitive I/O operations that the caller is responsible for calling at the right times.

> Makes it harder to accidentally introduce direct tokio calls into the rest of quinn-udp again.

We can address this with CI if necessary. Extra crates means extra maintenance effort, which we must keep to a minimum.

quinn-runtime/src/lib.rs
yu-re-ka: AsRawFd also only exists on Unix. I don't think we can get around this problem easily. The interface we are describing here is completely different for Unix and Non-Unix platforms.

quinn-runtime/src/lib.rs
yu-re-ka: > the use of callbacks could be dropped in favor of returning information sufficient to allow the caller to run appropriate code directly.

This will be hard because _after_ the code is run tokio makes some decisions based on if the recv/send attempt was successful or if it returned WouldBlock. tokio's interface for this is the `guard.try_io()` which also takes a FnOnce.

quinn-runtime/Cargo.toml
yu-re-ka: Implemented in the other PR

quinn-runtime/src/lib.rs
yu-re-ka: Actually I found a way to do this, will push to the other PR shortly

yu-re-ka:
@Ralith please have a look at the current state of #1346 and tell me if it addresses all issues you raised

PR #1346, opened 2022-04-17
async-std support with trait objects

yu-re-ka:
Alternative to #1345


yu-re-ka:
Surprisingly using the async-std/async-io based UdpSocket and Timer code leads to better performance, even though it spawns a background thread:

```
$ cargo run --bin bulk --release 

Client 0 stats:
Overall download stats:

Transferred 1073741824 bytes on 1 streams in 1.86s (549.73 MiB/s)

Stream download metrics:

      │  Throughput   │ Duration 
 AVG  │  549.75 MiB/s │     1.86s
 P0   │  549.50 MiB/s │     1.86s
 P10  │  550.00 MiB/s │     1.86s
 P50  │  550.00 MiB/s │     1.86s
 P90  │  550.00 MiB/s │     1.86s
 P100 │  550.00 MiB/s │     1.86s

$ sed "s/TokioRuntime/AsyncStdRuntime/g" -i bench/src/lib.rs
$ cargo run --bin bulk --release --features runtime-async-std
[...]

Client 0 stats:
Overall download stats:

Transferred 1073741824 bytes on 1 streams in 1.70s (602.03 MiB/s)

Stream download metrics:

      │  Throughput   │ Duration 
 AVG  │  602.25 MiB/s │     1.70s
 P0   │  602.00 MiB/s │     1.70s
 P10  │  602.50 MiB/s │     1.70s
 P50  │  602.50 MiB/s │     1.70s
 P90  │  602.50 MiB/s │     1.70s
 P100 │  602.50 MiB/s │     1.70s
```

Ralith:
> Surprisingly using the async-std/async-io based UdpSocket and Timer code (still running under the tokio runtime otherwise) leads to better performance in the bulk bench, even though it spawns a background thread.

Interesting! How consistent is that across runs?

> We could also try to choose the correct runtime for the user:

I actually really like the look of that helper, though I'm not sure it should be exposed directly. Let's use that to avoid complicating the `Endpoint::client` and `Endpoint::server` helpers (the entire purpose of those is to be simple), but take an explicit argument in `Endpoint::new` for when explicit selection is required.

quinn/examples/client-async-std.rs
```diff
@@ -0,0 +1,161 @@
+//! This example demonstrates an HTTP client that requests files from a server.
```
Ralith, line 1: I'm not sure the fairly trivial differences justify duplicating the maintenance cost of completely duplicating the client/server examples. Maybe a single dedicated, simpler example that just establishes a connection?

Ralith (commented):
Thanks for your continued work on this! I'm liking how it's shaping up.

I do wonder if we could refactor quinn-udp to be plain stupid-simple helper functions on `RawFd`/`RawHandle` and isolate all the runtime abstraction into the `quinn` crate. I don't think this is a hard blocker for me, though.

yu-re-ka:
> Interesting! How consistent is that across runs?

After trying a few more things I think it comes down to the amount of threads the load is distributed across. When using the single-threaded tokio executor (like the bench is), it is bottle-necked on that single thread, and off-loading thread to another thread takes load off the executor thread.
When using a multi-threaded runtime, a few threads (4-6 threads on my machine) do increase performance because it allows the load to be handled by multiple cores, but increasing the thread count even further will result in slower performance because of lots of moving tasks around.

> Let's use that to avoid complicating the Endpoint::client and Endpoint::server helpers (the entire purpose of those is to be simple), but take an explicit argument in Endpoint::new for when explicit selection is required.

Sounds good! I will implement that later.

yu-re-ka:
(Close was accidental. Pressed the wrong button)

parazyd:
@yu-re-ka This still seems to have a hard dependency on the tokio crate. Is that intended?

yu-re-ka:
Yes, that was the request from the maintainers:

> I also don't think tokio's sync primitives are runtime-specific, so those can be left in on the first pass to simplify review.

I tested that it only builds the sync feature of the tokio crate, so you don't get a dependency on any of the runtime/executor code.

yu-re-ka:
I have removed the runtime parameter from Endpoint::client, Endpoint::server (these now use the default mechanism described earlier), and Endpoint::connect, Endpoint::connect_with (these use the runtime the endpoint was created with).

parazyd:
Brilliant work

quinn-udp/src/unix.rs
```diff
@@ -68,11 +67,12 @@ impl UdpSocket {
...
             let last_send_error = &mut self.last_send_error;
-            let mut guard = ready!(self.io.poll_write_ready(cx))?;
-            if let Ok(res) =
-                guard.try_io(|io| send(state, io.get_ref(), last_send_error, transmits))
-            {
-                return Poll::Ready(res);
+            ready!(self.io.poll_write_ready(cx))?;
+            match send(state, self.io.get_ref(), last_send_error, transmits) {
+                Err(e) if e.kind() == io::ErrorKind::WouldBlock => {
+                    self.io.clear_write_ready(cx);
```
Ralith, line 73: What if the socket becomes ready in between the call to `poll_write_ready` and `clear_write_ready`? Could this miss a wakeup? Similarly for reads.

yu-re-ka: First, the clear_write_ready is a no-op on all implementations but tokio+unix (using AsyncFd), so we are only talking about this platform now. async-io doesn't require (or have a way of) clearing readiness events. tokio UdpSocket automatically clears readiness events for us in the call to `try_recv_from`, and it also has no way to manually do that.

The code here is following the instructions from the tokio documentation:

> Indicates to tokio that the file descriptor is no longer ready. The internal readiness flag will be cleared, and tokio will wait for the next edge-triggered readiness notification from the OS.
> It is critical that this function not be called unless your code actually observes that the file descriptor is not ready. Do not call it simply because, for example, a read succeeded; it should be called when a read is observed to block.

When we observe that the file descriptor is not ready (I/O operation returns WouldBlock), we clear the readiness status.
I don't know what would the alternative? This is exactly how tokio documentation describes it and how tokio does it itself internally.

Ralith: Thanks; I'll have to study this in some detail...

yu-re-ka: Your concern is indeed very justified:
https://github.com/microsoft/msquic/issues/2654

quinn/src/runtime/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+    fn reset(self: Pin<&mut Self>, i: Instant);
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<()>;
+}
+
+pub trait Runtime: udp::runtime::Runtime {
+    fn new_timer(&self, i: Instant) -> Pin<Box<dyn AsyncTimer>>;
+    fn spawn(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>);
+}
+
+pub fn make_runtime() -> Box<dyn Runtime> {
```
djc, line 27: I'd prefer not having this automatic selection, instead requiring callers to pass in an `impl Runtime` explicitly.

quinn/src/runtime/async_std_runtime.rs
```diff
@@ -0,0 +1,26 @@
+use super::{AsyncTimer, Runtime};
```
djc, line 1: Style nit: please split the imports as mentioned elsewhere.

djc (commented):
This is looking pretty good, thanks for working on it!

Some comments, mostly stylistic concerns -- hope that's okay.

quinn/src/runtime/mod.rs
yu-re-ka: You'll have to discuss this with @Ralith, who requested to reduce the changes to the user-facing API.

quinn/src/runtime/async_std_runtime.rs
yu-re-ka: I can't infer from the other comments how you would like imports to be structured. Is there some styling guidelines where I can read up on this?

yu-re-ka:
I can see from your comments that you would like things to be changed _somehow_, but there are so many options for how I could interpret these comments. I now just chose any one option, so you will have to make new comments if my interpretations did not match your comment intention - or send me a link to some guideline I can follow.

quinn/src/runtime/mod.rs
Ralith: I think it will be very rare for this not to do the expected thing, and in those cases `Endpoint::new` can be used. I'd rather not complicate every entry point with a detail that most users aren't interested in customizing.

quinn-udp/src/unix.rs
Ralith: Discussed with upstream tokio; this will miss wakeups. In particular, the implementation of `clear_write_ready` in terms of `poll_write_ready` will cause readiness events that occur after the initial `poll_write_ready` to be lost.

Our basic problem is that `AsyncUdpSocket` is trying to be too low-level. Rather than trying to make the methods of `quinn_udp::UdpSocket` abstract over the underlying async primitive, let's try converting the (already object-safe) `quinn_udp::UdpSocket` itself into a trait object, with separate implementations for tokio and async-std. Those separate implementations should then call into shared code to do the OS-specific heavy-lifting. We can then hoist the trait machinery up into the `quinn` crate and leave the shared OS-specific heavy lifting code in quinn_udp, concentrating awareness of async runtimes in one place.

quinn/src/runtime/async_std_runtime.rs
Ralith: Our convention is to split imports into three blocks: `std`, followed by external crates, followed by the local crate. We don't currently have a formal style guide written, though we should; thanks for bearing with us in the mean time.

dignifiedquire:
why was this closed?

yu-re-ka:
I don't have motivation to refactor this again. Anyone feel free to open a new PR and use these changes if it is helpful.

quinn-udp/src/unix.rs
Ralith: I've implemented the approach I outlined above in https://github.com/quinn-rs/quinn/pull/1364.

Ralith:
This was very close to completion, so I took it the last few steps in https://github.com/quinn-rs/quinn/pull/1364.

PR #1350, opened 2022-05-04
Enable UDP GRO

alessandrod:
Hello! Thanks for such a great lib! 🤗

As a way to learn quinn I started looking into the IO code and noticed that while GSO is supported, GRO isn't yet. This PR is my first stab at implementing it. Initial results seem promising:

No GRO:
```
Client 0 stats:
Overall download stats:

Transferred 1073741824 bytes on 1 streams in 2.96s (345.49 MiB/s)

Stream download metrics:

      │  Throughput   │ Duration 
 AVG  │  345.62 MiB/s │     2.96s
 P0   │  345.50 MiB/s │     2.96s
 P10  │  345.75 MiB/s │     2.96s
 P50  │  345.75 MiB/s │     2.96s
 P90  │  345.75 MiB/s │     2.96s
 P100 │  345.75 MiB/s │     2.96s

```

With GRO:
```
 Client 0 stats:
Overall download stats:

Transferred 1073741824 bytes on 1 streams in 2.51s (407.22 MiB/s)

Stream download metrics:

      │  Throughput   │ Duration 
 AVG  │  407.38 MiB/s │     2.52s
 P0   │  407.25 MiB/s │     2.51s
 P10  │  407.50 MiB/s │     2.52s
 P50  │  407.50 MiB/s │     2.52s
 P90  │  407.50 MiB/s │     2.52s
 P100 │  407.50 MiB/s │     2.52s
```

The existing tests pass, so hopefully I'm not benchmarking garbage.

Obviously the PR needs a lot more work - some things that come to mind:

- should this be conditionally enabled based on config? (GSO seems to be enabled unconditionally)
- should the GRO size be configurable? If the read buffers aren't large enough GRO is actually less performant than no GRO
- I'm testing on a 5.14 kernel, we need to find the minimum kernel version that supports GRO

### DIFF
--- a/quinn-udp/src/lib.rs
+++ b/quinn-udp/src/lib.rs
@@ -68,6 +68,7 @@ pub struct RecvMeta {
     pub ecn: Option<EcnCodepoint>,
     /// The destination IP address which was encoded in this datagram
     pub dst_ip: Option<IpAddr>,
+    pub gso_size: Option<usize>,
 }
 
 impl Default for RecvMeta {
@@ -78,6 +79,7 @@ impl Default for RecvMeta {
             len: 0,
             ecn: None,
             dst_ip: None,
+            gso_size: None,
         }
     }
 }
--- a/quinn-udp/src/lib.rs
+++ b/quinn-udp/src/lib.rs
@@ -65,10 +65,10 @@ impl Default for UdpState {
 pub struct RecvMeta {
     pub addr: SocketAddr,
     pub len: usize,
+    pub stride: usize,
     pub ecn: Option<EcnCodepoint>,
     /// The destination IP address which was encoded in this datagram
     pub dst_ip: Option<IpAddr>,
-    pub gso_size: Option<usize>,
 }
 
 impl Default for RecvMeta {
@@ -77,9 +77,9 @@ impl Default for RecvMeta {
         Self {
             addr: SocketAddr::new(Ipv6Addr::UNSPECIFIED.into(), 0),
             len: 0,
+            stride: 0,
             ecn: None,
             dst_ip: None,
-            gso_size: None,
         }
     }
 }
--- a/quinn-udp/src/lib.rs
+++ b/quinn-udp/src/lib.rs
@@ -37,6 +37,7 @@ pub const BATCH_SIZE: usize = imp::BATCH_SIZE;
 #[derive(Debug)]
 pub struct UdpState {
     max_gso_segments: AtomicUsize,
+    gro_segments: usize,
 }
 
 impl UdpState {
@@ -52,6 +53,15 @@ impl UdpState {
     #[inline]
     pub fn max_gso_segments(&self) -> usize {
         self.max_gso_segments.load(Ordering::Relaxed)
+    }
+
+    /// The number of segments to read when GRO is enabled. Used as a factor to
+    /// compute the receive buffer size.
+    ///
+    /// Returns 1 if the platform doesn't support GRO.
+    #[inline]
+    pub fn gro_segments(&self) -> usize {
+        self.gro_segments
     }
 }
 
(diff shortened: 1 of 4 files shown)


quinn-udp/src/lib.rs
```diff
@@ -68,6 +68,7 @@ pub struct RecvMeta {
     pub ecn: Option<EcnCodepoint>,
     /// The destination IP address which was encoded in this datagram
     pub dst_ip: Option<IpAddr>,
+    pub gso_size: Option<usize>,
```
Ralith, line 71: Should this be `gro_size`? Or maybe `stride`?

Does this need to be `Option`, or could we default it to be equal to `len` to make things simpler for the caller?

quinn/src/endpoint.rs
```diff
@@ -346,27 +346,31 @@ impl EndpointInner {
...
-                            .inner
-                            .handle(now, meta.addr, meta.dst_ip, meta.ecn, data)
-                        {
-                            Some((handle, DatagramEvent::NewConnection(conn))) => {
-                                let conn =
-                                    self.connections
-                                        .insert(handle, conn, self.udp_state.clone());
-                                self.incoming.push_back(conn);
+                        for buf in buf[0..meta.len].chunks(meta.gso_size.unwrap_or(meta.len)) {
+                            let data: BytesMut = buf.into();
```
Ralith, line 350: I wonder if it would be more efficient to convert the entire slice to a `BytesMut` up front, then split it with `split_off` or similar, rather than making separate allocations.

Ralith:
This is awesome, thanks!

> should this be conditionally enabled based on config? (GSO seems to be enabled unconditionally)

GSO is used based on `quinn_udp::UdpState::new()`, which detects GSO support at runtime. We should do similar here, and the same "create a dummy socket and setsockopt it" pattern should be applicable.

> should the GRO size be configurable?

We could experimentally determine a reasonable value, and leave customizing it to future work if anyone ever actually wants that.

> I'm testing on a 5.14 kernel, we need to find the minimum kernel version that supports GRO

I think it should be sufficient to see whether the setsockopt succeeds.

quinn-udp/src/lib.rs
alessandrod: I went for stride and made it default to len

quinn/src/endpoint.rs
alessandrod: That was actually what I did at first before switching to `chunks()`, then convinced myself I didn't like the while loop. I've now gone back to doing only one allocation but perf wise it doesn't seem to make a difference (these are all same-sized relatively small buffers so I'm assuming that the allocator is being smart)

alessandrod:
> This is awesome, thanks!
> 
> > should this be conditionally enabled based on config? (GSO seems to be enabled unconditionally)
> 
> GSO is used based on `quinn_udp::UdpState::new()`, which detects GSO support at runtime. We should do similar here, and the same "create a dummy socket and setsockopt it" pattern should be applicable.
> 
> > should the GRO size be configurable?
> 
> We could experimentally determine a reasonable value, and leave customizing it to future work if anyone ever actually wants that.
> 
> > I'm testing on a 5.14 kernel, we need to find the minimum kernel version that supports GRO
> 
> I think it should be sufficient to see whether the setsockopt succeeds.

Ok I've now implemented this. One thing to notice is that since we do `setsockopt` in `UdpSocket::from_std` but `UdpState` lives at another layer, the current implementation does this:

a) `UdpSocket::from_std` calls `init(&socket)` which opportunistically tries to enable `UDP_GRO`, ignoring the case in which setting the opt fails
b) `EndpointRef::new` sizes the recv buffer using `UdpState::gro_segments()`. If setting `UDP_GRO` failed during step a), it'll fail in this step too, therefore GRO will be disabled.

An alternative implementation could be:

- get rid of step a)
- keep step b), and from there do: `if udp_state.gro_segments() > 1 { socket.<new method that enables gro>() }`

quinn/src/endpoint.rs
Ralith: If there's empirically no difference then I'm happy either way.

Ralith (approved):
I imagined we'd pass down a `&UdpState` to `init`, but this seems fine too.

djc:
Awesome, thanks!

PR #1351, opened 2022-05-05
connection: change overly verbose info span to debug

alessandrod:
This moves a super verbose info span to debug. In addition to reducing clutter, it turns out that when running with RUST_LOG=info, this span actually takes significant cpu time


djc (approved):

Ralith (approved):
I don't think this has ever proved to be useful in practice; I honestly wouldn't mind stripping it out entirely.

djc:
Please submit a PR that backports this change to 0.8, and add a separate commit that bumps the version number for the relevant crate(s), and we will be happy to publish it to crates.io.

PR #1357, opened 2022-05-12
Named futures + replace NewConnection with Connection methods

Ralith:
Experimenting with a return to named 'static futures, motivated by downstream use cases. See e.g. https://github.com/hyperium/h3/pull/80.

Open questions:
- Which interfaces specifically need to be named?
- Is 'static needed, or just names?
- Are GATs + TAIT close enough that none of this matters?

Also included is an idea to drop `NewConnection` in favor of unobtrusive async methods on `Connection`, bundled because they would have to be rewritten to match this PR's approach if it move forwards. I'd like feedback on this change before I invest the effort to actually remove `NewConnection` and update all the examples/docs.

### DIFF
--- a/quinn/Cargo.toml
+++ b/quinn/Cargo.toml
@@ -39,6 +39,7 @@ futures-io = { version = "0.3.19", optional = true }
 # Implements futures::Stream for async streams such as `Incoming`
 futures-core = { version = "0.3.19", optional = true }
 rustc-hash = "1.1"
+pin-project-lite = "0.2"
 proto = { package = "quinn-proto", path = "../quinn-proto", version = "0.8", default-features = false }
 rustls = { version = "0.20.3", default-features = false, features = ["quic"], optional = true }
 thiserror = "1.0.21"
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -12,10 +12,11 @@ use std::{
 
 use crate::runtime::{AsyncTimer, Runtime};
 use bytes::Bytes;
+use pin_project_lite::pin_project;
 use proto::{ConnectionError, ConnectionHandle, ConnectionStats, Dir, StreamEvent, StreamId};
 use rustc_hash::FxHashMap;
 use thiserror::Error;
-use tokio::sync::{mpsc, oneshot, Notify};
+use tokio::sync::{futures::Notified, mpsc, oneshot, Notify};
 use tracing::debug_span;
 use udp::UdpState;
 
@@ -362,6 +363,30 @@ impl Connection {
         }
     }
 
+    /// Accept the next incoming uni-directional stream
+    pub fn accept_uni(&self) -> AcceptUni<'_> {
+        AcceptUni {
+            conn: &self.0,
+            notify: self.0.shared.stream_incoming[Dir::Uni as usize].notified(),
+        }
+    }
+
+    /// Accept the next incoming bidirectional stream
+    pub fn accept_bi(&self) -> AcceptBi<'_> {
+        AcceptBi {
+            conn: &self.0,
+            notify: self.0.shared.stream_incoming[Dir::Bi as usize].notified(),
+        }
+    }
+
+    /// Receive an application datagram
+    pub fn read_datagram(&self) -> ReadDatagram<'_> {
+        ReadDatagram {
+            conn: self.0.clone(),
+            notify: self.0.shared.datagrams.notified(),
+        }
+    }
+
     /// Wait for the connection to be closed for any reason
     ///
     /// Despite the return type's name, closed connections are often not an error condition at the
@@ -688,6 +713,103 @@ impl futures_core::Stream for IncomingBiStreams {
     }
 }
 
+pin_project! {
+    /// Future produced by [`Connection::accept_uni`]
+    pub struct AcceptUni<'a> {
+        conn: &'a ConnectionRef,
+        #[pin]
+        notify: Notified<'a>,
+    }
+}
+
+impl Future for AcceptUni<'_> {
+    type Output = Result<RecvStream, ConnectionError>;
+
+    fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        let (conn, id, is_0rtt) = ready!(poll_accept(ctx, this.conn, this.notify, Dir::Uni))?;
+        Poll::Ready(Ok(RecvStream::new(conn, id, is_0rtt)))
+    }
+}
+
+pin_project! {
+    /// Future produced by [`Connection::accept_bi`]
+    pub struct AcceptBi<'a> {
+        conn: &'a ConnectionRef,
+        #[pin]
+        notify: Notified<'a>,
+    }
+}
+
+impl Future for AcceptBi<'_> {
+    type Output = Result<(SendStream, RecvStream), ConnectionError>;
+
+    fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        let (conn, id, is_0rtt) = ready!(poll_accept(ctx, this.conn, this.notify, Dir::Bi))?;
+        Poll::Ready(Ok((
+            SendStream::new(conn.clone(), id, is_0rtt),
+            RecvStream::new(conn, id, is_0rtt),
+        )))
+    }
+}
+
+fn poll_accept<'a>(
+    ctx: &mut Context<'_>,
+    conn: &'a ConnectionRef,
+    mut notify: Pin<&mut Notified<'a>>,
+    dir: Dir,
+) -> Poll<Result<(ConnectionRef, StreamId, bool), ConnectionError>> {
+    let mut state = conn.state.lock("poll_accept");
+    // Check for incoming streams before checking `state.error` so that already-received streams,
+    // which are necessarily finite, can be drained from a closed connection.
+    if let Some(id) = state.inner.streams().accept(dir) {
+        let is_0rtt = state.inner.is_handshaking();
+        state.wake(); // To send additional stream ID credit
+        drop(state); // Release the lock so clone can take it
+        return Poll::Ready(Ok((conn.clone(), id, is_0rtt)));
+    } else if let Some(ref e) = state.error {
+        return Poll::Ready(Err(e.clone()));
+    }
+    loop {
+        match notify.as_mut().poll(ctx) {
+            // `state` lock ensures we didn't race with readiness
+            Poll::Pending => return Poll::Pending,
+            // Spurious wakeup, get a new future
+            Poll::Ready(()) => notify.set(conn.shared.stream_incoming[dir as usize].notified()),
+        }
+    }
+}
+
+pin_project! {
+    /// Future produced by [`Connection::read_datagram`]
+    pub struct ReadDatagram<'a> {
+        conn: ConnectionRef,
+        #[pin]
+        notify: Notified<'a>,
+    }
+}
+
+impl Future for ReadDatagram<'_> {
+    type Output = Result<Bytes, ConnectionError>;
+    fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        let mut conn = this.conn.state.lock("ReadDatagram::poll");
+        // Check for buffered datagrams before checking `state.error` so that already-received
+        // datagrams, which are necessarily finite, can be drained from a closed connection.
+        if let Some(x) = conn.inner.datagrams().recv() {
+            Poll::Ready(Ok(x))
+        } else if let Some(ref e) = conn.error {
+            Poll::Ready(Err(e.clone()))
+        } else if this.notify.poll(ctx).is_pending() {
+            // `conn` lock ensures we don't race with readiness
+            Poll::Pending
+        } else {
+            unreachable!("ReadDatagram notified with no datagrams pending");
+        }
+    }
+}
+
 /// Stream of unordered, unreliable datagrams sent by the peer
 #[derive(Debug)]
 pub struct Datagrams(ConnectionRef);
@@ -809,6 +931,8 @@ pub struct ConnectionInner {
 #[derive(Debug, Default)]
 pub(crate) struct Shared {
     stream_opening: [Notify; 2],
+    stream_incoming: [Notify; 2],
+    datagrams: Notify,
     closed: Notify,
 }
 
@@ -932,16 +1056,19 @@ impl State {
                     }
                 }
                 Stream(StreamEvent::Opened { dir: Dir::Uni }) => {
+                    shared.stream_incoming[Dir::Uni as usize].notify_waiters();
                     if let Some(x) = self.incoming_uni_streams_reader.take() {
                         x.wake();
                     }
                 }
                 Stream(StreamEvent::Opened { dir: Dir::Bi }) => {
+                    shared.stream_incoming[Dir::Bi as usize].notify_waiters();
                     if let Some(x) = self.incoming_bi_streams_reader.take() {
                         x.wake();
                     }
                 }
                 DatagramReceived => {
+                    shared.datagrams.notify_waiters();
                     if let Some(x) = self.datagram_reader.take() {
                         x.wake();
                     }
@@ -1047,12 +1174,15 @@ impl State {
         }
         shared.stream_opening[Dir::Uni as usize].notify_waiters();
         shared.stream_opening[Dir::Bi as usize].notify_waiters();
+        shared.stream_incoming[Dir::Uni as usize].notify_waiters();
+        shared.stream_incoming[Dir::Bi as usize].notify_waiters();
         if let Some(x) = self.incoming_uni_streams_reader.take() {
             x.wake();
         }
         if let Some(x) = self.incoming_bi_streams_reader.take() {
             x.wake();
         }
+        shared.datagrams.notify_waiters();
         if let Some(x) = self.datagram_reader.take() {
             x.wake();
         }
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -930,7 +930,8 @@ impl ConnectionInner {
                     }
                 }
                 Stream(StreamEvent::Available { dir }) => {
-                    self.stream_opening[dir as usize].notify_one();
+                    // Might mean any number of streams are ready, so we wake up everyone
+                    self.stream_opening[dir as usize].notify_waiters();
                 }
                 Stream(StreamEvent::Finished { id }) => {
                     if let Some(finishing) = self.finishing.remove(&id) {
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -96,7 +96,7 @@ impl Connecting {
     pub fn into_0rtt(mut self) -> Result<(NewConnection, ZeroRttAccepted), Self> {
         // This lock borrows `self` and would normally be dropped at the end of this scope, so we'll
         // have to release it explicitly before returning `self` by value.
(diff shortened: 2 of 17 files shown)


quinn/src/notify.rs
```diff
@@ -0,0 +1,89 @@
...
+use slab::Slab;
+
+/// Broadcasts an event to any number of waiters
+#[derive(Clone)]
+pub struct NotifyOwned {
+    shared: Arc<Mutex<Shared>>,
+}
+
+struct Shared {
+    version: u64,
```
djc, line 13: I feel like calling this `epoch` might better evoke how it's used? Perhaps the field here could get some docs on how/why it's supposed to be used?

quinn/src/notify.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+pub struct Waiter {
+    shared: Arc<Mutex<Shared>>,
+    state: Option<State>,
+}
+
+impl Waiter {
+    /// Register to be woken by the next `notify_all`. Must be called before readiness could occur,
```
djc, line 54: Short single no-period first line of docs?

djc:
Sorry for the slow review...

quinn/src/connection.rs
```diff
@@ -364,6 +366,30 @@ impl Connection {
         }
     }
 
+    /// Accept the next incoming uni-directional stream
```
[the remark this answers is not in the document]
Ralith: > Is this intended to replace the IncomingUniStreams/IncomingBiStreams stuff?

Right. My thinking is that this makes the API a bit simpler, primarily since `NewConnection` can be dispensed with entirely. On the other hand, it makes it much more difficult to use off-the-shelf stream combinators like [`buffer_unordered`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffer_unordered). For the most part I think combinators have rightly fallen out of fashion compared to imperative-style `async` blocks, but the constructs aren't always exactly equivalent; e.g. emulating `buffer_unordered` might require combining `tokio::spawn` with an async semaphore, which would not be as graceful about memory reuse, but saves a layer of indirection on the wake-up path.

> Should we deprecate that API?

In principle this would be preferred, but deprecating `NewConnection` gracefully seems difficult, since it's central to accepting/creating new connections; e.g. `Endpoint::connect` itself would need to be deprecated, which consumes an awfully appealing part of the namespace.

quinn/src/notify.rs
```diff
@@ -0,0 +1,89 @@
+use std::sync::{Arc, Mutex};
+use std::task::{Context, Waker};
+
+use slab::Slab;
```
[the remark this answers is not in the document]
Ralith: A slab is required so that `Waiter`s can reliably and efficiently update and remove their associated `Waker`s on poll and drop respectively without invalidating eachothers' indices. If wakers can't be updated, then we'll miss wakeups after e.g. tokio workstealing, and if they can't be removed, then memory use grows without bound in some use patterns; consider e.g. a `loop { select! { ... } }` which constructs new `Waiter`s on each cycle but usually selects a different branch.

quinn/src/notify.rs
```diff
@@ -0,0 +1,89 @@
...
+
+/// Broadcasts an event to any number of waiters
+#[derive(Clone)]
+pub struct NotifyOwned {
```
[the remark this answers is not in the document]
Ralith: You recall correctly. Tokio's `Notify` is appealing, but because `Notified` borrows from the `Notify` it cannot be easily used to implement `'static` futures, which I hypothesized might be useful downstream, based on https://github.com/hyperium/h3/pull/80 and https://github.com/libp2p/rust-libp2p/pull/2289#issuecomment-1119872606. However, I haven't been able to reach any of the h3 folks to discuss this in recent months, so some question remains as to whether it's really useful. We should prefer `tokio::Notify` if we don't actually need `'static`.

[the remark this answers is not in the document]
elenaf9: Enabling `'static` futures would definitely be desirable from a user perspective.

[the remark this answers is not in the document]
Ralith: Thanks for the feedback! Can you provide a list of the specific futures you'd find that useful for? For example, would you want it for the read/write futures as well?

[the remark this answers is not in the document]
elenaf9: Actually, I might have to take back what I said above. For new outgoing connections we need `'static` futures (which quinn already provides with the `Connecting` future). Futures for new streams on the other hand are driven in the same struct that owns the `Connection`, so I don't think the futures need to be 'static. However, I did not look into it in detail yet; as written in https://github.com/hyperium/h3/pull/80 it could still get complicated with the lifetimes. 
I'm sorry that I can not give more info at the current point; we are currently undergoing some internal refactoring of how we handle transports and multiplexing, and things are still in discussion. 
Either way, until #1219 is merged we'll probably stick with using `quinn_proto` anyway. So I don't think our use-case has to be considered right now, and thus using `Tokio's` Notify is sufficient. I am happy to do a PR to re-add `NotifyOwned` if it turns out that it's needed from our side. But of course I can't speak for the h3 folks and other users.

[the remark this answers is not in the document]
Ralith: Thanks for elaborating! I'll probably leave this open for now in to gather further feedback. I think we can release 0.9 without it and roll back to named futures without breaking, even.

quinn/src/notify.rs
Ralith: Mooted by adoption of `tokio::sync::Notify`.

quinn/src/notify.rs
Ralith: Mooted by adoption of `tokio::sync::Notify`.

quinn/src/notify.rs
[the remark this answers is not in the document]
Ralith: Went with tokio's `Notify` in the end because GATs are stabilizing and we haven't gotten any feedback clearly requesting `'static` regardles.

quinn/src/notify.rs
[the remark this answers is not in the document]
Ralith: Mooted by adoption of `tokio::sync::Notify`.

Ralith:
Still TODO: Actually getting rid of `NewConnection`, unless there's a compelling reason to keep it around.

Ralith:
That seems to have panned out to a pleasant simplification in the tests/examples. Less `unwrap` and fewer intermediate variables all 'round.

quinn/src/connection.rs
```diff
@@ -1012,7 +1037,7 @@ impl ConnectionInner {
     }
 
     /// Used to wake up all blocked futures when the connection becomes closed for any reason
-    fn terminate(&mut self, reason: ConnectionError) {
+    fn terminate(&mut self, shared: &Shared, reason: ConnectionError) {
```
djc, line 1018: I feel like inverting the argument order here would be more consistent with our style?

quinn/src/connection.rs
```diff
@@ -1043,18 +1068,18 @@ impl ConnectionInner {
         for (_, waker) in self.stopped.drain() {
             waker.wake();
         }
-        self.closed.notify_waiters();
+        shared.closed.notify_waiters();
     }
 
-    fn close(&mut self, error_code: VarInt, reason: Bytes) {
+    fn close(&mut self, shared: &Shared, error_code: VarInt, reason: Bytes) {
```
djc, line 1046: And here.

quinn/src/connection.rs
```diff
@@ -764,35 +775,47 @@ impl ConnectionRef {
...
 }
 
+#[derive(Debug)]
 pub struct ConnectionInner {
+    pub(crate) state: Mutex<State>,
+    pub(crate) shared: Shared,
+}
+
+#[derive(Debug)]
+pub(crate) struct Shared {
```
djc, line 798: Should we think of a more specific name for this? `Notifiers`? `Wakers`?

quinn/src/connection.rs
```diff
@@ -764,35 +775,47 @@ impl ConnectionRef {
...
+    stream_opening: [Notify; 2],
+    closed: Notify,
+}
+
+pub(crate) struct State {
```
djc, line 805: Maybe a more specific name? `ConnectionState`, or `ConnState`? (I think variable names can probably still be `state`, but still think this would help clarify.)

quinn/src/connection.rs
```diff
@@ -730,31 +737,35 @@ impl ConnectionRef {
...
-            finishing: FxHashMap::default(),
-            stopped: FxHashMap::default(),
-            closed: Arc::new(Notify::new()),
-            error: None,
-            ref_count: 0,
-            udp_state,
-            runtime,
-        })))
+        Self(Arc::new(ConnectionInner {
+            state: Mutex::new(State {
```
djc, line 727: Maybe `State` or `ConnectionInner` could do with an associated `new()` function (as a separate change), what do you think?

quinn/Cargo.toml
```diff
@@ -46,6 +46,7 @@ tracing = "0.1.10"
 tokio = { version = "1.0.1", features = ["sync"] }
 udp = { package = "quinn-udp", path = "../quinn-udp", version = "0.2" }
 webpki = { version = "0.22", default-features = false, optional = true }
+pin-project-lite = "0.2"
```
djc, line 49: Keep this in alphabetic order?

quinn/src/connection.rs
```diff
@@ -362,6 +363,30 @@ impl Connection {
         }
     }
 
+    /// Accept the next incoming uni-directional stream
+    pub fn accept_uni(&self) -> AcceptUni<'_> {
```
djc, line 293: So in terms of API design, I wonder if this is the most natural expression of what people need here. The alternative I have in mind would be an `accept()` that yields an enum with `Uni` or `Bi` newtype variants.

quinn/src/connection.rs
```diff
@@ -688,6 +713,99 @@ impl futures_core::Stream for IncomingBiStreams {
...
+    pub struct AcceptUni<'a> {
+        conn: &'a ConnectionRef,
+        #[pin]
+        notify: Notified<'a>,
+    }
+}
+
+impl Future for AcceptUni<'_> {
+    type Output = Result<RecvStream, ConnectionError>;
+    fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
```
djc, line 629: I like to put an empty line between associated type and `fn` definitions, thoughts?

quinn/src/connection.rs
```diff
@@ -688,6 +713,99 @@ impl futures_core::Stream for IncomingBiStreams {
...
+}
+
+fn poll_accept<'a>(
+    ctx: &mut Context<'_>,
+    conn: &'a ConnectionRef,
+    mut notify: Pin<&mut Notified<'a>>,
+    dir: Dir,
+) -> Poll<Result<(ConnectionRef, StreamId, bool), ConnectionError>> {
+    let mut state = conn.state.lock("poll_accept");
+    if let Some(id) = state.inner.streams().accept(dir) {
```
djc, line 664: Should we check the `state.error` before `accept()`?

quinn/src/connection.rs
```diff
@@ -688,6 +713,99 @@ impl futures_core::Stream for IncomingBiStreams {
...
+        Poll::Ready(Err(e.clone()))
+    } else {
+        loop {
+            if notify.as_mut().poll(ctx).is_pending() {
+                // `conn` lock ensures we don't race with readiness
+                return Poll::Pending;
+            } else {
+                // Spurious wakeup, get a new future
+                notify.set(conn.shared.stream_incoming[dir as usize].notified());
+            }
```
djc, line 679: If the `notify.set()` call fits on a single line, this might read better as a `match`.

quinn/src/connection.rs
```diff
@@ -688,6 +713,99 @@ impl futures_core::Stream for IncomingBiStreams {
...
+    let mut state = conn.state.lock("poll_accept");
+    if let Some(id) = state.inner.streams().accept(dir) {
+        let is_0rtt = state.inner.is_handshaking();
+        state.wake(); // To send additional stream ID credit
+        drop(state); // Release the lock so clone can take it
+        Poll::Ready(Ok((conn.clone(), id, is_0rtt)))
+    } else if let Some(ref e) = state.error {
+        Poll::Ready(Err(e.clone()))
+    } else {
+        loop {
```
djc, line 672: I'd probably opt to `return` explicitly from the previous blocks, so we can reduce the indentation by one level for this loop.

quinn/src/connection.rs
```diff
@@ -688,6 +713,99 @@ impl futures_core::Stream for IncomingBiStreams {
...
+        notify: Notified<'a>,
+    }
+}
+
+impl Future for ReadDatagram<'_> {
+    type Output = Result<Bytes, ConnectionError>;
+    fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        let mut conn = this.conn.state.lock("ReadDatagram::poll");
+        if let Some(x) = conn.inner.datagrams().recv() {
```
djc, line 698: Should we check `error` first? If not, why not?

quinn/src/connection.rs
```diff
@@ -623,6 +604,72 @@ impl Connection {
...
+}
+
+fn poll_open<'a>(
+    ctx: &mut Context<'_>,
+    conn: &'a ConnectionRef,
+    mut notify: Pin<&mut Notified<'a>>,
+    dir: Dir,
+) -> Poll<Result<(ConnectionRef, StreamId, bool), ConnectionError>> {
+    let mut state = conn.state.lock("poll_open");
+    if let Some(ref e) = state.error {
```
djc, line 599: Same comments here as on the `accept` side...

djc (commented):
Nice work, and actually -100 lines!

quinn/src/connection.rs
Ralith: My thinking is that we might move other data for which unique access isn't required into here to reduce synchronization in the future. Could always rename it then, of course.

quinn/src/connection.rs
Ralith: I think it's usually a mistake to repeat the module name in a type unless that type is reexported from a higher level. It duplicates context the module itself already establishes.

quinn/src/connection.rs
Ralith: Interesting thought. I think this form should be our null hypothesis, since it corresponds closely to what's in `main`.

One upside to the current form is that it makes it easy to implement an application protocol that assigns special significance to e.g. the first unidirectional stream. I often use this pattern for a "hello" message, which must be processed before any other work can be done, which would require bidirectional streams to be manually buffered if a single method was used. I would probably redefine my application protocols to require that no bidirectional streams be opened before the first unidirectional stream if we went that way, which is an additional constraint.

Conversely, a combined method makes it easier to define application protocols that handle both stream types uniformly. I haven't encountered any cases that require that myself; do you know of any?

We could actually support both just fine (the semantics of using them simultaneously wouldn't be any weirder than those of calling `accept_uni` concurrently) but it's probably better to keep the interface from growing any more than absolutely necessary.

quinn/src/connection.rs
Ralith: We should not. This ordering allows the application to drain any buffered incoming streams on a closed connection. A closed connection cannot receive additional streams, so that's guaranteed to terminate in a timely manner, and it reduces the amount of data lost in the event of an error.

quinn/src/connection.rs
Ralith: As above.

quinn/src/connection.rs
Ralith: Agreed.

quinn/src/connection.rs
Ralith: I'd rather not increase the number of places that the eight arguments here get passed through, and which must be updated for any change to them. Did realize we can `Default` `Shared` for now, though!

quinn/src/connection.rs
djc: > Conversely, a combined method makes it easier to define application protocols that handle both stream types uniformly. I haven't encountered any cases that require that myself; do you know of any?

I don't know of any. It seems interesting to me that a single `accept()` is (a) maybe more familiar for TCP users and (b) then forces application developers to decide how to deal with streams of the other type, rather than letting them potentially accumulate inside the connection.

quinn/src/connection.rs
djc: So maybe add a comment about why this is set up the way it is?

quinn/src/connection.rs
```diff
@@ -584,139 +542,172 @@ impl Connection {
...
-#[cfg(feature = "futures-core")]
-impl futures_core::Stream for IncomingUniStreams {
-    type Item = Result<RecvStream, ConnectionError>;
-
-    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
-        self.poll(cx)
+pin_project! {
+    /// Future produced by [`Connection::open_bi`]
+    pub struct OpenBi<'a> {
+        conn: &'a ConnectionRef,
```
kpp, line 573: This will make it harder to use. For example, I will have to make a self-ref struct, however in the previous version I didn't have to:

```
pub struct QuicMuxer {
    connection: quinn::Connection,
    incoming: quinn::IncomingBiStreams,
    outgoing: Option<quinn::OpenBi>,
}
```

Ralith: That is why I asked a month ago at https://github.com/libp2p/rust-libp2p/pull/2801#issuecomment-1229357955 which futures you wanted to be `'static`.

This form avoids bit of indirection and allocation, and is able to reuse an existing implementation of a complex synchronization structure, so I'd like to make it work. What's the motivation for storing an `OpenBi`? Could you drop in `Box::pin(async move { conn.open_bi().await })` instead (until TAIT stabilizes)?

kpp: > What's the motivation for storing an OpenBi

We got a poll_ based API.

> Could you drop in

I will figure that out tomorrow.

quinn/src/connection.rs
Ralith: I think the key here is that they *don't* accumulate. Each stream type is flow-controlled independently, so accepting only a single type safely backpressures the sender. A single method makes this impossible to take advantage of, and I think is therefore a disadvantage, although I agree it's a bit more familiar.

quinn/src/connection.rs
Ralith: Done.

quinn/src/connection.rs
Ralith: Added comment.

quinn/src/connection.rs
kpp: Well, no.

Box::pin gives `Pin<Box<dyn Future<Output = Result<(quinn::SendStream, quinn::RecvStream), quinn::ConnectionError>>>>`, which is not Send:

```
 `(dyn futures::Future<Output = Result<(quinn::send_stream::SendStream, quinn::recv_stream::RecvStream), quinn_proto::connection::ConnectionError>> + 'static)` cannot be sent between threads safely
```

Ralith: You need to add `+ Send` to the trait object type.

kpp: Thank you.

Ralith: To be clear, does that thereby address your requirements?

kpp: Yes

Ralith: Excellent, thanks for the feedback!

quinn/src/connection.rs
djc: That's a good argument!

djc (approved):

PR #1360, opened 2022-05-29
[README] update talk slides link

wmmc88:
https://dirkjan.ochtman.nl/files/quic-future-in-rust.pdf results in a 404


djc:
Thanks! Not sure how relevant that content still is, it's pretty old at this point...

djc (approved):

PR #1364, opened 2022-06-03
Abstract runtime support

### LINKED ISSUE #502: async_std support

Ralith:
Fixes #502. Draws very heavily on prior work in https://github.com/quinn-rs/quinn/pull/1346. I haven't intensively reviewed all of the inherited code yet, and perhaps there's opportunity to break things up further.

@yu-re-ka I'd like your review on this, if you don't mind? Especially the async-std stuff which I was mostly guessing on.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -19,25 +19,6 @@ type IpTosTy = libc::c_uchar;
 #[cfg(not(target_os = "freebsd"))]
 type IpTosTy = libc::c_int;
 
-fn only_v6(sock: &SockRef<'_>) -> io::Result<bool> {
-    let raw_fd = sock.as_raw_fd();
-    let mut val: libc::c_int = 0;
-    let mut len = mem::size_of::<libc::c_int>() as libc::socklen_t;
-    let res = unsafe {
-        libc::getsockopt(
-            raw_fd,
-            libc::IPPROTO_IPV6,
-            libc::IPV6_V6ONLY,
-            &mut val as *mut _ as *mut _,
-            &mut len,
-        )
-    };
-    match res {
-        -1 => Err(io::Error::last_os_error()),
-        _ => Ok(val != 0),
-    }
-}
-
 /// Tokio-compatible UDP socket with some useful specializations.
 ///
 /// Unlike a standard tokio UDP socket, this allows ECN bits to be read and written on some
@@ -107,7 +88,7 @@ fn init(io: SockRef<'_>) -> io::Result<()> {
     let is_ipv4 = addr.family() == libc::AF_INET as libc::sa_family_t;
 
     // macos and ios do not support IP_RECVTOS on dual-stack sockets :(
-    if is_ipv4 || ((!cfg!(any(target_os = "macos", target_os = "ios"))) && !only_v6(&io)?) {
+    if is_ipv4 || ((!cfg!(any(target_os = "macos", target_os = "ios"))) && !io.only_v6()?) {
         let on: libc::c_int = 1;
         let rc = unsafe {
             libc::setsockopt(


Ralith:
Thanks! Propagating UDP errors is possible, but we're not aware of any cases where it's actually appropriate to do so, and in some cases it actually exposes you to denial-of-service attacks.

yu-re-ka (dismissed):

DemiMarie:
> Thanks! Propagating UDP errors is possible, but we're not aware of any cases where it's actually appropriate to do so, and in some cases it actually exposes you to denial-of-service attacks.

Is that true in all cases?  At the very least I would like to be able to log these errors, as otherwise a firewall blocking traffic is indistinguishable from a timeout.  Also, some errors (such as EPERM on Linux) are a result of the kernel blocking the traffic deliberately and so are not temporary.

Ralith:
I have personally observed EPERM on Linux manifesting transiently. This is also beyond the scope of this PR.

quinn/src/endpoint.rs
```diff
@@ -342,8 +349,17 @@ impl EndpointInner {
...
-            match self.socket.poll_recv(cx, &mut iovs, &mut metas) {
-                Poll::Ready(Ok(msgs)) => {
+            match self.socket.poll_recv_ready(cx) {
+                Poll::Pending => break,
+                Poll::Ready(Err(e)) => return Err(e),
+                Poll::Ready(Ok(())) => {}
+            }
+            let socket_state = &mut self.socket_state;
+            let socket = &self.socket;
+            match socket.try_io(Interest::READABLE, || {
```
djc: Style nit: would like to avoid multi-line scrutinees.

Thought, I think 2021 edition is now in scope, so can we avoid all the borrow splitting before passing state into a closure?

quinn/src/endpoint.rs
```diff
@@ -416,20 +433,30 @@ impl EndpointInner {
...
-                .poll_send(&self.udp_state, cx, self.outgoing.as_slices().0)
-            {
-                Poll::Ready(Ok(n)) => {
+            match self.socket.poll_send_ready(cx) {
+                Poll::Pending => break Ok(false),
+                Poll::Ready(Err(e)) => return Err(e),
+                Poll::Ready(Ok(())) => {}
+            }
+            // Split borrows
+            let socket_state = &mut self.socket_state;
```
djc: (Same here.)

quinn/src/endpoint.rs
```diff
@@ -349,17 +383,8 @@ impl EndpointInner {
             });
         let mut iovs = unsafe { iovs.assume_init() };
         loop {
-            match self.socket.poll_recv_ready(cx) {
```
djc: It feels weird that this code got added in the previous commit and is removed here. How does the intermediate state make sense?

quinn/src/endpoint.rs
```diff
@@ -433,30 +458,20 @@ impl EndpointInner {
...
-            // Split borrows
-            let socket_state = &mut self.socket_state;
-            let socket = &self.socket;
-            let udp_state = &self.udp_state;
-            let transmits = self.outgoing.as_slices().0;
-            match socket.try_io(Interest::WRITABLE, || {
-                socket_state.send(socket.into(), udp_state, transmits)
-            }) {
-                Ok(n) => {
+            match self
```
djc, line 461: Style: multi-line scrutinee.

quinn/src/endpoint.rs
Ralith: The logic is initially moved out of quinn-udp to make that crate async-agnostic in the previous commit, then pushed down into the runtime abstraction in this commit. I think decoupling those steps is a net win to separate those two concerns, even though it requires this code to move around a bit.

quinn/src/endpoint.rs
Ralith: Oh yeah, forgot that was a thing now. Fixed.

djc (approved):
I forget, did we review benchmark impact for these changes?

yu-re-ka:
fe93a681081952bcb2515c791bdf3204d0061139

```
  Time (mean ± σ):      1.387 s ±  0.065 s    [User: 1.689 s, System: 0.767 s]
  Range (min … max):    1.236 s …  1.532 s    50 runs
```

59285257dfe3161d9c04930d9db2fb3a6a444463

```
  Time (mean ± σ):      1.384 s ±  0.048 s    [User: 1.683 s, System: 0.772 s]
  Range (min … max):    1.242 s …  1.495 s    50 runs
```

elenaf9:
Hi, would you mind releasing a new version of `quinn-udp` that includes this change? 
We are currently using `quinn-proto` with `UdpSocket` from `std::net`. Would love to switch to `quinn-udp` sockets for the ECN support, however since we support multiple runtimes (async-std and tokio) we depend on the changes of this PR.

Ralith:
Sure! I've published quinn-udp 0.2.

elenaf9:
Thanks!

PR #1366, opened 2022-06-10
Fixes on bench lib for 32-bit platforms.

matszpk:
Fixes on bench lib for 32-bit platforms. Change upload_size and download_size to u64.


djc (approved):

djc:
Thanks!

PR #1379, opened 2022-07-01
Allow initial maximum UDP payload size to be configured

Ralith:
Motivated by prospective user request. Also potentially useful for benchmark experimentation.


TimonPost (approved):
Looks good! And also a good comment to highlight potential issues.

djc (approved):

djc:
Hmm, what is pulling in time?

Ralith:
Looks like rcgen and tracing-subscriber. Is it time to bump MSRV anyway? Merging regardless since preexisting issue.

djc:
Yeah, I think it's become too hard to avoid 1.56.

PR #1383, opened 2022-07-02
Disable IP fragmentation on Windows

Ralith:
~~Depends on #1364.~~

### DIFF
--- a/quinn-udp/src/windows.rs
+++ b/quinn-udp/src/windows.rs
@@ -1,0 +1,156 @@
+use std::{
+    io::{self, IoSliceMut},
+    mem,
+    os::windows::io::AsRawSocket,
+    time::Instant,
+};
+
+use proto::Transmit;
+use windows_sys::Win32::Networking::WinSock;
+
+use super::{log_sendmsg_error, RecvMeta, UdpSockRef, UdpState, IO_ERROR_LOG_INTERVAL};
+
+/// QUIC-friendly UDP interface for Windows
+#[derive(Debug)]
+pub struct UdpSocketState {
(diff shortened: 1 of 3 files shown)


quinn-udp/src/windows.rs
```diff
@@ -0,0 +1,160 @@
...
+                // recurring on the next call.
+                Err(_) if sent != 0 => return Ok(sent),
+                Err(e) => {
+                    if e.kind() == io::ErrorKind::WouldBlock {
+                        return Err(e);
+                    }
+
+                    // Other errors are ignored, since they will ususally be handled
+                    // by higher level retransmits and timeouts.
+                    // - PermissionDenied errors have been observed due to iptable rules.
```
DemiMarie, line 109: `s/iptable/firewall/`?

quinn-udp/src/windows.rs
```diff
@@ -0,0 +1,160 @@
...
+                    //   Those are not fatal errors, since the
+                    //   configuration can be dynamically changed.
+                    // - Destination unreachable errors have been observed for other
```
DemiMarie, line 112: Looks like a truncated comment.

quinn-udp/src/windows.rs
Ralith: Deleted the Linux-specific details here entirely.

djc (approved):
Seems okay. Is there something I can do to get git to properly notice that `windows.rs` is a copy, so it's more obvious what changed here relative to the other implementations?

Ralith:
Not that I'm aware of, unfortunately.

PR #1384, opened 2022-07-15
Add data limits settings to connections

### LINKED ISSUE #1342: Per-connection data limits
I believe these are the knobs for data limits:

https://docs.rs/quinn/latest/quinn/struct.TransportConfig.html#method.stream_receive_window
https://docs.rs/quinn/latest/quinn/struct.TransportConfig.html#method.receive_window

Would like to be able to set these per-connection.

Also, would be great to release this: https://github.com/quinn-rs/quinn/pull/1315

I suppose if a client violates these settings, the connection is just dropped and it's up to the implementer to reject the next connection attempt from said user if so desired, does that sound right?

inflation:
Fix #1342

Although it's seemingly straightforward, I'm not sure `receive_window` is actually used to limit the data. There is a `received_max_data()` but it can only increase `max_data`, and I'm not seeing any reasoning why it can't be set to lower values. Anyone familiar with it are welcome to shed some light.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1108,6 +1108,28 @@ impl Connection {
         self.streams.set_max_concurrent(dir, count);
     }
 
+    /// Maximum number of bytes the peer may transmit across all streams of a connection before
+    /// becoming blocked.
+    ///
+    /// This should be set to at least the expected connection latency multiplied by the maximum
+    /// desired throughput. Larger values can be useful to allow maximum throughput within a
+    /// stream while another is blocked.
+    pub fn set_receive_window(&mut self, value: VarInt) {
+        self.streams.set_receive_window(value);
+    }
+
+    /// Maximum number of bytes the peer may transmit without acknowledgement on any one stream
+    /// before becoming blocked.
+    ///
+    /// This should be set to at least the expected connection latency multiplied by the maximum
+    /// desired throughput. Setting this smaller than `receive_window` helps ensure that a single
+    /// stream doesn't monopolize receive buffers, which may otherwise occur if the application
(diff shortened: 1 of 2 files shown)


Ralith (changes requested):
To be effective, `set_receive_window` must increase `StreamsState::local_max_data` by the difference between the new connection-level receive window and the old one, and set `pending.max_data = true`. If the window shrinks, then instead `local_max_data` must not be changed, but the difference in credit passed to `add_read_credits` in future calls should be silently absorbed.

Adjusting the stream-level receive window is a bit quirky, because the peer will assume the window advertised during the handshake for every new stream until we tell them otherwise, which is particularly significant for peer-initiated streams. This means we can never reduce the size below that level, and so long as the level differs from that we sent in the transport parameters, we must arrange for a `MAX_STREAM_DATA` frame to be sent whenever either side opens a new receive stream. That's all perfectly viable (with a modest bit of extra code), but is it consistent with your requirements?

`max_data` is the limit dictated by the peer and should not be changed.

inflation:
I would assume the main usage for connection level data limit is to adapt to changing constraints once the connection is established. If these two properties can only go up, maybe we just state it in the doc and don't provide the knobs at all. 

It's still a useful thing to have but would be inconsistent with the options in transport level.

Ralith:
> If these two properties can only go up

That isn't the case. Existing flow control credit, including flow control credit issued implicitly on new streams, cannot be rescinded, but future flow control credit can be issued according to whatever logic we like, which can include arbitrary decreases.

Ralith:
https://github.com/quinn-rs/quinn/pull/1386 implemented the connection-scoped configuration; feel free to open a new PR if you still want to pursue dynamic per-stream limits.

PR #1386, opened 2022-07-18
Configure receive window per connection

lijunwangs:
Problem:
Currently the receive_window on the server side is obtained from the ServerConfig stored in the Endpoint, which will make all connections having the same receive_window and making per connection data limits difficult.

Solution:
Create interfaces to set the receive_window per connection. 

This is to address https://github.com/quinn-rs/quinn/issues/1342 to make receive_window configurable per connection. It only handles receive_window not the stream_receive_window which can be addressed in a separate PR.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1107,6 +1107,29 @@ impl Connection {
         self.streams.set_max_concurrent(dir, count);
     }
 
+    /// Set the maximum number of bytes the peer may transmit across all streams of a connection before
+    /// becoming blocked.
+    ///
+    /// This should be set to at least the expected connection latency multiplied by the maximum
+    /// desired throughput. Larger values can be useful to allow maximum throughput within a
+    /// stream while another is blocked.
+    pub fn set_receive_window(&mut self, receive_window: VarInt) {
+        self.streams.set_receive_window(receive_window);
+    }
+
+    /// Set the maximum number of bytes the peer may transmit without acknowledgement on any one stream
+    /// before becoming blocked.
+    ///
+    /// This should be set to at least the expected connection latency multiplied by the maximum
+    /// desired throughput. Setting this smaller than `receive_window` helps ensure that a single
+    /// stream doesn't monopolize receive buffers, which may otherwise occur if the application
+    /// chooses not to read from a large stream for a time while still requiring data on other
+    /// streams.
+    pub fn set_stream_receive_window(&mut self, stream_receive_window: VarInt) {
+        self.streams
+            .set_stream_receive_window(stream_receive_window);
+    }
+
     fn on_ack_received(
         &mut self,
         now: Instant,
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1107,14 +1107,11 @@ impl Connection {
         self.streams.set_max_concurrent(dir, count);
     }
 
-    /// Set the maximum number of bytes the peer may transmit across all streams of a connection before
-    /// becoming blocked.
-    ///
-    /// This should be set to at least the expected connection latency multiplied by the maximum
-    /// desired throughput. Larger values can be useful to allow maximum throughput within a
-    /// stream while another is blocked.
+    /// See [`TransportConfig::set_receive_window`]
     pub fn set_receive_window(&mut self, receive_window: VarInt) {
-        self.streams.set_receive_window(receive_window);
+        if self.streams.set_receive_window(receive_window) {
+            self.spaces[SpaceId::Data].pending.max_data = true;
+        }
     }
 
     fn on_ack_received(
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1107,7 +1107,7 @@ impl Connection {
         self.streams.set_max_concurrent(dir, count);
     }
 
-    /// See [`quinn::TransportConfig::receive_window()`]
+    /// See [`TransportConfig::receive_window()`]
     pub fn set_receive_window(&mut self, receive_window: VarInt) {
         if self.streams.set_receive_window(receive_window) {
             self.spaces[SpaceId::Data].pending.max_data = true;
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1107,7 +1107,7 @@ impl Connection {
         self.streams.set_max_concurrent(dir, count);
     }
 
-    /// See [`TransportConfig::set_receive_window`]
+    /// See [`quinn::TransportConfig::receive_window()`]
     pub fn set_receive_window(&mut self, receive_window: VarInt) {
         if self.streams.set_receive_window(receive_window) {
             self.spaces[SpaceId::Data].pending.max_data = true;
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1117,19 +1117,6 @@ impl Connection {
         self.streams.set_receive_window(receive_window);
     }
 
-    /// Set the maximum number of bytes the peer may transmit without acknowledgement on any one stream
-    /// before becoming blocked.
-    ///
-    /// This should be set to at least the expected connection latency multiplied by the maximum
-    /// desired throughput. Setting this smaller than `receive_window` helps ensure that a single
-    /// stream doesn't monopolize receive buffers, which may otherwise occur if the application
-    /// chooses not to read from a large stream for a time while still requiring data on other
-    /// streams.
-    pub fn set_stream_receive_window(&mut self, stream_receive_window: VarInt) {
-        self.streams
-            .set_stream_receive_window(stream_receive_window);
-    }
-
     fn on_ack_received(
         &mut self,
         now: Instant,
--- a/quinn-proto/src/connection/streams/state.rs
+++ b/quinn-proto/src/connection/streams/state.rs
@@ -759,6 +759,14 @@ impl StreamsState {
         self.ensure_remote_streams(dir);
     }
 
+    pub fn set_receive_window(&mut self, receive_window: VarInt) {
+        self.receive_window = receive_window.into();
+    }
+
+    pub fn set_stream_receive_window(&mut self, stream_receive_window: VarInt) {
+        self.stream_receive_window = stream_receive_window.into();
+    }
+
     pub(super) fn insert(&mut self, remote: bool, id: StreamId) {
         let bi = id.dir() == Dir::Bi;
         if bi || !remote {
--- a/quinn-proto/src/connection/streams/state.rs
+++ b/quinn-proto/src/connection/streams/state.rs
@@ -763,15 +763,6 @@ impl StreamsState {
         self.receive_window = receive_window.into();
     }
 
-    pub fn set_stream_receive_window(&mut self, stream_receive_window: VarInt) {
-        self.stream_receive_window = stream_receive_window.into();
-
-        // correct the existing streams
-        for recv in self.recv.values_mut() {
-            recv.set_sent_max_stream_data(self.stream_receive_window);
-        }
-    }
-
     pub(super) fn insert(&mut self, remote: bool, id: StreamId) {
         let bi = id.dir() == Dir::Bi;
         if bi || !remote {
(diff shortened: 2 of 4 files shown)


lijunwangs:
Hi @Ralith, can you take a look at this?

Ralith:
The relevant review from #1384 still applies. Copying here for convenience:

To be effective, set_receive_window must increase StreamsState::local_max_data by the difference between the new connection-level receive window and the old one, and set pending.max_data = true. If the window shrinks, then instead local_max_data must not be changed, but the difference in credit passed to add_read_credits in future calls should be silently absorbed.

To limit the risk of drift, let's cite the methods on TransportConfig (with intra-doc links) rather than duplicating their docs.

lijunwangs:
> The relevant review from #1384 still applies. Copying here for convenience:
> 
> To be effective, set_receive_window must increase StreamsState::local_max_data by the difference between the new connection-level receive window and the old one, and set pending.max_data = true. If the window shrinks, then instead local_max_data must not be changed, but the difference in credit passed to add_read_credits in future calls should be silently absorbed.
> 
> To limit the risk of drift, let's cite the methods on TransportConfig (with intra-doc links) rather than duplicating their docs.

Thanks @Ralith  on the local_max_data when window is shrank, just to confirm that when we add_read_credits, we do not do the following:

Ralith:
> Instead we do self.local_max_data = self.local_max_data.saturating_add(credits - (shrink_diff)) where shrink_diff = (new_receive_window - old_receive_window)?

Close, but some state is required, because it could take many calls to `add_read_credits` until the excess window size from a single reduction is absorbed. You should compute the excess window size, then so long as that amount is nonzero, subtract from it instead of adding to `local_max_data`. Immediately after hits zero, every following credit goes back into adding to `local_max_data`. This will also need unit testing to verify that credit is handled correctly even when read credit increments are large/small/etc.

lijunwangs:
Hi @Ralith, on the following

```
To be effective, set_receive_window must increase StreamsState::local_max_data by the difference between the new connection-level receive window and the old one, and set pending.max_data = true.
```

Which `pending` are you referring to? The one in RecvStream? Or in in StreamsState? I do not see the `pending` in StreamsState has max_data field.

Ralith:
`self.spaces[SpaceId::Data].pending` in `Connection`, specifically.

quinn-proto/src/connection/mod.rs
```diff
@@ -1107,6 +1107,13 @@ impl Connection {
         self.streams.set_max_concurrent(dir, count);
     }
 
+    /// See [`TransportConfig::set_receive_window`]
```
lijunwangs, line 1110: TransportConfig::receive_window

quinn/src/connection.rs
```diff
@@ -546,6 +546,18 @@ impl Connection {
         conn.wake();
     }
 
+    /// Set the maximum number of bytes the peer may transmit across all streams of a connection before
```
lijunwangs, line 549: Ue doc link

lijunwangs:
@Ralith -- I have addressed your feedback. Could you take another look? Thanks

lijunwangs:
@Ralith -- a gentle reminder, can you review this again?

quinn-proto/src/connection/streams/state.rs
```diff
@@ -789,7 +811,13 @@ impl StreamsState {
     /// suppress sending further updates until the window increases significantly
     /// again.
     pub(super) fn add_read_credits(&mut self, credits: u64) -> ShouldTransmit {
-        self.local_max_data = self.local_max_data.saturating_add(credits);
+        if credits > self.receive_window_shrink_debt {
+            let net_credits = credits.saturating_sub(self.receive_window_shrink_debt);
```
Ralith, line 815: This `saturating_sub` should be a regular subtraction for clarity, since we already know the difference is at least 1 from the `if` condition.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1553,4 +1581,77 @@ mod tests {
             assert_eq!(client.max_remote[Dir::Bi as usize], 201);
         }
     }
+
+    #[test]
+    fn expand_receive_window() {
+        let mut server = make(Side::Server);
+        let new_receive_window = 2 * 1024 * 1024u32;
+        let explanded = server.set_receive_window(new_receive_window.into());
```
Ralith, line 1589: Typo:
```suggestion
        let expanded = server.set_receive_window(new_receive_window.into());
```

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1553,4 +1581,77 @@ mod tests {
             assert_eq!(client.max_remote[Dir::Bi as usize], 201);
         }
     }
+
+    #[test]
+    fn expand_receive_window() {
+        let mut server = make(Side::Server);
+        let new_receive_window = 2 * 1024 * 1024u32;
```
Ralith, line 1588: For robustness, let's make this `2 * server.receive_window` or similar, rather than relying on one hardcoded value having a particular relationship to another hardcoded value elsewhere.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1553,4 +1581,77 @@ mod tests {
...
+        let shuould_transmit = server.add_read_credits(credits);
+        assert_eq!(server.receive_window_shrink_debt, 0);
+        assert_eq!(server.local_max_data, prev_local_max_data + credits);
+        assert!(shuould_transmit.should_transmit());
+    }
+
+    #[test]
+    fn shrink_receive_window() {
+        let mut server = make(Side::Server);
+        let new_receive_window = 1024 * 512u32;
```
Ralith, line 1607: As above, let's derive this from the initial window rather than hardcoding.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1553,4 +1581,77 @@ mod tests {
...
+        assert!(!explanded);
+        assert_eq!(server.receive_window, new_receive_window as u64);
+        assert_eq!(server.local_max_data, prev_local_max_data);
+        assert_eq!(server.receive_window_shrink_debt, shrink_diff);
+        let prev_local_max_data = server.local_max_data;
+
+        // credit, local_max_data does not change as it is absorbed by receive_window_shrink_debt
+        let credits = 1024u64;
+        let expected_receive_window_shrink_debt = server.receive_window_shrink_debt - credits;
+        let shuould_transmit = server.add_read_credits(credits);
```
Ralith, line 1622: Typo:
```suggestion
        let should_transmit = server.add_read_credits(credits);
```

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1553,4 +1581,77 @@ mod tests {
...
+        assert_eq!(
+            server.receive_window_shrink_debt,
+            expected_receive_window_shrink_debt
+        );
+        assert_eq!(server.local_max_data, prev_local_max_data);
+        assert!(!shuould_transmit.should_transmit());
+
+        // credit again, local_max_data does not change as it is absorbed by receive_window_shrink_debt
+        let expected_receive_window_shrink_debt = server.receive_window_shrink_debt - credits;
+        let shuould_transmit = server.add_read_credits(credits);
```
Ralith, line 1632: Typo:
```suggestion
        let should_transmit = server.add_read_credits(credits);
```

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1553,4 +1581,77 @@ mod tests {
...
+        assert_eq!(
+            server.receive_window_shrink_debt,
+            expected_receive_window_shrink_debt
+        );
+        assert_eq!(server.local_max_data, prev_local_max_data);
+        assert!(!shuould_transmit.should_transmit());
```
Ralith, line 1638: This looks like the exact same code ran twice. Convert into a loop of the form `for _ in 0..2` for better clarity?

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1553,4 +1581,77 @@ mod tests {
...
+
+        // credit again which exceeds all remaining expected_receive_window_shrink_debt
+        let credits = 1024 * 512;
+        let expected_local_max_data =
+            server.local_max_data + (credits - server.receive_window_shrink_debt);
+        let shuould_transmit = server.add_read_credits(credits);
```
Ralith, line 1644: Typo

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1553,4 +1581,77 @@ mod tests {
...
+        assert_eq!(server.receive_window_shrink_debt, 0);
+        assert_eq!(server.local_max_data, expected_local_max_data);
+        assert!(!shuould_transmit.should_transmit());
+
+        // credit again, all should be added to local_max_data
+        let credits = 1024 * 512;
+        let expected_local_max_data = server.local_max_data + credits;
+        let shuould_transmit = server.add_read_credits(credits);
```
Ralith, line 1652: Typo

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1553,4 +1581,77 @@ mod tests {
...
+        assert!(!shuould_transmit.should_transmit());
+
+        // credit again which exceeds all remaining expected_receive_window_shrink_debt
+        let credits = 1024 * 512;
+        let expected_local_max_data =
+            server.local_max_data + (credits - server.receive_window_shrink_debt);
+        let shuould_transmit = server.add_read_credits(credits);
+        assert_eq!(server.receive_window_shrink_debt, 0);
+        assert_eq!(server.local_max_data, expected_local_max_data);
+        assert!(!shuould_transmit.should_transmit());
```
Ralith, line 1647: If `credits - server.receive_window_shrink_debt` is expected to be nonzero:
- we should assert explicitly that `local_max_data` grew
- we should not assert that `should_transmit` is false to reduce this test's sensitivity to the flow control significance threshold

Ralith (changes requested):
Thanks for the ping, this had slipped through the cracks.

Thank you for writing up those very thorough tests! Overall this looks great, just a few minor tweaks.

quinn-proto/src/connection/streams/state.rs
lijunwangs: Done

quinn-proto/src/connection/streams/state.rs
lijunwangs: Done

quinn-proto/src/connection/streams/state.rs
lijunwangs: Done

quinn-proto/src/connection/streams/state.rs
lijunwangs: Done

quinn-proto/src/connection/streams/state.rs
lijunwangs: Done

quinn-proto/src/connection/streams/state.rs
lijunwangs: Done

quinn-proto/src/connection/streams/state.rs
lijunwangs: Done

quinn-proto/src/connection/streams/state.rs
lijunwangs: Done

quinn-proto/src/connection/streams/state.rs
lijunwangs: Done

quinn-proto/src/connection/streams/state.rs
lijunwangs: Done

Ralith (approved):
LGTM, thanks!

lijunwangs:
Thanks @Ralith for your detailed reviews. Can you help merge this PR? I do not have permission to merge the PR against this repo. Thanks

Ralith:
Usually I prefer to give @djc an opportunity to weigh in on nontrivial changes. Do you specifically need it merged quickly?

lijunwangs:
> 



> Usually I prefer to give @djc an opportunity to weigh in on nontrivial changes. Do you specifically need it merged quickly?

Yes -- we have a project which we plan to use this feature tweak different data limits for different connections for the same endpoint from the server side -- related to issue https://github.com/quinn-rs/quinn/issues/1342.

djc:
I'll review this today.

djc (approved):

lijunwangs:
Hi @Ralith and @djc, I wonder if it is possible to merge this to the 0.8.x branch? Our application is based off 0.8.x branch.

djc:
Sure, please submit a PR to the 0.8.x branch (squashing all the changes).

PR #1387, opened 2022-07-21
Fix build errors so quinn-proto can be used under wasm32-unknown-unknown

thombles:
This fixes a handful of basic things so that `quinn-proto` (with no default features) cleanly builds under the `wasm32-unknown-unknown` target, and can be used with a browser-based `wasm-bindgen-test` runner.

* One of the transitive dependencies `getrandom` needs an additional feature to work in the browser
* There is a warning about an unused import when the `ring` feature is not in use
* There were some odd type inference failures:

```
   Compiling quinn-proto v0.8.0 (/home/tk/repos/quinn/quinn-proto)
error[E0283]: type annotations needed
   --> /home/tk/repos/quinn/quinn-proto/src/connection/streams/recv.rs:150:23
    |
150 |             if offset != final_offset.into() {
    |                       ^^ ------------------- this method call resolves to `T`
    |                       |
    |                       cannot infer type
    |
    = note: multiple `impl`s satisfying `u64: PartialEq<_>` found in the following crates: `core`, `serde_json`:
            - impl PartialEq for u64;
            - impl PartialEq<serde_json::value::Value> for u64;

error[E0283]: type annotations needed
   --> /home/tk/repos/quinn/quinn-proto/src/connection/streams/state.rs:285:26
    |
285 |         Ok(if bytes_read != final_offset.into() {
    |                          ^^ ------------------- this method call resolves to `T`
    |                          |
    |                          cannot infer type
    |
    = note: multiple `impl`s satisfying `u64: PartialEq<_>` found in the following crates: `core`, `serde_json`:
            - impl PartialEq for u64;
            - impl PartialEq<serde_json::value::Value> for u64;

error[E0282]: type annotations needed
   --> /home/tk/repos/quinn/quinn-proto/src/connection/mod.rs:746:62
    |
746 |                         == self.path.max_udp_payload_size as _),
    |                                                              ^ cannot infer type
    |
    = note: type must be known at this point
```

### DIFF
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -36,6 +36,10 @@ tinyvec = { version = "1.1", features = ["alloc"] }
 tracing = "0.1.10"
 webpki = { version = "0.22", default-features = false, optional = true }
 
+[target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
+# Choose appropriate feature set for getrandom, which is used via rand/rand_core
+getrandom = { version = "0.2", features = ["js"] }
+
 [dev-dependencies]
 assert_matches = "1.1"
 hex-literal = "0.3.0"
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -36,10 +36,6 @@ tinyvec = { version = "1.1", features = ["alloc"] }
 tracing = "0.1.10"
 webpki = { version = "0.22", default-features = false, optional = true }
 
-[target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
-# Choose appropriate feature set for getrandom, which is used via rand/rand_core
-getrandom = { version = "0.2", features = ["js"] }
-
 [dev-dependencies]
 assert_matches = "1.1"
 hex-literal = "0.3.0"
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -1,7 +1,9 @@
 use std::{convert::TryInto, fmt, num::TryFromIntError, sync::Arc, time::Duration};
 
-use rand::RngCore;
 use thiserror::Error;
+
+#[cfg(feature = "ring")]
+use rand::RngCore;
 
 use crate::{
     cid_generator::{ConnectionIdGenerator, RandomConnectionIdGenerator},
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -743,7 +743,7 @@ impl Connection {
                     && !can_send.acks
                     && can_send.other
                     && (buf_capacity - builder.datagram_start)
-                        == self.path.max_udp_payload_size as _),
+                        == self.path.max_udp_payload_size as usize),
                 "SendableFrames was {:?}, but only ACKs have been written",
                 can_send
             );
--- a/quinn-proto/src/connection/streams/recv.rs
+++ b/quinn-proto/src/connection/streams/recv.rs
@@ -147,7 +147,7 @@ impl Recv {
     ) -> Result<bool, TransportError> {
         // Validate final_offset
         if let Some(offset) = self.final_offset() {
-            if offset != final_offset.into() {
+            if offset != final_offset.into_inner() {
                 return Err(TransportError::FINAL_SIZE_ERROR("inconsistent value"));
             }
         } else if self.end > final_offset.into() {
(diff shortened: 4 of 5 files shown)


djc (dismissed):
Thanks!

quinn-proto/Cargo.toml
```diff
@@ -36,6 +36,10 @@ tinyvec = { version = "1.1", features = ["alloc"] }
 tracing = "0.1.10"
 webpki = { version = "0.22", default-features = false, optional = true }
 
+[target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
+# Choose appropriate feature set for getrandom, which is used via rand/rand_core
+getrandom = { version = "0.2", features = ["js"] }
```
Ralith, line 41: Are we sure this doesn't affect other targets? My understanding is that target-specific features don't actually work yet (https://github.com/rust-lang/cargo/issues/1197).

thombles: Good call-out - this one always trips me up so I've gone back for a closer look. I'm pretty sure this is okay. The deficiency in cargo is that if your dependency appears multiple times in `Cargo.toml` underneath different targets, you'll end up with the union of features rather than the specific set you wanted for your target. Choosing whether the dependency exists or not (with a given featureset) does work though.

As committed:
* `cargo tree` (for x86_64) shows no direct dependency on `getrandom` from `quinn-proto` (but there is one under `rand_core` of course, which doesn't have any JS support libs)
* `cargo tree --target wasm32-unknown-unknown` has a direct dependency, and has also brought in the JS-related libs

Interestingly, it also works fine if we take out the `[target...]` header entirely and leave `getrandom`+`js` as an always-on dependency.

* `cargo tree` (for x86_64) has the direct dependency on `getrandom` but it has no meaningful effect, and _doesn't_ pull in JS-related dependencies
* `cargo tree --target wasm32-unknown-unknown` is the same as before, and works as you would expect
* `getrandom` itself has target-specific dependencies for that extra JS support, which is why the feature is not harmful on a more typical platform

https://github.com/rust-random/getrandom/blob/7089766df0ae5b9377d5bf63138fad235823ea6d/Cargo.toml#L26-L28

So we could just pop it straight in `[dependencies]` if this target pruning is just going to create noise/confusion.

(For reference, this "add `getrandom` to your own `Cargo.toml`" is what they recommend [in their docs](https://docs.rs/getrandom/latest/getrandom/index.html#indirect-dependencies) to support WASM builds.)

Ralith: Thanks for the detailed investigation! It looks like `getrandom` is taking care to make this safe. In particular:

> This feature has no effect on targets other than `wasm32-unknown-unknown`.

is reassuring. On that basis, let's preserve this as currently written; might as well avoid spurious dependency edges.

Ralith (dismissed):

quinn-proto/Cargo.toml
Ralith: Although... any idea why this clause doesn't exist in `rand` or `rand_core` itself? It seems odd for it to be specifically our responsibility to apply this target detection.

thombles: Tbh I had the same thought but the the `rand` issues (https://github.com/rust-random/rand/issues/1110#issuecomment-812909245 https://github.com/rust-random/rand/issues/991#issuecomment-771386545) make it clear that they don't want to do feature-forwarding.

> We intentionally have removed such feature-forwarding and we do not recommend addition of such features in library crates dependent on `getrandom`.

thombles: Actually, now that I think a bit harder it means that I shouldn't be putting this in `quinn-proto` but rather in the project that depends on `quinn-proto`. I'll take it out!

Ralith: I mean, I'm confused as to why they don't just enable the feature themselves. It sounds like @newpavlov is against *any* library crates enabling this feature, which suggests that we shouldn't either, though I don't understand the reasoning.

newpavlov: The idea is that the `js` feature SHOULD be enabled only by application crates. Enabling it by default will make life miserable for those who target `wasm32-unknown-unknown`, but do not rely on `wasm-bindgen`. Ideally we would have a separate target for "web" WASM. Alternatively `js` could've been a configuration flag, but unfortunately those are somewhat unergonomic to use.

Ralith (approved):

quinn-proto/Cargo.toml
Ralith: Thank you for clarifying! I'm glad we were able to dodge that bullet at review time.

PR #1391, opened 2022-08-01
code still assumes memory layout?

leshow:
I saw that the IP types are changing in Rust... I also saw that this library updated in #987 but (and I could be wrong here) I think there's still an issue on recv? As far as I can tell, `decode_recv` still assumes the layout to be either `sockaddr_in` or `in_addr`

If I'm misunderstanding something here, sorry. I am not a libc or unsafe expert by any stretch.

I copied some of the conversion code from mio: https://github.com/tokio-rs/mio/blob/master/src/sys/unix/net.rs#L143

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -2,7 +2,7 @@ use std::{
     io,
     io::IoSliceMut,
     mem::{self, MaybeUninit},
-    net::{IpAddr, SocketAddr},
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6},
     os::unix::io::AsRawFd,
     ptr,
     sync::atomic::AtomicUsize,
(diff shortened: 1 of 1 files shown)


quinn-udp/src/unix.rs
```diff
@@ -337,7 +336,7 @@ fn recv(io: SockRef<'_>, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) ->
                 hdrs.as_mut_ptr(),
                 bufs.len().min(BATCH_SIZE) as libc::c_uint,
                 0,
-                ptr::null_mut(),
+                std::ptr::null_mut(),
```
paolobarbolini, line 339: Drive by review: is there a reason for changing this? Considering it may also need to be backported I'd reduce the changes to the minimum necessary to fix this

quinn-udp/src/unix.rs
```diff
@@ -515,8 +516,27 @@ fn decode_recv(
     }
 
     let addr = match libc::c_int::from(name.ss_family) {
-        libc::AF_INET => unsafe { SocketAddr::V4(ptr::read(&name as *const _ as _)) },
-        libc::AF_INET6 => unsafe { SocketAddr::V6(ptr::read(&name as *const _ as _)) },
+        libc::AF_INET => {
+            // Safety: if the ss_family field is AF_INET then storage must be a sockaddr_in.
+            let addr: &libc::sockaddr_in =
+                unsafe { &*(&name as *const _ as *const libc::sockaddr_in) };
+            let ip = Ipv4Addr::from(addr.sin_addr.s_addr.to_ne_bytes());
```
djc, line 523: Style nits: here and below, would be nice to inline the `ip` and `port` expressions into the `SocketAddr` initialization.

djc (approved):
Nice, some minor nits and I'd like to have @Ralith look at this, but thanks for fixing this up!

quinn-udp/src/unix.rs
leshow: There was a warning on macos after removing `ptr::read` from the other sections. If you're alright with the warning I can revert this line

quinn-udp/src/unix.rs
leshow: Sure

quinn-udp/src/unix.rs
djc: Can we put a cfg guard on the import matching the guard for this code?

leshow: Won't this create another diff line that you are trying to avoid?

Ralith (approved):
Good catch, thanks!

quinn-udp/src/unix.rs
djc: Yeah, that's fair.

brennanwatt:
Any idea when a release with this fix will be made? We're trying to move our Rust compiler version forward, but everything breaks starting w/ 2022-08-01 without this change

Ralith:
It looks like this was included in 0.8.4. Is that not working for you?

PR #1392, opened 2022-08-02
Port fix for address layout reliance to 0.8

djc:
I guess we should yank 0.1.{0,1,2}?

cc @est31 I'm not quite motivated to backport this to older versions than 0.8, but you might be interested.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,16 +12,16 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        rust: [stable, beta, 1.53.0]
+        rust: [stable, beta, 1.57.0]
         exclude:
           - os: macos-latest
             rust: beta
           - os: macos-latest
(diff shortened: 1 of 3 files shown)


Ralith (approved):

Ralith:
> I guess we should yank 0.1.{0,1,2}?

Makes sense.

djc:
Published 0.1.3 and yanked all previous versions.

djc:
On second thought, I unyanked 0.1.2 so folks can take a little longer to upgrade, I guess? I'll yank it tomorrow.

Ralith:
Doesn't yanking only affect new checkouts? It shouldn't be too disruptive unless people are pinning patch versions, which is unusual.

est31:
Yanking affects people who have the version in their Cargo.lock too, because Cargo prints a warning in that instance. But admittedly this is only a light warning. I don't think that waiting is needed.

djc:
Okay, yanked it again.

PR #1393, opened 2022-08-03
Backport: Made receive_window  tweakable per connection

lijunwangs:
Backport:  Made receive_window  tweak-able per connection on the server side.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1099,6 +1099,13 @@ impl Connection {
         self.streams.set_max_concurrent(dir, count);
     }
 
+    /// See [`TransportConfig::receive_window()`]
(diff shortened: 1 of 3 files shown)


lijunwangs:
Hi @djc @Ralith, this is the backport PR for https://github.com/quinn-rs/quinn/pull/1386. Thanks!

djc (approved):

lijunwangs:
Thank your for the review @djc -- could you also help merge this into 0.8.x branch? I do not have the write access to the repo.

djc:
I guess as a backport we probably don't need another review.

lijunwangs:
I wonder what is the plan for having a new release of 0.8.x like tag: 0.8.4 containing the fix? Our project depends on this and right now we are using explicit git dependency as workaround but which prevents us from publishing our own crates to the community. If this can be released soon, it would be great. @djc @Ralith  Thanks!

PR #1394, opened 2022-08-11
Bump versions to 0.8.4

djc:
cc @lijunwangs


Ralith (approved):

lijunwangs:
Thanks @djc and @Ralith I look forward to these crates to be published so that we can use directly!

djc:
Published.

lijunwangs:
Thanks a lot for your prompt support!

Issue #1395, opened 2022-08-12
A way of explicitly detecting a timeout/connection close


WilliamVenner:
Hi there,

Is there a way to explicitly detect if a connection has timed out or closed without affecting or opening streams?

I have a situation where I am waiting on a future that must be aborted if the connection times out or the connection closes. Unfortunately at this point I am unable to open, read or write to any streams of the connection, so I need to explicitly listen for a timeout/connection close.

Is there any way to do this in Quinn currently?

Many thanks

WilliamVenner:
I found https://docs.rs/quinn/latest/quinn/struct.SendStream.html#method.stopped !!

Ralith:
It's also common to use the end of incoming datagrams/streams/whatever from the streams in `NewConnection` to determine this.

PR #1396, opened 2022-08-15
Define `Connection::closed` helper to await connection termination

Ralith:
This has come up several times, most recently in https://github.com/quinn-rs/quinn/issues/1395. While equivalent behavior is already possible by waiting for e.g. `IncomingUniStreams` to yield `Err` or `None`, this is substantially more discoverable.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -366,6 +366,30 @@ impl Connection {
         }
     }
 
+    /// Wait for the connection to be closed for any reason
+    ///
+    /// Despite the return type's name, closed connections are often not an error condition at the
+    /// application layer. Cases that might be routine include [`ConnectionError::LocallyClosed`]
+    /// and [`ConnectionError::ApplicationClosed`].
+    pub async fn closed(&self) -> ConnectionError {
+        let closed;
+        {
+            let conn = self.0.lock("closed");
+            closed = conn.closed.clone();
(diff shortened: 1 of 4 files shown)


djc (dismissed):
Approved pending clippy fixes.

quinn/src/connection.rs
```diff
@@ -366,6 +366,26 @@ impl Connection {
         }
     }
 
+    /// Wait for the connection to be closed for any reason
+    pub async fn closed(&self) -> ConnectionError {
```
Ralith, line 370: I'm waffling over whether this should return `Result<(), ConnectionError>`, by mapping `ConnectionError::LocallyClosed` to `Ok(())`. It doesn't add any information, but it does make unexpected vs. expected failures a little clearer--except the peer closing a connection is also an "expected" failure and certainly can't be considered `Ok`, so maybe any attempt to interpret for the application's benefit is futile?

djc: Yeah, it seems like singling out the `LocallyClosed` doesn't quite make sense. I agree that it looks a little strange to unconditionally return a type with an `Error` suffix, but I guess we should solve that using some additional documentation?

Ralith: Added some brief discussion.

djc (approved):

PR #1404, opened 2022-08-31
Check QUIC Bit that Short Header Packet set too

thekuwayama:
Hi!
This PR would allow `Connection` to check Short Header Packets greasing QUIC Bit.

`PlainHeader::decode()` can check Long Header Packets with the QUIC Bit set to a value of 0. If `EndpointConfig` is not set grease_quic_bit with false, it raise `PacketDecodeError::InvalidHeader("fixed bit unset")`. But, in my understanding, Short Header Packets should be checked QUIC Bit too.

- https://www.rfc-editor.org/rfc/rfc9287.html#section-3.1-2
- https://www.rfc-editor.org/rfc/rfc9287.html#section-3.1-3

Thanks.

### DIFF
--- a/quinn-proto/src/packet.rs
+++ b/quinn-proto/src/packet.rs
@@ -575,7 +575,7 @@ impl PlainHeader {
                 });
             }
 
-            match LongHeaderType::from_byte(first, grease_quic_bit)? {
+            match LongHeaderType::from_byte(first)? {
                 LongHeaderType::Initial => {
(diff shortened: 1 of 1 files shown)


Ralith (approved):
Good catch, thanks! Confirmed by [RFC 9000 §17.3.1](https://www.rfc-editor.org/rfc/rfc9000.html#name-1-rtt-packet).

PR #1431, opened 2022-10-26
quinn-udp: bump version to 0.3

djc:
quinn-proto is a public dependency of quinn-udp (for example, `RecvMeta` exposes `EcnCodePoint`). #1429 bumped the version for quinn-proto and the dependency for quinn-udp, but not quinn-udp's version; do that here.

Since I was building in an old working directory I also noticed that I was unable to compile quinn with the tokio 1.9 that `Cargo.lock` was set to. Bump the tokio dependency up to 1.13 which is the oldest version that seems to compile.

### DIFF
--- a/quinn-udp/Cargo.toml
+++ b/quinn-udp/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quinn-udp"
-version = "0.2.0"
+version = "0.3.0"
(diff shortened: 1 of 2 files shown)


Ralith (approved):
Good catch, thanks! Figured we were okay due to the bump to 0.2, but of course quinn-proto is a public dep of quinn-udp, whoops.

PR #1434, opened 2022-10-27
Add --keylog command line option for perf client and server binaries.

stormshield-damiend:
This allow dumping crypto keys to disk using SSLKEYLOGFILE env var.

### DIFF
--- a/perf/src/bin/perf_client.rs
+++ b/perf/src/bin/perf_client.rs
@@ -60,6 +60,9 @@ struct Opt {
     #[cfg(feature = "json-output")]
     #[clap(long)]
     json: Option<PathBuf>,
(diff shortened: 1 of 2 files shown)


djc (approved):

djc:
This seems fine, but I'm a little confused why you need/want this for the perf binary?

stormshield-damiend:
> This seems fine, but I'm a little confused why you need/want this for the perf binary?

We were testing and debugging a "no packet encryption" patch and other optimizations and needed to check that no encryption was applied to packet and that optimization were working as expected. I plan to do some more PRs for the "no packet encryption" patch and also to some optimizations we did.

Ralith (approved):

PR #1436, opened 2022-10-27
Add dangerous_packet_null_cipher feature to disable packet encryption.

stormshield-damiend:
This can be use when doing performance tests to reveal other hot-spots than packet encryption and decryption.
This is for testing purpose and should not be used in production.

### DIFF
--- a/perf/src/bin/perf_client.rs
+++ b/perf/src/bin/perf_client.rs
@@ -11,8 +11,8 @@ use quinn::TokioRuntime;
 use tokio::sync::Semaphore;
 use tracing::{debug, error, info};
 
-use perf::bind_socket;
 use perf::stats::{OpenStreamStats, Stats};
+use perf::{bind_socket, noprotection::NoProtectionClientConfig};
 #[cfg(feature = "json-output")]
 use std::path::PathBuf;
 
@@ -66,6 +66,9 @@ struct Opt {
     /// UDP payload size that the network must be capable of carrying
     #[clap(long, default_value = "1200")]
     initial_max_udp_payload_size: u16,
+    /// Disable packet encryption/decryption (for debugging purpose)
+    #[clap(long = "no-protection")]
+    no_protection: bool,
 }
 
 #[tokio::main(flavor = "current_thread")]
@@ -128,7 +131,11 @@ async fn run(opt: Opt) -> Result<()> {
     let mut transport = quinn::TransportConfig::default();
     transport.initial_max_udp_payload_size(opt.initial_max_udp_payload_size);
 
-    let mut cfg = quinn::ClientConfig::new(Arc::new(crypto));
+    let mut cfg = if opt.no_protection {
+        quinn::ClientConfig::new(Arc::new(NoProtectionClientConfig::new(Arc::new(crypto))))
+    } else {
+        quinn::ClientConfig::new(Arc::new(crypto))
+    };
     cfg.transport_config(Arc::new(transport));
 
     let stream_stats = OpenStreamStats::default();
--- a/perf/src/bin/perf_server.rs
+++ b/perf/src/bin/perf_server.rs
@@ -6,7 +6,7 @@ use clap::Parser;
 use quinn::TokioRuntime;
 use tracing::{debug, error, info};
 
-use perf::bind_socket;
+use perf::{bind_socket, noprotection::NoProtectionServerConfig};
 
 #[derive(Parser)]
 #[clap(name = "server")]
@@ -35,6 +35,9 @@ struct Opt {
     /// UDP payload size that the network must be capable of carrying
     #[clap(long, default_value = "1200")]
     initial_max_udp_payload_size: u16,
+    /// Disable packet encryption/decryption (for debugging purpose)
+    #[clap(long = "no-protection")]
+    no_protection: bool,
 }
 
 #[tokio::main(flavor = "current_thread")]
@@ -87,7 +90,11 @@ async fn run(opt: Opt) -> Result<()> {
     let mut transport = quinn::TransportConfig::default();
     transport.initial_max_udp_payload_size(opt.initial_max_udp_payload_size);
 
-    let mut server_config = quinn::ServerConfig::with_crypto(Arc::new(crypto));
+    let mut server_config = if opt.no_protection {
+        quinn::ServerConfig::with_crypto(Arc::new(NoProtectionServerConfig::new(Arc::new(crypto))))
+    } else {
+        quinn::ServerConfig::with_crypto(Arc::new(crypto))
+    };
     server_config.transport_config(Arc::new(transport));
 
     let socket = bind_socket(opt.listen, opt.send_buffer_size, opt.recv_buffer_size)?;
--- a/perf/Cargo.toml
+++ b/perf/Cargo.toml
@@ -14,6 +14,7 @@ json-output = ["serde", "serde_json"]
 anyhow = "1.0.22"
 hdrhistogram = { version = "7.2", default-features = false }
 quinn = { path = "../quinn" }
+quinn-proto = { path = "../quinn-proto" }
 rcgen = "0.10.0"
 rustls = { version = "0.20", default-features = false, features = ["dangerous_configuration"] }
 rustls-pemfile = "1.0.0"
--- a/perf/src/lib.rs
+++ b/perf/src/lib.rs
@@ -7,6 +7,8 @@ use tracing::warn;
 #[cfg_attr(not(feature = "json-output"), allow(dead_code))]
 pub mod stats;
 
+pub mod noprotection;
+
 pub fn bind_socket(
     addr: SocketAddr,
     send_buffer_size: usize,
--- a/perf/src/noprotection.rs
+++ b/perf/src/noprotection.rs
@@ -1,0 +1,191 @@
+use std::sync::Arc;
+
+use bytes::BytesMut;
+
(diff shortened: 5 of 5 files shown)


Ralith:
Could you accomplish this with a custom `crypto::Session` implementation downstream? That's a bit more boilerplate since you'd have to manually forward several methods, but it would let us keep this hazardous code away from the main lib.

stormshield-damiend:
> Could you accomplish this with a custom `crypto::Session` implementation downstream? That's a bit more boilerplate since you'd have to manually forward several methods, but it would let us keep this hazardous code away from the main lib.

i will take a look and get back when done.

stormshield-damiend:
As a reminder the goal of the patch is to disable packet encryption on the production code path to find hotspot on the production code path.

I took some time to look at the various traits implemented in "quinn-proto/src/crypto/rustls.rs" and i think it will not achieve the previously listed goal (if i did not miss something obvious). We want the TLS session to be created and used as if packets were encrypted. The same patch could also be done to disable packet header protection (i plan to do a similar patch or update this one). If we implement a custom crypto Session, we still want to run the TLS handshake and ephemeral session key code as in production, this will need some kind of "inner: TlsSession" field to forward calls, doing this may/will lower performance as crypto::Session and associated Trait will need to be forwarded to call appropriate function on inner field. Doing so the tested code will not be the same as the production one without the packet encryption/decryption.

I totally understand that this code is dangerous to commit as it must not be enabled in production, if you want you can keep the code as a patch/PR and use it when doing performance measurement in order to test or develop some optimizations.

Ralith:
Why do you believe forwarding calls to an inner `TlsSession` would impact performance at all?

stormshield-damiend:
> Why do you believe forwarding calls to an inner `TlsSession` would impact performance at all?

i will do a POC and see how it goes, but i think i may need to have Traits forwarding thus one more level of calls.

stormshield-damiend:
Sorry for the delay but i was busy working on other topics. Here is an updated version with a custom `crypto::Session`

quinn-proto/src/crypto/rustls.rs
```diff
@@ -263,7 +266,11 @@ impl crypto::ClientConfig for rustls::ClientConfig {
                 )
                 .unwrap(),
             ),
-        }))
+        };
+        #[cfg(not(feature = "dangerous_packet_null_cipher"))]
```
Matthias247, line 270: The way this is implemented makes this a global setting. Either all quinn implementations inside an executable would use real crypto, or they would use no crypto. It's not a per-endpoint or per-connection choice anymore.

This is a bit inflexible.

But what is even more problematic is that it opens the door for supply-chain attacks. You can have any other crate define a dependency on quinn with the dangerous feature enabled, and as a result our code which directly uses quinn and the intention of being safe now was crypto enabled.

I think what would be safer is the rustls approach of requiring the feature flag to enable a new API which actually disables the security. So there should be an EndpointConfig/TransportConfig/etc that decides whether to return `NullCipherSession` or not.

Ralith:
The main benefit of achieving this with a custom `Session` is that no changes to quinn are required. You should be able to place this implementation in the downstream performance testing code that prompted its development, and use it with an unmodified quinn.

stormshield-damiend:
> The main benefit of achieving this with a custom `Session` is that no changes to quinn are required. You should be able to place this implementation in the downstream performance testing code that prompted its development, and use it with an unmodified quinn.

Hi Ralith,

just to be sure: you want me to go further and implement a full standalone session in perf sub-crate and add options to enable it ?

djc:
Yes, I think that's the right way to implement this if you want it merged upstream.

stormshield-damiend:
> Yes, I think that's the right way to implement this if you want it merged as part of this repo.

Ok il will take a look and update the patch. I need to figure out how TlsSession is magically used as a rustls glue layer.

stormshield-damiend:
New version available with the custom crypto session localized in perf folder. It can be enabled using ```--null-cipher``` command line option.

perf/src/null.rs
```diff
@@ -0,0 +1,191 @@
...
+    inner: Box<dyn crypto::PacketKey>,
+}
+
+impl NullCipherPacketKey {
+    fn new(key: Box<dyn crypto::PacketKey>) -> Self {
+        Self { inner: key }
+    }
+}
+
+pub struct NullCiperClientConfig {
```
Ralith, line 31: typo:
```suggestion
pub struct NullCipherClientConfig {
```

perf/src/bin/perf_server.rs
```diff
@@ -87,7 +90,12 @@ async fn run(opt: Opt) -> Result<()> {
     let mut transport = quinn::TransportConfig::default();
     transport.initial_max_udp_payload_size(opt.initial_max_udp_payload_size);
 
-    let mut server_config = quinn::ServerConfig::with_crypto(Arc::new(crypto));
+    // FIXME add command line option
```
Ralith, line 93: Outdated comment?

Ralith (changes requested):
Thanks, that's exactly what I had in mind.

djc:
So naming wise, I'm not so sure "null cipher" fits in? We don't use the term cipher anywhere else IIRC. Maybe no_crypto?

perf/src/bin/perf_client.rs
```diff
@@ -66,6 +66,9 @@ struct Opt {
     /// UDP payload size that the network must be capable of carrying
     #[clap(long, default_value = "1200")]
     initial_max_udp_payload_size: u16,
+    /// Disable packet encryption/decryption (for debugging purpose)
+    #[clap(long = "null-cipher")]
+    nullcipher: bool,
```
djc, line 71: The field here should follow the naming convention, so null_cipher rather than nullcipher.

perf/src/bin/perf_client.rs
```diff
@@ -128,7 +131,11 @@ async fn run(opt: Opt) -> Result<()> {
     let mut transport = quinn::TransportConfig::default();
     transport.initial_max_udp_payload_size(opt.initial_max_udp_payload_size);
 
-    let mut cfg = quinn::ClientConfig::new(Arc::new(crypto));
+    let mut cfg = if opt.nullcipher {
+        quinn::ClientConfig::new(Arc::new(NullCiperClientConfig::new(Arc::new(crypto))))
+    } else {
+        quinn::ClientConfig::new(Arc::new(crypto))
```
djc, line 137: If the use of crypto is going in this branch, we should also move the initialization of it here?

perf/src/null.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+}
+
+pub struct NullCiperServerConfig {
+    inner: Arc<rustls::ServerConfig>,
+}
+
+impl NullCiperServerConfig {
+    pub fn new(config: Arc<rustls::ServerConfig>) -> Self {
+        Self { inner: config }
```
djc, line 47: Why are we storing a rustls config in here??

perf/src/bin/perf_client.rs
Ralith: It's used in both branches; the new logic uses it to delegate to TLS for most operations.

perf/src/null.rs
Ralith: See references to `self.inner` below. In short, this logic creates a TLS session as usual but short-circuits the packet protection.

djc: Why don't we go whole hog and avoid all the TLS work?

perf/src/null.rs
stormshield-damiend: Fixed

perf/src/bin/perf_server.rs
stormshield-damiend: Exactly

stormshield-damiend:
> So naming wise, I'm not so sure "null cipher" fits in? We don't use the term cipher anywhere else IIRC. Maybe no_crypto?

i am open to any suggestion about naming, @Ralith what is your opinion ?

perf/src/bin/perf_client.rs
stormshield-damiend: Fixed (and will be updated later depending on name discussion outcome).

perf/src/null.rs
stormshield-damiend: There is two reasons for this:
- we want to do everything TLS related (handshake, key rolling, ...) to see it in profile/flamegraph but we don't want the profile to be overwhelmed with the crypto operation
- doing like this require far less modification

What could also be bypassed is header protection and also packet tags as both of those operation also do some kind of crypto operations.

djc: Ah, okay, thanks.

djc:
Maybe `NoPacketProtection`, or `NoProtection` or `Unprotected` if you also want to strip out the header protection?

stormshield-damiend:
> Maybe `NoPacketProtection`, or `NoProtection` or `Unprotected` if you also want to strip out the header protection?

Maybe NoEncryption to be close to this RFC/Draft https://datatracker.ietf.org/doc/html/draft-banks-quic-disable-encryption ?

stormshield-damiend:
> Maybe `NoPacketProtection`, or `NoProtection` or `Unprotected` if you also want to strip out the header protection?

I am fine with NoProtection and maybe later a level or bitfield could be included as parameter to choose which part could be disabled (header / packet / tag).

djc (approved):

Ralith (approved):

Issue #729, opened 2020-04-22
proto: write outgoing packets to caller-supplied memory


Ralith:
`Connection::poll_transmit` allocates and immediately returns a `Vec<u8>`. There's no reason it couldn't instead take a `&mut [MaybeUninit<u8>]` or similar, giving the caller much greater flexibility to reduce allocations/copies, particularly when working with e.g. registered io_uring buffers.

lmpn:
This PR https://github.com/quinn-rs/quinn/pull/1545 seems to have addressed this issue

Ralith:
It has not; refer to the signature of `Connetion::poll_transmit`.

lmpn:
I'm giving it a try 👍

cameronelliott:
Just surfing by, and saw this.
I'm just going to dump my two cents for what it's worth. 
Please ignore if it is not relevant for you.

This type of issue is super important for io_uring, AF_XDP, DPDK, (Networking APIs to get to 100Gbits/sec++)

Actually to take Quinn or any Rust Quic stack to the next level the whole RX & TX hot paths should be allocation-free except by the consumer of the Quic stack.

It would be great to have a Rust Quic stack that allows caller-supplied buffers for both the RX & TX paths, and is allocation-free on the hot path.
(When I say RX, I mean the post-decryption RX destination, a sans-io Quic has the caller do the raw RX)

Ralith:
This was fixed by @lmpn's generous contribution in #1697.

> I mean the post-decryption RX destination

We already decrypt in-place in caller-supplied memory, and only alloc/copy on the receive path when needed to reduce memory usage. Memory allocation is not currently a bottleneck for Quinn; if you want to hit massive throughput, parallelism and encryption are your main bottlenecks. If you're interested in hitting demanding performance targets with Quinn, feel free to open a new issue to discuss any obstacles you encounter.

PR #1108, opened 2021-05-02
Reserve the maximum transmit buffer capacity upfront

Matthias247:
The current implementation continously resizes the datagram buffer if
GSO is enabled and further datagrams are appended. Some benchmarking
and profiling showed that this doesn't have too much of an impact with
the glibc allocator, the strategy proved rather inefficient with pooling
memory allocators like jemalloc and mimalloc. For those, the cost of
calling `realloc` is rather high.

I benchmarked a bunch of strategies to determine the most efficient way forward:
1. Continously resize output buffer (Current approach)
2. Reserve maximum buffer size upfront
3. Reserve space for a single datagram to minimize over-allocation for
  tiny transmits. If this is not enough, reallocate once for maximum size

Based on the results of those, I am proposing to go for approach 2) and
simply allocate the maximum buffer size upfront, which yields maximum
efficiency for mimalloc + jemalloc.

### Glibc:

Baseline:

> Sent 1073741824 bytes on 1 streams in 1.79s (572.62 MiB/s)

Allocate for 1 MTU, then for `max_datagrams`:

> Sent 1073741824 bytes on 1 streams in 1.78s (576.31 MiB/s)

Allocate for `max_datagrams` upfront:

> Sent 1073741824 bytes on 1 streams in 1.79s (572.28 MiB/s)

### Mimalloc:

Baseline:

> Sent 1073741824 bytes on 1 streams in 1.84s (557.34 MiB/s)

Allocate for 1 MTU, then for `max_datagrams`:

> Sent 1073741824 bytes on 1 streams in 1.74s (587.76 MiB/s)

Allocate for `max_datagrams` upfront:

> Sent 1073741824 bytes on 1 streams in 1.71s (600.06 MiB/s)

### Jemalloc:

Baseline:

> Sent 1073741824 bytes on 1 streams in 1.86s (551.75 MiB/s)

Allocate for 1 MTU, then for `max_datagrams`:

> Sent 1073741824 bytes on 1 streams in 1.73s (592.50 MiB/s)

Allocate for `max_datagrams` upfront:

> Sent 1073741824 bytes on 1 streams in 1.72s (596.29 MiB/s)

Issue #915, opened 2020-11-13
Completion-based I/O


Ralith:
Modern high-performance I/O APIs (e.g. io_uring and I/O completion ports) are completion-oriented, unlike the traditional readiness-oriented `epoll` paradigm. Advantages include fewer syscalls and no copying. On Windows in particular, readiness-oriented I/O is poorly supported.

@Matthias247 has reported that a prototype variant of Quinn modified to use registered I/O on windows performs significantly (~20%?) better than our Linux backend using `sendmmsg` and `recvmmsg` for efficient batching, and drastically better than the fallback backend on Windows. Due to major changes in tokio/mio's windows support, we should re-evaluate the latter result after moving to tokio 0.3.

On Linux, io_uring is only available on recent kernels. Other platforms (e.g. macOS, BSDs) may not offer completion-based I/O at all. Retaining a readiness-oriented fallback will therefore remain necessary for some time. However, the performance benefits seem to justify making completion-based I/O the first-class target.

A complicating factor is that tokio itself is, currently, 100% readiness-oriented. On Linux, we may be able to bridge this gap gracefully due to io_uring/epoll interop, but it's unclear if something similar is possible on Windows. If not, a background thread may be necessary.

Matthias247:
I pushed my POC for RIO and completion based IO to https://github.com/quinn-rs/quinn/pull/918

This might help on getting an idea what is necessary. I think in a model where Quinn owns all IO and the associated thread it's not terribly complicated.
The fact that datagrams are received and transmitted in an all-or-nothing fashion, and that it mostly has to deal with a single socket and not lots of them makes it easier than e.g. implementing support for completion oriented TCP. But obviously one still has to be a bit careful around ownership.

KirillLykov:
I wonder if there is any plans about using `io_uring`? It looks like another way to proceed might be using AF_XDP which might be simpler(?) in a way

Ralith:
See also discussion at https://github.com/quinn-rs/quinn/issues/1319. This issue was originally opened before GSO/GRO support were implemented and it's unclear if io_uring will be a benefit in net on Linux, though we'd be happy to mentor someone who wants to experiment.

KirillLykov:
> See also discussion at #1319. This issue was originally opened before GSO/GRO support were implemented and it's unclear if io_uring will be a benefit in net on Linux, though we'd be happy to mentor someone who wants to experiment.

[...]

PR #1438, opened 2022-11-09
Preallocate output vector in poll_transmit().

stormshield-damiend:
This is a partial fix for #729 doing the whole optimization as suggested in this issue will require more work as output vector is put within Transmit { } struct, using a &[u8] will require adding a bunch of lifetime here and there. I started working on this but it will take more time.

This simple optimization lead to a performance increase of around 3% with default crypto and around 15% without packet encryption and decryption (using custom patch from PR1436).

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -489,7 +489,7 @@ impl Connection {
             _ => false,
         };
 
-        let mut buf = Vec::new();
+        let mut buf = Vec::with_capacity(max_datagrams * self.path.max_udp_payload_size as usize);
         // Reserving capacity can provide more capacity than we asked for.
         // However we are not allowed to write more than MTU size. Therefore
         // the maximum capacity is tracked separately.


djc (approved):
Nice, this makes sense to me!

Matthias247:
I remember I’ve looked at this before and already converted it to a more gradual allocation approach which goes fast towards a single allocation but not immediately. 

This change here is more aggressive, but it will also overallocate a lot for ack-only packets (and there’s lots of those) and even if there’s no data to send. 

I probably would recommend not to spend much more time into this at the moment. Once #1219 lands the path for Quinn to own the complete IO eventloop is pretty clear. And then one could directly go towards completion based IO (#915) with preallocated buffers which avoids all the dynamic Transmit allocation stuff.

Matthias247:
This was the previous change: https://github.com/quinn-rs/quinn/pull/1108

stormshield-damiend:
> This was the previous change: #1108
> 
> Reading that again I think it already does the same that you try to achieve here: Doing a single allocation and never resize the `Vec`. I'm wondering why you observed any performance difference. The only way where #1108 wouldn't do the same thing is no packet is produced at all (the code between `Vec::new()` and the resize detects we can't write a packet), and in that case not allocating should be more efficient thinking

Hi, i did a performance test using perf_client and perf_server on two VMs between two interfaces connected using a dedicated virtual switch, upload only and did some flamegraph on client side, if you do that you will see a lot of Vec allocation when pushing/appending data in poll_transmit() and its child. If you disabled packet cipher you will see it more in the flamegraph.

Ralith:
Thanks for working on this!

I second @Matthias247's confusion as to why this makes a difference. In the existing code, buffer allocation is performed here:
https://github.com/quinn-rs/quinn/blob/b80baa008d2eb0a585472772be449432a3c48fd8/quinn-proto/src/connection/mod.rs#L622-L624

Note that when `buf.capacity()` is zero, as it is initially, this should be requesting exactly the same amount of space. I wonder if what we're really seeing here is a difference in heuristics between `with_capacity` and `reserve`? Maybe `reserve_exact` would close the gap. Another possibility is that we have a bug that's leading to more incremental allocation elsewhere.

I'd like to be sure we understand the underlying mechanism before merging a change like this, as purely empirical optimizations are nearly impossible to maintain in the face of future refactoring.

stormshield-damiend:
> Thanks for working on this!
> 
> I second @Matthias247's confusion as to why this makes a difference. In the existing code, buffer allocation is performed here:
> 
> https://github.com/quinn-rs/quinn/blob/b80baa008d2eb0a585472772be449432a3c48fd8/quinn-proto/src/connection/mod.rs#L622-L624
> 
> Note that when `buf.capacity()` is zero, as it is initially, this should be requesting exactly the same amount of space. I wonder if what we're really seeing here is a difference in heuristics between `with_capacity` and `reserve`? Maybe `reserve_exact` would close the gap. Another possibility is that we have a bug that's leading to more incremental allocation elsewhere.
> 
> I'd like to be sure we understand the underlying mechanism before merging a change like this, as purely empirical optimizations are nearly impossible to maintain in the face of future refactoring.

i will take a look later this week and see if i can find an explanation.

djc:
@stormshield-damiend @aochavia do you still want to explore this?

stormshield-damiend:
> @stormshield-damiend @aochavia do you still want to explore this?

not for now, i am busy on other stuff. By the way i think it may be related to no-encryption patch as packet may be bigger than when encrypted, what do you think ?

You can close the PR for now, i can submit another one later on.

PR #1439, opened 2022-11-13
Fix duplicate MAX_STREAMS transmit

Ralith:
When the application observes a stream finish or reset, we dispose of its state via `StreamsState::stream_freed`, and queue a transmit of stream flow control credit directly on the `Connection`. `stream_freed` however also sets `StreamsState::max_streams_dirty`, which is polled each time we receive a packet to see whether a stream has become fully closed due to an incoming frame. Together, these led to a MAX_STREAMS frame being transmit both immediately, and following the next packet receipt. We prevent that by unsetting the appropriate `max_streams_dirty` flags when sending `MAX_STREAMS`.

### DIFF
--- a/quinn-proto/src/connection/streams/state.rs
+++ b/quinn-proto/src/connection/streams/state.rs
@@ -32,7 +32,7 @@ pub struct StreamsState {
     /// Maximum number of remotely-initiated streams that may be opened over the lifetime of the
     /// connection so far, per direction
     max_remote: [u64; 2],
-    /// Number of streams that we've given the peer permission to open
+    /// Number of streams that we've given the peer permission to open and which aren't fully closed
     allocated_remote_count: [u64; 2],
     /// Size of the desired stream flow control window. May be smaller than `allocated_remote_count`
     /// due to `set_max_concurrent` calls.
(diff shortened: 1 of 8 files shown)


quinn-proto/src/connection/streams/state.rs
```diff
@@ -443,32 +443,27 @@ impl StreamsState {
...
-            trace!(
-                value = self.max_remote[Dir::Bi as usize],
-                "MAX_STREAMS (bidirectional)"
-            );
-            buf.write(frame::Type::MAX_STREAMS_BIDI);
-            buf.write_var(self.max_remote[Dir::Bi as usize]);
-            stats.max_streams_bidi += 1;
+        for dir in Dir::iter() {
+            // MAX_STREAMS
+            if pending.max_stream_id[dir as usize] && buf.len() + 9 < max_size {
```
djc, line 448: Style: this looks like an opportunity for early continue?

djc (dismissed):
Nice!

quinn-proto/src/connection/streams/state.rs
Ralith: Good catch, fixed.

djc (approved):

PR #1440, opened 2022-11-13
Fix stream open futures sometimes waiting on the wrong notification

Ralith:
After a spurious wakeup, `Connection::open_uni` and `open_bi` would switch to waiting for notifications about incoming streams, rather than the intended notification about new stream ID flow control budget. This was probably a copy-paste error from `poll_accept`.

Let's get a patch release out as soon as this is merged.


Ralith:
(linter complaints addressed in #1439)

djc:

djc (approved):

Ralith:
Science says: not hard.

djc (approved):

PR #1446, opened 2022-11-20
Improve handling of small stateless resets

### LINKED ISSUE #1443: clarification on stateless reset
I was hoping to get some clarification on how the stateless reset functionality works in quinn/QUIC. My understanding of the stateless reset feature is that if a connection exists between two endpoints A and B and one side of the connection state is lost for some reason (say endpoint A crashed and was restarted) then the side that lost state would be able to inform the other side and the old connection would be closed quickly on the side that didn't lose state. In my limited testing i've found this to not be the case though, and instead (via logs) i'm able to see that the side the lost sate (A) is sending a stateless reset but the other side (B) doesn't seem to handle the reset and close the connection on its side, instead the connection hangs until the configured idle timeout period at which point the connection is timed out.

Is my understanding of how the stateless reset is supposed to work not correct? Is there some configuration that I need to be doing differently in order to have this work properly? (eg do you need to use a non-default EndpointConfig)

Ralith:
Fixes #1443, and tweaks padding of outgoing packets to follow the guidance of RFC 9000 §10.3 on minimum size.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1865,8 +1865,14 @@ impl Connection {
             None
         };
 
+        let packet = partial_decode.data();
+        let stateless_reset = packet.len() >= RESET_TOKEN_SIZE + 5
+            && self.peer_params.stateless_reset_token.as_deref()
+                == Some(&packet[packet.len() - RESET_TOKEN_SIZE..]);
+
         match partial_decode.finish(header_crypto) {
-            Ok(packet) => self.handle_packet(now, remote, ecn, packet),
+            Ok(packet) => self.handle_packet(now, remote, ecn, Some(packet), stateless_reset),
+            Err(_) if stateless_reset => self.handle_packet(now, remote, ecn, None, true),
             Err(e) => {
                 trace!("unable to complete packet decoding: {}", e);
             }
@@ -1878,15 +1884,18 @@ impl Connection {
         now: Instant,
         remote: SocketAddr,
         ecn: Option<EcnCodepoint>,
-        mut packet: Packet,
+        packet: Option<Packet>,
+        stateless_reset: bool,
     ) {
-        trace!(
-            "got {:?} packet ({} bytes) from {} using id {}",
-            packet.header.space(),
-            packet.payload.len() + packet.header_data.len(),
-            remote,
-            packet.header.dst_cid(),
-        );
+        if let Some(ref packet) = packet {
+            trace!(
+                "got {:?} packet ({} bytes) from {} using id {}",
+                packet.header.space(),
+                packet.payload.len() + packet.header_data.len(),
+                remote,
+                packet.header.dst_cid(),
+            );
+        }
 
         if self.is_handshaking() && remote != self.path.remote {
             debug!("discarding packet with unexpected remote during handshake");
@@ -1895,15 +1904,14 @@ impl Connection {
 
         let was_closed = self.state.is_closed();
         let was_drained = self.state.is_drained();
-        let stateless_reset = self
-            .peer_params
-            .stateless_reset_token
-            .map_or(false, |token| {
-                packet.payload.len() >= RESET_TOKEN_SIZE
-                    && packet.payload[packet.payload.len() - RESET_TOKEN_SIZE..] == token[..]
-            });
 
-        let result = match self.decrypt_packet(now, &mut packet) {
+        let decrypted = match packet {
+            None => Err(None),
+            Some(mut packet) => self
+                .decrypt_packet(now, &mut packet)
+                .map(move |number| (packet, number)),
+        };
+        let result = match decrypted {
             Err(Some(e)) => {
                 warn!("illegal packet: {}", e);
                 Err(e.into())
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -1865,8 +1865,11 @@ impl Connection {
             None
         };
 
+        let stateless_reset = is_stateless_reset(partial_decode.data(), &self.peer_params);
```
djc, line 1868: This introduces a single-caller function. In this case, why is that better than introducing a block of code inline here, maybe with some comments to demarcate different parts?

quinn-proto/src/connection/mod.rs
```diff
@@ -1878,15 +1881,18 @@ impl Connection {
         now: Instant,
         remote: SocketAddr,
         ecn: Option<EcnCodepoint>,
-        mut packet: Packet,
+        packet: Option<Packet>,
+        stateless_reset: bool,
```
djc, line 1885: It looks like the `Option` wrapper around `Packet` and the `stateless_reset` flag are actually redundant state (and it looks like the `stateless_reset` flag is unused).

quinn-proto/src/connection/mod.rs
```diff
@@ -1895,15 +1901,12 @@ impl Connection {
...
-            .map_or(false, |token| {
-                packet.payload.len() >= RESET_TOKEN_SIZE
-                    && packet.payload[packet.payload.len() - RESET_TOKEN_SIZE..] == token[..]
-            });
 
-        let result = match self.decrypt_packet(now, &mut packet) {
+        let decrypted = packet.map_or(Err(None::<TransportError>), |mut packet| {
+            let number = self.decrypt_packet(now, &mut packet)?;
+            Ok((packet, number))
+        });
```
djc, line 1908: I think this might be easier to follow as an explicit `match`?

quinn-proto/src/connection/mod.rs
```diff
@@ -3381,3 +3384,15 @@ impl SentFrames {
             && self.retransmits.is_empty(streams)
     }
 }
+
+/// Whether `packet` is a valid stateless reset for this connection
+fn is_stateless_reset(packet: &[u8], peer_params: &TransportParameters) -> bool {
+    if packet.len() < RESET_TOKEN_SIZE + 5 {
+        return false;
+    }
+    let token = match peer_params.stateless_reset_token {
```
djc, line 3393: I think we had this discussion before, but I think I would prefer `&` in the scrutinee to `ref` in the pattern -- you had a good rebuttal last time but I don't remember what it was and I can't think of a good reason in this case.

quinn-proto/src/connection/packet_builder.rs
```diff
@@ -128,13 +128,16 @@ impl PacketBuilder {
...
+        // Each packet must be large enough for header protection sampling, i.e. the combined
+        // lengths of the encoded packet number and protected payload must be at least 4 bytes
+        // longer than the sample required for header protection. Further, each packet should be at
+        // least tag_len + 6 bytes larger than the destination CID on incoming packets so that the
+        // peer may send stateless resets that are indistinguishable from regular traffic.
 
         // pn_len + payload_len + tag_len >= sample_size + 4
         // payload_len >= sample_size + 4 - pn_len - tag_len
-        let min_size = buffer.len() + (sample_size + 4).saturating_sub(number.len() + tag_len);
+        let min_size = (buffer.len() + (sample_size + 4).saturating_sub(number.len() + tag_len))
```
djc, line 139: Style nit: I usually find it hard to make sense of asymmetric `.max()` in my head -- how would you feel about making it `Ord::max(.., ..)` instead? (`Ord` is in the prelude.)

djc (commented):
Nice!

quinn-proto/src/connection/mod.rs
Ralith: For most stateless reset packets, `packet` will be `Some`. It is only `None` when the reset is shorter than the RFC recommends. The `stateless_reset` flag is read in the `Err(None)` case of `match decrypted`.

quinn-proto/src/connection/mod.rs
Ralith: At time of writing I thought it might end up with two callsites. Inlined.

quinn-proto/src/connection/mod.rs
Ralith: Done.

quinn-proto/src/connection/packet_builder.rs
Ralith: I like `.max` when chaining more than two, but I'm fine with switching to the qualified version here. Done.

quinn-proto/src/connection/mod.rs
Ralith: I prefer `ref` patterns because they unambiguously communicate that you are binding a reference to the place. When matching on a value whose type isn't immediately apparent to the reader, the "match ergonomics" style makes it difficult to tell whether your pattern variables are getting the values in the place or a reference to it.

Rewrote to use `as_deref` in the course of inlining, so this is moot regardless.

quinn-proto/src/connection/mod.rs
djc: Okay!

djc (approved):

PR #1448, opened 2022-11-21
fix: don't bail if setting IP_RECVTOS fails

jfro:
old macOS versions don't have it, so if you try say `connection` example on macOS 10.11, it'll error with IoError 42 - Protocol not available.

I'm not sure if you're wanting a tracing log, or if you're planning on another 0.8.x release but we're still on it until we do our next non-bugfix release where we'll go to 0.9. Easily can be applied to 0.9/main too.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -116,8 +116,9 @@ fn init(io: &std::net::UdpSocket) -> io::Result<()> {
 
     let addr = io.local_addr()?;
 
-    // macos and ios do not support IP_RECVTOS on dual-stack sockets :(
-    if addr.is_ipv4() || ((!cfg!(any(target_os = "macos", target_os = "ios"))) && !only_v6(io)?) {
+    // mac and ios do not support IP_RECVTOS on dual-stack sockets :(
+    // older macos versions also don't have the flag and will error out if we don't ignore it
+    if addr.is_ipv4() || !only_v6(io)? {
         let on: libc::c_int = 1;
         let rc = unsafe {
             libc::setsockopt(
@@ -129,9 +130,13 @@ fn init(io: &std::net::UdpSocket) -> io::Result<()> {
             )
         };
         if rc == -1 {
-            return Err(io::Error::last_os_error());
+            tracing::debug!(
+                "Ignoring error setting IP_RECVTOS on socket: {:?}",
+                io::Error::last_os_error()
+            );
         }
     }
+
     #[cfg(target_os = "linux")]
     {
         if addr.is_ipv4() {
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,16 +12,16 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        rust: [stable, beta, 1.57.0]
+        rust: [stable, beta, 1.59.0]
         exclude:
           - os: macos-latest
             rust: beta
           - os: macos-latest
-            rust: 1.57.0
+            rust: 1.59.0
           - os: windows-latest
             rust: beta
           - os: windows-latest
-            rust: 1.57.0
+            rust: 1.59.0
 
     runs-on: ${{ matrix.os }}
 
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Quinn is a pure-rust, async-compatible implementation of the IETF [QUIC][quic] t
   [rustls][rustls] and [*ring*][ring]
 - Application-layer datagrams for small, unreliable messages
 - Future-based async API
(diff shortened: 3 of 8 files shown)


quinn-udp/src/unix.rs
```diff
@@ -20,25 +20,6 @@ type IpTosTy = libc::c_uchar;
...
-            libc::IPV6_V6ONLY,
-            &mut val as *mut _ as *mut _,
-            &mut len,
-        )
-    };
-    match res {
-        -1 => Err(io::Error::last_os_error()),
-        _ => Ok(val != 0),
-    }
-}
```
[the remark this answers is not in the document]
jfro: was unused after my conditional deletions, but in `main` branch it looks like it got elevated somewhere shared

quinn-udp/src/unix.rs
```diff
@@ -117,21 +98,23 @@ fn init(io: &std::net::UdpSocket) -> io::Result<()> {
...
-                libc::IPPROTO_IP,
-                libc::IP_RECVTOS,
-                &on as *const _ as _,
-                mem::size_of_val(&on) as _,
-            )
-        };
-        if rc == -1 {
-            return Err(io::Error::last_os_error());
-        }
+    let on: libc::c_int = 1;
```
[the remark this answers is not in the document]
jfro: ah good point, I can just ignore it on macos/ios as to not change linux behavior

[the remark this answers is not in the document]
Ralith: I'm not opposed to making it warn-only on Linux too for consistency. If it fails, we only don't get IPv4 ECN, which is non-critical.

quinn-udp/src/unix.rs
```diff
@@ -116,22 +97,28 @@ fn init(io: &std::net::UdpSocket) -> io::Result<()> {
...
+            libc::IP_RECVTOS,
+            &on as *const _ as _,
+            mem::size_of_val(&on) as _,
+        )
+    };
+    if rc == -1 {
+        let err = io::Error::last_os_error();
+        // ignore on macos/ios but keep linux behavior the same as before
+        if cfg!(any(target_os = "macos", target_os = "ios")) {
+            tracing::warn!("Ignoring error setting IP_RECVTOS on socket: {:?}", err);
```
Ralith, line 116: Would it make more sense for this to be `info` or even `debug`? By design, Quinn is well-behaved if ECN data is missing, and it's routinely missing due to the data being stripped along the network path even if both hosts support it.

quinn-udp/src/unix.rs
```diff
@@ -116,22 +97,28 @@ fn init(io: &std::net::UdpSocket) -> io::Result<()> {
...
-                mem::size_of_val(&on) as _,
-            )
-        };
-        if rc == -1 {
-            return Err(io::Error::last_os_error());
+    // mac and ios do not support IP_RECVTOS on dual-stack sockets :(
+    // older macos versions also don't have the flag and will error out if we don't ignore it
+    let on: libc::c_int = 1;
+    let rc = unsafe {
+        libc::setsockopt(
```
Ralith, line 104: Because `IP_RECTOS` is an IPv4-specific option, we should only be running this when `only_v6(io)` is false, not unconditionally.

Ralith (changes requested):

quinn-udp/src/unix.rs
jfro: ah my bad! i'll restore that check

quinn-udp/src/unix.rs
jfro: yeah wasn't sure what was best, maybe debug if it's not important enough?

quinn-udp/src/unix.rs
```diff
@@ -116,8 +116,9 @@ fn init(io: &std::net::UdpSocket) -> io::Result<()> {
 
     let addr = io.local_addr()?;
 
-    // macos and ios do not support IP_RECVTOS on dual-stack sockets :(
-    if addr.is_ipv4() || ((!cfg!(any(target_os = "macos", target_os = "ios"))) && !only_v6(io)?) {
+    // mac and ios do not support IP_RECVTOS on dual-stack sockets :(
+    // older macos versions also don't have the flag and will error out if we don't ignore it
+    if !only_v6(io)? {
```
Ralith, line 121: Probably worth verifying that `only_v6` successfully returns false on a macOS IPv4 socket, rather than failing with a similar weird error.

jfro: looks like it might be invalid argument on IPv4 socket... so will need to tweak this

Ralith (approved):
Thanks!

djc:
Okay, can you please:

* Squash the existing commits together
* Add a commit that bumps the MSRV for time
* Add a commit that fixes the clippy issues?

PR #1453, opened 2022-11-23
Fix GSO / GRO auto detection when IPv6 is disabled.

stormshield-damiend:
One of my Linux test machine has IPv6 disabled. While doing some bench and test on it, i noticed that GSO / GRO were not enabled. Here is a small fix to fallback to an IPv4 loopback in case it failed on IPv6 loopback.


djc (approved):
Seems like an improvement!

Not a huge fan of the multiline `match` scrutinee but also can't think of a better spelling for this particular case, so...

Ralith:
Thanks!

PR #1455, opened 2022-11-27
Fix panic on concurrent read_datagram completion

### LINKED ISSUE #1454: "ReadDatagram notified with no datagrams pending" panic
Dear all,

I am trying to receive datagrams on a Quic connection using two futures handled by a FuturesUnordered structure. Each future clones the connection and call read_datagram() on the cloned connection. I got a panic in quinn/src/connection.rs:730. 

In order to reproduce the crash you can use my forked quinn repo and [this commit](https://github.com/stormshield-pj50/quinn/commit/060aa197dca5e289bd875c0269c66459e5ec41d0).  It basically setups a datagram client/server. You can reproduce the issue by running on the same machine:

cargo run --example datagram_server
cargo run --example datagram_client

Thanks for looking at this,
Philippe

Ralith:
The previous logic assumed that a datagram must be available if the future was notified, but this might not be the case if e.g. two futures are concurrently waiting for datagrams, but only one datagram has been received, since we use notify_waiters to wake all waiting futures to be robust in the face of cancellation.

Fixes #1454.

### DIFF
--- a/quinn/Cargo.toml
+++ b/quinn/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quinn"
-version = "0.9.1"
+version = "0.9.2"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/quinn-rs/quinn"
 description = "Versatile QUIC transport protocol implementation"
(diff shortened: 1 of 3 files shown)


djc:

djc (approved):

Ralith:
Added test.

djc (approved):

PR #1456, opened 2022-11-27
Allow initial MTU to be specified in bulk benchmark

Ralith:


rklaehn (approved):
certainly an interesting parameter to have configurable...

djc (approved):

PR #1458, opened 2022-11-29
connection: wake 'stopped' streams on stream finish events

bmwill:
In the event that the SendStream Stopped future is polled once and then dropped before the remote side has stopped the stream, and instead the stream has been finished, the waker registered in the connection stopped map is never removed and is effectively "leaked" until the connection is closed. This can lead to a large amount of memory being retained when a connection is very long lived and many streams are used over the life of that connection.

This particular issue can be seen with this type of usage of SendStream::stopped: https://github.com/bmwill/anemo/blob/main/crates/anemo/src/network/request_handler.rs#L170-L177.

This is one such way to fix this memory retention, although it may not be the project's preferred way. Another option would potentially be to remove the waker from the hash map in a `Drop` impl for the `Stopped` type.

### DIFF
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -307,6 +307,13 @@ impl Future for Stopped<'_> {
     }
 }
 
+impl<'a> Drop for Stopped<'a> {
+    fn drop(&mut self) {
+        let mut conn = self.stream.conn.state.lock("Stopped::drop");
+        conn.stopped.remove(&self.stream.stream);
+    }
+}
+
 /// Future produced by [`SendStream::write()`].
 ///
 /// [`SendStream::write()`]: crate::SendStream::write
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -963,6 +963,9 @@ impl State {
                         // If the finishing stream was already dropped, there's nothing more to do.
                         let _ = finishing.send(None);
                     }
+                    if let Some(stopped) = self.stopped.remove(&id) {
+                        stopped.wake();
+                    }
                 }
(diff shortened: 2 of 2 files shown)


bmwill:
> Another option would potentially be to remove the waker from the hash map in a Drop impl for the Stopped type.

Given `Stopped` retains an `&mut SendStream` internally, i'm thinking that using a Drop impl may actually be a preferable, more foolproof solution.

quinn/src/send_stream.rs
```diff
@@ -307,6 +307,19 @@ impl Future for Stopped<'_> {
...
+impl<'a> Drop for Stopped<'a> {
+    fn drop(&mut self) {
+        let mut conn = self.stream.conn.state.lock("Stopped::drop");
+
+        if let Some(stopped) = conn.stopped.remove(&self.stream.stream) {
+            // Given this is the future that would have registered the waker we may not need to
+            // actually call Waker::wake, calling here may result in a spurious wake up but at
+            // least we can be certain that there isn't a task that will never be woken up.
+            stopped.wake();
+        }
```
Ralith, line 319: Is there a specific case of a never-woken task that you envision? I'm having trouble coming up with any cases where calling wake here is necessary.

Ralith (commented):
Great catch, thanks!

quinn/src/send_stream.rs
bmwill: I don't have a specific case, and would be completely comfortable just ignoring the removed waker. Just let me know what you'd prefer we do here and i'll make the change.

Ralith: In the interest of maintainability, let's ignore it then.

Ralith (approved):
Thanks!

Issue #1459, opened 2022-11-29
`quinn::Connection` wakers should be cleaned up more aggressively


Ralith:
`quinn::Connection` stores a variety of wakers, particularly for dispatching stream events. In some cases (e.g. #1458) these might not get disposed of on all paths, especially if futures are canceled by dropping. We should ensure that all futures which manually register wakers (or equivalent) clean them up n `Drop`. Futures built on `tokio::sync::Notify` broadcasts don't have this issue as `Notified` takes care of itself.

PR #1460, opened 2022-11-29
send-stream, recv-stream: cleanup any wakers on drop

bmwill:
Fixes: https://github.com/quinn-rs/quinn/issues/1459

### DIFF
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -313,13 +313,6 @@ impl Future for Stopped<'_> {
     }
 }
 
-impl<'a> Drop for Stopped<'a> {
-    fn drop(&mut self) {
-        let mut conn = self.stream.conn.state.lock("Stopped::drop");
-        conn.stopped.remove(&self.stream.stream);
-    }
-}
-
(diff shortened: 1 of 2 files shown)


quinn/src/send_stream.rs
```diff
@@ -261,6 +261,12 @@ impl tokio::io::AsyncWrite for SendStream {
 impl Drop for SendStream {
     fn drop(&mut self) {
         let mut conn = self.conn.state.lock("SendStream::drop");
+
+        // clean up any previously registered wakers
+        conn.finishing.remove(&self.stream);
+        conn.stopped.remove(&self.stream);
```
Ralith, line 267: Provided people aren't leaking things themselves, these are redundant to the `Drop` impls of `Finish` and `Stopped`. I think I actually like them a bit better here than in the futures since it lets us hit the mutex less, but for clarity we shouldn't have them in both places.

Ralith (changes requested):
Thanks!

Ralith (approved):
Thanks!

PR #1447, opened 2022-11-21
add FreeBSD IP_RECVDSTADDR support

lucifer9:
Fixes #1142

PR #1461, opened 2022-12-01
udp: add safe wrapper for setsockopt()

djc:
Let's get #1447 in first, but I think something like this would be a bit nicer?

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -91,79 +91,33 @@ fn init(io: SockRef<'_>) -> io::Result<()> {
     // mac and ios do not support IP_RECVTOS on dual-stack sockets :(
     // older macos versions also don't have the flag and will error out if we don't ignore it
     if is_ipv4 || !io.only_v6()? {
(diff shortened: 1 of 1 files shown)


quinn-udp/src/unix.rs
```diff
@@ -91,110 +91,42 @@ fn init(io: SockRef<'_>) -> io::Result<()> {
...
-        unsafe {
-            libc::setsockopt(
-                io.as_raw_fd(),
-                libc::SOL_UDP,
-                libc::UDP_GRO,
-                &on as *const _ as _,
-                mem::size_of_val(&on) as _,
-            )
-        };
+        // TODO: handle error here?
```
Ralith, line 102: Per the comment ("opportunistically") ignoring an error is intended here. We want to turn it on if we can, but if we can't that's okay too.

Ralith (commented):
Nice!

quinn-udp/src/unix.rs
djc: 👍

Ralith (approved):

PR #1462, opened 2022-12-02
Add support for IP_RECVORIGDSTADDR socket option.

PiotrSikora:
When enabled on a socket with IP_TRANSPARENT, it populates metadata
with the original destination (IP:port) for traffic redirected with
TPROXY.

Signed-off-by: Piotr Sikora <piotr@aviatrix.com>

### DIFF
--- a/quinn-udp/src/fallback.rs
+++ b/quinn-udp/src/fallback.rs
@@ -86,6 +86,7 @@ impl UdpSocketState {
             addr: addr.as_socket().unwrap(),
             ecn: None,
             dst_ip: None,
+            dst_addr: None,
         };
         Ok(1)
     }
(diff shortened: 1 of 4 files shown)


Ralith:
What's the use case for this?

PiotrSikora:
> What's the use case for this?

This allows a QUIC/UDP proxy that intercepted traffic using `TPROXY` to forward traffic towards the original destination.

Note that if the original socket doesn't have `IP_TRANSPARENT` socket option enabled, then this won't receive any new CMSGs, so the overhead of having this enabled if pretty small.

Ralith:
quinn-udp is primarily intended to support the requirements of Quinn. Is there a way you imagine Quinn using this functionality? If not, forking/vendoring the logic your application needs might make more sense.

PR #1466, opened 2022-12-06
Clippy fixes

StygianLightning:
I checked the examples and found a typo in one of the examples and an allowed lint as a workaround to a bug that's since been fixed. The other commits fix some clippy lints that show up on nightly.


djc (approved):
Nice, thanks!

StygianLightning:
No worries! 
There's one thing I'm not sure about, and that's the use of these allow annotations:
>     #[allow(clippy::unnecessary_cast)] // self.hdr.msg_controllen defined as size_t
I thought size_t was platform specific, but I looked it up earlier, and [it's defined unconditionally as a type alias for usize](https://docs.rs/libc/latest/libc/type.size_t.html), so maybe it would make sense to remove these annotations and the cast.

Ralith:
Try it and see what CI thinks?

djc (approved):

StygianLightning:
Huh, Mac doesn't like that for some reason. Gonna fix it tonight.

StygianLightning:
I fixed the three instances that failed on macOS and reduced the scope of the two #[allow] annotations from my original commit.

Ralith (approved):
Thanks!

PR #1468, opened 2022-12-12
Update Wording About Quick Being a Draft

### LINKED ISSUE #1467: Is QUIC Still a Draft?
I see this line in the docs:

https://github.com/quinn-rs/quinn/blob/389d7ce5067fbef460f0730861340b15f13d4803/docs/book/src/quic.md?plain=1#L4

From what I can tell from the IETF tracker, it seems that now QUIC is no longer in the draft phase is that right?

zicklag:
Closes #1467.


Ralith (approved):
Thanks!

PR #1474, opened 2022-12-24
Fail gracefully on attempts to connect to an unspecified IP address

Ralith:
Silent failure here has tripped up multiple users, most recently in https://github.com/quinn-rs/quinn/issues/1472.


djc (approved):
Nice!

PR #1476, opened 2023-01-06
Re-export udp::{RecvMeta, UdpState} for AsyncUdpSocket

XOR-op:
To implement `AsyncUdpSocket` Trait from `quinn` crate, we must use the types `RecvMeta` and `UdpState`. However, current version of quinn does not re-export them, so anyone who wants to implement the trait needs importing another `quinn-udp` crate. This PR fixes this problem by re-export them in `quinn` crate.

### DIFF
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -63,6 +63,7 @@ pub use proto::{
     ConnectionClose, ConnectionError, EndpointConfig, IdleTimeout, ServerConfig, StreamId,
     Transmit, TransportConfig, VarInt,
 };
+pub use udp::{RecvMeta, UdpState};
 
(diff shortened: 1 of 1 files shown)


XOR-op:
Do you mean we should export the whole `udp` module? That is a good idea since they are all helpful when implementing a custom `AsyncUdpSocket`, and in fact, there are only 4 structs + 1 constant.

djc:
I'd prefer not to use a `*` import, but feel free to export all the items that are useful to you.

XOR-op:
I finally re-export `quinn-udp` as a submodule of quinn, which does not mess up the root namespace.

Ralith (approved):

PR #1482, opened 2023-01-24
Improve error when connecting with an unsupported version

### LINKED ISSUE #1481: The custom version of the api is weird
Hi, I recently used quinn-proto to replace rustls with noise protocol and successfully implemented PoC with `Noise_IKpsk2_25519_ChaChaPoly_BLAKE2b`.  But when I set up a custom version, I meet a problem.

Before I started setting up the custom version, I had the following test code
```rs
let bind_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 0);
let quinn_s_cfg = quinn::ServerConfig::with_crypto(Arc::new(s_cfg));
let quinn_c_cfg = quinn::ClientConfig::new(Arc::new(c_cfg));

let s_endpoint = quinn::Endpoint::server(quinn_s_cfg, bind_addr).unwrap();
let s_addr = s_endpoint.local_addr().unwrap();

let handle = task::spawn(async move {
    let mut c_endpoint = quinn::Endpoint::client(bind_addr).unwrap();
    c_endpoint.set_default_client_config(quinn_c_cfg);

    let conn = c_endpoint.connect(s_addr, "").unwrap().await.unwrap();

    let (mut send, mut recv) = conn.open_bi().await.unwrap();
    send.write_all(b"ping").await.unwrap();

    let mut buf = [0u8; 4];
    recv.read_exact(&mut buf).await.unwrap();
    assert_eq!(&buf[..4], b"pong");
});

let conn = s_endpoint.accept().await.unwrap().await.unwrap();
let (mut send, mut recv) = conn.accept_bi().await.unwrap();

let mut buf = [0u8; 4];
recv.read_exact(&mut buf).await.unwrap();
assert_eq!(&buf[..4], b"ping");

send.write_all(b"pong").await.unwrap();

handle.await.unwrap();
```

After a quick scan of the doc, I modified some of the code.

Set the default version for ClientConfig and the supported version list for ServerConfig's Endpoint.
```rs
let mut quinn_c_cfg = quinn::ClientConfig::new(Arc::new(c_cfg));
quinn_c_cfg.version(DEFAULT_QUIC_VERSION);

let mut endpoint_cfg = EndpointConfig::default();
endpoint_cfg.supported_versions(vec![DEFAULT_QUIC_VERSION]);
let s_endpoint = Endpoint::new(
    endpoint_cfg,
    Some(quinn_s_cfg),
    UdpSocket::bind(bind_addr).unwrap(),
    TokioRuntime,
)
.unwrap();
```

At this point, the test failed. When I used RUST_LOG=trace, I saw that the relevant packets were being dropped on the **client** side.

```
DEBUG quinn_proto::endpoint: dropping packet with unsupported version
```

Then I found it:

https://github.com/quinn-rs/quinn/blob/e5a5d3eeb80f637e6bb55730456efc452a9ac9d2/quinn-proto/src/endpoint.rs#L165-L168

It checks the server_config on the client side, but I explicitly set a specific version in the client_config.

The workaround is to add a server_config to the client as well, but the client doesn't have a real server private key, it has to mock an server_config.

Ralith:
If the cryptographic layer supports a version not in the local endpoint's supported versions list, a connection was initiated but all packets returned by a compatible peer were dropped. This is a relatively easy configuration error to make.

Fixes #1481.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -369,6 +369,9 @@ impl Endpoint {
         if remote.port() == 0 || remote.ip().is_unspecified() {
             return Err(ConnectError::InvalidRemoteAddress(remote));
         }
+        if !self.config.supported_versions.contains(&config.version) {
+            return Err(ConnectError::UnsupportedVersion);
+        }
 
         let remote_id = RandomConnectionIdGenerator::new(MAX_CID_SIZE).generate_cid();
         trace!(initial_dcid = %remote_id);
@@ -830,7 +833,7 @@ pub enum ConnectError {
     /// Use `Endpoint::connect_with` to specify a client configuration.
     #[error("no default client config")]
     NoDefaultClientConfig,
-    /// The cryptographic layer does not support the specified QUIC version
+    /// The local endpoint does not support the QUIC version specified in the client configuration
     #[error("unsupported QUIC version")]
     UnsupportedVersion,
 }


quinn-proto/src/endpoint.rs
```diff
@@ -369,6 +369,9 @@ impl Endpoint {
         if remote.port() == 0 || remote.ip().is_unspecified() {
             return Err(ConnectError::InvalidRemoteAddress(remote));
         }
+        if !self.config.supported_versions.contains(&config.version) {
```
djc, line 372: Style nit: I think `else if` would make sense here?

djc (approved):

quinn-proto/src/endpoint.rs
djc: Fair enough. I prefer the increased density over syntactic independence, I think. Also in my mind using `else if` would group these more as a set of preconditions, whereas the syntactic independence means I have to do more work parsing in my head because I consider each of them independently.

Anyway, either is fine.

Ralith: (needs reapproval though)

djc (approved):

PR #1484, opened 2023-01-27
Fix usage message

### LINKED ISSUE #1483: perf_server expects private key to be in DER format, but --help says PEM

dzvon:
Closes #1483


dzvon:
Fixed lint test in #1485.

PR #1493, opened 2023-02-16
Document how to close a stream correctly

flub:
It is not trivial to correctly close a stream without errors and knowing everything is fully acknowledged.  Describe some details around this.

### DIFF
--- a/quinn/src/recv_stream.rs
+++ b/quinn/src/recv_stream.rs
@@ -51,16 +51,18 @@ use crate::{
 /// // In the receiving task
 /// let mut buf = [0u8; 10];
 /// let data = recv_stream.read_exact(&mut buf).await?;
-/// if recv_stream.read_to_end(0).await.is_err() {
-///     recv_stream.stop(0u8.into()).ok();
-/// }
+/// recv_stream.read_to_end(0).await?;
 /// # Ok(())
 /// # }
 /// ```
 ///
-/// Note that in this example the receiver sends an error to the sender if it received some
-/// unexpected data, interrupting any further attempts to send data.  But crucially only
-/// after it attempted to read the end of the stream.
+/// Note that in this example the receiver is dropped because [`RecvStream::read_to_end`]
+/// takes ownership, this results in the implicit call to `stop(0)` interrupting any further
+/// attempts to send data.  Crucially the `stop` call only happens after it attempted to
+/// read the entire stream.
+///
+/// [`RecvStream::read_chunk`] could be used instead which does not take ownership and
+/// allows using an explit call to [`RecvStream::stop`] with a custom error code.
 ///
 /// [`ReadError`]: crate::ReadError
 /// [`stop()`]: RecvStream::stop
--- a/quinn/src/recv_stream.rs
+++ b/quinn/src/recv_stream.rs
@@ -57,9 +57,9 @@ use crate::{
 /// ```
 ///
 /// Note that in this example the receiver is dropped because [`RecvStream::read_to_end`]
-/// takes ownership, this results in the implicit call to `stop(0)` interrupting any further
-/// attempts to send data.  Crucially the `stop` call only happens after it attempted to
-/// read the entire stream.
+/// takes ownership.  This results in the implicit call to `stop(0)` if the stream was not
+/// finished, interrupting any further attempts to send data.  Crucially the `stop` call
+/// only happens after it attempted to read the entire stream.
 ///
 /// [`RecvStream::read_chunk`] could be used instead which does not take ownership and
 /// allows using an explit call to [`RecvStream::stop`] with a custom error code.
--- a/quinn/src/recv_stream.rs
+++ b/quinn/src/recv_stream.rs
@@ -25,28 +25,36 @@ use crate::{
 ///
 /// When a stream is expected to be closed gracefully the sender should call
 /// [`SendStream::finish`].  However there is no guarantee the connected [`RecvStream`] will
-/// receive this End Of File (EOF) indicator in the same QUIC frame as the last frame which
+/// receive the "finished" notification in the same QUIC frame as the last frame which
 /// carried data.
 ///
-/// Even if the application layer logic already knows it read all the data because it
-/// does its own framing, it must still read from the [`RecvStream`] until EOF.  Otherwise
-/// it risks inadvertently calling [`RecvStream::stop`] if it drops the stream.  And calling
-/// [`RecvStream::stop`] could result in the connected [`SendStream::finish`] call failing
-/// with a [`WriteError::Stopped`] error.
+/// Even if the application layer logic already knows it read all the data because it does
+/// its own framing, it must still read from the [`RecvStream`] until `AsyncRead` signals
+/// End Of File (EOF).  Otherwise it risks inadvertently calling [`RecvStream::stop`] if it
+/// drops the stream.  And calling [`RecvStream::stop`] could result in the connected
+/// [`SendStream::finish`] call failing with a [`WriteError::Stopped`] error.
 ///
 /// For example if exactly 10 bytes are to be read, you still need to explicitly read EOF:
 ///
-/// ```ignore
+/// ```no_run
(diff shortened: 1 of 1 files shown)


djc (dismissed):
Thanks, this looks great! I'll give @Ralith a chance to review as well.

quinn/src/recv_stream.rs
```diff
@@ -21,8 +21,40 @@ use crate::{
 /// - A variant of [`ReadError`] has been yielded by a read call
 /// - [`stop()`] was called explicitly
 ///
+/// # Closing a stream
+///
+/// When a stream is expected to be closed gracefully the sender should call
+/// [`SendStream::finish`].  However there is no guarantee the connected [`RecvStream`] will
+/// receive this End Of File (EOF) indicator in the same QUIC frame as the last frame which
+/// carried data.
```
Ralith, line 29: "EOF" isn't really a QUIC term. Maybe `the "finished" notification`?

quinn/src/recv_stream.rs
```diff
@@ -21,8 +21,40 @@ use crate::{
...
+///
+/// So even if the application layer logic already knows it read all the data because it
```
Ralith, line 31: nit:
```suggestion
/// Even if the application layer logic already knows it read all the data because it
```

quinn/src/recv_stream.rs
```diff
@@ -21,8 +21,40 @@ use crate::{
...
+/// does its own framing, it must still read from the [`RecvStream`] until EOF.  Otherwise
```
Ralith, line 32:
```suggestion
/// does its own framing, it must still read from the [`RecvStream`] until finished.  Otherwise
```

quinn/src/recv_stream.rs
```diff
@@ -21,8 +21,40 @@ use crate::{
...
+/// it risks inadvertently calling [`RecvStream::stop`] if it drops the stream.  And calling
+/// [`RecvStream::stop`] could result in the connected [`SendStream::finish`] call failing
+/// with a [`WriteError::Stopped`] error.
+///
+/// For example if exactly 10 bytes are to be read, you still need to explicitly read EOF:
```
Ralith, line 37:
```suggestion
/// For example if exactly 10 bytes are to be read, you still need to explicitly read until finished:
```

quinn/src/recv_stream.rs
```diff
@@ -21,8 +21,40 @@ use crate::{
...
+///
+/// ```ignore
```
Ralith, line 39: This is at risk of bitrot. Instead, we could start the block with something like 
````
```no_run
# fn foo(send_stream: &mut SendStream, recv_stream: &mut RecvStream) {
````

That way we're at least type-checking it.

File: quinn/src/recv_stream.rs [hunk 6]
```diff
@@ -21,8 +21,40 @@ use crate::{
...
+/// // In the sending task
+/// send_stream.write(&b"0123456789"[..]).await?;
+/// send_stream.finish()?;
+///
+/// // In the receiving task
+/// let mut buf = [0u8; 10];
+/// let data = recv_stream.read_exact(&mut buf);
+/// if let Some(chunk) = recv_stream.read_chunk(8, false).await? {
+///     reader.stop(0u8.into()).ok();
+/// }
```
Ralith, line 49: This is a little more terse, and has exactly the same effect if the `recv_stream` is dropped after `?` bails out:
```suggestion
/// recv_stream.read_to_end(0).await?;
```

Ralith (changes requested):
Thanks!

quinn/src/recv_stream.rs
flub: I did consider that option, but preferred the `.read_chunk()` approach:

- The amount of data being read is unlimited and allocated in a `Vec` when using `.read_to_end()`.  The receiver has no control over this.
- `.read_chunk(n, false)` has the benefit there won't even be any delays due to reordering or retransmission.
- Returning from the function may or may not drop the reader, that depends on the context of the caller.  For an example I prefer the explicit call to `.stop()`.

quinn/src/recv_stream.rs
flub: True, and here I agree because this describes QUIC behaviour.  Later when talking about how to use the stream I would keep the EOF terminology however because it is an `AsyncRead` concept and that is what you're interacting with.

flub:
Good call on actually compiling the example, there were quite a few errors!

I've left some EOF wording in there since I think that is the terminology when using stream I think.  Let me know if you'd still like to remove this.

quinn/src/recv_stream.rs
flub: Huh, sorry I did misread `read_to_end`.  Makes sense, I'd still leave in the explicit stop though for the example.

quinn/src/recv_stream.rs
Ralith: > it is an `AsyncRead` concept and that is what you're interacting with.

None of the methods in the example involve `AsyncRead`, nor is `AsyncRead` in any way necessary to use `RecvStream`s. Due to the baked in `io::Error` it's really a second-class API, exposed only for compatibility.

It's true that EOF is usually a familiar concept all the same, but I worry that it will be confusing to have inconsistent terminology.

quinn/src/recv_stream.rs
Ralith: I suppose it does make the intent significantly clearer; fine by me.

quinn/src/recv_stream.rs
Ralith: Alternatively, maybe ``read to the end of the [`RecvStream`]``? That's more consistent with the method name, while still avoiding introduction of inconsistent terms.

quinn/src/recv_stream.rs
```diff
@@ -21,8 +21,49 @@ use crate::{
...
+/// let data = recv_stream.read_exact(&mut buf).await?;
+/// if recv_stream.read_to_end(0).await.is_err() {
+///     recv_stream.stop(0u8.into()).ok();
+/// }
+/// # Ok(())
+/// # }
+/// ```
+///
+/// Note that in this example the receiver sends an error to the sender if it received some
+/// unexpected data, forcing the stream to finish in an error state.
```
Ralith, line 61: > forcing the stream to finish in an error state.

This might be a bit misleading. If the sender has finished the stream, our stopping it no longer has any meaning, so our having done so is unlikely to be visible. The core function of `stop` is to interrupt any future attempts to transmit on the stream, which is at that point moot.

A more correct statement might be "..., interrupting any further attempts to send data."

quinn/src/recv_stream.rs
flub: Ok, I've fully removed this.

flub:
Please take another look, I've adopted the suggested wording changes.

Ralith (changes requested):
Thanks!

Ralith (approved):

Ralith (approved):

quinn/src/recv_stream.rs
```diff
@@ -21,8 +21,51 @@ use crate::{
...
+/// # {
+/// // In the sending task
+/// send_stream.write(&b"0123456789"[..]).await?;
+/// send_stream.finish().await?;
+///
+/// // In the receiving task
+/// let mut buf = [0u8; 10];
+/// let data = recv_stream.read_exact(&mut buf).await?;
+/// if recv_stream.read_to_end(0).await.is_err() {
+///     recv_stream.stop(0u8.into()).ok();
```
Ralith, line 55: Ah, of course this call is impossible because `read_to_end` takes ownership. Perhaps we should change it to take `&mut self`...

flub:
Apologies for just using the github suggestions and letting CI find the bugs, which is just painful and slow.

I've fixed up the example locally and it all works again.  I stuck with `.read_to_end(0)` and updated the description saying when it calls `.stop()`.

Ralith (approved):
Thanks for seeing this through!

Issue #1488, opened 2023-01-30
Support BoringSSL


nmittler:
This is important for projects that require FIPS compliance. This would add an optional feature that would provide a BoringSSL implementation of the crypto APIs.

Can potentially leverage Cloudflare's [boring](https://github.com/cloudflare/boring), which already has FIPS as a feature.

nmittler:
FYI I've actually already started work on this in [my own branch](https://github.com/nmittler/quinn/tree/boring).

djc:
I'm not convinced that we'll want to maintain a boring backend in this repository, but obviously if the `crypto::Session` trait needs to be adapted to implement this we'd be open to any improvements that need to be made!

nmittler:
@djc understood, thanks. Would you be open to having a separate repository for this under  [quinn-rs](https://github.com/quinn-rs)?

djc:
Hmm, not sure. If we put it in quinn-rs, I wonder if that gives off an implicit endorsement from the maintainers of the org, which I'm not sure makes sense for this case. @Ralith what do you think?

BTW, I think we'd be happy to link it from the docs and/or README.

Ralith:
I'm not opposed to having it under the quinn-rs org so long as maintenance expectations/endorsement are clearly documented, since it'd ensure we can at least facilitate changes of maintainership down the line if needed. We also have other unpolished repos in the org already.

+1 to FIPSifying rustls potentially being a lower-maintenance option for everyone, though.

djc:
Seems reasonable to me...

stevenctl:
Would a branch on the main repo simplify piggy-backing on your test infra?

nmittler:
> Would a branch on the main repo simplify piggy-backing on your test infra?

Good question regarding test infra. I certainly don't want to reinvent CI for this effort. I'm not sure a branch is really sustainable, however, since we'd want it to be a versioned entity. Regardless, having the code separate (branch or repo) also raises the concern of keeping it in sync with the latest `quinn`. Lots of details to sort out.

djc:
I agree that we shouldn't have anything live on a branch forever.

Ralith:
A test suite crate that's parameterized on crypto stack sounds like a graceful solution, though it might take some effort. Graceful integration with the built-in unit testing harness seems unlikely, for one. Care would be needed to avoid significantly degrading the testing experience, though maybe there's also opportunity here to make things cleaner than the status quo.

nmittler:
[...]

PR #1494, opened 2023-02-16
[WIP] Introducing conformance testing crate

nmittler:
Based on an idea proposed in #1488. Creates a separate crate for testing crypto providers. The new crate `quinn-test` contains a copy of the original testing code from `quinn-proto`, but refactored to introduce the idea of a `CryptoProvider` that is used to provide crypto client/server config for the endpoints. A `Suite` combines a client and server `CryptoProvider` and contains all of the testing methods.

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -24,6 +24,7 @@ use crate::{
 /// for higher bandwidths and latencies increases worst-case memory consumption, but does not impair
 /// performance at lower bandwidths and latencies. The default configuration is tuned for a 100Mbps
 /// link with a 100ms round trip time.
+#[cfg_attr(any(test, feature = "integration-tests"), quinn_visibility::make(pub))]
 pub struct TransportConfig {
     pub(crate) max_concurrent_bidi_streams: VarInt,
     pub(crate) max_concurrent_uni_streams: VarInt,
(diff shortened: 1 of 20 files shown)


nmittler:
This is a work-in-progress, but I wanted to send it out ealy for feedback. I had to make a number of things public to make this work, so I expect that will be a focal point of discussion.

I included a single `example` test that runs the `lifecycle` test with the rustls provider.

quinn-proto/src/config.rs
```diff
@@ -25,27 +25,27 @@ use crate::{
...
-
-    pub(crate) persistent_congestion_threshold: u32,
-    pub(crate) keep_alive_interval: Option<Duration>,
-    pub(crate) crypto_buffer_size: usize,
-    pub(crate) allow_spin: bool,
-    pub(crate) datagram_receive_buffer_size: Option<usize>,
-    pub(crate) datagram_send_buffer_size: usize,
-
-    pub(crate) congestion_controller_factory: Box<dyn congestion::ControllerFactory + Send + Sync>,
+    pub max_concurrent_bidi_streams: VarInt,
```
djc, line 28: Why do we need this? This seems like a blocker for me -- if we need to make all the currently crate-private API public for this, I don't think this idea is workable.

nmittler: The tests call these methods directly and now that the tests are in a separate crate, I assume that means they have to be public (although I'm still pretty new to rust). I'm definitely open to alternatives, if there is a better approach.

nmittler:
@djc PTAL. I managed to get this working by using slightly modified version of the [visibility macro](https://docs.rs/visibility/latest/visibility/).

djc:
I'm not that excited about using the visibility macro for this. I still think the [reverse dependency](https://github.com/quinn-rs/quinn/issues/1488#issuecomment-1434883862) is probably a less invasive solution.

nmittler:
@djc understood. I'll go ahead and close this and will open others shortly.

PR #1496, opened 2023-02-22
Support external crypto implementations.

nmittler:
These changes are needed for the BoringSSL crypto provider (#1488), which will reside in a separate repository.

### DIFF
--- a/quinn-proto/src/crypto.rs
+++ b/quinn-proto/src/crypto.rs
@@ -56,7 +56,8 @@ pub trait Session: Send + 'static {
     ///
     /// This should be called with the contents of `CRYPTO` frames. If it returns `Ok`, the
     /// caller should call `write_handshake()` to check if the crypto protocol has anything
-    /// to send to the peer.
+    /// to send to the peer. This method will only return `true` the first time that
+    /// handshake data is available. Future calls will always return false.
     ///
     /// On success, returns `true` iff `self.handshake_data()` has been populated.
     fn read_handshake(&mut self, buf: &[u8]) -> Result<bool, TransportError>;
--- a/quinn-proto/src/transport_error.rs
+++ b/quinn-proto/src/transport_error.rs
@@ -49,7 +49,7 @@ pub struct Code(u64);
 
 impl Code {
     /// Create QUIC error code from TLS alert code
-    pub(crate) fn crypto(code: u8) -> Self {
+    pub fn crypto(code: u8) -> Self {
         Code(0x100 | u64::from(code))
     }
 }


nmittler:
@djc I think the visibility changes here make sense, since they're effectively part of the crypto provider API. WDYT?

quinn-proto/src/transport_error.rs
```diff
@@ -49,7 +49,7 @@ pub struct Code(u64);
 
 impl Code {
     /// Create QUIC error code from TLS alert code
-    pub(crate) fn crypto(code: u8) -> Self {
+    pub fn crypto(code: u8) -> Self {
```
nmittler, line 52: Not sure how you want to handle this one. If we want to be picky, we could make this a `try_from` pattern, as is done elsewhere in the code.

quinn-proto/src/transport_parameters.rs
```diff
@@ -68,30 +68,30 @@ macro_rules! make_struct {
         /// Transport parameters used to negotiate connection-level preferences between peers
         #[derive(Debug, Copy, Clone, Eq, PartialEq)]
         pub struct TransportParameters {
-            $($(#[$doc])* pub(crate) $name : VarInt,)*
+            $($(#[$doc])* pub $name : VarInt,)*
 
             /// Does the endpoint support active connection migration
-            pub(crate) disable_active_migration: bool,
+            pub disable_active_migration: bool,
```
djc, line 74: I'm inclined to make methods for these so we can guarantee that these aren't modified from the outside, what do you think? @Ralith any opinions?

perf/src/noprotection.rs
```diff
@@ -129,6 +130,10 @@ impl crypto::ClientConfig for NoProtectionClientConfig {
 
         Ok(Box::new(NoProtectionSession::new(tls)))
     }
+
+    fn as_any(&self) -> &dyn Any {
```
djc, line 134: `&dyn Any` + downcasting seems relatively ugly. Why can't we do a `&dyn crypto::ClientConfig` thing here?

quinn-proto/src/transport_parameters.rs
nmittler: @djc why not just rely on mutability of the struct?

perf/src/noprotection.rs
nmittler: Thanks @djc ... unfortunately, I'm still a bit of a rust newbie and all the downcasting examples I've seen use an intermediate conversion to Any. My attempts at hacking around it have so far been unsuccessful. Do you have an example to illustrate your suggestion?

nmittler: Nevermind, I think I found a work-around that no longer requires downcasting.

nmittler:
@djc I believe I've addressed your concerns. PTAL

quinn/src/endpoint.rs
```diff
@@ -107,7 +107,8 @@ impl Endpoint {
         Self::new_with_runtime(config, server_config, Box::new(socket), Arc::new(runtime))
     }
 
-    fn new_with_runtime(
+    /// Construct an endpoint with arbitrary configuration and pre-constructed async socket.
+    pub fn new_with_runtime(
```
nmittler, line 111: This was needed to get my copy of the quinn examples working with boringssl.

nmittler:
@djc any other comments here? I'd like to get this in quickly, since there are 2+ follow-on PRs that depend on this.

nmittler:
@Ralith did you want to take a look as well?

djc (commented):
@Ralith would be great to get your feedback on these changes, too.

quinn-proto/src/transport_parameters.rs
```diff
@@ -186,10 +247,14 @@ impl TransportParameters {
 ///
 /// This is communicated as a transport parameter during TLS session establishment.
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
-pub(crate) struct PreferredAddress {
+pub struct PreferredAddress {
+    /// The server's IPv4 address.
```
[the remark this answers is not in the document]
nmittler: Done.

quinn-proto/src/transport_error.rs
Ralith: I think this is infallible, and suitable for exposure as-is. From the RFC:

> CRYPTO_ERROR (0x0100-0x01ff):
>    The cryptographic handshake failed. A range of 256 values is reserved for carrying error codes specific to the cryptographic handshake that is used. Codes for errors occurring when TLS is used for the cryptographic handshake are described in [Section 4.8](https://www.rfc-editor.org/rfc/rfc9001#section-4.8) of [[QUIC-TLS](https://www.rfc-editor.org/rfc/rfc9000.html#QUIC-TLS)].

Note in particular that the cryptographic layer is free to define these as it sees fit.

quinn-proto/src/transport_parameters.rs
```diff
@@ -186,14 +247,34 @@ impl TransportParameters {
...
 
 impl PreferredAddress {
+    /// Returns the server's IPv4 address, if available.
+    pub fn address_v4(&self) -> Option<SocketAddrV4> {
+        self.address_v4.map_or(None, |addr| Some(addr))
+    }
+
+    /// Returns the server's IPv6 address, if available.
+    pub fn address_v6(&self) -> Option<SocketAddrV6> {
+        self.address_v6.map_or(None, |addr| Some(addr))
```
Ralith, line 265: Surely the `map_or` here and above can be discarded entirely?

quinn/src/endpoint.rs
Ralith: Can you elaborate on why?

Ralith (changes requested):

quinn-proto/src/transport_parameters.rs
```diff
@@ -94,6 +94,67 @@ macro_rules! make_struct {
...
+                self.stateless_reset_token
+            }
+
+            /// The server's preferred address for communication after handshake completion
+            #[inline]
+            pub fn preferred_address(&self) -> Option<PreferredAddress> {
+                self.preferred_address
+            }
+        }
+
```
Ralith, line 157: Can you elaborate on the motivation for exposing these? Most of these are used purely by the transport layer, IIRC.

quinn-proto/src/token.rs
```diff
@@ -119,6 +123,11 @@ impl ResetToken {
         result.copy_from_slice(&signature[..RESET_TOKEN_SIZE]);
         result.into()
     }
+
+    /// Returns an empty token.
+    pub fn none() -> Self {
+        Self([0; RESET_TOKEN_SIZE])
+    }
```
Ralith, line 130: Is this necessary, given the existing `From<[u8; RESET_TOKEN_SIZE]>` impl?

quinn/src/endpoint.rs
nmittler: I'm trying to make helper methods, similar to what quinn has for rustls. For example, this is a boring version of a helper method to create a client endpoint:

```rust
pub fn client_endpoint(addr: SocketAddr) -> io::Result<quinn::Endpoint> {
        let socket = std::net::UdpSocket::bind(addr)?;
        let runtime = default_runtime()
            .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "no async runtime found"))?;
        quinn::Endpoint::new_with_runtime(
            default_endpoint_config(),
            None,
            runtime.wrap_udp_socket(socket)?,
            runtime,
        )
    }
```

For this to work, `Endpoint::new_with_runtime` has to be public. The other method `Endpoint::new_with_abstract_socket` requires that the caller has a concrete, unboxed `tokio` udp socket. I'm still pretty new to rust and I suspect I can make something work, but it seemed (at least initially) logical for `new_with_runtime` to be public.

nmittler:
@Ralith PTAL

quinn-proto/src/transport_parameters.rs
djc: Wouldn't it be better to use the `TransportParameters::read()` and `write()` methods, which are already public? Looks like that API needs a byte buffer.

nmittler: Only select fields of transport parameters are included in the early data context, so I don't read/write will work.

djc: Have you found code that does the equivalent of this on the rustls code path? All other things being equal, adding code for this to `TransportParameters` seems like it could make sense.

nmittler: I copied the logic from Google's quiche library: https://github.com/google/quiche/blob/6a93efcba4e4339bb3b4fe14f340f4e3f60b28c7/quiche/quic/core/crypto/transport_parameters.cc#L1553

djc: @Ralith huh, do you remember how/where this is implemented for our rustls code path?

nmittler:
@djc if I'm reading the lint error correctly, I think it's an existing issue with `main`. Can you take a look?

djc:
Yup, new clippy lints from Rust 1.68. Addressed in #1506.

djc:
We'll get to it when we get to it -- we're maintaining Quinn on a voluntary basis and you're adding code paths here that are somewhat far out of the core use case. Some patience will be required, just have your downstream code depend on a Git commit for now?

nmittler:
@djc understood. Thanks!

quinn-proto/src/transport_parameters.rs
Ralith: Interesting; BoringSSL is taking responsibility for not accepting 0-RTT on the server if the server's configuration has changed, as a conservative solution. Our current behavior is to not check this; it's the user's responsibility not to change transport parameters in illegal ways without also invalidating resumption tokens via rustls. Quiche's approach is a bit less error prone, but also more restrictive than is strictly necessary since it prevents relaxing flow control limits.

A quick and easy way to make BoringSSL happy here would be to pass an empty or fixed context. I don't see a security risk to doing so as far as QUIC is concerned; the biggest hazard is that someone will illegally reduce a limit and cause connections resumed across the change to fail with `PROTOCOL_VIOLATION`. Perhaps there's a larger concern for higher-level protocols, but those can customize the rustls behavior directly already.

We could match Quiche's behavior without as much API expansion by offering an accessor for an opaque byte string, but I'm not convinced even that much is necessary. @nmittler do you have an opinion?

Ralith:
Thanks for your patience and persistence here; I've been preoccupied, but I do want to land this.

nmittler:
@Ralith @djc transport parameters have been reverted. PTAL

quinn/src/endpoint.rs
djc: If you build the `std::net::UdpSocket`, it looks like you can just call `new()` instead of `new_with_runtime()`? It wraps the socket for you.

nmittler: Ok, I guess the last issue is that `default_runtime` returns an `Arc<dyn Runtime>`, but `new` requires an `impl Runtime`.  I can dereference the `Arc`, but that gives me a `&dyn Runtime`. Anyway, this seems wrong, because `new` wraps the value in another `Arc`, which would effictively move ownership IIUC.

I've seen other callsites of`new` and just directly pass in `TokioRuntime`. However, if we theoretically want to support other runtimes, using `default_runtime` seems better.

Still learning rust, so I'm probably missing something.

Ralith: Ah, I think this is a minor flaw in our API design. We could hack around the problem by defining `impl Runtime for Arc<dyn Runtime>`, but this causes gratuitous double indirection since it'll get re-`Arc`ed. I think we should replace `impl Runtime` with `Arc<dyn Runtime>` in the existing public APIs.

I think it's fine for test code to use `new` and hardcode `TokioRuntime` for now, and we can keep `new_with_rutnime` private and pursue API cleanup in future work.

quinn-proto/src/crypto.rs
```diff
@@ -56,7 +56,8 @@ pub trait Session: Send + 'static {
     ///
     /// This should be called with the contents of `CRYPTO` frames. If it returns `Ok`, the
     /// caller should call `write_handshake()` to check if the crypto protocol has anything
-    /// to send to the peer.
+    /// to send to the peer. This method will only return `true` the first time that
+    /// handshake data is available. Future calls will always return false.
```
Ralith, line 60:
```suggestion
    /// handshake data is available. Future calls may return false.
```

Unless you specifically need this guarantee I'd rather preserve implementation flexibility.

Ralith (changes requested):

quinn/src/endpoint.rs
nmittler: SGTM. done.

quinn-proto/src/crypto.rs
Ralith: Ah, weird. No harm in documenting that for now then.

Ralith (approved):

nmittler:
@Ralith @djc thanks for the review! I guess this will be included in the next patch release?

djc:
Yes -- if you would like one sooner rather than later, can you send a PR to bump the quinn-proto version number?

nmittler:
@djc thanks ... I've opened https://github.com/quinn-rs/quinn/pull/1520.

PR #1497, opened 2023-02-24
docs(quinn): remove restriction to tokio

mxinden:
With https://github.com/quinn-rs/quinn/pull/1364 `quinn` is no longer restricted to `tokio`. Commit updates the crate heading accordingly.

Note that I don't feel strongly about this. Just a drive-by contribution.


Ralith (approved):
Thanks!

PR #1499, opened 2023-02-26
Clippy 1.67

djc:


perf/src/bin/perf_server.rs
```diff
@@ -53,7 +53,7 @@ async fn main() {
 
 async fn run(opt: Opt) -> Result<()> {
     let (key, cert) = match (&opt.key, &opt.cert) {
-        (&Some(ref key), &Some(ref cert)) => {
+        (Some(key), Some(cert)) => {
```
djc, line 56: @Ralith I remember you have opinions on this but they don't align with my own. I think you care more than I do so while this looks like a good change to me, happy to defer to your judgement here.

quinn-proto/src/connection/mod.rs
```diff
@@ -1337,7 +1337,7 @@ impl Connection {
         let loss_delay = cmp::max(rtt.mul_f32(self.config.time_threshold), TIMER_GRANULARITY);
 
         // Packets sent before this time are deemed lost.
-        let lost_send_time = now - loss_delay;
+        let lost_send_time = now.checked_sub(loss_delay).unwrap();
```
djc, line 1340: Seems fine to make this explicit? Seems pretty obviously safe -- we wouldn't worry about `loss_delay` overflowing a contemporary `Instant`.

bench/src/bin/bulk.rs
```diff
@@ -79,7 +79,7 @@ async fn server(endpoint: quinn::Endpoint, opt: Opt) -> Result<()> {
                 };
                 trace!("stream established");
 
-                let _: tokio::task::JoinHandle<Result<()>> = tokio::spawn(async move {
+                tokio::spawn(async move {
```
djc, line 82: I couldn't think of any reason to bind this?

perf/src/bin/perf_server.rs
Ralith: I'm generally not a fan of this lint (https://github.com/rust-lang/rust-clippy/issues/10245) because it obfuscates whether you're borrowing or moving out of the match scrutinee. I'm not going to die on this hill, though.

quinn-proto/src/connection/mod.rs
Ralith: :+1:

bench/src/bin/bulk.rs
Ralith: I'd guess this was required at some point in the past to unify the return type. If it builds without it, great!

Ralith (approved):

PR #1501, opened 2023-02-27
Provider-based crypto tests 

nmittler:
This allows the crypto integration tests to be run against any crypto provider. For now, only rustls is supported. This will be updated in the future once the boringssl provider has landed.

Requires #1496.


nmittler:
I'm going to give up for now on re-using the quinn-proto tests for boringssl. Instead, I'll make some basic integration tests that use the public API.

djc:
@nmittler probably a good idea. Sorry for being slow with the reviews!

PR #1502, opened 2023-03-01
Borrow `self` in read_to_end, rather than consuming

Ralith:
Allows `RecvStream::stop` to be called with custom error codes if `ReadToEndError::TooLong` is encountered.

See also discussion in #1493.


djc (approved):
Should we bump the version number proactively for this semver-incompatible change?

Ralith:
No strong feelings; is that the process you'd like to standardize on?

djc (approved):

PR #1504, opened 2023-03-03
Implement fallback for `sendmmsg` and `recvmmsg`

link2xt:
Some Android phones use older 2.x kernels
and do not support `sendmmsg`
which is available only since Linux 3.0.

Similarly, `recvmmsg` is only available
since Linux 2.6.34.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -194,7 +194,9 @@ fn send(
     let num_transmits = transmits.len().min(BATCH_SIZE);
 
     loop {
-        let n = unsafe { libc::sendmmsg(io.as_raw_fd(), msgs.as_mut_ptr(), num_transmits as _, 0) };
+        let n = unsafe {
+            sendmmsg_with_fallback(io.as_raw_fd(), msgs.as_mut_ptr(), num_transmits as _)
+        };
         if n == -1 {
             let e = io::Error::last_os_error();
             match e.kind() {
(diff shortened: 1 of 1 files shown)


djc:
I changed the repo setting to only flag contributors from new GitHub members.

quinn-udp/src/unix.rs
```diff
@@ -143,7 +143,7 @@ fn init(io: SockRef<'_>) -> io::Result<()> {
     Ok(())
 }
 
-#[cfg(not(any(target_os = "macos", target_os = "ios")))]
+#[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "android")))]
```
djc, line 146: This seems like a pretty heavy hammer, which impairs performance for all Android builds on account of what's presumably a pretty small minority of very old phones.

Ralith: I'd prefer the dynamic fallback. It's a common enough Linux pattern, and lets us avoid baking in assumptions about use cases. I wouldn't be surprised if we ended up needing similar tricks on other platforms sooner or later either.

Even for typical mobile applications where CPU might not be a bottleneck, power use is still important, for example.

quinn-udp/src/unix.rs
```diff
@@ -66,6 +66,100 @@ impl Default for UdpSocketState {
...
+    let flags = 0;
+    let timeout = ptr::null_mut::<libc::timespec>();
+    let ret =
+        libc::syscall(libc::SYS_recvmmsg, sockfd, msgvec, vlen, flags, timeout) as libc::c_int;
+
+    if ret == -1 {
+        let e = io::Error::last_os_error();
+        match e.raw_os_error() {
+            Some(libc::ENOSYS) => {
+                // Fallback to `sendmsg`.
```
Ralith, line 153:
```suggestion
                // Fallback to `recvmsg`.
```

Ralith (changes requested):
Very tidy, thanks!

Ralith (dismissed):
Thanks!

quinn-udp/src/unix.rs
```diff
@@ -66,6 +66,100 @@ impl Default for UdpSocketState {
...
+#[cfg(not(any(target_os = "macos", target_os = "ios")))]
+unsafe fn sendmmsg_with_fallback(
+    sockfd: libc::c_int,
+    msgvec: *mut libc::mmsghdr,
+    vlen: libc::c_uint,
+) -> libc::c_int {
+    let flags = 0;
+    let ret = libc::syscall(libc::SYS_sendmmsg, sockfd, msgvec, vlen, flags) as libc::c_int;
+
+    if ret == -1 {
```
djc, line 101: Style nit: would prefer an early return (with inverted condition here).

quinn-udp/src/unix.rs
```diff
@@ -66,6 +66,100 @@ impl Default for UdpSocketState {
     }
 }
 
+/// Fallback implementation of `sendmmsg` using `sendmsg`
```
djc, line 69: Style nit: we usually have more of a top-down ordering of items, so ideally we would move the added `sendmmsg_with_fallback()` directly below `send()`, with `sendmmsg_fallback` below that, and similar for  the receive side.

quinn-udp/src/unix.rs
```diff
@@ -66,6 +66,100 @@ impl Default for UdpSocketState {
...
+}
+
+#[cfg(not(any(target_os = "macos", target_os = "ios")))]
+unsafe fn sendmmsg_with_fallback(
+    sockfd: libc::c_int,
+    msgvec: *mut libc::mmsghdr,
+    vlen: libc::c_uint,
+) -> libc::c_int {
+    let flags = 0;
+    let ret = libc::syscall(libc::SYS_sendmmsg, sockfd, msgvec, vlen, flags) as libc::c_int;
```
djc, line 99: Out of curiosity, is the difference between `libc::syscall(libc::SYS_recvmmsg, ..)` and `libc::recvmmsg(..)` material for catching an `ENOSYS` error?

quinn-udp/src/unix.rs
```diff
@@ -66,6 +66,100 @@ impl Default for UdpSocketState {
...
+    sockfd: libc::c_int,
+    msgvec: *mut libc::mmsghdr,
+    vlen: libc::c_uint,
+) -> libc::c_int {
+    let flags = 0;
+    let timeout = ptr::null_mut::<libc::timespec>();
+    let ret =
+        libc::syscall(libc::SYS_recvmmsg, sockfd, msgvec, vlen, flags, timeout) as libc::c_int;
+
+    if ret == -1 {
```
djc, line 149: Style: early return?

quinn-udp/src/unix.rs
```diff
@@ -66,6 +66,100 @@ impl Default for UdpSocketState {
...
+        let e = io::Error::last_os_error();
+        match e.raw_os_error() {
+            Some(libc::ENOSYS) => {
+                // Fallback to `recvmsg`.
+                recvmmsg_fallback(sockfd, msgvec, vlen)
```
djc, line 154: Don't know how much you care, but presumably we *could* store state on whether mmsg support is available somewhere, maybe in the `UdpState`, so that we don't have to fallback everytime? (I suppose that would come at the cost of one extra branch even for environments where mmsg support *is* available, but that should be a pretty cheap branch with prediction.)

quinn-udp/src/unix.rs
djc: Ah, makes sense. Maybe add a short comment to explain the rationale behind using `syscall()` here?

djc (approved):

PR #1505, opened 2023-03-08
Switch CI coverage test to use cargo-llvm-cov

djc:


Ralith (approved):
Thanks for getting this going again!

PR #1510, opened 2023-03-14
Initial support for PLPMTUD

### LINKED ISSUE #69: Path MTU discovery
We should implement [packetization-layer path MTU discovery](https://tools.ietf.org/html/rfc4821) per the draft's instructions.

aochagavia:
This PR resurrects @djc's old https://github.com/quinn-rs/quinn/pull/804. As of yet, it is just a draft. IMO it makes sense to look at the commits separately (we might want to squash them in the end, though).

Below are some bits of feedback from the original PR that I'm planning to include before considering this PR as ready for review (feel free to suggest more):

- [x] Implement black hole detection
- [x] Use an accessor for `MtuDiscovery::max_udp_payload_size`, to better convey that only `MtuDiscovery` itself is meant to modify it

~~Looking at the original PR, there is one question that comes up: I am tracking ACKs for MTU probes manually (e.g. without using `PackageBuilder::finalize_and_track`), because otherwise things like automatic retransmission kick in. That seems different from what @djc did in the original PR (he seems to get ACK and packet loss detection automatically, without additional code). Am I working in the right direction, or did I miss a function I could use to get ACKs + lost packets for free like in the original PR?~~ I messed up and finally figured out how to get things working :)

Closes #69

### DIFF
--- a/bench/src/lib.rs
+++ b/bench/src/lib.rs
@@ -142,6 +142,8 @@ pub fn transport_config(opt: &Opt) -> quinn::TransportConfig {
     // High stream windows are chosen because the amount of concurrent streams
     // is configurable as a parameter.
     let mut config = quinn::TransportConfig::default();
+    #[cfg(any(windows, os = "linux"))]
+    config.mtu_discovery_config(Some(quinn::MtuDiscoveryConfig::default()));
     config.max_concurrent_uni_streams(opt.max_streams.try_into().unwrap());
     config.initial_max_udp_payload_size(opt.initial_mtu);
     config
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -10,6 +10,7 @@ use crate::{
     congestion,
     crypto::{self, HandshakeTokenKey, HmacKey},
     VarInt, VarIntBoundsExceeded, DEFAULT_SUPPORTED_VERSIONS, INITIAL_MAX_UDP_PAYLOAD_SIZE,
+    MAX_UDP_PAYLOAD,
 };
 
 /// Parameters governing the core QUIC state machine
@@ -37,6 +38,7 @@ pub struct TransportConfig {
     pub(crate) time_threshold: f32,
     pub(crate) initial_rtt: Duration,
     pub(crate) initial_max_udp_payload_size: u16,
+    pub(crate) mtu_discovery_config: Option<MtuDiscoveryConfig>,
 
     pub(crate) persistent_congestion_threshold: u32,
     pub(crate) keep_alive_interval: Option<Duration>,
@@ -154,20 +156,47 @@ impl TransportConfig {
         self
     }
 
-    /// UDP payload size that the network must be capable of carrying
-    ///
-    /// Effective max UDP payload size may change over the life of the connection in the future due
-    /// to path MTU detection, but will never fall below this value.
+    /// The initial value to be used as the maximum UDP payload size before running MTU discovery
+    /// (see [`TransportConfig::mtu_discovery_config`]).
     ///
     /// Must be at least 1200, which is the default, and known to be safe for typical internet
     /// applications. Larger values are more efficient, but increase the risk of unpredictable
-    /// catastrophic packet loss due to exceeding the network path's IP MTU.
+    /// catastrophic packet loss due to exceeding the network path's IP MTU. If the provided value
+    /// is higher than what the network path actually supports, packet loss will eventually trigger
+    /// black hole detection and bring it down to 1200.
     ///
     /// Real-world MTUs can vary according to ISP, VPN, and properties of intermediate network links
-    /// outside of either endpoint's control. Extreme caution should be used when raising this value
-    /// for connections outside of private networks where these factors are fully controlled.
+    /// outside of either endpoint's control. Caution should be used when raising this value for
+    /// connections outside of private networks where these factors are fully controlled.
     pub fn initial_max_udp_payload_size(&mut self, value: u16) -> &mut Self {
         self.initial_max_udp_payload_size = value.max(INITIAL_MAX_UDP_PAYLOAD_SIZE);
+        self
+    }
+
+    /// Specifies the MTU discovery config (see [`MtuDiscoveryConfig`] for details).
+    ///
+    /// Defaults to `None`, which disables MTU discovery altogether.
+    ///
+    /// # Important
+    ///
+    /// MTU discovery depends on platform support for disabling UDP packet fragmentation, which is
+    /// not always available. If the platform allows fragmenting UDP packets, MTU discovery may end
+    /// up "discovering" an MTU that is not really supported by the network, causing packet loss
+    /// down the line.
+    ///
+    /// The `quinn` crate provides the `Endpoint::server` and `Endpoint::client` constructors that
+    /// automatically disable UDP packet fragmentation on Linux and Windows. When using these
+    /// constructors, MTU discovery will reliably work, unless the code is compiled targeting an
+    /// unsupported platform (e.g. iOS). In the latter case, it is advisable to keep MTU discovery
+    /// disabled.
+    ///
+    /// Users of `quinn-proto` and authors of custom `AsyncUdpSocket` implementations should ensure
+    /// to disable UDP packet fragmentation (this is strongly recommended by [RFC
+    /// 9000](https://www.rfc-editor.org/rfc/rfc9000.html#section-14-7), regardless of MTU
+    /// discovery). They can build on top of the `quinn-udp` crate, used by `quinn` itself, which
+    /// provides Linux and Windows support for disabling packet fragmentation.
+    pub fn mtu_discovery_config(&mut self, value: Option<MtuDiscoveryConfig>) -> &mut Self {
+        self.mtu_discovery_config = value;
         self
     }
 
@@ -267,6 +296,7 @@ impl Default for TransportConfig {
             time_threshold: 9.0 / 8.0,
             initial_rtt: Duration::from_millis(333), // per spec, intentionally distinct from EXPECTED_RTT
             initial_max_udp_payload_size: INITIAL_MAX_UDP_PAYLOAD_SIZE,
+            mtu_discovery_config: None,
 
             persistent_congestion_threshold: 3,
             keep_alive_interval: None,
@@ -313,6 +343,113 @@ impl fmt::Debug for TransportConfig {
             .field("datagram_send_buffer_size", &self.datagram_send_buffer_size)
             .field("congestion_controller_factory", &"[ opaque ]")
             .finish()
+    }
+}
+
+/// Parameters governing MTU discovery.
+///
+/// # The why of MTU discovery
+///
+/// By design, QUIC ensures during the handshake that the network path between the client and the
+/// server is able to transmit unfragmented UDP packets with a body of 1200 bytes. In other words,
+/// once the connection is established, we know that the network path's maximum transmission unit
+/// (MTU) is of at least 1200 bytes (plus IP and UDP headers). Because of this, a QUIC endpoint can
+/// split outgoing data in packets of 1200 bytes, with confidence that the network will be able to
+/// deliver them (if the endpoint were to send bigger packets, they could prove too big and end up
+/// being dropped).
+///
+/// There is, however, a significant overhead associated to sending a packet. If the same
+/// information can be sent in fewer packets, that results in higher throughput. The amount of
+/// packets that need to be sent is inversely proportional to the MTU: the higher the MTU, the
+/// bigger the packets that can be sent, and the fewer packets that are needed to transmit a given
+/// amount of bytes.
+///
+/// Most networks have an MTU higher than 1200. Through MTU discovery, endpoints can detect the
+/// path's MTU and, if it turns out to be higher, start sending bigger packets.
+///
+/// # MTU discovery internals
+///
+/// Quinn implements MTU discovery through DPLPMTUD (Datagram Packetization Layer Path MTU
+/// Discovery), described in [section 14.3 of RFC
+/// 9000](https://www.rfc-editor.org/rfc/rfc9000.html#section-14.3). This method consists of sending
+/// QUIC packets padded to a particular size (called PMTU probes), and waiting to see if the remote
+/// peer responds with an ACK. If an ACK is received, that means the probe arrived at the remote
+/// peer, which in turn means that the network path's MTU is of at least the packet's size. If the
+/// probe is lost, it is sent another 2 times before concluding that the MTU is lower than the
+/// packet's size.
+///
+/// MTU discovery runs on a schedule (e.g. every 600 seconds) specified through
+/// [`MtuDiscoveryConfig::interval`]. The first run happens right after the handshake, and
+/// subsequent discoveries are scheduled to run when the interval has elapsed, starting from the
+/// last time when MTU discovery completed.
+///
+/// Since the search space for MTUs is quite big (the smallest possible MTU is 1200, and the highest
+/// is 65527), Quinn performs a binary search to keep the number of probes as low as possible. The
+/// lower bound of the search is equal to [`TransportConfig::initial_max_udp_payload_size`] in the
+/// initial MTU discovery run, and is equal to the currently discovered MTU in subsequent runs. The
+/// upper bound is determined by the minimum of [`MtuDiscoveryConfig::upper_bound`] and the
+/// `max_udp_payload_size` transport parameter received from the peer during the handshake.
+///
+/// # Black hole detection
+///
+/// If, at some point, the network path no longer accepts packets of the detected size, packet loss
+/// will eventually trigger black hole detection and reset the detected MTU to 1200. In that case,
+/// MTU discovery will be triggered after [`MtuDiscoveryConfig::black_hole_cooldown`] (ignoring the
+/// timer that was set based on [`MtuDiscoveryConfig::interval`]).
+///
+/// # Interaction between peers
+///
+/// There is no guarantee that the MTU on the path between A and B is the same as the MTU of the
+/// path between B and A. Therefore, each peer in the connection needs to run MTU discovery
+/// independently in order to discover the path's MTU.
+#[derive(Clone, Debug)]
+pub struct MtuDiscoveryConfig {
+    pub(crate) interval: Duration,
+    pub(crate) upper_bound: u16,
+    pub(crate) black_hole_cooldown: Duration,
+}
+
+impl MtuDiscoveryConfig {
+    /// Specifies the time to wait after completing MTU discovery before starting a new MTU
+    /// discovery run.
+    ///
+    /// Defaults to 600 seconds, as recommended by [RFC
+    /// 8899](https://www.rfc-editor.org/rfc/rfc8899).
+    pub fn interval(&mut self, value: Duration) -> &mut Self {
+        self.interval = value;
+        self
+    }
+
+    /// Specifies the upper bound to the max UDP payload size that MTU discovery will search for.
+    ///
+    /// Defaults to 1452, to stay within Ethernet's MTU when using IPv4 and IPv6. The highest
+    /// allowed value is 65527, which corresponds to the maximum permitted UDP payload on IPv6.
+    ///
+    /// It is safe to use an arbitrarily high upper bound, regardless of the network path's MTU. The
+    /// only drawback is that MTU discovery might take more time to finish.
+    pub fn upper_bound(&mut self, value: u16) -> &mut Self {
+        self.upper_bound = value.min(MAX_UDP_PAYLOAD);
+        self
+    }
+
+    /// Specifies the amount of time that MTU discovery should wait after a black hole was detected
+    /// before running again. Defaults to one minute.
+    ///
+    /// Black hole detection can be spuriously triggered in case of congestion, so it makes sense to
+    /// try MTU discovery again after a short period of time.
+    pub fn black_hole_cooldown(&mut self, value: Duration) -> &mut Self {
+        self.black_hole_cooldown = value;
+        self
+    }
+}
+
+impl Default for MtuDiscoveryConfig {
+    fn default() -> Self {
+        MtuDiscoveryConfig {
+            interval: Duration::from_secs(600),
+            upper_bound: 1452,
+            black_hole_cooldown: Duration::from_secs(60),
+        }
     }
 }
 
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -44,6 +44,7 @@ mod datagrams;
 use datagrams::DatagramState;
 pub use datagrams::{Datagrams, SendDatagramError};
 
+mod mtud;
 mod pacing;
 
 mod packet_builder;
@@ -263,6 +264,8 @@ impl Connection {
                 config.initial_rtt,
                 config.congestion_controller_factory.build(now),
                 config.initial_max_udp_payload_size,
+                None,
+                config.mtu_discovery_config.clone(),
                 now,
                 path_validated,
             ),
@@ -428,8 +431,8 @@ impl Connection {
                     SpaceId::Data,
                     "PATH_CHALLENGE queued without 1-RTT keys"
                 );
-                let mut buf = Vec::with_capacity(self.path.max_udp_payload_size as usize);
-                let buf_capacity = self.path.max_udp_payload_size as usize;
+                let mut buf = Vec::with_capacity(self.path.current_mtu() as usize);
+                let buf_capacity = self.path.current_mtu() as usize;
 
                 let mut builder = PacketBuilder::new(
                     now,
@@ -542,7 +545,7 @@ impl Connection {
                 // We need to send 1 more datagram and extend the buffer for that.
 
                 // Is 1 more datagram allowed?
-                if buf_capacity >= self.path.max_udp_payload_size as usize * max_datagrams {
+                if buf_capacity >= self.path.current_mtu() as usize * max_datagrams {
                     // No more datagrams allowed
                     break;
                 }
@@ -554,7 +557,7 @@ impl Connection {
                 // budget left, we always allow a full MTU to be sent
                 // (see https://github.com/quinn-rs/quinn/issues/1082)
                 if self.path.anti_amplification_blocked(
-                    self.path.max_udp_payload_size as u64 * num_datagrams as u64 + 1,
+                    self.path.current_mtu() as u64 * num_datagrams as u64 + 1,
                 ) {
                     trace!("blocked by anti-amplification");
                     break;
@@ -569,9 +572,9 @@ impl Connection {
                     } else {
                         0
                     } as u64;
-                    debug_assert!(untracked_bytes <= self.path.max_udp_payload_size as u64);
+                    debug_assert!(untracked_bytes <= self.path.current_mtu() as u64);
 
-                    let bytes_to_send = u64::from(self.path.max_udp_payload_size) + untracked_bytes;
+                    let bytes_to_send = u64::from(self.path.current_mtu()) + untracked_bytes;
                     if self.in_flight.bytes + bytes_to_send >= self.path.congestion.window() {
                         space_idx += 1;
                         congestion_blocked = true;
@@ -585,7 +588,7 @@ impl Connection {
                     if let Some(delay) = self.path.pacing.delay(
                         smoothed_rtt,
                         bytes_to_send,
-                        self.path.max_udp_payload_size,
+                        self.path.current_mtu(),
                         self.path.congestion.window(),
                         now,
                     ) {
@@ -601,7 +604,7 @@ impl Connection {
                 if let Some(mut builder) = builder.take() {
                     // Pad the packet to make it suitable for sending with GSO
                     // which will always send the maximum PDU.
-                    builder.pad_to(self.path.max_udp_payload_size);
+                    builder.pad_to(self.path.current_mtu());
 
                     builder.finish_and_track(now, self, sent_frames.take(), &mut buf);
 
@@ -609,7 +612,7 @@ impl Connection {
                 }
 
                 // Allocate space for another datagram
-                buf_capacity += self.path.max_udp_payload_size as usize;
+                buf_capacity += self.path.current_mtu() as usize;
                 if buf.capacity() < buf_capacity {
                     // We reserve the maximum space for sending `max_datagrams` upfront
                     // to avoid any reallocations if more datagrams have to be appended later on.
@@ -619,9 +622,7 @@ impl Connection {
                     // (e.g. purely containing ACKs), modern memory allocators
                     // (e.g. mimalloc and jemalloc) will pool certain allocation sizes
                     // and therefore this is still rather efficient.
-                    buf.reserve(
-                        max_datagrams * self.path.max_udp_payload_size as usize - buf.capacity(),
-                    );
+                    buf.reserve(max_datagrams * self.path.current_mtu() as usize - buf.capacity());
                 }
                 num_datagrams += 1;
                 coalesce = true;
@@ -666,7 +667,7 @@ impl Connection {
                 space_id,
                 &mut buf,
                 buf_capacity,
-                (num_datagrams - 1) * (self.path.max_udp_payload_size as usize),
+                (num_datagrams - 1) * (self.path.current_mtu() as usize),
                 ack_eliciting,
                 self,
                 self.version,
@@ -742,8 +743,7 @@ impl Connection {
                 !(sent.is_ack_only(&self.streams)
                     && !can_send.acks
                     && can_send.other
-                    && (buf_capacity - builder.datagram_start)
-                        == self.path.max_udp_payload_size as usize),
+                    && (buf_capacity - builder.datagram_start) == self.path.current_mtu() as usize),
                 "SendableFrames was {can_send:?}, but only ACKs have been written"
             );
             pad_datagram |= sent.requires_padding;
@@ -773,6 +773,48 @@ impl Connection {
 
         self.app_limited = buf.is_empty() && !congestion_blocked;
 
+        // Send MTU probe if necessary
+        if buf.is_empty() && self.state.is_established() {
+            let space_id = SpaceId::Data;
+            let probe_size = match self
+                .path
+                .mtud
+                .poll_transmit(now, self.spaces[space_id].next_packet_number)
+            {
+                Some(next_probe_size) => next_probe_size,
+                None => return None,
+            };
+
+            let buf_capacity = probe_size as usize;
+            buf.reserve(buf_capacity);
+
+            let mut builder = PacketBuilder::new(
+                now,
+                space_id,
+                &mut buf,
+                buf_capacity,
+                0,
+                true,
+                self,
+                self.version,
+            )?;
+
+            // We implement MTU probes as ping packets padded up to the probe size
+            buf.write(frame::Type::PING);
+            builder.pad_to(probe_size);
+            let sent_frames = SentFrames {
+                non_retransmits: true,
+                ..Default::default()
+            };
+            builder.finish_and_track(now, self, Some(sent_frames), &mut buf);
+
+            self.stats.frame_tx.ping += 1;
+            self.stats.path.sent_plpmtud_probes += 1;
+            num_datagrams = 1;
+
+            trace!(?probe_size, "writing MTUD probe");
+        }
+
         if buf.is_empty() {
             return None;
         }
@@ -794,7 +836,7 @@ impl Connection {
             },
             segment_size: match num_datagrams {
                 1 => None,
-                _ => Some(self.path.max_udp_payload_size as usize),
+                _ => Some(self.path.current_mtu() as usize),
             },
             src_ip: self.local_ip,
         })
@@ -1175,6 +1217,9 @@ impl Connection {
                     self.spaces[space].pending_acks.subtract_below(acked);
                 }
                 ack_eliciting_acked |= info.ack_eliciting;
+
+                // Notify MTU discovery that a packet was acked, because it might be an MTU probe
+                self.path.mtud.on_acked(space, packet, info.size);
                 self.on_packet_acked(now, space, info);
             }
         }
@@ -1342,6 +1387,8 @@ impl Connection {
 
     fn detect_lost_packets(&mut self, now: Instant, pn_space: SpaceId, due_to_ack: bool) {
         let mut lost_packets = Vec::<u64>::new();
+        let mut lost_mtu_probe = None;
+        let in_flight_mtu_probe = self.path.mtud.in_flight_mtu_probe();
         let rtt = self.path.rtt.conservative();
         let loss_delay = cmp::max(rtt.mul_f32(self.config.time_threshold), TIMER_GRANULARITY);
 
@@ -1371,24 +1418,30 @@ impl Connection {
 
             if info.time_sent <= lost_send_time || largest_acked_packet >= packet + packet_threshold
             {
-                lost_packets.push(packet);
-                size_of_lost_packets += info.size as u64;
-                if info.ack_eliciting && due_to_ack {
-                    match persistent_congestion_start {
-                        // Two ACK-eliciting packets lost more than congestion_period apart, with no
-                        // ACKed packets in between
-                        Some(start) if info.time_sent - start > congestion_period => {
-                            in_persistent_congestion = true;
+                if Some(packet) == in_flight_mtu_probe {
+                    // Lost MTU probes are not included in `lost_packets`, because they should not
+                    // trigger a congestion control response
+                    lost_mtu_probe = in_flight_mtu_probe;
+                } else {
+                    lost_packets.push(packet);
+                    size_of_lost_packets += info.size as u64;
+                    if info.ack_eliciting && due_to_ack {
+                        match persistent_congestion_start {
+                            // Two ACK-eliciting packets lost more than congestion_period apart, with no
+                            // ACKed packets in between
+                            Some(start) if info.time_sent - start > congestion_period => {
+                                in_persistent_congestion = true;
+                            }
+                            // Persistent congestion must start after the first RTT sample
+                            None if self
+                                .path
+                                .first_packet_after_rtt_sample
+                                .map_or(false, |x| x < (pn_space, packet)) =>
+                            {
+                                persistent_congestion_start = Some(info.time_sent);
+                            }
+                            _ => {}
                         }
-                        // Persistent congestion must start after the first RTT sample
-                        None if self
-                            .path
-                            .first_packet_after_rtt_sample
-                            .map_or(false, |x| x < (pn_space, packet)) =>
-                        {
-                            persistent_congestion_start = Some(info.time_sent);
-                        }
-                        _ => {}
                     }
                 }
             } else {
@@ -1423,6 +1476,7 @@ impl Connection {
                     self.streams.retransmit(frame);
                 }
                 self.spaces[pn_space].pending |= info.retransmits;
+                self.path.mtud.on_non_probe_lost(now, *packet, info.size);
             }
             // Don't apply congestion penalty for lost ack-only packets
             let lost_ack_eliciting = old_bytes_in_flight != self.in_flight.bytes;
@@ -1436,6 +1490,17 @@ impl Connection {
                     size_of_lost_packets,
                 );
             }
+        }
+
+        // Handle a lost MTU probe
+        if let Some(packet) = lost_mtu_probe {
+            let info = self.spaces[SpaceId::Data]
+                .sent_packets
+                .remove(&packet)
+                .unwrap(); // safe: lost_mtu_probe is omitted from lost_packets, and therefore must not have been removed yet
+            self.remove_in_flight(SpaceId::Data, &info);
+            self.path.mtud.on_probe_lost();
+            self.stats.path.lost_plpmtud_probes += 1;
         }
     }
 
@@ -2171,7 +2236,7 @@ impl Connection {
                 }
 
                 if self.side.is_client() {
-                    // Client-only beceause server params were set from the client's Initial
+                    // Client-only because server params were set from the client's Initial
                     let params =
                         self.crypto
                             .transport_parameters()?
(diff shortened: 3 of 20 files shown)


quinn-proto/src/tests/util.rs
```diff
@@ -18,10 +18,14 @@ use tracing::{info_span, trace};
 
 use super::*;
 
+pub const DEFAULT_MAX_UDP_PAYLOAD_SIZE: usize = 1200;
+
 pub struct Pair {
     pub server: TestEndpoint,
     pub client: TestEndpoint,
     pub time: Instant,
+    /// Simulates the maximum size allowed for UDP payloads (packets exceeding this size will be dropped)
```
Ralith, line 27:
```suggestion
    /// Simulates the maximum size allowed for UDP payloads by the link (packets exceeding this size will be dropped)
```

Ralith:
> I am tracking ACKs for MTU probes manually (e.g. without using PackageBuilder::finalize_and_track), because otherwise things like automatic retransmission kick in.

Per [RFC 9000 § Retransmission of Information](https://www.rfc-editor.org/rfc/rfc9000.html#name-retransmission-of-informati), QUIC retransmission operates at frame granularity, not packet granularity. In particular note:

> PING and PADDING frames contain no information, so lost PING or PADDING frames do not require repair.

So, there should be no need to bypass the normal loss detection machinery. However, note that there is a subtle interaction with congestion control:

> Loss of a QUIC packet that is carried in a PMTU probe is therefore not a reliable indication of congestion and SHOULD NOT trigger a congestion control reaction; see Item 7 in [Section 3](https://www.rfc-editor.org/rfc/rfc8899#section-3) of [[DPLPMTUD](https://www.rfc-editor.org/rfc/rfc9000.html#DPLPMTUD)]. However, PMTU probes consume congestion window, which could delay subsequent transmission by an application.

aochagavia:
@Ralith thanks for the feedback! I think I finally figured out how to use `finalize_and_track` :). The current status of the PR is that it replicates the original changes by @djc, plus black hole detection and tests (check out, for instance, `blackhole_after_mtu_change_repairs_itself`). Any suggestions so far?

Next week I plan to hunt down the test failure (unfortunately, the test runner doesn't give info about which test actually failed, other than `thread panicked while panicking. aborting.`)

Ralith:
> unfortunately, the test runner doesn't give info about which test actually failed

Hmm, we should probably set RUST_BACKTRACE=1 in the test environment.

quinn-proto/src/connection/mod.rs
```diff
@@ -773,6 +775,46 @@ impl Connection {
...
+                .mtud
+                .poll_transmit(self.spaces[space_id].next_packet_number)
+            {
+                Some(next_probe_size) => next_probe_size,
+                None => return None,
+            };
+
+            if buf.capacity() < probe_size as usize {
+                buf.reserve(probe_size as usize - buf.capacity());
+            }
```
Ralith, line 792: `reserve` is relative to the current `len`, not the capacity; further, we already know `len` is zero, and `reserve` already noops if redundant, so we might as well call it unconditionally.
```suggestion
            buf.reserve(probe_size as usize);
```

quinn-proto/src/connection/mod.rs
```diff
@@ -773,6 +775,46 @@ impl Connection {
...
+                0,
+                true,
+                self,
+                self.version,
+            )?;
+
+            // We implement MTU probes as ping packets padded up to the probe size
+            buf.write(frame::Type::PING);
+            builder.pad_to(probe_size);
+            builder.finish_and_track(now, self, Some(SentFrames::default()), &mut buf);
```
Ralith, line 810: We should set `non_retransmits` in `SentFrames` here to reflect the presence of non-retransmittable ACK-eliciting frames, similar to how other PINGs are handled:

https://github.com/quinn-rs/quinn/blob/d31597f181ad1357761c0179792e09a7803ba46e/quinn-proto/src/connection/mod.rs#L2729-L2735

I don't think any existing logic cares either way, but it might be a landmine in the future.

quinn-proto/src/connection/mod.rs
```diff
@@ -1428,6 +1480,16 @@ impl Connection {
                 );
             }
         }
+
+        // Handle a lost MTU probe
+        if let Some(packet) = lost_mtu_probe {
+            let info = self.spaces[SpaceId::Data]
+                .sent_packets
+                .remove(&packet)
+                .unwrap(); // safe: lost_packets is populated just above
```
Ralith, line 1489: It's not immediately obvious what lost_packets being populated has to do with this unwrap. Should this say something like "lost_mtu_probe is necessarily omitted from lost_packets, and therefore must not have been removed yet"?

quinn-proto/src/connection/mod.rs
```diff
@@ -3078,6 +3140,11 @@ impl Connection {
             .saturating_sub(self.in_flight.bytes)
     }
 
+    #[cfg(test)]
+    pub(crate) fn congestion_events(&self) -> u64 {
+        self.stats.path.congestion_events
+    }
```
Ralith, line 3146: This is already available via `pub fn stats(&self)`.

quinn-proto/src/connection/mod.rs
```diff
@@ -3127,6 +3194,12 @@ impl Connection {
         self.rem_cids.active_seq()
     }
 
+    /// Returns the detected max_udp_payload_size
+    #[cfg(test)]
+    pub(crate) fn max_udp_payload_size(&self) -> u16 {
```
Ralith, line 3199: This is very easy to confuse with the transport parameter of the same name. Should we call it something like `path_mtu`, perhaps?

Ralith (changes requested):
Quick incomplete pass, will spend some more time soon.

aochagavia:
Thanks for the suggestions! Next week I'll have time to go through them ;)

By the way, we'll probably want to tweak a few things (please let me know if you think differently):

* The PLPMTUD algorithm should use binary search, instead of going linearly through a hardcoded list. It probably makes sense to start the search at 1500 and have a user-configurable upper bound that defaults to 1500 as well (that is what [s2n-quic](https://github.com/aws/s2n-quic) is [doing](https://github.com/aws/s2n-quic/blob/e31715845bc3d93ffbeb614284147a20defbb0c3/quic/s2n-quic-transport/src/path/mtu.rs)).
* The RFC [recommends](https://www.rfc-editor.org/rfc/rfc8899#section-5.1.1) running PLPMTUD every 600 seconds. I guess we will want to implement that as well.

Also interesting: how do you think this change should affect versioning? Once this is merged, endpoints will be doing a few extra requests than usual, which might be unexpected for the crate's users. Related to this, do you think it is worthwhile to provide a way to disable PLPMTUD?

Ralith:
> do you think it is worthwhile to provide a way to disable PLPMTUD

I think it's fine to silently enable it by default, but I can imagine use cases where it might be unnecessary (known networks) or harmful (endpoints which want to sleep for very long periods). A good approach might be to make the interval configurable through an `Option<Duration>` or similar.

quinn-proto/src/connection/mod.rs
aochagavia: Good catch, thanks!

quinn-proto/src/connection/mod.rs
aochagavia: Sure, I was a bit confused by the idea of tracking capacity separately from the buffer itself. I might explore extracting that logic into its own struct in a future PR, to remove that bit of complexity

Ralith:
> `None` for cases where MTUD should run only once

Why once rather than never?

aochagavia:
Just pushed an updated (and rebased) version of the PR, with updated docs, passing tests and all the features we discussed above.

There are still a few things I want to look into (tomorrow), before marking this PR as "ready for review" (instead of it being a draft):

1. I'm not totally satisfied with the current form that `MtuDiscovery` has taken. I'll see if I can refactor it a bit to make it more readable.
2. I want to double check that our approach to black hole detection is sound, particularly its consequences in the presence of congestion. For that I'll have another peek at s2n-quic and maybe also ask around in the QUIC WG Slack.
3. I want to double check that all the code that is using `mtud.current_mtu()` doesn't mind that the returned value of the function is allowed (and expected!) to change.

Also, there is [this comment](https://github.com/quinn-rs/quinn/pull/804#issuecomment-1172950107) in the old PR, mentioning that we should only enable PLPMTUD on platforms where we are setting the DF IP bit on. Is that still relevant @Ralith?

Finally, I added a commit to make `Pair::drive` more robust (otherwise it would consider a connection idle even though there were packets waiting to be sent, as I discovered after thoroughly debugging the `cid_retirement` test).

Ralith:
> technically, the local EndpointConfig::max_udp_payload_size is about which sizes we are willing to receive, and not about the sizes we are willing to send, so I'm not totally sure it makes sense to reuse it here (i.e. maybe we should have a separate way of configuring the upper bound for our binary search). Do you have an opinion on this?

Do we strictly need an upper bound? We could do open-ended "binary search" by growing the probe size exponentially until it starts failing. That'd be cool for Just Working on localhost, LANs with arbitrarily large jumbo frames, and so forth. Though, we probably should take care not to overly sacrifice rate of convergence in a typical environment for these niche cases. I'm not strongly opposed to a dedicated knob, but it's nice to have universal logic when possible. We might want to tune it carefully to test typical internet limits first regardless.

> As mentioned on Matrix, the docs of [initial_max_udp_payload_size](https://docs.rs/quinn/0.9.3/quinn/struct.TransportConfig.html#method.initial_max_udp_payload_size) mention that "effective max UDP payload size (...) will never fall below [the provided] value". I think we actually should fall below the provided value in the case of blackholes (that is what the current implementation does). Or am I missing something?

Capturing our discussion in chat: I'm happy to turn this into an initial guess rather than a hard floor.

> Also, there is https://github.com/quinn-rs/quinn/pull/804#issuecomment-1172950107 in the old PR, mentioning that we should only enable PLPMTUD on platforms where we are setting the DF IP bit on. Is that still relevant @Ralith?

Yes; I believe we currently only set that bit on Linux and Windows.

aochagavia:
> Do we strictly need an upper bound? We could do open-ended "binary search" by growing the probe size exponentially until it starts failing. (...). We might want to tune it carefully to test typical internet limits first regardless.

The upper bound is not strictly needed, but since the peer and the local endpoint have a configured `max_udp_payload_size` transport parameter, I think we can take that as a hint. To give you an idea of the performance difference in a typical internet path (MTU = 1500), here are the amount of probes sent before converging depending on the upper bound used:

* [lower_bound = 1200, upper_bound = 1500] converges after 3 probes
* [lower_bound = 1200, upper_bound = 10000] converges after 22 probes
* [lower_bound = 1200, upper_bound = u16::MAX] converges after 29 probes

Note that each failed attempt at a particular MTU results in 3 probes being sent. After the third lost probe, we conclude that the probed MTU was too big.

I took the time to experiment using exponential search to find the upper bound before doing binary search. Here are the results:

* [lower_bound = 1200, link_mtu = 1500] converges after 14 probes (7 probes to find the bounds [1456, 1712], 7 probes performing binary search)
* [lower_bound = 1200, link_mtu = 10000] converges after 30 probes (12 probes to find the bounds [9392, 17584], 18 probes performing binary search)
* [lower_bound = 1200, link_mtu = u16::MAX] converges after 24 probes (14 probes to find the bounds [33968, 65535], 10 probes performing binary search)

This leads me to think we should keep the current approach.

Ralith:
Nice analysis! I concur that prioritizing convergence for realistic cases should take precedence.

> the peer and the local endpoint have a configured max_udp_payload_size transport parameter, I think we can take that as a hint

Hmm, the RFC [says](https://www.rfc-editor.org/rfc/rfc9000.html#section-18.2-4.12.1):
>  It is expected that this is the space an endpoint dedicates to holding incoming packets.

and specifies the default value as 65527, so this may not be a good hint in practice, and our current default of 1480 might be inappropriate. For the sake of fast convergence I'd be happy to have a dedicated transport config field for MTUD upper bound, and default it to the appropriate value for a typical internet path.

aochagavia:
Thanks for the feedback. I ended up introducing a `PlpmtudConfig` struct that contains the necessary config for PLPMTUD.

Here's another update and some additional questions:

1. Regarding black hole detection, I was unable to ask about it on the QUIC WG Slack because it is invite-only (I sent an email to the mailing list to request an invitation, but haven't heard anything yet). In any case, we are following the same approach as s2n-quic, which seems sound to me (it was introduced in [this PR](https://github.com/aws/s2n-quic/pull/696), but there is no discussion about _why_ they took that approach). The gist of it is that we keep a "black hole counter" that is incremented whenever a non-mtu-probe packet is lost that is larger than 1200 (unless we have received an ACK for a MTU-sized packet with a larger packet number, which would indicate that the path is all right). If the counter reaches 4, we assume there is a black hole. After a black hole is detected, current MTU is lowered to 1200 and a new PLPMTUD run is scheduled 60 seconds later (unless PLPMTUD is disabled). The counter is reset whenever we receive an MTU-sized packet. I'm inclined to follow s2n-quic's approach here, unless someone can come up with a better idea (or can arrange an invite to the QUIC WG Slack).
2. Regarding the interaction between PLPMTUD and congestion control... The RFCs I am following seem to contradict each other. RFC 8899 [specifies](https://www.rfc-editor.org/rfc/rfc8899.html#name-features-required-to-provid) that "_An update to the PLPMTU (or MPS) MUST NOT increase the congestion window measured in bytes_", while RFC 9002 [says](https://www.rfc-editor.org/rfc/rfc9002#section-7.2-2) that "_If the maximum datagram size changes during the connection, the initial congestion window SHOULD be recalculated with the new size._". Any ideas on which one we should follow? Or am I somehow misunderstanding the RFCs? There is an [issue](https://github.com/aws/s2n-quic/issues/959#issuecomment-1480852033) in s2n-quic's repository related to this, and this morning I posted my findings, so they might have ideas on what to do about it. Otherwise it would be good to bring it up on Slack.
3. Regarding the IP DF bit, I had a look at `quinn-udp` and I can't fully conclude from the code which platforms have the DF bit set. I'll assume that @Ralith's comment above is correct and make sure PLPMTUD is only allowed in Linux and Windows (it surprises me, though, that we are unable to do it in other mature operating systems).
4. Finally, are there any other things I should look at in order to bring this PR into shape, so it can land?

djc:
> 1. Regarding black hole detection, I was unable to ask about it on the QUIC WG Slack because it is invite-only (I sent an email to the mailing list to request an invitation, but haven't heard anything yet).

When did you request an invite?

> 3. Regarding the IP DF bit, I had a look at `quinn-udp` and I can't fully conclude from the code which platforms have the DF bit set. I'll assume that @Ralith's comment above is correct and make sure PLPMTUD is only allowed in Linux and Windows (it surprises me, though, that we are unable to do it in other mature operating systems).

Looks like it's `IP_DONTFRAG`/`IPV6_DONTFRAG` for Windows and `IP_PMTUDISC_PROBE` for Linux. https://stackoverflow.com/questions/4415725/ip-dont-fragment-bit-on-mac-os has some suggestions for macOS (that I suppose we don't implement yet).

> 4. Finally, are there any other things I should look at in order to bring this PR into shape, so it can land?

If you think it's reasonably close, mark it as ready and I'll do a full review pass.

quinn-proto/src/config.rs
```diff
@@ -154,23 +157,32 @@ impl TransportConfig {
...
     pub fn initial_max_udp_payload_size(&mut self, value: u16) -> &mut Self {
         self.initial_max_udp_payload_size = value.max(INITIAL_MAX_UDP_PAYLOAD_SIZE);
         self
     }
 
+    /// Specifies the PLPMTUD config (see [`PlpmtudConfig`] for details). Defaults to
+    /// [`PlpmtudConfig::default`].
+    ///
+    /// Setting this value to `None` disables PLPMTUD altogether.
+    pub fn plpmtud_interval(&mut self, value: Option<PlpmtudConfig>) -> &mut Self {
```
Ralith, line 181:
```suggestion
    pub fn plpmtud_config(&mut self, value: Option<PlpmtudConfig>) -> &mut Self {
```

quinn-proto/src/connection/mod.rs
```diff
@@ -2962,9 +3026,6 @@ impl Connection {
             }).expect("preferred address CID is the first received, and hence is guaranteed to be legal");
         }
         self.peer_params = params;
-        self.path.max_udp_payload_size = self.path.max_udp_payload_size.min(
-            u16::try_from(self.peer_params.max_udp_payload_size.into_inner()).unwrap_or(u16::MAX),
-        );
```
Ralith, line 2967: Shouldn't MTU detection be aware of the hard limit set by the peer? This might be lower than our configured upper bound.

quinn-proto/src/connection/paths.rs
```diff
@@ -163,3 +170,453 @@ impl RttEstimator {
...
+    /// The UDP payload size we are currently probing
+    probing_udp_payload_size: u16,
+    /// Packet number of an in-flight probe, if any
+    probe_packet_number: Option<u64>,
+    /// Lost probes at the current probe size
+    lost_probes: usize,
+    /// A count of the number of packets with a size > BASE_UDP_PAYLOAD_SIZE lost since
+    /// the last time a packet with size equal to the current MTU was acknowledged
+    black_hole_counter: u8,
+    /// The largest acked packet of size `max_udp_payload_size`
```
Ralith, line 197: Outdated comment? There's no field by that name in this struct.

quinn-proto/src/connection/paths.rs
```diff
@@ -163,3 +170,453 @@ impl RttEstimator {
...
+    /// Detected MTU for the path
+    current_mtu: u16,
+    /// The lower bound for the current binary search
+    search_lower_bound: u16,
+    /// The upper bound for the current binary search
+    search_upper_bound: u16,
+    /// The UDP payload size we are currently probing
+    probing_udp_payload_size: u16,
+    /// Packet number of an in-flight probe, if any
+    probe_packet_number: Option<u64>,
```
Ralith, line 191: Do any of these fields have meaning if `probe_packet_number` is `None`? If not, should they all be inside a struct together, inside a single `Option`?

quinn-proto/src/connection/paths.rs
```diff
@@ -163,3 +170,453 @@ impl RttEstimator {
...
+            self.probe_packet_number = None;
+        }
+
+        if let Some(probe_udp_payload_size) = self.next_mtu_to_probe(last_probe_succeeded) {
+            self.probe_packet_number = Some(next_packet_number);
+            self.probing_udp_payload_size = probe_udp_payload_size;
+            Some(probe_udp_payload_size)
+        } else {
+            self.phase = Phase::Complete;
+            self.next_mtud_activation = Some(now + self.config.as_ref().unwrap().interval);
```
Ralith, line 292: Should we randomize this a little to avoid storms if many clients start at the same time (e.g. due to a server restart)?

quinn-proto/src/connection/paths.rs
```diff
@@ -163,3 +170,453 @@ impl RttEstimator {
...
+    pub fn lost_non_probe_packet(&mut self, now: Instant, packet_number: u64, packet_size: u16) {
+        // Keep track of lost packets higher than the base size, and reset the current MTU
+        // if they reach the threshold
+        if packet_size > BASE_UDP_PAYLOAD_SIZE
+            && self
+                .largest_acked_mtu_sized_packet
+                .map_or(true, |pn| packet_number > pn)
+        {
+            self.black_hole_counter += 1;
+            if self.black_hole_counter > BLACK_HOLE_THRESHOLD {
```
Ralith, line 382: Should this threshold be configurable?

quinn-proto/src/connection/paths.rs
```diff
@@ -67,20 +74,15 @@ impl PathData {
         PathData {
             remote,
             rtt: prev.rtt,
-            pacing: Pacer::new(
-                smoothed_rtt,
-                congestion.window(),
-                prev.max_udp_payload_size,
-                now,
-            ),
+            pacing: Pacer::new(smoothed_rtt, congestion.window(), prev.current_mtu(), now),
```
Ralith, line 77: Do we need to update the pacer when the MTU changes?

quinn-proto/src/connection/paths.rs
aochagavia: Yes, they do have a meaning (I have renamed `probe_packet_number` to `in_flight_probe`, to make it clearer). It is true, though, that many fields (including these) only have a meaning if we are in the `Searching` phase, so I have modified the `Searching` variant to include them, instead of cluttering the top-level `MtuDiscovery`

quinn-proto/src/connection/paths.rs
aochagavia: True, thanks!

quinn-proto/src/connection/paths.rs
aochagavia: It would be easy to add, but it might make the API less easy to change in a backwards-compatible way (e.g. what if we want to use a different black hole detection mechanism in the future?)

Putting my API designer's hat on, I'd wait until the current approach has proven itself _and_ users ask for more control

quinn-proto/src/connection/mod.rs
aochagavia: Yes, I got confused because I didn't know whether this was a hard limit... Just pushed the necessary changes to make MTUD aware of it

quinn-proto/src/connection/mod.rs
```diff
@@ -773,6 +771,47 @@ impl Connection {
...
+            // We implement MTU probes as ping packets padded up to the probe size
+            buf.write(frame::Type::PING);
+            builder.pad_to(probe_size);
+            let sent_frames = SentFrames {
+                non_retransmits: true,
+                ..Default::default()
+            };
+            builder.finish_and_track(now, self, Some(sent_frames), &mut buf);
+
+            self.stats.frame_tx.ping += 1;
```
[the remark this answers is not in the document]
aochagavia: I think that makes sense. Would `PathStats` be the right place? I can add a `sent_mtud_probes` field to it

quinn-proto/src/connection/paths.rs
aochagavia: The current behavior is to run MTUD right after the connection is established, and when that is done, to run it on an interval. This line of code is scheduling MTUD to run again after the user-configured interval has elapsed, so I assume it shouldn't interact with something like a server restart. Or am I missing something?

In case you were referring to randomizing the moment of the initial MTUD run, I'm not convinced that it would be beneficial. For one, I assume in most cases you want to run MTUD as early as possible. Also, even if MTUD is disabled / randomized, a server will probably still need additional mechanisms to protect itself from a storm of connections... So my hunch here is to leave it as is and let servers deal with the problem through other means (e.g. load balancing)

quinn-proto/src/connection/paths.rs
aochagavia: I think we do. The pacer's capacity is derived from the rtt and the congestion window, but it might end up being a different value because it is clamped between `10 * MTU` and `256 * MTU`. Before this PR, the capacity was only recalculated when the size of the congestion window changes, but it is clear that it should also be recalculated when the MTU changes (I just pushed a commit to that effect). Does that make sense?

aochagavia:
@Ralith thanks for the review. I just finished addressing all your comments, and left a few questions / remarks here and there.

While looking at the pacer and its interaction with the congestion controller, I discovered that each congestion controller has a `max_datagram_size` field, so I assume that will need to be updated based on the detected MTU. I'll look into that tomorrow.

aochagavia:
So today I've been looking into updating the congestion controller's `max_datagram_size` when we have discovered a new MTU. Unfortunately, the current API design seems to go against it. We currently offer a direct way to set `max_datagram_size`, independently of the path's MTU, and it would be pretty unexpected if we would now link both together. IMO if we want the congestion controllers to use the path's MTU as their `max_datagram_size`, we would need to rethink the API. Basically, we would need to remove the `max_datagram_size` function from the controller's config struct (which is a breaking change), and start using the current MTU instead. This might also mean that we need to automatically derive the values of other config fields, like `initial_window` and `minimum_window`, because they depend on the value of `max_datagram_size`. Overall it feels like a pretty invasive change, and since we are already allowing `max_datagram_size` and the current MTU to diverge, I'd rather postpone it. I asked on the QUIC implementers Slack this morning and haven't heard back from them yet.

I also looked into enabling PLPMTUD only on platforms that set the IP DF bit. The technical side is straightforward, but the API design side less so. Assuming we want to keep `quinn-proto` free of platform-specific code, we can either:

1. Have `TransportConfig::default` enable PLPMTUD by default. Direct users of `quinn-proto` would need to make sure they are properly setting the DF bit on their IO implementation, or otherwise disable PLPMTUD through `TransportConfig::plpmtud_config(None)`. On our side, we would make `quinn::Endpoint` ignore the PLPMTUD configuration on unsupported platforms. Unfortunately, this requires a backwards-incompatible change (as far as I can see), because `quinn::Endpoint` has read-only access to the `TransportConfig` (through `Arc<TransportConfig>`) and we would need to mutate it.
1. Have `TransportConfig::default` disable PLPMTUD by default. The function to enable it should document the fact that PLPMTUD is platform-specific and that you should enable it at your own risk. In a later PR we could investigate the possibility of making `quinn::Endpoint` raise an error if PLPMTUD is enabled on unsupported platforms (it would require additions to `quinn-proto`'s public API, so I'd rather postpone this part).
1. Introduce a `PlatformConfig` type in `quinn-proto` that indicates platform support for particular features, such as setting the DF bit on outgoing UDP packets. We could have a `PlatformConfig::default_transport_config(&self) -> TransportConfig` method that generates a suitable default `TransportConfig`, with PLPMTUD enabled if setting the DF bit is supported. The `PlatformConfig` instance would be created by users of the `quinn-proto` library.

All in all I think option 2 is the most viable, so the current implementation disables PLPMTUD by default.

With all this said, @djc I think the PR is ready for a thorough review, so I have marked it as ready.

quinn-proto/src/connection/paths.rs
Ralith: SGTM.

quinn-proto/src/connection/paths.rs
Ralith: Makes sense. Probably this should be mitigated by higher level application logic acting on connection initiation timing.

quinn-proto/src/connection/mod.rs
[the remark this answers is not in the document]
Ralith: I think so.

quinn-proto/src/config.rs
```diff
@@ -316,6 +338,80 @@ impl fmt::Debug for TransportConfig {
...
+    fn default() -> Self {
+        PlpmtudConfig {
+            interval: Duration::from_secs(600),
+            max_udp_payload_size_upper_bound: 1452,
+            black_hole_cooldown: Duration::from_secs(60),
+        }
+    }
+}
+
+impl fmt::Debug for PlpmtudConfig {
```
Ralith, line 402: I don't see any custom behavior in this impl. Could it be `derive`d?

quinn-proto/src/connection/paths.rs
Ralith: LGTM.

Ralith (approved):

Ralith:
> Basically, we would need to remove the max_datagram_size function from the controller's config struct (which is a breaking change), and start using the current MTU instead.

That sounds correct to me. This breakage is fine; we already have breakage pending/planned, and most users probably aren't using this interface. That said, happy to leave this for future work.

> All in all I think option 2 is the most viable, so the current implementation disables PLPMTUD by default.

Since `quinn-proto` is independent form a specific I/O backend, this makes sense to me. A similar issue exists for segmentation offload configuration; future iterations on this should probably address both. I like the idea of a `PlatformConfig` struct to make this easier for users to get right, but it's fine to leave that for future work too. In the mean time, we should at least update the main examples and performance test/benchmarking tools to enable MTU on linux/windows.

quinn-proto/src/connection/mod.rs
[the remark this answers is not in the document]
aochagavia: Perfect, I just made the change

aochagavia:
Just rebased on top of latest main and pushed a commit modifying the benchmarks, perf, and examples. I think there isn't anything left to be done on my side.

The next things I plan to pick up in separate PRs:

* Change the default `max_udp_payload_size` used when creating an `Endpoint` (from `1480` to `65527`), as suggested in [RFC 9000](https://www.rfc-editor.org/rfc/rfc9000.html#section-18.2-4.8.1) and by @Ralith in [this comment](https://github.com/quinn-rs/quinn/pull/1510#issuecomment-1480486452).
* Make sure the congestion control algorithm uses the current PMTU, instead of a hardcoded constant. Modify the API to configure congestion control accordingly (see [this comment](https://github.com/quinn-rs/quinn/pull/1510#issuecomment-1493078724) for additional context).
* Provide an easy way for users to create a platform-specific `TransportConfig`, maybe in the form of a `PlatformConfig` struct as described in [this comment](https://github.com/quinn-rs/quinn/pull/1510#issuecomment-1486880195). Start using it in our examples and benchmarks.

quinn-proto/src/config.rs
```diff
@@ -316,6 +338,67 @@ impl fmt::Debug for TransportConfig {
...
+pub struct PlpmtudConfig {
+    pub(crate) interval: Duration,
+    pub(crate) max_udp_payload_size_upper_bound: u16,
+    pub(crate) black_hole_cooldown: Duration,
+}
+
+impl PlpmtudConfig {
+    /// Specifies the time to wait after completing PLPMTUD before starting a new MTU discovery
+    /// round. Defaults to 600 seconds, as recommended by
+    /// [RFC 8899](https://www.rfc-editor.org/rfc/rfc8899).
```
djc, line 363: Style nit: let's add an empty line between these two sentences to help rustdoc.

quinn-proto/src/config.rs
```diff
@@ -316,6 +338,67 @@ impl fmt::Debug for TransportConfig {
...
+/// supported by the current network path. The lower bound of the search is equal to
+/// [`TransportConfig::initial_max_udp_payload_size`] in the initial PLPMTUD round, and is equal
+/// to the currently discovered MTU in subsequent runs. The upper bound is determined by
+/// [`PlpmtudConfig::max_udp_payload_size_upper_bound`].
+///
+/// If, at some point, the network path no longer accepts packets of the detected size, packet
+/// loss will eventually trigger black hole detection and reset the detected MTU to 1200. In
+/// that case, PLPMTUD will be triggered after [`PlpmtudConfig::black_hole_cooldown`] (ignoring the
+/// timer that was set based on [`PlpmtudConfig::interval`]).
+#[derive(Clone, Debug)]
```
djc, line 353: Looks like we can make this `Copy`?

quinn-proto/src/connection/mod.rs
```diff
@@ -773,6 +772,48 @@ impl Connection {
...
+                Some(next_probe_size) => next_probe_size,
+                None => return None,
+            };
+
+            let buf_capacity = probe_size as usize;
+            buf.reserve(buf_capacity);
+
+            let mut builder = PacketBuilder::new(
+                now,
+                SpaceId::Data,
```
djc, line 792: Nit: use `space_id` here?

quinn-proto/src/connection/mod.rs
```diff
@@ -773,6 +772,48 @@ impl Connection {
...
+        // Send MTU probe if necessary
+        if buf.is_empty() && self.state.is_established() {
+            let space_id = SpaceId::Data;
+            let probe_size = match self
+                .path
+                .mtud
+                .poll_transmit(now, self.spaces[space_id].next_packet_number)
+            {
+                Some(next_probe_size) => next_probe_size,
+                None => return None,
```
djc, line 784: Not sure this `return None` makes sense, that would skip all the statistics stuff below? Might make sense to extract a method for the code below to avoid this drifting too far to the right.

quinn-proto/src/connection/mod.rs
```diff
@@ -1175,6 +1216,7 @@ impl Connection {
                     self.spaces[space].pending_acks.subtract_below(acked);
                 }
                 ack_eliciting_acked |= info.ack_eliciting;
+                self.path.mtud.on_acked(space, packet, info.size);
```
djc, line 1219: Any considerations for moving this into `self.on_packet_acked()` instead of keeping it outside?

quinn-proto/src/connection/paths.rs
```diff
@@ -1,7 +1,9 @@
 use std::{cmp, net::SocketAddr, time::Duration, time::Instant};
+use tracing::trace;
 
 use super::pacing::Pacer;
-use crate::{congestion, packet::SpaceId, TIMER_GRANULARITY};
+use crate::config::PlpmtudConfig;
+use crate::{congestion, packet::SpaceId, MAX_UDP_PAYLOAD, TIMER_GRANULARITY};
```
djc, line 6: Nit: merge the `config::PlpmtudConfig` import into the last line of this block?

quinn-proto/src/connection/paths.rs
```diff
@@ -163,3 +177,605 @@ impl RttEstimator {
         }
     }
 }
+
```
djc, line 180: I think this stuff should go in a separate module, say, `mtud`?

quinn-proto/src/connection/paths.rs
```diff
@@ -163,3 +177,605 @@ impl RttEstimator {
...
+/// Implements Datagram Packetization Layer Path Maximum Transmission Unit Discovery
+///
+/// See <https://www.rfc-editor.org/rfc/rfc8899> for details
+#[derive(Clone)]
```
djc, line 184: This looks like it can be `Copy`.

quinn-proto/src/connection/paths.rs
```diff
@@ -163,3 +177,605 @@ impl RttEstimator {
...
+            Some(EnabledMtuDiscovery {
+                phase: Phase::Searching(search_state),
+                ..
+            }) => search_state.in_flight_probe,
+            _ => None,
+        }
+    }
+
+    /// Notifies the [`MtuDiscovery`] that the in-flight MTU probe was lost
+    pub fn on_probe_packet_lost(&mut self) {
```
djc, line 299: Naming: maybe lose the `packet` word in both `on_probe_packet_lost()` and `on_non_probe_packet_lost()`?

quinn-proto/src/connection/paths.rs
```diff
@@ -163,3 +177,605 @@ impl RttEstimator {
...
+
+    /// Notifies the [`MtuDiscovery`] that a non-probe packet was lost
+    pub fn on_non_probe_packet_lost(&mut self, now: Instant, packet_number: u64, packet_size: u16) {
+        // Keep track of lost packets higher than the base size, and reset the current MTU
+        // if they reach the threshold
+        if packet_size > BASE_UDP_PAYLOAD_SIZE
+            && self
+                .largest_acked_mtu_sized_packet
+                .map_or(true, |pn| packet_number > pn)
+        {
```
djc, line 313: Style nit: invert this condition so we can return early?

quinn-proto/src/connection/paths.rs
```diff
@@ -163,3 +177,605 @@ impl RttEstimator {
...
+    /// We are currently searching for a higher PMTU
+    Searching(SearchState),
+    /// Searching has completed and will be triggered again at the provided instant
+    Complete(Instant),
+}
+
+#[derive(Debug, Clone)]
+struct SearchState {
+    /// The lower bound for the current binary search
+    search_lower_bound: u16,
```
djc, line 450: Nit: maybe drop the `search_` prefixes here?

quinn-proto/src/connection/paths.rs
```diff
@@ -163,3 +177,605 @@ impl RttEstimator {
...
+    in_flight_probe: Option<u64>,
+    /// Lost probes at the current probe size
+    lost_probe_count: usize,
+}
+
+impl SearchState {
+    /// Creates a new search state, with the specified lower bound (the upper bound is derived from
+    /// the config and the peer's `max_udp_payload_size` transport parameter)
+    fn new(lower_bound: u16, peer_max_udp_payload_size: u16, config: &PlpmtudConfig) -> Self {
+        let search_upper_bound = config
```
djc, line 465: Consider using `clamp()` here?

quinn-proto/src/config.rs
Ralith: We could, but that would cause the future addition of non-`Copy` fields to be breaking. Our other config structs aren't `Copy`, so I think it makes sense to be conservative here.

quinn-proto/src/connection/mod.rs
Ralith: I think this is correct as written; we haven't sent anything, so there's no stats to update. Compare the `if buf.is_empty()` case that immediately follows the new code. Might still be opportunity for more clarity though.

bench/src/lib.rs
```diff
@@ -142,6 +142,8 @@ pub fn transport_config(opt: &Opt) -> quinn::TransportConfig {
     // High stream windows are chosen because the amount of concurrent streams
     // is configurable as a parameter.
     let mut config = quinn::TransportConfig::default();
+    #[cfg(any(windows, os = "linux"))]
```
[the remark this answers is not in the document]
Ralith: A method seems like overkill since the value won't ever change for a given build. Applications not using quinn-udp at all logically shouldn't care about quinn-udp's constants.

quinn-proto/src/connection/mod.rs
aochagavia: Moving the call to `mtud.on_acked` into `self.on_packet_acked` seems to make sense, but requires modifying the signature of the latter, so it has access to the packet number (which it currently lacks). I'll do it ;)

quinn-proto/src/connection/paths.rs
aochagavia: If we follow @Ralith's suggestion in https://github.com/quinn-rs/quinn/pull/1510#discussion_r1156223697, `MtuDiscoveryConfig` won't be `Copy` and this struct can't be `Copy` either

quinn-proto/src/connection/mod.rs
djc: That might not be an improvement. I was just going off of, are there other callers of `on_packet_acked()`, how do these things conceptually relate? But it looks like this is the only caller that potentially uses the `Data` space, so it's probably fine to leave as is.

quinn-proto/src/connection/mod.rs
djc: Ahh yes, I guess I forgot about the `buf.is_empty()` precondition here.

quinn-proto/src/config.rs
djc: Fair enough. Let's make all the private types that can be `Copy` `Copy`, though.

quinn-proto/src/connection/paths.rs
aochagavia: Unfortunately, `clamp` will panic if `min > max`, which is something we should handle robustly here (i.e. we want `config.upper_bound.clamp(lower_bound, peer_max_udp_payload_size)`, but it is possible that `lower_bound` is higher than `peer_max_udp_payload_size`).

djc: Fair enough, maybe add a comment about this?

aochagavia: While adding the comment I figured out a way to use clamp after all, just pushed it ;)

quinn-proto/src/config.rs
aochagavia: Sure, just pushed a change for that

quinn-proto/src/connection/mod.rs
aochagavia: Ok, just reverted the change

aochagavia:
@djc I just pushed an amended (and rebased) version of the commits that address all your suggestions

quinn-proto/src/config.rs
```diff
@@ -154,23 +156,50 @@ impl TransportConfig {
         self
     }
 
-    /// UDP payload size that the network must be capable of carrying
-    ///
-    /// Effective max UDP payload size may change over the life of the connection in the future due
-    /// to path MTU detection, but will never fall below this value.
+    /// The initial value to be used as the maximum UDP payload size before running PLPMTUD (see
```
djc, line 159: Let's use "MTU discovery" instead of "PLPMTUD" here?

quinn-proto/src/config.rs
```diff
@@ -316,6 +346,113 @@ impl fmt::Debug for TransportConfig {
...
+/// By design, QUIC ensures during the handshake that the network path between the client and the
+/// server is able to transmit unfragmented UDP packets with a body of 1200 bytes. In other words,
+/// once the connection is established, we know that the network path's maximum transmission unit
+/// (MTU) is of at least 1200 bytes (plus IP and UDP headers). Because of this, a QUIC endpoint can
+/// split outgoing data in packets of 1200 bytes, with confidence that the network will be able to
+/// deliver them (if the endpoint were to send bigger packets, they could prove too big and end up
+/// being dropped).
+///
+/// There is, however, a significant overhead associated to sending a packet. If the same
+/// information can be sent in less packets, that results in higher throughput. The amount of
```
djc, line 362: "less" -> "fewer".

quinn-proto/src/config.rs
```diff
@@ -316,6 +346,113 @@ impl fmt::Debug for TransportConfig {
...
+/// packets that need to be sent is inversely proportional to the MTU: the higher the MTU, the
+/// bigger the packets that can be sent, and the less packets that are needed to transmit a given
```
djc, line 364: "less" -> "fewer"

quinn-proto/src/config.rs
```diff
@@ -316,6 +346,113 @@ impl fmt::Debug for TransportConfig {
...
+/// 9000](https://www.rfc-editor.org/rfc/rfc9000.html#section-14.3). This method consists of sending
+/// QUIC packets padded to a particular size (called PMTU probes), and waiting to see if the remote
+/// peer responds with an ACK. If an ACK is received, that means the probe arrived at the remote
+/// peer, which in turn means that the network path's MTU is of at least the packet's size. If the
+/// probe is lost, it is sent another 2 times before concluding that the MTU is lower than the
+/// packet's size.
+///
+/// MTU discovery runs on a schedule (e.g. every 600 seconds) specified through
+/// [`MtuDiscoveryConfig::interval`]. The first run happens right after the handshake, and
+/// subsequent discoveries are scheduled to run when the interval is elapsed, starting from the last
```
djc, line 383: I think "is" -> "has" is more correct?

quinn-proto/src/config.rs
```diff
@@ -316,6 +346,113 @@ impl fmt::Debug for TransportConfig {
     }
 }
 
+/// Parameters governing MTU discovery.
```
djc, line 349: This comment is great!

quinn-proto/src/connection/mtud.rs
```diff
@@ -0,0 +1,652 @@
...
+#[derive(Debug, Clone)]
+struct EnabledMtuDiscovery {
+    phase: Phase,
+    peer_max_udp_payload_size: u16,
+    config: MtuDiscoveryConfig,
+}
+
+impl EnabledMtuDiscovery {
+    fn new(config: MtuDiscoveryConfig) -> Self {
+        EnabledMtuDiscovery {
```
djc, line 173: Style nit: use `Self` here.

quinn-proto/src/connection/mtud.rs
```diff
@@ -0,0 +1,652 @@
...
+            // Start the first search
+            self.phase = Phase::Searching(SearchState::new(
+                current_mtu,
+                self.peer_max_udp_payload_size,
+                &self.config,
+            ));
+        } else if let Phase::Complete(next_mtud_activation) = &self.phase {
+            if now < *next_mtud_activation {
+                return None;
+            } else {
```
djc, line 197: Style nit: maybe drop the `else` here?

djc (approved):
Looking great! A few minor things found in the second round.

quinn-proto/src/config.rs
```diff
@@ -316,6 +346,113 @@ impl fmt::Debug for TransportConfig {
...
+
+    /// Specifies the upper bound to the max UDP payload size that MTU discovery will search for.
+    ///
+    /// Defaults to 1452, to stay within Ethernet's MTU when using IPv4 and IPv6. The highest
+    /// allowed value is 65527, which corresponds to the maximum permitted UDP payload on IPv6.
+    ///
+    /// It is safe to use an arbitrarily high upper bound, regardless of the network path's MTU. The
+    /// only drawback is that MTU discovery might take more time to finish.
+    pub fn upper_bound(&mut self, value: u16) -> &mut Self {
+        self.upper_bound = value.min(MAX_UDP_PAYLOAD);
```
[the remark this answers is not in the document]
aochagavia: I don't really have a preference about this, but I'd be inclined to follow whatever is the convention in the codebase. A quick search reveals 2 usages of `Ord::min` and 1 of `Ord::max`, against 44 usages of  `.min` and 27 of `.max`. Shall we keep the `.min` then?

quinn-proto/src/connection/mtud.rs
```diff
@@ -0,0 +1,652 @@
+use crate::{packet::SpaceId, MtuDiscoveryConfig, MAX_UDP_PAYLOAD};
+use std::time::Instant;
+use tracing::trace;
+
+/// Implements Datagram Packetization Layer Path Maximum Transmission Unit Discovery
+///
+/// See [`MtuDiscoveryConfig`] for details
+#[derive(Clone)]
+pub struct MtuDiscovery {
```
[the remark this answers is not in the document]
aochagavia: True, I assume using `pub(crate)` on the structs suffices, right?

aochagavia:
@djc just pushed changes addressing your comments

quinn-proto/src/config.rs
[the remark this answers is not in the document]
djc: Fair enough.

quinn-proto/src/connection/mtud.rs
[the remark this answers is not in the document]
djc: See #1526 about the standard I'd be going for.

djc (approved):

quinn-proto/src/connection/mtud.rs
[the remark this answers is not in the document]
aochagavia: I see, I just started using `pub(crate)` instead of `pub` in `mtud.mod`

djc (approved):
Nice work!

Ralith (approved):
Great stuff, thank you! This feature's been long-awaited.

PR #1512, opened 2023-03-15
Do not add IP_TOS cmsg on Android

### LINKED ISSUE #1511: sendmmsg fails with EINVAL on Android
While trying to get quinn-based file transfer (https://github.com/deltachat/deltachat-core-rust/pull/4007) working on Android, we discovered that no UDP packets are sent.

I compiled a CLI application `deltachat-repl` with NDK and ran it in Termux under `strace` to find out what is happening. Apparently all `sendmmsg` calls failed with EINVAL error.

Example of the first `sendmmsg` call:
```
2228  sendmmsg(18,  <unfinished ...>                                  
2229  futex(0x78c180a568, FUTEX_WAIT_PRIVATE, 1, NULL <unfinished ...>
2225  epoll_pwait(3,  <unfinished ...>           
2224  futex(0x78bfdf6068, FUTEX_WAIT_PRIVATE, 1, NULL <unfinished ...>
2228  <... sendmmsg resumed>[{msg_hdr={msg_name={sa_family=AF_INET, sin_port=htons(38795), sin_addr=inet_addr("192.168.178.181")}, msg_namelen=16, msg_iov=[{iov_base="\311\0\0\0\1\24\326\340\2350\360\376\2206O\230Q\317\326\274P\256\321yG\333\10\6\350\237\266\36"..., iov_len=1200}], msg_iovlen=1, msg_control=[{cmsg_len=20, cmsg_level=SOL_IP, cmsg_type=IP_TOS, cmsg_data=[0x2, 0, 0, 0]}], msg_controllen=24, msg_flags=0}}], 1, 0) = -1 EINVAL (Invalid argument)
```

Normally, on Linux, `sendmmsg` call looks like this:
```
413989 sendmmsg(17,  <unfinished ...>
413985 <... epoll_wait resumed>[{events=EPOLLOUT, data={u32=0, u64=0}}], 1024, 987) = 1  
413985 epoll_wait(3,  <unfinished ...>                                                                                                                        
413989 <... sendmmsg resumed>[{msg_hdr={msg_name={sa_family=AF_INET, sin_port=htons(33499), sin_addr=inet_addr("192.168.178.181")}, msg_namelen=16, msg_iov=[{iov_base="\306\0\0\0\1\24\371~\254V<\27\27\203\247\201sY\362\325\233\357\323\377\304\2\10\27\\\327\375\n"..., iov_len=1200}], msg_iovlen=1, msg_control=[{cmsg_len=20, cmsg_level=SOL_IP, cmsg_type=IP_TOS, cmsg_data=[0x2, 0, 0, 0]}], msg_controllen=24, msg_flags=0}, msg_len=1200}], 1, 0) = 1
```

The difference I see is `msg_len=1200` when running on Linux, but not on Android, but this is a return value, it is written by the syscall on success.

Sending UDP packets with `ncat` from Termux works, so it is not a problem with a general network permission.

link2xt:
Android usually comes with old Linux kernels,
and Linux <3.13 does not support IP_TOS cmsg type, returning EINVAL error from sendmmsg without sending anything.

To avoid this problem we do not try to set TOS on Android at all.

Fixes #1511

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -17,7 +17,7 @@ use super::{cmsg, log_sendmsg_error, RecvMeta, UdpSockRef, UdpState, IO_ERROR_LO
 
 #[cfg(target_os = "freebsd")]
 type IpTosTy = libc::c_uchar;
-#[cfg(not(target_os = "freebsd"))]
+#[cfg(not(any(target_os = "freebsd", target_os = "android")))]
 type IpTosTy = libc::c_int;
 
 /// Tokio-compatible UDP socket with some useful specializations.
@@ -493,6 +493,11 @@ fn prepare_msg(
     let mut encoder = unsafe { cmsg::Encoder::new(hdr) };
     let ecn = transmit.ecn.map_or(0, |x| x as libc::c_int);
     if transmit.destination.is_ipv4() {
+        // Android may use old kernel <3.13 which does not support IP_TOS cmsg type.
+        // On these kernels `sendmmsg` returns EINVAL and does not send anything
+        // if it encounters IP_TOS cmsg.
+        // To avoid this, we do not try to set TOS on Android at all.
+        #[cfg(not(target_os = "android"))]
         encoder.push(libc::IPPROTO_IP, libc::IP_TOS, ecn as IpTosTy);
     } else {
         encoder.push(libc::IPPROTO_IPV6, libc::IPV6_TCLASS, ecn);


djc (dismissed):

djc:
LGTM, thanks for following up!

dignifiedquire:
Maybe EINVAL should be not ignored to avoid swallowing potential issues ?

djc:
I'm honestly not very satisified with the approach of disabling ECN for all Android users when mostly very old devices are actually impacted, so I would like to at least discuss alternative strategies of dealing with this. I'd also like review from Ralith on this stuff before merging.

Given that we're both volunteers, pinging within 24h after submitting a PR seems a little pushy IMO.

(In general, I'd be fine with creating a 0.9.4 release. 0.10 will have to wait until after rustls 0.21 has been released.)

djc:
> > Given that we're both volunteers, pinging within 24h after submitting a PR seems a little pushy IMO.
> 
> Ok, I just saw that you approved and also left a reaction on the rebase comment, so if you decide not to merge then something is probably missing that I should fix with the PR.

Fair enough, missed that context -- too much GitHub stuff to keep all the context around. Sorry for changing my mind, I'd rather optimize for newer/future devices while still trying to support the old stuff.

> Regarding alternative approaches, we can try to set ECN by default but switch it off as soon as we get a single `io::ErrorKind::InvalidInput`, that could also work.

Yeah, that might be nicer. I guess we won't know for sure if the `InvalidInput` is referring to the `TOS` value, which seems a little tricky?

(Also relating to releases, note that this is actually part of quinn-udp, for which we can just push out further 0.3.x releases from the main branch.)

PR #1506, opened 2023-03-10
Apply suggestions from clippy 1.68

djc:

PR #1513, opened 2023-03-15
Update clippy lint name

aochagavia:
I saw that the CI was failing for #1510 because of an unrelated clippy issue. This is the fix 😉


aochagavia:
Oops, just saw #1506 ...

PR #1516, opened 2023-03-16
Fallback to not using ECN if IP_TOS is not supported

link2xt:
On Linux <3.13 sendmsg/sendmmsg system calls
return EINVAL without sending anything if they encounter an `IP_TOS` cmsg. To ensure we can still send
on such systems, remember if we got an EINVAL error and switch to "fallback mode" in which we do not try to transmit ECN at all and only use well-supported features.


djc:
Seems reasonable to me!

Ralith (approved):

Ralith:
Thanks!

PR #1518, opened 2023-03-18
Getters for stream concurrency metrics and related documentation

### LINKED ISSUE #1517: Proto: Provide explicit MAX_STREAMS control.
Quic gives explicit control over how many streams of each kind the remote can open via MAX_STREAMS frames. Afaik quinn-proto only exposes this as limit on the number of concurrent open streams, which is more permissive than the application may want to be. This allows for more precise and flexible back pressure control.

The config methods would probably look like this:
```Rust
Connection::set_max_bi_stream_id(&mut self, StreamId);
Connection::set_max_uni_stream_id(&mut self, StreamId);
```

The condition could either be ORed or ANDed with the concurrency set by `set_max_concurrent_*_streams`.

In practice OR would be used like this:
```Rust
conn.set_max_concurrent_uni_streams(0u32.into());
conn.set_max_uni_stream_id(StreamId::new(side, Dir::Uni, 30u32.into()));
```

In practice AND would be used like this:
```Rust
conn.set_max_uni_stream_id(StreamId::new(side, Dir::Uni, 30u32.into()));
conn.set_max_concurrent_uni_streams(VarInt::MAX);
```

I have no preference between AND or OR semantics, both give the application sufficient control to implement any strategy it wants.

I'm happy to send a PR.

FlorianUekermann:
Would you prefer VarInts here?

Fixes #1517

### DIFF
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -56,6 +56,7 @@ impl<'a> Streams<'a> {
     /// Accept a remotely initiated stream of a certain directionality, if possible
     ///
     /// Returns `None` if there are no new incoming streams for this connection.
+    /// Has no impact on the data flow-control or stream concurrency limits.
     pub fn accept(&mut self, dir: Dir) -> Option<StreamId> {
         if self.state.next_remote[dir as usize] == self.state.next_reported_remote[dir as usize] {
             return None;
@@ -78,6 +79,18 @@ impl<'a> Streams<'a> {
     /// The number of streams that may have unacknowledged data.
     pub fn send_streams(&self) -> usize {
         self.state.send_streams
+    }
+
+    /// The number of remotely initiated open streams of a certain directionality.
+    ///
+    /// Includes remotely initiated streams, which have not been accepted via [`accept`](Self::accept).
+    /// These streams count against the respective concurrency limit reported by
+    /// [`Connection::max_concurrent_streams`](super::Connection::max_concurrent_streams).
+    pub fn remote_open_streams(&self, dir: Dir) -> u64 {
+        // total opened - total closed = total opened - ( total permitted - total permitted unclosed )
+        self.state.next_remote[dir as usize]
+            - (self.state.max_remote[dir as usize]
+                - self.state.allocated_remote_count[dir as usize])
     }
 }
 
--- a/quinn-proto/src/connection/streams/state.rs
+++ b/quinn-proto/src/connection/streams/state.rs
@@ -31,9 +31,9 @@ pub struct StreamsState {
     pub(super) max: [u64; 2],
     /// Maximum number of remotely-initiated streams that may be opened over the lifetime of the
     /// connection so far, per direction
-    max_remote: [u64; 2],
+    pub(super) max_remote: [u64; 2],
     /// Number of streams that we've given the peer permission to open and which aren't fully closed
-    allocated_remote_count: [u64; 2],
+    pub(super) allocated_remote_count: [u64; 2],
     /// Size of the desired stream flow control window. May be smaller than `allocated_remote_count`
     /// due to `set_max_concurrent` calls.
     max_concurrent_remote_count: [u64; 2],
@@ -761,6 +761,10 @@ impl StreamsState {
         self.flow_control_adjusted = true;
         self.max_concurrent_remote_count[dir as usize] = count.into();
         self.ensure_remote_streams(dir);
+    }
+
+    pub fn max_concurrent(&self, dir: Dir) -> u64 {
+        self.allocated_remote_count[dir as usize]
     }
 
     /// Set the receive_window and returns wether the receive_window has been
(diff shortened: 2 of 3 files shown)


djc (dismissed):

djc:
Nice, thanks!

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -79,6 +80,17 @@ impl<'a> Streams<'a> {
     pub fn send_streams(&self) -> usize {
         self.state.send_streams
     }
+
+    /// The number of remotely initiated open streams of a certain directionality.
+    ///
+    /// Includes remotely initiated streams, which have not been accepted via [`accept`](Self::accept).
+    /// These streams count against the respective concurrency limit reported by
+    /// [`Connection::max_concurrent_streams`](super::Connection::max_concurrent_streams).
+    pub fn remote_open_streams(&self, dir: Dir) -> u64 {
```
Ralith, line 89: What's the use case for this? It seems difficult to use, since it might become out of date immediately.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -79,6 +80,17 @@ impl<'a> Streams<'a> {
...
+        self.state.next_remote[dir as usize]
+            - (self.state.max_remote[dir as usize]
+                - self.state.allocated_remote_count[dir as usize])
```
Ralith, line 92: It took me a few minutes to understand this math (number of streams ever opened minus number of streams ever closed), so it could probably use a comment.

Ralith (changes requested):
> Would you prefer VarInts here?

I lean towards no. I think the main benefit of `VarInt` is documenting range limitations on inputs; for outputs it's difficult to imagine any case where it wouldn't be unwrapped immediately, so we might as well skip the ceremony.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -31,9 +31,9 @@ pub struct StreamsState {
     pub(super) max: [u64; 2],
     /// Maximum number of remotely-initiated streams that may be opened over the lifetime of the
     /// connection so far, per direction
-    max_remote: [u64; 2],
+    pub(super) max_remote: [u64; 2],
     /// Number of streams that we've given the peer permission to open and which aren't fully closed
-    allocated_remote_count: [u64; 2],
+    pub(super) allocated_remote_count: [u64; 2],
```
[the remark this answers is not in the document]
FlorianUekermann: I stuck with the pattern that methods `Streams` tend to directly access `StreamsState` fields, like:
```
    pub fn send_streams(&self) -> usize {
        self.state.send_streams
    }
```
But I can move the method and wrap it in `Streams` for public access if you want.

quinn-proto/src/connection/streams/mod.rs
FlorianUekermann: The motivation is quite similar to the other one. It is hard to keep track of what is actually considered open and counts against the limit without knowing a bunch of implementation details. They complement each other nicely if you want to know the full picture of the current concurrency situation.

`max_concurrent_streams()` tells what the concurrency limit is right now.
`remote_open_streams()` tells you how much of it is in use.
The difference tells you how many free slots are left or if the remote hit the limit.

I personally don't need this one right now, so we can leave it for later. But knowing if the remote hit the limit is useful to make resource allocation decisions (especially if downstream resources are costly and need to be allocated/requested before advertising their availability upstream by increasing the concurrency limit).

quinn-proto/src/connection/streams/mod.rs
FlorianUekermann: Will do, if we keep it at all.

quinn-proto/src/connection/streams/state.rs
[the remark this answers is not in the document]
Ralith: Ah, right, this is inside the `streams` module. So long as we're not exposing fields all the way up to `connection` I'm happy regardless.

quinn-proto/src/connection/streams/mod.rs
Ralith: I usually prefer to limit API expansion to stuff that's immediately applicable, lest we end up committed to supporting a bunch of stuff that nobody ends up wanting. On the other hand, this does seem like a straightforward statistic, and it's hard to imagine it ever becoming a burden to expose. In the end I think I'm happy to have it, especially if @djc's not raising any eyebrows.

quinn-proto/src/connection/streams/mod.rs
FlorianUekermann: Changed to this:
```Rust
// total opened - total closed = total opened - ( total permitted - total permitted unclosed )
        self.state.next_remote[dir as usize]
            - (self.state.max_remote[dir as usize]
                - self.state.allocated_remote_count[dir as usize])
```

Ralith (approved):

djc (approved):

djc:
Thanks!

PR #1522, opened 2023-04-01
Don't rely on sleeping in 0-RTT test

Ralith:
This is unnecessarily racy, at least in principle.

### DIFF
--- a/quinn/src/tests.rs
+++ b/quinn/src/tests.rs
@@ -254,22 +254,30 @@ async fn zero_rtt() {
     let _guard = subscribe();
     let endpoint = endpoint();
 
-    const MSG: &[u8] = b"goodbye!";
+    const MSG0: &[u8] = b"zero";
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Nice! @nmittler FYI.

Ralith:
Tangentially: a common cause of tests getting stuck until a timeout and then passing (which came up in previous related discussion) is relying on the implicit close-on-drop behavior and failing to drop something before calling `wait_idle`. It can be avoided robustly by closing connections explicitly.

nmittler:
@Ralith we should probably also add a test for 0-RTT being rejected here. That's where I'm running into a timing issue in the [quinn-boring PR](https://github.com/quinn-rs/quinn-boring/pull/2). There is a lower-level test for rejection under quinn-proto, but it doesn't suffer from the same timing problem.

PR #1526, opened 2023-04-04
Warn on unreachable_pub

djc:
I have found this makes it easier to see whether changes affect the public API, which can be hard especially in reviews when context is limited.


Ralith (approved):
I'm not a *huge* fan of the extra verbosity, but the rationale makes sense to me.

djc:
> I'm not a _huge_ fan of the extra verbosity, but the rationale makes sense to me.

Yeah, especially to cover all of the fields when a struct has limited visibility is a bit annoying, but I generally really like that you can trust the explicit visibility without having to conduct a reachability analysis in your head.

Ralith (approved):

PR #1527, opened 2023-04-04
Warn on clippy's use_self lint

djc:
This is generally more concise and in case of a rename avoids having to change a bunch of extra locations.

(Note that clippy warnings fail our CI's lint job.)

### DIFF
--- a/quinn-proto/src/cid_generator.rs
+++ b/quinn-proto/src/cid_generator.rs
@@ -46,7 +46,7 @@ impl RandomConnectionIdGenerator {
         debug_assert!(cid_len <= MAX_CID_SIZE);
         Self {
             cid_len,
-            ..RandomConnectionIdGenerator::default()
+            ..Self::default()
         }
     }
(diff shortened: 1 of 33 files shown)


aochagavia:
In case anyone cares... I personally dislike using `Self`, because it means I have to mentally look up which type it refers to and therefore makes reading the code a bit more difficult (maybe it is just a lack of familiarity, though). I'd rather have a tiny bit less clean diffs when renames happen.

aochagavia:
Wow, didn't know this was debated to that length... At least I feel comforted that I'm not alone 😅

Ralith (approved):

Ralith:
> In case anyone cares...

Your input is absolutely valued!

> I have to mentally look up which type it refers to

To me, this seems less demanding than any use of `self` (the expression). The language already aggressively buys into that context, so getting a bit of extra mileage out of it seems like a win to me.

PR #1529, opened 2023-04-05
Tie up PLPMTUD's loose ends

aochagavia:
This PR introduces a few self-contained improvements to make PLPMTUD work better.

We will probably want to discuss the API impact of the change to the congestion controller code:

* New `on_mtu_update` function on the `Controller` trait
* Removed `max_datagram_size` from all controller configs (the current MTU is used now instead)
* Removed `minimum_window` from all controller configs (it is now calculated on demand based on the current MTU)

With all this in place, running the top-level `bench` crate with a MTUD upper bound of 9000 results in sending rates of ~900 MiB/s (on my machine ™).

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -1,4 +1,4 @@
-use std::{convert::TryInto, fmt, num::TryFromIntError, sync::Arc, time::Duration};
+use std::{fmt, num::TryFromIntError, sync::Arc, time::Duration};
 
 use thiserror::Error;
 
@@ -476,7 +476,7 @@ impl EndpointConfig {
             || Box::<RandomConnectionIdGenerator>::default();
         Self {
             reset_key,
-            max_udp_payload_size: 1480u32.into(), // Typical internet MTU minus IPv4 and UDP overhead, rounded up to a multiple of 8
+            max_udp_payload_size: MAX_UDP_PAYLOAD.into(), // See RFC 9000 (https://www.rfc-editor.org/rfc/rfc9000.html#section-18.2-4.10.1)
             connection_id_generator_factory: Arc::new(cid_factory),
             supported_versions: DEFAULT_SUPPORTED_VERSIONS.to_vec(),
             grease_quic_bit: true,
@@ -508,12 +508,17 @@ impl EndpointConfig {
         self
     }
 
-    /// Maximum UDP payload size accepted from peers. Excludes UDP and IP overhead.
+    /// Maximum UDP payload size accepted from peers (excluding UDP and IP overhead).
     ///
-    /// The default is suitable for typical internet applications. Applications which expect to run
-    /// on networks supporting Ethernet jumbo frames or similar should set this appropriately.
-    pub fn max_udp_payload_size(&mut self, value: u64) -> Result<&mut Self, ConfigError> {
-        self.max_udp_payload_size = value.try_into()?;
+    /// Must be greater or equal than 1200.
+    ///
+    /// Defaults to 65527, which is the maximum permitted UDP payload.
+    pub fn max_udp_payload_size(&mut self, value: u16) -> Result<&mut Self, ConfigError> {
+        if !(1200..=65_527).contains(&value) {
+            return Err(ConfigError::OutOfBounds);
+        }
+
+        self.max_udp_payload_size = value.into();
         Ok(self)
     }
 
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -38,6 +38,7 @@ pub struct TransportConfig {
     pub(crate) time_threshold: f32,
     pub(crate) initial_rtt: Duration,
     pub(crate) initial_max_udp_payload_size: u16,
+    pub(crate) max_guaranteed_udp_payload_size: u16,
     pub(crate) mtu_discovery_config: Option<MtuDiscoveryConfig>,
 
     pub(crate) persistent_congestion_threshold: u32,
@@ -160,16 +161,35 @@ impl TransportConfig {
     /// (see [`TransportConfig::mtu_discovery_config`]).
     ///
     /// Must be at least 1200, which is the default, and known to be safe for typical internet
-    /// applications. Larger values are more efficient, but increase the risk of unpredictable
-    /// catastrophic packet loss due to exceeding the network path's IP MTU. If the provided value
-    /// is higher than what the network path actually supports, packet loss will eventually trigger
-    /// black hole detection and bring it down to 1200.
-    ///
-    /// Real-world MTUs can vary according to ISP, VPN, and properties of intermediate network links
-    /// outside of either endpoint's control. Caution should be used when raising this value for
-    /// connections outside of private networks where these factors are fully controlled.
+    /// applications. Larger values are more efficient, but increase the risk of packet loss due to
+    /// exceeding the network path's IP MTU. If the provided value is higher than what the network
+    /// path actually supports, packet loss will eventually trigger black hole detection and bring
+    /// it down to [`TransportConfig::max_guaranteed_udp_payload_size`].
     pub fn initial_max_udp_payload_size(&mut self, value: u16) -> &mut Self {
         self.initial_max_udp_payload_size = value.max(INITIAL_MAX_UDP_PAYLOAD_SIZE);
+        self
+    }
+
+    pub(crate) fn get_initial_max_udp_payload_size(&self) -> u16 {
+        self.initial_max_udp_payload_size
+            .max(self.max_guaranteed_udp_payload_size)
+    }
+
+    /// The maximum UDP payload size guaranteed to be supported by the network.
+    ///
+    /// Must be at least 1200, which is the default, and lower than or equal to
+    /// [`TransportConfig::initial_max_udp_payload_size`].
+    ///
+    /// Real-world MTUs can vary according to ISP, VPN, and properties of intermediate network links
+    /// outside of either endpoint's control. Extreme care should be used when raising this value
+    /// outside of private networks where these factors are fully controlled. If the provided value
+    /// is higher than what the network path actually supports, the result will be unpredictable and
+    /// catastrophic packet loss, without a possibility of repair. Prefer
+    /// [`TransportConfig::initial_max_udp_payload_size`] together with
+    /// [`TransportConfig::mtu_discovery_config`] to set a maximum UDP payload size that robustly
+    /// adapts to the network.
+    pub fn max_guaranteed_udp_payload_size(&mut self, value: u16) -> &mut Self {
+        self.max_guaranteed_udp_payload_size = value.max(INITIAL_MAX_UDP_PAYLOAD_SIZE);
         self
     }
 
@@ -296,6 +316,7 @@ impl Default for TransportConfig {
             time_threshold: 9.0 / 8.0,
             initial_rtt: Duration::from_millis(333), // per spec, intentionally distinct from EXPECTED_RTT
             initial_max_udp_payload_size: INITIAL_MAX_UDP_PAYLOAD_SIZE,
+            max_guaranteed_udp_payload_size: INITIAL_MAX_UDP_PAYLOAD_SIZE,
             mtu_discovery_config: None,
 
             persistent_congestion_threshold: 3,
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -9,8 +9,7 @@ use crate::{
     cid_generator::{ConnectionIdGenerator, RandomConnectionIdGenerator},
     congestion,
     crypto::{self, HandshakeTokenKey, HmacKey},
-    VarInt, VarIntBoundsExceeded, DEFAULT_SUPPORTED_VERSIONS, INITIAL_MAX_UDP_PAYLOAD_SIZE,
-    MAX_UDP_PAYLOAD,
+    VarInt, VarIntBoundsExceeded, DEFAULT_SUPPORTED_VERSIONS, INITIAL_MTU, MAX_UDP_PAYLOAD,
 };
 
 /// Parameters governing the core QUIC state machine
@@ -37,8 +36,8 @@ pub struct TransportConfig {
     pub(crate) packet_threshold: u32,
     pub(crate) time_threshold: f32,
     pub(crate) initial_rtt: Duration,
-    pub(crate) initial_max_udp_payload_size: u16,
-    pub(crate) max_guaranteed_udp_payload_size: u16,
+    pub(crate) initial_mtu: u16,
+    pub(crate) min_guaranteed_mtu: u16,
     pub(crate) mtu_discovery_config: Option<MtuDiscoveryConfig>,
 
     pub(crate) persistent_congestion_threshold: u32,
@@ -164,32 +163,31 @@ impl TransportConfig {
     /// applications. Larger values are more efficient, but increase the risk of packet loss due to
     /// exceeding the network path's IP MTU. If the provided value is higher than what the network
     /// path actually supports, packet loss will eventually trigger black hole detection and bring
-    /// it down to [`TransportConfig::max_guaranteed_udp_payload_size`].
-    pub fn initial_max_udp_payload_size(&mut self, value: u16) -> &mut Self {
-        self.initial_max_udp_payload_size = value.max(INITIAL_MAX_UDP_PAYLOAD_SIZE);
+    /// it down to [`TransportConfig::min_guaranteed_mtu`].
+    pub fn initial_mtu(&mut self, value: u16) -> &mut Self {
+        self.initial_mtu = value.max(INITIAL_MTU);
         self
     }
 
-    pub(crate) fn get_initial_max_udp_payload_size(&self) -> u16 {
-        self.initial_max_udp_payload_size
-            .max(self.max_guaranteed_udp_payload_size)
+    pub(crate) fn get_initial_mtu(&self) -> u16 {
+        self.initial_mtu.max(self.min_guaranteed_mtu)
     }
 
     /// The maximum UDP payload size guaranteed to be supported by the network.
     ///
     /// Must be at least 1200, which is the default, and lower than or equal to
-    /// [`TransportConfig::initial_max_udp_payload_size`].
+    /// [`TransportConfig::initial_mtu`].
     ///
     /// Real-world MTUs can vary according to ISP, VPN, and properties of intermediate network links
     /// outside of either endpoint's control. Extreme care should be used when raising this value
     /// outside of private networks where these factors are fully controlled. If the provided value
     /// is higher than what the network path actually supports, the result will be unpredictable and
     /// catastrophic packet loss, without a possibility of repair. Prefer
-    /// [`TransportConfig::initial_max_udp_payload_size`] together with
+    /// [`TransportConfig::initial_mtu`] together with
     /// [`TransportConfig::mtu_discovery_config`] to set a maximum UDP payload size that robustly
     /// adapts to the network.
-    pub fn max_guaranteed_udp_payload_size(&mut self, value: u16) -> &mut Self {
-        self.max_guaranteed_udp_payload_size = value.max(INITIAL_MAX_UDP_PAYLOAD_SIZE);
+    pub fn min_guaranteed_mtu(&mut self, value: u16) -> &mut Self {
+        self.min_guaranteed_mtu = value.max(INITIAL_MTU);
         self
     }
 
@@ -315,8 +313,8 @@ impl Default for TransportConfig {
             packet_threshold: 3,
             time_threshold: 9.0 / 8.0,
             initial_rtt: Duration::from_millis(333), // per spec, intentionally distinct from EXPECTED_RTT
-            initial_max_udp_payload_size: INITIAL_MAX_UDP_PAYLOAD_SIZE,
-            max_guaranteed_udp_payload_size: INITIAL_MAX_UDP_PAYLOAD_SIZE,
+            initial_mtu: INITIAL_MTU,
+            min_guaranteed_mtu: INITIAL_MTU,
             mtu_discovery_config: None,
 
             persistent_congestion_threshold: 3,
@@ -406,7 +404,7 @@ impl fmt::Debug for TransportConfig {
 ///
 /// Since the search space for MTUs is quite big (the smallest possible MTU is 1200, and the highest
 /// is 65527), Quinn performs a binary search to keep the number of probes as low as possible. The
-/// lower bound of the search is equal to [`TransportConfig::initial_max_udp_payload_size`] in the
+/// lower bound of the search is equal to [`TransportConfig::initial_mtu`] in the
 /// initial MTU discovery run, and is equal to the currently discovered MTU in subsequent runs. The
 /// upper bound is determined by the minimum of [`MtuDiscoveryConfig::upper_bound`] and the
 /// `max_udp_payload_size` transport parameter received from the peer during the handshake.
--- a/quinn-proto/src/congestion.rs
+++ b/quinn-proto/src/congestion.rs
@@ -58,6 +58,9 @@ pub trait Controller: Send {
         lost_bytes: u64,
     );
 
+    /// The known MTU for the current network path has been updated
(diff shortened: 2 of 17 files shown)


quinn-proto/src/connection/mtud.rs
```diff
@@ -77,10 +77,15 @@ impl MtuDiscovery {
     }
 
     /// Notifies the [`MtuDiscovery`] that a packet has been ACKed
```
Ralith, line 79: Let's document the return value here.

quinn-proto/src/congestion/bbr/mod.rs
```diff
@@ -468,6 +469,13 @@ impl Controller for Bbr {
         self.loss_state.lost_bytes += lost_bytes;
     }
 
+    fn on_mtu_update(&mut self, current_mtu: u16) {
+        self.current_mtu = current_mtu as u64;
+        self.min_cwnd = calculate_min_window(self.current_mtu);
+        self.init_cwnd = self.init_cwnd.max(self.min_cwnd);
```
Ralith, line 475: This means `init_cwnd` can only ever increase, not shrink, even if the MTU decreases. Is that intended? We don't seem to be updating initial window at all in the other algorithms.

quinn-proto/src/connection/mtud.rs
```diff
@@ -350,6 +335,80 @@ impl SearchState {
...
+            self.loss_burst_counter = 0;
+            self.largest_acked_mtu_sized_packet = Some(packet_number);
+        }
+    }
+
+    /// Notifies the [`BlackHoleDetector`] of a packet loss burst
+    pub(crate) fn on_packet_loss_burst(
+        &mut self,
+        leading_packet_number: u64,
+        leading_packet_size: u16,
```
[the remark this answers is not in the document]
aochagavia: I assume the reason for looking only at the leading packet size is that it keeps the code in s2n-quic simple and provides enough robustness. However, in our case, I think there is no additional complexity in making it more consistent, by ensuring only MTU-sized packets were lost instead of only looking at the leading one. I'll implement it 😉

quinn-proto/src/congestion/bbr/mod.rs
aochagavia: Hmmm... There seems to be a difference in the way the initial window is used in bbr compared to cubic and new reno. The latter two only use it upon creating the controller, so there is no need to update it, but bbr is using it in more places, so it makes sense to update it.

You are right that the mentioned code will only increase `init_cwnd` and never shrink it, which is a mistake. We actually want to clamp `config.initial_window` between `min_cwnd` and `u64::max` (I just pushed a commit doing so)

aochagavia:
I just pushed updated commits with the changes you suggested 😉

aochagavia:
@djc @Ralith Here's a reminder just in case this slipped through the cracks. I'll need these changes for my work on ACK frequency.

Ralith:
Still high on my list :+1:

aochagavia:
The last commit adds a `base_udp_payload_size` to `TransportConfig`, meant to configure the payload size that is guaranteed to be supported by the network. Adding it to `MtuDiscoveryConfig`, as @Ralith suggested, doesn't work here, because this is a property of the black hole detector (i.e. we want to use it even if MTU discovery is is disabled).

By the way, I felt a bit limited by the fact that we aren't using the builder pattern for `TransportConfig`, with a `build` function or some such where to do validation. To keep things user-friendly, I decided to automatically raise the `initial_max_udp_payload_size` when the `base_udp_payload_size` is raised (instead of crashing, which would force the user to set the values in a particular order).

quinn-proto/src/config.rs
```diff
@@ -508,12 +508,17 @@ impl EndpointConfig {
...
     ///
-    /// The default is suitable for typical internet applications. Applications which expect to run
-    /// on networks supporting Ethernet jumbo frames or similar should set this appropriately.
-    pub fn max_udp_payload_size(&mut self, value: u64) -> Result<&mut Self, ConfigError> {
-        self.max_udp_payload_size = value.try_into()?;
+    /// Must be greater or equal than 1200.
+    ///
+    /// Defaults to 65527, which is the maximum permitted UDP payload.
+    pub fn max_udp_payload_size(&mut self, value: u16) -> Result<&mut Self, ConfigError> {
+        if value < 1200 {
```
djc, line 537: Let's also yield `OutOfBounds` for `> 65_527`?

quinn-proto/src/connection/mtud.rs
```diff
@@ -50,8 +47,7 @@ impl MtuDiscovery {
         Self {
             current_mtu,
             state,
-            black_hole_counter: 0,
-            largest_acked_mtu_sized_packet: None,
+            black_hole_detector: BlackHoleDetector::new(),
```
djc: Might not make sense now, but for the future: this would be easier to review as two commits, one that stuffs the current logic into the a `BlackHoleDetector` type, and one that replaces the `BlackHoleDetector` implementation.

quinn-proto/src/connection/mtud.rs
```diff
@@ -131,31 +128,31 @@ impl MtuDiscovery {
...
-        self.black_hole_counter += 1;
-        if self.black_hole_counter > BLACK_HOLE_THRESHOLD {
-            self.black_hole_counter = 0;
-            self.largest_acked_mtu_sized_packet = None;
+    /// Returns true if a black hole was detected
+    ///
+    /// Calling this function will close the previous loss burst. If a black hole is detected, the
+    /// current MTU will be reset to `BASE_UDP_PAYLOAD_SIZE`.
+    pub(crate) fn detect_black_hole(&mut self, now: Instant) -> bool {
+        if self.black_hole_detector.detect_black_hole() {
```
djc, line 156: Style nit: consider inverting the check here so you can early return for the other case. I'd probably aslo change the method name (as well as the one one `BlackHoleDetector`) to something like `black_hole_detected()`, since it doesn't seem like this method is actually running much logic to "detect" a black hole.

quinn-proto/src/congestion.rs
```diff
@@ -12,6 +12,8 @@ pub use bbr::{Bbr, BbrConfig};
 pub use cubic::{Cubic, CubicConfig};
 pub use new_reno::{NewReno, NewRenoConfig};
 
+const BASE_DATAGRAM_SIZE: u64 = 1200;
```
djc, line 15: Style nit: we usually put our constants near the bottom of the module, not near the top.

quinn-proto/src/connection/mtud.rs
```diff
@@ -87,7 +87,7 @@ impl MtuDiscovery {
     ) -> bool {
         // MTU probes are only sent in application data space
         if space != SpaceId::Data {
-            return;
+            return false;
```
djc, line 99: Looks like this change actually belongs in a different commit?

quinn-proto/src/connection/mtud.rs
aochagavia: Thanks, I'll keep it in mind

quinn-proto/src/connection/mtud.rs
```diff
@@ -409,27 +575,42 @@ mod tests {
...
 
     #[test]
-    fn mtu_discovery_disabled_lost_four_packets_triggers_black_hole_detection() {
+    fn mtu_discovery_disabled_lost_four_packet_bursts_triggers_black_hole_detection() {
         let mut mtud = MtuDiscovery::disabled(1_400);
         let now = Instant::now();
 
-        for _ in 0..4 {
-            mtud.on_non_probe_lost(now, 0, 1300);
+        for i in 0..4 {
```
Ralith, line 593: Not a new issue, but should this for loop and similar below be defined in terms of `BLACK_HOLE_THRESHOLD` rather than hardcoded?

quinn-proto/src/connection/mtud.rs
```diff
@@ -81,10 +77,17 @@ impl MtuDiscovery {
     }
 
     /// Notifies the [`MtuDiscovery`] that a packet has been ACKed
-    pub(crate) fn on_acked(&mut self, space: SpaceId, packet_number: u64, packet_bytes: u16) {
+    ///
+    /// Returns true if the packet was an MTU probe
```
Ralith, line 90: nit: should the change to return a value here be part of the next commit? I don't see anywhere black hole detection uses it. No big deal though.

quinn-proto/src/config.rs
```diff
@@ -160,16 +161,34 @@ impl TransportConfig {
...
+    ///
+    /// Real-world MTUs can vary according to ISP, VPN, and properties of intermediate network links
+    /// outside of either endpoint's control. Extreme care should be used when raising this value
+    /// outside of private networks where these factors are fully controlled. If the provided value
+    /// is higher than what the network path actually supports, the result will be unpredictable and
+    /// catastrophic packet loss, without a possibility of repair. Prefer
+    /// [`TransportConfig::initial_max_udp_payload_size`] together with
+    /// [`TransportConfig::mtu_discovery_config`] to set a maximum UDP payload size that robustly
+    /// adapts to the network.
+    pub fn base_udp_payload_size(&mut self, value: u16) -> &mut Self {
```
Ralith, line 187: I think "base" is unclear here. Maybe `max_guaranteed`? Or `max_safe`?

quinn-proto/src/config.rs
```diff
@@ -160,16 +161,34 @@ impl TransportConfig {
...
+        self.base_udp_payload_size = value.max(BASE_UDP_PAYLOAD_SIZE);
+        self.initial_max_udp_payload_size = self
+            .initial_max_udp_payload_size
+            .max(self.base_udp_payload_size);
```
Ralith, line 191: I'm not super happy with how this makes the builder ordering-sensitive. It's not actually harmful if the invariant here is violated, is it? Or, maybe we should clamp the effective value when consuming the config?

quinn-proto/src/connection/mtud.rs
```diff
@@ -143,18 +152,18 @@ impl MtuDiscovery {
     ///
     /// Calling this function will close the previous loss burst. If a black hole is detected, the
     /// current MTU will be reset to `BASE_UDP_PAYLOAD_SIZE`.
-    pub(crate) fn detect_black_hole(&mut self, now: Instant) -> bool {
-        if self.black_hole_detector.detect_black_hole() {
-            self.current_mtu = BASE_UDP_PAYLOAD_SIZE;
+    pub(crate) fn black_hole_detected(&mut self, now: Instant) -> bool {
+        if !self.black_hole_detector.black_hole_detected() {
+            return false;
+        }
```
Ralith: Looks like this refactoring (incl. rename) landed in the wrong commit.

Ralith (changes requested):
Thanks, overall this is looking great!

quinn-proto/src/connection/mtud.rs
aochagavia: If it's not big deal, I'd rather concentrate on the rest of the feedback. I'll keep it in mind for next PRs, though!

quinn-proto/src/config.rs
aochagavia: Sure, let's go with `max_guaranteed`

quinn-proto/src/config.rs
aochagavia: I agree that making the builder ordering-sensitive is undesirable. In this particular case, raising the `initial_max_udp_payload_size` is partly a matter of convenience (if a higher payload size is guaranteed, we might as well use it from the beginning), partly of consistency (if a black whole is detected using an MTU of `initial_max_udp_payload_size`, the MTU will be reset to `max_guaranteed_udp_payload_size`, which might actualy be higher).

Short-term, clamping the value when consuming the config would get rid of the ordering issue. The only drawback I see is that we are opening the door for future confusion, when working on Quinn internals, because now we have to remember that `initial_max_udp_payload_size` must be clamped at the start by `guaranteed_max_udp_payload_size`, every time it is used.

Long-term, I would prefer a less conservative approach, replacing the current config struct by a builder which validates the config and generates a ready-to-consume config struct (either implicitly correcting user errors, as we do now, or rejecting them). The resulting struct can then be used without worrying about correcting user mistakes.

Since this PR is blocking the 0.10 release, let's go with the short-term solution for now. However, I'd be interested in hearing your thoughts on the idea of a builder.

quinn-proto/src/connection/mtud.rs
aochagavia: I actually like the fact that the testing code is not tied to the constant. That way tests will fail if someone were to accidentally increase the threshold. Also, there is a chance we will make the threshold configurable in the future (if it turns out to be necessary to tweak), in which case there will be no constant to use.

quinn-proto/src/config.rs
Ralith: > partly a matter of convenience [...], partly of consistency

Wouldn't be the first case of weird behavior being possible if you supply a weird config, at least. To some extent that's a fact of rich configuration.

> now we have to remember that initial_max_udp_payload_size must be clamped at the start by guaranteed_max_udp_payload_size, every time it is used.

I think this could be mitigated by using a getter method, and removing `pub(crate)` from the fields to enforce its use, if we want to not just let weird configs be weird.

> I'd be interested in hearing your thoughts on the idea of a builder.

I'm not categorically opposed, but I'm not sure this alone is compelling motivation for one. I prefer detecting errors per-setter as it reduces the distance between an error occurring and being reported, and so far we've gotten away without many complex invariants spanning multiple fields.

On the other hand, maybe a builder provide opportunity clean up some of the other awkwardness, e.g. with difficulty constructing platform-specific defaults. I know djc leaned into builders really hard in rustls not long ago to allow their config structures to be built up more gracefully. Config API usability improvements are definitely an area worth exploring in the future.

quinn-proto/src/config.rs
Ralith: Hmm, that reads like "guaranteed to be the maximum value supported" rather than "maximum value guaranteed to be supported" to me. Do you disagree?

quinn-proto/src/connection/mtud.rs
aochagavia: I will move it to the original commit introducing robust black hole detection

quinn-proto/src/config.rs
aochagavia: Thanks! I ended up keeping the `pub(crate)` for the field, because our current testing code assumes all fields are accessible, but I added a `pub(crate)` getter for code reuse.

@djc any thoughts on the idea of having a proper builder?

quinn-proto/src/config.rs
Ralith: On second thought I see how this reading makes sense; it's still probably difficult to interpret in isolation, but the comment resolves that and most people shouldn't be touching this anyway.

aochagavia: Hmmm... I can see the difference, yes. Just pushed a change to use `max_guaranteed` after all

quinn-proto/src/config.rs
djc: It feels like it would be a PITA to have a builder for all of `TransportConfig` if we only need it for these two MTU discovery-related values. Can we isolate the MTU things into a separate type (should we move them into `MtuDiscoveryConfig`, maybe while renaming it) that we can use a builder for?

> I'm not categorically opposed, but I'm not sure this alone is compelling motivation for one. I prefer detecting errors per-setter as it reduces the distance between an error occurring and being reported, and so far we've gotten away without many complex invariants spanning multiple fields.

Definitely agree that this can be downside of builder patterns. The most important reason for rustls to use a builder pattern was to force the caller to set some values, which IIRC is not important for our use cases (in fact, we'd probably prefer for most users not to touch the `guaranteed_max_udp_payload_size`).

> On the other hand, maybe a builder provide opportunity clean up some of the other awkwardness, e.g. with difficulty constructing platform-specific defaults.

Can you explain what kind of awkwardness we have going on here?

Ralith: > should we move them into MtuDiscoveryConfig

I think this is a poor fit because `MtuDiscoveryConfig` is omitted entirely when MTUD is disabled, but the initial and minimum MTU are still significant in that case as black hole detection runs regardless.

> Can you explain what kind of awkwardness we have going on here?

MTUD is the worst of it, since we want it on by default iff the underlying socket supports disabling fragmentation, but `TransportConfig::default` doesn't know what backend it will be used with. Maybe we should just default it on globally since as of #1547 almost every modern platform should be covered?

Segmentation offload would be similar, except there's probably no need to let users tune it, so it can remain specified via sidechannels. Maybe MTUD should require a config *and* a sidechannel DONTFRAG-compliance flag...

Ralith: I've now implemented that last strategy in #1547; I'm pretty happy with it.

quinn-proto/src/config.rs
```diff
@@ -160,19 +161,38 @@ impl TransportConfig {
...
+    /// Real-world MTUs can vary according to ISP, VPN, and properties of intermediate network links
+    /// outside of either endpoint's control. Extreme care should be used when raising this value
+    /// outside of private networks where these factors are fully controlled. If the provided value
+    /// is higher than what the network path actually supports, the result will be unpredictable and
+    /// catastrophic packet loss, without a possibility of repair. Prefer
+    /// [`TransportConfig::initial_max_udp_payload_size`] together with
+    /// [`TransportConfig::mtu_discovery_config`] to set a maximum UDP payload size that robustly
+    /// adapts to the network.
+    pub fn max_guaranteed_udp_payload_size(&mut self, value: u16) -> &mut Self {
+        self.max_guaranteed_udp_payload_size = value.max(INITIAL_MAX_UDP_PAYLOAD_SIZE);
```
Ralith, line 192: preexisting nit: should we replace "udp_payload_size" here and above with "mtu" for consistency/clarity?

Ralith (approved):

quinn-proto/src/config.rs
aochagavia: You mean from `max_guaranteed_udp_payload_size` to `max_guaranteed_mtu`? I have been avoiding the term "MTU" because I find it ambiguous (e.g. it could also mean the max IP datagram size). If you are confident this change will clarify things, then I'm all for it, though. In that case, should we also rename `initial_max_udp_payload_size` to `initial_mtu`, and other usages of `udp_payload_size` throughout the codebase?

Ralith: I agree that there's some risk of confusion here and I'm interested in other perspectives. I think clarifying in the doc comments will help, as will consistent use of "MTU" throughout the codebase/docs. For example, I think the proposed renaming is more consistent with the use of "MTU" in `MtuDiscoveryConfig` and its documentation than the status quo. We should definitely keep `initial_max_udp_payload_size` consistent with whatever we do here.

Also happy to leave this to be resolved in a separate PR, if you just want to get the rest in.

aochagavia: I just went throught all usages of `max_udp_payload_size` and pushed a commit to update them. The main changes are:

* `initial_max_udp_payload_size` -> `initial_mtu`
* `max_guaranteed_udp_payload_size` -> `min_guaranteed_mtu` (note the change in the prefix from `max` to `min`, as this is the minimum MTU that is guaranteed to be supported).
* Variables called (or containing) `max_udp_payload_size` -> `mtu`

I left everything related to the `max_udp_payload_size` transport parameter untouched, so we keep using the names from the spec.

What do you think? Feel free to amend the last commit if there are any obvious improvements that would allow this being merged (I won't have time to tweak it anymore till next week).

Ralith (approved):
LGTM, thanks!

quinn-proto/src/config.rs
Ralith: I think that's a solid improvement in readability across the board, yeah :+1:

aochagavia:
Just FYI, I'd like to fininsh the ACK frequency PR no later than next week (it is not 100% sure yet whether funding will extend after that period), so it would be helpful if this can get merged soon.

quinn-proto/src/connection/mod.rs
```diff
@@ -1470,15 +1470,20 @@ impl Connection {
...
                 self.remove_in_flight(pn_space, &info);
                 for frame in info.stream_frames {
                     self.streams.retransmit(frame);
                 }
                 self.spaces[pn_space].pending |= info.retransmits;
-                self.path.mtud.on_non_probe_lost(now, *packet, info.size);
+                self.path.mtud.on_non_probe_lost(*packet, info.size);
             }
+
+            let black_hole_detected = self.path.mtud.black_hole_detected(now);
```
djc, line 1493: Style nit: I think this would be easier to understand if it didn't try to be fancy with the `bool` to `u64` conversion, instead writing this as an `if`-block.

quinn-proto/src/connection/mtud.rs
```diff
@@ -349,6 +348,127 @@ impl SearchState {
...
+impl BlackHoleDetector {
+    fn new() -> Self {
+        Self {
+            suspicious_loss_bursts: 0,
+            largest_non_probe_lost: None,
+            loss_burst_has_non_suspicious_packets: false,
+            largest_suspicious_packet_lost: None,
+            largest_acked_mtu_sized_packet: None,
+        }
+    }
```
djc: Can we ditch this in favor of a `Default` derive?

quinn-proto/src/connection/mtud.rs
```diff
@@ -131,31 +130,31 @@ impl MtuDiscovery {
...
     /// Notifies the [`MtuDiscovery`] that a non-probe packet was lost
-    pub(crate) fn on_non_probe_lost(&mut self, now: Instant, packet_number: u64, packet_size: u16) {
-        // Ignore packets smaller or equal than the base size
-        if packet_size <= BASE_UDP_PAYLOAD_SIZE {
-            return;
-        }
+    ///
+    /// When done notifying of lost packets, [`MtuDiscovery::black_hole_detected`] must be called, to
+    /// ensure the last loss burst is properly processed and to trigger black hole recovery logic if
+    /// necessary
```
djc, line 145: Nit: please end this sentence with a period.

quinn-proto/src/congestion.rs
```diff
@@ -58,6 +58,9 @@ pub trait Controller: Send {
         lost_bytes: u64,
     );
 
+    /// The known MTU for the current network path has been updated
+    fn on_mtu_update(&mut self, current_mtu: u16);
```
djc, line 62: Should we call the argument `new_mtu` instead?

quinn-proto/src/congestion/bbr/mod.rs
```diff
@@ -344,8 +345,8 @@ impl Bbr {
         if self.recovery_window >= bytes_lost {
             self.recovery_window -= bytes_lost;
         } else {
-            const K_MAX_SEGMENT_SIZE: u64 = MAX_DATAGRAM_SIZE;
-            self.recovery_window = K_MAX_SEGMENT_SIZE;
+            let k_max_segment_size: u64 = self.current_mtu;
```
djc, line 348: Style nit: the indirection through a named binding (with an unnecessary type annotation) seems pretty pointless, how about assigning directly and annotating with a comment to clarify the context here?

djc (approved):
Approved with nits.

Sorry for the slow review, I'm on vacation.

quinn-proto/src/connection/mtud.rs
aochagavia: We can't, because a later commit introduced an additional field for `min_mtu`, provided as an argument to `BlackHoleDetector::new`

quinn-proto/src/connection/mod.rs
aochagavia: I agree, just updated it

quinn-proto/src/congestion.rs
aochagavia: Sure, that seems indeed a bit clearer

quinn-proto/src/congestion/bbr/mod.rs
aochagavia: Good catch, it looks like I forgot to clean up after replacing the constant by the current mtu

aochagavia:
Just amended the relevant commits to apply @djc's suggestions

djc:
Thanks!

PR #1534, opened 2023-04-14
quinn: take Arc<dyn Runtime> directly

djc:
See #1530 and https://github.com/quinn-rs/quinn/pull/1496#discussion_r1142773904.


Ralith (approved):
Thanks!

PR #1535, opened 2023-04-14
Bump quinn-udp version to 0.4

link2xt:
It includes a fallback for old Linux and Android systems which do not support setting ECN flag on UDP packets.

### DIFF
--- a/quinn-udp/Cargo.toml
+++ b/quinn-udp/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quinn-udp"
-version = "0.3.2"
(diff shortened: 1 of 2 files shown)


Ralith:
A breaking public dependency version change is a breaking change, so this should be 0.4.

djc:
@link2xt can you bump this to 0.4.0, please?

Ralith (approved):

Ralith (changes requested):
Need to update the version number `quinn` asks for to match.

Ralith (approved):

PR #1543, opened 2023-04-19
Implement acknowledgement frequency

aochagavia:
_Note: this PR is no longer a draft, but I've left the text below untouched so people can trace how the PR was developed. You will probably want to jump to [this comment](https://github.com/quinn-rs/quinn/pull/1543#issuecomment-1531658463) for an overview of what this PR actually achieves._

---

This is a draft PR for now, because it is missing the following stuff:

* Regularly sending `ACK_FREQUENCY` frames to actively control the ACK frequency of the peer (I'm waiting for #1529 to be merged to avoid conflicts around the congestion control code)
* Polishing out-of-order packet detection (more on that below, but the current code to handle out-of-order ack-eliciting packets is probably wrong)
* Adding a few more tests

As to the things that _are_ properly implemented:

* `min_ack_delay` transport parameter
* `ACK_FREQUENCY` and `IMMEDIATE_ACK` frames
* Logic around delaying ACKs up to `max_ack_delay`, properly tweaked upon receiving an `ACK_FREQUENCY` frame
* Receiving an `IMMEDIATE_ACK` causes an ACK to be sent in the next `poll_transmit` call, even if that results in an ACK-only packet.
* If the remote peer supports it, we send an `IMMEDIATE_ACK` in the following cases:
    * When a PMTU probe is sent (see section 8.3 of the draft)
    * When a client wants to validate the path after connection migration (see section 8.4 of the draft)
    * When a loss probe is sent (see section 5 of the draft)

That being said, here are a few questions:

1. What do you think of the general direction of the PR? Do you see anything that seems out of place?
2. Is there a way to access unencrypted packet frames in tests that simulate a connection (`quinn-proto/src/test/mod.rs`)? It would be useful to ensure the ACK delay value in ACK frames is being properly set. Unit tests don’t give the same degree of confidence and have a higher maintenance cost. If this doesn't exist yet, would you welcome a commit adding it?
3. I'm confused about the way out-of-order packets are detected in the ACK frequency draft, which seems different than in RFC 9000. The draft gives an example that, to my understanding, is wrong (which probably means my understanding is wrong). I posted a detailed question in the [quicdev slack](https://quicdev.slack.com/archives/C68J3KSH3/p1681895302006479) to get more clarity, so please have a look if you think you can help.

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -39,6 +39,7 @@ pub struct TransportConfig {
     pub(crate) initial_mtu: u16,
     pub(crate) min_guaranteed_mtu: u16,
     pub(crate) mtu_discovery_config: Option<MtuDiscoveryConfig>,
+    pub(crate) ack_frequency_config: Option<AckFrequencyConfig>,
 
     pub(crate) persistent_congestion_threshold: u32,
     pub(crate) keep_alive_interval: Option<Duration>,
@@ -218,6 +219,17 @@ impl TransportConfig {
         self
     }
 
+    /// Specifies the ACK frequency config (see [`AckFrequencyConfig`] for details).
+    ///
+    /// The provided configuration will be ignored if the peer does not support the acknowledgement
+    /// frequency QUIC extension.
+    ///
+    /// Defaults to `None`, which disables the ACK frequency feature.
+    pub fn ack_frequency_config(&mut self, value: Option<AckFrequencyConfig>) -> &mut Self {
+        self.ack_frequency_config = value;
+        self
+    }
+
     /// Number of consecutive PTOs after which network is considered to be experiencing persistent congestion.
     pub fn persistent_congestion_threshold(&mut self, value: u32) -> &mut Self {
         self.persistent_congestion_threshold = value;
@@ -316,6 +328,7 @@ impl Default for TransportConfig {
             initial_mtu: INITIAL_MTU,
             min_guaranteed_mtu: INITIAL_MTU,
             mtu_discovery_config: None,
+            ack_frequency_config: None,
 
             persistent_congestion_threshold: 3,
             keep_alive_interval: None,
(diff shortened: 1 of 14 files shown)


Ralith:
> Is there a way to access unencrypted packet frames in tests that simulate a connection (quinn-proto/src/test/mod.rs)?

Not presently. We've generally relied on making assertions about resulting connection state. You could do that here by looking at the RTT estimate, but that is a bit indirect. I don't see anything fundamentally wrong with asserting about the transmitted plaintext (we do something morally similar in the version negotiation tests, after all), but it's not immediately obvious to me how it might be done gracefully.

aochagavia:
> I don't see anything fundamentally wrong with asserting about the transmitted plaintext (we do something morally similar in the version negotiation tests, after all), but it's not immediately obvious to me how it might be done gracefully.

I just pushed a [commit](https://github.com/quinn-rs/quinn/pull/1543/commits/2650a7489729aed14e65039da8aa0d6285c2c7d3) prototyping how this could work. It introduces a `Connection::decode_packet` function that returns the decoded bytes of the packet's payload, meant to be used from tests. It duplicates about 70 lines of code, though, and it requires deriving `Clone` for two structs (we could make the latter conditional for testing, if desired). It feels a bit awkward at the moment because of the code duplication, but with a bit of effort I might be able to extract that code into a dedicated module that can be shared. Do you think this is a good direction?

By the way, I have had no feedback on the implementer's Slack about detection of out-of-order packets in the ACK frequency draft, so my current plan is to go through the draft again next week and implement my own interpretation of it. I also plan to look around to see if there are open implementations of the current version of the draft that I can have a look.

Ralith:
> Do you think this is a good direction?

Seems reasonable in general (especially with `#[cfg(test)]`), but yeah factoring out the common logic would be nice. I have no problem with adding `Clone` to internal types; its absence generally isn't principled, we just haven't needed it before.

> I have had no feedback on the implementer's Slack about detection of out-of-order packets in the ACK frequency draft

I saw; that was disappointing. Maybe open an issue (or PR with a correction) on https://github.com/quicwg/ack-frequency? The WG has been good about engaging on github historically. I definitely wouldn't be shocked by an error in an example.

aochagavia:
And... here are the benchmarks!

### Practical details

I'm running the server using `cargo run --release --bin perf_server -- --no-protection` and the client using `cargo run --release --bin perf_client -- --no-protection --duration 15`.

For the scenarios with ACK frequency enabled I'm doing so through the following code:

```rust
let mut ack_freq = AckFrequencyConfig::default();
ack_freq.ack_eliciting_threshold(10);
transport.ack_frequency_config(Some(ack_freq));
```

### Baseline (before this PR)

```
 AVG  │          1.19ms │            1.12ms │    75.00µs │      896.07 MiB/s │        966.58 MiB/s
 P0   │        816.00µs │          625.00µs │     1.00µs │      210.25 MiB/s │         31.00 MiB/s
 P10  │        873.00µs │          818.00µs │     3.00µs │      627.00 MiB/s │        712.50 MiB/s
 P50  │          1.18ms │            1.08ms │    68.00µs │      846.50 MiB/s │        926.00 MiB/s
 P90  │          1.60ms │            1.40ms │   138.00µs │     1146.00 MiB/s │       1223.00 MiB/s
 P100 │          4.76ms │           32.26ms │   480.00µs │     1226.00 MiB/s │       1601.00 MiB/s
```

### ACK frequency disabled (after this PR)

```
 AVG  │        950.00µs │          857.00µs │    51.00µs │     1175.44 MiB/s │       1232.08 MiB/s
 P0   │        693.00µs │          106.00µs │     1.00µs │       34.59 MiB/s │         34.72 MiB/s
 P10  │        746.00µs │          733.00µs │    26.00µs │      965.50 MiB/s │       1035.00 MiB/s
 P50  │        827.00µs │          803.00µs │    46.00µs │     1210.00 MiB/s │       1246.00 MiB/s
 P90  │          1.04ms │          967.00µs │    81.00µs │     1341.00 MiB/s │       1365.00 MiB/s
 P100 │         28.90ms │           28.80ms │   799.00µs │     1444.00 MiB/s │       9440.00 MiB/s
```

There is a clear improvement in both upload and download throughput, because now we send an ACK every 2 ack-eliciting packets by default.

### ACK frequency enabled on the client (threshold = 10)

```
 AVG  │        869.00µs │          805.00µs │    50.00µs │     1191.91 MiB/s │       1466.07 MiB/s
 P0   │        684.00µs │           71.00µs │     1.00µs │       30.00 MiB/s │         19.44 MiB/s
 P10  │        751.00µs │          698.00µs │     5.00µs │      992.50 MiB/s │       1166.00 MiB/s
 P50  │        810.00µs │          756.00µs │    46.00µs │     1235.00 MiB/s │       1323.00 MiB/s
 P90  │          1.01ms │          858.00µs │    77.00µs │     1332.00 MiB/s │       1433.00 MiB/s
 P100 │         33.34ms │           51.46ms │   826.00µs │     1462.00 MiB/s │      14088.00 MiB/s
```

There is a clear improvement in download throughput, probably because the server now spends less time creating ACK packets and can send data at a higher rate (note that, when ACK frequency is enabled on the client-side, the client will ask the server to modify its ACK frequency).

### ACK frequency enabled on the server (threshold = 10)

```
 AVG  │        890.00µs │          844.00µs │    51.00µs │     1182.28 MiB/s │       1202.34 MiB/s
 P0   │        684.00µs │          122.00µs │     1.00µs │       34.59 MiB/s │        100.25 MiB/s
 P10  │        766.00µs │          753.00µs │     2.00µs │     1016.50 MiB/s │       1064.00 MiB/s
 P50  │        826.00µs │          828.00µs │    49.00µs │     1211.00 MiB/s │       1208.00 MiB/s
 P90  │        984.00µs │          940.00µs │    86.00µs │     1306.00 MiB/s │       1329.00 MiB/s
 P100 │         28.90ms │            9.98ms │   427.00µs │     1462.00 MiB/s │       8200.00 MiB/s
```

This one seems to make no difference, and maybe even worsen performance a bit. I haven't found out yet why this is the case. I expected to improvements in upload throughput, but somehow they aren't there.

aochagavia:
@djc it looks like you closed this PR by accident

djc:
Ahh, sorry -- I deleted the branch for #1552, which apparently makes it impossible to reopen this PR. Can you try repushing this branch so we can reopen it? We could open a new PR but would be better to keep this one going.

aochagavia:
Ah yes, because the PR was targetting that branch... I can also no longer edit the target branch of the PR, even after force-pushing a rebased version, so it looks like I'll have to open a new one.

aochagavia:
And here is the new PR: #1553

Issue #438, opened 2019-09-18
Implement delayed acknowledgement


Ralith:
[Generating Acknowledgements](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#generating-acks):
> In general, frequent feedback from a receiver improves loss and congestion
> response, but this has to be balanced against excessive load generated by a
> receiver that sends an ACK frame in response to every ack-eliciting packet.  The
> guidance offered below seeks to strike this balance.
> 
> An ACK frame SHOULD be generated for at least every second ack-eliciting packet.
> This recommendation is in keeping with standard practice for TCP [[RFC5681]](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#RFC5681).

Detailed advice for cases when delaying acknowledgement is and is not appropriate follows. Due to the complexity cost, we should probably confirm the benefit of doing so with benchmarks.

A heavily-loaded Quinn receiver will already only generate one acknowledgement per batch of packets it consumes, so this may be less important for us than for a naive implementation.

Ralith:
https://www.fastly.com/blog/measuring-quic-vs-tcp-computational-efficiency reports that the performance benefit of delayed ACKs is significant. We should pursue implementation.

Ralith:
Also of potential interest is a draft extension for coordinating ACK delay configuration: https://tools.ietf.org/html/draft-iyengar-quic-delayed-ack-00

Ralith:
We should also stop attaching outstanding the complete list of unacknowledged ACKs to every single outgoing packet, as this places excess load on the peer. Instead, we should only transmit an ACK if new packets have been received and either the maximum ACK delay has elapsed OR the number of new packets received exceeds a tunable threshold (default 2-10).

Ralith:
An implementation should immediately send an ACK upon receiving [an out of order packet](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#section-1.2-3.8).

stormshield-damiend:
New draft that replace previous ones: https://datatracker.ietf.org/doc/draft-ietf-quic-ack-frequency/

aochagavia:
I'll be working on this in the coming weeks

aochagavia:
[...]

PR #1553, opened 2023-05-03
Implement acknowledgement frequency

aochagavia:
Overview of the feature:

* The ack frequency extension allows one to control the ACK frequency of the remote peer.
* Quinn provides an `AckFrequencyConfig` to control the parameters, which are sent to the peer in an `ACK_FREQUENCY` frame at the beginning of the connection (technically, the extension allows modifying the ack frequency parameters multiple times during a connection, but then you would need to come up with an algorithm to automatically tune the parameters, which IMO is out of scope for this first implementation).
* On the other side of the connection, Quinn will receive and process the `ACK_FREQUENCY` frame, updating the local ack frequency parameters based on the new parameters sent by the peer.
* If the local peer has requested a custom ack frequency from the remote peer, Quinn will send an `IMMEDIATE_ACK` frame every RTT if there are any unacked ack-eliciting packets (this is a new kind of frame to immediately elicit an ACK). This is recommended by the draft as a way to ensure congestion control and loss detection work properly.
* The draft is ambiguous when it comes to out-of-order packet detection. I wrote the specific details in [this issue](https://github.com/quicwg/ack-frequency/issues/183), hoping it gets clarified. In the meantime, I am sticking to the text of the draft and ignoring the parts in the examples that deviate from the draft's definitions.

I have a bunch of questions / comments:

1. I introduced two timers, one to be notified when `max_ack_delay` is elapsed, and one to be notified when an RTT is elapsed. I tried to come up with an approach that wouldn't require additional timers, but I'm not sure it is possible. It would be nice if you can double-check that, given your knowledge of the codebase.
1. I introduced a `reset_rtt_timer` function that arms the `Timer::Rtt` for the first time. I wanted to arm it right after we have reached the Data space and have an RTT estimate. I ended up calling it inside `process_payload`. Please let me know if there is a more suitable place.
1. I rewrote a long comment inside `PendingAcks::acks_sent`, which I found difficult to understand. Hopefully the new version says the same in clearer terms (otherwise let me know and I'll update it)

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -39,6 +39,7 @@ pub struct TransportConfig {
     pub(crate) initial_mtu: u16,
     pub(crate) min_mtu: u16,
     pub(crate) mtu_discovery_config: Option<MtuDiscoveryConfig>,
+    pub(crate) ack_frequency_config: Option<AckFrequencyConfig>,
 
     pub(crate) persistent_congestion_threshold: u32,
     pub(crate) keep_alive_interval: Option<Duration>,
@@ -218,6 +219,17 @@ impl TransportConfig {
         self
     }
 
+    /// Specifies the ACK frequency config (see [`AckFrequencyConfig`] for details)
+    ///
+    /// The provided configuration will be ignored if the peer does not support the acknowledgement
+    /// frequency QUIC extension.
+    ///
+    /// Defaults to `None`, which disables the ACK frequency feature.
+    pub fn ack_frequency_config(&mut self, value: Option<AckFrequencyConfig>) -> &mut Self {
+        self.ack_frequency_config = value;
+        self
+    }
+
     /// Number of consecutive PTOs after which network is considered to be experiencing persistent congestion.
     pub fn persistent_congestion_threshold(&mut self, value: u32) -> &mut Self {
         self.persistent_congestion_threshold = value;
@@ -316,6 +328,7 @@ impl Default for TransportConfig {
             initial_mtu: INITIAL_MTU,
             min_mtu: INITIAL_MTU,
             mtu_discovery_config: Some(MtuDiscoveryConfig::default()),
+            ack_frequency_config: None,
 
             persistent_congestion_threshold: 3,
             keep_alive_interval: None,
@@ -362,6 +375,84 @@ impl fmt::Debug for TransportConfig {
             .field("datagram_send_buffer_size", &self.datagram_send_buffer_size)
             .field("congestion_controller_factory", &"[ opaque ]")
             .finish()
+    }
+}
+
+/// Parameters for controlling the peer's acknowledgement frequency
+///
+/// The parameters provided in this config will be sent to the peer at the beginning of the
+/// connection, so it can take them into account when sending acknowledgements (see each parameter's
+/// description for details on how it influences acknowledgement frequency).
+///
+/// Quinn's implementation follows the fourth draft of the
+/// [QUIC Acknowledgement Frequency extension](https://datatracker.ietf.org/doc/html/draft-ietf-quic-ack-frequency-04).
+/// The defaults produce behavior slightly different than the behavior without this extension,
+/// because they change the way reordered packets are handled (see
+/// [`AckFrequencyConfig::reordering_threshold`] for details).
+#[derive(Clone, Debug)]
+pub struct AckFrequencyConfig {
+    pub(crate) ack_eliciting_threshold: VarInt,
+    pub(crate) max_ack_delay: Option<Duration>,
+    pub(crate) reordering_threshold: VarInt,
+}
+
+impl AckFrequencyConfig {
+    /// The ack-eliciting threshold we will request the peer to use
+    ///
+    /// This threshold represents the number of ack-eliciting packets an endpoint may receive
+    /// without immediately sending an ACK.
+    ///
+    /// The remote peer should send at least one ACK frame when more than this number of
+    /// ack-eliciting packets have been received. A value of 0 results in a receiver immediately
+    /// acknowledging every ack-eliciting packet.
+    ///
+    /// Defaults to 1, which sends ACK frames for every other ack-eliciting packet.
+    pub fn ack_eliciting_threshold(&mut self, value: VarInt) -> &mut Self {
+        self.ack_eliciting_threshold = value;
+        self
+    }
+
+    /// The `max_ack_delay` we will request the peer to use
+    ///
+    /// This parameter represents the maximum amount of time that an endpoint waits before sending
+    /// an ACK when the ack-eliciting threshold hasn't been reached.
+    ///
+    /// The provided `max_ack_delay` will be clamped to be at least the peer's `min_ack_delay`
+    /// transport parameter.
+    ///
+    /// Defaults to `None`, in which case the peer's original `max_ack_delay` will be used, as
+    /// obtained from its transport parameters.
+    pub fn max_ack_delay(&mut self, value: Option<Duration>) -> &mut Self {
+        self.max_ack_delay = value;
+        self
+    }
+
+    /// The reordering threshold we will request the peer to use
+    ///
+    /// This threshold represents the amount of out-of-order packets that will trigger an endpoint
+    /// to send an ACK, without waiting for `ack_eliciting_threshold` to be exceeded or for
+    /// `max_ack_delay` to be elapsed.
+    ///
+    /// A value of 0 indicates out-of-order packets do not elicit an immediate ACK. A value of 1
+    /// immediately acknowledges any packets that are received out of order (this is also the
+    /// behavior when the extension is disabled).
+    ///
+    /// It is recommended to set this value to [`TransportConfig::packet_threshold`] minus one.
+    /// Since the default value for [`TransportConfig::packet_threshold`] is 3, this value defaults
+    /// to 2.
+    pub fn reordering_threshold(&mut self, value: VarInt) -> &mut Self {
+        self.reordering_threshold = value;
+        self
+    }
+}
+
+impl Default for AckFrequencyConfig {
+    fn default() -> Self {
+        Self {
+            ack_eliciting_threshold: VarInt(1),
+            max_ack_delay: None,
+            reordering_threshold: VarInt(2),
+        }
     }
 }
 
--- a/quinn-proto/src/connection/ack_frequency.rs
+++ b/quinn-proto/src/connection/ack_frequency.rs
@@ -1,0 +1,124 @@
+use crate::connection::spaces::PendingAcks;
+use crate::frame::AckFrequency;
+use crate::{AckFrequencyConfig, TransportError, VarInt, TIMER_GRANULARITY};
+use std::time::Duration;
+
+/// State associated to ACK frequency
+pub(super) struct AckFrequencyState {
+    //
+    // Sending ACK_FREQUENCY frames
+    //
+    in_flight_ack_frequency_frame: Option<(u64, Duration)>,
+    next_outgoing_sequence_number: VarInt,
+    pub(super) peer_max_ack_delay: Duration,
+
+    //
+    // Receiving ACK_FREQUENCY frames
+    //
+    last_ack_frequency_frame: Option<u64>,
+    pub(super) max_ack_delay: Duration,
+}
+
+impl AckFrequencyState {
+    pub(super) fn new(default_max_ack_delay: Duration) -> Self {
+        Self {
+            in_flight_ack_frequency_frame: None,
+            next_outgoing_sequence_number: VarInt(0),
+            peer_max_ack_delay: default_max_ack_delay,
+
+            last_ack_frequency_frame: None,
+            max_ack_delay: default_max_ack_delay,
+        }
+    }
+
+    /// Returns the `max_ack_delay` that should be requested of the peer when sending an
+    /// ACK_FREQUENCY frame
+    pub(super) fn candidate_max_ack_delay(&self, config: &AckFrequencyConfig) -> Duration {
+        // Use the peer's max_ack_delay if no custom max_ack_delay was provided in the config
+        config.max_ack_delay.unwrap_or(self.peer_max_ack_delay)
+    }
+
+    /// Returns the `max_ack_delay` for the purposes of calculating the PTO
+    ///
+    /// This `max_ack_delay` is defined as the maximum of the peer's current `max_ack_delay` and all
+    /// in-flight `max_ack_delay`s (i.e. proposed values that haven't been acknowledged yet, but
+    /// might be already in use by the peer).
+    pub(super) fn max_ack_delay_for_pto(&self) -> Duration {
+        // Note: we have at most one in-flight ACK_FREQUENCY frame
+        if let Some((_, max_ack_delay)) = self.in_flight_ack_frequency_frame {
+            self.peer_max_ack_delay.max(max_ack_delay)
+        } else {
+            self.peer_max_ack_delay
+        }
+    }
+
+    /// Returns the next sequence number for an ACK_FREQUENCY frame
+    pub(super) fn next_sequence_number(&mut self) -> VarInt {
+        assert!(self.next_outgoing_sequence_number <= VarInt::MAX);
+
+        let seq = self.next_outgoing_sequence_number;
+        self.next_outgoing_sequence_number.0 += 1;
+        seq
+    }
+
+    /// Returns true if we should send an ACK_FREQUENCY frame
+    pub(super) fn should_send_ack_frequency(&self) -> bool {
+        // Currently, we only allow sending a single ACK_FREQUENCY frame. There is no need to send
+        // more, because none of the sent values needs to be updated in the course of the connection
+        self.next_outgoing_sequence_number.0 == 0
+    }
+
+    /// Notifies the [`AckFrequencyState`] that a packet containing an ACK_FREQUENCY frame was sent
+    pub(super) fn ack_frequency_sent(&mut self, pn: u64, requested_max_ack_delay: Duration) {
+        self.in_flight_ack_frequency_frame = Some((pn, requested_max_ack_delay));
+    }
+
+    /// Notifies the [`AckFrequencyState`] that a packet has been ACKed
+    pub(super) fn on_acked(&mut self, pn: u64) {
+        match self.in_flight_ack_frequency_frame {
+            Some((number, requested_max_ack_delay)) if number == pn => {
+                self.in_flight_ack_frequency_frame = None;
+                self.peer_max_ack_delay = requested_max_ack_delay;
+            }
+            _ => {}
+        }
+    }
+
+    /// Notifies the [`AckFrequencyState`] that an ACK_FREQUENCY frame was received
+    ///
+    /// Updates the endpoint's params according to the payload of the ACK_FREQUENCY frame, or
+    /// returns an error in case the requested `max_ack_delay` is invalid.
+    ///
+    /// Returns `true` if the frame was processed and `false` if it was ignored because of being
+    /// stale.
+    pub(super) fn ack_frequency_received(
+        &mut self,
+        frame: &AckFrequency,
+        pending_acks: &mut PendingAcks,
+    ) -> Result<bool, TransportError> {
+        if self
+            .last_ack_frequency_frame
+            .map_or(false, |highest_sequence_nr| {
+                frame.sequence.into_inner() <= highest_sequence_nr
+            })
+        {
+            return Ok(false);
+        }
+
+        self.last_ack_frequency_frame = Some(frame.sequence.into_inner());
+
+        // Update max_ack_delay
+        let max_ack_delay = Duration::from_micros(frame.request_max_ack_delay.into_inner());
+        if max_ack_delay < TIMER_GRANULARITY {
+            return Err(TransportError::PROTOCOL_VIOLATION(
+                "Requested Max Ack Delay in ACK_FREQUENCY frame is less than min_ack_delay",
+            ));
+        }
+        self.max_ack_delay = max_ack_delay;
+
+        // Update the rest of the params
+        pending_acks.set_ack_frequency_params(frame);
+
+        Ok(true)
+    }
+}
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -19,7 +19,7 @@ use crate::{
     cid_queue::CidQueue,
     coding::BufMutExt,
     config::{ServerConfig, TransportConfig},
-    crypto::{self, HeaderKey, KeyPair, Keys, PacketKey},
+    crypto::{self, KeyPair, Keys, PacketKey},
     frame,
     frame::{Close, Datagram, FrameStruct},
     packet::{Header, LongType, Packet, PartialDecode, SpaceId},
@@ -31,7 +31,7 @@ use crate::{
     token::ResetToken,
     transport_parameters::TransportParameters,
     Dir, EndpointConfig, Frame, Side, StreamId, Transmit, TransportError, TransportErrorCode,
-    VarInt, MAX_STREAM_COUNT, MIN_INITIAL_SIZE, RESET_TOKEN_SIZE, TIMER_GRANULARITY,
+    VarInt, MAX_STREAM_COUNT, MIN_INITIAL_SIZE, TIMER_GRANULARITY,
 };
 
 mod ack_frequency;
@@ -52,6 +52,9 @@ mod pacing;
 
 mod packet_builder;
 use packet_builder::PacketBuilder;
+
+mod packet_crypto;
+use packet_crypto::ZeroRttCrypto;
 
 mod paths;
 use paths::PathData;
@@ -2027,40 +2030,13 @@ impl Connection {
         ecn: Option<EcnCodepoint>,
         partial_decode: PartialDecode,
     ) {
-        let header_crypto = if partial_decode.is_0rtt() {
-            if let Some(ref crypto) = self.zero_rtt_crypto {
-                Some(&*crypto.header)
-            } else {
-                debug!("dropping unexpected 0-RTT packet");
-                return;
-            }
-        } else if let Some(space) = partial_decode.space() {
-            if let Some(ref crypto) = self.spaces[space].crypto {
-                Some(&*crypto.header.remote)
-            } else {
-                debug!(
-                    "discarding unexpected {:?} packet ({} bytes)",
-                    space,
-                    partial_decode.len(),
-                );
-                return;
-            }
-        } else {
-            // Unprotected packet
-            None
-        };
-
-        let packet = partial_decode.data();
-        let stateless_reset = packet.len() >= RESET_TOKEN_SIZE + 5
-            && self.peer_params.stateless_reset_token.as_deref()
-                == Some(&packet[packet.len() - RESET_TOKEN_SIZE..]);
-
-        match partial_decode.finish(header_crypto) {
-            Ok(packet) => self.handle_packet(now, remote, ecn, Some(packet), stateless_reset),
-            Err(_) if stateless_reset => self.handle_packet(now, remote, ecn, None, true),
-            Err(e) => {
-                trace!("unable to complete packet decoding: {}", e);
-            }
+        if let Some(decoded) = packet_crypto::unprotect_header(
+            partial_decode,
+            &self.spaces,
+            self.zero_rtt_crypto.as_ref(),
+            self.peer_params.stateless_reset_token,
+        ) {
+            self.handle_packet(now, remote, ecn, decoded.packet, decoded.stateless_reset);
         }
     }
 
@@ -3676,11 +3652,6 @@ const MIN_PACKET_SPACE: usize = 40;
 /// memory allocations when calling `poll_transmit()`. Benchmarks have shown
 /// that numbers around 10 are a good compromise.
 const MAX_TRANSMIT_SEGMENTS: usize = 10;
-
-struct ZeroRttCrypto {
-    header: Box<dyn HeaderKey>,
-    packet: Box<dyn PacketKey>,
-}
 
 #[derive(Default)]
 struct SentFrames {
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -3293,6 +3293,43 @@ impl Connection {
         self.spaces[self.highest_space].immediate_ack_pending = true;
     }
 
+    /// Decodes a packet, returning its decrypted payload, so it can be inspected in tests
+    #[cfg(test)]
+    pub(crate) fn decode_packet(&self, event: &ConnectionEvent) -> Option<Vec<u8>> {
+        let (first_decode, remaining) = match &event.0 {
+            ConnectionEventInner::Datagram {
+                first_decode,
+                remaining,
+                ..
+            } => (first_decode, remaining),
+            _ => return None,
+        };
+
+        if remaining.is_some() {
+            panic!("Packets should never be coalesced in tests");
+        }
+
+        let decrypted_header = packet_crypto::unprotect_header(
+            first_decode.clone(),
+            &self.spaces,
+            self.zero_rtt_crypto.as_ref(),
+            self.peer_params.stateless_reset_token,
+        )?;
+
+        let mut packet = decrypted_header.packet?;
+        packet_crypto::decrypt_packet_body(
+            &mut packet,
+            &self.spaces,
+            self.zero_rtt_crypto.as_ref(),
+            self.key_phase,
+            self.prev_crypto.as_ref(),
+            self.next_crypto.as_ref(),
+        )
+        .ok()?;
+
+        Some(packet.payload.to_vec())
+    }
+
     /// The number of bytes of packets containing retransmittable frames that have not been
     /// acknowledged or declared lost.
     #[cfg(test)]
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -34,6 +34,9 @@ use crate::{
     VarInt, MAX_STREAM_COUNT, MIN_INITIAL_SIZE, RESET_TOKEN_SIZE, TIMER_GRANULARITY,
 };
 
+mod ack_frequency;
+use ack_frequency::AckFrequencyState;
+
 mod assembler;
 pub use assembler::Chunk;
 
@@ -191,6 +194,11 @@ pub struct Connection {
     close: bool,
 
     //
+    // ACK frequency
+    //
+    ack_frequency: AckFrequencyState,
+
+    //
     // Loss Detection
     //
     /// The number of times a PTO has been sent without receiving an ack.
@@ -307,6 +315,10 @@ impl Connection {
 
             path_response: None,
             close: false,
+
+            ack_frequency: AckFrequencyState::new(get_max_ack_delay(
+                &TransportParameters::default(),
+            )),
 
             pto_count: 0,
 
@@ -479,7 +491,9 @@ impl Connection {
 
         // If we need to send a probe, make sure we have something to send.
         for space in SpaceId::iter() {
-            self.spaces[space].maybe_queue_probe(&self.streams);
+            let request_immediate_ack =
+                space == SpaceId::Data && self.peer_supports_ack_frequency();
+            self.spaces[space].maybe_queue_probe(request_immediate_ack, &self.streams);
         }
 
         // Check whether we need to send a close message
@@ -499,6 +513,15 @@ impl Connection {
             }
             _ => false,
         };
+
+        // Check whether we need to send an ACK_FREQUENCY frame
+        if self.ack_frequency.should_send_ack_frequency()
+            && self.highest_space == SpaceId::Data
+            && self.peer_supports_ack_frequency()
+            && self.config.ack_frequency_config.is_some()
+        {
+            self.spaces[SpaceId::Data].pending.ack_frequency = true;
+        }
 
         let mut buf = BytesMut::new();
         // Reserving capacity can provide more capacity than we asked for.
@@ -535,7 +558,8 @@ impl Connection {
             }
 
             let mut ack_eliciting = !self.spaces[space_id].pending.is_empty(&self.streams)
-                || self.spaces[space_id].ping_pending;
+                || self.spaces[space_id].ping_pending
+                || self.spaces[space_id].immediate_ack_pending;
             if space_id == SpaceId::Data {
                 ack_eliciting |= self.can_send_1rtt();
             }
@@ -694,6 +718,7 @@ impl Connection {
                 // have gotten any other ACK for the data earlier on.
                 if !self.spaces[space_id].pending_acks.ranges().is_empty() {
                     Self::populate_acks(
+                        now,
                         self.receiving_ecn,
                         &mut SentFrames::default(),
                         &mut self.spaces[space_id],
@@ -740,7 +765,29 @@ impl Connection {
                 break;
             }
 
-            let sent = self.populate_packet(space_id, &mut buf, buf_capacity - builder.tag_len);
+            if builder.ack_eliciting
+                && builder.space == SpaceId::Data
+                && self.peer_supports_ack_frequency()
+                && self.config.ack_frequency_config.is_some()
+                && self.timers.get(Timer::Rtt).is_none()
+            {
+                // Request an immediate ACK to ensure we receive one within RTT, as recommended by
+                // the acknowledgement frequency draft
+                self.spaces[SpaceId::Data].immediate_ack_pending = true;
+                tracing::info!(
+                    "setting RTT timer with RTT = {} ms",
+                    self.path.rtt.get().as_millis()
+                );
+                self.timers.set(Timer::Rtt, now + self.path.rtt.get());
+            }
+
+            let sent = self.populate_packet(
+                now,
+                space_id,
+                &mut buf,
+                buf_capacity - builder.tag_len,
+                builder.exact_number,
+            );
 
             // ACK-only packets should only be sent when explicitly allowed. If we write them due
             // to any other reason, there is a bug which leads to one component announcing write
@@ -758,6 +805,7 @@ impl Connection {
 
             if sent.largest_acked.is_some() {
                 self.spaces[space_id].pending_acks.acks_sent();
+                self.timers.stop(Timer::MaxAckDelay);
             }
 
             // Keep information about the packet around until it gets finalized
@@ -809,6 +857,14 @@ impl Connection {
 
             // We implement MTU probes as ping packets padded up to the probe size
             buf.write(frame::Type::PING);
+            self.stats.frame_tx.ping += 1;
+
+            // If supported by the peer, we want no delays to the probe's ACK
+            if self.peer_supports_ack_frequency() {
+                buf.write(frame::Type::IMMEDIATE_ACK);
+                self.stats.frame_tx.immediate_ack += 1;
+            }
+
             builder.pad_to(probe_size);
             let sent_frames = SentFrames {
                 non_retransmits: true,
@@ -816,7 +872,6 @@ impl Connection {
             };
             builder.finish_and_track(now, self, Some(sent_frames), &mut buf);
 
-            self.stats.frame_tx.ping += 1;
             self.stats.path.sent_plpmtud_probes += 1;
             num_datagrams = 1;
 
@@ -1002,6 +1057,35 @@ impl Connection {
                         );
                         self.endpoint_events
                             .push_back(EndpointEventInner::NeedIdentifiers(now, num_new_cid));
+                    }
+                }
+                Timer::MaxAckDelay => {
+                    trace!("max ack delay reached");
+                    // This timer is only armed in the Data space
+                    self.spaces[SpaceId::Data]
+                        .pending_acks
+                        .on_max_ack_delay_timeout()
+                }
+                Timer::Rtt => {
+                    // This timer is only armed when ACK frequency has been enabled, the peer
+                    // supports it, and we are in the Data space
+                    let space = &mut self.spaces[SpaceId::Data];
+                    trace!(
+                        in_flight_ack_eliciting = space.sent_packets.len(),
+                        "rtt timer"
+                    );
+
+                    // It is recommended to request an immediate ACK once per RTT if there are
+                    // unacked packets. Packets with timer-triggered `IMMEDIATE_ACK` frames count
+                    // towards unacked packets, but we don't want to take them into account here.
+                    // For that reason, we only send an `IMMEDIATE_ACK` frame when there
+                    // are at least _two_ unacked packets. This is an approximation, but should work
+                    // well in practice (see
+                    // [this comment](https://github.com/quinn-rs/quinn/pull/1553#discussion_r1251268689)
+                    // for more details)
+                    if space.sent_packets.len() > 1 {
+                        space.immediate_ack_pending = true;
+                        self.timers.set(Timer::Rtt, now + self.path.rtt.get());
                     }
                 }
             }
@@ -1234,6 +1318,9 @@ impl Connection {
                         .on_mtu_update(self.path.mtud.current_mtu());
                 }
 
+                // Notify ack frequency that a packet was acked, because it might contain an ACK_FREQUENCY frame
+                self.ack_frequency.on_acked(packet);
+
                 self.on_packet_acked(now, space, info);
             }
         }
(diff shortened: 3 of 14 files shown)


aochagavia:
And... here are some benchmarks!

### Practical details

I'm running the server using `cargo run --release --bin perf_server -- --no-protection` and the client using `cargo run --release --bin perf_client -- --no-protection --duration 15` (note that this makes the benchmarks misleading for real-world use cases, because encryption is disabled). As you can deduce from the cargo commands used, I'm running both benchmarks on the same machine.

For the scenarios with ACK frequency enabled I'm doing so through the following code:

```rust
let mut ack_freq = AckFrequencyConfig::default();
ack_freq.ack_eliciting_threshold(10);
transport.ack_frequency_config(Some(ack_freq));
```

### Conclusion

There seems to be no significant performance difference in the benchmarks, though we probably need a more statistically rigorous approach to benchmarking if we want to be sure.

### Baseline (before this PR)

```
 AVG  │        862.00µs │          811.00µs │    50.00µs │     1199.97 MiB/s │       1293.80 MiB/s
 P0   │        661.00µs │          534.00µs │     1.00µs │       32.34 MiB/s │         34.34 MiB/s
 P10  │        741.00µs │          678.00µs │     2.00µs │      986.50 MiB/s │       1089.00 MiB/s
 P50  │        810.00µs │          762.00µs │    47.00µs │     1235.00 MiB/s │       1313.00 MiB/s
 P90  │          1.01ms │          919.00µs │    79.00µs │     1350.00 MiB/s │       1475.00 MiB/s
 P100 │         30.91ms │           29.12ms │   470.00µs │     1513.00 MiB/s │       1873.00 MiB/s
```

### ACK frequency disabled (threshold = 1)

```
 AVG  │        920.00µs │          805.00µs │    53.00µs │     1190.95 MiB/s │       1280.93 MiB/s
 P0   │        670.00µs │          134.00µs │     1.00µs │       33.44 MiB/s │         34.12 MiB/s
 P10  │        746.00µs │          699.00µs │     4.00µs │      996.50 MiB/s │       1113.00 MiB/s
 P50  │        819.00µs │          774.00µs │    49.00µs │     1222.00 MiB/s │       1292.00 MiB/s
 P90  │          1.00ms │          899.00µs │    82.00µs │     1341.00 MiB/s │       1431.00 MiB/s
 P100 │         29.92ms │           29.30ms │   847.00µs │     1493.00 MiB/s │       7464.00 MiB/s
```

### ACK frequency enabled on the client (threshold = 10)

```
 AVG  │        866.00µs │          767.00µs │    48.00µs │     1187.26 MiB/s │       1479.18 MiB/s
 P0   │        686.00µs │           70.00µs │     1.00µs │       33.25 MiB/s │         32.25 MiB/s
 P10  │        753.00µs │          680.00µs │     3.00µs │     1020.50 MiB/s │       1198.00 MiB/s
 P50  │        821.00µs │          749.00µs │    46.00µs │     1219.00 MiB/s │       1336.00 MiB/s
 P90  │        980.00µs │          835.00µs │    77.00µs │     1329.00 MiB/s │       1471.00 MiB/s
 P100 │         30.08ms │           30.99ms │   421.00µs │     1458.00 MiB/s │      14288.00 MiB/s
```

### ACK frequency enabled on the server (threshold = 10)

```
 AVG  │        890.00µs │          844.00µs │    51.00µs │     1182.28 MiB/s │       1202.34 MiB/s
 P0   │        684.00µs │          122.00µs │     1.00µs │       34.59 MiB/s │        100.25 MiB/s
 P10  │        766.00µs │          753.00µs │     2.00µs │     1016.50 MiB/s │       1064.00 MiB/s
 P50  │        826.00µs │          828.00µs │    49.00µs │     1211.00 MiB/s │       1208.00 MiB/s
 P90  │        984.00µs │          940.00µs │    86.00µs │     1306.00 MiB/s │       1329.00 MiB/s
 P100 │         28.90ms │            9.98ms │   427.00µs │     1462.00 MiB/s │       8200.00 MiB/s
```

aochagavia:
How do you feel about the benchmark results? It'd be nice to be able to measure things in a more principled way, to ensure any performance difference (or lack thereof) is statistically significant. Maybe it'd make sense to simulate the networking part in memory (I have a semi-working local prototype), to make the output more predictable. Any thoughts?

aochagavia:
Today I did another round of benchmarks using the changes from #1559, so I could simulate the link's delay (using `bench` instead of `perf`).

Command used: `cargo run --release --bin bulk -- --no-protection --simulate-network --simulated-link-delay [delay] --download-size 200M --stats`. Each run has its own value of `delay`, as you can see in the results below.

I ran the benchmarks on top of the main branch (equivalent to an ack-eliciting threshold of 0), and on top of this branch (using an ack-eliciting threshold of 1 and 10). In all cases, throughput remained similar, so I'm only reporting it a single time here, for each benchmarked link delay.

*I assume throughput is being constrained by pacing*. It would be interesting to tweak the config to start right away at a high transfer rate, because that would put real stress on Quinn and thereby reveal the true performance difference. I tried setting the initial congestion window to an absurdly high value, but somehow it didn't make any difference. Any clues on how to achieve higher throughput from the start? @Ralith, @djc

---

Delay = 5ms

```
      │  Throughput   │ Duration
 AVG  │   79.59 MiB/s │     2.51s
 P0   │   79.56 MiB/s │     2.51s
 P10  │   79.62 MiB/s │     2.51s
 P50  │   79.62 MiB/s │     2.51s
 P90  │   79.62 MiB/s │     2.51s
 P100 │   79.62 MiB/s │     2.51s
```

---

Delay = 10ms

```
      │  Throughput   │ Duration
 AVG  │   46.83 MiB/s │     4.27s
 P0   │   46.81 MiB/s │     4.27s
 P10  │   46.84 MiB/s │     4.27s
 P50  │   46.84 MiB/s │     4.27s
 P90  │   46.84 MiB/s │     4.27s
 P100 │   46.84 MiB/s │     4.27s
```

---

Delay = 50ms

```
      │  Throughput   │ Duration
 AVG  │   10.77 MiB/s │    18.57s
 P0   │   10.77 MiB/s │    18.56s
 P10  │   10.77 MiB/s │    18.58s
 P50  │   10.77 MiB/s │    18.58s
 P90  │   10.77 MiB/s │    18.58s
 P100 │   10.77 MiB/s │    18.58s
```

---

Delay = 200ms

```
      │  Throughput   │ Duration
 AVG  │    2.78 MiB/s │    72.03s
 P0   │    2.78 MiB/s │    72.00s
 P10  │    2.78 MiB/s │    72.06s
 P50  │    2.78 MiB/s │    72.06s
 P90  │    2.78 MiB/s │    72.06s
 P100 │    2.78 MiB/s │    72.06s
```

Ralith:
I concur that testing on a clean loopback is probably going to make it difficult to judge the impact of this change. 

> I assume throughput is being constrained by pacing

This should be easy to test by hacking `Pacer::delay` to always return `None`.

> do these results make any sense at all?

It's been a while since I've done manual WAN testing, but that seems much too slow. 

I think we should expect throughput equal to one congestion window per RTT, and the congestion window should grow reasonably quickly unless there's packet loss, which should be impossible in the simulated context. The bandwidth you report seems almost exactly inversely proportional to round trip time, which suggests the congestion window is not growing. I wonder if [`Connection::app_limited`](https://github.com/quinn-rs/quinn/blob/37e7b742cbd4cffc2f463358a6cb77149d3b78df/quinn-proto/src/connection/mod.rs#L210) is getting set spuriously? This is intended to prevent the congestion window from growing endlessly when sending less than the link's capacity. The logic is a bit fiddly (see e.g. https://github.com/quinn-rs/quinn/issues/1126).

It should be easy to verify whether the congestion window is growing by reporting `PathStats::cwnd` at intervals (or even just at start/end). A large initial congestion window really should have an obvious impact, though; even pacing is cwnd-based. When the initial window is set very high, and data is nonetheless not being sent, where does the `Connection::poll_transmit` call bail out?

aochagavia:
As it turns out, throughput was limited by the default send and receive windows. We were increasing the RTT, yet keeping the window sizes constant, so the same amount of traffic had to be spread over a longer period, leading to low throughput. The solution:

```rust
config.stream_receive_window(VarInt::MAX);
config.send_window(u64::MAX);
```

I re-benchmarked this with the config above, and now we are seeing proper performance. For instance, with delay = 5 and ack-eliciting threshold = 10:

```
      │  Throughput   │ Duration 
 AVG  │ 1160.82 MiB/s │  885.00ms
 P0   │  980.50 MiB/s │  808.00ms
 P10  │  981.00 MiB/s │  808.00ms
 P50  │ 1161.00 MiB/s │  876.00ms
 P90  │ 1250.00 MiB/s │  904.00ms
 P100 │ 1268.00 MiB/s │     1.04s
```

Unfortunately, testing with different ack frequencies and delays seemed to make no difference as regards to performance (on my system).

@Ralith @djc all this time I've been working under the assumption that this PR only makes sense if performance improves (otherwise it introduces complexity to the codebase for no benefit). Or do you think it is worthwhile to have this anyway, as long as we can show that performance doesn't worsen?

By the way, if you are interested I can prepare a branch you can use to easily benchmark locally, using the loopback interface and introducing delays with `tc` to ensure nothing fishy is going on with the simulated network code (unfortunately `tc` doesn't work on WSL, and trying it out in a cloud VM yielded weird results).

djc:
IIRC this draft was specifically introduced to improve performance, so if we can't reproduce that result there's likely some more digging to do. As such, I think as long as this doesn't introduce a (non-trivial) performance regression, I think it would still be okay to merge?

aochagavia:
Just rebased on top of latest main, to resolve a few conflicts. Any chance this PR can be reviewed soon? I don't think there is anything left to do on my side (please let me know otherwise).

Ralith:
Will get this in the next day or two; thanks for your patience!

quinn-proto/src/frame.rs
```diff
@@ -684,6 +691,19 @@ impl Iter {
                 token: self.take_len()?,
             },
             Type::HANDSHAKE_DONE => Frame::HandshakeDone,
+            Type::ACK_FREQUENCY => {
+                let sequence = self.bytes.get_var()?;
+                let ack_eliciting_threshold = self.bytes.get_var()?;
+                let request_max_ack_delay = self.bytes.get_var()?;
+                let reordering_threshold = self.bytes.get_var()?;
```
Ralith, line 698: nit: call these inline in the struct expression for concision? Intermediate variables don't seem necessary here.

quinn-proto/src/connection/mod.rs
```diff
@@ -2833,10 +2976,55 @@ impl Connection {
...
+            // Safe to unwrap because these are always provided when ACK frequency is enabled
+            let config = self.config.ack_frequency_config.as_ref().unwrap();
+            let min_ack_delay_micros = self.peer_params.min_ack_delay.unwrap().0;
+
+            // Ensure the delay is within bounds to avoid a PROTOCOL_VIOLATION error
+            let max_ack_delay = self.ack_frequency.candidate_max_ack_delay(config);
+            let max_ack_delay_micros = max_ack_delay
+                .as_micros()
+                .try_into()
+                .unwrap_or(u64::MAX)
```
Ralith, line 2989: `u64::MAX` is out of range for varints, so this will lead to a panic. Should we clamp to `VarInt::MAX.into_inner()`?

quinn-proto/src/frame.rs
```diff
@@ -871,9 +891,39 @@ impl Datagram {
...
+    pub(crate) request_max_ack_delay: u64,
+    pub(crate) reordering_threshold: u64,
+}
+
+impl AckFrequency {
+    pub(crate) fn encode<W: BufMut>(
+        sequence: u64,
+        ack_eliciting_threshold: u64,
+        request_max_ack_delay_micros: u64,
+        reordering_threshold: u64,
```
Ralith, line 907: Should these just be `&self`?

quinn-proto/src/frame.rs
```diff
@@ -871,9 +891,39 @@ impl Datagram {
     }
 }
 
+#[derive(Debug, Copy, Clone)]
+pub(crate) struct AckFrequency {
+    pub(crate) sequence: u64,
+    pub(crate) ack_eliciting_threshold: u64,
+    pub(crate) request_max_ack_delay: u64,
+    pub(crate) reordering_threshold: u64,
```
Ralith, line 899: Let's use `VarInt` rather than `u64` to reduce opportunity for error elsewhere. We're not totally consistent about that, but stronger types are generally preferred, making e.g. the `u64::MAX` bug above more obvious. This also allows `.write(x)` and `.get()` to be used directly.

quinn-proto/src/frame.rs
```diff
@@ -902,4 +952,30 @@ mod test {
...
+        let mut buf = Vec::new();
+        AckFrequency::encode(42, 20, 50_000, 1, &mut buf);
+        let frames = frames(buf);
+        assert_eq!(frames.len(), 1);
+        match &frames[0] {
+            Frame::AckFrequency(ack_frequency) => {
+                assert_eq!(ack_frequency.sequence, 42);
+                assert_eq!(ack_frequency.ack_eliciting_threshold, 20);
+                assert_eq!(ack_frequency.request_max_ack_delay, 50_000);
+                assert_eq!(ack_frequency.reordering_threshold, 1);
```
Ralith, line 967: If we `derive(Eq, PartialEq)` for the frame struct, and take `&self` in `encode`, this could be more concise.

quinn-proto/src/tests/mod.rs
```diff
@@ -1914,6 +1932,32 @@ fn malformed_token_len() {
...
+
+    let stats_after_connect = pair.client_conn_mut(client_ch).stats();
+
+    // Lose a ping
+    let default_mtu = mem::replace(&mut pair.mtu, 0);
+    pair.client_conn_mut(client_ch).ping();
+    pair.drive_client();
+    pair.mtu = default_mtu;
+
+    // Drive the connection further so a loss probe is sent and the ping is recovered
```
Ralith, line 1951: Pings actually never get retransmitted (see https://www.rfc-editor.org/rfc/rfc9000.html#name-retransmission-of-informati), though it doesn't seem like that's important to this test in practice.

quinn-proto/src/tests/mod.rs
```diff
@@ -2070,39 +2114,6 @@ fn connect_runs_mtud_again_after_600_seconds() {
     assert_eq!(pair.server_conn_mut(server_ch).path_mtu(), 1452);
 }
 
-#[test]
-fn packet_loss_and_retry_too_low_mtu() {
```
Ralith, line 2074: Why is this being removed?

quinn-proto/src/tests/mod.rs
```diff
@@ -2184,6 +2211,388 @@ fn packet_splitting_not_necessary_after_higher_mtu_discovered() {
...
+
+    pair.drive();
+    let stats_after_drive = pair.client_conn_mut(client_ch).stats();
+    assert_eq!(
+        stats_after_drive.frame_rx.acks - stats_after_ping.frame_rx.acks,
+        1
+    );
+
+    // The time is start + max_ack_delay
+    assert_eq!(pair.time, start + Duration::from_millis(25));
```
Ralith, line 2252: Can we avoid an implicit dependency on the default max ack delay value here?

quinn-proto/src/tests/mod.rs
```diff
@@ -2184,6 +2211,388 @@ fn packet_splitting_not_necessary_after_higher_mtu_discovered() {
...
+    let _guard = subscribe();
+    let mut pair = Pair::default();
+    let (client_ch, _) = pair.connect();
+    pair.drive();
+
+    let acks_after_connect = pair.client_conn_mut(client_ch).stats().frame_rx.acks;
+
+    pair.client_conn_mut(client_ch).immediate_ack();
+    pair.drive_client(); // Send ping
+    pair.drive_server(); // Process ping
```
Ralith, line 2285: There's no actual ping, right? Reword comments?

quinn-proto/src/tests/mod.rs
```diff
@@ -2184,6 +2211,388 @@ fn packet_splitting_not_necessary_after_higher_mtu_discovered() {
...
+    pair.drive_client(); // Give the client a chance to process the ack
+
+    let acks_after_ping = pair.client_conn_mut(client_ch).stats().frame_rx.acks;
+
+    assert_eq!(acks_after_ping - acks_after_connect, 1);
```
Ralith, line 2290: Shouldn't we be checking that the *server* sent an immediate ACK, upon receiving the request from the client? Why is the client sending an ACK at all in this period?

quinn-proto/src/tests/util.rs
```diff
@@ -515,6 +520,16 @@ fn packet_size(transmit: &Transmit) -> usize {
     transmit.contents.len()
 }
 
+fn ecn(x: Option<EcnCodepoint>, congestion_experienced: bool) -> Option<EcnCodepoint> {
+    x.map(|codepoint| {
+        if congestion_experienced {
+            EcnCodepoint::from_bits(0b11).unwrap()
```
Ralith, line 535: `EcnCodepoint::Ce`? Those variants are only hidden for the benefit of the public API.

quinn-proto/src/tests/util.rs
```diff
@@ -333,17 +338,17 @@ impl TestEndpoint {
...
                     for event in events.drain(..) {
                         conn.handle_event(event);
                     }
                 }
 
+                if self.timeout.map_or(false, |x| x <= now) {
+                    self.timeout = None;
+                    conn.handle_timeout(now);
+                }
+
```
Ralith, line 360: What's the significance of relocating this?

quinn-proto/src/transport_parameters.rs
```diff
@@ -418,6 +434,10 @@ impl TransportParameters {
             || params.max_udp_payload_size.0 < 1200
             || params.initial_max_streams_bidi.0 > MAX_STREAM_COUNT
             || params.initial_max_streams_uni.0 > MAX_STREAM_COUNT
+            || params.min_ack_delay.map_or(false, |min_ack_delay| {
+                // min_ack_delay uses nanoseconds, whereas max_ack_delay uses milliseconds
+                min_ack_delay.0 > params.max_ack_delay.0 * 1_000_000
```
Ralith, line 439: Isn't it microseconds, not nanoseconds?

Ralith (changes requested):
Went over everything carefully, save for a bit of skimming in the `ack_frequency` unit test. Overall this looks great, thank you!

quinn-proto/src/connection/mod.rs
```diff
@@ -2436,6 +2525,7 @@ impl Connection {
         number: u64,
         payload: Bytes,
     ) -> Result<(), TransportError> {
+        self.reset_rtt_timer(now);
```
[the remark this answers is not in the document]
aochagavia: The function name is misleading here... It is actually setting up the RTT timer for the first time. Do you think that belongs to a different place in the code? Or would it be enough to rename it to something like `setup_rtt_timer`?

quinn-proto/src/connection/mod.rs
```diff
@@ -3220,6 +3418,14 @@ impl Connection {
             .push_back(EndpointEventInner::NeedIdentifiers(now, n));
     }
 
+    /// Send an IMMEDIATE_ACK frame to the remote endpoint
+    ///
+    /// According to the spec, this will result in an error if the remote endpoint does not support
+    /// the Acknowledgement Frequency extension
+    pub fn immediate_ack(&mut self) {
```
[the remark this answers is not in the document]
aochagavia: Good catch, I'll make it `pub(crate)`

quinn-proto/src/connection/spaces.rs
```diff
@@ -233,6 +246,7 @@ pub struct Retransmits {
     pub(super) crypto: VecDeque<frame::Crypto>,
     pub(super) new_cids: Vec<IssuedCid>,
     pub(super) retire_cids: Vec<u64>,
+    pub(super) ack_frequency: Option<u64>,
```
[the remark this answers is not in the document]
aochagavia: I think this should be ok, as it is explicitly contemplated in the draft (the comment in brackets is mine):

> When an ACK_FREQUENCY frame is lost, the sender is encouraged to send another ACK_FREQUENCY frame [instead of resending the original frame].

I'll update the code

quinn-proto/src/connection/mod.rs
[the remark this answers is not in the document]
Ralith: Since the last relevant signal considered is whether the peer has declared support for the extension in its transport parameters, how about we do this when we process the peer's 1-RTT transport parameters, in `Connection::handle_peer_params`? Probably along with renaming it for clarity.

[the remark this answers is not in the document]
aochagavia: Will do, `handle_peer_params` definitely sounds like the right place

quinn-proto/src/connection/spaces.rs
```diff
@@ -414,51 +489,183 @@ impl SendableFrames {
...
+    ///
+    /// * `0`: no special action is taken
+    /// * `1`: an ACK is immediately sent if it is out-of-order according to RFC 9000
+    /// * `>1`: an ACK is immediately sent if it is out-of-order according to the ACK frequency draft
+    reordering_threshold: u64,
+    /// The earliest ack-eliciting packet since the last ACK was sent, used to calculate the moment
+    /// upon which `max_ack_delay` elapses
+    earliest_ack_eliciting_since_last_ack_sent: Option<Instant>,
+    /// The packet number ranges of ack-eliciting packets yet to be ACKed
+    ranges: ArrayRangeSet,
```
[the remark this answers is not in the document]
aochagavia: True, do you think it is worthwhile to reword the comment to make that explicit?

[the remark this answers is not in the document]
Ralith: Yes, I think this is dangerously misleading as-is. Perhaps "ACK-eliciting packets the peer hasn't confirmed receipt of ACKs for"?

quinn-proto/src/tests/util.rs
djc: (And relocating code should go in a separate commit.)

quinn-proto/src/tests/mod.rs
aochagavia: Thanks for the clarification, I have updated the comment

quinn-proto/src/tests/mod.rs
aochagavia: This test is mostly a sanity check I added when I was figuring out how everything worked. Now we have many tests that trigger packet loss and recovery, I deemed it unnecessary to keep it. I'll happily resurrect it, though, if you think it's worthwhile.

quinn-proto/src/tests/mod.rs
aochagavia: Hmm... The ideal thing here would be to configure the endpoint's max ack delay and assert against the provided value. However, it is currently not possible to specify a custom max ack delay, so the default is always used. To keep things simple and explicit, I'll assert against `TransportParameters::default().ack_delay`.

quinn-proto/src/tests/mod.rs
aochagavia: As far as I can see, the assertion is indeed checking that the server sent an immediate ACK and it was processed by the client (we are checking the `frame_rx` stats of the client).

quinn-proto/src/transport_parameters.rs
aochagavia: Good catch! Not sure how I got that one wrong... The draft is pretty clear about it:

> Note that while the endpoint's max_ack_delay transport parameter is in milliseconds ([Section 18.2](https://rfc-editor.org/rfc/rfc9000#section-18.2) of [[QUIC-TRANSPORT](https://www.rfc-editor.org/rfc/rfc9000)]), min_ack_delay is specified in microseconds

quinn-proto/src/tests/util.rs
aochagavia: I don't exactly remember why this was necessary, but I just reverted it and observed that at least one test triggers an infinite loop (probably related to the timers that are being added in this PR). I'll extract the change into its own commit.

aochagavia:
Thanks for the thorough review! Just rebased on top of latest main and pushed a new version that incorporates all suggestions (unless anything slipped through the cracks).

quinn-proto/src/tests/util.rs
```diff
@@ -515,6 +520,16 @@ fn packet_size(transmit: &Transmit) -> usize {
     transmit.contents.len()
 }
 
+fn ecn(x: Option<EcnCodepoint>, congestion_experienced: bool) -> Option<EcnCodepoint> {
+    x.map(|codepoint| {
+        if congestion_experienced {
```
djc, line 534: Style: would use `match` instead of `if` here.

Maybe the function name could be a bit more descriptive? `ecn_or()` or `ecn_and_then()` or something else...

quinn-proto/src/connection/mod.rs
```diff
@@ -2015,40 +2018,13 @@ impl Connection {
...
-            && self.peer_params.stateless_reset_token.as_deref()
-                == Some(&packet[packet.len() - RESET_TOKEN_SIZE..]);
-
-        match partial_decode.finish(header_crypto) {
-            Ok(packet) => self.handle_packet(now, remote, ecn, Some(packet), stateless_reset),
-            Err(_) if stateless_reset => self.handle_packet(now, remote, ecn, None, true),
-            Err(e) => {
-                trace!("unable to complete packet decoding: {}", e);
-            }
+        if let Some(decoded) = packet_crypto::unprotect_header(
```
djc, line 2021: Oh, I really want this as one separate commit with just the extraction of this code into a separate method.

quinn-proto/src/connection/mod.rs
```diff
@@ -3258,78 +3234,33 @@ impl Connection {
...
-            &self.next_crypto.as_ref().unwrap().remote
-        };
-
-        crypto
-            .decrypt(number, &packet.header_data, &mut packet.payload)
-            .map_err(|_| {
-                trace!("decryption failed with packet number {}", number);
-                None
-            })?;
+        let result = packet_crypto::decrypt_packet_body(
```
djc, line 3237: IMO this should also be its own separate commit.

quinn-proto/src/config.rs
```diff
@@ -365,6 +378,72 @@ impl fmt::Debug for TransportConfig {
...
+    /// transport parameter.
+    ///
+    /// Defaults to `None`, in which case the peer's original `max_ack_delay` will be used.
+    pub fn max_ack_delay(&mut self, value: Option<Duration>) -> &mut Self {
+        self.max_ack_delay = value;
+        self
+    }
+
+    /// The amount of out-of-order packets that will trigger a peer to send an ACK, without waiting
+    /// for `ack_eliciting_threshold` to be exceeded or for `max_ack_delay` to be elapsed.
```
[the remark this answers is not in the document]
aochagavia: > Clearly we can't configure the peer to do anything.

The way ACK frequency works in the current draft is that you actually "configure" the peer, telling it which parameters it should use to regulate sending of ACKs. Specifically, we tell the peer which `max_ack_delay`, `ack_eliciting_threshold`, and `reordering_threshold` to use.

The doc comment to the `AckFrequencyConfig` offers valuable context on how this works:

```rust
/// Parameters for controlling the peer's acknowledgement frequency
///
/// The parameters provided in this config will be sent to the peer at the beginni
/// connection, so it can take them into account when sending acknowledgements (se
/// description for details on how it influences acknowledgement frequency).
```

But maybe you are already aware of all this, and it's just the wording that feels odd? IMO with the above comment and the fact that a compliant peer is supposed to obey the `ACK_FREQUENCY` frame, I think it's OK to imply that we are configuring the peer.

quinn-proto/src/connection/mod.rs
```diff
@@ -190,6 +193,13 @@ pub struct Connection {
     path_response: Option<PathResponse>,
     close: bool,
 
+    //
+    // ACK frequency
+    //
+    ack_frequency: AckFrequencyState,
+    last_ack_frequency_frame: Option<u64>,
+    max_ack_delay: Duration,
```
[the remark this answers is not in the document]
aochagavia: Good point! `AckFrequencyState` contains the necessary state related to _sending_ `ACK_FREQUENCY` frames (e.g. is the frame still in-flight, what is the currently known `max_ack_delay` of the peer, etc). Next to that, we have some state in the `Connection` related to _receiving_ `ACK_FREQUENCY` frames (what is the sequence number of the last received frame? What is our `max_ack_delay`?). I thought it worthwhile to keep the separation of concerns here, but it felt like overkill to introduce a new struct with just two fields for the remaining state.

I can think of the following alternatives:

1. Include the explanation above in the form of comments next to the fields, and a doc comment to `AckFrequencyState`
2. Introduce a new struct, in `ack_frequency.rs`, containing the `last_ack_frequency_frame` and `max_ack_delay` fields. Not sure yet how to name it, though.
3. Bring everything together under `AckFrequencyState`, relying on comments to explain the difference

I'd go with option 1 or 3. Do you have any preference?

quinn-proto/src/connection/mod.rs
```diff
@@ -1683,6 +1757,12 @@ impl Connection {
         self.timers.set(Timer::KeepAlive, now + interval);
     }
 
+    fn setup_rtt_timer(&mut self, now: Instant) {
```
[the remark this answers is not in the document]
aochagavia: Sure, will do!

quinn-proto/src/connection/mod.rs
```diff
@@ -2523,7 +2608,11 @@ impl Connection {
                     if remote == self.path.remote {
                         // PATH_CHALLENGE on active path, possible off-path packet forwarding
                         // attack. Send a non-probing packet to recover the active path.
-                        self.ping();
+                        if self.peer_supports_ack_frequency() {
```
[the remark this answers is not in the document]
aochagavia: Do you mean something like this? I like it

```rust
match self.peer_supports_ack_frequency() {
    true => self.immediate_ack(),
    false => self.ping(),
}
```

quinn-proto/src/connection/mod.rs
```diff
@@ -3161,6 +3357,18 @@ impl Connection {
...
 
+    fn peer_supports_ack_frequency(&self) -> bool {
+        self.peer_params.min_ack_delay.is_some()
+    }
+
+    /// Send an IMMEDIATE_ACK frame to the remote endpoint
+    ///
+    /// According to the spec, this will result in an error if the remote endpoint does not support
+    /// the Acknowledgement Frequency extension
+    pub(crate) fn immediate_ack(&mut self) {
```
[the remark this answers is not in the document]
aochagavia: I'd say in this case the existence of the method is justified because it makes the code a bit more self-documenting: having a function called `immediate_ack` (plus docs explaining what it does) makes the code in my eyes easier to follow than writing `self.spaces[self.highest_space].immediate_ack_pending = true;`.

I can inline it, though, if you feel it really improves things.

quinn-proto/src/tests/util.rs
aochagavia: The method is setting the congestion experienced bit if ECN is supported and `congestion_experienced == true`. I agree that `ecn` is too generic. I'll use `set_congestion_experienced`, which is more descriptive.

quinn-proto/src/connection/mod.rs
aochagavia: Just to be sure, you'd like to split https://github.com/quinn-rs/quinn/pull/1553/commits/1bf73f9d76542dd6095d04f3bf35260f4dc51108 into three commits?

1. Extracting `unprotect_header`
2. Extracting `decrypt_packet_body`
3. Adding the tests that rely on these methods

aochagavia:
Just rebased and pushed changes to address some of your suggestions. For those that need more clarification I left some comments and questions

quinn-proto/src/connection/mod.rs
djc: Yes, please.

quinn-proto/src/connection/mod.rs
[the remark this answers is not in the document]
djc: Yup.

quinn-proto/src/connection/mod.rs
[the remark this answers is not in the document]
djc: Option 3 sounds good to me!

quinn-proto/src/config.rs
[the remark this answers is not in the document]
djc: Right -- it's mostly the wording that I'm objecting to here. I'd like us to be cognizant of the fact that we can request the peer to do things and we can expect the peer to do things a certain way, but "will trigger" seems too strong to me -- to some extent the peer is unreliably and we have to do the right thing even if it doesn't.

> Regarding style, is it just the first line that shouldn't end with a "."?

The idea is that the first line is usually a sentence fragment rather than a sentence (a sentence without a subject, a verb, or it doesn't include a "complete thought"). Because the rest of the docstring is not subject to the same conciseness constraints, they can usually be more fully formed.

quinn-proto/src/connection/mod.rs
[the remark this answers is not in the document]
djc: I mean, you can still add comments, right? I think putting this line in its own code block (surrounded by empty lines) with an explanatory comment makes the code just as well-documented, but still easier to follow due to less indirection.

I don't feel very strongly about it though, if you don't think it makes sense.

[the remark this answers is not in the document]
aochagavia: I was about to inline the method, but then I realized it is also being used in tests, so IMO it makes sense to keep it after all

quinn-proto/src/connection/packet_crypto.rs
```diff
@@ -0,0 +1,70 @@
+use tracing::{debug, trace};
+
+use crate::connection::spaces::PacketSpace;
+use crate::crypto::{HeaderKey, PacketKey};
+use crate::packet::{Packet, PartialDecode};
+use crate::token::ResetToken;
+use crate::RESET_TOKEN_SIZE;
+
+pub(super) struct UnprotectHeaderResult {
```
djc, line 10: Style nit: this should go below `unprotect_header()`.

quinn-proto/src/connection/mod.rs
```diff
@@ -3230,78 +3230,33 @@ impl Connection {
...
+            self.prev_crypto.as_ref(),
+            self.next_crypto.as_ref(),
+        )?;
 
-        if let Some(ref mut prev) = self.prev_crypto {
-            if prev.end_packet.is_none() && key_phase == self.key_phase {
-                // Outgoing key update newly acknowledged
-                prev.end_packet = Some((number, now));
-                self.set_key_discard_timer(now, space);
+        let pn = result.map(|d| {
```
djc, line 3242: Style: I'd prefer this as a `match` expression with an early return, so we don't have to nest (and indent) all this code into a closure. Also not a big fan of the `d` binding here.

quinn-proto/src/connection/mod.rs
```diff
@@ -3296,6 +3296,43 @@ impl Connection {
...
     }
 
+    /// Decodes a packet, returning its decrypted payload, so it can be inspected in tests
+    #[cfg(test)]
+    pub(crate) fn decode_packet(&self, event: &ConnectionEvent) -> Option<Vec<u8>> {
+        if let ConnectionEventInner::Datagram {
+            first_decode,
+            remaining,
+            ..
+        } = &event.0
```
djc, line 3306: Let's do a `match` with early return here?

quinn-proto/src/packet.rs
```diff
@@ -19,7 +19,7 @@ use crate::{
 // to inspect the version and packet type (which depends on the version).
 // This information allows us to fully decode and decrypt the packet.
 #[allow(unreachable_pub)] // fuzzing only
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
djc, line 22: Where/why do we need these `Clone` derives?

aochagavia: These are necessary for the tests, in this line: https://github.com/quinn-rs/quinn/blob/8244dedb0b5b1fda8ca3f9869fd299b6aac6cfdc/quinn-proto/src/connection/mod.rs#L3313

djc: Ah... How do you feel about doing a `cfg_attr()` thing here?

aochagavia: Sounds reasonable, will add it

quinn-proto/src/connection/packet_crypto.rs
aochagavia: The idea is to put it after the `fn`, right? I'll could do the same with `DecryptPacketResult`

djc: Yup.

quinn-proto/src/config.rs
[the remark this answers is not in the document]
aochagavia: Thanks, I have reworded the comments and made sure the first lines are shorter

aochagavia:
Thanks for the review! Just rebased an pushed fixes for all points we discussed 😉 

Almost there!

quinn-proto/src/tests/mod.rs
Ralith: Must've misread, whoops.

quinn-proto/src/tests/mod.rs
Ralith: Makes sense; I'm happy to leave it out.

quinn-proto/src/tests/mod.rs
Ralith: nit: Would be nice to have the change in this commit rather than the later one. NBD if the rebase is too much effort though.

quinn-proto/src/connection/ack_frequency.rs
```diff
@@ -0,0 +1,125 @@
...
+            return Ok(false);
+        }
+
+        self.last_ack_frequency_frame = Some(frame.sequence.into_inner());
+
+        // Update max_ack_delay
+        let max_ack_delay = Duration::from_micros(frame.request_max_ack_delay.into_inner());
+        if max_ack_delay < TIMER_GRANULARITY {
+            return Err(TransportError::PROTOCOL_VIOLATION(
+                "Request Max Ack Delay in ACK_FREQUENCY frame is less than min_ack_delay",
```
Ralith, line 114: Grammar nit: "requested"

quinn-proto/src/connection/ack_frequency.rs
```diff
@@ -0,0 +1,125 @@
...
+            ));
+        }
+        self.max_ack_delay = max_ack_delay;
+
+        // Update the rest of the params
+        pending_acks.set_reordering_threshold(frame.reordering_threshold.into_inner());
+        pending_acks.set_ack_eliciting_threshold(frame.ack_eliciting_threshold.into_inner());
```
Ralith, line 121: nit: Might be a bit tidier to make these a single setter that takes the frame?

quinn-proto/src/connection/ack_frequency.rs
```diff
@@ -0,0 +1,67 @@
+use crate::{AckFrequencyConfig, VarInt};
```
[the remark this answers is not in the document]
Ralith: Strongly agree that anywhere other than `connection/mod.rs` is appealing, and a standalone module seems fine to me.

quinn-proto/src/tests/mod.rs
```diff
@@ -2181,6 +2208,388 @@ fn packet_splitting_not_necessary_after_higher_mtu_discovered() {
...
+    let mut client_config = client_config();
+    let mut ack_freq_config = AckFrequencyConfig::default();
+    ack_freq_config.ack_eliciting_threshold(10u32.into());
+    Arc::get_mut(&mut client_config.transport)
+        .unwrap()
+        .ack_frequency_config(Some(ack_freq_config))
+        .mtu_discovery_config(None); // To keep traffic cleaner
+
+    let mut pair = Pair::default();
+    pair.latency = Duration::from_millis(10); // Need latency to avoid an RTT = 0
```
Ralith, line 2407: Why is RTT=0 a problem here? RTT=0 (to within timer precision) can happen in reality.

quinn-proto/src/tests/mod.rs
```diff
@@ -2181,6 +2208,388 @@ fn packet_splitting_not_necessary_after_higher_mtu_discovered() {
...
+        server_stats_after.frame_rx.ping - server_stats_before.frame_rx.ping,
+        3
+    );
+    assert_eq!(
+        server_stats_after.frame_tx.acks - server_stats_before.frame_tx.acks,
+        0
+    );
+
+    let server_stats_before = pair.server_conn_mut(server_ch).stats();
+    pair.time += Duration::from_millis(30); // > max_ack_delay
```
Ralith, line 2487: Can we derive this value from the default config rather than hardcoding it?

quinn-proto/src/tests/mod.rs
```diff
@@ -2181,6 +2208,388 @@ fn packet_splitting_not_necessary_after_higher_mtu_discovered() {
...
+        1
+    );
+    assert_eq!(
+        stats_after_ping.path.congestion_events - stats_after_connect.path.congestion_events,
+        1
+    );
+}
+
+#[test]
+fn ack_frequency() {
```
Ralith, line 2391: This test seems to cover a bunch of independent cases. Can we split them up, maybe with a helper function for common setup?

Ralith (changes requested):
This is looking really good; thanks for your persistence in driving it forward! Nits aside, just a few things left. I'll endeavor to be more responsive in the coming days, and if I don't manage that then I'll address any remaining items myself.

It didn't occur to me at first, but this also fixes #438 independently of support for the extension, which is a nice win.

quinn-proto/src/connection/mod.rs
```diff
@@ -1004,6 +1043,29 @@ impl Connection {
...
+                        in_flight_ack_eliciting = space.sent_packets.len(),
+                        "rtt timer"
+                    );
+
+                    if !space.sent_packets.is_empty() {
+                        // It is recommended to request an immediate ACK once per RTT if there are
+                        // unacked packets
+                        space.immediate_ack_pending = true;
+                    }
+                    self.timers.set(Timer::Rtt, now + self.path.rtt.get());
```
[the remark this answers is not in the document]
aochagavia: Somehow I can't reply inline to [this comment](https://github.com/quinn-rs/quinn/pull/1553#discussion_r1249004850), so I'll do it here, as both are related.

First of all, I'm glad you found the root cause of the infinite loop. Sorry I missed that. It'd been better if I'd persevered looking for the problem, instead of tweaking the testing code.

Regarding a solution, the simplest one I can come up with is to schedule the next timeout after `max(RTT, TIMER_GRANULARITY)`. That way we can sidestep additional complexity, and to me it seems like a sensible choice This code is executed right after the RTT timeout has elapsed and it doesn't make sense to schedule the new RTT timeout before the next tick, which I assume happens after `TIMER_GRANULARITY`. Do you agree?

Otherwise, setting the timeout when sending an ack-eliciting packet would also work, but requires some bookkeeping (i.e. tracking the `Instant` of the last ACK sent, so we can check whether an RTT has elapsed since).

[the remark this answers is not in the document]
Ralith: > I can't reply inline

The input box for that can be found in the original thread. Threading interacts kind of weirdly with review-scoped comments.

> schedule the next timeout after max(RTT, TIMER_GRANULARITY).

I'm not sure how this solves the problem of the infinite loop, or of spurious wakeups when there's no activity. In the case of the looping test, RTT is 10ms, so replacing `now + rtt` with `now + max(rtt, TIMER_GRANULARITY)` won't change behavior at all. Did you mean to propose `now + rtt + TIMER_GRANULARITY`? That seems like an improvement, but latency variance we'll still likely see some superfluous transmits, which is suboptimal.

> setting the timeout when sending an ack-eliciting packet [...] requires some bookkeeping

Can't we just set the timer to `now + rtt` if it's unset, and leave it alone if it's already set? Did you mean to say tracking the `Instant` of the last ACK *received*, so we don't request redundant ACKs from the peer? That sounds like an improvement, but is independent of the two separate issues identified, which are:
1. IMMEDIATE_ACKs count as in-flight for purposes of sending another IMMEDIATE_ACK, causing spurious IMMEDIATE_ACK sends if the RTT timer expiry is handled before the peer's ACK is processed, which is likely to happen as the peer's ACK will take about an RTT to arrive.
2. The RTT timer is always active, even when the connection is idle, which causes unnecessary wakeups.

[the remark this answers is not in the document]
Ralith: > Sorry I missed that.

And no worries there, sometimes an outside perspective makes it a lot easier to diagnose stuff :grinning:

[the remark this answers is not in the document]
aochagavia: Thanks for summarizing the problem again. It's helps load back part of the context in my head, after 2 months of working on other stuff 😅. Reading back what I wrote above, I'm not that sure it makes much sense. Now let me try to articulate my thoughts about both problems you mentioned.

### RTT is always active

The current implementation is too naive, waking the connection every RTT even if there is no work to do. Your suggested alternative sounds compelling. Here's how it would work in practice:

1. By default the timer should be inactive
2. When sending an ack-eliciting packet and the RTT timer is not set, we should set the timer to `now + RTT` _and_ request an immediate ack (otherwise there is a chance we won't get an ACK within RTT, and the draft says that we should _cause a receiver to send an acknowledgement at least once per RTT if there are unacknowledged ack-eliciting packets in flight_)
3. When the timer is elapsed _and_ there are ack-eliciting packets in-flight, we should set the timer to `now + RTT` _and_ request an immediate ack

Does that sound reasonable?

### Timeout-triggered immediate ack frames can cause an infinite loop

I'll reply inline in the thread 😉

quinn-proto/src/tests/mod.rs
aochagavia: I don't exactly remember the specific reason to avoid an RTT = 0 here. Intuitively, I'd think having an RTT > 0 makes it more straightforward to check that the RTT timeout is working properly in the different parts of the test.

Regarding a minimum value, we could modify the timer interval to be at least `TIMER_GRANULARITY`, since that is in theory the minimum possible interval and using a lower interval than that would have no effect. Not sure if it's worth it to do this, though, since it should have no observable effect.

quinn-proto/src/tests/util.rs
aochagavia: Assuming we tweak the timer as mentioned [here](https://github.com/quinn-rs/quinn/pull/1553#discussion_r1251222791), I'm afraid requiring two unacked packets is too relaxed. Think of the following scenario:

1. We send a single ack-eliciting packet after a period of inactivity. Because the RTT timer is unset, the packet automatically gets an `IMMEDIATE_ACK` frame and the RTT timer is set.
2. We receive an ACK before the RTT timer elapses.
3. We send a new packet. Because the RTT timer is set, the packet does not get an `IMMEDIATE_ACK` frame, and the RTT timer isn't touched.
4. The RTT timer elapses, and sees that there is only one packet in flight. The timer is unset and no `IMMEDIATE_ACK` frame is sent, _which is not according to spec_.

Maybe we should track timeout-triggered `IMMEDIATE_ACK` frames after all? Not sure about the best way to do it, but maybe we could modify `PacketSpace::immediate_ack_pending` to be of type `Option<ImmediateAckSource>`:

```rust
enum ImmediateAckSource {
    Timeout,
    Other,
}
```

That way, when sending the packet, we can record its number if the source of the immediate ack was a timeout.

quinn-proto/src/tests/mod.rs
Ralith: Fair point. On review, I think this tests needs RTT≠0 just so that we can reason about the progression of time within it.

quinn-proto/src/connection/mod.rs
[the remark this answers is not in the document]
Ralith: > Does that sound reasonable?

Sounds perfect!

quinn-proto/src/tests/util.rs
Ralith: Note also that tail loss probes will already be sent by the existing loss detection logic if we don't get an ACK within a constant factor of an RTT for a trailing flight.

aochagavia:
Thanks again for the review! Just rebased and pushed a new version with all your comments addressed.

Ralith (approved):
LGTM, thank you for seeing this through!

djc:
Sorry for taking so long to review this, thanks for all the work!

aochagavia:
No worries, thank _you_ and @Ralith for creating and maintaining Quinn!

Issue #1556, opened 2023-05-06
CI: FreeBSD run emits errors but does not mark Action as `failed`


Tirka:

Tirka:
Here is the possible explanation:

<details>
<summary>Running build, test, fuzz and checking process exit codes</summary>

```
➜  quinn git:(main) ✗ cargo build
   Compiling quinn-udp v0.4.0 (/usr/home/maksimv/quinn/quinn-udp)
error[E0425]: cannot find value `SYS_sendmmsg` in crate `libc`
    --> quinn-udp/src/unix.rs:325:35
     |
325  |       let ret = libc::syscall(libc::SYS_sendmmsg, sockfd, msgvec, vlen, flags) as libc::c_int;
     |                                     ^^^^^^^^^^^^ help: a function with a similar name exists: `sendmmsg`
     |
    ::: /home/maksimv/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.142/src/unix/bsd/freebsdlike/freebsd/mod.rs:5162:5
     |
5162 | /     pub fn sendmmsg(
➜  quinn git:(main) ✗ cargo build
   Compiling quinn-udp v0.4.0 (/usr/home/maksimv/quinn/quinn-udp)
error[E0425]: cannot find value `SYS_sendmmsg` in crate `libc`
    --> quinn-udp/src/unix.rs:325:35
     |
325  |       let ret = libc::syscall(libc::SYS_sendmmsg, sockfd, msgvec, vlen, flags) as libc::c_int;
     |                                     ^^^^^^^^^^^^ help: a function with a similar name exists: `sendmmsg`
     |
    ::: /home/maksimv/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.142/src/unix/bsd/freebsdlike/freebsd/mod.rs:5162:5
     |
5162 | /     pub fn sendmmsg(
5163 | |         sockfd: ::c_int,
5164 | |         msgvec: *mut ::mmsghdr,
➜  quinn git:(main) ✗ cargo build --all-targets
   Compiling quinn-udp v0.4.0 (/usr/home/maksimv/quinn/quinn-udp)
error[E0425]: cannot find value `SYS_sendmmsg` in crate `libc`
    --> quinn-udp/src/unix.rs:325:35
     |
325  |       let ret = libc::syscall(libc::SYS_sendmmsg, sockfd, msgvec, vlen, flags) as libc::c_int;
[...]

PR #1557, opened 2023-05-06
Fail FreeBSD CI if any step fails

Tirka:
FreeBSD CI Action gives false success results. Here is possible explanation: [#1556](https://github.com/quinn-rs/quinn/issues/1556)

Removing fuzz tests is not ideal, but better than getting false success CI check results on every run


Ralith:
Maybe combine all the commands with `&&` instead, to explicitly propagate failures?

Tirka:
> Maybe combine all the commands with `&&` instead, to explicitly propagate failures?

Done. Looks good to me

Ralith (approved):
Thanks!

PR #1558, opened 2023-05-08
Upgrade socket2 to 0.5

djc:
This should be okay now that we've bumped our MSRV?

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -34,16 +34,12 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        rust: [stable, beta, 1.63.0]
+        rust: [stable, beta]
         exclude:
           - os: macos-latest
             rust: beta
-          - os: macos-latest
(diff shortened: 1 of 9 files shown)


djc:
The CI failure here is confusing. Tokio's support for `AsFd` was added in https://github.com/quinn-rs/quinn/actions/runs/4924706904/jobs/8798057112?pr=1558 (released in 1.27), and we already have 1.63 as our MSRV.

.github/workflows/rust.yml
```diff
@@ -77,7 +77,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: check
-          args: --lib --all-features -p quinn-udp -p quinn-proto -p quinn
```
Ralith: Stray commit?

Ralith (changes requested):
Great improvements overall!

djc:
Bumped the tokio dependency in quinn to 1.28.1 which should make the MSRV test pass.

Ralith (approved):

PR #1559, opened 2023-05-08
Overhaul bench

aochagavia:
This PR introduces a few options to benchmark more effectively:

- Disable encryption (using the `--no-protection` flag, similar to the `perf` crate).
- Run on a simulated in-memory network, instead of going through the networking stack of the operating system (using the `--simulate-network` together with the `--simulated-link-latency` and `--simulated-link-capacity` options). This makes it trivial to benchmark for different latencies and link capacities out-of-the-box (instead of relying on platform-specific utilities like `tc`).
- Disable MTUD (using `--disable-mtud`) and set the minimum MTU (using `--min-mtu`), to make runs more deterministic

### DIFF
--- a/bench/src/lib.rs
+++ b/bench/src/lib.rs
@@ -169,6 +169,10 @@ pub fn transport_config(opt: &Opt) -> quinn::TransportConfig {
     let mut config = quinn::TransportConfig::default();
     config.max_concurrent_uni_streams(opt.max_streams.try_into().unwrap());
     config.initial_mtu(opt.initial_mtu);
+    config.min_mtu(opt.min_mtu);
+    if opt.disable_mtud {
+        config.mtu_discovery_config(None);
+    }
     config
 }
 
@@ -250,6 +254,12 @@ pub struct Opt {
     /// Starting guess for maximum UDP payload size
     #[clap(long, default_value = "1200")]
     pub initial_mtu: u16,
+    /// The MTU that is guaranteed to be supported by the network
+    #[clap(long, default_value = "1200")]
+    min_mtu: u16,
+    /// Disable MTU discovery, to make the output more predictable
+    #[clap(long)]
+    disable_mtud: bool,
     /// Disable packet encryption/decryption
     #[clap(long)]
     no_protection: bool,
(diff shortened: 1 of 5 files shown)


bench/src/lib.rs
```diff
@@ -221,6 +253,18 @@ pub struct Opt {
     /// Disable packet encryption/decryption
     #[clap(long)]
     no_protection: bool,
+    /// Simulate network in-memory, instead of using the network stack
+    #[clap(long)]
+    simulate_network: bool,
+    /// Simulated link delay (one way), in milliseconds
+    #[clap(long, default_value = "0")]
```
Ralith, line 270: `requires("simulate_network")` here and below? I think that's the syntax, at least.

bench/src/simulated_network.rs
```diff
@@ -0,0 +1,250 @@
...
+        Ok(self.addr)
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct InMemorySocket {
+    addr: SocketAddr,
+    inbound: Arc<Mutex<InboundQueue>>,
+    waker: Arc<Mutex<Option<Waker>>>,
+}
```
Ralith, line 107: Would it be simpler to `Arc` the whole `InMemorySocket`?

bench/src/simulated_network.rs
```diff
@@ -0,0 +1,250 @@
...
+
+use queue::InboundQueue;
+
+#[derive(Debug)]
+pub struct InMemorySocketHandle {
+    pub network: Arc<InMemoryNetwork>,
+    pub addr: SocketAddr,
+}
+
+impl AsyncUdpSocket for InMemorySocketHandle {
```
Ralith, line 19: Should this also implement the recently-introduced `fn may_fragment(&self) -> bool { false }`?

bench/src/simulated_network.rs
```diff
@@ -0,0 +1,250 @@
...
+                    *waker = Some(cx.waker().clone())
+                }
+            } else {
+                // Wake up next time we can read
+                let next_read = inbound.time_of_next_receive();
+                let waker = cx.waker().clone();
+                tokio::task::spawn(async move {
+                    tokio::time::sleep_until(next_read.into()).await;
+                    waker.wake();
+                });
```
Ralith, line 88: If we assume at most one reader, which I think is reasonable, we could store the `Sleep` future in `self` and poll it directly.

aochagavia:
Closing this, since there is little chance I'll be able to work on it anytime soon

PR #1560, opened 2023-05-09
Add perf_datagram_{server, client}

aochagavia:
This PR adds a `perf_datagram_server` and `perf_datagram_client` to benchmark datagram transmit performance. There are two modes of operation: sending all the data through datagrams, or sending all the data in a single stream. That makes it possible to compare performance between the two. I dedided to add them as separate binaries because they differ too much from the existing `perf_{client, server}`, but I reused the code where possible.

Interestingly, there is a huge discrepancy in performance between datagrams and streams. Somehow the throughput when using datagrams is much lower! Below follows a comparison. I won't have time to investigate the reason of the difference, but as far as I can see the benchmarking code is correct.

### Using datagrams

Server: `cargo run --release --bin perf_datagram_server -- --no-protection`

Client: `cargo run --release --bin perf_datagram_client -- --no-protection`

Output:

```
Transferred 2147484066 bytes in 6670 ms (307.04 MiB/s)
Bytes lost 2561048 (2.44 MiB)
Congestion events: 23
Congestion window: 454043
```

### Using a stream

Server: `cargo run --release --bin perf_datagram_server -- --no-protection --use-stream`

Client: `cargo run --release --bin perf_datagram_client -- --no-protection --use-stream`

Output:

```
Transferred 2147483648 bytes in 1185 ms (1727.60 MiB/s)
Bytes lost 627092 (0.60 MiB)
Congestion events: 7
Congestion window: 4069176
```

### DIFF
--- a/perf/src/bin/perf_client.rs
+++ b/perf/src/bin/perf_client.rs
@@ -1,5 +1,5 @@
 use std::{
-    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
+    net::{IpAddr, SocketAddr},
     sync::Arc,
     time::{Duration, Instant},
 };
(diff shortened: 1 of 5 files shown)


aochagavia:
Closing this, since there is little chance I'll be able to work on it anytime soon

PR #1563, opened 2023-05-11
deps: remove webpki dependency.

cpu:
This commit updates the existing Quinn sub-crate dependencies to remove the explicit `webpki` v0.22 dependency. It's not used by directly anywhere and can be a transitive dependency through rustls as required.

### DIFF
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -18,7 +18,7 @@ maintenance = { status = "experimental" }
 
 [features]
 default = ["tls-rustls", "log"]
-tls-rustls = ["rustls", "webpki", "ring"]
+tls-rustls = ["rustls", "ring"]
 # Provides `ClientConfig::with_native_roots()` convenience method
 native-certs = ["rustls-native-certs"]
 # Write logs via the `log` crate when no `tracing` subscriber exists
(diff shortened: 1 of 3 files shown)


quinn-proto/Cargo.toml
```diff
@@ -36,7 +36,7 @@ slab = "0.4"
 thiserror = "1.0.21"
 tinyvec = { version = "1.1", features = ["alloc"] }
 tracing = "0.1.10"
-webpki = { version = "0.22", default-features = false, optional = true }
+webpki = { version = "0.100.1", package = "rustls-webpki", default-features = false, optional = true }
```
djc, line 39: Oops... how do we use the direct `webpki` dependency, anyway? Do we still need it? Is releasing this as 0.10.2 semver-compatible?

cpu: Good questions :sweat_smile:  I spotted it in `cargo tree` looking at the trust-dns update and didn't dig deeper naively assuming `cargo` would warn on an unused explicit dep. 

Looking closer now I don't think it's used anywhere so I've pushed a reworked version of this branch that drops the explicit dependency. I think that will be semver compatible :crossed_fingers:

djc:
Mind adding a commit that bumps the version on the affected library crates?

Ralith:
Thanks for catching this!

cpu:
> Mind adding a commit that bumps the version on the affected library crates?

:ballot_box_with_check: d909c7143b75dde11eab80d5f0798c72d7999241

djc:
Only the library crates, please -- we just leave the perf crate at 0.1.0.

cpu:
> Only the library crates, please -- we just leave the perf crate at 0.1.0.

Ah, ok :-) Fixed in b6a90d7

djc (approved):

djc:
Published 0.10.1 for quinn and quinn-proto.

PR #1574, opened 2023-05-26
Allow compilation with `--cfg fuzzing` but not `--features=arbitrary`

smoelius:
Right now, if `quinn-proto` is built without `--features=arbitrary`, one gets an error about an unresolved import. The `compile_error` helps to clarify why.

An alternative would be to try and disentangle `--cfg fuzzing` and `--features=arbitrary`. But this change is less invasive.

### DIFF
--- a/quinn-proto/src/lib.rs
+++ b/quinn-proto/src/lib.rs
@@ -88,10 +88,12 @@ pub mod fuzzing {
     pub use crate::frame::ResetStream;
     pub use crate::packet::PartialDecode;
     pub use crate::transport_parameters::TransportParameters;
-    use crate::MAX_CID_SIZE;
-    use arbitrary::{Arbitrary, Result, Unstructured};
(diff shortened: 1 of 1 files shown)


djc:
> An alternative would be to try and disentangle `--cfg fuzzing` and `--features=arbitrary`. But this change is less invasive.

Did you try looking at this? I think I would usually prefer disentangling over this.

djc (approved):

smoelius:
> > An alternative would be to try and disentangle `--cfg fuzzing` and `--features=arbitrary`. But this change is less invasive.
> 
> Did you try looking at this? I think I would usually prefer disentangling over this.

I did not, but if you hold off on merging, I will.

smoelius:
@djc Is this better?

Ralith (approved):
Thanks!

djc:
Yes, thank you!

PR #1575, opened 2023-05-26
Return `UnknownStream` on `finish` a closed stream

### LINKED ISSUE #1569: Return `UnknownStream` instead of panic when calling `finish()` on an already closed `SendStream`
I think the behavior should be consist of `reset()`. The panic is caused by

https://github.com/quinn-rs/quinn/blob/4b309cd46b5a68c96c105b00dfbc65216523677f/quinn/src/send_stream.rs#L143C11-L160

which polls and unwraps a used oneshot receiver.

I created a wrapper over a pair of `SendStream` and `RecvStream` to use `tokio::io::copy_bidirectional` on them. Since `tokio::io::copy_bidirectional` returns error immediately after encountering error on any direction without gracefully shutting down the other direction, I sometimes needs to manually close the other one. However, there is no way to distinguish which one should I close, it is required to explicitly close both one. Unfortunately, the `SendStream` may already been closed by `tokio::io::copy_bidirectional`, and calling close on it multiple times would cause panic. Additionally, the panicked thread holds the `MutexGuard`, which causes the lock poisoning and crashes the whole program.

EAimTY:
Fixes #1569

### DIFF
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -27,6 +27,7 @@ pub struct SendStream {
     stream: StreamId,
     is_0rtt: bool,
     finishing: Option<oneshot::Receiver<Option<WriteError>>>,
+    finished: bool,
 }
 
 impl SendStream {
@@ -36,6 +37,7 @@ impl SendStream {
             stream,
             is_0rtt,
             finishing: None,
+            finished: false,
         }
     }
 
@@ -122,6 +124,10 @@ impl SendStream {
 
     #[doc(hidden)]
     pub fn poll_finish(&mut self, cx: &mut Context) -> Poll<Result<(), WriteError>> {
+        if self.finished {
+            return Poll::Ready(Err(WriteError::UnknownStream));
+        }
+
         let mut conn = self.conn.state.lock("poll_finish");
         if self.is_0rtt {
             conn.check_0rtt()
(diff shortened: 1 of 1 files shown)


quinn/src/send_stream.rs
```diff
@@ -122,6 +124,10 @@ impl SendStream {
 
     #[doc(hidden)]
     pub fn poll_finish(&mut self, cx: &mut Context) -> Poll<Result<(), WriteError>> {
+        if self.finished {
```
djc, line 127: Actually, is there any reason we shouldn't fuse this to `Ok(())`?

Ralith: My preference would be to return `UnknownStream` for consistency rather than tracking additional state. Easy enough for callers to ignore when they really do want to make potentially redundant calls. Also compare `Recv::stop` in proto.

Ralith (changes requested):
Thanks for the PR! I think we could avoid the extra `bool` by setting `self.finishing = None` after it succeeds. This will route to the next call through the `if self.finishing.is_none() {` branch, which will check the canonical stream state in the `proto` layer and return `UnknownStream` automatically.

Ralith:
Calling `finish` on a stream that was previously reset should be an error, since otherwise the caller might conclude that all data was transmitted. We could track whether a stream was reset, but that'd require extra state in `quinn` and could not be sensibly implemented in `quinn-proto`. Consistently returning `UnknownStream` for streams which may have been closed in any way seems like the best path.

Matthias247:
That makes sense. I assumed this was just for the case where the stream fully finished by having sent all the data, and not that it had been cancelled via a reset

Ralith:
This specific issue is, but a graceful fix (per above) means not tracking that information.

Ralith:
Went ahead and drafted the approach I described in https://github.com/quinn-rs/quinn/pull/1588, since it's affecting multiple users. Thanks for bringing this to our attention!

Issue #1576, opened 2023-05-27
Intra-endpoint horizontal scaling


Ralith:
Quinn, like any flexible QUIC implementation, supports horizontal scaling across multiple endpoints through the use of custom connection ID generators and QUIC-aware load-balancing front-ends. However, the requirement for a third party load balancer and custom logic makes this difficult to leverage. For the overwhelmingly common case of applications that fit comfortably on a single host, Quinn should allow a single endpoint to scale the number of concurrent near-linearly with respect to available CPU cores.

In the current architecture, separate connection drivers already allow significant cryptographic work, and all application-layer work, for independent connections to happen in parallel. A bottleneck remains at the endpoint driver, an async task responsible for driving all network I/O and timers. We can do better.

The Linux [`SO_REUSEPORT`](https://lwn.net/Articles/542629/) option will distribute incoming packets on a single port across multiple sockets. Windows "Receive Side Scaling" may be similar. On at least these platforms, we can therefore run up to one endpoint driver per core. Some architectural changes are required to prevent catastrophic contention.

Key `quinn_proto::Endpoint` methods take `&mut self`, preventing meaningful parallelization of datagram processing. We should embrace interior mutability to convert these to `&self` methods which will not contend on the hot path of handling datagrams of established connections. In particular:
- Independent `Connection` states *must* be concurrently mutable. Morally this could be as simple as `RwLock<Slab<Mutex<Connection>>>`.
- Global locking on connection creation/destruction is difficult to avoid, but probably tolerable. See also sharded-slab, a complex but fast storage option.
- An explicit "endpoint view" notion in quinn-proto might be useful to allow lock-free access to per-socket state such as transmit queues without pushing too much more complexity into the `quinn` layer.

SO_REUSEPORT consistently routes packets to sockets based on a 4-tuple hash, so we can rely on connections moving between drivers only in the event of migrations, minimizing contention outside of connection setup/teardown.

Tokio presently lacks a mechanism to ensure certain tasks run on separate threads, which may complicate improving parallelism for high-level users. We could address this by spawning our own threads, or by working with upstream to develop new APIs. It's also possible that simply spawning N drivers and letting work stealing do its thing might work out well enough in practice.

### Unified Drivers

[...]

PR #1578, opened 2023-05-28
Horizontally scalable endpoints

Ralith:
Having a proper go at #1576. This should ultimately reduce the inter-task communication required at the quinn layer by allowing connection tasks to manipulate the endpoint directly, and form a strong foundation for a more incremental and performant attempt at #1219.

Initial batch of changes doesn't touch the quinn layer yet, but have turned out pretty graceful and self contained, so no need to wait. Still, I'll probably add to this.


djc:
Maybe split off the incidental cleanups/fixes into a separate PR? Easier to get through, and less prone to change (as in your latest insights from #1576, which I assume might affect this PR). (Maybe also the refactoring of mutable state, not sure how much sense that makes on its own?)

PR #1579, opened 2023-06-01
Replace unmaintained actions-rs/* actions in CI workflows

striezel:
Basically all of the `actions-rs/*` actions are unmaintained. See <https://github.com/actions-rs/toolchain/issues/216> for more information. Due to their age they generate several warnings in CI runs, for example in https://github.com/quinn-rs/quinn/actions/runs/5100061676:

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions-rs/toolchain@v1, Swatinem/rust-cache@v1, actions-rs/cargo@v1. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

To get rid of some of those warnings the occurrences of `actions-rs/toolchain` are replaced by [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain), and the occurrences of `actions-rs/cargo` are replaced by direct invocations of `cargo`.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -9,11 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: stable
-          override: true
+      - uses: dtolnay/rust-toolchain@stable
       - uses: taiki-e/install-action@cargo-llvm-cov
       - run: cargo llvm-cov --all-features --workspace --lcov --output-path lcov.info
       - name: Upload coverage to Codecov
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -45,64 +45,35 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
         with:
-          profile: minimal
           toolchain: ${{ matrix.rust }}
-          override: true
       - uses: Swatinem/rust-cache@v1
-      - uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: --all-targets
-      - uses: actions-rs/cargo@v1
-        with:
-          command: test
-      - uses: actions-rs/cargo@v1
+      - run: cargo build --all-targets
+      - run: cargo test
+      - run: cargo test --manifest-path fuzz/Cargo.toml
         if: ${{ matrix.rust }} == "stable"
-        with:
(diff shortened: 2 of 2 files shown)


Ralith:
Thanks! I see you've used `@master` -- should this be `@v1` for stability?

.github/workflows/codecov.yml
```diff
@@ -9,11 +9,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
```
djc, line 12: IIRC `dtolnay/rust-toolchain` offers a `stable` version, which makes it possible to ditch the separate `toolchain` argument, would be a nice improvement.

djc (commented):
Thanks for taking care of this!

Can you also add a snippet to our Dependabot config to check Actions?

```
- package-ecosystem: github-actions
  directory: "/"
  schedule:
    interval: weekly
```

.github/workflows/codecov.yml
striezel: Yes, I've just changed that.

striezel:
@Ralith:

> Thanks! I see you've used `@master` -- should this be `@v1` for stability?

I would not recommend to do this, because [according to its creator that is unsupported](https://github.com/dtolnay/rust-toolchain/issues/45#issuecomment-1336100499):

> The `v1` tag only exists so that the action gets a page on https://github.com/marketplace/actions. (And it has to be a tag for that to happen.) I am not interested in updating it. `master` and all the other branches (`stable`, `1.65` etc) are the only supported and documented way to use the action.

@djc:
Concerning the Dependabot stuff:

> Can you also add a snippet to our Dependabot config to check Actions?
> 
> ```
> - package-ecosystem: github-actions
>   directory: "/"
>   schedule:
>     interval: weekly
> ```

Yes, I could do this, too, if you want it, but I would prefer to do this in a separate pull request.

Ralith:
> I would not recommend to do this, because https://github.com/dtolnay/rust-toolchain/issues/45#issuecomment-1336100499:

Ah, good catch. Why `@stable` in some places an `@master` elsewhere though?

djc (approved):

djc:
> > I would not recommend to do this, because [dtolnay/rust-toolchain#45 (comment)](https://github.com/dtolnay/rust-toolchain/issues/45#issuecomment-1336100499):
> 
> Ah, good catch. Why `@stable` in some places an `@master` elsewhere though?

`@stable` where we need the stable toolchain, `@master` where we explicitly select some other version.

PR #1581, opened 2023-06-03
Miscellaneous cleanup

Ralith:
Straightforward improvements extracted from #1578.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -40,23 +40,7 @@ use crate::{
 pub struct Endpoint {
     rng: StdRng,
     transmits: VecDeque<Transmit>,
-    /// Identifies connections based on the initial DCID the peer utilized
-    ///
-    /// Uses a standard `HashMap` to protect against hash collision attacks.
-    connection_ids_initial: HashMap<ConnectionId, ConnectionHandle>,
-    /// Identifies connections based on locally created CIDs
-    ///
-    /// Uses a cheaper hash function since keys are locally created
-    connection_ids: FxHashMap<ConnectionId, ConnectionHandle>,
-    /// Identifies connections with zero-length CIDs
-    ///
-    /// Uses a standard `HashMap` to protect against hash collision attacks.
-    connection_remotes: HashMap<FourTuple, ConnectionHandle>,
-    /// Reset tokens provided by the peer for the CID each connection is currently sending to
-    ///
-    /// Incoming stateless resets do not have correct CIDs, so we need this to identify the correct
-    /// recipient, if any.
-    connection_reset_tokens: ResetTokenTable,
+    index: ConnectionIndex,
     connections: Slab<ConnectionMeta>,
     local_cid_generator: Box<dyn ConnectionIdGenerator>,
     config: Arc<EndpointConfig>,
@@ -79,10 +63,7 @@ impl Endpoint {
         Self {
             rng: StdRng::from_entropy(),
             transmits: VecDeque::new(),
-            connection_ids_initial: HashMap::default(),
-            connection_ids: FxHashMap::default(),
-            connection_remotes: HashMap::default(),
-            connection_reset_tokens: ResetTokenTable::default(),
+            index: ConnectionIndex::default(),
             connections: Slab::new(),
             local_cid_generator: (config.connection_id_generator_factory.as_ref())(),
             config,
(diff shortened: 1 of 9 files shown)


quinn-proto/src/endpoint.rs
```diff
@@ -748,17 +691,115 @@ impl fmt::Debug for Endpoint {
...
             .field("connections", &self.connections)
             .field("config", &self.config)
             .field("server_config", &self.server_config)
             .finish()
     }
 }
 
+/// Maps packets to existing connections
+#[derive(Default, Debug)]
+struct ConnectionIndex {
```
djc, line 706: Not sure `Index` feels like the most obvious suffix here. Maybe `Map`?

quinn-proto/src/endpoint.rs
```diff
@@ -748,17 +691,115 @@ impl fmt::Debug for Endpoint {
...
+            })
+            .or_else(|| {
+                let data = datagram.data();
+                if data.len() < RESET_TOKEN_SIZE {
+                    return None;
+                }
+                self.connection_reset_tokens
+                    .get(addresses.remote, &data[data.len() - RESET_TOKEN_SIZE..])
+            })
+            .cloned()
```
djc, line 796: In this unnested context, I think this could benefit from rewriting with early returns (and the `?` operator) and explicit matches rather than this combinator festival...

quinn-proto/src/endpoint.rs
```diff
@@ -559,15 +548,18 @@ impl Endpoint {
...
                 trace!(id = ch.0, icid = %dst_cid, "connection incoming");
-                Some((ch, conn))
+                Some(DatagramEvent::NewConnection(ch, conn))
             }
             Err(e) => {
                 debug!("handshake failed: {}", e);
                 self.handle_event(ch, EndpointEvent(EndpointEventInner::Drained));
                 if let ConnectionError::TransportError(e) = e {
-                    self.initial_close(version, addresses, crypto, &src_cid, &loc_cid, e);
+                    Some(DatagramEvent::Response(self.initial_close(
```
djc: Style nit: maybe replace the `if let` with a `match` here?

quinn-proto/src/endpoint.rs
```diff
@@ -129,7 +121,7 @@ impl Endpoint {
         local_ip: Option<IpAddr>,
         ecn: Option<EcnCodepoint>,
         data: BytesMut,
-    ) -> Option<(ConnectionHandle, DatagramEvent)> {
+    ) -> Option<DatagramEvent> {
```
djc, line 124: I think this could be factored out into a separate commit? Seems like a somewhat independent change, would make it a little easier to review.

quinn-proto/src/tests/util.rs
```diff
@@ -310,26 +310,25 @@ impl TestEndpoint {
...
                             .push_back(event);
                     }
+                    DatagramEvent::Response(transmit) => {
+                        self.outbound.extend(split_transmit(transmit));
+                    }
                 }
             }
         }
 
-        while let Some(x) = self.poll_transmit() {
```
djc, line 329: I suppose this changes the order of things. Have you reasoned about why it's okay to change things? Wasn't there a reason to do "metadata" events before I/O?

quinn-proto/src/endpoint.rs
```diff
@@ -62,7 +61,6 @@ impl Endpoint {
     ) -> Self {
         Self {
             rng: StdRng::from_entropy(),
-            transmits: VecDeque::new(),
```
djc, line 81: Certainly a nice direction!

quinn-proto/src/endpoint.rs
```diff
@@ -575,7 +582,8 @@ impl Endpoint {
         tls: Box<dyn crypto::Session>,
         server_config: Option<Arc<ServerConfig>>,
         transport_config: Arc<TransportConfig>,
-    ) -> (ConnectionHandle, Connection) {
+        ch: ConnectionHandle,
```
djc, line 585: This feels out of order with respect to our usual style?

djc (commented):
Nice work! Should we proactively bump the quinn-proto version incompatibly?

quinn-proto/src/endpoint.rs
Ralith: I think `Index` is descriptive here. A map owns its values; an index (as in database index) merely refers to them.

quinn-proto/src/endpoint.rs
Ralith: Good call. We can't use `?` here because the desired fallback is always "try the next case", but it's still a fair bit more readable this way.

quinn-proto/src/endpoint.rs
Ralith: Done.

quinn-proto/src/endpoint.rs
Ralith: Done.

quinn-proto/src/endpoint.rs
Ralith: Fixed.

Ralith:
> Should we proactively bump the quinn-proto version incompatibly?

Ah, right, forgot the new policy.

quinn-proto/src/tests/util.rs
Ralith: I don't think this changes any behavior, because appending to `self.outbound` has no side effects.

djc: Ah, fair. So what about for other downstream users of the quinn-proto crate?

djc (approved):

quinn-proto/src/tests/util.rs
Ralith: Transmits not associated with any connection are by definition not associated with any persistent state, so no special care should be required, even leaving aside the fact that we could never rely on the ordering of real I/O to begin with.

PR #1585, opened 2023-06-06
Added Endpoint::reject_new_connections

### LINKED ISSUE #1584: How to reject incoming connections after quinn 0.9?
I'm [trying to upgrade a library](https://github.com/khonsulabs/fabruic/pull/58) that uses quinn under the hood, and the library previously offered the ability to close incoming connections to enable a way to gracefully shut down a server. This worked in 0.8 by passing the `Incoming` stream into its own task, and stopping that task once `close_incoming` was called. This dropped the `Incoming` type, which caused incoming connections to be refused. 

In quinn 0.9 and later, `Endpoint::server` no longer returns an `Incoming` type, and the `Endpoint` now has an `accept()` function instead. All of my attempts to prevent incoming connections have not worked: either the incoming connection is accepted before being terminated or the connection hangs rather than being refused.

Is there a way to preserve the functionality we were utilizing in 0.8 in the current version? Thank you, and apologies if I missed something in the docs for how to handle this!

ecton:
Closes #1584

This commit adds a new function that refuses new connections without impacting existing connections. Internally, this just sets the connection limit to 0, which causes incoming connections to be rejected. This is the same approach that was taken in 0.8.5 when `Incoming` was dropped.

Let me know if I should also add a unit test to the `quinn` crate. Since it's a simple pass-through, I elected not to initially.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -653,7 +653,12 @@ impl Endpoint {
         }
     }
 
-    /// Rejects new connections without affecting existing connections.
+    /// Reject new incoming connections without affecting existing connections
+    ///
+    /// Convenience short-hand for using
+    /// [`set_server_config`](Self::set_server_config) to update
+    /// [`concurrent_connections`](ServerConfig::concurrent_connections) to
+    /// zero.
     pub fn reject_new_connections(&mut self) {
         if let Some(config) = self.server_config.as_mut() {
             Arc::make_mut(config).concurrent_connections(0);
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -653,6 +653,13 @@ impl Endpoint {
         }
     }
 
+    /// Rejects new connections without affecting existing connections.
+    pub fn reject_new_connections(&mut self) {
+        if let Some(config) = self.server_config.as_mut() {
+            Arc::make_mut(config).concurrent_connections(0);
+        }
+    }
+
     /// Access the configuration used by this endpoint
     pub fn config(&self) -> &EndpointConfig {
         &self.config
--- a/quinn-proto/src/tests/util.rs
+++ b/quinn-proto/src/tests/util.rs
@@ -166,15 +166,6 @@ impl Pair {
         }
     }
 
-    /// Attempts to connect, asserting that the server doesn't accept the
-    /// connection and the client marks the connection as closed.
-    pub(super) fn assert_connection_refused(&mut self) {
-        let client_ch = self.begin_connect(client_config());
-        self.drive();
-        self.server.assert_no_accept();
-        assert!(self.client.connections.get(&client_ch).unwrap().is_closed());
-    }
-
     pub(super) fn connect(&mut self) -> (ConnectionHandle, ConnectionHandle) {
         self.connect_with(client_config())
     }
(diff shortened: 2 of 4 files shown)


Ralith:
Thanks for the PR! Have you tried instead using the existing `Endpoint::set_server_config` interface to accomplish this in your application?

djc:
@Ralith maybe we can pass a `bool` to `wait_idle()` that takes care of the `TransportConfig` trick?

quinn-proto/src/endpoint.rs
```diff
@@ -653,6 +653,13 @@ impl Endpoint {
         }
     }
 
+    /// Rejects new connections without affecting existing connections.
```
Ralith, line 656:
```suggestion
    /// Reject new incoming connections without affecting existing connections
    ///
    /// Convenience short-hand for using [`set_server_config`](Self::set_server_config) to update [`concurrent_connections`](ServerConfig::concurrent_connections) to zero.
```

quinn-proto/src/tests/util.rs
```diff
@@ -166,6 +166,15 @@ impl Pair {
         }
     }
 
+    /// Attempts to connect, asserting that the server doesn't accept the
+    /// connection and the client marks the connection as closed.
+    pub(super) fn assert_connection_refused(&mut self) {
```
Ralith, line 171: Factoring this into a helper method seems premature with only one caller, given that it's just a few lines.

quinn/src/endpoint.rs
```diff
@@ -236,6 +236,16 @@ impl Endpoint {
         self.inner.state.lock().unwrap().socket.local_addr()
     }
 
+    /// Rejects new connections without affecting existing connections.
```
Ralith, line 239: Reword as above

Ralith (changes requested):
Since this is an important API for certain graceful shutdown patterns, including it probably makes sense for discoverability.

Ralith:
> maybe we can pass a bool to `wait_idle()`

That doesn't seem very discoverable either, TBH. After discussion I think this is fine.

Ralith (approved):

djc (approved):

quinn-proto/src/endpoint.rs
```diff
@@ -653,6 +653,18 @@ impl Endpoint {
...
 
+    /// Reject new incoming connections without affecting existing connections
+    ///
+    /// Convenience short-hand for using
+    /// [`set_server_config`](Self::set_server_config) to update
+    /// [`concurrent_connections`](ServerConfig::concurrent_connections) to
+    /// zero.
+    pub fn reject_new_connections(&mut self) {
+        if let Some(config) = self.server_config.as_mut() {
+            Arc::make_mut(config).concurrent_connections(0);
```
MOZGIII, line 664: Should this me `Arc::make_mut` and also handle a case where the value is used concurrently more explicitly? I don't think clone-on-write is really what's meant here...

djc: For the purposes of this code, it seems sufficient to me that `self.server_config` holds the modified config. Seems to me that we don't really care that `Connection`s are holding on to a "stale" `ServerConfig` that still allows more `concurrent_connections`.

MOZGIII: I misread what `Arc::make_mut` actually does. Yeah, this design is good!

PR #1588, opened 2023-06-11
Fix panic on redundant finish calls

Ralith:
Obsoletes https://github.com/quinn-rs/quinn/pull/1575 which hasn't been updated in a while, plus bonus cosmetic cleanup.

### DIFF
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -144,8 +144,10 @@ impl SendStream {
             .poll(cx)
             .map(|x| x.unwrap())
         {
-            Poll::Ready(None) => Poll::Ready(Ok(())),
-            Poll::Ready(Some(e)) => Poll::Ready(Err(e)),
+            Poll::Ready(x) => {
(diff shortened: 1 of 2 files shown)


quinn/src/send_stream.rs
```diff
@@ -144,8 +144,13 @@ impl SendStream {
             .poll(cx)
             .map(|x| x.unwrap())
         {
-            Poll::Ready(None) => Poll::Ready(Ok(())),
-            Poll::Ready(Some(e)) => Poll::Ready(Err(e)),
+            Poll::Ready(x) => {
+                self.finishing = None;
+                match x {
```
djc, line 149: Nit: hoist the `Poll::Ready(..)` call?

djc (approved):
Is this worth a maintenance release?

quinn/src/send_stream.rs
Ralith: This made a nice concise combinator solution obvious! :tada:

PR #1595, opened 2023-06-29
Use boxed AsyncUdpSocket when creating abstract endpoint

tthebst:
Wrapping the std socket with either  `TokioRuntime` or `AsyncStdRuntime` returns `Box<dyn AsyncUdpSocket>`.  Creating an endpoint with an abstract socket requires `impl AsyncUdpSocket` and it is therefore not possible to use a wrapped socket together with `new_with_abstract_socket`.

This MR fixes this inconsistency by using `Box<dyn AsyncUdpSocket>` in `new_with_abstract_socket`.

### DIFF
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -101,10 +101,10 @@ impl Endpoint {
     pub fn new_with_abstract_socket(
         config: EndpointConfig,
         server_config: Option<ServerConfig>,
-        socket: impl AsyncUdpSocket,
+        socket: Box<dyn AsyncUdpSocket>,
         runtime: Arc<dyn Runtime>,
     ) -> io::Result<Self> {
-        Self::new_with_runtime(config, server_config, Box::new(socket), runtime)
+        Self::new_with_runtime(config, server_config, socket, runtime)
     }
 
     fn new_with_runtime(
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -56,7 +56,7 @@ impl Endpoint {
         let socket = std::net::UdpSocket::bind(addr)?;
         let runtime = default_runtime()
             .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "no async runtime found"))?;
-        Self::new_with_runtime(
+        Self::new_with_abstract_socket(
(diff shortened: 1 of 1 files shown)


djc (approved):

quinn/src/endpoint.rs
```diff
@@ -101,10 +101,10 @@ impl Endpoint {
     pub fn new_with_abstract_socket(
         config: EndpointConfig,
         server_config: Option<ServerConfig>,
-        socket: impl AsyncUdpSocket,
+        socket: Box<dyn AsyncUdpSocket>,
         runtime: Arc<dyn Runtime>,
     ) -> io::Result<Self> {
-        Self::new_with_runtime(config, server_config, Box::new(socket), runtime)
+        Self::new_with_runtime(config, server_config, socket, runtime)
```
Ralith, line 107: It looks like this signature has converged to being identical with `new_with_runtime`. Want to inline that and update its callsites to use `new_with_abstract_socket` directly for simplicity?

tthebst: Good idea. Removed `new_with_runtime` and replaced it with `new_with_abstract_socket`.

Ralith (approved):
Thanks!

quinn/src/endpoint.rs
```diff
@@ -91,23 +91,14 @@ impl Endpoint {
...
-        config: EndpointConfig,
-        server_config: Option<ServerConfig>,
-        socket: Box<dyn AsyncUdpSocket>,
-        runtime: Arc<dyn Runtime>,
-    ) -> io::Result<Self> {
-        Self::new_with_runtime(config, server_config, socket, runtime)
-    }
-
-    fn new_with_runtime(
+    fn new_with_abstract_socket(
```
djc, line 101: Um, this replaced the public constructor with only a private one. That doesn't seem like the right direction?

tthebst: Ops. Sorry. https://github.com/quinn-rs/quinn/pull/1596

djc: To be fair, it's also @Ralith's bad for not catching it in review! Thanks for following up.

PR #1597, opened 2023-07-03
Limiting transmit queue size from endpoint-generated transmit packets

lijunwangs:
Dropping endpoint generated transmit packets when the outgoing queue size exceeds the limit to bound memory usage. The selected value found to be limiting memory usage to about 1 G resident size for a server for sustained loads from clients and the queue to be stable around the set size.


quinn/src/endpoint.rs
```diff
@@ -436,7 +443,12 @@ impl State {
                                         .send(ConnectionEvent::Proto(event));
                                 }
                                 Some(DatagramEvent::Response(t)) => {
-                                    self.outgoing.push_back(udp_transmit(t));
+                                    if self.transmit_queue_size.load(Ordering::Relaxed)
+                                        < MAX_TRANSMIT_QUEUE_SIZE
+                                    {
+                                        self.outgoing.push_back(udp_transmit(t));
+                                        self.transmit_queue_size.fetch_add(1, Ordering::Relaxed);
```
Ralith, line 450: How about we track the number of payload bytes (`t.contents.len()`) rather than the number of packets? This lets us reason in terms of approximate memory use directly, and ensures consistency regardless of segmentation offload.

lijunwangs: Great suggestion. I'll give it a try.

quinn/src/endpoint.rs
```diff
@@ -523,7 +536,10 @@ impl State {
                                 .send(ConnectionEvent::Proto(event));
                         }
                     }
-                    Transmit(t) => self.outgoing.push_back(udp_transmit(t)),
+                    Transmit(t) => {
+                        self.outgoing.push_back(udp_transmit(t));
+                        self.transmit_queue_size.fetch_add(1, Ordering::Relaxed);
```
Ralith, line 541: The counter should be incremented when events are pushed onto the `events` queue by the connection, so that we count packets that are currently in the queue. That's why an atomic is needed.

Ralith: Alternatively, if we're only concerned about packets queued by the endpoint, trusting in congestion/flow control/pacing for connection-generated packets, we could just switch to a non-atomic counter that's fully internal to `Endpoint`.

lijunwangs: I think we do need to take consideration of connection generated packets for the connections in initial state -- I have found they could generate large amount of packets fast enough to generate a build up in the outgoing queue as well. So I think a Atomic variable is better for this purpose in the future and also for back porting to older version like 0.9.x.

quinn/src/endpoint.rs
lijunwangs: I have changed to based on the bytes count of transmit.contents.len()

quinn/src/endpoint.rs
```diff
@@ -523,7 +541,12 @@ impl State {
                                 .send(ConnectionEvent::Proto(event));
                         }
                     }
-                    Transmit(t) => self.outgoing.push_back(udp_transmit(t)),
+                    Transmit(t) => {
+                        let contents_len = t.contents.len();
+                        self.outgoing.push_back(udp_transmit(t));
+                        self.transmit_queue_contents_len
+                            .fetch_add(contents_len, Ordering::Relaxed);
```
Ralith, line 548: This only counts packets after they've been received by the connection, but they still occupy space while waiting in the `self.events` queue. Is that intended?

lijunwangs: Yes, in my test, I have not seen a build-up of items in the events  it is always the outgoing queue the problem. I think the difference here is the send-limiter when draining down the outgoing queue.

Ralith: Ah, good to know. We should document this in a comment, since it's not obvious otherwise. That also means there's no need for the counter to be an atomic, since we always have `&mut self` when accessing it; let's use a plain `usize` instead.

quinn/src/lib.rs
```diff
@@ -115,3 +115,8 @@ const RECV_TIME_BOUND: Duration = Duration::from_micros(50);
 
 /// The maximum amount of time that should be spent in `sendmsg()` calls per endpoint iteration
 const SEND_TIME_BOUND: Duration = Duration::from_micros(50);
+
+/// The maximum size of content length of packets in the outgoing transmit queue. Transmit packets
+/// generated from the endpoint (retry or initial close) can be dropped when this limit is being execeeded.
+/// Chose to represent 100 MB of data.
+const MAX_TRANSMIT_QUEUE_CONTENTS_LEN: usize = 100000000;
```
Ralith, line 122: Intersperse underscores every three digits (from the right), or write it as a product of smaller numbers, so that this is readable.

quinn/src/lib.rs
```diff
@@ -115,3 +115,8 @@ const RECV_TIME_BOUND: Duration = Duration::from_micros(50);
 
 /// The maximum amount of time that should be spent in `sendmsg()` calls per endpoint iteration
 const SEND_TIME_BOUND: Duration = Duration::from_micros(50);
+
+/// The maximum size of content length of packets in the outgoing transmit queue. Transmit packets
+/// generated from the endpoint (retry or initial close) can be dropped when this limit is being execeeded.
+/// Chose to represent 100 MB of data.
+const MAX_TRANSMIT_QUEUE_CONTENTS_LEN: usize = 10_0000_000;
```
Ralith, line 122: I think you meant to write `100_000_000`?

Ralith (approved):
Thanks!

quinn/src/lib.rs
lijunwangs: oh. yes

Ralith: You left the extra 0 in there, making this 1G, which I don't think was intended.

lijunwangs: The 100MB was intended not 10MB. With 100MB. Running a simple QUIC server which is under the persistent flood of initial packets actually shows RES memory usage of more than 2g.

Ralith: The value currently written is 1G, not 100MB.

lijunwangs: Oh. My bad! Sorry. Updating it to true 100MB

Ralith (approved):

PR #1598, opened 2023-07-10
Backport supress stateless packet to 0.10.x

lijunwangs:
Backport e652b6d999f053ffe21eeea247854882ae480281 from main to 0.10.x branch. 

The implementation is different from the main branch due to the refactoring happened to main branch. 
 
   1. The enforcement of the limit is done at the quinn-proto layer as it is where the endpoint generated packets being queued. 
   2. The transmit_queue_contents_len to track the buffer contents length at quinn-proto::Endpoint layer.
   3. The outgoing_queue_contents_len to track the buffer contents length at Quinn::Endpoint layer
   4. Quinn::Endpoint uses the new interface set_socket_buffer_fill exposed by proto::Endpoint to communicate the updated outgoing_queue_contents_len

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -66,6 +66,7 @@ pub struct Endpoint {
     server_config: Option<Arc<ServerConfig>>,
     /// Whether the underlying UDP socket promises not to fragment packets
     allow_mtud: bool,
+    max_transmit_queue_contents_len: usize,
     transmit_queue_contents_len: Arc<AtomicUsize>,
 }
 
@@ -79,6 +80,7 @@ impl Endpoint {
         config: Arc<EndpointConfig>,
         server_config: Option<Arc<ServerConfig>>,
         allow_mtud: bool,
+        max_transmit_queue_contents_len: usize,
         transmit_queue_contents_len: Arc<AtomicUsize>,
     ) -> Self {
         Self {
@@ -93,6 +95,7 @@ impl Endpoint {
             config,
             server_config,
             allow_mtud,
+            max_transmit_queue_contents_len,
             transmit_queue_contents_len,
         }
     }
@@ -462,6 +465,15 @@ impl Endpoint {
         }
     }
 
+    fn to_supresss_stateless_packet(&self) -> bool {
+        // Limiting the memory usage for items queued in the outgoing queue from endpoint
+        // generated packets. Otherwise, we may see a build-up of the queue under test with
+        // flood of initial packets against the endpoint. The sender with the sender-limiter
+        // may not keep up the pace of these packets queued into the queue.
+        self.transmit_queue_contents_len.load(Ordering::Relaxed)
+            >= self.max_transmit_queue_contents_len
+    }
+
     fn handle_first_packet(
         &mut self,
         now: Instant,
@@ -536,6 +548,9 @@ impl Endpoint {
 
         let (retry_src_cid, orig_dst_cid) = if server_config.use_retry {
             if token.is_empty() {
+                if self.to_supresss_stateless_packet() {
+                    return None;
+                }
                 // First Initial
                 let mut random_bytes = vec![0u8; RetryToken::RANDOM_BYTES_LEN];
                 self.rng.fill_bytes(&mut random_bytes);
@@ -697,6 +712,9 @@ impl Endpoint {
         local_id: &ConnectionId,
         reason: TransportError,
     ) {
+        if self.to_supresss_stateless_packet() {
+            return;
+        }
         let number = PacketNumber::U8(0);
         let header = Header::Initial {
             dst_cid: *remote_id,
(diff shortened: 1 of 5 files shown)


lijunwangs:
Hi Ben/Dirkjan, can you please review the backport implementation of https://github.com/quinn-rs/quinn/commit/e652b6d999f053ffe21eeea247854882ae480281 to the 0.10.x release?  @Ralith @djc

quinn-proto/src/endpoint.rs
```diff
@@ -75,6 +84,7 @@ impl Endpoint {
         config: Arc<EndpointConfig>,
         server_config: Option<Arc<ServerConfig>>,
         allow_mtud: bool,
+        transmit_queue_contents_len: Arc<AtomicUsize>,
```
Ralith, line 87: This is a breaking change, and hence defeats the purpose of the backport. A better solution might be to add a getter on `proto::Endpoint` for the current queue length, which could then be called from the `quinn::Endpoint`. This also avoids needing shared ownership of an atomic.

Ralith (changes requested):

quinn-proto/src/endpoint.rs
lijunwangs: @Ralith Can you elaborate how this is a breaking change? Older quinn crate with the newer quinn-proto version? Adding a getter at the proto::Endpoint cannot solve the issue as we are doing the suppression at proto::Endpoint not at quinn::Endpoint. Am I missing something?

lijunwangs: Maybe provide a getter to get the Arc<AtomicUsize> at proto::Endpoint? If so, that can work. For the breaking change, I guess you are referring to the proto::Endpoint::new interface change?

Ralith: Changing the signature of a public function is always a breaking change, and hence cannot be made in a patch release.

A getter accessing the current total size of the `proto::Endpoint` transmit queue would be much simpler than sharing an atomic.

lijunwangs: Got it. We will need to setters (increment and decrement)

Ralith: Rather than separate increment/decrement (which is error-prone), a setter for a single absolute count that the -proto layer should consider in addition to its internal count would be simpler and more robust.

lijunwangs: What should we call that counter?

lijunwangs: How about connection_transmit_queue_contents_len?

Ralith: Perhaps `socket_buffer_fill`? Doesn't matter too much, since this is only a backport.

lijunwangs: Done

Ralith: Doesn't look like you've pushed?

lijunwangs: ahh -- sorry,  re-pushed to the right branch!

quinn-proto/src/endpoint.rs
```diff
@@ -447,6 +463,35 @@ impl Endpoint {
         }
     }
 
+    /// Limiting the memory usage for items queued in the outgoing queue from endpoint
+    /// generated packets. Otherwise, we may see a build-up of the queue under test with
+    /// flood of initial packets against the endpoint. The sender with the sender-limiter
+    /// may not keep up the pace of these packets queued into the queue.
+    fn to_supresss_stateless_packets(&self) -> bool {
```
Ralith, line 470: nit: rename to `stateless_reset_packets_suppressed`?

Ralith (approved):
Ah, I see why you were thinking about separate methods now. I'm happy either way.

quinn-proto/src/tests/mod.rs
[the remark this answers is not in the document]
lijunwangs: Yes, will revert this

quinn-proto/src/endpoint.rs
lijunwangs: will do.

quinn-proto/src/tests/mod.rs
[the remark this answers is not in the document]
lijunwangs: Done

Ralith (commented):
This looks good so far, but did you also want to disable version negotiation and stateless reset packets under the heavily loaded condition?

lijunwangs:
> This looks good so far, but did you also want to disable version negotiation and stateless reset packets under the heavily loaded condition?

Great point! I reviewed the implementation in main -- we were suppressing version negotiation and stateless reset packets as well. I will add these into the consideration as well to keep it consistent and more robust.

Ralith (approved):
LGTM, thanks!

PR #1600, opened 2023-07-14
Reexport rustls types

jonatanzeidler:
This PR makes quinn reexport all `rustls` types as discussed in #300. This change helps fix the issue described in [this example](https://github.com/jonatanzeidler/quinn-rustls-issue) by allowing it to use the reexported types.

### DIFF
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -65,6 +65,8 @@ pub use proto::{
     ConnectionClose, ConnectionError, EndpointConfig, IdleTimeout, MtuDiscoveryConfig,
     ServerConfig, StreamId, Transmit, TransportConfig, VarInt,
 };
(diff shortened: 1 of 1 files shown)


jonatanzeidler:
Please check whether reexporting all of `rustls` is what we want, as well as whether the location and reexport name make sense to you.

Ralith:
The convention I've seen elsewhere is to reexport the crate by name at the top level, i.e. `pub extern crate rustls;` in `lib.rs`. I think that's a nice clean and intuitive approach.

jonatanzeidler:
Thanks for the hint. I changed it accordingly, but opted for the `pub use` syntax, which was already used for the `udp` crate, too.

djc (approved):

PR #1603, opened 2023-07-15
Backport MSRV bump to 0.9.x

Ralith:
Hopefully will make CI able to run tests.


Ralith:
Looks like this isn't even sufficient, and there'll be a boatload of lint issues anyway.

PR #1610, opened 2023-07-21
Expose more properties and setters on UdpState

### LINKED ISSUE #1609: Expose more `quinn-udp` UdpState properties
In the implementation of `quinn-udp` (I'm focusing on [UNIX/Linux](https://github.com/quinn-rs/quinn/blob/main/quinn-udp/src/unix.rs)), the following properties/methods on `UdpState` are used:

- `state.max_gso_segments.store`
- `state.set_sendmsg_einval()`
- `state.sendmsg_einval()`

They are ~required~ recommended when implementing one's own `AsyncUdpSocket`. Exposing them would enable more correct external implementations.

I can make a PR if you'd like :)

Icelk:
Allows downgrading system capabilities.
Closes #1609.

### DIFF
--- a/quinn-udp/src/lib.rs
+++ b/quinn-udp/src/lib.rs
@@ -78,6 +78,15 @@ impl UdpState {
         self.max_gso_segments.load(Ordering::Relaxed)
     }
 
+    /// Disables GSO, if your platform looks to support it but in reality doesn't.
+    pub fn disable_gso(&self) {
+        self.max_gso_segments.store(1, Ordering::SeqCst);
+    }
+    /// Disables GRO, if your platform looks to support it but in reality doesn't.
+    pub fn disable_gro(&mut self) {
+        self.gro_segments = 1;
+    }
+
     /// The number of segments to read when GRO is enabled. Used as a factor to
     /// compute the receive buffer size.
     ///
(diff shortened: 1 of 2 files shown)


Icelk:
Would it be possibly to release a patch version after this PR? I need to downgrade the system UDP capabilities to get `quinn` working with `tokio-uring` :)

Icelk:
Seems like 35089ee1c8708f6697767baf4f96fa6d7707e923 is the only commit since `0.11` with breaking changes. Could that maybe be cherry-picked out?

Ralith:
Thanks for looking into interop with tokio-uring! I'm not sure this is an approach we want to support. Aside from the (fixable) issues with exposing these implementation details, GSO and GRO are critical to performance under heavy load, and ECN support is required for conformance. io-uring support alone is unlikely to be balance those severe costs. tokio-uring support should require those features be available.

Ralith:
https://github.com/quinn-rs/quinn/pull/1612 will allow a similar level of control with less leakage, though it is a breaking change.

Ralith:
Closing as #1612 was merged. Let us know if that doesn't solve your issue.

Icelk:
Again, sorry for the late response. Your new code worked excellently and GSO & GRO are now properly implemented (https://github.com/Icelk/kvarn/tree/proper-quic/). So thank you! I completely agree with you, it would be bad not to use the offloads to maximize performance.

I blame my skepticism on my disregarding some error handling (i.e. trying again after failed GSO) and the fact that my server didn't support GSO (it's a VM), so thanks for showing me to the correct side!

Icelk:
Do you know when the next release is planned so these changes can be used in packages published to crates.io (my release is sort of blocking on this & a `tokio-uring` PR) :)

djc:
We don't have concrete plans right now. Given that I think a semver-incompatible rustls release is probably a few weeks away (https://github.com/rustls/rustls/issues/1435), I would be inclined to wait with doing a semver-incompatible Quinn release until that's ready. Would it help if we published an alpha release?

Icelk:
Waiting a few weeks is completely fine—my packages doesn't have a ton of users. And thanks for your awesome work. I've been coding this on/off for a few months now, so waiting another one is OK.

PR #1611, opened 2023-07-21
Fix excessive receive buffer allocation

### LINKED ISSUE #1608: `Endpoint`s with default config allocate too much memory on creation
Memory usage (RSS) of our app surges to hunders of MB in a few minutes after startup when building with quinn 0.10.1, comparing to ~50MB with quinn 0.9.X. With [dhat](https://crates.io/crates/dhat) we noticed large memory allocations during the creation `Endpoint`s:
image

We suspect [this commit](https://github.com/quinn-rs/quinn/pull/1529/commits/1c53b42056d5397ffb45d065d71f5d79fa1d071a), which increased the max udp payload size to 64KB, combining with the `recv_buf` size calculation,

https://github.com/quinn-rs/quinn/blob/0ae7c60b15637d7343410ba1e5cc3151e3814557/quinn/src/endpoint.rs#L673-L678

might be the root cause. On a GRO-enabled kernel we have [`udp_state.gro_segments() == 64`](https://github.com/quinn-rs/quinn/blob/0ae7c60b15637d7343410ba1e5cc3151e3814557/quinn-udp/src/unix.rs#L775), and [`BATCH_SIZE == 32`](https://github.com/quinn-rs/quinn/blob/0ae7c60b15637d7343410ba1e5cc3151e3814557/quinn-udp/src/unix.rs#L703), resulting in a 128MB `recv_buf` for each `Endpoint`.

Ralith:
Allocation logic previously considered a GRO batch to be one datagram, rather than (when supported) 64. This led to an unintended 128MiB allocation, now scaled back down to ~4MiB.

Fixes #1608.

### DIFF
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -384,20 +384,20 @@ pub(crate) struct Shared {
 impl State {
     fn drive_recv<'a>(&'a mut self, cx: &mut Context, now: Instant) -> Result<bool, io::Error> {
         self.recv_limiter.start_cycle();
+        let recv_count = recv_datagram_count(&self.udp_state) / self.udp_state.gro_segments();
         let mut metas = [RecvMeta::default(); BATCH_SIZE];
-        let mut iovs = MaybeUninit::<[IoSliceMut<'a>; BATCH_SIZE]>::uninit();
+        let mut iovs = std::array::from_fn::<MaybeUninit<IoSliceMut<'a>>, BATCH_SIZE, _>(|_| {
+            MaybeUninit::uninit()
+        });
         self.recv_buf
-            .chunks_mut(self.recv_buf.len() / BATCH_SIZE)
+            .chunks_mut(self.recv_buf.len() / recv_count)
             .enumerate()
-            .for_each(|(i, buf)| unsafe {
-                iovs.as_mut_ptr()
-                    .cast::<IoSliceMut>()
-                    .add(i)
-                    .write(IoSliceMut::<'a>::new(buf));
+            .for_each(|(i, buf)| {
+                iovs[i].write(IoSliceMut::<'a>::new(buf));
             });
-        let mut iovs = unsafe { iovs.assume_init() };
+        let iovs = unsafe { slice_assume_init_mut(&mut iovs[..recv_count]) };
         loop {
-            match self.socket.poll_recv(cx, &mut iovs, &mut metas) {
+            match self.socket.poll_recv(cx, iovs, &mut metas) {
                 Poll::Ready(Ok(msgs)) => {
                     self.recv_limiter.record_work(msgs);
                     for (meta, buf) in metas.iter().zip(iovs.iter()).take(msgs) {
@@ -552,6 +552,12 @@ impl State {
     }
 }
 
+// TODO: Remove once feature maybe_uninit_slice (https://github.com/rust-lang/rust/issues/63569)
+// stabilizes
+unsafe fn slice_assume_init_mut<T>(x: &mut [MaybeUninit<T>]) -> &mut [T] {
+    std::slice::from_raw_parts_mut(x.as_mut_ptr().cast(), x.len())
+}
+
 #[inline]
 fn udp_transmit(t: proto::Transmit) -> udp::Transmit {
     udp::Transmit {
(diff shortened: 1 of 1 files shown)


quinn/src/endpoint.rs
```diff
@@ -708,6 +713,13 @@ impl EndpointRef {
     }
 }
 
+/// Number of datagrams to allocate space for per recv
+fn recv_datagram_count(udp: &UdpState) -> usize {
```
djc, line 717: Style nit: move this down below remaining `EndpointRef` impls? Or maybe this should become a `UdpState` method? All the inputs are from quinn-udp...

djc (approved):

quinn/src/endpoint.rs
```diff
@@ -384,20 +384,20 @@ pub(crate) struct Shared {
...
-            .for_each(|(i, buf)| unsafe {
-                iovs.as_mut_ptr()
-                    .cast::<IoSliceMut>()
-                    .add(i)
-                    .write(IoSliceMut::<'a>::new(buf));
+            .for_each(|(i, buf)| {
+                iovs[i].write(IoSliceMut::<'a>::new(buf));
             });
-        let mut iovs = unsafe { iovs.assume_init() };
+        let iovs = unsafe { slice_assume_init_mut(&mut iovs[..recv_count]) };
```
mahkoh, line 398: If gro is available, then recv_count is 1. But AIUI, in a single iovec, you can only receive messages from the same source. So this means that in each syscall you can receive messages from at most one remote.

Would it make more sense to split this more dynamically? E.g. if we detect burst activity, create 1 big iovec with 64 gro segments. But if activity is more spread out, use many iovecs with one gro segment each.

quinn/src/endpoint.rs
Ralith: I'm trying to keep quinn-udp focused on the platform API, and exclude tuning parameters as much as possible.

quinn/src/endpoint.rs
Ralith: Good catch; this is probably a regression for endpoints with many active connections. However, per https://github.com/quinn-rs/quinn/pull/1354, if we're using GRO we must accept exactly 64 segments, or else we'll get inadvertent packet truncation. I guess we'll have to bite the bullet and lower the max datagram size...

mahkoh: Why is the current default 64k? Is this useful outside of communication on localhost?

Ralith: The default is [from RFC 9000](https://www.rfc-editor.org/rfc/rfc9000.html#transport-parameter-definitions), though we're free to pick something else. Whatever value we choose will impose a hard upper limit to MTU detection, and it'd be nice to Just Work on networks with jumbo frames. However, AFAIK such networks are rare, so I'm willing to drop this to a best-case internet MTU and let jumbo frame users adjust their configurations if necessary.

djc: Do we have to allocate per segment or can we allocate a large buffer which would allow some jumbo frames provided that most frames are smaller?

quinn/src/endpoint.rs
Ralith: GRO segment count is a property of the OS. The number of worst-case datagrams we want to be able to receive in one call is a tuning decision, except that it must be a multiple of the GRO batch size if we're going to use GRO.

quinn/src/endpoint.rs
Ralith: If we receive a GRO batch consisting of many jumbo frame packets, and we did not allocate enough space for all of them, packets may be truncated (i.e. lost) per https://github.com/quinn-rs/quinn/pull/1354.

Ralith: Consensus in the implementers' slack seems to be that `recvmmsg` isn't adding a lot here, i.e. we probably won't see much regression from doing one syscall per sender as in this PR. This is in principle workload dependent... it's tempting to leverage that to simplify our APIs a bunch, though.

djc: Yeah, sounds like ditching `recvmmsg()` could make sense.

djc:
Closing this, as we decided to merge #1615 instead.

PR #1612, opened 2023-07-21
Fold UdpState into AsyncUdpSocket

### LINKED ISSUE #1609: Expose more `quinn-udp` UdpState properties

Ralith:
Allows knowledge of UdpState to be isolated entirely within AsyncUdpSocket implementations, simplifying quinn::Endpoint and the poll_send API, and exposing more control over UDP feature checks to implementers.

Fixes #1609.

### DIFF
--- a/quinn-udp/src/windows.rs
+++ b/quinn-udp/src/windows.rs
@@ -8,7 +8,7 @@ use std::{
 
 use windows_sys::Win32::Networking::WinSock;
 
-use super::{log_sendmsg_error, RecvMeta, Transmit, UdpSockRef, UdpState, IO_ERROR_LOG_INTERVAL};
+use super::{log_sendmsg_error, RecvMeta, Transmit, UdpSockRef, IO_ERROR_LOG_INTERVAL};
 
 /// QUIC-friendly UDP interface for Windows
 #[derive(Debug)]
@@ -80,12 +80,7 @@ impl UdpSocketState {
         Ok(())
     }
 
-    pub fn send(
-        &self,
-        socket: UdpSockRef<'_>,
-        _state: &UdpState,
-        transmits: &[Transmit],
-    ) -> Result<usize, io::Error> {
+    pub fn send(&self, socket: UdpSockRef<'_>, transmits: &[Transmit]) -> Result<usize, io::Error> {
         let mut sent = 0;
         for transmit in transmits {
             match socket.0.send_to(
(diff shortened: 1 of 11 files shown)


quinn/src/runtime.rs
```diff
@@ -58,6 +54,16 @@ pub trait AsyncUdpSocket: Send + Debug + 'static {
     fn may_fragment(&self) -> bool {
         true
     }
+
+    /// Maximum number of datagrams that a [`Transmit`] may encode
```
djc, line 58: Style nit: should these go before or after `may_fragment()`?

quinn/src/runtime/async_std.rs
```diff
@@ -23,10 +24,11 @@ impl Runtime for AsyncStdRuntime {
         async_std::task::spawn(future);
     }
 
-    fn wrap_udp_socket(&self, sock: std::net::UdpSocket) -> io::Result<Box<dyn AsyncUdpSocket>> {
+    fn wrap_udp_socket(&self, sock: std::net::UdpSocket) -> io::Result<Arc<dyn AsyncUdpSocket>> {
         udp::UdpSocketState::configure((&sock).into())?;
-        Ok(Box::new(UdpSocket {
+        Ok(Arc::new(UdpSocket {
```
djc, line 29: Style suggestion (unrelated to the change): maybe `.into()` instead of `Arc::new()`?

quinn/src/runtime/async_std.rs
```diff
@@ -23,10 +24,11 @@ impl Runtime for AsyncStdRuntime {
...
             io: Async::new(sock)?,
+            state: udp::UdpState::new(),
```
djc, line 31: So, should we merge `udp::UdpState` and `udp::UdpSocketState` as well?

djc (approved):

djc:
Ohh, nice!

quinn/src/runtime.rs
Ralith: As far as I'm concerned they're getters with morally similar roles so the ordering isn't significant. Do you have a preference?

quinn/src/runtime/async_std.rs
Ralith: That doesn't work, because we're relying on unsizing to coerce `Arc<UdpSocket>` to `Arc<dyn AsyncUdpSocket>`, and the latter doesn't implement `From`. I like the explicitness, anyway.

quinn/src/runtime/async_std.rs
Ralith: Turned out great; added to the PR.

quinn-udp/src/windows.rs
```diff
@@ -137,6 +132,25 @@ impl UdpSocketState {
         };
         Ok(1)
     }
+
+    /// The maximum amount of segments which can be transmitted if a platform
+    /// supports Generic Send Offload (GSO).
+    ///
+    /// This is 1 if the platform doesn't support GSO. Subject to change if errors are detected
```
djc, line 139: The "Subject to change" sentence doesn't really make sense in this context, I think?

quinn/src/runtime.rs
djc: I'd judge getters that yield a number as trivially higher-complexity than a getter that yields `bool` so just for consistency I'd probably move these ones above `may_fragment()`. But that's definitely the nittest of nits.

djc (approved):

quinn-udp/src/windows.rs
Ralith: The intention is to describe the behavior of the platform-agnostic interface; the goal of quinn-udp is for the caller to not have to account for the quirks of any one platform. Portable code shouldn't assume this returns a constant, so the docs say as much.

PR #1613, opened 2023-07-23
Mitigate optimistic ACK attacks

Ralith:


quinn-proto/src/connection/spaces.rs
```diff
@@ -476,6 +479,86 @@ impl PendingAcks {
     }
 }
 
+/// Helper for mitigating opportunistic ACK attacks
```
djc, line 482: Is this approach new? What do other implementations do here? Should we seek some consensus/discussion in the context of the QUIC WG?

quinn-proto/src/connection/mod.rs
```diff
@@ -784,11 +786,12 @@ impl Connection {
         // Send MTU probe if necessary
         if buf.is_empty() && self.state.is_established() {
             let space_id = SpaceId::Data;
-            let probe_size = match self
-                .path
-                .mtud
-                .poll_transmit(now, self.spaces[space_id].next_packet_number)
-            {
+            // FIXME: Doesn't account for skipped PNs
```
djc, line 789: How big of an issue is this?

Ralith: Whoops, stale comment, that was fixed by the new `peek` call.

quinn-proto/src/connection/spaces.rs
Ralith: Skipping packets is [the recommendation in RFC 9000](https://www.rfc-editor.org/rfc/rfc9000.html#name-optimistic-ack-attack). I've updated the comment to cite this. The exponential distribution in particular is cribbed from quic-go.

quinn-proto/src/connection/spaces.rs
```diff
@@ -476,6 +479,88 @@ impl PendingAcks {
...
+        // First skipped PN is in 0..64
+        let exponent = 6;
+        Self {
+            next_skipped_packet_number: rng.gen_range(0..2u64.saturating_pow(exponent)),
+            prev_skipped_packet_number: None,
+            exponent,
+        }
+    }
+
+    pub(super) fn peek_packet_number(&self, space_id: SpaceId, space: &PacketSpace) -> u64 {
```
djc, line 518: So for the methods here, you've made the trade-off to hide `PacketSpace::get_tx_number()` from the `super` module in favor of passing in `SpaceId` here and checking it against `SpaceId::Data`. That gains us some compile-time safety from forcing calls to `get_tx_number()` to happen through `OpportunisticAckDetector`, but makes the code here a bit more complex due to having to pass in the `space_id`. Given that `PacketSpace::next_packet_number` is `pub(super)` anyway, maybe it isn't so bad to leave the decision to access it directly to callers from the `Connection` impl and avoid passing the argument?

quinn-proto/src/connection/spaces.rs
```diff
@@ -476,6 +479,88 @@ impl PendingAcks {
...
+/// Skipped packet numbers occur only in the application data space (where costly transfers might
+/// take place) and are distributed exponentially to reflect the reduced likelihood and impact of
+/// bad behavior from a peer that has been well-behaved for an extended period.
+///
+/// ACKs for packet numbers that have not yet been allocated are also a transport error, but an
+/// attacker with knowledge of the congestion control algorithm in use could time falsified ACKs to
+/// arrive after the packets they reference are sent.
+///
+/// [optimistic ACK attacks]: https://www.rfc-editor.org/rfc/rfc9000.html#name-optimistic-ack-attack
+pub(super) struct OptimisticAckDetector {
```
djc, line 498: Naming considerations: so the name here describes the purpose (detecting opportunistic ACKs) well but doesn't really describe the mechanism (skipping packet numbers). In turn, this suggests making the method names fairly verbose (adding the `_packet_number` suffix). Do you think that's the right trade-off?

djc (approved):

quinn-proto/src/connection/spaces.rs
```diff
@@ -476,6 +479,88 @@ impl PendingAcks {
...
+        rng: &mut (impl Rng + ?Sized),
+        space_id: SpaceId,
+        space: &mut PacketSpace,
+    ) -> u64 {
+        let n = space.get_tx_number();
+        if space_id != SpaceId::Data || n != self.next_skipped_packet_number {
+            return n;
+        }
+
+        trace!("skipping pn {}", n);
```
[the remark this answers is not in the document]
Ralith: Done.

quinn-proto/src/connection/spaces.rs
Ralith: Good thought; renamed to `PacketNumberFilter`.

quinn-proto/src/connection/spaces.rs
Ralith: This was a break-even on LoC, but revealed that `peek` is only called in the data space anyway, and narrows interfaces, so I think it's a win :+1:

PR #1614, opened 2023-07-23
Perform a key update early in a connection's lifetime

Ralith:
Inspired by quic-go's behavior as documented in https://mailarchive.ietf.org/arch/msg/quic/glen5JXVLU9BhmIB73E2UOUOp6M/.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -148,6 +148,8 @@ pub struct Connection {
     /// Set if 0-RTT is supported, then cleared when no longer needed.
     zero_rtt_crypto: Option<ZeroRttCrypto>,
     key_phase: bool,
+    /// How many packets are in the current key phase. Used only for `Data` space.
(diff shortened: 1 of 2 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -3477,6 +3491,11 @@ const MIN_PACKET_SPACE: usize = 40;
 /// that numbers around 10 are a good compromise.
 const MAX_TRANSMIT_SEGMENTS: usize = 10;
 
+/// Perform key updates this many packets before the AEAD confidentiality limit.
+///
+/// Chosen arbitrarily, intended to be large enough to prevent spurious connection loss.
+const KEY_UPDATE_MARGIN: u64 = 10000;
```
djc, line 3497:
```suggestion
const KEY_UPDATE_MARGIN: u64 = 10_000;
```

djc (approved):
Cool!

PR #1615, opened 2023-07-24
Reduce default maximum UDP payload size to the ethernet MTU

### LINKED ISSUE #1608: `Endpoint`s with default config allocate too much memory on creation

Ralith:
Drastically reduces the size and sparsity of the quinn receive buffer for no performance cost in typical applications.

Fixes #1608. Alternative to #1611 with less impact on endpoints that are heavily loaded by many peers. Not certain which we should prefer yet; `recvmmsg` might not be buying us all that much, but neither is supporting jumbo frames (outside of loopback). Maybe we should even do both!

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -567,7 +567,7 @@ impl EndpointConfig {
             || Box::<RandomConnectionIdGenerator>::default();
         Self {
             reset_key,
-            max_udp_payload_size: MAX_UDP_PAYLOAD.into(), // See RFC 9000 (https://www.rfc-editor.org/rfc/rfc9000.html#section-18.2-4.10.1)
+            max_udp_payload_size: (1500u32 - 28).into(), // Ethernet MTU minus IP + UDP headers
             connection_id_generator_factory: Arc::new(cid_factory),
             supported_versions: DEFAULT_SUPPORTED_VERSIONS.to_vec(),
             grease_quic_bit: true,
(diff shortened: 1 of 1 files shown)


djc (approved):
Makes sense to me.

lijunwangs:
Hi @djc can we backport this to 0.10.x and release a new version on that stream?

djc:
I think that makes sense, yes.

lijunwangs:
> I think that makes sense, yes.

Okay, I will create a backport PR

PR #1616, opened 2023-07-26
Preserve the tracing span in the EndpointDriver task

flub:
The EndpointDriver is spawned in a task.  Unless the span is copied into this task the current span is lost.

Copying the current span means any tracing calls inside the driver, from e.g. a custom AsyncUdpSocket, are preserved in the span.  This makes it easy to use different parent spans for different endpoints and have log messages appear with their parent span.  Otherwise these log messages are not attached to a span.

----

An alternative is to create a new span using `tracing::info_span!("endpointdriver")` instead of copying the current span.  This new span would also be correctly attached to the current span as a child.

### DIFF
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -20,6 +20,7 @@ use proto::{
 };
 use rustc_hash::FxHashMap;
 use tokio::sync::{futures::Notified, mpsc, Notify};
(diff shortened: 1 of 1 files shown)


djc (approved):

Ralith (approved):
Thanks, I've wanted this!

PR #1620, opened 2023-07-31
Backport to 0.10.x: Reduce default maximum UDP payload size to the ethernet MTU

lijunwangs:
Drastically reduces the size and sparsity of the quinn receive buffer for no performance cost in typical applications.

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -495,7 +495,7 @@ impl EndpointConfig {
             || Box::<RandomConnectionIdGenerator>::default();
         Self {
             reset_key,
-            max_udp_payload_size: MAX_UDP_PAYLOAD.into(), // See RFC 9000 (https://www.rfc-editor.org/rfc/rfc9000.html#section-18.2-4.10.1)
+            max_udp_payload_size: (1500u32 - 28).into(), // Ethernet MTU minus IP + UDP headers
             connection_id_generator_factory: Arc::new(cid_factory),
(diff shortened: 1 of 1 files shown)


lijunwangs:
@Ralith @djc This is backport for 89ff085cf5d1e0a3c52eaedf76e79cd34a85bf4f into 0.10.x

Ralith (approved):

lijunwangs:
Hi @Ralith and @djc, as this issue is fixed in 0.10.x -- can we publish a new version (0.10.2)?

djc:
@Ralith how do you feel about releasing this in a semver-compatible release? I suppose it has the potential to break users who were relying on the new behavior (though that might only be the party who asked us to implement it in the first place).

Ralith:
> though that might only be the party who asked us to implement it in the first place

Were they using an extra-large interface MTU?

I don't think this is breaking. At worst it will be a fixable performance regression for anyone who was employing jumbo frames or similar, and I'd be a bit surprised if anyone actually was.

djc:
Hmm, I guess it was just from #1529, so there wasn't really a user pushing this.

djc:
Published quinn-proto 0.10.2 and quinn 0.10.2.

lijunwangs:
> Published quinn-proto 0.10.2 and quinn 0.10.2.

Thanks @djc can we add a tag in the branch as well?

djc:
Done.

Issue #1621, opened 2023-08-02
Add rebind_with_abstract_socket


zzwxh:
#1432

djc:
Can you add a bit more context about what you need this for/what your use case is?

djc:
I still don't understand what problem you're trying to solve.

zzwxh:
Converting a synchronous socket to an asynchronous socket is weird, why not just use an asynchronous socket.

zzwxh:
```rust
pub fn tokio_server(
    config: quinn::ServerConfig,
    socket: tokio::net::UdpSocket,
) -> std::io::Result<quinn::Endpoint> {
    quinn::Endpoint::new_with_abstract_socket(
        create_ep_config(),
        Some(config),
        AsyncSocket {
            io: socket,
            inner: quinn::udp::UdpSocketState::new(),
        },
        Arc::new(AsyncRuntime),
    )
}

#[derive(Debug)]
struct AsyncRuntime;

impl quinn::Runtime for AsyncRuntime {
    fn new_timer(&self, i: std::time::Instant) -> std::pin::Pin<Box<dyn quinn::AsyncTimer>> {
        Box::pin(tokio::time::sleep_until(i.into()))
    }

    fn spawn(&self, future: std::pin::Pin<Box<dyn std::future::Future<Output = ()> + Send>>) {
        tokio::spawn(future);
    }

    fn wrap_udp_socket(
        &self,
        _: std::net::UdpSocket,
    ) -> std::io::Result<Box<dyn quinn::AsyncUdpSocket>> {
        unimplemented!()
    }
}
```

zzwxh:
I still can't understand why a protocol library needs to call the system api directly instead of changing the requirement of asynchronous sockets

djc:
tokio's `TcpStream` has an `into_std()` method as well, so if you have a tokio `TcpStream` that you'd like to convert to a `dyn AsyncUdpSocket` I think that should make that easier? IIRC the main reasons Quinn has its own abstraction on top of the system APIs are (a) support for Explicit Congestion Notifications (ECN) and (b) performance.

zzwxh:
If you want to support ecn, why not extend the AsyncUdpSocket trait

zzwxh:
> ```rust
> pub fn tokio_server(
>     config: quinn::ServerConfig,
>     socket: tokio::net::UdpSocket,
> ) -> std::io::Result<quinn::Endpoint> {
[...]

PR #1622, opened 2023-08-02
remove wrap_udp_socket

zzwxh:
#1621


Ralith:
I'm not interested in replacing `SocketAddr` arguments in convenience APIs with a much less obvious and more complex abstract structure. Users who need to use that structure can freely use existing lower-level APIs.

Issue #1624, opened 2023-08-09
Protocol not available (os error 42)


jieliangma:
version: 0.10.2
device: iPhone SE2, iOS 13.6.1
`quinn::Endpoint::client("0.0.0.0:0".parse()?)?` got an error: Protocol not available (os error 42)

djc:
That's pretty strange. Can you dig into the code inside `Endpoint::client()` and see which call is failing? Is it the `UdpSocket::bind()`, something in `quinn_udp::unix::init()` (called by `wrap_with_socket()`)? We have had people use the library on iOS before so it should work but maybe the version you have is too old. Have you tried earlier versions?

jieliangma:
in file src/unix.rs 
```
    #[cfg(any(target_os = "freebsd", target_os = "macos", target_os = "ios"))]
    {
        if is_ipv4 {
            set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_DONTFRAG, OPTION_ON)?; //  this line error
        }
    }
```

djc:
Yup, that's it. So we'll probably want to ignore this particular error value here, at least for macOS?

PR #1626, opened 2023-08-09
set may_fragment based on whether setting IP_DONTFRAG fails with ENOPROTOOPT.

jieliangma:
[fix issue](https://github.com/quinn-rs/quinn/issues/1624)

### DIFF
--- a/quinn-udp/src/lib.rs
+++ b/quinn-udp/src/lib.rs
@@ -32,14 +32,6 @@ mod imp;
 
 pub use imp::UdpSocketState;
 
-/// Whether transmitted datagrams might get fragmented by the IP layer
-///
-/// Returns `false` on targets which employ e.g. the `IPV6_DONTFRAG` socket option.
-#[inline]
-pub fn may_fragment() -> bool {
-    imp::may_fragment()
-}
-
 /// Number of UDP packets to send/receive at a time
 pub const BATCH_SIZE: usize = imp::BATCH_SIZE;
 
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -32,6 +32,7 @@ pub struct UdpSocketState {
     last_send_error: Mutex<Instant>,
     max_gso_segments: AtomicUsize,
     gro_segments: usize,
+    may_fragment: bool,
 
     /// True if we have received EINVAL error from `sendmsg` or `sendmmsg` system call at least once.
     ///
@@ -74,34 +75,39 @@ impl UdpSocketState {
             }
         }
 
+        let mut may_fragment = false;
         #[cfg(target_os = "linux")]
         {
             // opportunistically try to enable GRO. See gro::gro_segments().
             let _ = set_socket_option(&*io, libc::SOL_UDP, libc::UDP_GRO, OPTION_ON);
 
             // Forbid IPv4 fragmentation. Set even for IPv6 to account for IPv6 mapped IPv4 addresses.
-            set_socket_option(
+            let result = set_socket_option(
                 &*io,
                 libc::IPPROTO_IP,
                 libc::IP_MTU_DISCOVER,
                 libc::IP_PMTUDISC_PROBE,
-            )?;
+            );
+            may_fragment = is_result_noprotoopt_err(result) || may_fragment;
 
             if is_ipv4 {
                 set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_PKTINFO, OPTION_ON)?;
             } else {
-                set_socket_option(
+                let result = set_socket_option(
                     &*io,
                     libc::IPPROTO_IPV6,
                     libc::IPV6_MTU_DISCOVER,
                     libc::IP_PMTUDISC_PROBE,
-                )?;
+                );
+                may_fragment = is_result_noprotoopt_err(result) || may_fragment;
             }
         }
         #[cfg(any(target_os = "freebsd", target_os = "macos", target_os = "ios"))]
         {
             if is_ipv4 {
-                set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_DONTFRAG, OPTION_ON)?;
+                let result =
+                    set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_DONTFRAG, OPTION_ON);
+                may_fragment = is_result_noprotoopt_err(result) || may_fragment;
             }
         }
         #[cfg(any(target_os = "freebsd", target_os = "macos"))]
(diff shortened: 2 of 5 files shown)


quinn-udp/src/unix.rs
```diff
@@ -164,7 +164,7 @@ fn init(io: SockRef<'_>) -> io::Result<()> {
     #[cfg(any(target_os = "freebsd", target_os = "macos", target_os = "ios"))]
     {
         if is_ipv4 {
-            set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_DONTFRAG, OPTION_ON)?;
+            let _ = set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_DONTFRAG, OPTION_ON);
```
djc, line 167: @Ralith what do you think, should we only ignore this specific error value?

djc (commented):
Possible argument against this PR: iOS 13 (the last version of iOS that didn't support this, assuming iOS releases got support at the same time as macOS) has been out of security support since Sep 2020.

@jieliangma why has this device not been updated to a later OS? It looks like it ought to support iOS 16 (the current latest version) just fine.

quinn-udp/src/unix.rs
Ralith: Yeah, and only on the iOS (and perhaps macOS) versions expected to produce it. Otherwise we're liable to miss any similar future hazards, which could lead to degraded performance due to incorrect PMTU behavior.

djc:
> In short, this PR can't be merged without adding dynamic iOS version detection logic to `may_fragment()`. For robustness, we should also only ignore the specific expected error, and only in environments specifically expected to produce it.

If we try to set `IP_DONTFRAG` in `init()`, I suppose we can store the `may_fragment` bit as state after that, so that we don't need to interrogate the socket for every call to `may_fragment()`?

Ralith:
Good point, that's also a lot easier and less fragile than version checks. `may_fragment` should then probably be a method on `UdpSocketState`.

Ralith:
Thanks for the revision! Please split the cosmetic refactoring commits out from the semantic changes so they can be easily reviewed.

jieliangma:
> Thanks for the revision! Please split the cosmetic refactoring commits out from the semantic changes so they can be easily reviewed.

done

quinn-udp/src/lib.rs
```diff
@@ -32,14 +32,6 @@ mod imp;
 
 pub use imp::UdpSocketState;
 
-/// Whether transmitted datagrams might get fragmented by the IP layer
-///
-/// Returns `false` on targets which employ e.g. the `IPV6_DONTFRAG` socket option.
-#[inline]
-pub fn may_fragment() -> bool {
```
djc, line 39: As suggested in @Ralith's comment, please split this change (moving the `may_fragment()` function into the `UdpSocketState` impl) into a separate commit from the changes that actually make the answer stateful.

quinn-udp/src/unix.rs
```diff
@@ -33,6 +33,7 @@ pub struct UdpSocketState {
     last_send_error: Mutex<Instant>,
     max_gso_segments: AtomicUsize,
     gro_segments: usize,
+    ip_dontfrag_failing: bool,
```
djc, line 36: I think we should just call this `may_fragment`.

quinn-udp/src/unix.rs
```diff
@@ -49,12 +50,13 @@ impl UdpSocketState {
...
             gro_segments: gro::gro_segments(),
+            ip_dontfrag_failing: false,
             sendmsg_einval: AtomicBool::new(false),
         }
     }
 
-    pub fn configure(sock: UdpSockRef<'_>) -> io::Result<()> {
-        init(sock.0)
+    pub fn configure(&mut self, sock: UdpSockRef<'_>) -> io::Result<()> {
+        init(sock.0, &mut self.ip_dontfrag_failing)
```
djc, line 59: I think it would be cleaner to just run an `io::Result<bool>` from `init()` instead of passing a `&mut bool` into it.

djc:
> In short, this PR can't be merged without adding dynamic iOS version detection logic to `may_fragment()`. For robustness, we should also only ignore the specific expected error, and only in environments specifically expected to produce it.

It seems okay to me to set `may_fragment` based on whether setting `IP_DONTFRAG` fails with `NOPROTOOPT`, without separate implementations for FreeBSD or newer versions of the Apple OSes, do you see any reason?

Ralith:
Yeah, that sounds good to me, since we're no longer relying on an independent judgement in `may_fragment` to line up.

djc:
@jieliangma #1629 refactored part of this code a bit, should make the changes here easier but will require some rebasing or rework.

quinn-udp/src/unix.rs
jieliangma: Thanks for your commit, I have rebased the code.

quinn-udp/src/lib.rs
jieliangma: done

djc: My intent with this comment was to have the change to move `may_fragment()` part of `UdpSocketState` in the first commit. Currently it seems the first commit will just fail to compile, which is definitely not the intention.

quinn-udp/src/unix.rs
```diff
@@ -74,34 +75,42 @@ impl UdpSocketState {
...
 
+        let mut may_fragment: bool = false;
         #[cfg(target_os = "linux")]
         {
             // opportunistically try to enable GRO. See gro::gro_segments().
             let _ = set_socket_option(&*io, libc::SOL_UDP, libc::UDP_GRO, OPTION_ON);
 
             // Forbid IPv4 fragmentation. Set even for IPv6 to account for IPv6 mapped IPv4 addresses.
-            set_socket_option(
+            may_fragment = set_socket_option(
```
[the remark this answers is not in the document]
jieliangma: I added a function is_result_noprotoopt_err to check it.

quinn-udp/src/unix.rs
```diff
@@ -74,34 +75,42 @@ impl UdpSocketState {
             }
         }
 
+        let mut may_fragment: bool = false;
```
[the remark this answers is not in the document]
jieliangma: `:bool` was removed

quinn-udp/src/lib.rs
jieliangma: got it

djc (approved):

djc:
Going to merge this and fix it up a little.

PR #1636, opened 2023-08-20
proto: Allow GSO to be manually disabled

Ralith:


djc (approved):

djc:
Do we need to backport this?

Ralith:
Yep, see https://github.com/quinn-rs/quinn/pull/1638.

PR #1637, opened 2023-08-20
udp: disable GSO on EINVAL

Ralith:
Linux 5.15 was observed to emit an EINVAL when attempting to transmit UDP with GSO over IPv6 on a virtio net interface, which apparently only supports UDP GSO on IPv6 as of Linux 6.2. 6.1 was also observed to emit EINVAL, though only after EIO. In both cases, EINVAL was suppressed by disabling GSO.

Companion change to #1636. I don't fully understand why Linux is returning EINVAL in these cases, so I'm not 100% certain this is the correct fix, but at the very least it won't hurt.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -265,15 +265,15 @@ fn send(
                 }
                 io::ErrorKind::WouldBlock => return Err(e),
                 _ => {
-                    // Some network adapters do not support GSO. Unfortunately, Linux offers no easy way
-                    // for us to detect this short of an I/O error when we try to actually send
-                    // datagrams using it.
+                    // Some network adapters and drivers do not support GSO. Unfortunately, Linux
+                    // offers no easy way for us to detect this short of an EIO or sometimes EINVAL
+                    // when we try to actually send datagrams using it.
                     #[cfg(target_os = "linux")]
-                    if e.raw_os_error() == Some(libc::EIO) {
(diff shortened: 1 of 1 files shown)


quinn-udp/src/unix.rs
```diff
@@ -265,15 +265,16 @@ fn send(
...
                 _ => {
-                    // Some network adapters do not support GSO. Unfortunately, Linux offers no easy way
-                    // for us to detect this short of an I/O error when we try to actually send
-                    // datagrams using it.
+                    // Some network adapters and drivers do not support GSO. Unfortunately, Linux
+                    // offers no easy way for us to detect this short of an EIO or sometimes EINVAL
+                    // when we try to actually send datagrams using it.
                     #[cfg(target_os = "linux")]
-                    if e.raw_os_error() == Some(libc::EIO) {
+                    if e.raw_os_error() == Some(libc::EIO) || e.raw_os_error() == Some(libc::EINVAL)
```
djc, line 272: Is there some spelling here that concisely avoids the duplication and allows us to use `libc::EIO | libc::EINVAL`? Maybe `if let Some(libc::EIO | libc::EINVAL) = e.raw_os_err() {`?

djc (approved):

quinn-udp/src/unix.rs
Ralith: Went with `Some(libc::EIO) | Some(libc::EINVAL)` because it's really hard for me not to read `|` between two integer constants as "bitwise OR".

PR #1641, opened 2023-08-20
proto: Fix double-boxing of congestion::ControllerFactory 

Ralith:
Allows callers who already have a trait object to pass it on without yet more boxing.

This propagates a lesson we've learned a few times at the `quinn` layer.

### DIFF
--- a/quinn-proto/src/congestion.rs
+++ b/quinn-proto/src/congestion.rs
@@ -2,6 +2,7 @@
 
 use crate::connection::RttEstimator;
 use std::any::Any;
+use std::sync::Arc;
 use std::time::Instant;
 
 mod bbr;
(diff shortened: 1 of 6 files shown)


quinn-proto/src/congestion.rs
```diff
@@ -77,7 +78,7 @@ pub trait Controller: Send {
 /// Constructs controllers on demand
 pub trait ControllerFactory {
     /// Construct a fresh `Controller`
-    fn build(&self, now: Instant, current_mtu: u16) -> Box<dyn Controller>;
+    fn build(self: Arc<Self>, now: Instant, current_mtu: u16) -> Box<dyn Controller>;
```
Ralith, line 81: `&Arc<Self>` might be even nicer here, in case a future impl doesn't want to capture a reference to `self`, but that's not (yet?) object safe and all our actual impls want this pattern anyway.

djc (approved):

PR #1642, opened 2023-08-20
proto: add forgotten field to Debug for TransportConfig

Ralith:


quinn-proto/src/config.rs
```diff
@@ -374,6 +374,10 @@ impl fmt::Debug for TransportConfig {
             )
             .field("datagram_send_buffer_size", &self.datagram_send_buffer_size)
             .field("congestion_controller_factory", &"[ opaque ]")
+            .field(
```
djc, line 377: We could destruct `Self` here in order to make the compiler error for not handling all the fields?

djc (approved):

quinn-proto/src/config.rs
Ralith: Good idea. We were, in fact, missing several fields.

PR #1644, opened 2023-08-22
Replace Matrix link with Discord

Ralith:
Discord seems to be more popular in the Rust community, and bridging between the two is unstable.


Ralith:
Whoops, didn't see https://github.com/quinn-rs/quinn/pull/1643.

PR #1646, opened 2023-08-23
proto: detect stateless resets in authed and unprotected packets

Ralith:
RFC9000 says: endpoints MUST treat any packet ending in a valid stateless reset token as a Stateless Reset

Previously, we did not detect stateless resets that appeared to be unprotected packets (e.g. Retry or Version Negotiation) or which were successfully decrypted (astronomically unlikely with TLS, but possible with custom cryptographic layers).

We didn't detect this before because all current standard QUIC versions only issue stateless resets as short-header packets.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -2096,29 +2096,28 @@ impl Connection {
                 .map(move |number| (packet, number)),
         };
         let result = match decrypted {
+            _ if stateless_reset => {
+                debug!("got stateless reset");
+                Err(ConnectionError::Reset)
+            }
             Err(Some(e)) => {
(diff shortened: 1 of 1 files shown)


djc (approved):

Ralith:
Straightforward but not quite trivial. We'd need to set up a a server with a known `ResetKey`, connect to it, and then inject a hand-crafted reset packet to the client. Would be a good follow-up.

jeromegn:
Can this be backported + released for 0.10.x? :)

Ralith:
Backport/bump PR: https://github.com/quinn-rs/quinn/pull/1647

Ralith:
Published in 0.10.4.

Issue #1433, opened 2022-10-26
Performance under tokio multi_thread vs current_thread runtime


shaun-cox:
We've built a router of sorts using Quinn, and so far, have been thrilled with how easy it's been to get up and going with Quinn's nice API.

Now we're in the phase of performance measuring and our metric of interest is "megabits per core", or how much data can we move through our router per unit of CPU consumption.

Our use case is not a typical "bulk throughput" scenario where we have all the data up front and need to flow it all quickly and measure goodput. Our scenario is small packets sent by clients at paced intervals (e.g. 160 byte chunks sent every 2.5ms, such as by an Audio codec's output). The router receives many
streams of this nature and simply relays the chunks out on some other QUIC send streams.

The surprising anomaly we've discovered is that, for the identical offered load to the router, if we switch to using `#[tokio::main(flavor = "current_thread")]` from `#[tokio::main(flavor = "multi_thread", worker_threads = 4]`, we see a reduction in CPU consumed by about 71%. (1172 millicores to 335 millicores)

At first, we thought this was mainly due to the appearance of `Mutex::lock_contended` as the top-most hit showing in `perf`, but upon closer inspection, we've also noticed that when using the multi-threaded tokio runtime, Quinn ends up generating about 3x the number of outgoing UDP datagrams from the router.

We're currently able to achieve only about 466 Mbits/core in this single-threaded runtime scenario, which seems pretty low. We'd envisioned running pods with a 4 core CPU limit, and getting higher throughput, but if we do, we only achieve 133 Mbits/core, which is way less efficient and more costly.

We're wondering if anyone can offer an explanation for the 3x increase in outgoing UDP datagrams just by switching runtimes. We're assuming it has to do with fewer opportunities for internal batching or realizing that multiple stream frames or QUIC packets can be coalesced into the same UDP datagram?

We're also curious about the choice of having to acquire that `Mutex` protecting `state` in `quinn::ConnectionInner` for every `poll_read` and `poll_write` on any stream associated with a connection. Isn't this an immediate scalability killer if the tasks that read/write from Quinn streams get spread out across the multi-threaded runtime's thread pool?

Thanks.

Ralith:
> have been thrilled with how easy it's been to get up and going with Quinn's nice API.

Glad to hear it!

> Our scenario is small packets sent by clients at paced intervals (e.g. 160 byte chunks sent every 2.5ms, such as by an Audio codec's output)

[...]

Issue #599, opened 2020-01-15
Avoiding unbounded channels


DemiMarie-parity:
Is it possible to avoid using unbounded channels?  My attempts to do so have lead to intermittent deadlocks that I have not successfully been able to debug.

Ralith:
This should work fine. The unbounded channels are a matter of implementation convenience, not necessity. It's painful to wrap asynchronous operations in hand-written futures, but we'll likely move away from them whenever we migrate to async/await.

PR #1650, opened 2023-08-26
Communicate endpoint events in -proto w/ an intrusive queue

Ralith:
Work towards #1576, https://github.com/quinn-rs/quinn/issues/599.

This replaces the requirement for callers to pass `EndpointEvent`s around with shared memory managed by the proto layer, using a purpose-built lock-free intrusive mpsc queue called `SharedList`. This simplifies the proto API and may be more efficient, although endpoint events are not typically on a hot path.

Intensive review is required for the implementation of `SharedList`, which contains internal unsafety and relies on careful use of memory orderings for correctness. I couldn't find an off-the-shelf, well-audited data structure with similar semantics.

Alternative approaches to this API change include a std mpsc channel and the pre-existing `EndpointEvent` enum, or an mpsc channel of `Arc<EndpointEvents>` with an additional `AtomicBool` flag to minimize redundant messaging. These would avoid rolling our own data structure, at the cost using a more reviewed but needlessly complex and less specialized data structure. Further, `SharedList` may prove useful elsewhere within quinn: a similar queue-once MPSC pattern exists in streams, and might be leveraged to increase application parallelism when performing stream I/O (see discussion in https://github.com/quinn-rs/quinn/issues/1433).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -58,7 +58,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: dtolnay/rust-toolchain@1.63.0
+      - uses: dtolnay/rust-toolchain@1.67.0
       - uses: Swatinem/rust-cache@v2
       - run: cargo check --lib --all-features -p quinn-udp -p quinn-proto -p quinn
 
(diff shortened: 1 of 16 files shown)


Ralith:
In retrospect this might be more gracefully posed by first refactoring to use an internal mpsc queue, then replacing the mpsc queue with the custom one. Let me know if that's worth investing time in.

Ralith:
Closing in favor of #1726, though if we continue in that direction we may eventually want to port the intrusive queue over.

PR #1652, opened 2023-08-31
silent an info message related to ack frequency.

stormshield-damiend:
Change an info into debug to silent a message that show too frequently.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -799,7 +799,7 @@ impl Connection {
                 // Request an immediate ACK to ensure we receive one within RTT, as recommended by
                 // the acknowledgement frequency draft
                 self.spaces[SpaceId::Data].immediate_ack_pending = true;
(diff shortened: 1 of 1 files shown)


djc (approved):
Nit: maybe "silent" -> "Silence" in the commit message?

Thanks!

quinn-proto/src/connection/mod.rs
```diff
@@ -799,7 +799,7 @@ impl Connection {
                 // Request an immediate ACK to ensure we receive one within RTT, as recommended by
                 // the acknowledgement frequency draft
                 self.spaces[SpaceId::Data].immediate_ack_pending = true;
-                tracing::info!(
+                tracing::debug!(
```
Ralith, line 802:
```suggestion
                tracing::trace!(
```

Ralith (approved):
Thanks; this definitely wasn't intended. We should probably even drop it further -- `debug` isn't intended to have a lot of routine noise.

PR #1654, opened 2023-09-02
Clamp per-RTT ACK timeouts to at most once per 10ms

Ralith:
Avoids unreasonably high ACK rates on LANs/loopback.


djc (approved):

djc:
Nice work!

Ralith:
I think #1656 is actually a better way to go here. Will open a new PR with just the refactoring.

PR #1666, opened 2023-09-20
Pass byte read count to FinishEarly

### LINKED ISSUE #1663: RecvStream::read_exact() API insufficient
In Quinn, [`RecvStream::read_exact()`](https://docs.rs/quinn/latest/quinn/struct.RecvStream.html#method.read_exact) returns `Result<(), ReadExactError>`.

To detect end-of-stream, we can use [`ReadExactError::FinishedEarly`](https://docs.rs/quinn/latest/quinn/enum.ReadExactError.html):

> `FinishedEarly`
> The stream finished before all bytes were read

But then, we are unable to distinguish between end-of-stream at `read_exact()` boundaries or anywhere else.

For example, if I read packets of fixed-size 512 bytes, an EOS at 0, 512, 1024, 1536 or 2048… is correct (there is no more packets). But an EOS at 153, 781 or 1888 bytes is an error (partial packet), but we could not tell from `FinishedEarly` alone.

By contrast, in Tokio, [`AsyncReadExt::read_exact()`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#method.read_exact) returns `io::Result<usize>` (so the `usize` allows to distinguish both cases).

rom1v:
If `read_exact()` finishes early (end of stream), report the number of bytes actually read.

(note: this breaks the API)

Fixes #1663

### DIFF
--- a/quinn/src/recv_stream.rs
+++ b/quinn/src/recv_stream.rs
@@ -537,12 +537,14 @@ impl<'a> Future for ReadExact<'a> {
     type Output = Result<(), ReadExactError>;
     fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
         let this = self.get_mut();
-        let mut remaining = this.buf.remaining();
+        let total = this.buf.remaining();
+        let mut remaining = total;
         while remaining > 0 {
             ready!(this.stream.poll_read(cx, &mut this.buf))?;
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Thanks!

Ralith (approved):

PR #1668, opened 2023-09-20
Backport #1667 to 0.9.x

djc:


djc:
I'm not motivated enough to backport clippy fixes so I'd propose we just run with this.

Ralith (approved):

Ralith:
LGTM. Maybe we should disable linting outside of `main`...

djc:
Published quinn-proto 0.9.5.

PR #1669, opened 2023-09-20
Backport #1667 to 0.10.x

djc:


djc:
I'm not motivated enough to backport clippy fixes so I'd propose we just run with this.

Ralith (approved):

djc:
Published quinn-proto 0.10.5.

PR #1638, opened 2023-08-20
Backport gso mitigations

Ralith:
Backport of #1636 and #1637 to the release branch. Should block on review of those.

PR #1671, opened 2023-09-21
backport to 0.9: Allow GSO to be manually disabled

grooviegermanikus:
Hi,
we love quinn but we recently had a hard time with GSO and fly.io hosting and quinn 0.9.4!

This errors are piling up in the logs:
> sendmsg error: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }, Transmit: { destination: [dead:0:beef:a7b:17b:0101:2345:2]:11111, src_ip: None, enc: Some(Ect0), len: 1291, segment_size: Some(1200) }

After some research we concluded that quinn's GSO detection returns a false-positive - at least on fly.io machines!
Our reasoning:
1. The problem went away by disabling GSO with patch a06838.
2. fly.io support confirmed GSO is not supported on their platform!

### Proposed Solution
short term: backport a06838 to allow manual disabling GSO
long term: backport a fix for the feature detection

maybe the solution should be ported to 0.10 too

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -46,6 +46,8 @@ pub struct TransportConfig {
     pub(crate) datagram_send_buffer_size: usize,
 
     pub(crate) congestion_controller_factory: Box<dyn congestion::ControllerFactory + Send + Sync>,
+
+    pub(crate) enable_segmentation_offload: bool,
 }
 
 impl TransportConfig {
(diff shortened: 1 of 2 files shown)


djc:
Why are you not on 0.10 yet?

Ralith:
As an alternative workaround, did you try `ethtool -K $INTERFACE tx-udp-segmentation off`?

grooviegermanikus:
> As an alternative workaround, did you try `ethtool -K $INTERFACE tx-udp-segmentation off`?

Hi, let me check with ops!

yet I guess on PaaS-environments like fly.io that's tedious workaround. the deployment model on fly.io is to provide a docker image. access to OS level is possible but limited AFAIK.

djc (approved):

djc:
> maybe the solution should be ported to 0.10 too

This was done already, in #1638.

grooviegermanikus:
> Why are you not on 0.10 yet?

we depend on a component we don't control

djc:
> we depend on a component we don't control

Is that something public? Would like to have more information.

grooviegermanikus:
> > we depend on a component we don't control
> 
> Is that something public? Would like to have more information.

solana 1.16.x:
here https://github.com/solana-labs/solana/blob/c932953e7bd8f7db10a1c74c6ad29dd809307c70/Cargo.toml#L258
(master is on 0.10)

djc:
Ahh, okay, so released versions of Solana require 0.9 (master is on 0.10) and you depend on Solana. Thanks for the clarification!

djc:
If you submit a PR to bump the version on the 0.9.x branch I can publish a release right after that.

Ralith:
> yet I guess on PaaS-environments like fly.io that's tedious workaround. the deployment model on fly.io is to provide a docker image. access to OS level is possible but limited AFAIK.

I'm not much good at docker, but I think you can arrange for it to run a script on startup, which could include an `ethtool` invocation before invoking your actual service. This might be especially helpful if the quinn use is hidden in a dependency.

grooviegermanikus:
@djc [here's](https://github.com/quinn-rs/quinn/pull/1675) the PR for version bump
Thanks a lot !

PR #1675, opened 2023-10-04
Bump quinn-proto version to 0.9.6

grooviegermanikus:
includes:
* [opt-out for GSO](https://github.com/quinn-rs/quinn/pull/1671)

### DIFF
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quinn-proto"
-version = "0.9.5"
+version = "0.9.6"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/quinn-rs/quinn"
 description = "State machine for the QUIC transport protocol"


djc:
What's the point here? The minimum bound for quinn-proto already allows 0.9.4 which is semver-compatible with 0.9.5.

grooviegermanikus:
you mean that I should rather point to Quinn 0.9.4 and quinn-proto 0.9.5?

I expected that Quinn 0.9 5 without proto should do it.

or am I missing sth?

regards!

grooviegermanikus:
tried this:
` cargo update -p quinn-proto`

gives:
>     Updating quinn-proto v0.9.4 -> v0.9.5

yet this does not compile:
`transport_config.enable_segmentation_offload(false)`

djc:
Yes, you need to bump the quinn-proto version to 0.9.6 because we made another 0.9.5 release without this since my previous suggestion. No need to bump the quinn version though.

grooviegermanikus:
oops failed to rename the branch

djc (approved):

djc:
It looks like the FreeBSD failure was a fluke, and the clippy job has been failing on 0.9.x for a while.

djc:
Published 0.9.6.

PR #1682, opened 2023-10-12
Do not pre-allocate memory on streams

### LINKED ISSUE #1677: Don't pre-allocate uni/bi stream
I've come to realize high memory usage might be caused by pre-allocated streams as set by `max_concurrent_bidi_streams` and `max_concurrent_uni_streams`.

If it was possible to lazily allocate, it would use far less memory on startup.

jeromegn:
This PR wraps `Recv` and `Send` streams in `Option<Box<T>>` to prevent significant memory pre-allocation.

Still a few things to verify / do:
- [ ] Make sure there's never a `None` value when it should be `Some`
- [ ] Validate the logic here, not sure when to insert a `None` or a `Some` exactly...

(Hopefully) Fixes #1677

### DIFF
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -244,8 +244,9 @@ impl<'a> SendStream<'a> {
 
     /// Check if this stream was stopped, get the reason if it was
     pub fn stopped(&mut self) -> Result<Option<VarInt>, UnknownStream> {
-        match self.state.send.get(&self.id).and_then(|s| s.as_ref()) {
-            Some(s) => Ok(s.stop_reason),
+        match self.state.send.get(&self.id).as_ref() {
+            Some(Some(s)) => Ok(s.stop_reason),
+            Some(None) => Ok(None),
             None => Err(UnknownStream { _private: () }),
         }
     }
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -129,10 +129,13 @@ impl<'a> RecvStream<'a> {
     /// Discards unread data and notifies the peer to stop transmitting. Once stopped, further
     /// attempts to operate on a stream will yield `UnknownStream` errors.
     pub fn stop(&mut self, error_code: VarInt) -> Result<(), UnknownStream> {
-        let stream = match self.state.recv.get_mut(&self.id).and_then(|s| s.as_mut()) {
-            Some(s) => s,
-            None => return Err(UnknownStream { _private: () }),
-        };
+        let stream =
+            match self.state.recv.get_mut(&self.id).map(|s| {
+                s.get_or_insert_with(|| Box::new(Recv::new(self.state.stream_receive_window)))
+            }) {
+                Some(s) => s,
+                None => return Err(UnknownStream { _private: () }),
+            };
 
         let (read_credits, stop_sending) = stream.stop()?;
         if stop_sending.should_transmit() {
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -210,7 +210,6 @@ impl<'a> SendStream<'a> {
 
         let limit = self.state.write_limit();
 
-        // annoying to have to do this pre-emptively
         let max_send_data = self.state.max_send_data(&self.id);
 
         let stream = self
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -7,8 +7,10 @@ use bytes::Bytes;
 use thiserror::Error;
 use tracing::trace;
 
+use self::state::get_or_insert_recv;
+
 use super::spaces::{Retransmits, ThinRetransmits};
-use crate::{frame, Dir, StreamId, VarInt};
+use crate::{connection::streams::state::get_or_insert_send, frame, Dir, StreamId, VarInt};
 
 mod recv;
 use recv::Recv;
@@ -133,9 +135,7 @@ impl<'a> RecvStream<'a> {
             hash_map::Entry::Occupied(s) => s,
             hash_map::Entry::Vacant(_) => return Err(UnknownStream { _private: () }),
         };
-        let stream = entry
-            .get_mut()
-            .get_or_insert_with(|| Box::new(Recv::new(self.state.stream_receive_window)));
+        let stream = get_or_insert_recv(self.state.stream_receive_window)(entry.get_mut());
 
         let (read_credits, stop_sending) = stream.stop()?;
         if stop_sending.should_transmit() {
@@ -216,7 +216,7 @@ impl<'a> SendStream<'a> {
             .state
             .send
             .get_mut(&self.id)
-            .map(|s| s.get_or_insert_with(|| Box::new(Send::new(max_send_data))))
+            .map(get_or_insert_send(max_send_data))
             .ok_or(WriteError::UnknownStream)?;
 
         if limit == 0 {
@@ -262,7 +262,7 @@ impl<'a> SendStream<'a> {
             .state
             .send
             .get_mut(&self.id)
-            .map(|s| s.get_or_insert_with(|| Box::new(Send::new(max_send_data))))
+            .map(get_or_insert_send(max_send_data))
             .ok_or(FinishError::UnknownStream)?;
 
         let was_pending = stream.is_pending();
@@ -284,7 +284,7 @@ impl<'a> SendStream<'a> {
             .state
             .send
             .get_mut(&self.id)
-            .map(|s| s.get_or_insert_with(|| Box::new(Send::new(max_send_data))))
+            .map(get_or_insert_send(max_send_data))
             .ok_or(UnknownStream { _private: () })?;
 
         if matches!(stream.state, SendState::ResetSent) {
@@ -313,7 +313,7 @@ impl<'a> SendStream<'a> {
             .state
             .send
             .get_mut(&self.id)
-            .map(|s| s.get_or_insert_with(|| Box::new(Send::new(max_send_data))))
+            .map(get_or_insert_send(max_send_data))
             .ok_or(UnknownStream { _private: () })?;
 
         stream.priority = priority;
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -1,6 +1,6 @@
 use std::{
     cell::RefCell,
-    collections::{hash_map, BinaryHeap, VecDeque},
+    collections::{BinaryHeap, VecDeque},
 };
 
 use bytes::Bytes;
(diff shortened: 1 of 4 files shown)


jeromegn:
I have tested this in [corrosion](https://github.com/superfly/corrosion) and it significantly reduced memory usage. As far as I can tell everything is working as intended but I'd like somebody to thoroughly review it.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -129,11 +129,13 @@ impl<'a> RecvStream<'a> {
...
-            hash_map::Entry::Vacant(_) => return Err(UnknownStream { _private: () }),
-        };
-        let stream = entry.get_mut();
+        let stream =
+            match self.state.recv.get_mut(&self.id).map(|s| {
+                s.get_or_insert_with(|| Box::new(Recv::new(self.state.stream_receive_window)))
+            }) {
+                Some(s) => s,
+                None => return Err(UnknownStream { _private: () }),
+            };
```
Ralith, line 138: It's not clear to me why the old `Entry` logic had to go away here. Couldn't we slap a `let stream = stream,get_or_insert_with(...)` after the existing code to the same effect?

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -207,11 +209,17 @@ impl<'a> SendStream<'a> {
         }
 
         let limit = self.state.write_limit();
+
+        // annoying to have to do this pre-emptively
+        let max_send_data = self.state.max_send_data(&self.id);
```
Ralith, line 214: This is fine; LLVM will likely push it down into the branch after inlining.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -237,7 +245,7 @@ impl<'a> SendStream<'a> {
 
     /// Check if this stream was stopped, get the reason if it was
     pub fn stopped(&mut self) -> Result<Option<VarInt>, UnknownStream> {
-        match self.state.send.get(&self.id) {
+        match self.state.send.get(&self.id).and_then(|s| s.as_ref()) {
```
Ralith, line 248: This is incorrect. We should return `Ok(None)` if the stream is unallocated. Otherwise, it becomes impossible to distinguish open but unallocated send streams from closed and forgotten ones.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -253,6 +261,7 @@ impl<'a> SendStream<'a> {
             .state
             .send
             .get_mut(&self.id)
+            .and_then(|s| s.as_mut())
```
Ralith, line 264: This should allocate the stream and continue, not fail.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -273,6 +282,7 @@ impl<'a> SendStream<'a> {
...
```
Ralith, line 285: As above.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -300,6 +310,7 @@ impl<'a> SendStream<'a> {
...
```
Ralith, line 313: As above.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -315,6 +326,7 @@ impl<'a> SendStream<'a> {
             .state
             .send
             .get(&self.id)
+            .and_then(|s| s.as_ref())
```
Ralith, line 329: This should return the default priority, not fail.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -215,14 +215,18 @@ impl<'a> Chunks<'a> {
...
 
-        let mut recv = match entry.get().stopped {
+        let mut recv = match entry
+            .get_mut()
+            .get_or_insert_with(|| Box::new(Recv::new(streams.stream_receive_window)))
+            .stopped
+        {
             true => return Err(ReadableError::UnknownStream),
-            false => entry.remove(),
+            false => entry.remove().unwrap(), // this can't fail at this point
```
Ralith, line 229: Best practice is to say *why* it can't fail (due to `get_or_insert_with` above).

quinn-proto/src/connection/streams/state.rs
```diff
@@ -589,11 +605,11 @@ impl StreamsState {
     }
 
     pub(crate) fn received_ack_of(&mut self, frame: frame::StreamMeta) {
-        let mut entry = match self.send.entry(frame.id) {
-            hash_map::Entry::Vacant(_) => return,
-            hash_map::Entry::Occupied(e) => e,
+        let stream = match self.send.get_mut(&frame.id).and_then(|s| s.as_mut()) {
+            None => return,
+            Some(s) => s,
```
Ralith, line 610: As above, why do we need to get rid of the `Entry` scheme?

quinn-proto/src/connection/streams/state.rs
```diff
@@ -627,7 +643,7 @@ impl StreamsState {
         for dir in Dir::iter() {
             for index in 0..self.next[dir as usize] {
                 let id = StreamId::new(Side::Client, dir, index);
-                let stream = self.send.get_mut(&id).unwrap();
+                let stream = self.send.get_mut(&id).and_then(|s| s.as_mut()).unwrap();
```
Ralith, line 646: It's non-obvious that we'll never have `None`s in the map here. For robustness, let's handle those by `continue`ing rather than panicking, in the spirit of the branch below, since unallocated streams can't have outgoing data.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -797,24 +818,40 @@ impl StreamsState {
...
-                // the perspective of the peer.
-                Dir::Bi if remote => self.initial_max_stream_data_bidi_local,
-                Dir::Bi => self.initial_max_stream_data_bidi_remote,
-            };
-            let stream = Send::new(max_data);
-            assert!(self.send.insert(id, stream).is_none());
+            if remote {
+                assert!(self.send.insert(id, None).is_none());
+            } else {
+                assert!(self
```
Ralith, line 839: Could we get away without eagerly allocating locally-initiated streams? Downstream code might want to e.g. reserve certain stream IDs without using them immediately.

Ralith (changes requested):
Thanks, this looks like a nice improvement to memory use!

quinn-proto/src/connection/streams/mod.rs
jeromegn: A deallocated stream should return `None` from `self.state.send.get`, I'm not deallocating the inner `Option` ever. Maybe I'm misunderstanding though.

Ralith: By "unallocated stream" I mean a stream that exists in `send`, but is associated with a value of `None`, so `get` here returns `Some(&mut None)`. Such streams might be surfaced to the user e.g. after receiving a new remotely-initiated bidirectional stream.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -605,11 +605,16 @@ impl StreamsState {
...
-            Some(s) => s,
+        let max_send_data = self.max_send_data(&frame.id);
+        let mut entry = match self.send.entry(frame.id) {
+            hash_map::Entry::Vacant(_) => return,
+            hash_map::Entry::Occupied(e) => e,
         };
 
+        let stream = entry
+            .get_mut()
+            .get_or_insert_with(|| Box::new(Send::new(max_send_data)));
```
Ralith, line 616: Because we only call this after sending data on this stream, this closure *should* be unreachable. If we did somehow screw that up, then we might hit an underflow below with unpredictable effects down the line, so perhaps a short-circuit or even an unwrap would be better here?

quinn-proto/src/connection/streams/state.rs
```diff
@@ -833,25 +841,25 @@ impl StreamsState {
...
-                    .is_none());
-            }
+            // if remote {
+            assert!(self.send.insert(id, None).is_none());
+            // } else {
+            //     assert!(self
+            //         .send
+            //         .insert(id, Some(Box::new(Send::new(self.max_send_data(&id)))))
+            //         .is_none());
+            // }
```
Ralith, line 851: Were these comments meant to be included?

jeromegn: Deleted them locally but haven't pushed the commit yet, my bad!

quinn-proto/src/connection/streams/state.rs
```diff
@@ -605,15 +605,16 @@ impl StreamsState {
...
         };
 
-        let stream = entry
-            .get_mut()
-            .get_or_insert_with(|| Box::new(Send::new(max_send_data)));
+        // Because we only call this after sending data on this stream,
+        // this closure should be unreachable. If we did somehow screw that up,
+        // then we might hit an underflow below with unpredictable effects down
+        // the line. Best to short-circuit.
+        let stream = entry.get_mut().as_mut().unwrap();
```
Ralith, line 617: By "short-circuit" I meant return early, as was the behavior before.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -610,11 +610,16 @@ impl StreamsState {
             hash_map::Entry::Occupied(e) => e,
         };
 
-        // Because we only call this after sending data on this stream,
-        // this closure should be unreachable. If we did somehow screw that up,
-        // then we might hit an underflow below with unpredictable effects down
-        // the line. Best to short-circuit.
-        let stream = entry.get_mut().as_mut().unwrap();
+        let stream = match entry.get_mut().as_mut() {
+            Some(s) => s,
```
Ralith, line 614: nit: this would be a great case or let-else.

Ralith (approved):
Thanks!

quinn-proto/src/connection/streams/state.rs
```diff
@@ -797,24 +832,26 @@ impl StreamsState {
         expanded
     }
 
+    pub(super) fn max_send_data(&self, id: &StreamId) -> VarInt {
```
djc, line 835: It would be nice to squash all the commits and then extract this as a separate prefix commit.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -249,10 +257,12 @@ impl<'a> SendStream<'a> {
     ///
     /// [`StreamEvent::Finished`]: crate::StreamEvent::Finished
     pub fn finish(&mut self) -> Result<(), FinishError> {
+        let max_send_data = self.state.max_send_data(&self.id);
         let stream = self
             .state
             .send
             .get_mut(&self.id)
+            .map(|s| s.get_or_insert_with(|| Box::new(Send::new(max_send_data))))
```
djc, line 265: I feel like there are a few recurring patterns here that would make good candidates for moving into a method somewhere to raise the level of abstraction. We could address that after merging this, but would be great if you can take a whack at it.

djc (approved):
Nice!

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -216,7 +216,7 @@ impl<'a> SendStream<'a> {
             .state
             .send
             .get_mut(&self.id)
-            .map(|s| s.get_or_insert_with(|| Box::new(Send::new(max_send_data))))
+            .map(get_or_insert_send(max_send_data))
```
djc, line 219: Was actually hoping for something that can also abstract over the `self.state.send.get_mut()` of it all.

jeromegn:
Shall we merge or are we waiting for a commit history cleanup from me? :)

djc:
I have a cleaned up branch but was failing to push it up to your remote... let me just push it here.

djc:
See #1685. Might do more golfing later. @jeromegn would you mind forward porting?

jeromegn:
Made a PR for the forward port: https://github.com/quinn-rs/quinn/pull/1686

PR #1685, opened 2023-10-14
#1682 squashed + split

djc:
cc @jeromegn


quinn-proto/src/connection/streams/state.rs
```diff
@@ -865,6 +858,17 @@ impl StreamsState {
             self.send_streams -= 1;
         }
     }
+
+    pub(super) fn max_send_data(&self, id: StreamId) -> VarInt {
```
Ralith, line 895: We should call this `initial_max_...` to avoid confusion with live flow control data.

Ralith (approved):

PR #1691, opened 2023-10-18
Make `quinn-proto::{Connection, Endpoint}` deterministic

### LINKED ISSUE #1684: quinn-proto: Non-determinism in endpoint/connection/congestion
While debugging an issue I have, I found that `quinn-proto` protocol isn't fully deterministic. The source of randomness seems to be the call to `StdRng::from_entropy` in three places:

https://github.com/quinn-rs/quinn/blob/e1e1e6e392a382fbded42ca010505fecb8fe3655/quinn-proto/src/endpoint.rs#L62

https://github.com/quinn-rs/quinn/blob/e1e1e6e392a382fbded42ca010505fecb8fe3655/quinn-proto/src/connection/mod.rs#L270

https://github.com/quinn-rs/quinn/blob/e1e1e6e392a382fbded42ca010505fecb8fe3655/quinn-proto/src/congestion/bbr/mod.rs#L99

Are these intended? 

Thank you for the help!

michael-yxchen:
`quinn-proto::{Connection, Endpoint}` structs call `StdRng::from_entropy()` in their implementation. It makes it hard to reproduce errors in an otherwise deterministic test case.

This PR addresses the issue by adding a `StdRng` argument to the respective `new` functions. For other functions that may create an endpoint/connection, an argument of type `impl FnOnce() -> StdRng` is added. This avoids eager calls to generate fresh entropy. e.g. the `Endpoint::handle` function is frequently called but we'd only need the rng when handling a new incoming connection.

Fixes an unrelated lint error.

Closes #1684

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -252,7 +252,7 @@ impl Connection {
         now: Instant,
         version: u32,
         allow_mtud: bool,
-        rng_seed: <StdRng as SeedableRng>::Seed,
+        rng_seed: [u8; 32],
     ) -> Self {
         let side = if server_config.is_some() {
             Side::Server
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -10,7 +10,7 @@ use std::{
 
 use bytes::{Bytes, BytesMut};
 use frame::StreamMetaVec;
-use rand::{rngs::StdRng, Rng};
+use rand::{rngs::StdRng, Rng, SeedableRng};
 use thiserror::Error;
 use tracing::{debug, error, trace, trace_span, warn};
 
@@ -252,7 +252,7 @@ impl Connection {
         now: Instant,
         version: u32,
         allow_mtud: bool,
-        mut rng: StdRng,
+        rng_seed: <StdRng as SeedableRng>::Seed,
     ) -> Self {
         let side = if server_config.is_some() {
             Side::Server
@@ -268,6 +268,7 @@ impl Connection {
             expected_token: Bytes::new(),
             client_hello: None,
         });
+        let mut rng = StdRng::from_seed(rng_seed);
         let path_validated = server_config.as_ref().map_or(true, |c| c.use_retry);
         let mut this = Self {
             endpoint_config,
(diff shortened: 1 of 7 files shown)


quinn-proto/src/endpoint.rs
```diff
@@ -126,6 +127,7 @@ impl Endpoint {
         local_ip: Option<IpAddr>,
         ecn: Option<EcnCodepoint>,
         data: BytesMut,
+        rng: impl FnOnce() -> StdRng,
```
Ralith, line 130: Why doesn't this and below use `self.rng`?

Ralith (changes requested):
I'd prefer to avoid making `rand` a public dependency if possible. Could we accept a seed instead of an RNG?

quinn-proto/src/endpoint.rs
michael-yxchen: I got the impression that the endpoint and connection it creates are using freshly seeded rngs, as they were calling `StdRng::from_entropy` separately. Should I use one to seed the other in the upcoming fix?

djc: Yes, I think that would make sense.

michael-yxchen:
Put up another version that always use `Endpoint.rng` to seed `Connection.rng`. Let me know if you like it this way, or prefer to obtain fresh randomness from the entropy pool if `Endpoint.rng` is not seeded

quinn-proto/src/connection/mod.rs
```diff
@@ -252,6 +252,7 @@ impl Connection {
         now: Instant,
         version: u32,
         allow_mtud: bool,
+        rng_seed: <StdRng as SeedableRng>::Seed,
```
Ralith, line 255: Can we just name the concrete type here for clarity?

quinn-proto/src/endpoint.rs
```diff
@@ -57,9 +57,10 @@ impl Endpoint {
         config: Arc<EndpointConfig>,
         server_config: Option<Arc<ServerConfig>>,
         allow_mtud: bool,
+        rng_seed: Option<<StdRng as SeedableRng>::Seed>,
```
Ralith, line 60: As above

Ralith (changes requested):
Thanks, that looks cleaner!

quinn-proto/src/endpoint.rs
Ralith: Yeah, using fresh entropy in each was a matter of convenience, not a principled requirement. Using well-conditioned cryptographic RNGs to seed eachother is robust.

Ralith (approved):
Thanks!

djc (approved):

PR #1692, opened 2023-10-19
feat: use mutable transmits to AsyncUdpSocket::poll_send

dignifiedquire:
In our implementation (n0-computer/iroh) we currently have to maniuplate the destination addresses based on internal mappings. Currently we need to copy the full `Transmit` slice in order to update these addresses. To avoid this copy, this changes it such that a mutable reference is passed in.

### DIFF
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -490,7 +490,7 @@ impl State {
                 break Ok(true);
             }
 
-            match self.socket.poll_send(cx, self.outgoing.as_slices().0) {
+            match self.socket.poll_send(cx, self.outgoing.as_mut_slices().0) {
                 Poll::Ready(Ok(n)) => {
                     let contents_len: usize =
                         self.outgoing.drain(..n).map(|t| t.contents.len()).sum();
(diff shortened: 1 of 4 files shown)


Ralith:
What happens when a `poll_send` call isn't able to transmit all packets immediately, i.e. returns less than `transmits.len()`? It seems difficult to avoid unpredictably double-translating addresses with this change, which could be harmful if your translation isn't idempotent.

dignifiedquire:
That should be easy enough to handle on our side, we can either reset to the original addrs for those that weren't sent, or detect based on the addr if we have mapped it yet.

dignifiedquire:
The CI failures look unrelated to me: BSD seems to hang and the lint is complaining about unrelated code.

chayleaf:
this would be useful for us too (for rewriting the destination address as well)

Ralith:
I'd like to at least see a clearly documented contract for whether or not unsent entries may be changed after the call returns. I think it should specifically guarantee that they won't be to avoid unpredictable behavior. It's not clear to me if that guarantee can be respected without defeating the point by requiring storage on the side, though.

Is this copy showing up in profiling at all? It should be straightforward to reuse an allocation for it.

Ralith:
https://github.com/quinn-rs/quinn/pull/1729 will likely render this unnecessary by only accepting a single, trivially rewritten `Transmit` at a time.

dignifiedquire:
Closing in favor of #1729

PR #1693, opened 2023-10-20
Yield protocol violation for packets without frames

djc:
Received a private bug report from @QUICTester. Thanks!

This should probably have a test, which I don't have energy for right now. @Ralith any suggestion for where such a test should live?

I don't love that this equates "no frames" to "empty payload" which feels like a potential conceptual jump, but the for-loop formulation makes the alternative a decent bit more verbose...

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -2192,7 +2192,7 @@ impl Connection {
                 return Ok(());
             }
             State::Closed(_) => {
-                for result in frame::Iter::new(packet.payload.freeze()) {
+                for result in frame::Iter::new(packet.payload.freeze())? {
                     let frame = match result {
                         Ok(frame) => frame,
                         Err(err) => {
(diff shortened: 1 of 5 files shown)


Ralith:
> any suggestion for where such a test should live?

If we move this check into `frame::Iter::new` it becomes easy to test in isolation, but also so trivial that we might not bother.

> I don't love that this equates "no frames" to "empty payload" which feels like a potential conceptual jump

Aren't these equivalent? If payload data isn't a well-formed frame, that's an error too.

djc:
> If we move this check into `frame::Iter::new` it becomes easy to test in isolation, but also so trivial that we might not bother.

Yup, that's better -- done.

> Aren't these equivalent? If payload data isn't a well-formed frame, that's an error too.

Yup, I do think they're equivalent.

quinn-proto/src/connection/mod.rs
```diff
@@ -2494,12 +2494,21 @@ impl Connection {
...
     ) -> Result<(), TransportError> {
+        if payload.is_empty() {
+            // "An endpoint MUST treat receipt of a packet containing no frames as a
+            // connection error of type PROTOCOL_VIOLATION."
+            // https://www.rfc-editor.org/rfc/rfc9000.html#name-frames-and-frame-types
+            return Err(TransportError::PROTOCOL_VIOLATION(
+                "packet payload is empty",
+            ));
+        }
+
```
Ralith, line 2505: Leftover code:
```suggestion
```

Ralith (approved):

PR #1697, opened 2023-10-25
#729: proto: write outgoing packets to caller-supplied memory

lmpn:
I've used `BytesMut` for convenience but I can change to `&mut [u8]`.


quinn-proto/src/tests/util.rs
```diff
@@ -144,21 +144,19 @@ impl Pair {
                 );
                 continue;
             }
-            if x.contents[0] & packet::LONG_HEADER_FORM == 0 {
-                let spin = x.contents[0] & packet::SPIN_BIT != 0;
+            if x.1[0] & packet::LONG_HEADER_FORM == 0 {
```
djc, line 147: IMO we should unpack the tuple in the loop header here.

quinn/src/endpoint.rs
```diff
@@ -408,6 +408,7 @@ impl State {
                     .write(IoSliceMut::<'a>::new(buf));
             });
         let mut iovs = unsafe { iovs.assume_init() };
+        let mut buffer = BytesMut::with_capacity(1500);
```
djc, line 411: I think this (and other instances of this) should presumably use some call into quinn-proto to get the applicable MTU instead of hard-coding this.

djc (commented):
This seems pretty nice! What's your motivation? Have you measured the performance impact?

quinn-proto/src/connection/mod.rs
```diff
@@ -2,7 +2,8 @@ use std::{
     cmp,
     collections::VecDeque,
     convert::TryFrom,
-    fmt, io, mem,
+    fmt, io,
+    mem::{self},
```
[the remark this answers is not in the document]
lmpn: it isn't, fixed

quinn-proto/src/tests/util.rs
lmpn: done

quinn-proto/src/tests/util.rs
```diff
@@ -176,15 +174,13 @@ impl Pair {
                 continue;
             }
             if let Some(ref socket) = self.server.socket {
-                socket.send_to(&x.contents, x.destination).unwrap();
+                socket.send_to(&x.1, x.0.destination).unwrap();
```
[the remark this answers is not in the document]
lmpn: done

quinn/src/endpoint.rs
```diff
@@ -562,11 +565,11 @@ impl State {
 }
 
 #[inline]
-fn udp_transmit(t: proto::Transmit) -> udp::Transmit {
+fn udp_transmit(t: proto::Transmit, buffer: Bytes) -> udp::Transmit {
     udp::Transmit {
         destination: t.destination,
+        contents: buffer,
```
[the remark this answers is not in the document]
lmpn: Most likely is because initially i tried to do this with lifetime and this is leftover from it.
Fixed.

quinn/src/endpoint.rs
lmpn: Please check out my latest commit

quinn-proto/src/connection/mod.rs
```diff
@@ -477,13 +482,17 @@ impl Connection {
                     SpaceId::Data,
                     "PATH_CHALLENGE queued without 1-RTT keys"
                 );
-                let mut buf = BytesMut::with_capacity(self.path.current_mtu() as usize);
-                let buf_capacity = self.path.current_mtu() as usize;
+
+                if buf.capacity() < self.path.current_mtu() as usize {
```
djc, line 486: I don't think this check is right? Previously we were checking the capacity of a newly created `BytesMut`, but now we're getting a pre-existing `BytesMut` from another callers -- instead of the total capacity we should be making sure that enough capacity is remaining beyond the current contents.

quinn-proto/src/connection/mod.rs
```diff
@@ -901,14 +909,14 @@ impl Connection {
...
-            contents: buf.freeze(),
+            size: buf.len(),
             ecn: if self.path.sending_ecn {
                 Some(EcnCodepoint::Ect0)
             } else {
                 None
             },
             segment_size: match num_datagrams {
-                1 => None,
+                1 | 0 => None,
```
djc, line 919: Why is this necessary? It doesn't obviously make sense to me.

quinn/src/connection.rs
```diff
@@ -857,16 +857,17 @@ impl State {
         let mut transmits = 0;
 
         let max_datagrams = self.socket.max_transmit_segments();
-
-        while let Some(t) = self.inner.poll_transmit(now, max_datagrams) {
+        let mut buffer = BytesMut::with_capacity(1500);
```
djc, line 860: Use MTU.

quinn-proto/src/connection/mod.rs
lmpn: No it is not.
I had an issue running the tests while doing the first iteration.
Thanks for the catch 🙏

quinn/src/connection.rs
lmpn: I need a pointer regarding this... I've investigated but I'm could not understand from where should I get the MTU value in this case.

lmpn:
I've not forgotten about the perf. I will share the comparison later 👍

Ralith:
I'd be surprised if there was a significant performance improvement from this alone, but it brings us much closer to being able to reuse transmit buffers (perhaps by doing UDP transmits directly from connection tasks), which is more likely to be directly beneficial.

lmpn:
I've ran `cargo bench` for main and this branch. Results below:

poll_transmit_pre_alloc
test large_data_10_streams  ... bench:  30,968,912 ns/iter (+/- 11,592,202) = 338 MB/s
test large_data_1_stream    ... bench:  26,580,878 ns/iter (+/- 20,358,295) = 39 MB/s
test small_data_100_streams ... bench:  14,363,537 ns/iter (+/- 17,569,828)
test small_data_1_stream    ... bench:  19,154,729 ns/iter (+/- 5,427,772)

latest main
test large_data_10_streams  ... bench:  30,603,466 ns/iter (+/- 11,396,640) = 342 MB/s
test large_data_1_stream    ... bench:  21,525,054 ns/iter (+/- 19,954,110) = 48 MB/s
test small_data_100_streams ... bench:  17,087,725 ns/iter (+/- 22,559,371)
test small_data_1_stream    ... bench:  19,227,774 ns/iter (+/- 5,343,771)

If this is not enough please give me some pointers and I will try to do an analysis that fits the project.

lmpn:
@djc I've decided to expose the mtu from the connection.
If there is already an approach to retrieve or calculate such value tell me 🙏

djc:
So this shows as a ~18% regression for `small_data_100_streams`? That seems bad -- and surprising...

lmpn:
So I've executed `cargo bench` 10 times for main and this branch and compared the times(image below)
Screenshot 2023-10-27 at 15 16 06

There is indeed a regression (on average) for `large_data_1_stream` and `small_data_100_streams`, 3% and 6% respectively.

quinn-proto/src/connection/mod.rs
```diff
@@ -3423,6 +3423,11 @@ impl Connection {
         self.state = State::Drained;
         self.endpoint_events.push_back(EndpointEventInner::Drained);
     }
+
+    /// Returns the connection's current MTU
+    pub fn get_current_mtu(&self) -> u16 {
```
Ralith, line 3428: nit: Rust style is to omit `get_` prefixes.

quinn-proto/src/connection/mod.rs
```diff
@@ -3423,6 +3423,11 @@ impl Connection {
         self.state = State::Drained;
         self.endpoint_events.push_back(EndpointEventInner::Drained);
     }
+
+    /// Returns the connection's current MTU
```
Ralith, line 3427: This just repeats the function's name. Let's add information that helps a reader who doesn't understand the name, and which gives context for how it might be used. Maybe something like:

```
/// Storage required for the largest packet known to be supported by the current path
///
/// Buffers passed to [`Connection::poll_transmit`] should be at least this large.
```

Ralith (commented):
Exposing the MTU makes sense to me; other quinn-proto users will want it for the same reason.

Ralith:
@lmpn I'm having some difficulty reading your chart screenshot. Are the 4 aggregated figures for each separate test, in order? So there's a 20% speedup in `small_data_1_stream`? That's confusing since the original single run you posted indicated no change there, and a similar speedup in `small_data_100_streams`. That original run also indicates a *very* high variance, so maybe this data is too noisy to be meaningful anyway. I'll try to get some of my own numbers to compare.

We *should* be able to get identical performance to `main` by having the `quinn` layer create a fresh `BytesMut` for each `poll_transmit` call, right?

lmpn:
Sorry for the bad image. Check the one below and tell me if it is clearer.
The speedups are calculated as (1-main_time/poll_transmit_pre_alloc_time)*100

Screenshot 2023-10-28 at 14 28 24

The single run showed 18% regression for `small_data_100_streams`. I assumed this was because there is too much noise/bad run and because I'm running in my own computer, thus, I took 10 runs of each branch.

As you can see above I calculated the speedup for the:
- average of all times
- max of all times
- min of all times
- top 3 minimum times
- top 5 minimum times

Even with noise this values show:
- `small_data_1_stream` has improved in every scenario
- there is a regression for `large_data_1_stream` / `small_data_100_streams`
- (I think) `large_data_10_streams` the difference *seems* negligible.


> We _should_ be able to get identical performance to `main` by having the `quinn` layer create a fresh `BytesMut` for each `poll_transmit` call, right?

Yes I would assume so.

Ralith:
Thanks for the detailed analysis! To summarize: apparent massive speedup for `small_data_1_stream`, maybe a slight slowdown (or just noise) elsewhere. Perhaps reusing a single `BytesMut` is letting a single allocation serve multiple small packets.

I think this nets out as a win as written, personally. I've experimented on a couple of my machines (both Windows and Linux) and wasn't able to show a significant difference either way.

djc (approved):

quinn-proto/src/connection/mod.rs
```diff
@@ -477,13 +482,17 @@ impl Connection {
                     SpaceId::Data,
                     "PATH_CHALLENGE queued without 1-RTT keys"
                 );
-                let mut buf = BytesMut::with_capacity(self.path.current_mtu() as usize);
-                let buf_capacity = self.path.current_mtu() as usize;
+                let remaining_capacity = buf.capacity() - buf.len();
+                if remaining_capacity < self.path.current_mtu() as usize {
+                    buf.reserve(self.path.current_mtu() as usize - remaining_capacity);
+                }
```
Ralith, line 488: `BytesMut::reserve` is relative to the current length, not the current capacity (see [source](https://github.com/tokio-rs/bytes/blob/bde8c50703869f54b905560eb62c2478a0111885/src/bytes_mut.rs#L566-L574)), so this can under-reserve. Further, we're actually duplicating the short-circuit logic inside `reserve` here; we might as well rely on that directly for simplicity.

```suggestion
                buf.reserve(self.path.current_mtu() as usize);
```

quinn-proto/src/lib.rs
```diff
@@ -282,7 +281,7 @@ pub struct Transmit {
     /// Explicit congestion notification bits to set on the packet
     pub ecn: Option<EcnCodepoint>,
     /// Contents of the datagram
-    pub contents: Bytes,
+    pub size: usize,
```
Ralith, line 284: Outdated comment. Should now say something like "Amount of data written to the caller-supplied buffer".

quinn/src/connection.rs
```diff
@@ -857,16 +857,19 @@ impl State {
...
-                Some(s) => (t.contents.len() + s - 1) / s, // round up
+                Some(s) => (t.size + s - 1) / s, // round up
             };
             // If the endpoint driver is gone, noop.
-            let _ = self
-                .endpoint_events
-                .send((self.handle, EndpointEvent::Transmit(t)));
+            let _ = self.endpoint_events.send((
+                self.handle,
+                EndpointEvent::Transmit(t, buffer.clone().freeze()),
```
Ralith, line 871: `BytesMut::clone` always performs a deep copy. We should use [`split_to`](https://docs.rs/bytes/latest/bytes/struct.BytesMut.html#method.split_to) instead to have any possibility of reduced allocs/copies. That will also make `clear` below unnecessary.

Ralith (changes requested):

djc:
@lmpn was this motivated by Hacktoberfest, and if so, is there a merge deadline to make this count for your stats?

lmpn:
> @lmpn was this motivated by Hacktoberfest, and if so, is there a merge deadline to make this count for your stats?

No...
I just have interest in the networking and performance programming.

quinn/src/endpoint.rs
```diff
@@ -441,19 +444,23 @@ impl State {
...
-                                        self.outgoing.push_back(udp_transmit(t));
+                                        let contents_len = transmit.size;
+                                        self.outgoing.push_back(udp_transmit(
+                                            transmit,
+                                            buffer.split_to(contents_len).freeze(),
+                                        ));
                                         self.transmit_queue_contents_len = self
                                             .transmit_queue_contents_len
                                             .saturating_add(contents_len);
+                                        buffer.clear();
```
Ralith, line 463: Because we're using `split_to` above, this `clear` is now a no-op; let's drop it.

Ralith (approved):
Thanks!

lmpn:
Screenshot 2023-11-01 at 12 00 11

Above the results of the benchmark with the latest commit on this branch.
If someone can double check on their side it would be great.

Ralith:
My environment still registers too much noise to get meaningful data, but it's exciting that it had such an impact in yours! Maybe the impact is larger on Windows, where GSO isn't doing as much heavy lifting.

PR #1698, opened 2023-10-28
ensure LICENSE files are present in all published crates

decathorpe:
This PR adds symbolic links for the `LICENSE-APACHE` and `LICENSE-MIT` files into the `quinn`, `quinn-proto`, and `quinn-udp` workspace members so that `cargo package` / `cargo publish` picks them up and includes them when publishing crates.

Both the Apache-2.0 and the MIT license requires that (re)distributed sources contain a copy of the license text. Not including the license text is a problem, especially for Linux distributions that package this crate as a dependency, for example, for the HTTP/3 support in the `reqwest` crate.

I have not added a symlink into workspace members that have `publish = false` set in their cargo metadata.

### DIFF
--- a/quinn-proto/LICENSE-APACHE
+++ b/quinn-proto/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../LICENSE-APACHE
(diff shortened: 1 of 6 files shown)


djc (approved):
Thanks! Would you like to have this backported into a 0.10.x release?

decathorpe:
Thanks for merging! Backporting to 0.10.x and a new release are not necessary.

Issue #765, opened 2020-05-11
ECN on Windows


Demi-Marie:
This might be possible, although the APIs are undocumented.

Ralith:
It would be useful to cite evidence.

DemiMarie-parity:
Receiving ECN is possible via the [`IP_RECVTOS` option at `IPPROTO_IP`](https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options) and the [`IPV6_RECVTCLASS` option at `IPPROTO_IPV6`](https://docs.microsoft.com/en-us/windows/win32/winsock/ipproto-ipv6-socket-options).  Setting is definitely possible if one has administrative privileges.  I am not sure about otherwise.

DemiMarie-parity:
Setting it might be possible via the [undocumented `IPV6_ECN` and `IP_ECN` socket options](/aL3891/RioSharp/blob/4fdb3b5426ecab3ce60cccda2289bb8a1bc2cd14/RioSharp/WinSock.cs).

DemiMarie-parity:
As per https://github.com/microsoft/msquic/issues/559#issuecomment-650610256 I don’t believe this is supported on current versions of Windows.  Closing.

Demi-Marie:
Reopening in light of https://github.com/microsoft/msquic/issues/559#issuecomment-677783801 which actually implements ECN in MsQuic.  Since `msquic.dll` is not an OS-provided DLL, I am confident that this will continue to work in future releases of Windows.

panxy2013:
in microsoft document, we can write UDP client / server, and use wirshark can find, IP ECN 2bits can be modify by windows socket API WSASendMsg

but how to use in Rust, does quinn-udp can support such action?
1. send UDP with ECN bit = 0x01 or 0x02
2. Recieve UDP and can get ECN bit from receive UDP packet

https://learn.microsoft.com/en-us/windows/win32/winsock/winsock-ecn

Thanks

djc:
We don't support this yet in quinn-udp, would you be able to send a PR to use windows-sys APIs to implement this?

panxy2013:
Well, since I'm newer to Rust, I don't have the ability to write good Rust code at the moment
i am working with ALVR project, which use C++/Rust language.

i will use workaround to solve the current problem(Rust call C API to send ECN UDP packet)
Maybe i can try it, Thank you for your explanation

stormshield-damiend:
[...]

PR #1701, opened 2023-11-03
udp: add support for ECN, local addrs, GSO, and GRO on Windows

stormshield-damiend:
This adds support for ECN on Windows using `WSARecvMsg()` and `WSASendMsg()` from WinSock API.

This solves #765.

Some points need to be discussed or done :

- [X] `WSARecvMSG()` cannot be accessed directly, we need to retrieve its function pointer using a dedicated call to `WSAIoctl()`. This detection may fail on old versions of windows (it seems to have been added as an extension in Windows XP). Do we keep the old API as a fallback (as currently done in the draft PR) ?
-  [X] If we keep the fallback API, do we still send ECN if we cannot receive them ?
- [X] The previous detection is also done at every creation of a socket. Should this be done once on a dummy UDP socket in a lazy way (using OnceCell or OnceLock for example) ?
- [x] All unsafe code block must be checked and have a # Safety : documentation
- [x] wsa_cmsg.rs must be reviewed carefully
- [x] see if we could merge cmsg.rs and wsa_cmsg.rs
- [x] dual stack test must be done

Some possible long term tasks or benefits :
- We could try to merge cmsg and wsa_cmsg into socket2, as at the moment, there is no API to read or write socket control messages
- Using the new API gives us access to some other control messages / features, for example we could use [URO](https://github.com/microsoft/net-offloads/blob/main/udp-rsc-offload.md)

Feel free to comment on this draft.


quinn-udp/src/windows.rs
```diff
@@ -53,6 +90,34 @@ impl UdpSocketState {
             if rc == -1 {
                 return Err(io::Error::last_os_error());
             }
+
+            if wsa_recvmsg_ptr.is_some() {
+                let rc = unsafe {
+                    WinSock::setsockopt(
```
stormshield-damiend, line 96: Idea: add a helper function as done in unix.rs

quinn-udp/src/windows.rs
```diff
@@ -40,6 +47,36 @@ impl UdpSocketState {
 
         let sock_true: u32 = 1;
 
+        // Detect if OS expose WSARecvMsg API based on
```
stormshield-damiend, line 50: Idea: add some assert related to cmsg as done in unix.rs

quinn-udp/src/windows.rs
Ralith: That would be welcome for readability.

quinn-udp/src/windows.rs
```diff
@@ -40,6 +47,36 @@ impl UdpSocketState {
...
+                None,
+            )
+        };
+        if rc == -1 {
+            return Err(io::Error::last_os_error());
+        }
+
+        if len as usize != mem::size_of::<WinSock::LPFN_WSARECVMSG>() {
+            // consider that the function WSARecvMsg is not there
+            // FIXME log this ?
```
Ralith, line 76: A debug-level log seems reasonable.

quinn-udp/src/windows.rs
```diff
@@ -68,38 +133,162 @@ impl UdpSocketState {
             if rc == -1 {
                 return Err(io::Error::last_os_error());
             }
+
+            if wsa_recvmsg_ptr.is_some() {
```
Ralith, line 137: How do the below options interact with IPv4 packets received on dual-stack sockets? On Linux we have to set the IPv4 option `IP_RECVTOS` (probably analogous to Windows's `IP_ECN`) for dual-stack sockets to get ECN information for IPv4 packets.

quinn-udp/src/windows.rs
```diff
@@ -68,38 +133,162 @@ impl UdpSocketState {
...
     pub fn send(&self, socket: UdpSockRef<'_>, transmits: &[Transmit]) -> Result<usize, io::Error> {
         let mut sent = 0;
         for transmit in transmits {
-            match socket.0.send_to(
-                &transmit.contents,
-                &socket2::SockAddr::from(transmit.destination),
-            ) {
-                Ok(_) => {
-                    sent += 1;
+            // FIXME be symmetric ? do not send ECN if we cannot receive them ?
```
Ralith, line 176: Congestion is one-way, so there's no need to disable outgoing ECN just because we can't handle it incoming.

quinn-udp/src/windows.rs
```diff
@@ -40,6 +47,36 @@ impl UdpSocketState {
...
+        // Detect if OS expose WSARecvMsg API based on
+        // https://github.com/Azure/mio-uds-windows/blob/a3c97df82018086add96d8821edb4aa85ec1b42b/src/stdnet/ext.rs#L601
+        let guid = WinSock::WSAID_WSARECVMSG;
+        let mut wsa_recvmsg_ptr = None;
+        let mut len = 0;
+
+        // FIXME add Safety: Option handles the NULL pointer with a None value
+        let rc = unsafe {
+            WinSock::WSAIoctl(
+                socket.0.as_raw_socket() as _,
```
Ralith, line 59: I see we're passing in the socket here. Does that mean the function pointer might vary depending on the socket? If not, I wonder if we should use a static `OnceLock` or similar to avoid redundant work for every new endpoint.

quinn-udp/src/windows.rs
```diff
@@ -68,38 +133,162 @@ impl UdpSocketState {
...
-            match socket.0.send_to(
-                &transmit.contents,
-                &socket2::SockAddr::from(transmit.destination),
-            ) {
-                Ok(_) => {
-                    sent += 1;
+            // FIXME be symmetric ? do not send ECN if we cannot receive them ?
+            if self.wsa_recvmsg_ptr.is_some() {
+                // we cannot use [`socket2::sendmsg()`] and [`socket2::MsgHdr`] as we do not have access
+                // to the inner field which holds the WSAMSG
```
Ralith, line 179: Why do we need direct field access here, rather than using e.g. [`MsgHdr::with_control`](https://docs.rs/socket2/latest/socket2/struct.MsgHdr.html#method.with_control)?

quinn-udp/src/windows.rs
```diff
@@ -68,38 +133,162 @@ impl UdpSocketState {
...
                     // by higher level retransmits and timeouts.
-                    log_sendmsg_error(&self.last_send_error, e, transmit);
+                    log_sendmsg_error(&self.last_send_error, io::Error::last_os_error(), transmit);
                     sent += 1;
                 }
+            } else {
+                match socket.0.send_to(
+                    &transmit.contents,
+                    &socket2::SockAddr::from(transmit.destination),
+                ) {
```
Ralith, line 273: Let's not duplicate the body of this `match`. Maybe compute a `Result` from each possible send path, and then match on it once?

quinn-udp/src/windows.rs
```diff
@@ -68,38 +133,162 @@ impl UdpSocketState {
...
         let mut sent = 0;
         for transmit in transmits {
-            match socket.0.send_to(
-                &transmit.contents,
-                &socket2::SockAddr::from(transmit.destination),
-            ) {
-                Ok(_) => {
-                    sent += 1;
+            // FIXME be symmetric ? do not send ECN if we cannot receive them ?
+            if self.wsa_recvmsg_ptr.is_some() {
```
Ralith, line 177: Maybe move the body of this into a helper function for clarity and to limit indentation?

quinn-udp/src/windows.rs
```diff
@@ -111,22 +300,125 @@ impl UdpSocketState {
...
-        let (len, _flags, addr) = socket.0.recv_from_vectored(bufs)?;
-        meta[0] = RecvMeta {
-            len,
-            stride: len,
-            addr: addr.as_socket().unwrap(),
-            ecn: None,
-            dst_ip: None,
-        };
-        Ok(1)
+        if let Some(wsa_recvmsg_ptr) = self.wsa_recvmsg_ptr {
```
Ralith, line 303: As above, consider a helper function.

quinn-udp/src/windows.rs
```diff
@@ -111,22 +300,125 @@ impl UdpSocketState {
...
+                    None,
+                );
+                if rc == -1 {
+                    return Err(io::Error::last_os_error());
+                }
+            }
+
+            // FIXME add Safety: ?
+            let addr = unsafe {
+                let (_, addr) = socket2::SockAddr::try_init(|addr_storage, len| {
```
Ralith, line 343: Would `SockAddr::new` be a bit more concise here and below? I think `try_init` is intended for cases when you already have a (typically foreign) function that takes an output pointer.

quinn-udp/src/windows.rs
```diff
@@ -111,22 +300,125 @@ impl UdpSocketState {
...
+                            socket2::SockAddr::try_init(|addr_storage, len| {
+                                *len = mem::size_of_val(&destination) as _;
+                                std::ptr::copy_nonoverlapping(&destination, addr_storage as _, 1);
+                                Ok(())
+                            })?
+                        };
+                        dst_ip = address.as_socket().map(|s| s.ip());
+                    }
+                    (WinSock::IPPROTO_IP, WinSock::IP_ECN) => {
+                        ecn_bits = unsafe { cmsg::decode::<u8>(cmsg) };
```
Ralith, line 387: As above, can we justify this choice of payload type?

quinn-udp/src/windows.rs
```diff
@@ -68,38 +133,162 @@ impl UdpSocketState {
...
+                    lpBuffers: &mut data,
+                    Control: ctrl,
+                    dwBufferCount: 1,
+                    dwFlags: 0,
+                };
+
+                // Add control messages (ECN and PKTINFO)
+                let mut encoder = unsafe { cmsg::Encoder::new(&mut wsa_msg) };
+
+                let ecn = transmit.ecn.map_or(0, |x| x as u32);
```
Ralith, line 205: Can we cite docs attesting that `u32` is the proper cmsg payload type here? That's been a common source of bugs, across multiple QUIC impls.

quinn-udp/src/wsa_cmsg.rs
Ralith, line 1: This seems to have a lot in common with the existing cmsg.rs. Avoiding duplication is particularly valuable for tricky unsafe code. Can we factor out the common code somehow? Perhaps encapsulate the header types behind a platform-specific newtype?

Ralith (changes requested):
Thanks for digging into this!

quinn-udp/src/windows.rs
stormshield-damiend: It is the same and can be done on a [dummy socket](https://github.com/microsoft/Windows-classic-samples/blob/main/Samples/Win7Samples/netds/winsock/recvmsg/rmmc.cpp#L477).

I will change the code to use a `OnceLock`.

quinn-udp/src/windows.rs
stormshield-damiend: ok i will do it

quinn-udp/src/windows.rs
stormshield-damiend: ok i will do it

quinn-udp/src/windows.rs
stormshield-damiend: Nice catch, i will do some test to see how it works

quinn-udp/src/windows.rs
stormshield-damiend: ok i will update the code and thus drop the legacy part

quinn-udp/src/windows.rs
```diff
@@ -2,18 +2,25 @@ use std::{
...
 };
 
 use windows_sys::Win32::Networking::WinSock;
 
-use super::{log_sendmsg_error, RecvMeta, Transmit, UdpSockRef, IO_ERROR_LOG_INTERVAL};
+use crate::EcnCodepoint;
+
+use super::{cmsg, log_sendmsg_error, RecvMeta, Transmit, UdpSockRef, IO_ERROR_LOG_INTERVAL};
+
+const CMSG_LEN: usize = 128; // FIXME arbitrary value, same as unix ?, is it too small/much ?
```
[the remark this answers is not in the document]
stormshield-damiend: Ok i will keep the 128 which should be enough

quinn-udp/src/windows.rs
stormshield-damiend: ok i will do it

quinn-udp/src/windows.rs
stormshield-damiend: cause for now Encoder / Decoder / wsa_cmsg_xxx helpers are using the WSABuf, i will see if something can be done

quinn-udp/src/windows.rs
stormshield-damiend: Nice catch, this is a bug and should be a u32, i will fix the code and also link to a documentation if i found one. Otherwise i can also link to msquic userland implementation.

quinn-udp/src/windows.rs
stormshield-damiend: ok i will do it

quinn-udp/src/windows.rs
stormshield-damiend: ok i will do it

quinn-udp/src/windows.rs
stormshield-damiend: i will look to find a documentation to link, if i don't find one i can link to msquic userland code.

quinn-udp/src/wsa_cmsg.rs
stormshield-damiend: Yes i agree, i will first do the other proposed changes then take a look at this

quinn-udp/src/windows.rs
stormshield-damiend: This imply that we drop the legacy code (so Windows XP), is it ok with you ? if so i will also do it for the recv part.

quinn-udp/src/windows.rs
stormshield-damiend: i will take a look and see if it works

quinn-udp/src/windows.rs
Ralith: Windows XP is not supported by Microsoft, so I'm happy to not support it ourselves.

quinn-udp/src/windows.rs
stormshield-damiend: new code uses `OnceCell` from the crate due to rust MSRV, could you take a look and see if it is fine with you ? code will change a bit due to removal of legacy part but principle will stay the same.

quinn-udp/src/windows.rs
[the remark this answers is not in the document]
stormshield-damiend: I measured at max size of 64 for IPv6 case, so i kept 128.

quinn-udp/src/wsa_cmsg.rs
stormshield-damiend: i will start to work on this, after a first look at it, i do not think this will be possible without generic types.

I plan to add :
- and helper trait for cmsg operations (align, len, space, firsthdr, nexthdr, data, control len getter and setter). This helper use both Hdr and Cmsg native types, thus require two generic types.
- two implementation of this trait, one for windows with WSAMSG and WSACMG, one for Linux with the libc types
- a generic Encoder<Hdr, Cmsg, CmsgHelper>
- a generic decode<T, Cmsg> function
- a generic Iterator like struct
- i may also need to abstract some type for level and type in Cmsg

Are you fine with this approach or do you think it differently ?

stormshield-damiend: Or maybe add some platform wrappers for Hdr and Cmsg + a bunch of helper functions and rework Encoder / Decoder and Iterator ?

Ralith: Thanks! Could we arrange the platform-specific cmsg code into `cmsg/{unix, windows}.rs`? I'm thinking these might reexport/define the cmsg traversal functions (`NXTHDR`, etc) and header type aliases. If the Windows behavior is really as close a match as it seems, this should allow the existing `cmsg.rs` to be used almost as-is, just replacing `libc::` with `platform::` or w/e. I think that should work for both encoding and decoding.

stormshield-damiend: > Thanks! Could we arrange the platform-specific cmsg code into `cmsg/{unix, windows}.rs`? I'm thinking these might reexport/define the cmsg traversal functions (`NXTHDR`, etc) and header type aliases. If the Windows behavior is really as close a match as it seems, this should allow the existing `cmsg.rs` to be used almost as-is, just replacing `libc::` with `platform::` or w/e. I think that should work for both encoding and decoding.

Ok i will rework the code and see how it looks. I am busy on other work stuff until the end of December so it may take some time before i can rework the code.

> 
> I also wonder if we should ask Microsoft to expose the cmsg macros from windows-sys themselves.

Sure that is a good idea, do you want to do the PR or do you want me to do it ?

Ralith: > I am busy on other work stuff until the end of December so it may take some time before i can rework the code.

No worries! Will be a nice boost whenever we get it.

> Sure that is a good idea, do you want to do the PR or do you want me to do it ?

I think you've got the most context at the moment, so if you've got the time, please do!

stormshield-damiend: > > I am busy on other work stuff until the end of December so it may take some time before i can rework the code.
> 
> No worries! Will be a nice boost whenever we get it.
> 
> > Sure that is a good idea, do you want to do the PR or do you want me to do it ?
> 
> I think you've got the most context at the moment, so if you've got the time, please do!

They already had this asked and cannot do it due to the fact their rust wrappers function are generated from SDK  metadata which do not expose macros.

See https://github.com/microsoft/windows-rs/issues/2701 and also [Here](https://kennykerr.ca/rust-getting-started/where-are-the-macros.html)

They say that the C macros code must be replicated in Rust.

Maybe we could add those macros to socket2 ? If so i can propose a PR there.

stormshield-damiend: @Ralith i factorized the code for unix and tested it on linux. I am still working on windows one (my build VM is broken i will finish next Monday).

Could you take a look at `quinn-udp/src/cmsg/mod.rs` and `quinn-udp/src/cmsg/unix.rs` and their usage in `quinn-udp/src/unix.rs` and tell me if the approach is ok with you ?

Thanks

Ralith: This looks generally great, thanks! I have mixed feelings about using a trait when we only ever have exactly one implementation at compiletime, but I do like how it makes the common API explicit, allowing code to be validated without compiling on every supported platform. In any case, callsites are clear and concise.

quinn-udp/src/cmsg/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+
+#[cfg(windows)]
+#[path = "windows.rs"]
+mod imp;
+
+pub(crate) use imp::Aligned;
+
+// Helper traits for native types for control messages
+pub(crate) trait MsgHdrHelper {
+    type Output;
```
Ralith, line 18: Maybe name this `ControlHeader` for clarity? Also, could we add a `: CMsgHdrHelper` constraint to it? That should allow the rest of the code in this module to have only a single type variable rather than two.

quinn-udp/src/cmsg/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+mod imp;
+
+#[cfg(windows)]
+#[path = "windows.rs"]
+mod imp;
+
+pub(crate) use imp::Aligned;
+
+// Helper traits for native types for control messages
+pub(crate) trait MsgHdrHelper {
```
Ralith, line 17: style nit:
```suggestion
pub(crate) trait MsgHdr {
```

quinn-udp/src/cmsg/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn control_len(&self) -> usize;
+
+    fn set_control_len(&mut self, len: usize);
+
+    fn cmsg_firsthdr(&self) -> *mut Self::Output;
+
+    fn cmsg_nxthdr(&self, cmsg: &Self::Output) -> *mut Self::Output;
+}
+
+pub(crate) trait CMsgHdrHelper {
```
Ralith, line 29: style nit:
```suggestion
pub(crate) trait CMsgHdr {
```

Ralith (changes requested):

quinn-udp/src/cmsg/mod.rs
Ralith: An associated type is always uniquely identified by the type it's associated with:
```diff
diff --git a/quinn-udp/src/cmsg/mod.rs b/quinn-udp/src/cmsg/mod.rs
index 78d449f4..cc5ecdc7 100644
--- a/quinn-udp/src/cmsg/mod.rs
+++ b/quinn-udp/src/cmsg/mod.rs
@@ -15,7 +15,7 @@ pub(crate) use imp::Aligned;
 
 // Helper traits for native types for control messages
 pub(crate) trait MsgHdr {
-    type ControlMessage;
+    type ControlMessage: CMsgHdr;
 
     fn control_len(&self) -> usize;
 
@@ -43,13 +43,13 @@ pub(crate) trait CMsgHdr {
 ///
 /// The operation must be "finished" for the native msghdr to be usable, either by calling `finish`
 /// explicitly or by dropping the `Encoder`.
-pub(crate) struct Encoder<'a, M: MsgHdr, C: CMsgHdr> {
+pub(crate) struct Encoder<'a, M: MsgHdr> {
     hdr: &'a mut M,
-    cmsg: Option<&'a mut C>,
+    cmsg: Option<&'a mut M::ControlMessage>,
     len: usize,
 }
 
-impl<'a, M: MsgHdr<ControlMessage = C>, C: CMsgHdr> Encoder<'a, M, C> {
+impl<'a, M: MsgHdr> Encoder<'a, M> {
     /// # Safety
     /// - `hdr` must contains a suitably aligned pointer to a buffer big enought to hold the control messages bytes
     //    that can be safely written
@@ -68,8 +68,8 @@ impl<'a, M: MsgHdr<ControlMessage = C>, C: CMsgHdr> Encoder<'a, M, C> {
     /// - If insufficient buffer space remains.
     /// - If `T` has stricter alignment requirements than native `cmsghdr`
     pub(crate) fn push<T: Copy + ?Sized>(&mut self, level: c_int, ty: c_int, value: T) {
-        assert!(mem::align_of::<T>() <= mem::align_of::<C>());
-        let space = C::cmsg_space(mem::size_of_val(&value));
+        assert!(mem::align_of::<T>() <= mem::align_of::<M::ControlMessage>());
+        let space = M::ControlMessage::cmsg_space(mem::size_of_val(&value));
         assert!(
             self.hdr.control_len() >= self.len + space,
             "control message buffer too small. Required: {}, Available: {}",
@@ -77,7 +77,11 @@ impl<'a, M: MsgHdr<ControlMessage = C>, C: CMsgHdr> Encoder<'a, M, C> {
             self.hdr.control_len()
         );
         let cmsg = self.cmsg.take().expect("no control buffer space remaining");
-        cmsg.set(level, ty, C::cmsg_len(mem::size_of_val(&value)));
+        cmsg.set(
+            level,
+            ty,
+            M::ControlMessage::cmsg_len(mem::size_of_val(&value)),
+        );
         unsafe {
             ptr::write(cmsg.cmsg_data() as *const T as *mut T, value);
         }
@@ -93,7 +97,7 @@ impl<'a, M: MsgHdr<ControlMessage = C>, C: CMsgHdr> Encoder<'a, M, C> {
 
 // Statically guarantees that the encoding operation is "finished" before the control buffer is read
 // by `sendmsg` like API.
-impl<'a, M: MsgHdr, C: CMsgHdr> Drop for Encoder<'a, M, C> {
+impl<'a, M: MsgHdr> Drop for Encoder<'a, M> {
     fn drop(&mut self) {
         self.hdr.set_control_len(self.len as _);
     }
@@ -108,12 +112,12 @@ pub(crate) unsafe fn decode<T: Copy, C: CMsgHdr>(cmsg: &C) -> T {
     ptr::read(cmsg.cmsg_data() as *const T)
 }
 
-pub(crate) struct Iter<'a, M: MsgHdr, C: CMsgHdr> {
+pub(crate) struct Iter<'a, M: MsgHdr> {
     hdr: &'a M,
-    cmsg: Option<&'a C>,
+    cmsg: Option<&'a M::ControlMessage>,
 }
 
-impl<'a, M: MsgHdr<ControlMessage = C>, C: CMsgHdr> Iter<'a, M, C> {
+impl<'a, M: MsgHdr> Iter<'a, M> {
     /// # Safety
     ///
     /// `hdr` must hold a pointer to memory outliving `'a` which can be soundly read for the
@@ -127,8 +131,8 @@ impl<'a, M: MsgHdr<ControlMessage = C>, C: CMsgHdr> Iter<'a, M, C> {
     }
 }
 
-impl<'a, M: MsgHdr<ControlMessage = C>, C: CMsgHdr> Iterator for Iter<'a, M, C> {
-    type Item = &'a C;
+impl<'a, M: MsgHdr> Iterator for Iter<'a, M> {
+    type Item = &'a M::ControlMessage;
 
     fn next(&mut self) -> Option<Self::Item> {
         let current = self.cmsg.take()?;
diff --git a/quinn-udp/src/unix.rs b/quinn-udp/src/unix.rs
index 787679bb..d7d47814 100644
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -770,10 +770,7 @@ mod gso {
         }
     }
 
-    pub(crate) fn set_segment_size(
-        encoder: &mut cmsg::Encoder<libc::msghdr, libc::cmsghdr>,
-        segment_size: u16,
-    ) {
+    pub(crate) fn set_segment_size(encoder: &mut cmsg::Encoder<libc::msghdr>, segment_size: u16) {
         encoder.push(libc::SOL_UDP, libc::UDP_SEGMENT, segment_size);
     }
 }
```

djc:
I do think this should not be just one commit. I'd suggest at least some commits like this:

* Move `cmsg.rs` to `cmsg/mod.rs`
* Further splitting of `cmsg` into multiple modules without any other changes
* Change `cmsg` to new structure on the Unix side that can accomodate the changes for Windows
* Add changes for Windows

Instead of depending on once_cell, could we use [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) instead? We'd have to bump the MSRV a bunch...

stormshield-damiend:
> I do think this should not be just one commit. I'd suggest at least some commits like this:
> 
>     * Move `cmsg.rs` to `cmsg/mod.rs`
> 
Ok i can do that

>     * Further splitting of `cmsg` into multiple modules without any other changes

Something like `encoder.rs` / `decoder.rs` / `iterator.rs` ?

> 
>     * Change `cmsg` to new structure on the Unix side that can accomodate the changes for Windows

Sure i can do that

> 
>     * Add changes for Windows

In one commit ? 

> 
> 
> Instead of depending on once_cell, could we use [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) instead? We'd have to bump the MSRV a bunch...

Yes i can do that, but this is your and @Ralith calls.

djc:
> > I do think this should not be just one commit. I'd suggest at least some commits like this:
> > ```
> > * Move `cmsg.rs` to `cmsg/mod.rs`
> > ```
> 
> Ok i can do that
> 
> > ```
> > * Further splitting of `cmsg` into multiple modules without any other changes
> > ```
> 
> Something like `encoder.rs` / `decoder.rs` / `iterator.rs` ?

Sorry, I meant splitting along the lines of the split you have now (so in this commit, just splitting `unix.rs` out of `mod.rs`).

> 
> > ```
> > * Change `cmsg` to new structure on the Unix side that can accomodate the changes for Windows
> > ```
> 
> Sure i can do that
> 
> > ```
> > * Add changes for Windows
> > ```
> 
> In one commit?

If you think there are multiple stages that can each compile and pass tests, that might be nice?

> > Instead of depending on once_cell, could we use [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) instead? We'd have to bump the MSRV a bunch...
> 
> Yes i can do that, but this is your and @Ralith calls.

Let's see what @Ralith thinks.

stormshield-damiend:
I have taken into account @djc remarks. I still need to add Safety comments where needed.

quinn-udp/src/cmsg/windows.rs
```diff
@@ -0,0 +1,112 @@
...
+    fn cmsg_len(length: usize) -> usize {
+        self::wsa2::cmsg_len(length)
+    }
+
+    fn cmsg_data(&self) -> *mut c_uchar {
+        unsafe { self::wsa2::cmsg_data(self) }
+    }
+}
+
+mod wsa2 {
```
djc, line 61: What's the value of having this extra layer of abstractions? Why not inline all the functions defined in this module into the trait methods?

quinn-udp/src/windows.rs
```diff
@@ -1,14 +1,28 @@
...
+use super::{
+    cmsg::{self, CMsgHdr},
+    log_sendmsg_error, RecvMeta, Transmit, UdpSockRef, IO_ERROR_LOG_INTERVAL,
+};
+
+// Enough to store max(IP_PKTINFO + IP_ECN, IPV6_PKTINFO + IPV6_ECN) bytes (header + data) and some extra margin
+const CMSG_LEN: usize = 128;
+
+// FIXME this could use [`std::sync::OnceLock`] once the MSRV is bumped to 1.70 and upper
+static WSARECVMSG_PTR: OnceCell<WinSock::LPFN_WSARECVMSG> = OnceCell::new();
```
djc, line 25: Please move `const`/`static` declarations to the bottom of the module.

quinn-udp/src/windows.rs
```diff
@@ -1,14 +1,28 @@
...
     sync::Mutex,
     time::Instant,
 };
 
+use once_cell::sync::OnceCell;
 use windows_sys::Win32::Networking::WinSock;
 
-use super::{log_sendmsg_error, RecvMeta, Transmit, UdpSockRef, IO_ERROR_LOG_INTERVAL};
+use crate::EcnCodepoint;
+
```
djc, line 15: Nit: let's keep `use crate::` and `use super::` together in the same block.

quinn-udp/src/windows.rs
```diff
@@ -140,8 +155,240 @@ impl UdpSocketState {
     }
 }
 
+fn send(
```
djc, line 158: What's the value in this indirection?

quinn-udp/src/windows.rs
```diff
@@ -38,13 +62,30 @@ impl UdpSocketState {
         };
         let is_ipv4 = addr.as_socket_ipv4().is_some() || !v6only;
 
+        let wsa_recvmsg_ptr = WSARECVMSG_PTR.get_or_init(|| get_wsarecvmsg_fn(&*socket.0));
+
+        // We do not support anymore old version of windows that do not give access to WSARecvMsg() function
+        if wsa_recvmsg_ptr.is_none() {
+            tracing::error!("Network stack does not support WSARecvMsg function");
```
djc, line 69: Please lowercase all log messages.

quinn-udp/src/cmsg/mod.rs
```diff
@@ -1,113 +1,5 @@
-use std::{mem, ptr};
+#[cfg(unix)]
```
djc, line 6: Okay, let's squash this commit into the previous one.

quinn-udp/src/cmsg/mod.rs
```diff
@@ -1,5 +1,138 @@
...
+
 #[cfg(unix)]
 #[path = "unix.rs"]
 mod imp;
 
-pub(crate) use imp::{decode, Aligned, Encoder, Iter};
+pub(crate) use imp::Aligned;
+
+// Helper traits for native types for control messages
+pub(crate) trait MsgHdr {
```
djc, line 17: For this commit, I'd like to split it into two parts, one that extracts the traits from the surrounding code but leaves as much as possible in the same place in `cmsg/unix.rs`, minimizing the diff, and then another that moves the non-Unix specific stuff over to `mod.rs`. This should make it easier to see what is changed.

By prevailing style, the trait definitions should be near the bottom of the module, below the `Encoder` and other types that rely on them.

Within this trait, let's order the methods as `cmsg_firsthdr()`, then `cmsg_nxthdr()`, then `set_control_len()`, then `control_len()` (and use the same order in the impls).

I'm not a big fan of the mix of C naming conventions (`MsgHdr`, `cmsg_firsthdr()`) and Rust naming conventions (`set_control_len()`). Maybe go with Rust naming conventions for the traits? @Ralith thoughts?

djc (commented):
Looking pretty good!

Ralith:
> Instead of depending on once_cell, could we use [OnceLock](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) instead? We'd have to bump the MSRV a bunch...

Our rule of thumb is 6 months, right? 1.70 is just about there, so I'm fine with it. Tokio's still on 1.63, but probably they just haven't had any need to change, and there's no sense letting that hold us back arbitrarily. It's also a really trivial update, though, so I'm also happy to judge it non-blocking and handle it personally in a followup if @stormshield-damiend wants to stay focused.

quinn-udp/src/cmsg/mod.rs
Ralith: I think there's some value in clarity from a direct correspondence to C naming, so I prefer e.g. `MsgHdr` over `MessageHeader`. We're not trying to represent the abstract notion of a header here, but a specific POSIX-specified struct, its API, and their direct analogs in Win32.

That said, we should at least be consistent in either always preserving the literal C "words" (i.e. `controllen`), or always delimiting them (e.g. `cmsg_first_hdr`).

I don't feel too strongly either way, though, since suitable documentation should make this clear enough and it's not exactly a far-reaching abstraction.

quinn-udp/src/cmsg/windows.rs
stormshield-damiend: the goal was to isolate the porting of CMSG_XXX C macro in a private module in case we try to upstream it to socket2. I can remove it and flatten the code.

quinn-udp/src/windows.rs
stormshield-damiend: Sure, do you have a doc somewhere that describe the style you follow in quinn ?

quinn-udp/src/windows.rs
stormshield-damiend: it come from the time where we had both implementation (new one using `wsasendmsg` and old one using send_to), i can flatten the code.

stormshield-damiend:
> > Instead of depending on once_cell, could we use [OnceLock](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) instead? We'd have to bump the MSRV a bunch...
> 
> Our rule of thumb is 6 months, right? 1.70 is just about there, so I'm fine with it. Tokio's still on 1.63, but probably they just haven't had any need to change, and there's no sense letting that hold us back arbitrarily. It's also a really trivial update, though, so I'm also happy to judge it non-blocking and handle it personally in a followup if @stormshield-damiend wants to stay focused.

Fine with me, lets keep the `once_cell` dependency and remove it later.

quinn-udp/src/windows.rs
stormshield-damiend: done, can you tell me if it is fine ?

quinn-udp/src/cmsg/mod.rs
stormshield-damiend: i will work on this part tomorrow

djc: I would prefer delimiting the words everywhere.

quinn-udp/src/windows.rs
djc: The closest thing to a style guide we have so far is rustls: https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md.

quinn-udp/src/cmsg/windows.rs
djc: Yes, please flatten.

quinn-udp/src/windows.rs
stormshield-damiend: Maybe this could be link as a new section in main README.md ?

quinn-udp/src/cmsg/mod.rs
stormshield-damiend: I was also planning to choose this approach.

quinn-udp/src/windows.rs
```diff
@@ -62,30 +109,98 @@ impl UdpSocketState {
...
+
+            // Add control messages (ECN and PKTINFO)
+            let mut encoder = unsafe { cmsg::Encoder::new(&mut wsa_msg) };
+
+            if let Some(ip) = transmit.src_ip {
+                let ip = std::net::SocketAddr::new(ip, 0);
+                let ip = socket2::SockAddr::from(ip);
+                match ip.family() {
+                    WinSock::AF_INET => {
+                        let src_ip: WinSock::SOCKADDR_IN = unsafe { ptr::read(ip.as_ptr() as _) };
```
djc, line 147: Style nit: feels silly to have a type annotation combined with `as _`. For those cases, let's just write it as `as WinSock::SOCKADDR_IN`?

quinn-udp/src/cmsg/mod.rs
```diff
@@ -1,5 +1,138 @@
...
+    ffi::{c_int, c_uchar},
+    mem, ptr,
+};
+
 #[cfg(unix)]
 #[path = "unix.rs"]
 mod imp;
 
-pub(crate) use imp::{decode, Aligned, Encoder, Iter};
+pub(crate) use imp::Aligned;
```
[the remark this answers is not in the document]
stormshield-damiend: Do you have any suggestion ?

quinn-udp/src/windows.rs
```diff
@@ -18,6 +24,16 @@ pub struct UdpSocketState {
 
 impl UdpSocketState {
     pub fn new(socket: UdpSockRef<'_>) -> io::Result<Self> {
+        assert!(
```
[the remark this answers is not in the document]
stormshield-damiend: unix part has the same type of assert, i dunno if there are documented somewhere

quinn-udp/src/cmsg/mod.rs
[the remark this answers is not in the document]
djc: "udp: move newly generic code for reusability"?

quinn-udp/src/windows.rs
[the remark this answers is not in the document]
djc: I saw that at least some public methods that have a bunch of asserts have nice documentation of the possible panics (you adjust the docstrings). We should copy those for all public methods that have a bunch of assertions -- at least the new ones.

quinn-udp/src/cmsg/mod.rs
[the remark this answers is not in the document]
stormshield-damiend: i'll go for that, thanks

quinn-udp/src/windows.rs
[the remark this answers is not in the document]
stormshield-damiend: i will take a look

quinn-udp/src/windows.rs
```diff
@@ -163,4 +384,9 @@ fn set_socket_option(
 }
 
 pub(crate) const BATCH_SIZE: usize = 1;
+// Enough to store max(IP_PKTINFO + IP_ECN, IPV6_PKTINFO + IPV6_ECN) bytes (header + data) and some extra margin
+const CMSG_LEN: usize = 128;
 const OPTION_ON: u32 = 1;
+
+// FIXME this could use [`std::sync::OnceLock`] once the MSRV is bumped to 1.70 and upper
+static WSARECVMSG_PTR: OnceCell<WinSock::LPFN_WSARECVMSG> = OnceCell::new();
```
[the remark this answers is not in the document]
stormshield-damiend: ok i will do it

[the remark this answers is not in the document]
stormshield-damiend: to do that i need to create a dummy udp socket, then lookup the function ptr on it then close it, is it fine with you ?

quinn-udp/src/windows.rs
stormshield-damiend: I had no time to test this, the code is different from the unix side and i dunno how windows work with dual stack sockets. Does quinn has a dual stack regression tests ?

Ralith: We have basic end-to-end tests, but those won't detect missing metadata. Windows IPv6 sockets are not dual-stack by default but may be configured as such. We should probably add some quinn-udp unit tests for this.

I think we can consider this non-blocking, because the status quo doesn't support these features *ever*.

quinn-udp/src/unix.rs
```diff
@@ -43,6 +43,11 @@ pub struct UdpSocketState {
 }
 
 impl UdpSocketState {
+    /// Tokio-compatible UDP socket with some useful specializations
```
Ralith, line 46: nit: this crate is completely decoupled from async runtime concerns. Maybe just say "QUIC-friendly" as in Windows below?

quinn-udp/src/windows.rs
```diff
@@ -1,23 +1,46 @@
...
 pub struct UdpSocketState {
     last_send_error: Mutex<Instant>,
 }
 
 impl UdpSocketState {
+    /// Create a QUIC-friendly UDP socket for Windows
+    ///
+    /// # Panics
+    /// - If `CMSG_LEN` is too small to store a [`WinSock::IN6_PKTINFO`] and one `i32` (for ECN)
+    /// - If [`WinSock::CMSGHDR`] has stricter alignment requirements than [`cmsg::Aligned`]
```
Ralith, line 32: `cmsg::Aligned` is an impl detail, so it's not useful to mention in the public docs. CI should catch such a case before anyone hits it anyway, so let's just cut the second bullet point here.

quinn-udp/src/windows.rs
```diff
@@ -76,30 +114,100 @@ impl UdpSocketState {
...
+                        encoder.push(WinSock::IPPROTO_IPV6, WinSock::IPV6_PKTINFO, pktinfo);
+                    }
+                    _ => {
+                        return Err(io::Error::from(io::ErrorKind::InvalidInput));
+                    }
                 }
+            }
+
+            // ECN is a C integer https://learn.microsoft.com/en-us/windows/win32/winsock/winsock-ecn
+            let ecn = transmit.ecn.map_or(0, |x| x as i32);
```
Ralith, line 176: Shouldn't we use the C integer type, then?
```suggestion
            // ECN is a C integer https://learn.microsoft.com/en-us/windows/win32/winsock/winsock-ecn
            let ecn = transmit.ecn.map_or(0, |x| x as libc::c_int);
```

quinn-udp/src/windows.rs
```diff
@@ -111,20 +219,92 @@ impl UdpSocketState {
...
+                return Err(io::Error::last_os_error());
+            }
+        }
+
+        let addr = unsafe {
+            let (_, addr) = socket2::SockAddr::try_init(|addr_storage, len| {
+                *len = mem::size_of_val(&source) as _;
+                ptr::copy_nonoverlapping(&source, addr_storage as _, 1);
+                Ok(())
+            })?;
```
Ralith, line 265: nit: if we defined `source` as a `sockaddr_storage`, we could use the simpler `SockAddr::new` instead here.

quinn-udp/src/windows.rs
```diff
@@ -111,20 +219,92 @@ impl UdpSocketState {
...
+                    dst_ip = Some(IpAddr::from(unsafe { pktinfo.ipi6_addr.u.Byte }));
+                }
+                (WinSock::IPPROTO_IP, WinSock::IP_ECN) => {
+                    // ECN is a C integer https://learn.microsoft.com/en-us/windows/win32/winsock/winsock-ecn
+                    ecn_bits = unsafe { cmsg::decode::<i32, WinSock::CMSGHDR>(cmsg) };
+                }
+                (WinSock::IPPROTO_IPV6, WinSock::IPV6_ECN) => {
+                    // ECN is a C integer https://learn.microsoft.com/en-us/windows/win32/winsock/winsock-ecn
+                    ecn_bits = unsafe { cmsg::decode::<i32, WinSock::CMSGHDR>(cmsg) };
+                }
```
Ralith, line 297: As above:
```suggestion
                (WinSock::IPPROTO_IP, WinSock::IP_ECN) => {
                    // ECN is a C integer https://learn.microsoft.com/en-us/windows/win32/winsock/winsock-ecn
                    ecn_bits = unsafe { cmsg::decode::<libc::c_int, WinSock::CMSGHDR>(cmsg) };
                }
                (WinSock::IPPROTO_IPV6, WinSock::IPV6_ECN) => {
                    // ECN is a C integer https://learn.microsoft.com/en-us/windows/win32/winsock/winsock-ecn
                    ecn_bits = unsafe { cmsg::decode::<libc::c_int, WinSock::CMSGHDR>(cmsg) };
                }
```

quinn-udp/src/windows.rs
```diff
@@ -154,4 +334,72 @@ impl UdpSocketState {
...
 pub(crate) const BATCH_SIZE: usize = 1;
+// Enough to store max(IP_PKTINFO + IP_ECN, IPV6_PKTINFO + IPV6_ECN) bytes (header + data) and some extra margin
+const CMSG_LEN: usize = 128;
+const OPTION_ON: u32 = 1;
+
+// FIXME this could use [`std::sync::OnceLock`] once the MSRV is bumped to 1.70 and upper
+static WSARECVMSG_PTR: Lazy<WinSock::LPFN_WSARECVMSG> = Lazy::new(|| {
+    let s = unsafe { WinSock::socket(WinSock::AF_INET as _, WinSock::SOCK_DGRAM as _, 0) };
+    if s == WinSock::INVALID_SOCKET {
+        tracing::debug!("ignoring WSARecvMsg function pointer due to socket creation error");
```
Ralith, line 368: nit: it'd be nice to log the actual error here and below, presumably via `io::Error::last_os_error()`.

Ralith:
Rebased to include the new quinn-udp tests, but the cfg gates in quinn-udp/tests/tests.rs will need to be relaxed to validate the new functionality.

quinn-udp/src/windows.rs
Ralith: This is impossible because socket2 is still on an older version of windows-sys.

Ralith:
Addressed outstanding comments. Currently failing:
- `ecn_v4`: ECN codepoint is not successfully set on transmit (per Wireshark) though we do seem to correctly receive the unset state.
- `ecn_v6`: No packet is received for the dual-stack test.

Ralith:
Dropped ECN CE from the test and fixed the IPv4 ECN cmsg. Should be good to go now.

Ralith (approved):

Ralith:
Added GSO and GRO support while I was at it. I wasn't able to observe GRO working in loopback tests, but at least it isn't hurting.

quinn-udp/src/windows.rs
```diff
@@ -1,23 +1,42 @@
...
 pub struct UdpSocketState {
     last_send_error: Mutex<Instant>,
 }
 
 impl UdpSocketState {
     pub fn new(socket: UdpSockRef<'_>) -> io::Result<Self> {
+        assert!(
+            CMSG_LEN
+                >= WinSock::CMSGHDR::cmsg_space(mem::size_of::<WinSock::IN6_PKTINFO>())
+                    + WinSock::CMSGHDR::cmsg_space(mem::size_of::<i32>())
```
stormshield-damiend, line 35: c_int instead of i32

quinn-udp/src/windows.rs
```diff
@@ -17,15 +17,17 @@ use crate::{
...
 #[derive(Debug)]
 pub struct UdpSocketState {
     last_send_error: Mutex<Instant>,
 }
 
 impl UdpSocketState {
+    /// Create a helper for `sock`
+    ///
+    /// # Panics
+    /// - If `CMSG_LEN` is too small to store a [`WinSock::IN6_PKTINFO`] and one `i32` (for ECN)
```
stormshield-damiend, line 30: c_int instead of i32

quinn-udp/tests/tests.rs
```diff
@@ -87,6 +87,30 @@ fn ecn_v4() {
     }
 }
 
+#[test]
+#[cfg_attr(not(target_os = "linux"), ignore)]
+fn gso() {
+    let send = UdpSocket::bind("[::1]:0").unwrap();
```
stormshield-damiend, line 93: This will fail if IPv6 support is disabled at OS level, i can modify this and basic test to fallback to Ipv4 loopback in case we fail to create the Ipv6 socket.

quinn-udp/src/windows.rs
```diff
@@ -178,6 +178,15 @@ impl UdpSocketState {
                 encoder.push(WinSock::IPPROTO_IPV6, WinSock::IPV6_ECN, ecn);
             }
 
+            // Segment size is a u32 https://learn.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-wsasetudpsendmessagesize
```
stormshield-damiend, line 192: this should be documented close to `CMSG_LEN` L374 and also taken into account in the assert in the `UdpSocketState::new()` function

quinn-udp/tests/tests.rs
```diff
@@ -10,8 +10,12 @@ use socket2::Socket;
...
 #[test]
 fn basic() {
-    let send = UdpSocket::bind("[::1]:0").unwrap();
-    let recv = UdpSocket::bind("[::1]:0").unwrap();
+    let send = UdpSocket::bind("[::1]:0")
+        .or_else(|_| UdpSocket::bind("127.0.0.1:0"))
+        .unwrap();
+    let recv = UdpSocket::bind("[::1]:0")
+        .or_else(|_| UdpSocket::bind("127.0.0.1:0"))
+        .unwrap();
```
Ralith, line 18: Could we just use IPv4 here (and below) unconditionally for simplicity?

quinn-udp/src/windows.rs
```diff
@@ -1,23 +1,46 @@
...
 pub struct UdpSocketState {
     last_send_error: Mutex<Instant>,
 }
 
 impl UdpSocketState {
+    /// Create a helper for `sock`
+    ///
+    /// # Panics
+    /// - If `CMSG_LEN` is too small to store a [`WinSock::IN6_PKTINFO`], one `libc::c_int` (for ECN)
+    ///   and one `u32` (for Segment size)
```
Ralith, line 31: On review, I note this is *also* concerned with implementation details that don't mean anything to users. Although this contradicts @djc's guidance, I think this note risks adding more confusion than it avoids: a reader should immediately ask what `CMSG_LEN` is and what they have to do about it.

quinn-udp/src/windows.rs
```diff
@@ -154,4 +369,98 @@ impl UdpSocketState {
...
+    };
+
+    match rc == 0 {
+        true => Ok(()),
+        false => Err(io::Error::last_os_error()),
+    }
+}
+
 pub(crate) const BATCH_SIZE: usize = 1;
+// Enough to store max(IP_PKTINFO + IP_ECN, IPV6_PKTINFO + IPV6_ECN) + UDP_SEND_MSG_SIZE bytes (header + data) and some extra margin
```
Ralith, line 395: `max(UDP_SEND_MSG_SIZE, UDP_COALESCED_INFO)`, right?

quinn-udp/src/windows.rs
```diff
@@ -154,4 +369,98 @@ impl UdpSocketState {
...
+    unsafe {
+        WinSock::closesocket(s);
+    }
+
+    wsa_recvmsg_ptr
+});
+
+static MAX_GSO_SEGMENTS: Lazy<usize> = Lazy::new(|| {
+    let socket = match std::net::UdpSocket::bind("[::]:0")
+        .or_else(|_| std::net::UdpSocket::bind("127.0.0.1:0"))
```
Ralith, line 450: Should we use IPv4 unconditionally here? I guess maybe IPv4 could also be disabled, so we'd want to try both for robustness...

Ralith (approved):

quinn-udp/src/windows.rs
stormshield-damiend: Yes, good catch, fixing it

quinn-udp/src/windows.rs
stormshield-damiend: We have same code for [linux gso detection](https://github.com/quinn-rs/quinn/blob/55234e178fdca81cd51a5bfb520cb912de14f72e/quinn-udp/src/unix.rs#L758) i think it is more robust to keep it that way.

quinn-udp/src/windows.rs
stormshield-damiend: Ok, lets wait for @djc

stormshield-damiend: Maybe we could use a more generic documentation like for examples : 
# Panics
- If some operating system dependent invariant are not valid

Ralith: I dunno, that doesn't seem actionable for users. A panic here is always a bug on our part, IMO, and hence isn't really part of the public API, any more than a panic due to indexing out of bounds would be.

stormshield-damiend: We could replace these checks by [compile time errors](https://doc.rust-lang.org/std/macro.compile_error.html) ?

Ralith: That would be nice, but `libc`'s bindings of the cmsg macros aren't `const fn`s. In practice, test coverage on supported platforms is an adequate approximation of a compile-time error, IMO.

stormshield-damiend: So we could add this as two regression tests (one for windows and one for unix) instead of an assert within the `new()` ?

quinn-udp/src/windows.rs
```diff
@@ -38,68 +61,180 @@ impl UdpSocketState {
...
+        );
+
         let now = Instant::now();
         Ok(Self {
             last_send_error: Mutex::new(now.checked_sub(2 * IO_ERROR_LOG_INTERVAL).unwrap_or(now)),
         })
     }
 
-    pub fn send(&self, socket: UdpSockRef<'_>, transmits: &[Transmit]) -> Result<usize, io::Error> {
+    pub fn send(&self, socket: UdpSockRef<'_>, transmits: &[Transmit]) -> io::Result<usize> {
```
djc, line 128: Can we just revert the spelling change here from `Result<usize, io::Error>` to `io::Result<usize>`? IMO the former is more obvious.

djc (commented):
(Not a full review, just reviewing for semver incompatibility changes -- I didn't spot any.)

quinn-udp/src/windows.rs
stormshield-damiend: Sure i will do it

quinn-udp/src/windows.rs
djc: I'm on board with reverting the addition of this note if this is not actionable for the caller.

djc (approved):

quinn-udp/src/windows.rs
stormshield-damiend: ok il will do it

quinn-udp/tests/tests.rs
stormshield-damiend: I think keeping it like that make tests more robust, but i am fine to use ipv4 only if you want.

quinn-udp/src/windows.rs
stormshield-damiend: @djc `unix.rs` has `send()` use `Result<usize, io::Error>` and `recv()` use `io::Result<usize>`. Want me to make all (`unix.rs` and `windows.rs`) use `Result<usize, io::Error>`  or `io::Result<usize>` ?

Ralith: Using `io::Result` is always preferred, and style fixes to existing code in that vein would be absolutely welcome. Not a blocker for this PR though.

quinn-udp/tests/tests.rs
Ralith: Sure, I'm happy either way. It's pretty terse after all.

Ralith:
Thanks for taking care of this! This is a huge and long overdue improvement to our performance and reliability on Windows.

PR #1702, opened 2023-11-03
proto: add a way to configure the minimum MTU change needed to stop the binary search used during the MTU discovery phase.

stormshield-damiend:
Previously the binary search algorithm used during the MTU discovery phase was stopping itself if the discovered MTU was within -20 / +20 of the previous detected one.

This could lead to finding a MTU 40 bytes smaller of the real one in the worst case.

This change allow changing the default minimum change value if more precision is needed.

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -552,6 +552,7 @@ impl Default for AckFrequencyConfig {
 pub struct MtuDiscoveryConfig {
     pub(crate) interval: Duration,
     pub(crate) upper_bound: u16,
+    pub(crate) minimum_change: u16,
     pub(crate) black_hole_cooldown: Duration,
 }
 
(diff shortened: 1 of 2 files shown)


djc:
What value would you use for this in your project?

stormshield-damiend:
> What value would you use for this in your project?

Hi djc, as you know we do packet tunneling with Quic, thus the MTU is quite important for us as it is in direct link with the possible throughput of the tunnel.
With the default value during a real-life test with someone in Spain connecting to an AWS VM, Quinn find the estimated MTU (40 lower than the real one) in 3 search phases. With a value of 1 (so -1 / +1 from real MTU), Quinn find the real MTU with 2 more iterations.

We think that 1 is too much aggressive for the default value, thus we propose this API, which allow changing the default value when needed.

djc:
Okay, so you intend to configure this all the way to 1?

stormshield-damiend:
> Okay, so you intend to configure this all the way to 1?

Yes one or a small value, we need to run more real-life test to see how it goes in term of search convergence.

Ralith (approved):

PR #1703, opened 2023-11-06
Removed assert to check finalize is called before drop

boserohan:
Removed the debug_assert statement in fn finalize_inner() for checking finalize is called before drop on Chunks. Also removed the boolean 'drop' variable used for this check.

### DIFF
--- a/quinn-proto/src/connection/streams/recv.rs
+++ b/quinn-proto/src/connection/streams/recv.rs
@@ -281,15 +281,12 @@ impl<'a> Chunks<'a> {
 
     /// Finalize
     pub fn finalize(mut self) -> ShouldTransmit {
-        self.finalize_inner(false)
+        self.finalize_inner()
     }
 
-    fn finalize_inner(&mut self, drop: bool) -> ShouldTransmit {
+    fn finalize_inner(&mut self) -> ShouldTransmit {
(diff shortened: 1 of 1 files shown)


djc (commented):
The code changes look good. It would be nice to extend the documentation around this type to explain what `finalize()` does and what happens if it's not called.

boserohan91:
Thanks for reviewing. Sure, can you please point me to the location in the docs where I can make those changes?

djc:
Just the docstrings for the relevant methods.

boserohan91:
To me it seems `finalize()` needs to be called whenever all chunks of data available to the receiver on a given stream have been read (i.e. stream is finished or reset). Hence, disposing it's stored state. This might also involve situations where the sender is blocked due to flow control limits, and thus preventing the stream from being stalled by additionally granting these credits to the sender, and notifying the QUIC state machine that a frame needs to be transmitted for the same. Is my understanding correct with this? If not, please could you correct me if I am wrong.

boserohan91:
@djc can you help my understanding on this please?

djc:
@boserohan91 I don't have much time right now to dig into this. I suggest just updating your PR with some documentation based on your understanding and we'll review the suggested changes to see if they make sense.

PR #1708, opened 2023-11-10
Update docs for `Connection::accept_bi()` and `Connection::open_bi()`

rickvanprim:
If this is too link heavy, I can remove the links and just leave the plain backtick form.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -282,7 +282,12 @@ impl Connection {
     ///
     /// Streams are cheap and instantaneous to open unless blocked by flow control. As a
     /// consequence, the peer won't be notified that a stream has been opened until the stream is
-    /// actually used.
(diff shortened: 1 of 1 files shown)


quinn/src/connection.rs
```diff
@@ -282,7 +282,12 @@ impl Connection {
     ///
     /// Streams are cheap and instantaneous to open unless blocked by flow control. As a
     /// consequence, the peer won't be notified that a stream has been opened until the stream is
-    /// actually used.
+    /// actually used. Calling [`open_bi()`] then waiting on the [`RecvStream`] without writing
+    /// anything to [`SendStream`] will never return.
```
Ralith, line 286: nit: such a future *will* return when the connection is closed.
```suggestion
    /// anything to [`SendStream`] will never succeed.
```
(and similarly below)

Ralith (approved):
Thanks! That mistake has bitten a number of folks; hopefully this will help.

djc (approved):

PR #1713, opened 2023-11-30
Remove "max_tlps" configuration variable

link2xt:
It is not used since
<https://github.com/quinn-rs/quinn/pull/134>
(commit bdc9ac76a453208848a58bf3b24a633ade8b299a)
was merged in 2018.

RFC 9002 also does not mention Tail Loss Probe
except that it has been integrated into the PTO (Probe Timeout)
mechanism.


Ralith (approved):
Great catch, thanks! We have some breakage pending anyway.

PR #1721, opened 2023-12-07
proto: yield transport error for Initial packets with no CRYPTO

djc:


quinn-proto/src/connection/mod.rs
```diff
@@ -1812,10 +1812,25 @@ impl Connection {
...
         if let Some(data) = remaining {
             self.handle_coalesced(now, remote, ecn, data);
         }
+
+        if self.highest_space == SpaceId::Initial && self.state.is_handshake() {
+            // "The payload of an Initial packet includes a CRYPTO frame (or frames) containing a
+            // cryptographic handshake message, ACK frames, or both."
+            // https://www.rfc-editor.org/rfc/rfc9000.html#section-17.2.2
+            let space = &self.spaces[SpaceId::Initial];
+            if space.crypto_stream.bytes_read() == 0 {
```
Ralith, line 1826: There might be Initial packets other than the first packet. Any reason not to check those too? We could place the logic in `process_early_payload`.

djc: Hmm, the spec specifically talks about the first packet:

> The first packet sent by a client always includes a CRYPTO frame that contains the start or all of the first cryptographic handshake message.

For other frames, it looks like PING is also allowed, for example.

Ralith: Ah, you're right.

Ralith:
From Windows CI:

> test tests::handshake_anti_deadlock_probe has been running for over 60 seconds
> memory allocation of 24696061952 bytes failed

That's a bit spooky.

djc:
Reverted this to the initial approach, but updated the RFC comment to more closely match the code.

Ralith (approved):

PR #1715, opened 2023-12-04
Update rustls to 0.23 and ring to 0.17

djc:

PR #1723, opened 2023-12-09
Update rustls to 0.22

ibigbug:
so that the traits are implemented for rustls configs


Ralith:
See https://github.com/quinn-rs/quinn/pull/1715.

Issue #1131, opened 2021-05-26
Unbounded queues might lead to excessive memory usage


Matthias247:
The endpoint and per connection logic is running on individual tokio tasks. To communicate between those tasks, unbounded [`futures::channel::mpsc`](https://docs.rs/futures/0.3.15/futures/channel/mpsc/fn.unbounded.html) channels are used.
In specific there are 2 channels:
- One which carries endpoint events (e.g. received datagrams) from the endpoint to the connection task
- Another one which carries connection events (e.g. produced datagrams) to the endpoint task

The fact that these channels are unbounded raises the question on what exactly will happen if one of the tasks is producing data faster than the other task can consume it. The question probably became more important recently due to [adding additional fairness](https://github.com/quinn-rs/quinn/pull/1119) between tasks: Since the endpoint task will yield more often, the connection task might have more chances to enqueue new data that can't be sent right away.

The usual problems with this design are unbounded memory growth - and that as soon as the queue reached a certain size the components mostly work on outdated/non-interesting data, latencies will be terrible and reliability too.

Let's briefly look at each side individually:

### Endpoint task -> Connection task

This queue is used for forwarding received datagrams. In the connection task all received datagrams are currently immediately processed. As long as single-threaded runtime is used, the queue is not really unbounded since no new datagrams won't be enqueued by the endpoint as long as the connection is working on the existing ones. Instead if more datagrams would be enqueued in the UDP socket, they would get dropped there.

I think this is ok for the moment.

### Connection task -> Endpoint task

The connection produces datagrams whenever it is deemed ok by the congestion controller, and will enqueue them on the endpoint for transmission. I think this queue can get problematic in some situations. E.g. when either the connection task gets scheduled often enough between endpoint task iterations that it produces more datagrams than the endpoint can send in one iteration, or if multiple connections all have data to send.

I don't really think that letting the endpoint task perform `write`s until the queue is drained would be the right solution - it would again lead to long times where the stack doesn't process either peer data (ACKs) or user data.

**I can at the moment see a variety of ways to improve this:**
1. The simplest: Use a bounded channel between the components and `try_send`. If the queue is full because the endpoint can't keep up, the packet will get dropped and the congestion controller will deal with it. Downsides:
    1. We spend all the CPU time for producing a packet (it's expensive!), and then just drop it
    2. Might not scale will in a scenario where an endpoint is serving a high amount of connections. There doesn't seem any fairness regarding which packets will get dropped (although one could claim random == fair?).
[...]

PR #1726, opened 2023-12-17
Proto-layer endpoint/connection communication

Ralith:
Gives us freedom to adopt more specialized communication strategies (e.g. atomics or fine-grained locks) between endpoints and connections. These will need to be implemented at the proto layer, because optimal strategies will be sensitive to the specific data proto-private being communicated, e.g. `NeedIdentifiers` as an atomic counter.

Simplified take on https://github.com/quinn-rs/quinn/pull/1650, pushing cheeky data structures to future work.


quinn/src/connection.rs
```diff
@@ -899,11 +899,15 @@ impl State {
     }
 
     fn forward_endpoint_events(&mut self) {
-        while let Some(event) = self.inner.poll_endpoint_events() {
+        if self.inner.poll_endpoint_events() {
```
djc, line 913: Invert for early return?

quinn/Cargo.toml
```diff
@@ -46,6 +46,7 @@ thiserror = "1.0.21"
 tracing = "0.1.10"
 tokio = { version = "1.28.1", features = ["sync"] }
 udp = { package = "quinn-udp", path = "../quinn-udp", version = "0.5", default-features = false }
+atomic-waker = "1.1.2"
```
djc, line 49: Nit: alphabetic order?

djc (commented):
Do you mind writing a more detailed description of how this PR makes this more modular? I skimmed the commits but the new structure (or, how the structure changes) is still a bit hazy.

quinn/Cargo.toml
Ralith: Oops, I thought `cargo add` would preserve that.

Ralith:
The main effect of this PR is to replace the requirement that quinn-proto users pass endpoint and connection events around themselves with the requirement that users trigger wakeups when events have occurred, with quinn-proto managing the actual message-passing internally. This is not necessarily more modular, but could be useful because it allows future work to replace naive message-passing with more specialized structures as in #1650. A more specialized communications structure could be lighter: an `AtomicUsize` counter is much lighter than message-passing for requesting new connection IDs and has constant memory overhead, for example. We still need to dispatch wakeups, but unlike messages, those are idempotent and hence don't need to be buffered, which might simplify downstream code.

It's not obvious that the benefit here is significant as it stands. `quinn` wasn't simplified, because it still needs the same channels for other purposes. Perhaps those can be addressed in future work; e.g. `quinn::ConnectionEvent` could probably be reduced to a set of shared flags.

My real goal is to work towards fine-grained interior mutability in `quinn-proto`, particularly `Endpoint`, to eventually support parallel endpoint drivers. I don't think this PR moves us in that direction.

This is also a step towards resolving https://github.com/quinn-rs/quinn/issues/1131, but the real substance of the hazard described there is in handling datagrams, which is unchanged, and could be addressed without this PR.

This effort was ultimately exploratory, and I'm open to shelving it until it's more strongly motivated.

PR #1729, opened 2023-12-19
Transmit directly from connection tasks

Ralith:
This allows outgoing data to parallelize perfectly. Initial informal testing suggests a performance improvement for bulk data, likely due to reduced allocation and cross-task messaging. A larger performance benefit should be expected for endpoints hosting numerous connections.

I'd left the original userspace-multiplexed transmit strategy in place for a long time because I assumed the UDP socket had a mutex-guarded send buffer, so leaning into task-parallelism for sending would just lead to contention. This turns out to be complete nonsense, at least on Linux: outgoing UDP datagrams are buffered by the kernel with dynamically allocated memory, and the primitive NIC queuing operation is apparently scalable, as borne out by empirical testing. The following minimal test scales almost linearly with physical parallelism on Linux:

```rs
use std::{
    net::{SocketAddr, UdpSocket},
    thread,
    time::Instant,
};

fn main() {
    let sock = UdpSocket::bind("[::]:0").unwrap();
    let target: SocketAddr = "[::1]:1234".parse().unwrap();
    let start = Instant::now();
    const DATAGRAMS: usize = 1_000_000;
    thread::scope(|scope| {
        let threads = thread::available_parallelism().unwrap().get();
        dbg!(threads);
        let each = DATAGRAMS / threads;
        let sock = &sock;
        for _ in 0..threads {
            scope.spawn(move || {
                for _ in 0..each {
                    sock.send_to(b"hello, world", target).unwrap();
                }
            });
        }
    });
    let seconds = start.elapsed().as_secs_f32();
    println!("{} pps", DATAGRAMS as f32 / seconds);
}
```

It'd be interesting to see how this compares on other major platforms, but Linux's drastic improvement and the simplification of Quinn's internals are enough to satisfy me that we should go this way.

### DIFF
--- a/quinn-proto/src/tests/util.rs
+++ b/quinn-proto/src/tests/util.rs
@@ -350,7 +350,7 @@ impl TestEndpoint {
             }
         }
         let buffer_size = self.endpoint.config().get_max_udp_payload_size() as usize;
-        let mut buf = BytesMut::with_capacity(buffer_size);
+        let mut buf = Vec::with_capacity(buffer_size);
 
         while self.inbound.front().map_or(false, |x| x.0 <= now) {
             let (recv_time, ecn, packet) = self.inbound.pop_front().unwrap();
@@ -386,8 +386,8 @@ impl TestEndpoint {
                     }
                     DatagramEvent::Response(transmit) => {
                         let size = transmit.size;
-                        self.outbound
-                            .extend(split_transmit(transmit, buf.split_to(size).freeze()));
+                        self.outbound.extend(split_transmit(transmit, &buf[..size]));
+                        buf.clear();
                     }
                 }
             }
@@ -396,7 +396,7 @@ impl TestEndpoint {
 
     pub(super) fn drive_outgoing(&mut self, now: Instant) {
         let buffer_size = self.endpoint.config().get_max_udp_payload_size() as usize;
-        let mut buf = BytesMut::with_capacity(buffer_size);
+        let mut buf = Vec::with_capacity(buffer_size);
 
         loop {
             let mut endpoint_events: Vec<(ConnectionHandle, EndpointEvent)> = vec![];
@@ -417,8 +417,8 @@ impl TestEndpoint {
                 }
                 while let Some(transmit) = conn.poll_transmit(now, MAX_DATAGRAMS, &mut buf) {
                     let size = transmit.size;
-                    self.outbound
-                        .extend(split_transmit(transmit, buf.split_to(size).freeze()));
+                    self.outbound.extend(split_transmit(transmit, &buf[..size]));
+                    buf.clear();
                 }
                 self.timeout = conn.poll_timeout();
             }
@@ -460,7 +460,7 @@ impl TestEndpoint {
         incoming: Incoming,
         now: Instant,
     ) -> Result<ConnectionHandle, ConnectionError> {
-        let mut buf = BytesMut::new();
+        let mut buf = Vec::new();
         match self.endpoint.accept(incoming, now, &mut buf, None) {
             Ok((ch, conn)) => {
                 self.connections.insert(ch, conn);
@@ -470,8 +470,7 @@ impl TestEndpoint {
             Err(error) => {
                 if let Some(transmit) = error.response {
                     let size = transmit.size;
-                    self.outbound
-                        .extend(split_transmit(transmit, buf.split_to(size).freeze()));
+                    self.outbound.extend(split_transmit(transmit, &buf[..size]));
                 }
                 self.accepted = Some(Err(error.cause.clone()));
                 Err(error.cause)
@@ -480,19 +479,17 @@ impl TestEndpoint {
     }
 
     pub(super) fn retry(&mut self, incoming: Incoming) {
-        let mut buf = BytesMut::new();
+        let mut buf = Vec::new();
         let transmit = self.endpoint.retry(incoming, &mut buf).unwrap();
         let size = transmit.size;
-        self.outbound
-            .extend(split_transmit(transmit, buf.split_to(size).freeze()));
+        self.outbound.extend(split_transmit(transmit, &buf[..size]));
     }
 
     pub(super) fn reject(&mut self, incoming: Incoming) {
-        let mut buf = BytesMut::new();
+        let mut buf = Vec::new();
         let transmit = self.endpoint.refuse(incoming, &mut buf);
         let size = transmit.size;
-        self.outbound
-            .extend(split_transmit(transmit, buf.split_to(size).freeze()));
+        self.outbound.extend(split_transmit(transmit, &buf[..size]));
     }
 
     pub(super) fn assert_accept(&mut self) -> ConnectionHandle {
@@ -653,7 +650,8 @@ pub(super) fn min_opt<T: Ord>(x: Option<T>, y: Option<T>) -> Option<T> {
 /// The maximum of datagrams TestEndpoint will produce via `poll_transmit`
 const MAX_DATAGRAMS: usize = 10;
 
-fn split_transmit(transmit: Transmit, mut buffer: Bytes) -> Vec<(Transmit, Bytes)> {
+fn split_transmit(transmit: Transmit, buffer: &[u8]) -> Vec<(Transmit, Bytes)> {
+    let mut buffer = Bytes::copy_from_slice(buffer);
     let segment_size = match transmit.segment_size {
         Some(segment_size) => segment_size,
         _ => return vec![(transmit, buffer)],
--- a/quinn-proto/src/tests/util.rs
+++ b/quinn-proto/src/tests/util.rs
@@ -667,7 +667,7 @@ fn split_transmit(transmit: Transmit, mut buffer: Bytes) -> Vec<(Transmit, Bytes
         transmits.push((
             Transmit {
                 destination: transmit.destination,
-                size: buffer.len(),
+                size: contents.len(),
                 ecn: transmit.ecn,
                 segment_size: None,
                 src_ip: transmit.src_ip,
--- a/quinn-udp/src/fallback.rs
+++ b/quinn-udp/src/fallback.rs
@@ -24,37 +24,25 @@ impl UdpSocketState {
         })
     }
 
-    pub fn send(&self, socket: UdpSockRef<'_>, transmits: &[Transmit]) -> io::Result<usize> {
-        let mut sent = 0;
-        for transmit in transmits {
-            match socket.0.send_to(
-                &transmit.contents,
-                &socket2::SockAddr::from(transmit.destination),
-            ) {
-                Ok(_) => {
-                    sent += 1;
-                }
-                // We need to report that some packets were sent in this case, so we rely on
-                // errors being either harmlessly transient (in the case of WouldBlock) or
-                // recurring on the next call.
-                Err(_) if sent != 0 => return Ok(sent),
-                Err(e) => {
-                    if e.kind() == io::ErrorKind::WouldBlock {
-                        return Err(e);
-                    }
-
-                    // Other errors are ignored, since they will usually be handled
-                    // by higher level retransmits and timeouts.
-                    // - PermissionDenied errors have been observed due to iptable rules.
-                    //   Those are not fatal errors, since the
-                    //   configuration can be dynamically changed.
-                    // - Destination unreachable errors have been observed for other
-                    log_sendmsg_error(&self.last_send_error, e, transmit);
-                    sent += 1;
-                }
-            }
+    pub fn send(&self, socket: UdpSockRef<'_>, transmit: &Transmit) -> io::Result<()> {
+        let Err(e) = socket.0.send_to(
+            &transmit.contents,
+            &socket2::SockAddr::from(transmit.destination),
+        ) else {
+            return Ok(());
+        };
+        if e.kind() == io::ErrorKind::WouldBlock {
+            return Err(e);
         }
-        Ok(sent)
+
+        // Other errors are ignored, since they will usually be handled
+        // by higher level retransmits and timeouts.
+        // - PermissionDenied errors have been observed due to iptable rules.
+        //   Those are not fatal errors, since the
+        //   configuration can be dynamically changed.
+        // - Destination unreachable errors have been observed for other
+        log_sendmsg_error(&self.last_send_error, e, transmit);
+        Ok(())
     }
 
     pub fn recv(
--- a/quinn-udp/src/fallback.rs
+++ b/quinn-udp/src/fallback.rs
@@ -24,9 +24,9 @@ impl UdpSocketState {
         })
     }
 
-    pub fn send(&self, socket: UdpSockRef<'_>, transmit: &Transmit) -> io::Result<()> {
+    pub fn send(&self, socket: UdpSockRef<'_>, transmit: &Transmit<'_>) -> io::Result<()> {
         let Err(e) = socket.0.send_to(
-            &transmit.contents,
+            transmit.contents,
             &socket2::SockAddr::from(transmit.destination),
         ) else {
             return Ok(());
--- a/quinn-udp/src/fallback.rs
+++ b/quinn-udp/src/fallback.rs
@@ -4,9 +4,7 @@ use std::{
     time::Instant,
 };
 
-use proto::Transmit;
-
-use super::{log_sendmsg_error, RecvMeta, UdpSockRef, UdpState, IO_ERROR_LOG_INTERVAL};
+use super::{log_sendmsg_error, RecvMeta, Transmit, UdpSockRef, IO_ERROR_LOG_INTERVAL};
 
 /// Fallback UDP socket interface that stubs out all special functionality
 ///
@@ -18,20 +16,15 @@ pub struct UdpSocketState {
 }
 
 impl UdpSocketState {
-    pub fn new(socket: UdpSocketRef<'_>) -> io::Result<Self> {
+    pub fn new(socket: UdpSockRef<'_>) -> io::Result<Self> {
         socket.0.set_nonblocking(true)?;
         let now = Instant::now();
-        Self {
+        Ok(Self {
             last_send_error: Mutex::new(now.checked_sub(2 * IO_ERROR_LOG_INTERVAL).unwrap_or(now)),
-        }
+        })
     }
 
-    pub fn send(
-        &self,
-        socket: UdpSockRef<'_>,
-        _state: &UdpState,
-        transmits: &[Transmit],
-    ) -> io::Result<usize> {
+    pub fn send(&self, socket: UdpSockRef<'_>, transmits: &[Transmit]) -> io::Result<usize> {
         let mut sent = 0;
         for transmit in transmits {
             match socket.0.send_to(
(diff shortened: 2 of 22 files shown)


Ralith:
Hmm, this needs a mechanism to broadcast wakeups when the UDP socket is backpressured...

quinn/src/connection.rs
```diff
@@ -863,47 +876,56 @@ pub(crate) struct State {
...
+        }) {
+            let size = t.contents.len();
+            let segment_size = t.segment_size;
+            if let Poll::Ready(0) | Poll::Pending =
+                self.socket.poll_send(cx, std::slice::from_ref(&t))?
+            {
+                self.buffered_transmit = Some(t);
+                return Ok(false);
+            }
+            transmits += match segment_size {
```
djc, line 909: The diff here gets a bit confusing because I think you inlined this refactoring of moving around the `transmits` calculation? I think it's a good change, but if it can be a separate commit it would be good to do that.

quinn/src/endpoint.rs
```diff
@@ -452,22 +447,23 @@ impl State {
...
-                                    // Limiting the memory usage for items queued in the outgoing queue from endpoint
-                                    // generated packets. Otherwise, we may see a build-up of the queue under test with
-                                    // flood of initial packets against the endpoint. The sender with the sender-limiter
-                                    // may not keep up the pace of these packets queued into the queue.
-                                    if self.transmit_queue_contents_len
-                                        < MAX_TRANSMIT_QUEUE_CONTENTS_LEN
-                                    {
-                                        let contents_len = transmit.size;
-                                        self.outgoing.push_back(udp_transmit(
+                                    // Send if there's kernel buffer space; otherwise, drop it
```
djc, line 450: Ugh, I forgot everything about what this is. What are endpoint-generated datagrams used for? Maybe add some comments for context?

djc (commented):
Looks cool -- initial round of feedback trying to make sense of it all.

quinn/src/endpoint.rs
Ralith: Added substantially more detail.

quinn/src/connection.rs
```diff
@@ -863,47 +876,56 @@ pub(crate) struct State {
...
-        while let Some(t) = self.inner.poll_transmit(now, max_datagrams, &mut buffer) {
-            transmits += match t.segment_size {
+        while let Some(t) = self.buffered_transmit.take().or_else(|| {
+            self.inner
+                .poll_transmit(now, max_datagrams, &mut buffer)
+                .map(|t| {
+                    let buffer = buffer.split_to(t.size).freeze();
+                    udp_transmit(t, buffer)
+                })
+        }) {
```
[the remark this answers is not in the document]
Ralith: `match` works for me; applied.

quinn/src/endpoint.rs
```diff
@@ -533,35 +533,25 @@ impl State {
...
-                        }
-                        if let Some(event) = self.inner.handle_event(ch, e) {
-                            // Ignoring errors from dropped connections that haven't yet been cleaned up
-                            let _ = self
-                                .connections
-                                .senders
-                                .get_mut(&ch)
-                                .unwrap()
-                                .send(ConnectionEvent::Proto(event));
+                Poll::Ready(Some((ch, e))) => {
```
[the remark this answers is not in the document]
Ralith: Well spotted! IMO that makes sense independent of this change, though, so I've added it in a separate commit.

quinn/src/connection.rs
Ralith: Reduced the churn here. I don't think it actually matters much whether we count after successful `poll_transmit` or after `poll_send`, and the original pattern is a bit simpler.

Ralith:
I've replaced the original `try_clone_box` mechanism with the runtime's built-in concurrent I/O support, which exists in both tokio and async-std. This required a somewhat fiddly extension of the runtime abstraction layer, because for both runtimes, concurrent I/O is only possible if you use `async fn` interfaces, probably due to self-reference somewhere in the critical futures. Luckily, the fiddly bits are isolated almost exclusively within runtime.rs, with both implementations and callers remaining straightforward.

flub:
Hi, curious to try this out on various other platforms as well.  How have you been running benchmarks to compare this?

I guess any perf loss due to losing sendmmsg is offset by the perf gain of not doing the userspace-multiplexing.  sendmmsg does give some small benefit when you manage to fill the pipe to many peers in one syscall, but I'd have to spend more time figuring out how often quinn really manages to do that.

Ralith:
I haven't benchmarked this very rigorously yet, since I don't have convenient access to a machine that's both running Linux (for our most optimized UDP backend) and not a laptop (subject to unpredictable throttling). We could use more data on that, if you're interested. I do plan to do some benchmarking on Windows, but we have neither sendmmsg nor GSO there just yet, so it might be a bit unfair to the status quo.

Ralith:
Rebased to resolve conflicts with recent quinn-udp improvements.

PureWhiteWu:
Wow, I tried this pr, and this pr perfectly solves our problems, the server send bandwidth of 4500 connections raised from about 300MB/s to 900MB/s (the desired rate)!
Great work! Thank you very much! Really looking forward for this PR to be merged!

PureWhiteWu:
Also, just curious, is there any way to use `sendmmsg` back together with the perfect parallelization?

The CPU usage raised from about 400% to 680% with this PR (at a fixed sending rate of 900 MB/s), and the percent of syscall (previously `sendmmsg` and now `sendmsg`) raised from 25% to 45%.

If there's no way to achieve this, then I think the parallelization is more important.

PureWhiteWu:
Seems that when it comes to 9000 connections, still has some problems



Here's the flamegraph for 9000 connections: 

[flame.html.zip](https://github.com/quinn-rs/quinn/files/13928956/flame.html.zip)

Ralith:
> Wow, I tried this pr, and this pr perfectly solves our problems!

Awesome, thanks for the report! That's exactly the kind of data I've been hoping for, since this seems to be a no-op on our single-connection benchmarks despite reduced wakeups and allocation.

> Also, just curious, is there any way to use `sendmmsg` back together with the perfect parallelization?

When sending from a single connection at a time, we're almost invariably sending to a single address, so `sendmmsg` has nothing to offer compared to the segmentation offload we already have. It might be interesting to play with `MAX_TRANSMIT_SEGMENTS` in `quinn-proto/connection/mod.rs` to see if that improves your performance further; I think this change may have removed the drawbacks to raising that number, and perhaps we should do away with the additional limit entirely.

Hypothetically we could have one `sendmmsg` task per core, handling transmits for a proportional subset of connections. This could hypothetically reduce the number of system calls under some workloads, at the cost of greatly increased complexity to coordinate connections. It's unclear if this could actually realize a performance benefit in practice. QUIC seems to benefit far more from GSO than merely reducing system call rate; it seems that the slowest part of `sendmsg` is the per-datagram work in the kernel.

> The CPU usage raised from about 400% to 680% with this PR (at a fixed sending rate of 900 MB/s), and the percent of syscall (previously sendmmsg and now sendmsg) raised from 25% to 45%.

Given the speedup, this is mostly indicating that your cores are spending more time doing useful work, rather than sitting idle waiting for the endpoint task to do something.

> Seems that when it comes to 9000 connections(taskseted to 48C), still has some problems

There will always be some level of activity that can't be served within a given hardware budget, so it's not immediately obvious to me that this is a problem, unless your CPU is obviously being underutilized. There's always some wiggle room for tuning, of course. Try playing with `MAX_TRANSMIT_SEGMENTS` to get more mileage out of each call, and maybe your UDP socket's `SO_SNDBUF` and kernel receive buffer sizes in case packets are being lost there.

Nothing jumps out at me as too sketchy Quinn-wise on the flamegraph. However, it's interesting that `<tokio::time::sleep::Sleep as core::future::future::Future>::poll ` seems to take up almost 20% of CPU time per thread(?), and mostly in trying to acquire a mutex. This is probably contention around Tokio's global timer wheel. I brought this up briefly on the Tokio discord and it sounds like there's significant opportunity for reducing contention there, and they would likely welcome contributions to optimize it.

PureWhiteWu:
Thank you very much for your so detailed response!

> There will always be some level of activity that can't be served within a given hardware budget, so it's not immediately obvious to me that this is a problem, unless your CPU is obviously being underutilized.

Just a clarification, in the 9000 connections bench, the CPU limit is 4800% but only utilized about 1400%, so I think it's nothing to do with CPU usage.

I will try `MAX_TRANSMIT_SEGMENTS` later.

Really looking forward to this pr to be merged!

Ralith:
> Just a clarification, in the 9000 connections bench, the CPU limit is 4800% but only utilized about 1400%, so I think it's nothing to do with CPU usage.

Could be a buffer size issue, and/or receive-side bottlenecking on the endpoint task. I have some ideas for parallel receive tasks using `SO_REUSEADDR` which could help, though parallelizing intra-endpoint ops will take some care.

Scaling to truly large numbers will ultimately need some sort of load-balancing between multiple endpoints, potentially even on multiple machines, but we've definitely got room for improvement here still.

quinn/src/runtime.rs
```diff
@@ -31,10 +31,15 @@ pub trait AsyncTimer: Send + Debug + 'static {
...
-        -> Poll<Result<usize, io::Error>>;
+    /// Create a helper for awaiting I/O readiness
+    fn create_io_poller(self: Arc<Self>) -> Pin<Box<dyn UdpPoller>>;
+
+    /// Send UDP datagrams from `transmits`, or return `WouldBlock` and clear the underlying
+    /// socket's readiness, or return an I/O error
+    ///
+    /// If this returns [`io::ErrorKind::WouldBlock`], [`UdpPoller::poll_writable`] must be called
+    /// to register the calling task to be woken when a send should be attempted again.
+    fn try_send(&self, transmit: &Transmit) -> io::Result<()>;
```
dignifiedquire, line 42: I wish this would take `Transmit` owned (and if needed, returns `Transmit` in the error case)

djc: Why?

dignifiedquire: because we need to rewrite these before sending

Ralith: As of this PR, `Transmit` no longer owns any non-`Copy` data. Just `Clone` it.

dignifiedquire: fair enough 👍

quinn/src/runtime.rs
```diff
@@ -66,6 +71,53 @@ pub trait AsyncUdpSocket: Send + Sync + Debug + 'static {
...
+where
+    F: Fn() -> T + Send + Sync + 'static,
+    T: Future<Output = io::Result<()>> + Send + Sync + 'static,
+{
+    fn poll_writable(self: Pin<&mut Self>, cx: &mut Context) -> Poll<io::Result<()>> {
+        let mut this = self.project();
+        if this.fut.is_none() {
+            this.fut.set(Some((this.f)()));
+        }
+        let result = this.fut.as_mut().as_pin_mut().unwrap().poll(cx);
```
flub, line 107: I think you could do the option dance here to avoid using `.unwrap()`: take out the future or create it, poll it, put it back (or not if consumed).

Ralith: I think that would be unsound, because `T` may be `!Unpin`.

flub:
So I finally got round to doing some kind of decent tests on this.  I used our setup to notice performance regressions on our end.  It's not perfect, but it's some kind of indicator.

These numbers are all in gbps, but relative numbers indicate more than the absolute.
```
|         | 0.10 | 0.11 | parallel-transmit |
|---------+------+------+-------------------|
| 1-to-1  | 1.96 | 2.13 |              0.83 |
| 1-to-3  | 5.96 | 2.74 |              2.71 |
| 1-to-5  | 5.40 | 2.28 |              3.18 |
| 1-to-10 | 5.33 | 2.12 |              3.09 |
| 2-to-2  | 4.19 | 3.87 |              1.85 |
| 2-to-4  | 7.32 | 4.46 |              4.07 |
| 2-to-6  | 7.64 | 4.55 |              4.72 |
| 2-to-10 | 7.82 | 4.25 |              5.30 |
```

So I started with what is the current baseline, in the `0.10` column.  The `0.11` column is our stack ported to current quinn `main` branch.  Then the final column is it ported to the parallel-transmit branch.

There is already a pretty significant regression in the main branch, this porting was relatively easy so I'm fairly confident that is realistic.  This is a bit scary, I was hoping that the parallel-transmit branch would compensate for that.  Unfortunately it does not really.  This port was rather involved and it might well be that there are improvements in there, especially given that others have indicated this resulted in good speedups.  I know there are some places where I took shortcuts that won't do any good.  But I'm about to go on a holiday so wanted to post this early.

Now obviously we do weird things with our stack at https://github.com/n0-computer/iroh/.  We present a single `AsyncUdpSocket` to quinn and send via different sockets depending on the destination and whether we managed to holepunch etc.  We know that's not optimal and weird.  Regardless so far this is somewhat disappointing to me and will need further investigation on our end.

Ralith:
Thanks for the report! Consider opening a separate issue for regressions in 0.11 base, which are out of scope of this PR. A few questions:

- Can you reproduce this with the standard tokio-based `AsyncUdpSocket`?
- What is your sample size? Bandwidth tests tend to be noisy, especially if ran for short periods.
- How many threads do your endpoints have?
- How busy are those threads? Are you CPU-bound, synchronization-bound, congestion-bound, or something else?
- Does your custom `AsyncUdpSocket` have any internal synchronization? This wouldn't have been under any pressure previously, but could cause heavy contention in this branch.
- How does your `UdpPoller` work? Do you handle concurrent waiters correctly? Missed wakeups due to bugs here could cause arbitrarily bad performance with low CPU load.

Ralith:
Rebased.

quinn/src/connection.rs
```diff
@@ -59,8 +60,14 @@ impl Connecting {
...
 
+        let driver = ConnectionDriver(conn.clone());
         runtime.spawn(Box::pin(
-            ConnectionDriver(conn.clone()).instrument(Span::current()),
+            async {
+                if let Err(e) = driver.await {
+                    tracing::error!("I/O error: {e}");
+                }
+            }
+            .instrument(Span::current()),
```
djc, line 70: It feels like this commit has a lot going on, and could be split into smaller parts?

quinn/src/endpoint.rs
```diff
@@ -553,28 +553,28 @@ impl State {
...
                 }
+            };
+
+            if event.is_drained() {
+                self.connections.senders.remove(&ch);
+                if self.connections.is_empty() {
+                    shared.idle.notify_waiters();
+                }
+            }
+            if let Some(event) = self.inner.handle_event(ch, event) {
```
djc, line 528: `let-else` this to reduce drift even more?

quinn/src/endpoint.rs
```diff
@@ -582,40 +540,32 @@ impl State {
...
-    fn dequeue(&mut self, sent: usize) {
-        self.contents_len = self
-            .contents_len
-            .saturating_sub(self.outgoing.drain(..sent).map(|t| t.contents.len()).sum());
-    }
-
-    fn transmits(&self) -> &[udp::Transmit] {
-        self.outgoing.as_slices().0
-    }
+fn respond(socket: &dyn AsyncUdpSocket, transmit: proto::Transmit, response_buffer: &mut BytesMut) {
```
djc: I'm having trouble relating these changes to the commit message?

djc (commented):
Here's an initial round of feedback. Still not entirely sure I fully grok the way the commits build up.

quinn/src/endpoint.rs
Ralith: This commit makes transmits stateless, hence `TransmitState` is no longer needed. Instead of buffering up outgoing endpoint-generated datagrams, we attempt to send them immediately and discard them if that fails.

quinn/src/runtime.rs
```diff
@@ -31,10 +31,15 @@ pub trait AsyncTimer: Send + Debug + 'static {
 
 /// Abstract implementation of a UDP socket for runtime independence
 pub trait AsyncUdpSocket: Send + Sync + Debug + 'static {
-    /// Send UDP datagrams from `transmits`, or register to be woken if sending may succeed in the
-    /// future
-    fn poll_send(&self, cx: &mut Context, transmits: &[Transmit])
-        -> Poll<Result<usize, io::Error>>;
+    /// Create a helper for awaiting I/O readiness
```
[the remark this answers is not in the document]
Ralith: Substantially expanded docs on all the new items. Is it clearer now?

quinn/src/runtime.rs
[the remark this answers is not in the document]
Ralith: We cannot, due to `Pin`. This is the second time someone's proposed it, so I added a comment explaining why.

quinn/src/connection.rs
Ralith: Split into three parts:
- Adding an error path to `ConnectionDriver`
- Moving transmits into `ConnectionDriver`
- Removing `quinn::EndpointEvent`

Is that clearer? Neither of the newly split out commits are justified in isolation, but they do help isolate the actual behavior change.

Ralith:
Rebased.

quinn/src/connection.rs
```diff
@@ -933,40 +942,66 @@ pub(crate) struct State {
...
+            let result = self.socket.try_send(std::slice::from_ref(&t));
+            if result
+                .as_ref()
+                .map_or_else(|e| e.kind() == io::ErrorKind::WouldBlock, |&n| n == 0)
+            {
+                // Retry immediately
+                self.buffered_transmit = Some(t);
+                continue;
+            }
+            result?;
```
djc: This is pretty dense. This is one line more but maybe easier to follow?

```rust
let retry = match self.socket.try_send(std::slice::from_ref(&t)) {
    Ok(0) => true,
    Err(e) if e.kind() == io::ErrorKind::WouldBlock => true,
    Err(e) => return Err(err),
};

if retry {
    self.buffered_transmit = Some(t);
    continue;
}
```

quinn/src/endpoint.rs
```diff
@@ -580,40 +538,32 @@ impl State {
...
```
djc: Nit: make the `socket` go last?

quinn-udp/src/fallback.rs
```diff
@@ -26,12 +26,7 @@ impl UdpSocketState {
         }
     }
 
-    pub fn send(
-        &self,
-        socket: UdpSockRef<'_>,
-        _state: &UdpState,
-        transmits: &[Transmit],
-    ) -> io::Result<usize> {
+    pub fn send(&self, socket: UdpSockRef<'_>, transmits: &[Transmit]) -> io::Result<usize> {
```
djc: So we have no platforms in CI where this path is used? Maybe we should either get one or remove this code?

quinn-proto/src/tests/util.rs
```diff
@@ -620,7 +620,7 @@ fn split_transmit(transmit: Transmit, mut buffer: Bytes) -> Vec<(Transmit, Bytes
         transmits.push((
             Transmit {
                 destination: transmit.destination,
-                size: buffer.len(),
+                size: contents.len(),
```
djc, line 621: (Typo in commit message: "transimt" -> "transmit".)

djc (approved):
Looking good!

quinn/src/runtime.rs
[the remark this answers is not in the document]
djc: Yup, it's better. I still got a little confused because it's not that obvious that this increases waker state, but I'm not sure if there's a great fix for that.

quinn/src/connection.rs
Ralith: Oh, that's much better indeed.

quinn-udp/src/fallback.rs
Ralith: Strongly agree that we need CI coverage for this. In fact, I just hacked up udp/src/lib.rs to force use of this manually and found (and fixed) a bunch of other bitrot. Not sure what the best way to achieve that is, but for now let's get this merged.

PR #1730, opened 2023-12-21
Don't allocate an endpoint response buffer for every driver wakeup

Ralith:
Most packets (under good conditions), let alone most wakeups, don't require a response from the endpoint, so this was very wasteful. `BytesMut::new` [is guaranteed](https://docs.rs/bytes/latest/bytes/struct.BytesMut.html#growth) to grow automatically when needed.


djc (approved):
Nice!

PR #1731, opened 2023-12-22
Test UDP layer in isolation

Ralith:
Inspired by #1701. Should substantially reduce risk of breakage and ease supporting new platforms.

### DIFF
--- a/quinn-udp/tests/tests.rs
+++ b/quinn-udp/tests/tests.rs
@@ -1,0 +1,143 @@
+use std::{
+    io::IoSliceMut,
+    net::{Ipv4Addr, Ipv6Addr, SocketAddr, SocketAddrV4, SocketAddrV6, UdpSocket},
+    slice,
+};
+
+use bytes::Bytes;
+use quinn_udp::{EcnCodepoint, RecvMeta, Transmit, UdpSocketState};
+use socket2::Socket;
+
+#[test]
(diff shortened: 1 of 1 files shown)


Ralith:
Need someone with a mac to investigate the failures there. Not sure how they can be going wrong when our end-to-end tests pass...

Ralith:
Okay, it looks like it's just a bit racy. Weird.

Ralith:
There we go, just had to set it back in blocking mode.

quinn-udp/tests/tests.rs
```diff
@@ -0,0 +1,143 @@
...
+            (false, true) => assert_eq!(dst, Ipv4Addr::LOCALHOST.to_ipv6_mapped()),
+            (true, true) => assert_eq!(dst, Ipv6Addr::LOCALHOST),
+        }
+    }
+}
+
+fn to_v6_mapped(x: SocketAddr) -> SocketAddr {
+    match x {
+        SocketAddr::V4(x) => SocketAddrV6::new(x.ip().to_ipv6_mapped(), x.port(), 0, 0).into(),
+        x @ SocketAddr::V6(_) => x,
```
djc, line 141: Style nit: I think we don't need the `x` binding in this arm?

djc (approved):

PR #1734, opened 2024-01-04
Replace rustls-native-certs with rustls-platform-verifier

Ralith:
rustls-platform-verifier doesn't seem to support rustls 0.22 yet, so we might want to hold off on this in the hopes of not complicating #1715. Still, long-term, this seems like a safer default to promote.

### DIFF
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -19,8 +19,8 @@ maintenance = { status = "experimental" }
 [features]
 default = ["tls-rustls", "log"]
 tls-rustls = ["rustls", "ring"]
-# Provides `ClientConfig::with_native_roots()` convenience method
-native-certs = ["rustls-native-certs"]
+# Provides `ClientConfig::with_platform_verifier()` convenience method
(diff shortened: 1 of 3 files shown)


Ralith:
Looks like upstream is missing FreeBSD support :(

djc (approved):
Nice! I think rustls-platform-verifier 0.2 will be coming along soon enough.

Ralith:
FreeBSD support will be included in the next upstream release, which will probably also bump to rustls 0.22.

cpu:
> the next upstream release, which will probably also bump to rustls 0.22.

@Ralith We're planning to do a rustls-platform-verifier 0.1.1 release w/ FreeBSD support and Rustls 0.21 separate from the 0.2.0 release with Rustls 0.22. There's an important consumer of the library that will be on Rustls 0.21 for a while so the 0.1.x release tree will be maintained for a decent amount of time.

complexspaces:
@Ralith 0.1.1 is now out and [released](https://github.com/rustls/rustls-platform-verifier/releases/tag/v%2F0.1.1) 👍

PR #1736, opened 2024-01-08
udp: use io::Result<> where possible

stormshield-damiend:
Use `io::Result<>` in `quinn-udp` module. This was found as non blocking in #1701

### DIFF
--- a/quinn-udp/src/fallback.rs
+++ b/quinn-udp/src/fallback.rs
@@ -31,7 +31,7 @@ impl UdpSocketState {
         socket: UdpSockRef<'_>,
         _state: &UdpState,
         transmits: &[Transmit],
-    ) -> Result<usize, io::Error> {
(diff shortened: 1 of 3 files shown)


djc (approved):
Stylistically, I would prefer doing it the other way around (`Result<_, io::Error>`) because I find that more obvious at a pretty limited verboseness cost. In general, I've ended up not being a big fan of type aliases in almost all uses.

But, it seems @Ralith feels differently, and either way being consistent is probably better.

Ralith:
Thanks!

> I would prefer doing it the other way around

Oh, is that how we ended up with these? I feel like `Result` aliases are at least a well established convention, and in particular `io::Result` is very widely used.

djc:
I'm definitely not a fan of unprefixed `Result` aliases, though I agree the qualified `io::Result` is common enough that I don't mind it as much.

PR #1739, opened 2024-01-15
Add mutual_cluster example

alexchamberlain:
Add a more detailed example of running multiple processes that mutually authenticate with one another. This demonstrates:
- How to use `rcgen` and associated creates to generate a toy CA
- How to load certificates and configure connections for mutual authentication
- How to keep track of connections and reestablish them if they go down

There might be a couple of points of clean up on this one, but I think it's good enough to share. I understand this is a slighly bigger example than some of the others, but it took me a while to figure out how to get the crates to play together and I figured others might find it useful; is this of interest?

### DIFF
--- a/quinn/Cargo.toml
+++ b/quinn/Cargo.toml
@@ -53,13 +53,14 @@ crc = "3"
 bencher = "0.1.5"
 directories-next = "2"
 rand = "0.8"
-rcgen = "0.12.0"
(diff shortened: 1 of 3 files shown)


djc:
This seems pretty specific for the Quinn repo, and it spends quite a bit of code on things that would be better documented over in the rcgen repo instead of this repo.

Issue #1742, opened 2024-01-16
Odd Behaviour with QUINN and Apple Network Framework


paxsonsa:
Thanks for the great library, docs, and for all the work you do. Thanks in advance for having a peek. Trying to track down an interesting problem; I have a small example here of a Quinn server running on macOS (13.6.3) with the latest version of Quinn and rust.  

The core issue I am running into is when the server is the one accepting the bidirectional stream after the tunnel is established, no data is received, however if I change the connection from accept_bi and open_bi the stream works perfectly as expected.

Am I missing something obvious here? Happy to provide the simplified swift server, figured most people might not want that.

Example:
```rust
/*
rustc 1.75.0 (82e1608df 2023-12-21)
Cargo.toml
quinn = "*"
bytes = "1.5"
rustls = { version = "0.21", features = ["quic"] }
rcgen = "0.12.0"
tokio = { version = "1.35.1", features = ["full"] }
thiserror = "1.0"
tracing = "0.1"
tracing-subscriber = "0.3"
*/
use bytes::BytesMut;
use quinn::{Endpoint, ServerConfig};
use std::str;
use std::{net::SocketAddr, sync::Arc};

pub const ALPN_QUIC_HTTP: &[&[u8]] = &[b"demo"];

#[tokio::main]
async fn main() {
    // Set up the tracing subscriber
    tracing::subscriber::set_global_default(
        tracing_subscriber::FmtSubscriber::builder()
            .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
            .finish(),
    )
    .unwrap();

    // TODO: Create Simple Server to Echo Messages from Client
    // - It would appear the Swift QUIC in Networking is not capable of customized CA
    // this is an issue. That might mean we need to use Rust Binding with Swift?
    // Another option is to message the Apple Help.
    let addr = "127.0.0.1:4567".parse().unwrap();
    run_server(addr).await;
    // tokio::spawn(run_server(addr));
    // run_client(addr).await.unwrap();
}

async fn run_server(addr: SocketAddr) {
    let cert = rcgen::generate_simple_self_signed(vec![
        "127.0.0.1".into(),
        "localhost".into(),
        "0.0.0.0".into(),
    ])
    .unwrap();
    let cert_der = cert.serialize_der().unwrap();
    let private_key = cert.serialize_private_key_der();
    let private_key = rustls::PrivateKey(private_key);
    let cert_chain = vec![rustls::Certificate(cert_der.clone())];

[...]

PR #1743, opened 2024-01-19
Document two more commonly-observed mistakes in `RecvStream` use

Ralith:
Prompted by #1742 and assorted prior support requests. Should serve as a good reminder for us as well.


djc (approved):

djc:
Nice!

paxsonsa:
Thanks, that is helpful!

PR #1745, opened 2024-01-20
Fix parsing of IPV6 URLs in client example

gretchenfrage:
Hi all! My name's Phoenix. I've been interesting in contributing to this project for a while and thought I'd start with trying to get the examples working on my machine. Let me know if you have any feedback :)

I'm running Ubuntu 22.04.3, and even after enabling IPV6, when running the examples in the way suggested in the README:

```sh
$ cargo run --example server ./
$ cargo run --example client https://localhost:4433/Cargo.toml
```

The client timed out. If I passed `--listen 127.0.0.1:4433` or `--listen 0.0.0.0:4433` when running the server, it worked fine. But weirder yet, if I passed `--listen [::]:4433` on the server, it also works fine. So it seems like my system allows `127.0.0.1` to connect to `[::]` but not to `[::1]`.

In any case, I figured that I should be able to get the client to work without changing the server command by running:

```sh
cargo run --example client https://[::1]:4433/Cargo.toml --host localhost
```

But it failed to convert the URL to a socket address. So I fixed that, which is what this PR is.

Commit message
---

Currently, when trying to run the client example with an IPV6 address URL, such as by running:

    cargo run --example client https://[::1]:4433/Cargo.toml --host localhost

A "failed to lookup address information: Name or service not known" error is raised. This is because `url.host_str()` is `"[::1]"`, which is wrapped in brackets. These brackets, specified by by RFC 2732, are part of the URL syntax, not the IP address syntax.

Although this code succeeds, because the standard library treats this like a URL:

    use std::net::ToSocketAddrs;
    "[::1]:4433".to_socket_addrs()

This code does not:

    use std::net::Ipv6Addr;
    "[::1]".parse::<Ipv6Addr>()

As the stdlib expects to just receive "::1". Consequentially, this does not succeed, counterintuitively:

    use std::net::ToSocketAddrs;
    ("[::1]", 4433).to_socket_addrs()

This code fixes the client example's URL parsing behavior by stripping out such brackets in the same way as [is done in tokio-tungstenite][1].

[1]: https://github.com/snapview/tokio-tungstenite/blob/052d085aff4708b924b92becaa83923b15045841/src/lib.rs#L404

### DIFF
--- a/quinn/examples/client.rs
+++ b/quinn/examples/client.rs
@@ -63,7 +63,8 @@ fn main() {
 #[tokio::main]
 async fn run(options: Opt) -> Result<()> {
     let url = options.url;
-    let remote = (url.host_str().unwrap(), url.port().unwrap_or(4433))
+    let url_host = strip_ipv6_brackets(url.host_str().unwrap());
+    let remote = (url_host, url.port().unwrap_or(4433))
         .to_socket_addrs()?
         .next()
         .ok_or_else(|| anyhow!("couldn't resolve to an address"))?;
@@ -102,11 +103,7 @@ async fn run(options: Opt) -> Result<()> {
     let request = format!("GET {}\r\n", url.path());
     let start = Instant::now();
     let rebind = options.rebind;
-    let host = options
-        .host
-        .as_ref()
-        .map_or_else(|| url.host_str(), |x| Some(x))
-        .ok_or_else(|| anyhow!("no hostname specified"))?;
+    let host = options.host.as_deref().unwrap_or(url_host);
 
     eprintln!("connecting to {host} at {remote}");
     let conn = endpoint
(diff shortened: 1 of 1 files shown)


quinn/examples/client.rs
```diff
@@ -105,7 +108,7 @@ async fn run(options: Opt) -> Result<()> {
     let host = options
         .host
         .as_ref()
-        .map_or_else(|| url.host_str(), |x| Some(x))
+        .map_or_else(|| url.host_str().map(strip_ipv6_brackets), |x| Some(x))
```
Ralith, line 111: nit: maybe reuse the result from doing this above?

Ralith (approved):
Thanks!

> it seems like my system allows 127.0.0.1 to connect to [::] but not to [::1].

That's expected. `[::]` is the IPv6 wildcard address, so it listens on every local IP, and on a dual-stack socket that includes IPv4. `[::1]` is a specific IPv6 address, so a socket bound to it only accepts packets addressed explicitly to that address, which necessarily must be IPv6. We don't use the IPv6 wildcard by default in the examples because opening ports accessible from other hosts in demo code seems like poor form.

Note that whether sockets are dual-stack by default varies with platform and system configuration.

quinn/examples/client.rs
gretchenfrage: Sounds good. Incidentally, it looks like that error handling case was superfluous, because the function the function already began by unwrapping `url.host_str()`. So it's a bit simpler now overall.

PR #1746, opened 2024-01-21
Send path validation responses to the correct remote

Ralith:
We previously always sent these on the active path, which undermines a peer's validation of the return direction of a candidate new path.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -617,9 +617,10 @@ impl Connection {
                 // for starting another datagram. If there is any anti-amplification
                 // budget left, we always allow a full MTU to be sent
                 // (see https://github.com/quinn-rs/quinn/issues/1082)
-                if self.path.anti_amplification_blocked(
-                    self.path.current_mtu() as u64 * num_datagrams as u64 + 1,
-                ) {
+                if self
+                    .path
+                    .anti_amplification_blocked(self.path.current_mtu() as u64 * num_datagrams + 1)
+                {
                     trace!("blocked by anti-amplification");
                     break;
                 }
@@ -728,7 +729,7 @@ impl Connection {
                 space_id,
                 buf,
                 buf_capacity,
-                (num_datagrams - 1) * (self.path.current_mtu() as usize),
+                (num_datagrams as usize - 1) * (self.path.current_mtu() as usize),
                 ack_eliciting,
                 self,
                 self.version,
@@ -898,7 +899,7 @@ impl Connection {
         trace!("sending {} bytes in {} datagrams", buf.len(), num_datagrams);
         self.path.total_sent = self.path.total_sent.saturating_add(buf.len() as u64);
 
-        self.stats.udp_tx.on_sent(num_datagrams as u64, buf.len());
+        self.stats.udp_tx.on_sent(num_datagrams, buf.len());
 
         Some(Transmit {
             destination: self.path.remote,
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -796,6 +796,38 @@ impl Connection {
                 break;
             }
 
+            // Send an off-path PATH_RESPONSE. Prioritized over on-path data to ensure that path
+            // validation can occur while the link is saturated.
+            if space_id == SpaceId::Data && num_datagrams == 1 {
+                if let Some((token, remote)) = self.path_responses.pop_off_path(&self.path.remote) {
+                    // `unwrap` guaranteed to succeed because `builder_storage` was populated just
+                    // above.
+                    let mut builder = builder_storage.take().unwrap();
+                    trace!("PATH_RESPONSE {:08x} (off-path)", token);
+                    buf.write(frame::Type::PATH_RESPONSE);
+                    buf.write(token);
+                    self.stats.frame_tx.path_response += 1;
+                    builder.pad_to(MIN_INITIAL_SIZE);
+                    builder.finish_and_track(
+                        now,
+                        self,
+                        Some(SentFrames {
+                            non_retransmits: true,
+                            ..SentFrames::default()
+                        }),
+                        buf,
+                    );
+                    self.stats.udp_tx.on_sent(1, buf.len());
+                    return Some(Transmit {
+                        destination: remote,
+                        size: buf.len(),
+                        ecn: None,
+                        segment_size: None,
+                        src_ip: self.local_ip,
+                    });
+                }
+            }
+
             let sent = self.populate_packet(
                 now,
                 space_id,
@@ -2583,7 +2615,7 @@ impl Connection {
                     close = Some(reason);
                 }
                 Frame::PathChallenge(token) => {
-                    self.path_responses.push(number, token);
+                    self.path_responses.push(number, token, remote);
                     if remote == self.path.remote {
                         // PATH_CHALLENGE on active path, possible off-path packet forwarding
                         // attack. Send a non-probing packet to recover the active path.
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -796,6 +797,44 @@ impl Connection {
...
+                self.stats.udp_tx.ios += 1;
+                self.stats.udp_tx.bytes += buf.len() as u64;
+                self.path.congestion.on_sent(now, buf.len() as u64, pn);
+                return Some(Transmit {
+                    destination: remote,
+                    size: buf.len(),
+                    ecn: None,
+                    segment_size: None,
+                    src_ip: self.local_ip,
+                });
```
djc, line 835: This looks like it is duplicating a decent amount of stuff. Do you think it's worth deduplicating this a bit more?

quinn-proto/src/connection/mod.rs
```diff
@@ -796,6 +797,44 @@ impl Connection {
...
+            // Send an off-path PATH_RESPONSE. LIFO since old challenges might have been forgotten by the sender.
+            if space_id == SpaceId::Data
+                && num_datagrams == 1
+                && self
+                    .path_responses
+                    .last()
+                    .map_or(false, |(_, remote)| *remote != self.path.remote)
+            {
+                let (response, remote) = self.path_responses.pop().unwrap();
+                let mut builder = builder_storage.take().unwrap();
```
djc, line 809: How do we know these `unwrap`s to be safe? Should be use a nested `if let Some(..) = self.path_responses.last() { if space_id == SpaceId::Data && .. {` to avoid the first one?

quinn-proto/src/connection/mod.rs
```diff
@@ -2585,15 +2624,25 @@ impl Connection {
...
+                        .find(|(_, addr)| *addr == remote);
+                    if let Some((existing, _)) = existing {
+                        // Update an queued response
+                        if existing.packet <= number {
+                            *existing = response;
+                        }
+                    } else if self.path_responses.len() < MAX_PATH_RESPONSES {
+                        self.path_responses.push((response, remote));
+                    } else {
+                        trace!("ignoring excessive PATH_CHALLENGE");
```
djc, line 2645: How do you feel about wrapping this up in a `PathResponses` type? It feels like there is substantial logic on both the push and the pop case which it might be nice to outline. (If we store this in `PathData`, maybe it could be methods there instead of an additional type.)

djc (commented):
Thanks for picking this up!

quinn-proto/src/connection/mod.rs
```diff
@@ -197,7 +197,8 @@ pub struct Connection {
     //
     // Queued non-retransmittable 1-RTT data
     //
-    path_response: Option<PathResponse>,
+    /// Responses to PATH_CHALLENGE frames, and the addresses to send them to
+    path_responses: Vec<(PathResponse, SocketAddr)>,
```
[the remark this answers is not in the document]
Ralith: I think folding this into `PathData` makes sense as part of multi-path support. In the absence of multi-path, I don't think we can easily make use of the additional state that carries, so it's clearer to store only what we'll use. I'd prefer to isolate this fix from beginning to build multi-path out.

quinn-proto/src/connection/mod.rs
Ralith: Factored out the stats stuff and did some tweaking. I feel like what's left is pretty distinctive to this case, though I'm open to ideas if you still think it could be more concise.

quinn-proto/src/connection/mod.rs
Ralith: Removed one by refactoring, and documented the other.

quinn-proto/src/connection/mod.rs
Ralith: Seems reasonable. Done.

quinn-proto/src/connection/mod.rs
```diff
@@ -197,7 +197,8 @@ pub struct Connection {
     //
     // Queued non-retransmittable 1-RTT data
     //
-    path_response: Option<PathResponse>,
+    /// Responses to PATH_CHALLENGE frames, and the addresses to send them to
+    path_responses: PathResponses,
```
djc, line 201: It would be nice to extract the abstraction in one commit, then update it to do the new stuff in another.

quinn-proto/src/connection/mod.rs
```diff
@@ -3612,10 +3635,63 @@ pub enum Event {
...
+#[derive(Copy, Clone)]
 struct PathResponse {
     /// The packet number the corresponding PATH_CHALLENGE was received in
     packet: u64,
     token: u64,
+    /// The address the corresponding PATH_CHALLENGE was received from
+    remote: SocketAddr,
+}
+
+#[derive(Default)]
```
djc, line 3647: This should probably be above `PathResponse`?

Maybe move this and `PathResponse` into `paths` to avoid ballooning the size of `connection/mod.rs` even more?

quinn-proto/src/connection/mod.rs
djc: This is probably okay for now -- maybe it some point it makes sense to wrap a builder around `Transmit`...

quinn-proto/src/connection/mod.rs
Ralith: Done.

quinn-proto/src/connection/mod.rs
Ralith: Done.

quinn-proto/src/connection/paths.rs
```diff
@@ -167,27 +169,63 @@ impl RttEstimator {
...
+        /// Arbitrary permissive limit to prevent abuse
+        const MAX_PATH_RESPONSES: usize = 16;
+        let response = PathResponse {
+            packet,
+            token,
+            remote,
+        };
+        let existing = self.pending.iter_mut().find(|x| x.remote == remote);
+        if let Some(existing) = existing {
+            // Update an queued response
```
djc, line 186: "an" -> "a"?

quinn-proto/src/connection/paths.rs
```diff
@@ -167,27 +169,63 @@ impl RttEstimator {
...
+            // We don't expect to ever hit this with well-behaved peers, so we don't bother dropping
+            // older challenges.
+            trace!("ignoring excessive PATH_CHALLENGE");
         }
     }
 
-    pub(crate) fn pop(&mut self) -> Option<u64> {
-        Some(self.pending.take()?.token)
+    pub(crate) fn pop_off_path(&mut self, remote: &SocketAddr) -> Option<(u64, SocketAddr)> {
+        let response = *self.pending.last()?;
```
djc, line 202: For the `pop()` methods, shouldn't we be searching the list for challenges from the given `remote`, rather than only checking the last one? If not, why not?

quinn-proto/src/connection/mod.rs
```diff
@@ -796,6 +796,37 @@ impl Connection {
                 break;
             }
 
+            // Send an off-path PATH_RESPONSE
+            if space_id == SpaceId::Data && num_datagrams == 1 {
```
djc, line 800: So AIUI this means we're always yielding off-path `PATH_RESPONSE` frames before yielding any on-path packets. Is that optimal? I would (maybe naively) expect that we might want to prioritize on-path packets and only then yield off-path challenge responses in order to minimize latency for the former?

quinn-proto/src/connection/paths.rs
Ralith: Each call to `poll_transmit` on an otherwise healthy connection will call both `pop` variants, so forwards progress is guaranteed, and that seems like enough. Added a brief comment explaining.

quinn-proto/src/connection/mod.rs
Ralith: This prioritization is intended. Added an explanatory comment.

djc (approved):

Issue #1678, opened 2023-10-09
Update package ring to 0.17 for Windows arm64 support


lhyzf:
see https://github.com/briansmith/ring/issues/1614

djc:
We expose both *ring* and rustls as part of the public API, so in order to update *ring* to 0.17 we'll need to make a semver-incompatible bump to the API. However, I expect we'll also get a semver-incompatible rustls release in a few weeks from now so I think we'll want to wait until we can do both.

dignifiedquire:
rustls is already using `ring@0.17` on the latest `rustls@0.21.8`, which is a bit unfortunate, as this means currently `quinn` pulls in two versions of ring

djc:
Yup -- I don't think there's much that can be done, hopefully rustls 0.22 will be published soon.

PikuZheng:
any progress?

Ralith:
You can track the current progress at https://github.com/quinn-rs/quinn/pull/1715, where you previously commented.

Ralith:
Quinn 0.11 has been released, using ring 0.17.

PR #1751, opened 2024-01-28
Upgrade ring to 0.17.7

hack3ric:
ring 0.17.7 brings new platform support such as RISC-V and LoongArch, and is in line with rustls' ring version, thus avoiding multiple ring packages of different versions to be built.

### DIFF
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -29,7 +29,7 @@ arbitrary = { version = "1.0.1", features = ["derive"], optional = true }
 bytes = "1"
 rustc-hash = "1.1"
 rand = "0.8"
-ring = { version = "0.16.7", optional = true }
+ring = { version = "0.17.7", optional = true }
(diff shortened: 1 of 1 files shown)


djc:
See #1715.

hack3ric:
> See #1715.

Sorry about not digging deeper. I didn't find introducing solely ring upgrade causing any incompatibilities while working on building on RISC-V. I guess it could be better to merge the ring upgrade part and release the non-breaking changes first?

Ralith (approved):
Since rustls 0.21.10 switched to ring 0.17, this does seem like an easy win. Am I forgetting something?

zh-jq-b:
> Since rustls 0.21.10 switched to ring 0.17, this does seem like an easy win. Am I forgetting something?

https://github.com/quinn-rs/quinn/issues/1678

Ralith:
Ah, right, it's a breaking change, so we can't release it alone without excessive churn, and it's a trivial enough change that there's not really any benefit from merging it early otherwise. Still, thanks for bringing it up!

PR #1752, opened 2024-01-30
Allow app to accept/reject/retry before handshake begins

### LINKED ISSUE #1661: Automatic Retry heuristics
[QUIC's Retry mechanism](https://www.rfc-editor.org/rfc/rfc9000.html#name-address-validation-using-re) is (primarily) a defense against resource exhaustion attacks by forcing peers to prove that they can receive packets sent to their address before allocating connection state. It is not the default mode because it costs an additional round-trip of latency. Quinn currently requires Retry only at the application's explicit request, but applications are unlikely to make effective use of this in practice. To provide more robustness by default, we should automatically enable Retry whenever the number of incoming connections whose address has not yet been validated is above some (proportionate?) threshold.

gretchenfrage:
Closes #1661 


Ralith:
Thank you for working on this! I've wanted this feature for a long time, and I'm excited about the amount of discretion it gives applications for practically no increase in intrinsic API complexity.

It may be a few days before I have time to dig into reviewing this properly, but I can see it's turned out to be a fairly large change. Any chance you could break it up into a series of smaller incremental, self-contained commits?

quinn-proto/src/endpoint.rs
```diff
@@ -857,12 +804,218 @@ impl IndexMut<ConnectionHandle> for Slab<ConnectionMeta> {
...
+    pub fn may_retry(&self) -> bool {
+        // Currently, we do not produce retry tokens in a way that allows us to tell the difference
+        // between ones minted for a retry packet versus for a NEW_TOKEN frame, so to be safe we
+        // must simply never respond with a retry packet to an initial packet with a token in it.
+        // However, the QUIC protocol would allow for such a distinction to be implemented.
+        !self.is_validated()
+    }
+
+    /// Attempt to accept this incoming connection (an error may still occur).
+    pub fn accept(
```
djc, line 862: I think it would make more sense to do this the other way around, keeping an `accept()` method on `Endpoint` that takes ownership of the `IncomingConnection`.

quinn-proto/src/endpoint.rs
```diff
@@ -857,12 +804,218 @@ impl IndexMut<ConnectionHandle> for Slab<ConnectionMeta> {
...
+                        buf,
+                    ))),
+                    _ => Err(None),
+                }
+            }
+        }
+    }
+
+    /// Reject this incoming connection attempt.
+    pub fn reject(self, endpoint: &mut Endpoint, buf: &mut BytesMut) -> Transmit {
```
djc, line 945: Same for this.

quinn-proto/src/endpoint.rs
```diff
@@ -857,12 +804,218 @@ impl IndexMut<ConnectionHandle> for Slab<ConnectionMeta> {
...
+            &self.src_cid,
+            TransportError::CONNECTION_REFUSED(""),
+            buf,
+        )
+    }
+
+    /// Respond with a retry packet, requiring the client to retry with address validation.
+    ///
+    /// Panics if `may_retry` is false.
+    pub fn retry(self, endpoint: &mut Endpoint, buf: &mut BytesMut) -> Transmit {
```
djc, line 960: And this.

quinn/examples/client.rs
```diff
@@ -39,6 +39,10 @@ struct Opt {
     /// Simulate NAT rebinding after connecting
     #[clap(long = "rebind")]
     rebind: bool,
+
+    /// Address to bind on
+    #[clap(long = "bind", default_value = "[::]:0")]
+    bind: SocketAddr,
```
djc, line 45: This could be a separate commit.

quinn/src/connection.rs
```diff
@@ -27,44 +28,90 @@ use crate::{
...
+/// having begun. However, on incoming connections, this is given to the application before the
+/// server has begun performing its half of the handshake. This allows the application to check the
+/// remote address and reject the connection or have the client retry it with address validation
+/// before ever allocating state for it. Once this future is polled or other methods which take
+/// `&mut self` are called the server begins its half of the handshake and those methods can no
+/// longer be called.
 #[derive(Debug)]
 #[must_use = "futures/streams/sinks do nothing unless you `.await` or poll them"]
-pub struct Connecting {
+pub struct Connecting(Option<ConnectingState>);
```
djc, line 41: Why not expose `ConnectingState` directly?

quinn/src/connection.rs
```diff
@@ -27,44 +28,90 @@ use crate::{
...
+        Self(Some(ConnectingState::Handshaking(
+            ConnectingHandshaking::new(handle, conn, endpoint_events, conn_events, socket, runtime),
+        )))
+    }
 
-        Self {
-            conn: Some(conn),
-            connected: on_connected_recv,
-            handshake_data_ready: Some(on_handshake_data_recv),
+    fn accept(&mut self) -> Result<&mut ConnectingHandshaking, ConnectionError> {
```
djc, line 90: ... because this gets pretty messy.

quinn/src/connection.rs
```diff
@@ -92,20 +139,22 @@ impl Connecting {
...
-        if is_ok {
-            let conn = self.conn.take().unwrap();
-            Ok((Connection(conn), ZeroRttAccepted(self.connected)))
-        } else {
-            Err(self)
+    /// For incoming connections, a 0.5-RTT connection will always be successfully constructed,
+    /// unless an error occurs attempting to accept the connection which had not already been
+    /// caught but which would have happened upon attempting to accept this connection attempt
+    /// regardless of calling `into_0rtt`.
+    pub fn into_0rtt(mut self) -> Result<(Connection, ZeroRttAccepted), Into0RttError> {
```
djc, line 146: And this, too.

quinn/src/endpoint.rs
```diff
@@ -383,10 +383,34 @@ pub(crate) struct State {
     recv_buf: Box<[u8]>,
     send_limiter: WorkLimiter,
     runtime: Arc<dyn Runtime>,
+}
+
+#[derive(Debug)]
+struct TransmitState {
```
djc, line 389: This could also be a separate commit.

djc (commented):
I skimmed your changes to try and give some high-level feedback.

In particular, the changes to the -proto `Endpoint` are just too hard to review as one commit. Suggest you refactor into an API that better fits your proposal before making the final change to split off the API that allows the caller to make a decision on handling incoming connections.

gretchenfrage:
Thanks for the review! Splitting this into more incremental commits and making the other changes you mentioned makes sense to me. Just so you know, though, it might be a little bit until I can dedicate more time to this--but I'll get back to you when it's ready to be looked at again.

quinn-proto/src/endpoint.rs
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
gretchenfrage: Done.

quinn/examples/client.rs
gretchenfrage: Done.

quinn/src/endpoint.rs
gretchenfrage: Done.

gretchenfrage:
Should be ready for re-review now. I split this up into various commits which should make this a lot easier to see what I'm doing if you go through them one at a time, and made some of the other changes you requested too.

> Why not expose ConnectingState directly?

I don't really see how this would work--at least, I'm not sure how to reconcile that idea with Ralith's request that this functionality be placed into the existing `Connecting` API. Which, although that wasn't my original proposal, I do like the idea of. If you have something in mind though I'd be happy to talk about it.

> ... because this gets pretty messy.

Yeah that's fair--I've refactored that area in general, it should be a lot cleaner now.

quinn/src/endpoint.rs
gretchenfrage: Responded below in thread.

quinn/src/connection.rs
gretchenfrage: Responded below in thread.

quinn/src/connection.rs
gretchenfrage: Responded below in thread.

quinn/src/connection.rs
gretchenfrage: Responded below in thread.

quinn/src/connection.rs
```diff
@@ -122,20 +158,145 @@ impl Connecting {
...
+    ///
+    /// Will panic if called when the handshake has already begun or if `may_retry` is false.
+    pub fn may_retry(&self) -> bool {
+        self.unwrap_incoming_ref("may_retry").inner.is_validated()
+    }
+
+    /// Reject this incoming connection attempt
+    ///
+    /// Will panic if called when the handshake has already begun.
+    pub fn reject(self) {
```
Ralith, line 207: Let's also have a method to `ignore` a connection attempt.

quinn/src/tests.rs
```diff
@@ -159,17 +159,29 @@ fn export_keying_material() {
...
+                endpoint
+                    .accept()
+                    .await
+                    .expect("endpoint")
+                    .await
+                    .expect("connection")
+            }
+        });
+        let outgoing_conn = outgoing_conn_fut.await.unwrap();
+        let incoming_conn = incoming_conn_fut.await.unwrap();
```
Ralith, line 184: It'd be simpler to skip spawning and just `join` the two.

quinn/src/tests.rs
```diff
@@ -183,70 +195,88 @@ fn export_keying_material() {
...
-    let mut endpoint = Endpoint::server(
-        server_config,
-        SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 0),
-    )
-    .unwrap();
-    let mut client_config = ClientConfig::with_root_certificates(roots);
-    client_config.transport_config(transport_config);
-    endpoint.set_default_client_config(client_config);
+/// Constructs endpoints suitable for connecting to themselves and each other
+struct EndpointFactory(rcgen::Certificate);
```
Ralith, line 247: What's the purpose of this refactoring? It's not obviously required by any other changes in this commit.

Ralith (changes requested):

quinn-proto/src/endpoint.rs
```diff
@@ -542,44 +529,124 @@ impl Endpoint {
...
+            &incoming.src_cid,
+            TransportError::CONNECTION_REFUSED(""),
+            buf,
+        )
+    }
+
+    /// Respond with a retry packet, requiring the client to retry with address validation.
+    ///
+    /// Panics if `may_retry` is false.
+    pub fn retry(&mut self, incoming: IncomingConnection, buf: &mut BytesMut) -> Transmit {
```
Ralith, line 601: Rather than requiring the user to call `may_retry` and panicing if they get it wrong, could we make `may_retry` private and return an error here if it's `false`? That way it's impossible for users to forget to check.

quinn-proto/src/endpoint.rs
```diff
@@ -626,6 +705,31 @@ impl Endpoint {
...
     }
 
+    fn connection_refuse_if_connection_limit(
+        &mut self,
+        version: u32,
+        addresses: FourTuple,
+        crypto: &Keys,
+        src_cid: &ConnectionId,
+        buf: &mut BytesMut,
+    ) -> Option<Transmit> {
```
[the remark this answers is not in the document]
gretchenfrage: Made it return `Result<(), Transmit>` (when you write `Err<Transmit>` I don't know what you mean, did you mean something by that?)

quinn-proto/src/endpoint.rs
```diff
@@ -626,6 +705,31 @@ impl Endpoint {
         conn
     }
 
+    fn connection_refuse_if_connection_limit(
```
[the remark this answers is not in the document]
gretchenfrage: Renamed to `remote_address_validated` (it's goes nicely with `remote_address`)

quinn-proto/src/endpoint.rs
```diff
@@ -863,6 +967,70 @@ pub enum DatagramEvent {
...
+    }
+
+    /// Whether it is legal to require the client to retry.
+    ///
+    /// If `is_validated` is false, `may_retry` is necessarily true.
+    fn may_retry(&self) -> bool {
+        // Currently, we do not produce retry tokens in a way that allows us to tell the difference
+        // between ones minted for a retry packet versus for a NEW_TOKEN frame, so to be safe we
+        // must simply never respond with a retry packet to an initial packet with a token in it.
+        // However, the QUIC protocol would allow for such a distinction to be implemented.
```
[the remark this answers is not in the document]
gretchenfrage: Method removed.

quinn-proto/src/endpoint.rs
```diff
@@ -542,44 +538,125 @@ impl Endpoint {
...
             loc_cid,
-            src_cid,
-            addresses,
+            incoming.src_cid,
+            incoming.addresses,
             now,
             tls,
             Some(server_config),
             transport_config,
+            incoming.retry_src_cid.is_some(),
```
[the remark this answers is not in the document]
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
```diff
@@ -863,6 +967,70 @@ pub enum DatagramEvent {
...
+    fn may_retry(&self) -> bool {
+        // Currently, we do not produce retry tokens in a way that allows us to tell the difference
+        // between ones minted for a retry packet versus for a NEW_TOKEN frame, so to be safe we
+        // must simply never respond with a retry packet to an initial packet with a token in it.
+        // However, the QUIC protocol would allow for such a distinction to be implemented.
+        !self.is_validated()
+    }
+}
+
+impl fmt::Debug for IncomingConnection {
```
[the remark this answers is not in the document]
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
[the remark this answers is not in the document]
Ralith: Er, right.

quinn/src/connection.rs
```diff
@@ -29,43 +29,23 @@ use proto::congestion::Controller;
...
+pub struct Connecting(Option<ConnectingState>);
+
+#[derive(Debug)]
+pub(crate) enum ConnectingState {
+    Handshaking(ConnectingHandshaking),
+}
+
+#[derive(Debug)]
+pub(crate) struct ConnectingHandshaking {
     conn: Option<ConnectionRef>,
```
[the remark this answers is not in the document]
gretchenfrage: Obviated by larger refactor as requested below in thread. PR now doesn't actually change this file at all.

quinn-proto/src/endpoint.rs
```diff
@@ -987,29 +977,27 @@ impl IncomingConnection {
     ///
     /// This means that the sender of the initial packet has proved that they can receive traffic
     /// sent to `self.remote_address()`.
-    fn is_validated(&self) -> bool {
+    pub fn is_validated(&self) -> bool {
```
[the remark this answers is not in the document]
gretchenfrage: Resolved as mentioned above in thread.

quinn-proto/src/tests/mod.rs
```diff
@@ -2760,3 +2750,19 @@ fn reject_new_connections() {
     pair.server.assert_no_accept();
     assert!(pair.client.connections.get(&client_ch).unwrap().is_closed());
 }
+
+#[test]
+fn reject_remote_address() {
+    let _guard = subscribe();
+    let mut pair = Pair::default();
+    pair.server.retry_policy = RetryPolicy(Box::new(|incoming| {
+        incoming.remote_address();
```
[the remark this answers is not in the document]
gretchenfrage: Mostly was making sure it doesn't panic. But overall this test is weird. I removed the test outright because the new test added to `quinn` tests this behavior and is in a better position to do so what with it actually having real IP addresses in a normal way.

quinn-proto/src/endpoint.rs
```diff
@@ -521,17 +476,48 @@ impl Endpoint {
...
+        buf: &mut BytesMut,
+    ) -> Result<(ConnectionHandle, Connection), Option<Transmit>> {
+        self.check_connection_limit(
+            incoming.version,
+            incoming.addresses,
+            &incoming.crypto,
+            &incoming.src_cid,
+            buf,
+        )
+        .map_err(Some)?;
```
Ralith, line 517: This `map_err` is not required, because `Option<Transmit>` satisfies `From<Transmit>`.

quinn-proto/src/tests/util.rs
```diff
@@ -292,6 +292,34 @@ pub(super) struct TestEndpoint {
     conn_events: HashMap<ConnectionHandle, VecDeque<ConnectionEvent>>,
     pub(super) captured_packets: Vec<Vec<u8>>,
     pub(super) capture_inbound_packets: bool,
+    pub(super) retry_policy: RetryPolicy,
```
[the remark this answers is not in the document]
gretchenfrage: With the removal of the added test mentioned above in thread, I now simply changed this to a `use_retry` bool.

quinn/src/connection.rs
```diff
@@ -27,13 +28,30 @@ use crate::{
...
 
 /// In-progress connection attempt future
+///
+/// On outgoing connections, this is given to the application in the state of the handshake already
+/// having begun. However, on incoming connections, this is given to the application before the
+/// server has begun performing its half of the handshake. This allows the application to check the
+/// remote address and reject the connection or have the client retry it with address validation
+/// before ever allocating state for it. Once this future is polled or other methods which take
+/// `&mut self` are called the server begins its half of the handshake and those methods can no
+/// longer be called.
```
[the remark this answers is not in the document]
gretchenfrage: Obviated by larger refactor.

quinn/src/connection.rs
gretchenfrage: Done.

quinn/src/tests.rs
gretchenfrage: Done.

quinn/src/tests.rs
gretchenfrage: The new `ip_blocking` test creates 3 endpoints, specifically relying on the 2 that are designated client endpoint having different IP addresses, and tests the behavior of the server blocking one of their IP addresses and not the other. `EndpointFactory` is factored out so that their different endpoints can be constructed with compatible keys, which tests weren't previously doing. Creating multiple endpoints is necessary to give them different IP addresses, which other tests haven't been doing.

quinn-proto/src/endpoint.rs
gretchenfrage: Done (both in quinn-proto and in quinn).

quinn-proto/src/endpoint.rs
```diff
@@ -1052,6 +1050,23 @@ pub enum ConnectError {
     UnsupportedVersion,
 }
 
+/// Error for attempting to retry an [`IncomingConnection`] that can not be retried
+#[derive(Debug, Error)]
+pub struct RetryError(pub IncomingConnection);
```
Ralith, line 1055: Does `pub` here add much value, given that `into_incoming` exists? If not, let's keep the field private to reserve more space for non-breaking changes in the future.

quinn-proto/src/endpoint.rs
```diff
@@ -1052,6 +1050,23 @@ pub enum ConnectError {
...
+
+impl RetryError {
+    /// Get the [`IncomingConnection`]
+    pub fn into_incoming(self) -> IncomingConnection {
+        self.0
+    }
+}
+
+impl fmt::Display for RetryError {
```
Ralith, line 1064: Could this be achieved more concisely with an `#[error("...")]` attribute on the type? I think `thiserror` supports that.

quinn-proto/src/endpoint.rs
```diff
@@ -1052,6 +1050,23 @@ pub enum ConnectError {
     UnsupportedVersion,
 }
 
+/// Error for attempting to retry an [`IncomingConnection`] that can not be retried
```
Ralith, line 1053: Let's be as specific and actionable as possible: s/can not be retried/whose address was already validated by a previous retry/ or similar.

quinn-proto/src/tests/mod.rs
```diff
@@ -2750,19 +2750,3 @@ fn reject_new_connections() {
     pair.server.assert_no_accept();
     assert!(pair.client.connections.get(&client_ch).unwrap().is_closed());
 }
-
-#[test]
-fn reject_remote_address() {
```
Ralith: Can we easily test that explicitly rejecting a connection produces the expected error (i.e. not a timeout) on the client side? Coverage at the `quinn` layer is good, and adequate if a unit test is difficult for some reason, but unit tests have many fewer variables to worry about.

gretchenfrage:
> Should we introduce a new public future (perhaps Handshaking) returned by accept, and move this and into_0rtt there?
> 
> Pros:
> 
> - Reduced side effects here and in into_0rtt
> - No need for an AcceptError in into_0rtt
> - Less risk of panics in is_validated/may_retry/reject/retry
> 
> Cons:
> 
> - Duplication of local_ip/remote_address
> 
> Seems like a win to me. We could even get rid of the state enum entirely by implementing IntoFuture rather than Future to support direct await calls.

Alright, I've done a considerable refactor which I think achieves a good compromise between the various desires here, and results in a smaller PR previously. I'm no longer introducing different internal states to `Connecting`--in fact `quinn::Connecting` is completely untouched.

Awaiting `accept` on a `quinn::Endpoint` now returns a new `quinn::IncomingConnection` type. `quinn::IncomingConnection` has the following methods:

- All the methods `quinn_proto::IncomingConnection` has
- `fn accept(self) -> Result<Connecting, ConnectionError>`
- `fn reject(self)`
- `fn retry(self) -> Result<(), RetryError>`
  - (new `quinn::RetryError` is an error but also is just a wrapper around the `quinn::IncomingConnection`)
- `fn ignore(self)`

Furthermore, it implements `IntoFuture<Output=Result<Connection, ConnectionError>>`, the output type being the same as `Connecting`. This means that code which was just doing some basic `endpoint.accept().await.unwrap().await` (or similar) logic will still work without modification.

quinn/src/connection.rs
```diff
@@ -143,194 +152,25 @@ impl Connecting {
...
-            ConnectingState::AcceptError(_) => panic!("local_ip called after connecting errored"),
-        }
+        let conn = self.conn.as_ref().unwrap();
+        let inner = conn.state.lock("local_ip");
+
+        inner.inner.local_ip()
     }
 
-    /// The peer's UDP address
+    /// The peer's UDP address.
```
Ralith: nit: no trailing punctuation on the first line

quinn/src/connection.rs
```diff
@@ -143,194 +152,25 @@ impl Connecting {
     ///
     /// On all non-supported platforms the local IP address will not be available,
     /// and the method will return `None`.
-    ///
-    /// Will panic if called after `poll` has returned `Ready`.
```
Ralith: This is still true, isn't it?

quinn/src/incoming_connection.rs
```diff
@@ -0,0 +1,163 @@
...
+use proto::ConnectionError;
+use thiserror::Error;
+
+use crate::{
+    connection::{Connecting, Connection},
+    endpoint::EndpointRef,
+};
+
+/// An incoming connection for which the server has not yet begun its part of the handshake
+#[must_use = "futures/streams/sinks do nothing unless you `.await` or poll them"]
```
Ralith, line 19: `Future` has its own `must_use` annotation, so this `must_use` annotation isn't needed. The annotation on other quinn futures predates that Rust feature.

quinn/src/endpoint.rs
```diff
@@ -137,8 +137,9 @@ impl Endpoint {
 
     /// Get the next incoming connection attempt from a client
     ///
-    /// Yields [`Connecting`] futures that must be `await`ed to obtain the final `Connection`, or
-    /// `None` if the endpoint is [`close`](Self::close)d.
+    /// Yields [`IncomingConnection`]s, which can be `await`ed to obtain the final
+    /// [`Connection`](crate::Connection) or used in more complex ways such as to perform retries,
+    /// or `None` if the endpoint is [`close`](Self::close)d.
```
Ralith, line 142: nit: readability:
```suggestion
    /// Yields [`IncomingConnection`]s, or `None` if the endpoint is [`close`](Self::close)d. 
    /// [`IncomingConnection`]s can be `await`ed to obtain the final
    /// [`Connection`](crate::Connection) or used to e.g. filter connection attempts or force address validation.
```

quinn/src/incoming_connection.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub fn accept(mut self) -> Result<Connecting, ConnectionError> {
+        let state = self.0.take().unwrap();
+        state
+            .endpoint
+            .accept(state.inner, state.response_buffer)
+            .ok_or_else(|| {
+                ConnectionError::TransportError(proto::TransportError {
+                    code: proto::TransportErrorCode::PROTOCOL_VIOLATION,
+                    frame: None,
+                    reason: "Problem with initial packet".to_owned(),
```
Ralith, line 70: This isn't necessarily true. For example, `accept` can fail if the connection limit is exceeded, which isn't a transport error ("The peer violated the QUIC specification") at all. We should propagate an accurate error from the proto layer.

We could overload `LocallyClosed` for the limit-exceeded case, or add a new variant, or introduce an `AcceptError` that's either a `ConnectionError` or a limit-exceeded error. I lean towards a new variant as the clearest and most self-contained approach.

quinn/src/incoming_connection.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// Whether the socket address that is initiating this connection has been validated
+    ///
+    /// This means that the sender of the initial packet has proved that they can receive traffic
+    /// sent to `self.remote_address()`.
+    pub fn remote_address_validated(&self) -> bool {
+        self.0.as_ref().unwrap().inner.remote_address_validated()
+    }
+
+    /// Attempt to accept this incoming connection (an error may still occur)
+    pub fn accept(mut self) -> Result<Connecting, ConnectionError> {
```
Ralith, line 61: Could we simplify the API by returning `Connecting` unconditionally, but constructing it such that it immediately yields a `ConnectionError` when awaited? That seems like it'd reduce the number of distinct error paths in downstream code, and reduce churn. For example, the change to the accept loop in quinn/examples/server.rs, which introduces a regression where accept errors aren't logged, would be avoided.

quinn/src/tests.rs
```diff
@@ -183,70 +195,88 @@ fn export_keying_material() {
...
+            }
+        }
+    });
+    tokio::join!(
+        async move {
+            client_1
+                .connect(server_addr, "localhost")
+                .unwrap()
+                .await
+                .expect_err("server should have blocked this");
```
Ralith, line 224: We should assert that we get the correct error. For example, receiving a timeout here would be a bug.

Ralith (changes requested):
Thanks, this is looking really nice!

quinn/src/incoming_connection.rs
Ralith: On second thought, that's just `.await` again. No harm in being a little more fine-grained with the explicit API, I suppose!

Ralith:
I'm not sure how to feel about having both `Endpoint::accept` and `IncomingConnection::accept` which do rather different things. On the one hand, it's clear enough in context on each type, but in an abstract sense having to `accept` each connection *twice* is funny and might cause confusion in informal discussion. Would `IncomingConnection::connect` be a clearer name? But that wouldn't match the `proto` layer...

Ralith:
After sleeping on it I think the current naming is probably the sweet spot.

quinn/src/connection.rs
gretchenfrage: Presumably you were looking at solely the newer commits when reviewing--I reverted all changes to this file, and so the reason the period was re-added is because it is already like that on main. Regardless, I can drive-by fix it.

quinn/src/connection.rs
gretchenfrage: Similar to above comment--I reverted all changes to this file, and so if this comment is missing it's because it's missing on main.

gretchenfrage:
> After sleeping on it I think the current naming is probably the sweet spot.

Huh alright. I mean I'd be fine renaming it to `IncomingConnection::connect` on both the quinn and quinn-proto layer but yeah leaving as-is works.

gretchenfrage:
> Could we simplify the API by returning Connecting unconditionally, but constructing it such that it immediately yields a ConnectionError when awaited if the proto-layer accept failed? That seems like it'd reduce the number of distinct error paths in downstream code, and reduce churn. For example, the change to the accept loop in quinn/examples/server.rs, which introduces a regression where accept errors aren't logged, would be avoided.

> On second thought, that's just .await again. No harm in being a little more fine-grained with the explicit API, I suppose!

If it helps clarify the situation: when changing server applications, I suspect what might end up being done most naturally in most cases is just changing functions which accept a `Connecting` to instead accept an `IncomingConnection` but then awaiting it all the same. See for example the entire diff to `perf_server.rs`:

```diff
diff --git a/perf/src/bin/perf_server.rs b/perf/src/bin/perf_server.rs
index 169ca65e..4b710268 100644
--- a/perf/src/bin/perf_server.rs
+++ b/perf/src/bin/perf_server.rs
@@ -123,7 +123,7 @@ async fn run(opt: Opt) -> Result<()> {
     Ok(())
 }
 
-async fn handle(handshake: quinn::Connecting, opt: Arc<Opt>) -> Result<()> {
+async fn handle(handshake: quinn::IncomingConnection, opt: Arc<Opt>) -> Result<()> {
     let connection = handshake.await.context("handshake failed")?;
     debug!("{} connected", connection.remote_address());
     tokio::try_join!(
```

Thus it's the new `IncomingConnectionFuture` struct that implicitly does the work of acting as a little adapter which can be used as if it were "returning Connecting unconditionally, but constructing it such that it immediately yields a ConnectionError when awaited if the proto-layer accept failed"

quinn-proto/src/endpoint.rs
gretchenfrage: Obviated by changing this method's error type to `(ConnectionError, Option<Transmit>)` which means it does require a non-trivial `map_err`.

quinn-proto/src/endpoint.rs
gretchenfrage: Made private. I sometimes like pattern-destructuring things like this, but it's fine.

quinn-proto/src/endpoint.rs
gretchenfrage: Done. To both.

quinn-proto/src/endpoint.rs
gretchenfrage: Both rephrased to:

```
/// Error for attempting to retry an [`IncomingConnection`] which already bears an address
/// validation token from a previous retry
```

quinn-proto/src/tests/mod.rs
gretchenfrage: Test re-added, renamed to `reject_manually`, and now asserts that the correct error occurs.

quinn/src/incoming_connection.rs
gretchenfrage: Removed.

quinn/src/endpoint.rs
gretchenfrage: Changed to this:

```
/// Yields [`IncomingConnection`]s, or `None` if the endpoint is [`close`](Self::close)d.
/// [`IncomingConnection`] can be `await`ed to obtain the final
/// [`Connection`](crate::Connection), or used to eg. filter connection attempts or force
/// address validation, or converted into an intermediate `Connecting` future which can be
/// used to eg. send 0.5-RTT data.
```

quinn/src/incoming_connection.rs
gretchenfrage: New `ConnectionError::ConnectionLimitExceeded` variant added. Error when accepting `IncomingConnection` now originates in quinn-proto, rather than being synthesized in the quinn wrapper.

Return type of `quinn_proto::Endpoint::accept` changed from:

```rs
Result<(ConnectionHandle, Connection), Option<Transmit>>
```

To:

```rs
Result<(ConnectionHandle, Connection), (ConnectionError, Option<Transmit>)>
```

quinn/src/tests.rs
gretchenfrage: Done.

gretchenfrage:
I re-added the `RetryPolicy` callback to the quinn-proto testing utilities, due to me re-added the quinn-proto `reject_remote_address` test (now renamed to `reject_manually`). With regards to your previous comment on this topic:

> Could we reduce complexity by just taking an IncomingConnectionResponse directly here? No downstream code seems to need arbitrary logic.

It would not actually be that simple, because `RetryPolicy::yes()` is not as simple as unconditionally returning `IncomingConnectionResponse::Accept`--it only does so if the `IncomingConnection` is not validated.

quinn/src/connection.rs
Ralith: Oops, right.

quinn-proto/src/endpoint.rs
```diff
@@ -542,44 +538,125 @@ impl Endpoint {
...
+                        buf,
+                    ))),
+                    _ => Err(None),
                 }
             }
         }
     }
 
+    /// Reject this incoming connection attempt.
+    #[allow(dead_code)] // becomes used in subsequent commits in same PR
```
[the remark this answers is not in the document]
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
```diff
@@ -863,6 +967,70 @@ pub enum DatagramEvent {
...
+    fn local_ip(&self) -> Option<IpAddr> {
+        self.addresses.local_ip
+    }
+
+    /// The peer's UDP address.
+    #[allow(dead_code)] // becomes used in subsequent commits in same PR
+    fn remote_address(&self) -> SocketAddr {
+        self.addresses.remote
+    }
```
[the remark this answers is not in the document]
gretchenfrage: Done.

gretchenfrage:
All comments should be resolved, and I just refactored the commit history and force-pushed again so it should be more clean now.

quinn/src/endpoint.rs
```diff
@@ -449,14 +453,8 @@ impl State {
...
-                                Some(DatagramEvent::NewConnection(handle, conn)) => {
-                                    let conn = self.connections.insert(
-                                        handle,
-                                        conn,
-                                        self.socket.clone(),
-                                        self.runtime.clone(),
-                                    );
-                                    self.incoming.push_back(conn);
+                                Some(DatagramEvent::NewConnection(incoming)) => {
+                                    self.incoming.push_back((incoming, response_buffer));
```
Ralith, line 457: Do we have any limits on the length of `self.incoming`? This might be a memory exhaustion hazard, even though the individual elements shouldn't be large. Previously `ServerConfig::concurrent_connections` limited it, but now we don't check that until later.

`ServerConfig::concurrent_connections` would still be a natural conservative limit, though we'd need to expose a getter; perhaps a hidden one, similar to `get_max_udp_payload_size`. We could even make the limit exact by exposing the current active connection count (e.g. publicizing `quinn_proto::Endpoint::known_connections`, maybe renamed to `connection_count`), which could make `ConnectionLimitExceeded` errors from `accept` unreachable via the `quinn` API, though I think the benefit is small.

Alternatively, we could just set a generous-but-finite ad-hoc threshold; say, 2^16 elements? That should be low enough to prevent memory from growing beyond 100MiB, and high enough not to lose connections except under load conditions that are catastrophic regardless. This would be simplest and I don't think it has much of a drawback.

quinn-proto/src/endpoint.rs
```diff
@@ -592,10 +583,22 @@ impl Endpoint {
         }
     }
 
+    /// Reject this incoming connection attempt
+    pub fn reject(&mut self, incoming: IncomingConnection, buf: &mut BytesMut) -> Transmit {
```
Ralith, line 587: bikeshed: should we name this "refuse", to match the RFC terminology and the error code name?

Ralith (changes requested):
LGTM aside from a few last minor issues.

quinn/src/endpoint.rs
gretchenfrage: Hm... Well, if you say that a 2^16 limit seems good enough for now, then it is definitely the simplest, so I think I'll go with that. Nevertheless, there is an alternative approach (I'm a bit unclear on whether this is what you're suggesting in the second paragraph) where we count pending incoming connections towards the connection limit on the quinn-proto level. This diff would illustrate basically how we could do that:

<details><summary>Large diff</summary>

```diff
diff --git a/quinn-proto/src/endpoint.rs b/quinn-proto/src/endpoint.rs
index 7ce4d944..89a5c8b0 100644
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -40,6 +40,7 @@ pub struct Endpoint {
     rng: StdRng,
     index: ConnectionIndex,
     connections: Slab<ConnectionMeta>,
+    num_incoming_connections: usize,
     local_cid_generator: Box<dyn ConnectionIdGenerator>,
     config: Arc<EndpointConfig>,
     server_config: Option<Arc<ServerConfig>>,
@@ -63,6 +64,7 @@ impl Endpoint {
             rng: rng_seed.map_or(StdRng::from_entropy(), StdRng::from_seed),
             index: ConnectionIndex::default(),
             connections: Slab::new(),
+            num_incoming_connections: 0,
             local_cid_generator: (config.connection_id_generator_factory.as_ref())(),
             config,
             server_config,
@@ -434,14 +436,23 @@ impl Endpoint {
             return None;
         }
 
-        if let Err(initial_close) =
-            self.check_connection_limit(version, addresses, &crypto, &src_cid, buf)
+        let server_config = self.server_config.as_ref().unwrap();
+
+        if (self.connections.len() + self.num_incoming_connections)
+            >= server_config.concurrent_connections as usize
+            || self.is_full()
         {
-            return Some(DatagramEvent::Response(initial_close));
+            debug!("refusing connection");
+            return Some(DatagramEvent::Response(self.initial_close(
+                version,
+                addresses,
+                &crypto,
+                &src_cid,
+                TransportError::CONNECTION_REFUSED(""),
+                buf,
+            )));
         }
 
-        let server_config = self.server_config.as_ref().unwrap();
-
         if dst_cid.len() < 8 && dst_cid.len() != self.local_cid_generator.cid_len() {
             debug!(
                 "rejecting connection due to invalid DCID length {}",
@@ -485,6 +496,8 @@ impl Endpoint {
             }
         };
 
+        self.num_incoming_connections += 1;
+
         Some(DatagramEvent::NewConnection(IncomingConnection {
             addresses,
             ecn,
@@ -507,15 +520,7 @@ impl Endpoint {
         now: Instant,
         buf: &mut BytesMut,
     ) -> Result<(ConnectionHandle, Connection), (ConnectionError, Option<Transmit>)> {
-        self.check_connection_limit(
-            incoming.version,
-            incoming.addresses,
-            &incoming.crypto,
-            &incoming.src_cid,
-            buf,
-        )
-        .map_err(|response| (ConnectionError::ConnectionLimitExceeded, Some(response)))?;
-
+        self.num_incoming_connections -= 1;
         let server_config = self.server_config.as_ref().unwrap().clone();
 
         let ch = ConnectionHandle(self.connections.vacant_key());
@@ -585,6 +590,7 @@ impl Endpoint {
 
     /// Reject this incoming connection attempt
     pub fn reject(&mut self, incoming: IncomingConnection, buf: &mut BytesMut) -> Transmit {
+        self.num_incoming_connections -= 1;
         self.initial_close(
             incoming.version,
             incoming.addresses,
@@ -606,6 +612,7 @@ impl Endpoint {
         if incoming.remote_address_validated() {
             return Err(RetryError(incoming));
         }
+        self.num_incoming_connections -= 1;
         let server_config = self.server_config.as_ref().unwrap();
 
         // First Initial
@@ -653,6 +660,11 @@ impl Endpoint {
         })
     }
 
+    /// Call when ignoring an incoming connection attempt to stop counting it towards the connection limit
+    pub fn ignore_incoming(&mut self) {
+        self.num_incoming_connections -= 1;
+    }
+
     fn add_connection(
         &mut self,
         ch: ConnectionHandle,
@@ -701,30 +713,6 @@ impl Endpoint {
         conn
     }
 
-    fn check_connection_limit(
-        &mut self,
-        version: u32,
-        addresses: FourTuple,
-        crypto: &Keys,
-        src_cid: &ConnectionId,
-        buf: &mut BytesMut,
-    ) -> Result<(), Transmit> {
-        let server_config = self.server_config.as_ref().unwrap();
-        if self.connections.len() < server_config.concurrent_connections as usize && !self.is_full()
-        {
-            return Ok(());
-        }
-        debug!("refusing connection");
-        Err(self.initial_close(
-            version,
-            addresses,
-            crypto,
-            src_cid,
-            TransportError::CONNECTION_REFUSED(""),
-            buf,
-        ))
-    }
-
     fn initial_close(
         &mut self,
         version: u32,
@@ -957,6 +945,11 @@ pub enum DatagramEvent {
     /// The datagram is redirected to its `Connection`
     ConnectionEvent(ConnectionHandle, ConnectionEvent),
     /// The datagram may result in starting a new `Connection`
+    ///
+    /// This event corresponds to an incrementing of the incoming connection counter, which counts
+    /// towards the active connection limit. Therefore, this incoming connection should not be
+    /// dropped without a corresponding call to `accept`, `reject`, `retry`, or `ignore_incoming`,
+    /// so as to decrement the incoming connection counter.
     NewConnection(IncomingConnection),
     /// Response generated directly by the endpoint
     Response(Transmit),
diff --git a/quinn/src/endpoint.rs b/quinn/src/endpoint.rs
index d962c7d4..119504e2 100644
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -815,4 +815,8 @@ impl EndpointInner {
             Err(e) => Err((e, response_buffer)),
         }
     }
+
+    pub(crate) fn ignore_incoming(&self) {
+        self.state.lock().unwrap().inner.ignore_incoming();
+    }
 }
diff --git a/quinn/src/incoming_connection.rs b/quinn/src/incoming_connection.rs
index 247cf751..9376b426 100644
--- a/quinn/src/incoming_connection.rs
+++ b/quinn/src/incoming_connection.rs
@@ -87,7 +87,7 @@ impl IncomingConnection {
 
     /// Ignore this incoming connection attempt, not sending any packet in response
     pub fn ignore(mut self) {
-        self.0.take().unwrap();
+        self.0.take().unwrap().endpoint.ignore_incoming();
     }
 }
 
```

</details>

But it does introduce a very easy to make and annoying to catch possible usage error of quinn-proto wherein the user discards an `quinn_proto::IncomingConnection` in one case or another without somehow triggering a decrement `num_incoming_connections` (or otherwise fails to keep that counter synchronized with reality), and there are other places in quinn where these sorts of ad-hoc heuristic numbers seem to exist, so I think for now I'll just do your final suggestion yeah, but this is here as a note if we feel a need to re-visit this in the future I guess.

quinn-proto/src/endpoint.rs
```diff
@@ -626,6 +698,31 @@ impl Endpoint {
...
+    fn check_connection_limit(
+        &mut self,
+        version: u32,
+        addresses: FourTuple,
+        crypto: &Keys,
+        src_cid: &ConnectionId,
+        buf: &mut BytesMut,
+    ) -> Result<(), Transmit> {
+        let server_config = self.server_config.as_ref().unwrap();
+        if self.connections.len() >= server_config.concurrent_connections as usize || self.is_full()
```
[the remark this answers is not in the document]
gretchenfrage: Done.

quinn-proto/src/tests/util.rs
```diff
@@ -287,11 +287,39 @@ pub(super) struct TestEndpoint {
...
     pub(super) connections: HashMap<ConnectionHandle, Connection>,
     conn_events: HashMap<ConnectionHandle, VecDeque<ConnectionEvent>>,
     pub(super) captured_packets: Vec<Vec<u8>>,
     pub(super) capture_inbound_packets: bool,
+    pub(super) retry_policy: RetryPolicy,
+}
+
+pub(super) struct RetryPolicy(
+    pub(super) Box<dyn Fn(&IncomingConnection) -> IncomingConnectionResponse>,
+);
```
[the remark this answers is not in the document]
gretchenfrage: Done. Renamed to `IncomingConnectionBehavior` though.

quinn/src/endpoint.rs
```diff
@@ -137,8 +138,11 @@ impl Endpoint {
 
     /// Get the next incoming connection attempt from a client
     ///
-    /// Yields [`Connecting`] futures that must be `await`ed to obtain the final `Connection`, or
-    /// `None` if the endpoint is [`close`](Self::close)d.
+    /// Yields [`IncomingConnection`]s, or `None` if the endpoint is [`close`](Self::close)d.
+    /// [`IncomingConnection`] can be `await`ed to obtain the final
+    /// [`Connection`](crate::Connection), or used to eg. filter connection attempts or force
+    /// address validation, or converted into an intermediate `Connecting` future which can be
+    /// used to eg. send 0.5-RTT data.
```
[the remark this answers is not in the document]
gretchenfrage: Done.

quinn/src/endpoint.rs
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
gretchenfrage: I can if you want, but there's also a lot of existing places in the repository where the word "reject" is used, such as in `reject_new_connections`. Thoughts?

quinn/src/endpoint.rs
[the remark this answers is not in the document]
Ralith: Missed the first instance, "eg. filter connection attempts"

quinn/src/lib.rs
```diff
@@ -125,3 +127,7 @@ const SEND_TIME_BOUND: Duration = Duration::from_micros(50);
 /// generated from the endpoint (retry or initial close) can be dropped when this limit is being execeeded.
 /// Chose to represent 100 MB of data.
 const MAX_TRANSMIT_QUEUE_CONTENTS_LEN: usize = 100_000_000;
+
+/// The maximum number of `IncomingConnection`s we allow to be enqueued at a time before we start
+/// rejecting new `IncomingConnection`s automatically.
```
Ralith, line 132: Let's include a few words as to how this value was chosen: a nice round number that implies <100MiB or so, presuming `IncomingConnection` accounts for something like 1200 bytes of memory.

quinn-proto/src/endpoint.rs
Ralith: Hmm, good point. No strong opinion; we can revisit in a quick follow-up if we want.

quinn/src/lib.rs
gretchenfrage: Done.

quinn/src/endpoint.rs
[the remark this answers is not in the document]
gretchenfrage: Oh, right. Done.

Ralith (approved):
Great work, thanks!

gretchenfrage:
Yaaaaay

quinn/src/incoming_connection.rs
```diff
@@ -0,0 +1,149 @@
...
+        Self(Some(State {
+            inner,
+            endpoint,
+            response_buffer,
+        }))
+    }
+
+    /// The local IP address which was used when the peer established
+    /// the connection
+    pub fn local_ip(&self) -> Option<IpAddr> {
```
djc, line 42: Nit: these simple getters should live below the consuming functions (`accept()` et al).

quinn-proto/src/endpoint.rs
```diff
@@ -434,77 +435,32 @@ impl Endpoint {
             return None;
         }
 
-        let server_config = self.server_config.as_ref().unwrap().clone();
-
-        if self.connections.len() >= server_config.concurrent_connections as usize || self.is_full()
+        if let Err(initial_close) =
+            self.check_connection_limit(version, addresses, &crypto, &src_cid, buf)
```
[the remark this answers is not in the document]
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
```diff
@@ -542,44 +537,119 @@ impl Endpoint {
...
     }
 
+    /// Respond with a retry packet, requiring the client to retry with address validation
+    ///
+    /// Errors if `incoming.remote_address_validated()` is true.
+    fn retry(
+        &mut self,
+        incoming: IncomingConnection,
+        buf: &mut BytesMut,
+    ) -> Result<Transmit, RetryError> {
```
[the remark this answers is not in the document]
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
```diff
@@ -592,6 +662,7 @@ impl Endpoint {
         tls: Box<dyn crypto::Session>,
         server_config: Option<Arc<ServerConfig>>,
         transport_config: Arc<TransportConfig>,
+        path_validated: bool,
```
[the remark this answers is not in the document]
gretchenfrage: Note the removal of this line below:

```diff
-         let path_validated = server_config.as_ref().map_or(true, |c| c.use_retry);
```

Previously, when constructing a `Connection`, it based whether the address was validated on whether the server's `use_retry` was true. But now that the user can determine that dynamically we have to actually thread it through dynamically.

quinn-proto/src/endpoint.rs
```diff
@@ -863,6 +959,48 @@ pub enum DatagramEvent {
     Response(Transmit),
 }
 
+/// An incoming connection for which the server has not yet begun its part of the handshake.
+struct IncomingConnection {
```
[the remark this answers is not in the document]
gretchenfrage: I like that. Done.

quinn/src/endpoint.rs
```diff
@@ -383,10 +383,34 @@ pub(crate) struct State {
...
     transmit_queue_contents_len: usize,
 }
 
+impl TransmitState {
+    fn respond(&mut self, transmit: proto::Transmit, mut response_buffer: BytesMut) {
+        // Limiting the memory usage for items queued in the outgoing queue from endpoint
+        // generated packets. Otherwise, we may see a build-up of the queue under test with
+        // flood of initial packets against the endpoint. The sender with the sender-limiter
+        // may not keep up the pace of these packets queued into the queue.
+        if self.transmit_queue_contents_len < MAX_TRANSMIT_QUEUE_CONTENTS_LEN {
```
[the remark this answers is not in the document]
gretchenfrage: Done.

quinn/src/endpoint.rs
```diff
@@ -765,3 +773,52 @@ impl std::ops::Deref for EndpointRef {
         &self.0
     }
 }
+
+impl EndpointInner {
```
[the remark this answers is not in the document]
gretchenfrage: Done.

quinn/src/incoming_connection.rs
```diff
@@ -0,0 +1,149 @@
...
+};
+
+/// An incoming connection for which the server has not yet begun its part of the handshake
+pub struct IncomingConnection(Option<State>);
+
+struct State {
+    inner: proto::IncomingConnection,
+    endpoint: EndpointRef,
+    response_buffer: BytesMut,
+}
```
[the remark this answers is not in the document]
gretchenfrage: Done.

quinn/src/incoming_connection.rs
gretchenfrage: Done.

gretchenfrage:
Took me a bit longer than expected but it should be ready for another look now.

I also pushed back the adding of the connection limit check to proto `accept` until the commit which makes `accept` public, to be consistent with the spirit of the request to delay the adding of the error check to `retry`.

lijunwangs:
Hi @djc @Ralith, any plan on getting this into main? I found this capability very useful.

djc:
Yes, will be merged soon I think.

gretchenfrage:
Just rebased over main, resolving some non-trivial merge conflicts with [`95ac70d6`](https://github.com/quinn-rs/quinn/commit/95ac70d6b672eed58de053fa6f0b0d16f9ac73c5)

quinn-proto/src/endpoint.rs
```diff
@@ -490,24 +489,30 @@ impl Endpoint {
...
-                Ok((ch, conn)) => Some(DatagramEvent::NewConnection(ch, conn)),
-                Err((_, response)) => response.map(DatagramEvent::Response),
-            }
-        }
+        }))
     }
 
     /// Attempt to accept this incoming connection (an error may still occur)
-    fn accept(
+    pub fn accept(
```
djc, line 496: The return type here (`Result<(ConnectionHandle, Connection), (ConnectionError, Option<Transmit>)>`) seems pretty elaborate. Can we do something a little more friendly?

quinn-proto/src/endpoint.rs
```diff
@@ -602,8 +607,29 @@ impl Endpoint {
...
     /// Respond with a retry packet, requiring the client to retry with address validation
-    fn retry(&mut self, incoming: Incoming, buf: &mut BytesMut) -> Transmit {
+    ///
+    /// Errors if `incoming.remote_address_validated()` is true.
+    pub fn retry(
+        &mut self,
+        incoming: Incoming,
+        buf: &mut BytesMut,
+    ) -> Result<Transmit, RetryError> {
+        if incoming.remote_address_validated() {
```
djc, line 630: This feels a little silly. Should `Incoming` be an `enum` instead with variants that signal whether the remote address has been validated, so that we make it statically impossible to trigger this error?

djc (commented):
I took the liberty of (force-) pushing some changes to this branch to split out one of the commits and improve the `TransmitState` abstraction. I have a few more questions here, but looking forward to merging this soon!

quinn-proto/src/endpoint.rs
```diff
@@ -642,6 +639,16 @@ impl Endpoint {
         conn
     }
 
+    fn check_connection_limit(&mut self) -> Result<(), TransportError> {
```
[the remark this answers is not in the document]
Ralith: I've been considering that as well. I think the simplified API it produces is very appealing. I've hesitated because "limit to N connections" is probably a very reasonable common path to promote, but setting the config is probably not much easier than just rolling your own, as of this change. If removing `concurrent_connections` sounds good to you, then I'm on board.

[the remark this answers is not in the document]
gretchenfrage: I think that sounds like a good idea. It would allow users to implement it easily, and it's probably fine to allow arbitrarily many simultaneous connections for users who don't bother limiting it.

quinn-proto/src/endpoint.rs
gretchenfrage: Will experiment with.

quinn-proto/src/endpoint.rs
gretchenfrage: Hm. While I like the impulse, I can't think of a good way to make it work.

The first problem is that there would be methods duplicated between the two specific types. We'd want to implement `accept`, `reject`, `ignore`, `local_ip`, and `remote_address` on both the retryable and non-retryable variants, and possibly the parent enum too. Possibly we could factor this out into a trait which all three implement, but overall that seems over-complicated.

A second problem would come up if we start sending `NEW_TOKEN` frames. That would introduce the possibility of an `Incoming` which is both validated and retryable. We would want our API to allow those to be retried, but a typical user application would not actually want to retry them. This would create an awkward situation with any enum-based way of structuring this I can think of.

As one example, if we established this as:

```rs
enum Incoming {
    Validated(ValidatedIncoming),
    NotValidated(NotValidatedIncoming),
}
```

Then introducing `NEW_TOKEN` frames would mean we would want to make `NotValidatedIncoming` sometimes retryable, so we'd be back to the error not being statically prevented.

As an alternative example, if we established this as:

```rs
enum Incoming {
    Retryable(RetryableIncoming),
    NotRetryable(NotRetryableIncoming),
}
```

That would solve the above issue but make applications implementing retry logic awkwardly complicated, because they'd want to be doing:

```rs
match incoming {
    Incoming::Retryable(incoming) => if incoming.remote_address_validated() {
        handle(incoming.accept())
    } else {
        incoming.retry()
    },
    Incoming::NotRetryable(incoming) => handle(incoming.accept()),
}
```

Which seems gross.

Having there just be a `remote_address_validated` method is nice in that it avoids those problems. If we introduce `NEW_TOKEN` frame sending then we can just also introduce a `may_retry` method, such that `!remote_address_validated` implies `may_retry` but not necessarily the other way around (like I originally had in my prototype [[link]](https://github.com/gretchenfrage/quinn/pull/1#issuecomment-1909192157)). Which would be nice because we'd be exposing maximal freedom but without requiring existing applications to contort themselves to adapt to it.

As to the awkwardness of this solution not statically preventing the error, I might compare that to say, how `std::fs::File` still implements `Write` even if you opened it in read-only mode.

quinn-proto/src/endpoint.rs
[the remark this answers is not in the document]
djc: Okay, let's do it. @gretchenfrage are you able to pick this up? I'd suggest just replacing the first commit by just ripping out everything that has to do with connection limit, and then somewhere at the end tacking on a commit that allows the caller to get the `Endpoint`'s current `open_connections()`.

quinn-proto/src/endpoint.rs
djc: Thanks for the in-depth response!

> The first problem is that there would be methods duplicated between the two specific types. We'd want to implement `accept`, `reject`, `ignore`, `local_ip`, and `remote_address` on both the retryable and non-retryable variants, and possibly the parent enum too. Possibly we could factor this out into a trait which all three implement, but overall that seems over-complicated.

This doesn't seem like a big deal? We could use a private type that underlies both of the enum variant types and expose these methods only on the enum type.

> A second problem would come up if we start sending `NEW_TOKEN` frames. That would introduce the possibility of an `Incoming` which is both validated and retryable. We would want our API to allow those to be retried, but a typical user application would not actually want to retry them. This would create an awkward situation with any enum-based way of structuring this I can think of.

Not sure if this is true? 9000 says in 8.1.2:

> In response to processing an Initial packet containing a token that was provided in a Retry packet, a server cannot send another Retry packet; it can only refuse the connection or permit it to proceed.

So, if the client sends a token, the server is not allowed to send a retry. I'm not sure how you think this ties in with `NEW_TOKEN`, but from my skimming of the RFC it seems like `NEW_TOKEN` would never be sent in Initial packets, only on established connections -- but I can't find an authoritative reference for this.

quinn/src/incoming.rs
```diff
@@ -0,0 +1,149 @@
...
+pub struct RetryError(Incoming);
+
+impl RetryError {
+    /// Get the [`Incoming`]
+    pub fn into_incoming(self) -> Incoming {
+        self.0
+    }
+}
+
+/// Basic adapter to let [`Incoming`] be `await`-ed like a [`Connecting`]
```
ryleung-solana, line 127: I'm not sure we really need this adapter - perhaps only an explicit accept might be better

quinn/src/endpoint.rs
```diff
@@ -137,8 +138,10 @@ impl Endpoint {
 
     /// Get the next incoming connection attempt from a client
     ///
-    /// Yields [`Connecting`] futures that must be `await`ed to obtain the final `Connection`, or
-    /// `None` if the endpoint is [`close`](Self::close)d.
+    /// Yields [`Incoming`]s, or `None` if the endpoint is [`close`](Self::close)d. [`Incoming`]
```
ryleung-solana, line 141: Connecting to my comment on the IncomingFuture adapter not really having an obvious use - perhaps a more useful  convenience utility would be to be able to configure an Endpoint to accept, reject or verify with a supplied predicate. Something like 

```
enum FilterDecision {
    Accept,
    Reject,
    Validate,
}

filter_incoming: F where F: FnMut(incoming: Incoming, remote: IpAddr) -> FilterDecision
```

Ralith: That seems strictly less powerful, and also more difficult to use, than inserting whatever logic you like after a successful `Endpoint::accept`.

quinn/src/incoming.rs
Ralith: I like it, personally. It preserves syntactic compatibility with existing code, and is simultaneously reasonably obvious, concise, and doesn't cost us much to support.

quinn-proto/src/endpoint.rs
Ralith: > Not sure if this is true?

@gretchenfrage's analysis is correct. `NEW_TOKEN` frames are sent by the server to provide a client with tokens that the client may then choose to supply in future Initial packets. Per your quote, because these tokens were not provided in a Retry packet, the server may send a Retry in response, even though a token was present.

djc: Ah, alright then! I retract my objection.

quinn/src/incoming.rs
gretchenfrage: I think that if the adapter didn't exist it would be necessary for each user to basically reinvent its logic. What it does is abstracting over synchronous versus asynchronous failure to fully accept an `Incoming`, which probably is what almost any user is going to do with the exception of if they're using 0-RTT.

quinn/src/endpoint.rs
gretchenfrage: I think that sort of "accept/reject/retry logic as middleware" architecture would be good for a high level framework, but for a low-level transport implementation we should tend towards giving the user more control.

quinn-proto/src/endpoint.rs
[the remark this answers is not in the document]
gretchenfrage: Yep I'll do that.

quinn-proto/src/endpoint.rs
gretchenfrage: Yeah I'm thinking just replacing `(ConnectionError, Option<Transmit>)` with:

```rs
/// Error type for attempting to accept an [`Incoming`]
#[derive(Debug)]
pub struct AcceptError {
    /// Underlying error describing reason for failure
    pub cause: ConnectionError,
    /// Optional response to transmit back
    pub response: Option<Transmit>,
}
```

Should be sufficient

gretchenfrage:
There were some non-trivialities to removing `concurrent_connections`. I took the liberty of refactoring some things a tiny bit. Let me know if you disagree with these decisions.

Updated branch to remove `concurrent_connections` from `ServerConfig`. Adds `open_connections(&self) -> usize` method to `Endpoint` which user can use to enforce concurrent connection limit manually. Also, now removes existing `reject_new_connections` method on endpoint, as that relies on `concurrent_connections` and also can now just be done by user manually with the new "incoming" API we're adding.

Previous version of this branch added `ConnectionError::ConnectionLimitExceeded` variant. New version instead adds `ConnectionError::CidsExhausted` variant. In previous version of this branch, `check_connection_limit` method was factored out and called when accepting connections. It did two things: check concurrent connections, and check `is_full`, and cause `ConnectionLimitExceeded` error if either case true. In new version of this branch, we rename `is_full` to `cids_exhausted` and just directly call `is_full` when accepting connections and cause `CidsExhausted` error if true.

On main, `ConnectError` has `TooManyConnections` variant, with comment "The number of active connections on the local endpoint is at the limit." It is _not_ triggered if `concurrent_connections` limit is reached (which I think is confusing given the error name and message), but rather is only triggered if `is_full` returns true. We rename `ConnectError::TooManyConnections` to `ConnectError::CidsExhausted` so it looks basically the same as `ConnectionError::CidsExhausted`.

---

Other changes I made to this branch:

- Squashed your `quinn: improve TransmitState abstraction` into prev
- Adds quinn proto `AcceptError` as noted in previous comment
- Adds --connection-limit parameter to server example

quinn-proto/src/endpoint.rs
gretchenfrage: Did that.

quinn-proto/src/endpoint.rs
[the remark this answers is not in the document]
gretchenfrage: See comment below.

quinn-proto/src/endpoint.rs
```diff
@@ -609,6 +623,76 @@ impl Endpoint {
...
     }
 
+    /// Reject this incoming connection attempt
+    pub fn reject(&mut self, incoming: Incoming, buf: &mut BytesMut) -> Transmit {
+        self.initial_close(
+            incoming.version,
+            incoming.addresses,
+            &incoming.crypto,
+            &incoming.src_cid,
+            TransportError::CONNECTION_REFUSED(""),
```
lijunwangs, line 633: Is it possible to allow the app specify the reason so  that the upper layer client may do things differently?

quinn/src/incoming.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /// Attempt to accept this incoming connection (an error may still occur)
+    pub fn accept(mut self) -> Result<Connecting, ConnectionError> {
+        let state = self.0.take().unwrap();
+        state.endpoint.accept(state.inner, state.response_buffer)
+    }
+
+    /// Reject this incoming connection attempt
+    pub fn reject(mut self) {
```
lijunwangs, line 41: If the application simply drops the Incoming and not calling reject, will the server still bother sending the INITIAL_CLOSE?

Ralith: The `Drop` impl just below calls `reject` automatically. Call `ignore` if you don't want that.

quinn-proto/src/endpoint.rs
Ralith: The spec [specifically warns](https://www.rfc-editor.org/rfc/rfc9000.html#section-10.2.3-3) against this:

> Sending a CONNECTION_CLOSE of type 0x1d in an Initial or Handshake packet could expose application state or be used to alter application state. A CONNECTION_CLOSE of type 0x1d MUST be replaced by a CONNECTION_CLOSE of type 0x1c when sending the frame in Initial or Handshake packets. Otherwise, information about the application state might be revealed. Endpoints MUST clear the value of the Reason Phrase field and SHOULD use the APPLICATION_ERROR code when converting to a CONNECTION_CLOSE of type 0x1c

Ralith:
I went through and s/reject/refuse/ in the API for consistency w/ the spec and general IETF convention, now that we no longer need to worry about not matching `fn reject_new_connections`.

quinn-proto/src/endpoint.rs
```diff
@@ -609,6 +623,76 @@ impl Endpoint {
...
+    }
+
+    /// Respond with a retry packet, requiring the client to retry with address validation
+    ///
+    /// Errors if `incoming.remote_address_validated()` is true.
+    pub fn retry(
+        &mut self,
+        incoming: Incoming,
+        buf: &mut BytesMut,
+    ) -> Result<Transmit, RetryError> {
```
lijunwangs, line 645: In earlier logic, we were checking stateless_packets_supressed, and if it is, we are omitting sending the retry. Is this no longer relevant?

quinn-proto/src/endpoint.rs
lijunwangs: sounds good

lijunwangs:
What are the remaining outstanding issues? Can we get this merged into main if no more issues? @djc @Ralith

quinn/src/endpoint.rs
```diff
@@ -356,12 +359,57 @@ pub(crate) struct EndpointInner {
...
+    pub(crate) fn accept(
+        &self,
+        incoming: proto::Incoming,
+        mut response_buffer: BytesMut,
+    ) -> Result<Connecting, ConnectionError> {
+        let mut state = self.state.lock().unwrap();
+        state
+            .inner
+            .accept(incoming, Instant::now(), &mut response_buffer)
+            .map(|(handle, conn)| {
```
djc, line 372: Style nit: I prefer to avoid `map().map_err()` constructions in favor of just having a `match` expression, which uses more explicit syntax to do the same thing.

quinn-proto/src/tests/util.rs
```diff
@@ -428,8 +450,58 @@ impl TestEndpoint {
...
+            })
+            .map_err(|error| {
+                if let Some(transmit) = error.response {
+                    let size = transmit.size;
+                    self.outbound
+                        .extend(split_transmit(transmit, buf.split_to(size).freeze()));
+                }
+                self.accepted = Some(Err(error.cause.clone()));
+                error.cause
+            })
```
djc, line 474: Same here with `map().map_err()`.

quinn/src/endpoint.rs
```diff
@@ -413,14 +461,14 @@ impl State {
...
-                                Some(DatagramEvent::NewConnection(handle, conn)) => {
-                                    let conn = self.connections.insert(
-                                        handle,
-                                        conn,
-                                        self.socket.clone(),
-                                        self.runtime.clone(),
-                                    );
-                                    self.incoming.push_back(conn);
+                                Some(DatagramEvent::NewConnection(incoming)) => {
+                                    if self.incoming.len() < MAX_INCOMING_CONNECTIONS {
```
djc, line 465: Does it still make sense to check against `MAX_INCOMING_CONNECTIONS` here when we now allow the caller to decide whether to accept the connection?

djc (approved):
I have a few more minor questions/comments, but I think we should merge this soon.

quinn/src/endpoint.rs
gretchenfrage: Yeah. `MAX_INCOMING_CONNECTIONS` is a new thing we're introducing in this PR. When the endpoint driver task receives a potentially connection-creating packet, it pushes an `Incoming` to `quinn::endpoint::State`'s `incoming` `VecDeque`, from which the user the application can get it through `Endpoint::accept`. The size of this queue should be close to zero in healthy conditions. We're adding this conservative limit so that an endpoint doesn't just use indefinitely more memory in a situation where incoming connection attempts are happening faster than the application is processing them for a prolonged period of time.

quinn-proto/src/tests/util.rs
gretchenfrage: Done.

quinn/src/endpoint.rs
gretchenfrage: Done.

gretchenfrage:
Made requested tweaks and rebased over main.

gretchenfrage:
Tests passed, so should be ready to merge I think!

lijunwangs:
> Tests passed, so should be ready to merge I think!

@gretchenfrage , @djc there are some unresolved conversations -- I think we just need to resolve them. I do not see any real outstanding issues.

djc:
@lijunwangs there is no need to keep litigating this -- please assume I'm well aware of the state.

djc (approved):

lijunwangs:
@djc -- sorry for pushing for this -- it is a critical improvement for our use case. Thank you all!

PR #1758, opened 2024-02-16
Enable packet info and related flags for quinn-udp on Android

conectado:
We're using the info obtained with `PKT_INFO`, particularly the destination IP, on multiple platforms including on android.

Most of the socket options used for linux, save for GRO, are available on android, so it's simply adding it to the conditional compilation.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -46,7 +46,11 @@ impl UdpSocketState {
     pub fn new(sock: UdpSockRef<'_>) -> io::Result<Self> {
         let io = sock.0;
         let mut cmsg_platform_space = 0;
-        if cfg!(target_os = "linux") || cfg!(target_os = "freebsd") || cfg!(target_os = "macos") {
+        if cfg!(target_os = "linux")
+            || cfg!(target_os = "freebsd")
+            || cfg!(target_os = "macos")
+            || cfg!(target_os = "android")
+        {
             cmsg_platform_space +=
                 unsafe { libc::CMSG_SPACE(mem::size_of::<libc::in6_pktinfo>() as _) as usize };
         }
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -76,9 +76,10 @@ impl UdpSocketState {
         }
 
         let mut may_fragment = false;
-        #[cfg(target_os = "linux")]
+        #[cfg(any(target_os = "linux", target_os = "android"))]
         {
             // opportunistically try to enable GRO. See gro::gro_segments().
+            #[cfg(target_os = "linux")]
             let _ = set_socket_option(&*io, libc::SOL_UDP, libc::UDP_GRO, OPTION_ON);
 
             // Forbid IPv4 fragmentation. Set even for IPv6 to account for IPv6 mapped IPv4 addresses.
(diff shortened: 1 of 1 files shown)


djc (approved):

Ralith (changes requested):
Thanks! A few notes:
- Do we need to add android to the case on line 49 too?
- Are GSO/GRO ever available on Android?
- Would love to have CI coverage for this, if you're up for it.

Ralith:
@conectado do you plan to do any follow-up work here?

thomaseizinger:
@Ralith Unfortunately, your force-push just broke our CI because we are depending on this commit hash. If you'd like to update the PR, we'd prefer to do it via merge commits only.

thomaseizinger:
We do plan on following up here but we are preparing the product launch at the moment so there hasn't yet been time for this! :)

thomaseizinger:
> @Ralith Unfortunately, your force-push just broke our CI because we are depending on this commit hash. If you'd like to update the PR, we'd prefer to do it via merge commits only.

I'll make a dedicated fork for now to un-break CI.

thomaseizinger:
> * Are GSO/GRO ever available on Android?

We haven't tested this specifically but at least the destination IP is available with this PR.

> * Would love to have CI coverage for this, if you're up for it.

How do you envision to test this in CI? I guess we'd have to use something like QEMU to emulate common Android CPU architectures and then run the tests on it?

Ralith:
> We haven't tested this specifically but at least the destination IP is available with this PR.

GSO/GRO are unrelated to destination IPs, but are a big CPU efficiency win. They're well-established on Linux so I'd be surprised if they're absent from Android. Not a blocker for this PR, but maybe something you'd want to double-check if you're targeting that platform.

> How do you envision to test this in CI? I guess we'd have to use something like QEMU to emulate common Android CPU architectures and then run the tests on it?

I'm not familiar with Android workflows, so I'm not sure. Some sort of emulator makes sense. Maybe crib from e.g. https://github.com/rustls/rustls-platform-verifier/blob/main/.github/workflows/ci.yml? CI coverage is particularly important here as we're not Android experts ourselves.

Ralith:
Sure. The only blocking concern is updating the size check on line 49.

thomaseizinger:
> Sure. The only blocking concern is updating the size check on line 49.

Awesome, thank you! I'll make those changes and open an issue about adding Android CI. Eventually, we want a decent Android CI for the project so once we have to acquire that knowledge, I might be able to port some of it to `quinn`.

thomaseizinger:
> > Sure. The only blocking concern is updating the size check on line 49.
> 
> Awesome, thank you! I'll make those changes and open an issue about adding Android CI. Eventually, we want a decent Android CI for the project so once we have to acquire that knowledge, I might be able to port some of it to `quinn`.

I opened an issue and [linked](https://users.rust-lang.org/t/twir-call-for-participation/4821/621?u=thomaseizinger) it on the rust-lang forum to be included in next week's TWIR. Maybe somebody volunteers :)

I'll nudge @conectado to push a fix for the size check!

Ralith (approved):
Thanks!

conectado:
@thomaseizinger thanks for taking over! :)

@ralith Pushed the fix.

Also, it will probably just work enabling GRO/GSO in android since the socket options [exists in the kernel](https://cs.android.com/android/platform/superproject/main/+/main:prebuilts/runtime/mainline/runtime/sdk/android/arm/include/bionic/libc/kernel/uapi/linux/udp.h;l=20-21?q=udp_segment&ss=android%2Fplatform%2Fsuperproject%2Fmain) I can try it on a follow up PR when I have a bit more availability.

Issue #1756, opened 2024-02-15
Sent ACK state accumulates when acting as a pure receiver


honzasp:
I have a Quinn client and server, where the server sends a stream of datagrams to the client. I found that the client is leaking memory, and memory profiling with Heaptrack has shown that the culprit is the `BTreeMap::insert()` call in `quinn_proto::connection::spaces::PacketSpace::sent()`, called from `PacketBuilder::finish_and_track()`. Here is a screenshot from Heaptrack:

![Screenshot from 2024-02-15 21-06-45](https://github.com/quinn-rs/quinn/assets/64298/b6610de5-41ba-47e2-8ca6-25bfe42b1b44)

Here is a minimal reproducing example:
https://github.com/honzasp/quinn_leak

It seems that packets are added to `PacketSpace::sent_packets` and then never removed. Perhaps because the data is sent using datagrams and not via streams?

Ralith:
Thanks for the report! Packets are removed from `sent_packets` when they are acknowledged or deemed lost. This mechanism is not sensitive to the contents of packets, including whether they contain stream or datagrams data. Have you verified that `sent_packets.len()` grows without bound?

honzasp:
Thanks for your prompt response! I added a debug print of `self.sent_packets.len()` to `PacketSpace::sent()` and it grows without bound in the client (but not in the server).

Ralith:
Interesting; thanks for confirming. I'll try to have a look this weekend.

Ralith:
The packets accumulating on the client side, i.e. on the receiver (I missed that at first), are ACKs. Because the client never sends anything ACK-eliciting (i.e. application data) itself, the sender is never technically required to transmit its own ACKs, without which the client cannot discard state. We should transmit ACKs occasionally even when not required to. We used to, but broke it in https://github.com/quinn-rs/quinn/pull/1415. We should also defensively forget ACK-only packets that accumulate beyond a threshold.

As an immediate mitigation, you can arrange for an endpoint which is only receiving data at the application layer for long periods to occasionally send something. For example, in the repro, having the client send one datagram for every 1000 it receives prevents state accumulation.

[Relevant RFC text](https://www.rfc-editor.org/rfc/rfc9000.html#name-generating-acknowledgments):
[...]

PR #1761, opened 2024-02-19
Improve behavior when only one peer sends ACK-eliciting packets

### LINKED ISSUE #1760: ACKs should be sent unilaterally when the number of unacknowledged packets is large
Sending ACKs is only required after receiving an ACK-eliciting packet, but an endpoint that mostly sends data might go arbitrarily long without that happening. In those cases, for the benefit of the receiver we should occasionally include ACK frames when sending ACK-eliciting data.

Spec:
> When sending a packet for any reason, an endpoint SHOULD attempt to include an ACK frame if one has not been sent recently. Doing so helps with timely loss detection at the peer.

See also https://github.com/quinn-rs/quinn/issues/1756.

Ralith:
Fixes #1760, #1756.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1321,7 +1321,7 @@ impl Connection {
 
         let mut ack_eliciting_acked = false;
         for packet in newly_acked.elts() {
-            if let Some(info) = self.spaces[space].sent_packets.remove(&packet) {
+            if let Some(info) = self.spaces[space].take(packet) {
                 if let Some(acked) = info.largest_acked {
                     // Assume ACKs for all packets below the largest acknowledged in `packet` have
                     // been received. This can cause the peer to spuriously retransmit if some of
@@ -1593,14 +1593,14 @@ impl Connection {
                 size_of_lost_packets
             );
 
-            for packet in &lost_packets {
-                let info = self.spaces[pn_space].sent_packets.remove(packet).unwrap(); // safe: lost_packets is populated just above
+            for &packet in &lost_packets {
+                let info = self.spaces[pn_space].take(packet).unwrap(); // safe: lost_packets is populated just above
                 self.remove_in_flight(pn_space, &info);
                 for frame in info.stream_frames {
                     self.streams.retransmit(frame);
                 }
                 self.spaces[pn_space].pending |= info.retransmits;
-                self.path.mtud.on_non_probe_lost(*packet, info.size);
+                self.path.mtud.on_non_probe_lost(packet, info.size);
             }
 
             if self.path.mtud.black_hole_detected(now) {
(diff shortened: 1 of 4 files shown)


quinn-proto/src/connection/spaces.rs
```diff
@@ -695,6 +699,17 @@ impl PendingAcks {
     pub(super) fn ranges(&self) -> &ArrayRangeSet {
         &self.ranges
     }
+
+    /// Should be called immediately before a non-probing packet is composed
+    pub(super) fn on_send_scheduled(&mut self) {
+        // If we're going to send a packet anyway, and we've received a significant number of
+        // non-ACK-eliciting packets, then include an ACK to help the peer perform timely loss
+        // detection even if they're not sending any ACK-eliciting packets themselves.
+        const LAZY_ACK_THRESHOLD: u64 = 10;
```
djc, line 763: This seems fairly low to me?

quinn-proto/src/connection/mod.rs
```diff
@@ -2949,6 +2949,7 @@ impl Connection {
         let mut sent = SentFrames::default();
         let space = &mut self.spaces[space_id];
         let is_0rtt = space_id == SpaceId::Data && space.crypto.is_none();
+        space.pending_acks.on_send_scheduled();
```
djc, line 2950: If we only call this in `populate_packet()`, how does that help us sending ACKs even if the connection isn't otherwise sending?

djc (approved):

quinn-proto/src/connection/spaces.rs
Ralith: My thinking is that the impact is modest (ACK frames are usually small), so returns probably diminish rapidly. We're also constraining this to whenever the local endpoint has something to send regardless, so the actual rate may be lower.

I didn't think super hard about it, though. Want to propose another figure?

quinn-proto/src/connection/mod.rs
Ralith: Good question; I got some wires crossed here. As the comment in the definition suggests, we should be counting the number of packets we've *received*, not sent. Revising.

Ralith: This particular commit improves the behavior of the peer which is regularly sending ACK-eliciting data, but not receiving any. We're counting the number of non-ACK-eliciting packets received since we last sent an ACK, which is (probably) proportional to the state the receiver is keeping in memory. If `populate_packet` doesn't get called, then we're not generating traffic that must be ACKed, so there's no problem. If it does, then we emit an occasional spontaneous ACK to keep that count down.

Thoughts on how we might document this better?

quinn-proto/src/connection/spaces.rs
djc: I guess it's fine.

quinn-proto/src/connection/mod.rs
djc: I guess I find the `on_send_scheduled()` method name (and the docstring) kind of opaque -- it doesn't do much to clarify what it's actually doing. For methods that do a bunch of admin work, naming after trigger event seems okay, but I think here we could be more specific? `maybe_ack_non_eliciting()`?

Ralith: SGTM; renamed and expanded doc.

PR #1764, opened 2024-02-20
Document cancel-safety of stream I/O operations

### LINKED ISSUE #1763: Unclear which, if any `RecvStream::read_*` methods are cancel-safe
Presumably RecvStream::read` can be cancelled and retried without missing reads, but I'm not sure just from the docs.

Ralith:
Fixes #1763.

### DIFF
--- a/quinn/src/recv_stream.rs
+++ b/quinn/src/recv_stream.rs
@@ -21,6 +21,13 @@ use crate::{
 /// - A variant of [`ReadError`] has been yielded by a read call
 /// - [`stop()`] was called explicitly
 ///
+/// # Cancellation
+///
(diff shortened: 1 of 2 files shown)


djc (approved):
Great work!

How hard is it to explicitly describe what the behavior in the non-safe scenario is? I feel like that would help clarify why a particular operation is/cannot be cancel-safe.

Ralith:
I tried to cover that in the type-level documentation, to limit repetition:
> A `read` method is said to be *cancel-safe* when dropping its future before the future becomes ready cannot lead to loss of stream data.

i.e. the non-safe scenario does not provide that guarantee. Does that need to be further clarified? It's a fairly precise statement already.

djc:
Yeah, I guess it's good enough.

PR #1765, opened 2024-02-21
fix(quinn-udp): use TOS for IPv4-mapped IPv6 dst addrs

mxinden:
When handling [IPv4-mapped IPv6 addresses](https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html#ipv4-mapped-ipv6-addresses) (e.g. `::ffff:127.0.0.1`) use IPv4 TOS (type of service) instead of IPv6 Traffic Class to encode ECN value. Otherwise ECN value is ignored by OS.

Tested on Ubuntu 23.10.

Setting this as draft for now. I am not familiar enough with IPv4-mapped IPv6 addresses across operating systems to feel confident in the patch proposed here. Input welcome.

Needed for https://github.com/mozilla/neqo/pull/1604.

//CC @larseggert

(Thank you for `quinn-udp`! :heart: )


Ralith:
> Setting this as draft for now. I am not familiar enough with IPv4-mapped IPv6 addresses across operating systems to feel confident in the patch proposed here. Input welcome.

I think the change is reasonable on the face of it, and the rest comes down to empirical kernel behavior, which is what we have CI for. To wit, it looks like Windows and FreeBSD are unhappy, but I'm not sure it's with the meat of the change. Perhaps they don't like the test trying to bind a socket to an IPv6-mapped IPv4 address?

quinn-udp/src/lib.rs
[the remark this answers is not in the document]
mxinden: Much better. See https://github.com/quinn-rs/quinn/pull/1765/commits/539471dbb818420d53dbbe6aa0aa8ea7e86669bc.

quinn-udp/tests/tests.rs
```diff
@@ -91,6 +91,44 @@ fn ecn_v4() {
     }
 }
 
+#[test]
+#[cfg_attr(target_os = "windows", ignore)]
```
mxinden, line 95: I am failing to send an IPv4-mapped IPv6 packet to an IPv4 socket on windows thus far.

quinn-udp/tests/tests.rs
```diff
@@ -154,35 +193,34 @@ fn test_send_recv(send: &Socket, recv: &Socket, transmit: Transmit) {
...
+        for addr in [src, dst] {
+            match (send_v6, recv_v6) {
+                (_, false) => assert_eq!(addr, Ipv4Addr::LOCALHOST),
+                // Windows gives us real IPv4 addrs, whereas *nix use IPv6-mapped IPv4
+                // addrs. Canonicalize to IPv6-mapped for robustness.
+                (false, true) => {
+                    assert_eq!(ip_to_v6_mapped(addr), Ipv4Addr::LOCALHOST.to_ipv6_mapped())
+                }
+                (true, true) => {
+                    if addr != Ipv6Addr::LOCALHOST && addr != Ipv4Addr::LOCALHOST.to_ipv6_mapped() {
```
Ralith, line 213: nit: `assert`

quinn-udp/tests/tests.rs
Ralith: That definitely works fine in general; we have E2E test coverage for dual-stack servers talking to IPv4 clients. Looks like maybe Windows just doesn't like IP_TOS on IPv6 sockets? Maybe it works with `IPV6_TCLASS`? Let me know if you don't have a Windows machine you can easily test on and I can tinker with it myself.

Error for reference:

> called `Result::unwrap()` on an `Err` value: Os { code: 10022, kind: InvalidInput, message: "An invalid argument was supplied." }

mxinden: > Looks like maybe Windows just doesn't like IP_TOS on IPv6 sockets? Maybe it works with IPV6_TCLASS?

This is the hint I needed. I only applied the actual patch to `unix.rs`. Opposite to *nix, Windows actually errors, when using `IPV6_TCLASS` with an IPv4-mapped IPv6 address. Fixed with https://github.com/quinn-rs/quinn/pull/1765/commits/0b97f4638d059fb242768375284bed82f129528d. Thank you @Ralith!

quinn-udp/tests/tests.rs
mxinden: :see_no_evil: https://github.com/quinn-rs/quinn/pull/1765/commits/1294b7e25469d24efe6a6f0bbdca9d4f889a950c much better!

quinn-udp/tests/tests.rs
Ralith: I actually had that backwards; glad it was a useful prompt anyway!

mxinden:
Test is running and succeeding on all platforms now (no more `ignore`). @Ralith thank you for the help. Mind taking another look?

Ralith (approved):
Great fix, thanks!

PR #1769, opened 2024-02-28
Make `quinn_proto::Connection` and `quinn_proto::Endpoint` impl `Sync`

Pixelstormer:
The only things preventing these types from already being `Sync` were a lack of `Sync` bounds on trait objects they were storing, and a `RefCell` that was only being used to work around a limitation in the API of `BinaryHeap`.

Adding `Sync` bounds to the relevant traits was simple enough and didn't cause any warnings or errors, but to remove the `RefCell` it was necessary to rework the way that pending stream IDs were stored.

The specific motivation for this PR was so that I could store these types as components directly in the [Bevy ECS](https://bevyengine.org/), which requires component types be `Send + Sync` for multi-threaded system scheduling. The alternative would be to wrap the types in a `Mutex` or other synchronization primitive, but removing the limitation entirely seemed like the better option, given how the only non-trivial blocker was just a hacky workaround an API limitation.

The aforementioned rework to remove the `RefCell` also (IMO) simplifies the logic a fair bit, by utilizing the `BinaryHeap`'s sorting for both priority and fairness. Sorting for stream IDs with the same priority falls back to a `recency` counter, which is decremented each time that stream is processed. This causes the streams of the same priority to be 'cycled' through, in the same way that was achieved previously by popping from the front and pushing to the back of a `VecDeque` that has now been removed.  Technically the u64 could underflow and break the cycling, but that is highly unlikely to ever happen, due to the counter being initialized to `u64::MAX` and only being decremented by 1 at a time.

### DIFF
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -1,7 +1,4 @@
-use std::{
-    cell::RefCell,
-    collections::{hash_map, BinaryHeap, VecDeque},
-};
+use std::collections::{hash_map, BinaryHeap};
 
 use bytes::Bytes;
 use thiserror::Error;
@@ -237,7 +234,7 @@ impl<'a> SendStream<'a> {
         self.state.unacked_data += written.bytes as u64;
         trace!(stream = %self.id, "wrote {} bytes", written.bytes);
         if !was_pending {
-            push_pending(&mut self.state.pending, self.id, stream.priority);
+            self.state.pending.push_pending(self.id, stream.priority);
         }
         Ok(written)
     }
@@ -268,7 +265,7 @@ impl<'a> SendStream<'a> {
         let was_pending = stream.is_pending();
         stream.finish()?;
         if !was_pending {
-            push_pending(&mut self.state.pending, self.id, stream.priority);
+            self.state.pending.push_pending(self.id, stream.priority);
         }
 
         Ok(())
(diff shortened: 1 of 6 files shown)


Ralith:
> The aforementioned rework to remove the RefCell also (IMO) simplifies the logic a fair bit, by utilizing the BinaryHeap's sorting for both priority and fairness. Sorting for stream IDs with the same priority falls back to a recency counter, which is decremented each time that stream is processed.

I'm strongly in favor of getting rid of the `RefCell`, but the mechanism described here is more complex than rotating through a `VecDeque`, and more costly in the common case where most streams have the same priority. Some possible alternatives:
- Refactor the original design to store the `VecDeque`s in a `Slab` on the side and refer to them by (immutable) index
- Store a `BTreeMap<i32, VecDeque<StreamId>>` -- this is probably less efficient when many different priorities are used, but unless that's shown to be a bottleneck, it's pleasantly straightforward

Pixelstormer:
When implementing this originally, I was fond of the idea of using the `BinaryHeap`'s sorting for both priority and fairness, as well as reducing the number of collections that needed to be allocated, but I'll agree that it hasn't reduced complexity as much as I had hoped.

> - Refactor the original design to store the VecDeques in a Slab on the side and refer to them by (immutable) index

This would increase the number of layers of indirection to 3, needing to go through `BinaryHeap` -> `Slab` -> `VecDeque` to actually retrieve the `StreamId`. Of course, it's possible that this wouldn't actually impact performance that much, but regardless, it would still require a fair bit of complexity and bookkeeping.

> - Store a BTreeMap<i32, VecDeque<StreamId>> -- this is probably less efficient when many different priorities are used, but unless that's shown to be a bottleneck, it's pleasantly straightforward

In the common case of only having a few or even just 1 priority in use, this only has 2 layers of indirection, although this would increase with more priorities as the `BTreeMap` allocates more heap nodes. Again, and as you mentioned, this may not actually be a bottleneck, and it would also be a much simpler, and thus more maintainable, implementation.

I think that the latter alternative is better, as making the code simpler and more understandable would be very valuable, and it's not clear that it would even have a meaningful impact on performance - I know that I had some trouble understanding the original implementation because of the tricky bookkeeping for always keeping at least 1 empty `VecDeque` around.

Ralith:
Yeah, I'm not happy with the existing code either. Proceeding with the `BTreeMap` approach sounds good to me; we can always revisit again in the future, and in the mean time it'll be simpler/more readable/not incur a silly `!Sync`.

Pixelstormer:
So, I've decided to keep the current approach of removing the `VecDeque`s and implementing fairness by including a recency value in the key, as the main source of complexity was really just all the bookkeeping needed to work around how limited `BinaryHeap`'s API is. By comparison, `BTreeMap` crucially has a `range` method that does everything we want, with no need for any external bookkeeping. There's also a nightly method, `lower_bound`, that would make it even more terse if it ever gets stabilized.

I think this time round, it actually is much simpler than before, as this removes all of the aforementioned complexity working around `BinaryHeap`'s limited API, which both the previous implementations ran into, while still removing all of the bookkeeping to do with managing the `VecDeque`s. I'll admit that personally, I'm particularly averse to (the bookkeeping introduced by) the optimization of keeping 1 empty `VecDeque` around when there are no pending streams, as I think that was the hardest part of the original code to understand. Thus, I'm definitely biased towards wanting to remove the `VecDeque`s, to eliminate that unintuitive logic and related bookkeeping.

> the mechanism described here is more complex than rotating through a VecDeque

I don't think the idea of the recency value is actually any more complex that rotating a `VecDeque`, as fundamentally, they're both just doing the same thing - implementing round-robin scheduling. I'm hoping that this similarity in simplicity was just being hidden by the surrounding bookkeeping, and that is it now more apparent.

Pixelstormer:
So, CI is failing because of the use of the `pop_last` method, which was stabilised in 1.66, while Quinn's current MSRV is 1.65. I don't know what Quinn's MSRV policy is, but given that it's only a different of a single version, there is an argument for bumping the MSRV for this, as the [original motivation](https://github.com/rust-lang/rust/issues/62924#issue-472054719) for adding `pop_last` and other related functions was that they are meaningfully faster than the old way of calling `.iter().next()`/`.iter().next_back()`.

djc:
> So, CI is failing because of the use of the `pop_last` method, which was stabilised in 1.66, while Quinn's current MSRV is 1.65. I don't know what Quinn's MSRV policy is, but given that it's only a different of a single version, there is an argument for bumping the MSRV for this, as the [original motivation](https://github.com/rust-lang/rust/issues/62924#issue-472054719) for adding `pop_last` and other related functions was that they are meaningfully faster than the old way of calling `.iter().next()`/`.iter().next_back()`.

I think bumping to 1.66 is probably fine at this point -- please either squash your other commits and add a separate commit for this or do this in a separate PR.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -335,60 +335,43 @@ impl<'a> SendStream<'a> {
...
+        Bound::Included(PendingPriority {
+            priority,
+            recency: u64::MAX,
+        }),
+    ));
+
+    pending.insert(
+        PendingPriority {
+            priority,
+            recency: range.next().map(|(p, _)| p.recency - 1).unwrap_or(u64::MAX),
```
djc, line 357: Could `p.recency - 1` wrap? If not, why not? Maybe comment that this is unlikely to wrap because `recency` gets initialized to `u64::MAX`?

djc (commented):
I think this generally looks good. It would be nice to squash this into 3 commits:

* MSRV bump
* Changes for the new priority management
* Other changes that add `Send + Sync` bounds

Ralith:
I went ahead and opened https://github.com/quinn-rs/quinn/pull/1810 for the MSRV bump, since I was excited to get rid of some janky old `BTreeMap` iterator use.

Pixelstormer:
Rebased ontop of #1810

djc (approved):
It doesn't look like this was rebased correctly? The commits from #1810 should not longer appear in this PR since they have been merged into main already, and there's a conflict left.

Otherwise, your new commits look good to me, thanks!

Pixelstormer:
It looks like #1810 was updated inbetween me rebasing onto it and it getting merged into main. Should be fixed now.

djc:
Great -- will leave it to @Ralith to have another look before merging, thanks!

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -335,60 +335,43 @@ impl<'a> SendStream<'a> {
...
+        }),
+    ));
+
+    // Determine the recency value for this stream
+    // Setting it to 1 below all preexisting streams of this priority causes it to be sorted after all of those streams in
+    // the `BTreeMap`. This implements round-robin scheduling for streams that are still pending even after being handled,
+    // as in that case they are removed and then immediately reinserted.
+    // If this is the only/first stream for this priority, recency is initialised to u64::MAX instead. As this function is
+    // the only place that recency is initialised or mutated, this ensures that it will practically never underflow.
+    let recency = range.next().map(|(p, _)| p.recency - 1).unwrap_or(u64::MAX);
```
Ralith, line 358: Could we simplify this by using a single `Connection`-scoped recency value that we decrement on each insert? `u64` is big enough that this still won't ever overflow, and it'd save a traversal.

Also might as well use wrapping arithmetic to update it, since a rollover would be harmless even if it did somehow happen.

Ralith (changes requested):
Thanks, this is a great simplification!

quinn-proto/src/connection/streams/mod.rs
Pixelstormer: > Could we simplify this by using a single Connection-scoped recency value that we decrement on each insert? u64 is big enough that this still won't ever overflow, and it'd save a traversal.

This is an excellent suggestion, and actually lets us switch back to using `BinaryHeap`, as needing to do lookups on the collection was the whole motivation for replacing it with a `BTreeMap`, even though `BinaryHeap` is a better fit for the other ways the collection is used.

> Also might as well use wrapping arithmetic to update it, since a rollover would be harmless even if it did somehow happen.

An underflow would break fairness, as the recency of the stream that caused the underflow would wrap around to `u64::MAX`, while all other streams of that priority would still have recency values close to 0, so only that one stream would be handled until its recency gets close to 0 again.

Ralith: Oh, right, we might never touch the pre-underflow streams again. Good thing `u64` is big enough that this won't happen in practice.

Ralith (approved):
Thanks, I'm really happy with how this has turned out! Simultaneously cleaner, more efficient, and presenting a better API 🎉

PR #1774, opened 2024-03-06
Fix constant of frame_type IMMEDIATE_ACK

nemethf:
Both the draft and the IANA registry say the current value for frame_type IMMEDIATE_ACK is 0x1F.

https://www.iana.org/assignments/quic/quic.xhtml
https://datatracker.ietf.org/doc/html/draft-ietf-quic-ack-frequency#section-10.2

### DIFF
--- a/quinn-proto/src/frame.rs
+++ b/quinn-proto/src/frame.rs
@@ -131,7 +131,7 @@ frame_types! {
     HANDSHAKE_DONE = 0x1e,
     // ACK Frequency
     ACK_FREQUENCY = 0xaf,
-    IMMEDIATE_ACK = 0xac,
+    IMMEDIATE_ACK = 0x1f,
     // DATAGRAM
 }
 


djc (approved):
Oops, thanks! Do you want/need a maintenance release for this?

djc:
Who is we? What's your interest/use case?

Ralith (approved):
Looks like this changed in draft 7. We'll need to keep an eye out for future revisions!

levaitamas:
> Who is we? What's your interest/use case?

I guess @nemethf was referring to me. :smile:  We are experimenting with real-time media traffic, and it turns out that both [gstreamer quicsink/quicsrc](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/merge_requests/1036) and [moq-rs](https://github.com/kixelated/moq-rs) uses `quinn`.

PR #1776, opened 2024-03-07
Use the correct MTU for prev_path

nemethf:
I've just bumped into this.  I'm not confident this is really a bug, but using the MTU of a different path seems a wrong thing to do.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -480,7 +480,7 @@ impl Connection {
                     SpaceId::Data,
                     "PATH_CHALLENGE queued without 1-RTT keys"
                 );
-                buf.reserve(self.path.current_mtu() as usize);
+                buf.reserve(MIN_INITIAL_SIZE as usize);
 
                 let buf_capacity = buf.capacity();
 


djc (approved):
This does seem to make more sense, thanks for pointing it out!

quinn-proto/src/connection/mod.rs
```diff
@@ -483,7 +483,7 @@ impl Connection {
                     SpaceId::Data,
                     "PATH_CHALLENGE queued without 1-RTT keys"
                 );
-                buf.reserve(self.path.current_mtu() as usize);
+                buf.reserve(prev_path.current_mtu() as usize);
```
Ralith, line 486: We're actually always sending a fixed-size packet here; see the `pad_to` call below, and the absence of any frames other than the `PATH_CHALLENGE`. Maybe `reserve(MIN_INITIAL_SIZE)` would be clearer still?

djc:
@nemethf did you want to follow up on this?

nemethf:
> @nemethf did you want to follow up on this?

Should I  update the PR with `reserve(MIN_INITIAL_SIZE)`?  (I'm okay with @Ralith or you do it.)

djc:
> > @nemethf did you want to follow up on this?
> 
> Should I update the PR with `reserve(MIN_INITIAL_SIZE)`? (I'm okay with @Ralith or you do it.)

Would be nice!

nemethf:
> > > @nemethf did you want to follow up on this?

I force-pushed a version suggested by @Ralith.   Thanks.

Ralith (approved):
Thanks!

PR #1777, opened 2024-03-10
Maintain in-flight counters separately per path

### LINKED ISSUE #1775: PATH_CHALLENGE is never sent
My client host has two network interfaces.  I bind the client to 0.0.0.0:0 and bring down the primary interface with `ip link set dev eth0 down`  during a download.  This does not result in a connection migration.

Firstly, because the client has nothing to send, so the server won't be notified about the new path.  So I change the client-config to make it send pings periodically:

```rust
    let mut transport_config = TransportConfig::default();
    transport_config
        .keep_alive_interval(Some(Duration::from_secs(1)));
    client_config.transport_config(Arc::new(transport_config));
```

Now when the server notices the new path, it creates two path challenges in Connection::migrate(), one for the prev_path, one for the new_path.  However, it never sends the PATH_CHALLENGE for new_path.  I'm guessing this is because `poll_transmit()` always has something better to send.  If I modify poll_transmit to prioritize both PATH_CHALLENGEs, then the connection migration works as expected. 

I copy my patch below.  But most probably this is not the right fix, because probably it is not a good idea to send a PATH_CHALLENGE frame alone in a packet and before other more important frames.

```diff
diff --git a/quinn-proto/src/connection/mod.rs b/quinn-proto/src/connection/mod.rs
index 03a6050..25b2631 100644
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -519,6 +519,54 @@ impl Connection {
                 });
             }
         }
+        // Send PATH_CHALLENGE for a current path if necessary
+        if self.path.challenge_pending {
+            self.path.challenge_pending = false;
+            let token = self.path
+                .challenge
+                .expect("path challenge pending without token");
+            let destination = self.path.remote;
+            debug_assert_eq!(
+                self.highest_space,
+                SpaceId::Data,
+                "PATH_CHALLENGE queued without 1-RTT keys"
+            );
+            buf.reserve(self.path.current_mtu() as usize);
+
+            let buf_capacity = buf.capacity();
+
+            let mut builder = PacketBuilder::new(
+                now,
+                SpaceId::Data,
+                buf,
+                buf_capacity,
+                0,
+                false,
+                self,
+                self.version,
+            )?;
+            trace!("validating path with PATH_CHALLENGE {:08x}", token);
+            buf.write(frame::Type::PATH_CHALLENGE);
+            buf.write(token);
+            self.stats.frame_tx.path_challenge += 1;
+
+            // An endpoint MUST expand datagrams that contain a
+            // PATH_CHALLENGE frame to at least the smallest allowed
+            // maximum datagram size of 1200 bytes, unless the
+            // anti-amplification limit for the path does not permit
+            // sending a datagram of this size
+            builder.pad_to(MIN_INITIAL_SIZE);
+
+            builder.finish(self, buf);
+            self.stats.udp_tx.on_sent(1, buf.len());
+            return Some(Transmit {
+                destination,
+                size: buf.len(),
+                ecn: None,
+                segment_size: None,
+                src_ip: self.local_ip,
+            });
+        }
 
         // If we need to send a probe, make sure we have something to send.
         for space in SpaceId::iter() {
```

Ralith:
Fixes #1775, theoretically. @nemethf can you confirm?

RFC9000 §9.4:

> The capacity available on the new path might not be the same as the old path. Packets sent on the old path MUST NOT contribute to congestion control or RTT estimation for the new path.

We accomplish this by tracking the first packet number on each path, which allows the logic that removes acked/lost packets from congestion control counters to determine which, if any, known path a packet was sent on.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1430,7 +1430,7 @@ impl Connection {
     // Not timing-aware, so it's safe to call this for inferred acks, such as arise from
     // high-latency handshakes
     fn on_packet_acked(&mut self, now: Instant, info: SentPacket) {
-        self.remove_in_flight(&info);
+        self.in_flight.remove(&info);
         if info.ack_eliciting && self.path.challenge.is_none() {
             // Only pass ACKs to the congestion controller if we are not validating the current
             // path, so as to ignore any ACKs from older paths still coming in.
@@ -1595,7 +1595,7 @@ impl Connection {
 
             for &packet in &lost_packets {
                 let info = self.spaces[pn_space].take(packet).unwrap(); // safe: lost_packets is populated just above
-                self.remove_in_flight(&info);
+                self.in_flight.remove(&info);
                 for frame in info.stream_frames {
                     self.streams.retransmit(frame);
                 }
@@ -1624,7 +1624,7 @@ impl Connection {
         // Handle a lost MTU probe
         if let Some(packet) = lost_mtu_probe {
             let info = self.spaces[SpaceId::Data].take(packet).unwrap(); // safe: lost_mtu_probe is omitted from lost_packets, and therefore must not have been removed yet
-            self.remove_in_flight(&info);
+            self.in_flight.remove(&info);
             self.path.mtud.on_probe_lost();
             self.stats.path.lost_plpmtud_probes += 1;
         }
@@ -2023,7 +2023,7 @@ impl Connection {
         space.in_flight = 0;
         let sent_packets = mem::take(&mut space.sent_packets);
         for (_, packet) in sent_packets.into_iter() {
-            self.remove_in_flight(&packet);
+            self.in_flight.remove(&packet);
         }
         self.set_loss_detection_timer(now)
     }
@@ -2321,7 +2321,7 @@ impl Connection {
                 // Retransmit all 0-RTT data
                 let zero_rtt = mem::take(&mut self.spaces[SpaceId::Data].sent_packets);
                 for (_, info) in zero_rtt {
-                    self.remove_in_flight(&info);
+                    self.in_flight.remove(&info);
                     self.spaces[SpaceId::Data].pending |= info.retransmits;
                 }
                 self.streams.retransmit_all_for_0rtt();
@@ -2384,7 +2384,7 @@ impl Connection {
                             let sent_packets =
                                 mem::take(&mut self.spaces[SpaceId::Data].sent_packets);
                             for (_, packet) in sent_packets {
-                                self.remove_in_flight(&packet);
+                                self.in_flight.remove(&packet);
                             }
                         } else {
                             self.accepted_0rtt = true;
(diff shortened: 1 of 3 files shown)


nemethf:
> Fixes #1775, theoretically. @nemethf can you confirm?

Yes, it does solve my original issue and the traffic quickly migrates to the new path.  Thank you.

---

Looking at the captured traffic, I was a bit surprised:
```
 No. -     Time -             Source -           Dest -             Proto -        Length -         Info -
 278       2.897193           10.0.2.1           10.0.3.1           QUIC           78               Protected Payload (KP0), DCID=7abe9d2420aaddb1, PKN: 101, PING, PADDING
 279       2.907467           10.0.3.1           10.0.1.1           QUIC           1248             Protected Payload (KP0), DCID=033dda038eeebb0b, PKN: 219, PC, PADDING
 280       2.907472           10.0.3.1           10.0.2.1           HTTP3          1248             Protected Payload (KP0), DCID=033dda038eeebb0b, PKN: 220, ACK_ECN, PC, RC, STREAM(3)
```
In other words the server sends payload alongside with the PATH_CHALLENGE. I did not know this was possible, but RFC9000 permits it: " An endpoint MAY send data to an unvalidated peer address, but it MUST protect against potential attacks as described in Sections 9.3.1 and 9.3.2."   So it's all good.  Thanks again.

Ralith:
Yep, that's the intended behavior. Separate logic applies an anti-amplification budget to all unvalidated paths to prevent this from being too much of an amplification hazard if someone attempts the rather esoteric off-path packet forwarding attack. Thanks for confirming!

quinn-proto/src/connection/packet_builder.rs
```diff
@@ -203,8 +203,11 @@ impl PacketBuilder {
             stream_frames: sent.stream_frames,
         };
 
-        conn.in_flight.insert(&packet);
-        conn.in_flight.bytes -= conn.spaces[space_id].sent(exact_number, packet);
+        conn.path.in_flight.insert(&packet);
+        if conn.path.first_packet.is_none() {
+            conn.path.first_packet = Some(exact_number);
+        }
+        conn.path.in_flight.bytes -= conn.spaces[space_id].sent(exact_number, packet);
```
djc, line 210: Should this be in a method on `PathData`?

quinn-proto/src/connection/mod.rs
```diff
@@ -3437,6 +3434,17 @@ impl Connection {
...
     }
 
+    /// Update counters to account for a packet becoming acknowledged, lost, or abandoned
+    fn remove_in_flight(&mut self, pn: u64, packet: &SentPacket) {
+        // Visit known paths from newest to oldest to find the one `pn` was sent on
+        for path in [&mut self.path].into_iter().chain(self.prev_path.as_mut()) {
+            if path.first_packet.map_or(false, |first| first <= pn) {
+                path.in_flight.remove(packet);
+                return;
+            }
```
djc, line 3444: Should this also be a method on `PathData`?

Feels like `first_packet` could/should remain private.

djc (approved):
Nice!

quinn-proto/src/connection/mod.rs
Ralith: Yep, that works out nicely and also lets us privatize more of `InFlight`.

PR #1781, opened 2024-03-16
Add Connection::is_closed method

### LINKED ISSUE #1780: Add Connection::is_closed method
## Problem

Multiple threads are using the same connection when the connection dies. All the threads race to restore the connection, potentially, reconnecting multiple times.

## Solution

Put the connection behind a RwLock, acquire the writer lock, check if the connection is closed via `is_closed` method, if it is, reconnect. This way the reconnection happens at most once.

jean-airoldie:
Closes #1780

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -582,6 +582,15 @@ impl Connection {
         // May need to send MAX_STREAMS to make progress
         conn.wake();
     }
+
+    /// Whether the connection is closed
+    ///
+    /// Closed connections cannot transport any further data. A connection becomes closed when
+    /// either peer application intentionally closes it, or when either transport layer detects an
+    /// error such as a time-out or certificate validation failure.
+    pub fn is_closed(&self) -> bool {
+        self.0.state.lock("is_closed").inner.is_closed()
+    }
 }
 
 pin_project! {


Ralith:
Thanks for the PR! This information is already available via [`Connection::close_reason`](https://docs.rs/quinn/latest/quinn/struct.Connection.html#method.close_reason).

zphrs:
Hello! I just spent the past 10 minutes or so tracking down the behavior of close_reason() and I would have benefited from having an `is_closed()` helper function (even if internally the helper is merely `close_reason().is_some()`). I only am leaving a comment here because I had to go digging through the source code in order to verify that `close_reason()` in fact is set to `Some` even if the Connection is not closed with the `close()` method and is instead closed by Quinn, such as in the case of an idle timeout.

I additionally would appreciate adding more to the documentation for `close_reason()` to clarify that `close_reason()` is always set when a socket is closed. Currently the documentation merely says:

> If the connection is closed, the reason why. Returns None if the connection is still open. 

This does not make it clear that the reason is always set when a socket is closed. For that assurance I had to look into the source code where I found [src/connection.rs#L982-L983](https://github.com/quinn-rs/quinn/blob/b2b930a0662b18b2e351264a21e175478bb3c3f1/quinn/src/connection.rs#L982-L983):
```rust
    /// Always set to Some before the connection becomes drained
    pub(crate) error: Option<ConnectionError>,
```
Is it possible you'd reconsider adding this convenience function and add to the documentation that close_reason is always set to Some when the connection is closed?

djc:
I don't think we want to add more API, but happy to review a PR that extends/clarifies the documentation!

PR #1783, opened 2024-03-16
Expose SendStream::poll_write and poll_finish methods

### LINKED ISSUE #1782: Expose poll_write & poll_close
Since `SendStream` methods no longer return a named future, the only current way of doing polling is via `futures_io::AsyncWrite` or `tokio::io::AsyncWrite`. I would like to be able to call `poll_write` and `poll_close` without any of these libraries features and without needing to use a `std::io::Error` which can be undesirable.

jean-airoldie:
Closes #1782

### DIFF
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -142,12 +142,6 @@ impl SendStream {
     ///
     /// No new data may be written after calling this method. Completes when the peer has
     /// acknowledged all sent data, retransmitting data as needed.
-    ///
-    /// On success, return Poll::Ready(Ok(())).
-    ///
-    /// If the stream is not ready for writing, the method returns Poll::Pending and arranges
-    /// for the current task (via cx.waker().wake_by_ref()) to receive a notification when the
-    /// stream becomes writable or is closed.
     pub fn poll_finish(&mut self, cx: &mut Context) -> Poll<Result<(), WriteError>> {
         let mut conn = self.conn.state.lock("poll_finish");
         if self.is_0rtt {
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -138,7 +138,16 @@ impl SendStream {
         Finish { stream: self }.await
     }
 
-    #[doc(hidden)]
+    /// Attempt to shut down the send stream gracefully.
+    ///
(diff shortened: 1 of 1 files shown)


quinn/src/send_stream.rs
```diff
@@ -138,7 +138,16 @@ impl SendStream {
...
+    /// Attempt to shut down the send stream gracefully.
+    ///
+    /// No new data may be written after calling this method. Completes when the peer has
+    /// acknowledged all sent data, retransmitting data as needed.
+    ///
+    /// On success, return Poll::Ready(Ok(())).
+    ///
+    /// If the stream is not ready for writing, the method returns Poll::Pending and arranges
+    /// for the current task (via cx.waker().wake_by_ref()) to receive a notification when the
+    /// stream becomes writable or is closed.
```
Ralith, line 150: This paragraph is confusing. Perhaps it was copy-pasted in error from `poll_write`? I think the first two paragraphs are sufficient documentation.

Ralith (changes requested):
I'm open to re-exposing named futures. That would provide access to the full breadth of write variations, rather than just the `&[u8]` case. On the other hand, it's a very good point that the various `AsyncWrite` impls commit us to supporting this pattern anyway, so we might as well expose it directly.

quinn/src/send_stream.rs
jean-airoldie: Indeed it was.

jean-airoldie:
Personally I prefer having access to the polling functions directly because it makes multi-stage futures much simpler to implement.

Ralith (approved):

Ralith:
Figured; been there myself. That can break down if you want multiple futures to concurrently manipulate the same value, but writes require `&mut SendStream` anyway, so that isn't a concern here.

jean-airoldie:
Cool. I'll do another one for the read polling, if you don't mind.

Ralith:
SGTM. Best to be consistent, after all.

ibigbug:
is this removed again in the latest release? this one is relying on it https://github.com/hyperium/h3/blob/master/h3-quinn/src/lib.rs#L600

@djc @Ralith

djc:
https://github.com/hyperium/h3/pull/238 already ports h3-quinn to quinn 0.11.

PR #1784, opened 2024-03-17
Expose RecvStream::poll_read & poll_read_chunk

### LINKED ISSUE #1782: Expose poll_write & poll_close

jean-airoldie:
Closes #1782

### DIFF
--- a/quinn/src/recv_stream.rs
+++ b/quinn/src/recv_stream.rs
@@ -141,7 +141,25 @@ impl RecvStream {
         .await
     }
 
-    fn poll_read(
+    /// Attempts to read from the stream into buf.
+    ///
+    /// On success, returns Poll::Ready(Ok(num_bytes_read)) and places data in
+    /// the buf. If no data was read, it implies that EOF has been reached.
+    ///
+    /// If no data is available for reading, the method returns Poll::Pending
+    /// and arranges for the current task (via cx.waker()) to receive a notification
+    /// when the stream becomes readable or is closed.
+    pub fn poll_read(
+        &mut self,
+        cx: &mut Context,
+        buf: &mut [u8],
+    ) -> Poll<Result<usize, ReadError>> {
+        let mut buf = ReadBuf::new(buf);
+        ready!(self.poll_read_buf(cx, &mut buf))?;
+        Poll::Ready(Ok(buf.filled().len()))
+    }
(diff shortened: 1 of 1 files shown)


jean-airoldie:
Now for this one, the `poll_read_chunk` is not part of `AsyncRead`, so technically this would force you to support that new method, but I figured that if it might result in improved read performance, that might be desirable to expose too.

jean-airoldie:
In my use case, I'm thinking of using `poll_read` to read the size prefix, followed by `poll_read_chunk` to read the actual message, for instance.

quinn/src/recv_stream.rs
```diff
@@ -196,8 +214,15 @@ impl RecvStream {
         .await
     }
 
-    /// Foundation of [`Self::read_chunk`]
-    fn poll_read_chunk(
+    /// Attempts to read a chunk from the stream into buf.
```
Ralith, line 217: There is no `buf` in this context.

Ralith (changes requested):
> I figured that if it might result in improved read performance, that might be desirable to expose too.

This is speculative. I'd prefer to avoid further complicating our API without a concrete use case that benefits, particularly as using `poll_read_buf` correctly is itself complex. For example, your use case would need to handle consuming any number of chunks before reaching the desired total length, and you'd then have to either deal with having non-contiguous data in whatever downstream processing you do, or copy everything into a single buffer, at best duplicating the logic already provided by `poll_read`.

jean-airoldie:
> This is speculative. I'd prefer to avoid further complicating our API without a concrete use case that benefits, particularly as using poll_read_buf correctly is itself complex.

Fair enough. In my understanding, the potential benefit was from the unordered read, even if you have to copy the chunks into a separate buffer.

Ralith (approved):

djc (approved):

djc:
(Needs a rebase. Would be nice if you can squash the commits here.)

djc (approved):

PR #1788, opened 2024-03-23
Enable `ios` for the same os_type checks as `macos`

jamilbk:
It appears that IPv4-only is [broken on iOS](https://github.com/firezone/firezone/issues/4276).

Enabling the same code paths for ios that exist for macos fixes all issues. In our testing, we've found the network stacks to be nearly identical between iOS and macOS.

This fork has been tested on 3 physical iOS devices with IPv4-only, IPv6-only, and dual stack configurations on iOS 17.

Thanks for the great library. Happy to do any further testing!

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -683,10 +683,8 @@ fn decode_recv(
                 // Temporary hack around broken macos ABI. Remove once upstream fixes it.
                 // https://bugreport.apple.com/web/?problemID=48761855
                 #[allow(clippy::unnecessary_cast)] // cmsg.cmsg_len defined as size_t
-                if cfg!(target_os = "macos")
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Thanks! Looks like this causes some issues in CI?

jamilbk:
Hmm, is the CI expected to fail for this? I would not have expected this change to affect `macos` tests. I'll take a closer look.

jamilbk:
https://github.com/quinn-rs/quinn/pull/1788/commits/248ac33ff79ba3173e794aa63cc814d70f4714b3 fixes the issue. Sorry about that! Glad CI caught it. Please give it a once-over.

Thanks for accepting!

Ralith (approved):
Thanks!

PR #1789, opened 2024-03-23
Perform early first-packet validation before decryption

Ralith:


lijunwangs (approved):
Looks good

quinn-proto/src/endpoint.rs
```diff
@@ -415,11 +409,13 @@ impl Endpoint {
         version: u32,
         addresses: FourTuple,
         crypto: &Keys,
-        src_cid: &ConnectionId,
-        dst_cid: &ConnectionId,
+        first_decode: &PartialDecode,
```
djc, line 412: I suppose this change could be made in the initial commit?

quinn-proto/src/endpoint.rs
```diff
@@ -396,6 +396,50 @@ impl Endpoint {
         }
     }
 
+    /// Check if we should refuse a connection attempt regardless of the packet's contents
+    fn early_validate_first_packet(
```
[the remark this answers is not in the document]
Ralith: Good call, that's much better.

quinn-proto/src/endpoint.rs
```diff
@@ -231,6 +231,19 @@ impl Endpoint {
                     return None;
                 }
             };
+
+            if let Some(response) = self.early_validate_first_packet(
+                version,
+                addresses,
+                &crypto,
+                // Necessarily `Some(_)` when `first_decode.initial_version()` is `Some(_)`
```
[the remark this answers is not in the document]
Ralith: Yep, that helped.

quinn-proto/src/endpoint.rs
Ralith: Mooted by other refactoring.

quinn-proto/src/packet.rs
```diff
@@ -612,6 +610,15 @@ impl PlainHeader {
     }
 }
 
+#[derive(Clone, Debug)]
+pub(crate) struct PlainHeaderInitial {
```
djc, line 614: Naming nit: maybe `InitialHeader`? I don't think `Plain` provides sufficient disambiguation value here, and would be nice to have a more natural word order.

quinn-proto/src/endpoint.rs
```diff
@@ -232,6 +232,18 @@ impl Endpoint {
                         return None;
                     }
                 };
+
+            if let Err(reason) = self.early_validate_first_packet(dst_cid) {
+                return Some(DatagramEvent::Response(self.initial_close(
```
djc, line 237: Should/could we have `initial_close()` take `&InitialHeader` instead of separate `version` and `src_cid`?

quinn-proto/src/endpoint.rs
```diff
@@ -575,12 +575,17 @@ impl Endpoint {
             return Err(TransportError::CONNECTION_REFUSED(""));
         }
 
-        if dst_cid.len() < 8
-            && (!server_config.use_retry || dst_cid.len() != self.local_cid_generator.cid_len())
+        // RFC9000 §7.2 dictates that initial (client-chosen) destination CIDs must be at least 8
+        // bytes. If this might be a Retry packet, then the length may instead match our usual CID
```
djc, line 579: "If this might be" sounds a bit strange to me? Maybe just if "If this is" or "In case this is"?

djc (approved):
LGTM!

quinn-proto/src/packet.rs
Ralith: This is the pre-decryption view of the header; see also `enum PlainHeader` vs. `enum Header`. I'll go with `PlainInitialHeader`.

quinn-proto/src/endpoint.rs
Ralith: We sometimes call `initial_close` after decrypting the packet, when a plain header is no longer convenient.

PR #1790, opened 2024-03-23
Accept connections with unrecognized address validation tokens

Ralith:
If we share an identity with a different QUIC implementation, e.g. across an update or through a load balancer, and retries are enabled, we might receive address validation tokens generated by it and fail to decode them. Restarting address validation with a fresh Retry packet allows the connection attempt to proceed.

This involves refactoring of the Retry-mediated address validation cryptosystem, for which particularly careful review is required. In particular, we ditch the random plaintext in favor of directly encrypting the client address, and instead use the (server-chosen) retry source CID to derive a unique token key.

The circumstances where this matters are niche, so no particular rush to get this in before 0.11.


quinn-proto/src/token.rs
```diff
@@ -67,20 +67,30 @@ impl RetryToken {
...
         })
     }
+}
 
-    fn put_additional_data<'b>(address: &SocketAddr, additional_data: &'b mut [u8]) -> &'b [u8] {
-        let mut cursor = &mut *additional_data;
-        match address.ip() {
-            IpAddr::V4(x) => cursor.put_slice(&x.octets()),
-            IpAddr::V6(x) => cursor.put_slice(&x.octets()),
+fn encode_addr(buf: &mut Vec<u8>, address: &SocketAddr) {
```
djc, line 77: Should this be an `impl coding::Codec` for `SocketAddr` instead?

quinn-proto/src/endpoint.rs
```diff
@@ -466,8 +467,41 @@ impl Endpoint {
             )));
         }
 
-        let (retry_src_cid, orig_dst_cid) = if server_config.use_retry {
-            if token.is_empty() {
+        let (retry_src_cid, orig_dst_cid) = 'accept: {
```
djc, line 470: At about 7-8 levels of indentation, maybe it's time to split this logic out?

quinn-proto/src/token.rs
Ralith: I considered that, but opted not to because:
- there's only one callsite
- the type tag makes serialization fallible for reasons other than length, which can't be expressed in `Codec`'s existing API.

quinn-proto/src/endpoint.rs
Ralith: Split the `match` out, which simplified the control flow considerably. I didn't split the retry composition out because it needs a lot of locals.

gretchenfrage:
Hm. Please tell me if I'm missing context, but my understanding is that the spec forbids this.

[8.1.2. ](https://www.rfc-editor.org/rfc/rfc9000.html#section-8.1.2)[Address Validation Using Retry Packets](https://www.rfc-editor.org/rfc/rfc9000.html#name-address-validation-using-re)

> In response to processing an Initial packet containing a token that was provided in a Retry packet, a server cannot send another Retry packet; it can only refuse the connection or permit it to proceed.

> If a server receives a client Initial that contains an invalid Retry token but is otherwise valid, it knows the client will not accept another Retry token. The server can discard such a packet and allow the client to time out to detect handshake failure, but that could impose a significant latency penalty on the client. Instead, the server SHOULD immediately close ([Section 10.2](https://www.rfc-editor.org/rfc/rfc9000.html#immediate-close)) the connection with an INVALID_TOKEN error. Note that a server has not established any state for the connection at this point and so does not enter the closing period.

[17.2.5.1. ](https://www.rfc-editor.org/rfc/rfc9000.html#section-17.2.5.1)[Sending a Retry Packet](https://www.rfc-editor.org/rfc/rfc9000.html#name-sending-a-retry-packet)

> A client MUST accept and process at most one Retry packet for each connection attempt. After the client has received and processed an Initial or Retry packet from the server, it MUST discard any subsequent Retry packets that it receives.

Also relevant context:

[8.1.1. ](https://www.rfc-editor.org/rfc/rfc9000.html#section-8.1.1)[Token Construction](https://www.rfc-editor.org/rfc/rfc9000.html#name-token-construction)

> A token sent in a NEW_TOKEN frame or a Retry packet MUST be constructed in a way that allows the server to identify how it was provided to a client. These tokens are carried in the same field but require different handling from servers.

I need to look more into how exactly quinn is handling retries, but I think the appropriate way to facilitate clustering like this would be to make the important part of retry tokens be an encrypted and signed message containing the client's address, and allow the key which encrypts and signs that message to be configured so it can be set the same between different servers in a cluster, so that their retry tokens will be intercompatible

Ralith:
My incorrect word choice in the PR/commit message may have made this needlessly confusing, and has been revised.

It is true that we must never perform a repeated Retry. However, we may receive address validation tokens that the peer obtained from a previous connection's `NEW_TOKEN` frame, i.e. *not* a Retry token. If we fail to decode those for any reason, it is still legal to perform a Retry. See [8.1.3](https://www.rfc-editor.org/rfc/rfc9000.html#section-8.1.3-10):

> If the [address validation] token is invalid, then the server SHOULD proceed as if the client did not have a validated address, including potentially sending a Retry packet.

> Note: The rationale for treating the client as unvalidated rather than discarding the packet is that the client might have received the token in a previous connection using the NEW_TOKEN frame, and if the server has lost state, it might be unable to validate the token at all, leading to connection failure if the packet is discarded.

gretchenfrage:
Hmmmm. Ah, ok, I guess that, in the phrase "If a server receives a client Initial that contains an invalid Retry token but is otherwise valid", they mean to imply by the phrase "Retry token" that it is an address validation token which is valid enough to be unambiguously identified as having been minted as for a retry rather than a new_token frame but which is otherwise in some way invalid? That makes sense, although in my opinion the RFC's language is a bit confusing there. Sounds good though.

Ralith:
Yeah, this came up as an ambiguous case in the implementers' slack. Perhaps we'll get an errata for it.

quinn-proto/src/token.rs
```diff
@@ -93,6 +98,22 @@ fn decode_addr<B: Buf>(buf: &mut B) -> Option<SocketAddr> {
     Some(SocketAddr::new(ip, port))
 }
 
+/// Reasons why a retry token might fail to validate a client's address
+#[derive(Debug, Copy, Clone)]
+pub(crate) enum RetryError {
```
djc, line 103: We have a `RetryError` now in `proto::endpoint` too, we should rename either one of these to help disambiguate?

djc (approved):

quinn-proto/src/token.rs
djc: @Ralith ping, want to finish this up?

Ralith: Renamed this one `TokenDecodeError`. We can tolerate a bit more verbosity for an internal type, after all.

Ralith:
Rebased, and dropped the "factor out token decoding" commit since `handle_first_packet` is much cleaner after #1752.

Ralith:
MSRV failure is not related.

djc (approved):

djc:
Disabled the flag that requires all status checks to pass before merging for a bit.

PR #1791, opened 2024-03-25
Add missing transport parameter validation checks

Ralith:
As [required by the RFC](https://www.rfc-editor.org/rfc/rfc9000.html#section-18.2-8).

### DIFF
--- a/quinn-proto/src/transport_parameters.rs
+++ b/quinn-proto/src/transport_parameters.rs
@@ -433,18 +433,33 @@ impl TransportParameters {
         }
 
         // Semantic validation
+
+        // https://www.rfc-editor.org/rfc/rfc9000.html#section-18.2-4.26.1
         if params.ack_delay_exponent.0 > 20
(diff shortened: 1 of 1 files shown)


quinn-proto/src/transport_parameters.rs
```diff
@@ -444,7 +444,13 @@ impl TransportParameters {
                 min_ack_delay.0 > params.max_ack_delay.0 * 1_000
             })
             || (side.is_server()
-                && (params.stateless_reset_token.is_some() || params.preferred_address.is_some()))
+                && (params.original_dst_cid.is_some()
```
djc, line 447: Should we add a link here to the RFC section you linked in the PR description?

djc (approved):

quinn-proto/src/transport_parameters.rs
Ralith: Added citations for all checks.

PR #1796, opened 2024-03-30
Allow packets with impossible CIDs to be ignored rather than reset

Ralith:
Reduces the likelihood of Quinn endpoints responding to non-QUIC packets.

I'm not sure whether this is needed, especially after #1794. However, the cost seems low, it's a more direct solution, and defense in depth is nice.


quinn-proto/src/cid_generator.rs
```diff
@@ -86,3 +90,93 @@ impl ConnectionIdGenerator for RandomConnectionIdGenerator {
...
+}
+
+impl KeyedConnectionIdGenerator {
+    /// Create a generator with a random key
+    #[cfg(feature = "ring")]
+    pub fn new() -> Self {
+        let mut reset_key = [0; 64];
+        rand::thread_rng().fill_bytes(&mut reset_key);
+
+        let key = Arc::new(ring::hmac::Key::new(ring::hmac::HMAC_SHA256, &reset_key));
```
djc, line 108: Nit: elide the `key` binding?

quinn-proto/src/cid_generator.rs
```diff
@@ -86,3 +90,93 @@ impl ConnectionIdGenerator for RandomConnectionIdGenerator {
...
+/// [`validate`](ConnectionIdGenerator::validate)d
+pub struct KeyedConnectionIdGenerator {
+    key: Arc<dyn crypto::HmacKey>,
+    lifetime: Option<Duration>,
+}
+
+impl KeyedConnectionIdGenerator {
+    /// Create a generator with a random key
+    #[cfg(feature = "ring")]
+    pub fn new() -> Self {
```
djc, line 104: Maybe just skip `new()` and move the implementation into `Default`?

quinn-proto/src/cid_generator.rs
```diff
@@ -86,3 +90,93 @@ impl ConnectionIdGenerator for RandomConnectionIdGenerator {
...
+        self.key
+            .sign(nonce, &mut expected_signature[..self.key.signature_len()]);
+        (expected_signature[..SIGNATURE_LEN] == signature[..])
+            .then_some(())
+            .ok_or(InvalidCid)
+    }
+}
+
+const NONCE_LEN: usize = 3; // Good for more than 16 million connections
+const SIGNATURE_LEN: usize = 5; // 8-byte total CID length
```
djc, line 168: Maybe `8 - NONCE_LEN`?

quinn-proto/src/cid_generator.rs
```diff
@@ -20,8 +20,20 @@ pub trait ConnectionIdGenerator: Send {
     ///
     /// Connection IDs will be retired after the returned `Duration`, if any. Assumed to be constant.
     fn cid_lifetime(&self) -> Option<Duration>;
+
+    /// Quickly determine whether `cid` could have been generated by this generator
+    ///
+    /// False positives are permitted, but will reduce the efficiency with which invalid packets are
+    /// discarded.
+    fn validate(&self, _cid: &ConnectionId) -> Result<(), InvalidCid> {
```
djc, line 29: Nit: maybe move this before `cid_len()`?

quinn-proto/src/cid_generator.rs
```diff
@@ -86,3 +90,93 @@ impl ConnectionIdGenerator for RandomConnectionIdGenerator {
...
+    }
+
+    fn validate(&self, cid: &ConnectionId) -> Result<(), InvalidCid> {
+        let (nonce, signature) = cid.split_at(NONCE_LEN);
+        let mut expected_signature = [0; MAX_SIGNATURE_LEN];
+        self.key
+            .sign(nonce, &mut expected_signature[..self.key.signature_len()]);
+        (expected_signature[..SIGNATURE_LEN] == signature[..])
+            .then_some(())
+            .ok_or(InvalidCid)
```
djc, line 163: Style: don't love the `then_some(()).ok_or(InvalidCid)` chain. Explicit `match` at the cost of one more line?

```rust
match expected_signature[..SIGNATURE_LEN] == signature[..] {
     true => Ok(()),
     false => Err(InvalidCid),
}
```

quinn-proto/src/endpoint.rs
```diff
@@ -278,6 +268,17 @@ impl Endpoint {
         // If we got this far, we're a server receiving a seemingly valid packet for an unknown
         // connection. Send a stateless reset if possible.
         //
+
+        if !first_decode.is_initial()
```
djc, line 272: Is there a reason not to introduce it in this spot in the first place? This move seems somewhat independent of using the keyed generator by default.

djc (approved):
Nice!

djc:
I also wonder if the cost of crypto here actually makes it less attractive than the random generator?

Ralith:
That's a good question. A more thoughtful choice of hash function, rather than just using whatever *ring* has lying around, might help. Maybe even just `std`'s hasher, since we're not trying to be super secure here...

Ralith:
For clarity, extra cost is paid here in exactly two places:
- when generating a CID, which is gated behind the decision to respond to a connection attempt
- sending a stateless reset for an unrecognized packet

quinn-proto/src/cid_generator.rs
Ralith: I like how `new` is much more discoverable in rustdocs, though I recognize that we're not really consistent about having it.

Ralith:
Refactored to use a non-cryptographic hash.

djc (approved):

lijunwangs:
> I also wonder if the cost of crypto here actually makes it less attractive than the random generator? @lijunwangs do you think this would help your use case (on top of the other options we've merged recently)?

Thanks @djc/ @Ralith  -- this will be a great add to strengthen  the security. Is the change backward compatible? Can it break existing client code?

Ralith:
It should be backwards-compatible, yeah.

PR #1800, opened 2024-04-02
Break linkability on client-side after planned migration

nemethf:
The server switches to a new CID as well, but it sends a path-challenge on the old path with the new CID.  So it is still possible to trace back the connection to the old path after the client switches to a new socket.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -2909,6 +2909,12 @@ impl Connection {
         );
     }
 
+    /// Handle a change in the local address, i.e. an active migration
+    pub fn local_address_changed(&mut self) {
+        self.update_rem_cid();
+        self.ping();
+    }
+
     /// Switch to a previously unused remote connection ID, if possible
     fn update_rem_cid(&mut self) {
         let (reset_token, retired) = match self.rem_cids.next() {
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -96,8 +96,8 @@ enum ConnectionEvent {
         error_code: VarInt,
         reason: bytes::Bytes,
     },
+    LocalAddressChanged,
     Proto(proto::ConnectionEvent),
-    Ping,
 }
 
 #[derive(Debug)]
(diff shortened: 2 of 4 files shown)


quinn/src/lib.rs
```diff
@@ -96,6 +96,7 @@ enum ConnectionEvent {
     },
     Proto(proto::ConnectionEvent),
     Ping,
+    UpdateRemoteConnectionId,
```
Ralith, line 99: It looks like these two are only ever sent in sequence. Maybe we should have a single more abstract event, e.g. `Rebind` or `LocalAddressChanged`?

Ralith (commented):
Good catch, thanks! Please open a separate issue for the server-side bug if you don't plan to address it immediately yourself.

quinn/src/lib.rs
nemethf: I've merged the two into LocalAddressChanged, and force-pushed a new version of the PR.  The commit message  references a newly created issue about the server-side case.  Thanks.

quinn-proto/src/connection/mod.rs
```diff
@@ -2907,7 +2907,7 @@ impl Connection {
     }
 
     /// Switch to a previously unused remote connection ID, if possible
-    fn update_rem_cid(&mut self) {
+    pub fn update_rem_cid(&mut self) {
```
djc, line 2910: Do we still need this? Why?

djc (approved):

quinn-proto/src/connection/mod.rs
nemethf: Without it, I get:
```
error[E0624]: method `update_rem_cid` is private
    --> ~/quinn/quinn/src/connection.rs:919:32
     |
919  |                     self.inner.update_rem_cid();
     |                                ^^^^^^^^^^^^^^ private method
     |
    ::: ~/quinn/quinn-proto/src/connection/mod.rs:2910:5
     |
2910 |     fn update_rem_cid(&mut self) {
     |     ---------------------------- private method defined here

For more information about this error, try `rustc --explain E0624`.
error: could not compile `quinn` (lib) due to 1 previous error
```

Ralith:
Tweaked to encapsulate `update_rem_cid` inside a higher level, more semantic API at the proto layer.

PR #1801, opened 2024-04-02
Socket rebind: drain old socket

nemethf:
During a planned connection migration allow the client to receive
trafic via the old, abandoned socket until the first packet arrives on
the new socket.

---

I wrote this PR in a "cargo cult"-style without really knowing what I do. It seems to work as the client starts to acknowledge stream frames of the old path on the new path.  However, the server does not send stream frames right after it notices the new path even if  I modify the server to process acknowledgements when there's an outstanding path challenge [here](https://github.com/quinn-rs/quinn/blob/990203b4776c19a259e9b41fe6298ef0e17a2a08/quinn-proto/src/connection/mod.rs#L1431).  

Packet No. 1828, below, is the first packet arriving to the server on the new path.  The server sends path-challenges in 1891 (allowing linkability) and 1893.  The path validation for the new path arrives in 1910.   The server sends the first stream frame on the new path in 1913 with a repeated path challenge, so I assume it haven't processed the path response frame yet.

I'd like to see stream frames as early as packet 1893.  What should I do?  Thank you.

This is the server-side packet trace:
```
#pkt time     src      dst            len
1820 6.204678 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1234, STREAM(3)
1821 6.205635 10.0.1.1 10.0.3.1 QUIC   84 DCID=0b28501a55611291, PKN: 598, ACK_ECN
1822 6.207027 10.0.1.1 10.0.3.1 QUIC   84 DCID=0b28501a55611291, PKN: 599, ACK_ECN
1823 6.208060 10.0.1.1 10.0.3.1 QUIC   84 DCID=0b28501a55611291, PKN: 600, ACK_ECN
1824 6.209568 10.0.1.1 10.0.3.1 QUIC   84 DCID=0b28501a55611291, PKN: 601, ACK_ECN
1825 6.211314 10.0.1.1 10.0.3.1 QUIC   84 DCID=0b28501a55611291, PKN: 602, ACK_ECN
1826 6.213122 10.0.1.1 10.0.3.1 QUIC   84 DCID=0b28501a55611291, PKN: 603, ACK_ECN
1827 6.215704 10.0.1.1 10.0.3.1 QUIC   84 DCID=0b28501a55611291, PKN: 604, ACK_ECN
1828 6.221486 10.0.2.1 10.0.3.1 QUIC   78 DCID=b97c2e5111f44552, PKN: 605, PING, RC, PADDING
1829 6.229023 10.0.3.1 10.0.1.1 QUIC  834 DCID=65cc97c40b1817d1, PKN: 1235, STREAM(3)
1830 6.229031 10.0.3.1 10.0.1.1 QUIC 1500 DCID=65cc97c40b1817d1, PKN: 1236, STREAM(3)
1831 6.230494 10.0.3.1 10.0.1.1 QUIC  918 DCID=65cc97c40b1817d1, PKN: 1237, STREAM(3)
1832 6.232322 10.0.3.1 10.0.1.1 QUIC  834 DCID=65cc97c40b1817d1, PKN: 1238, STREAM(3)
1833 6.233964 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1239, STREAM(3)
1834 6.234850 10.0.3.1 10.0.1.1 QUIC  270 DCID=65cc97c40b1817d1, PKN: 1240, STREAM(3)
1835 6.235388 10.0.3.1 10.0.1.1 QUIC  270 DCID=65cc97c40b1817d1, PKN: 1241, STREAM(3)
1836 6.235915 10.0.3.1 10.0.1.1 QUIC  270 DCID=65cc97c40b1817d1, PKN: 1242, STREAM(3)
1837 6.236435 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1243, STREAM(3)
1838 6.237339 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1244, STREAM(3)
1839 6.237523 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 606, ACK_ECN
1840 6.238242 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1245, STREAM(3)
1841 6.239146 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1246, STREAM(3)
1842 6.240050 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1247, STREAM(3)
1843 6.240954 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1248, STREAM(3)
1844 6.241858 10.0.3.1 10.0.1.1 QUIC  646 DCID=65cc97c40b1817d1, PKN: 1249, STREAM(3)
1845 6.241887 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 607, ACK_ECN
1846 6.243158 10.0.3.1 10.0.1.1 QUIC  646 DCID=65cc97c40b1817d1, PKN: 1250, STREAM(3)
1847 6.243241 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 608, ACK_ECN
1848 6.244286 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 609, ACK_ECN
1849 6.244419 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1251, STREAM(3)
1850 6.245322 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1252, STREAM(3)
1851 6.245769 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 610, ACK_ECN
1852 6.246226 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1253, STREAM(3)
1853 6.247131 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1254, STREAM(3)
1854 6.247890 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 611, ACK_ECN
1855 6.248035 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1255, STREAM(3)
1856 6.248936 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 612, ACK_ECN
1857 6.248939 10.0.3.1 10.0.1.1 QUIC  270 DCID=65cc97c40b1817d1, PKN: 1256, STREAM(3)
1858 6.250358 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 613, ACK_ECN
1859 6.252547 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 614, ACK_ECN
1860 6.254742 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 615, ACK_ECN
1861 6.270770 10.0.3.1 10.0.1.1 QUIC 1221 DCID=65cc97c40b1817d1, PKN: 1257, ACK_ECN, STREAM(3)
1862 6.270803 10.0.3.1 10.0.1.1 QUIC 1500 DCID=65cc97c40b1817d1, PKN: 1258, STREAM(3)
1863 6.273014 10.0.3.1 10.0.1.1 QUIC 1294 DCID=65cc97c40b1817d1, PKN: 1259, STREAM(3)
1864 6.275579 10.0.3.1 10.0.1.1 QUIC 1210 DCID=65cc97c40b1817d1, PKN: 1260, STREAM(3)
1865 6.277988 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1261, STREAM(3)
1866 6.278874 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1262, STREAM(3)
1867 6.279219 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 616, ACK_ECN
1868 6.279789 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1263, STREAM(3)
1869 6.280699 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1264, STREAM(3)
1870 6.281603 10.0.3.1 10.0.1.1 QUIC  270 DCID=65cc97c40b1817d1, PKN: 1265, STREAM(3)
1871 6.282122 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1266, STREAM(3)
1872 6.282454 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 617, ACK_ECN
1873 6.283022 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1267, STREAM(3)
1874 6.283921 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1268, STREAM(3)
1875 6.284825 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1269, STREAM(3)
1876 6.284937 10.0.2.1 10.0.3.1 QUIC   85 DCID=b97c2e5111f44552, PKN: 618, ACK_ECN
1877 6.285729 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1270, STREAM(3)
1878 6.285981 10.0.2.1 10.0.3.1 QUIC   85 DCID=b97c2e5111f44552, PKN: 619, ACK_ECN
1879 6.286633 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1271, STREAM(3)
1880 6.287428 10.0.2.1 10.0.3.1 QUIC   85 DCID=b97c2e5111f44552, PKN: 620, ACK_ECN
1881 6.287538 10.0.3.1 10.0.1.1 QUIC  270 DCID=65cc97c40b1817d1, PKN: 1272, STREAM(3)
1882 6.288064 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1273, STREAM(3)
1883 6.288969 10.0.3.1 10.0.1.1 QUIC  646 DCID=65cc97c40b1817d1, PKN: 1274, STREAM(3)
1884 6.289233 10.0.2.1 10.0.3.1 QUIC   85 DCID=b97c2e5111f44552, PKN: 621, ACK_ECN
1885 6.290249 10.0.3.1 10.0.1.1 QUIC  270 DCID=65cc97c40b1817d1, PKN: 1275, STREAM(3)
1886 6.290777 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1276, STREAM(3)
1887 6.291027 10.0.2.1 10.0.3.1 QUIC   85 DCID=b97c2e5111f44552, PKN: 622, ACK_ECN
1888 6.291682 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1277, STREAM(3)
1889 6.292585 10.0.3.1 10.0.1.1 QUIC  458 DCID=65cc97c40b1817d1, PKN: 1278, STREAM(3)
1890 6.293295 10.0.2.1 10.0.3.1 QUIC   85 DCID=b97c2e5111f44552, PKN: 623, ACK_ECN
1891 6.293490 10.0.3.1 10.0.1.1 QUIC 1248 DCID=4d405bcceea196bd, PKN: 1279, PC, PADDING
1892 6.295375 10.0.2.1 10.0.3.1 QUIC   85 DCID=b97c2e5111f44552, PKN: 624, ACK_ECN
1893 6.295974 10.0.3.1 10.0.2.1 QUIC 1248 DCID=4d405bcceea196bd, PKN: 1280, PC, RC, PADDING
1894 6.297188 10.0.2.1 10.0.3.1 QUIC   85 DCID=b97c2e5111f44552, PKN: 625, ACK_ECN
1895 6.298468 10.0.3.1 10.0.2.1 QUIC 1374 DCID=4d405bcceea196bd, PKN: 1281, PING, PADDING
1896 6.299029 10.0.2.1 10.0.3.1 QUIC   85 DCID=b97c2e5111f44552, PKN: 626, ACK_ECN
1897 6.310328 10.0.2.1 10.0.3.1 QUIC   80 DCID=b97c2e5111f44552, PKN: 627, MSD(3)
1898 6.320945 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 628, ACK_ECN
1899 6.325686 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 629, ACK_ECN
1900 6.328957 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 630, ACK_ECN
1901 6.330771 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 631, ACK_ECN
1902 6.332201 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 632, ACK_ECN
1903 6.333996 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 633, ACK_ECN
1904 6.335832 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 634, ACK_ECN
1905 6.337617 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 635, ACK_ECN
1906 6.339041 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 636, ACK_ECN
1907 6.340860 10.0.2.1 10.0.3.1 QUIC   84 DCID=b97c2e5111f44552, PKN: 637, ACK_ECN
1908 6.342657 10.0.2.1 10.0.3.1 QUIC   85 DCID=b97c2e5111f44552, PKN: 638, ACK_ECN
1909 6.343609 10.0.2.1 10.0.3.1 QUIC 1248 DCID=b97c2e5111f44552, PKN: 639, PR, PADDING
1910 6.346113 10.0.2.1 10.0.3.1 QUIC 1248 DCID=b97c2e5111f44552, PKN: 640, ACK_ECN, PR, PADDING
1911 6.346132 10.0.2.1 10.0.3.1 QUIC  102 DCID=b97c2e5111f44552, PKN: 641, NCI
1912 6.348709 10.0.2.1 10.0.3.1 QUIC   85 DCID=b97c2e5111f44552, PKN: 642, ACK_ECN
1913 6.371125 10.0.3.1 10.0.2.1 QUIC 1248 DCID=4d405bcceea196bd, PKN: 1282, ACK_ECN, PC, NCI, STREAM(3)
1914 6.393867 10.0.3.1 10.0.2.1 QUIC 1248 DCID=4d405bcceea196bd, PKN: 1283, ACK_ECN, PC, STREAM(3)
```

### DIFF
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -437,7 +437,7 @@ impl State {
             now,
         );
         self.recv_state.recv_limiter.finish_cycle();
-        poll_res
+        poll_res.map(|x| x.keep_going)
     }
 
     fn drive_send(&mut self, cx: &mut Context) -> Result<bool, io::Error> {
@@ -764,7 +764,7 @@ impl RecvState {
         transmit_state: &mut TransmitState,
         socket: &dyn AsyncUdpSocket,
         now: Instant,
-    ) -> Result<bool, io::Error> {
+    ) -> Result<PollProgress, io::Error> {
         let mut metas = [RecvMeta::default(); BATCH_SIZE];
         let mut iovs: [IoSliceMut; BATCH_SIZE] = {
             let mut bufs = self
@@ -823,7 +823,7 @@ impl RecvState {
                     }
                 }
                 Poll::Pending => {
-                    break;
+                    return Ok(PollProgress { keep_going: false });
                 }
                 // Ignore ECONNRESET as it's undefined in QUIC and may be injected by an
                 // attacker
@@ -835,10 +835,14 @@ impl RecvState {
                 }
             }
             if !self.recv_limiter.allow_work() {
-                return Ok(true);
+                return Ok(PollProgress { keep_going: true });
             }
         }
-
-        Ok(false)
     }
+}
+
+#[derive(Default)]
+struct PollProgress {
+    /// Whether datagram handling was interrupted early by the work limiter for fairness
+    keep_going: bool,
 }
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -3,6 +3,7 @@ use std::{
     future::Future,
     io,
     io::IoSliceMut,
+    mem,
     net::{SocketAddr, SocketAddrV6},
     pin::Pin,
     str,
@@ -213,7 +214,7 @@ impl Endpoint {
         let addr = socket.local_addr()?;
         let socket = self.runtime.wrap_udp_socket(socket)?;
         let mut inner = self.inner.state.lock().unwrap();
-        inner.socket = socket;
+        inner.prev_socket = Some(mem::replace(&mut inner.socket, socket));
         inner.ipv6 = addr.is_ipv6();
 
         // Generate some activity so peers notice the rebind
@@ -407,6 +408,9 @@ impl EndpointInner {
 #[derive(Debug)]
 pub(crate) struct State {
     socket: Arc<dyn AsyncUdpSocket>,
+    /// During an active migration, abandoned_socket receives traffic
+    /// until the first packet arrives on the new socket.
+    prev_socket: Option<Arc<dyn AsyncUdpSocket>>,
     inner: proto::Endpoint,
     transmit_state: TransmitState,
     recv_state: RecvState,
@@ -429,6 +433,19 @@ pub(crate) struct Shared {
 impl State {
     fn drive_recv(&mut self, cx: &mut Context, now: Instant) -> Result<bool, io::Error> {
         self.recv_state.recv_limiter.start_cycle();
+        if let Some(socket) = &self.prev_socket {
+            // We don't care about the `PollProgress` from old sockets.
+            let poll_res = self.recv_state.poll_socket(
+                cx,
+                &mut self.inner,
+                &mut self.transmit_state,
+                &**socket,
+                now,
+            );
+            if poll_res.is_err() {
+                self.prev_socket = None;
+            }
+        };
         let poll_res = self.recv_state.poll_socket(
             cx,
             &mut self.inner,
@@ -437,7 +454,13 @@ impl State {
             now,
         );
         self.recv_state.recv_limiter.finish_cycle();
-        poll_res.map(|x| x.keep_going)
+        let poll_res = poll_res?;
+        if poll_res.received_connection_packet {
+            // Traffic has arrived on self.socket, therefore there is no need for the abandoned
+            // one anymore. TODO: Account for multiple outgoing connections.
+            self.prev_socket = None;
+        }
+        Ok(poll_res.keep_going)
     }
 
     fn drive_send(&mut self, cx: &mut Context) -> Result<bool, io::Error> {
@@ -679,6 +702,7 @@ impl EndpointRef {
             },
             state: Mutex::new(State {
                 socket,
+                prev_socket: None,
                 inner,
                 transmit_state: TransmitState::default(),
                 ipv6,
@@ -765,6 +789,7 @@ impl RecvState {
         socket: &dyn AsyncUdpSocket,
         now: Instant,
     ) -> Result<PollProgress, io::Error> {
+        let mut received_connection_packet = false;
         let mut metas = [RecvMeta::default(); BATCH_SIZE];
         let mut iovs: [IoSliceMut; BATCH_SIZE] = {
             let mut bufs = self
@@ -807,6 +832,7 @@ impl RecvState {
                                 }
                                 Some(DatagramEvent::ConnectionEvent(handle, event)) => {
                                     // Ignoring errors from dropped connections that haven't yet been cleaned up
+                                    received_connection_packet = true;
                                     let _ = self
                                         .connections
                                         .senders
@@ -823,7 +849,10 @@ impl RecvState {
                     }
                 }
                 Poll::Pending => {
-                    return Ok(PollProgress { keep_going: false });
+                    return Ok(PollProgress {
+                        received_connection_packet,
+                        keep_going: false,
+                    });
                 }
                 // Ignore ECONNRESET as it's undefined in QUIC and may be injected by an
                 // attacker
@@ -835,7 +864,10 @@ impl RecvState {
                 }
             }
             if !self.recv_limiter.allow_work() {
-                return Ok(PollProgress { keep_going: true });
+                return Ok(PollProgress {
+                    received_connection_packet,
+                    keep_going: true,
+                });
             }
         }
     }
@@ -843,6 +875,8 @@ impl RecvState {
 
 #[derive(Default)]
 struct PollProgress {
+    /// Whether a datagram was routed to an existing connection
+    received_connection_packet: bool,
     /// Whether datagram handling was interrupted early by the work limiter for fairness
     keep_going: bool,
 }
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -310,21 +310,14 @@ impl Future for EndpointDriver {
 
     #[allow(unused_mut)] // MSRV
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
-        let mut endpoint_guard = self.0.state.lock().unwrap();
-        let endpoint = &mut *endpoint_guard;
+        let mut endpoint = self.0.state.lock().unwrap();
         if endpoint.driver.is_none() {
             endpoint.driver = Some(cx.waker().clone());
         }
 
         let now = Instant::now();
         let mut keep_going = false;
-        keep_going |= endpoint.recv_state.drive_recv(
-            cx,
-            &mut endpoint.inner,
-            &mut endpoint.transmit_state,
-            &*endpoint.socket,
-            now,
-        )?;
+        keep_going |= endpoint.drive_recv(cx, now)?;
         keep_going |= endpoint.handle_events(cx, &self.0.shared);
         keep_going |= endpoint.drive_send(cx)?;
 
@@ -335,7 +328,7 @@ impl Future for EndpointDriver {
         if endpoint.ref_count == 0 && endpoint.recv_state.connections.is_empty() {
             Poll::Ready(Ok(()))
         } else {
-            drop(endpoint_guard);
+            drop(endpoint);
             // If there is more work to do schedule the endpoint task again.
             // `wake_by_ref()` is called outside the lock to minimize
             // lock contention on a multithreaded runtime.
@@ -434,6 +427,19 @@ pub(crate) struct Shared {
 }
 
 impl State {
+    fn drive_recv(&mut self, cx: &mut Context, now: Instant) -> Result<bool, io::Error> {
+        self.recv_state.recv_limiter.start_cycle();
+        let poll_res = self.recv_state.poll_socket(
+            cx,
+            &mut self.inner,
+            &mut self.transmit_state,
+            &*self.socket,
+            now,
+        );
+        self.recv_state.recv_limiter.finish_cycle();
+        poll_res
+    }
+
     fn drive_send(&mut self, cx: &mut Context) -> Result<bool, io::Error> {
         self.send_limiter.start_cycle();
 
@@ -751,7 +757,7 @@ impl RecvState {
         }
     }
 
-    fn drive_recv(
+    fn poll_socket(
         &mut self,
         cx: &mut Context,
         endpoint: &mut proto::Endpoint,
@@ -759,7 +765,6 @@ impl RecvState {
         socket: &dyn AsyncUdpSocket,
         now: Instant,
     ) -> Result<bool, io::Error> {
-        self.recv_limiter.start_cycle();
         let mut metas = [RecvMeta::default(); BATCH_SIZE];
         let mut iovs: [IoSliceMut; BATCH_SIZE] = {
             let mut bufs = self
@@ -830,12 +835,10 @@ impl RecvState {
                 }
             }
             if !self.recv_limiter.allow_work() {
-                self.recv_limiter.finish_cycle();
                 return Ok(true);
             }
         }
 
-        self.recv_limiter.finish_cycle();
         Ok(false)
     }
 }
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -180,7 +180,7 @@ impl Endpoint {
         server_name: &str,
     ) -> Result<Connecting, ConnectError> {
         let mut endpoint = self.inner.state.lock().unwrap();
-        if endpoint.driver_lost || endpoint.connections.close.is_some() {
+        if endpoint.driver_lost || endpoint.recv_state.connections.close.is_some() {
             return Err(ConnectError::EndpointStopping);
         }
         if addr.is_ipv6() && !endpoint.ipv6 {
@@ -198,6 +198,7 @@ impl Endpoint {
 
         let socket = endpoint.socket.clone();
         Ok(endpoint
+            .recv_state
             .connections
             .insert(ch, conn, socket, self.runtime.clone()))
     }
@@ -216,7 +217,7 @@ impl Endpoint {
         inner.ipv6 = addr.is_ipv6();
 
         // Generate some activity so peers notice the rebind
-        for sender in inner.connections.senders.values() {
+        for sender in inner.recv_state.connections.senders.values() {
             // Ignoring errors from dropped connections
             let _ = sender.send(ConnectionEvent::LocalAddressChanged);
         }
@@ -254,8 +255,8 @@ impl Endpoint {
     pub fn close(&self, error_code: VarInt, reason: &[u8]) {
         let reason = Bytes::copy_from_slice(reason);
         let mut endpoint = self.inner.state.lock().unwrap();
-        endpoint.connections.close = Some((error_code, reason.clone()));
-        for sender in endpoint.connections.senders.values() {
+        endpoint.recv_state.connections.close = Some((error_code, reason.clone()));
+        for sender in endpoint.recv_state.connections.senders.values() {
             // Ignoring errors from dropped connections
             let _ = sender.send(ConnectionEvent::Close {
                 error_code,
@@ -279,7 +280,7 @@ impl Endpoint {
         loop {
             {
                 let endpoint = &mut *self.inner.state.lock().unwrap();
-                if endpoint.connections.is_empty() {
+                if endpoint.recv_state.connections.is_empty() {
                     break;
                 }
                 // Construct future while lock is held to avoid race
@@ -309,25 +310,32 @@ impl Future for EndpointDriver {
 
     #[allow(unused_mut)] // MSRV
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
-        let mut endpoint = self.0.state.lock().unwrap();
+        let mut endpoint_guard = self.0.state.lock().unwrap();
+        let endpoint = &mut *endpoint_guard;
         if endpoint.driver.is_none() {
             endpoint.driver = Some(cx.waker().clone());
         }
 
         let now = Instant::now();
         let mut keep_going = false;
-        keep_going |= endpoint.drive_recv(cx, now)?;
+        keep_going |= endpoint.recv_state.drive_recv(
+            cx,
+            &mut endpoint.inner,
+            &mut endpoint.transmit_state,
+            &*endpoint.socket,
+            now,
+        )?;
         keep_going |= endpoint.handle_events(cx, &self.0.shared);
         keep_going |= endpoint.drive_send(cx)?;
 
-        if !endpoint.incoming.is_empty() {
+        if !endpoint.recv_state.incoming.is_empty() {
             self.0.shared.incoming.notify_waiters();
         }
 
-        if endpoint.ref_count == 0 && endpoint.connections.is_empty() {
+        if endpoint.ref_count == 0 && endpoint.recv_state.connections.is_empty() {
             Poll::Ready(Ok(()))
         } else {
-            drop(endpoint);
+            drop(endpoint_guard);
             // If there is more work to do schedule the endpoint task again.
             // `wake_by_ref()` is called outside the lock to minimize
             // lock contention on a multithreaded runtime.
@@ -346,7 +354,7 @@ impl Drop for EndpointDriver {
         self.0.shared.incoming.notify_waiters();
         // Drop all outgoing channels, signaling the termination of the endpoint to the associated
         // connections.
-        endpoint.connections.senders.clear();
+        endpoint.recv_state.connections.senders.clear();
     }
 }
 
@@ -373,7 +381,10 @@ impl EndpointInner {
             Ok((handle, conn)) => {
                 let socket = state.socket.clone();
                 let runtime = state.runtime.clone();
-                Ok(state.connections.insert(handle, conn, socket, runtime))
+                Ok(state
+                    .recv_state
+                    .connections
+                    .insert(handle, conn, socket, runtime))
             }
             Err(error) => {
                 if let Some(transmit) = error.response {
@@ -405,16 +416,13 @@ pub(crate) struct State {
     socket: Arc<dyn AsyncUdpSocket>,
     inner: proto::Endpoint,
     transmit_state: TransmitState,
-    incoming: VecDeque<proto::Incoming>,
+    recv_state: RecvState,
     driver: Option<Waker>,
     ipv6: bool,
-    connections: ConnectionSet,
     events: mpsc::UnboundedReceiver<(ConnectionHandle, EndpointEvent)>,
     /// Number of live handles that can be used to initiate or handle I/O; excludes the driver
     ref_count: usize,
     driver_lost: bool,
-    recv_limiter: WorkLimiter,
-    recv_buf: Box<[u8]>,
     send_limiter: WorkLimiter,
     runtime: Arc<dyn Runtime>,
 }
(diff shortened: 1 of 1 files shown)


quinn/src/endpoint.rs
```diff
@@ -408,13 +413,62 @@ impl State {
                     .write(IoSliceMut::<'a>::new(buf));
             });
         let mut iovs = unsafe { iovs.assume_init() };
+        if self.abandoned_socket.is_some() {
```
Ralith, line 416: This is a big block of code that's mostly shared with the normal path. Can we factor it out?

quinn/src/endpoint.rs
```diff
@@ -369,6 +373,7 @@ pub(crate) struct EndpointInner {
 #[derive(Debug)]
 pub(crate) struct State {
     socket: Arc<dyn AsyncUdpSocket>,
+    abandoned_socket: Option<Arc<dyn AsyncUdpSocket>>,
```
Ralith, line 376: Explain what this is with a comment.

Ralith (changes requested):
Thanks for the PR! This generally makes sense. Minor note, the diff will be a bit more readable if you set #1800's branch as the target, or just rebase onto main if the changes don't actually interact.

I'm wondering if it might be cleaner to use a single unchanging socket bound to a wildcard IP, and have each outgoing datagram set its source IP address explicitly. This would avoid the need for two separate receive paths, and allow connections to operate concurrently on multiple addresses on the same endpoint. It would break support for active migration on platforms which don't support `IPV6_PKTINFO` or equivalent, but most platforms do. This would probably be a larger change, so perhaps best left for future work even if it sounds good.

nemethf:
> the diff will be a bit more readable if you set #1800's branch as the target, 

I haven't known that's possible.  Thanks.

>or just rebase onto main if the changes don't actually interact.

I'll probably go this way.


> I'm wondering if it might be cleaner to use a single unchanging socket bound to a wildcard IP, and have each outgoing datagram set its source IP address explicitly. This would avoid the need for two separate receive paths, and allow connections to operate concurrently on multiple addresses on the same endpoint. It would break support for active migration on platforms which don't support `IPV6_PKTINFO` or equivalent, but most platforms do. 

I'm OK with this.

But does it eliminate use-cases when the client does not want to receive traffic on a specific network interface due to security reasons?  Hm.  Maybe the client could filter incoming packets based on their destination address. 

And what if the client wants to move from port A to port B?  We used to test migration with moving from 127.0.0.1:A to 127.0.0.1:B.  Although this might not be relevant in practice.

> This would probably be a larger change, so perhaps best left for future work even if it sounds good.

If there are multiple socket to listen to, one can use the `select` libc call.  It seems rust's [selecting](https://docs.rs/selecting) crate only support [AsRawFd](https://docs.rs/selecting/latest/selecting/trait.AsRawFd.html), which is not available on windows.  [libselect](https://crates.io/crates/libselect/0.1.0)  is cross-platform, but I'm not sure it is acceptable to use it in quinn.  Or is there any other way currently in quinn to poll two sockets at once?

At any rate, maybe it is enough for this pull request to query the two sockets in sequence.

Ralith:
> But does it eliminate use-cases when the client does not want to receive traffic on a specific network interface due to security reasons?

Good question. This can be solved at the host firewall level, but people are accustomed to specifying a narrow listen address automatically getting you strong isolation without any extra effort, so this could be a footgun. Packets from unexpected peers do get dropped quickly, but this would arguably remove a layer of defense-in-depth.

> And what if the client wants to move from port A to port B?

Another good question. It's not obvious to me if there's a use case here, but maybe if there's weird firewall rules, or the application is actively trying to break linkability? We're getting this at the cost of the capability to have different connections on different paths in the same endpoint, but an applications can always choose to have multiple endpoints.

It sounds like sticking with multiple sockets along the lines drafted here is the safer bet, even if it is a little ugly. We may want to revisit this for multipath in the future when concurrent outgoing paths are more important.

> At any rate, maybe it is enough for this pull request to query the two sockets in sequence.

Correct. The underlying async runtime is responsible for notifying us when any operation that we got a `Poll::Pending` (morally EWOULDBLOCK) for on the previous `Future::poll` becomes newly ready.

nemethf:
I force-pushed a new version, which eliminates the code duplication of the previous version as requested.  Due to my limited rust knowledge I struggled with the refactorization, and  the result is probably not idiomatic rust.  

Just like upstream main version, it does not call `self.recv_limiter.finish_cycle()`  in case of an error.  Maybe there's a good reason for it.

quinn/src/endpoint.rs
```diff
@@ -490,25 +504,59 @@ impl State {
...
-                self.recv_limiter.finish_cycle();
-                return Ok(true);
+                return (received_data, Ok(true));
             }
         }
+    }
 
-        self.recv_limiter.finish_cycle();
-        Ok(false)
+    fn drive_recv<'a>(&'a mut self, cx: &mut Context, now: Instant) -> Result<bool, io::Error> {
```
djc, line 524: We generally prefer to order functions top-down, so `drive_recv()` should come before `poll_socket()`.

quinn/src/endpoint.rs
```diff
@@ -432,27 +436,31 @@ pub(crate) struct Shared {
...
-            });
-        let mut iovs = unsafe { iovs.assume_init() };
+    fn poll_socket(
+        &mut self,
+        primary_socket: bool,
+        cx: &mut Context,
+        now: Instant,
+        iovs: &mut [IoSliceMut; BATCH_SIZE],
+        mut metas: [RecvMeta; BATCH_SIZE],
+    ) -> (
```
djc, line 446: I'm finding this return type pretty hard to parse. Maybe it should be a three-field `struct` or something?

quinn/src/endpoint.rs
```diff
@@ -482,6 +493,9 @@ impl State {
                                         .send(ConnectionEvent::Proto(event));
                                 }
                                 Some(DatagramEvent::Response(transmit)) => {
+                                    if !primary_socket {
+                                        continue;
```
djc, line 497: It's not obvious to me why this should happen -- maybe add a comment?

quinn/src/endpoint.rs
```diff
@@ -432,27 +436,31 @@ pub(crate) struct Shared {
...
+        mut metas: [RecvMeta; BATCH_SIZE],
+    ) -> (
+        bool, // Recevied data
+        Result<bool, io::Error>,
+    ) {
+        let mut received_data = false;
+        let socket = if primary_socket {
+            &self.socket
+        } else {
+            self.abandoned_socket.as_ref().unwrap()
```
djc, line 454: This `unwrap()` and the fact that we're passing in 5 arguments here makes me think we should restructure this. Maybe a short-lived `ReceiveContext` type that holds the `metas` and `iovs` as well as a `&dyn AsyncUdpSocket` and `&mut` references into the members of `State` that it needs?

quinn/src/endpoint.rs
nemethf: I tried to avoid doing anything non-essential on the abandoned path.  But on second though, it was unnecessarily cautious.  `DatagramEvent::NewConnection` cannot arrive to the client on the abandoned path during an client migration.  And to tell the truth, I don't understand when a `DatagramEvent::Response` is generated. 

In the latest version, I removed these tests.

quinn/src/endpoint.rs
nemethf: I force-pushed a new version addressing this and the rest of your comments.  The code is definitely easier to read.  (I had trouble defining the correct lifetimes for `struct ReceiveContext`.  In the end, it compiles and works fine.  But I have  feeling this could have been written more clearly.) 

Thanks again.

quinn/src/endpoint.rs
```diff
@@ -805,3 +855,16 @@ impl std::ops::Deref for EndpointRef {
...
 }
+
+struct ReceiveContext<'a, 'b: 'c, 'c> {
+    cx: &'c mut Context<'b>,
+    now: Instant,
+    iovs: &'a mut [IoSliceMut<'a>; BATCH_SIZE],
+    metas: &'c mut [RecvMeta; BATCH_SIZE],
+}
+
+struct PollResult {
```
djc, line 866: Please derive `Default` on this, which should do the right thing to avoid initializing this manually.

quinn/src/endpoint.rs
```diff
@@ -443,20 +447,57 @@ impl State {
...
+                // there is no need for the abandoned one anymore.
+                self.abandoned_socket = None;
+            }
+        }
+        if let Some(err) = poll_res.err {
+            Err(err)
+        } else {
+            self.recv_limiter.finish_cycle();
+            Ok(poll_res.keep_going)
+        }
```
djc, line 480: Nit: suggest writing this differently:

```rust
match poll_res.err {
    None => {
         self.recv_limiter.finish_cycle();
         Ok(poll_res.keep_going)
    }
    Some(err) => Err(err),
}
```

quinn/src/endpoint.rs
```diff
@@ -805,3 +855,16 @@ impl std::ops::Deref for EndpointRef {
...
+    cx: &'c mut Context<'b>,
+    now: Instant,
+    iovs: &'a mut [IoSliceMut<'a>; BATCH_SIZE],
+    metas: &'c mut [RecvMeta; BATCH_SIZE],
+}
+
+struct PollResult {
+    received_data: bool,
+    keep_going: bool,
+    err: Option<io::Error>,
```
djc, line 869: Nit: in Rust, it would be more idiomatic to call this `error` (we usually write out field and type names).

quinn/src/endpoint.rs
```diff
@@ -443,20 +447,57 @@ impl State {
...
+        } else {
+            self.recv_limiter.finish_cycle();
+            Ok(poll_res.keep_going)
+        }
+    }
+
+    fn poll_socket(
+        &mut self,
+        socket: &Arc<dyn AsyncUdpSocket>,
+        rc: &mut ReceiveContext,
```
djc, line 486: Nit: let's call this `rcx`?

quinn/src/endpoint.rs
nemethf: Nice.  I still find the explicit initialization more readable in the return statements of `poll_socket()`.  It is okay to only use `default()` in `drive_recv()`?

djc: Sure, that's what I had in mind, too!

quinn/src/endpoint.rs
```diff
@@ -805,3 +855,16 @@ impl std::ops::Deref for EndpointRef {
         &self.0
     }
 }
+
+struct ReceiveContext<'a, 'b: 'c, 'c> {
```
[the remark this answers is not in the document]
djc: Okay, let's leave as is then, thanks for working it out!

quinn/src/endpoint.rs
nemethf: Sure, I pushed the updated version.  It's a bit unfortunate that I learn rust while I work on quinn.

djc (approved):

quinn/src/endpoint.rs
djc: No problem, we're also just pretty picky, thanks for sticking with it!

djc:
Still needs a rebase onto current main. If you're in it, can you change the order of the methods to have `drive_recv()` before `poll_socket()` -- I seem to recall it was like that at some point, but maybe that got lost in the shuffle.

quinn/src/endpoint.rs
```diff
@@ -425,6 +429,38 @@ impl State {
...
+            cx,
+            now,
+            metas: &mut metas,
+        };
+        let mut poll_res = PollResult::default();
+        if let Some(socket) = &self.abandoned_socket {
+            poll_res = self.poll_socket(&socket.clone(), rcx);
+        };
+        if !poll_res.keep_going {
+            poll_res = self.poll_socket(&self.socket.clone(), rcx);
```
Ralith, line 442: Calling `Arc::clone`, and then `Arc::drop`, on every transmit feels a bit ugly. Alternative ideas, in order of preference:
- Group the fields accessed by `poll_socket` into a smaller struct owned by `State`, then move `poll_socket` into a method on that struct and pass in a plain reference to the socket
- Lift `poll_socket` out into a method on `ReceiveContext` instead, adding fields as needed to refer to values owned by `State`

quinn/src/endpoint.rs
```diff
@@ -425,6 +429,38 @@ impl State {
     fn drive_recv(&mut self, cx: &mut Context, now: Instant) -> Result<bool, io::Error> {
         self.recv_limiter.start_cycle();
         let mut metas = [RecvMeta::default(); BATCH_SIZE];
+        let rcx = &mut ReceiveContext {
+            cx,
+            now,
+            metas: &mut metas,
```
Ralith, line 435: Could we move the definition of `metas` into `poll_socket`? One less thing to pass down that way. Reusing a PoD stack allocation probably doesn't buy us much.

quinn/src/endpoint.rs
```diff
@@ -437,16 +473,17 @@ impl State {
...
         loop {
-            match self.socket.poll_recv(cx, &mut iovs, &mut metas) {
+            match socket.poll_recv(rcx.cx, &mut iovs, rcx.metas) {
                 Poll::Ready(Ok(msgs)) => {
                     self.recv_limiter.record_work(msgs);
-                    for (meta, buf) in metas.iter().zip(iovs.iter()).take(msgs) {
+                    for (meta, buf) in rcx.metas.iter().zip(iovs.iter()).take(msgs) {
                         let mut data: BytesMut = buf[0..meta.len].into();
                         while !data.is_empty() {
+                            received_data = true;
```
Ralith, line 482: This will trigger on any UDP datagram, including random garbage that isn't from the remote server or even a valid QUIC packet. Would it make more sense to rename it `received_connection_packet` or similar, and only set it in the `ConnectionEvent` case?

Ralith (changes requested):
What if the endpoint has multiple outgoing connections to multiple servers? Should we hold on to `abandoned_socket` until we've authenticated a post-migration packet from each live connection?

This PR is an improvement over the status quo regardless, so it's okay to leave that for future work, but something to consider.

quinn/src/endpoint.rs
nemethf: Commit 41cb514 implements the first idea.

quinn/src/endpoint.rs
nemethf: Commit 59516f7 implements this.

quinn/src/endpoint.rs
```diff
@@ -795,3 +758,140 @@ impl std::ops::Deref for EndpointRef {
         &self.0
     }
 }
+
+#[derive(Debug)]
+struct PollState {
```
Ralith, line 763: Document. Maybe "State involved in processing an incoming datagram"?

quinn/src/endpoint.rs
```diff
@@ -795,3 +758,140 @@ impl std::ops::Deref for EndpointRef {
...
+            transmit_state: TransmitState::default(),
+            incoming: VecDeque::new(),
+            recv_buf: recv_buf.into(),
+            recv_limiter: WorkLimiter::new(RECV_TIME_BOUND),
+            metas: [RecvMeta::default(); BATCH_SIZE],
+        }
+    }
+    fn poll_socket(
+        &mut self,
+        socket: &Arc<dyn AsyncUdpSocket>,
```
Ralith, line 801: We don't need to clone the `Arc`, so we should deref eagerly for clarity:
```suggestion
        socket: &dyn AsyncUdpSocket,
```

quinn/src/endpoint.rs
```diff
@@ -214,11 +217,12 @@ impl Endpoint {
         let addr = socket.local_addr()?;
         let socket = self.runtime.wrap_udp_socket(socket)?;
         let mut inner = self.inner.state.lock().unwrap();
+        inner.abandoned_socket = Some(inner.socket.clone());
         inner.socket = socket;
```
Ralith, line 221: We can avoid a `clone` here and clarify intent while we're at it:
```suggestion
        inner.abandoned_socket = Some(mem::replace(&mut inner.socket, socket));
```

Ralith (changes requested):
Thanks, I think the end state of this looks good now.  I think rewriting history as follows, with roughly the same end state, would provide the greatest clarity to support review for correctness:

1. Introduce `PollState`, moving `drive_recv` onto it (plus a socket argument) without changing behavior
1. Reintroduce a method on `State` and move the logic to start/end the `recv_limiter` cycle into it
1. Introduce `abandoned_socket` and associated logic

If that's beyond what you're comfortable with/motivated for, let me know and I can have a go at it myself.

> Instead of force-pushing a new version, I pushed separate commits to address your review points.

Quinn is large enough that we can't usually keep the state of every open PR memorized, so it's much easier to re-review from scratch than to try to judge incremental revisions based on half-remembered context. As a bonus, this makes it easier for new reviewers to get involved in a PR, and allows more useful history to be merged.

This workflow does make it harder to see how a specific comment was addressed. A better code review tool would address both requirements, but we make do.

> As an additional improvement it would be good not to create iovs with each poll_socket() call, but I was not able to factor this out into PollState::new().

That would be out of scope for this PR anyway, no worries.

nemethf:
> I think rewriting history as follows, with roughly the same end state, would provide the greatest clarity to support review for correctness:
> 
>     1. Introduce `PollState`, moving `drive_recv` onto it (plus a socket argument) without changing behavior
> 
>     2. Reintroduce a method on `State` and move the logic to start/end the `recv_limiter` cycle into it
> 
>     3. Introduce `abandoned_socket` and associated logic
> 
> 
> If that's beyond what you're comfortable with/motivated for, let me know and I can have a go at it myself.

These steps sound logical, but I just cannot implement step 1 without step 2.  So it would be great if you cloud take it over.  Thanks.

Ralith:
Split up the changes as described, and addressed my own outstanding feedback. I still need to do a final once-over for correctness, though.

Ralith (approved):

Ralith:
Did a little more readability cleanup and satisfied myself as far as correctness goes.

quinn/src/endpoint.rs
```diff
@@ -428,100 +441,24 @@ pub(crate) struct Shared {
...
-        self.recv_limiter.finish_cycle();
-        Ok(false)
-    }
-
     fn drive_send(&mut self, cx: &mut Context) -> Result<bool, io::Error> {
         self.send_limiter.start_cycle();
 
         let result = loop {
-            if self.transmit_state.outgoing.is_empty() {
+            if self.recv_state.transmit_state.outgoing.is_empty() {
```
djc, line 471: Feels weird for `drive_send()` to reference `recv_state`... Should it be named something else?

quinn/src/endpoint.rs
```diff
@@ -441,6 +437,15 @@ pub(crate) struct Shared {
 }
 
 impl State {
+    fn drive_recv(&mut self, cx: &mut Context, now: Instant) -> Result<bool, io::Error> {
+        self.recv_state.recv_limiter.start_cycle();
+        let poll_res = self
```
djc, line 455: Maybe add a comment about why it's correct to skip `finish_cycle()` on error returns of `poll_socket()`?

quinn/src/endpoint.rs
```diff
@@ -215,7 +216,7 @@ impl Endpoint {
         let addr = socket.local_addr()?;
         let socket = self.runtime.wrap_udp_socket(socket)?;
         let mut inner = self.inner.state.lock().unwrap();
-        inner.socket = socket;
+        inner.abandoned_socket = Some(mem::replace(&mut inner.socket, socket));
```
djc, line 219: I find this name a little unwieldy, how do you feel about `prev_socket`?

quinn/src/endpoint.rs
```diff
@@ -841,10 +866,19 @@ impl RecvState {
...
+                });
             }
         }
-
-        Ok(false)
     }
 }
+
+#[derive(Default)]
+struct PollProgress {
```
djc, line 879: Would be nice to make the introduction of this type a separate commit.

quinn/src/endpoint.rs
Ralith: It doesn't matter either way because errors here are fatal to the endpoint driver. I refactored to not skip it because that was more concise than a comment explaining why we can.

djc (approved):

PR #1808, opened 2024-04-04
bump rustls 

gabrik:
Trying to make it pass the CI

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -898,14 +898,14 @@ impl ClientConfig {
     /// Create a client configuration that trusts the platform's native roots
     #[cfg(feature = "platform-verifier")]
     pub fn with_platform_verifier() -> Self {
-        Self::new(Arc::new(crypto::rustls::client_config(
+        Self::new(Arc::new(crypto::rustls::QuicClientConfig::new(
             ServerVerifier::Platform,
         )))
     }
 
     /// Create a client configuration that trusts specified trust anchors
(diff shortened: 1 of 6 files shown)


djc:
Have rebased my PR so you can work on top of that.

gabrik:
Maybe it is a stupid question, but what's the point of `QuicClientConfig` ? 
the `crypto::ClientConfig` can be easily implemented for `rustls::ClientConfig` I'm missing something?

djc:
> Maybe it is a stupid question, but what's the point of `QuicClientConfig` ? the `crypto::ClientConfig` can be easily implemented for `rustls::ClientConfig` I'm missing something?

The goal is to do some validation that the `ClientConfig`'s crypto provider can support QUIC connections:

* Has cipher suites for TLS 1.3
* The 1.3 cipher suites support QUIC
* Specifically supports AES-128-GCM-SHA256 which QUIC mandates for initial packets

I've squashed your changes into my commit and will make some further changes. Would you be up for making a similar changes for the `ServerConfig`?

gabrik:
Sure, I can do the same for the `ServerConfig`

Btw, I think keeping the `impl crypto::ClientConfig for rustls::ClientConfig` (and the server equivalent) would be good as modifying the `QuicClientConfig.inner` is not ergonomic as it is behind an `Arc` as you can see from the tests.

We use for instance to enable client authentication or other stuff in the TLS config.

djc:
See the current state of #1715 for a bunch of changes I made to how we handle the client side that might apply to the server side as well. Please base your next PR on the current state of that branch.

PR #1809, opened 2024-04-05
feat(server-config): added QuicServerConfig

gabrik:


gabrik:
4 tests fails, I cannot tell why.

Seems that I generate the configuration in the same way it was done before, so it is unclear to me.

gabrik:
Was the `alpn` configuration, everything should be ok now.

Btw, I've seen a bit too much `unwrap` in both `QuicClientConfig` and `QuicServerConfig` I think their constructors should return `Result<Self, ConfigurationError>` or something similar.

It is not a problem for the way we use the library as we will use the `try_from` passing directly rustls types, but for a user that wants to use those utils it can indeed be an issue.

djc:
Thanks, pulled this into #1715 as well.

PR #1810, opened 2024-04-06
Update MSRV to 1.66

Ralith:


quinn-proto/src/range_set/tests.rs
```diff
@@ -149,6 +149,16 @@ macro_rules! common_set_tests {
                 }
             }
 
+            #[test]
+            fn minmax() {
```
djc, line 153: Should this be named `min_max`?

djc (approved):

PR #1811, opened 2024-04-07
Allocate Incoming response buffers as needed

Ralith:
Threading buffers through alongside the `Incoming` does not reduce the number of allocations.


djc (approved):
Nice!

PR #1821, opened 2024-04-14
Fix race condition between 0-RTT and Incoming

### LINKED ISSUE #1820: 0-RTT packets can be lost due to race condition introduced along with Incoming
Unfortunately it looks like in #1752 I accidentally introduced a bug. Apologies for not catching it.

Bug description: If a client sends a 0-RTT datagram to the server after sending the datagram with its Initial packet, and it is received and processed by the endpoint before the corresponding `Incoming` is `accept`-ed, the 0-RTT packet will not find any associated connection and thus be dropped. As a consequence, this seems to cause request/response pairs on connections where this occurs to actually take 3 round trips to go through successfully.

I am working on a fix, and I will see if I can get a PR opened for it tomorrow.

gretchenfrage:
Closes #1820

The fix:

- Endpoint now maintains a slab with an entry for each pending Incoming to buffer received data.
- ConnectionIndex now maps initial DCID to that slab key immediately upon construction of Incoming.
- If Incoming is accepted, association is overridden with association with ConnectionHandle, and all buffered datagrams are fed to newly constructed Connection.
- If Incoming is refused/retried/ignored, or accepting errors, association and slab entry are cleaned up to prevent memory leak.

Additional considerations:

- The Incoming::ignore operation can no longer be implemented as just dropping it. To help prevent incorrect API usage, proto::Incoming is modified to log a warning if it is dropped without being passed to Endpoint::accept/refuse/retry/ignore.
- To help protect against memory exhaustion attacks, per-Incoming buffered data is limited to twice the receive window or 10 KB, which- ever is larger. Excessive packets silently dropped.

  - Does this introduce a new vulnerability to an attack in which an attacker could spam a server with 0-RTT packets with the same connection ID as it observed a client attempting to initiate a 0-RTT connection to the server? I do think so.

    Is this a severe problem? Here's two reasons I don't think so:

    1. The default receive window is set to max value, so this won't actually kick in unless the user is already hardening against adverse conditions. 2. It is already possible for an on-path attacker to distrupt a connection handshake if 0.5-RTT data is being used, so this probably doesn't actually expand the set of situations in which it's vulnerable to this kind of vulnerability.

    Could this be avoided? Possibly by introducing additional state to the buffering state to validate whether these packets are validly encrypted for the associated connection? However, that may risk making these operations costly enough that they start to defeat the DDOS-resistance abilities of the Incoming API.

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -776,6 +776,10 @@ pub struct ServerConfig {
 
     pub(crate) preferred_address_v4: Option<SocketAddrV4>,
     pub(crate) preferred_address_v6: Option<SocketAddrV6>,
+
+    pub(crate) max_incoming: usize,
+    pub(crate) incoming_buffer_size: u64,
+    pub(crate) incoming_buffer_size_total: u64,
 }
 
 impl ServerConfig {
@@ -795,6 +799,10 @@ impl ServerConfig {
 
             preferred_address_v4: None,
             preferred_address_v6: None,
+
+            max_incoming: 1 << 16,
+            incoming_buffer_size: 10 << 20,
+            incoming_buffer_size_total: 100 << 20,
         }
     }
 
@@ -836,6 +844,54 @@ impl ServerConfig {
     /// If the client is able to reach this address, it will switch to it.
     pub fn preferred_address_v6(&mut self, address: Option<SocketAddrV6>) -> &mut Self {
         self.preferred_address_v6 = address;
+        self
+    }
+
+    /// Maximum number of [`Incoming`][crate::Incoming] to allow to exist at a time
+    ///
+    /// An [`Incoming`][crate::Incoming] comes into existence when an incoming connection attempt
+    /// is received and stops existing when the application either accepts it or otherwise disposes
+    /// of it. While this limit is reached, new incoming connection attempts are immediately
+    /// refused. Larger values have greater worst-case memory consumption, but accommodate greater
+    /// application latency in handling incoming connection attempts.
+    ///
+    /// The default value is set to 65536. With a typical Ethernet MTU of 1500 bytes, this limits
+    /// memory consumption from this to under 100 MiB--a generous amount that still prevents memory
+    /// exhaustion in most contexts.
+    pub fn max_incoming(&mut self, max_incoming: usize) -> &mut Self {
+        self.max_incoming = max_incoming;
+        self
+    }
+
+    /// Maximum number of received bytes to buffer for each [`Incoming`][crate::Incoming]
+    ///
+    /// An [`Incoming`][crate::Incoming] comes into existence when an incoming connection attempt
+    /// is received and stops existing when the application either accepts it or otherwise disposes
+    /// of it. This limit governs only packets received within that period, and does not include
+    /// the first packet. Packets received in excess of this limit are dropped, which may cause
+    /// 0-RTT or handshake data to have to be retransmitted.
+    ///
+    /// The default value is set to 10 MiB--an amount such that in most situations a client would
+    /// not transmit that much 0-RTT data faster than the server handles the corresponding
+    /// [`Incoming`][crate::Incoming].
+    pub fn incoming_buffer_size(&mut self, incoming_buffer_size: u64) -> &mut Self {
+        self.incoming_buffer_size = incoming_buffer_size;
+        self
+    }
+
+    /// Maximum number of received bytes to buffer for all [`Incoming`][crate::Incoming]
+    /// collectively
+    ///
+    /// An [`Incoming`][crate::Incoming] comes into existence when an incoming connection attempt
+    /// is received and stops existing when the application either accepts it or otherwise disposes
+    /// of it. This limit governs only packets received within that period, and does not include
+    /// the first packet. Packets received in excess of this limit are dropped, which may cause
+    /// 0-RTT or handshake data to have to be retransmitted.
+    ///
+    /// The default value is set to 100 MiB--a generous amount that still prevents memory
+    /// exhaustion in most contexts.
+    pub fn incoming_buffer_size_total(&mut self, incoming_buffer_size_total: u64) -> &mut Self {
+        self.incoming_buffer_size_total = incoming_buffer_size_total;
         self
     }
 }
@@ -880,6 +936,12 @@ impl fmt::Debug for ServerConfig {
             .field("migration", &self.migration)
             .field("preferred_address_v4", &self.preferred_address_v4)
             .field("preferred_address_v6", &self.preferred_address_v6)
+            .field("max_incoming", &self.max_incoming)
+            .field("incoming_buffer_size", &self.incoming_buffer_size)
+            .field(
+                "incoming_buffer_size_total",
+                &self.incoming_buffer_size_total,
+            )
             .finish()
     }
 }
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -1,7 +1,7 @@
 use std::{
     collections::{hash_map, HashMap},
     convert::TryFrom,
-    fmt,
+    fmt, mem,
     net::{IpAddr, SocketAddr},
     ops::{Index, IndexMut},
     sync::Arc,
@@ -51,6 +51,9 @@ pub struct Endpoint {
     allow_mtud: bool,
     /// Time at which a stateless reset was most recently sent
     last_stateless_reset: Option<Instant>,
+    /// Buffered Initial and 0-RTT messages for pending incoming connections
+    incoming_buffers: Slab<IncomingBuffer>,
+    all_incoming_buffers_total_bytes: u64,
 }
 
 impl Endpoint {
@@ -74,6 +77,8 @@ impl Endpoint {
             server_config,
             allow_mtud,
             last_stateless_reset: None,
+            incoming_buffers: Slab::new(),
+            all_incoming_buffers_total_bytes: 0,
         }
     }
 
@@ -189,17 +194,42 @@ impl Endpoint {
         //
 
         let addresses = FourTuple { remote, local_ip };
-        if let Some(ch) = self.index.get(&addresses, &first_decode) {
-            return Some(DatagramEvent::ConnectionEvent(
-                ch,
-                ConnectionEvent(ConnectionEventInner::Datagram(DatagramConnectionEvent {
-                    now,
-                    remote: addresses.remote,
-                    ecn,
-                    first_decode,
-                    remaining,
-                })),
-            ));
+        if let Some(route_to) = self.index.get(&addresses, &first_decode) {
+            let event = DatagramConnectionEvent {
+                now,
+                remote: addresses.remote,
+                ecn,
+                first_decode,
+                remaining,
+            };
+            match route_to {
+                RouteDatagramTo::Incoming(incoming_idx) => {
+                    let incoming_buffer = &mut self.incoming_buffers[incoming_idx];
+                    let config = &self.server_config.as_ref().unwrap();
+
+                    if incoming_buffer
+                        .total_bytes
+                        .checked_add(datagram_len as u64)
+                        .map_or(false, |n| n <= config.incoming_buffer_size)
+                        && self
+                            .all_incoming_buffers_total_bytes
+                            .checked_add(datagram_len as u64)
+                            .map_or(false, |n| n <= config.incoming_buffer_size_total)
+                    {
+                        incoming_buffer.datagrams.push(event);
+                        incoming_buffer.total_bytes += datagram_len as u64;
+                        self.all_incoming_buffers_total_bytes += datagram_len as u64;
+                    }
+
+                    return None;
+                }
+                RouteDatagramTo::Connection(ch) => {
+                    return Some(DatagramEvent::ConnectionEvent(
+                        ch,
+                        ConnectionEvent(ConnectionEventInner::Datagram(event)),
+                    ))
+                }
+            }
         }
 
         //
@@ -489,6 +519,10 @@ impl Endpoint {
             }
         };
 
+        let incoming_idx = self.incoming_buffers.insert(IncomingBuffer::default());
+        self.index
+            .insert_initial_incoming(orig_dst_cid, incoming_idx);
+
         Some(DatagramEvent::NewConnection(Incoming {
             addresses,
             ecn,
@@ -501,6 +535,8 @@ impl Endpoint {
             crypto,
             retry_src_cid,
             orig_dst_cid,
+            incoming_idx,
+            improper_drop_warner: IncomingImproperDropWarner,
         }))
     }
 
@@ -512,6 +548,11 @@ impl Endpoint {
         buf: &mut Vec<u8>,
         server_config: Option<Arc<ServerConfig>>,
     ) -> Result<(ConnectionHandle, Connection), AcceptError> {
+        let remote_address_validated = incoming.remote_address_validated();
+        incoming.improper_drop_warner.dismiss();
+        let incoming_buffer = self.incoming_buffers.remove(incoming.incoming_idx);
+        self.all_incoming_buffers_total_bytes -= incoming_buffer.total_bytes;
+
         let packet_number = incoming.packet.header.number.expand(0);
         let InitialHeader {
             src_cid,
@@ -522,6 +563,7 @@ impl Endpoint {
 
         if self.cids_exhausted() {
             debug!("refusing connection");
+            self.index.remove_initial(incoming.orig_dst_cid);
             return Err(AcceptError {
                 cause: ConnectionError::CidsExhausted,
                 response: Some(self.initial_close(
@@ -550,6 +592,7 @@ impl Endpoint {
             .is_err()
         {
             debug!(packet_number, "failed to authenticate initial packet");
+            self.index.remove_initial(incoming.orig_dst_cid);
             return Err(AcceptError {
                 cause: TransportError::PROTOCOL_VIOLATION("authentication failed").into(),
                 response: None,
@@ -596,11 +639,12 @@ impl Endpoint {
             tls,
             Some(server_config),
             transport_config,
-            incoming.remote_address_validated(),
+            remote_address_validated,
         );
         if dst_cid.len() != 0 {
             self.index.insert_initial(dst_cid, ch);
         }
+
         match conn.handle_first_packet(
             now,
             incoming.addresses.remote,
@@ -611,6 +655,11 @@ impl Endpoint {
         ) {
             Ok(()) => {
                 trace!(id = ch.0, icid = %dst_cid, "new connection");
+
+                for event in incoming_buffer.datagrams {
+                    conn.handle_event(ConnectionEvent(ConnectionEventInner::Datagram(event)))
+                }
+
                 Ok((ch, conn))
             }
             Err(e) => {
@@ -637,7 +686,8 @@ impl Endpoint {
         &mut self,
         header: &PlainInitialHeader,
     ) -> Result<(), TransportError> {
-        if self.cids_exhausted() {
+        let config = &self.server_config.as_ref().unwrap();
+        if self.cids_exhausted() || self.incoming_buffers.len() >= config.max_incoming {
             return Err(TransportError::CONNECTION_REFUSED(""));
         }
 
@@ -663,6 +713,9 @@ impl Endpoint {
 
     /// Reject this incoming connection attempt
     pub fn refuse(&mut self, incoming: Incoming, buf: &mut Vec<u8>) -> Transmit {
+        self.clean_up_incoming(&incoming);
+        incoming.improper_drop_warner.dismiss();
+
         self.initial_close(
             incoming.packet.header.version,
             incoming.addresses,
@@ -680,6 +733,10 @@ impl Endpoint {
         if incoming.remote_address_validated() {
             return Err(RetryError(incoming));
         }
+
+        self.clean_up_incoming(&incoming);
+        incoming.improper_drop_warner.dismiss();
+
         let server_config = self.server_config.as_ref().unwrap();
 
         // First Initial
@@ -722,6 +779,22 @@ impl Endpoint {
             segment_size: None,
             src_ip: incoming.addresses.local_ip,
         })
+    }
+
+    /// Ignore this incoming connection attempt, not sending any packet in response
+    ///
+    /// Doing this actively, rather than merely dropping the [`Incoming`], is necessary to prevent
+    /// memory leaks due to state within [`Endpoint`] tracking the incoming connection.
+    pub fn ignore(&mut self, incoming: Incoming) {
+        self.clean_up_incoming(&incoming);
+        incoming.improper_drop_warner.dismiss();
+    }
+
+    /// Clean up endpoint data structures associated with an `Incoming`.
+    fn clean_up_incoming(&mut self, incoming: &Incoming) {
+        self.index.remove_initial(incoming.orig_dst_cid);
+        let incoming_buffer = self.incoming_buffers.remove(incoming.incoming_idx);
+        self.all_incoming_buffers_total_bytes -= incoming_buffer.total_bytes;
     }
 
     fn add_connection(
@@ -870,8 +943,28 @@ impl fmt::Debug for Endpoint {
             .field("connections", &self.connections)
             .field("config", &self.config)
             .field("server_config", &self.server_config)
+            // incoming_buffers too large
+            .field("incoming_buffers.len", &self.incoming_buffers.len())
+            .field(
+                "all_incoming_buffers_total_bytes",
+                &self.all_incoming_buffers_total_bytes,
+            )
             .finish()
     }
+}
+
+/// Buffered Initial and 0-RTT messages for a pending incoming connection
+#[derive(Default)]
+struct IncomingBuffer {
+    datagrams: Vec<DatagramConnectionEvent>,
+    total_bytes: u64,
+}
+
+/// Part of protocol state incoming datagrams can be routed to
+#[derive(Copy, Clone, Debug)]
+enum RouteDatagramTo {
+    Incoming(usize),
+    Connection(ConnectionHandle),
 }
 
 /// Maps packets to existing connections
@@ -880,7 +973,7 @@ struct ConnectionIndex {
     /// Identifies connections based on the initial DCID the peer utilized
     ///
     /// Uses a standard `HashMap` to protect against hash collision attacks.
-    connection_ids_initial: HashMap<ConnectionId, ConnectionHandle>,
+    connection_ids_initial: HashMap<ConnectionId, RouteDatagramTo>,
     /// Identifies connections based on locally created CIDs
     ///
     /// Uses a cheaper hash function since keys are locally created
@@ -897,9 +990,21 @@ struct ConnectionIndex {
 }
 
 impl ConnectionIndex {
+    /// Associate an incoming connection with its initial destination CID
+    fn insert_initial_incoming(&mut self, dst_cid: ConnectionId, incoming_key: usize) {
+        self.connection_ids_initial
+            .insert(dst_cid, RouteDatagramTo::Incoming(incoming_key));
+    }
+
+    /// Remove an association with an initial destination CID
+    fn remove_initial(&mut self, dst_cid: ConnectionId) {
+        self.connection_ids_initial.remove(&dst_cid);
+    }
+
     /// Associate a connection with its initial destination CID
     fn insert_initial(&mut self, dst_cid: ConnectionId, connection: ConnectionHandle) {
-        self.connection_ids_initial.insert(dst_cid, connection);
+        self.connection_ids_initial
+            .insert(dst_cid, RouteDatagramTo::Connection(connection));
     }
 
     /// Associate a connection with its first locally-chosen destination CID if used, or otherwise
@@ -940,10 +1045,10 @@ impl ConnectionIndex {
     }
 
     /// Find the existing connection that `datagram` should be routed to, if any
-    fn get(&self, addresses: &FourTuple, datagram: &PartialDecode) -> Option<ConnectionHandle> {
+    fn get(&self, addresses: &FourTuple, datagram: &PartialDecode) -> Option<RouteDatagramTo> {
         if datagram.dst_cid().len() != 0 {
             if let Some(&ch) = self.connection_ids.get(datagram.dst_cid()) {
-                return Some(ch);
+                return Some(RouteDatagramTo::Connection(ch));
             }
         }
         if datagram.is_initial() || datagram.is_0rtt() {
@@ -953,7 +1058,7 @@ impl ConnectionIndex {
         }
         if datagram.dst_cid().len() == 0 {
             if let Some(&ch) = self.connection_remotes.get(addresses) {
-                return Some(ch);
+                return Some(RouteDatagramTo::Connection(ch));
             }
         }
         let data = datagram.data();
@@ -963,6 +1068,7 @@ impl ConnectionIndex {
         self.connection_reset_tokens
             .get(addresses.remote, &data[data.len() - RESET_TOKEN_SIZE..])
             .cloned()
+            .map(RouteDatagramTo::Connection)
     }
 }
 
@@ -1025,6 +1131,8 @@ pub struct Incoming {
     crypto: Keys,
     retry_src_cid: Option<ConnectionId>,
     orig_dst_cid: ConnectionId,
+    incoming_idx: usize,
+    improper_drop_warner: IncomingImproperDropWarner,
 }
 
 impl Incoming {
@@ -1059,7 +1167,24 @@ impl fmt::Debug for Incoming {
             // rest is too big and not meaningful enough
             .field("retry_src_cid", &self.retry_src_cid)
             .field("orig_dst_cid", &self.orig_dst_cid)
+            .field("incoming_idx", &self.incoming_idx)
+            // improper drop warner contains no information
             .finish_non_exhaustive()
+    }
+}
+
+struct IncomingImproperDropWarner;
+
+impl IncomingImproperDropWarner {
+    fn dismiss(self) {
+        mem::forget(self);
+    }
+}
+
+impl Drop for IncomingImproperDropWarner {
+    fn drop(&mut self) {
+        warn!("quinn_proto::Incoming dropped without passing to Endpoint::accept/refuse/retry/ignore \
+               (may cause memory leak and eventual inability to accept new connections)");
     }
 }
 
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -27,8 +27,8 @@ use crate::{
         PartialDecode, PlainInitialHeader,
     },
     shared::{
-        ConnectionEvent, ConnectionEventInner, ConnectionId, EcnCodepoint, EndpointEvent,
-        EndpointEventInner, IssuedCid,
+        ConnectionEvent, ConnectionEventInner, ConnectionId, DatagramConnectionEvent, EcnCodepoint,
+        EndpointEvent, EndpointEventInner, IssuedCid,
     },
     token::TokenDecodeError,
     transport_parameters::{PreferredAddress, TransportParameters},
@@ -192,13 +192,13 @@ impl Endpoint {
         if let Some(ch) = self.index.get(&addresses, &first_decode) {
             return Some(DatagramEvent::ConnectionEvent(
                 ch,
-                ConnectionEvent(ConnectionEventInner::Datagram {
+                ConnectionEvent(ConnectionEventInner::Datagram(DatagramConnectionEvent {
                     now,
                     remote: addresses.remote,
                     ecn,
                     first_decode,
                     remaining,
-                }),
+                })),
             ));
         }
 
--- a/quinn-proto/src/shared.rs
+++ b/quinn-proto/src/shared.rs
@@ -11,15 +11,19 @@ pub struct ConnectionEvent(pub(crate) ConnectionEventInner);
 #[derive(Debug)]
 pub(crate) enum ConnectionEventInner {
     /// A datagram has been received for the Connection
-    Datagram {
-        now: Instant,
-        remote: SocketAddr,
-        ecn: Option<EcnCodepoint>,
-        first_decode: PartialDecode,
-        remaining: Option<BytesMut>,
-    },
+    Datagram(DatagramConnectionEvent),
     /// New connection identifiers have been issued for the Connection
     NewIdentifiers(Vec<IssuedCid>, Instant),
+}
+
+/// Variant of [`ConnectionEventInner`].
+#[derive(Debug)]
+pub(crate) struct DatagramConnectionEvent {
+    pub(crate) now: Instant,
+    pub(crate) remote: SocketAddr,
+    pub(crate) ecn: Option<EcnCodepoint>,
+    pub(crate) first_decode: PartialDecode,
+    pub(crate) remaining: Option<BytesMut>,
 }
 
 /// Events sent from a Connection to an Endpoint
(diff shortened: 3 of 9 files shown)


quinn-proto/src/endpoint.rs
```diff
@@ -50,6 +50,8 @@ pub struct Endpoint {
     allow_mtud: bool,
     /// Time at which a stateless reset was most recently sent
     last_stateless_reset: Option<Instant>,
+    /// Buffered 0-rtt messages for pending incoming connections
```
Ralith, line 53: These aren't necessarily 0-RTT. They might be Initial packets bearing extra handshake information, or even just padding intended to mitigate anti-amplification limits.

quinn-proto/src/endpoint.rs
```diff
@@ -188,17 +191,41 @@ impl Endpoint {
...
+                remaining,
+            };
+            match route_to {
+                RouteDatagramTo::Incoming(incoming_idx) => {
+                    let incoming_buffer = &mut self.incoming_buffers[incoming_idx];
+                    let server_config = self.server_config.as_ref().unwrap();
+                    let buffer_limit = (server_config.transport.receive_window.into_inner()
+                        as usize)
+                        .saturating_mul(2)
+                        .max(10_000);
```
Ralith, line 209: Why not use the receive window unmodified? If there's a strong justification, include it in a comment for posterity.

quinn-proto/src/endpoint.rs
```diff
@@ -878,13 +942,27 @@ impl fmt::Debug for Endpoint {
     }
 }
 
+/// Buffered 0-rtt messages for a pending incoming connection
```
Ralith, line 945: Not necessarily 0-RTT.

quinn-proto/src/endpoint.rs
```diff
@@ -878,13 +942,27 @@ impl fmt::Debug for Endpoint {
...
+#[derive(Default)]
+struct IncomingBuffer {
+    datagrams: Vec<DatagramConnectionEvent>,
+    total_bytes: usize,
+}
+
+/// Part of protocol state incoming datagram can be routed to
```
Ralith, line 952:
```suggestion
/// Part of protocol state incoming datagrams can be routed to
```

quinn-proto/src/endpoint.rs
```diff
@@ -901,9 +979,21 @@ struct ConnectionIndex {
...
 
 impl ConnectionIndex {
+    /// Associate an incoming connection with its initial destination CID
+    fn insert_initial_incoming(&mut self, dst_cid: ConnectionId, incoming_key: usize) {
+        self.connection_ids_initial
+            .insert(dst_cid, RouteDatagramTo::Incoming(incoming_key));
+    }
+
+    /// Remove an association between an incoming connection and its initial destination CID
+    fn remove_initial_incoming(&mut self, dst_cid: ConnectionId) {
```
Ralith, line 989: nit: the behavior here doesn't seem to be `Incoming`-specific
```suggestion
    fn remove_initial(&mut self, dst_cid: ConnectionId) {
```

quinn-proto/src/endpoint.rs
```diff
@@ -1037,36 +1131,46 @@ impl Incoming {
...
+            .field("retry_src_cid", &inner.retry_src_cid)
+            .field("orig_dst_cid", &inner.orig_dst_cid)
+            .field("incoming_idx", &inner.incoming_idx)
             .finish_non_exhaustive()
     }
 }
 
+impl Drop for Incoming {
+    fn drop(&mut self) {
+        if self.0.is_some() {
```
Ralith, line 1168: Is this the only reason we've wrapped the contents in an `Option`? I'd rather avoid the readability cost of traversing and `unwrap`ing the `Option` everywhere. Maybe add a ZST field that logs on drop, and `mem::forget` it on the intended path?

quinn-proto/src/endpoint.rs
```diff
@@ -486,19 +513,26 @@ impl Endpoint {
...
-        Some(DatagramEvent::NewConnection(Incoming {
-            addresses,
-            ecn,
-            packet: InitialPacket {
-                header,
-                header_data: packet.header_data,
-                payload: packet.payload,
+        let incoming_idx = self.incoming_buffers.insert(Default::default());
+        self.index
+            .insert_initial_incoming(orig_dst_cid, incoming_idx);
```
Ralith, line 518: How do we limit the number of these? Buffer size limits will not be an effective defense if the number of buffers is unbounded.

Ralith (changes requested):
Great catch, thanks! This was very likely to be an issue in practice because 0-RTT packets are likely to sent, and hence received, immediately after the first Initial, and might hence routinely be processed by the `Endpoint` before the application has even had a chance to see the `Incoming`.

I think the strategy here is on the right track, but we should consider carefully what the default buffering limit should be. While it's true that the default `receive_window` is unlimited, the default *stream* receive window is not, and neither is the default stream concurrency limit, so there are effective limits by default, and we should preserve that pattern here.

quinn-proto/src/endpoint.rs
gretchenfrage: There's some overhead, right? For packet and frame headers, crypto data, etc? Genuine question, tell me if I'm wrong. So I figured for safety: take the maximum amount of stream data allowed to be received, double it to be safe, and also lower-bound it at 10 KB for edge cases where the initial maximum stream data is configured so low that it's mostly overhead.

quinn-proto/src/endpoint.rs
gretchenfrage: Good idea.

quinn-proto/src/endpoint.rs
Ralith: There is some overhead, but it'll tend to be a single-digit percentage of the packet. At most I'd round up to the next packet size. If the sender wants to pack packets wastefully, then wasted space is ultimately on them, anyway.

Ralith: A clever way to round up to the next packet size without having to think too hard about what that *is* would be to drop a new packet iff the *existing* buffer is already above the limit.

gretchenfrage:
> This was very likely to be an issue in practice because 0-RTT packets are likely to sent, and hence received, immediately after the first Initial, and might hence routinely be processed by the `Endpoint` before the application has even had a chance to see the Incoming.

Indeed, I discovered this not through thinking about the code really hard but through empirical testing. I was a bit embarrassed when I realized that the problem was caused by me, I initially thought it was pre-existing.

fd22d144b8b65d9b

---

> How do we limit the number of these? Buffer size limits will not be an effective defense if the number of buffers is unbounded.

Good catch. And solving this is kind of tricky. But here's an approach I've implemented now that I think works, let me know what you think:

Firstly, we can now move the `MAX_INCOMING_CONNECTIONS` check from quinn to proto. (Also, rename it to just `MAX_INCOMING`, which we sort of forgot to do in the original PR).

However, that is not sufficient to prevent memory exhaustion via filling many incoming with buffered early data, because these limits multiply to too high of a number. With the default transport config:

(100 max concurrent bidi streams + 100 max concurrent uni streams) × 1.25 MB stream receive window = 250 MB

× 2^16 max incoming = 1.64 TB

Which I'm pretty sure is a lot of RAM. So to avoid that, we implement another `MAX_ALL_INCOMING_BUFFERED` limit set to 100 MiB which is a limit to the total number of bytes buffered for incomings across all incoming within the connection. If either the per-incoming limit or the `MAX_ALL_INCOMING_BUFFERED` limit are exceeded, the packet is dropped.

So to summarize:

- Upon receipt of an incoming-_creating_ packet, if the number of pending incoming exceeds `MAX_INCOMING` = 1^16, endpoint automatically responds with refuse packet.
- Upon receipt of an incoming-_associated_ packet, if the number of buffered packet bytes for that incoming exceeds the per-incoming limit calculated as `min(receive_window, (max_concurrent_bidi_streams + max_concurrent_uni_streams) * stream_receive_window)`, endpoint drops the packet.
- Upon receipt of an incoming-_associated_ packet, if the number of buffered packet bytes for all incoming collectively exceeds `MAX_ALL_INCOMING_BUFFERED` = 100 MiB, endpoint drops the packet.

quinn-proto/src/endpoint.rs
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
gretchenfrage: See comment below.

gretchenfrage:
I'm not sure how to debug the freebsd CI failure, I don't see logs for it

Ralith:
That's weird, normally it logs just fine. GitHub infra flake, perhaps? I've restarted it manually.

gretchenfrage:
Thanks. It passed this time, so I guess it was just a spurious CI failure.

quinn-proto/src/endpoint.rs
```diff
@@ -1063,10 +1185,26 @@ impl fmt::Debug for Incoming {
...
+
+impl IncomingImproperDropWarner {
+    fn dismiss(self) {
+        mem::forget(self);
+    }
+}
+
+impl Drop for IncomingImproperDropWarner {
+    fn drop(&mut self) {
+        warn!("quinn_proto::Incoming dropped without passing to Endpoint::accept/refuse/retry/ignore (may cause memory leak)");
```
Ralith, line 1204: More seriously, repeated drops could eventually lead to inability to accept new connections.

quinn-proto/src/endpoint.rs
```diff
@@ -188,17 +195,52 @@ impl Endpoint {
...
+
+                    // By checking the buffer limit before rather than after adding the current
+                    // datagram size, we naturally allow for around one datagram worth of QUIC
+                    // overhead before we start dropping packets.
+
+                    if incoming_buffer.total_bytes <= buffer_limit
+                        && self
+                            .all_incoming_buffers_total_bytes
+                            .saturating_add(datagram_len as u64)
+                            < MAX_ALL_INCOMING_BUFFERED
```
Ralith, line 226: It's confusing that this clause checks after adding the next datagram whereas the first clause checks before.

quinn-proto/src/endpoint.rs
```diff
@@ -188,17 +195,52 @@ impl Endpoint {
...
+                        config
+                            .max_concurrent_bidi_streams
+                            .into_inner()
+                            .saturating_add(config.max_concurrent_uni_streams.into_inner())
+                            .saturating_mul(config.stream_receive_window.into_inner()),
+                    );
+
+                    // By checking the buffer limit before rather than after adding the current
+                    // datagram size, we naturally allow for around one datagram worth of QUIC
+                    // overhead before we start dropping packets.
```
Ralith, line 220: Rounding up to the next packet isn't necessarily about allowing for overhead (which might vary dramatically according to sender discretion), just erring on the side of being permissive.

quinn-proto/src/endpoint.rs
```diff
@@ -188,17 +195,52 @@ impl Endpoint {
...
+
+                    if incoming_buffer.total_bytes <= buffer_limit
```
Ralith, line 222: If we use `<` here, then 0 has a more intuitive effect.

quinn-proto/src/endpoint.rs
```diff
@@ -50,6 +50,10 @@ pub struct Endpoint {
     allow_mtud: bool,
     /// Time at which a stateless reset was most recently sent
     last_stateless_reset: Option<Instant>,
+    /// Buffered Initial and 0-RTT messages for pending incoming connections
+    incoming_buffers: Slab<IncomingBuffer>,
+    incoming_count: usize,
```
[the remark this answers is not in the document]
gretchenfrage: Mistake on my part. I was aware that slab doesn't maintain a free list and thus has O(used entries + free entries) iteration time. My brain made the intuitive but incorrect leap from that to thinking that `.len()` must return `used entries + free entries`.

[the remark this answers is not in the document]
gretchenfrage: Field removed.

quinn-proto/src/endpoint.rs
gretchenfrage: Warning updated.

gretchenfrage:
> Should we expose the new limits on ServerConfig? 100MiB makes sense for a lot of applications, but someone might want to use Quinn in a memory-constrained environment (e.g. a router).

Good idea. Now that this is all happening in proto, there's no friction to us doing that. Added three new settings to `ServerConfig`: `max_incoming`, `max_buffer_bytes_per_incoming`, and `max_buffer_bytes_all_incoming`.

> What if a connection is configured for use with application datagrams (or some other future messaging primitive) only, and stream limits are hence set to zero? Might be simplest to only use a dedicated limit.

I made it so the default `ServerConfig` constructor calculates `max_buffer_bytes_per_incoming` as was previously calculated there. Although I did switch it to just hard-code the overhead to 1200 bytes--let me know if you want me to not do it like that, but it seems more straightforward than the trick with comparison. Anyways, if a user configures the stream limits to zero, but still wants 0-RTT datagrams in excess of 1200 bytes to work more-so than they otherwise would, they can manually set `max_buffer_bytes_per_incoming` to some appropriately high number. Let me know if you think this is a pitfall sufficiently likely to be worth documenting.

> It's confusing that this clause checks after adding the next datagram whereas the first clause checks before.

> Rounding up to the next packet isn't necessarily about allowing for overhead (which might vary dramatically according to sender discretion), just erring on the side of being permissive.

> If we use < here, then 0 has a more intuitive effect.

I believe these should all have been fixed in this change.

quinn-proto/src/endpoint.rs
gretchenfrage: Should be addressed, see comment below.

quinn-proto/src/endpoint.rs
gretchenfrage: Should be addressed, see comment below.

quinn-proto/src/endpoint.rs
gretchenfrage: Should be addressed, see comment below.

quinn-proto/src/endpoint.rs
```diff
@@ -188,17 +195,52 @@ impl Endpoint {
...
+            match route_to {
+                RouteDatagramTo::Incoming(incoming_idx) => {
+                    let incoming_buffer = &mut self.incoming_buffers[incoming_idx];
+                    let config = &self.server_config.as_ref().unwrap().transport;
+                    let buffer_limit = config.receive_window.into_inner().min(
+                        config
+                            .max_concurrent_bidi_streams
+                            .into_inner()
+                            .saturating_add(config.max_concurrent_uni_streams.into_inner())
+                            .saturating_mul(config.stream_receive_window.into_inner()),
```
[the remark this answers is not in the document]
gretchenfrage: Should be addressed, see comment below.

quinn-proto/src/config.rs
```diff
@@ -836,6 +873,45 @@ impl ServerConfig {
         self.preferred_address_v6 = address;
         self
     }
+
+    /// Maximum number of [`Incoming`][crate::Incoming] to allow to exist at a time before
+    /// rejecting new [`Incoming`][crate::Incoming] automatically.
+    ///
+    /// Should hover around zero unless connection-creating packets are being received faster than
+    /// the application is calling [`Endpoint::accept`][crate::Endpoint::accept] and dealing with
+    /// them.
```
Ralith, line 882: - A quick skim of this wording seems to suggest that it should be set to a value around 0, which is incorrect.
- Whether the number *of incoming connections* hovers around zero or not is heavily dependent on the latency with which the application layer handles them, which we can't predict.
- Incoming connections might be consumed by means other than `accept`.

How about something like:

```suggestion
    /// Maximum permitted number of [`Incoming`][crate::Incoming] connections that have not yet been [`accept`][crate::Endpoint::accept]ed or otherwise disposed
    ///
    /// While this limit is reached, new incoming connection attempts will be immediately refused. Larger values have greater worst-case memory consumption, but allow the application layer more time to handle an incoming connection attempt.
```

plus mention and justification of the default value to provide a starting point for tuning.

quinn-proto/src/config.rs
```diff
@@ -836,6 +873,45 @@ impl ServerConfig {
...
+        self.max_incoming = max_incoming;
+        self
+    }
+
+    /// Maximum number of received bytes to buffer for each [`Incoming`][crate::Incoming] between
+    /// when the [`Incoming`][crate::Incoming] is created and when it is used to
+    /// accept/refuse/retry/ignore the connection attempt (not including the first packet).
+    ///
+    /// Packets received in excess of this limit are dropped. This may cause 0-RTT data to have to
+    /// be retransmitted, and, in some cases, failure of the handshake to complete.
```
Ralith, line 893: I don't see why this should cause connection failure except in the same sense that failing to handle an `Incoming` in a timely fashion would, which probably doesn't bear repeating here.

quinn-proto/src/config.rs
```diff
@@ -836,6 +873,45 @@ impl ServerConfig {
...
+    pub fn max_buffer_bytes_per_incoming(
```
Ralith, line 894: bikeshed for improved consistency with e.g. `crypto_buffer_size`, `datagram_receive_buffer_size`:
```suggestion
    pub fn incoming_buffer_size(
```

quinn-proto/src/config.rs
```diff
@@ -836,6 +873,45 @@ impl ServerConfig {
...
+        self
+    }
+
+    /// Maximum number of received bytes to buffer for all [`Incoming`][crate::Incoming]
+    /// collectively between when the [`Incoming`][crate::Incoming] is created and when it is used
+    /// to accept/refuse/retry/ignore the connection attempt (not including the first packet).
+    ///
+    /// Packets received in excess of this limit are dropped. This may cause 0-RTT data to have to
+    /// be retransmitted, and, in some cases, failure of the handshake to complete.
+    pub fn max_buffer_bytes_all_incoming(
```
Ralith, line 908: bikeshed:
```suggestion
    pub fn incoming_buffer_size_total(
```

quinn-proto/src/endpoint.rs
```diff
@@ -188,17 +193,46 @@ impl Endpoint {
...
+                            .saturating_add(datagram_len as u64)
+                            < config.max_buffer_bytes_all_incoming
+                    {
+                        incoming_buffer.datagrams.push(event);
+                        incoming_buffer.total_bytes = incoming_buffer
+                            .total_bytes
+                            .saturating_add(datagram_len as u64);
+                        self.all_incoming_buffers_total_bytes += self
+                            .all_incoming_buffers_total_bytes
+                            .saturating_add(datagram_len as u64);
```
Ralith, line 224: This doubles the current value on each operation, irreversibly. Are we missing a unit test that would have caught that?

quinn-proto/src/endpoint.rs
```diff
@@ -188,17 +193,46 @@ impl Endpoint {
...
+                        < config.max_buffer_bytes_per_incoming
+                        && self
+                            .all_incoming_buffers_total_bytes
+                            .saturating_add(datagram_len as u64)
+                            < config.max_buffer_bytes_all_incoming
+                    {
+                        incoming_buffer.datagrams.push(event);
+                        incoming_buffer.total_bytes = incoming_buffer
+                            .total_bytes
+                            .saturating_add(datagram_len as u64);
```
Ralith, line 221: `saturating_add` here and below is superfluous, because any summation that would wrap would have produced `u64::MAX` in the condition above and necessarily failed a `<` test. I wouldn't mind keeping it for symmetry, but just using `+= datagram_len as u64` is more readable and avoids bugs like below.

quinn-proto/src/config.rs
```diff
@@ -782,8 +806,17 @@ impl ServerConfig {
...
     ) -> Self {
+        let transport = Arc::new(TransportConfig::default());
+        let max_buffer_bytes_per_incoming = transport.receive_window.into_inner().min(
+            transport
+                .max_concurrent_bidi_streams
+                .into_inner()
+                .saturating_add(transport.max_concurrent_uni_streams.into_inner())
+                .saturating_mul(transport.stream_receive_window.into_inner())
+                .saturating_add(1200),
+        );
```
Ralith, line 817: This is complex and prone to misconfiguration (e.g. datagram-only connection). Maybe just set it to, say, 1/10th of the default global max?

quinn-proto/src/config.rs
```diff
@@ -793,6 +826,10 @@ impl ServerConfig {
 
             preferred_address_v4: None,
             preferred_address_v6: None,
+
+            max_incoming: 1 << 16,
+            max_buffer_bytes_per_incoming,
+            max_buffer_bytes_all_incoming: 100 << 20,
```
Ralith, line 832: Document/justify these defaults so that it's easier to revisit them in the future if need be.

Ralith (changes requested):

quinn-proto/src/endpoint.rs
gretchenfrage: :(

quinn-proto/src/config.rs
```diff
@@ -774,6 +774,30 @@ pub struct ServerConfig {
...
+    /// be retransmitted, and, in some cases, failure of the handshake to complete.
+    pub(crate) max_buffer_bytes_per_incoming: u64,
+
+    /// Maximum number of received bytes to buffer for all [`Incoming`][crate::Incoming]
+    /// collectively between when the [`Incoming`][crate::Incoming] is created and when it is used
+    /// to accept/refuse/retry/ignore the connection attempt (not including the first packet).
+    ///
+    /// Packets received in excess of this limit are dropped. This may cause 0-RTT data to have to
+    /// be retransmitted, and, in some cases, failure of the handshake to complete.
+    pub(crate) max_buffer_bytes_all_incoming: u64,
```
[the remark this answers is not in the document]
gretchenfrage: Removed.

quinn-proto/src/config.rs
gretchenfrage: > - A quick skim of this wording seems to suggest that it should be set to a value around 0, which is incorrect.

I see how it could be interpreted as that, yeah.

> - Whether the number _of incoming connections_ hovers around zero or not is heavily dependent on the latency with which the application layer handles them, which we can't predict.

Ah, yeah, that's an error on my part.

> - Incoming connections might be consumed by means other than `accept`.

The method I linked was `Endpoint::accept`, not `Incoming::accept`, so it would still be on the path to refuse/retry/ignore. But it is a confusion hazard, yeah.

I largely rewrote the docs for these three settings, I think it should be more clear and correct now. Let me know what you think.

quinn-proto/src/config.rs
gretchenfrage: You mentioned before the possibility of receiving Initial packets after the first initial packet, rather than solely just 0-RTT packets. My understanding isn't entirely clear on how that works and whether the loss of those packets could lead to handshake failure. I probably should've highlighted it more as something I was uncertain about. I removed that claim from the docs for now.

quinn-proto/src/config.rs
gretchenfrage: Done.

quinn-proto/src/config.rs
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
gretchenfrage: I've reworked this logic to be based on `checked_add` now--I think it's a lot more intuitive and thus easier to read and less prone to these kids of programming mistakes and gives me more confidence that there's not some edge case.

A unit test that would've caught that would also have probably caught this bug in the first place when we were adding the Incoming API, given that their consequence is not connection failure outright but rather the 0-RTT feature appearing to engage successfully but not actually in practice achieving a request/response cycle in 1 round trip. I think that would be a good thing to do, but I'm not sure how easy it would be to do, and to what degree it would require essentially creating a basic network emulator. I might be really easy.

quinn-proto/src/endpoint.rs
gretchenfrage: Reworked this logic as described in comment above. Problem should be fixed.

quinn-proto/src/config.rs
gretchenfrage: Done.

quinn-proto/src/config.rs
gretchenfrage: Documented defaults and their reasons in the method docs.

gretchenfrage:
Thanks for the feedback. My last round of changes was a bit sloppier than I would've liked it to be, especially with me missing the `+=` bug. I think the feedback should be addressed now though.

- Reworked the limit checking logic to be based on `checked_add` in a way I think is a lot cleaner.
- Largely rewrote the added config docs.
- Made other more minor changes.

quinn-proto/src/endpoint.rs
```diff
@@ -1059,10 +1167,27 @@ impl fmt::Debug for Incoming {
             // rest is too big and not meaningful enough
             .field("retry_src_cid", &self.retry_src_cid)
             .field("orig_dst_cid", &self.orig_dst_cid)
+            .field("incoming_idx", &self.incoming_idx)
+            // improper drop warner contains no information
             .finish_non_exhaustive()
     }
 }
 
+struct IncomingImproperDropWarner;
```
djc, line 1176: By destructuring `Incoming` at its consumption site, can we implement the warning directly on `Incoming::drop()`? If not, can we use destructuring instead of `mem::forget()` inside `IncomingImproperDropWarner::dismiss()`?

quinn-proto/src/shared.rs
```diff
@@ -11,17 +11,21 @@ pub struct ConnectionEvent(pub(crate) ConnectionEventInner);
...
-        first_decode: PartialDecode,
-        remaining: Option<BytesMut>,
-    },
+    Datagram(DatagramConnectionEvent),
     /// New connection identifiers have been issued for the Connection
     NewIdentifiers(Vec<IssuedCid>, Instant),
 }
 
+/// Variant of [`ConnectionEventInner`].
+#[derive(Debug)]
```
djc, line 20: Would be nice to make this a separate commit.

quinn-proto/src/endpoint.rs
```diff
@@ -482,6 +512,10 @@ impl Endpoint {
             }
         };
 
+        let incoming_idx = self.incoming_buffers.insert(Default::default());
```
djc, line 515: Please avoid `Default::default()` in favor of using the type name.

quinn-proto/src/config.rs
```diff
@@ -838,6 +846,54 @@ impl ServerConfig {
...
+    /// An [`Incoming`][crate::Incoming] comes into existence when an incoming connection attempt
+    /// is received and stops existing when the application either accepts it or otherwise disposes
+    /// of it. This limit governs only packets received within that period, and does not include
+    /// the first packet. Packets received in excess of this limit are dropped, which may cause
+    /// 0-RTT data to have to be retransmitted.
+    ///
+    /// The default value is set to 10 MiB--an amount such that in most situations a client would
+    /// not transmit that much 0-RTT data faster than the server handles the corresponding
+    /// [`Incoming`][crate::Incoming].
+    pub fn incoming_buffer_size(&mut self, incoming_buffer_size: u64) -> &mut Self {
```
djc, line 877: Do we need/want this? Seems to me most servers will mostly/only care about `_total`?

djc (commented):
Some minorish feedback.

This all seems like substantial complexity. How do we feel about alternative solutions where we force callers to address one `Incoming` at a time? Do we really need the flexibility of keeping multiple incoming packets in flight at the same time? It feels like a larger change that is sort of flying in here under the radar as a very flexible solution to a problem (the race condition identified in #1820), with the flexibility also causing an increase in fragility.

(The answer might well be yes, but wanted to make this more explicit anyway.)

Minor nit: with the frequent force-pushes, keeping an issue reference in the commit message is a bit of a pain because it adds a whole bunch of backreferences in the issue. Consider leaving the issue reference out of the commit message in the future, in favor of keeping it in the PR description.

quinn-proto/src/endpoint.rs
gretchenfrage: Types which implement drop cannot have their ownership taken by destructuring

```rs
// this compiles:
struct A(Box<i32>);
fn a(a: A) {
    let A(i) = a;
}

// this does not compile:
/*error[E0509]: cannot move out of type `B`, which implements the `Drop` trait
  --> src/lib.rs:12:16
   |
12 |     let B(i) = b;
   |           -    ^ cannot move out of here
   |           |
   |           data moved here
   |           move occurs because `i` has type `Box<i32>`, which does not implement the `Copy` trait
   |
help: consider borrowing the pattern binding
   |
12 |     let B(ref i) = b;
   |           +++*/
struct B(Box<i32>);
impl Drop for B { fn drop(&mut self) {} }
fn b(b: B) {
    let B(i) = b;
}
```

(This is what you're referring to, right?)

quinn-proto/src/config.rs
gretchenfrage: I'm not sure. I might be off on this comparison, but it seems abstractly analogous to the transport parameters having an initial max data limit both per each stream and for all streams in total.

Earlier, before concerns around interaction with raw datagrams were raised, I was computing the default for this limit based on the maximum valid amount of stream data bytes a 0-rtt client could transmit.

The scenarios I'm trying to imagine where it's useful _are_ seeming rather esoteric. I wonder what @Ralith thinks about whether this limit is useful enough to justify its existence?

gretchenfrage:
> This all seems like substantial complexity. How do we feel about alternative solutions where we force callers to address one `Incoming` at a time? Do we really need the flexibility of keeping multiple incoming packets in flight at the same time? It feels like a larger change that is sort of flying in here under the radar as a very flexible solution to a problem (the race condition identified in #1820), with the flexibility also causing an increase in fragility.

It's a worthwhile question.

Merely limiting there to being only one `Incoming` at a time would not let us fully avoid buffering early datagrams not associated with a connection handle. It would allow us to indiscriminately put all early datagrams not associated with a connection handle into a single buffer rather than in per-`Incoming` buffers, but that doesn't seem like much of a simplification.

To avoid having to buffer early datagrams not associated with a connection handle, we would need to be able make a decision whether to accept/refuse/retry/ignore an incoming connection attempt immediately and synchronously when the endpoint receives the connection-creating packet and before it tries to process any further packets (as the decision affects which subsequently received packets get routed to a connection and which get discarded).

This could be achieved by removing the `Incoming` API as it exists now and instead just putting some sort of `incoming_logic: Box<dyn FnMut(&IncomingConnectionInfo) -> IncomingConnectionDecision>` field in the server config which the endpoint can call synchronously. I don't think we should do that.

One example of a situation where allowing multiple `Incoming` to be in motion at the same time would be useful is if there is an IP block list (or allow list) that's stored in a database but has an in-memory Bloom filter to accelerate it:

```rs
while let Some(incoming) = endpoint.accept().await {
    let ip = incoming.remote_address().ip();
    if !block_list_bloom_filter.maybe_contains(ip) {
        task::spawn(async move {
            handle_incoming(incoming.accept().expect("TODO put real error handling here")).await;
        });
    } else {
        task::spawn(async move {
            if !block_list_database.contains(ip).await {
                handle_incoming(incoming.accept().expect("TODO put real error handling here")).await;
            }
        });
    }
}
```

This would be a situation where allowing multiple `Incoming` in motion simultaneously, and even allowing the application to make decisions on them in a different order than they were produced, could improve performance and/or attack mitigation effectiveness.

quinn-proto/src/endpoint.rs
gretchenfrage: Done.

quinn-proto/src/shared.rs
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
djc: Ah yes -- thanks for trying!

djc:
> To avoid having to buffer early datagrams not associated with a connection handle, we would need to be able make a decision whether to accept/refuse/retry/ignore an incoming connection attempt immediately and synchronously when the endpoint receives the connection-creating packet and before it tries to process any further packets (as the decision affects which subsequently received packets get routed to a connection and which get discarded).

I was thinking we might have `accept()` take ownership of the `Endpoint`, storing the `Endpoint` in `Incoming`, and "giving it back" after accept/reject/ignore. This would still allow the caller to handle incoming connections asynchronously, but not in parallel (thus avoiding buffering issues). It's less flexible but simpler in the end.

quinn-proto/src/config.rs
Ralith: Having separate limits is desirable in principle because it enables better behavior when a single badly behaved connection is saturating the buffer. With a single global limit, it's easier for an attacker to cause 0-RTT or initial packets for other connections to be lost, even when the total limit is the same.

Maybe offering a separate knob is overkill in this case. Well-behaved servers should consume `Incoming`s in a timely manner, so the period for which any single badly behaved connection can saturate the buffer is short. We could leave this off for simplicity's sake in the absence of specific demand; I don't have a strong opinion at the moment.

djc: Okay, given that we (mostly @gretchenfrage!) have already done the work and that additional complexity is probably limited, let's keep it for now.

Ralith:
> have accept() take ownership of the Endpoint

This would complicate the async layer considerably. We would need to move the `proto::Endpoint` out of the `quinn::endpoint::State` each time a connection is incoming, then restore it and wake the endpoint driver after a decision is made. It's also difficult to allow such an `accept` to be used concurrently with any other endpoint access (e.g. `connect`, `close`, assorted getters). Because typical servers will always be waiting on `accept`, surfacing such a limitation to the public API is likely to degrade usability.

More importantly, it wouldn't solve anything: between GRO and recvmmsg, we may receive many datagrams instantaneously. One batch might include both the first datagram for an incoming connection and follow-up initial or 0-RTT packets for that same connection. These must be either buffered or (as in the status quo) dropped, and it's most convenient to do so at the proto layer, where we at least have the option to correlate them and discard non-QUIC datagrams.

Finally, if we *could* suspend receipt of datagrams immediately after receiving a connection's first datagram, that would be at odds with future work to parallelize datagram receipt and other endpoint driver work, which is our major remaining milestone for intra-endpoint scaling.

In sum, the current form of this PR is strongly motivated.

quinn-proto/src/config.rs
```diff
@@ -838,6 +846,54 @@ impl ServerConfig {
...
+
+    /// Maximum number of [`Incoming`][crate::Incoming] to allow to exist at a time
+    ///
+    /// An [`Incoming`][crate::Incoming] comes into existence when an incoming connection attempt
+    /// is received and stops existing when the application either accepts it or otherwise disposes
+    /// of it. While this limit is reached, new incoming connection attempts are immediately
+    /// refused. Larger values have greater worst-case memory consumption, but accommodate greater
+    /// application latency in handling incoming connection attempts.
+    ///
+    /// The default value is set to 65536. Assuming each [`Incoming`][crate::Incoming] accounts for
```
Ralith, line 858: "Assuming ..." isn't very helpful to the user, because they don't know whether they should assume that or not. Maybe reword in terms of the the MTU? Something like "With a typical Ethernet MTU of 1500 bytes, this implies a worst-case memory consumption under 100MiB".

quinn-proto/src/config.rs
```diff
@@ -838,6 +846,54 @@ impl ServerConfig {
...
+        self
+    }
+
+    /// Maximum number of received bytes to buffer for each [`Incoming`][crate::Incoming]
+    ///
+    /// An [`Incoming`][crate::Incoming] comes into existence when an incoming connection attempt
+    /// is received and stops existing when the application either accepts it or otherwise disposes
+    /// of it. This limit governs only packets received within that period, and does not include
+    /// the first packet. Packets received in excess of this limit are dropped, which may cause
+    /// 0-RTT data to have to be retransmitted.
```
Ralith, line 872: "0-RTT or handshake data", or maybe just "may temporarily delay the connection".

quinn-proto/src/config.rs
```diff
@@ -838,6 +846,54 @@ impl ServerConfig {
...
+    }
+
+    /// Maximum number of received bytes to buffer for all [`Incoming`][crate::Incoming]
+    /// collectively
+    ///
+    /// An [`Incoming`][crate::Incoming] comes into existence when an incoming connection attempt
+    /// is received and stops existing when the application either accepts it or otherwise disposes
+    /// of it. This limit governs only packets received within that period, and does not include
+    /// the first packet. Packets received in excess of this limit are dropped, which may cause
+    /// 0-RTT data to have to be retransmitted.
```
Ralith, line 889: As above

quinn-proto/src/config.rs
Ralith: In short, everything important is subject to retransmission until the connection times out completely. A lot of thought was put into the handshake process to avoid situations where transient packet loss could cause a deadlock.

Ralith (changes requested):
Implementation LGTM. Remaining items:
- Some final nits on the docs
- ~Decide whether to discard per-`Incoming` buffer limits for simplicity~
- Test coverage

quinn-proto/src/config.rs
gretchenfrage: I like specifying the specific types of packets because it seems noteworthy that in some cases (specific patterns of handshakes and no early data) it might be fine to set this to 0.

quinn-proto/src/config.rs
gretchenfrage: Done.

quinn-proto/src/config.rs
gretchenfrage: Done.

quinn-proto/src/endpoint.rs
gretchenfrage: Done.

gretchenfrage:
Comments tweaked. Tests added. As suggested, new tests are based on `zero_rtt_happypath` test and validate the number of dropped packets. Added to `TestEndpoint` / `IncomingConnectionBehavior` the ability to have `Incoming` be pushed to a `waiting_incoming` vec to be dealt with later rather than dealt with immediately.

quinn-proto/src/tests/mod.rs
```diff
@@ -666,6 +666,114 @@ fn zero_rtt_rejection() {
...
+    // the first packet is stored in the Incoming
+    // the next three are incoming-buffered, bringing the incoming buffer size to 3600 bytes
+    // the last four are dropped due to the buffering limit and must be retransmitted
+    const EXPECTED_DROPPED: u64 = 4;
+
+    let _guard = subscribe();
+    let mut server_config = server_config();
+    configure_server(&mut server_config);
+    let mut pair = Pair::new(Arc::new(EndpointConfig::default()), server_config);
+    pair.server.incoming_connection_behavior = IncomingConnectionBehavior::Validate;
```
Ralith, line 683: Do we need to be overriding the default here?

Ralith (approved):

quinn-proto/src/tests/mod.rs
gretchenfrage: I'm not sure why it would be important. It was just copied from the `zero_rtt_happypath` test, so I'd be curious why it was originally added there.

Ralith: Digging through history, it looks like it was added in 3b01662324a14b605f5e7f54922c445212835b9f, to give test coverage for the case where the client must retransmit 0-RTT when responding to a Retry during a 0-RTT connection attempt. I think this is valuable for the happypath test, but unnecessary here, especially since using `Wait` for the second (0-RTT) connection attempt undermines it anyway.

gretchenfrage: Removed.

djc:
Thanks for all the effort here!

PR #1826, opened 2024-04-16
Revise and add additional 0-rtt doc comments

gretchenfrage:
I had a hard time getting 0-RTT actually working in practice. It escalated to me forking rustls so I could put in a bunch of print statements. Although I did find the 0-rtt tests, there were some subtle enough pitfalls that I didn't notice the relevant ways my attempts were diverging.

In hopes of improving understandability and discoverability of this, this PR adds notes about these pitfalls to some of the doc comments, and also largely rewrites the doc comment for `Connecting::into_0rtt` to make its behavior more clear,  especially across various edge cases.

This PR does not change any code or APIs.

### DIFF
--- a/quinn-proto/src/crypto/rustls.rs
+++ b/quinn-proto/src/crypto/rustls.rs
@@ -257,10 +257,19 @@ pub struct HandshakeData {
 
 /// A QUIC-compatible TLS client configuration
 ///
-/// Can be constructed via [`ClientConfig::with_root_certificates()`][root_certs],
-/// [`ClientConfig::with_platform_verifier()`][platform] or by using the [`TryFrom`] implementation with a
-/// custom [`rustls::ClientConfig`]. A pre-existing `ClientConfig` must have TLS 1.3 support enabled for
-/// this to work. 0-RTT support is available if `enable_early_data` is set to `true`.
+/// Quinn implicitly constructs a `QuicClientConfig` with reasonable defaults within
+/// [`ClientConfig::with_root_certificates()`][root_certs] and [`ClientConfig::with_platform_verifier()`][platform].
+/// Alternatively, `QuicClientConfig`'s [`TryFrom`] implementation can be used to wrap around a
+/// custom [`rustls::ClientConfig`], in which case care should be taken around certain points:
+///
+/// - If `enable_early_data` is not set to true, then sending 0-RTT data will not be possible on
+///   outgoing connections.
+/// - The [`rustls::ClientConfig`] must have TLS 1.3 support enabled for conversion to succeed.
+///
+/// The object in the `resumption` field of the inner [`rustls::ClientConfig`] determines whether
+/// calling `into_0rtt` on outgoing connections returns `Ok` or `Err`. It typically allows
+/// `into_0rtt` to proceed if it recognizes the server name, and defaults to an in-memory cache of
+/// 256 server names.
 ///
 /// [root_certs]: crate::config::ClientConfig::with_root_certificates()
 /// [platform]: crate::config::ClientConfig::with_platform_verifier()
@@ -386,10 +395,14 @@ impl std::error::Error for NoInitialCipherSuite {}
 
 /// A QUIC-compatible TLS server configuration
 ///
-/// Can be constructed via [`ServerConfig::with_single_cert()`][single] or by using the
-/// [`TryFrom`] implementation with a custom [`rustls::ServerConfig`]. A pre-existing
-/// `ServerConfig` must have TLS 1.3 support enabled for this to work. 0-RTT support is
-/// available to clients if `max_early_data_size` is set to `u32::MAX`.
+/// Quinn implicitly constructs a `QuicServerConfig` with reasonable defaults within
+/// [`ServerConfig::with_single_cert()`][single]. Alternatively, `QuicServerConfig`'s [`TryFrom`]
+/// implementation or `with_initial` method can be used to wrap around a custom
+/// [`rustls::ServerConfig`], in which case care should be taken around certain points:
+///
+/// - If `max_early_data_size` is not set to `u32::MAX`, the server will not be able to accept
+///   incoming 0-RTT data. QUIC prohibits `max_early_data_size` values other than 0 or `u32::MAX`.
+/// - The `rustls::ServerConfig` must have TLS 1.3 support enabled for conversion to succeed.
 ///
 /// [single]: crate::config::ServerConfig::with_single_cert()
 pub struct QuicServerConfig {
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -79,29 +79,48 @@ impl Connecting {
 
     /// Convert into a 0-RTT or 0.5-RTT connection at the cost of weakened security
     ///
-    /// Opens up the connection for use before the handshake finishes, allowing the API user to
-    /// send data with 0-RTT encryption if the necessary key material is available. This is useful
-    /// for reducing start-up latency by beginning transmission of application data without waiting
-    /// for the handshake's cryptographic security guarantees to be established.
+    /// Returns `Ok` immediately if the local endpoint is able to attempt sending 0/0.5-RTT data.
+    /// If so, the returned [`Connection`] can be used to send application data without waiting for
+    /// the rest of the handshake to complete, at the cost of weakened cryptographic security
+    /// guarantees. The returned [`ZeroRttAccepted`] future resolves when the handshake does
+    /// complete, at which point subsequently opened streams and written data will have full
+    /// cryptographic protection.
     ///
-    /// When the `ZeroRttAccepted` future completes, the connection has been fully established.
+    /// ## Outgoing
     ///
-    /// # Security
+    /// For outgoing connections, the initial attempt to convert to a [`Connection`] which sends
+    /// 0-RTT data will proceed if the [`crypto::ClientConfig`][crate::crypto::ClientConfig]
+    /// attempts to resume a previous TLS session. However, **the remote endpoint may not actually
+    /// _accept_ the 0-RTT data**--yet still accept the connection attempt in general. This
+    /// possibility is conveyed through the [`ZeroRttAccepted`] future--when the handshake
+    /// completes, it resolves to true if the 0-RTT data was accepted and false if it was rejected.
+    /// If it was rejected, the existence of streams opened and other application data sent prior
+    /// to the handshake completing will not be conveyed to the remote application, and local
+    /// operations on them will return `ZeroRttRejected` errors.
     ///
-    /// On outgoing connections, this enables transmission of 0-RTT data, which might be vulnerable
-    /// to replay attacks, and should therefore never invoke non-idempotent operations.
+    /// A server may reject 0-RTT data at its discretion, but accepting 0-RTT data requires the
+    /// relevant resumption state to be stored in the server, which servers may limit or lose for
+    /// various reasons including not persisting resumption state across server restarts.
     ///
-    /// On incoming connections, this enables transmission of 0.5-RTT data, which might be
-    /// intercepted by a man-in-the-middle. If this occurs, the handshake will not complete
-    /// successfully.
+    /// If manually providing a [`crypto::ClientConfig`][crate::crypto::ClientConfig], check your
+    /// implementation's docs for 0-RTT pitfalls.
     ///
-    /// # Errors
+    /// ## Incoming
     ///
-    /// Outgoing connections are only 0-RTT-capable when a cryptographic session ticket cached from
-    /// a previous connection to the same server is available, and includes a 0-RTT key. If no such
-    /// ticket is found, `self` is returned unmodified.
+    /// For incoming connections, conversion to 0.5-RTT will always fully succeed. `into_0rtt` will
+    /// always return `Ok` and the [`ZeroRttAccepted`] will always resolve to true.
     ///
-    /// For incoming connections, a 0.5-RTT connection will always be successfully constructed.
+    /// If manually providing a [`crypto::ServerConfig`][crate::crypto::ServerConfig], check your
+    /// implementation's docs for 0-RTT pitfalls.
+    ///
+    /// ## Security
+    ///
+    /// On outgoing connections, this enables transmission of 0-RTT data, which is vulnerable to
+    /// replay attacks, and should therefore never invoke non-idempotent operations.
+    ///
+    /// On incoming connections, this enables transmission of 0.5-RTT data, which may be sent
+    /// before TLS client authentication has occurred, and should therefore not be used to send
+    /// data for which client authentication is being used.
     pub fn into_0rtt(mut self) -> Result<(Connection, ZeroRttAccepted), Self> {
         // This lock borrows `self` and would normally be dropped at the end of this scope, so we'll
         // have to release it explicitly before returning `self` by value.
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -189,12 +189,16 @@ impl SendStream {
         conn.inner.send_stream(self.stream).priority()
     }
 
-    /// Completes when the stream is stopped or read to completion by the peer
+    /// Completes when the peer stops the stream or reads the stream to completion
     ///
-    /// Yields `Some` with the stop error code when the stream is stopped by the peer. Yields `None`
-    /// when the stream is [`finish()`](Self::finish)ed locally and all stream data has been
-    /// received (but not necessarily processed) by the peer, after which it is no longer meaningful
-    /// for the stream to be stopped.
+    /// Yields `Some` with the stop error code if the peer stops the stream. Yields `None` if the
+    /// local side [`finish()`](Self::finish)es the stream and then the peer acknowledges receipt
+    /// of all stream data (although not necessarily the processing of it), after which the peer
+    /// closing the stream is no longer meaningful.
+    ///
+    /// For a variety of reasons, the peer may not send acknowledgements immediately upon receiving
+    /// data. As such, relying on `stopped` to know when the peer has read a stream to completion
+    /// may introduce more latency than using an application-level response of some sort.
     pub async fn stopped(&mut self) -> Result<Option<VarInt>, StoppedError> {
         Stopped { stream: self }.await
     }


quinn/src/connection.rs
```diff
@@ -70,29 +70,55 @@ impl Connecting {
...
+    /// existence will typically not be conveyed to the remote application.
     ///
-    /// On incoming connections, this enables transmission of 0.5-RTT data, which might be
-    /// intercepted by a man-in-the-middle. If this occurs, the handshake will not complete
-    /// successfully.
+    /// By default, quinn uses [`rustls::ClientConfig`] as the client crypto implementation. It
+    /// allows the initial attempt to convert to 0-RTT to proceed if its `enable_early_data` field
+    /// is set to true and its `resumption` field recognizes the server name (which defaults to an
+    /// in-memory cache of 256 server names). **If you are manually providing the client crypto
+    /// config, you must set `enable_early_data` to true to get 0-RTT working.**
```
djc, line 98: I don't think this should be documented here.

quinn/src/connection.rs
```diff
@@ -70,29 +70,55 @@ impl Connecting {
...
+    /// ---
     ///
-    /// Outgoing connections are only 0-RTT-capable when a cryptographic session ticket cached from
-    /// a previous connection to the same server is available, and includes a 0-RTT key. If no such
-    /// ticket is found, `self` is returned unmodified.
+    /// For incoming connections, conversion to 0.5-RTT will always fully succeed. `into_0rtt` will
+    /// always return `Ok` and the [`ZeroRttAccepted`] will always resolve to true.
     ///
-    /// For incoming connections, a 0.5-RTT connection will always be successfully constructed.
+    /// By default, quinn uses [`rustls::ServerConfig`] as the server crypto implementation. Its
```
djc, line 106: rustls-specific.

quinn/src/send_stream.rs
```diff
@@ -134,6 +134,17 @@ impl SendStream {
     ///
     /// No new data may be written after calling this method. Completes when the peer has
     /// acknowledged all sent data, retransmitting data as needed.
+    ///
+    /// 0-rtt
+    /// ---
+    ///
+    /// Because 0-RTT packets cannot contain ACKs, if this is called from a 0-RTT stream opened by
+    /// the client, it cannot resolve for at least two round trips after the creation of the
+    /// connection.
```
djc, line 143: Can you briefly describe why "it cannot resolve for at least two round trips"?

djc (commented):
Thanks for fleshing out these docs, great idea!

I have lots of little feedback about how we should document this stuff, but would be happy to get something like this merged.

quinn-proto/src/config.rs
```diff
@@ -778,6 +778,14 @@ pub struct ServerConfig {
 
 impl ServerConfig {
     /// Create a default config with a particular handshake token key
+    ///
+    /// 0-rtt
+    /// ---
+    ///
+    /// In general, where quinn automatically constructs a `ServerConfig`, it uses as the
```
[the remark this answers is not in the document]
Ralith: > I don't think max_early_data_size has to be specifically u32::MAX

It does. [RFC 9001](https://www.rfc-editor.org/rfc/rfc9001.html#name-enabling-0-rtt):

> Servers MUST NOT send the early_data extension with a max_early_data_size field set to any value other than 0xffffffff. A client MUST treat receipt of a NewSessionTicket that contains an early_data extension with any other value as a connection error of type PROTOCOL_VIOLATION.

gretchenfrage:
Thanks for the feedback! Happy to workshop this into something we both think is good.

quinn/src/connection.rs
```diff
@@ -70,29 +70,55 @@ impl Connecting {
...
     ///
-    /// When the `ZeroRttAccepted` future completes, the connection has been fully established.
+    /// Outgoing
+    /// ---
     ///
-    /// # Security
+    /// For outgoing connections, the initial attempt to convert to 0-RTT will proceed if the
+    /// [`crypto::ClientConfig`][crate::crypto::ClientConfig] attempts to resume a previous TLS
+    /// session. However, **the remote endpoint may not actually _accept_ the 0-RTT data**--yet
+    /// still accept the connection attempt in general. This may occur, for example, if the remote
```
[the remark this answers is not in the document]
gretchenfrage: While I agree we should be technically correct, I also think it's pedagogically important here to ground it in a concrete and common-to-encounter example (server restart). If I had slightly less background knowledge here I might not know what a server "losing resumption state" means or whether it was something common or esoteric. Also, if my understanding is correct, the condition described in this sentence could occur in cases other than lost resumption state--even, hypothetically, a server for some reason being programmed to reject early data based on a random coin toss.

I'm trying to think of ways to rephrase this to make it feel more general. Perhaps: "One common reason this may occur is if the remote endpoint has restarted and thus lost its resumption state since the local endpoint last connected to it." I feel like that's a lot more clunky, grammatically convoluted, and lengthy, though. What do you think?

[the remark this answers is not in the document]
djc: Maybe something like:

"However, the server may at its discretion decide to reject 0-RTT data while otherwise accepting the connection. The ability to accept 0-RTT data depends on resumption state being stored in the server, which servers will limit due to any number of reasons. Some servers only store resumption state in memory, such that a restart of the server will reset all resumption state."

quinn/src/send_stream.rs
gretchenfrage: After further inspection of the RFCs, I realized I am confused about this bit. I will pick this back up when I understand better.

quinn-proto/src/config.rs
```diff
@@ -778,6 +778,14 @@ pub struct ServerConfig {
 
 impl ServerConfig {
     /// Create a default config with a particular handshake token key
+    ///
+    /// 0-rtt
```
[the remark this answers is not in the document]
gretchenfrage: Changed in all locations.

quinn-proto/src/config.rs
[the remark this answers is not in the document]
gretchenfrage: All should be addressed.

quinn-proto/src/config.rs
```diff
@@ -899,6 +907,14 @@ pub struct ClientConfig {
 
 impl ClientConfig {
     /// Create a default config with a particular cryptographic config
+    ///
+    /// 0-rtt
+    /// ---
+    ///
+    /// In general, where quinn automatically constructs a `ClientConfig`, it uses as the
```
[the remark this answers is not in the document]
gretchenfrage: Should be addressed.

quinn/src/connection.rs
```diff
@@ -70,29 +70,55 @@ impl Connecting {
...
-    /// for the handshake's cryptographic security guarantees to be established.
+    /// Returns `Ok` immediately if the local endpoint is able to attempt 0/0.5-RTT conversion. If
+    /// so, the returned [`Connection`] can be used to open streams and write data to them without
+    /// waiting for the rest of the handshake to complete, at the cost of weakened cryptographic
+    /// security guarantees. The returned [`ZeroRttAccepted`] future resolves when the handshake
+    /// does complete, at which point subsequently opened streams and written data will have full
+    /// cryptographic protection.
     ///
-    /// When the `ZeroRttAccepted` future completes, the connection has been fully established.
+    /// Outgoing
```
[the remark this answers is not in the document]
gretchenfrage: Done.

quinn/src/connection.rs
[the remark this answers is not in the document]
gretchenfrage: I changed it to something based on that.

quinn/src/connection.rs
gretchenfrage: Moved.

quinn/src/connection.rs
gretchenfrage: Moved.

quinn/src/send_stream.rs
gretchenfrage: I spent a while investigating this, and I was wrong about it. I was able to get it to resolve faster by creating a co-routine in the client that opens a new stream and writes a byte to it every millisecond, to trigger the server's ack threshold sooner. I think the moral is that `stopped` shouldn't be used for latency-sensitive things since acks can be delayed, so I added a note to that effect.

gretchenfrage:
Thanks to some other changes that have been made to the code base since this was last reviewed, some of the "where should this be documented" questions sort of solved themselves. I've shuffled things around, changed some things, and tried to make sure the current revision reflects the feedback given.

Current change set:

- Rustls-specific documentation on `crypto::rustls::ServerConfig` and `crypto::rustls::ClientConfig`
- Non-Rustls-specific documentation `Connection::into_0rtt`
- A warning regarding latency on `SendStream::stopped`

Thanks!

quinn/src/send_stream.rs
Ralith: Right, relying on transport-layer ACKs to drive application-layer behavior is almost always a bad idea.

djc (approved):
This looks great, thanks for spending time on this!

quinn-proto/src/crypto/rustls.rs
```diff
@@ -386,10 +395,15 @@ impl std::error::Error for NoInitialCipherSuite {}
...
-/// `ServerConfig` must have TLS 1.3 support enabled for this to work. 0-RTT support is
-/// available to clients if `max_early_data_size` is set to `u32::MAX`.
+/// Quinn implicitly constructs a `QuicServerConfig` with reasonable defaults within
+/// [`ServerConfig::with_single_cert()`][single]. Alternatively, `QuicServerConfig`'s [`TryFrom`]
+/// implementation or `with_initial` method can be used to wrap around a custom
+/// [`rustls::ServerConfig`], in which case care should be taken around certain points:
+///
+/// - If `max_early_data_size` is not set to `u32::MAX`, the server may "accept" 0-RTT conversion
+///   attempts by clients but have all attempts by the clients to send 0-RTT data blocked until the
+///   handshake completes. QUIC prohibits `max_early_data_size` values other than 0 or `u32::MAX`.
```
Ralith, line 405: I don't think this is correct. If `max_early_data_size` is set to zero, clients may not attempt 0-RTT in future connection attempts at all. See [RFC 9001 § Enabling 0-RTT](https://www.rfc-editor.org/rfc/rfc9001.html#name-enabling-0-rtt). If a server configuration change disables 0-RTT after a client has already cached keys, then 0-RTT data sent with a future connection attempt will be actively rejected (see `WriteError::ZeroRttRejected`), not merely blocked until the handshake completes. This is not to be confused with flow control limits, which may cause 0-RTT to be available but block on use as described here, but this is no different than the 1-RTT behavior.

quinn/src/connection.rs
```diff
@@ -79,29 +79,47 @@ impl Connecting {
 
     /// Convert into a 0-RTT or 0.5-RTT connection at the cost of weakened security
     ///
-    /// Opens up the connection for use before the handshake finishes, allowing the API user to
-    /// send data with 0-RTT encryption if the necessary key material is available. This is useful
-    /// for reducing start-up latency by beginning transmission of application data without waiting
-    /// for the handshake's cryptographic security guarantees to be established.
+    /// Returns `Ok` immediately if the local endpoint is able to attempt 0/0.5-RTT conversion. If
```
Ralith, line 82: "conversion" is strange word choice here. Maybe "communication" or "... able to attempt sending 0/0.5-RTT data"?

quinn/src/connection.rs
```diff
@@ -79,29 +79,47 @@ impl Connecting {
...
+    /// so, the returned [`Connection`] can be used to open streams and write data to them without
```
Ralith, line 83: s/open streams and write data/send application data/

Streams are not the only form of application data supported in 0-RTT. Application datagrams may be supported, and other forms might be added in future extensions.

quinn/src/connection.rs
```diff
@@ -79,29 +79,47 @@ impl Connecting {
...
     ///
-    /// # Security
+    /// For outgoing connections, the initial attempt to convert to 0-RTT will proceed if the
+    /// [`crypto::ClientConfig`][crate::crypto::ClientConfig] attempts to resume a previous TLS
+    /// session. However, **the remote endpoint may not actually _accept_ the 0-RTT data**--yet
+    /// still accept the connection attempt in general. This possibility is conveyed through the
+    /// [`ZeroRttAccepted`] future--when the handshake completes, it resolves to true if the 0-RTT
+    /// data was accepted and false if it was rejected. If it was rejected, streams opened prior
+    /// to the handshake completing will _hang_, and their existence will typically not be conveyed
+    /// to the remote application.
```
Ralith, line 98: Rejected 0-RTT streams should fail immediately with `ZeroRttRejected` errors, not hang. If this isn't happening, please open a bug.

quinn/src/connection.rs
```diff
@@ -79,29 +79,47 @@ impl Connecting {
...
+    /// implementation's docs for 0-RTT pitfalls.
+    ///
+    /// ## Security
+    ///
+    /// On outgoing connections, this enables transmission of 0-RTT data, which is vulnerable to
+    /// replay attacks, and should therefore never invoke non-idempotent operations.
+    ///
+    /// On incoming connections, this enables transmission of 0.5-RTT data, which is vulnerable
+    /// to an attack in which a man-in-the-middle may prevent the handshake from successfully
+    /// completing.
```
Ralith, line 122: I don't understand what distinct attack this is describing. A MITM can always fail a connection at any time by dropping traffic.

Ralith (changes requested):
Thanks for working on this!

quinn/src/connection.rs
gretchenfrage: Oh, I was just deferring to the pre-existing comment added by @djc in [this commit](https://github.com/quinn-rs/quinn/commit/2fb15b9a0721a46ea2336b9f079ffdd9f3154a85)

>    /// On incoming connections, this enables transmission of 0.5-RTT data, which might be
>    /// intercepted by a man-in-the-middle. If this occurs, the handshake will not complete
>    /// successfully.

It's possible I misinterpreted it. Do you know what that comment is referring to?

quinn/src/connection.rs
gretchenfrage: > Also, "typically not be conveyed" is confusing, because it suggests that sometimes they will be made visible, which I think is false.

I wrote this as such because that's ultimately an implementation detail of the software on the other side. It hypothetically could have some QUIC implementation that conveys information about rejected 0-RTT data. I could rephrase to make it more clear that that's what I'm talking about, or alternatively, just remove 'typically' altogether if we think it's not a detail worth mentioning?

Ralith: > or alternatively, just remove 'typically' altogether if we think it's not a detail worth mentioning?

Let's. I think it's more confusing than helpful.

quinn/src/connection.rs
Ralith: My reading of that comment is that a MITM can *intercept and decode* the data, which is more serious.

gretchenfrage: Huh, I'm not aware of that being the case. @djc I don't suppose you'd be able to recall what that comment was referring to?

quinn/src/connection.rs
gretchenfrage: > Rejected 0-RTT streams should fail immediately with `ZeroRttRejected` errors, not hang. If this isn't happening, please open a bug.

Opened #1904.

Ralith: Looks like a typical missed wakeup bug; opened #1905 to fix.

quinn-proto/src/crypto/rustls.rs
gretchenfrage: I thought I observed that behavior, but I can no longer reproduce it. Either my initial observation was mistaken or it's been fixed in recent updates.

quinn/src/connection.rs
djc: Not really -- I guess this is ultimately [`ServerConfig::send_half_rtt_data`](https://docs.rs/rustls/latest/rustls/server/struct.ServerConfig.html#structfield.send_half_rtt_data), but I don't find its documentation all that clear either:

> Whether the server should send “0.5RTT” data. This means the server sends data after its first flight of handshake messages, without waiting for the client to complete the handshake.

> This can improve TTFB latency for either server-speaks-first protocols, or client-speaks-first protocols when paired with “0RTT” data. This comes at the cost of a subtle weakening of the normal handshake integrity guarantees that TLS provides. Note that the initial ClientHello is indirectly authenticated because it is included in the transcript used to derive the keys used to encrypt the data.
> 
> This only applies to TLS1.3 connections. TLS1.2 connections cannot do this optimisation and this setting is ignored for them. It is also ignored for TLS1.3 connections that even attempt client authentication.
> 
> This defaults to false. This means the first application data sent by the server comes after receiving and validating the client’s handshake up to the `Finished` message. This is the safest option.

djc: See also https://bblanche.gitlabpages.inria.fr/publications/BhargavanBlanchetKobeissiSP2017.pdf (search for "0.5-RTT Weak Authentication").

ctz: Yes a MITM can always prevent a handshake from completing (like dropping, reordering, corrupting any message in it). The 0.5RTT flag is enabling the server to send data before it can confirm the handshake was successful in the server -> client direction; that means it hasn't completed any client authentication (rustls outlaws this case) but also that any corruption in the server -> client handshake won't be detected by the server yet. That is not a problem if the data that would be sent is always 100% public (ie, an attacker could just connect to the server and get the same data).

Ralith: It sounds like the only thing we need to warn here is that client authentication can't have been performed yet.

gretchenfrage: Thanks for the links. Helped me understand a bit more.

gretchenfrage:
Thanks for the feedback! All comments should be addressed.

quinn-proto/src/crypto/rustls.rs
Ralith: Looks like this hasn't been addressed?

Ralith (changes requested):

quinn-proto/src/crypto/rustls.rs
gretchenfrage: Sorry, I confused myself! This should be fixed, and also I made a couple other tweaks to the rustls config docs to remove lingering uses of the word "conversion" as a description of what `into_0rtt` does, to be consistent with your other request.

quinn-proto/src/crypto/rustls.rs
```diff
@@ -386,10 +395,14 @@ impl std::error::Error for NoInitialCipherSuite {}
...
-/// [`TryFrom`] implementation with a custom [`rustls::ServerConfig`]. A pre-existing
-/// `ServerConfig` must have TLS 1.3 support enabled for this to work. 0-RTT support is
-/// available to clients if `max_early_data_size` is set to `u32::MAX`.
+/// Quinn implicitly constructs a `QuicServerConfig` with reasonable defaults within
+/// [`ServerConfig::with_single_cert()`][single]. Alternatively, `QuicServerConfig`'s [`TryFrom`]
+/// implementation or `with_initial` method can be used to wrap around a custom
+/// [`rustls::ServerConfig`], in which case care should be taken around certain points:
+///
+/// - If `max_early_data_size` is not set to `u32::MAX`, the server will not be able to accept
+///   incoming 0-RTT data. QUIC prohibits `max_early_data_size` values other than 0 or `u32::MAX`.
```
Ralith, line 404:
```suggestion
///   incoming 0-RTT data. QUIC prohibits `max_early_data_size` values other than disabled or `u32::MAX`.
```
Strictly speaking, RFC 9001 [forbids](https://www.rfc-editor.org/rfc/rfc9001.html#section-4.6.1-2) *any* value other than `u32::MAX`; to disable it, the extension is not sent at all. It's an idiosyncrasy of rustls that setting the value to 0 means not sending it.

Ralith (approved):

quinn-proto/src/crypto/rustls.rs
Ralith: Eh, we're documenting the rustls config here, this is fine.

PR #1827, opened 2024-04-18
perf: change throughput units from MiB/s into Mb/s.

stormshield-damiend:
Use a more commonly used bandwidth unit in perf_client and perf_server.


Ralith (approved):
I always end up doing this conversion by hand anyway.

Issue #1832, opened 2024-04-25
GSO padding has high overhead for application datagrams larger than half MTU


Ralith:
For example, if an application is on a path with a 1500 byte MTU and is sending large numbers of 1200 byte datagrams (e.g. tunneling traffic from a non-MTU-aware protocol), up to 25% of bandwidth will be spent padding 1200 byte QUIC packets to 1500 bytes to allow them to be sent in GSO batches with a predictable segment size. This is an unreasonably high overhead for a realistic scenario, so we should find ways to reduce the amount of padding used.

https://github.com/quinn-rs/quinn/blob/5f63104b6a1a843d7627a13e491b24228eb0312f/quinn-proto/src/connection/mod.rs#L664-L666

Directions to explore:
- End the GSO batch whenever continuing it would require an excessive amount of padding.
  On its own this would prevent GSO from operating entirely in the above scenario, which isn't ideal. It's not obvious what "excessive" should mean, but it probably doesn't need to be precise.
- Additionally, set the GSO segment size to the length of the first packet in the batch, rather than the MTU.
  Combined with the above, this allows GSO to optimize flights of packets of any uniform size. Increases in size within a batch can be avoided by using the GSO segment size, rather than the MTU, as the limit for individual packet size in packet assembly.

An unusually small first packet in a flight might lead to wastefully small packets in the remainder of the flight when traffic is heterogeneous. It's not obvious if this is a meaningful risk in practice: if a small datagram is followed by a datagram too large to fit into the same packet, then the large datagram wouldn't fit within the small segment size either, so the GSO batch would naturally terminate. If a small datagram is followed by stream data, the stream data can be fragmented and the segment size needn't be small to begin with.

PR #1836, opened 2024-04-27
Fix minor datagram send bugs

Ralith:
Noticed while exploring solutions to #1832. Also tossing in any other simple wins I stumble across.

### DIFF
--- a/quinn-proto/src/connection/datagrams.rs
+++ b/quinn-proto/src/connection/datagrams.rs
@@ -149,6 +149,28 @@ impl DatagramState {
         Ok(was_empty)
     }
 
+    /// Discard outgoing datagrams with a payload larger than `max_payload` bytes
+    ///
+    /// Used to ensure that reductions in MTU don't get us stuck in a state where we have a datagram
+    /// queued but can't send it.
+    pub(super) fn drop_oversized(&mut self, max_payload: usize) {
+        self.outgoing.retain(|datagram| {
+            let result = datagram.data.len() < max_payload;
+            if !result {
+                trace!(
+                    "dropping {} byte datagram violating {} byte limit",
+                    datagram.data.len(),
+                    max_payload
+                );
+            }
+            result
+        });
+    }
+
+    /// Attempt to write a datagram frame into `buf`, consuming it from `self.outgoing`
+    ///
+    /// Returns whether a frame was written. At most `max_size` bytes will be written, including
+    /// framing.
     pub(super) fn write(&mut self, buf: &mut Vec<u8>, max_size: usize) -> bool {
         let datagram = match self.outgoing.pop_front() {
             Some(x) => x,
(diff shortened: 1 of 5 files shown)


quinn-proto/src/connection/datagrams.rs
```diff
@@ -68,11 +68,19 @@ impl<'a> Datagrams<'a> {
     ///
     /// Not necessarily the maximum size of received datagrams.
     pub fn max_size(&self) -> Option<usize> {
+        let key = self.conn.spaces[SpaceId::Data].crypto.as_ref().map_or_else(
```
djc, line 71: Would be nice to split the large expression in a separate commit and then fix the panic from there.

I suspect that, as in the other PR, `match` would be easier to follow than `map_or_else()`.

quinn-proto/src/connection/datagrams.rs
```diff
@@ -149,10 +149,28 @@ impl DatagramState {
         Ok(was_empty)
     }
 
-    pub(super) fn write(&mut self, buf: &mut Vec<u8>, max_size: usize) -> bool {
-        let datagram = match self.outgoing.pop_front() {
-            Some(x) => x,
-            None => return false,
+    pub(super) fn write(
+        &mut self,
+        buf: &mut Vec<u8>,
```
djc, line 154: Can you add a docstring that elucidates the distinction between `max_size` and `drop_above_size`?

djc (approved):

quinn-proto/src/connection/datagrams.rs
Ralith: I've instead rewritten this logic to proactively drop all oversized datagrams when a black hole is detected, rather than checking them lazily. It's a cosmetic difference for purposes of this PR, but it preempts a bug in #1837 where datagrams could become unrecoverably clogged by an over-MTU datagram due to `can_send_1rtt` skipping datagrams when an over-segment-size datagram is at the front of the queue, before packet assembly takes place.

quinn-proto/src/connection/datagrams.rs
```diff
@@ -149,6 +149,27 @@ impl DatagramState {
         Ok(was_empty)
     }
 
+    /// Discard outgoing datagrams with a payload larger than `max_payload` bytes
+    ///
+    /// Used to ensure that reductions in MTU don't get us stuck in a state where we have a datagram
+    /// queued but can't send it.
+    pub(super) fn drop_oversized(&mut self, max_payload: usize) {
+        for i in (0..self.outgoing.len()).rev() {
```
djc, line 157: Couldn't this be a call to `self.outgoing.retain()`?

djc (approved):
Nice!

quinn-proto/src/connection/datagrams.rs
djc: Makes sense.

quinn-proto/src/connection/datagrams.rs
Ralith: Good catch.

PR #1837, opened 2024-04-27
Match GSO segment size to the first datagram, not the MTU

### LINKED ISSUE #1832: GSO padding has high overhead for application datagrams larger than half MTU

Ralith:
Fixes #1832.

~Still working on~ how to handle the case where a datagram is queued but won't fit with the current segment size. We need to either detect this in advance and finish the GSO batch before beginning another packet, or gracefully abandon packet assembly after failing to fit any frames. ~Currently leaning towards the latter.~ This will need careful testing.

~Also still TODO: End the GSO batch early if the amount of padding in the current packet would be excessive.~

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -474,7 +474,7 @@ impl Connection {
         // Position in `buf` of the first byte of the current UDP datagram. When coalescing QUIC
         // packets, this can be earlier than the start of the current QUIC packet.
         let mut datagram_start = 0;
-        let segment_size = usize::from(self.path.current_mtu());
+        let mut segment_size = usize::from(self.path.current_mtu());
 
         // Send PATH_CHALLENGE for a previous path if necessary
         if let Some(ref mut prev_path) = self.prev_path {
@@ -666,13 +666,30 @@ impl Connection {
 
                 // Finish current packet
                 if let Some(mut builder) = builder_storage.take() {
-                    // Pad the packet to make it suitable for sending with GSO
-                    // which will always send the maximum PDU.
-                    builder.pad_to(self.path.current_mtu());
+                    if pad_datagram {
+                        builder.pad_to(MIN_INITIAL_SIZE);
+                    }
+
+                    if num_datagrams > 1 {
+                        // Pad the current packet to GSO segment size so it can be included in the
+                        // GSO batch.
+                        builder.pad_to(segment_size as u16);
+                    }
 
                     builder.finish_and_track(now, self, sent_frames.take(), buf);
 
-                    debug_assert_eq!(buf.len(), buf_capacity, "Packet must be padded");
+                    if num_datagrams == 1 {
+                        // Set the segment size for this GSO batch to the size of the first UDP
+                        // datagram in the batch. Larger data that cannot be fragmented
+                        // (e.g. application datagrams) will be included in a future batch. When
+                        // sending large enough volumes of data for GSO to be useful, we expect
+                        // packet sizes to usually be consistent, e.g. populated by max-size STREAM
+                        // frames or uniformly sized datagrams.
+                        segment_size = buf.len();
+                        // Clip the unused capacity out of the buffer so future packets don't
+                        // overrun
+                        buf_capacity = buf.len();
+                    }
                 }
 
                 // Allocate space for another datagram
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -3530,6 +3530,29 @@ impl Connection {
     pub fn current_mtu(&self) -> u16 {
         self.path.current_mtu()
     }
+
+    /// Upper bound for number of bytes of overhead in the next 1-RTT packet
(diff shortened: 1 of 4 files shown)


Ralith:
On second thought, maybe judging whether a datagram fits into a certain segment size in advance of beginning packet assembly isn't so bad. We already have the logic to make a conservative estimate in `max_datagram_size`, and (as of the preceding PR) drop datagrams that exceed that limit.

Ralith:
This is now feature-complete, but still needs testing.

Ralith:
Went ahead and banged out the tests, finding and fixing a number of bugs in the process.

quinn-proto/src/connection/mod.rs
```diff
@@ -475,6 +475,7 @@ impl Connection {
         // Position in `buf` of the first byte of the current UDP datagram. When coalescing QUIC
         // packets, this can be earlier than the start of the current QUIC packet.
         let mut datagram_start = 0;
+        let mut segment_size = self.path.current_mtu();
```
djc, line 481: I suspect this might be easier to review if you split the mere naming of `self.path.current_mtu()` (where used as a `segment_size`) into a separate commit. Also, since we're naming this, based on this diff it looks like the natural type for this might be `usize`, not `u16`?

quinn-proto/src/connection/mod.rs
```diff
@@ -3527,6 +3527,27 @@ impl Connection {
     pub fn current_mtu(&self) -> u16 {
         self.path.current_mtu()
     }
+
+    /// Upper bound for number of bytes in the next 1-RTT packet on the current path consumed by the
+    /// QUIC header and AEAD tag
+    ///
+    /// All other bytes in a packet are frames.
+    fn overhead_bound_1rtt(&self) -> usize {
```
djc, line 3569: Don't love this name -- maybe `max_1rtt_overhead()`? Would also stop the initial docstrings sentence after "packet" and leave the details of "current path" and "consumed by" to a second sentence.

quinn-proto/src/connection/mod.rs
```diff
@@ -3531,22 +3546,37 @@ impl Connection {
     /// Upper bound for number of bytes in the next 1-RTT packet on the current path consumed by the
     /// QUIC header and AEAD tag
     ///
-    /// All other bytes in a packet are frames.
+    /// All other bytes in a packet are frames. Only changes if the length of the remote connection
```
djc, line 3567: Put this in the original commit?

djc (commented):
Initial round, got interrupted by some bored kids.

quinn-proto/src/connection/mod.rs
```diff
@@ -3537,21 +3551,35 @@ impl Connection {
...
     fn max_1rtt_overhead(&self) -> usize {
+        // 1 byte for flags, 4 for worst-case packet number
+        1 + self.rem_cids.active().len() + 4 + self.tag_len_1rtt()
+    }
+
+    /// Exact size of non-frame data for the next 1-RTT packet
+    ///
+    /// Packet number size is sensitive to latency and packet loss, so unlike `max_1rtt_overhead`
+    /// this may change unpredictably under normal conditions.
+    fn exact_1rtt_overhead(&self) -> usize {
```
djc, line 3580: Maybe consider using a single function for `exact_1rtt_overhead()` and `max_1rtt_overhead()` with an `Option` argument for `pn`? Reduces duplication and makes the data dependency on the packet number more obvious in the caller.

djc (approved):

djc:
Good stuff!

PR #1839, opened 2024-04-30
Look up the time through trait Runtime

Ralith:
Allows end-to-end testing with a simulated flow of time, improving determinism and allowing extreme latencies to be modeled more easily. Inspired by @marcblanchet's multi-day experiments with simulated latencies for deep space networking.


marcblanchet:
thanks a lot!  (FYI, we have a running simulation with 5 days one-way delay. takes a month to get the results! so any improvement here would be so useful!

djc (approved):

flub:
Nice!  This can probably also help with android sleep issues messing up timing.

PR #1840, opened 2024-05-01
Make SendStream::finish synchronous

Ralith:
Followup to https://github.com/quinn-rs/quinn/pull/1699. `SendStream::finish` as a future has proved to be a footgun, so this converts it into a synchronous method, delegating "wait until received" duties to `SendStream::stopped`. We expect that to rarely be required in real applications, though it does come up several times in our tests where we specifically want some data to be received before we drop a connection or take a time measurement.

### DIFF
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -10,7 +10,7 @@ use proto::{ConnectionError, FinishError, StreamId, Written};
 use thiserror::Error;
 
 use crate::{
-    connection::{ConnectionRef, UnknownStream},
+    connection::{ClosedStream, ConnectionRef},
     VarInt,
 };
 
@@ -120,8 +120,8 @@ impl SendStream {
             Err(Stopped(error_code)) => {
                 return Poll::Ready(Err(WriteError::Stopped(error_code)));
             }
-            Err(UnknownStream) => {
-                return Poll::Ready(Err(WriteError::UnknownStream));
+            Err(ClosedStream) => {
+                return Poll::Ready(Err(WriteError::ClosedStream));
             }
         };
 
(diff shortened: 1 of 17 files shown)


quinn/src/send_stream.rs
```diff
@@ -130,64 +129,43 @@ impl SendStream {
...
-            conn.wake();
-        }
-        match Pin::new(self.finishing.as_mut().unwrap())
-            .poll(cx)
-            .map(|x| x.unwrap())
-        {
-            Poll::Ready(x) => {
-                self.finishing = None;
-                Poll::Ready(x.map_or(Ok(()), Err))
+    /// To wait for the peer to receive all buffered stream data, see [`stopped`](Self::stopped).
```
djc, line 138: This improved documentation is great!

quinn/src/send_stream.rs
```diff
@@ -130,64 +129,43 @@ impl SendStream {
...
+    ///
+    /// May fail if [`finish`](Self::finish) or [`reset`](Self::reset) was previously called. This
+    /// error is harmless and serves only to indicate that the caller may have incorrect assumptions
+    /// about the stream's state.
+    pub fn finish(&mut self) -> Result<(), UnknownStream> {
+        let mut conn = self.conn.state.lock("finish");
+        match conn.inner.send_stream(self.stream).finish() {
+            Ok(()) => {
```
djc, line 146: Style: I'd maybe suggest making the `match` the tail expression of the function, and hoisting the `Ok(())` into the two arms, removing the explicit `return` in the third? Maybe also move the comment for the second arm onto instead of in the arm, which would help formatting conciseness.

djc (approved):
Looks great!

Style suggestion: I like using the `()` suffix on method/function names in doc links, and rustdoc copes with it. Maybe consider it?

PR #1843, opened 2024-05-02
Apply clippy suggestions from Rust 1.78

djc:


quinn/src/runtime.rs
```diff
@@ -167,7 +167,7 @@ pub fn default_runtime() -> Option<Arc<dyn Runtime>> {
 
     #[cfg(feature = "runtime-async-std")]
     {
-        return Some(Arc::new(AsyncStdRuntime));
+        Some(Arc::new(AsyncStdRuntime))
```
Ralith, line 170: Why here and not also just below?

Ralith (approved):

PR #1851, opened 2024-05-08
Clarify Connection::space_can_send logic

Ralith:


Ralith:
Testing and re-examination reveals that this doesn't actually change any behavior, because if ACKs were sendable then `SendableFrames::acks` would have been set by the first `PacketSpace::can_send` call. We only reach the case that hard-codes that to false if we already know it would have returned false. Cleaned up the logic a bunch to avoid future confusion; please review carefully for absence of semantic changes.

quinn-proto/src/connection/mod.rs
```diff
@@ -1011,27 +1011,15 @@ impl Connection {
...
-
-        if self.spaces[space_id].crypto.is_some() && self.can_send_1rtt(frame_space_1rtt) {
-            return SendableFrames {
-                other: true,
-                acks: false,
-            };
-        }
-
-        if self.zero_rtt_crypto.is_some() && self.side.is_client() {
+        if self.spaces[space_id].crypto.is_some()
```
djc: This commit has too many steps coalesced for me to easily review, do you mind splitting it up so that we can both review more carefully for avoiding semantic changes? In particular, is `other: true, acks: false` the same as `empty()`?

quinn-proto/src/connection/mod.rs
```diff
@@ -1011,21 +1011,19 @@ impl Connection {
...
+        if self.spaces[space_id].crypto.is_none()
+            && (space_id != SpaceId::Data
+                || self.zero_rtt_crypto.is_none()
+                || self.side.is_server())
         {
-            let mut can_send = self.spaces[space_id].can_send(&self.streams);
-            if space_id == SpaceId::Data {
-                can_send.other |= self.can_send_1rtt(frame_space_1rtt);
-            }
-            if !can_send.is_empty() {
```
djc, line 1035: This condition seems to have been dropped, is that a noop? It seems unrelated to the condition inversion.

Ralith: Split into a separate commit for clarity.

quinn-proto/src/connection/mod.rs
Ralith: > do you mind splitting it up

Good thought, done.

> In particular, is other: true, acks: false the same as empty()?

No. In the original logic, if `acks` should be true then we would have returned from the first `if`, and `can_send_1rtt` in `Data` space always means `other` should be true.

djc (approved):

djc:
Much easier to review, thanks!

PR #1852, opened 2024-05-08
make UdpPollHelper public available

zh-jq:
It will be useful when impl custom AsyncUdpSocket


Ralith:
Given the simplicity and tangential nature of this helper, my inclination is to expect folks to copy-paste it rather than commit to it as part of our public API.

zh-jq:
OK. I will copy that code.
Do you have any suggestions on how to write the copyright note?

Ralith:
Attribution to "the Quinn project" should be fine, I guess? It's a pretty trivial code fragment, but I'm not a copyright lawyer.

Issue #1850, opened 2024-05-07
"SendableFrames was SendableFrames { acks: false, other: true }, but only ACKs have been written"


nemethf:
I get the backtrace below using the latest commit 1e54758. Can you, please, tell me how should I debug the issue?  Thanks.

```
thread 'gst-quic-runtime' panicked at ~/quinn/quinn-proto/src/connection/mod.rs:903:13:
SendableFrames was SendableFrames { acks: false, other: true }, but only ACKs have been written
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'gst-quic-runtime' panicked at ~/quinn/quinn/src/mutex.rs:140:42:
called `Result::unwrap()` on an `Err` value: PoisonError { .. }
stack backtrace:
   0:     0x7f7fab854486 - std::backtrace_rs::backtrace::libunwind::trace::hbee8a7973eeb6c93
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/../../backtrace/src/backtrace/libunwind.rs:104:5
   1:     0x7f7fab854486 - std::backtrace_rs::backtrace::trace_unsynchronized::hc8ac75eea3aa6899
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x7f7fab854486 - std::sys_common::backtrace::_print_fmt::hc7f3e3b5298b1083
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/sys_common/backtrace.rs:68:5
   3:     0x7f7fab854486 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hbb235daedd7c6190
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x7f7fab87d3e0 - core::fmt::rt::Argument::fmt::h76c38a80d925a410
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/fmt/rt.rs:142:9
   5:     0x7f7fab87d3e0 - core::fmt::write::h3ed6aeaa977c8e45
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/fmt/mod.rs:1120:17
[...]

PR #1854, opened 2024-05-08
Suppress ACK-only packet sanity-check when datagrams are queued

Ralith:
Fixes https://github.com/quinn-rs/quinn/issues/1850.

This bug caused occasional panics in debug builds in applications which send large application datagrams. The `debug_assert` did not account for the possibility of ACK-only packets arising due to queued non-ACK being too large to fit into the same packet, which became a problem when 6a8758ade2b81c2fc529a425722288dcca21130f introduced the possibility of ACK frames being sent alongside application data (such as large application datagrams) at times when it would be illegal to send an ACK frame spontaneously on its own.

### DIFF
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quinn-proto"
-version = "0.11.0"
+version = "0.11.1"
 edition = "2021"
 rust-version = "1.66"
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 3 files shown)


djc (approved):

djc:
How hard is it to add a test for this? (No need to block a release on this.) Do you want to add a version bump?

Ralith:
> How hard is it to add a test for this?

Should be feasible.

> Do you want to add a version bump?

Sure, soon as I finish diagnosing some related funny behavior...

Ralith:
Fixed and tested a related issue (stealing a variation on the unit test that was previously in #1851), still need a test for the original issue.

djc (approved):

Issue #1803, opened 2024-04-03
After planned migration, connection is linkable to prev. path


nemethf:
The server switches to a new CID, but it sends a path-challenge on the old path with the new CID. So it is still possible to trace back the connection to the old path after the client switches to a new socket.

See #1800 for breaking linkability on the client side.

Ralith:
https://github.com/quinn-rs/quinn/pull/1857 should sort this out.

PR #1857, opened 2024-05-10
Fix previous path validation allowing flows to be linked across migration

Ralith:
Fixes https://github.com/quinn-rs/quinn/issues/1803.


quinn-proto/src/connection/packet_builder.rs
```diff
@@ -41,6 +41,7 @@ impl PacketBuilder {
         datagram_start: usize,
         ack_eliciting: bool,
         conn: &mut Connection,
+        dst_cid: ConnectionId,
```
djc, line 44: Style nit: this feels off wrt our usual argument order?

djc (approved):
Nice!

quinn-proto/src/connection/packet_builder.rs
Ralith: Fixed.

PR #1859, opened 2024-05-12
proto: allow rustls::ClientConfig without initial suite

djc:


djc:
Hmm, maybe we should also implement `TryFrom` for `Arc<rustls::ClientConfig>`?

Ralith (approved):

Ralith:
> Hmm, maybe we should also implement TryFrom for Arc<rustls::ClientConfig>?

Oops, yes, I think we should have done this from the start. Luckily I don't *think* there's a drawback to having both... maybe slightly worse inference in rare cases.

djc:
> Oops, yes, I think we should have done this from the start. Luckily I don't _think_ there's a drawback to having both... maybe slightly worse inference in rare cases.

Added this in a second commit.

Also changed the first commit to actually check that the specified `Suite` is actually AES-128-GCM-SHA256.

Ralith (approved):

Issue #1469, opened 2022-12-13
help build err in openbsd


littlesum:
### log

```
sur# cargo build
   Compiling quinn-udp v0.3.2 (/home/build/quinn-0.9.3/quinn-udp)
error[E0425]: cannot find value `IP_RECVTOS` in crate `libc`
    --> quinn-udp/src/unix.rs:94:75
     |
94   |         if let Err(err) = set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_RECVTOS, OPTION_ON) {
     |                                                                           ^^^^^^^^^^ help: a constant with a similar name exists: `IP_RECVIF`
     |
    ::: /root/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.138/src/unix/bsd/netbsdlike/openbsd/mod.rs:1068:1
     |
1068 | pub const IP_RECVIF: ::c_int = 30;
     | ---------------------------------- similarly named constant `IP_RECVIF` defined here

error[E0412]: cannot find type `mmsghdr` in crate `libc`
   --> quinn-udp/src/unix.rs:166:26
    |
166 |       let mut msgs: [libc::mmsghdr; BATCH_SIZE] = unsafe { mem::zeroed() };
    |                            ^^^^^^^ help: a struct with a similar name exists: `cmsghdr`
    |
   ::: /root/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.138/src/unix/bsd/mod.rs:10:1
    |
10  | / s! {
11  | |     pub struct sockaddr {
12  | |         pub sa_len: u8,
13  | |         pub sa_family: sa_family_t,
...   |
124 | |     }
125 | | }
    | |_- similarly named struct `cmsghdr` defined here

error[E0425]: cannot find function `sendmmsg` in crate `libc`
   --> quinn-udp/src/unix.rs:197:32
    |
197 |         let n = unsafe { libc::sendmmsg(io.as_raw_fd(), msgs.as_mut_ptr(), num_transmits as _, 0) };
    |                                ^^^^^^^^ help: a function with a similar name exists: `sendmsg`
    |
   ::: /root/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.138/src/unix/bsd/mod.rs:743:5
    |
743 |     pub fn sendmsg(fd: ::c_int, msg: *const ::msghdr, flags: ::c_int) -> ::ssize_t;
[...]

PR #1863, opened 2024-05-15
Support openbsd

flub:
This makes Quinn compile on openbsd and the test suite passes as well.

As far as I understand ECN is not supported as part of a control message on recvmsg.  So I disabled the ECN tests for it.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -1,4 +1,4 @@
-#[cfg(not(any(target_os = "macos", target_os = "ios")))]
+#[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "openbsd")))]
 use std::ptr;
 use std::{
     io::{self, IoSliceMut},
@@ -17,6 +17,14 @@ use socket2::SockRef;
 use super::{
     cmsg, log_sendmsg_error, EcnCodepoint, RecvMeta, Transmit, UdpSockRef, IO_ERROR_LOG_INTERVAL,
 };
+
+// Defined in netinet6/in6.h on OpenBSD, this is not yet exported by the libc crate
+// directly.  See https://github.com/rust-lang/libc/issues/3704 for when we might be able to
+// rely on this from the libc crate.
+#[cfg(target_os = "openbsd")]
+const IPV6_DONTFRAG: libc::c_int = 62;
+#[cfg(not(target_os = "openbsd"))]
+const IPV6_DONTFRAG: libc::c_int = libc::IPV6_DONTFRAG;
 
 #[cfg(target_os = "freebsd")]
 type IpTosTy = libc::c_uchar;
@@ -48,6 +56,7 @@ impl UdpSocketState {
         let mut cmsg_platform_space = 0;
         if cfg!(target_os = "linux")
             || cfg!(target_os = "freebsd")
+            || cfg!(target_os = "openbsd")
             || cfg!(target_os = "macos")
             || cfg!(target_os = "ios")
             || cfg!(target_os = "android")
@@ -73,6 +82,7 @@ impl UdpSocketState {
 
         // mac and ios do not support IP_RECVTOS on dual-stack sockets :(
         // older macos versions also don't have the flag and will error out if we don't ignore it
+        #[cfg(not(target_os = "openbsd"))]
         if is_ipv4 || !io.only_v6()? {
             if let Err(err) = set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_RECVTOS, OPTION_ON)
             {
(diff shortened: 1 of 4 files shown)


quinn-udp/src/unix.rs
```diff
@@ -138,12 +145,19 @@ impl UdpSocketState {
...
+                    &*io,
+                    libc::IPPROTO_IPV6,
+                    libc::IPV6_DONTFRAG,
+                    OPTION_ON,
+                )?;
+            }
+            #[cfg(target_os = "openbsd")]
+            {
+                may_fragment |= true;
+            }
```
Ralith, line 160: Why are we skipping this? OpenBSD seems to define the symbol:

https://github.com/openbsd/src/blob/35659bf2453d7eeb12fd24089cf668ca13361c18/sys/netinet6/in6.h#L336

quinn-udp/src/unix.rs
```diff
@@ -293,7 +307,7 @@ fn send(
     }
 }
 
-#[cfg(any(target_os = "macos", target_os = "ios"))]
+#[cfg(any(target_os = "macos", target_os = "ios", target_os = "openbsd"))]
```
Ralith, line 310: `IP_RECVDSTADDR` seems to be defined in OpenBSD; should we add it to the cases where the `encode_ip` argument to `prepare_msg` is true below?

quinn-udp/src/unix.rs
flub: The libc crate does not have this symbol, I would prefer to have something sub-optimal working that can later be improved than wait until this is fully supported.

djc (approved):
Seems reasonable to merge this even without ECN support for now.

quinn-udp/src/unix.rs
flub: good catch, yes.  testing that shows it also works.

flub:
> OpenBSD does seem to define `IPV6_RECVTCLASS`: https://github.com/openbsd/src/blob/35659bf2453d7eeb12fd24089cf668ca13361c18/sys/netinet6/in6.h#L330C9-L330C24

Indeed, and the socket option is being set as well still.  Let me try again if I can get ENC to work for just IPv6.

flub:
> Thanks for working on this!
> 
> > As far as I understand ECN is not supported as part of a control message on recvmsg. So I disabled the ECN tests for it.
> 
> OpenBSD does seem to define `IPV6_RECVTCLASS`: https://github.com/openbsd/src/blob/35659bf2453d7eeb12fd24089cf668ca13361c18/sys/netinet6/in6.h#L330C9-L330C24

Ok, ECN now works and is tested on IPv6 but not IPv4.  I've had to split some tests for that.  I left the existing ecn_ipv6 test unmodified in the end but renamed it rather than fully splitting it in two.  It still makes sense on it's own.

flub:
This will close #1469 I believe.

quinn-udp/src/unix.rs
Ralith: We don't need to wait for `libc` to define the constant for us; it's just a number. Let's define the constant ourselves, ideally with a TODO pointing to a libc issue or PR.

quinn-udp/src/unix.rs
```diff
@@ -579,7 +603,11 @@ fn decode_recv(
     for cmsg in cmsg_iter {
         match (cmsg.cmsg_level, cmsg.cmsg_type) {
             // FreeBSD uses IP_RECVTOS here, and we can be liberal because cmsgs are opt-in.
-            (libc::IPPROTO_IP, libc::IP_TOS) | (libc::IPPROTO_IP, libc::IP_RECVTOS) => unsafe {
+            (libc::IPPROTO_IP, libc::IP_TOS) => unsafe {
+                ecn_bits = cmsg::decode::<u8, libc::cmsghdr>(cmsg);
+            },
+            #[cfg(not(target_os = "openbsd"))]
+            (libc::IPPROTO_IP, libc::IP_RECVTOS) => unsafe {
```
Ralith, line 610: The comment mentioning `IP_RECVTOS` should probably remain attached to the case that matches it

quinn-udp/tests/tests.rs
```diff
@@ -31,6 +33,46 @@ fn basic() {
...
+                segment_size: None,
+                src_ip: None,
+            },
+        );
+    }
+}
+
+#[test]
+#[cfg(not(target_os = "openbsd"))]
+fn enc_v6_dualstack() {
```
Ralith, line 75: typo
```suggestion
fn ecn_v6_dualstack() {
```

quinn/src/tests.rs
```diff
@@ -532,6 +532,7 @@ fn run_echo(args: EchoArgs) {
             // platforms in the doc comments of the various `local_ip` functions.
```
Ralith, line 532: Please update docs as discussed in this above comment.

Ralith (changes requested):

quinn/src/tests.rs
Ralith: Note these docs are being centralized in https://github.com/quinn-rs/quinn/pull/1868.

quinn-udp/src/unix.rs
flub: Cool, I've updated this now to just define the constant.  All still works.

flub:
It is not clear to me if the failing test is flaky or due to something I did, any hints?

Ralith (approved):
LGTM, thanks!

flub:
@Ralith thanks for the prodding, definitely better than the first version!

Issue #1860, opened 2024-05-13
Expose Packet Decoder？


thynson:
I'm trying to implement a simple QUIC Load Balancer, following [QUIC-LB] draft, in Rust.
It would be great if the `PartialDecode` related symbol can be exposed to the public.

[QUIC-LB]: https://datatracker.ietf.org/doc/draft-ietf-quic-load-balancers

djc:
Can you make a PR with the changes you need? I'd probably be open to exposing a little more API around this (especially since, based on history, our `PartialDecode` is unlikely to change much).

thynson:
I'm implementing it with s2n-quic for now, which already expose its packet decoder. And I'll try it out with quinn later when I finished a prototype.

PR #1865, opened 2024-05-17
proto: make PartialDecode API public

thynson:
Context: #1860.

This PR did a bit more than just expose symbols to public, a trait `ConnectionIdValidator` was introduced, due to the fact that a QUIC Load balancer have to support different kind of Connection ID Schemes, per [QUIC-LB-DRAFT Secion 4.1]( https://www.ietf.org/archive/id/draft-ietf-quic-load-balancers-19.html#section-4.1). 
Also some documents was added to suppress the clippy warning after those symbol being made public.

### DIFF
--- a/quinn-proto/src/packet.rs
+++ b/quinn-proto/src/packet.rs
@@ -14,7 +14,7 @@ use crate::{
 /// (which includes a variable-length packet number) without crypto context.
 /// The crypto context (represented by the `Crypto` type in Quinn) is usually
 /// part of the `Connection`, or can be derived from the destination CID for
-// Initial packets.
+/// Initial packets.
 ///
 /// To cope with this, we decode the invariant header (which should be stable
 /// across QUIC versions), which gives us the destination CID and allows us
@@ -32,13 +32,13 @@ impl PartialDecode {
     /// Begin decoding a QUIC packet from `bytes`, returning any trailing data not part of that packet
     pub fn new(
         bytes: BytesMut,
-        local_cid_len: usize,
+        cid_parser: &impl ConnectionIdParser,
         supported_versions: &[u32],
         grease_quic_bit: bool,
     ) -> Result<(Self, Option<BytesMut>), PacketDecodeError> {
         let mut buf = io::Cursor::new(bytes);
         let plain_header =
-            PlainHeader::decode(&mut buf, local_cid_len, supported_versions, grease_quic_bit)?;
+            PlainHeader::decode(&mut buf, cid_parser, supported_versions, grease_quic_bit)?;
         let dgram_len = buf.get_ref().len();
         let packet_len = plain_header
             .payload_len()
@@ -564,7 +564,7 @@ impl PlainHeader {
     /// Decode a plain header from given buffer, with given [`ConnectionIdParser`].
     pub fn decode(
         buf: &mut io::Cursor<BytesMut>,
-        local_cid_len: usize,
+        cid_parser: &impl ConnectionIdParser,
         supported_versions: &[u32],
         grease_quic_bit: bool,
     ) -> Result<Self, PacketDecodeError> {
@@ -574,13 +574,10 @@ impl PlainHeader {
         }
         if first & LONG_HEADER_FORM == 0 {
             let spin = first & SPIN_BIT != 0;
-            if buf.remaining() < local_cid_len {
-                return Err(PacketDecodeError::InvalidHeader("cid out of bounds"));
-            }
 
             Ok(Self::Short {
                 spin,
-                dst_cid: ConnectionId::from_buf(buf, local_cid_len),
+                dst_cid: cid_parser.parse(buf)?,
             })
         } else {
             let version = buf.get::<u32>()?;
@@ -770,6 +767,32 @@ impl PacketNumber {
     }
 }
 
+/// A [`ConnectionIdParser`] implementation that assumes the connection ID is of fixed length
+pub struct FixedLengthConnectionIdParser {
+    expected_len: usize,
+}
+
+impl FixedLengthConnectionIdParser {
+    /// Create a new instance of `FixedLengthConnectionIdParser`
+    pub fn new(expected_len: usize) -> Self {
+        Self { expected_len }
+    }
+}
+
+impl ConnectionIdParser for FixedLengthConnectionIdParser {
+    fn parse(&self, buffer: &mut impl Buf) -> Result<ConnectionId, PacketDecodeError> {
+        (buffer.remaining() >= self.expected_len)
+            .then(|| ConnectionId::from_buf(buffer, self.expected_len))
+            .ok_or(PacketDecodeError::InvalidHeader("packet too small"))
+    }
+}
+
+/// Parse connection id in short header packet
+pub trait ConnectionIdParser {
+    /// Parse a connection id from given buffer
+    fn parse(&self, buf: &mut impl Buf) -> Result<ConnectionId, PacketDecodeError>;
+}
+
 /// Long packet type including non-uniform cases
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub(crate) enum LongHeaderType {
@@ -945,9 +968,14 @@ mod tests {
 
         let server = initial_keys(Version::V1, &dcid, Side::Server, &suite);
         let supported_versions = DEFAULT_SUPPORTED_VERSIONS.to_vec();
-        let decode = PartialDecode::new(buf.as_slice().into(), 0, &supported_versions, false)
-            .unwrap()
-            .0;
+        let decode = PartialDecode::new(
+            buf.as_slice().into(),
+            &FixedLengthConnectionIdParser::new(0),
+            &supported_versions,
+            false,
+        )
+        .unwrap()
+        .0;
         let mut packet = decode.finish(Some(&*server.header.remote)).unwrap();
         assert_eq!(
             packet.header_data[..],
--- a/quinn-proto/src/packet.rs
+++ b/quinn-proto/src/packet.rs
@@ -23,7 +23,7 @@ use crate::{
 #[cfg_attr(test, derive(Clone))]
 #[derive(Debug)]
 pub struct PartialDecode {
-    plain_header: PlainHeader,
+    plain_header: ProtectedHeader,
     buf: io::Cursor<BytesMut>,
 }
 
@@ -38,7 +38,7 @@ impl PartialDecode {
     ) -> Result<(Self, Option<BytesMut>), PacketDecodeError> {
         let mut buf = io::Cursor::new(bytes);
         let plain_header =
-            PlainHeader::decode(&mut buf, cid_parser, supported_versions, grease_quic_bit)?;
+            ProtectedHeader::decode(&mut buf, cid_parser, supported_versions, grease_quic_bit)?;
         let dgram_len = buf.get_ref().len();
         let packet_len = plain_header
             .payload_len()
@@ -61,12 +61,12 @@ impl PartialDecode {
         self.buf.get_ref()
     }
 
-    pub(crate) fn initial_header(&self) -> Option<&PlainInitialHeader> {
+    pub(crate) fn initial_header(&self) -> Option<&ProtectedInitialHeader> {
         self.plain_header.as_initial()
     }
 
     pub(crate) fn has_long_header(&self) -> bool {
-        !matches!(self.plain_header, PlainHeader::Short { .. })
+        !matches!(self.plain_header, ProtectedHeader::Short { .. })
     }
 
     pub(crate) fn is_initial(&self) -> bool {
@@ -74,7 +74,7 @@ impl PartialDecode {
     }
 
     pub(crate) fn space(&self) -> Option<SpaceId> {
-        use self::PlainHeader::*;
+        use self::ProtectedHeader::*;
         match self.plain_header {
             Initial { .. } => Some(SpaceId::Initial),
             Long {
@@ -92,7 +92,7 @@ impl PartialDecode {
 
     pub(crate) fn is_0rtt(&self) -> bool {
         match self.plain_header {
-            PlainHeader::Long { ty, .. } => ty == LongType::ZeroRtt,
+            ProtectedHeader::Long { ty, .. } => ty == LongType::ZeroRtt,
             _ => false,
         }
     }
@@ -112,13 +112,13 @@ impl PartialDecode {
         self,
         header_crypto: Option<&dyn crypto::HeaderKey>,
     ) -> Result<Packet, PacketDecodeError> {
-        use self::PlainHeader::*;
+        use self::ProtectedHeader::*;
         let Self {
             plain_header,
             mut buf,
         } = self;
 
-        if let Initial(PlainInitialHeader {
+        if let Initial(ProtectedInitialHeader {
             dst_cid,
             src_cid,
             token_pos,
@@ -490,9 +490,9 @@ impl PartialEncode {
 
 /// Plain packet header
 #[derive(Clone, Debug)]
-pub enum PlainHeader {
+pub enum ProtectedHeader {
     /// An Initial packet header
-    Initial(PlainInitialHeader),
+    Initial(ProtectedInitialHeader),
     /// A Long packet header, as used during the handshake
     Long {
         /// Type of the Long header packet
@@ -533,8 +533,8 @@ pub enum PlainHeader {
     },
 }
 
-impl PlainHeader {
-    fn as_initial(&self) -> Option<&PlainInitialHeader> {
+impl ProtectedHeader {
+    fn as_initial(&self) -> Option<&ProtectedInitialHeader> {
         match self {
             Self::Initial(x) => Some(x),
             _ => None,
@@ -543,7 +543,7 @@ impl PlainHeader {
 
     /// The destination Connection ID of the packet
     pub fn dst_cid(&self) -> &ConnectionId {
-        use self::PlainHeader::*;
+        use self::ProtectedHeader::*;
         match self {
             Initial(header) => &header.dst_cid,
             Long { dst_cid, .. } => dst_cid,
@@ -554,9 +554,9 @@ impl PlainHeader {
     }
 
     fn payload_len(&self) -> Option<u64> {
-        use self::PlainHeader::*;
+        use self::ProtectedHeader::*;
         match self {
-            Initial(PlainInitialHeader { len, .. }) | Long { len, .. } => Some(*len),
+            Initial(ProtectedInitialHeader { len, .. }) | Long { len, .. } => Some(*len),
             _ => None,
         }
     }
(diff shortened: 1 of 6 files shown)


thynson:
I'm not sure if `ConnectionIdValidator` should be move to `quinn_proto::shared`.

quinn-proto/src/cid_generator.rs
```diff
@@ -31,6 +31,33 @@ pub trait ConnectionIdGenerator: Send + Sync {
     fn cid_lifetime(&self) -> Option<Duration>;
 }
 
+/// Validates connection id in short header packet
+pub trait ConnectionIdValidator {
```
djc, line 35: The trait should be defined below the implementator.

quinn-proto/src/cid_generator.rs
```diff
@@ -31,6 +31,33 @@ pub trait ConnectionIdGenerator: Send + Sync {
...
+    /// Validate and regconize a connection id from given buffer, return the length of connection id
+    /// or [`None`] if it is invalid
+    fn validate(&self, buffer: &[u8]) -> Option<usize>;
```
djc, line 38: If we're going to do this, I think we should deduplicate this with `ConnectionIdGenerator::validate()` somehow?

quinn-proto/src/cid_generator.rs
```diff
@@ -31,6 +31,33 @@ pub trait ConnectionIdGenerator: Send + Sync {
...
+
+impl FixedLengthConnectionIdValidator {
+    pub(crate) fn new(expected_len: usize) -> Self {
+        Self { expected_len }
+    }
+}
+
+impl ConnectionIdValidator for FixedLengthConnectionIdValidator {
+    fn validate(&self, buffer: &[u8]) -> Option<usize> {
+        if buffer.len() >= self.expected_len {
```
djc, line 53: Nit: write this as `(buffer.len() >= self.expected_len).then_some(self.expected_len)`.

quinn-proto/src/packet.rs
```diff
@@ -18,17 +19,17 @@ use crate::{
 // across QUIC versions), which gives us the destination CID and allows us
 // to inspect the version and packet type (which depends on the version).
 // This information allows us to fully decode and decrypt the packet.
-#[allow(unreachable_pub)] // fuzzing only
+#[allow(unreachable_pub, missing_docs)] // fuzzing only
```
djc, line 22: If this is no longer only for fuzzing we should get rid of the `missing_docs` and `unreachable_pub` allows.

quinn-proto/src/packet.rs
```diff
@@ -18,17 +19,17 @@ use crate::{
...
 #[cfg_attr(test, derive(Clone))]
 #[derive(Debug)]
 pub struct PartialDecode {
     plain_header: PlainHeader,
     buf: io::Cursor<BytesMut>,
 }
 
-#[allow(clippy::len_without_is_empty)]
+#[allow(clippy::len_without_is_empty, missing_docs)]
```
djc, line 30: Don't add `missing_docs` here.

Ralith:
What does this new trait add over the caller performing whatever validation they like themselves after decoding a header, as is already done with `ConnectionIdGenerator::validate`?

quinn-proto/src/packet.rs
```diff
@@ -618,17 +655,43 @@ impl PlainHeader {
             }
         }
     }
+
+    fn decode(
```
djc: This should be defined above `decode_with_cid_validator()`.

quinn-proto/src/packet.rs
```diff
@@ -55,12 +56,44 @@ impl PartialDecode {
         }
     }
 
+    /// Create an instance of [`PartialDecode`], with connection id validated by given
+    ///  [`ConnectionIdValidator`]
+    pub fn new_with_cid_validator<CidValidator: ConnectionIdValidator>(
```
djc, line 61: This has way too much duplicated code.

djc (commented):
I'm not convinced this `ConnectionIdValidator` is the right abstraction. Given that we're dealing with parsing, it seems like that should be front and center, maybe something like:

```rust
trait ConnectionIdParser {
     fn parse(&self, buf: &mut impl Buf) -> Result<ConnectionId, PacketDecodeError>;
}
```

(Feels like it could even be `TryFrom` impl?)

Also in terms of commit history, I'd like to see one commit that makes all the existing API that you need public, and then a separate commit that adds any new stuff you need (with as little as possible duplication).

djc:
> What does this new trait add over the caller performing whatever validation they like themselves after decoding a header, as is already done with `ConnectionIdGenerator::validate`?

It looks like parsing the header needs the expected CID len as input.

thynson:
Thanks for review. I'm going to polish it based on your comments

thynson:
The idea of `ConnectionIdValidator` comes from s2n-quic, but they have `Validator` and `Generator` and `Format: Validator + Generator`(see [here](https://github.com/aws/s2n-quic/blob/main/quic/s2n-quic-core/src/connection/id.rs) is what s2n-quic does). 

While in quinn, `ConnectionIdGenerator` has `validate` for now, so the duplication of functionality seems unavoidable unless we change trait `ConnectionIdGenerator` to requires `ConnectionIdParser`. Shall we go in this way?

thynson:
> While in quinn, `ConnectionIdGenerator` has `validate` for now, so the duplication of functionality seems unavoidable unless we change trait `ConnectionIdGenerator` to requires `ConnectionIdParser`. Shall we go in this way?

After further digging, I believe `ConnectionIdGenerator.parse` and `ConnectionIdValidator.validate` provides different semantics. The former is used to decide whether [a state-less reset should be sent](https://github.com/quinn-rs/quinn/blob/8bd0600089fa8bcf333df4cad2e4cac23b514a99/quinn-proto/src/endpoint.rs#L305), so rename `ConnectionIdValidator` seems right to me.

quinn-proto/src/cid_generator.rs
thynson: See https://github.com/quinn-rs/quinn/pull/1865#issuecomment-2117104025

quinn-proto/src/cid_generator.rs
thynson: `ConnectionIdParser` was introduced in `shred.rs` instead of this, I'm not sure whether the new define order is what you want.

quinn-proto/src/shared.rs
```diff
@@ -2,7 +2,36 @@ use std::{fmt, net::SocketAddr, time::Instant};
...
+    fn parse(&self, buf: &mut impl Buf) -> Result<ConnectionId, PacketDecodeError>;
+}
+
+impl ConnectionIdParser for FixedLengthConnectionIdParser {
+    fn parse(&self, buffer: &mut impl Buf) -> Result<ConnectionId, PacketDecodeError> {
+        (buffer.remaining() >= self.expected_len)
+            .then(|| ConnectionId::from_buf(buffer, self.expected_len))
+            .ok_or(PacketDecodeError::InvalidHeader("unexpected end of packet"))
+    }
+}
```
djc, line 34: This `impl` should live above the trait definition.

quinn-proto/src/packet.rs
```diff
@@ -539,25 +541,38 @@ impl PlainHeader {
...
+            &FixedLengthConnectionIdParser::new(local_cid_len),
+            supported_versions,
+            grease_quic_bit,
+        )
+    }
+
+    /// Decode a plain header from given buffer, with given [`ConnectionIdParser`].
+    pub fn decode_with_cid_parser<Parser: ConnectionIdParser>(
+        buf: &mut io::Cursor<BytesMut>,
+        cid_validator: &Parser,
```
djc, line 587: Nit: spell this as `cid_parser: &impl ConnectionIdParser`.

quinn-proto/src/packet.rs
```diff
@@ -10,16 +10,16 @@ use crate::{
...
-// (which includes a variable-length packet number) without crypto context.
-// The crypto context (represented by the `Crypto` type in Quinn) is usually
-// part of the `Connection`, or can be derived from the destination CID for
-// Initial packets.
-//
-// To cope with this, we decode the invariant header (which should be stable
-// across QUIC versions), which gives us the destination CID and allows us
-// to inspect the version and packet type (which depends on the version).
-// This information allows us to fully decode and decrypt the packet.
+/// Due to packet number encryption, it is impossible to fully decode a header
```
djc, line 13: If we're going to make this into a doc string, it should start with a concise description that fits on a single line, separated from the detailed rest of the text by an empty line.

quinn-proto/src/packet.rs
```diff
@@ -539,25 +541,38 @@ impl PlainHeader {
         }
     }
 
-    fn decode(
+    /// Decode a plain header from given buffer, assuming a fixed-length CID.
+    pub fn decode(
```
djc, line 570: It looks like this only has a single caller. I don't see a good reason to maintain two variants?

quinn-proto/src/packet.rs
```diff
@@ -30,16 +30,35 @@ pub struct PartialDecode {
...
+        Self::new_with_cid_parser(
+            bytes,
+            &FixedLengthConnectionIdParser::new(local_cid_len),
+            supported_versions,
+            grease_quic_bit,
+        )
+    }
+
+    /// Create an instance of [`PartialDecode`], with given [`ConnectionIdParser`].
+    pub fn new_with_cid_parser<Parser: ConnectionIdParser>(
```
djc, line 49: Same here: there are just two callers, both of which are passing in `self.local_cid_state.cid_len()`. I think we can just have one variant, and maybe implement `ConnectionIdParser` for `CidState` and `ConnectionIdGenerator`?

djc (commented):
This is looking much better already.

quinn-proto/src/packet.rs
thynson: I would like to keep them as is, if we really want to remove `PartialDecoder::new`, construct a parser when invoke `PartialDecoder::new_with_cid_parser`, impl ConnectionIdParser for them is less readable in my opinion.

quinn-proto/src/packet.rs
```diff
@@ -30,16 +30,35 @@ pub struct PartialDecode {
 
 #[allow(clippy::len_without_is_empty)]
 impl PartialDecode {
-    #[allow(unreachable_pub)] // fuzzing only
+    /// Create an instance of [`PartialDecode`], expecting a fixed-length local connection id
```
Ralith: If we're making this public, we should document the return type here and below. In particular, it's not obvious what the `Option<BytesMut>` is.

quinn-proto/src/packet.rs
```diff
@@ -10,6 +10,8 @@ use crate::{
     ConnectionId,
 };
 
+/// A decoder that decods invariant header from packet
+///
```
Ralith, line 14: This change should probably be in the previous commit, where the comment was made a doc comment.

Also:
- Typo: "decodes", not "decods"
- "A decoder that decodes" is redundant. Just say "Decodes a QUIC packet's invariant header" or similar.

quinn-proto/src/packet.rs
```diff
@@ -654,15 +661,27 @@ impl PlainHeader {
     }
 }
 
+/// A Plain QUIC Header
 #[derive(Clone, Debug)]
-pub(crate) struct PlainInitialHeader {
+pub struct PlainInitialHeader {
```
Ralith, line 632: If all the other enum variants' fields are public, these probably should be too, and then the accessors won't be needed.

Ralith (changes requested):
Mostly cosmetic issues, general direction looks good. Could you squash the constructor-unification commits down into the commits that introduce the new constructors?

quinn-proto/src/shared.rs
```diff
@@ -2,7 +2,36 @@ use std::{fmt, net::SocketAddr, time::Instant};
...
+
+/// Validates connection id in short header packet
+pub trait ConnectionIdParser {
+    /// Validate and regconize a connection id from given buffer, return the length of connection id
```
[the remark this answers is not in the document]
thynson: I think "Validate" should changed to "Parse" actually, `ConnectionIdParser` was originally named `ConntionIdValidator` when this PR was summitted.

quinn-proto/src/shared.rs
```diff
@@ -2,7 +2,36 @@ use std::{fmt, net::SocketAddr, time::Instant};
...
+    pub(crate) fn new(expected_len: usize) -> Self {
+        Self { expected_len }
+    }
+}
+
+impl ConnectionIdParser for FixedLengthConnectionIdParser {
+    fn parse(&self, buffer: &mut impl Buf) -> Result<ConnectionId, PacketDecodeError> {
+        (buffer.remaining() >= self.expected_len)
+            .then(|| ConnectionId::from_buf(buffer, self.expected_len))
+            .ok_or(PacketDecodeError::InvalidHeader("unexpected end of packet"))
```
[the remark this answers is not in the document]
thynson: I checked the original error string, at https://github.com/quinn-rs/quinn/blob/9c2c553d95bf1a9e3046f255d5e243b2f7d7ddc1/quinn-proto/src/packet.rs#L555, which is "cid out of bounds".
Should I follow the original one?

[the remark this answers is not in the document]
Ralith: I think "packet too small" is probably the most helpful for an end user, though none of the three options are terrible.

quinn-proto/src/packet.rs
```diff
@@ -539,25 +541,38 @@ impl PlainHeader {
...
+            &FixedLengthConnectionIdParser::new(local_cid_len),
+            supported_versions,
+            grease_quic_bit,
+        )
+    }
+
+    /// Decode a plain header from given buffer, with given [`ConnectionIdParser`].
+    pub fn decode_with_cid_parser(
+        buf: &mut io::Cursor<BytesMut>,
+        cid_validator: &impl ConnectionIdParser,
```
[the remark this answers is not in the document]
thynson: My opinion is that unless dynamic dispatch is required, `&impl ConnectionIdParser` should be preferred.

quinn-proto/src/packet.rs
Ralith: This doesn't seem to have been addressed?

quinn-proto/src/packet.rs
```diff
@@ -28,16 +31,16 @@ pub struct PartialDecode {
 
 #[allow(clippy::len_without_is_empty)]
 impl PartialDecode {
-    #[allow(unreachable_pub)] // fuzzing only
+    /// Decode `bytes` to create an instance of [`PartialDecode`], returning any remaining bytes
```
Ralith, line 34: This documentation is not clear about what "remaining" bytes are. How about something like this?

```rs
/// Begin decoding a QUIC packet from `bytes`, returning any trailing data not part of that packet
```

thynson: Updated based on your suggestion.

Ralith (approved):
LGTM, thanks!

quinn-proto/src/packet.rs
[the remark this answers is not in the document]
djc: The name is still wrong.

quinn-proto/src/packet.rs
djc: I'm not convinced. @Ralith any thoughts? Feels like `CidState` and `ConnectionIdGenerator` are plausible implementors of `ConnectionIdParser`.

Ralith: > I'm not convinced.

I think @thynson's comment above is out of date, and/or they decided to concede the point. The current form of the PR no longer introduces new constructors, instead tweaking the pre-existing ones.

> Feels like CidState and ConnectionIdGenerator are plausible implementors of ConnectionIdParser.

I think it would make sense for `ConnectionIdGenerator` to require `ConnectionIdParser`, yeah. We should probably take care to not perform redundant validation work for CIDs that have already been implicitly validated by finding them in the endpoint's live connection index.

This could be pursued in followup work, although we should do so before publishing this change, as it could render `FixedLengthConnectionIdGenerator` redundant.

I'm not sure `CidState` should implement `ConnectionIdParser`. It looks like it only knows about `cid_len` for the sake of judging whether CIDs are in use at all, and thereby detecting certain transport errors. User-defined CID generators might have more complex parsing requirements and, unlike `ConnectionIdGenerator`, `CidState` doesn't seem to be in a position to know about them.

quinn-proto/src/packet.rs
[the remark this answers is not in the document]
Ralith: Are you sure you're looking at the right code? The current revision has:

```rs
    /// Decode a plain header from given buffer, with given [`ConnectionIdParser`].
    pub fn decode(
        buf: &mut io::Cursor<BytesMut>,
        cid_parser: &impl ConnectionIdParser,
```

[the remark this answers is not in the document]
djc: I was expecting it to be correct from 6d8451cc9a336ad343ec22962a7f04883843d21e onwards.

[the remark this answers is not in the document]
djc: Fixed this up myself.

djc:
I ended up squashing and reordering these changes. I changed the commits to (a) first make the APIs public and then (b) add the `ConnectionIdParser` trait and implementation. The previous approach had some intermediate states that did not make sense. @thynson can you check that this still satisfies your use case (there should be no changes in terms of the final public API, but still good to check)?

thynson:
It's totally fine to me .

quinn-proto/src/packet.rs
```diff
@@ -620,13 +640,19 @@ impl PlainHeader {
...
 
+/// Header of an Initial packet, before decryption
 #[derive(Clone, Debug)]
-pub(crate) struct PlainInitialHeader {
-    pub(crate) dst_cid: ConnectionId,
-    pub(crate) src_cid: ConnectionId,
-    pub(crate) token_pos: Range<usize>,
-    pub(crate) len: u64,
-    pub(crate) version: u32,
+pub struct PlainInitialHeader {
```
djc, line 645: @Ralith before this becomes public I think we should give this name some more thought? As I understand it this is before removing header protection, which makes the `Plain` prefix a little confusing. Internally we now have `PlainInitialHeader` and `InitialHeader` which could also do with better disambiguation...

thynson: `PlainInitialHeader` makes people think of plain text, which is unencrypted in context of encryption/decryption.
Maybe we should follow RFC9000/RFC9001 to use the term "Protected" ?

djc: I like it! Added a commit that renames `PlainHeader` and `PlainInitialHeader` to `ProtectedHeader` and `ProtectedInitialHeader` -- both of these were previously private so should be fine.

quinn-proto/src/packet.rs
```diff
@@ -770,6 +767,32 @@ impl PacketNumber {
     }
 }
 
+/// An [`ConnectionIdParser`] implementation that assumes the connection ID is of fixed length
```
thynson, line 770: Nits: should be "A [`ConnectionIdParser`] implementation"

djc: Good call, fixed.

quinn-proto/src/packet.rs
Ralith: SGTM.

PR #1872, opened 2024-05-19
Trivial SendStream error cleanup

Ralith:


quinn/src/send_stream.rs
```diff
@@ -488,3 +488,14 @@ pub enum StoppedError {
     #[error("0-RTT rejected")]
     ZeroRttRejected,
 }
+
+impl From<StoppedError> for io::Error {
```
djc, line 492: I'm not sure why this is added? I don't see any rationale in the PR description and it doesn't really fit the PR title of "Trivial .. error cleanup" in my mind...

Ralith: We implement `From<_> for io::Error` for the rest of our error types, for the convenience of users who might be adapting QUIC to interfaces that bake in `io::Error` (e.g. various `AsyncWrite` traits), or combining QUIC with other I/O. It's an odd and unnecessary oversight to exclude `StoppedError` from this pattern.

djc (approved):

PR #1873, opened 2024-05-19
Introduce RecvStream::reset

### LINKED ISSUE #1870: API for awaiting for stream reset on the reader
Would be great to have an API that would allow waiting for a stream reset from the reader end. This is something that WebTransport API for the Web has, and it would be nice for supporting cross-platform interfaces for WebTransport (like `xwt`).

WebTransport is relevant in the context of QUIC and `quinn` because WebTransport is a protocol that builds on top of HTTP3, which builds on top of QUIC.

Concretely, I need this because I am having issues properly modeling interfaces at https://github.com/MOZGIII/xwt/pull/152.

Ralith:
Fixes #1870. I'm pretty sure I handled the state cleanup and flow control implications properly, but close review desired there all the same.

### DIFF
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -159,6 +159,33 @@ impl<'a> RecvStream<'a> {
 
         Ok(())
     }
+
+    /// Check whether this stream has been reset by the peer, returning the reset error code if so
+    ///
+    /// After returning `Ok(Some(_))` once, stream state will be discarded and all future calls will
+    /// return `Err(ClosedStream)`.
+    pub fn received_reset(&mut self) -> Result<Option<VarInt>, ClosedStream> {
+        let hash_map::Entry::Occupied(entry) = self.state.recv.entry(self.id) else {
+            return Err(ClosedStream { _private: () });
+        };
+        let Some(s) = entry.get().as_ref() else {
+            return Ok(None);
+        };
+        if s.stopped {
+            return Err(ClosedStream { _private: () });
+        }
+        let Some(code) = s.reset_code() else {
+            return Ok(None);
+        };
+
+        // Clean up state after application observes the reset, since there's no reason for the
+        // application to attempt to read or stop the stream once it knows it's reset
+        entry.remove_entry();
+        self.state.stream_freed(self.id, StreamHalf::Recv);
+        self.state.queue_max_stream_id(self.pending);
+
+        Ok(Some(code))
+    }
 }
 
 /// Access to streams
--- a/quinn/src/recv_stream.rs
+++ b/quinn/src/recv_stream.rs
@@ -308,7 +308,7 @@ impl RecvStream {
 
         // If we stored an error during a previous call, return it now. This can happen if a
         // `read_fn` both wants to return data and also returns an error in its final stream status.
-        let status = match self.reset.take() {
+        let status = match self.reset {
             Some(code) => ReadStatus::Failed(None, Reset(code)),
             None => {
                 let mut recv = conn.inner.recv_stream(self.stream);
@@ -340,6 +340,7 @@ impl RecvStream {
             ReadStatus::Failed(read, Reset(error_code)) => match read {
                 None => {
                     self.all_data_read = true;
+                    self.reset = Some(error_code);
                     Poll::Ready(Err(ReadError::Reset(error_code)))
                 }
                 done => {
--- a/quinn/src/recv_stream.rs
+++ b/quinn/src/recv_stream.rs
@@ -1,5 +1,5 @@
 use std::{
-    future::Future,
+    future::{poll_fn, Future},
     io,
     pin::Pin,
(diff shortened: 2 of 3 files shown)


quinn/src/recv_stream.rs
```diff
@@ -284,6 +284,52 @@ impl RecvStream {
...
+    ///
+    /// Yields `Some` with the reset error code when the stream is reset by the peer. Yields `None`
+    /// when the stream was previously [`stop()`](Self::stop)ed, or when the stream was
+    /// [`finish()`](crate::SendStream::finish)ed by the peer and all data has been received, after
+    /// which it is no longer meaningful for the stream to be reset.
+    ///
+    /// Notification of a reset is delivered exactly once. After a reset error code has been
+    /// delivered by *any* means, including calls to `reset()` or any operation that returns
+    /// [`ReadError::Reset`], future calls to `reset()` will return `Ok(None)`, and future calls to
+    /// methods that can return [`ReadError`] will return [`ReadError::ClosedStream`].
```
MOZGIII, line 300: I think it would be a bit confusing - might be better to explicitly maintain a notion that the error code has been handed over and return an error.
From the looks of it - it might not be that difficult to copy the error code out instead of consuming it.

---

Is this call cancellation-safe?

Ralith: Made resets persistent and documented cancel-safety.

MOZGIII (approved):
Looks good

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -156,6 +156,29 @@ impl<'a> RecvStream<'a> {
 
         Ok(())
     }
+
+    /// Check whether this stream has been reset by the peer, returning the reset error code if so
+    pub fn reset(&mut self) -> Result<Option<VarInt>, ClosedStream> {
```
djc, line 161: I think this is a bad name? It reads too much like an active verb, when the semantics are ostensibly supposed to be more like `wait_for_reset()`?

quinn/src/recv_stream.rs
```diff
@@ -284,6 +284,54 @@ impl RecvStream {
...
+    ///
+    /// This operation is cancel-safe.
+    pub async fn reset(&mut self) -> Result<Option<VarInt>, ResetError> {
```
djc, line 303: Same here for the name.

quinn/src/recv_stream.rs
```diff
@@ -284,6 +284,54 @@ impl RecvStream {
...
+        poll_fn(|cx| {
+            let mut conn = self.conn.state.lock("RecvStream::reset");
+            if self.is_0rtt && conn.check_0rtt().is_err() {
+                return Poll::Ready(Err(ResetError::ZeroRttRejected));
+            }
+
+            if let Some(code) = self.reset {
+                return Poll::Ready(Ok(Some(code)));
+            }
```
djc, line 312: You documented that notification of a reset would be delivered exactly once, but that doesn't seem to match this code?

quinn/src/recv_stream.rs
```diff
@@ -506,6 +554,15 @@ impl From<ReadableError> for ReadError {
     }
 }
 
+impl From<ResetError> for ReadError {
```
djc, line 557: What's the motivation for this impl?

MOZGIII: This would work nicely for the apps that want to either `read` or wait for `reset` to unify the error type without adding extra types

Ralith: Right, the intention is just to minimize friction for users who might be aggregating errors.

quinn-proto/src/connection/streams/mod.rs
Ralith: I agree that this name sucks. I wish `wait_for_reset().await` wasn't weird and stuttery, though.

djc: Yup. Other people must have run into this...

MOZGIII: I do have the names like this at my `xwt` codebase - `wait_for_bi` when bidirectional connection is opening for instance. I do not even have the irregular verb there, just a different pattern to follow.
So, I would say `wait_for_[...]().await` is not that bad.

---

`has_reset`?

djc: seen_reset/reset_seen/was_reset...

Ralith: Renamed to `received_reset`.

quinn/src/recv_stream.rs
Ralith: This was stale from a previous revision. Corrected, and added a similar warning to the `proto` layer where it *does* hold.

Ralith:
Forgot to actually queue a `max_stream_id` when needed after calling `stream_freed` in proto; fixed.

djc (approved):

PR #1874, opened 2024-05-19
Simpler and more precise ID flow control update check in Chunks

Ralith:
Noticed this looked like a redundant approximation while working on #1873. Careful review desired. The original logic here seems to date back to 2021 or earlier, but e.g. a3b35f93c25045b00849e846713cb9b94bc681c5 looks like a somewhat shorter-sighted hack around this same quirk. See commit description for additional rationale.

### DIFF
--- a/quinn-proto/src/connection/streams/recv.rs
+++ b/quinn-proto/src/connection/streams/recv.rs
@@ -313,10 +313,9 @@ impl<'a> Chunks<'a> {
 
         let mut should_transmit = false;
         // We issue additional stream ID credit after the application is notified that a previously
-        // open stream has finished or been reset and we've therefore disposed of its state.
-        if matches!(state, ChunksState::Finished | ChunksState::Reset(_))
-            && self.streams.side != self.id.initiator()
-        {
+        // open stream has finished or been reset and we've therefore disposed of its state, as
+        // recorded by `stream_freed` calls in `next`.
+        if self.streams.take_max_streams_dirty(self.id.dir()) {
             self.pending.max_stream_id[self.id.dir() as usize] = true;
             should_transmit = true;
         }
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/streams/recv.rs
```diff
@@ -299,10 +299,9 @@ impl<'a> Chunks<'a> {
 
         let mut should_transmit = false;
         // We issue additional stream ID credit after the application is notified that a previously
-        // open stream has finished or been reset and we've therefore disposed of its state.
-        if matches!(state, ChunksState::Finished | ChunksState::Reset(_))
-            && self.streams.side != self.id.initiator()
-        {
+        // open stream has finished or been reset and we've therefore disposed of its state, as
+        // recorded by `stream_freed` calls in `next`.
+        if self.streams.take_max_streams_dirty(self.id.dir()) {
```
djc, line 304: Hmm, maybe we should bump the abstraction level on `take_max_streams_dirty()`, in terms of its name and also maybe have it take a `&mut Retransmits` as an argument?

Ralith: Good idea; done in a separate commit.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -742,8 +742,17 @@ impl StreamsState {
         self.events.pop_front()
     }
 
-    pub(crate) fn take_max_streams_dirty(&mut self, dir: Dir) -> bool {
-        mem::replace(&mut self.max_streams_dirty[dir as usize], false)
+    /// Queues MAX_STREAM_ID frames in `pending` if needed
+    ///
+    /// Returns whether any frames were queued
```
djc, line 747: Nit: end this sentence in a period?

djc (approved):

djc:
Nice!

PR #1869, opened 2024-05-18
Fix stopped recv stream flow control underflow under reordering

Ralith:
Fixes #1818 (probably). Second commit isn't necessary but feels a bit tidier.

PR #1876, opened 2024-05-21
proto: make NoInitialCipherSuite Copy

djc:
`hickory-proto::ProtoErrorKind`, which will include this, implements `Clone`.

Downstream PR: https://github.com/hickory-dns/hickory-dns/pull/2217.

### DIFF
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -1,4 +1,7 @@
-use std::collections::{hash_map, BinaryHeap};
+use std::{
+    collections::{hash_map, BinaryHeap},
+    io,
+};
 
 use bytes::Bytes;
 use thiserror::Error;
@@ -441,6 +444,19 @@ impl ShouldTransmit {
 #[error("closed stream")]
 pub struct ClosedStream {
     _private: (),
+}
+
+impl ClosedStream {
+    #[doc(hidden)] // For use in quinn only
+    pub fn new() -> Self {
+        Self { _private: () }
+    }
+}
+
+impl From<ClosedStream> for io::Error {
+    fn from(x: ClosedStream) -> Self {
+        Self::new(io::ErrorKind::NotConnected, x)
+    }
 }
 
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
--- a/quinn/Cargo.toml
+++ b/quinn/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quinn"
-version = "0.11.0"
+version = "0.11.1"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/quinn-rs/quinn"
 description = "Versatile QUIC transport protocol implementation"
@@ -38,7 +38,7 @@ bytes = "1"
 futures-io = { version = "0.3.19", optional = true }
 rustc-hash = "1.1"
 pin-project-lite = "0.2"
-proto = { package = "quinn-proto", path = "../quinn-proto", version = "0.11", default-features = false }
+proto = { package = "quinn-proto", path = "../quinn-proto", version = "0.11.2", default-features = false }
 rustls = { version = "0.23", default-features = false, features = ["ring", "std"], optional = true }
 smol = { version = "2", optional = true }
 thiserror = "1.0.21"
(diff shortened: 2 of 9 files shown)


quinn/src/connection.rs
```diff
@@ -1255,7 +1255,7 @@ pub enum SendDatagramError {
 const MAX_TRANSMIT_DATAGRAMS: usize = 20;
 
 /// Error indicating that a stream has already been finished or reset
-#[derive(Debug, Error, Clone, PartialEq, Eq)]
+#[derive(Debug, Error, Clone, Copy, PartialEq, Eq)]
```
djc, line 1258: Do you remember why this is a separate type?

Ralith: If you mean to ask why it's separate from the `proto` definition, perhaps that was motivated by trait impls that would otherwise have coherency issues? I'm not sure. If we can reuse the proto-layer type here I don't immediately see a drawback to doing so.

Ralith (approved):
Would it be sufficient to impl just `Clone`? That gives us more freedom to add non-`Copy` fields in the future. It's not obvious to me why we'd ever need to, so no worries if `Copy` confers significant benefit on its own.

quinn/src/connection.rs
djc: > If you mean to ask why it's separate from the `proto` definition, perhaps that was motivated by trait impls that would otherwise have coherency issues? I'm not sure. If we can reuse the proto-layer type here I don't immediately see a drawback to doing so.

I wonder if we can do that semver-compatibly...

Ralith: I'd file it under "technically breaking but probably fine." Downstream could implement a downstream-defined trait separately for `proto::ClosedStream` and `quinn::ClosedStream`, which would break, but:
- why would you ever do that?
- it's trivial to fix by deleting either impl

djc: Okay, please have a look.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -443,6 +446,24 @@ pub struct ClosedStream {
     _private: (),
 }
 
+impl ClosedStream {
+    #[doc(hidden)] // For use in quinn only
```
Ralith, line 450: Oh, this must've been why we split it out before. I don't mind this solution, though. There's also the fact that "has not been opened" in the doc comment doesn't apply at the `quinn` layer, where you can never get a handle to an unopened stream, but I don't think that's a big deal either.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -443,6 +446,24 @@ pub struct ClosedStream {
...
+    pub fn new() -> Self {
+        Self { _private: () }
+    }
+
+    /// Returns whether the stream is closed
+    pub fn is_closed(&self) -> bool {
```
Ralith, line 456: What's this for?

djc: Not sure! Appears to be unused, looks like a bit of debris from an earlier refactoring. Will deprecate.

Ralith: It looks to me like it's newly added in this commit.

quinn/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "quinn"
-version = "0.11.0"
```
Ralith, line 3: Should we skip to 0.11.2 for consistency/less confusing git tags?

quinn-proto/src/connection/streams/mod.rs
djc: Oh, huh.

djc: Killed it.

quinn/Cargo.toml
djc: Skipping versions just feels differently confusing. I think we already have plenty of tags that are confusing in exactly the same way?

Ralith (approved):
If you're planning on publishing a release in the immediate future, note https://github.com/quinn-rs/quinn/pull/1869 is outstanding with a fix for a bug that's a panic in debug mode and possible transport error in release. Can always fast-follow with that if you don't have the bandwidth for review right now though.

djc:
I'll try to review your open PRs before releasing.

PR #1879, opened 2024-05-23
Variable length cids

Ralith:


quinn-proto/src/endpoint.rs
```diff
@@ -561,22 +558,6 @@ impl Endpoint {
             ..
         } = incoming.packet.header;
 
-        if self.cids_exhausted() {
```
thynson, line 564: One option, changing return type of the `ConnectionIdGenerator.generate` to `Result<ConnectionId, SomeError>`, so that if the implementation of `ConnectionIdGenerator` is willing to perform exhausted check, it could return an `Err(CidExhausted)`.  But since it must be a breaking change, I'm not sure if it's worth to do it in this way.

Ralith: These changes are breaking regardless, so we might as well take the cleanest approach. However:
- The generator would need to know the total number of live CIDs, requiring additional complexity
- Large deployments might have a difficult time tracking the number of CIDs in use in a shared space across multiple endpoints
- It's not clear if this check is all that useful to begin with

I'm presently leaning towards a simpler approach that just gives up if it can't generate a unique CID after some number of tries, to at least handle the case where someone is using an extremely small CID space for some reason.

djc:
Since we're open for breaking changes now, time to rebase this?

PR #1884, opened 2024-05-28
Add support for NetBSD, fix OpenBSD

flub:
This adds support for NetBSD.  As mentioned in the commit, it does not support DONTFRAG or ECN for IPv4, but does for IPv6.  It has recvmmsg so I implemented that recv branch.

Also included is a fix for OpenBSD which also does not have IP_DONTFRAG for IPv4.  Not sure how that slipped into #1863 / 98d8513e656467b32898402a372e33ea2108c478.

I have tested this on all of OpenBSD, NetBSD and FreeBSD and all pass the test suite now.

### DIFF
--- a/quinn-udp/src/cmsg/mod.rs
+++ b/quinn-udp/src/cmsg/mod.rs
@@ -124,6 +124,10 @@ pub(crate) trait MsgHdr {
 
     fn cmsg_nxt_hdr(&self, cmsg: &Self::ControlMessage) -> *mut Self::ControlMessage;
 
+    /// Sets the number of control messages added to this `struct msghdr`.
+    ///
+    /// Note that this is a destructive operation and should only be done as a finalisation
+    /// step.
     fn set_control_len(&mut self, len: usize);
 
     fn control_len(&self) -> usize;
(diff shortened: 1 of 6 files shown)


quinn-udp/src/cmsg/mod.rs
```diff
@@ -124,6 +124,10 @@ pub(crate) trait MsgHdr {
 
     fn cmsg_nxt_hdr(&self, cmsg: &Self::ControlMessage) -> *mut Self::ControlMessage;
 
+    /// Sets the number of control messages added to this `struct msghdr`.
+    ///
+    /// Note that this is a destructive operation and should only be done as a finalisation
+    /// step.
```
flub, line 130: This is perhaps a bit unfortunate.  It works as it is used now but I could not figure out a nice way to actually force this into the type system.

The `struct msghdr` is part of a larger struct here, so adding a finalisation step that takes ownership does not work.

The other option I considered was adding a separate `.finish()` method, but apart from being slightly more explicit it has the same problem with being purely conventional.

djc (approved):
Nice, thanks!

Ralith (approved):
Thanks!

quinn-udp/src/cmsg/mod.rs
Ralith: There's a lot in this module that relies on careful usage, so I think this is reasonable as-is, though of course improvements are always welcome.

PR #1887, opened 2024-06-06
During a client connect handshake, wait for Frame::HandshakeDone before considering the connection established

billylindeman:
As it stands today, a quinn client will consider itself established/connected prior to receving the HandshakeDone frame.  

If a connection fails because the server rejects the client (such as a tls certificate issue, or some other reason), the client will have already resolved the connecting.await? into a valid Connection.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -2465,11 +2465,13 @@ impl Connection {
                     // Server-only
                     self.spaces[SpaceId::Data].pending.handshake_done = true;
                     self.discard_space(now, SpaceId::Handshake);
-                }
 
-                self.events.push_back(Event::Connected);
-                self.state = State::Established;
-                trace!("established");
+                    // on server we consider the connection established at this point
+                    // but on client we must wait for HandshakeDone
+                    self.events.push_back(Event::Connected);
+                    self.state = State::Established;
+                    trace!("established");
+                }
                 Ok(())
             }
             Header::Initial(InitialHeader {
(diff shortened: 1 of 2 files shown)


djc:
Do you have a reference to the RFC section that describes this requirement?

billylindeman:
@djc The specific section I see is in the QUIC-TLS rfc: https://www.rfc-editor.org/rfc/rfc9001.html#name-handshake-confirmed

djc:
That section also has this:

> Additionally, a client MAY consider the handshake to be confirmed when it receives an acknowledgment for a 1-RTT packet. This can be implemented by recording the lowest packet number sent with 1-RTT keys and comparing it to the Largest Acknowledged field in any received 1-RTT ACK frame: once the latter is greater than or equal to the former, the handshake is confirmed.

Maybe that is applicable to our earlier implementation?

billylindeman:
@djc, I don't believe quinn implements that section / behavior because the precipitating issue that led to this investigation is in a situation where a client attempts to connect with a server and the server rejects the TLS keys.

In that case, the client considers itself established and then later fails once the client receives the close message - however if quinn was properly waiting for a 1-rtt ACK during the handshake I don't believe it would get to that point?

Ralith:
Thanks for the PR!

It is by design that Quinn yields the `Connection` upon entering the [`Handshake Complete`](https://www.rfc-editor.org/rfc/rfc9001.html#name-handshake-complete) state. The effect of this change is to wait for `Handshake Confirmed` instead. This takes an additional round-trip, *doubling* the expected connection establishment latency. See also RFC9000's [Example Handshake Flows](https://www.rfc-editor.org/rfc/rfc9000.html#name-example-handshake-flows).

Still, I appreciate that it's awkward for there not to be a well-defined location for clients to detect authentication failures when communicating with servers that require client authentication. Would it be sufficient to expose a new method on `Connection` that waits for handshake confirmation?

billylindeman:
@Ralith Thanks for the explanation! I re-read that part of the spec and that makes sense to me.  Having a well-defined place to await handshake confirmation would be great

djc:
> @Ralith Thanks for the explanation! I re-read that part of the spec and that makes sense to me. Having a well-defined place to await handshake confirmation would be great

Do you want to change this PR to address that?

Ralith:
Closing in favor of https://github.com/quinn-rs/quinn/pull/1944.

Issue #1896, opened 2024-06-14
Allow a client to specify Initial Connection ID


thynson:
I'm trying to implement a QUIC-LB routing policy that deals with Initial Packets, which aimed to offer affinity between clients and servers by encoding a client identifier into somewhere either in Source CID or in Destination CID of the Initial Packet.

However encoding it into SourceCID comes with the security concern of link-ability, since the `CidGenerator` has no idea about whether a connection is established yet, such kind of CIDs will be generated and used within the whole lifetime of a connection.

On the other hand, it's better to encode it into Destination CID of a Initial Packet, as long as it satisfy the requirements in [Section 7.2 of the RFC9000](https://www.rfc-editor.org/rfc/rfc9000#name-negotiating-connection-ids). Since it won't be used anymore once the CID negotiation is done, and then the server will provide a Routable CID (defined in https://datatracker.ietf.org/doc/draft-ietf-quic-load-balancers/) so that the load balancer could later route packets based on it.

djc:
Sounds okay, want to propose a PR?

thynson:
Plan to do it this weekend.

PR #1897, opened 2024-06-15
proto: make initial destination cid configurable 

thynson:
This PR implements #1896.

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -20,7 +20,9 @@ use crate::{
     cid_generator::{ConnectionIdGenerator, HashedConnectionIdGenerator},
     congestion,
     crypto::{self, HandshakeTokenKey, HmacKey},
-    VarInt, VarIntBoundsExceeded, DEFAULT_SUPPORTED_VERSIONS, INITIAL_MTU, MAX_UDP_PAYLOAD,
+    shared::ConnectionId,
+    RandomConnectionIdGenerator, VarInt, VarIntBoundsExceeded, DEFAULT_SUPPORTED_VERSIONS,
+    INITIAL_MTU, MAX_CID_SIZE, MAX_UDP_PAYLOAD,
 };
 
 /// Parameters governing the core QUIC state machine
@@ -963,6 +965,9 @@ pub struct ClientConfig {
     /// Cryptographic configuration to use
     pub(crate) crypto: Arc<dyn crypto::ClientConfig>,
 
+    /// Provider that populates the destination connection ID of Initial Packets
+    pub(crate) initial_dst_cid_provider: Arc<dyn Fn() -> ConnectionId + Send + Sync>,
+
     /// QUIC protocol version to use
     pub(crate) version: u32,
 }
@@ -973,8 +978,27 @@ impl ClientConfig {
         Self {
             transport: Default::default(),
             crypto,
+            initial_dst_cid_provider: Arc::new(|| {
+                RandomConnectionIdGenerator::new(MAX_CID_SIZE).generate_cid()
+            }),
             version: 1,
         }
+    }
+
+    /// Configure how to populate the destination CID of the initial packet when attempting to
+    /// establish a new connection.
+    ///
+    /// By default, it's populated with random bytes with reasonable length, so unless you have
+    /// a good reason, you do not need to change it.
+    ///
+    /// When prefer to override the default, please note that the generated connection ID MUST be
+    /// at least 8 bytes long and unpredictable, as per section 7.2 of RFC 9000.
+    pub fn initial_dst_cid_provider(
+        &mut self,
+        initial_dst_cid_provider: Arc<dyn Fn() -> ConnectionId + Send + Sync>,
+    ) -> &mut Self {
+        self.initial_dst_cid_provider = initial_dst_cid_provider;
+        self
     }
 
     /// Set a custom [`TransportConfig`]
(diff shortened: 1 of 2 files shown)


Ralith (approved):
Looks reasonable, thanks!

quinn-proto/src/config.rs
```diff
@@ -623,6 +624,12 @@ pub struct EndpointConfig {
     /// Create a cid generator for local cid in Endpoint struct
     pub(crate) connection_id_generator_factory:
         Arc<dyn Fn() -> Box<dyn ConnectionIdGenerator> + Send + Sync>,
+
+    /// Initial remote CID generator factory
```
djc, line 628: Would it be more correct/comprehensible to call this `dst` instead of `remote` everywhere? IIRC we mostly use `dst` for packets and `remote` for the connection state. Since these initial `dst_cid`s are never acknowledged by the remote peer, I think it would be more clear to refer to this as "destination CID" or `dst_cid` throughout.

@Ralith thoughts?

Ralith: That makes sense to me. Better matches RFC9000 terminology, too.

Ralith (changes requested):
On second thought, should this be a field of `ClientConfig`? It's only needed for outgoing connections, after all.

quinn-proto/src/config.rs
thynson: Indeed, I was hesitated on its name.

thynson:
> On second thought, should this be a field of ClientConfig? It's only needed for outgoing connections, after all.

I tried but was stuck with the mutability of the ClientConfig. Maybe some interior mutability pattern should applied here.


I'll find some time to implement a `ClientConfig` version in a couple of days.

djc:
> > On second thought, should this be a field of ClientConfig? It's only needed for outgoing connections, after all.
> 
> I tried but was stuck with the mutability of the ClientConfig. ~Maybe some interior mutability pattern should applied here~. A mutex is required to do so.
> 
> I'll find some time to implement a `ClientConfig` version in a couple of days.

Not sure what you mean? `EndpointConfig` ends up in an `Arc` inside `Endpoint::config`, too, so how is the mutability of `ClientConfig` different/an issue?

thynson:
> > > On second thought, should this be a field of ClientConfig? It's only needed for outgoing connections, after all.
> > 
> > 
> > I tried but was stuck with the mutability of the ClientConfig. ~Maybe some interior mutability pattern should applied here~. A mutex is required to do so.
> > I'll find some time to implement a `ClientConfig` version in a couple of days.
> 
> Not sure what you mean? `EndpointConfig` ends up in an `Arc` inside `Endpoint::config`, too, so how is the mutability of `ClientConfig` different/an issue?

The `EndpointConfig` holds an `Arc` of a factory, when building an `Endpoint`, an instance of `Box<dyn ConnectionIdGenerator>` is obtained from the factory, so the mutability issue is not applied there.

thynson:
Updated to move it to ClientConfig

quinn-proto/src/endpoint.rs
```diff
@@ -395,7 +395,11 @@ impl Endpoint {
             return Err(ConnectError::UnsupportedVersion);
         }
 
-        let remote_id = RandomConnectionIdGenerator::new(MAX_CID_SIZE).generate_cid();
+        let remote_id = config
+            .initial_dst_cid_generator
+            .lock()
```
thynson, line 400: It would probably be better to change the type of `initial_dst_cid_generator` to `Arc<dyn Fn() -> ConnectionId>`, since `generate_cid` is the only member of `ConnectionIdGenerator` that is used here.

quinn-proto/src/config.rs
```diff
@@ -1015,6 +1039,10 @@ impl fmt::Debug for ClientConfig {
         fmt.debug_struct("ClientConfig<T>")
             .field("transport", &self.transport)
             .field("crypto", &"ClientConfig { elided }")
+            .field(
+                "initial_dst_cid_generator",
+                &"ConnectionIdGenerator { elided }",
```
Ralith, line 1044: `ConnectionIdGenerator` isn't a type, so this is a bit confusing. Maybe just leave it out and switch to `finish_non_exhaustive`?

quinn-proto/src/config.rs
```diff
@@ -973,10 +978,29 @@ impl ClientConfig {
...
+    /// establish a new connection.
+    ///
+    /// By default, it's populated with random bytes with reasonable length, so unless you have
+    /// a good reason, you do not need to change it.
+    ///
+    /// When prefer to override the default, please note that the generated connection ID MUST be
+    /// at least 8 bytes long and unpredictable, as per section 7.2 of RFC 9000.
+    pub fn initial_dst_cid_provider<Provider: Fn() -> ConnectionId + Send + Sync + 'static>(
+        &mut self,
+        provider: Provider,
```
Ralith, line 998: Prefer to take an `Arc<dyn _>` here rather than wrapping it internally. That way if the user already has a boxed trait object, it doesn't have to be re-boxed.

Ralith (changes requested):
In the long run I want to make `ConnectionIdGenerator::generate_cid` take `&self` so we can share them more readily and do away with factories, but that'll be a breaking change regardless, so for simplicity in the short term I think I agree that a function trait object makes the most sense.

djc (approved):
LGTM, thanks!

Ralith (approved):
Thanks!

PR #1898, opened 2024-06-15
Only send MAX_STREAMS when >1/8 of flow control window is consumed 

Ralith:
Minor optimization to reduce outgoing traffic when receiving large numbers of streams. Mirrors existing logic for data flow control, e.g.

https://github.com/quinn-rs/quinn/blob/7b97a3fbd875903568249862f6d2ef4b7d30cfe8/quinn-proto/src/connection/streams/state.rs#L849-L855


djc (approved):
Nice!

PR #1900, opened 2024-06-16
Endpoint stats interface

ryleung-solana:
In Anza, we need to get more insight into the resources being used by quinn in servicing quic handshakes. This PR seeks to add an initial interface to the Endpoint struct to allow for getting more detailed stats than just the open connections, and some initial data relevant to the handshakes. On a side note to the Quinn devs, if there is a good way to support getting the current number of open handshakes, we would be interested in know it. Currently, it seems that the Endpoint is not signaled when a Connecting future finishes, and to add such an event would require some refactoring, not to mention handling an extra event, locking, etc.

### DIFF
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -92,11 +92,15 @@ impl Endpoint {
         let open_connections = state.inner.open_connections() as u64;
         let incoming_handshakes = state.incoming_handshakes;
         let outgoing_handshakes = state.outgoing_handshakes;
+        let refused_handshakes = state.refused_handshakes;
+        let ignored_handshakes = state.ignored_handshakes;
         let incoming_buffer_bytes = state.inner.incoming_buffer_bytes();
         EndpointStats {
             open_connections,
             incoming_handshakes,
             outgoing_handshakes,
+            refused_handshakes,
+            ignored_handshakes,
             incoming_buffer_bytes,
         }
     }
@@ -342,6 +346,8 @@ pub struct EndpointStats {
     pub open_connections: u64,
     pub incoming_handshakes: u64,
     pub outgoing_handshakes: u64,
+    pub refused_handshakes: u64,
+    pub ignored_handshakes: u64,
     pub incoming_buffer_bytes: u64,
 }
 
@@ -443,6 +449,8 @@ impl EndpointInner {
 
     pub(crate) fn refuse(&self, incoming: proto::Incoming) {
         let mut state = self.state.lock().unwrap();
+        state.incoming_handshakes += 1;
+        state.refused_handshakes += 1;
         let mut response_buffer = Vec::new();
         let transmit = state.inner.refuse(incoming, &mut response_buffer);
         respond(transmit, &response_buffer, &*state.socket);
@@ -457,7 +465,10 @@ impl EndpointInner {
     }
 
     pub(crate) fn ignore(&self, incoming: proto::Incoming) {
-        self.state.lock().unwrap().inner.ignore(incoming);
+        let mut state = self.state.lock().unwrap();
+        state.incoming_handshakes += 1;
+        state.ignored_handshakes += 1;
+        state.inner.ignore(incoming);
     }
 }
 
@@ -478,6 +489,8 @@ pub(crate) struct State {
     runtime: Arc<dyn Runtime>,
     incoming_handshakes: u64,
     outgoing_handshakes: u64,
+    refused_handshakes: u64,
+    ignored_handshakes: u64,
 }
 
 #[derive(Debug)]
@@ -697,6 +710,8 @@ impl EndpointRef {
                 runtime,
                 incoming_handshakes: 0,
                 outgoing_handshakes: 0,
+                refused_handshakes: 0,
+                ignored_handshakes: 0,
             }),
         }))
     }
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -332,15 +332,21 @@ impl Endpoint {
     }
 }
 
-/// Struct for holding the stats returned from Endpoint::stats
+/// Statistics on [Endpoint] activity
 #[non_exhaustive]
 #[derive(Debug, Default, Copy, Clone)]
 pub struct EndpointStats {
+    /// Number of open connecctions on this [Endpoint]
     pub open_connections: u64,
+    /// Cummulative number of Quic handshakes accepted by this [Endpoint]
     pub accepted_handshakes: u64,
+    /// Cummulative number of Quic handshakees sent from this [Endpoint]
     pub outgoing_handshakes: u64,
+    /// Cummulative number of Quic handshakes refused on this [Endpoint]
     pub refused_handshakes: u64,
+    /// Cummulative number of Quic handshakes ignored on this [Endpoint]
     pub ignored_handshakes: u64,
+    /// Number of bytes used in all incoming buffers on this [Endpoint]
     pub incoming_buffer_bytes: u64,
 }
 
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -85,8 +85,7 @@ impl Endpoint {
         )
     }
 
-    // Returns relevant stats from this Endpoint
-    //
+    /// Returns relevant stats from this Endpoint
     pub fn stats(&self) -> EndpointStats {
         let state = self.inner.state.lock().unwrap();
         let open_connections = state.inner.open_connections() as u64;
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -47,14 +47,6 @@ pub struct Endpoint {
     pub(crate) default_client_config: Option<ClientConfig>,
     runtime: Arc<dyn Runtime>,
 }
-#[derive(Debug, Default, Copy, Clone)]
-pub struct EndpointStats {
-    pub open_connections: usize,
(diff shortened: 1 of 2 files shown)


quinn/src/endpoint.rs
```diff
@@ -47,6 +47,14 @@ pub struct Endpoint {
     runtime: Arc<dyn Runtime>,
 }
 
+pub struct EndpointStats {
```
Ralith, line 50: To avoid risk of overflow on long-running 32-bit applications and for consistency with `ConnectionStats`, counters should be `u64`.

quinn/src/endpoint.rs
```diff
@@ -47,6 +47,14 @@ pub struct Endpoint {
...
+    open_connections: usize,
+    incoming_handshakes_since_last_sample: usize,
+    total_incoming_handshakes: usize,
+    total_outgoing_handshakes: usize,
+    outgoing_handshakes_since_last_sample: usize,
```
Ralith, line 55: `_since_last_sample` counters are redundant: a caller interested in progression over time can store the previous total and compute changes. This both simpler for us to implement and less lossy, because a caller can aggregate over any time interval. Similar logic probably applies to `open_connections` too.

Ralith, line 55: You probably want these fields to be public.

quinn/src/endpoint.rs
Ralith, line 50:
```suggestion
#[derive(Debug, Default, Copy, Clone)]
pub struct EndpointStats {
```

Ralith (changes requested):
> if there is a good way to support getting the current number of open handshakes, we would be interested in know it

This isn't readily available, because endpoint behavior isn't sensitive to connection state once the application has decided to accept it. What *is* readily available is the number of incoming connection attempts which haven't yet been accepted/rejected/ignored. You could always maintain your own atomic counter that you update by whatever logic you like, of course.

quinn/src/endpoint.rs
```diff
@@ -46,6 +46,14 @@ pub struct Endpoint {
     pub(crate) default_client_config: Option<ClientConfig>,
     runtime: Arc<dyn Runtime>,
 }
+#[derive(Debug, Default, Copy, Clone)]
```
djc, line 49: This should be defined further down, probably between `impl Endpoint` and `struct EndpointDriver(..)`.

ryleung-solana:
@Ralith @djc one thing that we're particularly interested in is related to the resource usage from crypto, especially in accept() because we don't currently reject any incoming connection requests. We merely accept all incoming connections and then immediately drop some Connecting future based on our own logic.

djc:
> @Ralith @djc one thing that we're particularly interested in is related to the resource usage from crypto, especially in accept() because we don't currently reject any incoming connection requests.

Why not? I think we had some Solana people look at that as an option for this exact reason.

ryleung-solana:
> > @Ralith @djc one thing that we're particularly interested in is related to the resource usage from crypto, especially in accept() because we don't currently reject any incoming connection requests.
> 
> Why not? I think we had some Solana people look at that as an option for this exact reason.

Mostly because we currently only do some basic rate-limiting and nothing more. It's probably a good idea to look at using reject when over the rate limit instead of accepting and immediately dropping :sweat: but the thing is that we currently don't really have a great idea of what the limit should be, since we don't have insight into the resource usage from crypto (but we think it's significant).

ryleung-solana:
Would you guys have any ideas of how we might be able to get some better numbers on e.g. the mem usage from crypto in accept() before the future is returned?

Ralith:
Can you elaborate on "mem usage from crypto"? I don't expect cryptographic operations to consume significant memory, or really anything but O(input). Usually the concern with handshakes is CPU usage, which could in principle be measured with some sort of high resolution timer around `accept` and, most importantly, the `poll` calls on the internal connection driver future.

quinn/src/endpoint.rs
```diff
@@ -304,6 +321,16 @@ impl Endpoint {
     }
 }
 
+// Struct for holding the stats returned from Endpoint::stats
+//
+#[derive(Debug, Default, Copy, Clone)]
+pub struct EndpointStats {
+    pub open_connections: u64,
```
Ralith, line 328: Would it be more useful/consistent to lean into the monotonic counter paradigm and count the lifetime number of successfully established and lost connections, rather than reporting the instantaneous live count? I have limited experience with large-scale infrastructure monitoring so maybe I'm getting carried away.

quinn/src/endpoint.rs
```diff
@@ -304,6 +321,16 @@ impl Endpoint {
...
+    pub total_incoming_handshakes: u64,
```
Ralith, line 329: Should we also count total accepted/rejected/ignored handshakes?

quinn/src/endpoint.rs
```diff
@@ -304,6 +321,16 @@ impl Endpoint {
...
+    pub total_outgoing_handshakes: u64,
+    pub incoming_buffers_total_bytes: u64,
```
Ralith, line 331: nit: So far 3/4 of these fields have "total" in the name. Is that adding any clarity? Maybe something like "handshakes_{received, sent}" and "incoming_buffer_bytes"?

quinn/src/endpoint.rs
ryleung-solana: I think we would find it more useful to just sample the number of open connection at that moment.

Ralith: Would that be adequately addressed by using the existing [`Endpoint::open_connections`](https://docs.rs/quinn-proto/latest/quinn_proto/struct.Endpoint.html#method.open_connections) rather than committing to it as part of `EndpointStats`?

quinn/src/endpoint.rs
```diff
@@ -304,6 +325,18 @@ impl Endpoint {
     }
 }
 
+// Struct for holding the stats returned from Endpoint::stats
+//
+#[derive(Debug, Default, Copy, Clone)]
+pub struct EndpointStats {
```
Ralith, line 331:
```suggestion
#[non_exhaustive]
pub struct EndpointStats {
```

ryleung-solana:
> Can you elaborate on "mem usage from crypto"? I don't expect cryptographic operations to consume significant memory, or really anything but O(input). Usually the concern with handshakes is CPU usage, which could in principle be measured with some sort of high resolution timer around `accept` and, most importantly, the `poll` calls on the internal connection driver future.

I think that's (CPU usage) also something to look at as well. We're looking specifically at x509 verification, as it happens before we get the future returned to us when we accept() making it a prime target for rate limiting given some data. It would also allow us to determine once and for all, whether crypto does indeed take a significant amount of memory under attack (e.g. getting spammed with lots of handshake attempts, or some variation of this), which we suspect from a recent security advisory, but don't know for sure.

Ralith:
> We're looking specifically at x509 verification, as it happens before we get the future returned to us when we accept() making it a prime target for rate limiting given some data.

This is no longer the case as of 0.11. No CPU-intensive work is performed until you accept a connection.

ryleung-solana:
CC @t-nelson for input. From the above, it's a little difficult to directly measure memory usage from x509 verification and apparently there are decent ways to rate-limit this in newer versions anyway. I also understand that it was requested to track various state transitions in the handshake, but this doesn't look super feasible.

quinn/src/endpoint.rs
ryleung-solana: Don't really have a strong opinion on this. I just thought it was something we might as well include in the stats interface - maybe eventually superseding [Endpoint::open_connections](https://docs.rs/quinn-proto/latest/quinn_proto/struct.Endpoint.html#method.open_connections) though you will know better than me whether this is appropriate.

quinn/src/endpoint.rs
```diff
@@ -318,6 +338,19 @@ impl Endpoint {
     }
 }
 
+// Struct for holding the stats returned from Endpoint::stats
+//
+#[non_exhaustive]
+#[derive(Debug, Default, Copy, Clone)]
+pub struct EndpointStats {
+    pub open_connections: u64,
+    pub incoming_handshakes: u64,
```
djc, line 347: If `incoming_handshakes` is merely the sum of accepted, refused and ignored handshakes I think we should track those three numbers separately without additionally tracking `incoming_handshake`, since it can trivially derived from the other three (that is, replace this with `accepted_handshakes`).

quinn/src/endpoint.rs
```diff
@@ -318,6 +338,19 @@ impl Endpoint {
     }
 }
 
+// Struct for holding the stats returned from Endpoint::stats
```
djc, line 341: This should be an actual docstring (`///`) without the trailing empty lines.

There should also be a docstring for each of the public fields, in particular taking care to document which of these are monotonically increasing vs other behavior (perhaps we can reflect the difference in the field names, too).

djc: "Struct for holding the stats returned from Endpoint::stats" is a very mechanical description. Maybe something more like "Statistics on [`Endpoint`] activity"?

Also, note that I requested you document each field here.

djc (approved):
LGTM, thanks.

ryleung-solana:
Hmm, I'm unable to merge. Perhaps it's because "@Ralith requested changes"?

djc:
You won't be able to merge either way, only we do. But, I'll wait for @Ralith's approval.

Ralith (approved):
Thanks!

PR #1901, opened 2024-06-17
Allow configuring rng seed through `EndpointConfig`

aochagavia:
@Ralith this is based on your [suggestion](https://discord.com/channels/976380008299917365/1063547094863978677/1247827802896470046) from a while ago

### DIFF
--- a/quinn-proto/src/config.rs
+++ b/quinn-proto/src/config.rs
@@ -629,6 +629,8 @@ pub struct EndpointConfig {
     pub(crate) grease_quic_bit: bool,
     /// Minimum interval between outgoing stateless reset packets
     pub(crate) min_reset_interval: Duration,
+    /// Optional seed to be used internally for random number generation
+    pub(crate) rng_seed: Option<[u8; 32]>,
 }
 
 impl EndpointConfig {
@@ -643,6 +645,7 @@ impl EndpointConfig {
             supported_versions: DEFAULT_SUPPORTED_VERSIONS.to_vec(),
             grease_quic_bit: true,
             min_reset_interval: Duration::from_millis(20),
+            rng_seed: None,
         }
     }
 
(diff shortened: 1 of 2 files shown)


quinn-proto/src/endpoint.rs
```diff
@@ -66,10 +66,11 @@ impl Endpoint {
         config: Arc<EndpointConfig>,
         server_config: Option<Arc<ServerConfig>>,
         allow_mtud: bool,
-        rng_seed: Option<[u8; 32]>,
```
djc, line 69: This is a semver-incompatible change. Not sure we want to merge this as is now?

Ralith: We could retain semver compat by letting this override the endpoint config.

aochagavia: I don't really have a preference, so whatever works for you ;)

Shall I modify the code to apply @Ralith 's suggestion? Or did you have a different idea in mind? I'm also ok waiting till enough breaking changes accumulate to roll this PR up with the rest (I can keep using my fork in the meantime).

djc: Yeah, that idea works for me! We should potentially file a "breaking change" issue for ourselves to consider when we go to release 0.12.

aochagavia: Just reverted the breaking change and updated the documentation. Let me know if there's anything else to be done

quinn-proto/src/config.rs
```diff
@@ -727,6 +730,17 @@ impl EndpointConfig {
         self.min_reset_interval = value;
         self
     }
+
+    /// Optional seed to be used internally for random number generation
+    ///
+    /// By default, quinn will initialize and endpoint's rng using a random seed. However, you can
```
Ralith, line 736: Typo, and "a random seed" is a bit confusing because it could be understood to describe what the data *is* instead of how it's populated.
```suggestion
    /// By default, quinn will initialize an endpoint's rng using a platform entropy source. However, you can
```

Ralith (approved):

quinn-proto/src/config.rs
aochagavia: That's much better wording, thanks! I amended the commit (instead of using GitHub's UI to accept your suggestion) to keep the git history clean.

Issue #1749, opened 2024-01-22
Suggestions for quinn-udp


larseggert:
A few suggestions that would IMO make quinn-udp more useful, in no particular order:
* Conversion from `EcnCodepoint` into `u8`.
* Setting/getting IPv4 DSCP (IPv6 traffic class) info, i.e., the other six bits not used by ECN.
   * With similar `u8` conversion for the whole field and the ECN/DSCP components individually.
* Setting/getting IPv4 TTL (IPv6 hop count) information on packets.
* Pacing offload via GSO (see [quiche](https://github.com/cloudflare/quiche/blob/master/apps/src/sendto.rs#L52-L85))

(NB: Rust newbie here, so some of these things might be doable already, but I couldn't quite figure out how).

Ralith:
> Conversion from EcnCodepoint into u8.

`x as u8` will do this alrady, though the expectation is that you shouldn't need to do that yourself, since quinn-udp's responsible for passing it down

> Setting/getting IPv4 DSCP (IPv6 traffic class) info, i.e., the other six bits not used by ECN. 

This would be a straightforward extension. Any thoughts on how this should be exposed to the application layer by a QUIC impl?

> Setting/getting IPv4 TTL (IPv6 hop count) information on packets.

Is this relevant to QUIC? We've been reluctant to let quinn-udp scope-creep into a generic UDP library.

> Pacing offload via GSO (see [quiche](https://github.com/cloudflare/quiche/blob/master/apps/src/sendto.rs#L52-L85))

Pacing offload be cool to have, but seems pretty niche because it requires manual configuration of the kernel to be effective, so won't be useful for end users/unsophisticated deployments. Is there an easy way to automatically detect when it's available? Also, you say "via GSO", but it looks to me like this entire GSO batch is sent with a single txtime there?

larseggert:
> `x as u8` will do this alrady, though the expectation is that you shouldn't need to do that yourself, since quinn-udp's responsible for passing it down

Doh. Of course.

> > Setting/getting IPv4 DSCP (IPv6 traffic class) info, i.e., the other six bits not used by ECN.
> 
> This would be a straightforward extension. Any thoughts on how this should be exposed to the application layer by a QUIC impl?

"It depends." Some DSCPs such as LE have straightforward semantics for an entire connection. Others may be better used for streams or individual QUIC datagrams (modulo the overall uncertainty if the network is equipped to deal with intra-flow DSCP changes.)

[...]

PR #1903, opened 2024-06-20
fix: remove unused dependency tracing-attributes

mxinden:
Various `quinn*` crates depend on `tracing`. `tracing` in its default feature set includes `tracing-attributes` in order to expose `#[instrument]` procedural macro in `tracing::instrument`.

None of `quinn*` make use of the `#[instrument]` macro. Thus the `tracing-attributes` dependency is unused.

To remove the dependency from the tree, this commit:

1. Restricts `tracing`'s `features` to `std`, effectively removing the `attributes` feature.
2. Consolidates the various `tracing*` imports across `quinn*` crates into workspace dependencies.

Let me know if you don't want to use workspace dependencies?

---

Context:

- Mozilla is currently using NSPR to send and receive QUIC datagrams.
- There is an ongoing effort to replace NSPR with `quinn-udp`, see https://github.com/quinn-rs/quinn/issues/1749.
- This effort is becoming more concrete, see [working implementation in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1901295) and corresponding [tracking issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1901295).
- Each new Rust dependency needs to be audited. The less dependencies (e.g. here `tracing-attributes`) the better.

As always, thank you for this project!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,10 @@ members = ["quinn", "quinn-proto", "quinn-udp", "bench", "perf", "fuzz"]
 default-members = ["quinn", "quinn-proto", "quinn-udp", "bench", "perf"]
 resolver = "2"
 
+[workspace.dependencies]
+tracing = { version = "0.1.10", default-features = false, features = ["std"] }
+tracing-subscriber = { version = "0.3.0", default-features = false, features = ["env-filter", "fmt", "ansi", "time", "local-time"] }
+
 [profile.bench]
 debug = true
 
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -3,6 +3,10 @@ members = ["quinn", "quinn-proto", "quinn-udp", "bench", "perf", "fuzz"]
 default-members = ["quinn", "quinn-proto", "quinn-udp", "bench", "perf"]
 resolver = "2"
 
+[workspace.dependencies]
```
djc, line 6: Would you mind doing the two steps in two commits? Makes it a lot easier to review.

djc:
Very cool to see that quinn-udp is being considered for use in Firefox!

Cargo.toml
mxinden: Done. Mind taking another look @djc?

djc (approved):
LGTM!

Ralith (approved):
Thanks! Always happy to see unused code get trimmed down.

Issue #931, opened 2020-12-03
Allow user to retrieve RTT estimation


jean-airoldie:
From what I understand, by doing a quick grep search, quinn seem to be keeping track of the rtt latency for an active connection.

https://github.com/quinn-rs/quinn/blob/1ea0b897fa50abe97f02ae1b2b8b5c9c4ba24a30/quinn-proto/src/connection/paths.rs#L53

AFAIK this data is currently not publicly accessible, but I think it would make sense to allow the user to retrieve that information.

My specific use case is that I would like to keep track of the rtt latency per connection over time using a histogram. For this to be possible, I would need to both be able to retrieve the latency estimation, but also be notified when it is updated.

With that in mind, a possibility (although a bit complex), would be to allow the user to provide a type per connection that implements a trait such as the following:
```rust
pub trait RttUpdate {
  fn update_rtt(&self, rtt: Duration) {
    // Default impl is a noop.
  }
}
```
Where `RttUpdate::update_rtt` would be called each time the RTT is updated.

djc:
Did you find the code added in #889?

jean-airoldie:
Thx, I didn't see that.

This kinda works for my use case, however its not optimal because I can't know when the rtt was updated. For instance, if I parse the rtt at a constant interval (say every 10 seconds), I can't know whether the rtt was updated or not since the last interval, which would lead to data point duplication.

jean-airoldie:
Looking at ralith's comment:
https://github.com/quinn-rs/quinn/pull/889#pullrequestreview-524708003
> LGTM. I wonder if we might eventually want to switch to exposing a single aggregate statistics structure to avoid churning the lock excessively, but that's an issue for another time.

I think it would be interesting to allow setting a consumer for the quinn connection statistics. IMO this would be way better than having to manually parse the values for each connection at a regular interval, and it would reduce contention for the lock.

Something like:
```rust
trait StatsConsumer {
  fn rrt_update(&self, rtt: Duration);
  fn bytes_recv_update(&self, nb_bytes: usize);
  fn bytes_sent_update(&self, nb_bytse: usze);
  // etc. (idk what statistics would be exposed)
}
```
And then a way to set the consumer for each connection (although setting the consumer before the connection is initiated would be better).
```rust
impl Connection {
  fn set_consumer(&self, consumer: Box<StatsConsumer>) {
    //...
  }
}
```

Ralith:
> its not optimal because I can't know when the rtt was updated

[...]

PR #1910, opened 2024-06-24
Add rtt estimator snapshots, if more accurate rtt analysis is demanded

StarStarJ:
Motivation:

This would allow users to get more accurate RTT information for specific needs like the range of jitter in the rtt. Additionally it was not possible to request the other member variables of the rtt estimator before.
(https://github.com/quinn-rs/quinn/issues/931#issuecomment-2185200803)

This can be useful for more accurate timer tracking between server & client on less stable connections (or a dos attack on the server or similar things)

For example here, we see that the smoothed is _kinda_ outdated
```
[quinn/examples/client.rs:147:5] conn.rtt_estimator_snapshot() = RttEstimatorSnapshot {
    latest: 126.72µs,
    smoothed: Some(
        391.996µs,
    ),
    var: 237.002µs,
    min: 126.72µs,
}
```

On linux you can also use:
`sudo tc qdisc add dev lo root netem delay 300ms 300ms`

to simulate a massive lag, and then the smooth rtt value will only provide a rough estimate of the average latency, not about the range of the rtts.

If you think a whole snapshot is too much, I'd already be happy to get access to just the `latest` member of the rtt estimator, so I could at least fire up my own logic without applying a custom congestion controller.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -60,7 +60,8 @@ mod packet_crypto;
 use packet_crypto::{PrevCrypto, ZeroRttCrypto};
 
 mod paths;
-pub use paths::RttEstimator;
+pub use paths::{RttEstimator, RttEstimatorSnapshot};
+
 use paths::{PathData, PathResponses};
 
 mod send_buffer;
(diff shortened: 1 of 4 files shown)


StarStarJ:
I just included timing information into some of my packets that are send & recv by the client, which is very accurate now.

Since I am probably the only one with my use case, I don't want to bother you with this anymore.

Thanks again for this nice lib.

Ralith:
I'm glad you found a satisfactory solution! I'm not categorically opposed to exposing more granular data, but indeed it's a difficult API design problem to come up with something that's broadly useful without being overly complex.

PR #1913, opened 2024-07-02
quinn: Make `Endpoint::client` dual-stack V6 by default

gretchenfrage:
Following a conversation in the crate discord [[link]](https://discord.com/channels/976380008299917365/1063547094863978677/1230352131924688926) in which @Ralith mentioned:

> maybe `Endpoint::client` should do this by default? I suspect real applications inevitably need to use a lower level constructor, but it's one less footgun

This PR makes makes `quinn::Endpoint::client` use `socket2` to automatically try to set `IPV6_V6ONLY` to false if binding to a V6 address. Modifies the documentation to match. If the call to `set_only_v6` errors, it logs the error as a warning, but proceeds with the rest of the `client` method anyways.

Removes the OS-based gating from `quinn`'s `tests::echo_dualstack` test. Unfortunately(?), that tests seems to work on main for me anyways on my current system, so I'm leaning on the CI to make sure that there's not some common system where dual-stack conversion is always impossible.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ default-members = ["quinn", "quinn-proto", "quinn-udp", "bench", "perf"]
 resolver = "2"
 
 [workspace.dependencies]
+socket2 = "0.5"
 tracing = { version = "0.1.10", default-features = false, features = ["std"] }
 tracing-subscriber = { version = "0.3.0", default-features = false, features = ["env-filter", "fmt", "ansi", "time", "local-time"] }
 
--- a/quinn/Cargo.toml
+++ b/quinn/Cargo.toml
@@ -41,6 +41,7 @@ pin-project-lite = "0.2"
 proto = { package = "quinn-proto", path = "../quinn-proto", version = "0.11.2", default-features = false }
 rustls = { version = "0.23", default-features = false, features = ["ring", "std"], optional = true }
 smol = { version = "2", optional = true }
+socket2 = { workspace = true }
 thiserror = "1.0.21"
 tracing =  { workspace = true }
 tokio = { version = "1.28.1", features = ["sync"] }
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -25,6 +25,7 @@ use proto::{
     EndpointEvent, ServerConfig,
 };
 use rustc_hash::FxHashMap;
+use socket2::{Domain, Protocol, Socket, Type};
 use tokio::sync::{futures::Notified, mpsc, Notify};
 use tracing::{Instrument, Span};
 use udp::{RecvMeta, BATCH_SIZE};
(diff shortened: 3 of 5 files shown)


quinn/src/endpoint.rs
```diff
@@ -54,19 +55,33 @@ impl Endpoint {
...
+    /// the address `[::]:0` is a reasonable default to maximize the ability to connect to other
+    /// address. For example:
+    ///
+    /// ```
+    /// quinn::Endpoint::client((std::net::Ipv6Addr::UNSPECIFIED, 0).into());
+    /// ```
+    ///
+    /// If unable to make the socket dual-stack, a warning is logged, and behavior depends on
+    /// platform defaults, but may involve clients bound to an IPV6 address being unable to connect
+    /// to IPV4 servers and vice versa.
```
Ralith, line 69: This is vague. How about:
```suggestion
    /// Some environments may not allow creation of dual-stack sockets, in which case an IPv6 client will only be able to connect to IPv6 servers.
```
IPv4 sockets are never dual-stack, and if `set_only_v6` fails we can reasonably assume that the socket is not dual-stack.

quinn/src/endpoint.rs
```diff
@@ -54,19 +55,33 @@ impl Endpoint {
...
     #[cfg(feature = "ring")]
     pub fn client(addr: SocketAddr) -> io::Result<Self> {
-        let socket = std::net::UdpSocket::bind(addr)?;
+        let socket = Socket::new(Domain::for_address(addr), Type::DGRAM, Some(Protocol::UDP))?;
+        socket.bind(&addr.into())?;
+        if addr.is_ipv6() {
+            if let Err(e) = socket.set_only_v6(false) {
+                tracing::warn!(%e, "unable to make socket dual-stack");
```
Ralith, line 76: I'm not sure this merits a warning. Consider that some environments will trigger this by design on every single call. Maybe `debug`?

Ralith (changes requested):
Thanks, I think this makes sense.

quinn/src/endpoint.rs
gretchenfrage: Changed to `debug`.

quinn/src/endpoint.rs
gretchenfrage: Added that plus the sentence "An IPv4 client is never dual-stack." at the end to help clarify things.

Cargo.toml
```diff
@@ -6,6 +6,7 @@ resolver = "2"
 [workspace.dependencies]
 tracing = { version = "0.1.10", default-features = false, features = ["std"] }
 tracing-subscriber = { version = "0.3.0", default-features = false, features = ["env-filter", "fmt", "ansi", "time", "local-time"] }
+socket2 = "0.5"
```
djc, line 9: Nit: keep these in alphabetic order please?

quinn/Cargo.toml
```diff
@@ -45,6 +45,7 @@ thiserror = "1.0.21"
 tracing =  { workspace = true }
 tokio = { version = "1.28.1", features = ["sync"] }
 udp = { package = "quinn-udp", path = "../quinn-udp", version = "0.5", default-features = false }
+socket2 = { workspace = true }
```
djc, line 48: This too.

djc (approved):
Thanks!

Ralith (approved):
Thanks!

PR #1915, opened 2024-07-05
bench(udp): measure non-GSO & GSO on localhost

mxinden:
This commit adds a basic benchmark, measuring the time to transfer 10 MiB with and without GSO on localhost.

```
gso_false/throughput    time:   [23.455 ms 25.230 ms 27.004 ms]
                        thrpt:  [370.31 MiB/s 396.35 MiB/s 426.35 MiB/s]

gso_true/throughput     time:   [1.0168 ms 1.1239 ms 1.2496 ms]
                        thrpt:  [7.8153 GiB/s 8.6889 GiB/s 9.6044 GiB/s]
```

```
$ cat /proc/cpuinfo
model name      : 12th Gen Intel(R) Core(TM) i9-12900HK
```

---

Given various optimizations for `localhost` traffic on Linux, we were wondering whether GSO actually has  an impact on `localhost`. According to this benchmark: Yes it does.

For what a `localhost` network benchmark is worth, here it is. Feel free to close in case it isn't useful.

### DIFF
--- a/quinn-udp/Cargo.toml
+++ b/quinn-udp/Cargo.toml
@@ -30,6 +30,6 @@ once_cell = "1.19.0"
 [dev-dependencies]
 criterion = "0.5"
 
-[[bench]]
+[target.'cfg(any(target_os = "linux", target_os = "windows"))'.bench]
 name = "throughput"
 harness = false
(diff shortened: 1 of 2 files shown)


Ralith (approved):
Neat! Seems worth having to me; quinn-udp has some internal logic, so being able to measure the impact of any large changes will help.

PR #1917, opened 2024-07-12
proto: ignore incoming packets for closed connections

djc:


Ralith (changes requested):
This is not the specified behavior. [RFC 9000 §10.2.1](https://www.rfc-editor.org/rfc/rfc9000.html#closing):
> An endpoint in the closing state sends a packet containing a CONNECTION_CLOSE frame in response to any incoming packet that it attributes to the connection.

What's the motivation?

PR #1920, opened 2024-07-17
Fix lack of reexport of ConnectionStats and ResetError

TirushOne:
ConnectionStats and ResetError are now reexported for use with the public interface, they weren't before.

Respecfully, please take better care of your libeary. I encountered these issues and thought I had something wrong because surley something as obvious as not reexporting your own types would have been cought and fixed before publishing the latest release of quinn, but I guess not. But regaurdless, there it is, it's fixed.

### DIFF
--- a/quinn/src/lib.rs
+++ b/quinn/src/lib.rs
@@ -63,8 +63,8 @@ mod work_limiter;
 
 pub use proto::{
     congestion, crypto, AckFrequencyConfig, ApplicationClose, Chunk, ClientConfig, ClosedStream,
-    ConfigError, ConnectError, ConnectionClose, ConnectionError, EndpointConfig, IdleTimeout,
-    MtuDiscoveryConfig, ServerConfig, StreamId, Transmit, TransportConfig, VarInt, ConnectionStats
+    ConfigError, ConnectError, ConnectionClose, ConnectionError, ConnectionStats, EndpointConfig,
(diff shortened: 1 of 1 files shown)


djc (approved):
Thanks for the PR!

> Respecfully, please take better care of your libeary. I encountered these issues and thought I had something wrong because surley something as obvious as not reexporting your own types would have been cought and fixed before publishing the latest release of quinn, but I guess not.

Unfortunately this kind of issue (due to the difference between being "visible" and "nameable") is quite often not all that obvious, I've seen similar issues in several different projects.

FWIW, feedback like "take better care of your library" isn't very constructive, I suggest finding a different way of giving feedback.

TirushOne:
> FWIW, feedback like "take better care of your library" isn't very constructive, I suggest finding a different way of giving feedback.

Well I did make this pr in a fit of frustration and I the feed back isn't the most contructive lol. So I will try to be a bit more contructive next time I have a bug fix

djc:
> Also will this change be coming out in a bug fix patach soon? Or will it be coming with the next scheduled update? I have a project using quinn atm and having this fix available soon would be very conviniant.

There is no schedule for updates, but we plan to push out some maintenance releases soon. Want to get some in-flight PRs merged before we do so, but I think it should happen soon (probably this week).

TirushOne:
> There is no schedule for updates, but we plan to push out some maintenance releases soon. Want to get some in-flight PRs merged before we do so, but I think it should happen soon (probably this week).
Good to hear

PR #1921, opened 2024-07-18
[breaking] deps(udp): make tracing optional and add optional log

mxinden:
This commit makes the `tracing` dependency in `quinn-udp` optional, but enabled by default. If disabled, `log` is used. By adding it to the `default` feature set, this is not a breaking change.

---

Context:
- I am working on integrating `quinn-udp` into Firefox. See [mozilla-central pull request](https://phabricator.services.mozilla.com/D212959).
- Each Rust dependency in Firefox needs to be audited via [`cargo vet`](https://github.com/mozilla/cargo-vet).
    ```
    ➜  mozilla-unified git:(quic-rust-udp-io) ✗ ./mach cargo vet 
    Vetting Failed!
    
    3 unvetted dependencies:
      tracing:0.1.37 missing ["safe-to-deploy"]
      tracing-attributes:0.1.24 missing ["safe-to-deploy"]
      tracing-core:0.1.30 missing ["safe-to-deploy"]
    
    recommended audits for safe-to-deploy:
        Command                                      Publisher  Used By                                   Audit Size
        cargo vet inspect tracing-attributes 0.1.24  hawkw      tracing                                   5025 lines
        cargo vet inspect tracing-core 0.1.30        hawkw      tracing                                   7032 lines
        cargo vet inspect tracing 0.1.34             hawkw      h2, warp, hyper, quinn-udp, and 2 others  11016 lines
    
    estimated audit backlog: 23073 lines
    ```

- `quinn-udp` optionally not depending on `tracing` will safe Mozilla a lot of work auditing all of `tracing`.

Questions:
- Would you be open to making `tracing` optional in `quinn-udp`? 
- If yes, how would you want to proceed with the `log` feature? See inline comment below.

Follow-up to https://github.com/quinn-rs/quinn/pull/1473 and https://github.com/quinn-rs/quinn/pull/1903

### DIFF
--- a/quinn-udp/Cargo.toml
+++ b/quinn-udp/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quinn-udp"
-version = "0.5.2"
+version = "0.6.0"
 edition = "2021"
 rust-version = "1.66"
 license = "MIT OR Apache-2.0"
--- a/quinn-udp/Cargo.toml
+++ b/quinn-udp/Cargo.toml
@@ -14,14 +14,14 @@ workspace = ".."
 all-features = true
 
 [features]
-default = ["log"]
-# Write logs via the `log` crate when no `tracing` subscriber exists
-log = ["tracing/log"]
+default = ["tracing-log"]
+tracing-log = ["tracing", "tracing/log"]
 
 [dependencies]
 libc = "0.2.113"
+log = { workspace = true, optional = true }
 socket2 = { workspace = true }
-tracing = { workspace = true }
+tracing = { workspace = true, optional = true }
 
 [target.'cfg(windows)'.dependencies]
 once_cell = { workspace = true }
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -13,6 +13,7 @@ use std::{
 };
 
 use socket2::SockRef;
+use tracing::{debug, error};
 
 use super::{
     cmsg, log_sendmsg_error, EcnCodepoint, RecvMeta, Transmit, UdpSockRef, IO_ERROR_LOG_INTERVAL,
@@ -87,7 +88,7 @@ impl UdpSocketState {
         if is_ipv4 || !io.only_v6()? {
             if let Err(err) = set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_RECVTOS, OPTION_ON)
             {
-                tracing::debug!("Ignoring error setting IP_RECVTOS on socket: {err:?}",);
+                debug!("Ignoring error setting IP_RECVTOS on socket: {err:?}",);
             }
         }
 
@@ -283,7 +284,7 @@ fn send(
                         // Prevent new transmits from being scheduled using GSO. Existing GSO transmits
                         // may already be in the pipeline, so we need to tolerate additional failures.
                         if state.max_gso_segments() > 1 {
-                            tracing::error!("got transmit error, halting segmentation offload");
+                            error!("got transmit error, halting segmentation offload");
                             state
                                 .max_gso_segments
                                 .store(1, std::sync::atomic::Ordering::Relaxed);
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -12,7 +12,10 @@ use std::{
     time::Instant,
 };
 
+#[cfg(all(feature = "log", not(feature = "tracing")))]
+use log::{debug, error};
 use socket2::SockRef;
+#[cfg(feature = "tracing")]
 use tracing::{debug, error};
 
 use super::{
@@ -86,9 +89,11 @@ impl UdpSocketState {
         // older macos versions also don't have the flag and will error out if we don't ignore it
         #[cfg(not(any(target_os = "openbsd", target_os = "netbsd")))]
         if is_ipv4 || !io.only_v6()? {
+            #[allow(unused_variables)]
             if let Err(err) = set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_RECVTOS, OPTION_ON)
             {
-                debug!("Ignoring error setting IP_RECVTOS on socket: {err:?}",);
+                #[cfg(any(feature = "tracing", feature = "log"))]
+                debug!("Ignoring error setting IP_RECVTOS on socket: {err:?}");
             }
         }
 
@@ -284,6 +289,7 @@ fn send(
                         // Prevent new transmits from being scheduled using GSO. Existing GSO transmits
                         // may already be in the pipeline, so we need to tolerate additional failures.
                         if state.max_gso_segments() > 1 {
+                            #[cfg(any(feature = "tracing", feature = "log"))]
                             error!("got transmit error, halting segmentation offload");
                             state
                                 .max_gso_segments
--- a/quinn/Cargo.toml
+++ b/quinn/Cargo.toml
@@ -45,7 +45,7 @@ socket2 = { workspace = true }
 thiserror = { workspace = true }
 tracing =  { workspace = true }
 tokio = { workspace = true }
-udp = { package = "quinn-udp", path = "../quinn-udp", version = "0.5", default-features = false, features = ["tracing"] }
+udp = { package = "quinn-udp", path = "../quinn-udp", version = "0.6", default-features = false, features = ["tracing"] }
 
 [dev-dependencies]
 anyhow = { workspace = true }
--- a/quinn/Cargo.toml
+++ b/quinn/Cargo.toml
@@ -28,7 +28,7 @@ runtime-async-std = ["async-io", "async-std"]
 runtime-smol = ["async-io", "smol"]
 
 # Write logs via the `log` crate when no `tracing` subscriber exists
-log = ["tracing/log", "proto/log", "udp/log"]
+log = ["tracing/log", "proto/log", "udp/tracing-log"]
 
 [dependencies]
 async-io = { workspace = true, optional = true }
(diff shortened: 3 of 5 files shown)


quinn-udp/Cargo.toml
```diff
@@ -14,14 +14,13 @@ workspace = ".."
 all-features = true
 
 [features]
-default = ["log"]
-# Write logs via the `log` crate when no `tracing` subscriber exists
-log = ["tracing/log"]
+default = ["tracing"]
```
mxinden, line 17: Note that this pull request proposes dropping the `log` **feature** from `quinn-udp`. A user using `quinn-udp` through `quinn` can enable the `tracing/log` feature through `quinn`'s `log` feature.

Now this does introduce some inconsistency. Alternatively we could:

- Rename the `log` feature to `tracing-log` to disambiguate from the `log` crate. Though I assume you would count this as a breaking change.
- Keep the `log` feature on `quinn-udp`, accepting the confusion that the `log` feature does not enable the `log` crate in `quinn-udp`, but actually forwards to the `tracing/log` feature.

djc: Yes, this is definitely a semver-incompatible change for quinn-udp. I think that's probably acceptable, but this PR should contain a commit that bumps the quinn-udp version number to 0.6.0 (and similar for quinn's dependency on quinn-udp).

Because I don't think we want to make a semver-incompatible release for the quinn crate right now, we should take care to make sure that the behavior for quinn is unchanged, but that should give us the freedom to change the quinn-udp interface as we see fit.

quinn-udp/src/unix.rs
```diff
@@ -87,7 +92,7 @@ impl UdpSocketState {
         if is_ipv4 || !io.only_v6()? {
             if let Err(err) = set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_RECVTOS, OPTION_ON)
             {
-                tracing::debug!("Ignoring error setting IP_RECVTOS on socket: {err:?}",);
+                debug!("Ignoring error setting IP_RECVTOS on socket: {err:?}",);
```
djc, line 95: IMO it would be nice to isolate these changes (that import the tracing event functions) in a separate commit from the introduction of the alternative impl.

djc:
Definitely unexcited about this change, but I suppose I can live with it.

Does the cargo vet output imply that mozilla-central already depends on h2 and warp, or just that those are the most popular crates.io crates that depend on tracing?

quinn-udp/src/unix.rs
mxinden: Sounds good. Extracted as a separate commit with https://github.com/quinn-rs/quinn/pull/1921/commits/fd4ede82da3669453aa97a4e580f7b75dc0e1537. Let me know if this is what you meant.

quinn-udp/Cargo.toml
djc: Per my previous comment, for compatibility with the previous situation we should make sure `default` in quinn-udp enables `tracing/log`. I think we should make that a separate feature `tracing-log = ["tracing", "tracing/log"]`, and change `default` to enable `tracing-log` instead of just `tracing`.

quinn-udp/src/unix.rs
```diff
@@ -14,6 +14,8 @@ use std::{
 
 use socket2::SockRef;
 
+use tracing::{debug, error};
```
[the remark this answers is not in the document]
mxinden: @djc note that that would have `cargo fmt` format it as:

``` rust
#[cfg(not(feature = "tracing"))]
use log::{debug, error};
use socket2::SockRef;
#[cfg(feature = "tracing")]
use tracing::{debug, error};
```

I opted for the empty line to have the two logging imports close to each other. What is your preference?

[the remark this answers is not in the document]
djc: Ah, that makes sense. Given that there's only one other external import, I think it's probably still fine to keep them together.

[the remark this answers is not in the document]
mxinden: > I think it's probably still fine to keep them together.

"them" as in `tracing` and `log` or as in `socket2`, `tracing` and `log`?

quinn-udp/Cargo.toml
mxinden: That makes sense. Thanks for the suggestion. Can you take a look at https://github.com/quinn-rs/quinn/pull/1921/commits/55f4a8d124a23ff5df881fc8c8b1a03f58d733a6?

quinn-udp/src/unix.rs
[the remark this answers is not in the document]
djc: Sorry -- I don't want the empty line.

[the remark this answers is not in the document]
mxinden: Thanks for clarifying. I updated https://github.com/quinn-rs/quinn/pull/1921/commits/2f197c088150aef7e2652024274cff71de11aa17.

mxinden:
I don't think the `freebsd` CI check failure is related:

```
  thread 'tests::key_update_simple' panicked at quinn-proto/src/tests/mod.rs:996:5:
  assertion failed: `None` does not match `Some(Event::Stream(StreamEvent::Readable { id })) if id == s`
  
  
  failures:
      tests::key_update_simple
  
  test result: FAILED. 232 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.94s
  
  error: test failed, to rerun pass `-p quinn-proto --lib`
```

quinn-udp/Cargo.toml
```diff
@@ -14,14 +14,14 @@ workspace = ".."
...
 [features]
-default = ["log"]
-# Write logs via the `log` crate when no `tracing` subscriber exists
-log = ["tracing/log"]
+default = ["tracing-log"]
+tracing-log = ["tracing", "tracing/log"]
 
 [dependencies]
 libc = "0.2.113"
+log = { workspace = true }
```
djc, line 22: Oh, forgot that this should be `optional = true`.

quinn/Cargo.toml
```diff
@@ -45,7 +45,7 @@ socket2 = { workspace = true }
 thiserror = { workspace = true }
 tracing =  { workspace = true }
 tokio = { workspace = true }
-udp = { package = "quinn-udp", path = "../quinn-udp", version = "0.5", default-features = false }
+udp = { package = "quinn-udp", path = "../quinn-udp", version = "0.5", default-features = false, features = ["tracing"] }
```
djc: As mentioned previously, this should be `tracing-log`.

quinn-udp/Cargo.toml
mxinden: @djc what should happen if the user enables neither the `tracing`, `tracing-log` nor `log` feature? Should there simply be no logging at all?

Fine by me. Though to me not intuitive. Given that `log` is so ubiquitous, I don't think anyone minds the non-optional dependency.

quinn/Cargo.toml
mxinden: Setting `features = ["tracing-log"]` would effectively make `tracing/log` non-optional. Am I missing something?

quinn-udp/Cargo.toml
djc: tracing is mostly an alternative to logging, and most modern applications that I have seen get rid of log in favor of tracing. So yes, if both log and tracing are disabled, downstream crate will get no logging at all, which seems right.

quinn/Cargo.toml
djc: I guess we should not (unconditionally) enable `tracing` here in general, it is enabled by default via the quinn `default` features anyway as before.

quinn-udp/Cargo.toml
mxinden: Reasoning makes sense. I updated https://github.com/quinn-rs/quinn/pull/1921/commits/91a667c3c00a1e63224c2dea0f0dfb9f7ba6f32f where both `tracing` and `log` are optional. `tracing` is the default. `tracing` takes precedence over `log`. Let me know if this looks good to you @djc. Thank you for bearing with me.

djc (approved):

djc:
Thanks, this looks good to me.

Ralith (approved):
Thanks!

Ralith:
Note that `quinn-udp` is presently a public dependency of `quinn`. I don't recall offhand if there's a very good reason for this, but it does mean we can't update quinn to use quinn-udp 0.6 without a breaking change.

I think we could avoid the breaking change in quinn-udp in this change. For example, we could have a feature `native-log = ["dep:log"]` and retain the existing `log` feature. That may or may not be worth the trouble/weirdness depending on whether we expect other breaking quinn-udp changes soon.

djc:
> I think we could avoid the breaking change in quinn-udp in this change. For example, we could have a feature `native-log = ["dep:log"]` and retain the existing `log` feature. That may or may not be worth the trouble/weirdness depending on whether we expect other breaking quinn-udp changes soon.

I think that makes sense as a temporary solution until the next breaking change? I think the only downside it the slightly weird feature name, which seems manageable if that lets us retain semver compatibility.

mxinden:
Good catch @Ralith. Thanks.

I created https://github.com/quinn-rs/quinn/pull/1923 implementing your suggestion above.

> That may or may not be worth the trouble/weirdness depending on whether we expect other breaking quinn-udp changes soon.

Happy to prepare an additional clean-up pull request that can be merged once a breaking change release is in sight.

What do you think @djc and @Ralith?

Ralith:
LGTM! Merged https://github.com/quinn-rs/quinn/pull/1923.

PR #1924, opened 2024-07-19
Update Connection::close docs for new API

flub:
In the new API the finish call no longer returns a future and to achieve the same behaviour the future from stopped needs to be used.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -280,6 +280,11 @@ impl Future for ConnectionDriver {
 /// automatically closed with an `error_code` of 0 and an empty `reason`. You can also close the
 /// connection explicitly by calling [`Connection::close()`].
 ///
+/// Closing the connection immediately abandons efforts to deliver data to the peer.  Upon
+/// receiving CONNECTION_CLOSE the peer *may* drop any stream data not yet delivered to the
+/// application. [`Connection::close()`] describes in more detail how to gracefully close a
+/// connection without losing application data.
+///
 /// May be cloned to obtain another handle to the same connection.
 ///
 /// [`Connection::close()`]: Connection::close
@@ -400,19 +405,35 @@ impl Connection {
 
     /// Close the connection immediately.
     ///
-    /// Pending operations will fail immediately with [`ConnectionError::LocallyClosed`]. Delivery
-    /// of data on unfinished streams is not guaranteed, so the application must call this only
-    /// when all important communications have been completed, e.g. by calling [`finish`] on
-    /// outstanding [`SendStream`]s and waiting for the resulting futures to complete.
+    /// Pending operations will fail immediately with [`ConnectionError::LocallyClosed`]. No
+    /// more data is sent to the peer and the peer may drop buffered data upon receiving
+    /// the CONNECTION_CLOSE frame.
     ///
     /// `error_code` and `reason` are not interpreted, and are provided directly to the peer.
     ///
     /// `reason` will be truncated to fit in a single packet with overhead; to improve odds that it
     /// is preserved in full, it should be kept under 1KiB.
     ///
+    /// # Gracefully closing a connection
+    ///
+    /// Only the peer last receiving application data can be certain that all data is
+    /// delivered. The only reliable action it can then take is to close the connection,
+    /// potentially with a custom error code. The delivery of the final CONNECTION_CLOSE
+    /// frame is very likely if both endpoints stay online long enough, and
+    /// [`Endpoint::wait_idle()`] can be used to provide sufficient time. Otherwise, the
+    /// remote peer will time out the connection, provided that the idle timeout is not
+    /// disabled.
+    ///
+    /// The sending side can not guarantee all stream data is delivered to the remote
+    /// application. It only knows the data is delivered to the QUIC stack of the remote
+    /// endpoint. Once the local side sends a CONNECTION_CLOSE frame in response to calling
+    /// [`close()`] the remote endpoint may drop any data it received but is as yet
+    /// undelivered to the application, including data that was acknowledged as received to
+    /// the local endpoint.
+    ///
     /// [`ConnectionError::LocallyClosed`]: crate::ConnectionError::LocallyClosed
-    /// [`finish`]: crate::SendStream::finish
-    /// [`SendStream`]: crate::SendStream
+    /// [`Endpoint::wait_idle()`]: crate::Endpoint::wait_idle
+    /// [`close()`]: Connection::close
     pub fn close(&self, error_code: VarInt, reason: &[u8]) {
         let conn = &mut *self.0.state.lock("close");
         conn.close(error_code, Bytes::copy_from_slice(reason), &self.0.shared);


djc (approved):

quinn/src/connection.rs
```diff
@@ -387,7 +387,8 @@ impl Connection {
     /// Pending operations will fail immediately with [`ConnectionError::LocallyClosed`]. Delivery
     /// of data on unfinished streams is not guaranteed, so the application must call this only
     /// when all important communications have been completed, e.g. by calling [`finish`] on
-    /// outstanding [`SendStream`]s and waiting for the resulting futures to complete.
+    /// outstanding [`SendStream`]s followed by waiting for the future returned by [`stopped`]
+    /// to complete.
```
Ralith, line 391: While technically correct, I think this recommends an anti-pattern. As discussed in https://github.com/quinn-rs/quinn/issues/1926#issuecomment-2239719487, inspecting the state of a stream after you call `finish` on it is highly error-prone. I think it would be best to drop the example entirely, or recommend a more idiomatic approach such as "..., e.g. by waiting to receive application-layer acknowledgements", which also accounts for ensuring that transmitted data is actually processed by the peer application.

Ralith (changes requested):

quinn/src/connection.rs
flub: I still think `.finish()` and waiting until all stream packets are ACKed is an important usecase.  I think this must be documented clearly somehow.

It is important because if you drop the `Connection` it will implicitly close it, dropping any un-acknowledged data.  And afaik there is no way to gracefully know when a connection can be safely closed otherwise.  So you must be able to know you have no more un-ACKed data for your connection somehow.

I think it makes sense to add a note explaining that an application-level acknowledgement could be preferred.  But than you still might want to use this mechanism of being told the transport has ACKed everything for that application-level ACK.

I think it's also fine to expand on this to explain that an application-level ack could be preferred.  I'm all for giving lots of guidance in API docs.  Here my attempt at a draft, as extra paragraph:

> Note that an application-level acknowledgement is often preferred.  When QUIC says all data is acknowledged this only means the remote QUIC stack has received the data, not that the application has received the data, which can also be important for your application to know.

flub: I should also note we have some tests which follow this pattern to signal from application-space that both sides are done with the connection:

```rust
send_stream.finish()?;
send_stream.stopped().await?;
recv_stream.read_to_end(0).await?;
```

I believe this essentially is application-space signalling in both directions that the bidirectional stream is done.  Is there a problem with this pattern?

Ralith: > if you drop the Connection it will implicitly close it, dropping any un-acknowledged data

In general, data that has been acknowledged at the transport level *may still be lost* if you close a connection before the peer application has processed it. The peer transport layer might not bother preserving data for closed connections, or there might be a buffering proxy in between you and the peer, or the peer application might abandon remaining work as soon as the connection becomes closed.

> you still might want to use this mechanism of being told the transport has ACKed everything for that application-level ACK

You can't use a transport-level ACK as an application-level ACK. The point of an application-level ACK is that it's only sent after the application has processed the data it acknowledges, which necessarily happens at some point after the transport-level ACK might have already been sent. That's exactly why relying on a transport-level ACK for application semantics is usually incorrect.

>  Is there a problem with this pattern?

It won't cause spurious failures, but it might not be accomplishing anything. If both sides always call `SendStream::finish` and then `RecvStream::read_to_end` before dropping the associated `SendStream`, then neither side will ever issue a stop, because the `RecvStream` will never be dropped while open. Further, if you already know at the application layer that no more data will arrive on the `RecvStream`, it probably doesn't matter to you whether it gets finished or stopped at the transport layer, so you can likely drop the `read_to_end(0)` as well.

flub: My apologies that an attempt to fix the docs resulted in another discussion about terminating connections.  However since it is about what should be documented I guess it is still on-topic.  Let me know however if this should move elsewhere.

If there is a bi-directional stream and the connection needs to be closed, there is always going to be one side which **receives** the last bit of data and never sends application data back.  This side has it easy: it knows things are done and can close the connection.  It can even do so with a custom error code to allow the peer to know things exited cleanly rather than unexpectedly, not that it can do much about it anyway.

If I understand you correctly this is about the only way you can correctly terminate a connection.  I think this is a good pattern.  Maybe it should be documented clearly as a section under `Connection` that extensively discusses orderly connection termination.

I don't think this is the only safe way to terminate.  [Section 3.1. "Sending  Stream States" from RFC 9000](https://www.rfc-editor.org/rfc/rfc9000.html#name-sending-stream-states) documents the `Data Recvd` state in figure 2.  If you consider a connection which has exactly one `SendStream` between connections (or also combined with a single `RecvStream` as part of a bidirectional stream), when that `SendStream` reached the `Data Recvd` state there is nothing the sender can do anymore.  This is a terminal state, the stream state is dropped.  This is a useful point for the application to know.  It knows at this point there is no other stream, and no more state for this stream.  If it closes the connection now there is no in transit data which can get lost.

You can argue that this application protocol is badly designed and it should allow for new streams to be initiated by the peer in case it did not manage to reach the `Data Read` state from section "3.2 Receiving Stream States".  I consider there to be plenty of situations where it is entirely fine to only care about not losing in transit data however (e.g. almost our entire test suite - which is a bunch of ad-hoc application level protocols).

> Further, if you already know at the application layer that no more data will arrive on the RecvStream, it probably doesn't matter to you whether it gets finished or stopped at the transport layer, so you can likely drop the read_to_end(0) as well.

The final read_to_end(0) asserts that the peer followed the protocol correctly and finished the stream.  It is I think the only way to read the End Of Stream in the API.  (on traditional filedescriptors I would expect a successful read of empty data to signal EOF/EOS).  If that EOS is not received you can log an error and debug why the peer did not behave as expected.


Going back to this doc comment, maybe you would prefer both options to be explained in more neutral language:

> Close the connection immediately.
>
> Pending operations will fail immediately with `ConnectionError::LocallyClosed`. Delivery of data on unfinished streams is not guaranteed, so the application must call this only when all important communications have been completed, e.g. after the last received data from the peer indicated the connection can be closed.
>
> If it is only needed to know there is no data in transit before closing the connection [`RecvStream::read_to_end`] can be used for incoming data. While for outgoing data [`SendStream::finish`] together with [`SendStream::stopped`] can be used. Note especially on the sending side this does not guarantee the peer **application** has processed the data, only that no more data is in transit for the streams.
>
> error_code and reason are not interpreted, and are provided directly to the peer.
>
> reason will be truncated to fit in a single packet with overhead; to improve odds that it is preserved in full, it should be kept under 1KiB.

djc: @flub can't speak for @Ralith but I think the attempt of depth/research here to improve our documentation is great!

Ralith: +1. I appreciate the thoughtful discussion; it's exactly what we need to work out the best documentation and/or API design to provide the best user experience in the long run.

> Maybe it should be documented clearly as a section under Connection that extensively discusses orderly connection termination.

I think discussing close patterns on `Connection::close` makes a lot of sense, but a pointer from the main `Connection` docs for visibility wouldn't hurt.

> [`Data Recvd`] is a terminal state, the stream state is dropped. This is a useful point for the application to know. [...] If it closes the connection now there is no in transit data which can get lost.

This is exposed by `SendStream::stopped`, and while we discourage relying on that signal, we will continue to expose it. If we assume that "data received by peer transport" implies "peer has the opportunity to read it", then it is indeed a useful signal. However, that's at odds with the specified semantics of closing a QUIC connection:

[RFC 9000 § 10.2-7](https://www.rfc-editor.org/rfc/rfc9000.html#section-10.2-7):

> Once its closing or draining state ends, an endpoint SHOULD discard all connection state

In other words, data received by the peer's transport layer *can still be lost* if the connection is closed before the peer's application layer consumes it. A conforming implementation "SHOULD" return `Err(ConnectionLost)` on reads immediately rather than continuing to yield data.

While some implementations, Quinn included, may err on the side of retaining state, application protocols which rely on this capability will not be portable between QUIC implementations. Such protocols may even be broken by reverse proxies that may not be under the developer's control. This is a strong reason to discourage such designs.

> The final read_to_end(0) asserts that the peer followed the protocol correctly and finished the stream. 

Right. This is generally not *necessary* to correctly implement an application protocol, but may be desirable. On the other hand, you might want to be permissive per the robustness principle.

> It is I think the only way to read the End Of Stream in the API

Most read operations can yield `Ok(None)` to indicate that the stream is finished. However, `read_to_end(0)` is a particularly convenient way to make this specific assertion.

> maybe you would prefer both options to be explained in more neutral language:

So long as I'm unconvinced that the `Data Recvd` state is useful to robust applications, I would prefer to explicitly discourage patterns that rely on it, if we describe them at all.

flub: > > Maybe it should be documented clearly as a section under Connection that extensively discusses orderly connection termination.
> 
> I think discussing close patterns on `Connection::close` makes a lot of sense, but a pointer from the main `Connection` docs for visibility wouldn't hurt.

Ok, let's continue to figure out what to write here first.


> [RFC 9000 § 10.2-7](https://www.rfc-editor.org/rfc/rfc9000.html#section-10.2-7):
> 
> > Once its closing or draining state ends, an endpoint SHOULD discard all connection state

- The **closing state** is entered upon *initiating* an immediate close, i.e. sending the CONNECTION_CLOSE frame.
- The **draining state** is entered upon *receiving* a CONNECTION_CLOSE frame.

It is unclear to me from the RFC is what the draining state really entails.  I apparently interpreted this to assume it would drain the local data.  The name clearly threw me off.

[§ 10.2.2](https://www.rfc-editor.org/rfc/rfc9000.html#name-draining-connection-state) says that the only difference is that when draining an endpoint must not send any packets beyond an optional single CONNECTION_CLOSE frame upon first entering the draining state.  While in the closing state it can respond with CONNECTION_CLOSE frames to any packet it receives on the connection.
 
> In other words, data received by the peer's transport layer _can still be lost_ if the connection is closed before the peer's application layer consumes it. A conforming implementation "SHOULD" return `Err(ConnectionLost)` on reads immediately rather than continuing to yield data.

Does the RFC say this somewhere?  It would be a lot clearer if it did.  I find 10.2.2 unnecessarily vague right now.

> While some implementations, Quinn included, may err on the side of retaining state, application protocols which rely on this capability will not be implementable on other QUIC implementations. Such protocols may even be broken by reverse proxies that may not be under the developer's control. This is a strong reason to discourage such designs.

I guess there's an argument for dropping the state sooner than to avoid surprises.  It fooled me.


Here a new draft:


> Close the connection immediately.
>
> Pending operations will fail immediately with `ConnectionError::LocallyClosed`. No more data is sent to the peer and the peer may drop any stream data upon receiving the 
CONNECTION_CLOSE frame.
>
> error_code and reason are not interpreted, and are provided directly to the peer.
>
> reason will be truncated to fit in a single packet with overhead; to improve odds that it is preserved in full, it should be kept under 1KiB.
>
> # Gracefully closing a connection
>
> To close a connection without data loss the application protocol has to be explicit in it's close notification.  Only the peer last receiving application data can be certain that all data is delivered.  The only reliable action it can take now is to close the connection, potentially with a custom error code. The delivery of the final CONNECTION_CLOSE frame is very likely if both endpoints stay online long enough. If not however the remote peer will eventually abandon the connection anyway.
>
> In particular be aware the sending side can not guarantee all stream data is delivered to the remote application, only to the remote QUIC stack of the endpoint. And once the local side sends a CONNECTION_CLOSE frame in response to calling `close` the remote endpoint may drop any undelivered data. Even data that was acknowledged as received to the local endpoint.

I guess the last paragraph isn't strictly necessary, but it is nice to explain the reasoning better.

I was considering "will drop data" rather than "may drop data.  But I guess "may" is clear enough here.

Ralith: > It is unclear to me from the RFC is what the draining state really entails

"draining" refers to "draining in-flight QUIC packets from the network path", not any aspect of local connection state.

> Does the RFC say [received data should be dropped on close] somewhere?

Yes, the text I quoted. Data which has been received by the transport but not read by the application is part of the connection state. "discard all connection state" means dropping that. I ran this by the implementers' slack as a sanity-check, and indeed this is the commonly accepted reading, and other major QUIC implementations (e.g. quic-go) do implement those semantics.

> I guess there's an argument for dropping the state sooner than to avoid surprises.

Yeah, I agree we should consider that, especially given the precedent from other impls.

> Here a new draft:

LGTM overall. I'm not sure what "be explicit in it's close notification" means (and you want "its", not "it's"), but the remaining text seems clear; maybe drop that sentence? Also, rather than "will eventually abandon the connection anyway", maybe "will time out the connection, provided that the idle timeout is not disabled", to put it in more concrete terms.

Other editorial nits: s/In particular be aware the/The/, s/And once/Once/, s/data. Even data/data, including data/

> I guess the last paragraph isn't strictly necessary

I think it's really valuable context, and will be helpful for folks to understand why the prior recommendations are made.

> I was considering "will drop data" rather than "may drop data. But I guess "may" is clear enough here.

"may" is also more correct. We can't in general make guarantees about peer behavior, and e.g. Quinn doesn't drop data like that currently.

flub: Trying to write code that does **lose** data according to what was discussed here I am still finding things I can not explain.

As section 10.2.1 states:
> In the closing state, an endpoint retains only enough information to generate a packet containing a CONNECTION_CLOSE frame and to identify packets as belonging to the connection.

And the draining state, entered by those receiving this CONNECTION_CLOSE frame, also follows this behaviour IIUC.

From what we said earlier this means as soon as a CONNECTION_CLOSE frame is received the acknowledged-but-undelivered application data can be dropped and the application will receive an error code.

However I could not find this behaviour, in this example the server reads application data after the client has closed the connection.  Yet this code panics as provided, but adding the `send_stream.stopped().await.unwrap()` to the client - that is waiting for all data to be ACKed - makes it run fine.  I would expect this to fail in both cases.

```rust
async fn run_server(endpoint: Endpoint) {
    let incoming = endpoint.accept().await.unwrap();
    let conn = incoming.await.unwrap();
    let mut recv_stream = conn.accept_uni().await.unwrap();
    let mut buf = [0u8; 5];
    recv_stream.read_exact(&mut buf).await.unwrap();

    println!("[server] sleeping");
    tokio::time::sleep(Duration::from_secs(3)).await;

    println!("[server] reading remainder");
    recv_stream.read_exact(&mut buf).await.unwrap();
    recv_stream.read_to_end(0).await.unwrap();

    conn.close(1u8.into(), b"the end");
}

async fn run_client(endpoint: Endpoint, server_addr: SocketAddr) {
    let conn = endpoint
        .connect(server_addr, "localhost")
        .unwrap()
        .await
        .unwrap();
    let mut send_stream = conn.open_uni().await.unwrap();
    send_stream.write_all(b"hello").await.unwrap();
    send_stream.write_all(b"world").await.unwrap();
    send_stream.finish().unwrap();
    // send_stream.stopped().await.unwrap();
    conn.close(2u8.into(), b"boom");
    println!("[client] closed")
}
```

With or without the `.stopped()` call I expect the server to panic, because of the sleep it should have received and acknowledged all data already.

I wonder if delivering the "world" and FIN data to the server is compliant with the spec in 10.2.1?  It is also not entirely clear to me why without the `.stopped()` call the error is exactly when reading "world".

Full code:
https://gist.github.com/flub/3f6d19fb126d1f9dd39fbf58349ec744

Ralith: > However I could not find this behaviour

Right, Quinn is more permissive than the spec requires here. That may or may not be a good thing.

> With or without the .stopped() call I expect the server to panic, because of the sleep it should have received and acknowledged all data already.

With the `.stopped().await`, you ensure that all of the data was received by the peer at the transport layer before sending a close. Without it, it may not even be sent. Quinn's current behavior on the read side is to allow applications to consume any data that was already received regardless of connection state. This behavior is neither required nor forbidden by the RFC.

> It is also not entirely clear to me why without the .stopped() call the error is exactly when reading "world".

Probably just a race condition between application and transport logic on the client. Data is sent concurrently with application logic, but this halts as soon as the connection is closed.

flub: thanks, makes sense.

flub:
Any comments on the updated text in the PR?  Is this the right track, does it need further tweaking?

quinn/src/connection.rs
```diff
@@ -280,6 +280,10 @@ impl Future for ConnectionDriver {
 /// automatically closed with an `error_code` of 0 and an empty `reason`. You can also close the
 /// connection explicitly by calling [`Connection::close()`].
 ///
+/// Closing the connection will make both peers immediately drop all the stream data not yet
+/// delivered to the application. [`Connection::close()`] describes in more detail how to
```
Ralith, line 284: This claim is too strong: Quinn currently does, and other conforming implementations may, allow previously received data to be read after a connection is closed, although as we discussed this is not safe to rely on. How about something like:

> Closing the connection will immediately abandon efforts to deliver data to the peer.

quinn/src/connection.rs
```diff
@@ -400,19 +404,35 @@ impl Connection {
 
     /// Close the connection immediately.
     ///
-    /// Pending operations will fail immediately with [`ConnectionError::LocallyClosed`]. Delivery
-    /// of data on unfinished streams is not guaranteed, so the application must call this only
-    /// when all important communications have been completed, e.g. by calling [`finish`] on
-    /// outstanding [`SendStream`]s and waiting for the resulting futures to complete.
+    /// Pending operations will fail immediately with [`ConnectionError::LocallyClosed`]. No
+    /// more data is sent to the peer and the peer may drop any stream data upon receiving
```
Ralith, line 408:
```suggestion
    /// more data is sent to the peer and the peer may drop buffered data upon receiving
```
Streams aren't the only form of application data, and we should be clear that we're talking about data the peer already received at the transport layer, i.e. buffered data.

quinn/src/connection.rs
```diff
@@ -400,19 +404,35 @@ impl Connection {
...
     ///
     /// `error_code` and `reason` are not interpreted, and are provided directly to the peer.
     ///
     /// `reason` will be truncated to fit in a single packet with overhead; to improve odds that it
     /// is preserved in full, it should be kept under 1KiB.
     ///
+    /// # Gracefully closing a connection
+    ///
+    /// Only the peer last receiving application data can be certain that all data is
+    /// delivered. The only reliable action it can take now is to close the connection,
```
Ralith, line 419: editorial:
```suggestion
    /// delivered. The only reliable action it can then take is to close the connection,
```

quinn/src/connection.rs
```diff
@@ -400,19 +404,35 @@ impl Connection {
...
+    /// potentially with a custom error code. The delivery of the final CONNECTION_CLOSE
+    /// frame is very likely if both endpoints stay online long enough,
```
Ralith, line 421: editorial:
```suggestion
    /// frame is very likely if both endpoints stay online long enough, and
```

quinn/src/connection.rs
```diff
@@ -400,19 +404,35 @@ impl Connection {
...
+    /// [`Endpoint::wait_idle()`] can be used to provide sufficient time. If not however, the
```
Ralith, line 422: editorial:
```suggestion
    /// [`Endpoint::wait_idle()`] can be used to provide sufficient time. Otherwise, the
```

quinn/src/connection.rs
```diff
@@ -400,19 +404,35 @@ impl Connection {
...
+    /// remote peer will time out the connection, provided that the idle timeout is not
+    /// disabled.
+    ///
+    /// The sending side can not guarantee all stream data is delivered to the remote
+    /// application, it only knows the data is delivered to the QUIC stack of the remote
```
Ralith, line 427:
```suggestion
    /// application. It only knows the data is delivered to the QUIC stack of the remote
```

quinn/src/connection.rs
```diff
@@ -400,19 +404,35 @@ impl Connection {
...
+    /// endpoint. Once the local side sends a CONNECTION_CLOSE frame in response to calling
+    /// [`close()`] the remote endpoint may drop any data it received but is as yet
+    /// undelivered to the application. Including data that was acknowledged as received to
```
Ralith, line 430:
```suggestion
    /// undelivered to the application, including data that was acknowledged as received to
```

Ralith (changes requested):
Looks good overall! Some editorial nits, and I think you forgot to rework the `Connection` docs after our discussion.

Ralith:
> Could you be more specific?

Just the stuff you fixed in your latest commit. Thanks!

Ralith (approved):
LGTM, thanks for working through this with us! I think this will be a big help to clarify good patterns for users going forwards, and save us all time in explaining the subtleties.

djc (approved):
This is looking great, thanks for all the efforts!

PR #1925, opened 2024-07-19
Bump some dependencies (tests done)

Sovenok-Hacker:


djc:
All of these are semver-compatible, and Cargo defaults to allowing semver-compatible updates by default (similar to specifying `^` explicitly in other semver tools like npm). As such, Cargo will select newer dependencies by default without us making these changes, so this PR is unnecessary.

Sovenok-Hacker:
> All of these are semver-compatible, and Cargo defaults to allowing semver-compatible updates by default (similar to specifying `^` explicitly in other semver tools like npm). As such, Cargo will select newer dependencies by default without us making these changes, so this PR is unnecessary.

Ok

PR #1923, opened 2024-07-19
[non-breaking] deps(udp): make tracing optional and add optional log

mxinden:
Alternative to https://github.com/quinn-rs/quinn/pull/1921.

Implementation of @Ralith suggestion in https://github.com/quinn-rs/quinn/pull/1921#issuecomment-2237230567.

Not breaking `quinn-udp`. I.e. only bumping patch version to `v0.5.3`.

PR #1930, opened 2024-07-22
refactor!: rename log feature to tracing-log

mxinden:
Previously `quinn*` would provide the `log` feature to log events via `log` if no `tracing` subscriber exists.

Later https://github.com/quinn-rs/quinn/pull/1923 allowed `quinn-udp` to log via `log` directly, making `tracing` an optional dependency. For that, it introduced the `direct-log` feature, a workaround name in order to not introduce a breaking change.

This commit cleans up the above, renaming the `log` feature to `tracing-log` and the `direct-log` to `log`. This is a breaking change and thus `quinn-udp` is bumped to `v0.6.0`.

See https://github.com/quinn-rs/quinn/pull/1921 for the full history.

---

Only merge before the next breaking change of `quinn`.

I am opening this up early as it is (a) still fresh in my mind, (b) in order to keep track of it and (c) make the next breaking change easier. Feel free to close if you don't think it is useful.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -27,7 +27,7 @@ jobs:
         #       | paste -sd ',' -
         run: |
           cargo llvm-cov \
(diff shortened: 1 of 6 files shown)


mxinden:
Assuming based on https://github.com/quinn-rs/quinn/pull/2371/ that latest `main` is open for breaking changes. If so, @Ralith and @djc mind including this patch?

djc:
LGTM, thanks for the ping!

Ralith (approved):

PR #1931, opened 2024-07-22
udp: avoid warnings about unused import on non-Linux platforms

djc:
This is a little ugly, but I don't like to see spurious warnings when developing on macOS.

Alternatively, I suppose we could suppress `unused_imports` here?

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -13,10 +13,14 @@ use std::{
 };
 
 #[cfg(all(feature = "direct-log", not(feature = "tracing")))]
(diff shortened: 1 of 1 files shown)


mxinden (approved):

Ralith:
I think a more complete and maintainable to this and a number of other hazards would be to introduce our own trivial internal logging facade as described in https://github.com/quinn-rs/quinn/pull/1932#pullrequestreview-2192415008.

Ralith:
Closing in favor of https://github.com/quinn-rs/quinn/pull/1935.

PR #1933, opened 2024-07-23
chore(quinn): feature flag socket2 imports

mxinden:
The `socket2` imports are only used by the feature flagged `client` function. Feature flag the imports as well.

Avoids the following warning:

```
  --> quinn/src/endpoint.rs:28:15
   |
28 | use socket2::{Domain, Protocol, Socket, Type};
   |               ^^^^^^  ^^^^^^^^  ^^^^^^  ^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
```

### DIFF
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -25,6 +25,7 @@ use proto::{
     EndpointEvent, ServerConfig,
 };
 use rustc_hash::FxHashMap;
+#[cfg(feature = "ring")]
 use socket2::{Domain, Protocol, Socket, Type};
 use tokio::sync::{futures::Notified, mpsc, Notify};
 use tracing::{Instrument, Span};


quinn/src/endpoint.rs
```diff
@@ -68,6 +67,8 @@ impl Endpoint {
     /// client will only be able to connect to IPv6 servers. An IPv4 client is never dual-stack.
     #[cfg(feature = "ring")]
     pub fn client(addr: SocketAddr) -> io::Result<Self> {
+        use socket2::{Domain, Protocol, Socket, Type};
```
djc, line 70: I usually prefer to have guards on the imports -- better to keep this noise outside the main code and make the import blocks a little noisier IMO.

djc:
CI failures are due to #1936.

djc (approved):

djc:
Thanks!

Ralith (approved):

PR #1932, opened 2024-07-22
fix(udp): feature flag tracing in windows.rs

mxinden:
8712910a made `tracing` optional and added optional `log`, but forgot to update `quinn-udp/src/windows.rs`. This commit fixes the oversight and bumps the `quinn-udp` version to `v0.5.4`.

---

Discovered in https://github.com/mozilla/neqo/pull/2002.

I assume that this is not an issue for most `quinn` users as they don't run `quinn-udp` with `default-features = false"`.

I am sorry for the oversight. I can look into a GitHub CI workflow that prevents this in the future tomorrow.

PR #1934, opened 2024-07-23
ci(rust.yml): add job testing feature permutations

mxinden:
This pull request adds a GitHub job, testing all feature permutations (including optional dependencies), across all default workspace members. It uses [`cargo hack`](https://github.com/taiki-e/cargo-hack).

This workflow would for example have caught the following recent bugs before landing in `main`:
- https://github.com/quinn-rs/quinn/pull/1932
- https://github.com/quinn-rs/quinn/pull/1933

Right now this job takes 5m 1s on Ubuntu, 4m 42s on MacOS  and 8m 29s Windows. In case you consider that too long, we could drop `--optional-deps` or exclude some features (`--exclude-features`).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -149,3 +149,16 @@ jobs:
         api-level: ${{ matrix.api-level }}
         arch: ${{ matrix.emulator-arch }}
         script: .github/workflows/rust-android-run-tests-on-emulator.sh
+
+  features:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+    env:
+      RUSTFLAGS: -Dwarnings
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - uses: taiki-e/install-action@cargo-hack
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -91,3 +91,16 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - uses: EmbarkStudios/cargo-deny-action@v2
+
+  check:
```
djc, line 95: I think this name is confusing. My proposal would be just `features`.

djc (approved):
I think this is okay, and the runtime is probably worth it.

.github/workflows/rust.yml
mxinden: Note that this does actually execute a `cargo check`. The only difference to a plain `cargo check` is that it does so across all feature permutations. Thus the name is consistent with the above `test` which executes a `cargo test`.

With that in mind, what do you think @djc? I don't feel strongly about it. Happy to change to `features`.

djc: There are lots of CI jobs that employ (or could employ) `check` so IMO `check` is not distinctive enough to describe the goal of this job, which is to exercise the Cargo features.

mxinden:
```suggestion
  features:
```

mxinden:
@djc I did the rename. Mind taking another look?

djc (approved):
LGTM.

Ralith (approved):
Thanks!

Ralith:
Looks like this needs a rebase.

PR #1935, opened 2024-07-23
refactor(udp): introduce log facade

mxinden:
Add `log` facade that either dispatches to `tracing` or `log`, or to a no-op implementation.

As suggested in https://github.com/quinn-rs/quinn/pull/1932#pullrequestreview-2192415008.

### DIFF
--- a/quinn-udp/src/lib.rs
+++ b/quinn-udp/src/lib.rs
@@ -37,11 +37,6 @@ use std::{
     time::{Duration, Instant},
 };
 
-#[cfg(all(feature = "direct-log", not(feature = "tracing")))]
-use log::warn;
-#[cfg(feature = "tracing")]
-use tracing::warn;
-
 #[cfg(any(unix, windows))]
 mod cmsg;
 
@@ -57,6 +52,29 @@ mod imp;
 #[cfg(not(any(unix, windows)))]
 #[path = "fallback.rs"]
 mod imp;
+
+#[allow(unused_imports, unused_macros)]
+mod log {
+    #[cfg(all(feature = "direct-log", not(feature = "tracing")))]
+    pub(crate) use log::{debug, error, info, trace, warn};
+
+    #[cfg(feature = "tracing")]
+    pub(crate) use tracing::{debug, error, info, trace, warn};
+
+    #[cfg(not(any(feature = "direct-log", feature = "tracing")))]
+    mod no_op {
+        macro_rules! trace    ( ($($tt:tt)*) => {{}} );
+        macro_rules! debug    ( ($($tt:tt)*) => {{}} );
+        macro_rules! info     ( ($($tt:tt)*) => {{}} );
+        macro_rules! log_warn ( ($($tt:tt)*) => {{}} );
+        macro_rules! error    ( ($($tt:tt)*) => {{}} );
+
+        pub(crate) use {debug, error, info, log_warn as warn, trace};
+    }
+
+    #[cfg(not(any(feature = "direct-log", feature = "tracing")))]
+    pub(crate) use no_op::*;
+}
 
 pub use imp::UdpSocketState;
 
@@ -139,7 +157,7 @@ fn log_sendmsg_error(
     let last_send_error = &mut *last_send_error.lock().expect("poisend lock");
     if now.saturating_duration_since(*last_send_error) > IO_ERROR_LOG_INTERVAL {
         *last_send_error = now;
-        warn!(
+        log::warn!(
         "sendmsg error: {:?}, Transmit: {{ destination: {:?}, src_ip: {:?}, enc: {:?}, len: {:?}, segment_size: {:?} }}",
             err, transmit.destination, transmit.src_ip, transmit.ecn, transmit.contents.len(), transmit.segment_size);
     }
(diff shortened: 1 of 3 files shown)


quinn-udp/src/lib.rs
```diff
@@ -38,9 +38,20 @@ use std::{
 };
 
 #[cfg(all(feature = "direct-log", not(feature = "tracing")))]
-use log::warn;
+use log;
 #[cfg(feature = "tracing")]
-use tracing::warn;
+use tracing as log;
```
Ralith, line 43: This punning makes it easy to inadvertently use an item that doesn't exist under a different feature configuration. A more robust approach would be to have an unconditionally defined internal log module that only exports exactly the intended interface, implemented as macros that are unconditionally defined, and only have variation in their bodies.

Ralith (changes requested):

mxinden:
Good point @Ralith. Thanks. Mind taking another look?

quinn-udp/src/log.rs
```diff
@@ -0,0 +1,21 @@
+#![allow(unused_imports, unused_macros)]
```
djc, line 1: I'd prefer to just keep all this in `lib.rs` (somewhere near where the previous imports where) instead of in a small separate file.

quinn-udp/src/log.rs
```diff
@@ -0,0 +1,21 @@
...
+
+#[cfg(all(feature = "direct-log", not(feature = "tracing")))]
+pub(crate) use log::{debug, error, info, trace, warn};
```
Ralith, line 4: This is still more fragile than necessary. Rather than reexporting macros based on cfg flags, we should define our own macros that have a uniform interface, with cfg-dependent expansions.

mxinden: @Ralith are you suggesting something along the lines of:

``` rust
macro_rules! log_trace {
    ($($arg:tt)*) => {
        #[cfg(all(feature = "direct-log", not(feature = "tracing")))]
        log::trace!($($arg)*);
        #[cfg(feature = "tracing")]
        tracing::trace!($($arg)*);
        #[cfg(not(any(feature = "direct-log", feature = "tracing")))]
        // no-op
        {};
    };
}
```

Both `tracing` as well as `log` are very liberal in what they accept, e.g. both accept any number of _Token Trees_ (i.e. `($arg::tt)*`):

https://docs.rs/log/latest/src/log/macros.rs.html#152-159

https://docs.rs/tracing/latest/src/tracing/macros.rs.html#1823-1830

While we could define our own macros as above, exposing one uniform interface, namely `($arg::tt)*`, that would not actually be any stricter than what this pull request implements today, as `($arg::tt)*` already allows any input already.

Am I missing something? What source of bugs would your suggestion prevent @Ralith?

Ralith: My thinking was that while they both expose very general interfaces, this facade probably shouldn't, because their generalities may not have much in common. Reviewing the docs, though, I guess they actually have a lot of overlap? Maybe it's fine. CI will help in any case, and this is a strict improvement.

quinn-udp/src/unix.rs
```diff
@@ -289,8 +285,7 @@ fn send(
                         // Prevent new transmits from being scheduled using GSO. Existing GSO transmits
                         // may already be in the pipeline, so we need to tolerate additional failures.
                         if state.max_gso_segments() > 1 {
-                            #[cfg(any(feature = "tracing", feature = "direct-log"))]
-                            error!("got transmit error, halting segmentation offload");
+                            crate::log::error!("got transmit error, halting segmentation offload");
```
Ralith, line 288: Why import `debug` but qualify `error`?

Ralith (approved):

quinn-udp/src/unix.rs
mxinden: This section has a `cfg` above, namely `#[cfg(target_os = "linux")]`. Importing `error!` would require:

``` diff
+ #[cfg(target_os = "linux")]
+ use super::error;
  use super::{
      cmsg, log::debug, log_sendmsg_error, EcnCodepoint, RecvMeta, Transmit, UdpSockRef,
      IO_ERROR_LOG_INTERVAL,
  };
```

Your call. I am fine with either.

quinn-udp/src/log.rs
mxinden: With the recent changes, `log.rs` increased in size. @djc would you still like me to move it into `lib.rs`? Fine by me.

quinn-udp/src/unix.rs
Ralith: That seems like a good reason to me.

quinn-udp/src/log.rs
mxinden: Friendly ping @djc. Do you still want me to move the code above to `lib.rs`?

djc: Yes, please.

djc (approved):

mxinden:
As discussed, I moved the `log` `mod` back into `lib.rs`. @djc would you mind taking another look?

PR #1937, opened 2024-07-24
Discard pre-handshake packets after the handshake

Ralith:
Attempting to parse frames from these non-frame-bearing packets was responsible for some spurious connection failures in interop testing under high packet loss.


djc (approved):

djc:
MSRV failure is unrelated.

PR #1938, opened 2024-07-24
Raise default idle timeout to 30 seconds

Ralith:
Recommended in RFC9308 § 3.2 for use on the public internet when a NAT may be on the path.


djc (approved):

djc:
(MSRV failure is fixed in #1939.)

PR #1939, opened 2024-07-24
Bump MSRV to 1.70 following tokio 1.39

djc:


quinn-proto/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "quinn-proto"
 version = "0.11.4"
 edition = "2021"
-rust-version = "1.67"
+rust-version = "1.70"
```
mxinden, line 5: Maybe move to workspace `Cargo.toml`?

mxinden (commented):
:+1:

Ralith (approved):

quinn-proto/Cargo.toml
Ralith: Is that a thing? Would be neat if so.

mxinden: Yes. See https://github.com/quinn-rs/quinn/pull/1940.

PR #1944, opened 2024-07-27
Expose the Handshake Confirmed state

Ralith:
This is convenient for some client authentication use cases, and for tests to know when it's safe to force a key update.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -349,6 +349,22 @@ impl Connection {
         }
     }
 
+    /// Wait for the connection to enter the [`Handshake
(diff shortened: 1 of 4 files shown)


Ralith:
This probably needs a test that it actually fails if client auth is required and not successful...

Ralith:
Ah, `reject_missing_client_cert` already covers that :tada:

quinn/src/connection.rs
```diff
@@ -349,6 +349,22 @@ impl Connection {
         }
     }
 
+    /// Wait for the connection to entire the [`Handshake
```
djc, line 352: "entire" -> "enter"?

djc (approved):
Nice!

PR #1946, opened 2024-07-31
Fix incorrect initial DCID indexing on retried connections

Ralith:
This could also cause other packets received before the server accepts the connection to be lost.

Noticed while checking some unrelated assumptions.


djc:
Is this a regression? Why do you think no one has noticed so far?

Ralith:
This was a regression from 0.10, yeah. I'm guessing folks aren't using retries (they add an RTT, after all), and/or the memory leak and lost packets are insignificant.

PR #1948, opened 2024-07-31
refactor(fuzz): add `#[cfg(fuzzing)]`

mxinden:
`quinn-proto` exposes some hooks for fuzzing only. These are behind a `#[cfg(fuzzing)]`.

The `fuzz` crate imports these hooks, but without the `#[cfg(fuzzing)]`.

Running `cargo check --workspace` thus fails importing each of these hooks. For example:

```
error[E0432]: unresolved import `proto::fuzzing`
  --> fuzz/fuzz_targets/packet.rs:7:5
   |
7  |     fuzzing::{PacketParams, PartialDecode},
   |     ^^^^^^^ could not find `fuzzing` in `proto`
```

This commit adds the various `#[cfg(fuzzing)]` to `fuzz/`, preventing the above errors.

---

Needed for https://github.com/quinn-rs/quinn/pull/1934.

### DIFF
--- a/fuzz/fuzz_targets/streams.rs
+++ b/fuzz/fuzz_targets/streams.rs
@@ -1,13 +1,19 @@
 #![no_main]
 
+#[cfg(fuzzing)]
 use arbitrary::Arbitrary;
+#[cfg(fuzzing)]
 use libfuzzer_sys::fuzz_target;
 
 extern crate proto;
+#[cfg(fuzzing)]
 use proto::fuzzing::{ConnectionState, ResetStream, Retransmits, StreamsState};
+#[cfg(fuzzing)]
 use proto::{Dir, Side, StreamId, VarInt};
+#[cfg(fuzzing)]
 use proto::{SendStream, Streams};
 
+#[cfg(fuzzing)]
 #[derive(Arbitrary, Debug)]
 struct StreamParams {
     side: Side,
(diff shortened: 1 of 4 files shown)


fuzz/fuzz_targets/streams.rs
```diff
@@ -1,13 +1,19 @@
 #![no_main]
 
+#[cfg(fuzzing)]
```
mxinden, line 3: The `cfg` statements here are noisy. Would you prefer a single `mod` behind a `cfg`?

Ralith:
I'm not sure this is the right approach. `#[cfg(...)]` *disables* code, so you're really just skipping the crate entirely by verbose means. What do other projects using this fuzzer do? Maybe we should exclude the crate from the workspace by default?

djc:
> I'm not sure this is the right approach. `#[cfg(...)]` _disables_ code, so you're really just skipping the crate entirely by verbose means. What do other projects using this fuzzer do? Maybe we should exclude the crate from the workspace by default?

Yeah, I think using the workspace `default-members` config value might be a better approach.

mxinden:
I wasn't aware of `default-members`. Thank you. Agreed, the status quo is much cleaner.

https://github.com/quinn-rs/quinn/blob/2d06eef43fec927b0cf8f960bedb814bf3e4cc79/Cargo.toml#L1-L3

PR #1954, opened 2024-08-02
quinn: bump version to 0.11.3

djc:


Ralith (approved):
Shouldn't we get -proto too?

djc:
It was done already in #1918.

PR #1955, opened 2024-08-03
No workspace lints

Ralith:
After publishing quinn-proto 0.11.4, building downstream crates (even quinn itself) fails with:

```
error: failed to download `quinn-proto v0.11.4`

Caused by:
  unable to get packages from source

Caused by:
  failed to parse manifest at `/home/ralith/.cargo/registry/src/index.crates.io-6f17d22bba15001f/quinn-proto-0.11.4/Cargo.toml`

Caused by:
  error inheriting `lints` from workspace root manifest's `workspace.lints`

Caused by:
  failed to find a workspace root
```

This reproduces on cargo nightly, 1.80, and 1.79. quinn-proto 0.11.4 has been yanked.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,3 @@ debug = true
 
 [profile.release]
 debug = true
-
-[workspace.lints.rust]
-# https://rust-fuzz.github.io/book/cargo-fuzz/guide.html#cfgfuzzing
-unexpected_cfgs = { level = "warn", check-cfg = ['cfg(fuzzing)'] }
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -43,5 +43,6 @@ rcgen = { workspace = true }
 tracing-subscriber = { workspace = true }
 lazy_static = "1"
 
-[lints]
-workspace = true
+[lints.rust]
+# https://rust-fuzz.github.io/book/cargo-fuzz/guide.html#cfgfuzzing
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(fuzzing)'] }
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quinn-proto"
-version = "0.11.4"
+version = "0.11.5"
 edition.workspace = true
 rust-version.workspace = true
 license.workspace = true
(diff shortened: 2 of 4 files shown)


digizeph:
I encountered the same issue. Cargo error messages as below:

```
error: failed to download `quinn-proto v0.11.4`

Caused by:
  unable to get packages from source

Caused by:
  failed to parse manifest at `/Users/mingwei/.cargo/registry/src/index.crates.io-6f17d22bba15001f/quinn-proto-0.11.4/Cargo.toml`

Caused by:
  error inheriting `lints` from workspace root manifest's `workspace.lints`

Caused by:
  failed to find a workspace root
```


**Temporary workaround**: removing the lines from the cargo registry cache's toml file fixed this issue.
```
[lints]
workspace = true
```

Ralith:
If you're affected by this, `cargo update` will roll you back to the working 0.11.3 due to the yank.

djc (approved):

djc:
Pinged the Cargo team about this on Zulip.

weihanglo:
Some versions of 1.79-beta and 1.80-nightly are indeed broken. @Ralith could you share which toolchain version (`cargo -vV` for every toolchain on your machine) you used when publishing `quinn-proto`? We need to make sure if it has been fixed or is a new regression. Thank you.

djc:
(It looks like an 1.80.1 is going out soon so might be interesting if we can identify this soonish.)

weihanglo:
Just note that the upcoming 1.80.1 change in Cargo (rust-lang/cargo#14325) has nothing to do with this issue. It was <https://github.com/rust-lang/cargo/issues/13843> and fixed by <https://github.com/rust-lang/cargo/pull/13852> before any stable releases.

Ralith:
I probably have tens if not hundreds of Rust toolchains on my machine, and went through several investigating this, but I *think* the original `publish` operation was with:

```
> cargo -vV
cargo 1.80.0-nightly (431db31d0 2024-05-28)
release: 1.80.0-nightly
commit-hash: 431db31d0dbeda320caf8ef8535ea48eb3093407
commit-date: 2024-05-28
host: x86_64-unknown-linux-gnu
libgit2: 1.7.2 (sys:0.18.3 vendored)
libcurl: 8.6.0-DEV (sys:0.4.72+curl-8.6.0 vendored ssl:OpenSSL/1.1.1w)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: NixOS 23.11.0 [64-bit]
```

I suppose I should take care to publish with stable cargo versions in the future.

PR #1956, opened 2024-08-04
Bump quinn -> quinn-proto version requirement

Ralith:
Needed for e.g. the HandshakeConfirmed event.

Noticed this *just* before publishing.


Ralith:
This may be unnecessary as of #1957, but at this point I want to either carefully audit all the changes since 0.11.3, or just play it safe.

Ralith:
Reviewing those commits I think we're okay now.

PR #1962, opened 2024-08-08
Enable use of aws-lc-rs instead of ring

djc:
Still needs a CI workflow to ensure that this actually works.

### DIFF
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -14,8 +14,14 @@ workspace = ".."
 all-features = true
 
 [features]
-default = ["rustls", "log"]
-rustls = ["dep:rustls", "ring"]
+default = ["rustls-ring", "log"]
+aws-lc-rs = ["dep:aws-lc-rs", "aws-lc-rs/aws-lc-sys", "aws-lc-rs/prebuilt-nasm"]
+# For backwards compatibility, `rustls` forwards to `rustls-ring`
+rustls = ["rustls-ring"]
+# Enable rustls with the `aws-lc-rs` crypto provider
+rustls-aws-lc-rs = ["dep:rustls", "rustls/aws-lc-rs", "aws-lc-rs"]
+# Enable rustls with the `ring` crypto provider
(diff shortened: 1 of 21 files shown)


quinn-proto/Cargo.toml
```diff
@@ -15,7 +15,10 @@ all-features = true
 
 [features]
 default = ["rustls", "log"]
-rustls = ["dep:rustls", "ring"]
+aws-lc-rs = ["dep:aws-lc-rs", "aws-lc-rs/aws-lc-sys"]
+rustls = ["rustls-ring"]
+rustls-aws-lc-rs = ["dep:rustls", "rustls/aws-lc-rs", "aws-lc-rs"]
+rustls-ring = ["dep:rustls", "rustls/ring", "ring"]
```
Ralith, line 21: Document?

quinn/Cargo.toml
Ralith, line 1: Document new features?

quinn-proto/src/cid_generator.rs
```diff
@@ -171,7 +171,6 @@ mod tests {
     use super::*;
 
     #[test]
-    #[cfg(feature = "ring")]
```
Ralith, line 174: Separate commit?

Berrysoft:
Would you provide an option to disable both aws-lc-rs and ring? I would like to try another third-party adapter.

djc:
> Would you provide an option to disable both aws-lc-rs and ring? I would like to try another third-party adapter.

This should already be the case both in released versions and once this PR is merged.

djc:
Addressed comments and added CI job to test this.

Ralith (approved):
aws-lc-rs is faster, right? Should we favor it over ring when both are enabled? There's an argument for preferring it by default, even, though that would reduce portability somewhat.

djc:
I think we should consider that for 0.12. For now, I don't think we want to switch the default crypto provider in a semver-compatible release, and I also think it might be surprising if the default is not the implementation that gets used if both providers are enabled (since other crates in the dependency graph can do that "behind your back").

PR #1963, opened 2024-08-10
Bump versions

Ralith:
I'd like to get #1961 published for interop reasons, and we might as well get the docs update out there while we're at it.


Ralith:
Actually, let's get https://github.com/quinn-rs/quinn/pull/1962 in first.

djc (approved):

PR #1950, opened 2024-08-01
ci: add Android job

mxinden:
Runs unit tests on Android emulator for both `x86` and `x86_64`.

---

Tracking issue: https://github.com/quinn-rs/quinn/issues/1778

Foundation to test https://github.com/quinn-rs/quinn/issues/1947.

PR #1964, opened 2024-08-10
fix(udp): use libc::recvmmsg on Android x86

### LINKED ISSUE #1947: `recvmmsg` is disallowed by seccomp on Android x86
#### What happened

We attempted switching Firefox Nightly to use `quinn-udp` for QUIC UDP I/O by default recently, though we had to roll back due to an error on Android x86. Investigating further, it seems like seccomp on Android x86 disallows `sendmsg` and `recvmmsg`. All details are in Bugzilla Bug [1910594](https://bugzilla.mozilla.org/show_bug.cgi?id=1910594) and [1910360](https://bugzilla.mozilla.org/show_bug.cgi?id=1910360#c5).

#### Reasoning

Looking at e.g. Android 13's [`SYSCALLS.TXT`](https://android.googlesource.com/platform/bionic/+/refs/tags/android-13.0.0_r83/libc/SYSCALLS.TXT) we can see support for `sendmsg` and `recvmmsg` on 64bit (`lp64`) and ARM:

```
# sockets
// [...]
ssize_t       __sendmsg:sendmsg(int, const struct msghdr*, unsigned int)  arm,lp64
// [...]
int           __recvmmsg:recvmmsg(int, struct mmsghdr*, unsigned int, int, const struct timespec*)   arm,lp64
```

While on x86, only the indirect calls through `socketcall` are allowed:

```
# sockets for x86. These are done as an "indexed" call to socketcall syscall.
// [...]
int           __sendmsg:socketcall:16(int, const struct msghdr*, unsigned int)  x86
// [...]
int           __recvmmsg:socketcall:19(int, struct mmsghdr*, unsigned int, int, const struct timespec*)   x86
```

#### Potential Solution

`libuv` has faced the same issue (see https://github.com/libuv/libuv/issues/2923). On x86 they use the indirect syscalls through `socketcall` (see https://github.com/libuv/libuv/pull/2925).

#### Historical context

The above might be due to historical reasons:

> On x86-32, socketcall() was historically the only entry point for
       the sockets API.  However, starting in Linux 4.3, direct system
       calls are provided on x86-32 for the sockets API.

https://man7.org/linux/man-pages/man2/socketcall.2.html

---

I still have to investigate a bit before proposing a fix. Opening this issue early to track progress.

mxinden:
> On x86-32, socketcall() was historically the only entry point for
> the sockets API.  However, starting in Linux 4.3, direct system
> calls are provided on x86-32 for the sockets API.

https://man7.org/linux/man-pages/man2/socketcall.2.html

Androids x86 even goes as far as automatically dispatching advanced socket calls (e.g. `socket` or `recvmmsg`) through `socketcall` in its Bionic libc:

> For example, socket() syscall on i386 actually becomes:
> socketcall(__NR_socket, 1, *(rest of args on stack)).

https://android.googlesource.com/platform/bionic/+/refs/tags/android-vts-14.0_r5/libc/SYSCALLS.TXT#19

In addition Android enables seccomp filtering since Android 8:

https://android-developers.googleblog.com/2017/07/seccomp-filter-in-android-o.html

Currently `quinn-udp` calls `recvmmsg` directly on Android x86:

https://github.com/quinn-rs/quinn/blob/c0b1e281e3167d4f4c8496082cb1117c4a270ad8/quinn-udp/src/unix.rs#L447-L448

A direct `recvmmsg` through `libc::syscall(libc::SYS_recvmmsg` does not trigger the automatic Android x86 Bionic libc dispatch logic through `socketcall`, but instead calls the unimplemented `recvmmsg` syscall. Instead of triggering a `libc::ENOSYS`, seccomp disallows the call, thus leading to a panic of the application.

This commit changes `quinn-udp` to use `libc::recvmmsg` on Android x86, thus leveraging Bionic's automatic dispatch of `recvmmsg` through `socketcall`.

Note that this commit only uses `libc::recvmmsg` on Android x86, not any other Android or Linux variant. Thus quinn-udp would still support missing `libc::recvmmsg` on those systems. See https://github.com/quinn-rs/quinn/pull/1504.

---

Fixes #1947.

Problem itself and fix reproducible on Firefox CI. See e.g. [latest CI run](https://treeherder.mozilla.org/jobs?repo=try&revision=fe7f8ef18948611619d9a67bacb4fff7f7f66e87) `startup-x86`.

Still unable to reproduce via https://github.com/quinn-rs/quinn/pull/1950. I assume this is due to still using a 64bit toolchain when testing `x86`.

//CC @link2xt following up on https://github.com/quinn-rs/quinn/pull/1504, this might be of interest for you.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -447,7 +447,11 @@ unsafe fn recvmmsg_with_fallback(
     let flags = 0;
     let timeout = ptr::null_mut::<libc::timespec>();
 
-    #[cfg(not(any(target_os = "freebsd", target_os = "netbsd")))]
+    #[cfg(not(any(
+        target_os = "freebsd",
+        target_os = "netbsd",
+        all(target_os = "android", target_arch = "x86")
+    )))]
     {
         let ret =
             libc::syscall(libc::SYS_recvmmsg, sockfd, msgvec, vlen, flags, timeout) as libc::c_int;
@@ -462,6 +466,19 @@ unsafe fn recvmmsg_with_fallback(
     {
         #[cfg(target_os = "freebsd")]
         let vlen = vlen as usize;
+        let ret = libc::recvmmsg(sockfd, msgvec, vlen, flags, timeout) as libc::c_int;
+        if ret != -1 {
+            return ret;
+        }
+    }
+
+    // On Android x86 seccomp only allows `recvmmsg` dispatched through
+    // `socketcall`. The bionic libc implementation automatically dispatches
+    // `recvmmsg` through `socketcall` when calling `libc::recvmmsg`. Calling
+    // `SYS-recvmmsg` directly (i.e. `syscall(SYS_recvmmsg)`) does not dispatch
+    // through `socketcall` and is thus disallowed by seccomp.
+    #[cfg(all(target_os = "android", target_arch = "x86"))]
+    {
         let ret = libc::recvmmsg(sockfd, msgvec, vlen, flags, timeout) as libc::c_int;
         if ret != -1 {
             return ret;


djc (approved):

djc:
Thanks! If you'd like a release with this, please add a commit that bumps the version number

quinn-udp/src/unix.rs
```diff
@@ -468,6 +472,19 @@ unsafe fn recvmmsg_with_fallback(
...
     }
 
+    // On Android x86 seccomp only allows `recvmmsg` dispatched through
+    // `socketcall`. The bionic libc implementation automatically dispatches
+    // `recvmmsg` through `socketcall` when calling `libc::recvmmsg`. Calling
+    // `SYS-recvmmsg` directly (i.e. `syscall(SYS_recvmmsg)`) does not dispatch
+    // through `socketcall` and is thus disallowed by seccomp.
+    #[cfg(all(target_os = "android", target_arch = "x86"))]
+    {
+        let ret = libc::recvmmsg(sockfd, msgvec, vlen, flags, timeout) as libc::c_int;
```
Ralith, line 482: This looks identical to the *BSD call above. Should we fold them together?

mxinden:
Alternative implementation in https://github.com/quinn-rs/quinn/pull/1966. Let me know which way forward you prefer.

mxinden:
Closing here since https://github.com/quinn-rs/quinn/pull/1966 is merged and released. Thanks for taking a look!

Issue #1503, opened 2023-03-03
`libc::sendmmsg` is not available on old Android phones


link2xt:
`sendmmsg` was added in Linux 3.0 and `recvmmsg` was added in Linux 2.6.33 according to their man pages.

It is used in
https://github.com/quinn-rs/quinn/blob/389d7ce5067fbef460f0730861340b15f13d4803/quinn-udp/src/unix.rs#L197

djc:
Ugh, how old are these Android phones that you are targeting?

If it's not too much complexity, we're happy to take a PR to enable support on older kernels.

Ralith:
Fixed by #1504.

mxinden:
I wonder whether the patch merged via https://github.com/quinn-rs/quinn/pull/1504 is worth the trouble it is causing. In other words, I would like to discuss switching back to calling `libc::recvmmsg` instead of `libc::syscall(libc::SYS_recvmmsg`.

- On Android x86 calling `libc::syscall(libc::SYS_recvmmsg` instead of `libc::recvmmsg` leads to a SECCOMP panic. See https://github.com/quinn-rs/quinn/pull/1964.
- At this point Android API Level 16 is [12 years old](https://apilevels.com/). Are there any known use-cases running `quinn` on such old devices? With the security concerns alone, I don't see a use-case outside a completely isolated environment.
- Android 4 (up to API Level 20) has been end-of-life'd 01 Oct 2017. Should `quinn` support an OS that has been unsupported for 7 years?

//CC @link2xt, @flub and @dignifiedquire (since you have been active on https://github.com/deltachat/deltachat-core-rust/pull/4007)

PR #1966, opened 2024-08-14
Revert "Implement fallback for `sendmmsg` and `recvmmsg`"

mxinden:
This reverts commit e6f184489e703bcf70850f99373c88746c250662.

e6f18448 called `sendmmsg` and `recvmmsg` through `libc::syscall` instead of `libc::sendmmsg` and `libc::recvmmsg`, thus preventing linking issues on old Android systems where `libc::sendmmsg` and `libc::recvmmsg` isn't available.

In https://github.com/quinn-rs/quinn/issues/1503#issuecomment-2285962727 the suggestion was made to no longer support these old Android systems (API level 16).

This commit reverts e6f18448. Given that `sendmmsg` support was previously dropped in ee088265, only the `recvmmsg` calls are reverted.

---

Alternative to https://github.com/quinn-rs/quinn/pull/1964. Given the additional simplicity, my preference is for this change.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -395,10 +395,12 @@ fn recv(io: SockRef<'_>, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) -> 
     }
     let msg_count = loop {
         let n = unsafe {
-            recvmmsg_with_fallback(
+            libc::recvmmsg(
                 io.as_raw_fd(),
                 hdrs.as_mut_ptr(),
                 bufs.len().min(BATCH_SIZE) as _,
+                0,
+                ptr::null_mut::<libc::timespec>(),
             )
         };
         if n == -1 {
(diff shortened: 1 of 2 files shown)


quinn-udp/src/unix.rs
```diff
@@ -442,25 +442,9 @@ unsafe fn recvmmsg_with_fallback(
...
-    // `recvmsg`, thus `SYS_recvmmsg` constant and direct system call do not exist
-    #[cfg(any(target_os = "freebsd", target_os = "netbsd"))]
-    {
-        #[cfg(target_os = "freebsd")]
-        let vlen = vlen as usize;
-        let ret = libc::recvmmsg(sockfd, msgvec, vlen, flags, timeout) as libc::c_int;
-        if ret != -1 {
-            return ret;
-        }
+    let ret = libc::recvmmsg(sockfd, msgvec, vlen as _, flags, timeout) as libc::c_int;
```
mxinden, line 445: Note the `as _` here. The alternative would be:

``` diff
modified   quinn-udp/src/unix.rs
@@ -442,7 +442,11 @@ unsafe fn recvmmsg_with_fallback(
     let flags = 0;
     let timeout = ptr::null_mut::<libc::timespec>();
 
-    let ret = libc::recvmmsg(sockfd, msgvec, vlen as _, flags, timeout) as libc::c_int;
+    let ret = {
+        #[cfg(target_os = "freebsd")]
+        let vlen = vlen as usize;
+        libc::recvmmsg(sockfd, msgvec, vlen, flags, timeout) as libc::c_int
+    };
     if ret != -1 {
         return ret;
     }
```

The additional scope is need in order to preserve the original `vlen` type (`c_int`) (i.e. not shadow `vlen`) for the `recvmmsg_fallback` below.

mxinden: Or we don't pass `vlen` to `recvmmsg_fallback` at all:

``` diff
modified   quinn-udp/src/unix.rs
@@ -181,6 +181,10 @@ impl UdpSocketState {
         bufs: &mut [IoSliceMut<'_>],
         meta: &mut [RecvMeta],
     ) -> io::Result<usize> {
+        if bufs.len() == 0 {
+            return Ok(0);
+        }
+
         recv(socket.0, bufs, meta)
     }
 
@@ -442,7 +446,9 @@ unsafe fn recvmmsg_with_fallback(
     let flags = 0;
     let timeout = ptr::null_mut::<libc::timespec>();
 
-    let ret = libc::recvmmsg(sockfd, msgvec, vlen as _, flags, timeout) as libc::c_int;
+    #[cfg(target_os = "freebsd")]
+    let vlen = vlen as usize;
+    let ret = libc::recvmmsg(sockfd, msgvec, vlen, flags, timeout) as libc::c_int;
     if ret != -1 {
         return ret;
     }
@@ -451,7 +457,7 @@ unsafe fn recvmmsg_with_fallback(
     match e.raw_os_error() {
         Some(libc::ENOSYS) => {
             // Fallback to `recvmsg`.
-            recvmmsg_fallback(sockfd, msgvec, vlen)
+            recvmmsg_fallback(sockfd, msgvec)
         }
         _ => -1,
     }
@@ -461,15 +467,8 @@ unsafe fn recvmmsg_with_fallback(
 /// for systems which do not support `recvmmsg`
 /// such as Linux <2.6.33.
 #[cfg(not(any(target_os = "macos", target_os = "ios", target_os = "openbsd")))]
-unsafe fn recvmmsg_fallback(
-    sockfd: libc::c_int,
-    msgvec: *mut libc::mmsghdr,
-    vlen: libc::c_uint,
-) -> libc::c_int {
+unsafe fn recvmmsg_fallback(sockfd: libc::c_int, msgvec: *mut libc::mmsghdr) -> libc::c_int {
     let flags = 0;
-    if vlen == 0 {
-        return 0;
-    }
 
     let n = libc::recvmsg(sockfd, &mut (*msgvec).msg_hdr, flags);
     if n == -1 {
```

Your call. I am fine either way.

mxinden:
@djc @Ralith do you have some time to give this pull request a review?

djc (approved):
This makes sense to me, but I'll want to wait for @Ralith's opinion.

mxinden:
Friendly ping @Ralith. Do you have some time to take a look?

Ralith (changes requested):
Do we still need the `ENOSYS` path and the `recvmmsg_fallback` definition? Prior to e6f184489e703bcf70850f99373c88746c250662 we just called `libc::recvmmgs` directly rather than wrapping it at all -- can we return to that?

mxinden:
Thank you for the review!

I am fine with no longer supporting both:

1. Android API Level 16 (i.e. Android 4)
2. Linux Kernel < 2.6.33

(1) allows us to drop indirection `libc::syscall(libc::SYS_sendmmsg` in favor of a direct call to `libc::recvmmsg`. (2) allows us to drop `recvmsg` (one `m`) fallback.

I updated this pull request accordingly.

@Ralith would you mind taking another look?

Ralith (approved):
LGTM! I don't think Rust officially supports targeting linux 2.x anyway.

djc:
* Published quinn-udp v0.5.5 at registry `crates-io`
* [new tag]           quinn-udp-0.5.5 -> quinn-udp-0.5.5
* [Release notes](https://github.com/quinn-rs/quinn/releases/tag/quinn-udp-0.5.5)

PR #1973, opened 2024-08-22
fix(.github/android): pass matrix.target and increase api to v26 

mxinden:
When extracting the emulator script into a separate file, the `matrix.target` substitution wasn't replaced with an environment variable.

Follow-up to https://github.com/quinn-rs/quinn/pull/1950.

Also the `ecn_*` unit tests are failing with any Android API version < v26 with:

```
2024-08-22T14:41:30.636494Z  WARN quinn_udp: sendmsg error: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }, Transmit: { destination: 127.0.0.1:50196, src_ip: None, enc: Some(Ect0), len: 5, segment_size: None }
```

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -144,6 +144,8 @@ jobs:
         sudo udevadm trigger --name-match=kvm
 
     - name: Set up Android Emulator and run tests
+      env:
+        TARGET: ${{ matrix.target }}
       uses: reactivecircus/android-emulator-runner@v2
       with:
         api-level: ${{ matrix.api-level }}
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -102,10 +102,10 @@ jobs:
             emulator-arch: x86_64
             # Note that x86_64 image is only available for API 21+. See
             # https://github.com/ReactiveCircus/android-emulator-runner?tab=readme-ov-file#configurations.
-            api-level: 21
+            api-level: 26
           - target: i686-linux-android
             emulator-arch: x86
-            api-level: 19
+            api-level: 26
 
     steps:
     - name: Checkout code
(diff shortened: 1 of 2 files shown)


djc (approved):

.github/workflows/rust.yml
```diff
@@ -102,10 +102,10 @@ jobs:
             emulator-arch: x86_64
             # Note that x86_64 image is only available for API 21+. See
             # https://github.com/ReactiveCircus/android-emulator-runner?tab=readme-ov-file#configurations.
-            api-level: 21
+            api-level: 26
```
djc, line 105: I think this is conceptually unrelated and thus merits a separate commit.

.github/workflows/rust.yml
```diff
@@ -102,10 +102,10 @@ jobs:
...
           - target: i686-linux-android
             emulator-arch: x86
-            api-level: 19
+            api-level: 26
```
mxinden, line 108: The `ecn_*` unit tests are failing with any Android API version < v26 with:

```
2024-08-22T14:41:30.636494Z  WARN quinn_udp: sendmsg error: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }, Transmit: { destination: 127.0.0.1:50196, src_ip: None, enc: Some(Ect0), len: 5, segment_size: None }
```

I will open a separate pull request discussing this failure.

I suggest we merge here already. Some Android tests (i.e. >=26) are better than none.

.github/workflows/rust.yml
mxinden: :+1: sorry for not doing that earlier. I updated the pull request @djc.

.github/workflows/rust.yml
mxinden: > I will open a separate pull request discussing this failure.

For the record, the separate pull request: https://github.com/quinn-rs/quinn/pull/1975

djc (approved):

PR #1977, opened 2024-09-02
proto: avoid panicking on rustls server config errors

djc:
As discussed on Discord.


Ralith (approved):
Nice, I worried this might break something.

PR #1979, opened 2024-09-02
Version bump `quinn` to enforce patched `quinn-proto`

finnbear:
Now crates like `wtransport`, that only depend on `quinn` and use `quinn-proto` via `quinn::proto`, can point to a [patched](https://github.com/quinn-rs/quinn/security/advisories/GHSA-vr26-jcq5-fjj8) version.

### DIFF
--- a/quinn/Cargo.toml
+++ b/quinn/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quinn"
-version = "0.11.4"
+version = "0.11.5"
 license.workspace = true
 repository.workspace = true
 description = "Versatile QUIC transport protocol implementation"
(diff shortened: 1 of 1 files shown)


Ralith (approved):
`cargo update` (or the default behavior when adding a new dependency) will be sufficient to get the latest `quinn-proto` regardless, so I'm not sure this is strictly needed. Still, no harm, and I wouldn't mind getting https://github.com/quinn-rs/quinn/pull/1977 published.

finnbear:
> cargo update

FWIW, once your `nightly` app code base has hundreds of dependencies, `cargo update` (or removing `Cargo.lock`) becomes hazardous (e.g. introduces incompatibilities due to SEMVER mistakes and changing `nightly` features) and it's safer to depend on exact versions of dependencies and manually update the important ones occasionally (but you can only do that with direct dependencies, not transitives ones).

Ralith:
`cargo update` can also target individual transitive dependencies if you don't want to do everything at once.

finnbear:
> cargo update can also target individual transitive dependencies if you don't want to do everything at once.

You're right, but if you use your library in 5 different places (say you use your game engine in 5 different games), then you would have to remember to run `cargo update` in those 5 places.

Maybe this is really niche :joy:

djc (approved):

djc:
* Published quinn-proto v0.11.8 at registry `crates-io`
* Published quinn v0.11.5 at registry `crates-io`
* [new tag]           quinn-0.11.5 -> quinn-0.11.5
* [new tag]           quinn-proto-0.11.8 -> quinn-proto-0.11.8
* [Release notes](https://github.com/quinn-rs/quinn/releases/tag/quinn-proto-0.11.8)

PR #1980, opened 2024-09-03
use orig_dst_cid to add into index to avoid corruption

lijunwangs:
Problem:

When the initial packet is a stateless retry packet, we use the orig_dst_cid to create the quinn_proto::Incoming. However we added the connection into the index using the header.dst_cid. These two connection CID can be different. When the connection is later removed from the index, we use the one populated in the Incoming. This will fail to properly remove it. This causes discrepancy of the Quinn_proto::Endpoint::index and Quinn_proto::Endpoint::incoming_buffers and cause crashes when the incoming buffer is accessed.  A scenario

1. A stateless connect retry initial packet is received, we add (header.dst_cid, index) into the connection index and incoming buffers.
2. The connection is ignored by the application which triggers clean_up_incoming and cleanup, This removes the corresponding index from the incoming_buffers but not the Connection from the index (which was added using header.dst_cid).
3. Client sends the connection retry packet again with the same header.dst_cid, in this case we found the connection from Endpoint::index, and obtained RouteDatagramTo::Incoming, with the index value in it, when accessing 
let incoming_buffer = &mut self.incoming_buffers[incoming_idx]; 
it will blows up due to wrong index as it was removed in step 2.

Changes:
in Quinn_proto::Endpoint::handle_first_packet use the orig_dst_cid variable to the index to keep consistent with the Incoming created.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -531,7 +531,7 @@ impl Endpoint {
 
         let incoming_idx = self.incoming_buffers.insert(IncomingBuffer::default());
         self.index
-            .insert_initial_incoming(header.dst_cid, incoming_idx);
+            .insert_initial_incoming(orig_dst_cid, incoming_idx);
 
         Some(DatagramEvent::NewConnection(Incoming {
             addresses,


lijunwangs:
This seems to have been addressed by e01609ccd8738bd438d86fa7185a0f85598cb58f.

Ralith:
Thanks for the diagnosis, all the same!

PR #1982, opened 2024-09-06
Apply suggestions from clippy 1.81

djc:


quinn-udp/src/cmsg/mod.rs
```diff
@@ -42,7 +42,7 @@ impl<'a, M: MsgHdr> Encoder<'a, M> {
     /// # Panics
     /// - If insufficient buffer space remains.
     /// - If `T` has stricter alignment requirements than `M::ControlMessage`
-    pub(crate) fn push<T: Copy + ?Sized>(&mut self, level: c_int, ty: c_int, value: T) {
+    pub(crate) fn push<T: Copy>(&mut self, level: c_int, ty: c_int, value: T) {
```
djc, line 45: `Copy` requires `Sized`, so a `?Sized` bound doesn't make sense here.

mxinden (approved):

Ralith (approved):

quinn-udp/src/cmsg/mod.rs
Ralith: Technically this doesn't need `Copy` and `?Sized` makes it more permissive, but we don't actually use `?Sized` arguments, and `Copy` is a nice guardrail against unintended use, so this seems reasonable.

PR #1983, opened 2024-09-07
remove panic-on-drop in `quinn_proto::Chunks`

### LINKED ISSUE #1531: quinn-proto: `Chunks` panic-on-drop behavior is a footgun and makes control flow annoying
I was just bitten by this assert while trying to debug a silent failure of my application during testing: https://github.com/quinn-rs/quinn/blob/main/quinn-proto/src/connection/streams/recv.rs#L289-L292

This was in a spawned Tokio task ~which apparently suppresses the normal printing of panic backtraces to stderr~ and I was running this as a test which buffered the panic backtrace but not letting it complete. And I hadn't yet wired it up to check the task's exit status as it's supposed to be long-running anyway and I'm still working on the principal implementation here.

On `StreamEvent::Opened` or `StreamEvent::Readable` I essentially call into code that looks like this: 

```rust
impl<'a> StreamReadContext<'a> {
    fn read(mut self) -> anyhow::Result<()> {
        // `self.stream: quinn_proto::RecvStream`
        let mut chunks = self.stream.read(true)?;

        loop {
            match self.activity {
                StreamActivity::PendingHeader => {
                    if let Some(header) = self.reader.try_peek_header().context("invalid header")? {
                        Notification::validate_header(header).context("invalid header")?;
                        *self.activity = StreamActivity::ReadingNotification { header };
                    };
                }
                StreamActivity::ReadingNotification { header } => {
                    if self.reader.input_len() >= header.length {
                        let notification = Notification::decode(&mut *self.reader)
                            .context("invalid notification")?;

                        // Handle `Notification`
                    }
                }
            }

            // Lazily read chunks so we don't buffer a ton of data up-front.
            if let Some(chunk) = chunks.next(MAX_CHUNK_SIZE)? { // 8 KiB
                tracing::trace!("read {} bytes", chunk.bytes.len());
                self.reader.push(chunk.bytes)?;
            } else {
                break;
            }
        }

        // If we need to transmit it'll be checked when we return to the event loop.
        let _ = chunks.finalize();

        Ok(())
    }
}
```

As implied by the code here, `Notification` messages have a header that gives their length so we know how much data we need to buffer to fully decode them. `reader` is the type that buffers chunks and implements the right traits for `Notification::decode()`.

Due to some bug in my code, `.try_peek_header()` is returning an error which normally _would_ be logged when returned from this method. However, because the `impl Drop for Chunks` executes on the return, and because I hadn't called `.finalize()` yet, this is turned into a `panic!()`, which in the current state of my code just causes the task to silently exit.

The thing is, I don't _need_ the `ShouldTransmit` flag because this immediately returns to an event loop where I call all the appropriate `.handle*()` and `.poll*()` methods on the `quinn_proto::Connection`. So the implicit finalize triggered by the `Drop` impl would otherwise be sufficient for me.

Because the panic message was being suppressed, it took me an annoyingly long time to figure this out. It wasn't until I ran the code in a debugger (which helpfully automatically breaks on panics) to figure this out. While I understand the point of the `debug_assert!()` is to teach the user how to use this API correctly, in my case it was much more of a hindrance than anything.

(~Tokio's partly to blame here... why suppress backtraces in debug mode? That doesn't make any sense.~ However, the panic doesn't strictly _need_ to happen either. `quinn-proto` is by design a low-level API, so why is it assuming the user might not know what they're doing here?)

abonander:
closes #1703
closes #1531 

cc @djc

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -82,7 +82,7 @@ pub use streams::StreamsState;
 use streams::StreamsState;
 pub use streams::{
     BytesSource, Chunks, ClosedStream, FinishError, ReadError, ReadableError, RecvStream,
-    SendStream, StreamEvent, Streams, WriteError, Written,
+    SendStream, ShouldTransmit, StreamEvent, Streams, WriteError, Written,
 };
 
 mod timer;
--- a/quinn-proto/src/connection/streams/recv.rs
+++ b/quinn-proto/src/connection/streams/recv.rs
@@ -220,7 +220,17 @@ impl Recv {
     }
 }
 
-/// Chunks
+/// Chunks returned from [`RecvStream::read()`][crate::RecvStream::read].
+///
+/// ### Note: Finalization Needed
+/// Bytes read from the stream are not released from the congestion window until
+/// either [`Self::finalize()`] is called, or this type is dropped.
+///
+/// It is recommended that you call [`Self::finalize()`] because it returns a flag
+/// telling you whether reading from the stream has resulted in the need to transmit a packet.
+///
+/// If this type is leaked, the stream will remain blocked on the remote peer until
+/// another read from the stream is done.
 pub struct Chunks<'a> {
     id: StreamId,
     ordered: bool,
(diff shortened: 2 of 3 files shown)


abonander:
I might also recommend making the `bool` field of `ShouldTransmit` public to allow destructuring.

djc:
> I might also recommend making the `bool` field of `ShouldTransmit` public to allow destructuring.

Do you want to add a commit for this?

abonander:
Sure, I just assumed there was a reason it wasn't already public.

abonander:
Strictly speaking, it is a potential SemVer hazard if you foresee other fields being added later, e.g. `total_bytes_read` or something like that.

I don't really care either way. My implementation always checks `poll_transmit()` if there isn't one pending already.

djc:
Then why would you recommend making the `ShouldTransmit` field public?

abonander:
I did say "I _might_ recommend", because some people _might_ find it useful.

If `ShouldTransmit` was only ever intended to be a wrapper for a `bool`, making the field public makes sense from an API design standpoint.

I was just saying I don't care to die on this hill either way because I don't even use the type. 🤷

djc (approved):

Ralith (approved):

Issue #1988, opened 2024-09-11
ACK Receive Timestsamps


sterlingdeng:
Hi quinn maintainers, this issue contains a proposal and details to implement packet receive timestamps[^1].

## Motivation

We would like to implement the GCC[^2] congestion control algorithm, commonly used for WebRTC, as a controller implementation in quinn. The GCC algorithm is both loss and delay based and require measurements in order to provide an estimated bitrate for the network link. ACK's currently provide enough information to the loss based controller because a loss can be measured by any skips in packet numbers, but the current design of ACK's lacks adequete information for the loss based controller. In the WebRTC domain, GCC is commonly used with TWCC[^3] an RTP/RTCP extension that provides feedback on the arrival time and sequence number of RTP packets. The Receive Timestamp draft[^1] is a QUIC extension that's analogous to TWCC.

The draft also contains a [motivation](https://www.ietf.org/archive/id/draft-smith-quic-receive-ts-00.html#name-motivation) section. Meta also seems to have their own implementation thats based off of this draft[^4] and presented[^5] their work to the IETF 118 meeting in Nov of 2023.

## Implementation

This section contains the components and their implementation details needed to implement this feature. It would be useful to read through the proposal first to familiarize oneself with how it works.

### Update TransportParmeters

| TransportParameter               | Description                                                                            | Code |
| -------------------------------- | -------------------------------------------------------------------------------------- | ---- |
| `max_receive_timestamps_per_ack` | Limits the number of timestamps per ACK frame                                          | tbd  |
| `receive_timestamp_exponent`     | Trades off precision of time for space. Exponent of 0 indicates microsecond precision. | tbd  |
| `receive_timestamp_basis`        | The basis for which all provided timestamps deltas are anchored on.                    | tbd  |

**Question**

The transport parameter values are not determined. How should we handle this? An option is to use a sufficiently large value that we believe there shouldn’t be a collision in a reasonable time if the official values are incremented sequentially.

### Update AckFrame

Update the AckFrame to include the timestamp information.

```rust
pub struct Ack {
    pub largest: u64,
    pub delay: u64,
    pub additional: Bytes,
    pub ecn: Option<EcnCounts>,
   +pub timestamps: Option<Bytes>,
}
```

[...]

PR #1992, opened 2024-09-19
feat: ack receiver timestamps

sterlingdeng:
This PR is the implementation that addresses https://github.com/quinn-rs/quinn/issues/1988

Manual E2E Testing/Validation instructions: https://github.com/sterlingdeng/quinn/pull/2


djc:
I don't think a feature makes sense for this, what's the motivation for that? Suggest squashing all of these commits because I don't think they make much sense on their own.

billylindeman:
@djc I suggested using a feature bc it is a draft specification, perhaps we should name the feature after the specification ```draft-smith-quic-receive-ts```  https://www.ietf.org/archive/id/draft-smith-quic-receive-ts-00.html

djc:
> @djc I suggested using a feature bc it is a draft specification, perhaps we should name the feature after the specification `draft-smith-quic-receive-ts` https://www.ietf.org/archive/id/draft-smith-quic-receive-ts-00.html

I don't think that's a good reason to make it a feature. The current API surface will still become part of our public API, and it just adds a bunch of complexity. We've added draft functionality before, should be fine.

sterlingdeng:
I can remove the feature flag in a follow up commit. Regarding squashing, my intention for the commits were to keep separate different parts of the PR for review and then squash at the end. If you want me to squash it now, I can do that too.

djc:
> I can remove the feature flag in a follow up commit. Regarding squashing, my intention for the commits were to keep separate different parts of the PR for review and then squash at the end. If you want me to squash it now, I can do that too.

While we are fans of clean commit histories with atomic commits, I think this falls short of that because the commits are not conceptually independent -- it doesn't make sense to add configuration that doesn't do anything, for example.

quinn-proto/src/config.rs
```diff
@@ -416,10 +430,51 @@ impl fmt::Debug for TransportConfig {
...
+    /// Timestamp values are divided by the exponent value provided. This reduces the size of the
+    /// VARINT for loss in precision. A exponent of 0 represents microsecond precision.
+    pub fn exponent(&mut self, value: VarInt) -> &mut Self {
+        self.exponent = value;
+        self
+    }
+
+    /// Sets the time base for which all timestamps are anchored on.
+    /// Defaults to Instant::now of when the default struct was created.
+    pub fn basis(&mut self, instant: std::time::Instant) -> &mut Self {
```
djc, line 462: Nit: please import `Instant` at the top of the module.

quinn-proto/src/connection/mod.rs
```diff
@@ -37,6 +36,8 @@ use crate::{
     VarInt, MAX_STREAM_COUNT, MIN_INITIAL_SIZE, TIMER_GRANULARITY,
 };
 
+use crate::config::AckTimestampsConfig;
```
djc, line 39: Nit: this should go in the `crate` import block, above.

quinn-proto/src/connection/mod.rs
```diff
@@ -227,6 +231,11 @@ pub struct Connection {
     /// no outgoing application data.
     app_limited: bool,
 
+    //
+    // Ack Receive Timestamps
+    //
+    peer_ack_timestamp_cfg: Option<AckTimestampsConfig>,
```
djc, line 237: Nit: would prefer to just call this `ack_timestamps_config`.

quinn-proto/src/connection/mod.rs
```diff
@@ -1343,6 +1358,13 @@ impl Connection {
         if ack.largest >= self.spaces[space].next_packet_number {
             return Err(TransportError::PROTOCOL_VIOLATION("unsent packet acked"));
         }
+
+        if ack.timestamps.is_some() != self.config.ack_timestamp_config.is_some() {
+            return Err(TransportError::PROTOCOL_VIOLATION(
+                "ack with timestamps expectation mismatched",
+            ));
+        }
```
djc, line 1366: Is this in the spec?

quinn-proto/src/connection/mod.rs
```diff
@@ -1371,13 +1393,21 @@ impl Connection {
             }
         }
 
+        let mut timestamp_iter = self.config.ack_timestamp_config.as_ref().map(|cfg| {
+            let decoder = ack.timestamp_iter(cfg.basis, cfg.exponent.0).unwrap();
```
djc, line 1397: Should at least have a comment about why this `unwrap()` is safe.

quinn-proto/src/connection/mod.rs
```diff
@@ -1371,13 +1393,21 @@ impl Connection {
...
+            let mut v: tinyvec::TinyVec<[PacketTimestamp; 10]> = tinyvec::TinyVec::new();
```
djc, line 1398: Nit: import `TinyVec` directly, and avoid type annotations. The code here is looking pretty unidiomatic... Can use `rev()` to reverse the direction on the source iterator directly which should be faster, and `collect()` into the `TinyVec` (are you sure we need to allocate here instead of yielding the iterator directly)?

quinn-proto/src/connection/mod.rs
```diff
@@ -1399,6 +1429,25 @@ impl Connection {
                 // Notify ack frequency that a packet was acked, because it might contain an ACK_FREQUENCY frame
                 self.ack_frequency.on_acked(packet);
 
+                if let Some(timestamp_iter) = timestamp_iter.as_mut() {
+                    while let Some(v) = timestamp_iter.peek() {
```
djc, line 1433: Why do we need/want to `peek()` here?

quinn-proto/src/connection/mod.rs
```diff
@@ -3231,6 +3291,7 @@ impl Connection {
         space: &mut PacketSpace,
         buf: &mut Vec<u8>,
         stats: &mut ConnectionStats,
+        peer_timestamp_config: Option<AckTimestampsConfig>,
```
djc, line 3294: Nit: just call this `ack_timestamps_config`?

quinn-proto/src/connection/mod.rs
```diff
@@ -3309,6 +3385,26 @@ impl Connection {
...
+                }
+                Some(AckTimestampsConfig {
+                    exponent,
+                    max_timestamps_per_ack,
+                    basis: self.epoch,
+                })
+            } else {
+                None
+            };
+        }
```
djc, line 3407: Suggest spelling this as:

`let (Some(max_timestamps_per_ack), Some(exponent)) = (params.max_recv_timestamps_per_ack, params.receive_timestamps_exponent) else { return None };`

quinn-proto/src/connection/receiver_timestamps.rs
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+}
+
+pub struct ReceiverTimestamps {
+    data: VecDeque<PacketTimestamp>,
+    max: usize,
+}
+
+impl std::fmt::Debug for ReceiverTimestamps {
```
djc, line 26: Nit: please import `fmt`.

quinn-proto/src/connection/receiver_timestamps.rs
```diff
@@ -0,0 +1,131 @@
+use std::collections::VecDeque;
+use std::time::Instant;
+
+use tracing::warn;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
```
djc, line 6: Suggest deriving `Copy` here.

quinn-proto/src/connection/receiver_timestamps.rs
```diff
@@ -0,0 +1,131 @@
...
+        self.data.iter().cloned()
+    }
+
+    pub(crate) fn len(&self) -> usize {
+        self.data.len()
+    }
+
+    pub(crate) fn inner(&self) -> &VecDeque<PacketTimestamp> {
+        &self.data
+    }
```
djc, line 80: Where is this used?

quinn-proto/src/connection/receiver_timestamps.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    pub(crate) fn subtract_below(&mut self, packet_number: u64) {
```
djc, line 82: Nit: move this up above `iter()`, please.

quinn-proto/src/connection/spaces.rs
```diff
@@ -16,6 +16,8 @@ use crate::{
     shared::IssuedCid, Dir, StreamId, TransportError, VarInt,
 };
 
+use crate::connection::receiver_timestamps::ReceiverTimestamps;
```
djc, line 19: Nit: move this into the `crate` import block.

quinn-proto/src/frame.rs
```diff
@@ -403,14 +421,111 @@ impl Ack {
...
+        self.timestamps
+            .as_ref()
+            .map(|v| AckTimestampIter::new(self.largest, basis, exponent, &v[..]))
+    }
+
+    // https://www.ietf.org/archive/id/draft-smith-quic-receive-ts-00.html#ts-ranges
+    fn encode_timestamps<W: BufMut>(
+        timestamps: &ReceiverTimestamps,
+        mut largest: u64,
+        buf: &mut W,
```
djc, line 445: Nit: spell this as `&mut impl BufMut`, please.

quinn-proto/src/frame.rs
```diff
@@ -403,14 +421,111 @@ impl Ack {
...
+        mut basis: Instant,
+        exponent: u64,
+        max_timestamps: u64,
+    ) {
+        if timestamps.len() == 0 {
+            buf.write_var(0);
+            return;
+        }
+        let mut prev: Option<u64> = None;
```
djc, line 454: Nit: is this type annotation necessary?

quinn-proto/src/frame.rs
```diff
@@ -767,12 +891,45 @@ fn scan_ack_blocks(buf: &mut io::Cursor<Bytes>, largest: u64, n: usize) -> Resul
...
+}
+
+#[derive(PartialEq, Eq)]
 enum IterErr {
     UnexpectedEnd,
     InvalidFrameId,
     Malformed,
 }
 
+impl std::fmt::Debug for IterErr {
```
djc, line 927: `Nit: import `fmt`.

quinn-proto/src/frame.rs
```diff
@@ -790,14 +947,81 @@ impl From<UnexpectedEnd> for IterErr {
...
+    deltas_remaining: usize,
+    first: bool,
+    next_pn: u64,
+}
+
+impl<'a> AckTimestampIter<'a> {
+    fn new(largest: u64, basis_instant: Instant, exponent: u64, mut data: &'a [u8]) -> Self {
+        // We read and throw away the Timestamp Range Count value because
+        // it was already used to properly slice the data.
+        let _ = data.get_var().unwrap();
```
djc, line 965: Is this `unwrap()` safe?

quinn-proto/src/frame.rs
```diff
@@ -790,14 +947,81 @@ impl From<UnexpectedEnd> for IterErr {
...
+        AckTimestampIter {
+            timestamp_basis: 0,
+            timestamp_exponent: exponent,
+            timestamp_instant_basis: basis_instant,
```
djc, line 969: Suggest giving the function arguments the target names in this type.

quinn-proto/src/frame.rs
```diff
@@ -790,14 +947,81 @@ impl From<UnexpectedEnd> for IterErr {
...
+            next_pn: largest,
+        }
+    }
+}
+
+impl<'a> Iterator for AckTimestampIter<'a> {
+    type Item = PacketTimestamp;
+    fn next(&mut self) -> Option<Self::Item> {
+        if !self.data.has_remaining() {
+            debug_assert!(
```
djc, line 982: Nit: use `debug_assert_eq!()`.

quinn-proto/src/frame.rs
```diff
@@ -790,14 +947,81 @@ impl From<UnexpectedEnd> for IterErr {
...
+            return None;
+        }
+        if self.deltas_remaining == 0 {
+            let gap = self.data.get_var().unwrap();
+            self.deltas_remaining = self.data.get_var().unwrap() as usize;
+            if self.first {
+                self.next_pn -= gap;
+            } else {
+                self.next_pn -= gap + 2;
+            }
```
djc, line 995: Nit: suggest writing this as `self.next_pn -= match self.first { .. }`.

quinn-proto/src/frame.rs
```diff
@@ -790,14 +947,81 @@ impl From<UnexpectedEnd> for IterErr {
...
+        } else {
+            self.next_pn -= 1;
+        }
+
+        let delta = self.data.get_var().unwrap();
```
djc, line 1000: Why is this safe?

quinn-proto/src/lib.rs
```diff
@@ -48,6 +48,7 @@ pub use crate::connection::{
 };
 
 mod config;
+pub use config::AckTimestampsConfig;
```
djc, line 51: Nit: move this into the block below.

quinn-proto/src/transport_parameters.rs
```diff
@@ -160,6 +166,17 @@ impl TransportParameters {
...
+            receive_timestamps_exponent: config
+                .ack_timestamp_config
+                .as_ref()
+                .map(|cfg| cfg.exponent),
+
+            max_recv_timestamps_per_ack: config
+                .ack_timestamp_config
+                .as_ref()
+                .map(|cfg| cfg.max_timestamps_per_ack),
+
```
djc, line 179: I think these maybe shouldn't be separate `Option`s, but a tuple wrapped in a single `Option`?

djc (commented):
Here's a bunch of mostly stylistic feedback.

quinn-proto/src/connection/mod.rs
sterlingdeng: It doesn't seem possible to `rev()` on the source iterator because its yielding elements directly from the frame. It's similar to the `AckIter` implementation where each iteration yields a value as it reads from the `Ack` frame. The values yielded by the `AckIter` are added to an `ArrayRangeSet` which is space optimized for handling ranges, and is backed by a `TinyVec`.

What I'm essentially trying to do is match a packet number and it's timestamp (if available) together by "zipping" both together. When reading the values from the frame (for both packet number and timestamp), the values are produced from high packet number to low, but when we use them later on in the code, they're used from low to high, hence why I do a reverse.

One option to reduce an allocation and avoid the need of `reverse`ing is to attempt to zip the packet number and timestamps together as the values are yielded when reading from the frame. However, the issue that I run into is that the `array_range_set` is space optimized so it only records the ranges. If I wanted to zip the packet number and timestamps together, the `array_range_set` would have to be modified to associate a packet number value to a timestamp; example tuple `(packet_number, Option<timestamp>)`.

However, I don't think it makes sense to modify the existing `array_range_set` for this draft, so I opted to handle it this way, which is to read all the timestamps value into a Vec, and then handle the timestamp and packet_number matching later in the code.

sterlingdeng: Okay nvm, I think I found a better solution - going to test it and report back..

quinn-proto/src/connection/mod.rs
sterlingdeng: I fixed it so it does a `if let Some` check

quinn-proto/src/connection/mod.rs
sterlingdeng: Removed - it wasn't.

quinn-proto/src/connection/mod.rs
sterlingdeng: No longer needed since the algorithm was reworked.

quinn-proto/src/connection/receiver_timestamps.rs
sterlingdeng: No longer needed after refactor

quinn-proto/src/transport_parameters.rs
djc: > It could be worth proposing for the draft that the transport parameters related to this feature are grouped together similar to how PreferredAddress does it.

Yes, that would probably be good.

quinn-proto/src/transport_parameters.rs
```diff
@@ -99,6 +99,9 @@ macro_rules! make_struct {
             pub(crate) stateless_reset_token: Option<ResetToken>,
             /// The server's preferred address for communication after handshake completion
             pub(crate) preferred_address: Option<PreferredAddress>,
+
+
+            pub(crate) ack_timestamps_cfg: AckTimestampsConfig,
```
Ralith, line 104: New transport parameters must document whether they are preserved across 0-RTT or not. What is the case here? Note that e.g. `ack_delay_exponent` is *not* preserved.

Ralith, line 104: `Option`?

quinn-proto/src/packet.rs
```diff
@@ -969,7 +969,6 @@ mod tests {
         for byte in &buf {
             print!("{byte:02x}");
         }
-        println!();
```
Ralith, line 972: Accidental change?

quinn-proto/src/frame.rs
```diff
@@ -381,13 +391,21 @@ impl Ack {
         delay: u64,
         ranges: &ArrayRangeSet,
         ecn: Option<&EcnCounts>,
+        timestamps: Option<(
+            &ReceiverTimestamps,
+            Instant,
+            u64, // exponent
+            u64, // max_timestamps
+        )>,
```
Ralith, line 399: Large tuples are unreadable. If you must bundle these together, use a struct.

quinn-proto/src/frame.rs
```diff
@@ -403,14 +421,111 @@ impl Ack {
...
+    }
+
+    // https://www.ietf.org/archive/id/draft-smith-quic-receive-ts-00.html#ts-ranges
+    fn encode_timestamps(
+        timestamps: &ReceiverTimestamps,
+        mut largest: u64,
+        buf: &mut impl BufMut,
+        mut basis: Instant,
+        exponent: u64,
+        max_timestamps: u64,
```
Ralith, line 448: This doesn't seem to be used for anything but a `debug_assert`. If we trust the caller to pass the right limit down, we might as well just trust the caller to truncate the timestamp list.

quinn-proto/src/connection/receiver_timestamps.rs
```diff
@@ -0,0 +1,126 @@
...
+pub struct PacketTimestamp {
+    pub packet_number: u64,
+    pub timestamp: Instant,
+}
+
+impl Default for PacketTimestamp {
+    fn default() -> Self {
+        Self {
+            packet_number: 0,
+            timestamp: Instant::now(),
```
Ralith, line 15: It's weird for `default()` to return different values over time. Maybe encode the delta here instead? Or, do we even need a `Default` impl?

quinn-proto/src/connection/receiver_timestamps.rs
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+
+    pub(crate) fn add(&mut self, packet_number: u64, timestamp: Instant) {
+        if self.data.len() == self.max {
+            self.data.pop_front();
+        }
+        if let Some(v) = self.data.back() {
+            if packet_number <= v.packet_number {
+                warn!("out of order packets are not supported");
```
Ralith, line 59: This will be very noisy on some real networks. It should be `trace` at most, or omitted entirely.

quinn-proto/src/connection/receiver_timestamps.rs
```diff
@@ -0,0 +1,126 @@
...
+                return;
+            }
+        }
+        self.data.push_back(PacketTimestamp {
+            packet_number,
+            timestamp,
+        });
+    }
+
+    pub(crate) fn subtract_below(&mut self, packet_number: u64) {
```
Ralith, line 69: Document.

quinn-proto/src/connection/receiver_timestamps.rs
```diff
@@ -0,0 +1,126 @@
...
+            .data
+            .partition_point(|v| v.packet_number < packet_number);
+        if idx == self.data.len() {
+            self.data.clear();
+        } else {
+            let _ = self.data.drain(0..=idx);
+        }
+    }
+
+    pub(crate) fn iter(&self) -> impl DoubleEndedIterator<Item = PacketTimestamp> + '_ {
```
Ralith, line 83: Document order.

quinn-proto/src/connection/mod.rs
```diff
@@ -3309,6 +3362,12 @@ impl Connection {
         self.path.mtud.on_peer_max_udp_payload_size_received(
             u16::try_from(self.peer_params.max_udp_payload_size.into_inner()).unwrap_or(u16::MAX),
         );
+        self.ack_timestamps_cfg = params.ack_timestamps_cfg;
+        if let Some(max) = params.ack_timestamps_cfg.max_timestamps_per_ack {
+            for space in self.spaces.iter_mut() {
+                space.pending_acks.set_receiver_timestamp(max.0 as usize);
```
Ralith, line 3368: Clamp this maximum to a reasonable value to guard against memory exhaustion attacks.

Ralith (changes requested):
As noted in the issue, please work with the IETF to reserve IDs. To avoid interop problems, public implementations must not use unreserved IDs

quinn-proto/src/frame.rs
Ralith: > I think it's considered safe because this is defined in the spec.

What's in the spec doesn't matter. Quinn must not panic just because a peer is non-conformant. That would be a DoS vulnerability.

I imagine this is safe because `scan_ack_timestamps_blocks` validated the ranges already. If that was the intent, document it, even if we've failed to do so elsewhere.

djc:
This doesn't seem to be going anywhere -- closing it for now, feel free to reopen if you want to work on this again.

Issue #1971, opened 2024-08-21
Don't silence errors from `quinn-udp` code


thomaseizinger:
Currently, `quinn-udp` logs certain errors and returns `Ok(())` even if the packet did not get sent. For example, a common one we see is 

```
2024-08-21T00:34:51.104902Z  WARN quinn_udp: sendmsg error: Os { code: 101, kind: NetworkUnreachable, message: "Network unreachable" }, Transmit: { destination: [2600:1900:40b0:1504:0:94::]:3478, src_ip: None, enc: None, len: 108, segment_size: None }
```

This happens when we have an IPv6 interface but only with a link-local address. Thus, binding the socket to the unspecified address works but sending any kind of packet fails.

I would like to catch this in my application and disable the socket but I cannot because `quinn-udp` returns `Ok(())`.

Would you be willing to change this? It appears that ignoring a send error within library code is too opinionated and perhaps a layer above (`quinn` itself) would be better suited for this.

Note that I am not using `quinn`, we only use `quinn-udp` as a socket library.

Related: #1750.

Ralith:
I'm not sure exposing more errors is useful. The motivation behind the current behavior is that UDP network errors are almost invariably one of two cases: those that can be trivially spoofed by an attacker, and those that may be transient. For example, do you want to disable your socket just because the user's wifi drops out for a moment, or they changed networks?

thomaseizinger:
> I'm not sure exposing more errors is useful. The motivation behind the current behavior is that UDP network errors are almost invariably one of two cases: those that can be trivially spoofed by an attacker, and those that may be transient. For example, do you want to disable your socket just because the user's wifi drops out for a moment, or they changed networks?

We don't hard-fail in our app for any IO errors for exactly that reason :)

As far as I know, `NetworkReachable` gets emitted if the socket is incapable of routing the packet at all. That appears to be permanent (until network connections change at which point we make a new socket).

I am test-driving a change in a fork right now. I can report how that is going in a couple of days!

In general though, I think it wouldn't hurt to expose the errors at this layer. It gives more flexibility to users which makes it IMO a better library.

Ralith:
> permanent (until network connections change at which point we make a new socket).

"Permanent until it changes" isn't very permanent. Further, if your socket is bound to the wildcard address, as typical, there's no point recreating it in response to network changes. Detecting network changes is also difficult, especially in portable code.

> It gives more flexibility to users

[...]

PR #1993, opened 2024-09-20
feat: Faster UDP/IO on Apple platforms

larseggert:
This uses Apple's private [`sendmsg_x`](https://github.com/apple-oss-distributions/xnu/blob/94d3b452840153a99b38a3a9659680b2a006908e/bsd/sys/socket.h#L1457-L1487) and [`recvmsg_x`](https://github.com/apple-oss-distributions/xnu/blob/94d3b452840153a99b38a3a9659680b2a006908e/bsd/sys/socket.h#L1425-L1455) system calls for multi-packet UDP I/O.

CC @mxinden

### DIFF
--- a/quinn-udp/benches/throughput.rs
+++ b/quinn-udp/benches/throughput.rs
@@ -23,7 +23,7 @@ pub fn criterion_benchmark(c: &mut Criterion) {
     let mut permutations = vec![];
     for gso_enabled in [
         false,
-        #[cfg(any(target_os = "linux", target_os = "windows"))]
+        #[cfg(any(target_os = "linux", target_os = "windows", apple))]
         true,
     ] {
         for gro_enabled in [false, true] {
--- a/quinn-udp/build.rs
+++ b/quinn-udp/build.rs
@@ -1,0 +1,26 @@
+use cfg_aliases::cfg_aliases;
+
+fn main() {
+    // Setup cfg aliases
+    cfg_aliases! {
+        // Platforms
+        apple: {
+            any(
+                target_os = "macos",
+                target_os = "ios",
+                target_os = "tvos",
+                target_os = "visionos"
+            )
+        },
+        bsd: {
+            any(
+                target_os = "freebsd",
+                target_os = "openbsd",
+                target_os = "netbsd"
+            )
+        },
+        // Convenience aliases
+        apple_fast: { all(apple, feature = "fast-apple-datapath") },
+        apple_slow: { all(apple, not(feature = "fast-apple-datapath")) },
+    }
+}
--- a/quinn-udp/src/cmsg/unix.rs
+++ b/quinn-udp/src/cmsg/unix.rs
@@ -32,6 +32,29 @@ impl MsgHdr for libc::msghdr {
     }
 }
 
+#[cfg(apple_fast)]
+impl MsgHdr for crate::imp::msghdr_x {
+    type ControlMessage = libc::cmsghdr;
+
+    fn cmsg_first_hdr(&self) -> *mut Self::ControlMessage {
+        let selfp = self as *const _ as *mut libc::msghdr;
+        unsafe { libc::CMSG_FIRSTHDR(selfp) }
+    }
+
+    fn cmsg_nxt_hdr(&self, cmsg: &Self::ControlMessage) -> *mut Self::ControlMessage {
+        let selfp = self as *const _ as *mut libc::msghdr;
+        unsafe { libc::CMSG_NXTHDR(selfp, cmsg) }
+    }
+
+    fn set_control_len(&mut self, len: usize) {
+        self.msg_controllen = len as _;
+    }
+
+    fn control_len(&self) -> usize {
+        self.msg_controllen as _
+    }
+}
+
 /// Helpers for [`libc::cmsghdr`]
 impl CMsgHdr for libc::cmsghdr {
     fn cmsg_len(length: usize) -> usize {
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -1,11 +1,4 @@
-#[cfg(not(any(
-    target_os = "macos",
-    target_os = "ios",
-    target_os = "tvos",
-    target_os = "visionos",
-    target_os = "openbsd",
-    target_os = "solaris",
-)))]
+#[cfg(not(any(apple, target_os = "openbsd", target_os = "solaris")))]
 use std::ptr;
 use std::{
     io::{self, IoSliceMut},
@@ -25,6 +18,38 @@ use super::{
     cmsg, log::debug, log_sendmsg_error, EcnCodepoint, RecvMeta, Transmit, UdpSockRef,
     IO_ERROR_LOG_INTERVAL,
 };
+
+// Adapted from https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/bsd/sys/socket_private.h
+#[cfg(apple_fast)]
+#[repr(C)]
+#[allow(non_camel_case_types)]
+pub(crate) struct msghdr_x {
+    pub msg_name: *mut libc::c_void,
+    pub msg_namelen: libc::socklen_t,
+    pub msg_iov: *mut libc::iovec,
+    pub msg_iovlen: libc::c_int,
+    pub msg_control: *mut libc::c_void,
+    pub msg_controllen: libc::socklen_t,
+    pub msg_flags: libc::c_int,
+    pub msg_datalen: usize,
+}
+
+#[cfg(apple_fast)]
+extern "C" {
+    fn recvmsg_x(
+        s: libc::c_int,
+        msgp: *const msghdr_x,
+        cnt: libc::c_uint,
+        flags: libc::c_int,
+    ) -> isize;
+
+    fn sendmsg_x(
+        s: libc::c_int,
+        msgp: *const msghdr_x,
+        cnt: libc::c_uint,
+        flags: libc::c_int,
+    ) -> isize;
+}
 
 // Defined in netinet6/in6.h on OpenBSD, this is not yet exported by the libc crate
 // directly.  See https://github.com/rust-lang/libc/issues/3704 for when we might be able to
@@ -63,13 +88,8 @@ impl UdpSocketState {
         let io = sock.0;
         let mut cmsg_platform_space = 0;
         if cfg!(target_os = "linux")
-            || cfg!(target_os = "freebsd")
-            || cfg!(target_os = "openbsd")
-            || cfg!(target_os = "netbsd")
-            || cfg!(target_os = "macos")
-            || cfg!(target_os = "ios")
-            || cfg!(target_os = "tvos")
-            || cfg!(target_os = "visionos")
+            || cfg!(bsd)
+            || cfg!(apple)
             || cfg!(target_os = "android")
             || cfg!(target_os = "solaris")
         {
@@ -131,13 +151,7 @@ impl UdpSocketState {
                 )?;
             }
         }
-        #[cfg(any(
-            target_os = "freebsd",
-            target_os = "macos",
-            target_os = "ios",
-            target_os = "tvos",
-            target_os = "visionos"
-        ))]
+        #[cfg(any(target_os = "freebsd", apple))]
         {
             if is_ipv4 {
                 // Set `may_fragment` to `true` if this option is not supported on the platform.
@@ -149,16 +163,7 @@ impl UdpSocketState {
                 )?;
             }
         }
-        #[cfg(any(
-            target_os = "freebsd",
-            target_os = "openbsd",
-            target_os = "netbsd",
-            target_os = "macos",
-            target_os = "ios",
-            target_os = "tvos",
-            target_os = "visionos",
-            target_os = "solaris",
-        ))]
+        #[cfg(any(bsd, apple, target_os = "solaris"))]
         // IP_RECVDSTADDR == IP_SENDSRCADDR on FreeBSD
         // macOS uses only IP_RECVDSTADDR, no IP_SENDSRCADDR on macOS (the same on Solaris)
         // macOS also supports IP_PKTINFO
@@ -236,27 +241,13 @@ impl UdpSocketState {
     }
 
     /// Sets the flag indicating we got EINVAL error from `sendmsg` syscall.
-    #[cfg(not(any(
-        target_os = "macos",
-        target_os = "ios",
-        target_os = "tvos",
-        target_os = "visionos",
-        target_os = "openbsd",
-        target_os = "netbsd"
-    )))]
+    #[cfg(not(any(apple, target_os = "openbsd", target_os = "netbsd")))]
     fn set_sendmsg_einval(&self) {
         self.sendmsg_einval.store(true, Ordering::Relaxed)
     }
 }
 
-#[cfg(not(any(
-    target_os = "macos",
-    target_os = "ios",
-    target_os = "tvos",
-    target_os = "visionos",
-    target_os = "openbsd",
-    target_os = "netbsd"
-)))]
+#[cfg(not(any(apple, target_os = "openbsd", target_os = "netbsd")))]
 fn send(
     #[allow(unused_variables)] // only used on Linux
     state: &UdpSocketState,
@@ -341,14 +332,67 @@ fn send(
     }
 }
 
-#[cfg(any(
-    target_os = "macos",
-    target_os = "ios",
-    target_os = "tvos",
-    target_os = "visionos",
-    target_os = "openbsd",
-    target_os = "netbsd"
-))]
+#[cfg(apple_fast)]
+fn send(state: &UdpSocketState, io: SockRef<'_>, transmit: &Transmit<'_>) -> io::Result<()> {
+    let mut hdrs = unsafe { mem::zeroed::<[msghdr_x; BATCH_SIZE]>() };
+    let mut iovs = unsafe { mem::zeroed::<[libc::iovec; BATCH_SIZE]>() };
+    let mut ctrls = [cmsg::Aligned([0u8; CMSG_LEN]); BATCH_SIZE];
+    let addr = socket2::SockAddr::from(transmit.destination);
+    let segment_size = transmit.segment_size.unwrap_or(transmit.contents.len());
+    let mut cnt = 0;
+    debug_assert!(transmit.contents.len().div_ceil(segment_size) <= BATCH_SIZE);
+    for (i, chunk) in transmit
+        .contents
+        .chunks(segment_size)
+        .enumerate()
+        .take(BATCH_SIZE)
+    {
+        prepare_msg(
+            &Transmit {
+                destination: transmit.destination,
+                ecn: transmit.ecn,
+                contents: chunk,
+                segment_size: Some(chunk.len()),
+                src_ip: transmit.src_ip,
+            },
+            &addr,
+            &mut hdrs[i],
+            &mut iovs[i],
+            &mut ctrls[i],
+            true,
+            state.sendmsg_einval(),
+        );
+        hdrs[i].msg_datalen = chunk.len();
+        cnt += 1;
+    }
+    let n = unsafe { sendmsg_x(io.as_raw_fd(), hdrs.as_ptr(), cnt as u32, 0) };
+    if n >= 0 {
+        return Ok(());
+    }
+    let e = io::Error::last_os_error();
+    match e.kind() {
+        io::ErrorKind::Interrupted => {
+            // Retry the transmission
+        }
+        io::ErrorKind::WouldBlock => return Err(e),
+        _ => {
+            // Other errors are ignored, since they will usually be handled
+            // by higher level retransmits and timeouts.
+            // - PermissionDenied errors have been observed due to iptable rules.
+            //   Those are not fatal errors, since the
+            //   configuration can be dynamically changed.
+            // - Destination unreachable errors have been observed for other
+            // - EMSGSIZE is expected for MTU probes. Future work might be able to avoid
+            //   these by automatically clamping the MTUD upper bound to the interface MTU.
+            if e.raw_os_error() != Some(libc::EMSGSIZE) {
+                log_sendmsg_error(&state.last_send_error, e, transmit);
+            }
+        }
+    }
+    Ok(())
+}
+
+#[cfg(any(target_os = "openbsd", target_os = "netbsd", apple_slow))]
 fn send(state: &UdpSocketState, io: SockRef<'_>, transmit: &Transmit<'_>) -> io::Result<()> {
     let mut hdr: libc::msghdr = unsafe { mem::zeroed() };
     let mut iov: libc::iovec = unsafe { mem::zeroed() };
@@ -360,12 +404,7 @@ fn send(state: &UdpSocketState, io: SockRef<'_>, transmit: &Transmit<'_>) -> io:
         &mut hdr,
         &mut iov,
         &mut ctrl,
-        cfg!(target_os = "macos")
-            || cfg!(target_os = "ios")
-            || cfg!(target_os = "tvos")
-            || cfg!(target_os = "visionos")
-            || cfg!(target_os = "openbsd")
-            || cfg!(target_os = "netbsd"),
+        cfg!(apple) || cfg!(target_os = "openbsd") || cfg!(target_os = "netbsd"),
         state.sendmsg_einval(),
     );
     let n = unsafe { libc::sendmsg(io.as_raw_fd(), &hdr, 0) };
@@ -394,14 +433,7 @@ fn send(state: &UdpSocketState, io: SockRef<'_>, transmit: &Transmit<'_>) -> io:
     Ok(())
 }
 
-#[cfg(not(any(
-    target_os = "macos",
-    target_os = "ios",
-    target_os = "tvos",
-    target_os = "visionos",
-    target_os = "openbsd",
-    target_os = "solaris",
-)))]
+#[cfg(not(any(apple, target_os = "openbsd", target_os = "solaris")))]
 fn recv(io: SockRef<'_>, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) -> io::Result<usize> {
     let mut names = [MaybeUninit::<libc::sockaddr_storage>::uninit(); BATCH_SIZE];
     let mut ctrls = [cmsg::Aligned(MaybeUninit::<[u8; CMSG_LEN]>::uninit()); BATCH_SIZE];
@@ -440,14 +472,35 @@ fn recv(io: SockRef<'_>, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) -> 
     Ok(msg_count as usize)
 }
 
-#[cfg(any(
-    target_os = "macos",
-    target_os = "ios",
-    target_os = "tvos",
-    target_os = "visionos",
-    target_os = "openbsd",
-    target_os = "solaris",
-))]
+#[cfg(apple_fast)]
+fn recv(io: SockRef<'_>, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) -> io::Result<usize> {
+    let mut names = [MaybeUninit::<libc::sockaddr_storage>::uninit(); BATCH_SIZE];
+    let mut ctrls = [cmsg::Aligned(MaybeUninit::<[u8; CMSG_LEN]>::uninit()); BATCH_SIZE];
+    let mut hdrs = unsafe { mem::zeroed::<[msghdr_x; BATCH_SIZE]>() };
+    let max_msg_count = bufs.len().min(BATCH_SIZE);
+    for i in 0..max_msg_count {
+        prepare_recv(&mut bufs[i], &mut names[i], &mut ctrls[i], &mut hdrs[i]);
+    }
+    let msg_count = loop {
+        let n = unsafe { recvmsg_x(io.as_raw_fd(), hdrs.as_mut_ptr(), max_msg_count as _, 0) };
+        match n {
+            -1 => {
+                let e = io::Error::last_os_error();
+                if e.kind() == io::ErrorKind::Interrupted {
+                    continue;
+                }
+                return Err(e);
+            }
+            n => break n,
+        }
+    };
+    for i in 0..(msg_count as usize) {
+        meta[i] = decode_recv(&names[i], &hdrs[i], hdrs[i].msg_datalen as usize);
+    }
+    Ok(msg_count as usize)
+}
+
+#[cfg(any(target_os = "openbsd", target_os = "netbsd", apple_slow))]
 fn recv(io: SockRef<'_>, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) -> io::Result<usize> {
     let mut name = MaybeUninit::<libc::sockaddr_storage>::uninit();
     let mut ctrl = cmsg::Aligned(MaybeUninit::<[u8; CMSG_LEN]>::uninit());
@@ -476,7 +529,8 @@ const CMSG_LEN: usize = 88;
 fn prepare_msg(
     transmit: &Transmit<'_>,
     dst_addr: &socket2::SockAddr,
-    hdr: &mut libc::msghdr,
+    #[cfg(not(apple_fast))] hdr: &mut libc::msghdr,
+    #[cfg(apple_fast)] hdr: &mut msghdr_x,
     iov: &mut libc::iovec,
     ctrl: &mut cmsg::Aligned<[u8; CMSG_LEN]>,
     #[allow(unused_variables)] // only used on FreeBSD & macOS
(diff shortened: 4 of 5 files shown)


Ralith:
> Is there interest in seeing TX support via sendmsg_x?

We found there wasn't much performance benefit, and was considerable difficulty taking advantage of, `sendmmsg`-style batching. IIRC the `_x` functions on macOS have more to offer than that, though. Will this unblock segmentation offload or other incidental optimizations?

quinn-udp/build.rs
```diff
@@ -0,0 +1,22 @@
+#![cfg(any(target_os = "macos", target_os = "ios"))]
+
+use std::env;
+use std::path::PathBuf;
+
+fn main() {
+    // Generate the bindings for Apple's private `recvmsg_x` from
+    // https://github.com/apple-oss-distributions/xnu/blob/main/bsd/sys/socket.h.
+    let bindings = bindgen::Builder::default()
```
Ralith, line 8: Bindgen is a pretty big and unstable dependency. Would it be simpler to commit the generated bindings?

larseggert:
Bench on `main`:
```
test large_data_10_streams  ... bench:  27,558,791 ns/iter (+/- 13,459,810) = 380 MB/s
test large_data_1_stream    ... bench:  24,324,266 ns/iter (+/- 19,219,937) = 43 MB/s
test small_data_100_streams ... bench:  19,437,900 ns/iter (+/- 20,065,941)
test small_data_1_stream    ... bench:  11,465,128 ns/iter (+/- 8,699,934)
```

Bench with this PR:
```
test large_data_10_streams  ... bench:  28,829,216 ns/iter (+/- 15,924,956) = 363 MB/s
test large_data_1_stream    ... bench:  14,354,999 ns/iter (+/- 20,039,122) = 73 MB/s
test small_data_100_streams ... bench:  14,061,741 ns/iter (+/- 17,311,517)
test small_data_1_stream    ... bench:  19,194,441 ns/iter (+/- 5,012,070)
```

Surprised that `large_data_10_streams` and `small_data_1_stream` are slower...

larseggert:
We've also found on neqo that multi-packet RX without multi-packet TX has limited benefits, since the RX batch size will be very small.

larseggert:
I added `sendmsg_x` support, mostly to see what the performance difference would be. But it seems that none of the benches or tests call `send` with a `Transmit` struct where `segment_size` is not `None`?

mxinden:
> A targeted quinn-udp benchmark might be more useful.

How about using the `throughput.rs` benchmark @larseggert?

https://github.com/quinn-rs/quinn/blob/main/quinn-udp/benches/throughput.rs

larseggert:
With @mxinden's benchmark. Baseline:
```
gso_true/throughput     time:   [58.076 ms 58.230 ms 58.387 ms]
                        thrpt:  [171.27 MiB/s 171.73 MiB/s 172.19 MiB/s]
```

Only `sendmsg_x`:
```
gso_true/throughput     time:   [15.143 ms 15.189 ms 15.236 ms]
                        thrpt:  [656.35 MiB/s 658.36 MiB/s 660.37 MiB/s]
                 change:
                        time:   [-74.028% -73.915% -73.808%] (p = 0.00 < 0.05)
                        thrpt:  [+281.80% +283.36% +285.04%]
                        Performance has improved.
```

Both `sendmsg_x` and `recvmsg_x`:
```
gso_true/throughput     time:   [12.632 ms 12.682 ms 12.731 ms]
                        thrpt:  [785.46 MiB/s 788.53 MiB/s 791.61 MiB/s]
                 change:
                        time:   [-78.321% -78.221% -78.112%] (p = 0.00 < 0.05)
                        thrpt:  [+356.88% +359.16% +361.27%]
                        Performance has improved.
```

quinn-udp/benches/throughput.rs
```diff
@@ -27,8 +27,13 @@ pub fn criterion_benchmark(c: &mut Criterion) {
     // Reverse non-blocking flag set by `UdpSocketState` to make the test non-racy
     recv.set_nonblocking(false).unwrap();
 
-    let mut receive_buffer = vec![0; MAX_BUFFER_SIZE];
-    let mut meta = RecvMeta::default();
+    let gro_segments = UdpSocketState::new((&send).into()).unwrap().gro_segments();
+    let mut receive_buffers = vec![[0; SEGMENT_SIZE]; gro_segments];
+    let mut receive_slives = receive_buffers
```
djc: `slives` is probably not what you were going for?

quinn-udp/src/unix.rs
```diff
@@ -325,12 +358,64 @@ fn send(
...
+                &mut ctrls[i],
+                true,
+                state.sendmsg_einval(),
+            );
+            hdrs[i].msg_datalen = chunk.len();
+        })
+        .count();
+    let n = unsafe { sendmsg_x(io.as_raw_fd(), hdrs.as_ptr(), cnt as u32, 0) };
+    // print!("o{} ", n);
+    if n == -1 {
```
djc, line 393: Nit: please flip the condition and make this an early return.

quinn-udp/src/unix.rs
```diff
@@ -325,12 +358,64 @@ fn send(
...
+                &mut iovs[i],
+                &mut ctrls[i],
+                true,
+                state.sendmsg_einval(),
+            );
+            hdrs[i].msg_datalen = chunk.len();
+        })
+        .count();
+    let n = unsafe { sendmsg_x(io.as_raw_fd(), hdrs.as_ptr(), cnt as u32, 0) };
+    // print!("o{} ", n);
```
djc, line 392: I think we can get rid of the commented out debug cruft now?

quinn-udp/src/unix.rs
```diff
@@ -418,12 +500,41 @@ fn recv(io: SockRef<'_>, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) ->
...
+    let msg_count = loop {
+        let n = unsafe {
+            recvmsg_x(
+                io.as_raw_fd(),
+                hdrs.as_mut_ptr(),
+                bufs.len().min(BATCH_SIZE) as _,
+                0,
+            )
+        };
+        if n == -1 {
```
djc, line 521: Nit: suggest spelling it like this:

```rust
match n {
    -1 => {
         ..
    }
    n => break n,
}
```

quinn-udp/src/unix.rs
larseggert: This always returns `Ok(())`, just as in the BSD version below. (Should both of these actually return an Error in this case?)

quinn-udp/src/unix.rs
```diff
@@ -325,12 +357,63 @@ fn send(
...
+                    segment_size: Some(chunk.len()),
+                    src_ip: transmit.src_ip,
+                },
+                &addr,
+                &mut hdrs[i],
+                &mut iovs[i],
+                &mut ctrls[i],
+                true,
+                state.sendmsg_einval(),
+            );
```
mxinden, line 386: Nit: It is not idiomatic to execute a side-effect in an `Iterator::map`.

> If you’re doing some sort of looping for a side effect, it’s considered more idiomatic to use [for](https://doc.rust-lang.org/stable/book/ch03-05-control-flow.html#looping-through-a-collection-with-for) than map().

https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.map

While this might seem dogmatic, I do find reasoning about an Iterator chain much easier, when being able to rely on the convention of no-side-effects-in-map.

Instead I suggest either using a standard `for` loop here, or `Iterator::for_each`, updating `cnt` on each iteration.

quinn-udp/src/unix.rs
mxinden: > Nit: please flip the condition and make this an early return.

Agreed with @larseggert. I suggest keeping as is, as both paths (`== -1` and `>-1`) have the same return value `Ok(())` and given that it mirrors the BSD implementation below.

> Should both of these actually return an Error in this case?

In case we do want to return an error here, I suggest doing this as part of https://github.com/quinn-rs/quinn/issues/1971 across all platforms (e.g. as you mention BSD) and not in this pull request.

quinn-udp/src/unix.rs
```diff
@@ -690,7 +816,18 @@ mod gso {
...
 }
 
-#[cfg(not(target_os = "linux"))]
+#[cfg(any(target_os = "macos", target_os = "ios"))]
+mod gso {
+    use super::*;
+
+    pub(super) fn max_gso_segments() -> usize {
+        BATCH_SIZE
+    }
```
mxinden, line 825: I think it is worth documenting here, that UDP Datagram segmentation is not offloaded to the OS or even better the NIC, but instead done here in user space in `fn send()` and then passed to the OS as individual `iovec`s.

mxinden (commented):
Impressive results. Great to see the MacOS `_x` syscalls work for QUIC UDP IO.

quinn-udp/benches/throughput.rs
```diff
@@ -27,8 +27,13 @@ pub fn criterion_benchmark(c: &mut Criterion) {
...
 
-    let mut receive_buffer = vec![0; MAX_BUFFER_SIZE];
-    let mut meta = RecvMeta::default();
+    let gro_segments = UdpSocketState::new((&send).into()).unwrap().gro_segments();
+    let mut receive_buffers = vec![[0; SEGMENT_SIZE]; gro_segments];
+    let mut receive_slices = receive_buffers
+        .iter_mut()
+        .map(|buf| IoSliceMut::new(buf))
+        .collect::<Vec<_>>();
+    let mut meta = vec![RecvMeta::default(); gro_segments];
```
Ralith, line 36: This change seems to be confusing GRO with multi-message receive. GRO batches are passed to userspace as a single packet; that's why we used a single large buffer. By switching to many small buffers, you've actually disabled GRO.

Ralith (dismissed):
Can we enable real GSO/GRO using these interfaces?

larseggert:
No. They are the equivalent of the `mmsg` Linux calls. AFAIK Apple doesn't have GSO/GRO via the socket interface.

quinn-udp/benches/throughput.rs
larseggert: Apple doesn't **do** GSO/GRO, at least not via the socket interface. The `_x` calls are the semantic equivalent of the Linux `mmsg` calls. I'm borrowing the GSO/GRO helpers set up multi-message TX/RX in the bench.

quinn-udp/src/unix.rs
larseggert: Can do either. What do the maintainers prefer?

quinn-udp/src/cmsg/unix.rs
```diff
@@ -32,6 +32,34 @@ impl MsgHdr for libc::msghdr {
...
+    fn cmsg_nxt_hdr(&self, cmsg: &Self::ControlMessage) -> *mut Self::ControlMessage {
+        let selfp = self as *const _ as *mut libc::msghdr;
+        unsafe { libc::CMSG_NXTHDR(selfp, cmsg) }
+    }
+
+    fn set_control_len(&mut self, len: usize) {
+        self.msg_controllen = len as _;
+        if len == 0 {
+            // netbsd is particular about this being a NULL pointer if there are no control
+            // messages.
```
djc, line 53: I guess this comment isn't relevant for the macOS implementation?

quinn-udp/src/unix.rs
```diff
@@ -24,6 +24,38 @@ use super::{
     IO_ERROR_LOG_INTERVAL,
 };
 
+// Adapted from https://github.com/apple-oss-distributions/xnu/blob/main/bsd/sys/socket.h
+#[cfg(any(target_os = "macos", target_os = "ios"))]
+#[repr(C)]
+#[allow(non_camel_case_types)]
+pub(crate) struct msghdr_x {
+    pub msg_name: *mut std::ffi::c_void,
```
djc, line 32: Looks like so far we've imported symbols from `std::ffi` directly, keeping `libc` symbols as prefixed. I'd be in favor of sticking with that style.

quinn-udp/src/unix.rs
```diff
@@ -24,6 +24,38 @@ use super::{
...
+    pub msg_iov: *mut libc::iovec,
+    pub msg_iovlen: std::ffi::c_int,
+    pub msg_control: *mut std::ffi::c_void,
+    pub msg_controllen: libc::socklen_t,
+    pub msg_flags: std::ffi::c_int,
+    pub msg_datalen: usize,
+}
+
+#[cfg(any(target_os = "macos", target_os = "ios"))]
+extern "C" {
```
djc, line 43: Are you contributing this to upstream libc? (We can carry it until it gets released.)

quinn-udp/src/unix.rs
djc: Yes, agreed that a `for` loop would be better here.

quinn-udp/src/unix.rs
djc: > Agreed with @larseggert. I suggest keeping as is, as both paths (`== -1` and `>-1`) have the same return value `Ok(())` and given that it mirrors the BSD implementation below.

I don't find this a very strong argument. This saves some rightward drift for the error handling code and duplicating the `Ok(())` return value doesn't seem like a particularly big deal here. @Ralith opinions?

quinn-udp/src/unix.rs
```diff
@@ -770,7 +907,16 @@ fn set_socket_option(
 
 const OPTION_ON: libc::c_int = 1;
 
-#[cfg(not(target_os = "linux"))]
+#[cfg(any(target_os = "macos", target_os = "ios"))]
+mod gro {
+    use super::BATCH_SIZE;
+
+    pub(super) fn gro_segments() -> usize {
+        BATCH_SIZE
```
mxinden, line 915: `quinn-udp` **does not** expose `sendmmsg` style sends, i.e. only supports multiple sends via GSO from a single buffer with equal size segments. Thus the above hack of returning `BATCH_SIZE` on `max_gso_segments`, despite the fact that MacOS does not support GSO, while not ideal, seems reasonable. The large multi-segment datagram received from the user is segmented in user-space and then passed to MacOS as individual `iovec`s, `sendmmsg` style.

`quinn-udp` **does** expose `recvmmsg` style reads, i.e. supports receiving multiple datagrams both in separate `iovec`s and via GRO. Given that MacOS does not support GRO, why return `BATCH_SIZE` here and not `1`?

```suggestion
        1
```

quinn-udp/src/unix.rs
larseggert: I wasn't planning on, since these are private APIs.

quinn-udp/src/unix.rs
larseggert: So that `UdpSocketState::gro_segments` is initialized to something other than 1? What am I missing?

quinn-udp/src/unix.rs
djc: Ohh, had missed that. I guess they might not take them. So given that these are documented as

> NOTE: This a private system call, the API is subject to change.

Should we make this an opt-in feature? Does private here only mean "subject to change" or also "will prevent your app from being allowed in the App Store"?

larseggert: I don't know. I do know that these calls have been there for 10+ years. I can put them behind a feature if you prefer.

larseggert: Maybe @ekinnear knows...

quinn-udp/src/unix.rs
mxinden: > So that `UdpSocketState::gro_segments` is initialized to something other than 1?

Why does it need to be initialized to something other than `1`?

If `UdpSocketState::gro_segments` is `1`, a user can still provide multiple buffers to `UdpSocketState::recv` and, on the MacOS code-path, each of those buffers can be passed to the OS to be filled in a `recvmmsg` style.

mxinden: Put differently, the value of `UdpSocketState::gro_segments` only determines, how many segments a user can receive in one `iovec`, not how many `iovec`s a user can pass to `UdpSocketState::recv()`.

In the case of quinn as a user, it simply determines how large each `iovec` should be, i.e. how many segments each `iovec` should be able to fit:

https://github.com/quinn-rs/quinn/blob/9a656d2d0b3031390e71a24258b5432115cfa381/quinn/src/endpoint.rs#L745-L750

Note that `recv_buf` here represents all `iovec`s in sequence passed to `UdpSocketState::recv()`.

(Quinn maintainers, please correct me if I am missing something.)

quinn-udp/benches/throughput.rs
```diff
@@ -39,7 +43,7 @@ pub fn criterion_benchmark(c: &mut Criterion) {
 
         let transmit = Transmit {
             destination: dst_addr,
-            ecn: None,
+            ecn: Some(quinn_udp::EcnCodepoint::Ect0),
```
Ralith, line 46: Is this relevant to throughput testing? We already cover the ability to send/receive ECN in the unit tests.

quinn-udp/benches/throughput.rs
Ralith: Sure, but this isn't an Apple-specific test; it runs on all platforms quinn-udp supports, some of which do have real GRO/GSO. This change will undermine performance testing on those platforms.

larseggert: This changed to
```rust
    let mut meta = vec![RecvMeta::default(); BATCH_SIZE];
```
in the meantime; is this still a concern?

(Because it seems that for `recvmmsg` platforms, you already needed that change?)

quinn-udp/benches/throughput.rs
larseggert: Sorry this snuck in from local testing.

quinn-udp/benches/throughput.rs
mxinden: > This changed to
> 
> ```rust
>     let mut meta = vec![RecvMeta::default(); BATCH_SIZE];
> ```

:+1: Thanks.

> in the meantime; is this still a concern?
> 
> (Because it seems that for `recvmmsg` platforms, you already needed that change?)

I initially wrote this benchmark (https://github.com/quinn-rs/quinn/pull/1915) to test GSO/GRO only. I am in favor of your addition, namely to leverage `recvmmsg` as well, i.e. pass more than one `iovec` to the OS to read into. Thanks @larseggert. I assume that on MacOS this benchmark now returns the same result for both `gro_result` `true` and `false`, correct?

To avoid any confusion, other than MacOS and iOS, I don't think there are any supported platforms that do `recvmmsg` but not `GRO`.

Linux / Android does [recvmmsg](https://man7.org/linux/man-pages/man2/recvmmsg.2.html) with GRO,  and Windows does [WSARecvMsg](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/ms741687(v=vs.85)) (equivalent to `recvmmsg`) with GRO. The BSDs do neither `recvmmsg` nor `GRO`, at least as implemented in `quinn-udp`.

The only exception now, with this pull request, is MacOS and iOS with `recvmsg_x`, acting like `recvmmsg`, but not supporting GRO.

quinn-udp/src/unix.rs
mxinden: Yes. Thank you @larseggert.

larseggert:
Are you waiting on anything from me on this?

quinn-udp/src/unix.rs
```diff
@@ -690,7 +807,21 @@ mod gso {
     }
 }
 
-#[cfg(not(target_os = "linux"))]
+#[cfg(any(target_os = "macos", target_os = "ios"))]
+// On Apple platforms using the `sendmsg_x` call, UDP datagram segmentation is not
+// offloaded to the NIC or even the kernel, but instead done here in user space in
+// [`send`]) and then passed to the OS as individual `iovec`s (up to `BATCH_SIZE`).
```
mxinden, line 813:
```suggestion
// On Apple platforms using the `sendmsg_x` call, UDP datagram segmentation is not
// offloaded to the NIC or even the kernel, but instead done here in user space in
// [`send`]) and then passed to the OS as individual `iovec`s (up to `BATCH_SIZE`).
#[cfg(any(target_os = "macos", target_os = "ios"))]
```

I believe in this codebase convention is to put comments before attributes.

quinn-udp/src/unix.rs
mxinden: > I can put them behind a feature if you prefer.

No strong opinion.

Anyone aware of any other project using these APIs across apple platforms (e.g. MacOS and iOS) and platform versions? If not, I think putting these behind a feature flag, thus allowing for some experimentation first, is the safest way to go.

We can e.g. enable the feature flag on Firefox Nightly, thus at least get to test it on a large set of MacOS machines, before always enabling it.

mxinden (commented):
`quinn-udp/benches/throughput.rs` will need more changes to still support non-apple platform. @larseggert I believe we will need to either run it multi-threaded, or use some kind of executor, e.g. `tokio`. I can prepare a commit in the next couple of days. Sorry for missing this in earlier reviews.

Changes itself look good to me.

quinn-udp/src/unix.rs
```diff
@@ -568,6 +753,17 @@ fn prepare_msg(
...
 
+#[cfg(all(
+    not(feature = "fast-apple-datapath"),
+    any(
+        target_os = "macos",
+        target_os = "ios",
+        target_os = "tvos",
+        target_os = "visionos",
+        target_os = "openbsd",
+        target_os = "netbsd"
```
mxinden, line 764:
```suggestion
        target_os = "netbsd",
        target_os = "linux",
```

Also, what about solaris?

quinn-udp/src/unix.rs
```diff
@@ -568,6 +753,17 @@ fn prepare_msg(
     encoder.finish();
 }
 
+#[cfg(all(
+    not(feature = "fast-apple-datapath"),
```
mxinden, line 757: I think it would be good for `fast-apple-datapath` to be a no-op on non-apple platforms. As it is now, on linux, with `fast-apple-datapath`, this won't compile.

How about something along the lines of:

```
#[cfg(any(
  target_os = // all non apple platforms,
  all(
        not(feature = "fast-apple-datapath"),
        any( /* all apple platoforms */),
  )
))]

File: quinn-udp/benches/throughput.rs [hunk 17]
```diff
@@ -27,8 +27,12 @@ pub fn criterion_benchmark(c: &mut Criterion) {
     // Reverse non-blocking flag set by `UdpSocketState` to make the test non-racy
     recv.set_nonblocking(false).unwrap();
 
-    let mut receive_buffer = vec![0; MAX_BUFFER_SIZE];
-    let mut meta = RecvMeta::default();
+    let mut receive_buffers = vec![[0; SEGMENT_SIZE]; BATCH_SIZE];
```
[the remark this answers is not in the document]
mxinden: @larseggert will be fixed with https://github.com/larseggert/quinn/pull/1.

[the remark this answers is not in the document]
larseggert: Thanks! 

Also, since all the tests are green, I guess quinn is missing a unit test to detect this breakage...

djc:
@Ralith can you do another round on this one?

larseggert:
Once @mxinden's fix to the bench is in, I will squash the PR.

quinn-udp/benches/throughput.rs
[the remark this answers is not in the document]
mxinden: Above referenced pull request addressing this issue moved to https://github.com/quinn-rs/quinn/pull/2010.

AndrewDryga:
Hey guys 👋, is there any chance Apple won't approve apps that are using those private syscalls in the App Store? They are notorious for doing so and even de-listing apps for using anything "undocumented". See 2.5.1 here: https://developer.apple.com/app-store/review/guidelines/

larseggert:
The use of the private syscalls is now behind a non-default feature.

AndrewDryga:
@larseggert should we add a big fat warning saying that if you enable this flag you will violate Apple ToS so it's only should be enabled if app is not distributed (or notarized for EU)?

larseggert:
I have no opinion here - we don't distribute via the App Store. 

Could you make a suggestion on what you think would be good to add?

thomaseizinger:
> The use of the private syscalls is now behind a non-default feature.

Thank you for making it optional! It can be an issue to toggle code-paths like these with cargo features because they get aggregated across the entire dependency tree. Simply adding another dependency that also happens to depend on `quinn-udp` can thus activate this silently. (As a rule of thumb, cargo features should only add new APIs, not modify existing ones so this doesn't happen.)

For some prior art, `curve25519-dalek` had to solve similar challenges: https://github.com/dalek-cryptography/curve25519-dalek/tree/main/curve25519-dalek#backends

It looks like it is already abstracted away pretty well. Could we change this to a "plain" rustc cfg that needs to be set at build-time?
I am happy for it to be opt-out. Most apps using this probably aren't in the app store and toggling it off is a one-liner in the build process. Ultimately, I got no opinion though on opt-in/opt-out.

mxinden:
> Simply adding another dependency that also happens to depend on `quinn-udp` can thus activate this silently.

Ignoring the case where the activation is unintentional, if that other dependency can use `quinn-udp` with `sendmsg_x` and `recvmsg_x`, I don't see a reason why other consumers of `quinn-udp` within the same binary, shouldn't use `sendmsg_x` and `recvmsg_x`. In other words, I don't think this needs to be optional per consumer of `quinn-udp`, but instead only optional for the entire binary. Thus I think a feature flag suffices.

> Could we change this to a "plain" rustc cfg that needs to be set at build-time?

I might be missing something. Doesn't the same concern apply both to plain cargo features and rustc cfgs? Aren't rustc cfgs also global, as in apply to the entire binary?

If so, I would opt for the, in my eyes, more idiomatic cargo feature instead of the less intuitive rustc cfg.

djc:
> > Could we change this to a "plain" rustc cfg that needs to be set at build-time?
> 
> I might be missing something. Doesn't the same concern apply both to plain cargo features and rustc cfgs? Aren't rustc cfgs also global, as in apply to the entire binary?
> 
> If so, I would opt for the, in my eyes, more idiomatic cargo feature instead of the less intuitive rustc cfg.

The `cfg` flag is set at build time. Cargo features are propagated through the dependency graph so that if you add a dependency to your binary that has a feature-enabled dependency on one of your dependencies, that feature now also gets enabled for your project. This can "silently" happen without you noticing, so it's a bit of a footgun.

mxinden:
> The `cfg` flag is set at build time.

I don't understand why this is relevant. Both cargo features and rustc cfgs happen at build time, i.e. not at runtime.

> Cargo features are propagated through the dependency graph so that if you add a dependency to your binary that has a feature-enabled dependency on one of your dependencies, that feature now also gets enabled for your project.

Correct. My argument above is, that this isn't a problem here. I argue that if it is safe for one consumer of `quinn-udp` to use `sendmsg_x` and `recvmsg_x`, it is safe for all consumers of `quinn-udp` within the same binary to use `sendmsg_x` and `recvmsg_x`. 

Can you think of a scenario where that is not the case?

>  This can "silently" happen without you noticing, so it's a bit of a footgun.

In this particular case, if you don't trust your dependency to make wise choices on their `quinn-udp` feature selection, I don't think one should use that dependency in the first place.

> I am happy for it to be opt-out.

I suggest for it to be opt-in for now, given that we have little to no experience with these APIs, nor are we aware of any other applications using these APIs.

djc:
> > The `cfg` flag is set at build time.
> 
> I don't understand why this is relevant. Both cargo features and rustc cfgs happen at build time, i.e. not at runtime.

Because `cfg` flags are set from the command/environment where the end artifact is built, whereas Cargo features can be set by other dependencies which often aren't under your control.

> > Cargo features are propagated through the dependency graph so that if you add a dependency to your binary that has a feature-enabled dependency on one of your dependencies, that feature now also gets enabled for your project.
> 
> Correct. My argument above is, that this isn't a problem here. I argue that if it is safe for one consumer of `quinn-udp` to use `sendmsg_x` and `recvmsg_x`, it is safe for all consumers of `quinn-udp` within the same binary to use `sendmsg_x` and `recvmsg_x`.
> 
> Can you think of a scenario where that is not the case?
> 
> > This can "silently" happen without you noticing, so it's a bit of a footgun.
> 
> In this particular case, if one doesn't trust one's dependency to make wise choices on their `quinn-udp` feature selection, I don't think one should use that dependency in the first place.

I don't think we're going to solve the crates.io supply chain issues here, and I think not everyone is as careful as Mozilla as vetting their dependencies (and any updates happening to them over time). As such, IMO it is a best practice that additional functionality enabled via features should additionally be toggled by API choices. On the other hand, (a) I'm not sure how ergonomic that would be in this case, (b) I don't like the ergonomics of `cfg` as an alternative.

> > I am happy for it to be opt-out.
> 
> I suggest for it to be opt-in for now, given that we have little to no experience with these APIs, nor are we aware of any other applications using these APIs.

This seems reasonable to me.

mxinden:
> > > The `cfg` flag is set at build time.
> > 
> > 
> > I don't understand why this is relevant. Both cargo features and rustc cfgs happen at build time, i.e. not at runtime.
> 
> Because `cfg` flags are set from the command/environment where the end artifact is built, whereas Cargo features can be set by other dependencies which often aren't under your control.

Thank you for expanding on this. That makes sense.

larseggert:
See 4f86c5058740a9295b74f87c8cab6206a60e9971. Let me know if I misunderstood something.

quinn-udp/src/unix.rs
```diff
@@ -440,14 +472,42 @@ fn recv(io: SockRef<'_>, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) ->
...
+    let max_msg_count = bufs.len().min(BATCH_SIZE);
+    for i in 0..max_msg_count {
+        prepare_recv(&mut bufs[i], &mut names[i], &mut ctrls[i], &mut hdrs[i]);
+    }
+    let msg_count = loop {
+        let n = unsafe {
+            recvmsg_x(
+                io.as_raw_fd(),
+                hdrs.as_mut_ptr(),
+                bufs.len().min(BATCH_SIZE) as _,
```
Ralith, line 489:
```suggestion
                max_msg_count as _,
```

Ralith (approved):

djc:
> See [4f86c50](https://github.com/quinn-rs/quinn/commit/4f86c5058740a9295b74f87c8cab6206a60e9971). Let me know if I misunderstood something.

Discussed this with @Ralith privately, we feel that the use of `cfg` has a negative impact on the ergonomics and the chance of issues arising from silent feature enabling is pretty limited in this case, so would prefer to stick with a feature for this for now.

djc:
Sorry it took a while to get the benchmark PR merged, can you rebase on top of main?

larseggert:
Shouldn't the `features` CI workflow have included `fast-apple-datapath`? I don't see it in the logs. (Or I am missing something.)

mxinden:
Should be fine I believe:

```
running `cargo check --no-default-features --features direct-log,fast-apple-datapath,log,tracing` on quinn-udp (1754/1772)
running `cargo check --no-default-features --features default` on quinn-udp (1755/1772)
running `cargo check --no-default-features --features direct-log` on quinn-udp (1756/1772)
running `cargo check --no-default-features --features default,direct-log` on quinn-udp (1757/1772)
running `cargo check --no-default-features --features fast-apple-datapath` on quinn-udp (1758/1772)
running `cargo check --no-default-features --features default,fast-apple-datapath` on quinn-udp (1759/1772)
running `cargo check --no-default-features --features direct-log,fast-apple-datapath` on quinn-udp (1760/1772)
running `cargo check --no-default-features --features default,direct-log,fast-apple-datapath` on quinn-udp (1761/1772)
running `cargo check --no-default-features --features log` on quinn-udp (1762/1772)
running `cargo check --no-default-features --features direct-log,log` on quinn-udp (1763/1772)
running `cargo check --no-default-features --features fast-apple-datapath,log` on quinn-udp (1764/1772)
running `cargo check --no-default-features --features direct-log,fast-apple-datapath,log` on quinn-udp (1765/1772)
running `cargo check --no-default-features --features tracing` on quinn-udp (1766/1772)
running `cargo check --no-default-features --features direct-log,tracing` on quinn-udp (1767/1772)
running `cargo check --no-default-features --features fast-apple-datapath,tracing` on quinn-udp (1768/1772)
running `cargo check --no-default-features --features direct-log,fast-apple-datapath,tracing` on quinn-udp (1769/1772)
running `cargo check --no-default-features --features log,tracing` on quinn-udp (1770/1772)
running `cargo check --no-default-features --features direct-log,log,tracing` on quinn-udp (1771/1772)
running `cargo check --no-default-features --features fast-apple-datapath,log,tracing` on quinn-udp (1772/1772)
```

quinn-udp/src/unix.rs
```diff
@@ -341,14 +332,67 @@ fn send(
...
+    }
+    let n = unsafe { sendmsg_x(io.as_raw_fd(), hdrs.as_ptr(), cnt as u32, 0) };
+    if n >= 0 {
+        return Ok(());
+    }
+    let e = io::Error::last_os_error();
+    match e.kind() {
+        io::ErrorKind::Interrupted => {
+            // Retry the transmission
+        }
```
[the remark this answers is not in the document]
mxinden: Agreed. @larseggert I will follow-up with a pull request.

mxinden:
For those interested in using the `fast-apple-datapath` feature:
- All MacOS Firefox Nightly users [are now using it](https://phabricator.services.mozilla.com/D234901).
- We see [~11% performance improvement](https://github.com/mozilla/neqo/pull/2302) on our localhost CPU bound throughput benchmark.
- Every other (50th percentile) QUIC UDP `recvmsg_x` call reads 2 datagrams.


image

https://glam.telemetry.mozilla.org/fog/probe/networking_http_3_udp_datagram_segments_received/explore?aggType=avg&os=Darwin&timeHorizon=WEEK&visiblePercentiles=%5B99.9%2C99%2C95%2C75%2C50%2C25%2C5%5D

PR #1995, opened 2024-09-24
Expose orignial dest cid

thynson:
Expose the original dest cid, which is generated by client when connecting.

The use case is described in #1896, and it would be convenient to expose it for logging and e2e testing.


Ralith (approved):
Seems reasonable.

djc (approved):

PR #1996, opened 2024-09-26
feat(quinn-proto): Make compilable in wasm32-unknown-unknown and run wasm tests in CI

matheus23:
### Summary of Changes

- Use `web_time` dependency instead of `std::time` & re-export `Instant`, `Duration`, `SystemTime` and `UNIX_EPOCH` in `lib.rs` so imports are simpler
- Adjust imports to use `crate::{Instant, Duration, ...}` instead of `std::time::{Instant, Duration, ...}`
- Added `wasm-bindgen-test` to enable running tests from a Wasm build
- Added a CI task to test quinn-proto tests in Wasm with node.js

### Running `quinn-proto` tests in Wasm

```sh
# install wasm-bindgen-test-runner
$ cargo install wasm-bindgen-cli
# run tests
$ cargo test -p quinn-proto --target wasm32-unknown-unknown
```

<details>
<summary>Old open questions, now resolved</summary>

- Should the added code/dependencies additionally be feature-gated?
  
  I personally don't think so. As far as I know, there's practically no one asking for wasm32-unknown-unknown compilability who isn't targeting nodejs/browsers/cloudflare workers/etc. and in all of these cases wasm-bindgen is the de-facto standard.
  There's also targeting wasmtime and other runtimes, but IMO those needs are met by `wasm32-wasi-*` targets, not `wasm32-unknown-unknown`.
  
  That said, for my intents and purposes that's mostly a cosmetic question, I don't mind having to enable another feature flag. Other concerns could be "what non-web `wasm32-unknown-unknown` becomes a thing and we need to break compatibility by introducing features?".
- Should the `Duration`, `Instant`, etc. re-exports live in `lib.rs`?
  
  They're `pub(crate)` only, so I don't think it matters much. I can move them into a file if you want, or we keep them here.
- Should we just depend on `web_time` instead of `std::time` everywhere instead of cfg-gating & re-exporting?
  
  `web_time` itself already multiplexes between `std::time` and a web-based implementation. So in a way, we're doing it twice here. What we gain is not depending on `web_time` most of the time.
  However, once we work on Wasm support for e.g. `quinn` itself, we'd need to do this whole re-export dance, again. So perhaps it'd be better to just depend on `web_time` everywhere?
  
</details>

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -69,6 +69,50 @@ jobs:
       - run: RUST_BACKTRACE=1 cargo test --manifest-path quinn-proto/Cargo.toml --no-default-features --features rustls-aws-lc-rs-fips
       - run: RUST_BACKTRACE=1 cargo test --manifest-path quinn/Cargo.toml --no-default-features --features rustls-aws-lc-rs-fips,runtime-tokio
 
+  wasm_test:
+    name: test wasm32-unknown-unknown
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Add wasm target
+        run: rustup target add wasm32-unknown-unknown
+
+      - name: Install nodejs v20
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Setup `wasm-tools`
+        uses: bytecodealliance/actions/wasm-tools/setup@v1
+
+      - name: Install cargo binstall
+        uses: cargo-bins/cargo-binstall@main
+      
+      # We need to downgrade cc to version 1.1.31 for ring Wasm compilation to work.
+      # See the upstream issue https://github.com/rust-lang/cc-rs/issues/1275
+      - name: Use working `cc` version 1.1.31
+        run: cargo update -p cc --precise 1.1.31
+
+      - name: build wasm32 tests (quinn-proto)
+        run: cargo test -p quinn-proto --target wasm32-unknown-unknown --no-run
+
+      # If the Wasm file contains any 'import "env"' declarations, then
+      # some non-Wasm-compatible code made it into the final code.
+      - name: Check for 'import "env"' in Wasm
+        run: |
+          ! wasm-tools print --skeleton target/wasm32-unknown-unknown/debug/deps/quinn_proto-*.wasm | grep 'import "env"'
+
+      - name: Install wasm-bindgen-test-runner
+        run: cargo binstall wasm-bindgen-cli --locked --no-confirm
+
+      - name: wasm32 test (quinn-proto)
+        run: cargo test -p quinn-proto --target wasm32-unknown-unknown
+
   msrv:
     runs-on: ubuntu-latest
     steps:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ clap = { version = "4", features = ["derive"] }
 crc = "3"
 directories-next = "2"
 futures-io = "0.3.19"
+getrandom = { version = "0.2", default-features = false }
 hdrhistogram = { version = "7.2", default-features = false }
 hex-literal = "0.4"
 lazy_static = "1"
@@ -37,6 +38,7 @@ rustc-hash = "2"
 rustls = { version = "0.23.5", default-features = false, features = ["std"] }
 rustls-pemfile = "2"
 rustls-platform-verifier = "0.4"
+rustls-pki-types = "1.7"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1"
 slab = "0.4.6"
@@ -49,6 +51,8 @@ tracing = { version = "0.1.10", default-features = false, features = ["std"] }
 tracing-futures = { version = "0.2.0", default-features = false, features = ["std-future"] }
 tracing-subscriber = { version = "0.3.0", default-features = false, features = ["env-filter", "fmt", "ansi", "time", "local-time"] }
 url = "2"
+wasm-bindgen-test = { version = "0.3.45" }
+web-time = "1"
 windows-sys = { version = ">=0.52, <=0.59", features = ["Win32_Foundation", "Win32_System_IO", "Win32_Networking_WinSock"] }
 
 [profile.bench]
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -45,12 +45,21 @@ thiserror = { workspace = true }
 tinyvec = { workspace = true, features = ["alloc"] }
 tracing = { workspace = true }
 
+# Feature flags & dependencies for wasm
+# wasm-bindgen is assumed for a wasm*-*-unknown target
+[target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dependencies]
+ring = { workspace = true, features = ["wasm32_unknown_unknown_js"] }
+getrandom = { workspace = true, features = ["js"] }
+rustls-pki-types = { workspace = true, features = ["web"] } # only added as dependency to enforce the `web` feature for this target
+web-time = { workspace = true }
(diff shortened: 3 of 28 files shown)


Cargo.toml
```diff
@@ -50,6 +50,9 @@ tracing-futures = { version = "0.2.0", default-features = false, features = ["st
 tracing-subscriber = { version = "0.3.0", default-features = false, features = ["env-filter", "fmt", "ansi", "time", "local-time"] }
 url = "2"
 windows-sys = { version = "0.59", features = ["Win32_Foundation", "Win32_System_IO", "Win32_Networking_WinSock"] }
+getrandom = { version = "0.2", default-features = false }
```
djc, line 53: Please keep the dependencies in alphabetic order. Also, it doesn't look like `rustls-pki-types` actually gets used here? And also please simplify `{ version = "x" }` to just `"x"`.

quinn-proto/Cargo.toml
```diff
@@ -43,6 +43,14 @@ thiserror = { workspace = true }
 tinyvec = { workspace = true, features = ["alloc"] }
 tracing = { workspace = true }
 
+# Feature flags & dependencies for wasm
+# wasm-bindgen is assumed for a wasm*-*-unknown target
+[target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dependencies]
+ring = { workspace = true, features = ["wasm32_unknown_unknown_js"] }
+getrandom = { workspace = true, features = ["js"] }
+rustls-pki-types = { workspace = true, features = ["web"] }
```
djc, line 51: Please add a comment here that we're only adding an edge for rustls-pki-types to activate the `web` feature.

.github/workflows/rust.yml
```diff
@@ -65,6 +65,33 @@ jobs:
...
+        uses: actions/checkout@v4
+
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Add wasm target
+        run: rustup target add wasm32-unknown-unknown
+
+      - name: wasm32 build (quinn)
+        run: cargo build -p iroh-quinn --no-default-features --features log,ring,rustls,wasm,runtime-wasm --target wasm32-unknown-unknown
```
djc: Pretty sure these crates names aren't quite right...

.github/workflows/rust.yml
```diff
@@ -65,8 +65,8 @@ jobs:
       - run: RUST_BACKTRACE=1 cargo test --manifest-path quinn-proto/Cargo.toml --no-default-features --features rustls-aws-lc-rs
       - run: RUST_BACKTRACE=1 cargo test --manifest-path quinn/Cargo.toml --no-default-features --features rustls-aws-lc-rs,runtime-tokio
 
-  wasm_build:
-    name: Build wasm32
+  wasm_test:
```
djc, line 68: I think we should just squash these changes into the first commit?

djc (commented):
Please squash all the changes into a single commit, as these commits are not conceptually separate.

quinn-proto/src/tests/mod.rs
```diff
@@ -27,11 +26,21 @@ use crate::{
...
-#[test]
+#[cfg(all(target_family = "wasm", target_os = "unknown"))]
+use wasm_bindgen_test::wasm_bindgen_test;
+
+// Enable this if you want to run these tests in the browser.
+// Unfortunately it's either-or: Enable this and you can run in the browser, disable to run in nodejs.
+// #[cfg(all(target_family = "wasm", target_os = "unknown"))]
+// wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
+
+#[cfg_attr(all(target_family = "wasm", target_os = "unknown"), wasm_bindgen_test)]
```
djc, line 42: I think we should find a way to avoid repeating these on every test.

djc (commented):
This is looking pretty good!

quinn-proto/src/tests/mod.rs
matheus23: Unfortunately I don't know of a good way :/
What works:
- Writing our own proc macro
- Writing our own declarative macro (I think?)
- Splitting the test into 3 modules: one with `pub fn`s, one with `#[test]`s calling the `pub fn`s, and one with `#[wasm_bindgen_test]`s running the `pub fn`s.

I don't like either of these options. Would love some good ideas.

What doesn't work:
- Use only `#[wasm_bindgen_test]` (it doesn't resolve to just `#[test]` on other platforms, which is a shame IMO)

djc: > * Use only `#[wasm_bindgen_test]` (it doesn't resolve to just `#[test]` on other platforms, which is a shame IMO)

Have you investigated whether there is an open issue or some discussion of this in the upstream repo? It seems pretty straightforward...

matheus23: Yeah, I looked, there was nothing. Anyhow, [I opened an issue](https://github.com/rustwasm/wasm-bindgen/issues/4139).
Do you think we should block this PR on that issue & a subsequent release of wasm-bindgen-test?

djc: Let's see if there's any movement on your issue until Monday or something? Looks like at least one of the active maintainers in that repo is also a Quinn contributor (👋 @daxpedda), so let's see what happens.

daxpedda: Thank you for the ping djc!
https://github.com/rustwasm/wasm-bindgen/pull/4150, part of the latest release, should resolve this issue.
If not please let me know!

matheus23:
Thanks for the review. Thoughts on this open question?

> Should we just depend on web_time instead of std::time everywhere instead of cfg-gating & re-exporting?

I'm on the fence. I think it may be better to depend on `web_time` on all targets. Otherwise I'll be repeating the same `lib.rs` changes in `quinn` itself (to avoid having `quinn_proto::Duration` be a public reexport).

matheus23:
@djc @Ralith what's left to do here to get this over the finish line?
I'd love to also contribute `quinn` wasm32-unknown-unknown support after this.

djc (approved):
LGTM, please squash the commits?

Ralith:
> I'm inclined to stick with depending on web_time only for WASM targets. IMO mainstream platforms should not bear the cost of niche platforms even if the overhead is small. @Ralith any thoughts?

Given that the amount of extra code is pretty trivial, this stance seems reasonable to me. I know I'd be a bit confused as a non-web user to see a web crate show up in my environment.

Ralith (approved):
The `wasm_bindgen_test` churn is unfortunate, but I guess there's nothing for it.

djc:
@matheus23 this is currently failing the WASM test, I guess that's not supposed to be intentional?

matheus23:
Seems like there's potentially some ring issues with wasm32.
Investigating this. Also need to make sure that `main` doesn't have any commits added that regress wasm32 support (So I need to merge in main continuously).

matheus23:
Okay, got to somewhat of the bottom on this. The stacktraces you saw were v8 crashing for some reason.
Updating nodejs from 18 to 20 fixes the issue.

matheus23:
Also seems like `cc` version 1.1.32 broke the ring Wasm build (but in github actions only?).
I've submitted an [upstream issue](https://github.com/rust-lang/cc-rs/issues/1275) and forced version 1.1.31 in CI for now.

quinn-proto/src/tests/mod.rs
matheus23: Not sure why I didn't think of this, but this works, too, at least in our case where we don't need async tests:
```rs
#[cfg(all(target_family = "wasm", target_os = "unknown"))]
use wasm_bindgen_test::wasm_bindgen_test as test;
```

I've changed the PR to use that. The diff is much smaller this way.

matheus23:
FWIW, this PR is ready from my perspective. Is there anything left to do? Do you want to re-review?

djc (approved):

djc:
Thanks, this is looking good!

PR #1999, opened 2024-09-29
Reduce memory allocations on servers that process a large number of streams

alessandrod:
The solana blockchain uses uni streams to deliver transactions. A transaction can be up to 1232 bytes, each transaction is a stream. A validator accepts thousands of connections and each connection sends millions of streams.

The Solana validator Agave uses Quinn for transaction processing, with jemalloc as the global allocator. We're encountering an issue where, due to the high volume of allocations while ingesting transactions, jemalloc struggles during network spikes. It blows out the tcache, spends excessive time managing arenas, allocating large amounts of memory, and then attempting to free it during decay. This results in a significant amount of time spent on syscalls and arena locking, leading to contention and random stalls in other threads.

Quinn is one of the largest source of allocations, so I've been working on improving things.

This is a profile of a single node validator ingesting 100k transactions with master, with a 512 streams receive window:

image
 
And this is with the PR applied:

image

As you can see with master we get 2 allocations for each stream _ingested_. With the PR we get 2 allocations for each stream _in the window_. 

The PR essentially introduces a free list of Recvs. When it's time to create a new Recv, it tries to pop one from the free list. 

This approach works great for us, but it relies on a few key assumptions: that there will be a large number of streams, the stream receive window will generally be fully utilized, and each stream will receive a similar amount of data. If these assumptions aren't met, this PR may lead to some memory inefficiency. 

I'm wondering whether you'd be interested in merging something like this, otherwise happy to keep in our fork. Also this is just a first pass, if you're interested I could try to make this behavior opt-in, or tunable somehow to limit the amount of spare memory allocated. I'm open to ideas!

### DIFF
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -29,6 +29,14 @@ impl Assembler {
         Self::default()
     }
 
+    /// Reset to the initial state
+    pub(super) fn reinit(&mut self) {
+        let old_data = mem::take(&mut self.data);
+        *self = Self::default();
+        self.data = old_data;
+        self.data.clear();
+    }
+
     pub(super) fn ensure_ordering(&mut self, ordered: bool) -> Result<(), IllegalOrderedRead> {
         if ordered && !self.state.is_ordered() {
             return Err(IllegalOrderedRead);
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -7,10 +7,11 @@ use bytes::Bytes;
 use thiserror::Error;
 use tracing::trace;
 
-use self::state::get_or_insert_recv;
-
 use super::spaces::{Retransmits, ThinRetransmits};
-use crate::{connection::streams::state::get_or_insert_send, frame, Dir, StreamId, VarInt};
+use crate::{
+    connection::streams::state::{get_or_insert_recv, get_or_insert_send},
+    frame, Dir, StreamId, VarInt,
+};
 
 mod recv;
 use recv::Recv;
@@ -149,8 +150,8 @@ impl<'a> RecvStream<'a> {
         // connection-level flow control to account for discarded data. Otherwise, we can discard
         // state immediately.
         if !stream.final_offset_unknown() {
-            entry.remove();
-            self.state.stream_freed(self.id, StreamHalf::Recv);
+            let recv = entry.remove().expect("must have recv when stopping");
+            self.state.stream_recv_freed(self.id, recv);
         }
 
         if self.state.add_read_credits(read_credits).should_transmit() {
@@ -168,7 +169,7 @@ impl<'a> RecvStream<'a> {
         let hash_map::Entry::Occupied(entry) = self.state.recv.entry(self.id) else {
             return Err(ClosedStream { _private: () });
         };
-        let Some(s) = entry.get().as_ref() else {
+        let Some(s) = entry.get().as_ref().and_then(|s| s.as_open_recv()) else {
             return Ok(None);
         };
         if s.stopped {
@@ -180,8 +181,9 @@ impl<'a> RecvStream<'a> {
 
         // Clean up state after application observes the reset, since there's no reason for the
         // application to attempt to read or stop the stream once it knows it's reset
-        entry.remove_entry();
-        self.state.stream_freed(self.id, StreamHalf::Recv);
+        let (_, recv) = entry.remove_entry();
+        self.state
+            .stream_recv_freed(self.id, recv.expect("must have recv on reset"));
         self.state.queue_max_stream_id(self.pending);
 
         Ok(Some(code))
--- a/quinn-proto/src/connection/streams/recv.rs
+++ b/quinn-proto/src/connection/streams/recv.rs
@@ -5,12 +5,14 @@ use thiserror::Error;
 use tracing::debug;
 
 use super::state::get_or_insert_recv;
-use super::{ClosedStream, Retransmits, ShouldTransmit, StreamHalf, StreamId, StreamsState};
+use super::{ClosedStream, Retransmits, ShouldTransmit, StreamId, StreamsState};
 use crate::connection::assembler::{Assembler, Chunk, IllegalOrderedRead};
+use crate::connection::streams::state::StreamRecv;
 use crate::{frame, TransportError, VarInt};
 
 #[derive(Debug, Default)]
 pub(super) struct Recv {
+    // NB: when adding or removing fields, remember to update `reinit`.
     state: RecvState,
     pub(super) assembler: Assembler,
     sent_max_stream_data: u64,
@@ -27,6 +29,15 @@ impl Recv {
             end: 0,
             stopped: false,
         })
+    }
+
+    /// Reset to the initial state
+    pub(super) fn reinit(&mut self, initial_max_data: u64) {
+        self.state = RecvState::default();
+        self.assembler.reinit();
+        self.sent_max_stream_data = initial_max_data;
+        self.end = 0;
+        self.stopped = false;
     }
 
     /// Process a STREAM frame
@@ -255,7 +266,7 @@ impl<'a> Chunks<'a> {
         let mut recv =
             match get_or_insert_recv(streams.stream_receive_window)(entry.get_mut()).stopped {
                 true => return Err(ReadableError::ClosedStream),
-                false => entry.remove().unwrap(), // this can't fail due to the previous get_or_insert_with
+                false => entry.remove().unwrap().into_inner(), // this can't fail due to the previous get_or_insert_with
             };
 
         recv.assembler.ensure_ordering(ordered)?;
@@ -292,14 +303,24 @@ impl<'a> Chunks<'a> {
         match rs.state {
             RecvState::ResetRecvd { error_code, .. } => {
                 debug_assert_eq!(self.read, 0, "reset streams have empty buffers");
-                self.streams.stream_freed(self.id, StreamHalf::Recv);
-                self.state = ChunksState::Reset(error_code);
+                let state = mem::replace(&mut self.state, ChunksState::Reset(error_code));
+                // At this point if we have `rs` self.state must be `ChunksState::Readable`
+                let recv = match state {
+                    ChunksState::Readable(recv) => StreamRecv::Open(recv),
+                    _ => unreachable!("state must be ChunkState::Readable"),
+                };
+                self.streams.stream_recv_freed(self.id, recv);
                 Err(ReadError::Reset(error_code))
             }
             RecvState::Recv { size } => {
                 if size == Some(rs.end) && rs.assembler.bytes_read() == rs.end {
-                    self.streams.stream_freed(self.id, StreamHalf::Recv);
-                    self.state = ChunksState::Finished;
+                    let state = mem::replace(&mut self.state, ChunksState::Finished);
+                    // At this point if we have `rs` self.state must be `ChunksState::Readable`
+                    let recv = match state {
+                        ChunksState::Readable(recv) => StreamRecv::Open(recv),
+                        _ => unreachable!("state must be ChunkState::Readable"),
+                    };
+                    self.streams.stream_recv_freed(self.id, recv);
                     Ok(None)
                 } else {
                     // We don't need a distinct `ChunksState` variant for a blocked stream because
@@ -345,7 +366,9 @@ impl<'a> Chunks<'a> {
                 self.pending.max_stream_data.insert(self.id);
             }
             // Return the stream to storage for future use
-            self.streams.recv.insert(self.id, Some(rs));
+            self.streams
+                .recv
+                .insert(self.id, Some(StreamRecv::Open(rs)));
         }
 
         // Issue connection-level flow control credit for any data we read regardless of state
--- a/quinn-proto/src/connection/streams/state.rs
+++ b/quinn-proto/src/connection/streams/state.rs
@@ -20,12 +20,58 @@ use crate::{
     Dir, Side, StreamId, TransportError, VarInt, MAX_STREAM_COUNT,
 };
 
+/// Wrapper around `Recv` that facilitates reusing `Recv` instances
+#[derive(Debug)]
+pub(super) enum StreamRecv {
+    /// A `Recv` that is ready to be opened
+    Free(Box<Recv>),
+    /// A `Recv` that has been opened
+    Open(Box<Recv>),
+}
+
+impl StreamRecv {
+    /// Returns a reference to the inner `Recv` if the stream is open
+    pub(super) fn as_open_recv(&self) -> Option<&Recv> {
+        match self {
+            Self::Open(r) => Some(r),
+            _ => None,
+        }
+    }
+
+    // Returns a mutable reference to the inner `Recv` if the stream is open
+    pub(super) fn as_open_recv_mut(&mut self) -> Option<&mut Recv> {
+        match self {
+            Self::Open(r) => Some(r),
+            _ => None,
+        }
+    }
+
+    // Returns the inner `Recv`
+    pub(super) fn into_inner(self) -> Box<Recv> {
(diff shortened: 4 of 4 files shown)


quinn-proto/src/connection/assembler.rs
```diff
@@ -29,6 +29,16 @@ impl Assembler {
         Self::default()
     }
 
+    /// Reset to the initial state
+    pub(super) fn reinit(&mut self) {
```
alessandrod, line 33: I'm not super happy with `reinit`. `reset` is already used for other purposes. I had `recycle` at some point but I like that even less

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -180,8 +181,9 @@ impl<'a> RecvStream<'a> {
 
         // Clean up state after application observes the reset, since there's no reason for the
         // application to attempt to read or stop the stream once it knows it's reset
-        entry.remove_entry();
-        self.state.stream_freed(self.id, StreamHalf::Recv);
+        let (_, recv) = entry.remove_entry();
```
alessandrod, line 184: I think that this path isn't covered by tests

quinn-proto/src/connection/streams/state.rs
```diff
@@ -20,12 +20,58 @@ use crate::{
...
+}
+
 #[allow(unreachable_pub)] // fuzzing only
 pub struct StreamsState {
     pub(super) side: Side,
     // Set of streams that are currently open, or could be immediately opened by the peer
     pub(super) send: FxHashMap<StreamId, Option<Box<Send>>>,
-    pub(super) recv: FxHashMap<StreamId, Option<Box<Recv>>>,
+    pub(super) recv: FxHashMap<StreamId, Option<StreamRecv>>,
+    pub(super) free_recv: Vec<StreamRecv>,
```
alessandrod, line 74: Having `free_recv` allows me to keep most of the internal API as is and grab spare streams in StreamState::insert. If you like the idea behind this PR, I'd actually want to get rid of free_recv (an extra allocation :P), and plumb streams directly into `ensure_remote_streams()` somehow

djc:
Haven't looked closely at your changes yet but conceptually the idea of having some kind of free list seems conceptually sound. If we want to merge this I think we should probably come up with some heuristics that make this an improvement in a majority of downstream users (in the default configuration), though we could presumably add some configuration knobs.

alessandrod:
Thanks for the super quick feedback, definitely appreciated!

> Haven't looked closely at your changes yet but conceptually the idea of having some kind of free list seems conceptually sound. If we want to merge this I think we should probably come up with some heuristics that make this an improvement in a majority of downstream users (in the default configuration), though we could presumably add some configuration knobs.

Sweet! Haven't thought about heuristics yet, but things that we might need to tune are:

- num of recvs in the freelist: is there a reason to ever want this lower than max_*_streams?
- num of elements in the Assembler BinaryHeap: I guess the larger the stream, the more elements it'll contain. We might need to shrink $at_some_point

quinn-proto/src/connection/assembler.rs
```diff
@@ -29,6 +29,16 @@ impl Assembler {
...
+        self.state = State::default();
+        self.data.clear();
+        self.buffered = 0;
+        self.allocated = 0;
+        self.bytes_read = 0;
+        self.end = 0;
```
Ralith, line 39: Could this be simplified to something like the below?
```rs
let old = mem::take(self);
self.data = old.data;
self.data.clear();
```
An empty `BinaryHeap` shouldn't allocate, and will likely get optimized out entirely. That way we avoid the hazard of forgetting to update this with new fields.

quinn-proto/src/connection/streams/recv.rs
```diff
@@ -29,6 +30,15 @@ impl Recv {
         })
     }
 
+    /// Reset to the initial state
+    pub(super) fn reinit(&mut self, initial_max_data: u64) {
+        self.state = RecvState::default();
+        self.assembler.reinit();
+        self.sent_max_stream_data = initial_max_data;
+        self.end = 0;
+        self.stopped = false;
```
Ralith, line 39: As above.

quinn-proto/src/connection/assembler.rs
Ralith: Maybe `clear`? That seems logically equivalent. I don't mind `reinit`, though.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -20,12 +20,58 @@ use crate::{
...
+    pub(super) fn into_inner(self) -> Box<Recv> {
+        match self {
+            Self::Ready(r) | Self::Open(r) => r,
+        }
+    }
+
+    // Reinitialize the stream so the inner `Recv` can be reused
+    pub(super) fn reinit(self, initial_max_data: u64) -> Self {
+        match self {
+            Self::Ready(_) => unreachable!("Self::reset on Ready"),
```
Ralith, line 59: Outdated text

quinn-proto/src/connection/streams/state.rs
Ralith: > plumb streams directly into ensure_remote_streams() somehow

Be mindful that `Recv` streams may also be locally-initiated bidirectional streams.

> an extra allocation :P

A (more or less) one-time allocation doesn't matter much, though I concur that expanding on the existing pattern of pre-allocated entries in `send`/`recv` seems graceful.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -20,12 +20,58 @@ use crate::{
     Dir, Side, StreamId, TransportError, VarInt, MAX_STREAM_COUNT,
 };
 
+/// Wrapper around `Recv` that facilitates reusing `Recv` instances
+#[derive(Debug)]
+pub(super) enum StreamRecv {
+    /// A `Recv` that is ready to be opened
+    Ready(Box<Recv>),
```
Ralith, line 27: bikeshed: maybe `Free` for clarity?

quinn-proto/src/connection/streams/state.rs
```diff
@@ -240,8 +287,12 @@ impl StreamsState {
 
         // Stopped streams become closed instantly on FIN, so check whether we need to clean up
         if closed {
-            self.recv.remove(&id);
-            self.stream_freed(id, StreamHalf::Recv);
+            let recv = self
+                .recv
+                .remove(&id)
+                .flatten()
+                .expect("must have recv when closing");
```
Ralith, line 294: Feel free to use a plain `unwrap` when it's provably impossible to reach.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -240,8 +287,12 @@ impl StreamsState {
 
         // Stopped streams become closed instantly on FIN, so check whether we need to clean up
         if closed {
-            self.recv.remove(&id);
-            self.stream_freed(id, StreamHalf::Recv);
+            let recv = self
```
Ralith, line 290: Maybe `let rs` for clarity that this is the same value, just owned rather than borrowed?

quinn-proto/src/connection/streams/state.rs
```diff
@@ -20,12 +20,58 @@ use crate::{
...
+pub(super) enum StreamRecv {
+    /// A `Recv` that is ready to be opened
+    Ready(Box<Recv>),
+    /// A `Recv` that has been opened
+    Open(Box<Recv>),
+}
+
+impl StreamRecv {
+    /// Returns a reference to the inner `Recv` if the stream is open
+    pub(super) fn open_recv(&self) -> Option<&Recv> {
```
Ralith, line 34: "open" is a verb, so this is a bit confusing. Maybe `as_open_recv`?

quinn-proto/src/connection/streams/state.rs
```diff
@@ -899,8 +966,16 @@ pub(super) fn get_or_insert_send(
...
-    move |opt| opt.get_or_insert_with(|| Recv::new(initial_max_data))
+) -> impl FnMut(&mut Option<StreamRecv>) -> &mut Recv {
+    move |opt| {
+        *opt = opt.take().map(|s| match s {
+            StreamRecv::Ready(recv) => StreamRecv::Open(recv),
+            s => s,
+        });
+        opt.get_or_insert_with(|| StreamRecv::Open(Recv::new(initial_max_data)))
+            .open_recv_mut()
+            .unwrap()
```
Ralith, line 977: Fold these together? Maybe `opt.get_or_insert_with(...).map(|x| x.ensure_open())`.

Ralith:
I'm happy to improve our performance for this use case, because numerous disposable streams are a major QUIC design pattern, and one I use personally in downstream projects besides.

I think the downside of this change is modest. Connections which use a consistent proportion of their stream window (whether that's all, none, or in between) will receive benefits without drawbacks. Only for connections which usually leave their stream window mostly empty but on rare occasion fill it will we see an increase in average memory use by the pooled `Recv`s, and that memory use should still be much smaller than the worst-case memory when the stream window is actually full of streams which are individually full of unread stream data. I'd therefore prefer to avoid introducing new configuration unless someone has a concrete use case that we can't cover automatically.

I'm interested in input from @jeromegn, responsible for previous work to reduce average memory use due to `Recv`s in https://github.com/quinn-rs/quinn/pull/1682.

> It blows out the tcache, spends excessive time managing arenas, allocating large amounts of memory, and then attempting to free it during decay. This results in a significant amount of time spent on syscalls and arena locking, leading to contention and random stalls in other threads.

How much of an impact do you see to these factors? The profile excerpts you showed demonstrate unsurprisingly that this change reduces allocator load from the affected code, but it's not obvious to me what proportion of Quinn's allocator load this accounts for. We should still have at least one allocation per incoming datagram for the datagram itself, for example.

quinn-proto/src/connection/streams/state.rs
alessandrod: Because leaders rotate very frequently, in solana every validator gets a huge spike in connections right before its leader slots. After I'm done minimizing per-stream allocations, I'll want to minimize per-connection allocations too, although it's still unclear how upstreamable that part will be.

alessandrod:
> > It blows out the tcache, spends excessive time managing arenas, allocating large amounts of memory, and then attempting to free it during decay. This results in a significant amount of time spent on syscalls and arena locking, leading to contention and random stalls in other threads.
> 
> How much of an impact do you see to these factors? The profile excerpts you showed demonstrate unsurprisingly that this change reduces allocator load from the affected code, but it's not obvious to me what proportion of Quinn's allocator load this accounts for. We should still have at least one allocation per incoming datagram for the datagram itself, for example.

In the screenshot you can see that the binary heap allocation is 44% of all allocations (blocks) made by quinn, and the Recv box is another 44%. Transactions can be _up to_ 1232 bytes, but many are much smaller, therefore you get a few txs in a single datagram.

The datagram allocation you're referring to, in that particular run was 5%:

Screenshot 2024-09-30 at 6 36 01 pm

I'm going to send a PR for this one too :)

quinn-proto/src/connection/streams/state.rs
alessandrod: done

quinn-proto/src/connection/streams/state.rs
alessandrod: done

quinn-proto/src/connection/streams/state.rs
alessandrod: fixed

quinn-proto/src/connection/assembler.rs
alessandrod: Assembler already has a clear(), and it doesn't clear all the fields

quinn-proto/src/connection/streams/recv.rs
alessandrod: If I do this, in the generated code in --release mode I do see the old value being copied on the stack, and then assembler copied back into self. Since this is in a relatively hot path for us, I'd rather not have that. But also given that we are in micro-opt territory, happy to do it if you feel strongly about it.

quinn-proto/src/connection/streams/state.rs
alessandrod: I don't think I can do what you're suggesting: ensure_open() would need to take by value to go from Free to Open, but get_or_insert_with returns a reference. And the closure takes `&mut Option<StreamRecv>` which is why I have the weird take().map()

quinn-proto/src/connection/streams/recv.rs
Ralith: Maybe a slightly different construct would work better? e.g.:
```rs
let old_data = mem:;take(&mut self.data);
*self = Self::default();
self.data = old_data;
self.data.clear();
```

Ralith:
> In the screenshot you can see that the binary heap allocation is 44% of all allocations (blocks) made by quinn, and the Recv box is another 44%.

Wow, awesome find!

> I'm going to send a PR for [reducing datagram allocations] too :)

While very welcome, that may be a harder problem, since ownership of datagram memory is extremely dynamic: it may move between threads, hang around until application data is consumed, etc. There might be some low hanging fruit for handling frames that don't carry application data without allocating, at least.

quinn-proto/src/connection/streams/state.rs
Ralith: Ah, you're right.

quinn-proto/src/connection/streams/recv.rs
alessandrod: for Assembler::reinit, this works. For Recv::reinit, we still have the same issue that it copies Assembler twice, and since Assembler is relatively large (88 bytes), that takes a few insns.

Ralith: Oh well, I don't feel that strongly. Maybe leave a comment inside the struct definition as a reminder to update `reinit` when adding or modifying fields.

quinn-proto/src/connection/assembler.rs
Ralith: Then `reinit` seems fine.

quinn-proto/src/connection/streams/recv.rs
alessandrod: done

quinn-proto/src/connection/assembler.rs
alessandrod: done

alessandrod:
I think I've addressed the nits. If it's ok for you, I'd like to remove the free_recv alloc (https://github.com/quinn-rs/quinn/pull/1999#discussion_r1780095798) in a follow up PR. I did just try to do it, and it's more work than expected, and I'm working on another PR that is more important for us (reducing the Bytes allocations).

Ralith:
> I'd like to remove the free_recv alloc (https://github.com/quinn-rs/quinn/pull/1999#discussion_r1780095798) in a follow up PR

SGTM, I expect that to take significant additional review anyway.

quinn-proto/src/connection/streams/mod.rs
Ralith: Improved coverage would be welcome, but doesn't need to be part of the PR, I think.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -20,12 +20,58 @@ use crate::{
...
+
+    // Returns the inner `Recv`
+    pub(super) fn into_inner(self) -> Box<Recv> {
+        match self {
+            Self::Free(r) | Self::Open(r) => r,
+        }
+    }
+
+    // Reinitialize the stream so the inner `Recv` can be reused
+    pub(super) fn reinit(self, initial_max_data: u64) -> Self {
```
Ralith, line 57: bikeshed: maybe `fn free`?

Ralith (approved):

quinn-proto/src/connection/streams/state.rs
alessandrod: done

djc (approved):

djc:
Thanks!

PR #2000, opened 2024-10-02
Add visionOS support

rofferom:
Just enable the same code blocks than on iOS.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -1,6 +1,7 @@
 #[cfg(not(any(
     target_os = "macos",
     target_os = "ios",
+    target_os = "tvos",
     target_os = "visionos",
     target_os = "openbsd",
     target_os = "solaris",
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Hmm, I wonder if there is something better we can do, like treating all the Darwin stuff the same?

rofferom:
Yes it would be great, especially because tvos support is missing too. I haven't tested it yet, that's why it isn't included in the MR.

djc:
Do you know the right incantation for supporting everything Darwin at once?

rofferom:
I have seen this one [`#[cfg(target_vendor = "apple")]`](https://github.com/rust-lang/rust/blob/1.81.0/library/std/src/sys/pal/unix/time.rs#L89). Other projects are using the same strategy than Quinn (ring, socket2).

Actually, I don't know the possible limitations of `#[cfg(target_vendor = "apple")]`.

djc:
`#[cfg(target_vendor = "apple")]` looks about right to me. @Ralith what do you think?

rofferom:
I have just added tvOS support with the current way to declare it. Waiting for Ralith's feedback to update the MR if required.

djc:
Curious to hear what you're building!

Ralith (approved):

Ralith:
> #[cfg(target_vendor = "apple")] looks about right to me (I noticed std/core use it in a few places). @Ralith what do you think?

That seems very appealing. I don't know how we'd verify that it's exactly the semantics we'd want short of digging into rustc, but it's hard to imagine it's too far off. Maybe switch to it in a separate commit so we can easily revert if needed?

PR #2002, opened 2024-10-07
Add alternative - unfair but lower latency - send stream scheduling strategy

alessandrod:
Solana leader slots are 400ms. When sending and ingesting transactions, minimizing latency is therefore key. Quinn currently tries to implement fairness when writing out send streams, which is a good default, but not great for our use case.

We want to pack as many transactions in a datagram as possible, without any fragmentation if not at the very end if a transaction doesn't fit in the remaining space. In that case, we want the end (fin) of the transaction to come _immediately after_ in order to minimize latency. The current round-robin algorithm doesn't allow this, and in fact leads to very high latency if the stream receive window is large.  

This PR tries to address this problem. It introduces a `TransportConfig::send_fairness(bool)` config. When set to false, streams are still scheduled based on priority, but once a chunk of a stream has been written out, we'll try to complete the stream instead of trying to round-robin balance it among the streams with the same priority.

This gets rid of fragmentation, and effectively allows API clients to precisely control the order in which streams are written out. 

Here's a server log without the patch:

```
[2024-10-07T13:51:43.960817939Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 7460 offset=0 len=255 fin=true
[2024-10-07T13:51:43.960824640Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T13:51:43.960829630Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 7461 offset=0 len=255 fin=true
[2024-10-07T13:51:43.960836460Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T13:51:43.960841420Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 7462 offset=0 len=110 fin=false
[2024-10-07T13:51:43.960849560Z TRACE quinn_proto::connection] got Data packet (1452 bytes) from 127.0.0.1:8014 using id f7b061311008438f
[2024-10-07T13:51:43.960859210Z TRACE quinn_proto::connection] recv; space=Data pn=1258
[2024-10-07T13:51:43.960865050Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T13:51:43.960869920Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 7463 offset=0 len=255 fin=true

[snip]

[2024-10-07T13:51:44.352021247Z TRACE quinn_proto::connection] recv; space=Data pn=7175
[2024-10-07T13:51:44.352024037Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T13:51:44.352026447Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 7420 offset=110 len=145 fin=true
[2024-10-07T13:51:44.352029877Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T13:51:44.352032267Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 7426 offset=110 len=145 fin=true
[2024-10-07T13:51:44.352036647Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T13:51:44.352039017Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 7432 offset=110 len=145 fin=true
[2024-10-07T13:51:44.352042398Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T13:51:44.352044778Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 7438 offset=110 len=145 fin=true
[2024-10-07T13:51:44.352048378Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T13:51:44.352050758Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 7444 offset=110 len=145 fin=true
[2024-10-07T13:51:44.352055058Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T13:51:44.352057518Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 7450 offset=110 len=145 fin=true
[2024-10-07T13:51:44.352061078Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T13:51:44.352063438Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 7456 offset=110 len=145 fin=true
[2024-10-07T13:51:44.352067018Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T13:51:44.352069528Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 7462 offset=110 len=145 fin=true
[2024-10-07T13:51:44.352073708Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T13:51:44.352076208Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 7468 offset=110 len=145 fin=true
[2024-10-07T13:51:44.352079558Z TRACE quinn_proto::connection] frame; ty=STREAM
```

Take a look at stream `7462`. It starts at `2024-10-07T13:51:43.960` `(pn=1257)`, and it's completed at `2024-10-07T13:51:44.352` `(pn=7175)` together with a bunch of other segmented transactions (note this is on localhost, so over the internet would be even worse). 

Here's a log with the PR instead:

```
2024-10-07T14:06:53.171353122Z TRACE quinn_proto::connection] recv; space=Data pn=1894
[2024-10-07T14:06:53.171356272Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T14:06:53.171358852Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 10179 offset=0 len=255 fin=true
[2024-10-07T14:06:53.171362262Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T14:06:53.171365652Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 10180 offset=0 len=255 fin=true
[2024-10-07T14:06:53.171369062Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T14:06:53.171371622Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 10181 offset=0 len=255 fin=true
[2024-10-07T14:06:53.171375192Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T14:06:53.171378072Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 10182 offset=0 len=255 fin=true
[2024-10-07T14:06:53.171382332Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T14:06:53.171384892Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 10183 offset=0 len=255 fin=true
[2024-10-07T14:06:53.171390133Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T14:06:53.171392773Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 10184 offset=0 len=110 fin=false
[2024-10-07T14:06:53.171397623Z TRACE quinn_proto::connection] got Data packet (1452 bytes) from 127.0.0.1:8004 using id 01dfdefc072f6a67
[2024-10-07T14:06:53.171401733Z TRACE quinn_proto::connection] recv; space=Data pn=1895
[2024-10-07T14:06:53.171404703Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T14:06:53.171407263Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 10184 offset=110 len=145 fin=true
[2024-10-07T14:06:53.171414553Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T14:06:53.171417323Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 10185 offset=0 len=255 fin=true
[2024-10-07T14:06:53.171420713Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T14:06:53.171423283Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 10186 offset=0 len=255 fin=true
[2024-10-07T14:06:53.171428433Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T14:06:53.171432083Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 10187 offset=0 len=255 fin=true
[2024-10-07T14:06:53.171435713Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T14:06:53.171438304Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 10188 offset=0 len=255 fin=true
[2024-10-07T14:06:53.171441674Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T14:06:53.171445274Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 10189 offset=0 len=218 fin=false
[2024-10-07T14:06:53.171449364Z TRACE quinn_proto::connection] got Data packet (1452 bytes) from 127.0.0.1:8004 using id 01dfdefc072f6a67
[2024-10-07T14:06:53.171453354Z TRACE quinn_proto::connection] recv; space=Data pn=1896
[2024-10-07T14:06:53.171456324Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T14:06:53.171460384Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 10189 offset=218 len=37 fin=true
[2024-10-07T14:06:53.171464434Z TRACE quinn_proto::connection] frame; ty=STREAM
[2024-10-07T14:06:53.171467174Z TRACE quinn_proto::connection] got stream frame id=client unidirectional stream 10190 offset=0 len=255 fin=true
```

As you can see the fin=false transactions get completed immediately in the next packet.

### DIFF
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -391,6 +391,23 @@ impl PendingStreamsQueue {
             id,
         });
     }
+
+    fn pop(&mut self) -> Option<PendingStream> {
+        self.streams.pop()
+    }
+
+    fn clear(&mut self) {
+        self.streams.clear();
+    }
+
+    fn iter(&self) -> impl Iterator<Item = &PendingStream> {
+        self.streams.iter()
+    }
+
+    #[cfg(test)]
+    fn len(&self) -> usize {
+        self.streams.len()
+    }
 }
 
 /// The [`StreamId`] of a stream with pending data queued, ordered by its priority and recency
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -365,6 +365,9 @@ impl<'a> SendStream<'a> {
 /// A queue of streams with pending outgoing data, sorted by priority
 struct PendingStreamsQueue {
     streams: BinaryHeap<PendingStream>,
+    /// The next stream to write out. This is `Some` when `TransportConfig::send_fairness(false)` and writing a stream is
+    /// interrupted while the stream still has some pending data. See `reinsert_pending()`.
+    next: Option<PendingStream>,
     /// A monotonically decreasing counter, used to implement round-robin scheduling for streams of the same priority.
     /// Underflowing is not a practical concern, as it is initialized to u64::MAX and only decremented by 1 in `push_pending`
     recency: u64,
@@ -374,12 +377,28 @@ impl PendingStreamsQueue {
     fn new() -> Self {
         Self {
             streams: BinaryHeap::new(),
+            next: None,
             recency: u64::MAX,
         }
     }
 
+    /// Reinsert a stream that was pending and still contains unsent data.
+    fn reinsert_pending(&mut self, id: StreamId, priority: i32) {
+        assert!(self.next.is_none());
+
+        self.next = Some(PendingStream {
+            priority,
+            recency: self.recency, // the value here doesn't really matter
+            id,
+        });
+    }
+
     /// Push a pending stream ID with the given priority, queued after any already-queued streams for the priority
     fn push_pending(&mut self, id: StreamId, priority: i32) {
+        // Note that in the case where fairness is disabled, if we have a reinserted stream we don't
+        // bump it even if priority > next.priority. In order to minimize fragmentation we
+        // always try to complete a stream once part of it has been written.
+
         // As the recency counter is monotonically decreasing, we know that using its value to sort this stream will queue it
         // after all other queued streams of the same priority.
         // This is enough to implement round-robin scheduling for streams that are still pending even after being handled,
@@ -393,25 +412,26 @@ impl PendingStreamsQueue {
     }
 
     fn pop(&mut self) -> Option<PendingStream> {
-        self.streams.pop()
+        self.next.take().or_else(|| self.streams.pop())
     }
 
     fn clear(&mut self) {
+        self.next = None;
         self.streams.clear();
     }
 
     fn iter(&self) -> impl Iterator<Item = &PendingStream> {
-        self.streams.iter()
+        self.next.iter().chain(self.streams.iter())
     }
 
     #[cfg(test)]
     fn len(&self) -> usize {
-        self.streams.len()
+        self.streams.len() + self.next.is_some() as usize
     }
 }
 
 /// The [`StreamId`] of a stream with pending data queued, ordered by its priority and recency
-#[derive(PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
 struct PendingStream {
     /// The priority of the stream
     // Note that this field should be kept above the `recency` field, in order for the `Ord` derive to be correct
--- a/quinn-proto/src/connection/streams/state.rs
+++ b/quinn-proto/src/connection/streams/state.rs
@@ -192,7 +192,7 @@ impl StreamsState {
             }
         }
 
-        self.pending.streams.clear();
+        self.pending.clear();
         self.send_streams = 0;
         self.data_sent = 0;
         self.connection_blocked.clear();
@@ -345,7 +345,7 @@ impl StreamsState {
     /// Whether any stream data is queued, regardless of control frames
     pub(crate) fn can_send_stream_data(&self) -> bool {
         // Reset streams may linger in the pending stream list, but will never produce stream frames
-        self.pending.streams.iter().any(|stream| {
+        self.pending.iter().any(|stream| {
             self.send
                 .get(&stream.id)
                 .and_then(|s| s.as_ref())
@@ -503,7 +503,7 @@ impl StreamsState {
 
             // Pop the stream of the highest priority that currently has pending data
             // If the stream still has some pending data left after writing, it will be reinserted, otherwise not
-            let Some(stream) = self.pending.streams.pop() else {
+            let Some(stream) = self.pending.pop() else {
                 break;
             };
 
@@ -1312,7 +1312,7 @@ mod tests {
         assert_eq!(meta[2].id, id_low);
 
         assert!(!server.can_send_stream_data());
-        assert_eq!(server.pending.streams.len(), 0);
+        assert_eq!(server.pending.len(), 0);
     }
 
     #[test]
@@ -1341,7 +1341,7 @@ mod tests {
             conn_state: &state,
         };
         assert_eq!(mid.write(b"mid").unwrap(), 3);
-        assert_eq!(server.pending.streams.len(), 1);
+        assert_eq!(server.pending.len(), 1);
 
         let mut high = SendStream {
             id: id_high,
@@ -1351,7 +1351,7 @@ mod tests {
         };
         high.set_priority(1).unwrap();
         assert_eq!(high.write(&[0; 200]).unwrap(), 200);
-        assert_eq!(server.pending.streams.len(), 2);
+        assert_eq!(server.pending.len(), 2);
 
         // Requeue the high priority stream to lowest priority. The initial send
         // still uses high priority since it's queued that way. After that it will
@@ -1370,7 +1370,7 @@ mod tests {
         assert_eq!(meta[0].id, id_high);
 
         // After requeuing we should end up with 2 priorities - not 3
-        assert_eq!(server.pending.streams.len(), 2);
+        assert_eq!(server.pending.len(), 2);
 
         // Send the remaining data. The initial mid priority one should go first now
         let meta = server.write_stream_frames(&mut buf, 1000);
@@ -1379,7 +1379,7 @@ mod tests {
         assert_eq!(meta[1].id, id_high);
 
         assert!(!server.can_send_stream_data());
-        assert_eq!(server.pending.streams.len(), 0);
+        assert_eq!(server.pending.len(), 0);
     }
 
     #[test]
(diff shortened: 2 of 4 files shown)


quinn-proto/src/connection/streams/mod.rs
```diff
@@ -365,15 +365,19 @@ impl<'a> SendStream<'a> {
 /// A queue of streams with pending outgoing data, sorted by priority
 struct PendingStreamsQueue {
     streams: BinaryHeap<PendingStream>,
+    fair: bool,
+    reinserted: Option<PendingStream>,
```
Ralith, line 369: Document new fields. Also, bikeshed: `next`?

quinn-proto/src/connection/streams/state.rs
```diff
@@ -536,7 +537,7 @@ impl StreamsState {
                 // If the stream still has pending data, reinsert it, possibly with an updated priority value
                 // Fairness with other streams is achieved by implementing round-robin scheduling,
                 // so that the other streams will have a chance to write data before we touch this stream again.
-                self.pending.push_pending(id, stream.priority);
+                self.pending.reinsert_pending(id, stream.priority);
```
Ralith, line 540: nit: pass `fair` down here to save the extra field in `PendingStreamsQueue`?

Ralith:
This change seems well motivated, thanks!

I wonder if we could have the best of both worlds, and avoid yet another esoteric configuration knob, with a more heuristic approach. It sounds like your main problem is that, when sending many individually sub-MTU streams, streams that get split across packets are subject to a much higher maximum latency than you might otherwise see. What if we special-cased those? E.g. don't advance the round-robin state if:
- a fragmented stream has less than an MTU of data remaining
- alternatively, a fragmented stream was not the first stream in a packet

Either should let us retain fairness for larger-than-MTU streams, while reducing the maximum latency for sub-MTU streams, without requiring users to understand any of these concerns.

alessandrod:
> This change seems well motivated, thanks!
> 
> I wonder if we could have the best of both worlds, and avoid yet another esoteric configuration knob, with a more heuristic approach. It sounds like your main problem is that, when sending many individually sub-MTU streams, streams that get split across packets are subject to a much higher maximum latency than you might otherwise see. What if we special-cased those? E.g. don't advance the round-robin state if:
> 
>     * a fragmented stream has less than an MTU of data remaining
> 
>     * alternatively, a fragmented stream was not the first stream in a packet
> 
> 
> Either should let us retain fairness for larger-than-MTU streams, while reducing the maximum latency for sub-MTU streams, without requiring users to understand any of these concerns.

For the general case, this sounds like a good idea, I'm happy to implement it if you want.

For our case specifically though, we have a proposal to 3x the max transaction size. We haven't really been able to even evaluate its feasibility so far because latency was already pretty bad, but once we deploy this change, I think we could reasonably enable 3x, therefore spanning multiple MTUs. In that case, I'd still want all datagrams of a tx to follow each other sequentially. 

> without requiring users to understand any of these concerns

Yeah this is a valid point ofc. How about I implement the heuristic to disable RR for < MTU, and then keep an option to always disable RR but instead of exposing it as a flag, I make it an opt-in feature flag?

alessandrod:
> keep an option to always disable RR but instead of exposing it as a flag, I make it an opt-in feature flag?

I can also always keep this out of tree ofc! I'd rather avoid having to have a vendor fork just for this tho.

djc:
Heuristics sounds good to me, but given the limited complexity of the additional configuration I feel like we could also accept the configuration upstream if it's well-motivated even in the presence of the heuristics.

Ralith:
> maybe we should avoid adding heuristics if they don't address the only actual use case we've seen?

This is compelling. Heuristics are a greater maintenance burden, and if they're not actually addressing the motivating case, why pay that cost?

I'm happy to move ahead with a global config flag, as originally proposed. I do suspect there's a more flexible middle ground here somewhere (maybe the setting should be per-stream?) but we don't necessarily have to work that out here and now when there's a clear win on the table already.

quinn-proto/src/connection/streams/mod.rs
alessandrod: done

quinn-proto/src/connection/streams/state.rs
alessandrod: good idea, done

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -378,6 +378,23 @@ impl PendingStreamsQueue {
...
+        self.streams.clear();
+    }
+
+    fn iter(&self) -> impl Iterator<Item = &PendingStream> {
+        self.streams.iter()
+    }
+
+    fn pop(&mut self) -> Option<PendingStream> {
+        self.streams.pop()
+    }
```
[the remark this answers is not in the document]
djc: I'd like to have this change in the first commit instead of the second, to keep them conceptually atomic.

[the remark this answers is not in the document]
alessandrod: I forgot that I had two commits 🤦‍♂️😅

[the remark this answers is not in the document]
alessandrod: fixed! also note I fixed len() to also include self.next if Some. (len is only used by tests but still)

djc (approved):
Thanks!

Ralith (approved):
Beautiful, thanks!

Ralith:
@lijunwangs this change may be of interest to you as well -- I recall you had similar concerns.

PR #2003, opened 2024-10-07
feat(quinn, quinn-proto): add aws-lc-rs-fips feature flag

M0dEx:
Adds support for aws-lc-rs FIPS mode using a feature flag in quinn and quinn-proto.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -172,8 +172,11 @@ jobs:
     runs-on: ${{ matrix.os }}
     env:
       RUSTFLAGS: -Dwarnings
+      # skip FIPS features outside of Linux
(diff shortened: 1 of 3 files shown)


djc (approved):

M0dEx:
Seems the tests finished, but the `features` workflow has an issue due to a missing `golang` dependency of the FIPS mode.

djc:
> The rustls project uses only the `ubuntu-latest` for feature combination checks, but I assume this project had a reason to have those run in a matrix over different OS's.

Yes, but it's fine to exempt the `fips` feature from the OS matrix and only support it on Linux given that's where it mainly makes sense per upstream guidance.

M0dEx:
The `features` job should now skip FIPS on WIndows and macOS.

Ralith (approved):

djc (approved):
Thanks, looks great!

PR #2009, opened 2024-10-09
#2008: Make max_idle_timeout negotiation commutative.

### LINKED ISSUE #2008: Incorrect `max_idle_timeout` negotiation.
## Steps to reproduce
1. Configure client side with custom transport parameters by overriding `max_idle_timeout` to `Duration::ZERO`
2. Connect to quinn-based remote server which configured `max_idle_timeout` some some value, e.g. 180 seconds. During handshake client side will omit `max_idle_timeout` parameter inside TLS handshake to server which seems to correspond to this code: https://github.com/quinn-rs/quinn/blob/15a4dcef42bf10c84535ec7e8331db9e97918856/quinn-proto/src/transport_parameters.rs#L36-L37

## Actual result
Quinn on server side will negotiate `idle_timeout` to be `180` seconds because not value from client received, but client side connection will actually compute `idle_timeout` to be `0`:
https://github.com/quinn-rs/quinn/blob/15a4dcef42bf10c84535ec7e8331db9e97918856/quinn-proto/src/connection/mod.rs#L3293-L3298
effectively falling back to `3 * PTO` timeout:
https://github.com/quinn-rs/quinn/blob/15a4dcef42bf10c84535ec7e8331db9e97918856/quinn-proto/src/connection/mod.rs#L1845-L1856

Due to low effective idle timeout connection is frequently locally closed and need to be re-established.

## Expected result
I'd expect that both sides agree on the same value of `max_idle_timeout` so in case local side configured with `0` timeout and peer side have non-zero idle timeout configured both endpoints will use same non-zero idle timeout.

This expectation seems to match the RFC where `0` value is threaten as disabled timeout according to [section 18.2](https://www.rfc-editor.org/rfc/rfc9000#section-18.2) which already respected while writing transport parameters to wire format.
> The maximum idle timeout is a value in milliseconds that is encoded as an integer; see ([Section 10.1](https://www.rfc-editor.org/rfc/rfc9000#idle-timeout)). Idle timeout is disabled when both endpoints omit this transport parameter or specify a value of 0.

And according to [section 10.1](https://www.rfc-editor.org/rfc/rfc9000#section-10.1) during negotiation in case of only one side have enabled `max_idle_timeout` both endpoints must negotiate to the value provided by other side.

> Each endpoint advertises a max_idle_timeout, but the effective value at an endpoint is computed as the minimum of the two advertised values (or the sole advertised value, if only one endpoint advertises a non-zero value).

mstyura:
Fixes #2008.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -186,7 +186,7 @@ pub struct Connection {
     /// Whether the idle timer should be reset the next time an ack-eliciting packet is transmitted.
     permit_idle_reset: bool,
     /// Negotiated idle timeout
-    idle_timeout: Option<VarInt>,
+    idle_timeout: Option<Duration>,
     timers: TimerTable,
     /// Number of packets received which could not be authenticated
     authentication_failures: u64,
@@ -317,7 +317,10 @@ impl Connection {
             next_crypto: None,
             accepted_0rtt: false,
             permit_idle_reset: true,
-            idle_timeout: config.max_idle_timeout,
+            idle_timeout: match config.max_idle_timeout {
+                None | Some(VarInt(0)) => None,
+                Some(dur) => Some(Duration::from_millis(dur.0)),
+            },
             timers: TimerTable::default(),
             authentication_failures: 0,
             error: None,
@@ -1845,7 +1848,7 @@ impl Connection {
     fn reset_idle_timeout(&mut self, now: Instant, space: SpaceId) {
         let timeout = match self.idle_timeout {
             None => return,
-            Some(x) => Duration::from_millis(x.0),
+            Some(dur) => dur,
         };
         if self.state.is_closed() {
             self.timers.stop(Timer::Idle);
@@ -3290,12 +3293,9 @@ impl Connection {
 
     fn set_peer_params(&mut self, params: TransportParameters) {
         self.streams.set_params(&params);
-        self.idle_timeout = match (self.config.max_idle_timeout, params.max_idle_timeout) {
-            (None, VarInt(0)) => None,
-            (None, x) => Some(x),
-            (Some(x), VarInt(0)) => Some(x),
-            (Some(x), y) => Some(cmp::min(x, y)),
-        };
+        self.idle_timeout =
+            negotiate_max_idle_timeout(self.config.max_idle_timeout, Some(params.max_idle_timeout));
+        trace!("negotiated max idle timeout {:?}", self.idle_timeout);
         if let Some(ref info) = params.preferred_address {
             self.rem_cids.insert(frame::NewConnectionId {
                 sequence: 1,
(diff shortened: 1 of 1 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -317,7 +317,10 @@ impl Connection {
             next_crypto: None,
             accepted_0rtt: false,
             permit_idle_reset: true,
-            idle_timeout: config.max_idle_timeout,
+            idle_timeout: config
+                .max_idle_timeout
+                .filter(|dur| dur.0 != 0)
+                .map(|dur| Duration::from_millis(dur.0)),
```
djc, line 323: Nit: instead of the `filter().map()` chaining, I think a `match` expression would make it easier to understand the logic here.

quinn-proto/src/connection/mod.rs
```diff
@@ -3290,12 +3290,11 @@ impl Connection {
...
     fn set_peer_params(&mut self, params: TransportParameters) {
         self.streams.set_params(&params);
-        self.idle_timeout = match (self.config.max_idle_timeout, params.max_idle_timeout) {
-            (None, VarInt(0)) => None,
-            (None, x) => Some(x),
-            (Some(x), VarInt(0)) => Some(x),
-            (Some(x), y) => Some(cmp::min(x, y)),
-        };
+        self.idle_timeout = negotiate_max_idle_timeout(
+            self.config.max_idle_timeout.unwrap_or_default(),
```
djc, line 3297: Nit: I think I'd prefer avoiding the `unwrap_or_default()` here in favor of explicitly matching `Some(VarInt(0)) | None`?

djc (approved):
These changes make sense to me, thanks!

quinn-proto/src/connection/mod.rs
mstyura: Could you please clarify you mean change sinagure of `negotiate_max_idle_timeout` to accept both or only local argument as `Option<VarInt>` and do all checks inside or having something like outside:
```
match config.max_idle_timeout {
    None | Some(VarInt(0)) => None,
    Some(dur) => Some(dur)),
}
```
?

mstyura: Changed signature of `negotiate_max_idle_timeout` to accept both arguments as `Option<VarInt>` and perform checks inside to treat `Some(VarInt(0))` and `None` as disabled timeout.

djc: LGTM, thanks! Would you mind squashing the review feedback commit into the original one(s)?

mstyura: Squashed MR to single commit.

djc: Well, I actually liked the split you had previously between the commutative negotiation and changing the timeout type, but I can live with this. Should at least fix up the first line of the commit message to be more descriptive of the change though.

mstyura: Ah, sorry, the initial split was still imperfect because both commits mixed change of type and signature& implementation of negotiation function.
I've also update commit message to make it more informative so hope it's now clear what was the purpose of the commit for anyone who will see it in git history.

Ralith (approved):
Thanks!

The original intent here was that users pass `None` to `TransportConfig::max_idle_timeout`, rather than `Some(0)`, if no maximum timeout is desired. I wonder if we should revisit that interface to make this less of a footgun?

Ralith:
This is a strict non-breaking improvement, anyway.

PR #2010, opened 2024-10-09
bench(udp): run GSO, GRO and recvmmsg permutations

mxinden:
- Benchmark all permutations of GSO, GRO and `recvmmsg`.
- Add support for all platforms.
- Add test-run (i.e. `cargo test --benches`, single execution in debug mode) to CI.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -30,7 +30,7 @@ jobs:
             $HOME/.cargo/bin/rustc --version
             echo "~~~~ freebsd-version ~~~~"
             freebsd-version
-          run: $HOME/.cargo/bin/cargo build --all-targets && $HOME/.cargo/bin/cargo test && $HOME/.cargo/bin/cargo test --manifest-path fuzz/Cargo.toml
+          run: $HOME/.cargo/bin/cargo build --all-targets && $HOME/.cargo/bin/cargo test && $HOME/.cargo/bin/cargo test --manifest-path fuzz/Cargo.toml && $HOME/.cargo/bin/cargo test -p quinn-udp --benches
   test:
     strategy:
       matrix:
@@ -54,6 +54,7 @@ jobs:
       - run: cargo test
       - run: cargo test --manifest-path fuzz/Cargo.toml
         if: ${{ matrix.rust }} == "stable"
+      - run: cargo test -p quinn-udp --benches
 
   test-aws-lc-rs:
     runs-on: ubuntu-latest
--- a/quinn-udp/Cargo.toml
+++ b/quinn-udp/Cargo.toml
@@ -30,8 +30,13 @@ once_cell = { workspace = true }
 windows-sys = { workspace = true }
 
 [dev-dependencies]
-criterion = "0.5"
+criterion = { version = "0.5", default-features = false, features = ["async_tokio"] }
+tokio = { workspace = true, features = ["rt", "rt-multi-thread", "net"] }
 
-[target.'cfg(any(target_os = "linux", target_os = "windows"))'.bench]
+[lib]
+# See https://github.com/bheisler/criterion.rs/blob/master/book/src/faq.md#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options
+bench = false
+
+[[bench]]
 name = "throughput"
 harness = false
--- a/quinn-udp/benches/throughput.rs
+++ b/quinn-udp/benches/throughput.rs
@@ -7,7 +7,7 @@ use std::{
 use criterion::{criterion_group, criterion_main, Criterion};
 use tokio::{io::Interest, runtime::Runtime};
 
-use quinn_udp::{RecvMeta, Transmit, UdpSocketState};
+use quinn_udp::{RecvMeta, Transmit, UdpSocketState, BATCH_SIZE};
 
 pub fn criterion_benchmark(c: &mut Criterion) {
     const TOTAL_BYTES: usize = 10 * 1024 * 1024;
@@ -39,12 +39,17 @@ pub fn criterion_benchmark(c: &mut Criterion) {
                 continue;
             }
 
-            permutations.push((gso_enabled, gro_enabled));
+            for recvmmsg_enabled in [false, true] {
+                permutations.push((gso_enabled, gro_enabled, recvmmsg_enabled));
+            }
         }
     }
 
-    for (gso_enabled, gro_enabled) in permutations {
-        let mut group = c.benchmark_group(format!("gso_{}_gro_{}", gso_enabled, gro_enabled));
+    for (gso_enabled, gro_enabled, recvmmsg_enabled) in permutations {
+        let mut group = c.benchmark_group(format!(
+            "gso_{}_gro_{}_recvmmsg_{}",
+            gso_enabled, gro_enabled, recvmmsg_enabled
+        ));
         group.throughput(criterion::Throughput::Bytes(TOTAL_BYTES as u64));
 
         let gso_segments = if gso_enabled {
@@ -65,7 +70,7 @@ pub fn criterion_benchmark(c: &mut Criterion) {
         } else {
             1
         };
-        let batch_size = 1;
+        let batch_size = if recvmmsg_enabled { BATCH_SIZE } else { 1 };
 
         group.bench_function("throughput", |b| {
             b.to_async(&rt).iter(|| async {
--- a/quinn-udp/benches/throughput.rs
+++ b/quinn-udp/benches/throughput.rs
@@ -1,38 +1,57 @@
+use std::{
+    io::{ErrorKind, IoSliceMut},
+    net::{Ipv4Addr, Ipv6Addr, UdpSocket},
+};
+
 use criterion::{criterion_group, criterion_main, Criterion};
+use tokio::{io::Interest, runtime::Runtime};
+
 use quinn_udp::{RecvMeta, Transmit, UdpSocketState};
-use std::cmp::min;
-use std::net::{Ipv4Addr, Ipv6Addr};
-use std::{io::IoSliceMut, net::UdpSocket, slice};
 
 pub fn criterion_benchmark(c: &mut Criterion) {
     const TOTAL_BYTES: usize = 10 * 1024 * 1024;
-    // Maximum GSO buffer size is 64k.
-    const MAX_BUFFER_SIZE: usize = u16::MAX as usize;
     const SEGMENT_SIZE: usize = 1280;
+
+    let rt = Runtime::new().unwrap();
+    let _guard = rt.enter();
 
     let (send_state, send_socket) = new_socket();
     let (recv_state, recv_socket) = new_socket();
-    // Reverse non-blocking flag set by `UdpSocketState` to make the test non-racy
-    recv_socket.set_nonblocking(false).unwrap();
-
-    let max_segments = min(
-        UdpSocketState::new((&send_socket).into())
-            .unwrap()
-            .max_gso_segments(),
-        MAX_BUFFER_SIZE / SEGMENT_SIZE,
-    );
     let dst_addr = recv_socket.local_addr().unwrap();
 
-    let mut receive_buffer = vec![0; MAX_BUFFER_SIZE];
-    let mut meta = RecvMeta::default();
+    let mut permutations = vec![];
+    for gso_enabled in [
+        false,
+        #[cfg(any(target_os = "linux", target_os = "windows"))]
+        true,
+    ] {
+        for gro_enabled in [false, true] {
+            #[cfg(target_os = "windows")]
+            if gso_enabled && !gro_enabled {
+                // Windows requires receive buffer to fit entire datagram on GRO
+                // enabled socket.
+                //
+                // OS error: "A message sent on a datagram socket was larger
+                // than the internal message buffer or some other network limit,
+                // or the buffer used to receive a datagram into was smaller
+                // than the datagram itself."
+                continue;
+            }
 
-    for gso_enabled in [false, true] {
-        let mut group = c.benchmark_group(format!("gso_{}", gso_enabled));
+            permutations.push((gso_enabled, gro_enabled));
+        }
+    }
+
+    for (gso_enabled, gro_enabled) in permutations {
+        let mut group = c.benchmark_group(format!("gso_{}_gro_{}", gso_enabled, gro_enabled));
         group.throughput(criterion::Throughput::Bytes(TOTAL_BYTES as u64));
 
-        let segments = if gso_enabled { max_segments } else { 1 };
-        let msg = vec![0xAB; SEGMENT_SIZE * segments];
-
+        let gso_segments = if gso_enabled {
+            send_state.max_gso_segments()
+        } else {
+            1
+        };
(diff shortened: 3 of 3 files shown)


quinn-udp/Cargo.toml
```diff
@@ -30,8 +30,13 @@ once_cell = { workspace = true }
 windows-sys = { workspace = true }
 
 [dev-dependencies]
-criterion = "0.5"
+criterion = { version = "0.5", default-features = false, features = ["async_tokio"] }
+tokio = { workspace = true, features = ["rt", "rt-multi-thread", "net"] }
```
mxinden, line 34: On Linux blocking `recvmmsg` calls will wait for all `iovec`s to be filled. 

>  A blocking recvmmsg() call blocks until vlen messages have been
       received or until the timeout expires.  A nonblocking call reads
       as many messages as are available (up to the limit specified by
       vlen) and returns immediately.

https://man7.org/linux/man-pages/man2/recvmmsg.2.html

This is e.g. problematic when combined with a GSO send with less segments than `iovec`s in its `recvmmsg` counterpart.

Thus this pull request changes the benchmark to use non-blocking instead of blocking `recvmmsg`. To do so, it uses `tokio`.

See also discussion in https://github.com/quinn-rs/quinn/pull/1993#discussion_r1786705730.

quinn-udp/benches/throughput.rs
```diff
@@ -1,76 +1,124 @@
...
+    let (recv_socket, recv_state) = new_socket(&mut rt);
+    let dst_addr = recv_socket.local_addr().unwrap();
 
-    let mut receive_buffer = vec![0; MAX_BUFFER_SIZE];
-    let mut meta = RecvMeta::default();
+    let mut permutations = vec![];
+    for gso_enabled in [
+        false,
+        #[cfg(any(target_os = "linux", target_os = "windows"))]
+        true,
```
mxinden, line 27: Other platforms don't support GSO.

Needs to add apple platforms once https://github.com/quinn-rs/quinn/pull/1993 merges.

quinn-udp/benches/throughput.rs
```diff
@@ -1,76 +1,124 @@
...
+use std::{
+    cmp::min,
+    io::{ErrorKind, IoSliceMut},
+    net::{Ipv4Addr, Ipv6Addr, UdpSocket},
+};
+use tokio::io::Interest;
+use tokio::runtime::Runtime;
+
+const MAX_IP_UDP_HEADER_SIZE: usize = 48;
+const MAX_DATAGRAM_SIZE: usize = u16::MAX as usize - MAX_IP_UDP_HEADER_SIZE;
```
mxinden, line 12: Linux allows for up to 64KiB per GSO and GRO call. That includes IP and UDP header.

quinn-udp/benches/throughput.rs
```diff
@@ -1,76 +1,124 @@
...
-                        received_segments += meta.len / meta.stride;
+                    while received < sent {
+                        recv_socket.readable().await.unwrap();
+                        let n = match recv_socket.try_io(Interest::READABLE, || {
+                            recv_state.recv((&recv_socket).into(), &mut receive_slices, &mut meta)
+                        }) {
+                            Ok(n) => n,
+                            // recv.readable() can lead to false positives. Try again.
+                            Err(e) if e.kind() == ErrorKind::WouldBlock => continue,
+                            e => e.unwrap(),
```
mxinden, line 103: `UdpSocket::readable` is allowed to return false-positives.

> The function may complete without the socket being readable. This is a false-positive and attempting a try_recv() will return with io::ErrorKind::WouldBlock.

https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.readable

.github/workflows/rust.yml
```diff
@@ -54,6 +54,7 @@ jobs:
       - run: cargo test
       - run: cargo test --manifest-path fuzz/Cargo.toml
         if: ${{ matrix.rust }} == "stable"
+      - run: cargo test --benches
```
mxinden, line 57: Note that I did not simply extend the existing `cargo test` call, but instead added an additional `cargo test --benches`.

`cargo test --benches` executes some but not all unit tests in addition to all benchmark tests. See also https://github.com/rust-lang/cargo/issues/6454.

mxinden:
@larseggert this pull request should resolve all issues raised in https://github.com/quinn-rs/quinn/pull/1993#pullrequestreview-2346382573. Can you take a look?

Targeting `main` directly given that changes are not tied to https://github.com/quinn-rs/quinn/pull/1993.

larseggert (approved):

mxinden:
@djc @Ralith ready for a full review. Let me know if you find this pull request helpful. 

Will unblock https://github.com/quinn-rs/quinn/pull/1993.

quinn-udp/Cargo.toml
djc: Suggest you keep a separate commit that makes the change from sync to async.

quinn-udp/benches/throughput.rs
```diff
@@ -1,76 +1,124 @@
 use criterion::{criterion_group, criterion_main, Criterion};
-use quinn_udp::{RecvMeta, Transmit, UdpSocketState};
-use std::cmp::min;
-use std::net::{Ipv4Addr, Ipv6Addr};
-use std::{io::IoSliceMut, net::UdpSocket, slice};
+use quinn_udp::{RecvMeta, Transmit, UdpSocketState, BATCH_SIZE};
```
djc, line 2: Nit: would be good to do `StdExternalCrate` import blocks here.

quinn-udp/benches/throughput.rs
```diff
@@ -1,76 +1,124 @@
...
+use std::{
+    cmp::min,
+    io::{ErrorKind, IoSliceMut},
+    net::{Ipv4Addr, Ipv6Addr, UdpSocket},
+};
+use tokio::io::Interest;
+use tokio::runtime::Runtime;
+
+const MAX_IP_UDP_HEADER_SIZE: usize = 48;
```
djc, line 11: Nit: constants should go near the bottom.

quinn-udp/benches/throughput.rs
```diff
@@ -1,76 +1,124 @@
...
-use std::cmp::min;
-use std::net::{Ipv4Addr, Ipv6Addr};
-use std::{io::IoSliceMut, net::UdpSocket, slice};
+use quinn_udp::{RecvMeta, Transmit, UdpSocketState, BATCH_SIZE};
+use std::{
+    cmp::min,
+    io::{ErrorKind, IoSliceMut},
+    net::{Ipv4Addr, Ipv6Addr, UdpSocket},
+};
+use tokio::io::Interest;
```
djc, line 8: Nit: for consistency, there should be one import statement for all `tokio` imports.

quinn-udp/benches/throughput.rs
```diff
@@ -1,76 +1,124 @@
...
+                        received += meta.iter().map(|m| m.len).take(n).sum::<usize>();
                     }
-                    assert_eq!(received_segments, segments);
                 }
             })
         });
     }
 }
 
+fn new_socket(rt: &mut Runtime) -> (tokio::net::UdpSocket, UdpSocketState) {
```
djc, line 113: Please keep the extraction of this function in a separate commit.

djc (commented):
I think this is probably a good change but would like to see more organization in terms of separate commits -- here's some early feedback.

quinn-udp/benches/throughput.rs
mxinden: :+1: c5172542

quinn-udp/benches/throughput.rs
mxinden: :+1: 5609a1d3

quinn-udp/benches/throughput.rs
mxinden: :+1: 5609a1d3

quinn-udp/benches/throughput.rs
mxinden: :+1: 5609a1d3 among others.

quinn-udp/Cargo.toml
mxinden: :+1: 5609a1d3

mxinden:
Thank you for the review @djc. I applied all suggestions. In addition, I split the one large commit into smaller atomic commits. Let me know what you think.

mxinden:
Windows CI failure seems intermittent. See e.g. https://github.com/mxinden/quinn/actions/runs/11424992884/job/31786093629 for a succeeding run, simply adding a `RUST_BACKTRACE=1`.

quinn-udp/benches/throughput.rs
```diff
@@ -1,79 +1,120 @@
...
+    for (gso_enabled, gro_enabled) in permutations {
+        let mut group = c.benchmark_group(format!("gso_{}_gro_{}", gso_enabled, gro_enabled));
+        group.throughput(criterion::Throughput::Bytes(TOTAL_BYTES as u64));
 
+        let gso_segments = if gso_enabled {
+            send_state.max_gso_segments()
+        } else {
+            1
+        };
+        let msg = vec![0xAB; min(MAX_DATAGRAM_SIZE, SEGMENT_SIZE * gso_segments)];
```
Ralith, line 58: Why was this `min` introduced? It doesn't seem related to the intent of the commit.

Ralith (changes requested):

quinn-udp/benches/throughput.rs
```diff
@@ -1,79 +1,120 @@
...
+        let gro_segments = if gro_enabled {
+            recv_state.gro_segments()
+        } else {
+            1
+        };
+        let batch_size = 1;
 
         group.bench_function("throughput", |b| {
-            b.iter(|| {
+            b.to_async(Runtime::new().unwrap()).iter(|| async {
```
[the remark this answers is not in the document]
mxinden: Good idea. Done in https://github.com/quinn-rs/quinn/pull/2010/commits/160a22ca0c071308c677d01aaa53c6ce9f34732f.

quinn-udp/benches/throughput.rs
```diff
@@ -1,79 +1,120 @@
...
 }
 
-fn new_socket() -> (UdpSocket, UdpSocketState) {
+fn new_socket(rt: &mut Runtime) -> (tokio::net::UdpSocket, UdpSocketState) {
     let socket = UdpSocket::bind((Ipv6Addr::LOCALHOST, 0))
         .or_else(|_| UdpSocket::bind((Ipv4Addr::LOCALHOST, 0)))
         .unwrap();
 
     let state = UdpSocketState::new((&socket).into()).unwrap();
+    let socket = rt.block_on(async { tokio::net::UdpSocket::from_std(socket).unwrap() });
```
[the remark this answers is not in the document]
mxinden: Good idea. Done in https://github.com/quinn-rs/quinn/pull/2010/commits/160a22ca0c071308c677d01aaa53c6ce9f34732f.

quinn-udp/benches/throughput.rs
mxinden: The `min` is need on Linux, which requires a segmented UDP datagram not to exceed the UDP datagram size limit, i.e. ~64k minus headers.

You are right, this does not directly relates to the intent of https://github.com/quinn-rs/quinn/pull/2010/commits/160a22ca0c071308c677d01aaa53c6ce9f34732f. I extracted it into a separate commit. 

See https://github.com/quinn-rs/quinn/pull/2010/commits/51ef00cc8b36103b1c88174042d67d355f266f96 and commit description for details.

.github/workflows/rust.yml
```diff
@@ -54,6 +54,7 @@ jobs:
       - run: cargo test
       - run: cargo test --manifest-path fuzz/Cargo.toml
         if: ${{ matrix.rust }} == "stable"
+      - run: cargo test -p quinn-udp --benches
```
mxinden, line 57: See https://github.com/quinn-rs/quinn/pull/2010#issuecomment-2424749007 for reason for `-p quinn-udp`.

mxinden:
Thank you for the review @Ralith. Would you mind taking another look?

quinn-udp/benches/throughput.rs
```diff
@@ -72,5 +67,13 @@ pub fn criterion_benchmark(c: &mut Criterion) {
     }
 }
 
+fn new_socket() -> (UdpSocket, UdpSocketState) {
+    let socket = UdpSocket::bind((Ipv6Addr::LOCALHOST, 0))
+        .or_else(|_| UdpSocket::bind((Ipv4Addr::LOCALHOST, 0)))
+        .unwrap();
+
+    let state = UdpSocketState::new((&socket).into()).unwrap();
+    (socket, state)
```
djc: Nits: I think this might read better without intermediate bindings, just yielding the tuple directly?

Also, add an empty line below the function please.

quinn-udp/benches/throughput.rs
```diff
@@ -1,79 +1,118 @@
...
-        UdpSocketState::new((&send_socket).into())
-            .unwrap()
-            .max_gso_segments(),
-        MAX_BUFFER_SIZE / SEGMENT_SIZE,
-    );
     let dst_addr = recv_socket.local_addr().unwrap();
 
-    let mut receive_buffer = vec![0; MAX_BUFFER_SIZE];
-    let mut meta = RecvMeta::default();
+    let mut permutations = vec![];
```
djc, line 23: Instead of this loop, I'd propose turning this into a `const PERMUTATIONS: [(bool, bool); 4]`, which I think will end up being easier to understand. Even after the addition of recvmmsg, I think it's about as compact as the current approach but fully declarative.

(Also seems to me like it would be worthwhile to split the changes to permutations from the commit that makes things async.)

djc (approved):
Thanks, this looks better!

quinn-udp/benches/throughput.rs
mxinden: @djc are you suggesting something along the lines of:

``` rust
    const PERMUTATIONS: [(bool, bool, bool); 6] = [
        (false, false, false),
        (false, false, true),
        (false, true, false),
        (false, true, true),
        (true, false, true),
        (true, true, false),
        // Windows requires receive buffer to fit entire datagram on GRO
        // enabled socket.
        //
        // OS error: "A message sent on a datagram socket was larger
        // than the internal message buffer or some other network limit,
        // or the buffer used to receive a datagram into was smaller
        // than the datagram itself."
        #[cfg(target_os = "windows")]
        (true, false, false),
        // As above.
        #[cfg(target_os = "windows")]
        (true, true, true),
    ];
```

I don't find this easier to read. It is not obvious to me, that the above covers all permutations of `(bool, bool, bool)`, or contains any duplicates. Nor is it obvious which one would be for `gso_enabled`, `gro_enabled` and `recvmmsg_enabled`.

quinn-udp/benches/throughput.rs
mxinden: :+1: done with the latest force push across the 4 commits.

quinn-udp/benches/throughput.rs
djc: Fair enough, feel free to ignore that.

Ralith (approved):
Thanks!

PR #2015, opened 2024-10-17
fix `rustls-aws-lc-rs` feature

stormshield-gt:
In the current Rust editions, if a dependency is declared as optional and it also exists a feature with the same name, `dep:name` will activate both the feature and the optional dependency.

That is the case for the `rustls` feature, so sadly 
```toml
rustls-aws-lc-rs = ["dep:rustls", "rustls/aws-lc-rs", "aws-lc-rs"]
```
will also activate the `rustls` feature, which also activates `ring`.

There is a zulip discussion about this surprising behavior [here](https://rust-lang.zulipchat.com/#narrow/channel/246057-t-cargo/topic/Odd.20cargo.20resolver.20behavior/near/449102225).

Hopefully, in Rust 2024 edition, with [this change](https://doc.rust-lang.org/edition-guide/rust-2024/cargo-remove-implicit-features.html), the behavior will be less surprising.

The PR introduce a breaking change, but as far as I know, there is no other way to fix this

Edit: I misunderstood the `zulip` thread; it's the `crate/feature` syntax that activates both the optional dependency and the feature if there is one with the same name. Not the `dep:crate` one. We can just use the `crate?/feature` syntax here and it should be non-breaking.

### DIFF
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -15,7 +15,7 @@ all-features = true
 
 [features]
 default = ["rustls-ring", "log"]
-aws-lc-rs = ["dep:aws-lc-rs", "aws-lc-rs/aws-lc-sys", "aws-lc-rs/prebuilt-nasm"]
+aws-lc-rs = ["dep:aws-lc-rs", "aws-lc-rs?/aws-lc-sys", "aws-lc-rs?/prebuilt-nasm"]
 aws-lc-rs-fips = ["aws-lc-rs", "aws-lc-rs?/fips"]
 # For backwards compatibility, `rustls` forwards to `rustls-ring`
 rustls = ["rustls-ring"]
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -20,10 +20,10 @@ aws-lc-rs-fips = ["aws-lc-rs", "aws-lc-rs?/fips"]
 # For backwards compatibility, `rustls` forwards to `rustls-ring`
 rustls = ["rustls-ring"]
 # Enable rustls with the `aws-lc-rs` crypto provider
-rustls-aws-lc-rs = ["dep:rustls", "rustls/aws-lc-rs", "aws-lc-rs"]
+rustls-aws-lc-rs = ["dep:rustls", "rustls?/aws-lc-rs", "aws-lc-rs"]
 rustls-aws-lc-rs-fips = ["rustls-aws-lc-rs", "aws-lc-rs-fips"]
 # Enable rustls with the `ring` crypto provider
-rustls-ring = ["dep:rustls", "rustls/ring", "ring"]
+rustls-ring = ["dep:rustls", "rustls?/ring", "ring"]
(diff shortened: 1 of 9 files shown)


djc:
> The PR introduce a breaking change, but as far as I know, there is no other way to fix this

I don't think we want to do a breaking change for this.

stormshield-gt:
> I don't think we want to do a breaking change for this.

I see, do you have any ideas about other alternatives? 

Right now the `aws-lc-rs` `rustls-aws-lc-rs` features cannot be used in the current state because [`quinn` still uses the `ring` provider when they are activated](https://github.com/quinn-rs/quinn/blob/41850c8a304f09c7d009a6e70e48f35bd737e1b5/quinn-proto/src/crypto/rustls.rs#L551-L556).

By executing this command on `main` we can see that the ring feature is activated on `quinn-proto`, so in case of using `rustls-aws-lc-rs` we will fallback to use `ring`:

```sh
$ cargo tree -p quinn -f "{p} {f}" --no-default-features --features rustls-aws-lc-rs
quinn v0.11.5 (/home/tudyg/dev/forks/quinn/quinn) aws-lc-rs,rustls-aws-lc-rs
├── bytes v1.7.2 default,std
├── pin-project-lite v0.2.14
├── quinn-proto v0.11.8 (/home/tudyg/dev/forks/quinn/quinn-proto) aws-lc-rs,ring,rustls,rustls-aws-lc-rs,rustls-ring
```

djc:
Maybe we can rename the rustls dependency to something else (like, `rustls-core` or something), adapting the code, to avoid `dep:rustls` triggering the feature of the same name?

djc:
Note that we haven't released #1962 yet, so changing the aws-lc-rs features in backwards-incompatible ways could also be an option.

stormshield-gt:
Some job failures (https://github.com/quinn-rs/quinn/actions/runs/11383037824/job/31667830692?pr=2015) seems unrelated to this PR, maybe this PR https://github.com/quinn-rs/quinn/pull/2007 didn't pass fully the PR CI?


> Maybe we can rename the rustls dependency to something else (like, rustls-core or something), adapting the code, to avoid dep:rustls triggering the feature of the same name?

Great idea, I will try this out

stormshield-gt:
I misunderstood the zulip thread; it's the crate/feature syntax that activates both the optional dependency and the feature if there is one with the same name. Not the dep:crate one. We can just use the crate?/feature syntax here and it should be non-breaking.

I've added an edit to the PR description just in case

Ralith:
> Why the smoll and async-std are optional but not tokio? Some people, not us, might want to use the smoll and async-std runtime without compiling tokio.

We use some portable synchronization primitives provided by tokio. When the `runtime-tokio` feature is not enabled, then the runtime components of tokio will not be compiled, so such a hypothetical user needn't be concerned about wasted compile time or code size.

> Maybe it could be better to just use the runtime-smol everywhere

Seems reasonable.

Ralith (approved):

djc:
Looks great, thanks!

PR #2017, opened 2024-10-21
feat: introduce `UdpSocketState::try_send` APIs

thomaseizinger:
Resolves: #1971.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -190,8 +190,24 @@ impl UdpSocketState {
         })
     }
 
+    /// Sends a [`Transmit`] on the given socket.
+    ///
+    /// This function will only ever return errors of kind [`io::ErrorKind::WouldBlock`].
+    /// All other errors will be logged and converted to `Ok`.
+    ///
+    /// UDP transmission errors are considered non-fatal because higher-level protocols must
+    /// employ retransmits and timeouts anyway in order to deal with UDP's unreliable nature.
+    /// Thus, logging is most likely the only thing you can do with these errors.
     pub fn send(&self, socket: UdpSockRef<'_>, transmit: &Transmit<'_>) -> io::Result<()> {
-        send(self, socket.0, transmit)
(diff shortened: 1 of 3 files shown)


quinn-udp/src/unix.rs
```diff
@@ -191,7 +191,21 @@ impl UdpSocketState {
...
+        match send(self, socket.0, transmit) {
+            Ok(()) => Ok(()),
+            Err(e) if e.kind() == io::ErrorKind::WouldBlock => Err(e),
+            Err(e) => {
+                // Other errors are ignored, since they will usually be handled
+                // by higher level retransmits and timeouts.
+                // - PermissionDenied errors have been observed due to iptable rules.
+                //   Those are not fatal errors, since the
+                //   configuration can be dynamically changed.
+                // - Destination unreachable errors have been observed for other
```
djc, line 203: Feels like this should be a docstring on the public API?

quinn-udp/src/windows.rs
```diff
@@ -358,7 +360,13 @@ fn send(socket: UdpSockRef<'_>, transmit: &Transmit<'_>) -> i32 {
             ptr::null_mut(),
             None,
         )
+    };
+
+    if rc != 0 {
+        return Err(io::Error::last_os_error());
     }
+
+    Ok(())
```
djc: Nit: prefer spelling this like

```rust
match rc {
     0 => Ok(()),
     _ => Err(io::Error::last_os_error()),
}
```

djc (approved):
This looks reasonable to me, thanks!

thomaseizinger:
@djc Can the feedback be in follow-up commits or should I append?

djc:
> @djc Can the feedback be in follow-up commits or should I amend?

Please amend so that we end up with a clean commit series to review.

thomaseizinger:
Just fixed up CI, sorry for the hickup!

Ralith (approved):
Thanks! We should probably consolidate the public definitions somehow one of these days...

PR #2020, opened 2024-10-28
Fix poll transmit and min mtu

stormshield-gt:
This PR adds a test for post-quantum key exchange. For now the test is failing to demonstrate the presence of a bug; I will then propose a fix.

Note that I added a commit that exposed the key exchange group, like this was done for cipher in https://github.com/quinn-rs/quinn/pull/2001


quinn-proto/src/crypto/rustls.rs
```diff
@@ -263,6 +268,8 @@ pub struct HandshakeData {
     pub server_name: Option<String>,
     /// The ciphersuite negotiated with the peer
     pub negotiated_cipher_suite: CipherSuite,
+    /// The key exchange group negotiated with the peer
+    pub negociated_key_exchange_group: NamedGroup,
```
djc, line 272: `negociated` -> `negotiated`, please.

quinn/tests/post_quantum.rs
```diff
@@ -0,0 +1,105 @@
...
+use rustls::{
+    pki_types::{CertificateDer, PrivatePkcs8KeyDer},
+    NamedGroup,
+};
+
+use std::{
+    error::Error,
+    net::{Ipv4Addr, SocketAddr},
+    sync::Arc,
+};
```
djc, line 16: Please sort imports in `StdExternalCrate` blocks where both `quinn` and `proto` count as part of the crate.

quinn/tests/post_quantum.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    let (endpoint, server_cert) = make_server_endpoint(server_addr)?;
+    // accept a single connection
+    tokio::spawn(async move {
+        let incoming_conn = endpoint.accept().await.unwrap();
+        let conn = incoming_conn.await.unwrap();
+        println!(
+            "[server] connection accepted: addr={}",
+            conn.remote_address()
+        );
```
djc, line 31: Use tracing events instead of `println!()` (and set up the usual subscriber).

quinn/tests/post_quantum.rs
```diff
@@ -0,0 +1,105 @@
...
+    });
+
+    let endpoint = make_client_endpoint((Ipv4Addr::UNSPECIFIED, 0).into(), &[&server_cert])?;
+    // connect to server
+    let connection = endpoint
+        .connect(server_addr, "localhost")
+        .unwrap()
+        .await
+        .unwrap();
+    println!("[client] connected: addr={}", connection.remote_address());
```
djc, line 50: Tracing.

quinn-proto/src/crypto/rustls.rs
stormshield-gt: oups, thanks for spoting this

quinn/tests/post_quantum.rs
stormshield-gt: I've done this; I've also added the `with_test_writer` to the `subscriber` because I noticed otherwise the log weren't printed.

quinn/tests/post_quantum.rs
stormshield-gt: I've changed it. In general, do you prefer I close the thread or you do it?

quinn/tests/post_quantum.rs
stormshield-gt: thanks that's cleaner, I've done it

quinn/tests/post_quantum.rs
djc: If you're reasonably sure you've addressed the feedback, feel free to resolve the thread.

Ralith:
https://github.com/quinn-rs/quinn/pull/2027 gets this test passing.

stormshield-gt:
I've rebased and split into 2 tests, one for the `substract with overflow` error and one for the `debug_assert`

stormshield-gt:
The freebsd job seems to failed because `cmake` is not installed on the runner

quinn-proto/src/crypto/rustls.rs
```diff
@@ -11,7 +11,7 @@ use rustls::{
     client::danger::ServerCertVerifier,
     pki_types::{CertificateDer, PrivateKeyDer, ServerName},
     quic::{Connection, HeaderProtectionKey, KeyChange, PacketKey, Secrets, Suite, Version},
-    CipherSuite,
+    CipherSuite, NamedGroup,
```
djc, line 14: Tiny nit: can you also fix the "negociated" in the commit message?

quinn/tests/post_quantum.rs
```diff
@@ -0,0 +1,118 @@
...
+use rustls::{
+    pki_types::{CertificateDer, PrivatePkcs8KeyDer},
+    NamedGroup,
+};
+use std::{
+    error::Error,
+    net::{Ipv4Addr, SocketAddr},
+    sync::Arc,
+};
+use tracing::info;
```
djc, line 15: Nit: please use separate blocks for `std`, external, and `quinn` imports (in that order).

quinn/tests/post_quantum.rs
```diff
@@ -0,0 +1,118 @@
...
+    // Waiting for a stream will complete with an error when the server closes the connection
+    let _ = connection.accept_uni().await;
+
+    // Make sure the server has a chance to clean up
+    endpoint.wait_idle().await;
+}
+
+fn make_client_endpoint(
+    bind_addr: SocketAddr,
+    server_certs: &[&[u8]],
```
djc, line 75: Let's do a single `server_cert: CertificateDer<'_>` instead, to match what we actually need to pass in here?

djc:
> The freebsd job seems to failed because `cmake` is not installed on the runner

Why does this PR change the build requirements? Can you change the job to install `cmake`?

stormshield-gt:
That's because `rustls-post-quantum` is not optional and unconditionally brings `aws-lc-rs`. Sadly, it's not possible to have an optional dev dependency currently in `cargo`. 

I used a "private" feature hack to work around this, like they do, for instance, in the `reqwest` crate. Please let me know if you have a better solution.

quinn-proto/src/crypto/rustls.rs
```diff
@@ -69,6 +69,11 @@ impl crypto::Session for TlsSession {
                 .negotiated_cipher_suite()
                 .expect("cipher is negotiated")
                 .suite(),
+            negotiated_key_exchange_group: self
+                .inner
+                .negotiated_key_exchange_group()
+                .expect("key exchange group is negociated")
```
BiagioFesta, line 75: typo: `negociated`

stormshield-gt: Thanks for catching this

stormshield-gt:
I've discovered a new corner case with `min_mtu` of 1433,
with the current feature setup (that avoids compiling `aws-lc-rs` on freebsd and android), the command to test locally is

```rust
RUST_LOG=trace cargo t --test post_quantum -F __rustls-post-quantum-test -- --show-output
```

Ralith:
Fixed the new case in #2046.

stormshield-gt:
Thanks again for the fixes, I can confirm it works also on my computer

quinn-proto/src/crypto/rustls.rs
```diff
@@ -256,6 +261,8 @@ pub struct HandshakeData {
     ///
     /// Always `None` for outgoing connections
     pub server_name: Option<String>,
+    /// The key exchange group negotiated with the peer
+    pub negotiated_key_exchange_group: NamedGroup,
```
djc, line 265: Sorry, we don't want to land this on main right now due to semver compatibility issues.

stormshield-gt: Which option do you prefer: 
- Remove the commit that adds this. In that case I will only test that the handshake succeeds, not the specific key exchange group.
- Keep this PR open until the next breaking release

djc: First option sounds okay to me, I'll shortly reopen a PR to add back the API and then we can tack on a commit to add this back to the test as well?

stormshield-gt: @stormshield-damiend suggested, as we use a `__rustls-post-quantum-test` for testing, maybe I can declare this field under this feature flag?

djc: Ohh, that makes sense to me, yes!

stormshield-gt: I did this in my last push :tada:

quinn/tests/post_quantum.rs
```diff
@@ -0,0 +1,118 @@
...
+};
+use tracing::info;
+
+use quinn::{
+    crypto::rustls::{HandshakeData, QuicClientConfig, QuicServerConfig},
+    Endpoint,
+};
+
+#[tokio::test]
+async fn post_quantum_key_worse_case_header() {
```
djc, line 21: `worse_case` -> `worst_case`?

quinn/tests/post_quantum.rs
```diff
@@ -0,0 +1,118 @@
...
+    check_post_quantum_key_exchange(1433, 8082).await;
+}
+
+async fn check_post_quantum_key_exchange(min_mtu: u16, server_port: u16) {
+    let _ = tracing_subscriber::FmtSubscriber::builder()
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .with_test_writer()
+        .try_init();
+
+    let _ = rustls_post_quantum::provider().install_default();
```
djc, line 36: Would prefer we pass this into the `rustls::ClientConfig` and `rustls::ServerConfig` builders directly.

quinn/tests/post_quantum.rs
```diff
@@ -0,0 +1,118 @@
...
+    let server_addr = (Ipv4Addr::LOCALHOST, server_port).into();
```
djc, line 37: Nit: spell this as `SocketAddr::from((Ipv4Addr::LOCALHOST, server_port));`?

djc (approved):

djc:
Thanks!

djc:
@Ralith want to review this? Should be straightforward. @stormshield-gt up for a rebase?

stormshield-gt:
From what I understand, when activating `aws-lc-rs` feature of `rustls` 0.24, PQ key exchange would be the default thanks to https://github.com/rustls/rustls/pull/2552

For now I think I will use the `prefer-post-quantum` feature of `rustls` 0.23

djc:
> From what I understand, when activating `aws-lc-rs` feature of `rustls` 0.24, PQ key exchange would be the default thanks to [rustls/rustls#2552](https://github.com/rustls/rustls/pull/2552)
> 
> For now I think I will use the `prefer-post-quantum` feature of `rustls` 0.23

`prefer-post-quantum` was enabled by default in [0.23.27](https://github.com/rustls/rustls/releases/tag/v%2F0.23.27) so I think you don't need to use it here (since our `Cargo.lock` has a newer version already).

stormshield-gt:
Thanks for spotting this, I don't enable the feature explicitly anymore.

quinn/tests/post_quantum.rs
```diff
@@ -0,0 +1,126 @@
...
+use tracing::info;
+
+use quinn::{
+    Endpoint,
+    crypto::rustls::{HandshakeData, QuicClientConfig, QuicServerConfig},
+};
+
+#[tokio::test]
+async fn post_quantum_key_worst_case_header() {
+    check_post_quantum_key_exchange(1274, 8081).await;
```
Ralith: nit: Using hard-coded port numbers is prone to spurious failure, since that port may be in used by an unrelated service on the system. A more robust approach is to specify port 0 (wildcard) when binding a `UdpSocket`, then query the actual `local_addr` it ends up with, allowing the OS to allocate an arbitrary free port.

Ralith (approved):

Cargo.toml
```diff
@@ -42,6 +42,7 @@ rustls = { version = "0.23.5", default-features = false, features = ["std"] }
 rustls-pemfile = "2"
 rustls-platform-verifier = "0.6"
 rustls-pki-types = "1.7"
+rustls-post-quantum = "0.1"
```
Ralith: I'm a bit confused by the mechanism here. This dependency doesn't seem to actually be referenced by anything.

djc: Yeah, it should no longer be used AIUI.

stormshield-gt: Yep, I forgot to remote it when I used the `prefered-post-quantum` feature

quinn/tests/post_quantum.rs
stormshield-gt: Thanks, that's a lot better

PR #2023, opened 2024-10-29
fix(.github/codecov): Ignore aws-lc-rs-fips for codecov

dongcarl:
An attempt to fix https://github.com/quinn-rs/quinn/actions/runs/11562272211/job/32183071070 on `main` and (unreleased) `0.5.6`

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -14,7 +14,21 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
       - uses: taiki-e/install-action@cargo-llvm-cov
-      - run: cargo llvm-cov --all-features --workspace --lcov --output-path lcov.info
+      - shell: bash
+        # Run llvm-cov _without_ the "aws-lc-rs-fips" or "rustls-aws-lc-rs-fips" features, since
+        # they have complex build requirements:
+        #     https://github.com/aws/aws-lc/blob/3263ce2a553e4e917217fb487f8c6f488fcb1866/BUILDING.md#build-prerequisites
+        #
+        # This list of features was determined using:
+        #     cargo metadata --format-version 1 --no-deps \
+        #       | jq -r ' .packages[].features | keys[]' \
+        #       | sort -u \
+        #       | grep -vFx -e 'default' -e 'aws-lc-rs-fips' -e 'rustls-aws-lc-rs-fips' \
+        #       | paste -sd ',' -
+        run: |
+          cargo llvm-cov \
+            --features="arbitrary,async-io,async-std,aws-lc-rs,direct-log,fast-apple-datapath,futures-io,json-output,lock_tracking,log,platform-verifier,ring,runtime-async-std,runtime-smol,runtime-tokio,rustls,rustls-aws-lc-rs,rustls-log,rustls-ring,serde,serde_json,smol,tracing" \
+            --workspace --lcov --output-path lcov.info
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v5
         with:
--- a/perf/Cargo.toml
+++ b/perf/Cargo.toml
@@ -6,6 +6,8 @@ license = "MIT OR Apache-2.0"
 publish = false
 
 [features]
+# NOTE: Please keep this in sync with the feature list in `.github/workflows/codecov.yml`, see
+# comment in that file for more information.
 default = ["json-output"]
 # Allow for json output from the perf client
 json-output = ["serde", "serde_json"]
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -11,6 +11,8 @@ categories.workspace = true
 workspace = ".."
 
 [features]
+# NOTE: Please keep this in sync with the feature list in `.github/workflows/codecov.yml`, see
+# comment in that file for more information.
 default = ["rustls-ring", "log"]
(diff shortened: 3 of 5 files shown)


djc (commented):
What goal are you trying to achieve here?

dongcarl:
@djc Just to fix https://github.com/quinn-rs/quinn/actions/runs/11562272211/job/32183071070

dongcarl:
@djc Should be good to go now. Checks are passing: https://github.com/quinn-rs/quinn/actions/runs/12588265064

I can drop the last commit (meant to be temporary just for testing) if this looks good.

gretchenfrage (approved):
Yeah just copying what Rustls is doing makes sense, and the second commit which makes the coverage CI run in PRs combined with the fact that CI passed for this PR seems to indicate that the change works. Thank you for working on this irksome thing!

(**_We must remember not to merge until the second commit is removed._**)

.github/workflows/codecov.yml
```diff
@@ -14,6 +16,18 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
       - uses: taiki-e/install-action@cargo-llvm-cov
+      # From: https://github.com/rustls/rustls/blob/3751e24bbc1fef4f8d0bb3711a3be2a3d2a7b793/.github/workflows/build.yml#L59-L71
+      - name: "rustls: Install NASM for aws-lc-rs on Windows"
```
djc, line 20: I think this first one might not be needed anymore, since rustls no enables `prebuilt-nasm` on aws-lc-rs?

djc:
So looking at this more, maybe we should exclude `fips` from the features used for coverage? The feature has no effect on Quinn itself, so it should not change anything about coverage. The disadvantage would be that we sort of need to keep features in sync manually, but IMO that might be better than showing everything down for a relatively niche use case.

Ralith:
+1 in favor of excluding it, since our CI is getting pretty big already.

dongcarl:
Updated to include all features except for `aws-lc-rs-fips` and `rustls-aws-lc-rs-fips` by parsing `cargo metadata` with `jq`. Let me know if this looks good.

Note that the last commit needs to be dropped.

gretchenfrage (approved):
I think this makes sense, although I wish there was a way built in to cargo to do this. But it looks like there might not be.

.github/workflows/codecov.yml
```diff
@@ -14,7 +14,11 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
       - uses: taiki-e/install-action@cargo-llvm-cov
-      - run: cargo llvm-cov --all-features --workspace --lcov --output-path lcov.info
+      - shell: bash
+        run: |
+          LLVMCOV_FEATURES=$(cargo metadata --format-version 1 --no-deps \
+            | jq -r '[ .packages[].features | keys[] | select(. != "default" and . != "aws-lc-rs-fips" and . != "rustls-aws-lc-rs-fips") ] | unique | join(",")')
```
djc, line 22: Can we just get the flattened version of this, without all the shell/jq magic? That stuff is unfamiliar (at least to me) and thus harder to maintain, and I don't think we change our features all that often.

dongcarl: Happy to do so. I could also make it so that `jq` only minimally does `.packages[].features | keys[]`, and the rest is done using `grep -v`, `sort -u` and `paste -sd` if that's clearer.

Let me know what you decide.

djc: I think all of that is still more complexity than is warranted here.

.github/workflows/codecov.yml
```diff
@@ -14,7 +16,11 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
       - uses: taiki-e/install-action@cargo-llvm-cov
-      - run: cargo llvm-cov --all-features --workspace --lcov --output-path lcov.info
+      - shell: bash
+        run: |
+          cargo llvm-cov \
```
djc, line 21: Can you add a comment here to explain how you arrived at the list of features, and a comment on the `[features]` sections in the relevant `Cargo.toml` files with a reminder to keep this in sync?

djc (approved):
Thanks! Looks ugly but hopefully not too hard to maintain.

.github/workflows/codecov.yml
dongcarl: Should I include my `cargo metadata` invocation in the comment in this file or no?

djc: No, I don't think that's necessary -- making changes should mostly be pretty straightforward, I guess?

dongcarl: What I mean is, you said I should "explain how you arrived at the list of features", and that invocation is how I did it.

So I'm wondering what comment you're looking for in this particular file?

djc: Ah, sure, let's include your `cargo metadata` invocation as a comment.

dongcarl: All done, let me know if this is what you wanted!

gretchenfrage: This looks fine to me.

gretchenfrage:
I believe @djc's final thread was properly resolved. I'm going to try and remove the final commit and merge this now.

gretchenfrage:
Thank you for dealing with this!

Issue #2018, opened 2024-10-24
Notifying Quinn about network path changes


flub:
Sometimes the application knows about network path changes that Quinn may not detect immediately to.  If the network changed significantly it can often be faster to reach optimal throughput and latency again by resetting things like the congestion controller, rtt estimator and mtu detection to initial states.

This was explored a bit in #1842 but that was ultimately closed.  However this topic was recently brought up again in another issue and there was an indication that something like this could belong in Quinn.  So this is an attempt to find out if we can get any closer to that.

It seems the biggest difference in thinking is whether this should be per `Connection` or for the entire `Endpoint`.  The motivation for per-connection is that these details are stored per-connection since each connection has its own network path.  While for the entire endpoint the argument is that an endpoint only has a single socket and this is probably the most common way for an application to know the network path changed.

Perhaps it is easier to agree on what the API and implementation in quinn-proto should be, because there an API on the `Connection` seems better suited.  Currently we have https://github.com/n0-computer/quinn/blob/02f3b33de039951bfaf4cc383d9f0ec6cdf8dcb1/quinn-proto/src/connection/mod.rs#L1317 as the entrypoint to this in the fork for iroh.  The [implementation](https://github.com/n0-computer/quinn/blob/02f3b33de039951bfaf4cc383d9f0ec6cdf8dcb1/quinn-proto/src/connection/paths.rs#L121) currently does 3 things:

- Creates a new `RttEstimator` from the original `TransportConfig` settings.
- Creates a new congestion controller from the original `TransportConfig` settings.
- [Resets the MTU discovery](https://github.com/n0-computer/quinn/blob/02f3b33de039951bfaf4cc383d9f0ec6cdf8dcb1/quinn-proto/src/connection/mtud.rs#L59), essentially re-creating a new one with the same settings it was originally created with.

What is the opinion on adding this in some form to quinn-proto?

djc:
It seems reasonable to me, let's see what @Ralith thinks.

Ralith:
Thanks for the summary, and for digging up the old discussion! The proposal sounds good to me. Even if we only wanted to expose this on the endpoint, `quinn` would still need to iterate over all connections and reset those states, which is exactly what the proposed interface allows.

> whether this should be per Connection or for the entire Endpoint.

[...]

PR #2025, opened 2024-10-31
feat(quinn-proto): Allow notifying of network path changes

flub:
This gives the proto state the ability to be externally notified about network path changes it might not be aware off.  In this case it will re-set the RTT, Congestion Controller and MTU settings to the initial states based on the TransportConfig.

This addresses the quinn-proto part of #2018.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1314,6 +1314,20 @@ impl Connection {
         self.path.congestion.as_ref()
     }
 
+    /// Resets path-specific settings.
+    ///
+    /// This will force-reset several subsystems related to a specific network path.
+    /// Currently this is the congestion controller, round-trip estimator, and the MTU
+    /// discovery.
+    ///
+    /// This is useful when it is known the underlying network path has changed and the old
+    /// state of these subsystems is no longer valid or optimal. In this case it might be
+    /// faster or reduce loss to settle on optimal values by restarting from the initial
+    /// configuration in the [`TransportConfig`].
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -1309,6 +1309,20 @@ impl Connection {
...
     }
 
+    /// Resets path-specific settings.
+    ///
+    /// This will force-reset several subsystems related to a specific network path.
+    /// Currently this is the congestion controller, round-trip estimator, and the MTU
+    /// discovery.
+    ///
+    /// This is useful when it is known the underlying network path has changed and the old
+    /// state of these subsystems is no longer valid or optimal. In which case it might be
```
Ralith, line 1319:
```suggestion
    /// state of these subsystems is no longer valid or optimal, in which case it might be
```

Ralith (commented):
Thanks!

quinn-proto/src/connection/mod.rs
flub: This creates a very long sentence, I'm used to technical writing style guides recommending shorter sentences.  I've reworded this slightly differently, hopefully also improving this sentence, but feel free to ask for more tweaks.

flub:
I'll auto-squash the commits once it gets an LGTM.  In the meantime it's easier to track changes by using fixup commits.

Ralith (approved):
LGTM, thanks!

flub:
thanks, rebased on main and squashed

djc (approved):

PR #2030, opened 2024-11-09
Feature flag for rustls logging

crazyscot:
I needed to dig into what was going on within rustls. Somebody else might find this useful (and I might need it again in future).
This feature flag also works with `tracing_subscriber`.

### DIFF
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -30,6 +30,8 @@ ring = ["dep:ring"]
 platform-verifier = ["dep:rustls-platform-verifier"]
 # Configure `tracing` to log events via `log` if no `tracing` subscriber exists.
 log = ["tracing/log"]
+# Enable rustls logging
+rustls-log = ["rustls?/logging"]
 
 [dependencies]
 arbitrary = { workspace = true, optional = true }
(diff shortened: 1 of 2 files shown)


djc (approved):

quinn-proto/Cargo.toml
```diff
@@ -30,6 +30,8 @@ ring = ["dep:ring"]
 platform-verifier = ["dep:rustls-platform-verifier"]
 # Configure `tracing` to log events via `log` if no `tracing` subscriber exists.
 log = ["tracing/log"]
+# Enable rustls logging
+rustls-log = ["rustls/logging"]
```
Ralith, line 34: Should this be `rustls?/logging`?

Ralith (approved):

quinn-proto/Cargo.toml
crazyscot: Good catch! Thanks, I'm still pretty new to Rust. Updated.

Ralith: Seems like you only updated one of them?

crazyscot: So I did. Apologies. This time, for sure?

PR #2031, opened 2024-11-09
Fix missing re-exports

### LINKED ISSUE #2012: Some types appear in `quinn`'s public API which are not accessible through `quinn` directly
One example of this was noted in #1757:

> - quinn does not re-export `quinn_proto::TransportError` and other downstream-API-exposed types despite that being a public part of quinn's API via [quinn's re-exporting of `quinn_proto::ConnectionError`](https://docs.rs/quinn/latest/quinn/enum.ConnectionError.html)

Another example is that [`StreamId`](https://docs.rs/quinn/latest/quinn/struct.StreamId.html)'s public API involves `Side` and `Dir`, but `Side` and `Dir` cannot be referenced without adding `quinn_proto` in addition to `quinn` as a dependency.

If this is indeed something we wish to avoid, it would be nice to have some automated tool which could detect such cases.

gretchenfrage:
Closes #2012 

In that issue, we noted that there doesn't seem to exists any automated tooling to automatically detect types from external crates which are visible in a crate's public API but which the crate does not re-export. So I went ahead and attempted to create such a tool: [[cargo should-be-public-checker]](https://github.com/gretchenfrage/cargo-should-be-public-checker). It's still pretty crude right now but it was enough to detect a handful of items that I confirmed were missing reexports in Quinn and proto. That said, it's possible that there are some things it missed.

should-be-public-checker output for quinn-proto (on main):

```
visible but not importable:
- quinn_proto::Chunk::bytes::Bytes
- quinn_proto::ClientConfig::with_root_certificates::RootCertStore
- quinn_proto::ClientConfig::with_root_certificates::VerifierBuilderError
- quinn_proto::ConnectionClose::frame_type::Type
- quinn_proto::ConnectionHandle::`<_ as Index<Some(AngleBracketed { args: [Type(ResolvedPath(Path { name: "ConnectionHandle", id: Id(4627), args: Some(AngleBracketed { args: [], constraints: [] }) }))], constraints: [] })>>`::Output::ConnectionMeta
- quinn_proto::ConnectionId::from_buf::Buf
- quinn_proto::PartialDecode::new::BytesMut
- quinn_proto::ServerConfig::with_single_cert::CertificateDer
- quinn_proto::ServerConfig::with_single_cert::PrivateKeyDer
- quinn_proto::crypto::rustls::HandshakeData::negotiated_cipher_suite::CipherSuite
- quinn_proto::crypto::rustls::QuicClientConfig::with_initial::ClientConfig
- quinn_proto::crypto::rustls::QuicServerConfig::with_initial::ServerConfig
- quinn_proto::crypto::rustls::QuicServerConfig::with_initial::Suite
- quinn_proto::crypto::rustls::TlsSession::`<_ as VZip<Some(AngleBracketed { args: [Type(Generic("V"))], constraints: [] })>>`::MultiLane
- quinn_proto::transport_parameters::TransportParameters::write::BufMut
```

should-be-public-checker output for quinn (on main):

```
visible but not importable:
- quinn::ApplicationClose::reason::Bytes
- quinn::ClientConfig::initial_dst_cid_provider::ConnectionId
- quinn::ClientConfig::initial_dst_cid_provider::ConnectionId::from_buf::Buf
- quinn::ConnectionClose::error_code::Code
- quinn::ConnectionClose::frame_type::Type
- quinn::ConnectionStats::frame_tx::FrameStats
- quinn::ConnectionStats::path::PathStats
- quinn::ConnectionStats::udp_tx::UdpStats
- quinn::EndpointConfig::cid_generator::ConnectionIdGenerator
- quinn::SendStream::write_chunks::Written
- quinn::StreamId::new::Dir
- quinn::StreamId::new::Side
- quinn::Transmit::ecn::EcnCodepoint
- quinn::VarInt::from_u64::VarIntBoundsExceeded
- quinn::crypto::CryptoError::`<_ as VZip<Some(AngleBracketed { args: [Type(Generic("V"))], constraints: [] })>>`::MultiLane
- quinn::rustls::ConfigSide::Sealed
- quinn::rustls::RootCertStore::roots::TrustAnchor
- quinn::rustls::client::EchConfig::new::EchConfigListBytes
- quinn::rustls::client::ServerCertVerifierBuilder::with_crls::CertificateRevocationListDer
- quinn::rustls::client::Tls13ClientSessionValue::`<_ as Deref<Some(AngleBracketed { args: [], constraints: [] })>>`::Target::ClientSessionCommon
- quinn::rustls::client::verify_server_cert_signed_by_trust_anchor::CertificateDer
- quinn::rustls::client::verify_server_cert_signed_by_trust_anchor::SignatureVerificationAlgorithm
- quinn::rustls::client::verify_server_cert_signed_by_trust_anchor::UnixTime
- quinn::rustls::client::verify_server_name::ServerName
- quinn::rustls::crypto::cipher::OutboundOpaqueMessage::read::MessageError
- quinn::rustls::crypto::cipher::OutboundOpaqueMessage::read::Reader
- quinn::rustls::crypto::cipher::PlainMessage::payload::Payload
- quinn::rustls::crypto::hpke::HpkeSuite::kem::HpkeKem
- quinn::rustls::crypto::hpke::HpkeSuite::sym::HpkeSymmetricCipherSuite
- quinn::rustls::crypto::hpke::HpkeSuite::sym::HpkeSymmetricCipherSuite::aead_id::HpkeAead
- quinn::rustls::crypto::hpke::HpkeSuite::sym::HpkeSymmetricCipherSuite::kdf_id::HpkeKdf
- quinn::rustls::crypto::ring::sign::any_ecdsa_type::PrivateKeyDer
- quinn::rustls::crypto::ring::sign::any_eddsa_type::PrivatePkcs8KeyDer
- quinn::rustls::crypto::verify_tls13_signature_with_raw_key::SubjectPublicKeyInfoDer
- quinn::rustls::server::ClientHello::server_cert_types::CertificateType
- quinn::rustls::unbuffered::TransmitTlsData::may_encrypt_early_data::MayEncryptEarlyData
```

Summary of changes:

- `quinn_proto::frame::Type` becomes public. So, in addition to simply re-exporting it:
  - Renamed it to `FrameType`
  - Added doc comment
  - Made its associated constants `pub(crate)` rather than `pub`
- In terms of crate re-exports: Quinn already re-exports `rustls` and `udp`. Following this trend, I also added re-export of `rustls` to proto, and re-export of `bytes` to both.
- I changed crate re-exports from `pub use` to `pub extern crate`. I'm fine changing this back if you like, but I think this is better because it's more clear that it's more clear that it's a crate being re-exported.
- Other than that, simply `pub use`d all the missing items. 

Of course, fixing the missing re-exports in `rustls` is out of scope, and the thing about `MultiLane` is some irrelevant blanket-impl.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -501,7 +501,7 @@ impl Connection {
                     self,
                 )?;
                 trace!("validating previous path with PATH_CHALLENGE {:08x}", token);
-                buf.write(frame::Type::PATH_CHALLENGE);
+                buf.write(frame::FrameType::PATH_CHALLENGE);
                 buf.write(token);
                 self.stats.frame_tx.path_challenge += 1;
 
@@ -883,7 +883,7 @@ impl Connection {
                     // above.
                     let mut builder = builder_storage.take().unwrap();
                     trace!("PATH_RESPONSE {:08x} (off-path)", token);
-                    buf.write(frame::Type::PATH_RESPONSE);
+                    buf.write(frame::FrameType::PATH_RESPONSE);
                     buf.write(token);
                     self.stats.frame_tx.path_response += 1;
                     builder.pad_to(MIN_INITIAL_SIZE);
@@ -979,12 +979,12 @@ impl Connection {
             )?;
 
             // We implement MTU probes as ping packets padded up to the probe size
-            buf.write(frame::Type::PING);
+            buf.write(frame::FrameType::PING);
             self.stats.frame_tx.ping += 1;
 
             // If supported by the peer, we want no delays to the probe's ACK
             if self.peer_supports_ack_frequency() {
-                buf.write(frame::Type::IMMEDIATE_ACK);
+                buf.write(frame::FrameType::IMMEDIATE_ACK);
                 self.stats.frame_tx.immediate_ack += 1;
             }
 
@@ -3037,7 +3037,7 @@ impl Connection {
 
         // HANDSHAKE_DONE
         if !is_0rtt && mem::replace(&mut space.pending.handshake_done, false) {
-            buf.write(frame::Type::HANDSHAKE_DONE);
+            buf.write(frame::FrameType::HANDSHAKE_DONE);
             sent.retransmits.get_or_create().handshake_done = true;
             // This is just a u8 counter and the frame is typically just sent once
             self.stats.frame_tx.handshake_done =
@@ -3047,7 +3047,7 @@ impl Connection {
         // PING
         if mem::replace(&mut space.ping_pending, false) {
             trace!("PING");
-            buf.write(frame::Type::PING);
+            buf.write(frame::FrameType::PING);
             sent.non_retransmits = true;
             self.stats.frame_tx.ping += 1;
         }
@@ -3055,7 +3055,7 @@ impl Connection {
         // IMMEDIATE_ACK
         if mem::replace(&mut space.immediate_ack_pending, false) {
             trace!("IMMEDIATE_ACK");
-            buf.write(frame::Type::IMMEDIATE_ACK);
+            buf.write(frame::FrameType::IMMEDIATE_ACK);
             sent.non_retransmits = true;
             self.stats.frame_tx.immediate_ack += 1;
         }
@@ -3111,7 +3111,7 @@ impl Connection {
                 sent.non_retransmits = true;
                 sent.requires_padding = true;
                 trace!("PATH_CHALLENGE {:08x}", token);
-                buf.write(frame::Type::PATH_CHALLENGE);
+                buf.write(frame::FrameType::PATH_CHALLENGE);
                 buf.write(token);
                 self.stats.frame_tx.path_challenge += 1;
             }
@@ -3123,7 +3123,7 @@ impl Connection {
                 sent.non_retransmits = true;
                 sent.requires_padding = true;
                 trace!("PATH_RESPONSE {:08x}", token);
-                buf.write(frame::Type::PATH_RESPONSE);
+                buf.write(frame::FrameType::PATH_RESPONSE);
                 buf.write(token);
                 self.stats.frame_tx.path_response += 1;
             }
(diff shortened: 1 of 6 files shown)


djc:
> * I changed crate re-exports from `pub use` to `pub extern crate`. I'm fine changing this back if you like, but I think this is better because it's more clear that it's more clear that it's a crate being re-exported.

This definitely does not seem like an improvement, IMO `extern crate` should basically never be used in modern Rust, and I think it's plenty clear enough that these are crate re-exports. If not, just add some comments.

I don't think we should re-export bytes. It is a 1.0 dependency that IMO is pretty unlikely to ever get a semver-incompatible bump, so there's not the same benefit like with rustls that matching the used version is hard.

My remaining question here is, are these changes semver-compatible? (In particular, the renaming of `frame::Type`.)

gretchenfrage:
> If not, just add some comments.

Unfortunately, doc comments on re-export items doesn't seem to be working with rustdoc? It's fine, though.

gretchenfrage:
:)

- Removed reexport of `bytes`.
- Changed `pub extern crate` back to `pub use`.

> My remaining question here is, are these changes semver-compatible? (In particular, the renaming of `frame::Type`.)

`quinn_proto::frame::Type` is not importable on main. the `struct Type` is `pub`, but the `mod frame` is default visibility, and `Type` is never re-exported. You can confirm that in [the current docs.rs for eg. `ConnectionClose`](https://docs.rs/quinn/latest/quinn/struct.ConnectionClose.html), the `Type` in `frame_type: Option<Type>` is not a link, and `Type` cannot be found by searching. Therefore, renaming it should not cause semver problems.

The only changes other than that rename are adding new re-exports. I don't imagine this is something that's considered to cause semver problems.

It may be hard to tell at a glance because `cargo fmt` alphabetizes it, but the new re-exports added to Quinn in [the last commit](https://github.com/quinn-rs/quinn/pull/2031/commits/2a9450ca8b6d38f5a1cb93750ff02cecec6bc40d) are:

- `quinn_proto::Dir`
- `quinn_proto::Side`
- `quinn_proto::ConnectionId`
- `quinn_proto::Written`
- `quinn_proto::TransportErrorCode`
- `quinn_proto::FrameStats`
- `quinn_proto::VarIntBoundsExceeded`
- `quinn_proto::PathStats`
- `quinn_proto::UdpStats`
- `quinn_proto::ConnectionIdGenerator`
- `quinn_proto::EcnCodepoint`
- `quinn_proto::FrameType`

djc (approved):
Thanks, this looks good to me!

Ralith (approved):
Thanks! Would be cool to have this running in CI.

gretchenfrage:
> (If you want to do a similar pass for rustls, that would be much appreciated!)

> Thanks! Would be cool to have this running in CI.

Appreciate the positive reception! I'll do both at the point that I get around to making the tool a bit better than it is now.

PR #2037, opened 2024-11-12
feat(quinn,quinn-udp): Avoid `socket2` and `std::net::UdpSocket` dependencies in `wasm32-unknown-unknown` target

matheus23:
## Changes

- Adds a `cfg_alias` `wasm_browser` for `all(target_family = "wasm", target_os = "unknown")` (matching `wasm*-*-unknown`)
- Makes `socket2` a target-dependent dependency
- Code depending on `std::net::UdpSocket` (e.g. `Runtime::wrap_udp_socket` or `Endpoint::rebind`) becomes gated as `not(wasm_browser)`
- adds `web_time` as target-dependent dependency in `quinn` to replace `std::time::{Instant, Duration}`, which panic in Wasm otherwise.

## Testing

- These changes were verified outside this PR with a custom `Runtime` implementation in the browser and nodejs.
  I'm not contributing this test (yet?) as discussed.
- The "test (wasm32-unknown-unknown)" github action job was adjusted with a check making sure there's no `(import "env" ...)` in the `quinn` `.wasm` file when built.
  This is a good, even if not quite sufficient effort in making sure the Wasm build doesn't break.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -144,25 +144,14 @@ jobs:
     name: test wasm32-unknown-unknown
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout sources
-        uses: actions/checkout@v4
-
-      - name: Install stable toolchain
-        uses: dtolnay/rust-toolchain@stable
-
-      - name: Add wasm target
-        run: rustup target add wasm32-unknown-unknown
-
-      - name: Install nodejs v20
-        uses: actions/setup-node@v4
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - run: rustup target add wasm32-unknown-unknown
+      - uses: actions/setup-node@v4
         with:
           node-version: 20
-
-      - name: Setup `wasm-tools`
-        uses: bytecodealliance/actions/wasm-tools/setup@v1
-
-      - name: Install cargo binstall
-        uses: cargo-bins/cargo-binstall@main
+      - uses: bytecodealliance/actions/wasm-tools/setup@v1
+      - uses: cargo-bins/cargo-binstall@main
 
       - run: cargo test -p quinn-proto --target wasm32-unknown-unknown --no-run
       - run: cargo check -p quinn-udp --target wasm32-unknown-unknown --no-default-features --features=tracing,log
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -164,20 +164,21 @@ jobs:
       - name: Install cargo binstall
         uses: cargo-bins/cargo-binstall@main
 
-      - name: build wasm32 tests (quinn-proto)
-        run: cargo test -p quinn-proto --target wasm32-unknown-unknown --no-run
+      - run: cargo test -p quinn-proto --target wasm32-unknown-unknown --no-run
+      - run: cargo check -p quinn-udp --target wasm32-unknown-unknown --no-default-features --features=tracing,log
+      - run: cargo rustc -p quinn --target wasm32-unknown-unknown --no-default-features --features=log,platform-verifier,rustls-ring --crate-type=cdylib
 
       # If the Wasm file contains any 'import "env"' declarations, then
       # some non-Wasm-compatible code made it into the final code.
-      - name: Check for 'import "env"' in Wasm
+      - name: Ensure no 'import "env"' in quinn_proto Wasm
         run: |
           ! wasm-tools print --skeleton target/wasm32-unknown-unknown/debug/deps/quinn_proto-*.wasm | grep 'import "env"'
+      - name: Ensure no 'import "env"' in quinn Wasm
+        run: |
+          ! wasm-tools print --skeleton target/wasm32-unknown-unknown/debug/quinn.wasm | grep 'import "env"'
 
-      - name: Install wasm-bindgen-test-runner
-        run: cargo binstall wasm-bindgen-cli --locked --no-confirm
-
-      - name: wasm32 test (quinn-proto)
-        run: cargo test -p quinn-proto --target wasm32-unknown-unknown
+      - run: cargo binstall wasm-bindgen-cli --locked --no-confirm
+      - run: cargo test -p quinn-proto --target wasm32-unknown-unknown
 
   msrv:
     runs-on: ubuntu-latest
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -282,4 +282,4 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
       - uses: taiki-e/install-action@cargo-hack
-      - run: cargo hack check --feature-powerset --optional-deps --no-dev-deps --ignore-unknown-features --ignore-private --group-features runtime-async-std,async-io,async-std --group-features runtime-smol,async-io,smol --skip "${{env.SKIP_FEATURES}}"
+      - run: cargo hack check --feature-powerset --optional-deps --clean-per-run --no-dev-deps --ignore-unknown-features --ignore-private --group-features runtime-async-std,async-io,async-std --group-features runtime-smol,async-io,smol --skip "${{env.SKIP_FEATURES}}"
--- a/quinn-udp/Cargo.toml
+++ b/quinn-udp/Cargo.toml
@@ -21,8 +21,10 @@ fast-apple-datapath = []
 [dependencies]
 libc = "0.2.158"
 log = { workspace = true, optional = true }
-socket2 = { workspace = true }
 tracing = { workspace = true, optional = true }
+
+[target.'cfg(not(all(target_family = "wasm", target_os = "unknown")))'.dependencies]
+socket2 = { workspace = true }
 
 [target.'cfg(windows)'.dependencies]
 once_cell = { workspace = true }
@@ -33,7 +35,7 @@ criterion = { version = "0.5", default-features = false, features = ["async_toki
 tokio = { workspace = true, features = ["rt", "rt-multi-thread", "net"] }
 
 [build-dependencies]
-cfg_aliases = "0.2"
+cfg_aliases = { workspace = true }
 
 [lib]
 # See https://github.com/bheisler/criterion.rs/blob/master/book/src/faq.md#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options
--- a/quinn-udp/src/lib.rs
+++ b/quinn-udp/src/lib.rs
@@ -27,12 +27,13 @@
 #![warn(unreachable_pub)]
 #![warn(clippy::use_self)]
 
+use std::net::{IpAddr, Ipv6Addr, SocketAddr};
 #[cfg(unix)]
 use std::os::unix::io::AsFd;
 #[cfg(windows)]
 use std::os::windows::io::AsSocket;
+#[cfg(not(wasm_browser))]
 use std::{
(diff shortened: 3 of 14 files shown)


.github/workflows/rust.yml
```diff
@@ -73,45 +73,35 @@ jobs:
...
-
-      - name: Install stable toolchain
-        uses: dtolnay/rust-toolchain@stable
-
-      - name: Add wasm target
-        run: rustup target add wasm32-unknown-unknown
-
-      - name: Install nodejs v20
-        uses: actions/setup-node@v4
+      - uses: actions/checkout@v4
```
djc, line 76: What's the rationale for removing all the step names here? Would prefer to split these cosmetic/superficial changes into a separate commit.

quinn/src/runtime.rs
```diff
@@ -6,18 +6,20 @@ use std::{
...
+use crate::Instant;
+
 /// Abstracts I/O and timer operations for runtime independence
 pub trait Runtime: Send + Sync + Debug + 'static {
     /// Construct a timer that will expire at `i`
     fn new_timer(&self, i: Instant) -> Pin<Box<dyn AsyncTimer>>;
     /// Drive `future` to completion in the background
     fn spawn(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>);
     /// Convert `t` into the socket type used by this runtime
+    #[cfg(feature = "net")]
```
djc, line 22: Hmm, I feel like trait methods that appear/disappear based on a Cargo feature aren't great. Are you still using the `Runtime` trait in your WASM setup?

quinn/src/tests.rs
```diff
@@ -160,7 +158,7 @@ fn read_after_close() {
             .unwrap()
             .await
             .expect("connect");
-        tokio::time::sleep_until(Instant::now() + Duration::from_millis(100)).await;
+        tokio::time::sleep(Duration::from_millis(100)).await;
```
djc, line 161: This seems completely unrelated, ideally pull it out into a separate commit?

quinn-udp/src/lib.rs
```diff
@@ -27,29 +27,35 @@
 #![warn(unreachable_pub)]
 #![warn(clippy::use_self)]
 
-#[cfg(unix)]
+use std::net::{IpAddr, Ipv6Addr, SocketAddr};
```
djc, line 30: Does it still make sense to depend on quinn-udp in the WASM world?

matheus23:
I realized this PR is missing some rationale, so I'm writing this on the top level for better visibility:

> Does it still make sense to depend on quinn-udp in the WASM world?

This was one of the first things I was thinking initially, too (and asked about in the crates discord).

The only definitions that `quinn` without the `net` feature needs from `quinn-udp` now are `BATCH_SIZE`, `RecvMeta`, `EcnCodepoint` and `Transmit`.

The fundamental problem IMO is that `AsyncUdpSocket` is tied to `quinn` and its `Runtime`, but then `AsyncUdpSocket` depends on e.g. `RecvMeta`. `quinn-udp` and `quinn` are thus tightly coupled at the moment.

The solution would be to separate out the type definitions of `RecvMeta`, etc. into another crate that both `quinn-udp` and `quinn` depend on. I didn't want to do this as that's quite an invasive refactor.
It is *not* possible to move `RecvMeta` etc. to `quinn`, because then `quinn-udp` would like to depend on `quinn` and the other way around.

> Are you still using the `Runtime` trait in your WASM setup?

Yes, there's no way of using `quinn` without providing a `Runtime` (e.g. the `Endpoint` keeps a reference to it).

The alternative would be integrating with `quinn-proto` only, which basically means reimplementing `quinn`s connection and endpoint drivers, timers, API, etc. That's practically not an option.

quinn/src/runtime.rs
matheus23: I've gone back and forth between having this gated on a feature and having this gated on the `wasm32-unknown-unknown` target.

Another option is to keep this method but then error out in all cases if invoked. This looks clean on the `quinn` side, but is error prone when e.g. some new `quinn` contribution suddenly makes use of the `wrap_udp_socket` fn outside of the `net` feature.

.github/workflows/rust.yml
matheus23: 1. I introduced the step names in the old PR but don't really think they're helping much.
2. All other jobs don't use step names. This is more consistent.

djc (approved):

djc (approved):
Okay, I think this all makes sense. Thanks!

quinn-udp/Cargo.toml
```diff
@@ -14,23 +14,27 @@ workspace = ".."
...
 
 [features]
-default = ["tracing", "log"]
+default = ["tracing", "log", "net"]
 # Configure `tracing` to log events via `log` if no `tracing` subscriber exists.
 log = ["tracing/log"]
 direct-log = ["dep:log"]
 # Use private Apple APIs to send multiple packets in a single syscall.
 fast-apple-datapath = []
+net = ["dep:socket2"]
```
dignifiedquire, line 23: maybe add a short description of what exactly the feature does?

djc:
@Ralith ping?

quinn-udp/src/lib.rs
mxinden, line 1: Instead of the many `cfg(feature = "net")`, wouldn't something along the following be simpler?

``` rust
#[cfg(feature = "net")]
use net::*;

#[cfg(feature = "net")]
mod net {
  // Everything that was previously feature flagged behind net.
}
```

mxinden (commented):
I understand the rational above for importing `quinn-udp` in a WASM context.

That said, for what my opinion is worth here, I can't judge whether it is worth the complexity it introduces. Will defer to the actual Quinn maintainers.

quinn-udp/src/lib.rs
matheus23: I've got a version where I've experimented with that.
It's a much more invasive change, I'm not sure if it's much better?
I can add it to this PR if other people agree: https://github.com/n0-computer/quinn/pull/23/commits/d06b71b9772bb828bf02b21796825a2ba6b7887b

mxinden: Thanks for the commit. I didn't think of the fact that `unix.rs` and `windows.rs` would need to be under the `net/` module.

At this point I don't have a better idea, nor a preference for the options listed here.

Ralith:
You mentioned that you also considered gating on target instead of a cargo feature. Can you elaborate on why you went with the feature instead? I think exposing a subset API on some targets is less hazardous than a cargo feature, particularly as this is technically a semver break affecting users who have `default-features = false` and use only a custom runtime.

quinn-udp/src/lib.rs
RealAstolfo: couldnt windows and unix stuff be stuffed under a "system" "sys" or something like that.

matheus23:
Main reason was that I noticed this was similar to e.g. tokio's `net` feature (something you disable when you build for Wasm similarly).
The `cfg`-gating also gets a little more complex, e.g:
```rs
#[cfg(all(not(all(target_family = "wasm", target_os = "unknown")), unix))]
use std::os::unix::io::AsFd;
```

But I see that you have cfg-aliases enabled for quinn-udp, so I'll try it with that.

matheus23:
I've updated the PR to use a cfg alias `wasm_browser`, which stands for `all(target_family = "wasm", target_os = "unknown")`.
With this, a bunch of cfgs get simpler. E.g. `all(not(wasm_browser), unix)` simplifies to just `unix`, since `wasm_browser` and `unix` are exclusive.

quinn-udp/src/lib.rs
matheus23: The newest version that avoids the feature in favor of a cfg alias `wasm_browser` makes lots of these cfgs simpler, as that cfg is exclusive with unix and windows cfgs.
Please take a look at the latest diff @mxinden I think you'll like it more.

matheus23:
The CI failed with:

> `/usr/bin/ld: final link failed: No space left on device`

gretchenfrage:
> The CI failed with:
> 
> > `/usr/bin/ld: final link failed: No space left on device`

[Indeed](https://github.com/quinn-rs/quinn/actions/runs/12596988919/job/35135265022?pr=2037)... Re-running it, it would be convenient if it's spurious but exciting if it's not.

gretchenfrage:
@matheus23 I know you saw this on Discord, but could you add [this commit](https://github.com/quinn-rs/quinn/pull/2124/commits/bc58859de84088a49a67f7834ac836bccf583c5d) to the PR by whatever means you find most convenient to make the CI pass?

gretchenfrage:
> The only definitions that `quinn` without the `net` feature needs from `quinn-udp` now are `BATCH_SIZE`, `RecvMeta`, `EcnCodepoint` and `Transmit`.
> 
> The fundamental problem IMO is that `AsyncUdpSocket` is tied to `quinn` and its `Runtime`, but then `AsyncUdpSocket` depends on e.g. `RecvMeta`. `quinn-udp` and `quinn` are thus tightly coupled at the moment.
> 
> The solution would be to separate out the type definitions of `RecvMeta`, etc. into another crate that both `quinn-udp` and `quinn` depend on. I didn't want to do this as that's quite an invasive refactor. It is _not_ possible to move `RecvMeta` etc. to `quinn`, because then `quinn-udp` would like to depend on `quinn` and the other way around.

To be honest, I'm not sure this feature gating approach is the best solution. I find it hard to follow, and I worry about our ability to remember why it is that way and what we need to do to keep respecting the needs of the WASM environment in the same way in the future.

If the problem is that a user might need to depend on these items that are used in both `quinn` and `quinn-udp`, while not having `quinn-udp` in the dependency tree (only having `quinn` in the dependency tree), I see 2 alternatives to feature gating as such:

- As you suggest, factor out some common dependency crate between `quinn` and `quinn-udp`. "Cleanest", but I understand the hesitation.
- Create duplicate versions of these items in `quinn` (e.g. essentially copy/paste `quinn_udp::RecvMeta` into somewhere in `quinn`), so that `AsyncUdpSocket` only references the versions defined in Quinn. Implement `From`/`Into` for conversion between them (little bit of boilerplate), and have `quinn::runtime::{tokio, async_io}` use those conversions.

@ralith @djc @mxinden thoughts on either of these suggestions?

djc:
> To be honest, I'm not sure this feature gating approach is the best solution. I find it hard to follow, and I worry about our ability to remember why it is that way and what we need to do to keep respecting the needs of the WASM environment in the same way in the future.

IMO the feature gating approach is okay even if it's not the cleanest. Personally, I feel like both alternatives (splitting out yet another smaller crate, or duplicating `RecvMeta` and `Transmit`) are probably worse.

Would be good to get another round of @Ralith's review.

djc:
@gretchenfrage what is necessary to get you to "Approve" this PR? Would be good to move forward.

gretchenfrage (commented):
> @gretchenfrage what is necessary to get you to "Approve" this PR? Would be good to move forward.

I was hoping to abstain and let @Ralith give a second approval, if a second approval is necessary. I was skeptical of the approach but I'm not inclined to veto it--it should not be considered blocked by me. I probably could have communicated that better.

If my review is needed I can schedule another look soon.

Ralith (approved):

Ralith:
LGTM, thanks! We can always revisit if someone comes up with a cleaner solution.

gretchenfrage:
Thank you for working on this!

matheus23:
Thank y'all for the reviews Railith, mxinden and djc.
Also specifically thank you to djc for keeping the ball up on this one :pray:

PR #2043, opened 2024-11-14
feat: QUIC Address discovery extension

divagant-martian:
Implements <https://datatracker.ietf.org/doc/draft-seemann-quic-address-discovery/>

Description TBD


quinn-proto/src/connection/paths.rs
```diff
@@ -37,6 +40,11 @@ pub(super) struct PathData {
     /// Used in persistent congestion determination.
     pub(super) first_packet_after_rtt_sample: Option<(SpaceId, u64)>,
     pub(super) in_flight: InFlight,
+    /// Whether this path has had it's remote address reported back to the peer. This only happens
+    /// if both peers agree to so based on their transport parameters.
+    pub(super) observed_addr_sent: bool,
```
Ralith, line 45: Could we avoid tracking this extra state by queueing a frame immediately whenever we change the path?

quinn-proto/src/connection/mod.rs
```diff
@@ -3140,6 +3241,19 @@ impl Connection {
                 buf.write(frame::FrameType::PATH_RESPONSE);
                 buf.write(token);
                 self.stats.frame_tx.path_response += 1;
+
+                // NOTE: this is technically not required but might be useful to ride the
+                // request/response nature of path challenges to refresh an observation
```
Ralith, line 3246: When is refreshing an observation useful, given that we don't believe the path has changed?

quinn-proto/src/connection/mod.rs
```diff
@@ -3140,6 +3241,19 @@ impl Connection {
...
+                // Since PATH_RESPONSE is a probing frame, this is allowed by the spec.
```
Ralith, line 3247: Why does it matter whether `PATH_RESPONSE` is a probing frame?

quinn-proto/src/connection/mod.rs
```diff
@@ -3058,6 +3103,53 @@ impl Connection {
                 self.stats.frame_tx.handshake_done.saturating_add(1);
         }
 
+        // OBSERVED_ADDR
+        let mut send_observed_address =
```
Ralith, line 3107: Should this be a helper method on `Connection`? Large inline lambdas aren't great for readability.

Ralith, line 3107: This name should reflect that it might not do anything.
```suggestion
        let mut maybe_send_observed_address =
```

quinn-proto/src/connection/mod.rs
```diff
@@ -3058,6 +3103,53 @@ impl Connection {
...
+             skip_sent_check: bool| {
+                // should only be sent within Data space and only if allowed by extension
+                // negotiation
+                // send is also skipped if the path has already sent an observed address
+                let send_allowed = self
+                    .config
+                    .address_discovery_role
+                    .should_report(&self.peer_params.address_discovery_role);
+                let send_required =
+                    space.pending.observed_addr || !self.path.observed_addr_sent || skip_sent_check;
```
Ralith, line 3123: Rather than having a bool to skip these checks in all but one case, lift the checks out to that one case.

quinn-proto/src/connection/mod.rs
```diff
@@ -3127,7 +3219,16 @@ impl Connection {
                 trace!("PATH_CHALLENGE {:08x}", token);
                 buf.write(frame::FrameType::PATH_CHALLENGE);
                 buf.write(token);
-                self.stats.frame_tx.path_challenge += 1;
+
+                send_observed_address(
```
Ralith, line 3223: How do we know this doesn't result in redundant frames being included in a single packet? Would it be easier to make the decision to include this frame in exactly one place?

quinn-proto/src/frame.rs
```diff
@@ -929,8 +940,86 @@ impl AckFrequency {
     }
 }
 
+/* Address Discovery https://datatracker.ietf.org/doc/draft-seemann-quic-address-discovery/ */
+
+/// Conjuction of the information contained in the address discovery frames
+/// ([`FrameType::OBSERVED_IPV4_ADDR`], [`FrameType::OBSERVED_IPV6_ADDR`]).
+#[derive(Debug, PartialEq, Eq, Clone)]
```
Ralith, line 947:
```suggestion
#[derive(Debug, PartialEq, Eq, Copy, Clone)]
```

quinn-proto/src/connection/mod.rs
```diff
@@ -3760,6 +3874,8 @@ pub enum Event {
     DatagramReceived,
     /// One or more application datagrams have been sent after blocking
     DatagramsUnblocked,
+    /// Received an observation of our external address from the peer.
```
Ralith, line 3877: Explicitly document whether this is necessarily different from the previous one, and the required configuration and peer support.

quinn-proto/src/connection/mod.rs
```diff
@@ -2909,6 +2921,33 @@ impl Connection {
...
+                        ));
+                    }
+
+                    if remote == self.path.remote {
+                        if let Some(updated) = self.path.update_observed_addr_report(observed) {
+                            self.events.push_back(Event::ObservedAddr(updated));
+                        }
+                    } else {
+                        // include in migration
+                        migration_observed_addr = Some(observed)
```
Ralith, line 2948: Why is this a special case?

quinn-proto/src/frame.rs
```diff
@@ -929,8 +940,86 @@ impl AckFrequency {
...
+    /// Monotonically increasing integer within the same connection.
+    pub(crate) seq_no: VarInt,
+    /// Reported observed address.
+    pub(crate) ip: IpAddr,
+    /// Reported observed port.
+    pub(crate) port: u16,
+}
+
+impl ObservedAddr {
+    pub(crate) fn new<N: Into<VarInt>>(remote: std::net::SocketAddr, seq_no: N) -> Self {
```
Ralith, line 958: Do we ever want to pass something that isn't already a `VarInt` in here? If not, skip the generic for simplicity.

quinn-proto/src/address_discovery.rs
```diff
@@ -0,0 +1,128 @@
...
+
+use crate::VarInt;
+
+pub(crate) const TRANSPORT_PARAMETER_CODE: u64 = 0x9f81a176;
+
+/// The role of each participant.
+///
+/// When enabled, this is reported as a transport parameter.
+#[derive(PartialEq, Eq, Clone, Copy, Debug, Default)]
+pub(crate) enum Role {
```
Ralith, line 12: It seems like we go to a lot of trouble to convert this to and from a pair of `bool`s. Would it it be simpler to represent it that way in the first place?

quinn-proto/src/connection/mod.rs
```diff
@@ -2974,6 +3013,12 @@ impl Connection {
                 &self.config,
             )
         };
+        new_path.last_observed_addr_report = self.path.last_observed_addr_report.clone();
+        if let Some(report) = observed_addr {
+            if let Some(updated) = new_path.update_observed_addr_report(report) {
+                self.events.push_back(Event::ObservedAddr(updated));
+            }
+        }
```
djc, line 3021: Can we push all/most of this logic into a method on `PathData`?

quinn-proto/src/frame.rs
djc: Also, import `SocketAddr` instead of qualifying it.

quinn-proto/src/frame.rs
```diff
@@ -929,8 +940,86 @@ impl AckFrequency {
...
+    /// [`FrameType::OBSERVED_IPV4_ADDR`] or [`FrameType::OBSERVED_IPV6_ADDR`].
+    pub(crate) fn read<R: Buf>(bytes: &mut R, is_ipv6: bool) -> coding::Result<Self> {
+        let seq_no = bytes.get()?;
+        let ip = if is_ipv6 {
+            IpAddr::V6(bytes.get()?)
+        } else {
+            IpAddr::V4(bytes.get()?)
+        };
+        let port = bytes.get()?;
+        Ok(Self { seq_no, ip, port })
```
djc, line 1011:
```rust
Ok(Self {
    seq_no: bytes.get()?,
    ip: match is_ipv6 {
        true => IpAddr::V6(bytes.get()?),
        false => IpAddr::V4(bytes.get()?),
    },
    port: bytes.get()?,
})
```

quinn-proto/src/transport_parameters.rs
```diff
@@ -421,6 +434,21 @@ impl TransportParameters {
...
+                    }
+                    let value: VarInt = r.get()?;
+                    if len != value.size() {
+                        return Err(Error::Malformed);
+                    }
+                    params.address_discovery_role = value.try_into()?;
+                    tracing::debug!(
+                        role = ?params.address_discovery_role,
+                        "address discovery enabled for peer"
+                    );
```
djc, line 450: Suggest rolling this up into a `Role::from_parameter()` (name TBD) constructor.

quinn/examples/client.rs
```diff
@@ -117,6 +122,18 @@ async fn run(options: Opt) -> Result<()> {
...
+    tokio::spawn(async move {
+        loop {
+            if let Some(new_addr) = *external_addresses.borrow_and_update() {
+                info!(%new_addr, "new external address report");
+            }
+            if external_addresses.changed().await.is_err() {
+                break;
+            }
+        }
+    });
```
djc, line 135: I don't think this is widely used enough that it merits real estate in these examples.

quinn/src/connection.rs
```diff
@@ -636,6 +636,12 @@ impl Connection {
         // May need to send MAX_STREAMS to make progress
         conn.wake();
     }
+
+    /// Track changed on our external address as reported by the peer.
+    pub fn observed_external_addr(&self) -> watch::Receiver<Option<SocketAddr>> {
```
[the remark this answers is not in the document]
djc: Not sure about exposing the `watch::Receiver` as part of the public API, that seems like it's leaking a bunch of implementation details?

quinn-proto/src/frame.rs
```diff
@@ -929,8 +940,86 @@ impl AckFrequency {
...
+/// Conjuction of the information contained in the address discovery frames
+/// ([`FrameType::OBSERVED_IPV4_ADDR`], [`FrameType::OBSERVED_IPV6_ADDR`]).
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub(crate) struct ObservedAddr {
+    /// Monotonically increasing integer within the same connection.
+    pub(crate) seq_no: VarInt,
+    /// Reported observed address.
+    pub(crate) ip: IpAddr,
+    /// Reported observed port.
+    pub(crate) port: u16,
```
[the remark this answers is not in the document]
divagant-martian: I think this is cleaner since socket addresses carry more information than just the ip and port for ipv6, making comparisons between two (seemingly) equal sockets address potentially problematic when the other fields differ. If you prefer I can change it nonetheless

quinn-proto/src/connection/mod.rs
divagant-martian: the reason why we added this is that we need to refresh the observation when the path (our end, their end) changes. We can identify that their end changed but I don't think we can identify 100% of the cases when there is a local change. Since the frame is cheap this seemed useful to keep information as up to date as possible

quinn-proto/src/connection/mod.rs
divagant-martian: I might be mistaken since I need to refresh my memory about the spec, but: the reason was to ensure this was sent on probing packets only

quinn-proto/src/connection/paths.rs
```diff
@@ -156,6 +171,37 @@ impl PathData {
...
+
+    /// Updates the last observed address report received on this path.
+    ///
+    /// If the address was updated, it's returned to be informed to the application.
+    #[must_use = "updated observed address must be reported to the application"]
+    pub(super) fn update_observed_addr_report(
+        &mut self,
+        observed: ObservedAddr,
+    ) -> Option<SocketAddr> {
+        match self.last_observed_addr_report.as_mut() {
```
gretchenfrage, line 183: I think this could be shorted a lot. `ObservedAddr` is copy. Maybe something like this?

```rs
let prev = self.last_observed_addr_report;
if prev.map(|prev| prev.seq_no < observed.seq_no).unwrap_or(true) {
    self.last_observed_addr_report = Some(observed);
}
if Some(observed.socket_addr()) != prev.map(|prev| prev.socket_addr()) {
    Some(observed.socket_addr())
} else {
    None
}
```
It could be golfed more but something like the above perhaps is at a nice spot for readability.

quinn-proto/src/frame.rs
```diff
@@ -922,8 +933,86 @@ impl AckFrequency {
...
+        buf.write(self.get_type());
+        buf.write(self.seq_no);
+        match self.ip {
+            IpAddr::V4(ipv4_addr) => {
+                buf.write(ipv4_addr);
+            }
+            IpAddr::V6(ipv6_addr) => {
+                buf.write(ipv6_addr);
+            }
+        }
```
gretchenfrage, line 988: Nit: Maybe just this?

```rs
        match self.ip {
            IpAddr::V4(ipv4_addr) => buf.write(ipv4_addr),
            IpAddr::V6(ipv6_addr) => buf.write(ipv6_addr),
        }
```

quinn-proto/src/frame.rs
```diff
@@ -922,8 +933,86 @@ impl AckFrequency {
...
+        let ip = if is_ipv6 {
+            IpAddr::V6(bytes.get()?)
+        } else {
+            IpAddr::V4(bytes.get()?)
+        };
+        let port = bytes.get()?;
+        Ok(Self { seq_no, ip, port })
+    }
+
+    /// Gives the [`SocketAddr`] reported in the frame.
```
gretchenfrage, line 1007: This project's convention is to have the first line of a doc comment not end with a period

quinn-proto/src/address_discovery.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    /// Whether this peer's role accepts observed address reports.
+    fn receives_reports(&self) -> bool {
+        matches!(self, Self::ReceiveOnly | Self::Both)
+    }
+
+    /// Whether this peer should report observed addresses to the other peer.
+    pub(crate) fn should_report(&self, other: &Self) -> bool {
+        self.is_reporter() && other.receives_reports()
+    }
```
gretchenfrage, line 56: I think maybe this should just be inlined into its call sites, I don't think the abstraction is helping

quinn-proto/src/address_discovery.rs
gretchenfrage: Since this doesn't really get copied around either, I'm thinking maybe it can just be:

```rs
pub struct TransportConfig {
    ...
    pub(crate) send_address_discovery_reports: bool,
    pub(crate) receive_address_discovery_reports: bool,
    ...
}
```

And the methods here can either be removed, inlined, or turned into free-floating functions

gretchenfrage (changes requested):
Looks promising so far!

quinn-proto/src/connection/mod.rs
divagant-martian: I would have loved this, and that's the first thing I tried. But since the `self.space` is used thought the code rust is not able to tell that the new function modifies disjoint fields so it believes we are trying to get two mutable references to `self` at the same time

quinn-proto/src/frame.rs
divagant-martian: I was a bit hesitant to do this ensuring read order but the read-write test is happy and now looks simpler, thanks

quinn-proto/src/address_discovery.rs
divagant-martian: > I also don't love the Role name for something arguably pretty niche, should we make the type name AddressDiscoveryRole instead

as far as I'm aware, in general it's preferred to not prefix names with their module name unless this repo has a different standard? the point is to refer to paths as `use module;` and then `module::Thing` instead of `use module::ModuleThing;`. If the standard for the repo is not this happy to change it.

I swaped the enum for the struct with two fields. This makes more sense in hindsight.

quinn-proto/src/frame.rs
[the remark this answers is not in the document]
Ralith: Good enough for me.

quinn-proto/src/address_discovery.rs
djc: Fair enough.

Ralith: I'm fine with this naming pattern. Since we only ever use it qualified, it's plenty clear.

quinn-proto/src/connection/mod.rs
```diff
@@ -3053,6 +3098,53 @@ impl Connection {
...
+                if buf.len() + observed.size() < max_size {
+                    observed.write(buf);
+
+                    self.next_observed_addr_seq_no =
+                        self.next_observed_addr_seq_no.saturating_add(1u8);
+                    self.path.observed_addr_sent = true;
+
+                    stats.frame_tx.observed_addr += 1;
+                    sent.retransmits.get_or_create().observed_addr = true;
+                    space.pending.observed_addr = false;
```
flub, line 3135: Somewhere in this block it should log that this frame is being written into the packet, to be consistent with the rest of the logging:
```rust
trace!(?observed, "OBSERVED_ADDRESS");
```

PR #2044, opened 2024-11-15
chore(ci): Remove workaround for broken `cc` version

matheus23:
We've previously had this workaround in CI for `cc` versions between version `1.1.32` and `1.1.37` which broke the `ring` build for wasm32-unknown-unknown in github actions.

Version `1.2.0` is now released and doesn't break `ring` anymore.

Related PRs:
- https://github.com/rust-lang/cc-rs/pull/1284
- https://github.com/rust-lang/cc-rs/pull/1286

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -121,11 +121,6 @@ jobs:
 
       - name: Install cargo binstall
         uses: cargo-bins/cargo-binstall@main
-      
-      # We need to downgrade cc to version 1.1.31 for ring Wasm compilation to work.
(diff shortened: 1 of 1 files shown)


djc (approved):
Should we force the use of cc 1.2 via a dependency requirement?

matheus23:
Honestly, I'm not sure. I've only seen these tests break in github actions so far.

Locally, running the cargo tests for wasm32-unknown-unknown with cc version 1.1.32-37 was just fine.

Ralith (approved):
Thanks!

PR #2051, opened 2024-11-18
chore: Fix `cargo clippy` issues

larseggert:


djc:
Are you sticking to nightly clippy for some reason? If so might make sense to have a CI job for that?

perf/src/bin/perf_client.rs
```diff
@@ -308,7 +308,7 @@ async fn request(
 
     let send_stream_stats = stream_stats.new_sender(&send, upload);
 
-    const DATA: [u8; 1024 * 1024] = [42; 1024 * 1024];
+    static DATA: [u8; 1024 * 1024] = [42; 1024 * 1024];
```
djc, line 311: Why is this? I would usually assume const is better than static.

djc (approved):

perf/src/bin/perf_client.rs
larseggert: https://rust-lang.github.io/rust-clippy/master/index.html#large_const_arrays

larseggert:
> Are you sticking to nightly clippy for some reason? If so might make sense to have a CI job for that?

I just run the Nightly toolchain by default. CI would work but `clippy --fix` isn't super stable and so this usually requires manual interaction.

PR #2055, opened 2024-11-20
Improve time accounting for Incoming

Ralith:
Plus some bonus MSRV cleanup.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -31,8 +31,8 @@ use crate::{
     },
     token::TokenDecodeError,
     transport_parameters::{PreferredAddress, TransportParameters},
-    Instant, ResetToken, RetryToken, Side, SystemTime, Transmit, TransportConfig, TransportError,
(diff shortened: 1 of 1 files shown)


quinn-udp/src/windows.rs
```diff
@@ -403,7 +402,7 @@ const CMSG_LEN: usize = 128;
 const OPTION_ON: u32 = 1;
 
 // FIXME this could use [`std::sync::OnceLock`] once the MSRV is bumped to 1.70 and upper
```
mxinden, line 404:
```suggestion
```

mxinden (approved):
+1 on the `quinn-udp/` changes. Thanks.

djc (approved):
Nice!

djc:
LazyLock is documented as stable from 1.80, but we have an older MSRV, right? Why isn't CI flagging this?

Ralith:
Good question. Dropping the MSRV stuff for now.

djc (approved):

PR #2056, opened 2024-11-20
feat(quinn-udp): support illumos

AsakuraMizu:
This PR adds support for Illumos and tests on CI. Also fixes a warning on Solaris and Illumos.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -82,6 +82,24 @@ jobs:
             export PATH=$HOME/.rust_solaris/bin:$PATH
             cargo build --all-targets && cargo test && cargo test --manifest-path fuzz/Cargo.toml && cargo test -p quinn-udp --benches
 
+  test-illumos:
+    name: test on illumos
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: test on Illumos
+        uses: vmactions/omnios-vm@v1
+        with:
+          usesh: true
+          mem: 4096
+          copyback: false
+          prepare: |
+            pkg install gcc14 curl pkg-config glib2
+            curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile minimal
+          run: |
+            . "$HOME/.cargo/env"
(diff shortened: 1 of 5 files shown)


.github/workflows/rust.yml
```diff
@@ -193,8 +211,8 @@ jobs:
   audit:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
-    - uses: EmbarkStudios/cargo-deny-action@v2
+      - uses: actions/checkout@v4
+      - uses: EmbarkStudios/cargo-deny-action@v2
```
mxinden, line 215: I suggest doing these formatting changes in a separate pull request.

.github/workflows/rust.yml
```diff
@@ -212,49 +230,49 @@ jobs:
             api-level: 26
 
     steps:
-    - name: Checkout code
```
mxinden, line 215: Same here, better done in a separate pull request.

mxinden (commented):
@psumbera given your expertise with Solaris, want to take a look?

.github/workflows/rust.yml
AsakuraMizu: I believe it's done accidentally since I have "Format On Save" enabled.

psumbera:
> @psumbera given your expertise with Solaris, want to take a look?

This looks good to me. Thank you!

Ralith (approved):

quinn/src/tests.rs
```diff
@@ -387,7 +387,10 @@ async fn zero_rtt() {
 }
 
 #[test]
-#[cfg_attr(target_os = "solaris", ignore = "Fails on Solaris")]
+#[cfg_attr(
+    any(target_os = "solaris", target_os = "illumos"),
```
djc, line 391: Nit: could we use `solarish` here, too?

djc (approved):

quinn/src/tests.rs
mxinden: Friendly ping @AsakuraMizu.

AsakuraMizu: I suggest doing this in a separate pull request since the `quinn` crate currently does not have a custom build script or cfg aliases setup.

Issue #2057, opened 2024-11-20
Usage of constant value for GREASE transport parameter make quinn vulnerable to fingerprinting by quic transport parameters.


mstyura:
Currently `quinn` uses constant value as `GREASE` reserved random parameter.
https://github.com/quinn-rs/quinn/blob/9386cde871c750464073772409615e90344b80e9/quinn-proto/src/transport_parameters.rs#L303-L305

This make `quinn` client side users vulnerable to fingerprinting by predictable patterns during handshake. Thanks to ability to inject custom TLS backend like [quinn-boring](https://github.com/quinn-rs/quinn-boring/) most of TLS handshake is configurable, except the content of quic transport parameters extension.

As a prevention actions I see the following steps:
1. Generate random transport parameter, like it is done [quic-go](https://github.com/quic-go/quic-go/blob/3e0a67b2476e1819752f04d75968de042b197b56/internal/wire/transport_parameters.go#L338-L344) or [quiche](https://github.com/google/quiche/blob/cb1090b20c40e2f0815107857324e99acf6ec567/quiche/quic/core/crypto/transport_parameters.cc#L843-L860)];
2. Make it optional;
3. Implement permutation of transport parameters, like it is done in [quiche](https://github.com/google/quiche/blob/cb1090b20c40e2f0815107857324e99acf6ec567/quiche/quic/core/crypto/transport_parameters.cc#L868-L873).

mstyura:
Correct me if I'm wrong, the general philosophy of library is not to provide an API which can potentially reduce default security & privacy while providing not clear benefits. So PRs to [opt out grease](https://github.com/quinn-rs/quinn/pull/2061) & random permutation (no such PR, just potentially) are not welcomed? I'm ok with any decision, just basically asking should I close above-mentioned PR or not yet (rebase).

Ralith:
I'd like to additionally see some concrete motivation. The bar can be pretty low for simple features that don't take much code/documentation, but if we're adding new paths and taking up space even in documentation, I want *some* evidence that they're useful, ideally for multiple applications.

For example, "imitate Apple's implementation" is niche enough that I'm not sure it clears the bar. If the two proposed changes are the only necessary ones to achieve that end, then I'm ambivalent and happy to defer to @djc's opinion. If more invasive changes we're unlikely to merge are additionally required, then the proposed changes don't seem to serve a purpose on their own and I'd rather do without.

dongcarl:
Perhaps a good middle-ground that:

[...]

PR #2058, opened 2024-11-20
#2057: Use randomly generated GREASE transport parameter.

mstyura:
As a first step to make `quinn` be less vulnerable to fingerprinting I propose to generate GREASE transport parameter randomly.


quinn-proto/src/transport_parameters.rs
```diff
@@ -476,6 +475,34 @@ fn decode_cid(len: usize, value: &mut Option<ConnectionId>, r: &mut impl Buf) ->
...
+    // [1] https://github.com/quic-go/quic-go/blob/3e0a67b2476e1819752f04d75968de042b197b56/internal/wire/transport_parameters.go#L338-L344
+    // [2] https://github.com/google/quiche/blob/cb1090b20c40e2f0815107857324e99acf6ec567/quiche/quic/core/crypto/transport_parameters.cc#L843-L860
+
+    let id = {
+        let rand = rng.next_u64();
+        let n = (rand % ((1 << 62) - 27)) / 31;
+        31 * n + 27
+    };
+    debug_assert!(id < (1 << 62));
+    debug_assert!(id % 31 == 27);
```
Ralith, line 491: Please document the intent behind this math.

quinn-proto/src/transport_parameters.rs
```diff
@@ -300,9 +301,7 @@ impl TransportParameters {
         }
         apply_params!(write_params);
 
-        // Add a reserved parameter to keep people on their toes
-        w.write_var(31 * 5 + 27);
-        w.write_var(0);
+        write_random_grease_reserved_parameter(w, &mut rand::thread_rng());
```
Ralith, line 304: Can we pipe in the endpoint/connection rng rather than using a global one? That should help with reproducibility in testing.

Ralith (changes requested):
Thanks, this seems nice to have.

quinn-proto/src/transport_parameters.rs
Ralith: `Endpoint` and `Connection` both already have `rng` fields, which are used for randomness in logic owned by those components. `RandomConnectionIdGenerator` and the config structs are constructed by the user outside of the lifetime of an endpoint or connection, and their global randomness is intended only as a convenience for typical use.

For this change, taking a generic `Rng` parameter in `TransportParameters::new` would probably be the most graceful approach, as it's called directly from `Endpoint` where the endpoint's RNG is readily available.

quinn-proto/src/transport_parameters.rs
```diff
@@ -476,6 +475,43 @@ fn decode_cid(len: usize, value: &mut Option<ConnectionId>, r: &mut impl Buf) ->
...
+fn write_random_grease_reserved_parameter<W: BufMut, R: RngCore>(w: &mut W, rng: &mut R) {
+    // See: https://datatracker.ietf.org/doc/html/rfc9000#section-22.3
+    // Inspired by quic-go & quiche implementation:
+    // [1] https://github.com/quic-go/quic-go/blob/3e0a67b2476e1819752f04d75968de042b197b56/internal/wire/transport_parameters.go#L338-L344
+    // [2] https://github.com/google/quiche/blob/cb1090b20c40e2f0815107857324e99acf6ec567/quiche/quic/core/crypto/transport_parameters.cc#L843-L860
+
+    // The randomly generated identifier of the form `31 * N + 27`;
+    let id = {
+        // To generate random ID it first necessary to pick random `N` such that `31 * N + 27 < 2^62`,
+        // It is achieved by picking random number from range [0; 2^62 - 27) and dividing it by 31.
```
Ralith, line 488: This describes the math being done, but we can already see that from the actual code. Please instead document *why* this is necessary. Presumably the idea is to satisfy the RFC's definition of IDs that must never be assigned? If so, explain that, and cite https://www.rfc-editor.org/rfc/rfc9000.html#name-quic-transport-parameters-r.

quinn-proto/src/transport_parameters.rs
```diff
@@ -467,6 +475,81 @@ impl TransportParameters {
...
+
+    /// The number of bytes to include into wire format from the `payload` buffer
+    payload_len: usize,
+}
+
+impl ReservedTransportParameter {
+    /// The maximum length of payload to include as parameter payload.
+    /// This value is not a specification-imposed limit and picked such that
+    /// it matches the limit by other implementations of quic, e.g. quic-go & quiche.
+    const MAX_PAYLOAD_LEN: usize = 16;
```
djc, line 502: Nit: move this to the bottom of the `impl` please?

quinn-proto/src/transport_parameters.rs
```diff
@@ -467,6 +475,81 @@ impl TransportParameters {
...
+
+    // Generate a random reserved identifier of the form `31 * N + 27` as required by RFC 9000.
+    // Reserved transport parameter identifiers are used to test compliance with the requirement
+    // that unknown transport parameters be ignored by peers.
+    // See: <https://www.rfc-editor.org/rfc/rfc9000.html#section-18.1> and <https://www.rfc-editor.org/rfc/rfc9000.html#section-22.3>
+    fn generate_reserved_id<R: rand::RngCore>(rng: &mut R) -> VarInt {
```
djc, line 508: Nit: move `generate_reserved_id()` below `random()` (since the latter depends on the former, and same point about taking the generic parameter for all three of the associated functions here.

djc (approved):
A few nits, thanks for working on this!

mstyura:
@djc thanks you very much for review! I've fixed code according to suggestions and also additionally validated comments grammar with ChatGPT and fixed several other mistakes highlighted by it.

Ralith (approved):
Thanks!

mstyura:
Is there anything else I can do to help move this PR toward being merged?

quinn-proto/src/transport_parameters.rs
```diff
@@ -467,6 +475,82 @@ impl TransportParameters {
     }
 }
 
+/// A reserved transport parameter.
+///
+/// It has an identifier of the form 31 * N + 27 for the integer value of N.
+/// Such identifiers are reserved to exercise the requirement that unknown transport parameters be ignored.
+/// The reserved transport parameter has no semantics and can carry arbitrary values.
+/// It may be included in `quic_transport_parameters` TLS extensions when sent and should be ignored when received.
```
Ralith, line 483:
```suggestion
/// It may be included in transport parameters sent to the peer, and should be ignored when received.
```
Transport parameters are not TLS extensions, and Quinn defines no identifier by the name `quic_transport_parameters`.

mstyura: I've made mistake with plural form of word extensions, I wanted to say that this reserved transport parameter may be included and carried inside [QUIC transport parameters extension](https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#tls-extensiontype-values-1) of TLS https://www.rfc-editor.org/rfc/rfc9001#section-8.2 (at least according to my current understanding of the subject). So should I fix plural form or just accept your suggestion or maybe rephrase the sentence in a more clear way?

djc:
> Is there anything else I can do to help move this PR toward being merged?

Pay for the maintenance you're getting for free?

quinn-proto/src/transport_parameters.rs
djc: I think @Ralith's suggestion is simpler and more clear.

quinn-proto/src/transport_parameters.rs
```diff
@@ -467,6 +475,82 @@ impl TransportParameters {
...
+///
+/// See spec: <https://www.rfc-editor.org/rfc/rfc9000.html#section-18.1>
+
```
djc, line 486: Nit: no empty line between docstring and type definition, please.

djc (approved):

djc:
Thanks for working on this!

mstyura:
> Pay for the maintenance you're getting for free? We're volunteer maintainers so please be a little more patient.

Sorry, I didn't intend to make a pressure on you. I understand you are volunteers with a lot of Rust projects to maintain. I just wanted to make sure I didn't miss anything required. Regarding the donations I'll talk again to company I work for to allocate some budget next year to support OSS project it rely on.

djc:
> Sorry, I didn't intend to make a pressure on you. I understand you are volunteers with a lot of Rust projects to maintain. I just wanted to make sure I didn't miss anything required.

No worries, I can see how it might have been confusing that this wasn't merged yet after we both approved it -- I get a bit stingy when I get the feeling of being pressured.

> Regarding the donations I'll talk again to company I work for to allocate some budget next year to support OSS project it rely on.

Any efforts in that direction are much appreciated, thanks!

PR #2060, opened 2024-11-21
docs: Correct MSRV in README

gretchenfrage:
Djc bumped MSRV to 1.70 in #1939 in July. However, the README still claims that the MSRV is 1.66. This PR fixes the README and adds a note to the CI config.


Ralith (approved):
Thanks!

PR #2066, opened 2024-11-24
#2057: Shuffle outgoing transport parameters.

mstyura:
I propose to remember the order in which write of the params should be performed when `TransportParameters` are crated. Then this order is used to write to wire format.
PR currently split into two commits:
1. Do the permutation using hard-coded IDs of prams;
2. Some refactoring to store the IDs into enum.

### DIFF
--- a/quinn-proto/src/transport_parameters.rs
+++ b/quinn-proto/src/transport_parameters.rs
@@ -12,7 +12,7 @@ use std::{
 };
 
 use bytes::{Buf, BufMut};
-use rand::{Rng as _, RngCore};
+use rand::{seq::SliceRandom as _, Rng as _, RngCore};
 use thiserror::Error;
 
 use crate::{
@@ -104,6 +104,12 @@ macro_rules! make_struct {
             /// of transport parameter extensions.
             /// When present, it is included during serialization but ignored during deserialization.
             pub(crate) grease_transport_parameter: Option<ReservedTransportParameter>,
+
+            /// Defines the order in which transport parameters are serialized.
+            ///
+            /// This field is initialized only for outgoing `TransportParameters` instances and
+            /// is set to `None` for `TransportParameters` received from a peer.
+            pub(crate) write_order: Option<[u8; TransportParameterId::SUPPORTED.len()]>,
         }
 
         // We deliberately don't implement the `Default` trait, since that would be public, and
@@ -126,6 +132,7 @@ macro_rules! make_struct {
                     stateless_reset_token: None,
                     preferred_address: None,
                     grease_transport_parameter: None,
+                    write_order: None,
                 }
             }
         }
@@ -168,6 +175,11 @@ impl TransportParameters {
                 VarInt::from_u64(u64::try_from(TIMER_GRANULARITY.as_micros()).unwrap()).unwrap(),
             ),
             grease_transport_parameter: Some(ReservedTransportParameter::random(rng)),
+            write_order: Some({
+                let mut order = std::array::from_fn(|i| i as u8);
+                order.shuffle(rng);
+                order
+            }),
             ..Self::default()
         }
     }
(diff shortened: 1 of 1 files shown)


quinn-proto/src/transport_parameters.rs
```diff
@@ -35,37 +70,61 @@ macro_rules! apply_params {
         $macro! {
             // #[doc] name (id) = default,
             /// Milliseconds, disabled if zero
-            max_idle_timeout(0x0001) = 0,
+            max_idle_timeout(TransportParameterId::MaxIdleTimeout) = 0,
```
Ralith, line 73: Factor the `TransportParameterId` qualifier out into the macro for readability.

quinn-proto/src/transport_parameters.rs
```diff
@@ -295,68 +363,101 @@ impl From<UnexpectedEnd> for Error {
...
-            (0x10, &self.retry_src_cid),
-        ] {
-            if let Some(ref cid) = *cid {
-                w.write_var(tag);
-                w.write_var(cid.len() as u64);
-                w.put_slice(cid);
+                }
+                TransportParameterId::StatelessResetToken => {
+                    if let Some(ref x) = self.stateless_reset_token {
+                        w.write_var(TransportParameterId::StatelessResetToken as u64);
```
Ralith, line 379: style: prefer `id as u64` here and below so there's less risk of copy-paste errors in the future.

quinn-proto/src/transport_parameters.rs
```diff
@@ -25,6 +25,41 @@ use crate::{
...
+    ReservedTransportParameter = 0x1B,
+
+    // https://www.rfc-editor.org/rfc/rfc9221.html#section-3
+    MaxDatagramFrameSize = 0x20,
+
+    // https://www.rfc-editor.org/rfc/rfc9287.html#section-3
+    GreaseQuicBit = 0x2AB2,
+
+    // https://datatracker.ietf.org/doc/html/draft-ietf-quic-ack-frequency#section-10.1
+    MinAckDelayDraftIetfQuicAckFrequency07 = 0xFF04DE1B,
```
Ralith, line 60: nit: This seems much more verbose than necessary.
```suggestion
    MinAckDelayDraft07 = 0xFF04DE1B,
```

djc:
> I propose to remember the order in which write of the params should be performed when `TransportParameters` are crated. Then this order is used to write to wire format.

Why do you want to store the order? That seems largely unrelated to the stated goal of this PR to shuffle transport params.

I think the commit introducing an enum should come first. Suggest extracting the macro from its current location in a separate commit so that it can get only the necessary modifications in place in the further commits.

mstyura:
> Why do you want to store the order? That seems largely unrelated to the stated goal of this PR to shuffle transport params.

@djc maybe I'm missing something, but inside `TransportParameters::write` I don't have access to random generator passed form `Connection` to perform shuffle at that time and in my previous MR it was suggested to not to use `rand::thread_rng()` directly. So the solution was to store the write order inside constructor where I have an access to random generator injected from `Connection` and use this order during `write`. Let me know if you think this could be done differently.

> I think the commit introducing an enum should come first. Suggest extracting the macro from its current location in a separate commit so that it can get only the necessary modifications in place in the further commits.

Do you mean only re-order commits in this MR or I should better to put refactoring related to `TransportParameterId` into separate PR and wait until it merged and update this MR with only shuffling implemented?

djc:
> > Why do you want to store the order? That seems largely unrelated to the stated goal of this PR to shuffle transport params.
> 
> @djc maybe I'm missing something, but inside `TransportParameters::write` I don't have access to random generator passed form `Connection` to perform shuffle at that time and in my previous MR it was suggested to not to use `rand::thread_rng()` directly. So the solution was to store the write order inside constructor where I have an access to random generator injected from `Connection` and use this order during `write`. Let me know if you think this could be done differently.

Ah, makes sense -- sorry for the noise!

> > I think the commit introducing an enum should come first. Suggest extracting the macro from its current location in a separate commit so that it can get only the necessary modifications in place in the further commits.
> 
> Do you mean only re-order commits in this MR or I should better to put refactoring related to `TransportParameterId` into separate PR and wait until it merged and update this MR with only shuffling implemented?

Reordering commits in this PR should be okay.

mstyura:
> Reordering commits in this PR should be okay.

Re-ordered commits and hopefully a bit enhanced implementation by allowing `==` between `TransportParameterId` and `u64` to avoid several usages `as u64` casts inside guard of match expression.

> Ah, makes sense -- sorry for the noise!

Also a bit changed implementation of reordering. Instead of storing all transport parameter IDs inside `write_order` array inside `TransportPrameters` which occupied `21 * u64 = 168` bytes spaces in struct I stored `21 * u8` values which treated as indices of `SUPPORTED_TRANSPORT_PARAMETERS` array.

quinn-proto/src/transport_parameters.rs
```diff
@@ -145,6 +170,9 @@ macro_rules! make_struct {
             /// of transport parameter extensions.
             /// When present, it is included during serialization but ignored during deserialization.
             pub(crate) grease_transport_parameter: Option<ReservedTransportParameter>,
+
+            /// The order in which transport parameters are serialized
+            pub(crate) write_order: Option<[u8; SUPPORTED_TRANSPORT_PARAMETERS.len()]>,
```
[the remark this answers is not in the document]
mstyura: Not really mandatory, I though `write_order` only make sense for instances of `TransportParameters` which are created to be then send, so received `TransportParameters` has `None` as `write_order`. I can change code to avoid `Option`, should I?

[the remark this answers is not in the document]
djc: Fair -- feels a little ugly that we're sticking it into `TransportParameters` but I see your point that it's not trivial to pass the RNG around to where we need it. Can you at least more clearly document on the field why it is `Option` and when it is expected to be `Some` vs `None`? (Keep the current comment on its own line, then an empty line, then the additional context.)

quinn-proto/src/transport_parameters.rs
```diff
@@ -387,45 +431,53 @@ impl TransportParameters {
             let len = len as usize;
 
             match id {
-                0x00 => decode_cid(len, &mut params.original_dst_cid, r)?,
-                0x02 => {
+                id if TransportParameterId::OriginalDestinationConnectionId == id => {
```
djc, line 434: One more thing! Could we do a `TryFrom<u64> for `TransportParameterId` impl with `Err = ()` so that we can simplify these match arms?

Ralith (approved):
Thanks!

djc (approved):

mstyura:
@djc, @Ralith thank you very much for review, feedback and patience! And accepting this change!

Ralith:
Thanks for your contribution! I've been specifically wanting this for a while.

PR #2067, opened 2024-11-25
proto: split config module

djc:


Ralith (approved):
I wouldn't mind seeing endpoint/server split out as well.

djc:
Let's get this merged, I'll take a look your proposal after.

PR #2068, opened 2024-11-25
Bump MSRV to 1.71

djc:
This is the MSRV current rustls has adopted (on the back of current env_logger requiring 1.71).


mxinden (approved):

Ralith (changes requested):
Missing README update?

Ralith (approved):

PR #2072, opened 2024-11-29
fix(udp): use IPV6_PMTUDISC_PROBE instead of IP_PMTUDISC_PROBE on v6

mxinden:
Both resolve to `3`, thus not an actual bug.

Caught thanks to @marten-seemann's [DF draft](https://github.com/marten-seemann/draft-seemann-tsvwg-udp-fragmentation).

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -146,7 +146,7 @@ impl UdpSocketState {
                     &*io,
                     libc::IPPROTO_IPV6,
                     libc::IPV6_MTU_DISCOVER,
-                    libc::IP_PMTUDISC_PROBE,
+                    libc::IPV6_PMTUDISC_PROBE,
                 )?;
             }
         }


Ralith (approved):
Thanks! Never know when we might hit a weird environment where these differ.

mxinden:
> Never know when we might hit a weird environment where these differ.

Agreed. The more complexity we can offload to other crates (here `libc`) the better.

mstyura:
Sorry for late comment, maybe it worth adding comment explaining why `IP_PMTUDISC_PROBE` is chosen over `IP_PMTUDISC_DO`
```
Path MTU discovery value   Meaning
              IP_PMTUDISC_WANT           Use per-route settings.
              IP_PMTUDISC_DONT           Never do Path MTU Discovery.
              IP_PMTUDISC_DO             Always do Path MTU Discovery.
              IP_PMTUDISC_PROBE          Set DF but ignore Path MTU.
```
E.g. `Chromium` [uses later option](https://source.chromium.org/chromium/chromium/src/+/main:net/socket/udp_socket_posix.cc;l=475?q=IP_PMTUDISC_DO) in context when `DF` flag is desired, `WebRTC` [also](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/rtc_base/physical_socket_server.cc;l=362?q=IP_PMTUDISC_DO%20webrtc). So there is should be something important which probably deserves explanation why `quinn` uses `IP_PMTUDISC_PROBE` rather than `IP_PMTUDISC_DO`.

Ralith:
It seems pretty self-explanatory to me? We do our own MTUD, so we don't want the kernel's built-in MTUD confusing things.

mstyura:
> We do our own MTUD, so we don't want the kernel's built-in MTUD confusing things.

Thanks, this is exact explanation I missed! WebRTC uses `IP_PMTUDISC_DO` option probably due to by default conservative packet size 1200 and lack of it's own MTU discovery.

mxinden:
@mstyura related discussion https://github.com/marten-seemann/draft-seemann-tsvwg-udp-fragmentation/issues/13. Might be worth moving the Chromium discussion there.

PR #2073, opened 2024-11-29
fix(udp): propagate error on apple_fast

mxinden:
With https://github.com/quinn-rs/quinn/pull/2017, the concrete `send` implementations per platform are supposed to propagate `io::Error`s. Those errors are then eiter logged and dropped in `UdpSocketState::send` or further propagated in `UdpSocketState::try_send`.

The `fast_apple` `send` implementation added in
https://github.com/quinn-rs/quinn/pull/1993 does not follow this pattern.

This commit adjusts the `fast_apple` implementation accordingly.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -395,16 +395,10 @@ fn send(state: &UdpSocketState, io: SockRef<'_>, transmit: &Transmit<'_>) -> io:
         }
         io::ErrorKind::WouldBlock => return Err(e),
         _ => {
-            // Other errors are ignored, since they will usually be handled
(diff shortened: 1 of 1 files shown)


Ralith (approved):
Thanks!

thomaseizinger (approved):
Thank you Max!

thomaseizinger:
FWIW, Rust 1.83 (released yesterday) stabilised new `io::ErrorKind`s like `HostUnreachable` and `NetworkUnreachable` (https://doc.rust-lang.org/std/io/enum.ErrorKind.html) which correspond to ICMP errors generated by the kernel for unroutable packets.

PR #2074, opened 2024-11-29
fix(udp): retry on ErrorKind::Interrupted

mxinden:
On Linux and Android, `quinn-udp` will retry a send on `ErrorKind::Interrupted`:

https://github.com/quinn-rs/quinn/blob/e318cc4a80436fd9fa19c02886d682c49efca185/quinn-udp/src/unix.rs#L305-L312

Do the same on all other unix platforms.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -383,25 +383,27 @@ fn send(state: &UdpSocketState, io: SockRef<'_>, transmit: &Transmit<'_>) -> io:
         hdrs[i].msg_datalen = chunk.len();
         cnt += 1;
     }
-    let n = unsafe { sendmsg_x(io.as_raw_fd(), hdrs.as_ptr(), cnt as u32, 0) };
-    if n >= 0 {
-        return Ok(());
-    }
-    let e = io::Error::last_os_error();
-    match e.kind() {
-        io::ErrorKind::Interrupted => {
-            // Retry the transmission
-        }
-        io::ErrorKind::WouldBlock => return Err(e),
-        _ => {
-            // - EMSGSIZE is expected for MTU probes. Future work might be able to avoid
-            //   these by automatically clamping the MTUD upper bound to the interface MTU.
-            if e.raw_os_error() != Some(libc::EMSGSIZE) {
-                return Err(e);
+    loop {
+        let n = unsafe { sendmsg_x(io.as_raw_fd(), hdrs.as_ptr(), cnt as u32, 0) };
+        if n == -1 {
+            let e = io::Error::last_os_error();
+            match e.kind() {
+                io::ErrorKind::Interrupted => {
+                    // Retry the transmission
+                    continue;
+                }
+                io::ErrorKind::WouldBlock => return Err(e),
+                _ => {
+                    // - EMSGSIZE is expected for MTU probes. Future work might be able to avoid
+                    //   these by automatically clamping the MTUD upper bound to the interface MTU.
+                    if e.raw_os_error() != Some(libc::EMSGSIZE) {
+                        return Err(e);
+                    }
+                }
             }
         }
+        return Ok(());
     }
-    Ok(())
 }
 
 #[cfg(any(target_os = "openbsd", target_os = "netbsd", apple_slow))]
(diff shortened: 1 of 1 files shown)


quinn-udp/src/unix.rs
```diff
@@ -383,31 +383,33 @@ fn send(state: &UdpSocketState, io: SockRef<'_>, transmit: &Transmit<'_>) -> io:
...
-            //   Those are not fatal errors, since the
-            //   configuration can be dynamically changed.
-            // - Destination unreachable errors have been observed for other
-            // - EMSGSIZE is expected for MTU probes. Future work might be able to avoid
-            //   these by automatically clamping the MTUD upper bound to the interface MTU.
-            if e.raw_os_error() != Some(libc::EMSGSIZE) {
-                log_sendmsg_error(&state.last_send_error, e, transmit);
+    loop {
+        let n = unsafe { sendmsg_x(io.as_raw_fd(), hdrs.as_ptr(), cnt as u32, 0) };
+        if n < 0 {
```
mxinden, line 388:
```suggestion
        if n == -1 {
```

On all other Unix platforms we check for `n == -1`. Would the same suffice on `fast_apple`? @larseggert do you know?

larseggert: I think so

Ralith (approved):
Thanks!

quinn-udp/src/unix.rs
```diff
@@ -383,31 +383,33 @@ fn send(state: &UdpSocketState, io: SockRef<'_>, transmit: &Transmit<'_>) -> io:
...
+                    // Other errors are ignored, since they will usually be handled
+                    // by higher level retransmits and timeouts.
+                    // - PermissionDenied errors have been observed due to iptable rules.
+                    //   Those are not fatal errors, since the
+                    //   configuration can be dynamically changed.
+                    // - Destination unreachable errors have been observed for other
+                    // - EMSGSIZE is expected for MTU probes. Future work might be able to avoid
+                    //   these by automatically clamping the MTUD upper bound to the interface MTU.
+                    if e.raw_os_error() != Some(libc::EMSGSIZE) {
+                        log_sendmsg_error(&state.last_send_error, e, transmit);
```
Ralith, line 406: Pre-existing issue, but we should return here; `log_sendmsg_error` vs. propagation is now decided at a higher level to support both use cases.

Ralith: Oh that's https://github.com/quinn-rs/quinn/pull/2073#event-15477015767.

quinn-udp/src/unix.rs
mxinden: Thanks. Switched to `== -1`.

quinn-udp/src/unix.rs
```diff
@@ -383,31 +383,33 @@ fn send(state: &UdpSocketState, io: SockRef<'_>, transmit: &Transmit<'_>) -> io:
...
-            //   Those are not fatal errors, since the
-            //   configuration can be dynamically changed.
-            // - Destination unreachable errors have been observed for other
-            // - EMSGSIZE is expected for MTU probes. Future work might be able to avoid
-            //   these by automatically clamping the MTUD upper bound to the interface MTU.
-            if e.raw_os_error() != Some(libc::EMSGSIZE) {
-                log_sendmsg_error(&state.last_send_error, e, transmit);
+    loop {
+        let n = unsafe { sendmsg_x(io.as_raw_fd(), hdrs.as_ptr(), cnt as u32, 0) };
+        if n == -1 {
```
djc, line 388: For both of these, how do we feel about inverting this condition to reduce rightward drift at the cost of duplicating the `return Ok(());`?

And/or have you considered deduplicating more of this logic? These `match` expressions are pretty much the same...

djc (approved):

quinn-udp/src/unix.rs
thomaseizinger: The entire loop could probably be extracted into a function with the syscall being an `FnMut` parameter?

quinn-udp/src/unix.rs
```diff
@@ -425,24 +427,27 @@ fn send(state: &UdpSocketState, io: SockRef<'_>, transmit: &Transmit<'_>) -> io:
...
+            match e.kind() {
+                io::ErrorKind::Interrupted => {
+                    // Retry the transmission
+                    continue;
+                }
+                io::ErrorKind::WouldBlock => return Err(e),
+                _ => {
+                    // - EMSGSIZE is expected for MTU probes. Future work might be able to avoid
+                    //   these by automatically clamping the MTUD upper bound to the interface MTU.
+                    if e.raw_os_error() != Some(libc::EMSGSIZE) {
```
thomaseizinger, line 443: I know we are doing it this way in the other impls too but this just tripped me up again, thinking that errors are not returned here.

I think this would be clearer if the bool is flipped and we special-case an `Ok(())` return here and default to returning the error.

Let me know if you are up for it, otherwise I can spend some cycles on this next week.

thomaseizinger: This could also be a match-arm guard in which case it would more naturally slot into handling different variants of this error.

(I guess we are using the raw error value here because the equivalent `ErrorKind` is still unstable?)

thomaseizinger: > (I guess we are using the raw error value here because the equivalent `ErrorKind` is still unstable?)

It actually doesn't even seem to be mapped. Sad.

quinn-udp/src/unix.rs
thomaseizinger: We can of have the same loop three times actually.

quinn-udp/src/unix.rs
thomaseizinger: I wonder if this case should actually still be special-cased. Should we move this condition further up as well? Sending too large of a packet seems like a worth-while error to report.

quinn-udp/src/unix.rs
mxinden: I can provide another patch de-duplicating the loop logic. That said, I suggest doing that in a separate pull request.

@djc @thomaseizinger fine by you?

quinn-udp/src/unix.rs
mxinden: I don't have a strong opinion, though prefer for any of such change to happen in a separate pull request.

quinn-udp/src/unix.rs
djc: Works for me.

djc:
Just needs rebasing!

PR #2075, opened 2024-11-29
tests(quinn): Ignore stress tests by default

gretchenfrage:
Per [Discord messages](https://discord.com/channels/976380008299917365/1063547094863978677/1310833978957234228), ignores long-running stress tests by default.

With this commit, `cargo test` runs in less than 2 seconds on my machine, whereas on main it takes over 60.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -33,7 +33,7 @@ jobs:
             echo "===== freebsd-version ====="
             freebsd-version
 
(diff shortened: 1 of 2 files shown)


Ralith (approved):

djc (approved):
Should we set it up so that these are still run in CI?

Ralith (approved):

gretchenfrage:
Should be done.

PR #2078, opened 2024-12-10
feat(quinn-proto): unhide `quinn_proto::coding`

### LINKED ISSUE #1277: Expose encoding and decoding of VarInt
The `VarInt` type is currently exposed, including `VarInt::encoded_size`.  However there seems no public API to encode and decode it and I wanted to use it for my own wire-format inside QUIC streams.

A simple wrapper would probably already be sufficient:
```
impl VarInt {
    /// Parses a VarInt from the byte slice.
    pub fn parse_bytes(mut buf: &[u8]) -> coding::Result<VarInt> {
        Self::decode(&mut buf)
    }

    /// Writes the VarInt encoded as bytes.
    pub fn write_bytes<B: BufMut>(&self, w: &mut B) {
        self.encode(w)
    }
}
```

Could adding something like this be considered?

I also noticed that the `quinn-proto::coding::Codec` trait seems to be public-but-hidden.  I assume we should not start depending on this either, but maybe making it public, at least for people directly depending on quinn-proto is also an alternative?

dignifiedquire:
It turned out currently the `coding` mod is already public, simply hidden from the docs.

It seems to be the minimal change to allow using this functionality for `VarInt` is to document it and not hide it.

Fixes #1277

### DIFF
--- a/quinn-proto/src/coding.rs
+++ b/quinn-proto/src/coding.rs
@@ -1,3 +1,5 @@
+//! Coding related traits.
+
 use std::net::{Ipv4Addr, Ipv6Addr};
 
 use bytes::{Buf, BufMut};
@@ -5,14 +7,19 @@ use thiserror::Error;
 
 use crate::VarInt;
 
+/// Error indicating that the provided buffer was too small
 #[derive(Error, Debug, Copy, Clone, Eq, PartialEq)]
 #[error("unexpected end of buffer")]
 pub struct UnexpectedEnd;
 
+/// Coding result type
 pub type Result<T> = ::std::result::Result<T, UnexpectedEnd>;
 
+/// Infallible encoding and decoding of QUIC primitives
 pub trait Codec: Sized {
+    /// Decode a `Self` from the provided buffer, if the buffer is large enough
     fn decode<B: Buf>(buf: &mut B) -> Result<Self>;
+    /// Append the encoding of `self` to the provided buffer
     fn encode<B: BufMut>(&self, buf: &mut B);
 }
 
@@ -92,7 +99,7 @@ impl Codec for Ipv6Addr {
     }
 }
 
-pub trait BufExt {
+pub(crate) trait BufExt {
     fn get<T: Codec>(&mut self) -> Result<T>;
     fn get_var(&mut self) -> Result<u64>;
 }
@@ -107,7 +114,7 @@ impl<T: Buf> BufExt for T {
     }
 }
 
-pub trait BufMutExt {
+pub(crate) trait BufMutExt {
     fn write<T: Codec>(&mut self, x: T);
     fn write_var(&mut self, x: u64);
 }
--- a/quinn-proto/src/lib.rs
+++ b/quinn-proto/src/lib.rs
@@ -27,7 +27,6 @@ use std::{
 };
 
 mod cid_queue;
-#[doc(hidden)]
 pub mod coding;
 mod constant_time;
 mod range_set;


djc (approved):
Seems reasonable to me. @Ralith thoughts?

quinn-proto/src/coding.rs
```diff
@@ -1,18 +1,25 @@
...
 
+/// Unexpected end of buffer error.
 #[derive(Error, Debug, Copy, Clone, Eq, PartialEq)]
 #[error("unexpected end of buffer")]
 pub struct UnexpectedEnd;
 
+#[doc(hidden)]
 pub type Result<T> = ::std::result::Result<T, UnexpectedEnd>;
 
+/// Trait to handle encoding and decoding from byte buffers.
```
Ralith, line 18: This is not a general-purpose serialization interface, and we don't need to restate that it's a trait.

```suggestion
/// Infallible encoding and decoding of QUIC primitives
```

quinn-proto/src/coding.rs
```diff
@@ -92,6 +99,7 @@ impl Codec for Ipv6Addr {
     }
 }
 
+#[doc(hidden)]
```
Ralith, line 102: Could we make this and below `pub(crate)` instead?

quinn-proto/src/coding.rs
```diff
@@ -1,18 +1,25 @@
...
 use thiserror::Error;
 
 use crate::VarInt;
 
+/// Unexpected end of buffer error.
 #[derive(Error, Debug, Copy, Clone, Eq, PartialEq)]
 #[error("unexpected end of buffer")]
 pub struct UnexpectedEnd;
 
+#[doc(hidden)]
```
Ralith, line 15: Why is this hidden? It's part of `Codec`'s API.

quinn-proto/src/coding.rs
```diff
@@ -1,18 +1,25 @@
+//! Coding related traits.
+
 use std::net::{Ipv4Addr, Ipv6Addr};
 
 use bytes::{Buf, BufMut};
 use thiserror::Error;
 
 use crate::VarInt;
 
+/// Unexpected end of buffer error.
```
Ralith, line 10: Try to use different terms than those in the item's name to help people for whom the name isn't self-explanatory.
```suggestion
/// Error indicating that the provided buffer was too small
```

quinn-proto/src/coding.rs
```diff
@@ -1,18 +1,25 @@
...
 #[derive(Error, Debug, Copy, Clone, Eq, PartialEq)]
 #[error("unexpected end of buffer")]
 pub struct UnexpectedEnd;
 
+#[doc(hidden)]
 pub type Result<T> = ::std::result::Result<T, UnexpectedEnd>;
 
+/// Trait to handle encoding and decoding from byte buffers.
 pub trait Codec: Sized {
+    /// Try to decode `Self` from the provided buffer.
```
Ralith, line 20:
```suggestion
    /// Decode `Self` from the provided buffer, if the buffer is large enough
```

quinn-proto/src/coding.rs
```diff
@@ -1,18 +1,25 @@
...
     fn decode<B: Buf>(buf: &mut B) -> Result<Self>;
+    /// Encode `self` into the provided buffer.
```
Ralith, line 22: We should document exactly where the bytes end up in the buffer. A user might reasonably assume that they get written at the start.
```suggestion
    /// Append the encoding of `self` to the provided buffer
```

Ralith (changes requested):
The idea of exposing this seems reasonable. Some cosmetic feedback to help with clarity.

quinn-proto/src/lib.rs
```diff
@@ -27,7 +27,6 @@ use std::{
 };
 
 mod cid_queue;
-#[doc(hidden)]
```
djc, line 30: Should we re-export this in quinn?

Ralith:
Worth noting that this is *not* the most minimal possible way to export `VarInt` encoding/decoding, since it also exposes logic for a handful of other simple std types, but I think it is a fairly natural place to redraw the encapsulation boundary for the benefit of protocols that reuse simple encoding patterns from QUIC, which are not necessarily limited to `VarInt`.

quinn-proto/src/lib.rs
dignifiedquire: would be nice, but not necessary from my side

quinn-proto/src/coding.rs
dignifiedquire: done, seems to work out

quinn-proto/src/coding.rs
dignifiedquire: oversight, unhidden

Ralith (approved):

djc (approved):

djc:
Can you squash these commits?

PR #2079, opened 2024-12-13
fix(udp): retry send on first EINVAL

mxinden:
Android API level < 26 does not support the `libc::IP_TOS` control message. `sendmsg` calls with `libc::IP_TOS` return `libc::EINVAL`.

https://github.com/quinn-rs/quinn/pull/1516 added a fallback, not setting `libc::IP_TOS` on consecutive calls to `sendmsg` after a failure with `libc::EINVAL`. The current datagram would be dropped. Consecutive datagrams passed to `sendmsg` would succeed as they would be sent without `libc::IP_TOS` through the fallback.

Instead of dropping the first datagram on `libc::EINVAL`, this commit adds a retry for it without `libc::IP_TOS`.

This is e.g. relevant for Neqo. When establishing a QUIC connection, dropping the first datagram [delays connection establishment by 100ms](https://github.com/mozilla/neqo/blob/3001a3a56f2274eaafaa956fb394f0817f526ae7/neqo-transport/src/rtt.rs#L28). With the retry introduced in this commit, delay due to unsupported `libc::IP_TOS` should be negligible.

Closes https://github.com/quinn-rs/quinn/pull/1975.

//CC @KershawChang

---

This pull request contains 3 commits:
1. Add Android API level 25 CI step.
2. Add retry on `libc::EINVAL`.
3. Bump `quinn-udp` patch version to `v0.5.9`.

### DIFF
--- a/quinn-udp/tests/tests.rs
+++ b/quinn-udp/tests/tests.rs
@@ -225,9 +225,12 @@ fn test_send_recv(send: &Socket, recv: &Socket, transmit: Transmit) {
         );
         let send_v6 = send.local_addr().unwrap().as_socket().unwrap().is_ipv6();
         let recv_v6 = recv.local_addr().unwrap().as_socket().unwrap().is_ipv6();
-        let src = meta.addr.ip();
-        let dst = meta.dst_ip.unwrap();
-        for addr in [src, dst] {
+        let mut addresses = vec![meta.addr.ip()];
+        // Not populated on every OS. See `RecvMeta::dst_ip` for details.
+        if let Some(addr) = meta.dst_ip {
+            addresses.push(addr);
+        }
+        for addr in addresses {
             match (send_v6, recv_v6) {
                 (_, false) => assert_eq!(addr, Ipv4Addr::LOCALHOST),
                 // Windows gives us real IPv4 addrs, whereas *nix use IPv6-mapped IPv4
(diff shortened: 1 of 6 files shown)


mxinden:
>   error: failed to verify the checksum of `anstyle v1.0.10`

I will look into the failure on Solaris. I don't think it is related and thus doesn't need to block reviewing here.

Ralith (approved):
Nice!

quinn-udp/tests/tests.rs
```diff
@@ -240,7 +243,23 @@ fn test_send_recv(send: &Socket, recv: &Socket, transmit: Transmit) {
...
+        if match transmit.destination.ip() {
+            IpAddr::V4(_) => true,
+            IpAddr::V6(a) => a.to_ipv4_mapped().is_some(),
+        } && cfg!(target_os = "android")
+            && std::env::var("API_LEVEL")
+                .ok()
+                .and_then(|v| v.parse::<u32>().ok())
+                .expect("API_LEVEL environment variable to be set on Android")
+                <= 25
+        {
```
djc, line 256: Nit: don't love this multi-line, multi-level-of-indent `if` condition. Maybe bind the first clause and the `api_level` value?

I find the `expect()` explanation a little confusing since `[cfg_attr(
    not(any(target_os = "linux", target_os = "windows", target_os = "android")),
    ignore
)]` does not seem to restrict this test to Android only?

djc (approved):
How common is 25 out there?

mxinden:
> How common is 25 out there?

After a quick search, <= Android 7 (i.e. API level 25) should be somewhere between 1% and 10%.

https://gs.statcounter.com/android-version-market-share

Given the prevalence of Android in general, single digit percentage would justifies the additional complexity introduced here.

mxinden:
Agreed. Forgot to mention this earlier. Thank you for the question!

(Will address the other comment in a bit. Appreciate the reviews!)

quinn-udp/tests/tests.rs
mxinden: > Maybe bind the first clause

Done.

> and the `api_level` value?

Can't do that, as `API_LEVEL` will only be available on Android.



> I find the `expect()` explanation a little confusing since `[cfg_attr( not(any(target_os = "linux", target_os = "windows", target_os = "android")), ignore )]` does not seem to restrict this test to Android only?

Given that Rust short-circuits boolean operations (aka. [lazy boolean operators](https://doc.rust-lang.org/reference/expressions/operator-expr.html#lazy-boolean-operators)) the third boolean expression is only evaluated if the first and second is `true`.

mxinden:
@djc can you take another look? I restructured the `if` statement in the unit test.

PR #2080, opened 2024-12-13
refactor(udp): remove unnecessary `return`

mxinden:
No need to `return Ok(())` early, given the final `return Ok(())` at the end of the function. Makes it consistent with other `send` implementations. Makes it consistent with early returns being errors only.

Follow-up to https://github.com/quinn-rs/quinn/pull/2074#discussion_r1863944160. I have not yet been able to come up with a better refactoring.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -341,8 +341,6 @@ fn send(
                     if e.raw_os_error() != Some(libc::EMSGSIZE) {
                         return Err(e);
                     }
-
-                    return Ok(());
                 }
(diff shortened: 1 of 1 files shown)


djc (approved):

Ralith (changes requested):
I don't think this is equivalent. If `sendmsg` fails with `EMSGSIZE`, then with this change the loop will continue rather than bailing out with success.

Ralith (approved):
Ah, you're right, miscounted the brackets. Thanks!

PR #2083, opened 2024-12-14
Make SystemTime mockable

gretchenfrage:
Only 1 of this and #2089 should be merged, and the other should be closed. To help automate this process, I am marking them as closing each other.

Closes #2089

---

We already do this for Instants. However, there are currently two places in main that we call SystemTime::now from proto. This PR makes them mockable for tests.

- `quinn::Runtime` gains new `system_now(&self) -> SystemTime` method. This has a default implementation, so no semver breakage for quinn.
- `proto::Endpoint::handle` gains a `system_now: SystemTime` argument. This causes semver breakage for proto.
- The `SystemTime` provided in `handle` is stored in `Incoming.received_at_system`, just like `Incoming.received_at`, and may be utilized in `Endpoint::retry`.
- Tests are adapted by making each `TestEndpoint` have its own `Instant` -> `SystemTime` conversion.

### DIFF
--- a/quinn/src/runtime.rs
+++ b/quinn/src/runtime.rs
@@ -6,7 +6,7 @@ use std::{
     pin::Pin,
     sync::Arc,
     task::{Context, Poll},
-    time::Instant,
+    time::{Instant, SystemTime},
 };
 
 use udp::{RecvMeta, Transmit};
@@ -15,15 +15,25 @@ use udp::{RecvMeta, Transmit};
 pub trait Runtime: Send + Sync + Debug + 'static {
     /// Construct a timer that will expire at `i`
     fn new_timer(&self, i: Instant) -> Pin<Box<dyn AsyncTimer>>;
+
     /// Drive `future` to completion in the background
     fn spawn(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>);
+
     /// Convert `t` into the socket type used by this runtime
     fn wrap_udp_socket(&self, t: std::net::UdpSocket) -> io::Result<Arc<dyn AsyncUdpSocket>>;
+
     /// Look up the current time
     ///
     /// Allows simulating the flow of time for testing.
     fn now(&self) -> Instant {
         Instant::now()
+    }
+
+    /// Look up the current [`SystemTime`]
+    ///
+    /// Allows simulating the flow of time for testing.
+    fn system_now(&self) -> SystemTime {
+        SystemTime::now()
     }
 }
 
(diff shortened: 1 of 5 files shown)


quinn/src/runtime.rs
```diff
@@ -25,6 +25,12 @@ pub trait Runtime: Send + Sync + Debug + 'static {
     fn now(&self) -> Instant {
         Instant::now()
     }
+    /// Look up the current [`SystemTime`]
```
Ralith, line 28: style: empty line between definitions

Ralith (changes requested):
Unfortunately, as `quinn-proto` is a public dependency, breaking changes to it are necessarily breaking changes to `quinn`. That said, minimizing API churn is still valuable regardless of semver.

It feels a little funny to be invoking `SystemTime::now` on every received packet receipt when it is almost never actually needed. I wonder if a trait object in `EndpointConfig` might be a more proportionate interface, and coincidentally a more semver-friendly one besides? Interested in perspectives.

gretchenfrage:
> I wonder if a trait object in EndpointConfig might be a more proportionate interface, and coincidentally a more semver-friendly one besides?

Maybe, but it makes it more complicated to mock. At worst, mocking could resort to feeling like they have to create an implementation that accesses thread-locals to thread this through method calls in a side channel.

gretchenfrage:
> > I wonder if a trait object in EndpointConfig might be a more proportionate interface, and coincidentally a more semver-friendly one besides?
> 
> Maybe, but it makes it more complicated to mock. At worst, mocking could resort to feeling like they have to create an implementation that accesses thread-locals to thread this through method calls in a side channel.

A secret third option would be to pass in a callback to `handle_packet` (an `impl FnOnce() -> SystemTime` or equivalent trait).

quinn/src/runtime.rs
gretchenfrage: Then I shall fix it for the others in this file too.

gretchenfrage:
> > > I wonder if a trait object in EndpointConfig might be a more proportionate interface, and coincidentally a more semver-friendly one besides?
> > 
> > 
> > Maybe, but it makes it more complicated to mock. At worst, mocking could resort to feeling like they have to create an implementation that accesses thread-locals to thread this through method calls in a side channel.
> 
> A secret third option would be to pass in a callback to `handle_packet` (an `impl FnOnce() -> SystemTime` or equivalent trait).

See 8f1408e008dcefcef15fb62b0d0188d8db946017 (pushed to side branch as a demonstration of what it could be like)

Ralith:
> At worst, mocking could resort to feeling like they have to create an implementation that accesses thread-locals to thread this through method calls in a side channel.

Surely simple shared state would be adequate?

gretchenfrage:
Yeah we should probably just go with #2089 it's way simpler.

Ralith:
That does look much more manageable, thanks!

gretchenfrage:
Closing in favor of #2089

PR #1912, opened 2024-06-30
Utilize NEW_TOKEN frames

gretchenfrage:
The server now sends the client NEW_TOKEN frames, and the client now stores and utilizes them.

The main motivation is that this allows 0.5-RTT data to not be subject to anti-amplification limits. This is a scenario likely to occur in HTTP/3 requests, as one example: a client makes a 0-RTT GET request for something like a jpeg, such that the response will be much bigger than the request, and so unless NEW_TOKEN frames are used, the response may begin to be transmitted but then hit the anti-amplification limit and have to pause until the full 1-RTT handshake completes.

For example, here's some experimental data that should be similar in the relevant ways:

- The client sends the server an integer and the server responds with that number of bytes
- They do it in 0-RTT if they can
- For each iteration the client endpoint does it twice and measures its request/response time from the second time it does it (so it will have 0-RTT and NEW_TOKEN material)
- 100ms localhost latency was simulated by running `sudo tc qdisc add dev lo root netem delay 100ms` (and undone with `sudo tc qdisc del dev lo root netem`)

This experiment was performed on Nov/24 with 2edf192511873a52093dd57b9e70eb4b27c442cd as main and 478b3259e62d8a5472e08759a7143c34b59718f1 as feature. 

new-token-nov24-graph

For responses in a certain size range, avoiding the anti-amplification limits by using NEW_TOKEN frames made the request/response complete in 1 RTT on this branch versus 2 RTT on main.

<details>
<summary>Reproducible experimental setup</summary>

`newtoken.rs` can be placed into `quinn/examples/`:

```rs

use std::{
    sync::Arc,
    net::ToSocketAddrs as _,
};
use anyhow::Error;
use quinn::*;
use tracing::*;
use tracing_subscriber::prelude::*;


#[tokio::main]
async fn main() -> Result<(), Error> {
    // init logging
    let log_fmt = tracing_subscriber::fmt::format()
        .compact()
        .with_timer(tracing_subscriber::fmt::time::uptime())
        .with_line_number(true);
    let stdout_log = tracing_subscriber::fmt::layer()
        .event_format(log_fmt)
        .with_writer(std::io::stderr);
    let log_filter = tracing_subscriber::EnvFilter::new(
        std::env::var(tracing_subscriber::EnvFilter::DEFAULT_ENV).unwrap_or("info".into())
    );
    let log_subscriber = tracing_subscriber::Registry::default()
        .with(log_filter)
        .with(stdout_log);
[...]

PR #2084, opened 2024-12-14
proto: Connection side enum

gretchenfrage:
Some notable differences from how I approached this in #1912:

- Calling it `ConnectionSide` rather than `SideState`
- Call the field `Connection.side` rather than `Connection.side_state`. The causes changes to be necessary to a different, but smaller set of lines of code, since now every existing instance of `self.side.is_client()` or `self.side.is_server()` just works as-is. And it's terser.
- Offloaded more logic to methods on `SideArgs` and an `impl From<SideArgs> for ConnectionSide`, in the interest of taking some complexity load off of `Connection::new`, which is already over 100 lines.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -129,7 +129,6 @@ use timer::{Timer, TimerTable};
 /// events or timeouts with different instants must not be interleaved.
 pub struct Connection {
     endpoint_config: Arc<EndpointConfig>,
-    server_config: Option<Arc<ServerConfig>>,
     config: Arc<TransportConfig>,
     rng: StdRng,
     crypto: Box<dyn crypto::Session>,
@@ -145,7 +144,7 @@ pub struct Connection {
     allow_mtud: bool,
     prev_path: Option<(ConnectionId, PathData)>,
     state: State,
-    side: Side,
+    side: ConnectionSide,
     /// Whether or not 0-RTT was enabled during the handshake. Does not imply acceptance.
     zero_rtt_enabled: bool,
     /// Set if 0-RTT is supported, then cleared when no longer needed.
@@ -191,9 +190,6 @@ pub struct Connection {
     authentication_failures: u64,
     /// Why the connection was lost, if it has been
     error: Option<ConnectionError>,
-    /// Sent in every outgoing Initial packet. Always empty for servers and after Initial keys are
-    /// discarded.
-    retry_token: Bytes,
     /// Identifies Data-space packet numbers to skip. Not used in earlier spaces.
     packet_number_filter: PacketNumberFilter,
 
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -1076,7 +1075,10 @@ impl Connection {
                 // forbids migration, drop the datagram. This could be relaxed to heuristically
                 // permit NAT-rebinding-like migration.
                 if remote != self.path.remote
-                    && self.server_config.as_ref().map_or(true, |x| !x.migration)
+                    && match &self.side {
+                        ConnectionSide::Server { server_config } => !server_config.migration,
+                        ConnectionSide::Client { .. } => true,
+                    }
```
Ralith, line 1073: nit: lift this out into a getter on `ConnectionSide`?

Ralith (approved):

quinn-proto/src/connection/mod.rs
```diff
@@ -3618,6 +3612,89 @@ impl fmt::Debug for Connection {
...
+
+    fn is_client(&self) -> bool {
+        self.side().is_client()
+    }
+
+    fn is_server(&self) -> bool {
+        self.side().is_server()
+    }
+
+    fn migration(&self) -> bool {
```
Ralith, line 3642: This name seems to be the inverse of its behavior and use. Also, `is_` for predicates.

gretchenfrage: Good point. Inverted both the method and its usage, and gave more descriptive name:

```diff
diff --git a/quinn-proto/src/connection/mod.rs b/quinn-proto/src/connection/mod.rs
index 1f2e803f..6dc4157e 100644
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1074,7 +1074,7 @@ impl Connection {
                 // If this packet could initiate a migration and we're a client or a server that
                 // forbids migration, drop the datagram. This could be relaxed to heuristically
                 // permit NAT-rebinding-like migration.
-                if remote != self.path.remote && self.side.migration() {
+                if remote != self.path.remote && !self.side.remote_may_migrate() {
                     trace!("discarding packet from unrecognized peer {}", remote);
                     return;
                 }
@@ -3647,10 +3647,10 @@ impl ConnectionSide {
         self.side().is_server()
     }
 
-    fn migration(&self) -> bool {
+    fn remote_may_migrate(&self) -> bool {
         match self {
-            ConnectionSide::Server { server_config } => !server_config.migration,
-            ConnectionSide::Client { .. } => true,
+            ConnectionSide::Server { server_config } => server_config.migration,
+            ConnectionSide::Client { .. } => false,
         }
     }
 }
```

Ralith (approved):

quinn-proto/src/connection/mod.rs
```diff
@@ -3618,6 +3620,41 @@ impl fmt::Debug for Connection {
...
+    fn is_server(&self) -> bool {
+        self.side().is_server()
+    }
+
+    fn remote_may_migrate(&self) -> bool {
+        match self {
+            Self::Server { server_config } => server_config.migration,
+            Self::Client { .. } => false,
+        }
+    }
```
djc, line 3647: Ordering nit: let's have `remote_may_migrate()` first, then order `is_client()` and `is_server()` before `side()` (since they depend on it).

quinn-proto/src/connection/mod.rs
```diff
@@ -3655,6 +3647,54 @@ impl ConnectionSide {
...
+    Client,
+    Server {
+        server_config: Arc<ServerConfig>,
+        pref_addr_cid: Option<ConnectionId>,
+        path_validated: bool,
+    },
+}
+
+impl SideArgs {
+    pub(crate) fn side(&self) -> Side {
```
djc, line 3676: Ordering nit: let's order this last.

djc (approved):

quinn-proto/src/connection/mod.rs
gretchenfrage: Both done.

PR #2085, opened 2024-12-14
proto: Refactor TokenDecodeError

gretchenfrage:
This PR consists solely of renames and docs changes.

- Renames enum `TokenDecodeError` -> `ValidationError`

  `TokenDecodeError` is used when that there was an error decoding a token, but is also when there was no error decoding the token but the token contained an address not applicable to the current situation. Thus, this rename more accurately reflects the purpose of the error.
- Renames variant `UnknownToken` -> `Unusable` ~~`Ignore`~~

  `UnknownToken` is used when a token cannot be decoded or deserialized, and triggers the token to be ignored. Thus, it's name is currently an accurate description of how it's used, however, I found it very confusing why we use it like that, and had to get Ralith to [explain](https://discord.com/channels/976380008299917365/1063547094863978677/1251350017529090049) it to me. Moreover, in the future, this error would be applicable to situations where a token _can_ be unambiguously decrypted/decoded as a token from a NEW_TOKEN frame, but is simply not valid. In that situation, the variant name would cease to be an accurate description of how it's used. Thus, we rename it to a more accurately describe the sorts of situations it would apply to, and leave a detailed comment on the variant explaining the rationale. Relatedly, we remove a comment from `endpoint.rs` for being now-redundant.

  At one point, in #1912, I [tried](https://github.com/quinn-rs/quinn/pull/1912#discussion_r1694059352) the name `InvalidMaybeNewToken`, but Ralith seemed hesitant. While that name was more descriptive, it's also pretty cumbersome.
- Renames variant `WrongAddress` -> `InvalidRetry`

  `WrongAddress` is used when a token can be successfully decoded and deserialized but has the wrong client address, and triggers the connection to be closed. Thus, it's name is currently an accurate description of how it's used. However, if in the future we start processing tokens from NEW_TOKEN frames, this error would _not_ be applicable to situations where a token can be unambiguously decrypted/decoded as a token from a NEW_TOKEN frame but has the wrong client address, as the appropriate way to handle that situation would be to ignore the token rather than close the connection, so we would have to use a different error. In that situation, the variant name would cease to be an accurate description of how it's used. Thus, we rename it to a more accurately describe the sorts of situations it would apply to.

  An additional conceivable benefit of this name is that it would make it easier to potentially add future cases where a token can be unambiguously _identified_ as a token from a Retry packet, but still cannot be fully decrypted or deserialized, or where it is otherwise for some reason unusable.


quinn-proto/src/endpoint.rs
```diff
@@ -29,7 +29,7 @@ use crate::{
         ConnectionEvent, ConnectionEventInner, ConnectionId, DatagramConnectionEvent, EcnCodepoint,
         EndpointEvent, EndpointEventInner, IssuedCid,
     },
-    token::TokenDecodeError,
+    token::ValidationError,
```
Ralith, line 32: Either name this `TokenValidationError`, or always refer to it qualified as `token::ValidationError`. Bare `ValidationError` is ambiguous outside of its module of origin.

quinn-proto/src/token.rs
```diff
@@ -99,19 +93,36 @@ fn decode_addr<B: Buf>(buf: &mut B) -> Option<SocketAddr> {
...
+    /// balancer. In such cases we proceed as if there was no token.
+    ///
+    /// [_RFC 9000 § 8.1.3:_](https://www.rfc-editor.org/rfc/rfc9000.html#section-8.1.3-10)
+    ///
+    /// > If the token is invalid, then the server SHOULD proceed as if the client did not have a
+    /// > validated address, including potentially sending a Retry packet.
+    ///
+    /// That said, this may also be used when a token _can_ be unambiguously decrypted/decoded as a
+    /// token from a NEW_TOKEN frame, but is simply not valid.
+    Ignore,
```
Ralith, line 116: nit: It feels weird for this to be a verb, when the other variant is (and error variants in general are) a noun. I like `Unknown` a little more in that respect, though I'm open to other ideas.

Ralith (approved):

quinn-proto/src/token.rs
gretchenfrage: Yeah, it's a bit unfortunate. I touched on this above, but to recap: "Unknown" has the problem of, "why do we return Unknown for a token from NEW_TOKEN that's expired or has the wrong address, we clearly know what this token is?", and something both nouny and accurate would have to capture "this is not valid, and it could conceivably be not from a Retry," which is cumbersome.

Ralith: Maybe `Ignored`?

gretchenfrage: Maybe, but if I saw `Ignored` decontextualized, I would imagine it to mean "the problem is that this has been ignored" before I'd imagine it to mean 'the error is that this thing should be ignored", or something

quinn-proto/src/endpoint.rs
gretchenfrage: Leaving as `ValidationError`, and changed the (1) usage in `endpoint.rs` to specify `token::ValidationError::Ignore`. Because:

- `TokenValidationError` is long.
- I anticipate future PRs likely being able to make this error entirely private to `token.rs` anyways.

quinn-proto/src/token.rs
gretchenfrage: Maybe `Unusable`?

gretchenfrage: I kind of like `Unusable`...

Ralith: Works for me!

Ralith (approved):

quinn-proto/src/endpoint.rs
Ralith: I've always liked terse names that get their context from the module system anyway.

djc (approved):

djc:
Nice!

Issue #1757, opened 2024-02-16
Misc cleanup points noticed in #1752


gretchenfrage:
I'm creating this issue as a list of miscellaneous things that we're noticing should be cleaned up in the course of developing #1752  but which would not be in scope for that issue itself:

- ~~[`quinn::Connecting` has a `#[must_use]` annotation denoting that it is a future which is obsolete as Rust now does that automatically](https://github.com/quinn-rs/quinn/pull/1752#discussion_r1488869073)~~ Fixed by #2087
- ~~[`quinn::Connecting::local_ip` is missing a comment stating that it will panic if called after `poll` has returned `Ready` like its `remote_address` method has](https://github.com/quinn-rs/quinn/pull/1752#discussion_r1488867521)~~ Fixed by #2088
- ~~[`quinn::Connecting::remote_address` has a trailing punctuation on the first line of its doc comment, which is inconsistent with general project style](https://github.com/quinn-rs/quinn/pull/1752#discussion_r1488867226)~~ Fixed by #2088
- ~~quinn does not re-export `quinn_proto::TransportError` and other downstream-API-exposed types despite that being a public part of quinn's API via [quinn's re-exporting of `quinn_proto::ConnectionError`](https://docs.rs/quinn/latest/quinn/enum.ConnectionError.html)~~ Fixed by #2031

gretchenfrage:
Yay we fixed it all! :D

PR #2087, opened 2024-12-14
quinn: Remove obsolete must_use for futures

gretchenfrage:
Various future types had boilerplate `must_use` annotations. Rust now does that automatically for futures. Removes all of them that didn't have some more specific message.

Noted in #1757.


Ralith (approved):
Thanks!

PR #2091, opened 2024-12-15
Gate PLPMTUD should_panic test on debug_assertions

### LINKED ISSUE #2090: PLPMTUD test fails to panic in release mode
On current main (73545b65b3b99659f84beff2a46c143d5ee9b99d), running:

```sh
cargo test --release -p quinn-proto -- connection::mtud::tests::mtu_discovery_with_peer_max_udp_payload_size_after_search_panics
```

Gives the following failure:

```
---- connection::mtud::tests::mtu_discovery_with_peer_max_udp_payload_size_after_search_panics stdout ----
note: test did not panic as expected
```

This only occurs if the `--release` flag is passed into the test run. The cause of it is `quinn-proto/src/connection/mtud.rs:88` being a debug assertion:

```rs
// MTUD is only active after the connection has been fully established, so it is
// guaranteed we will receive the peer's transport parameters before we start probing
debug_assert!(matches!(state.phase, Phase::Initial));
```

Related to #1510

gretchenfrage:
Closes #2090

### DIFF
--- a/quinn-proto/src/connection/mtud.rs
+++ b/quinn-proto/src/connection/mtud.rs
@@ -731,6 +731,7 @@ mod tests {
         assert!(completed(&mtud));
     }
 
+    #[cfg(debug_assertions)]
     #[test]
     #[should_panic]
     fn mtu_discovery_with_peer_max_udp_payload_size_after_search_panics() {


Ralith (approved):

djc:
Thanks!

PR #2071, opened 2024-11-28
fix(udp): do not enable URO on Windows on ARM

mxinden:
On Windows on ARM with Windows Subsystem for Linux (WSL) `WSA_RECVMSG` does not return the segment size of coalesced UDP datagrams. See https://github.com/quinn-rs/quinn/issues/2041 for details.

While lacking a fix for the root cause, don't enable URO, i.e. don't coalesce UDP datagrams, on Windows on ARM.

Issue #2041, opened 2024-11-14
Potential bug in GRO code for Windows?


thomaseizinger:
We are seeing an issue on a Windows machine where `quinn-udp` gives us packets larger than the configured MTU (2540 and 3810). Here is how we split the packets: https://github.com/firezone/firezone/blob/5437c3e2df001c8c491749c7bff0c063ccdf39e1/rust/socket-factory/src/lib.rs#L210-L255

This code has been working on Linux for several months and we have also not had reports that it doesn't work on Windows. Recently however a customer filed a bug report and looking through the logs, I discovered that we receive packets (i.e. `DatagramIn` in the above snippet) that are much larger than expected.

Going through `quinn-udp`'s source code, I developed two theories and would be interested in hearing your thoughts on it:

1. `stride` is being read incorrectly (endianess-issue?) and instead of being smaller than `len`, it is bigger than `len`, meaning we never end up splitting the packets despite it being coalesced. https://github.com/quinn-rs/quinn/blob/d23e4e494f7446e21184bf58acd17a861ae73bba/quinn-udp/src/windows.rs#L241
2. The buffer for control messages is too small, thus Windows fails to fit the `UDP_COALESCED_INFO` message into it and can't tell us that the buffer is coalesced even though it is.

Unfortunately, I've not been able to reproduce this on my own hardware :(

stormshield-damiend:
Does the windows run on intel CPU or on ARM CPU ?

mxinden:
We see a similar issue on Windows on ARM, see summary in [last comment on this thread](https://bugzilla.mozilla.org/show_bug.cgi?id=1916558#c39).

thomaseizinger:
> Does the windows run on intel CPU or on ARM CPU ?

We see this issue on an x64_86 CPU.

thomaseizinger:
> > The buffer for control messages is too small, thus Windows fails to fit the `UDP_COALESCED_INFO` message into it and can't tell us that the buffer is coalesced even though it is.
> 
> Interesting. I will try reproducing the issue with a larger buffer. For what it is worth, here is [MSQUIC's buffer length](https://github.com/microsoft/msquic/blob/e3ffc71a375d85b1216348b2ffa1e13653213e26/src/platform/datapath_winuser.c#L149-L158).

I thought about this a bit more and I am no longer sure that is a viable possibility unless we receive a lot of other control messages that we don't match on.

stormshield-damiend:
[...]

PR #2092, opened 2024-12-16
fix(udp): make GRO (i.e. URO) optional on Windows, off by default

mxinden:
We have multiple bug reports for GRO (i.e. URO) on Windows with no solution in sight.

- https://github.com/quinn-rs/quinn/issues/2041
- https://bugzilla.mozilla.org/show_bug.cgi?id=1916558

While it previously seemed like this might be bound to _Windows on ARM_ only, that assumption can no longer be made, see e.g. https://github.com/quinn-rs/quinn/issues/2041#issuecomment-2542435062 and [Bugzilla comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1916558#c54). Thus our previous attempt (https://github.com/quinn-rs/quinn/pull/2071) is not enough.

I suggest by default disabling GRO (i.e. URO) on all of Windows, with an advanced option for users to enable it via `UdpSocketState::set_gro(true)`.

Users that control their hardware, i.e. know the hardware supports URO, can enable it.

Users that want to implement a heuristic, as e.g. suggested by @thomaseizinger  in https://github.com/quinn-rs/quinn/issues/2041#issuecomment-2542435062, can enable or disable it programmatically. 

Reverts https://github.com/quinn-rs/quinn/pull/2071.
Tracked in https://github.com/quinn-rs/quinn/issues/2041

### DIFF
--- a/quinn-udp/src/windows.rs
+++ b/quinn-udp/src/windows.rs
@@ -10,13 +10,7 @@ use std::{
 
 use libc::{c_int, c_uint};
 use once_cell::sync::Lazy;
-use windows_sys::Win32::{
-    Networking::WinSock,
-    System::{
-        SystemInformation::IMAGE_FILE_MACHINE_ARM64,
-        Threading::{GetCurrentProcess, IsWow64Process2},
-    },
-};
+use windows_sys::Win32::Networking::WinSock;
 
 use crate::{
     cmsg::{self, CMsgHdr},
@@ -113,33 +107,16 @@ impl UdpSocketState {
             )?;
         }
 
-        match &*IS_WINDOWS_ON_ARM {
-            Ok(true) => {
-                // Bug on Windows on ARM, not receiving `UDP_COALESCED_INFO` `CMSG`
-                // when _Virtual Machine Platform_ feature enabled. See
-                // <https://github.com/quinn-rs/quinn/issues/2041> for details.
-                debug!("detected Windows on ARM host thus not enabling URO")
-            }
-            Ok(false) => {
-                // Opportunistically try to enable URO
-                let result = set_socket_option(
-                    &*socket.0,
-                    WinSock::IPPROTO_UDP,
-                    WinSock::UDP_RECV_MAX_COALESCED_SIZE,
-                    // u32 per
-                    // https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-udp-socket-options.
-                    // Choice of 2^16 - 1 inspired by msquic.
-                    u16::MAX as u32,
-                );
-
-                if let Err(_e) = result {
-                    debug!("failed to enable URO: {_e}");
-                }
-            }
-            Err(_e) => {
-                debug!("failed to detect host system thus not enabling URO: {_e}");
-            }
-        }
+        // Opportunistically try to enable GRO
+        _ = set_socket_option(
+            &*socket.0,
+            WinSock::IPPROTO_UDP,
+            WinSock::UDP_RECV_MAX_COALESCED_SIZE,
+            // u32 per
+            // https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-udp-socket-options.
+            // Choice of 2^16 - 1 inspired by msquic.
+            u16::MAX as u32,
+        );
 
         let now = Instant::now();
         Ok(Self {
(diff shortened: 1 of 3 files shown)


quinn-udp/src/windows.rs
```diff
@@ -107,23 +107,35 @@ impl UdpSocketState {
...
+    ///
+    /// <https://learn.microsoft.com/en-us/windows-hardware/drivers/network/udp-rsc-offload>
+    ///
+    /// Disabled by default due to <https://github.com/quinn-rs/quinn/issues/2041>.
+    pub fn set_gro(&self, enable: bool) -> io::Result<()> {
+        set_socket_option(
+            &*socket.0,
+            WinSock::IPPROTO_UDP,
+            WinSock::UDP_RECV_MAX_COALESCED_SIZE,
+            if enable {
```
djc, line 128: Style nit: `match` is probably nicer here.

quinn-udp/src/windows.rs
```diff
@@ -107,23 +107,35 @@ impl UdpSocketState {
...
-            u16::MAX as u32,
-        );
-
         let now = Instant::now();
         Ok(Self {
             last_send_error: Mutex::new(now.checked_sub(2 * IO_ERROR_LOG_INTERVAL).unwrap_or(now)),
         })
     }
 
+    /// Enable or disable receive offloading.
```
djc, line 116: Are there constraints on when this could/should be called?

mxinden: As in which state of the lifecycle of the socket? As far as I am aware, there are no constraints, e.g. you could toggle it between each read. That said, there might be some delay, given that, I assume, the NIC might have already coalesced some datagrams.

djc: Okay, that's fine then.

quinn-udp/src/windows.rs
thomaseizinger: Could also be written as `enable.then_some(u16::MAX).or(0)`.

thomaseizinger (approved):
LGTM.

Thanks Max!

quinn-udp/src/windows.rs
mxinden: I am fine either way. Updated with the second (`then_some`) suggestion. Happy to change to the first (`match`).

mxinden: @djc switched to `match` in the latest iteration.

quinn-udp/src/windows.rs
```diff
@@ -107,23 +107,34 @@ impl UdpSocketState {
...
+    ///
+    /// Also referred to as UDP Receive Segment Coalescing Offload (URO) on Windows.
+    ///
+    /// <https://learn.microsoft.com/en-us/windows-hardware/drivers/network/udp-rsc-offload>
+    ///
+    /// Disabled by default due to <https://github.com/quinn-rs/quinn/issues/2041>.
+    pub fn set_gro(&self, socket: UdpSockRef<'_>, enable: bool) -> io::Result<()> {
```
Ralith, line 123: Should we also expose this method on other platforms, just with a different default? Platform-specific APIs are a bit of a footgun.

quinn-udp/src/windows.rs
```diff
@@ -107,23 +107,34 @@ impl UdpSocketState {
...
         })
     }
 
+    /// Enable or disable receive offloading.
+    ///
+    /// Also referred to as UDP Receive Segment Coalescing Offload (URO) on Windows.
+    ///
+    /// <https://learn.microsoft.com/en-us/windows-hardware/drivers/network/udp-rsc-offload>
+    ///
+    /// Disabled by default due to <https://github.com/quinn-rs/quinn/issues/2041>.
```
Ralith, line 122: s/by default/by default on Windows/

Ralith (changes requested):

quinn-udp/src/windows.rs
mxinden: > Platform-specific APIs are a bit of a footgun.

I see your point. That said, the following arguments made me choose a platform specific API:

- What would `set_gro` do on platforms that don't do receive offloading, e.g. MacOS, BSDs, ...? In my eyes e.g. adding a no-op `set_gro` is a larger inconsistency (or surprise) than a platform specific API.

- I am not aware of any issues related to GRO on Linux and Android. Why offer and more importantly maintain a `fn` without a use-case?

Given the subtle issues with https://github.com/quinn-rs/quinn/issues/2041 across a wide variety of machines, I don't expect `set_gro` to be used by anyone but a tiny (brave) minority of folks that either control their hardware all the way, or implement a sophisticated heuristic to disable it programmatically.

@Ralith with the above in mind, do you still want me to add `set_gro` to the other platforms?

Ralith: > In my eyes e.g. adding a no-op set_gro is a larger inconsistency (or surprise) than a platform specific API. [...] Why offer and more importantly maintain a fn without a use-case?

I don't think this is much different from e.g. the `ecn` and `src_ip` fields of `Transmit`, which are silent no-ops on some platforms. The intent is that you can write platform-agnostic code against a fixed API and get the best possible behavior for each platform. Calling `set_gro(true)` and not getting GRO isn't harmful, or even reliably detectable.

For completeness, note one use-case for disabling GRO exists: Wireshark [still cannot handle](https://gitlab.com/wireshark/wireshark/-/issues/19109) GRO reliably, seemingly due to limitations in the kernel packet capture interface. Users trying to gather useful packet captures may wish to disable GSO/GRO to mitigate this. This comes up very rarely in practice, though. 

> I don't expect set_gro to be used by anyone but a tiny (brave) minority

That's a fair point. Idiomatic use as outlined above probably shouldn't ever call this API in the first place, since it's generally quinn-udp's responsibility to do the right thing by default. On that basis I'm happy to defer to your judgement.

mxinden: > > In my eyes e.g. adding a no-op set_gro is a larger inconsistency (or surprise) than a platform specific API. [...] Why offer and more importantly maintain a fn without a use-case?
> 
> I don't think this is much different from e.g. the `ecn` and `src_ip` fields of `Transmit`, which are silent no-ops on some platforms.

That is a good point!

mxinden:
Thank you for the quick help here @djc @Ralith and @thomaseizinger!

@djc @Ralith could either of you publish `quinn-udp` `v0.5.9` to crates.io? The crate's patch version is already increased as part of this pull request.

djc:
Yup, getting to it in ~15m!

djc:
* Published quinn-udp v0.5.9 at registry `crates-io`
* [new tag]           quinn-udp-0.5.9 -> quinn-udp-0.5.9
* [Release notes](https://github.com/quinn-rs/quinn/releases/tag/quinn-udp-0.5.9)

PR #2094, opened 2024-12-16
Factor out `IncomingToken`

gretchenfrage:
Improved decoupling between modules, makes code flow easier to understand IMO, and sets us up to have an easier time adding more internal variants to `RetryToken`, for NEW_TOKEN stuff.

### DIFF
--- a/quinn-proto/src/token.rs
+++ b/quinn-proto/src/token.rs
@@ -8,8 +8,9 @@ use bytes::{Buf, BufMut};
 use crate::{
     coding::{BufExt, BufMutExt},
     crypto::{CryptoError, HandshakeTokenKey, HmacKey},
+    packet::InitialHeader,
     shared::ConnectionId,
-    Duration, SystemTime, RESET_TOKEN_SIZE, UNIX_EPOCH,
+    Duration, ServerConfig, SystemTime, RESET_TOKEN_SIZE, UNIX_EPOCH,
 };
 
 pub(crate) struct RetryToken {
@@ -43,7 +44,7 @@ impl RetryToken {
         buf
     }
 
-    pub(crate) fn from_bytes(
+    fn from_bytes(
         key: &dyn HandshakeTokenKey,
         address: &SocketAddr,
         retry_src_cid: &ConnectionId,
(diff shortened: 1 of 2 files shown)


quinn-proto/src/token.rs
```diff
@@ -71,6 +72,22 @@ impl RetryToken {
             issued,
         })
     }
+
+    /// Ensure that this token validates an `Incoming`, and construct its token state
+    fn validate(
```
gretchenfrage, line 77: Djc, I know you've been [interested](https://github.com/quinn-rs/quinn/pull/1912#discussion_r1886597384) in in-lining this into `validate`. However, the place in `validate` where this is called is within an `and_then` closure, and you [didn't](https://github.com/quinn-rs/quinn/pull/1912#discussion_r1856334781) really like of having a big amount of logic in-lined into an `and_then` either. On that matter, going through with NEW_TOKEN usage would likely add even more logic here--although of course I can do it one way for this PR and then change it back in the future.

In any case, I think that having this method be separate makes the code easier to understand, because it very naturally reifies how I mentally compartmentalize different concerns in processing tokens.

Still, though, I can inline it if wanted.

quinn-proto/src/token.rs
```diff
@@ -185,6 +202,45 @@ pub(crate) struct IncomingToken {
...
+            orig_dst_cid: header.dst_cid,
+        }
+    }
+
+    /// Construct for an `Incoming` given the first packet header, or error if the connection
+    /// cannot be established
+    pub(crate) fn from_header(
+        header: &InitialHeader,
+        server_config: &ServerConfig,
+        remote_address: SocketAddr,
```
gretchenfrage, line 219: Note: Most logic in this module is passing around `SocketAddr` by reference. However, in my opinion, for small `Copy` types, one should just pass them by value and let the compiler figure out what it wants to do.

djc: Yup, we should probably rely on `Copy` more for this.

gretchenfrage: Great. I will likely create a "refactor token encoding" PR after this one that will do that among other things.

Ralith (approved):

gretchenfrage:
CI fail is due to flake.

djc (approved):

quinn-proto/src/token.rs
djc: Thanks for flagging this, going to merge this as-is and see if there's further refactoring that makes sense to me.

PR #2099, opened 2024-12-20
proto: replace hidden field with From impl

djc:
r? @gretchenfrage


gretchenfrage (commented):
Can we get a constructor as well?

djc:
Why? It doesn't seem like we need it, and that feels like a feature.

gretchenfrage:
Just seems like it could be useful for any real or hypothetical protocols built on top of QUIC which involve application data transmitting and then utilizing QUIC stream IDs. If you don't wanna do it now it can be added later if it ever comes up.

gretchenfrage (approved):

Ralith (approved):

Issue #2097, opened 2024-12-20
Usages of `#[doc(hidden)] pub`


gretchenfrage:
I found 13 places in this repo that use `#[doc(hidden)]` on pub types for various reasons. In my opinion, there is usually a better way to do something than that, and from what I'm seeing, most of these occurrences seem to date back to the 2018-2019 era of the project.

The way I see it:

 - If it's needed for tests, mark it `#[cfg(test)]`.
 - If it's needed for some weird thing, like perf testing or an interop container, gate it behind a feature flag, and document that that feature flag is not covered by semver commitments. It extra defensiveness desired, mimic [the thing the Hyper project does](https://github.com/hyperium/hyper/blob/30f2961e89eb306780d856e6e2c1ee10ffbbafd2/src/trace.rs#L5) with their `RUSTFLAGS='--cfg hyper_unstable_xyz'` thing.
 - If it's needed for Quinn, then that means that proto is favoring quinn as a privileged user, which is bad ideologically.

---

I found these usages of `#[doc(hidden)]`:

- [x] `quinn::send_stream::SendStream::poll_stopped`
  - As far as I can tell, the only usage of this anywhere is literally in the same module
  - Suggestion: Just make it private
  - #2100 resolves
- [x] `quinn::connection::Connection::force_key_update`
  - Used in interop container
  - Suggestion: Either make part of public API, or gate behind some sort of "interop-container" feature
  - #2101 resolves
- [x] `quinn_proto::connection::Connection::initiate_key_update`
  - Used in tests
  - Used in Quinn, in the `force_key_update` function, which is the previous item on this list
  - Suggestion: If make `force_key_update` part of public API, make this part of public API. If gate `force_key_update` behind "interop-container" feature, make this `#[cfg(any(test, feature = "interop-container"))]`
  - #2101 resolves
- [x] `quinn_proto::config::EndpointConfig::get_max_udp_payload_size`
  - Used in tests
  - Used in Quinn
  - Suggestion: Make part of public API
    - Possibly add getters for all config values?
  - #2102 resolves
- [x] `quinn_proto::connection::streams::ClosedStream::new`
  - Used in Quinn
  - Suggestion: Make part of public API, or rework Quinn usage to not need it
  - #2103 resolves
- [x] `quinn_proto::fuzzing`
  - Already marked `#[cfg(fuzzing)]`
  - Presumably used in fuzzing
[...]

PR #2100, opened 2024-12-20
quinn: Make SendStream::poll_stopped private

gretchenfrage:
Previously, this method was public, but marked as `#[doc(hidden)]`. However, nothing anywhere was using it. This commit makes it private.

---

See #2097


djc (approved):

Ralith (approved):
This was probably left over from quinn-h3. I've verified that the successor, h3-quinn, doesn't use it.

PR #2101, opened 2024-12-20
Un-hide force_key_update / initiate_key_update

gretchenfrage:
See #2097.

~~Note: `initiate_key_update` (the one in proto) is called from various places internally, which is the primary reason for differences in their names and comment.~~

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1233,7 +1233,7 @@ impl Connection {
     /// Update traffic keys spontaneously
     ///
     /// This can be useful for testing key updates, as they otherwise only happen infrequently.
-    pub fn initiate_key_update(&mut self) {
+    pub fn force_key_update(&mut self) {
         self.update_keys(None, false);
     }
 
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1230,7 +1230,9 @@ impl Connection {
         self.spaces[self.highest_space].ping_pending = true;
     }
 
-    #[doc(hidden)]
+    /// Update traffic keys spontaneously
+    ///
+    /// This can be useful for testing key updates, as they otherwise only happen infrequently.
     pub fn initiate_key_update(&mut self) {
         self.update_keys(None, false);
     }
--- a/quinn-proto/src/connection/packet_builder.rs
+++ b/quinn-proto/src/connection/packet_builder.rs
@@ -47,7 +47,7 @@ impl PacketBuilder {
         let sent_with_keys = conn.spaces[space_id].sent_with_keys;
         if space_id == SpaceId::Data {
             if sent_with_keys >= conn.key_phase_size {
-                conn.initiate_key_update();
+                conn.force_key_update();
             }
         } else {
             let confidentiality_limit = conn.spaces[space_id]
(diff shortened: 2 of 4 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -1230,7 +1230,9 @@ impl Connection {
         self.spaces[self.highest_space].ping_pending = true;
     }
 
-    #[doc(hidden)]
+    /// Update traffic keys
+    ///
+    /// The reason this is public is primarily for testing purposes.
     pub fn initiate_key_update(&mut self) {
```
Ralith, line 1236: Should we unify the naming for consistency? I lean towards `force_`.

Ralith (approved):

quinn-proto/src/connection/mod.rs
djc: Yes, agreed that would be nicer.

quinn-proto/src/connection/mod.rs
```diff
@@ -1230,7 +1230,9 @@ impl Connection {
         self.spaces[self.highest_space].ping_pending = true;
     }
 
-    #[doc(hidden)]
+    /// Update traffic keys
+    ///
+    /// The reason this is public is primarily for testing purposes.
```
djc, line 1235: I feel like we should ditch this note? It seems like this could be useful in other kinds of scenarios too, and I don't think this note is helpful to our users.

Ralith: What other situations?

I think API docs should generally provide context describing, or some sort of breadcrumb for, when and how they should be used. This is pretty mysterious without that, and in crypto that can be dangerous.

djc: That's fair, but the "The reason this is public" wording is a little roundabout for that kind of thing? Maybe something like "This can be used to help test key updates, which normally happen after a long time."?

quinn-proto/src/connection/mod.rs
gretchenfrage: Done.

quinn-proto/src/connection/mod.rs
gretchenfrage: Changed to

```
/// This can be useful for testing key updates, as they otherwise only happen infrequently.
```

gretchenfrage: Changed to:

```rs
    /// Update traffic keys spontaneously
    ///
    /// This can be useful for testing key updates, as they otherwise only happen infrequently.
```

djc (approved):

djc:
Hmm, CI seems stuck on this one? Maybe rebase?

gretchenfrage:
> Hmm, CI seems stuck on this one? Maybe rebase?

Nothing to rebase over ATM.

```
[phoenix@leliel:~/repos/h3stuff/quinn]$ git fetch upstream main
git From github.com:quinn-rs/quinn
 * branch              main       -> FETCH_HEAD

[phoenix@leliel:~/repos/h3stuff/quinn]$ git rebase upstream/main
Current branch force-key-update-public is up to date.

[phoenix@leliel:~/repos/h3stuff/quinn]$ git push --force origin force-key-update-public
Everything up-to-date
```

PR #2104, opened 2024-12-20
proto: Remove superfluous `#[doc(hidden)]` fuzzing

gretchenfrage:
The module is already `#[cfg(fuzzing)]`, which should make it not appear in docs.

---

See #2097 

I manually confirmed that the module doesn't appear when running `cargo doc` locally. However, I'm aware that docs.rs does some stuff about annotating which conditions different items are feature-gated behind, platform-gated behind, etc, which running `cargo doc` doesn't. I don't know how that works, whether it's a feature of cargo doc or something docs.rs has hacked on top, or how to recreate it locally. I'm hoping that it being `#[cfg(fuzzing)]`--notably, not a feature--will make it just not appear. But if we notice that it's popped up on the next release then we might have to revert.


Ralith (approved):
I don't think docs.rs is capable of setting random cfg flags.

djc (approved):

PR #2105, opened 2024-12-20
Unhide ecn variants

gretchenfrage:
See #2097

~~I took the liberty of committing b11f2bf5e71daeac93827541183be1e0a1fdab9c. Unfortunately, it is semver breaking for everything. Users must convert `x.is_ce()` to `x == EcnCodepoint::Ce`.~~


Ralith (changes requested):
I don't think we should get rid of the getters. They're a reasonably common idiom even with public variants because they're nice and terse, and the churn doesn't seem justified anyway.

Can you touch on the motivation for exposing the variants? I don't mind doing so, but I'd like to understand why.

gretchenfrage:
Removed that commit.

Ralith (approved):

djc (approved):

PR #2108, opened 2024-12-21
Remove `use self::` occurrences

gretchenfrage:
I noticed this was occurring here and there in the repo. I don't recall if there was a motivation for it in some old Rust version, but I don't think there is now.


djc (approved):

gretchenfrage:
New push should fix MSRV failure. Added comments to relevant places like such:

```rs
#[cfg(feature = "async-std")]
// Due to MSRV, we must specify `self::` where there's crate/module ambiguity
pub use self::async_std::AsyncStdRuntime;
```

Ralith (approved):
Yeah, this code just predates that Rust change.

PR #2109, opened 2024-12-21
proto: Pass `ConnectionId` by value internally

gretchenfrage:
Same idea as #2107 but for `ConnectionId`. `ConnectionId` is, and probably always will be, `Copy`. `ConnectionId` is not remotely large enough to trigger stack overflows or severe performance hazards by being copied. So, we should just simplify the code and let the compiler figure out if it wants to optimize it elsewise.

~~Plus, this shaves off a few lines by convincing the rustfmt to collapse some multi-line method signatures into one line.~~

Edit: Changes this PR to only do so internally, and not change the API.

### DIFF
--- a/quinn-proto/src/crypto/rustls.rs
+++ b/quinn-proto/src/crypto/rustls.rs
@@ -51,7 +51,7 @@ impl TlsSession {
 
 impl crypto::Session for TlsSession {
     fn initial_keys(&self, dst_cid: &ConnectionId, side: Side) -> Keys {
-        initial_keys(self.version, dst_cid, side, &self.suite)
+        initial_keys(self.version, *dst_cid, side, &self.suite)
     }
 
     fn handshake_data(&self) -> Option<Box<dyn Any>> {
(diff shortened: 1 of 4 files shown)


djc:
I don't think we want to change the public API for this right now.

gretchenfrage:
> I don't think we want to change the public API for this right now.

Understandable. I changed this PR so it only changes proto internally, without touching its API. Maybe we just prefer to pass by value in new APIs?

Ralith (approved):

Ralith:
Feel free to stash the public changes in a separate PR so we don't forget to bring them back in on the next breaking release.

djc (approved):

djc:
Maybe file an issue with a label so that we don't forget to clean up the public API instances of this when we get ready to release 0.12?

gretchenfrage:
> Maybe file an issue with a label so that we don't forget to clean up the public API instances of this when we get ready to release 0.12?

@djc to make sure I understand the reasoning: we haven't added semver-breaking changes since the last release, so we currently have the ability for our next release to be a minor version bump, and this change here isn't valuable enough to want to sacrifice that ability, so we want to wait to make such a change until some other PR appears which creates more motivation to do a breaking change. Is my understanding correct?

PR #2110, opened 2024-12-21
proto: Token encode refactors

gretchenfrage:

If "Make address a field of RetryToken" is too controversial, I can split it off into a subsequent PR to help the ones before it get through.


quinn-proto/src/token.rs
```diff
@@ -272,13 +272,15 @@ mod test {
         let addr = SocketAddr::new(Ipv6Addr::LOCALHOST.into(), 4433);
         let retry_src_cid = RandomConnectionIdGenerator::new(MAX_CID_SIZE).generate_cid();
         let token = RetryToken {
+            address: addr,
```
djc, line 254: Nit: suggest you rename the local binding to match the field name.

djc (approved):
Nice!

quinn-proto/src/token.rs
```diff
@@ -102,7 +102,7 @@ impl RetryToken {
         let mut sealed_token = raw_token_bytes.to_vec();
 
         let data = aead_key.open(&mut sealed_token, &[])?;
-        let mut reader = io::Cursor::new(data);
+        let mut reader = &mut &data[..];
```
Ralith, line 117: I think the `&mut` here is redundant. The variable itself is already mutable, and that's what we take a ref to when calling `decode_addr`.

Ralith (approved):
The weird `address` handling was leftover from of an old design where the address was AAD rather than explicitly stored. We migrated away from that because that made it impossible to distinguish invalid tokens from valid tokens with invalid addresses. Glad to see it tidied up.

PR #2111, opened 2024-12-22
proto: Refactor `Endpoint.handle`

gretchenfrage:
Just a draft for now.

---

Main motivation:

I think `Endpoint.handle` is too long. It's 173 lines and it contains section divider comments like this:

```rs
//
// Potentially create a new connection
//
```

The general architecture of this function dates back to a [very long time ago](https://github.com/gretchenfrage/quinn/commit/b919cc7100a1807413d5ee562231694df5a54831#diff-b7dd6c6dc06c9c5da00a0c048df395e5fed282392d70a222b1d0e34706d9baf3R47-R111) and it used to be a lot smaller. I find it difficult to read and think it should be split up more.

Also: take opportunities to reduce LOC where found. However, some overhead is to be expected due to additional function signatures.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -26,8 +26,8 @@ use crate::{
     },
     range_set::ArrayRangeSet,
     shared::{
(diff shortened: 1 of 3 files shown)


gretchenfrage:
Opting to split off into more PRs instead.

PR #2112, opened 2024-12-22
proto: Minor refactors to `Endpoint.handle`

gretchenfrage:
Please see individual commits.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -301,12 +301,7 @@ impl Endpoint {
         // connection. Send a stateless reset if possible.
         //
 
-        if !first_decode.is_initial()
-            && self
-                .local_cid_generator
-                .validate(first_decode.dst_cid())
-                .is_err()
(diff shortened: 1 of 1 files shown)


Ralith (approved):

quinn-proto/src/endpoint.rs
```diff
@@ -286,7 +286,9 @@ impl Endpoint {
                     None
                 }
             };
-        } else if first_decode.has_long_header() {
+        }
```
djc, line 289: While they are technically equivalent, I feel like the `else` here (in addition to being slightly more compact) is useful because it makes clear that there's no follow-through from the previous branch without relying on context (that the previous branch always diverges).

djc (approved):
Good stuff!

quinn-proto/src/endpoint.rs
gretchenfrage: Yeah I can see it. Removed commit.

PR #2113, opened 2024-12-22
Remove various obselete `allow` directives

gretchenfrage:


quinn-proto/src/congestion/cubic.rs
```diff
@@ -180,7 +180,6 @@ impl Controller for Cubic {
         self.recovery_start_time = Some(now);
 
         // Fast convergence
-        #[allow(clippy::branches_sharing_code)]
         // https://github.com/rust-lang/rust-clippy/issues/7198
```
Ralith, line 183: We should delete the comment that justifies the annotation along with the annotation.

Ralith (approved):

PR #2116, opened 2024-12-23
proto: Reduce size of `Endpoint.handle`

gretchenfrage:
IMO, `Endpoint.handle` is currently too large. This PR contains refactors relating to making it smaller by shunting more of its logic to `handle_first_packet`.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -237,15 +237,7 @@ impl Endpoint {
         let dst_cid = event.first_decode.dst_cid();
 
         if event.first_decode.initial_header().is_some() {
-            return self.handle_first_packet(
-                event.first_decode,
-                datagram_len,
-                addresses,
-                ecn,
-                event.remaining,
-                buf,
-                now,
-            );
+            return self.handle_first_packet(datagram_len, event, addresses, buf);
         } else if event.first_decode.has_long_header() {
             debug!(
                 "ignoring non-initial packet for unknown connection {}",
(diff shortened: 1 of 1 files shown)


quinn-proto/src/endpoint.rs
```diff
@@ -235,9 +232,7 @@ impl Endpoint {
             };
         }
 
-        //
         // Potentially create a new connection
-        //
```
djc, line 240: Extremely nitty: where the comment line describes a longer block (especially if the block it describes is itself split up into multiple blocks by empty lines), I prefer to keep an empty line between the comment and the block -- and for consistency, I might prefer keeping it for all of these.

djc (approved):
Good stuff!

quinn-proto/src/endpoint.rs
gretchenfrage: It gets a bit strange with the if-else but oh well

quinn-proto/src/endpoint.rs
```diff
@@ -238,54 +238,18 @@ impl Endpoint {
         //
         // Potentially create a new connection
         //
-
         let dst_cid = first_decode.dst_cid();
-        let Some(server_config) = &self.server_config else {
```
Ralith, line 243: If we're a client receiving a short-header packet, this change will cause us to reach the end of the function rather than exiting early. That should be fine, since the behavior is to send a stateless reset regardless, but the comment "If we got this far, we're a server ..." must be updated.

quinn-proto/src/endpoint.rs
```diff
@@ -200,11 +201,13 @@ impl Endpoint {
...
         };
 
-        // Handle packet on existing connection, if any
-
         let addresses = FourTuple { remote, local_ip };
+        let dst_cid = event.first_decode.dst_cid();
+
         if let Some(route_to) = self.index.get(&addresses, &event.first_decode) {
-            return match route_to {
+            // Handle packet on existing connection, if any
```
Ralith, line 207: In this branch, we know there's an existing connection [attempt], so "if any" no longer makes sense.

Ralith (approved):

PR #2117, opened 2024-12-23
Add gretchenfrage as a maintainer

djc:


gretchenfrage (approved):
Hooray!

.github/CODEOWNERS
```diff
@@ -1,2 +1,2 @@
-* @djc @Ralith
+* @djc @Ralith @gretchenfrage
 /quinn-udp @djc @Ralith @mxinden
```
Ralith, line 2: Do we need to modify line 2 as well, or is it redundant as-is?

Ralith (approved):
Happy to have you!

.github/CODEOWNERS
gretchenfrage: I think it might be redundant as-is?

gretchenfrage: GitHub's official example seems to indicate it's redundant as-is. https://github.com/github/docs/blob/main/content/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners.md#example-of-a-codeowners-file

Ralith: Cool, we can fix that in a followup then.

Ralith: Ah, then it's not redundant, and we should probably add @gretchenfrage there too?

djc: Yup!

PR #2119, opened 2024-12-26
proto: remove redundant cursors

Ralith:
Noticed this old TODO while reviewing #1912. Minor simplification/cleanup, no behavior change.

### DIFF
--- a/quinn-proto/src/frame.rs
+++ b/quinn-proto/src/frame.rs
@@ -1,6 +1,6 @@
 use std::{
     fmt::{self, Write},
-    io, mem,
+    mem,
     ops::{Range, RangeInclusive},
 };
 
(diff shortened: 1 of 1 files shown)


quinn-proto/src/frame.rs
```diff
@@ -732,7 +724,7 @@ impl Iterator for Iter {
             Ok(x) => Some(Ok(x)),
             Err(e) => {
                 // Corrupt frame, skip it and everything that follows
-                self.bytes = io::Cursor::new(Bytes::new());
+                self.bytes = Bytes::new();
```
gretchenfrage, line 727: Suggest changing to `self.bytes.clear()`.

gretchenfrage (approved):

PR #2120, opened 2024-12-26
proto: Replace calls to Duration::new

gretchenfrage:
Replaces them with more modern and appropriate constructors or constants.


Ralith (approved):
Thanks!

Issue #2121, opened 2024-12-26
`poll_transmit` is large


gretchenfrage:
`Connection.poll_transmit` is 582 lines. I think it should be split up somehow.

Maybe a good place to start would be to factor out the body of the loop over spaces? It accounts for 380 lines, and factoring it out would reduce rightward drift.

djc:
This drifted by on HN recently, and I've definitely noticed this:

https://www.16elt.com/2024/09/25/first-book-of-byte-sized-tech/#Idea-2-Smaller-components-are-not-necessarily-better-for-modularity

I've long had a guideline that single-call functions are little suspect, so not sure this is the right direction (but, I haven't looked at your draft PR yet).

gretchenfrage:
The "`DatagramBuilder`" idea Ralith noted in #2122 seems worth exploring

PR #2122, opened 2024-12-27
proto: Factor out poll_transmit_space

gretchenfrage:
First stab at #2121. This is the simplest possible change I could do to separate the loop body out into its own function. I began by copy/pasting it into a function, and then went through the compiler errors one-by-one adding parameters and replacing `break`/`continue`/`return` statements until it compiled. I hope this is the first step in the right direction of being able to figure out more sophisticated ways to gradually make it more tractable without breaking it.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -560,391 +560,30 @@ impl Connection {
         let mut congestion_blocked = false;
 
         // Iterate over all spaces and find data to send
-        let mut space_idx = 0;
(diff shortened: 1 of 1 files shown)


gretchenfrage:
@Ralith interested in your thoughts on whether this is the right direction, etc?

Ralith:
I like the idea of splitting this function up into well-encapsulated components in theory, but the complexity of the interface so far doesn't seem promising. If too much is shared across boundaries, then we're making things harder, not easier, to reason about.

Maybe we could introduce a `DatagramBuilder` that owns and better encapsulates much of this state, sort of like how we've got `PacketBuilder`?

PR #2123, opened 2024-12-27
fix(udp): support sending empty datagram

mxinden:
Work in progress.


mxinden:
Closing here. I don't think in the context of QUIC one ever has to send an empty UDP datagram. When testing, one can use `std::net::UdpSocket` instead.

Ralith:
It's not obvious to me that it would require any complexity for us to support, in which case I feel like we might as well.

PR #2130, opened 2025-01-09
ci: run quinn-udp tests with fast-apple-datapath

mxinden:
I suggest we test the feature introduced in https://github.com/quinn-rs/quinn/pull/1993 on CI.

---

Related, we are currently integrating it into Neqo (https://github.com/mozilla/neqo/pull/2302). On our localhost Download benchmark it shows ~10% throughput improvement.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -126,6 +126,8 @@ jobs:
       - run: cargo test --manifest-path fuzz/Cargo.toml
         if: ${{ matrix.rust }} == "stable"
       - run: cargo test -p quinn-udp --benches
+      - run: cargo test -p quinn-udp --benches --features fast-apple-datapath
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Ohh, beta ICEd in CI. I will file...

gretchenfrage:
Would it make sense to _only_ run it with this feature on macos, rather than both without and with?

Ralith:
IIRC we're deliberately supporting both cases, since the fast path is a bit sketchy.

gretchenfrage (approved):

gretchenfrage:
FWIW, the compiler bug also manifested from me rebasing #1912 over main. https://github.com/quinn-rs/quinn/actions/runs/12700393147/job/35402953810

djc:
@Ralith did you want to review this?

PR #2132, opened 2025-01-16
Associate a PathId to each Connection

divagant-martian:
adds a path id to the connection and sets the initial path id as 0


divagant-martian:
Wrong fork, I'm really sorry :sweat_smile:

PR #2136, opened 2025-01-26
Add `BloomTokenLog`

gretchenfrage:
- Adds default optional dependency on [`fastbloom`](https://lib.rs/crates/fastbloom)
- Adds `BloomTokenLog`, a new `TokenLog` implementation:
  - At first, it just stores elements in a hash set. In this phase, it experiences linear memory growth and neither false positives nor false negatives.
  - When the hash set would consume more memory than a configurable limit, it converts it to a bloom filter. This essentially makes it so that, rather than more elements causing it to consume more memory, more elements cause its false positive rate to keep going up. This way, its memory usage stays constant, and it treats old and newly added elements "fairly" rather than say, discarded new elements or something more questionable.
    - The `fastbloom` crate is a popular bloom filter implementation with SIMD acceleration.
  - The reason why it doesn't just use a bloom filter from the beginning is because then it would consume its maximum memory usage from the time it's first initialized. Overall, this two-phase mechanism means that its memory profile its linear with a ceiling, which avoids making the user pay for what they're not using, and has a false negative rate of zero in both phases, which is perhaps a safer default for users than one that doesn't.
  - The overall token log actually maintains two filters at any given point in time, each of which can independently contain elements as described above. It divides time into periods equal to the token lifetime, and stores a filter for each of the two phases in which currently non-expired tokens could expire in. This means the filters get reset over time, which prevents the hash set from consuming memory forever / the bloom filter from getting infinitely saturating and its false positive rate rising to 100%.
  - Note on false negatives: Although the bloom token log _itself_ doesn't experience false negatives, a server operationalizing it could experience false negatives if it is using the same token keys across multiple token log instances. This could occur over time (token keys persisted, but token log not persisted, and server restarts) and/or over space (multiple servers behind a load balancer sharing the same token key but each with their own token log). These cases probably wouldn't be security vulnerabilities--RFC 9000 does permit false negatives to occur so long as it's mitigated in some sense, and an attacker which gets to use the servers for amplification attacks only to the extent that there are multiple load-balanced servers / it can get the servers to reboot is probably still acceptably mitigated. But due to this complication I avoid mentioning the lack of false negatives in the documentation to avoid giving the user's a misleading impression.
- BloomTokenLog is made the default token log. Also, the default tokens sent is raised from 0 to 2.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ bytes = "1"
 clap = { version = "4", features = ["derive"] }
 crc = "3"
 directories-next = "2"
+fastbloom = "0.9"
 futures-io = "0.3.19"
 getrandom = { version = "0.3", default-features = false }
 hdrhistogram = { version = "7.2", default-features = false }
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -13,9 +13,11 @@ workspace = ".."
 [features]
 # NOTE: Please keep this in sync with the feature list in `.github/workflows/codecov.yml`, see
 # comment in that file for more information.
-default = ["rustls-ring", "log"]
+default = ["rustls-ring", "log", "bloom"]
 aws-lc-rs = ["dep:aws-lc-rs", "aws-lc-rs?/aws-lc-sys", "aws-lc-rs?/prebuilt-nasm"]
 aws-lc-rs-fips = ["aws-lc-rs", "aws-lc-rs?/fips"]
+# Enables BloomTokenLog, and uses it by default
+bloom = ["dep:fastbloom"]
 # For backwards compatibility, `rustls` forwards to `rustls-ring`
 rustls = ["rustls-ring"]
 # Enable rustls with the `aws-lc-rs` crypto provider
@@ -36,6 +38,7 @@ rustls-log = ["rustls?/logging"]
 arbitrary = { workspace = true, optional = true }
 aws-lc-rs = { workspace = true, optional = true }
 bytes = { workspace = true }
+fastbloom = { workspace = true, optional = true }
 lru-slab = { workspace = true }
 rustc-hash = { workspace = true }
 rand = { workspace = true }
@@ -57,8 +60,8 @@ web-time = { workspace = true }
 
 [dev-dependencies]
 assert_matches = { workspace = true }
-rand_pcg = "0.9"
 hex-literal = { workspace = true  }
+rand_pcg = "0.9"
 rcgen = { workspace = true }
 tracing-subscriber = { workspace = true }
 lazy_static = "1"
--- a/quinn-proto/src/bloom_token_log.rs
+++ b/quinn-proto/src/bloom_token_log.rs
@@ -1,0 +1,368 @@
+use std::{
+    collections::HashSet,
+    f64::consts::LN_2,
+    hash::{BuildHasher, Hasher},
+    mem::{size_of, take},
+    sync::Mutex,
+};
+
+use fastbloom::BloomFilter;
+use rustc_hash::FxBuildHasher;
+use tracing::{trace, warn};
+
+use crate::{Duration, SystemTime, TokenLog, TokenReuseError, UNIX_EPOCH};
+
+/// Bloom filter-based [`TokenLog`]
+///
+/// Parameterizable over an approximate maximum number of bytes to allocate. Starts out by storing
+/// used tokens in a hash set. Once the hash set becomes too large, converts it to a bloom filter.
+/// This achieves a memory profile of linear growth with an upper bound.
+///
+/// Divides time into periods based on `lifetime` and stores two filters at any given moment, for
+/// each of the two periods currently non-expired tokens could expire in. As such, turns over
+/// filters as time goes on to avoid bloom filter false positive rate increasing infinitely over
+/// time.
+pub struct BloomTokenLog(Mutex<State>);
+
+impl BloomTokenLog {
+    /// Construct with an approximate maximum memory usage and expected number of validation token
+    /// usages per expiration period
+    ///
+    /// Calculates the optimal bloom filter k number automatically.
+    pub fn new_expected_items(max_bytes: usize, expected_hits: u64) -> Self {
+        Self::new(max_bytes, optimal_k_num(max_bytes, expected_hits))
+    }
+
+    /// Construct with an approximate maximum memory usage and a [bloom filter k number][bloom]
+    ///
+    /// [bloom]: https://en.wikipedia.org/wiki/Bloom_filter
+    ///
+    /// If choosing a custom k number, note that `BloomTokenLog` always maintains two filters
+    /// between them and divides the allocation budget of `max_bytes` evenly between them. As such,
+    /// each bloom filter will contain `max_bytes * 4` bits.
+    pub fn new(max_bytes: usize, k_num: u32) -> Self {
+        Self(Mutex::new(State {
+            config: FilterConfig {
+                filter_max_bytes: max_bytes / 2,
+                k_num,
+            },
+            period_1_start: UNIX_EPOCH,
+            filter_1: Filter::default(),
+            filter_2: Filter::default(),
+        }))
+    }
+}
+
+impl TokenLog for BloomTokenLog {
+    fn check_and_insert(
+        &self,
+        nonce: u128,
+        issued: SystemTime,
+        lifetime: Duration,
+    ) -> Result<(), TokenReuseError> {
+        trace!(%nonce, "check_and_insert");
+
+        if lifetime.is_zero() {
+            // avoid divide-by-zero if lifetime is zero
+            return Err(TokenReuseError);
+        }
+
+        let mut guard = self.0.lock().unwrap();
+        let state = &mut *guard;
+
+        // calculate how many periods past period 1 the token expires
+        let expires_at = issued + lifetime;
+        let Ok(periods_forward) = expires_at
+            .duration_since(state.period_1_start)
+            .map(|duration| duration.as_nanos() / lifetime.as_nanos())
+        else {
+            // shouldn't happen unless time travels backwards or lifetime changes or the current
+            // system time is before the Unix epoch
+            warn!("BloomTokenLog presented with token too far in past");
+            return Err(TokenReuseError);
+        };
+
+        // get relevant filter
+        let filter = match periods_forward {
+            0 => &mut state.filter_1,
+            1 => &mut state.filter_2,
+            2 => {
+                // turn over filter 1
+                state.filter_1 = take(&mut state.filter_2);
+                state.period_1_start += lifetime;
+                &mut state.filter_2
+            }
+            _ => {
+                // turn over both filters
+                state.filter_1 = Filter::default();
+                state.filter_2 = Filter::default();
+                state.period_1_start = expires_at;
+                &mut state.filter_1
+            }
+        };
+
+        // insert into the filter
+        //
+        // the token's nonce needs to guarantee uniqueness because of the role it plays in the
+        // encryption of the tokens, so it is 128 bits. but since the token log can tolerate false
+        // positives, we trim it down to 64 bits, which would still only have a small collision
+        // rate even at significant amounts of usage, while allowing us to store twice as many in
+        // the hash set variant.
+        //
+        // token nonce values are uniformly randomly generated server-side and cryptographically
+        // integrity-checked, so we don't need to employ secure hashing to trim it down to 64 bits,
+        // we can simply truncate.
+        //
+        // per the Rust reference, we can truncate by simply casting:
+        // https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html#numeric-cast
+        filter.check_and_insert(nonce as u64, &state.config)
+    }
+}
+
+/// Default to 20 MiB max memory consumption and expected one million hits
+///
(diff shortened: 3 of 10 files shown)


quinn-proto/src/tests/util.rs
```diff
@@ -561,10 +561,12 @@ impl Write for TestWriter {
 
 pub(super) fn server_config() -> ServerConfig {
     let mut config = ServerConfig::with_crypto(Arc::new(server_crypto()));
-    config
-        .validation_token
-        .sent(2)
-        .log(Arc::new(SimpleTokenLog::default()));
+    if cfg!(feature = "fastbloom") {
```
gretchenfrage, line 564: My bad, I need to invert this.

gretchenfrage:
@djc and @ralith, some meta-discussion I'd like to explicitly invite, and this applies to #2137 as well: It may be worth evaluating whether we want to go this route in the first place (of adding these defaults). It's possible we decide we want to go a different route.

Some alternatives could include:

- We could simply leave Quinn's default `TokenLog` / `TokenStore` to be implementations that can't utilize tokens. I could create my own crate with `BloomTokenLog` / `TokenMemoryCache` in it.
  - We could experiment with me creating them in a crate, and consider merging them in to Quinn later.
- We could try to come up with an alternative implementation of these traits which is able to use tokens in a general-purpose context but is simpler than these ones.
- We could keep these implementations but try to shrink them a little bit through removing some of my "clever"ness, e.g. removing the "identity hasher."

Points in favor of merging these PRs:

- It has the potential to make some Quinn traffic faster "for free"
- I believe that these designs should generally avoid degrading performance a notable amount, and that the worst-case performance scenario is basically that they fail to benefit the user
- Users can easily disable these

Points against merging these PRs:

- It does give the server another worst case 20 MiB memory buffer that could consume RAM
- Increased complexity on Quinn for us to maintain
- Additional dependency on a third party crate we aren't involved in the creation of maintenance of

I think it's worth putting these in here, but I'd like to hear what you think.

Ralith:
re: meta-discussion: I think I'm on board with including this functionality out of the box. It seems to strike a good balance: typical users won't be affected by a 20MiB worst-case memory use increase, and will benefit from better handshake performance. Conversely, anyone who can't afford that probably already needs to customize e.g. flow control limits, so a little more tuning is not much to ask of them.

As far as the dep goes, I think a third-party dep is less maintenance load for us than becoming experts in bloom token implementation. That said, I did feel obligated to have a brief skim, and unfortunately [immediately found UB](https://github.com/tomtomwombat/fastbloom/pull/12). Let's consider that a blocker for now.

quinn-proto/src/bloom_token_log.rs
```diff
@@ -0,0 +1,337 @@
...
+    }
+}
+
+fn optimal_k_num(num_bytes: usize, expected_hits: u64) -> u32 {
+    // be more forgiving rather than panickey here. excessively high num_bits may occur if the user
+    // wishes it to be unbounded, so just saturate. expected_hits of 0 would cause divide-by-zero,
+    // so just fudge it up to 1 in that case.
+    let num_bits = (num_bytes as u64).saturating_mul(8);
+    let expected_hits = expected_hits.max(1);
+    (((num_bits as f64 / expected_hits as f64) * LN_2).round() as u32).max(1)
```
gretchenfrage, line 225: Saturate on overflow in ` as u32`

quinn-proto/src/bloom_token_log.rs
```diff
@@ -0,0 +1,337 @@
...
+                    return Ok(());
+                }
+
+                // convert to bloom
+                // avoid panicking if user passed in filter_max_bytes of 0. we document that this
+                // limit is approximate, so just fudge it up to 1.
+                let mut bloom = BloomFilter::with_num_bits((config.filter_max_bytes * 8).max(1))
+                    .hasher(FxBuildHasher)
+                    .hashes(config.k_num);
+                for item in hset.iter() {
```
gretchenfrage, line 159: Avoid `.iter()`

gretchenfrage (commented):
Self-review with fresh eyes.

quinn-proto/src/bloom_token_log.rs
Ralith, line 225: Explain or cite a reference for this math.

quinn-proto/src/bloom_token_log.rs
```diff
@@ -0,0 +1,337 @@
...
+
+/// The token's nonce needs to guarantee uniqueness because of the role it plays in the encryption
+/// of the tokens, so it is 128 bits. But since the token log can tolerate false positives, we trim
+/// it down to 64 bits, which would still only have a small collision rate even at significant
+/// amounts of usage, while allowing us to store twice as many in the hash set variant.
+///
+/// Token nonce values are uniformly randomly generated server-side and cryptographically
+/// integrity-checked, so we don't need to employ secure hashing for this, we can simply truncate.
+fn nonce_to_fingerprint(nonce: u128) -> u64 {
+    (nonce & u64::MAX as u128) as u64
```
Ralith, line 236: I don't think the mask does anything here; `as` truncates regardless.

quinn-proto/src/bloom_token_log.rs
```diff
@@ -0,0 +1,337 @@
...
+    fn optimal_k_num_test() {
+        assert_eq!(optimal_k_num(10 << 20, 1_000_000), 58);
+        assert_eq!(optimal_k_num(10 << 20, 1_000_000_000_000_000), 1);
+        // assert that these don't panic:
+        optimal_k_num(10 << 20, 0);
+        optimal_k_num(usize::MAX, 1_000_000);
+    }
+
+    #[test]
+    fn bloom_token_log_conversion() {
```
Ralith, line 274: How do we know this test actually hits the conversion path?

Ralith (changes requested):

Cargo.toml
```diff
@@ -23,6 +23,7 @@ bytes = "1"
 clap = { version = "4", features = ["derive"] }
 crc = "3"
 directories-next = "2"
+fastbloom = "0.8"
```
Ralith, line 26: 0.9 has been released with the UB fix.

quinn-proto/src/bloom_token_log.rs
```diff
@@ -0,0 +1,337 @@
...
+                state.period_1_start = expires_at;
+                &mut state.filter_1
+            }
+        };
+
+        filter.check_and_insert(fingerprint, &state.config)
+    }
+}
+
+/// Default to 20 MiB max memory consumption and expected one million hits
```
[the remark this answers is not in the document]
gretchenfrage: Default validation token lifetime is 2 weeks, so it's one token usage per 1.21 seconds.

quinn-proto/src/bloom_token_log.rs
gretchenfrage: It looks like `f64 as u32` does that automatically. https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=dd52657cbeb93bf56b8a3df6d0b159d0

quinn-proto/Cargo.toml
```diff
@@ -13,7 +13,7 @@ workspace = ".."
 [features]
 # NOTE: Please keep this in sync with the feature list in `.github/workflows/codecov.yml`, see
 # comment in that file for more information.
-default = ["rustls-ring", "log"]
+default = ["rustls-ring", "log", "fastbloom"]
```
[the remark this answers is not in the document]
gretchenfrage: Introduced feature named `bloom`.

quinn-proto/src/bloom_token_log.rs
```diff
@@ -0,0 +1,337 @@
...
+impl BloomTokenLog {
+    /// Construct with an approximate maximum memory usage and expected number of validation token
+    /// usages per expiration period
+    ///
+    /// Calculates the optimal bloom filter k number automatically.
+    pub fn new_expected_items(max_bytes: usize, expected_hits: u64) -> Self {
+        Self::new(max_bytes, optimal_k_num(max_bytes, expected_hits))
+    }
+
+    /// Construct with an approximate maximum memory usage and a bloom filter k number
```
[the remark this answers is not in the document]
gretchenfrage: Linked to Wikipedia.

quinn-proto/src/bloom_token_log.rs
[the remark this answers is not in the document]
gretchenfrage: Made a note of this in the doc comment.

quinn-proto/src/bloom_token_log.rs
gretchenfrage: Done

quinn-proto/src/bloom_token_log.rs
gretchenfrage: Nice catch. Also, I inlined this function.

quinn-proto/src/bloom_token_log.rs
gretchenfrage: Added new assertion at end to test for that

Cargo.toml
gretchenfrage: Switched to it!

quinn-proto/Cargo.toml
```diff
@@ -13,9 +13,11 @@ workspace = ".."
 [features]
 # NOTE: Please keep this in sync with the feature list in `.github/workflows/codecov.yml`, see
 # comment in that file for more information.
-default = ["rustls-ring", "log"]
+default = ["rustls-ring", "log", "bloom"]
 aws-lc-rs = ["dep:aws-lc-rs", "aws-lc-rs?/aws-lc-sys", "aws-lc-rs?/prebuilt-nasm"]
 aws-lc-rs-fips = ["aws-lc-rs", "aws-lc-rs?/fips"]
+# Enable things that use bloom filter dependency
```
Ralith, line 19: This documentation isn't very informative. How about something like "Enables `BloomTokenLog`, and uses it by default"

quinn-proto/src/config/mod.rs
```diff
@@ -498,7 +507,8 @@ impl ValidationTokenConfig {
     ///
     /// This refers only to tokens sent in NEW_TOKEN frames, in contrast to retry tokens.
     ///
-    /// Defaults to 0.
+    /// If the `fastbloom` feature is enabled (which it is by default), defaults to 2. Otherwise,
```
Ralith, line 510:
```suggestion
    /// If the `bloom` feature is enabled (which it is by default), defaults to 2. Otherwise,
```

Ralith (approved):

gretchenfrage:
@djc do you want to review this as well? It's fine either way.

djc:
Yes, would like to take a look.

quinn-proto/src/bloom_token_log.rs
```diff
@@ -0,0 +1,368 @@
...
+    /// each bloom filter will contain `max_bytes * 4` bits.
+    pub fn new(max_bytes: usize, k_num: u32) -> Self {
+        Self(Mutex::new(State {
+            config: FilterConfig {
+                filter_max_bytes: max_bytes / 2,
+                k_num,
+            },
+            period_1_start: UNIX_EPOCH,
+            filter_1: Filter::default(),
+            filter_2: Filter::default(),
```
djc, line 51: Nit: suggest moving this into a private `State::new()` constructor.

quinn-proto/src/bloom_token_log.rs
```diff
@@ -0,0 +1,368 @@
...
+                    bloom.insert(item);
+                }
+                *self = Self::Bloom(bloom);
+            }
+            Self::Bloom(bloom) => {
+                if bloom.insert(&fingerprint) {
+                    return Err(TokenReuseError);
+                }
+            }
+        }
```
djc, line 186: Nit: reduce rightward drift here by handling the `Set` case outside of the `match`?

```rust
let hset = match self {
    Self::Set(hset) => hset,
    Self::Bloom(bloom) => return match bloom.insert(&fingerprint) {
        true => Err(TokenReuseError),
        false => Ok(()),
    },
};
```

quinn-proto/src/bloom_token_log.rs
```diff
@@ -0,0 +1,368 @@
...
+const DEFAULT_EXPECTED_HITS: u64 = 1_000_000;
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use rand::prelude::*;
+    use rand_pcg::Pcg32;
+
+    fn new_rng() -> impl Rng {
+        Pcg32::from_seed(0xdeadbeefdeadbeefdeadbeefdeadbeefu128.to_le_bytes())
```
djc, line 262: Nit: use `_u128` to more cleanly separate the value from the type.

djc (approved):
Looks great, thanks! (And sorry for the slow cycle times.)

quinn-proto/src/bloom_token_log.rs
gretchenfrage: `State` isn't really extricable from this struct, it's basically just a hack to put all the fields in the same mutex. If we were to do that it would feel best to also create a `State::check_and_insert(&mut self, ...` for consistency. I'm not sure it's necessary at this point.

quinn-proto/src/bloom_token_log.rs
gretchenfrage: I'm inclined not to. I like to short-circuit when either the short-circuit seems like more of an exceptional case than the other part, or if the rightward drift is becoming unmanageable. Neither is really the case here.

PR #2137, opened 2025-01-26
Add `TokenMemoryCache`

gretchenfrage:
Implements `TokenMemoryCache`, and makes it the default `TokenStore` for clients.

### DIFF
--- a/quinn-proto/Cargo.toml
+++ b/quinn-proto/Cargo.toml
@@ -36,6 +36,7 @@ rustls-log = ["rustls?/logging"]
 arbitrary = { workspace = true, optional = true }
 aws-lc-rs = { workspace = true, optional = true }
 bytes = { workspace = true }
+lru-slab = { workspace = true }
 rustc-hash = { workspace = true }
 rand = { workspace = true }
 ring = { workspace = true, optional = true }
@@ -56,6 +57,7 @@ web-time = { workspace = true }
 
 [dev-dependencies]
 assert_matches = { workspace = true }
+rand_pcg = "0.9"
 hex-literal = { workspace = true  }
 rcgen = { workspace = true }
 tracing-subscriber = { workspace = true }
--- a/quinn-proto/src/config/mod.rs
+++ b/quinn-proto/src/config/mod.rs
@@ -14,8 +14,8 @@ use thiserror::Error;
 #[cfg(any(feature = "rustls-aws-lc-rs", feature = "rustls-ring"))]
 use crate::crypto::rustls::{QuicServerConfig, configured_provider};
 use crate::{
-    DEFAULT_SUPPORTED_VERSIONS, Duration, MAX_CID_SIZE, NoneTokenLog, NoneTokenStore,
-    RandomConnectionIdGenerator, SystemTime, TokenLog, TokenStore, VarInt, VarIntBoundsExceeded,
+    DEFAULT_SUPPORTED_VERSIONS, Duration, MAX_CID_SIZE, NoneTokenLog, RandomConnectionIdGenerator,
+    SystemTime, TokenLog, TokenMemoryCache, TokenStore, VarInt, VarIntBoundsExceeded,
     cid_generator::{ConnectionIdGenerator, HashedConnectionIdGenerator},
     crypto::{self, HandshakeTokenKey, HmacKey},
     shared::ConnectionId,
@@ -553,7 +553,7 @@ impl ClientConfig {
         Self {
             transport: Default::default(),
             crypto,
-            token_store: Arc::new(NoneTokenStore),
+            token_store: Arc::new(TokenMemoryCache::default()),
             initial_dst_cid_provider: Arc::new(|| {
                 RandomConnectionIdGenerator::new(MAX_CID_SIZE).generate_cid()
             }),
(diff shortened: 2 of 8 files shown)


quinn-proto/Cargo.toml
```diff
@@ -56,6 +56,7 @@ web-time = { workspace = true }
 
 [dev-dependencies]
 assert_matches = { workspace = true }
+rand_pcg = "0.3"
```
Ralith, line 59: Please update to latest.

quinn-proto/src/token_memory_cache.rs
```diff
@@ -0,0 +1,539 @@
...
+    sync::{Arc, Mutex},
+};
+use tracing::trace;
+
+use crate::token::TokenStore;
+
+/// `TokenStore` implementation that stores up to `N` tokens per server name for up to a
+/// limited number of server names, in-memory
+#[derive(Debug)]
+pub struct TokenMemoryCache<const N: usize>(Mutex<State<N>>);
```
Ralith, line 17: Does `N` really need to be a compile-time constant? Dynamic values are simpler and mean less work for the compiler downstream.

quinn-proto/src/token_memory_cache.rs
```diff
@@ -0,0 +1,539 @@
...
+            older: None,
+            newer: None,
+            tokens,
+        }
+    }
+}
+
+/// In-place vector queue of up to `N` `Bytes`
+#[derive(Debug)]
+struct Queue<const N: usize> {
```
Ralith, line 215: Would it be simpler for this to be a `VecDeque`?

quinn-proto/src/config/mod.rs
```diff
@@ -585,8 +585,8 @@ impl ClientConfig {
 
     /// Set a custom [`TokenStore`]
     ///
-    /// Defaults to [`NoneTokenStore`], which disables the use of tokens from NEW_TOKEN frames as a
-    /// client.
+    /// Default to [`TokenMemoryCache`] with 2 tokens per server, which is suitable for most
```
gretchenfrage, line 588: `Default` -> `Defaults`.

quinn-proto/src/token_memory_cache.rs
```diff
@@ -0,0 +1,539 @@
...
+            Some(token)
+        } else {
+            None
+        }
+    }
+}
+
+/// Slab-based linked LRU cache of `CacheEntry`
+#[derive(Debug, Default)]
+struct LinkedCache<const N: usize> {
```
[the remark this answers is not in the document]
gretchenfrage: I like it :D I'm worried about this though https://github.com/Ralith/lru-slab/issues/1

quinn-proto/src/token_memory_cache.rs
gretchenfrage: Yeahhhhhhhhhhh I guess not. My argument of avoiding a heap allocation seems kinda flimsy since there's already going to be 1 mandatory heap allocation for the server name and then another for each token.

gretchenfrage:
Switched from bespoke LRU cache implementation to Ralith's `lru_slab` crate. Saves a lot of code for sure, although I don't wanna merge until we decide what we wanna do about https://github.com/Ralith/lru-slab/issues/1

Switched from bespoke in-place queue implementation to just using `VecDeque`.

Together, these changes cut the size of this PR roughly in half. Nice!

quinn-proto/src/token_memory_cache.rs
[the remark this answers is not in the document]
Ralith: Fixed in 0.1.2. Not sure anyone should be storing billions of tokens in memory regardless.

quinn-proto/src/token_memory_cache.rs
Ralith: Heap allocations aren't something to be avoided for their own sake, in any case.

quinn-proto/src/token_memory_cache.rs
```diff
@@ -0,0 +1,247 @@
...
+                if let Some(removed_slot) = removed_key {
+                    let removed = self.lookup.remove(&removed_slot);
+                    debug_assert!(removed.is_some());
+                }
+            }
+        };
+    }
+
+    fn take(&mut self, server_name: &str) -> Option<Bytes> {
+        if let hash_map::Entry::Occupied(hmap_entry) = self.lookup.entry(server_name.into()) {
```
Ralith, line 115: style nit: use let-else to reduce rightward drift

Ralith, line 115: Here you're allocating a fresh `Arc<str>` and using a less common API to avoid performing multiple lookups into the hash table in the rare case of token exhaustion. Is that worthwhile?

Ralith (approved):

quinn-proto/src/token_memory_cache.rs
gretchenfrage: Probably not--unless the user set tokens_per_server to 1. I'll switch it.

quinn-proto/src/token_memory_cache.rs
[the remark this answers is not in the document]
gretchenfrage: Increased dependency version in this PR.

gretchenfrage:
@djc do you want to review this as well? It's fine either way.

djc:
Yes, will try to take a look soon.

quinn-proto/src/config/mod.rs
```diff
@@ -585,8 +585,8 @@ impl ClientConfig {
 
     /// Set a custom [`TokenStore`]
     ///
-    /// Defaults to [`NoneTokenStore`], which disables the use of tokens from NEW_TOKEN frames as a
-    /// client.
+    /// Defaults to a default [`TokenMemoryCache`], which is suitable for most internet
```
djc, line 588: Nit: "Defaults to a default" seems a little ugly? I'd probably omit "a default", since I think that is the reasonable assumption anyway.

djc (approved):
Nice!

PR #2141, opened 2025-01-27
fix(quinn): `impl tokio::io::AsyncWrite for SendStream`

matheus23:
We don't actually need the `runtime-tokio` feature to implement `tokio::io::AsnycWrite`, as we import enough of tokio for that already.

In a similar way, there already exists an `impl tokio::io::AsyncRead for RecvStream`, so this would just make the implementations between `SendStream` and `RecvStream` consistent.

### DIFF
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -260,7 +260,6 @@ impl futures_io::AsyncWrite for SendStream {
     }
 }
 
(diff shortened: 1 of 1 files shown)


djc (approved):
Makes sense, thanks!

djc:
(CI failure should be fixed in #2139.)

PR #2144, opened 2025-01-28
Upgrade to rand 0.9

djc:
Also, getrandom 0.3.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,3 @@
 [target.wasm32-unknown-unknown]
 runner = "wasm-bindgen-test-runner"
+rustflags = ["--cfg", 'getrandom_backend="wasm_js"']
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,14 +24,14 @@ clap = { version = "4", features = ["derive"] }
 crc = "3"
 directories-next = "2"
 futures-io = "0.3.19"
-getrandom = { version = "0.2", default-features = false }
+getrandom = { version = "0.3", default-features = false }
 hdrhistogram = { version = "7.2", default-features = false }
 hex-literal = "0.4"
 lazy_static = "1"
 log = "0.4"
 once_cell = "1.19"
 pin-project-lite = "0.2"
-rand = "0.8"
+rand = "0.9"
 rcgen = "0.13"
 ring = "0.17"
 rustc-hash = "2"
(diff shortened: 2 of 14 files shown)


gretchenfrage (changes requested):
WASM CI error. Per message, it [looks](https://docs.rs/getrandom/latest/getrandom/#webassembly-support) like we have to pass a new `RUSTFLAGS` env var in WASM CI?

gretchenfrage:
> WASM CI error. Per message, it [looks](https://docs.rs/getrandom/latest/getrandom/#webassembly-support) like we have to pass a new `RUSTFLAGS` env var in WASM CI?

Note @matheus23 's [suggested patch](https://discord.com/channels/976380008299917365/1063547094863978677/1333765135772160021)

Ralith (approved):

gretchenfrage (changes requested):
Diff itself looks good! But interested in the 2 remaining CI failures. I'm manually re-running the Solaris test because it seems like it was just flaky. But there's also [this](https://github.com/quinn-rs/quinn/actions/runs/13029035184/job/36396258325?pr=2144) audit CI failure relating to libfuzzer licensing:

```
error[rejected]: failed to satisfy license requirements
  ┌─ registry+https://github.com/rust-lang/crates.io-index#libfuzzer-sys@0.4.9:4:36
  │
4 │ license = "(MIT OR Apache-2.0) AND NCSA"
  │            ────────────────────────━━━━
  │            │                       │
  │            │                       rejected: license is not explicitly allowed
  │            license expression retrieved via Cargo.toml `license`
  │
  ├ NCSA - University of Illinois/NCSA Open Source License:
  ├   - OSI approved
  ├   - FSF Free/Libre
  ├ libfuzzer-sys v0.4.9
    └── fuzz v0.1.0
```

It's not entirely clear to me why that failure would be new, though.

Ralith:
Also probably not very important since it's just a dev-dep (right?)

djc:
Addressed the audit failure in #2148.

gretchenfrage (approved):

torokati44:
While `quinn-udp` already has a release published with this PR in it, `quinn-proto` and `quinn` itself does not.
Is there an expected timeline for releases of the latter two?

Ralith:
Nothing formal, but I think we're about due; it'd be nice to unbreak docs.rs besides. I think someone just needs to make a patch version bump PR.

PR #2146, opened 2025-01-29
quinn: `endpoint_wait_idle` failing test

ramfox:
I found what I think is a bug in quinn, and this PR adds a test illustrating the bug.

The `endpoint_wait_idle` test shows that if an endpoint attempts to connect to an address, but the connection never gets beyond the `Connecting` stage, it causes the `wait_idle` call (when trying to gracefully shutdown the endpoint) to always take around 3s to return.

I ran into this personally when a call to `Endpoint::connect` was made in a task that was aborted early. Suddenly, calls to `wait_idle` were taking around 3s, when they were previously <200ms. 

I'm hoping to get insight on to what could be causing this, so that either I can take a stab at fixing, or someone with more quinn experience can.

### DIFF
--- a/quinn/src/tests.rs
+++ b/quinn/src/tests.rs
@@ -112,6 +112,70 @@ async fn close_endpoint() {
     }
 }
 
+#[tokio::test]
+async fn endpoint_wait_idle() {
+    let _guard = subscribe();
+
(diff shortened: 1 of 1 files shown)


djc:
See discussion in

- #2147

PR #2149, opened 2025-01-31
add test for retry token expired

### LINKED ISSUE #2118: Add test for retry token expiration
Now that #2089 is merged, we should add a test for Retry token expiration. One could start by modifying this test: https://github.com/quinn-rs/quinn/blob/bfbeecdc1c23c4ba4e7697b67e4888a80b533fdb/quinn-proto/src/tests/mod.rs#L186

yohachiSuga:
- This PR closes #2118.
- This PR adds test code to verify the expiration of the retry token.

### DIFF
--- a/quinn-proto/src/tests/token.rs
+++ b/quinn-proto/src/tests/token.rs
@@ -24,6 +24,43 @@ fn stateless_retry() {
 }
 
 #[test]
+fn retry_token_expired() {
+    let _guard = subscribe();
+
+    let fake_time = Arc::new(FakeTimeSource::new());
+    let retry_token_lifetime = Duration::from_secs(1);
+
+    let mut pair = Pair::default();
+    pair.server.handle_incoming = Box::new(validate_incoming);
+
+    let mut config = server_config();
+    config
+        .time_source(Arc::clone(&fake_time) as _)
+        .retry_token_lifetime(retry_token_lifetime);
+    pair.server.set_server_config(Some(Arc::new(config)));
+
+    let client_ch = pair.begin_connect(client_config());
+    pair.drive_client();
+    pair.drive_server();
+    pair.drive_client();
+
+    // to expire retry token
+    fake_time.advance(retry_token_lifetime + Duration::from_millis(1));
+
+    pair.drive();
+    assert_matches!(
+        pair.client_conn_mut(client_ch).poll(),
+        Some(Event::ConnectionLost { reason: ConnectionError::ConnectionClosed(err) })
+        if err.error_code == TransportErrorCode::INVALID_TOKEN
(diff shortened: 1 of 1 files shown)


quinn-proto/src/tests/token.rs
```diff
@@ -23,6 +23,40 @@ fn stateless_retry() {
     assert_eq!(pair.server.known_cids(), 0);
 }
 
+#[test]
+fn stateless_retry_token_expired() {
```
gretchenfrage, line 27: Could you rename to avoid usage of the phrase "stateless retry", and just call it a retry token? [I've been pushing Quinn to avoid that](https://github.com/quinn-rs/quinn/pull/2086), because I think it's too easy to confuse with stateless resets, and it's not a phrase that's used in RFC 9000.

quinn-proto/src/tests/token.rs
```diff
@@ -23,6 +23,40 @@ fn stateless_retry() {
...
+    config.time_source(Arc::clone(&fake_time) as _);
+    pair.server.set_server_config(Some(Arc::new(config)));
+
+    let client_ch = pair.begin_connect(client_config());
+    pair.drive_client();
+    pair.drive_server();
+    pair.drive_client();
+
+    // to expire retry token
+    fake_time.advance(Duration::from_secs(100));
```
gretchenfrage, line 45: Would it be better to test that it counts as expired even if it's just a tiny bit late, like a second? The retry expiration period is configurable in the config.

gretchenfrage (changes requested):
Thanks for working on!

quinn-proto/src/tests/token.rs
yohachiSuga: Thank you for your information! I will rename this testcase.

quinn-proto/src/tests/token.rs
yohachiSuga: As you mentioned, I will update this to wait for the retry token lifetime to expire without using `fake_time`.

gretchenfrage: I still think using fake time is a good idea! I just think that 100 seconds is probably too high, and more importantly, arbitrary.

yohachiSuga: I see what you meant in your previous comment!
I've updated the test to verify expiration by advancing fake time to the retry token lifetime plus a small additional duration.
https://github.com/quinn-rs/quinn/pull/2149/files#diff-85a9a9679a7643217b4f2d5e0af2db99104ab928e995e102163ef62cc6b02fc5R48

gretchenfrage (approved):
Thanks!

PR #2151, opened 2025-02-10
Adding packet receipt timestamp collection into RecvMeta for linux

recatek:


quinn-udp/src/unix.rs
```diff
@@ -149,6 +157,11 @@ impl UdpSocketState {
                 )?;
             }
         }
+        #[cfg(target_os = "linux")]
+        {
+            // TODO: Should have a way to opt in to this.
```
Ralith, line 162: Is there a drawback to having this enabled? cc @mxinden

quinn-udp/src/lib.rs
```diff
@@ -115,6 +115,12 @@ pub struct RecvMeta {
     /// Populated on platforms: Windows, Linux, Android (API level > 25),
     /// FreeBSD, OpenBSD, NetBSD, macOS, and iOS.
     pub dst_ip: Option<IpAddr>,
+    /// A timestamp for when the given packet was received by the operating system.
+    /// Timestamping must be enabled for this information to be collected.
+    ///
+    /// Populated on platforms: Linux
+    #[cfg(not(wasm_browser))]
+    pub timestamp: Option<Duration>,
```
Ralith, line 123: Document what this is relative to. If it's unspecified, document that.

quinn-udp/src/unix.rs
```diff
@@ -149,6 +157,11 @@ impl UdpSocketState {
...
+            set_socket_option(&*io, libc::SOL_SOCKET, libc::SO_TIMESTAMP_NEW, OPTION_ON)?;
```
Ralith, line 163: An error here probably shouldn't be fatal.

quinn-udp/src/cmsg/mod.rs
```diff
@@ -83,6 +83,7 @@ impl<M: MsgHdr> Drop for Encoder<'_, M> {
 /// `cmsg` must refer to a native cmsg containing a payload of type `T`
 pub(crate) unsafe fn decode<T: Copy, C: CMsgHdr>(cmsg: &impl CMsgHdr) -> T {
     assert!(mem::align_of::<T>() <= mem::align_of::<C>());
+    // If you're hitting this, you might need to raise the CMSG_LEN constants
```
Ralith, line 86: Would this be more helpful as an assert message?

quinn-udp/src/unix.rs
recatek: I pulled it out into an optional function on the socket state anyway for Linux.

quinn-udp/src/lib.rs
recatek: Yep, added a note that it's sourced from CLOCK_REALTIME

quinn-udp/src/unix.rs
recatek: The standalone set function now returns an `io::Result<()>`

quinn-udp/src/unix.rs
```diff
@@ -725,6 +746,13 @@ fn decode_recv(
             (libc::SOL_UDP, gro::UDP_GRO) => unsafe {
                 stride = cmsg::decode::<libc::c_int, libc::cmsghdr>(cmsg) as usize;
             },
+            #[cfg(target_os = "linux")]
+            (libc::SOL_SOCKET, libc::SCM_TIMESTAMP) => {
+                let timeval = unsafe { cmsg::decode::<libc::timeval, libc::cmsghdr>(cmsg) };
+                let secs = Duration::from_secs(timeval.tv_sec as u64);
+                let usecs = Duration::from_micros(timeval.tv_usec as u64);
```
djc, line 753: Feels like [`Duration::new()`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.new) might be more idiomatic for this?

quinn-udp/src/lib.rs
```diff
@@ -115,6 +115,13 @@ pub struct RecvMeta {
     /// Populated on platforms: Windows, Linux, Android (API level > 25),
     /// FreeBSD, OpenBSD, NetBSD, macOS, and iOS.
     pub dst_ip: Option<IpAddr>,
+    /// A timestamp for when the given packet was received by the operating system.
+    /// Timestamping must be enabled for this information to be collected.
+    ///
+    /// Populated on platforms with clock sources:
+    /// - Linux: CLOCK_REALTIME (see docs for SO_TIMESTAMP)
+    #[cfg(not(wasm_browser))]
```
djc, line 123: Can we just make this `cfg(os = "linux")` only for now? Not much point in making this available as `None`-only on platforms where we don't have support for it.

djc (commented):
Mostly LGTM, please squash into a single commit.

quinn-udp/src/unix.rs
```diff
@@ -257,6 +265,17 @@ impl UdpSocketState {
         self.may_fragment
     }
 
+    /// Sets the socket to receive packet receipt timestamps.
+    #[cfg(target_os = "linux")]
+    pub fn set_enable_rx_timestamps(&self, sock: UdpSockRef<'_>, enabled: bool) -> io::Result<()> {
```
mxinden, line 270: I think `_enable_` is misleading, given that I can call `set_enable_rx_timestamps(some_sock, false)`.

How about:

```suggestion
    pub fn set_rx_timestamps(&self, sock: UdpSockRef<'_>, enabled: bool) -> io::Result<()> {
```

Consistent with e.g. `set_gro` on Windows.

mxinden (commented):
I am in favor of this feature. Thanks for the work.

recatek:
I changed the timestamp type a bit to be an enum now based on the clock and type it's from. This way it'll ZST out on platforms that don't support it.

djc (commented):
(And don't forget to squash your changes.)

mxinden:
Using `SystemTime` sounds like the reasonable choice to me.

I don't understand the benefit of nesting a `SystemTime` in an additional `enum`. Can you expand @recatek?

quinn-udp/src/lib.rs
```diff
@@ -126,10 +132,18 @@ impl Default for RecvMeta {
             stride: 0,
             ecn: None,
             dst_ip: None,
+            timestamp: None,
         }
     }
 }
 
+#[derive(Debug, Copy, Clone)]
```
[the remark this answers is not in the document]
mxinden: Note that technically adding a field to `RecvMeta` is already a breaking change, given that `RecvMeta` is not marked as `#[non_exhaustive]`.

I  don't feel strongly about it.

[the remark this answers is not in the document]
djc: Oh, good point. I think we should probably add a commit here (or soon after) that bumps the version to 0.6.0, then.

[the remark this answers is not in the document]
mxinden: Works for me. We can then also include https://github.com/quinn-rs/quinn/pull/1930 in v0.6.0.

recatek:
> I don't understand the benefit of nesting a `SystemTime` in an additional `enum`. Can you expand @recatek?

I'm eyeing future support for FreeBSD (which I am considering using for one use case). FreeBSD has a [SO_TS_CLOCK sockopt that lets you set SO_TS_MONOTONIC](https://github.com/nix-rust/nix/pull/2187) to receive timestamps from CLOCK_MONOTONIC instead of CLOCK_REALTIME. In this case we'd want a different type in there (ideally Instant, but probably another time type like Monotonic since Instant ["can't" be constructed](https://internals.rust-lang.org/t/os-specific-extensions-for-creating-instant/22408)).

So if/when I propose that change, it would look like

```rust
pub enum RecvTime {
    Realtime(SystemTime),
    Monotonic(Monotonic),
}
```

It's a good call to make RecvTime `#[non_exhaustive]` -- I'll do that as well.

recatek:
I'm using the receipt times to modify the Instant that I provide to quinn-proto's `now` Instant value, to get more accurate ack_delay calculation.

Ralith:
I'm not sure if that's necessary or correct. ACK delay describes the time that the QUIC endpoint deliberately waits, not delays forced on it by lower layers or intermediaries. Any time you pass a `now` in the past into a Quinn API, you're risking that Quinn will expect to be able to schedule additional work at a time that it thinks is the future, but which is actually the past, causing obscure performance degradation elsewhere.

recatek:
Ah, yes, you're right. I missed the distinction between intentional delays and unintentional delays [in the spec](https://www.rfc-editor.org/rfc/rfc9000#section-13.2.5). However, it's still important to be able to compare the timestamps with a now() call (be it SystemTime, Instant, or some clock_gettime) when you go to send for manually computing your own RTT calculations on top of QUIC. With that in mind, it's important to control/specify what clock those packet timestamps are sourced from rather than an arbitrary epoch.

recatek:
After discussion on Discord, I converted the timestamp back to a Duration. It'll make it easier to interface with quinn-proto in my next PR for getting the timestamp info into Datagrams

quinn-udp/src/lib.rs
```diff
@@ -115,6 +115,13 @@ pub struct RecvMeta {
     /// Populated on platforms: Windows, Linux, Android (API level > 25),
     /// FreeBSD, OpenBSD, NetBSD, macOS, and iOS.
     pub dst_ip: Option<IpAddr>,
+    /// A timestamp for when the given packet was received by the operating system.
+    /// Controlled by the `set_recv_timestamping` option on the source socket where available.
+    ///
+    /// Populated on platforms with varying clock sources, as follows:
+    /// - Linux: `CLOCK_REALTIME` (see `SO_TIMESTAMPNS` in `man 7 socket` for more information)
```
mxinden, line 122: To me the main benefit of `quinn-udp` is its single interface abstraction over the various concrete OS-specific UDP mechanisms. Thus I would like to strive for a single interface to the user, abstracting over many potentially different OS-specific implementations. Strive, as in not always achievable.

I am fine with `RecvMeta::timestamp` only being available on certain OSs, e.g. via `#[cfg(`. Bugs on the user side will be detected at compile-time.

What I would like to prevent is the value of `timestamp` to have different meanings across OSs, but no difference in its type signature. Bugs on the user side will only be detected at runtime.

If I understand the above correctly, the `Duration` embedded in `RecvMeta::timestamp` should be interpreted as the time since `UNIX_EPOCH` on Linux. What would we expose on other OSs in the future? Can we find a single abstraction?

recatek:
Added a test as well.

quinn-udp/src/lib.rs
mxinden: Thank you for the above explanation.

I still have concerns with this type signature.
- Representing a timestamp as a `Duration` is counter-intuitive to me.
- Given that quinn-udp controls the socket, does quinn-udp also control the timer configuration, i.e. CLOCK_REALTIME, CLOCK_MONOTONIC, or CLOCK_BOOTTIME?
- While MacOS and FreeBSD might provide these different time sources, do we as `quinn-udp` have to support them? Asked differently, for what use-case are we designing this API? Does this use-case need to support all of these different timers?
- Is one of these timer configurations the de-facto standard, while others are rather niche?

I don't have a strong opinion on this matter. I don't have experience with the syscalls. I don't see myself using it in the next couple of months, though I am eying it for future use in Firefox. Thus I will leave the final decision to the other maintainers.

djc: So @mxinden argued above that semantic differences (that is, in the value of `timestamp`) should be represented in the type system, which on the one hand seems reasonable, but on the other hand it seems unlikely that people are going to want to compare these values across operating systems?

I do agree that at least conceptually, the use of `Duration` as a `timestamp` is not the most intuitive. On the other hand, it seems like a somewhat reasonable representation of a timestamp relative to any kind of epoch. A `u64` newtype (or even a `secs: u64, nsecs: u64` newtype) doesn't really seem like an improvement, except for the name? I suppose we could wrap `Duration` in a newtype and implement `Deref` for it which seems like a low-complexity way to make this look a little more obvious?

As for which configuration/source is used, it seems like quinn-udp would want to select the single best available source for the platform -- which assumes there is a single metric for "best" that works for all relevant use cases; not sure I can think of a counterexample where that is not the case (maybe range vs accuracy?).

djc: @recatek @mxinden should we find a way forward on this?

mxinden: Fine moving forward here. My opinions above still stand. None of them are strong, i.e. please don't block on any of them.

There seems to be lots of open questions above, I assume due to us designing an API for all the use-cases instead of a specific one. May I suggest designing for a single use-case for now? Does a concrete use-case already exist? If so, would you mind sharing (some of) the source code using this API @recatek?

PR #2154, opened 2025-02-12
fix(udp/fast-apple): ignore empty cmsghdr

mxinden:
On MacOS < 14, with `fast-apple-datapath` feature, calls to `libc::CMSG_NXTHDR` might continuously return empty (i.e. all zero) `libc::cmsghdr` instead of a null pointer. This results in a busy loop in `decode_recv`:

``` rust
let cmsg_iter = unsafe { cmsg::Iter::new(hdr) };
for cmsg in cmsg_iter {
	match (cmsg.cmsg_level, cmsg.cmsg_type) {
```
https://github.com/quinn-rs/quinn/blob/b4378bb39dab4b58a1e6a3fea4fff9f87033dab6/quinn-udp/src/unix.rs#L685C1-L687C50

This commit fixes the above, returning a `null_mut()` pointer on an empty `libc::cmsgdhr`, thus terminating the `cmsg_iter`.

See also https://github.com/mozilla/neqo/pull/2427 for details.

### DIFF
--- a/quinn-udp/src/cmsg/unix.rs
+++ b/quinn-udp/src/cmsg/unix.rs
@@ -43,7 +43,18 @@ impl MsgHdr for crate::imp::msghdr_x {
 
     fn cmsg_nxt_hdr(&self, cmsg: &Self::ControlMessage) -> *mut Self::ControlMessage {
         let selfp = self as *const _ as *mut libc::msghdr;
-        unsafe { libc::CMSG_NXTHDR(selfp, cmsg) }
+        let next = unsafe { libc::CMSG_NXTHDR(selfp, cmsg) };
+
+        // On MacOS < 14 CMSG_NXTHDR might continuously return a zeroed cmsg. In
+        // such case, return a null pointer instead, thus indicating the end of
+        // the cmsghdr chain.
+        if unsafe { next.as_ref() }
+            .is_some_and(|n| (n.cmsg_len as usize) < std::mem::size_of::<libc::cmsghdr>())
+        {
+            return std::ptr::null_mut();
+        }
+
+        next
     }
 
     fn set_control_len(&mut self, len: usize) {
--- a/quinn-udp/Cargo.toml
+++ b/quinn-udp/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "quinn-udp"
-version = "0.5.9"
+version = "0.5.10"
 edition.workspace = true
 rust-version.workspace = true
 license.workspace = true


mxinden:
For the record, here is a Mozilla CI run, testing this patch on MacOS 10.15:

https://treeherder.mozilla.org/jobs?repo=try&revision=cc4cbf5f7bd0a48dcab741bb09c0c99cc94fa6da

quinn-udp/src/cmsg/unix.rs
```diff
@@ -43,7 +43,18 @@ impl MsgHdr for crate::imp::msghdr_x {
 
     fn cmsg_nxt_hdr(&self, cmsg: &Self::ControlMessage) -> *mut Self::ControlMessage {
         let selfp = self as *const _ as *mut libc::msghdr;
```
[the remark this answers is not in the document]
larseggert: `libc::CMSG_FIRSTHDR` and `libc::CMSG_NXTHDR` obviously need the `libc::msghdr`. We'd need to implement those for `msghdr_x` equivalent how they are in https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/bsd/sys/socket.h#L690-L713

larseggert:
```
Eijatuuli@eija quinn % RUST_LOG=trace cargo test -p quinn-udp --features fast-apple-datapath basic
   Compiling quinn-udp v0.5.9 (/Users/Eijatuuli/quinn/quinn-udp)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 3.21s
     Running unittests src/lib.rs (target/debug/deps/quinn_udp-921787c1926eb24c)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/tests.rs (target/debug/deps/tests-ba553e2506d92958)

running 1 test
test basic ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 5 filtered out; finished in 0.00s

Eijatuuli@eija quinn % RUST_LOG=trace cargo test -p quinn-udp --features fast-apple-datapath
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.16s
     Running unittests src/lib.rs (target/debug/deps/quinn_udp-921787c1926eb24c)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/tests.rs (target/debug/deps/tests-ba553e2506d92958)

running 6 tests
test basic ... ok
test ecn_v6 ... ok
test ecn_v4 ... ok
test ecn_v4_mapped_v6 ... ok
test gso ... ignored
test ecn_v6_dualstack ... ok

test result: ok. 5 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 0.00s

   Doc-tests quinn_udp

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

quinn-udp/src/cmsg/unix.rs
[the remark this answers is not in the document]
mxinden: @larseggert do I understand correctly that this can be done in a follow-up pull request? I.e. that it is not urgent?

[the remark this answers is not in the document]
mxinden: Also, note that Apple seems to use `CMSG_FIRSTHDR` with `recvmsg_x`:

https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/tests/recvmsg_x_test.c#L146

quinn-udp/src/cmsg/unix.rs
```diff
@@ -43,7 +43,15 @@ impl MsgHdr for crate::imp::msghdr_x {
...
-        unsafe { libc::CMSG_NXTHDR(selfp, cmsg) }
+        let next = unsafe { libc::CMSG_NXTHDR(selfp, cmsg) };
+
+        if next.is_null()
+            || unsafe { (*next).cmsg_len as usize } < std::mem::size_of::<libc::cmsghdr>()
```
thomaseizinger, line 49: Perhaps this is better expressed as:

```rust
if next.as_mut().is_some_and(|n| n.cmesg_len < ..)
```

Alternatively, you could extract the if null return null into a separate branch.

I think whilst relying on short-circuting here is safe, it is also not very inuitive.

quinn-udp/src/cmsg/unix.rs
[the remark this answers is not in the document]
larseggert: If there are any related changes you want to make, we could include them in this PR.

quinn-udp/src/cmsg/unix.rs
```diff
@@ -43,7 +43,15 @@ impl MsgHdr for crate::imp::msghdr_x {
...
+        {
+            return std::ptr::null_mut();
+        }
```
larseggert, line 52: Maybe leave a comment here about why we're doing this?

quinn-udp/src/cmsg/unix.rs
mxinden: Good idea. Thanks @thomaseizinger !

quinn-udp/src/cmsg/unix.rs
mxinden: :+1: Done. Unfortunately I was not able to find an official bug report for this.

mxinden:
> For the record, here is a Mozilla CI run, testing this patch on MacOS 10.15:
> 
> https://treeherder.mozilla.org/jobs?repo=try&revision=cc4cbf5f7bd0a48dcab741bb09c0c99cc94fa6da

All green :tada: 

With quinn-udp [unit tests succeeding](https://github.com/quinn-rs/quinn/pull/2154#issuecomment-2653548258) and and Mozilla CI succeeding, this pull request is ready for review.

quinn-udp/src/cmsg/unix.rs
```diff
@@ -43,7 +43,18 @@ impl MsgHdr for crate::imp::msghdr_x {
 
     fn cmsg_nxt_hdr(&self, cmsg: &Self::ControlMessage) -> *mut Self::ControlMessage {
         let selfp = self as *const _ as *mut libc::msghdr;
-        unsafe { libc::CMSG_NXTHDR(selfp, cmsg) }
+        let next = unsafe { libc::CMSG_NXTHDR(selfp, cmsg) };
+
+        // On MacOS < 14 CMSG_NXTHDR might continuously return a zeroed cmsg. In
+        // such case, return a null pointer instead, thus indicating the end of
+        // the cmsghdr chain.
+        if unsafe { next.as_ref() }
```
djc, line 51: Tiny nit: if you import `size_of()` at the top of the module, does this fit on one line?

djc (approved):

quinn-udp/src/cmsg/unix.rs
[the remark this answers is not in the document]
mxinden: Taking a deeper look:

Given the Apple example explicitly using[ `CMSG_FIRSTHDR` and `CMSG_NXTHDR` with a `msghdr_x`](https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/tests/recvmsg_x_test.c#L146), I don't think we need to re-implement either of these methods.

Given that `libc::CMSG_NXTHDR` takes [`msghdr`](https://docs.rs/libc/latest/libc/struct.msghdr.html) and not `msghdr_x`, I don't think my initial suggestion was valid.

Thus closing here as resolved. Sorry for the noise.

quinn-udp/src/cmsg/unix.rs
mxinden: We could import `size_of` at the top of the module. But, this code is feature flagged behind the `fast-apple-datapath` feature. Thus, when the feature isn't enabled, rustc would warn for an unused import.

We could as well feature flag the import, but that, in my eyes, just adds more noise overall.

Thoughts @djc?

djc: I often prefer to have guarded imports at the top of the module if that reduces noise at the actual call-site. Here, I think if that would avoid the wrapping it's probably worth it to me.

mxinden: Unfortunately `rustfmt` would not remove the line wrap if `size_of` is imported at the module level:

``` diff
diff --git a/quinn-udp/src/cmsg/unix.rs b/quinn-udp/src/cmsg/unix.rs
index 69b01756..1734c353 100644
--- a/quinn-udp/src/cmsg/unix.rs
+++ b/quinn-udp/src/cmsg/unix.rs
@@ -1,4 +1,7 @@
-use std::ffi::{c_int, c_uchar};
+use std::{
+    ffi::{c_int, c_uchar},
+    mem::size_of,
+};
 
 use super::{CMsgHdr, MsgHdr};
 
@@ -49,7 +52,7 @@ impl MsgHdr for crate::imp::msghdr_x {
         // such case, return a null pointer instead, thus indicating the end of
         // the cmsghdr chain.
         if unsafe { next.as_ref() }
-            .is_some_and(|n| (n.cmsg_len as usize) < std::mem::size_of::<libc::cmsghdr>())
+            .is_some_and(|n| (n.cmsg_len as usize) < size_of::<libc::cmsghdr>())
         {
             return std::ptr::null_mut();
         }
```

Still want the import @djc?

djc: Nah, let's skip it then.

djc:
> @djc given that this is not a breaking change and given that this change is behind the fast-apple-datapath feature only, would you be opposed to merging and publishing here soon? I can then include this patch in Firefox and thus we get more data.

Sounds good to me!

Ralith (approved):

thomaseizinger (commented):
What is the likelihood that other BSDs have a similar bug? Should we perhaps guard against this in the `Iterator` implementation?

https://github.com/quinn-rs/quinn/blob/9bcd53d5bacde18fbf667efab1479f98addbe5fe/quinn-udp/src/cmsg/mod.rs#L115

It would also result in slightly less code because we already do the dereferencing and `as_ref` there.

thomaseizinger:
> Thus I assume this bug is strictly related to how `recvmsg_x` writes to the control messages on Macos < 14, and thus I don't think we need a general fix in `Iterator`.
> 
> Would you agree @thomaseizinger?

Makes sense to me!

djc:
@mxinden I have invited you as a crates.io owner for quinn-udp. Want to take care of merging/publishing?

mxinden:
That is great news. Thank you @djc. I will merge, tag and publish now.

mxinden:
[Tagged](https://github.com/quinn-rs/quinn/releases/tag/quinn-udp-0.5.10) and [published](https://crates.io/crates/quinn-udp/0.5.10).

Thank you for the help everyone!

PR #2157, opened 2025-02-18
Remove use of Instant::now() in BBR bandwidth estimation code

recatek:
This was the only case where `Instant::now()` is used in quinn-proto or quinn-udp for anything other than tests or logging. Making this change to keep these two crates more IO-agnostic.

### DIFF
--- a/quinn-proto/src/congestion/bbr/bw_estimation.rs
+++ b/quinn-proto/src/congestion/bbr/bw_estimation.rs
@@ -3,7 +3,7 @@ use std::fmt::{Debug, Display, Formatter};
 use super::min_max::MinMax;
 use crate::{Duration, Instant};
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 pub(crate) struct BandwidthEstimation {
     total_acked: u64,
     prev_total_acked: u64,
@@ -11,7 +11,7 @@ pub(crate) struct BandwidthEstimation {
     prev_acked_time: Option<Instant>,
     total_sent: u64,
     prev_total_sent: u64,
-    sent_time: Instant,
+    sent_time: Option<Instant>,
     prev_sent_time: Option<Instant>,
     max_filter: MinMax,
     acked_at_last_window: u64,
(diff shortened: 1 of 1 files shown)


quinn-proto/src/congestion/bbr/bw_estimation.rs
```diff
@@ -43,10 +43,12 @@ impl BandwidthEstimation {
             None => return,
         };
 
-        let send_rate = if self.sent_time > prev_sent_time {
+        let send_rate = if self.sent_time > Some(prev_sent_time) {
+            // self.sent_time can't be None if it's greater than a Some value
```
recatek, line 47: Ref: https://doc.rust-lang.org/src/core/option.rs.html#2196

quinn-proto/src/congestion/bbr/bw_estimation.rs
```diff
@@ -43,10 +43,12 @@ impl BandwidthEstimation {
...
+            let sent_time = self.sent_time.unwrap();
```
recatek, line 48: Ref: https://doc.rust-lang.org/src/core/option.rs.html#2196

Ralith: I'd prefer to find a way to phrase this without `unwrap`. Maybe something like:
```rs
match self.send_time {
    Some(sent_time) if ... => ...,
    _ => u64::MAX,
}
```

Ralith (approved):
Thanks!

quinn-proto/src/congestion/bbr/bw_estimation.rs
gretchenfrage: I would also prefer something more natural than this combination of unwrapping and being clever about `Option<T>: Ord` in a way that requires a comment to explain, e.g. would `let send_rate = self.sent_time.filter(|sent_time| sent_time > prev_sent_time).unwrap_or(u64::MAX)` work?.

Edit: I see you added another commit, thanks! Just to be clear, it would be good if you'd squash it into the relevant commit to form a clean history. Maybe you know that, but GitHub doesn't provide me a great way to tell when you're done making changes in this situation.

Feel free to go with either the `match` based approach or the `filter` based approach. I think condensing it to a one-liner sounds neat but it's a minor difference.

gretchenfrage (changes requested):
Thanks!

PR #2159, opened 2025-02-21
Clippy 1.85, 2024 style

djc:


Ralith (approved):

djc:
@Ralith added a commit to ignore https://rustsec.org/advisories/RUSTSEC-2025-0007 for now.

(Discussions in the rustls project about how to cope with this in the slightly longer term are ongoing.)

Ralith (approved):

Issue #2160, opened 2025-02-22
Make RecvStream::poll_read_buf public


Kannen:
Hello,

While I was implementing a type that encapsulates `RecvStream` in order to make a `Stream` whose error type is `ReadError`, I could not find a way to read bytes on uninitialized bytes: `RecvStream::poll_read` takes a `&mut [u8]` and the other poll functions implemented in `AsyncRead`, return a `std::io::Error`.

The private function `RecvStream::poll_read_buf` takes a `tokio::io::ReadBuf` that can be initialized with `[MaybeUninit<u8>]` and return a `ReadError`. This is precisely what is needed!

I suppose that may use case is quite common, so probably other will find this helpfull.

It would be great if this function could be made public!

djc:
Sounds okay to me.

Kannen:
Yes, I ll do it soon.

gretchenfrage:
Closed by #2161

PR #2161, opened 2025-02-22
Makes poll_read_buf public

Kannen:
That closes issue #2160

### DIFF
--- a/quinn/src/recv_stream.rs
+++ b/quinn/src/recv_stream.rs
@@ -112,7 +112,16 @@ impl RecvStream {
         Poll::Ready(Ok(buf.filled().len()))
     }
 
-    fn poll_read_buf(
+    /// Attempts to read from the stream into buf.
+    ///
+    /// On success, returns Poll::Ready(Ok(())) and places data in
+    /// the buf. If no data was read and the buffer had a non null capacity,
+    /// it implies that EOF has been reached.
+    ///
+    /// If no data is available for reading, the method returns Poll::Pending
+    /// and arranges for the current task (via cx.waker()) to receive a notification
+    /// when the stream becomes readable or is closed.
+    pub fn poll_read_buf(
         &mut self,
         cx: &mut Context,
         buf: &mut ReadBuf<'_>,


gretchenfrage (changes requested):
Thank for the work! I want to question the premise a bit though...

It looks like the pre-existing `impl tokio::io::AsyncRead for RecvStream` block already basically just delegates to this method?

On your issue you said:

> the other poll functions implemented in `AsyncRead`, return a `std::io::Error`.

Could you elaborate about that? Ideally provide a runnable example?

That doesn't sound like what should be happening, and if it is happening sounds like a thing we should simply fix rather than exposing this method directly, but it's not entirely clear to  me what you're describing there.

Kannen:
Thank you for the question.

The premise is that it is not possible to implement a "poll_read" around a `RecvStream` that would be both efficient and would returned the `ReadError`. For that, `RecvStream` must provide a "poll_read" function that:
1. write on an uninitialized buffer 
2. return the `ReadError` or an equivalent type (injectively mapped).

What are the solution provided to user of quinn?

### `impl tokio::io::AsyncRead for RecvStream` 

1. ✅ Writes a potentialy uninitialized buffer
2. ❌ it maps the `ReadError` to a `std::io::Error`. Doing so it erases part of the information that is in the `ReadError`. For exemple the VarInt in `ReadError::Stopped(var_int)` is lost during the conversion of `ReadError to std::io::Error`.

### `RecvStream::poll_read`

1. ❌ Writes on an initialized buffer
2. ✅ Returns the ReadError

### Solutions

So *indeed*, it is not necessary to make `RecvStream::buf_poll` public to solve this issue.

For example, the buffer type of `RecvStream::poll_read` could be changed.

quinn/src/recv_stream.rs
```diff
@@ -112,7 +112,16 @@ impl RecvStream {
         Poll::Ready(Ok(buf.filled().len()))
     }
 
-    fn poll_read_buf(
+    /// Attempts to read from the stream into buf.
```
djc, line 115: Docstring nits: drop the period from the first sentence, and use backticks to quote the code parts (`Poll::Ready(Ok(())`, `Poll::Pending`, `cx.waker()`).

djc (approved):
This makes sense to me.

Suggest rebasing to fix the lint issues.

gretchenfrage (approved):
I understand now. Thanks!

@djc the comment is copy-pasted from the pre-existing method above--style flaws are essentially pre-existing. I have a diff drafted to copy edit both comments, which I plan to open in a PR after this is merged.

PR #2162, opened 2025-02-23
Fix minor inter-version build issues

Ralith:


quinn-proto/src/connection/mod.rs
```diff
@@ -1240,6 +1240,12 @@ impl Connection {
         self.update_keys(None, false);
     }
 
+    // Compatibility wrapper for quinn < 0.11.7. Remove for 0.12.
+    #[doc(hidden)]
```
djc, line 1244: Should this be `#[deprecated]`, too?

djc (approved):

quinn-proto/src/connection/mod.rs
Ralith: Sure, might as well.

PR #2163, opened 2025-02-23
docs(quinn): Copy edit poll_read(_buf?) docs

gretchenfrage:
Looks to the following for writing inspiration:

- https://docs.rs/tokio/latest/tokio/io/trait.AsyncRead.html#tymethod.poll_read
- https://doc.rust-lang.org/stable/std/io/trait.Read.html#tymethod.read

Achieves:

- Reflowing to 100 columns.
- Style-consistency with code-escaping and initial-line period.
- Distinguishing first lines between the two methods' doc comments.
- Hyperlink to SendStream::finish
- General writing style.

### DIFF
--- a/quinn/src/recv_stream.rs
+++ b/quinn/src/recv_stream.rs
@@ -94,14 +94,17 @@ impl RecvStream {
         .await
     }
 
-    /// Attempts to read from the stream into buf.
+    /// Attempts to read from the stream into the provided buffer
     ///
-    /// On success, returns Poll::Ready(Ok(num_bytes_read)) and places data in
-    /// the buf. If no data was read, it implies that EOF has been reached.
+    /// On success, returns `Poll::Ready(Ok(num_bytes_read))` and places data into `buf`. If this
+    /// returns zero bytes read (and `buf` has a non-zero length), that indicates that the remote
+    /// side has [`finish`]ed the stream and the local side has already read all bytes.
     ///
-    /// If no data is available for reading, the method returns Poll::Pending
-    /// and arranges for the current task (via cx.waker()) to receive a notification
-    /// when the stream becomes readable or is closed.
+    /// If no data is available for reading, this returns `Poll::Pending` and arranges for the
+    /// current task (via `cx.waker()`) to be notified when the stream becomes readable or is
+    /// closed.
+    ///
+    /// [`finish`]: crate::SendStream::finish
     pub fn poll_read(
         &mut self,
         cx: &mut Context,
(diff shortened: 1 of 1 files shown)


quinn/src/recv_stream.rs
```diff
@@ -112,15 +115,17 @@ impl RecvStream {
...
 
-    /// Attempts to read from the stream into buf.
+    /// Attempt to read from the stream into the provided buffer, which may be uninitialized
     ///
-    /// On success, returns Poll::Ready(Ok(())) and places data in
-    /// the buf. If no data was read and the buffer had a non null capacity,
-    /// it implies that EOF has been reached.
+    /// On success, returns `Poll::Ready(Ok(num_bytes_read))` and places data into `buf`. If this
+    /// returns zero bytes read (and `buf` has a non-zero length), that indicates that the remote
+    /// side has [`finish`]ed the stream and the local side has already read all bytes.
```
Ralith, line 122: This needs to be reworded to account for returning a `()` rather than a `usize` on success.

Ralith (changes requested):
Thanks!

djc (approved):
Thanks for following up on this!

quinn/src/recv_stream.rs
```diff
@@ -94,14 +94,17 @@ impl RecvStream {
         .await
     }
 
-    /// Attempts to read from the stream into buf.
+    /// Attempt to read from the stream into the provided buffer
```
flub, line 97: FWIW this goes against the Rust style guide, which I think is nice to follow generally: https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#appendix-a-full-conventions-text

> The summary line should be written in third person singular present indicative form. Basically, this means write ‘Returns’ instead of ‘Return’.

Ralith: This would also be more consistent with the following paragraph 👍

Ralith (approved):

quinn/src/recv_stream.rs
```diff
@@ -94,14 +94,17 @@ impl RecvStream {
...
-    /// On success, returns Poll::Ready(Ok(num_bytes_read)) and places data in
-    /// the buf. If no data was read, it implies that EOF has been reached.
+    /// On success, returns `Poll::Ready(Ok(num_bytes_read))` and places data into `buf`. If this
+    /// returns zero bytes read (and `buf` has a non-zero length), that indicates that the remote
+    /// side has [`finish`]ed the stream and the local side has already read all bytes.
     ///
-    /// If no data is available for reading, the method returns Poll::Pending
-    /// and arranges for the current task (via cx.waker()) to receive a notification
-    /// when the stream becomes readable or is closed.
+    /// [`finish`]: crate::SendStream::finish
```
djc, line 103: Nit: move the reference below the second paragraph, to the bottom of the docstring (same for the other method).

PR #2165, opened 2025-02-26
refactor: move sending path_challenge to its own function

divagant-martian:
While implementing multipath the logic of `poll_transmit` is becoming increasingly involved, and it's already quite a long function with 530+ lines _after_ this refactor. 

This PR aims to reduce the complexity of `poll_transmit` by a simple refactor that moves sending path challenge to its own function. It will help us also reduce the diff size and hopefully the cognitive strain of reviewing multipath for when its ready. In the meantime, this change provides a small improvement that stands on its own.


quinn-proto/src/connection/mod.rs
```diff
@@ -462,61 +520,10 @@ impl Connection {
...
-                    destination,
-                    size: buf.len(),
-                    ecn: None,
-                    segment_size: None,
-                    src_ip: self.local_ip,
-                });
-            }
+        let maybe_with_challenge = self.send_path_challenge(now, buf);
+        if maybe_with_challenge.is_some() {
+            return maybe_with_challenge;
```
Ralith, line 525: tiny nit: this can be a little more terse and narrowly scoped:
```suggestion
        if let Some(challenge) = self.send_path_challenge(now, buf) {
            return Some(challenge);
```

Ralith (approved):
Thanks! Strong agree that `poll_transmit` is getting unmanageable, and I'm very happy to accept a straightforward step towards breaking it up.

quinn-proto/src/connection/mod.rs
djc: Also by our style, the newly extracted function should go below poll_transmit() rather than above it.

divagant-martian:
Thanks for the quick response you all, I've updated the function call code and moved the function to below `poll_transmit`

djc (approved):

PR #2167, opened 2025-03-03
fix: Do not produce tail-loss probes larger than segment size

flub:
When needing to send tail-loss probes and building a GSO batch the tail-loss probe should not exceed the GSO segment size.  At the same time a GSO batch needs to be broken when the segment size exceeds the maximum size for a tail-loss probe.

Somehow we have a reliable reproducer of this in https://github.com/n0-computer/callme/tree/no-graceful-close, but that is quite a bit away from Quinn main.  Applying this fix solves it, but I am not really sure how to test this in the test suite.  There isn't any example in the file of tests constructing their state to test particular poll_transmit behaviours.  Any hints on how I might be able to add tests to this?

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -699,7 +699,7 @@ impl Connection {
                         // Clamp the datagram to at most the minimum MTU to ensure that loss probes
                         // can get through and enable recovery even if the path MTU has shrank
                         // unexpectedly.
-                        usize::from(INITIAL_MTU)
+                        std::cmp::min(segment_size, usize::from(INITIAL_MTU))
                     }
                 };
                 buf_capacity += next_datagram_size_limit;
--- a/quinn-proto/src/tests/mod.rs
+++ b/quinn-proto/src/tests/mod.rs
@@ -1780,6 +1780,55 @@ fn congested_tail_loss() {
     pair.client_send(client_ch, s).write(&[42; 1024]).unwrap();
 }
 
+// Send a tail-loss probe when GSO segment_size is less than INITIAL_MTU
+#[test]
+fn tail_loss_small_segment_size() {
+    let _guard = subscribe();
+    let mut pair = Pair::default();
+    let (client_ch, server_ch) = pair.connect();
+
+    // No datagrams frames received in the handshake.
+    let server_stats = pair.server_conn_mut(server_ch).stats();
+    assert_eq!(server_stats.frame_rx.datagram, 0);
+
+    const DGRAM_LEN: usize = 1000; // Below INITIAL_MTU after packet overhead.
+    const DGRAM_NUM: u64 = 5; // Enough to build a GSO batch.
+
+    info!("Sending an ack-eliciting datagram");
+    pair.client_conn_mut(client_ch).ping();
+    pair.drive_client();
+
+    // Drop these packets on the server side.
+    assert!(!pair.server.inbound.is_empty());
+    pair.server.inbound.clear();
+
+    // Doing one step makes the client advance time to the PTO fire time.
+    info!("stepping forward to PTO");
+    pair.step();
+
+    // Still no datagrams frames received by the server.
+    let server_stats = pair.server_conn_mut(server_ch).stats();
+    assert_eq!(server_stats.frame_rx.datagram, 0);
+
+    // Now we can send another batch of datagrams, so the PTO can send them instead of
+    // sending a ping.  These are small enough that the segment_size is less than the
+    // INITIAL_MTU.
+    info!("Sending datagram batch");
+    for _ in 0..DGRAM_NUM {
+        pair.client_datagrams(client_ch)
+            .send(vec![0; DGRAM_LEN].into(), false)
+            .unwrap();
+    }
+
+    // If this succeeds the datagrams are received by the server and the client did not
+    // crash.
+    pair.drive();
+
+    // Finally the server should have received some datagrams.
+    let server_stats = pair.server_conn_mut(server_ch).stats();
+    assert_eq!(server_stats.frame_rx.datagram, DGRAM_NUM);
+}
+
 #[test]
 fn datagram_send_recv() {
     let _guard = subscribe();


quinn-proto/src/connection/mod.rs
```diff
@@ -695,11 +695,16 @@ impl Connection {
                 let next_datagram_size_limit = match self.spaces[space_id].loss_probes {
                     0 => segment_size,
                     _ => {
-                        self.spaces[space_id].loss_probes -= 1;
+                        if num_datagrams >= 1 && segment_size > usize::from(INITIAL_MTU) {
+                            // When building a GSO batch we can not reduce the size of a
+                            // datagram for a tail-loss probe.
+                            break;
```
Ralith, line 701: We can and do:

https://github.com/quinn-rs/quinn/blob/8d6e48c20b71f7e915c13c33b66e2a09b6d59888/quinn-proto/src/connection/mod.rs#L639-L657

The trailing packet in a GSO batch may be shorter than the others, so we terminate a GSO batch *after* producing such a packet.

Ralith (changes requested):
Thanks for looking into this! Great catch about smaller segment sizes. What exactly is the symptom of this failure? I wonder if it might explain some other difficult-to-reproduce reports we've received.

quinn-proto/src/connection/mod.rs
flub: Oh right, and `buf_capacity` gets set from the `next_datagram_size` which this PR is fixing.  The whole flow of poll_transmit is still rather complex to me, but it's improving :)

I removed the early break here again.

flub:
> What exactly is the symptom of this failure?

The `debug_assert!()` at https://github.com/quinn-rs/quinn/blob/8d6e48c20b71f7e915c13c33b66e2a09b6d59888/quinn-proto/src/connection/mod.rs#L595 fails because `unracked_bytes` was 1200 bytes (in turn because of how `buf_capacity` is set), which is larger then the segment size when this happens.

flub:
I suspect this was so reproducible because that example sends a lot of fixed-size user datagrams (audio data), so the segment size is always the same and the GSO batching is also very similar between runs.  And it needed ungraceful shutdown to trigger so that loss probes are being sent.

Ralith (approved):
Thanks! I suspect this was indeed the underlying issue behind https://github.com/quinn-rs/quinn/issues/2127 as well; glad to have that explained.

Ralith:
>  Any hints on how I might be able to add tests to this?

I think the workload you've described is exactly how we should test this: contrive a connection which sends flights of fixed-size application datagrams that don't evenly divide the MTU, and cause the receiver to stop responding so that a tail loss probe is sent. That should be expressible in a quinn-proto-level test, I think? An integration test might work as well but would be more fragile.

quinn-proto/src/connection/mod.rs
```diff
@@ -699,7 +699,7 @@ impl Connection {
                         // Clamp the datagram to at most the minimum MTU to ensure that loss probes
                         // can get through and enable recovery even if the path MTU has shrank
                         // unexpectedly.
-                        usize::from(INITIAL_MTU)
+                        std::cmp::min(segment_size, usize::from(INITIAL_MTU))
```
djc, line 702: Nit: I like to spell this as `Ord::min(..)`, which is a little less noisy.

djc (approved):
Thanks!

quinn-proto/src/connection/mod.rs
kixelated:
```suggestion
                        segment_size.min(INITIAL_MTU as _)
```

djc: Personally I find the symmetric spelling easier to understand.

flub:
Ah, I think I might have a sense of how to test this in quinn-proto.  It'll be a good exercise for me.  I'm still aiming to add tests to this PR.

quinn-proto/src/connection/mod.rs
flub: Given it is a nit and there seems to be no consensus on the spelling I'm going to leave it.  Personally I have no preference and find all of the ways to write this fine.  I left the `usize::from()` part because that was how the code already did the cast.

flub:
I think I managed to write a decent test for it.  I would appreciate if you could check that as well.

Potentially the test surfaces another bug though.  I believe it shows that the first loss probe sent in this case contains 3 frames: PING, IMMEDIATE_ACK and DATAGRAM.  It was my understanding that if there was any legitimate data to send no ping frame should be included, so I was expecting only IMMEDIATE_ACK and DATAGRAM frames in the first loss probe.

Here the bit of logs:
```
TRACE client: quinn_proto::connection: timeout timer=LossDetection
TRACE client: quinn_proto::connection: PTO fired in_flight=5145 count=0 space=Data
TRACE client:send{space=Data pn=15}: quinn_proto::connection: PING
TRACE client:send{space=Data pn=15}: quinn_proto::connection: IMMEDIATE_ACK
TRACE client:send{space=Data pn=15}: quinn_proto::connection::datagrams: DATAGRAM len=1000
TRACE client:send{space=Data pn=16}: quinn_proto::connection::datagrams: DATAGRAM len=1000
TRACE client:send{space=Data pn=16}: quinn_proto::connection::packet_builder: PADDING * 2
TRACE client:send{space=Data pn=17}: quinn_proto::connection::datagrams: DATAGRAM len=1000
TRACE client:send{space=Data pn=17}: quinn_proto::connection::packet_builder: PADDING * 2
TRACE client:send{space=Data pn=18}: quinn_proto::connection::datagrams: DATAGRAM len=1000
TRACE client:send{space=Data pn=18}: quinn_proto::connection::packet_builder: PADDING * 2
TRACE client:send{space=Data pn=19}: quinn_proto::connection::datagrams: DATAGRAM len=1000
TRACE client: quinn_proto::connection: sending 5153 bytes in 5 datagrams
TRACE server: quinn_proto::connection: got Data packet (1031 bytes) from [::1]:44433 using id 7f201446e7820b1a
TRACE server:recv{space=Data pn=15}: quinn_proto::connection: got frame Ping
TRACE server:recv{space=Data pn=15}: quinn_proto::connection: got frame ImmediateAck
TRACE server:recv{space=Data pn=15}: quinn_proto::connection: got datagram frame len=1000
[... server showing the other frames received all containing a DATAGRAM frame ...]
```

This isn't a very harmful bug.  It is wasting exactly one byte in the loss probe (and thus in the entire GSO batch as you can see by the padding).  But it probably is still a bug?

Ralith:
Good catch. `PING` is actually redundant to `IMMEDIATE_ACK` too, there. Drafted https://github.com/quinn-rs/quinn/pull/2169 to resolve both cases.

quinn-proto/src/tests/mod.rs
```diff
@@ -1780,6 +1780,59 @@ fn congested_tail_loss() {
...
+    let _guard = subscribe();
+    let mut pair = Pair::default();
+    let (client_ch, server_ch) = pair.connect();
+
+    // No datagrams frames received in the handshake.
+    let server_stats = pair.server_conn_mut(server_ch).stats();
+    assert_eq!(server_stats.frame_rx.datagram, 0);
+
+    const DGRAM_LEN: usize = 1000; // Below INITIAL_MTU after packet overhead.
+    const DGRAM_NUM: usize = 5; // Enough to build a GSO batch.
```
Ralith, line 1795: Does this need to be larger than 2?

quinn-proto/src/tests/mod.rs
```diff
@@ -1780,6 +1780,59 @@ fn congested_tail_loss() {
...
+
+    info!("Sending 1st datagram batch");
+    for _ in 0..DGRAM_NUM {
+        pair.client_datagrams(client_ch)
+            .send(vec![0; DGRAM_LEN].into(), false)
+            .unwrap();
+    }
+    pair.drive_client();
```
Ralith, line 1803: Does this flight need to have any particular structure, or could any ACK-eliciting packet (e.g. a single ping) do the trick?

quinn-proto/src/tests/mod.rs
```diff
@@ -1780,6 +1780,59 @@ fn congested_tail_loss() {
...
+            .unwrap();
+    }
+
+    // If this succeeds the datagrams are received by the server and the client did not
+    // crash.
+    pair.drive();
+
+    // Finally the server should have received some datagrams.
+    let server_stats = pair.server_conn_mut(server_ch).stats();
+    assert_eq!(server_stats.frame_rx.datagram, 5);
```
Ralith, line 1833:
```suggestion
    assert_eq!(server_stats.frame_rx.datagram, DGRAM_NUM);
```

Ralith (approved):
Looks good overall, thanks! You've validated that the test fails without the fix?

djc (approved):

quinn-proto/src/tests/mod.rs
flub: Yes, it needs to be 3 at least because the assertion happens in the next iteration when it is getting ready to build the 3rd datagram.  Rather than than making it fail exactly I thought adding a few datagrams would make it more reliable.  We want to make sure the entire batch works correctly.

quinn-proto/src/tests/mod.rs
flub: Indeed, any ack-eliciting packet is fine.  I went with sending the same twice but maybe ack is even more clear.  Changed this.

flub:
> Looks good overall, thanks! You've validated that the test fails without the fix?

Indeed, removing the fix makes the test fail.

quinn-proto/src/tests/mod.rs
Ralith: Actually, why wouldn't the issue be triggered after a *single* datagram? That should begin a GSO batch of size 1000-ish, and if a TLP is queued, that'll be the next packet, and end up too large.

flub:
pushed the git autosquash now

quinn-proto/src/tests/mod.rs
flub: On the very first datagram it sets the size limit to 1200 because that is the maximum size it is allowed to produce.  Once the first datagram is populated however the first datagram is only 1030 bytes.  So when `next_datagram_size_limit` is being set for the 2nd datagram it is being set to 1030 because this is now the segment size.

The assert that notices the error happens earlier however, when congestion controller and pacing are checked if another datagram is still allows.  So the assertion failure only happens on the 3rd iteration.

dlon:
Hey @djc. Can we get a patch release with this fix?

djc:
Someone want to submit a version bump PR? Sorry, have been having a very busy week.

faern:
Done https://github.com/quinn-rs/quinn/pull/2209

PR #2168, opened 2025-03-07
refactor: Introduce a TransmitBuf for poll_transmit

flub:
Hi, I've been looking around in `poll_receive` recently and started wondering if it could be simplified.  This is my first attempt at a step in this direction, it pulls together the state about datagrams and their boundaries in the write buffer.  I think this helps understand the interactions between a bunch of state changes that were previously spread across many lines.  See the commit message for a detailed description.

If the general direction is considered useful I think I'd like to continue this further.  The next thing I'd probably look at is the buffer management for a packet.  I expect a lot of places don't need to know much about the full transmit buffer they're writing into but rather need to know the (start, len, end) space they can write their frames.  Probably reducing the need to do everything with offsets from the transmit buffer.

Anyway, my aim is the make things clearer and easier to manage.  Let's first see if this is step in that direction and then see what the future steps turn out to be.  I think any change like this should be able to stand on its own rather than rely on a future change, because the future is always uncertain and might never happen :)

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -89,6 +89,7 @@ mod timer;
 use timer::{Timer, TimerTable};
 
 mod transmit_builder;
+pub(crate) use transmit_builder::BufSlice;
 use transmit_builder::TransmitBuilder;
 
 /// Protocol state and logic for a single QUIC connection
@@ -536,13 +537,13 @@ impl Connection {
                 ack_eliciting |= self.can_send_1rtt(frame_space_1rtt);
             }
 
-            // Can we append more data into the current buffer?
-            // It is not safe to assume that `buf.len()` is the end of the data,
-            // since the last packet might not have been finished.
+            // Can we append more data into the current datagram?
+            // It is not safe to assume that `transmits.datagram().len()` is the end of the
+            // data, since the last packet might not have been finished.
             let buf_end = if let Some(builder) = &builder_storage {
-                transmit.len().max(builder.min_size) + builder.tag_len
+                transmit.datagram().len().max(builder.min_size) + builder.tag_len
             } else {
-                transmit.len()
+                transmit.datagram().len()
             };
 
             let tag_len = if let Some(ref crypto) = self.spaces[space_id].crypto {
@@ -558,7 +559,9 @@ impl Connection {
             // We are NOT coalescing (the default is we are, so this was turned off in an
             // earlier iteration) OR there is not enough space for another *packet* in this
             // datagram (buf_capacity - buf_end == unused space in datagram).
-            if !coalesce || transmit.datagram_max_offset() - buf_end < MIN_PACKET_SPACE + tag_len {
+            if !coalesce
+                || transmit.datagram_mut().capacity() - buf_end < MIN_PACKET_SPACE + tag_len
+            {
                 // We need to send 1 more datagram and extend the buffer for that.
 
                 // Is 1 more datagram allowed?
@@ -585,7 +588,7 @@ impl Connection {
                 if ack_eliciting && self.spaces[space_id].loss_probes == 0 {
                     // Assume the current packet will get padded to fill the segment
                     let untracked_bytes = if let Some(builder) = &builder_storage {
-                        transmit.datagram_max_offset() - builder.partial_encode.start
+                        transmit.datagram().len() - builder.partial_encode.start
                     } else {
                         0
                     } as u64;
@@ -639,9 +642,10 @@ impl Connection {
                         // MTU. Loss probes are the only packets for which we might grow
                         // `buf_capacity` by less than `segment_size`.
                         const MAX_PADDING: usize = 16;
-                        let packet_len_unpadded = cmp::max(builder.min_size, transmit.len())
-                            - transmit.datagram_start_offset()
-                            + builder.tag_len;
+                        let packet_len_unpadded =
+                            cmp::max(builder.min_size, transmit.datagram().len())
+                                - builder.partial_encode.start
+                                + builder.tag_len;
                         if packet_len_unpadded + MAX_PADDING < transmit.segment_size()
                             || transmit.datagram_start_offset() + transmit.segment_size()
                                 > transmit.datagram_max_offset()
@@ -766,11 +770,13 @@ impl Connection {
                 // to encode the ConnectionClose frame too. However we still have the
                 // check here to prevent crashes if something changes.
                 debug_assert!(
-                    transmit.len() + frame::ConnectionClose::SIZE_BOUND < builder.max_size,
-                    "ACKs should leave space for ConnectionClose"
+                    transmit.datagram().len() + frame::ConnectionClose::SIZE_BOUND
+                        < builder.max_size,
+                    "ACKS should leave space for ConnectionClose"
                 );
-                if transmit.len() + frame::ConnectionClose::SIZE_BOUND < builder.max_size {
-                    let max_frame_size = builder.max_size - transmit.len();
+                if transmit.datagram().len() + frame::ConnectionClose::SIZE_BOUND < builder.max_size
+                {
+                    let max_frame_size = builder.max_size - transmit.datagram().len();
                     match self.state {
                         State::Closed(state::Closed { ref reason }) => {
                             if space_id == SpaceId::Data || reason.is_transport_layer() {
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -760,6 +760,8 @@ impl Connection {
 
             if close {
                 trace!("sending CONNECTION_CLOSE");
+                let mut datagram = transmit.datagram_mut();
+
                 // Encode ACKs before the ConnectionClose message, to give the receiver
                 // a better approximate on what data has been processed. This is
                 // especially important with ack delay, since the peer might not
@@ -770,7 +772,7 @@ impl Connection {
                         self.receiving_ecn,
                         &mut SentFrames::default(),
                         &mut self.spaces[space_id],
-                        &mut transmit.datagram_mut(),
+                        &mut datagram,
                         &mut self.stats,
                     );
                 }
@@ -779,24 +781,22 @@ impl Connection {
                 // to encode the ConnectionClose frame too. However we still have the
                 // check here to prevent crashes if something changes.
                 debug_assert!(
-                    transmit.datagram().len() + frame::ConnectionClose::SIZE_BOUND
-                        < builder.max_size,
+                    datagram.len() + frame::ConnectionClose::SIZE_BOUND < builder.max_size,
                     "ACKS should leave space for ConnectionClose"
                 );
-                if transmit.datagram().len() + frame::ConnectionClose::SIZE_BOUND < builder.max_size
-                {
-                    let max_frame_size = builder.max_size - transmit.datagram().len();
+                if datagram.len() + frame::ConnectionClose::SIZE_BOUND < builder.max_size {
+                    let max_frame_size = builder.max_size - datagram.len();
                     match self.state {
                         State::Closed(state::Closed { ref reason }) => {
                             if space_id == SpaceId::Data || reason.is_transport_layer() {
-                                reason.encode(&mut transmit.datagram_mut(), max_frame_size)
+                                reason.encode(&mut datagram, max_frame_size)
                             } else {
                                 frame::ConnectionClose {
                                     error_code: TransportErrorCode::APPLICATION_ERROR,
                                     frame_type: None,
                                     reason: Bytes::new(),
                                 }
-                                .encode(&mut transmit.datagram_mut(), max_frame_size)
+                                .encode(&mut datagram, max_frame_size)
                             }
                         }
                         State::Draining => frame::ConnectionClose {
@@ -804,7 +804,7 @@ impl Connection {
                             frame_type: None,
                             reason: Bytes::new(),
                         }
-                        .encode(&mut transmit.datagram_mut(), max_frame_size),
+                        .encode(&mut datagram, max_frame_size),
                         _ => unreachable!(
                             "tried to make a close packet when the connection wasn't closed"
                         ),
@@ -827,15 +827,14 @@ impl Connection {
             // Send an off-path PATH_RESPONSE. Prioritized over on-path data to ensure that path
             // validation can occur while the link is saturated.
             if space_id == SpaceId::Data && transmit.num_datagrams() == 1 {
+                let mut datagram = transmit.datagram_mut();
                 if let Some((token, remote)) = self.path_responses.pop_off_path(self.path.remote) {
                     // `unwrap` guaranteed to succeed because `builder_storage` was populated just
                     // above.
                     let mut builder = builder_storage.take().unwrap();
                     trace!("PATH_RESPONSE {:08x} (off-path)", token);
-                    transmit
-                        .datagram_mut()
-                        .write(frame::FrameType::PATH_RESPONSE);
-                    transmit.datagram_mut().write(token);
+                    datagram.write(frame::FrameType::PATH_RESPONSE);
+                    datagram.write(token);
                     self.stats.frame_tx.path_response += 1;
                     builder.pad_to(MIN_INITIAL_SIZE);
                     builder.finish_and_track(
@@ -845,7 +844,7 @@ impl Connection {
                             non_retransmits: true,
                             ..SentFrames::default()
                         }),
-                        &mut transmit.datagram_mut(),
+                        &mut datagram,
                     );
                     self.stats.udp_tx.on_sent(1, transmit.len());
                     return Some(Transmit {
@@ -859,8 +858,9 @@ impl Connection {
             }
 
             let sent = {
-                let frame_space = builder.max_size - transmit.datagram_mut().len();
-                let mut buf = transmit.datagram_mut().limit(frame_space);
+                let datagram = transmit.datagram_mut();
+                let frame_space = builder.max_size - datagram.len();
+                let mut buf = datagram.limit(frame_space);
                 self.populate_packet(now, space_id, &mut buf, builder.exact_number)
             };
 
@@ -916,26 +916,23 @@ impl Connection {
 
             debug_assert_eq!(transmit.num_datagrams(), 0);
             transmit.start_new_datagram_with_size(probe_size as usize);
-
-            debug_assert!(transmit.datagram().is_empty());
+            let mut datagram = transmit.datagram_mut();
             let mut builder = PacketBuilder::new(
                 now,
                 space_id,
                 self.rem_cids.active(),
-                &mut transmit.datagram_mut(),
+                &mut datagram,
                 true,
                 self,
             )?;
 
             // We implement MTU probes as ping packets padded up to the probe size
-            transmit.datagram_mut().write(frame::FrameType::PING);
+            datagram.write(frame::FrameType::PING);
             self.stats.frame_tx.ping += 1;
 
             // If supported by the peer, we want no delays to the probe's ACK
             if self.peer_supports_ack_frequency() {
-                transmit
-                    .datagram_mut()
-                    .write(frame::FrameType::IMMEDIATE_ACK);
+                datagram.write(frame::FrameType::IMMEDIATE_ACK);
                 self.stats.frame_tx.immediate_ack += 1;
             }
 
@@ -944,7 +941,7 @@ impl Connection {
                 non_retransmits: true,
                 ..Default::default()
             };
-            builder.finish_and_track(now, self, Some(sent_frames), &mut transmit.datagram_mut());
+            builder.finish_and_track(now, self, Some(sent_frames), &mut datagram);
 
             self.stats.path.sent_plpmtud_probes += 1;
 
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -647,8 +647,7 @@ impl Connection {
                                 - builder.partial_encode.start
                                 + builder.tag_len;
                         if packet_len_unpadded + MAX_PADDING < transmit.segment_size()
-                            || transmit.datagram_start_offset() + transmit.segment_size()
-                                > transmit.datagram_max_offset()
+                            || transmit.datagram_mut().capacity() < transmit.segment_size()
                         {
                             trace!(
                                 "GSO truncated by demand for {} padding bytes or loss probe",
@@ -712,7 +711,7 @@ impl Connection {
                 }
             }
 
-            debug_assert!(transmit.datagram_max_offset() - transmit.len() >= MIN_PACKET_SPACE);
+            debug_assert!(transmit.datagram_mut().remaining_mut() >= MIN_PACKET_SPACE);
 
             //
             // From here on, we've determined that a packet will definitely be sent.
@@ -867,8 +866,7 @@ impl Connection {
                 !(sent.is_ack_only(&self.streams)
                     && !can_send.acks
                     && can_send.other
-                    && (transmit.datagram_max_offset() - builder.datagram_start)
-                        == self.path.current_mtu() as usize
+                    && transmit.datagram_mut().capacity() == self.path.current_mtu() as usize
                     && self.datagrams.outgoing.is_empty()),
                 "SendableFrames was {can_send:?}, but only ACKs have been written"
             );
@@ -911,7 +909,7 @@ impl Connection {
             debug_assert_eq!(transmit.num_datagrams(), 0);
             transmit.start_new_datagram_with_size(probe_size as usize);
 
-            debug_assert_eq!(transmit.datagram_start_offset(), 0);
+            debug_assert!(transmit.datagram().is_empty());
             let mut builder = PacketBuilder::new(
                 now,
                 space_id,
@@ -1003,7 +1001,7 @@ impl Connection {
         // sent once, immediately after migration, when the CID is known to be valid. Even
         // if a post-migration packet caused the CID to be retired, it's fair to pretend
         // this is sent first.
-        debug_assert_eq!(buf.datagram_start_offset(), 0);
+        debug_assert!(buf.datagram().is_empty());
         let mut builder = PacketBuilder::new(now, SpaceId::Data, *prev_cid, buf, false, self)?;
         trace!("validating previous path with PATH_CHALLENGE {:08x}", token);
         buf.datagram_mut().write(frame::FrameType::PATH_CHALLENGE);
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -89,7 +89,7 @@ mod timer;
 use timer::{Timer, TimerTable};
 
 mod transmit_builder;
-pub(crate) use transmit_builder::BufSlice;
+pub(crate) use transmit_builder::DatagramBuffer;
 use transmit_builder::TransmitBuilder;
 
 /// Protocol state and logic for a single QUIC connection
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -757,7 +757,7 @@ impl Connection {
                         self.receiving_ecn,
                         &mut SentFrames::default(),
                         &mut self.spaces[space_id],
-                        &mut transmit,
+                        &mut transmit.datagram_mut(),
                         &mut self.stats,
                     );
                 }
@@ -774,14 +774,14 @@ impl Connection {
                     match self.state {
                         State::Closed(state::Closed { ref reason }) => {
                             if space_id == SpaceId::Data || reason.is_transport_layer() {
-                                reason.encode(&mut transmit, max_frame_size)
+                                reason.encode(&mut transmit.datagram_mut(), max_frame_size)
                             } else {
                                 frame::ConnectionClose {
                                     error_code: TransportErrorCode::APPLICATION_ERROR,
                                     frame_type: None,
                                     reason: Bytes::new(),
                                 }
-                                .encode(&mut transmit, max_frame_size)
+                                .encode(&mut transmit.datagram_mut(), max_frame_size)
                             }
                         }
                         State::Draining => frame::ConnectionClose {
@@ -789,7 +789,7 @@ impl Connection {
                             frame_type: None,
                             reason: Bytes::new(),
                         }
-                        .encode(&mut transmit, max_frame_size),
+                        .encode(&mut transmit.datagram_mut(), max_frame_size),
                         _ => unreachable!(
                             "tried to make a close packet when the connection wasn't closed"
                         ),
@@ -817,8 +817,10 @@ impl Connection {
                     // above.
                     let mut builder = builder_storage.take().unwrap();
                     trace!("PATH_RESPONSE {:08x} (off-path)", token);
-                    transmit.write(frame::FrameType::PATH_RESPONSE);
-                    transmit.write(token);
+                    transmit
+                        .datagram_mut()
+                        .write(frame::FrameType::PATH_RESPONSE);
+                    transmit.datagram_mut().write(token);
                     self.stats.frame_tx.path_response += 1;
                     builder.pad_to(MIN_INITIAL_SIZE);
                     builder.finish_and_track(
@@ -844,7 +846,7 @@ impl Connection {
             let sent = self.populate_packet(
                 now,
                 space_id,
-                &mut transmit,
+                &mut transmit.datagram_mut(),
                 builder.max_size,
                 builder.exact_number,
             );
@@ -914,12 +916,14 @@ impl Connection {
             )?;
 
             // We implement MTU probes as ping packets padded up to the probe size
-            transmit.write(frame::FrameType::PING);
+            transmit.datagram_mut().write(frame::FrameType::PING);
             self.stats.frame_tx.ping += 1;
 
             // If supported by the peer, we want no delays to the probe's ACK
             if self.peer_supports_ack_frequency() {
-                transmit.write(frame::FrameType::IMMEDIATE_ACK);
+                transmit
+                    .datagram_mut()
+                    .write(frame::FrameType::IMMEDIATE_ACK);
                 self.stats.frame_tx.immediate_ack += 1;
             }
 
@@ -996,8 +1000,8 @@ impl Connection {
         debug_assert_eq!(buf.datagram_start_offset(), 0);
         let mut builder = PacketBuilder::new(now, SpaceId::Data, *prev_cid, buf, false, self)?;
         trace!("validating previous path with PATH_CHALLENGE {:08x}", token);
-        buf.write(frame::FrameType::PATH_CHALLENGE);
-        buf.write(token);
+        buf.datagram_mut().write(frame::FrameType::PATH_CHALLENGE);
+        buf.datagram_mut().write(token);
         self.stats.frame_tx.path_challenge += 1;
 
         // An endpoint MUST expand datagrams that contain a PATH_CHALLENGE frame
@@ -3982,6 +3986,12 @@ pub(crate) trait BufLen {
 impl BufLen for Vec<u8> {
     fn len(&self) -> usize {
         self.len()
+    }
+}
+
+impl BufLen for bytes::buf::Limit<&mut Vec<u8>> {
+    fn len(&self) -> usize {
+        self.get_ref().len()
     }
 }
 
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -964,20 +964,12 @@ impl Connection {
             .udp_tx
             .on_sent(transmit.num_datagrams() as u64, transmit.len());
 
-        Some(Transmit {
-            destination: self.path.remote,
-            size: transmit.len(),
-            ecn: if self.path.sending_ecn {
-                Some(EcnCodepoint::Ect0)
-            } else {
-                None
-            },
-            segment_size: match transmit.num_datagrams() {
-                1 => None,
-                _ => Some(transmit.segment_size()),
-            },
-            src_ip: self.local_ip,
-        })
+        let ecn = if self.path.sending_ecn {
+            Some(EcnCodepoint::Ect0)
+        } else {
+            None
+        };
+        Some(transmit.build(self.path.remote, ecn, self.local_ip))
     }
 
     /// Send PATH_CHALLENGE for a previous path if necessary
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -519,7 +519,7 @@ impl Connection {
             // always spill into the next datagram.
             let pn = self.packet_number_filter.peek(&self.spaces[SpaceId::Data]);
             let frame_space_1rtt = transmit
-                .segment_size
+                .segment_size()
                 .saturating_sub(self.predict_1rtt_overhead(Some(pn)));
 
             // Is there data or a close message to send in this space?
@@ -558,11 +558,11 @@ impl Connection {
             // We are NOT coalescing (the default is we are, so this was turned off in an
             // earlier iteration) OR there is not enough space for another *packet* in this
             // datagram (buf_capacity - buf_end == unused space in datagram).
-            if !coalesce || transmit.buf_capacity - buf_end < MIN_PACKET_SPACE + tag_len {
+            if !coalesce || transmit.datagram_max_offset() - buf_end < MIN_PACKET_SPACE + tag_len {
                 // We need to send 1 more datagram and extend the buffer for that.
 
                 // Is 1 more datagram allowed?
-                if transmit.num_datagrams >= transmit.max_datagrams {
+                if transmit.num_datagrams() >= transmit.max_datagrams() {
                     // No more datagrams allowed
                     break;
                 }
@@ -574,7 +574,7 @@ impl Connection {
                 // budget left, we always allow a full MTU to be sent
                 // (see https://github.com/quinn-rs/quinn/issues/1082)
                 if self.path.anti_amplification_blocked(
-                    (transmit.segment_size * transmit.num_datagrams) as u64 + 1,
+                    (transmit.segment_size() * transmit.num_datagrams()) as u64 + 1,
                 ) {
                     trace!("blocked by anti-amplification");
                     break;
@@ -585,13 +585,13 @@ impl Connection {
                 if ack_eliciting && self.spaces[space_id].loss_probes == 0 {
                     // Assume the current packet will get padded to fill the segment
                     let untracked_bytes = if let Some(builder) = &builder_storage {
-                        transmit.buf_capacity - builder.partial_encode.start
+                        transmit.datagram_max_offset() - builder.partial_encode.start
                     } else {
                         0
                     } as u64;
-                    debug_assert!(untracked_bytes <= transmit.segment_size as u64);
+                    debug_assert!(untracked_bytes <= transmit.segment_size() as u64);
 
-                    let bytes_to_send = transmit.segment_size as u64 + untracked_bytes;
+                    let bytes_to_send = transmit.segment_size() as u64 + untracked_bytes;
                     if self.path.in_flight.bytes + bytes_to_send >= self.path.congestion.window() {
                         space_idx += 1;
                         congestion_blocked = true;
@@ -625,7 +625,7 @@ impl Connection {
                         builder.pad_to(MIN_INITIAL_SIZE);
                     }
 
-                    if transmit.num_datagrams > 1 {
+                    if transmit.num_datagrams() > 1 {
                         // If too many padding bytes would be required to continue the GSO batch
                         // after this packet, end the GSO batch here. Ensures that fixed-size frames
                         // with heterogeneous sizes (e.g. application datagrams) won't inadvertently
@@ -640,15 +640,15 @@ impl Connection {
                         // `buf_capacity` by less than `segment_size`.
                         const MAX_PADDING: usize = 16;
                         let packet_len_unpadded = cmp::max(builder.min_size, transmit.len())
-                            - transmit.datagram_start
+                            - transmit.datagram_start_offset()
                             + builder.tag_len;
-                        if packet_len_unpadded + MAX_PADDING < transmit.segment_size
-                            || transmit.datagram_start + transmit.segment_size
-                                > transmit.buf_capacity
+                        if packet_len_unpadded + MAX_PADDING < transmit.segment_size()
+                            || transmit.datagram_start_offset() + transmit.segment_size()
+                                > transmit.datagram_max_offset()
                         {
                             trace!(
                                 "GSO truncated by demand for {} padding bytes or loss probe",
-                                transmit.segment_size - packet_len_unpadded
+                                transmit.segment_size() - packet_len_unpadded
                             );
                             builder_storage = Some(builder);
                             break;
@@ -656,27 +656,29 @@ impl Connection {
 
                         // Pad the current datagram to GSO segment size so it can be included in the
                         // GSO batch.
-                        builder.pad_to(transmit.segment_size as u16);
+                        builder.pad_to(transmit.segment_size() as u16);
                     }
 
                     builder.finish_and_track(now, self, sent_frames.take(), transmit.buf);
 
-                    if transmit.num_datagrams == 1 && space_id == SpaceId::Data {
-                        // Now that we know the size of the first datagram, check whether
-                        // the data we planned to send will fit in the next segment.  If
-                        // not, bails out and leave it for the next GSO batch.  We can't
-                        // easily compute the right segment size before the original call to
-                        // `space_can_send`, because at that time we haven't determined
-                        // whether we're going to coalesce with the first datagram or
-                        // potentially pad it to `MIN_INITIAL_SIZE`.
-
+                    if transmit.num_datagrams() == 1 {
                         transmit.clip_datagram_size();
+                        if space_id == SpaceId::Data {
+                            // Now that we know the size of the first datagram, check
+                            // whether the data we planned to send will fit in the next
+                            // segment.  If not, bails out and leave it for the next GSO
+                            // batch.  We can't easily compute the right segment size before
+                            // the original call to `space_can_send`, because at that time
+                            // we haven't determined whether we're going to coalesce with
+                            // the first datagram or potentially pad it to
+                            // `MIN_INITIAL_SIZE`.
 
-                        let frame_space_1rtt = transmit
-                            .segment_size
-                            .saturating_sub(self.predict_1rtt_overhead(Some(pn)));
-                        if self.space_can_send(space_id, frame_space_1rtt).is_empty() {
-                            break;
+                            let frame_space_1rtt = transmit
+                                .segment_size()
+                                .saturating_sub(self.predict_1rtt_overhead(Some(pn)));
+                            if self.space_can_send(space_id, frame_space_1rtt).is_empty() {
+                                break;
+                            }
                         }
                     }
                 }
@@ -691,7 +693,7 @@ impl Connection {
                         // unexpectedly.
                         transmit.start_new_datagram_with_size(std::cmp::min(
                             usize::from(INITIAL_MTU),
-                            transmit.segment_size,
+                            transmit.segment_size(),
                         ));
                     }
                 };
@@ -706,7 +708,7 @@ impl Connection {
                 }
             }
 
-            debug_assert!(transmit.buf_capacity - transmit.len() >= MIN_PACKET_SPACE);
+            debug_assert!(transmit.datagram_max_offset() - transmit.len() >= MIN_PACKET_SPACE);
 
             //
             // From here on, we've determined that a packet will definitely be sent.
@@ -809,7 +811,7 @@ impl Connection {
 
             // Send an off-path PATH_RESPONSE. Prioritized over on-path data to ensure that path
             // validation can occur while the link is saturated.
-            if space_id == SpaceId::Data && transmit.num_datagrams == 1 {
+            if space_id == SpaceId::Data && transmit.num_datagrams() == 1 {
                 if let Some((token, remote)) = self.path_responses.pop_off_path(self.path.remote) {
                     // `unwrap` guaranteed to succeed because `builder_storage` was populated just
                     // above.
@@ -857,7 +859,7 @@ impl Connection {
                 !(sent.is_ack_only(&self.streams)
                     && !can_send.acks
                     && can_send.other
-                    && (transmit.buf_capacity - builder.datagram_start)
+                    && (transmit.datagram_max_offset() - builder.datagram_start)
                         == self.path.current_mtu() as usize
                     && self.datagrams.outgoing.is_empty()),
                 "SendableFrames was {can_send:?}, but only ACKs have been written"
@@ -898,10 +900,10 @@ impl Connection {
                 .mtud
                 .poll_transmit(now, self.packet_number_filter.peek(&self.spaces[space_id]))?;
 
-            debug_assert_eq!(transmit.num_datagrams, 0);
+            debug_assert_eq!(transmit.num_datagrams(), 0);
             transmit.start_new_datagram_with_size(probe_size as usize);
 
-            debug_assert_eq!(transmit.datagram_start, 0);
+            debug_assert_eq!(transmit.datagram_start_offset(), 0);
             let mut builder = PacketBuilder::new(
                 now,
                 space_id,
@@ -940,13 +942,13 @@ impl Connection {
         trace!(
             "sending {} bytes in {} datagrams",
             transmit.len(),
-            transmit.num_datagrams
+            transmit.num_datagrams()
         );
         self.path.total_sent = self.path.total_sent.saturating_add(transmit.len() as u64);
 
         self.stats
             .udp_tx
-            .on_sent(transmit.num_datagrams as u64, transmit.len());
+            .on_sent(transmit.num_datagrams() as u64, transmit.len());
 
         Some(Transmit {
             destination: self.path.remote,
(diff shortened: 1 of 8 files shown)


djc:
Initial feedback from skimming:

* I like the direction!
* In order to make it easy to review this, I'd want the change to be split into smaller commits. I'd suggest by isolating all the state (fields) into `TransmitBuf` without doing anything else in the first commit, then add the trait(s) in a separate commit and start moving over methods/adding abstractions.
* The `BufMut + BufOffset` bounds like repetitive, I guess `BufOffset` could have `BufMut` as a supertrait probably?
* (Nit: let's start moving towards using `impl Trait` instead of `T: Trait` in argument position.)

flub:
I've attempted to split this up, hopefully the commits are more or less what you had in mind.  Still makes the same change essentially.

quinn-proto/src/connection/transmit_buf.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    /// The number of bytes written into the buffer so far
+    pub(super) fn len(&self) -> usize {
+        self.buf.len()
+    }
+
+    /// Returns `true` if the buffer did not have anything written into it
+    pub(super) fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
```
djc, line 78: Nit: move this before `len()`, since it depends on it.

quinn-proto/src/connection/mod.rs
```diff
@@ -926,19 +924,12 @@ impl Connection {
                 .mtud
                 .poll_transmit(now, self.packet_number_filter.peek(&self.spaces[space_id]))?;
 
-            let buf_capacity = probe_size as usize;
-            buf.buf.reserve(buf_capacity);
```
djc: Should this be in the previous commit?

quinn-proto/src/connection/mod.rs
```diff
@@ -661,68 +661,42 @@ impl Connection {
...
-                            let frame_space_1rtt = buf
-                                .segment_size
-                                .saturating_sub(self.predict_1rtt_overhead(Some(pn)));
-                            if self.space_can_send(space_id, frame_space_1rtt).is_empty() {
-                                break;
-                            }
+                    if buf.num_datagrams == 1 && space_id == SpaceId::Data {
+                        // Now that we know the size of the first datagram, check whether
+                        // the data we planned to send will fit in the next segment.  If
+                        // not, bails out andleave it for the next GSO batch.  We can't
```
djc, line 667: Nit: "andleave".

quinn-proto/src/connection/transmit_buf.rs
```diff
@@ -74,6 +74,80 @@ impl<'a> TransmitBuf<'a> {
...
+
+    /// Clips the datagram size to the current size
+    ///
+    /// Only valid for the first datagram, when the datagram might be smaller than the
+    /// segment size.  Needed before estimating the available space in the next datagram
+    /// based on [`TransmitBuf::segment_size`].
+    ///
+    /// Use [`TransmitBuf::start_new_datagram_with_size`] if you need to reduce the size of
+    /// the last datagram in a batch.
+    pub(super) fn clip_datagram_size(&mut self) {
```
djc, line 125: Ordering nit: `is_empty()` and `len()` should stay last, `&mut self` before `&self`, callers before callees. So `start_new_datagram_with_size()`, `start_new_datagram()`, `new_datagram_inner`, `clip_datagram_size()`.

quinn-proto/src/connection/transmit_buf.rs
```diff
@@ -74,6 +74,80 @@ impl<'a> TransmitBuf<'a> {
...
+    /// If this starts the second datagram the segment size will be set to the size of the
+    /// first datagram.
+    ///
+    /// If the underlying buffer does not have enough capacity yet this will allocate enough
+    /// capacity for all the datagrams allowed in a single batch.  Use
+    /// [`TransmitBuf::start_new_datagram_with_size`] if you know you will need less.
+    pub(super) fn start_new_datagram(&mut self) {
+        // We reserve the maximum space for sending `max_datagrams` upfront to avoid any
+        // reallocations if more datagrams have to be appended later on.  Benchmarks have
+        // shown a 5-10% throughput improvment compared to continuously resizing the
```
djc, line 141: "improvment"

quinn-proto/src/connection/transmit_buf.rs
```diff
@@ -74,6 +74,80 @@ impl<'a> TransmitBuf<'a> {
...
+        // (e.g. purely containing ACKs), modern memory allocators (e.g. mimalloc and
+        // jemalloc) will pool certain allocation sizes and therefore this is still rather
+        // efficient.
+        let max_capacity_hint = self.max_datagrams * self.segment_size;
+        self.new_datagram_inner(self.segment_size, max_capacity_hint)
+    }
+
+    /// Starts a datagram with a custom datagram size
+    ///
+    /// This is a specialised version of [`TransmitBuf::start_new_datagram`] which sets the
```
djc, line 152: "specialised" -> "specialized"

quinn-proto/src/connection/transmit_buf.rs
```diff
@@ -74,6 +74,80 @@ impl<'a> TransmitBuf<'a> {
...
+    fn new_datagram_inner(&mut self, datagram_size: usize, max_capacity_hint: usize) {
+        debug_assert!(self.num_datagrams < self.max_datagrams);
+        if self.num_datagrams == 1 {
+            // Set the segment size to the size of the first datagram.
+            self.segment_size = self.buf.len();
+        }
+        if self.num_datagrams >= 1 {
+            debug_assert!(datagram_size <= self.segment_size);
+            if datagram_size < self.segment_size {
+                // If This is a GSO batch and this datagram is smaller than the segment
```
djc, line 172: "This" -> "this"

quinn-proto/src/connection/transmit_buf.rs
```diff
@@ -75,6 +75,37 @@ impl<'a> TransmitBuf<'a> {
         self.len() == 0
     }
 
+    /// Returns the GSO segment size
+    ///
+    /// This is also the maximum size datagrams are allowed to be.  The first and last
+    /// datagram in a batch are allowed to be smaller however.  After the first datagram the
+    /// segment size is clipped to the size of the first datagram.
+    pub(super) fn segment_size(&self) -> usize {
```
djc, line 85: I guess? I don't love the LoC overhead this relative to the benefit this provide over an implicit/documentation contract that callers shouldn't be mutating `TransmitBuf`'s fields directly. It is a private API, of course, which we can trivially change over time.

quinn-proto/src/connection/mod.rs
```diff
@@ -3963,6 +3963,23 @@ fn negotiate_max_idle_timeout(x: Option<VarInt>, y: Option<VarInt>) -> Option<Du
     }
 }
 
+/// A buffer that can tell how much has been written to it already
+///
+/// This is commonly used for when a buffer is passed and the user may not write past a
+/// given size.  It allows the user of such a buffer to know the current cursor position in
+/// the buffer.  The maximum write size is usually passed in the same unit as
+/// [`BufLen::len`]: bytes since the buffer start.
+pub(crate) trait BufLen {
```
djc, line 3972: As suggested, can we make `BufLen: BufMut`?

quinn-proto/src/packet.rs
```diff
@@ -281,7 +282,7 @@ pub(crate) enum Header {
 }
 
 impl Header {
-    pub(crate) fn encode(&self, w: &mut Vec<u8>) -> PartialEncode {
+    pub(crate) fn encode<W: BufMut + BufLen>(&self, w: &mut W) -> PartialEncode {
```
djc, line 285: Nit: `impl BufMut + BufLen`.

djc (commented):
LGTM, did a more detailed pass although I have not completely reviewed all the comments and code changes.

Despite your preferences, please stick to single spaces after periods and US English in this project. 👍

quinn-proto/src/connection/transmit_buf.rs
flub: done

quinn-proto/src/connection/mod.rs
flub: sure

quinn-proto/src/connection/mod.rs
flub: done

quinn-proto/src/connection/transmit_buf.rs
flub: done.

For my understanding of your logic: what's the reasoning for the `start_new_datagram_with_size` to be before `start_new_datagram`?  And why does `clip_datagram_size` have an order other than in the `&mut self` group?

Also, what if `is_empty()` would be modified to use `self.buf.len()` instead of `self.len()`, would that have changed the order between `is_empty()` and `len()`?

As an aside, my ordering was not completely random either: constructors, `&self`, `&mut self` and within those blocks ordering was dependent on what is most important for the users (which is subjective).  But happy to follow local conventions.  Maybe it is worth considering writing a CONTRIBUTING.txt though?

quinn-proto/src/connection/transmit_buf.rs
flub: done

quinn-proto/src/connection/transmit_buf.rs
flub: done

quinn-proto/src/connection/transmit_buf.rs
flub: That's essentially a comment on this entire commit.  I do feel fairly strongly that the encapsulation matters, in Rust this means you have to write accessors and that's fine for me.  I don't think a few more lines of code are any overhead, it is even kind of nice you get to write a clear doc comment for it.

The encapsulation is a large part of this PR: the intricate logic of how all these fields relate to each other the invariants that need to be upheld are now entirely encoded in the 15 lines of `new_datagram_inner`.  And forcing non-mut access to the fields makes it easy to audit this.

quinn-proto/src/connection/transmit_buf.rs
djc: Good questions!

> For my understanding of your logic: what's the reasoning for the `start_new_datagram_with_size` to be before `start_new_datagram`?

Yeah, I misremembered and might go the other way on that based on `start_new_datagram_with_size()` is slightly higher-complexity than `start_new_datagram()`.

> And why does `clip_datagram_size` have an order other than in the `&mut self` group?

It should be in the `&mut self` group, but at the bottom (high complexity to low complexity).

> Also, what if `is_empty()` would be modified to use `self.buf.len()` instead of `self.len()`, would that have changed the order between `is_empty()` and `len()`?

No, because it'd still be (slightly) higher complexity.

> As an aside, my ordering was not completely random either: constructors, &self, &mut self and within those blocks ordering was dependent on what is most important for the users (which is subjective). But happy to follow local conventions. Maybe it is worth considering writing a CONTRIBUTING.txt though?

So we do constructors, `&mut self`, `&self` and within those mostly callers before callees and higher-complexity to lower-complexity.

There's a [style guide](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#style-guide) in rustls that I wrote that's mostly the same stuff.

quinn-proto/src/connection/mod.rs
flub: I chose not to do so because I don't think `BufLen` is that nice a trait.  It is more of a plug to bridge the use of `TransmitBuf` and the existing concept of passing around a buffer with a bunch of offsets describing it.  I hope that later it can be removed again, but also added this to limit the scope of this refactor.

And finally I think the naming would be much less clear, right now places get `BufMut + BufLen` and you immediately have very clear expectations.  If it were to be a supertrait it'd be something like `WriteBuffer`, `BufMutWithLen` or I don't know.  Naming it would become much trickier.

I do appreciate that this could become unwieldy as more traits are added.  Though all the uses for now are just two traits, which I don't think is problematic yet.

quinn-proto/src/packet.rs
flub: done

gretchenfrage:
Thank you for working on this! I think this should be considered to close #2121

quinn-proto/src/connection/mod.rs
```diff
@@ -643,97 +640,73 @@ impl Connection {
...
-                        // first datagram or potentially pad it to `MIN_INITIAL_SIZE`.
-                        if space_id == SpaceId::Data {
-                            let frame_space_1rtt =
-                                segment_size.saturating_sub(self.predict_1rtt_overhead(Some(pn)));
-                            if self.space_can_send(space_id, frame_space_1rtt).is_empty() {
-                                break;
-                            }
+                    builder.finish_and_track(now, self, sent_frames.take(), &mut buf);
+
+                    if buf.num_datagrams() == 1 && space_id == SpaceId::Data {
```
flub, line 664: This if condition shouldn't have been collapsed.  The buffer needs to be clipped in all packet spaces.  The extra size check is only needed in the data packet space.

quinn-proto/src/connection/transmit_buf.rs
```diff
@@ -0,0 +1,215 @@
...
+    ///
+    /// Only valid for the first datagram, when the datagram might be smaller than the
+    /// segment size. Needed before estimating the available space in the next datagram
+    /// based on [`TransmitBuf::segment_size`].
+    ///
+    /// Use [`TransmitBuf::start_new_datagram_with_size`] if you need to reduce the size of
+    /// the last datagram in a batch.
+    pub(super) fn clip_datagram_size(&mut self) {
+        debug_assert_eq!(self.num_datagrams, 1);
+        self.segment_size = self.buf.len();
```
flub, line 141: Probably also want to set `self.buf_capacity` to this size here.

gretchenfrage:
Thanks for your patience--I have ambitions of trying to take a closer look at this next weekend.

gretchenfrage:
Yes, I think this is a good direction! Some thoughts so far...

#### Naming

First a minor thing: Perhaps this should be called `TransmitBuilder` rather than `TransmitBuf`? That way it matches `PacketBuilder`, and I feel that more accurately represents that that is tracking more complex state than just a byte sequence.

#### `impl BufMut`

With regard to this:

```rs
unsafe impl BufMut for TransmitBuf<'_> {
    fn remaining_mut(&self) -> usize {
        self.buf.remaining_mut()
    }

    unsafe fn advance_mut(&mut self, cnt: usize) {
        self.buf.advance_mut(cnt);
    }

    fn chunk_mut(&mut self) -> &mut bytes::buf::UninitSlice {
        self.buf.chunk_mut()
    }
}
```

I'm not sure that I like that. It adds an `unsafe` statement, as well as a handful of LOC, for the sake of being able to write `buf` instead of `buf.buf` in perhaps a dozen places. Plus, it's a slightly surprising thing to keep in mind.

#### `trait BufLen`

You're introducing this new trait for the sake of being able to pass around `&mut (impl BufMut + BufLen)` instead of `&mut TransmitBuf`. I understand that your motivation for this is to abstract things, but in this circumstance I think that directly passing around `&mut TransmitBuf` or `&mut Vec<u8>` to various functions would be a net-win for it being simple to think about.

This is related to my comment about `impl BufMut`, as creating a version of the "Do not pass buffer around as a Vec" commit which avoids relying on this trait would be easier if we're also not trying to make our code leverage an `impl BufMut for TransmitBuf`.

#### Getter methods

I agree with djc [here](https://github.com/quinn-rs/quinn/pull/2168#discussion_r1995367137)--I'm pretty hesitant about having these getter methods on `TransmitBuf`: `segment_size`, `num_datagrams`, `max_datagrams`, `datagram_start_offset`, and `datagram_max_offset`.

These variables aren't really getting abstracted away if the users of `TransmitBuf` still have to know which of them exist and care about their specific values. That leaves the only remaining motivation of using getters like this as making sure we don't accidentally mutate them wrongly from outside the module. But it's pretty easy to check where we're writing to these variables and I don't think making these variables read-only from outside this module really gets at the bottleneck of us being able to make the code correct. And it has to be weighed against the cons of increased LOC, and us remembering which ones of these functions are simply reading a field versus which ones are doing something more transformative.

I was experimenting with an alternative way of trying to abstract some of these things, and I noticed that the `num_datagrams` value is used in similar ways in a few places repeatedly. So I tried to see if I could factor some of them away into more transformative methods on `TransmitBuf` until the `num_datagrams` value no longer needed to be accessed outside of `TransmitBuf` at all and I came up with this:

```diff
diff --git a/quinn-proto/src/connection/mod.rs b/quinn-proto/src/connection/mod.rs
index 3a39a433..62c42e73 100644
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -562,7 +562,7 @@ impl Connection {
                 // We need to send 1 more datagram and extend the buffer for that.
 
                 // Is 1 more datagram allowed?
-                if buf.num_datagrams() >= buf.max_datagrams() {
+                if !buf.more_datagrams_allowed() {
                     // No more datagrams allowed
                     break;
                 }
@@ -573,9 +573,7 @@ impl Connection {
                 // for starting another datagram. If there is any anti-amplification
                 // budget left, we always allow a full MTU to be sent
                 // (see https://github.com/quinn-rs/quinn/issues/1082)
-                if self.path.anti_amplification_blocked(
-                    (buf.segment_size() * buf.num_datagrams()) as u64 + 1,
-                ) {
+                if self.path.anti_amplification_blocked(buf.bytes_to_send()) {
                     trace!("blocked by anti-amplification");
                     break;
                 }
@@ -625,7 +623,7 @@ impl Connection {
                         builder.pad_to(MIN_INITIAL_SIZE);
                     }
 
-                    if buf.num_datagrams() > 1 {
+                    if !buf.first_datagram() {
                         // If too many padding bytes would be required to continue the GSO batch
                         // after this packet, end the GSO batch here. Ensures that fixed-size frames
                         // with heterogeneous sizes (e.g. application datagrams) won't inadvertently
@@ -661,7 +659,7 @@ impl Connection {
 
                     builder.finish_and_track(now, self, sent_frames.take(), &mut buf);
 
-                    if buf.num_datagrams() == 1 {
+                    if buf.first_datagram() {
                         buf.clip_datagram_size();
                         if space_id == SpaceId::Data {
                             // Now that we know the size of the first datagram, check
@@ -811,7 +809,7 @@ impl Connection {
 
             // Send an off-path PATH_RESPONSE. Prioritized over on-path data to ensure that path
             // validation can occur while the link is saturated.
-            if space_id == SpaceId::Data && buf.num_datagrams() == 1 {
+            if space_id == SpaceId::Data && buf.first_datagram() {
                 if let Some((token, remote)) = self.path_responses.pop_off_path(self.path.remote) {
                     // `unwrap` guaranteed to succeed because `builder_storage` was populated just
                     // above.
@@ -900,7 +898,6 @@ impl Connection {
                 .mtud
                 .poll_transmit(now, self.packet_number_filter.peek(&self.spaces[space_id]))?;
 
-            debug_assert_eq!(buf.num_datagrams(), 0);
             buf.start_new_datagram_with_size(probe_size as usize);
 
             debug_assert_eq!(buf.datagram_start_offset(), 0);
@@ -933,16 +930,7 @@ impl Connection {
             return None;
         }
 
-        trace!(
-            "sending {} bytes in {} datagrams",
-            buf.len(),
-            buf.num_datagrams()
-        );
-        self.path.total_sent = self.path.total_sent.saturating_add(buf.len() as u64);
-
-        self.stats
-            .udp_tx
-            .on_sent(buf.num_datagrams() as u64, buf.len());
+        buf.update_stats(&mut self.path.total_sent, &mut self.stats.udp_tx);
 
         Some(Transmit {
             destination: self.path.remote,
@@ -952,9 +940,10 @@ impl Connection {
             } else {
                 None
             },
-            segment_size: match buf.num_datagrams() {
-                1 => None,
-                _ => Some(buf.segment_size()),
+            segment_size: if buf.first_datagram() {
+                None
+            } else {
+                Some(buf.segment_size())
             },
             src_ip: self.local_ip,
         })
diff --git a/quinn-proto/src/connection/transmit_buf.rs b/quinn-proto/src/connection/transmit_buf.rs
index 61d3217c..71383945 100644
--- a/quinn-proto/src/connection/transmit_buf.rs
+++ b/quinn-proto/src/connection/transmit_buf.rs
@@ -1,4 +1,7 @@
 use bytes::BufMut;
+use tracing::trace;
+
+use crate::{FrameStats, UdpStats};
 
 use super::BufLen;
 
@@ -151,16 +154,27 @@ impl<'a> TransmitBuf<'a> {
         self.segment_size
     }
 
-    /// Returns the number of datagrams written into the buffer
-    ///
-    /// The last datagram is not necessarily finished yet.
-    pub(super) fn num_datagrams(&self) -> usize {
-        self.num_datagrams
+    pub(super) fn more_datagrams_allowed(&self) -> bool {
+        self.num_datagrams < self.max_datagrams
+    }
+
+    pub(super) fn bytes_to_send(&self) -> u64 {
+        (self.segment_size * self.num_datagrams) as u64 + 1
     }
 
-    /// Returns the maximum number of datagrams allowed to be written into the buffer
-    pub(super) fn max_datagrams(&self) -> usize {
-        self.max_datagrams
+    pub(super) fn first_datagram(&self) -> bool {
+        self.num_datagrams == 1
+    }
+
+    pub(super) fn update_stats(&self, path_total_sent: &mut u64, udp_tx: &mut UdpStats) {
+        trace!(
+            "sending {} bytes in {} datagrams",
+            self.buf.len(),
+            self.num_datagrams,
+        );
+        *path_total_sent = path_total_sent.saturating_add(self.buf.len() as u64);
+
+        udp_tx.on_sent(self.num_datagrams as u64, self.buf.len());
     }
 
     /// Returns the start offset of the current datagram in the buffer
```
</details>

Bear in mind that there might be some flaws in that still at this point, but I think my intent should be clear. No need to incorporate this actual concept into this PR, but I think this approach to abstraction has certain advantages--it more successfully offloads conceptual complexity from the overly large `poll_transmit` method by extracting inherently meaningful sub-units of its logic into methods `more_datagrams_allowed`, `bytes_to_send`, `first_datagram` and `update_stats`--and at the end of this, the number of usages of `num_datagrams` and `max_datagrams` outside of `transmit_buf.rs` has dropped to zero and so we can just make them private / delete the getters as a reward.

Similarly, I'm inclined to leaving `TransmitBuf.buf` as `pub(super)` and thus removing `is_empty`, `len`, and `as_mut_slice`.

gretchenfrage (changes requested):
Yes, I think this is a good direction! Some thoughts so far...

#### Naming

First a minor thing: Perhaps this should be called `TransmitBuilder` rather than `TransmitBuf`? That way it matches `PacketBuilder`, and I feel that more accurately represents that that is tracking more complex state than just a byte sequence.

#### `impl BufMut`

With regard to this:

```rs
unsafe impl BufMut for TransmitBuf<'_> {
    fn remaining_mut(&self) -> usize {
        self.buf.remaining_mut()
    }

    unsafe fn advance_mut(&mut self, cnt: usize) {
        self.buf.advance_mut(cnt);
    }

    fn chunk_mut(&mut self) -> &mut bytes::buf::UninitSlice {
        self.buf.chunk_mut()
    }
}
```

I'm not sure that I like that. It adds an `unsafe` statement, as well as a handful of LOC, for the sake of being able to write `buf` instead of `buf.buf` in perhaps a dozen places. Plus, it's a slightly surprising thing to keep in mind.

#### `trait BufLen`

You're introducing this new trait for the sake of being able to pass around `&mut (impl BufMut + BufLen)` instead of `&mut TransmitBuf`. I understand that your motivation for this is to abstract things, but in this circumstance I think that directly passing around `&mut TransmitBuf` or `&mut Vec<u8>` to various functions would be a net-win for it being simple to think about.

This is related to my comment about `impl BufMut`, as creating a version of the "Do not pass buffer around as a Vec" commit which avoids relying on this trait would be easier if we're also not trying to make our code leverage an `impl BufMut for TransmitBuf`.

#### Getter methods

I agree with djc [here](https://github.com/quinn-rs/quinn/pull/2168#discussion_r1995367137)--I'm pretty hesitant about having these getter methods on `TransmitBuf`: `segment_size`, `num_datagrams`, `max_datagrams`, `datagram_start_offset`, and `datagram_max_offset`.

These variables aren't really getting abstracted away if the users of `TransmitBuf` still have to know which of them exist and care about their specific values. That leaves the only remaining motivation of using getters like this as making sure we don't accidentally mutate them wrongly from outside the module. But it's pretty easy to check where we're writing to these variables and I don't think making these variables read-only from outside this module really gets at the bottleneck of us being able to make the code correct. And it has to be weighed against the cons of increased LOC, and us remembering which ones of these functions are simply reading a field versus which ones are doing something more transformative.

I was experimenting with an alternative way of trying to abstract some of these things, and I noticed that the `num_datagrams` value is used in similar ways in a few places repeatedly. So I tried to see if I could factor some of them away into more transformative methods on `TransmitBuf` until the `num_datagrams` value no longer needed to be accessed outside of `TransmitBuf` at all and I came up with this:

```diff
diff --git a/quinn-proto/src/connection/mod.rs b/quinn-proto/src/connection/mod.rs
index 3a39a433..62c42e73 100644
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -562,7 +562,7 @@ impl Connection {
                 // We need to send 1 more datagram and extend the buffer for that.
 
                 // Is 1 more datagram allowed?
-                if buf.num_datagrams() >= buf.max_datagrams() {
+                if !buf.more_datagrams_allowed() {
                     // No more datagrams allowed
                     break;
                 }
@@ -573,9 +573,7 @@ impl Connection {
                 // for starting another datagram. If there is any anti-amplification
                 // budget left, we always allow a full MTU to be sent
                 // (see https://github.com/quinn-rs/quinn/issues/1082)
-                if self.path.anti_amplification_blocked(
-                    (buf.segment_size() * buf.num_datagrams()) as u64 + 1,
-                ) {
+                if self.path.anti_amplification_blocked(buf.bytes_to_send()) {
                     trace!("blocked by anti-amplification");
                     break;
                 }
@@ -625,7 +623,7 @@ impl Connection {
                         builder.pad_to(MIN_INITIAL_SIZE);
                     }
 
-                    if buf.num_datagrams() > 1 {
+                    if !buf.first_datagram() {
                         // If too many padding bytes would be required to continue the GSO batch
                         // after this packet, end the GSO batch here. Ensures that fixed-size frames
                         // with heterogeneous sizes (e.g. application datagrams) won't inadvertently
@@ -661,7 +659,7 @@ impl Connection {
 
                     builder.finish_and_track(now, self, sent_frames.take(), &mut buf);
 
-                    if buf.num_datagrams() == 1 {
+                    if buf.first_datagram() {
                         buf.clip_datagram_size();
                         if space_id == SpaceId::Data {
                             // Now that we know the size of the first datagram, check
@@ -811,7 +809,7 @@ impl Connection {
 
             // Send an off-path PATH_RESPONSE. Prioritized over on-path data to ensure that path
             // validation can occur while the link is saturated.
-            if space_id == SpaceId::Data && buf.num_datagrams() == 1 {
+            if space_id == SpaceId::Data && buf.first_datagram() {
                 if let Some((token, remote)) = self.path_responses.pop_off_path(self.path.remote) {
                     // `unwrap` guaranteed to succeed because `builder_storage` was populated just
                     // above.
@@ -900,7 +898,6 @@ impl Connection {
                 .mtud
                 .poll_transmit(now, self.packet_number_filter.peek(&self.spaces[space_id]))?;
 
-            debug_assert_eq!(buf.num_datagrams(), 0);
             buf.start_new_datagram_with_size(probe_size as usize);
 
             debug_assert_eq!(buf.datagram_start_offset(), 0);
@@ -933,16 +930,7 @@ impl Connection {
             return None;
         }
 
-        trace!(
-            "sending {} bytes in {} datagrams",
-            buf.len(),
-            buf.num_datagrams()
-        );
-        self.path.total_sent = self.path.total_sent.saturating_add(buf.len() as u64);
-
-        self.stats
-            .udp_tx
-            .on_sent(buf.num_datagrams() as u64, buf.len());
+        buf.update_stats(&mut self.path.total_sent, &mut self.stats.udp_tx);
 
         Some(Transmit {
             destination: self.path.remote,
@@ -952,9 +940,10 @@ impl Connection {
             } else {
                 None
             },
-            segment_size: match buf.num_datagrams() {
-                1 => None,
-                _ => Some(buf.segment_size()),
+            segment_size: if buf.first_datagram() {
+                None
+            } else {
+                Some(buf.segment_size())
             },
             src_ip: self.local_ip,
         })
diff --git a/quinn-proto/src/connection/transmit_buf.rs b/quinn-proto/src/connection/transmit_buf.rs
index 61d3217c..71383945 100644
--- a/quinn-proto/src/connection/transmit_buf.rs
+++ b/quinn-proto/src/connection/transmit_buf.rs
@@ -1,4 +1,7 @@
 use bytes::BufMut;
+use tracing::trace;
+
+use crate::{FrameStats, UdpStats};
 
 use super::BufLen;
 
@@ -151,16 +154,27 @@ impl<'a> TransmitBuf<'a> {
         self.segment_size
     }
 
-    /// Returns the number of datagrams written into the buffer
-    ///
-    /// The last datagram is not necessarily finished yet.
-    pub(super) fn num_datagrams(&self) -> usize {
-        self.num_datagrams
+    pub(super) fn more_datagrams_allowed(&self) -> bool {
+        self.num_datagrams < self.max_datagrams
+    }
+
+    pub(super) fn bytes_to_send(&self) -> u64 {
+        (self.segment_size * self.num_datagrams) as u64 + 1
     }
 
-    /// Returns the maximum number of datagrams allowed to be written into the buffer
-    pub(super) fn max_datagrams(&self) -> usize {
-        self.max_datagrams
+    pub(super) fn first_datagram(&self) -> bool {
+        self.num_datagrams == 1
+    }
+
+    pub(super) fn update_stats(&self, path_total_sent: &mut u64, udp_tx: &mut UdpStats) {
+        trace!(
+            "sending {} bytes in {} datagrams",
+            self.buf.len(),
+            self.num_datagrams,
+        );
+        *path_total_sent = path_total_sent.saturating_add(self.buf.len() as u64);
+
+        udp_tx.on_sent(self.num_datagrams as u64, self.buf.len());
     }
 
     /// Returns the start offset of the current datagram in the buffer
```
</details>

Bear in mind that there might be some flaws in that still at this point, but I think my intent should be clear. No need to incorporate this actual concept into this PR, but I think this approach to abstraction has certain advantages--it more successfully offloads conceptual complexity from the overly large `poll_transmit` method by extracting inherently meaningful sub-units of its logic into methods `more_datagrams_allowed`, `bytes_to_send`, `first_datagram` and `update_stats`--and at the end of this, the number of usages of `num_datagrams` and `max_datagrams` outside of `transmit_buf.rs` has dropped to zero and so we can just make them private / delete the getters as a reward.

Similarly, I'm inclined to leaving `TransmitBuf.buf` as `pub(super)` and thus removing `is_empty`, `len`, and `as_mut_slice`.

flub:
> Yes, I think this is a good direction! Some thoughts so far...
> #### Naming
> 
> First a minor thing: Perhaps this should be called `TransmitBuilder` rather than `TransmitBuf`? That way it matches `PacketBuilder`, and I feel that more accurately represents that that is tracking more complex state than just a byte sequence.

I don't have very strong feelings about the naming, this was really just a working name I thought would change but then just stuck as it did end up making sense.  I think both work, but indeed `TransmitBuilder` would match `PacketBuilder` (which retains it's name in #2195).  The current naming does match the fact it implements `BufMut` directly.  I guess the variable name would also have to change if this was renamed, probably to `transmit` or so.

> #### `impl BufMut`
> 
> With regard to this:
> 
> ```rust
> unsafe impl BufMut for TransmitBuf<'_>
> ```
> 
> I'm not sure that I like that. It adds an `unsafe` statement, as well as a handful of LOC, for the sake of being able to write `buf` instead of `buf.buf` in perhaps a dozen places. Plus, it's a slightly surprising thing to keep in mind.

What would you think of:
```rust
impl TransmitBuf<'_> {
    /// Returns a buffer into which the datagram can be written
    pub(super) fn datagram_mut(&mut self) -> bytes::buf::Limit<&mut Vec<u8>> {
        self.buf.limit(self.buf_capacity)
    }
}
```

This is very similar to what I did in #2195 which has a `PacketBuilder::frame_space_mut` method which I do rather like how it reads.  And it gives protection against writing too many bytes into the buffer.  Doing this would indeed alight well with renaming `TransmitBuf` to `TransmitBuilder`.

(I'd like to return return a generic `&mut impl BufMut` there, but that doesn't currently work without `use`-lifetime annotations which are too recent a rust feature.  Also that doesn't let me implement `BufLen`.)

> #### `trait BufLen`
> 
> You're introducing this new trait for the sake of being able to pass around `&mut (impl BufMut + BufLen)` instead of `&mut TransmitBuf`. I understand that your motivation for this is to abstract things, but in this circumstance I think that directly passing around `&mut TransmitBuf` or `&mut Vec<u8>` to various functions would be a net-win for it being simple to think about.

I did this trait for two reasons:
- A number of APIs expect this "I get large buffer with a max write offset (and sometimes need a start offset)" notion, which is what I'm trying to move away from.  But to keep the scope of this PR smaller I can't change all of that at once, that would get out of hand.  So this `BufLen` is a stopgap measure.
- I assumed that the functions which took traits rather than the `&Vec<u8>` wanted to keep doing this.  There is some logic to this, because a number of these functions really don't have any business being able to change the state of the datagrams.  They only need to write some stuff.

So I think if you receive an explicit `&TransmitBuf` (or whatever it is called) you need to have some business with datagrams.  Only writing some frames is not that.

I think it might be possible to remove `BufLen` entirely and instead switch to `BufMut::remaining_mut` for these functions.  But:
- I'm not really statisfied from the `BufMut::remaining_mut` docs that this can be reliably used for this.  I think the implementations we use, and certainly `Limit`, do.  But it still leaves me a bit worried.
- It would switch around a lot more logic.  Using this `BufLen` keeps the changes to places where this stop-gap measure is needed much simpler and easier to review for correctness.

After #2195 there is only one `BufLen` remaining.  I think it's very reasonable to also remove that and I'd be happy to look at doing that in another PR.  Though I'd probably wait to figure out a nice way here until that has all been merged so I know what the final shape of things is and have more of a feel for what everyone agrees on for abstraction style.


> #### Getter methods
> 
> I agree with djc [here](https://github.com/quinn-rs/quinn/pull/2168#discussion_r1995367137)--I'm pretty hesitant about having these getter methods on `TransmitBuf`: `segment_size`, `num_datagrams`, `max_datagrams`, `datagram_start_offset`, and `datagram_max_offset`.
> 
> These variables aren't really getting abstracted away if the users of `TransmitBuf` still have to know which of them exist and care about their specific values. That leaves the only remaining motivation of using getters like this as making sure we don't accidentally mutate them wrongly from outside the module. But it's pretty easy to check where we're writing to these variables and I don't think making these variables read-only from outside this module really gets at the bottleneck of us being able to make the code correct.

This last sentence I am really not that sure about.  Abstractions the compiler provides are great at ensuring correctness, having to manually check on each PR if something is being mutated that shouldn't be is very fallible.  Having explicit contracts like this will make it clear if that is being changed.

> And it has to be weighed against the cons of increased LOC, and us remembering which ones of these functions are simply reading a field versus which ones are doing something more transformative.
> 
> I was experimenting with an alternative way of trying to abstract some of these things, and I noticed that the `num_datagrams` value is used in similar ways in a few places repeatedly. So I tried to see if I could factor some of them away into more transformative methods on `TransmitBuf` until the `num_datagrams` value no longer needed to be accessed outside of `TransmitBuf` at all and I came up with this:

Yes, I agree there may be more scope here.  And I was unsure whether to go further or not.  For the sake of being more incremental with changes I stopped here for this PR though.

For example I think the `bytes_to_send` is probably a good one to add.  But others I expect to be easier to remove in the future entirely.  E.g. `datagram_max_offset` is very tempting to remove after #2195 where it is only 3 uses remain.

So, from what I understand maybe next steps for this PR could be:
- Rename `TransmitBuf` to `TransmitBuilder`
- Add `TranmitBuiilder::datagram_mut` function
- See which helpers might easily disappear with this, e.g. `.len()`, `.is_empty()` etc.

If you think that's helpful I'll give those changes a go.

flub:
@gretchenfrage I've pushed some new commits.  Mainly this finally introduces a buffer abstraction that I've long wanted but resisted creating: the `BufSlice` (I'm not attached to this name, it could be better).  To get access to the datagram being written into you need `TransmitBuilder::datagram_mut()` (or `TransmitBuilder::datagram()`, but in return you get a nice access to the buffer.

This means the `PacketBuilder` now needs to keep track of sizes relative to the datagram buffer, a change I've also long wanted.  In turn this means some more of the questionable methods on `TransmitBuilder` can be removed, which is probably a good gain.

The main reason I was trying to avoid these changes so far was to try and keep the changes smaller.  The abstractions are a fair number of lines of code again, but have a straight forward interface.

flub:
FWIW, `TransmitBuilder::datagram_mut()` is similar to the approach I already took in #2195 where I used `PacketBuilder::frame_space_mut()` to access the buffer.  Give that model also works well there it gives me confidence this is a good approach.

It does leave a manual `BufMut` impl around, which I agree is unfortunate due to the `unsafe`.  Though as before it mostly forwards calls to the the underlying buffer implementation.  So I don't think this is so bad to have, and Quinn does not seem to have a history of avoiding unsafe.

flub:
Also note that `BufSlice` can probably completely remove the need for the `BufLen` trait from this PR.  I'll do that once the `BufSlice`/`DatagramBuffer` idea has been accepted and possibly reshaped to reduce churn.

gretchenfrage:
> I'm not really statisfied from the BufMut::remaining_mut docs that this can be reliably used for this. I think the implementations we use, and certainly Limit, do. But it still leaves me a bit worried.

Can you expand on why not? From looking at it, it actually seems like it might be fine.

[`BufMut::remaining_mut`](https://docs.rs/bytes/latest/bytes/buf/trait.BufMut.html#tymethod.remaining_mut):

> Returns the number of bytes that can be written from the current position until the end of the buffer is reached.
>
> This value is greater than or equal to the length of the slice returned by chunk_mut().
>
> #### Implementer notes
> 
> Implementations of remaining_mut should ensure that the return value does not change unless a call is made to advance_mut or any other function that is documented to change the BufMut’s current position.
>
> #### Note
>
> remaining_mut may return value smaller than actual available space.

Notable source code bits exemplifying the expected usage of this:

- [`BufMut::put_slice`](https://docs.rs/bytes/latest/src/bytes/buf/buf_mut.rs.html#246-264) panics if `remaining_mut() < slice.len()`
- [`<Vec<u8> as BufMut>::remaining_mut`](https://docs.rs/bytes/latest/src/bytes/buf/buf_mut.rs.html#1599-1667) returns `isize::MAX - vec.len()`

The comment "This value is greater than or equal to the length of the slice returned by chunk_mut()" clarifies that it's not merely meant as the remaining length of the next chunk, and the `put_slice` implementation establishes that it's expected to indicate how many more bytes can be written. The Vec<u8> implementation seems to further reinforce this understanding. Given this, the comment of "remaining_mut may return value smaller than actual available space" seems to be a pretty low-information note that it can be lower than, say, the extent of the underlying heap allocation or equivalent.

If `remaining_mut` can be used instead of introducing a new `BufLen` trait, that sounds quite preferable.

flub:
> > I'm not really statisfied from the BufMut::remaining_mut docs that this can be reliably used for this. I think the implementations we use, and certainly Limit, do. But it still leaves me a bit worried.
> 
> Can you expand on why not? From looking at it, it actually seems like it might be fine.
> 
> [`BufMut::remaining_mut`](https://docs.rs/bytes/latest/bytes/buf/trait.BufMut.html#tymethod.remaining_mut):
> 
> > Returns the number of bytes that can be written from the current position until the end of the buffer is reached.
> > This value is greater than or equal to the length of the slice returned by chunk_mut().
> > #### Implementer notes
> > Implementations of remaining_mut should ensure that the return value does not change unless a call is made to advance_mut or any other function that is documented to change the BufMut’s current position.
> > #### Note
> > remaining_mut may return value smaller than actual available space.
> 
> Notable source code bits exemplifying the expected usage of this:
> 
>     * [`BufMut::put_slice`](https://docs.rs/bytes/latest/src/bytes/buf/buf_mut.rs.html#246-264) panics if `remaining_mut() < slice.len()`
> 
>     * [`<Vec<u8> as BufMut>::remaining_mut`](https://docs.rs/bytes/latest/src/bytes/buf/buf_mut.rs.html#1599-1667) returns `isize::MAX - vec.len()`
> 
> 
> The comment "This value is greater than or equal to the length of the slice returned by chunk_mut()" clarifies that it's not merely meant as the remaining length of the next chunk, and the `put_slice` implementation establishes that it's expected to indicate how many more bytes can be written. The Vec implementation seems to further reinforce this understanding. Given this, the comment of "remaining_mut may return value smaller than actual available space" seems to be a pretty low-information note that it can be lower than, say, the extent of the underlying heap allocation or equivalent.
> 
> If `remaining_mut` can be used instead of introducing a new `BufLen` trait, that sounds quite preferable.

My worry is with the `Implementer notes`: it says `remaining_mut` is allowed to change whenever `advance_mut` is called.  But it does not say that it must strictly decrease with the amount that was advanced with.  So I imagine it would be legal to re-allocate inside `advance_mut` so that `remaining_mut` is now larger then before the call to `advance_mut`.  I'm not sure if this was intentional or not though.

What do you think about the `BufSlice`/`DatagramBuffer` abstraction though?  That is another way of avoiding this, and it's kind of an abstraction that fits nicely to write fixed-sized datagrams (and maybe packets).  Though I was worried about the amount of code added for it, now that it's there I do rather like it.

flub:
> What do you think about the `BufSlice`/`DatagramBuffer` abstraction though? That is another way of avoiding this, and it's kind of an abstraction that fits nicely to write fixed-sized datagrams (and maybe packets). Though I was worried about the amount of code added for it, now that it's there I do rather like it.

To clarify this: while `BufLen` still exists in this PR, in #2195 `Header::encode` is the only place that still has `BufLen` in it.  Since the `BufSlice` trait here removes the need for `BufLen` in `Header::encode` `BufLen` will be gone entirely once #2195 is updated to be based on this.

quinn-proto/src/packet.rs
```diff
@@ -281,7 +282,11 @@ pub(crate) enum Header {
 }
 
 impl Header {
-    pub(crate) fn encode(&self, w: &mut Vec<u8>) -> PartialEncode {
+    /// Encodes the QUIC packet header into the buffer
+    ///
+    /// The current position of the buffer is stored in the [`PartialEncode`] as the start
+    /// of the packet in the buffer.
+    pub(crate) fn encode(&self, w: &mut impl BufSlice) -> PartialEncode {
```
flub, line 289: This is really the only reason the `BufSlice` trait exists for now.  It is of course possible to make this `w: &mut DatagramBuffer`, the cost is a bunch of callers on the receive path would have to be adapted to construct a `DatagramBuffer` where they currently pass a `&mut Vec<u8>`.

There probably is an argument that this would be better anyway, rather than letting those callers create arbitrarily large datagrams.

So I created `BufSlice` for two reasons:

- Avoiding updating the callers to `Header::encode` on the receive path.
- Maybe it's a useful trait later, e.g. I could imagine a `PacketBuffer`, which would probably be not much more than a newtype around `DatagramBuffer`.

The 2nd reason is of course rather weak, if that's helpful at some point you can always create a trait at that point.

gretchenfrage:
> To clarify this: while BufLen still exists in this PR, in https://github.com/quinn-rs/quinn/pull/2195 Header::encode is the only place that still has BufLen in it. Since the BufSlice trait here removes the need for BufLen in Header::encode BufLen will be gone entirely once https://github.com/quinn-rs/quinn/pull/2195 is updated to be based on this.

One thing to keep in mind is that I think to the extent practical we want to avoid approving something based on the assumption that future work will be approved in any very specific form. This sort of plan is highly likely to be transformed through peer review. Not to mention that you could get busy / etc.

gretchenfrage:
> My worry is with the Implementer notes: it says remaining_mut is allowed to change whenever advance_mut is called. But it does not say that it must strictly decrease with the amount that was advanced with. So I imagine it would be legal to re-allocate inside advance_mut so that remaining_mut is now larger then before the call to advance_mut. I'm not sure if this was intentional or not though.

I feel pretty good at this point about just considering that to be less-than-perfect documentation of the `bytes` crate and assuming that `advance_mut` is supposed to reduce `remaining_mut` by exactly `cnt`. Especially since it's something that's only being used internally.

gretchenfrage:
> What do you think about the BufSlice/DatagramBuffer abstraction though? That is another way of avoiding this, and it's kind of an abstraction that fits nicely to write fixed-sized datagrams (and maybe packets). Though I was worried about the amount of code added for it, now that it's there I do rather like it.

I quite like the `DatagramBuffer` struct! It seems like a good new and simple abstraction that aligns well with what a lot of this code is expecting. "Growable vector of bytes, with a maximum length, that allocation-wise is actually a suffix of a `Vec<u8>` that may already have stuff in it." Moreover, if it becomes useful for more things than just datagrams, it could easily be renamed to something more generic.

I don't think it warrants all these traits though, and I think the traits are over-complicated. AFAICT this is the situation that's been created:

```mermaid
flowchart TD
    DatagramBuffer --> BufMut
    DatagramBuffer --> BufSlice
    DatagramBuffer --> BufLen
    V["Vec&lt;u8>"] --> BufMut
    V["Vec&lt;u8>"] --> BufSlice
    V["Vec&lt;u8>"] --> BufLen
    BufSlice --> Deref
    BufSlice --> DerefMut
  ```

And, from checking out bce96531fa640ab7153667e6b84eaefa08081478, I can just remove the `BufSlice: Deref + DerefMut` requirement and it compiles fine:

```diff
diff --git a/quinn-proto/src/connection/transmit_builder.rs b/quinn-proto/src/connection/transmit_builder.rs
index dac79621..f79da0ef 100644
--- a/quinn-proto/src/connection/transmit_builder.rs
+++ b/quinn-proto/src/connection/transmit_builder.rs
@@ -203,7 +203,7 @@ impl<'a> TransmitBuilder<'a> {
 ///
 /// The [`Deref`] impl must return the already written bytes as a slice.  [`DerefMut`] must
 /// allow modification of these already written bytes.
-pub(crate) trait BufSlice: BufMut + Deref<Target = [u8]> + DerefMut {
+pub(crate) trait BufSlice: BufMut {
     /// Returns the number of already written bytes in the buffer
     fn len(&self) -> usize;
 ```

Also the `DerefMut` implementation overall seems overkill since it's only used in 2 places for minor things, it seems a fair bit simpler to just give it a `fn written(&mut self) -> &mut [u8]`:

```diff
diff --git a/quinn-proto/src/connection/packet_builder.rs b/quinn-proto/src/connection/packet_builder.rs
index 3052efb2..b1e3484d 100644
--- a/quinn-proto/src/connection/packet_builder.rs
+++ b/quinn-proto/src/connection/packet_builder.rs
@@ -122,7 +122,7 @@ impl PacketBuilder {
         };
         let partial_encode = header.encode(&mut transmits.datagram_mut());
         if conn.peer_params.grease_quic_bit && conn.rng.random() {
-            transmits.datagram_mut()[partial_encode.start] ^= FIXED_BIT;
+            transmits.datagram_mut().written()[partial_encode.start] ^= FIXED_BIT;
         }
 
         let (sample_size, tag_len) = if let Some(ref crypto) = space.crypto {
@@ -255,7 +255,7 @@ impl PacketBuilder {
             .put_bytes(0, packet_crypto.tag_len());
         let encode_start = self.partial_encode.start;
         let mut datagram_buf = transmits.datagram_mut();
-        let packet_buf = &mut datagram_buf[encode_start..];
+        let packet_buf = &mut datagram_buf.written()[encode_start..];
         self.partial_encode.finish(
             packet_buf,
             header_crypto,
diff --git a/quinn-proto/src/connection/transmit_builder.rs b/quinn-proto/src/connection/transmit_builder.rs
index dac79621..968cfaed 100644
--- a/quinn-proto/src/connection/transmit_builder.rs
+++ b/quinn-proto/src/connection/transmit_builder.rs
@@ -1,5 +1,3 @@
-use std::ops::{Deref, DerefMut};
-
 use bytes::BufMut;
 
 use super::BufLen;
@@ -203,7 +201,7 @@ impl<'a> TransmitBuilder<'a> {
 ///
 /// The [`Deref`] impl must return the already written bytes as a slice.  [`DerefMut`] must
 /// allow modification of these already written bytes.
-pub(crate) trait BufSlice: BufMut + Deref<Target = [u8]> + DerefMut {
+pub(crate) trait BufSlice: BufMut {
     /// Returns the number of already written bytes in the buffer
     fn len(&self) -> usize;
 
@@ -242,6 +240,10 @@ impl<'a> DatagramBuffer<'a> {
             max_offset,
         }
     }
+
+    pub(crate) fn written(&mut self) -> &mut [u8] {
+        &mut self.buf[self.start_offset..]
+    }
 }
 
 unsafe impl BufMut for DatagramBuffer<'_> {
@@ -258,20 +260,6 @@ unsafe impl BufMut for DatagramBuffer<'_> {
     }
 }
 
-impl Deref for DatagramBuffer<'_> {
-    type Target = [u8];
-
-    fn deref(&self) -> &Self::Target {
-        &self.buf[self.start_offset..]
-    }
-}
-
-impl DerefMut for DatagramBuffer<'_> {
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        &mut self.buf[self.start_offset..]
-    }
-}
-
 impl BufSlice for DatagramBuffer<'_> {
     /// Returns the number of already written bytes in the buffer
     fn len(&self) -> usize {

```

</details>

From there I started trying to remove the need to have `BufSlice::len` / `BufLen::len` by just being able to take `.written().len()`, but the way the various traits are integrated together made it more laborious than I'm probably willing to do in the course of writing this comment.

> This is really the only reason the BufSlice trait exists for now. It is of course possible to make this w: &mut DatagramBuffer, the cost is a bunch of callers on the receive path would have to be adapted to construct a DatagramBuffer where they currently pass a &mut Vec<u8>.

That doesn't seem that hard. But also, could you clarify why that would be necessary on the receive path?

Overall I quite like where the `DatagramBuffer` abstraction is going. But I think the system of traits surrounding it is over-complicated.

flub:
[9da0da3](https://github.com/quinn-rs/quinn/pull/2168/commits/9da0da308f932b9be8942a568048e3dee5eb4568) continues the `DatagramBuffer` but removes the `BufSlice` trait.  As you can see in the commit it affects a number of calls on the receive side that pass in a `Vec<u8>` for a possible response.  These calls have been updated, but the `DatagramBuffer` type is kept internal in `quinn-proto`, so the public APIs still use `Vec<u8>` for this.

Regarding the `Deref` and `DerefMut`: I think these traits are very natural for this kind of buffer.  Removing them now results in 48 failures, because a few more places use the buffer like that now.  I now also use `Deref<Target = [u8]>` for the `len` implemenation which is needed.  If you manually add a `DatagramBuffer::len` instead of relying on the deref there are only 20 places that use the (mutable) slice access.  Still, I think that's enough to show this is a natural API for this so I'm leaning towards keeping this.

[f7110c7](https://github.com/quinn-rs/quinn/pull/2168/commits/f7110c74dc7e4e56b9167f267c0cec55ff55315d) removes the `BufLen` trait again, since no one seems to like this.  I did this by essentially moving one of the commits from #2195 here.  It uses `BufMut::remaining_mut`, in a way that even avoids the ambiguity that was discussed above :) (that edge case discussed is avoided thanks to `DatagramBuffer` being used where that problem occurred).  Anyway, I like this result and was only worried about that change making this PR larger than needed.

gretchenfrage:
This is high on my list on things to look at! Just been a bit busy.

djc:
Feedback from skimming the current state:

- It looks mostly pretty good
- I think the trait hierarchy looks okay
- I don't like the trivial getters on `TransmitBuilder`, IMO we can just make the fields `pub(super)` instead
- The `start_with_datagram_size()` methods and related naming seems a bit unwieldy

I think it's definitely worth getting the commit history in shape here.

flub:
>     * I don't like the trivial getters on `TransmitBuilder`, IMO we can just make the fields `pub(super)` instead

I really would like to keep the access to fields read-only, and rust only allows this by using getter methods.  When unfamiliar with the code, having everything access everything and mutate from various places make it rather hard to understand what is going on at times.  It also results in logic about fields of structs being encoded in surprising places because everywhere can access it.  I understand that originally this was convenient when first writing Quinn, but also think that it is fine to be more restrictive with the abstractions now.

>     * The `start_with_datagram_size()` methods and related naming seems a bit unwieldy

Yes, I was exploring this a bit more to see what can be done.  This is basically an artifact of how the code was structured before.  Having this code abstracted in one place makes it more obvious that this is a bit of a soar point.  I think it would totally make sense to further refactor this.  E.g. I can easily imagine inverting the api calls from "start new datagram" to 
"finish datagram", it would probably remove the need to keep track of variables like `num_datagrams` as well.  But I do think this comes up *because* inserting this abstraction makes is more easy to recognise this.  I would prefer to not to add more to the current PR, I think it already is an improvement on what was there before.

djc:
> I really would like to keep the access to fields read-only, and rust only allows this by using getter methods. When unfamiliar with the code, having everything access everything and mutate from various places make it rather hard to understand what is going on at times. It also results in logic about fields of structs being encoded in surprising places because everywhere can access it.

I understand this, but I think we can rely on callers in the limited visibility area to respect this if you just document it. Getters are just not worth it in this case.

flub:
Hi, just letting you know I'm not forgetting about this PR.  Am still waiting for a round of feedback from @gretchenfrage before doing followup changes.

quinn-proto/src/connection/mod.rs
```diff
@@ -577,7 +575,7 @@ impl Connection {
                 // (see https://github.com/quinn-rs/quinn/issues/1082)
                 if self
                     .path
-                    .anti_amplification_blocked(segment_size as u64 * (num_datagrams as u64) + 1)
+                    .anti_amplification_blocked((buf.segment_size * buf.num_datagrams) as u64 + 1)
```
gretchenfrage: Factoring out the cast is not a pure refactor. Consider this snippet, wherein the first debug line succeeds, but the experiences overflow:

```rs
#![allow(arithmetic_overflow)]

fn main () {
    const N: u32 = 2_000_000;
    dbg!((N as u64) * (N as u64));
    dbg!((N * N) as u64);
}
```

Do we have reason to believe that's not a problem here?

quinn-proto/src/connection/transmit_buf.rs
```diff
@@ -0,0 +1,91 @@
...
+        self.len() == 0
+    }
+
+    /// The number of bytes written into the buffer so far
+    pub(super) fn len(&self) -> usize {
+        self.buf.len()
+    }
+}
+
+unsafe impl BufMut for TransmitBuf<'_> {
```
gretchenfrage: I'm uncertain about whether it's better to have this delegate implementation of `BufMut`, or to simply pass in `&mut buf.buf` rather than `&mut buf` to methods that expect an `impl BufMut`.

quinn-proto/src/connection/mod.rs
```diff
@@ -3022,7 +3022,7 @@ impl Connection {
         &mut self,
         now: Instant,
         space_id: SpaceId,
-        buf: &mut Vec<u8>,
+        buf: &mut (impl BufMut + BufLen),
```
gretchenfrage: If there were already a single externally defined trait that could be fully sufficient for replacing `&mut Vec<u8>`/`&mut TransmitBuf` with `&mut (impl ThatTrait)`, I would enjoy doing so. But in this case where 1. we are taking on the complexity of defining an additional `BufLen` trait and its impl blocks 2. we are using compound `&mut (impl X + Y)` blocks to utilize it and 3. these methods are nevertheless only called with a one implementation, I'm don't really think it's worth it compared to just passing `&mut Vec<u8>` or `&mut TransmitBuf`.

quinn-proto/src/connection/transmit_buf.rs
```diff
@@ -193,23 +196,3 @@ impl<'a> TransmitBuf<'a> {
         self.buf.as_mut_slice()
     }
 }
-
```
gretchenfrage: It would be easier to review with our development workflow if you edited your PR history to undo this addition rather than including a commit that adds these lines and a subsequent one that removes them

quinn-proto/src/connection/transmit_buf.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    /// The number of bytes written into the buffer so far
+    pub(super) fn len(&self) -> usize {
+        self.buf.len()
+    }
+}
+
+unsafe impl BufMut for TransmitBuf<'_> {
+    fn remaining_mut(&self) -> usize {
+        self.buf.remaining_mut()
```
gretchenfrage: +1 on djc's comments on not being fan of the trivial getters for fully internal things

gretchenfrage (changes requested):
Thanks for your patience. First off, some things I want to highlight as positives:

- Your clarification of the code comments describing the logic of some of this code is appreciated.
- Renaming to `TransmitBuilder` seems good.

Some of the comments I attached here I believe are things you fix in later commits, and thus might not be relevant. But yes, as djc mentioned, it would be helpful for you to craft the PR commit history in a way that incorporates changes to commits into those commits, rather than by changing them with additional commits added to the end, because doing so meshes better with the git usage style of this project.

But yeah I like the direction where this is going.

Ralith:
> it would be helpful for you to craft the PR commit history in a way that incorporates changes to commits into those commits, rather than by changing them with additional commits added to the end, because doing so meshes better with the git usage style of this project.

Just a quick drive-by note -- it's not (just) that this is the style, but concretely it makes reviewing larger PRs *much* easier, since each commit can be reviewed in isolation.

gretchenfrage:
> > it would be helpful for you to craft the PR commit history in a way that incorporates changes to commits into those commits, rather than by changing them with additional commits added to the end, because doing so meshes better with the git usage style of this project.
> 
> Just a quick drive-by note -- it's not (just) that this is the style, but concretely it makes reviewing larger PRs *much* easier, since each commit can be reviewed in isolation.

Well, if a project utilizes a "never rebase" workflow, that _can_ make it easier to see what has changed about the PR since last review

Ralith:
Which is great for the brief period where you can exactly remember what happened last review :grin:

flub:
(Still massaging commits to take feedback into account, these newly pushed commits are not yet ready for a round of review.  Will re-request review once ready again.)

flub:
I'm afraid this had already diverged so far from what we have in noq by now, and noq has only evolved further in other directions. I was hoping this would be part of independent bits of changes that would lead up to the multipath implementation, but this hasn't panned out like that. So realistically I don't think I'll manage to pick this up and finish this PR. It was certainly interesting exploring the various choices one can make here, thanks for the feedback received during this review!

PR #2169, opened 2025-03-07
Minor TLP composition improvements

Ralith:


djc:
Probably ignore the paste advisory for now? aws-lc-rs are on it.

quinn-proto/src/connection/mod.rs
```diff
@@ -3535,7 +3539,7 @@ impl Connection {
     pub(crate) fn is_idle(&self) -> bool {
         Timer::VALUES
             .iter()
-            .filter(|&&t| t != Timer::KeepAlive && t != Timer::PushNewCid)
+            .filter(|&&t| t != Timer::KeepAlive && t != Timer::PushNewCid && t != Timer::KeyDiscard)
```
djc, line 3542: matches!() might be nicer at this point?

deny.toml
```diff
@@ -18,4 +18,8 @@ expression = "ISC AND MIT AND OpenSSL"
 license-files = [{ path = "LICENSE", hash = 0xbd0eed23 }]
 
 [advisories]
-ignore = ["RUSTSEC-2025-0007"]
+ignore = [
+    "RUSTSEC-2025-0007",
```
djc, line 22: IIRC this is for ring, which was withdrawn?

djc (approved):

djc:
Nice!

quinn-proto/src/connection/spaces.rs
```diff
@@ -129,12 +130,12 @@ impl PacketSpace {
             self.immediate_ack_pending = true;
         }
 
-        // Retransmit the data of the oldest in-flight packet
-        if !self.pending.is_empty(streams) {
+        if !self.pending.is_empty(streams) || !datagrams.outgoing.is_empty() {
```
flub, line 133: I don't think this is sufficient, because the next user datagram might not be able to fit in the next QUIC packet.  And in that case you'd still have to include a ping frame.  I guess the bug in this scenario is worse than sending the extra ping frame.

I also think it would be really good to write a test for this.

Ralith: Hmm, good catch. I think we actually want exactly the same judgement as `Connection::can_send_1rtt` here. It looks like that would actually fix another pre-existing bug here, since lazy cleanup of reset streams means `self.pending.is_empty` doesn't actually necessarily imply the ability to transmit any stream frames.

flub: TIL, I assumed the stream side of this was correct.

Ralith: I'm actually waffling a bit on whether the datagram side is an issue or not. The interaction of TLP queue accounting and size limits is nontrivial. I think I'll need to build some tests to draw empirical conclusions and validate a solution.

flub: I'm wondering whether the best thing is to revert this change and create an issue for the extra ping.  That way we have a correct-but-slightly-wasteful implementation and don't forget to follow up.

flub: I'm wondering whether the best thing is to revert this change and create an issue for the extra ping.  That way we have a correct-but-slightly-wasteful implementation and don't forget to follow up.

PR #2178, opened 2025-03-19
[quinn-udp] Support vectored sends

kixelated:
Initial attempt for linux. Not sure what to do about Windows.

### DIFF
--- a/quinn-udp/src/fallback.rs
+++ b/quinn-udp/src/fallback.rs
@@ -1,5 +1,5 @@
 use std::{
-    io::{self, IoSliceMut},
+    io::{self, IoSlice, IoSliceMut},
     sync::Mutex,
     time::Instant,
 };
@@ -93,8 +93,8 @@ impl UdpSocketState {
 }
 
 fn send(socket: UdpSockRef<'_>, transmit: &Transmit<'_>) -> io::Result<()> {
-    socket.0.send_to(
-        transmit.contents,
+    socket.0.send_to_vectored(
+        transmit.buffers,
         &socket2::SockAddr::from(transmit.destination),
     )
 }


quinn-udp/benches/throughput.rs
```diff
@@ -90,7 +90,7 @@ pub fn criterion_benchmark(c: &mut Criterion) {
                             send_state.send((&send_socket).into(), &transmit)
                         })
                         .unwrap();
-                    sent += transmit.contents.len();
```
kixelated, line 93: This is why I renamed `contents`, otherwise it wouldn't be a breaking change.

quinn/src/lib.rs
```diff
@@ -99,16 +99,6 @@ enum ConnectionEvent {
     Rebind(Arc<dyn AsyncUdpSocket>),
 }
 
-fn udp_transmit<'a>(t: &proto::Transmit, buffer: &'a [u8]) -> udp::Transmit<'a> {
-    udp::Transmit {
-        destination: t.destination,
-        ecn: t.ecn.map(udp_ecn),
-        contents: buffer,
```
kixelated, line 106: Errors with a reference to a temporary variable if you do:

```rust
  buffers: &[IoSlice::new(buffer)],
```

Ralith:
This looks good to me overall.

> Initial attempt for linux. Not sure what to do about Windows.

I think Windows should look very similar to Unix here, with easy first-class support. [`WSAMSG`](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/struct.WSAMSG.html) takes an array of `WSABUF`s, aka `IoSlice`s.

djc:
@kixelated do you intend to work on this more? If not, maybe close it?

kixelated:
> @kixelated do you intend to work on this more? If not, maybe close it?

Nah, it's a premature optimization and I'm now working on something else, so not really in a position to benchmark/optimize.

PR #2179, opened 2025-03-20
feat: Allow changing the UDP send/receive buffer sizes

larseggert:
Add `UdpSocketState::set_send_buffer_size` and `UdpSocketState::set_recv_buffer_size`, which allow resizing the respective socket buffers.

CC @mxinden

### DIFF
--- a/quinn-udp/src/fallback.rs
+++ b/quinn-udp/src/fallback.rs
@@ -86,6 +86,30 @@ impl UdpSocketState {
         1
     }
 
+    /// Resize the send buffer of `socket` to `bytes`
+    #[inline]
+    pub fn set_send_buffer_size(&self, socket: UdpSockRef<'_>, bytes: usize) -> io::Result<()> {
+        socket.0.set_send_buffer_size(bytes)
+    }
+
+    /// Resize the receive buffer of `socket` to `bytes`
+    #[inline]
+    pub fn set_recv_buffer_size(&self, socket: UdpSockRef<'_>, bytes: usize) -> io::Result<()> {
+        socket.0.set_recv_buffer_size(bytes)
+    }
+
+    /// Get the size of the `socket` send buffer
+    #[inline]
+    pub fn send_buffer_size(&self, socket: UdpSockRef<'_>) -> io::Result<usize> {
+        socket.0.send_buffer_size()
+    }
+
+    /// Get the size of the `socket` receive buffer
+    #[inline]
+    pub fn recv_buffer_size(&self, socket: UdpSockRef<'_>) -> io::Result<usize> {
+        socket.0.recv_buffer_size()
+    }
+
     #[inline]
     pub fn may_fragment(&self) -> bool {
         true
--- a/quinn-udp/tests/tests.rs
+++ b/quinn-udp/tests/tests.rs
@@ -1,8 +1,8 @@
 #[cfg(not(any(target_os = "openbsd", target_os = "netbsd", solarish)))]
-use std::net::{SocketAddr, SocketAddrV4, SocketAddrV6};
+use std::net::{SocketAddr, SocketAddrV6};
 use std::{
     io::IoSliceMut,
-    net::{IpAddr, Ipv4Addr, Ipv6Addr, UdpSocket},
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddrV4, UdpSocket},
     slice,
 };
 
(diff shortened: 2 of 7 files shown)


quinn-udp/tests/tests.rs
```diff
@@ -186,6 +186,36 @@ fn gso() {
     );
 }
 
+#[test]
+fn socket_buffers() {
```
mxinden, line 190: Could this test also check that the send and recv buffer size was updated via [`socket2` `send_buffer_size`](https://docs.rs/socket2/latest/socket2/struct.Socket.html#method.send_buffer_size) etc.?

mxinden (commented):
Long story short, I am in favor of this patch.

Increasing the UDP socket send and receive buffer can have a positive performance impact. That said, the additional memory footprint might not be an option for all `quinn-udp` users. Thus an opt-in set of functions on `UdpSocketState` seems ideal to me.

Long term, once we gathered more experience, we might want to bump the buffer sizes by default.

mxinden:
Worth having the code below use the new functions?

https://github.com/quinn-rs/quinn/blob/434c35861e68aac1da568bcd0b1523603f73f255/perf/src/lib.rs#L28-L49

mxinden:
Changes look good to me.

CI is failing, e.g. on Windows with:

```
---- socket_buffers stdout ----

thread 'socket_buffers' panicked at quinn-udp\tests\tests.rs:211:58:
called `Result::unwrap()` on an `Err` value: Os { code: 10022, kind: InvalidInput, message: "An invalid argument was supplied." }
```

thomaseizinger (commented):
I have played with this before as well and noticed some interesting behaviour on several platforms. Linux for example will give you however much buffer size it can if you request a really large size and _not_ fail the function.

I am not sure if this is common knowledge around socket APIs :)

I am curious to learn about any performance gains you get from this. In my testing, this wasn't a bottleneck but that was also ~6 months ago and I've landed other perf improvements in our codebase since.

larseggert:
I am thinking that I will change the test to check that a "set" operation results in _any_ change of the buffer size in the right direction (increase or decrease).

larseggert:
I can't figure out why `UdpSocketState::new` fails on WIndows 🤷

mxinden:
Do you know which of the syscalls in `UdpSocketState::new` fails?

https://github.com/quinn-rs/quinn/blob/434c35861e68aac1da568bcd0b1523603f73f255/quinn-udp/src/windows.rs#L31-L120

larseggert:
> Do you know which of the syscalls in `UdpSocketState::new` fails?

`thread 'socket_buffers' panicked at quinn-udp\src\windows.rs:44:42:`
https://github.com/quinn-rs/quinn/actions/runs/13993457832/job/39182608862?pr=2179#step:6:350

larseggert:
Turns out one needs to bind both sockets on Windows.

Ralith (changes requested):
Don'g forget `fallback.rs`

Ralith (approved):

quinn-udp/src/fallback.rs
```diff
@@ -86,6 +86,18 @@ impl UdpSocketState {
         1
     }
 
+    /// Resize the send buffer of `socket` to `bytes`.
```
djc, line 89: Tiny nit: please remove the periods from the first sentences of docstrings.

quinn-udp/tests/tests.rs
```diff
@@ -186,6 +186,79 @@ fn gso() {
...
+    )
+    .unwrap();
+    for sock in [&send, &recv] {
+        sock.bind(&socket2::SockAddr::from(SocketAddrV4::new(
+            Ipv4Addr::LOCALHOST,
+            0,
+        )))
+        .unwrap();
+
+        let sockstate = UdpSocketState::new(sock.into()).expect("created socket state");
```
djc, line 217: Tiny nit: `sockstate` -> `sock_state` or even `socket_state`.

djc (approved):
LGTM. Can you squash your changes into a single commit?

Do we want a quinn-udp version bump so this can be published?

larseggert:
I'll squash Monday. I also want to do a follow-up with methods for reading the buffer sizes. Bump afterwards would be nice.

djc:
> Bump afterwards would be nice.

Just add a commit with a version bump when you're ready.

larseggert:
I have no idea why (unrelated) tests are failing in CI, they are succeeding locally.

djc:
* Published quinn-udp v0.5.11 at registry `crates-io`
* [new tag]           quinn-udp-0.5.11 -> quinn-udp-0.5.11
* [Release notes](https://github.com/quinn-rs/quinn/releases/tag/quinn-udp-0.5.11)

PR #2185, opened 2025-03-23
fix: respecting max_datagrams in poll_transmit

filipe-cantarelli:
Hello,

I've been playing with quinn-proto recently and I noticed that, sometimes, when the maximum segment is large enough quinn will go over the max_datagrams limit of poll_transmit. I believe this PR will fix this issue.

Also, considering that quin-proto is a low level crate, it feels like this code:
https://github.com/quinn-rs/quinn/blob/434c35861e68aac1da568bcd0b1523603f73f255/quinn-proto/src/connection/mod.rs#L454-L457
should be removed and left for the upper level code to decide what max limit they truly want to use.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -564,7 +564,7 @@ impl Connection {
                 // We need to send 1 more datagram and extend the buffer for that.
 
                 // Is 1 more datagram allowed?
-                if buf_capacity >= segment_size * max_datagrams {
+                if num_datagrams >= max_datagrams {
                     // No more datagrams allowed
                     break;
                 }
@@ -577,7 +577,7 @@ impl Connection {
                 // (see https://github.com/quinn-rs/quinn/issues/1082)
                 if self
                     .path
-                    .anti_amplification_blocked(segment_size as u64 * num_datagrams + 1)
+                    .anti_amplification_blocked(segment_size as u64 * (num_datagrams as u64) + 1)
                 {
                     trace!("blocked by anti-amplification");
                     break;
@@ -965,7 +965,7 @@ impl Connection {
         trace!("sending {} bytes in {} datagrams", buf.len(), num_datagrams);
         self.path.total_sent = self.path.total_sent.saturating_add(buf.len() as u64);
 
-        self.stats.udp_tx.on_sent(num_datagrams, buf.len());
+        self.stats.udp_tx.on_sent(num_datagrams as u64, buf.len());
 
         Some(Transmit {
             destination: self.path.remote,
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -453,7 +453,7 @@ impl Connection {
         assert!(max_datagrams != 0);
         let max_datagrams = match self.config.enable_segmentation_offload {
             false => 1,
-            true => max_datagrams.min(MAX_TRANSMIT_SEGMENTS),
+            true => max_datagrams,
         };
 
         let mut num_datagrams = 0;
@@ -3949,13 +3949,6 @@ const MIN_PACKET_SPACE: usize = MAX_HANDSHAKE_OR_0RTT_HEADER_SIZE + 32;
 // scid len + scid + length + pn
 const MAX_HANDSHAKE_OR_0RTT_HEADER_SIZE: usize =
     1 + 4 + 1 + MAX_CID_SIZE + 1 + MAX_CID_SIZE + VarInt::from_u32(u16::MAX as u32).size() + 4;
-
-/// The maximum amount of datagrams that are sent in a single transmit
-///
-/// This can be lower than the maximum platform capabilities, to avoid excessive
-/// memory allocations when calling `poll_transmit()`. Benchmarks have shown
-/// that numbers around 10 are a good compromise.
-const MAX_TRANSMIT_SEGMENTS: usize = 10;
 
 /// Perform key updates this many packets before the AEAD confidentiality limit.
 ///
--- a/quinn-proto/src/tests/mod.rs
+++ b/quinn-proto/src/tests/mod.rs
@@ -1829,6 +1829,51 @@ fn tail_loss_small_segment_size() {
     assert_eq!(server_stats.frame_rx.datagram, DGRAM_NUM);
 }
 
+// Respect max_datagrams when TLP happens
+#[test]
+fn tail_loss_respect_max_datagrams() {
+    let _guard = subscribe();
+    let client_config = {
+        let mut c_config = client_config();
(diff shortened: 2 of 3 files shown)


Ralith:
> Also, considering that quin-proto is a low level crate, it feels like this code:

That code exists so that the application layer can easily opt out of GSO without needing to replace the UDP socket abstraction.

quinn-proto/src/connection/mod.rs
```diff
@@ -456,7 +456,7 @@ impl Connection {
             true => max_datagrams.min(MAX_TRANSMIT_SEGMENTS),
         };
 
-        let mut num_datagrams = 0;
+        let mut num_datagrams = 0usize;
```
Ralith, line 459: This and related changes don't seem to have anything to do with the intent of the commit. Please keep cosmetic refactoring separate.

quinn-proto/src/connection/mod.rs
```diff
@@ -564,7 +564,7 @@ impl Connection {
                 // We need to send 1 more datagram and extend the buffer for that.
 
                 // Is 1 more datagram allowed?
-                if buf_capacity >= segment_size * max_datagrams {
+                if buf_capacity >= segment_size * max_datagrams || num_datagrams >= max_datagrams {
```
Ralith, line 567: Do we need both of these conditions? I'm guessing the issue here is that, for flights ending in a TLP, `buf_capacity` is less than a segment below `segment_size * max_datagrams`. Alternatively, maybe the condition should be `buf_capacity > segment_size * (max_datagrams - 1)`, to reflect N segment-size datagrams plus a possibly shorter final datagram?

Ralith (changes requested):

quinn-proto/src/connection/mod.rs
filipe-cantarelli: The proposed change `buf_capacity > segment_size * (max_datagrams - 1)` does pass all the tests and works in my case (using max_datagrams=1).

But after some debugging, I feel a little bit hard to understand, the issue here being that segment_size is incorrect for the first datagram (if it's a TLP).

In the first datagram case(if it's a loss), what seems to be happening is:
buff_capacity is set to INITIAL_MTU, 1200:
```rust
let next_datagram_size_limit = match self.spaces[space_id].loss_probes {
    0 => segment_size,
    _ => {
        self.spaces[space_id].loss_probes -= 1;
        // Clamp the datagram to at most the minimum MTU to ensure that loss probes
        // can get through and enable recovery even if the path MTU has shrank
        // unexpectedly.
        std::cmp::min(segment_size, usize::from(INITIAL_MTU))
    }
};
buf_capacity += next_datagram_size_limit;
```
Even though we know that segment_size cannot possibly be the default anymore, and the variable segment_size will only be updated after we read the first datagram.

So this current conditional:
```rust
// Is 1 more datagram allowed?
if buf_capacity >= segment_size * max_datagrams {
    // No more datagrams allowed
    break;
}
```
Actually fails because segment_size is not correct, and IMO this conditional is much easier to understand than the proposed.

Maybe another way to fix this would be change the first piece of code to:

```rust
// Allocate space for another datagram
segment_size = match self.spaces[space_id].loss_probes {
    0 => segment_size,
    _ => {
        self.spaces[space_id].loss_probes -= 1;
        // Clamp the datagram to at most the minimum MTU to ensure that loss probes
        // can get through and enable recovery even if the path MTU has shrank
        // unexpectedly
        std::cmp::min(segment_size, usize::from(INITIAL_MTU))
    }
};
buf_capacity += segment_size;
```
With that I don't think we need to change the conditional at all.

filipe-cantarelli:
> > Also, considering that quin-proto is a low level crate, it feels like this code:
> 
> That code exists so that the application layer can easily opt out of GSO without needing to replace the UDP socket abstraction.

How about `max_datagrams.min(MAX_TRANSMIT_SEGMENTS)`, is it possible to remove this limitation so the app code request more than 10 segments?

Ralith:
Oh, you were concerned about the upper bound? We could probably get rid of that entirely, yeah.

quinn-proto/src/connection/mod.rs
Ralith: > But after some debugging, I feel a little bit hard to understand

Sure, I think only checking `num_datagrams` is clearer. Does that work?

> the issue here being that segment_size is incorrect for the first datagram (if it's a TLP).

I'm not sure what you mean here. `segment_size` describes how datagrams are laid out in the buffer. If it was incorrect the payload would be corrupted.

> Maybe another way to fix this would be change the first piece of code to:

That would yield an incorrect segment size when the last packet in a batch is smaller than the rest.

filipe-cantarelli: Ok, I went with the num_datagrams only check and managed to add a test to cover that case so it doesn't regress  it in the future, as I see there's a big refactor coming for this method.

filipe-cantarelli:
> Oh, you were concerned about the upper bound? We could probably get rid of that entirely, yeah.

Is this something that could be done here in this PR?

djc:
(Be sure to rebase on main to appease CI.)

filipe-cantarelli:
Could I then move the 10 limit to quinn in this PR and then open an issue to follow up relaxing it? I could maybe work on that later.

As far as where to cap it, maybe here is a good place:
https://github.com/quinn-rs/quinn/blob/main/quinn/src/connection.rs#L984

Ralith:
> Could I then move the 10 limit to quinn in this PR and then open an issue to follow up relaxing it? I could maybe work on that later.

SGTM, thanks!

> As far as where to cap it, maybe here is a good place:

Yep, that's the spot.

quinn-proto/src/tests/mod.rs
```diff
@@ -1829,6 +1829,50 @@ fn tail_loss_small_segment_size() {
...
+            .send(vec![0; DGRAM_LEN].into(), false)
+            .unwrap();
+    }
+
+    pair.drive();
+
+    // Finally checking the number of sent udp datagrams match the number of iops
+    let client_stats = pair.client_conn_mut(client_ch).stats();
+    assert_eq!(client_stats.udp_tx.ios, client_stats.udp_tx.datagrams);
+}
```
djc, line 1875: Nit: please leave an empty line here.

djc (approved):

filipe-cantarelli:
@Ralith Follow up issue created https://github.com/quinn-rs/quinn/issues/2189

filipe-cantarelli:
@djc, @Ralith what are the next steps for this pr?

djc:
Would be good to get another round of review from @Ralith before merging.

Ralith (approved):
Thanks!

Issue #2186, opened 2025-03-24
CI jobs broken due to conflicting implementation from deranged


djc:
This seems to be causing current CI failures:

https://github.com/jhpratt/deranged/issues/18

cc @larseggert

larseggert:
Should this have been fixed by #2187?

djc:
Well, it should no longer break CI at least. It could still be broken for downstreams that select the broken deranged version.

larseggert:
Hm. I merged `main` into #2179 after #2187 was merged there and CI still fails.

djc:
Yeah, you need to regenerate the lockfile due to the version bump.

PR #2187, opened 2025-03-24
Version Cargo lockfile

djc:
To mitigate #2186.


Ralith (approved):
AFAIK this is idiomatic these days :+1: 

We may need to fiddle with dependabot to make it not spammy.

PR #2188, opened 2025-03-25
Avoid inference hazard for integer comparisons

### LINKED ISSUE #2186: CI jobs broken due to conflicting implementation from deranged

djc:
Fixes #2186.

The deranged maintainer appears to be [unlikely](https://github.com/jhpratt/deranged/issues/18#issuecomment-2749704917) to revert the new `PartialEq` impls that are causing this. I'm guessing this is not causing downstream issues because only dev-dependencies are pulling in the deranged crate, so there should be no urgent need for a release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -665,9 +665,9 @@ checksum = "43da5946c66ffcc7745f48db692ffbb10a83bfe0afd96235c5c2a4fb23994929"
 
 [[package]]
 name = "deranged"
-version = "0.4.0"
+version = "0.4.1"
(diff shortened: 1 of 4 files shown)


flub (approved):

Ralith (approved):
LGTM, honestly seems more readable anyway.

> The deranged maintainer

That seems a bit harsh! Oh, wait.

PR #2191, opened 2025-03-30
docs: update document with example code separated

kota-yata:
context: https://github.com/quinn-rs/quinn/issues/2152

- set up docs/book so that example code in the document can be separated into .rs files and can be checked by the compiler
- update document to align with the latest version

PR #2192, opened 2025-03-30
Improve book code samples

djc:
Followup from #2191.


Ralith (approved):
Nice to finally have this tidied up!

PR #2199, opened 2025-04-11
quinn-udp: handle EMSGSIZE in a common place

thomaseizinger:
At present, each of the various `send` functions within `quinn-dup` handle the error case of `EMSGSIZE` individually. This code duplication can be avoided by moving this error handling to the `UdpSocketState::send` function.

Additionally, contrary to what the comment states, this error code can also be returned if the provided datagram is simply to large which can happen with some very big GSO batches.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -208,6 +208,9 @@ impl UdpSocketState {
         match send(self, socket.0, transmit) {
             Ok(()) => Ok(()),
             Err(e) if e.kind() == io::ErrorKind::WouldBlock => Err(e),
+            // - EMSGSIZE is expected for MTU probes. Future work might be able to avoid
+            //   these by automatically clamping the MTUD upper bound to the interface MTU.
+            Err(e) if e.raw_os_error() == Some(libc::EMSGSIZE) => Ok(()),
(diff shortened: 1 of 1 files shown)


thomaseizinger:
> Additionally, contrary to what the comment states, this error code can also be returned if the provided datagram is simply to large which can happen with some very big GSO batches.

I am still drilling down the problem here but I've noticed that with the default GSO batch size of 64, it is quite easy to hit this if you naively chunk your datagram into `segment_size * max_gso_segments`. I am not sure if that is what you are meant to do but that is what I did in my code. This created packet loss because `quinn-udp` was silently discarding this error.

I think there might be some more work we can do here by adding a warning / error if the `contents` of `Transmit` passed to `quinn_udp` are for sure to large to send, i.e. > `u16::MAX`.

Ralith (approved):
Thanks!

mxinden (approved):
A lot better.

larseggert:
We see some failures (https://github.com/mozilla/neqo/issues/2752) but I'm not sure this is because of this PR?

mxinden:
My bad. Unrelated to `quinn-udp`. Will be addressed in https://github.com/mozilla/neqo/pull/2755.

Issue #2206, opened 2025-04-15
`quinn-udp` panics with `no control buffer space remaining` and `apple-fast-path`


thomaseizinger:
~It works fine on MacOS 15.3.1. We only noticed this after upgrading one of our dev-machines to the 15.4 release.~

I believe this interacts with #2154, I added some additional logs and just before it panics, we can see the following:

```
2025-04-15T00:34:02.790Z DEBUG quinn_udp::cmsg: level=0 ty=7 space=16 control_len=88 len=16
2025-04-15T00:34:02.790Z DEBUG quinn_udp::cmsg: level=41 ty=46 space=32 control_len=88 len=16
```

That is the first occurrence of attempting to push a cmsg of type 46, which is [`IPV6_PKTINFO`](https://github.com/rust-lang/libc/blob/ce2edbbaa9bbc3273e356694b6dd2c39fc7b788b/src/unix/bsd/apple/mod.rs#L4326).

I don't yet understand what is going on here. We seem to have enough space? Here are the logs I added: https://github.com/thomaseizinger/quinn/commit/0b961822954130b538b57730923ea7ac90eefcd2

This reproduces in the unit-tests locally for me as soon as you set `src_ip`. I've opened a draft PR to see if CI catches it too: https://github.com/quinn-rs/quinn/pull/2205

thomaseizinger:
Oh, just noticed that this reproduces on `macos-latest` as well, which is 14.7.1: https://github.com/quinn-rs/quinn/actions/runs/14459487435/job/40549305291?pr=2205#step:7:36

thomaseizinger:
I think this was broken by #2154. If I comment this code out, the tests are passing. Perhaps this should be gated just on _reading_ cmsg headers rather that also on the write path?

PR #2207, opened 2025-04-15
fix(quinn-udp): only inspect `cmsg_len` when reading cmsg's

thomaseizinger:
This PR is a version of https://github.com/quinn-rs/quinn/pull/2205 with a clean commit history that should be ready for merging.

Resolves: #2206


thomaseizinger:
An alternative approach here would be to move this to the `Iterator` implementation. That would allow us to safely reuse `cmsg_nxt_hdr` for advancing it in the write-path.

Ralith:
Closing in favor of #2208, which indeed looks cleaner. Thanks!

PR #2208, opened 2025-04-15
fix(quinn-udp): move cmsg-len check to Iterator

thomaseizinger:
This is an alternative version to #2207 that has a smaller change to the API surface.

Resolves: #2206

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -127,7 +127,7 @@ jobs:
         if: ${{ matrix.rust }} == "stable"
       - run: cargo test --locked -p quinn-udp --benches
       - run: cargo test --locked -p quinn-udp --benches --features fast-apple-datapath
-        if: ${{ matrix.os }} == "macos-latest"
+        if: ${{ runner.os }} == "macOS"
 
   test-aws-lc-rs:
     runs-on: ubuntu-latest
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -122,6 +122,8 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo build --locked --all-targets
       - run: cargo test --locked
+      - run: cargo test --locked -p quinn-udp --features fast-apple-datapath
+        if: ${{ runner.os }} == "macOS"
       - run: cargo test --locked -- --ignored stress
       - run: cargo test --locked --manifest-path fuzz/Cargo.toml
         if: ${{ matrix.rust }} == "stable"
--- a/quinn-udp/src/cmsg/mod.rs
+++ b/quinn-udp/src/cmsg/mod.rs
@@ -112,6 +112,17 @@ impl<'a, M: MsgHdr> Iterator for Iter<'a, M> {
     fn next(&mut self) -> Option<Self::Item> {
         let current = self.cmsg.take()?;
         self.cmsg = unsafe { self.hdr.cmsg_nxt_hdr(current).as_ref() };
+
+        #[cfg(apple_fast)]
+        {
+            // On MacOS < 14 CMSG_NXTHDR might continuously return a zeroed cmsg. In
+            // such case, return `None` instead, thus indicating the end of
+            // the cmsghdr chain.
+            if current.len() < mem::size_of::<M::ControlMessage>() {
+                return None;
+            }
+        }
+
         Some(current)
     }
 }
--- a/quinn-udp/tests/tests.rs
+++ b/quinn-udp/tests/tests.rs
@@ -32,6 +32,29 @@ fn basic() {
 }
 
 #[test]
+fn basic_src_ip() {
+    let send = UdpSocket::bind((Ipv6Addr::LOCALHOST, 0))
+        .or_else(|_| UdpSocket::bind((Ipv4Addr::LOCALHOST, 0)))
+        .unwrap();
+    let recv = UdpSocket::bind((Ipv6Addr::LOCALHOST, 0))
+        .or_else(|_| UdpSocket::bind((Ipv4Addr::LOCALHOST, 0)))
+        .unwrap();
+    let src_ip = send.local_addr().unwrap().ip();
+    let dst_addr = recv.local_addr().unwrap();
+    test_send_recv(
+        &send.into(),
+        &recv.into(),
+        Transmit {
+            destination: dst_addr,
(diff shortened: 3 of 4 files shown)


quinn-udp/src/cmsg/mod.rs
```diff
@@ -112,6 +112,17 @@ impl<'a, M: MsgHdr> Iterator for Iter<'a, M> {
     fn next(&mut self) -> Option<Self::Item> {
         let current = self.cmsg.take()?;
         self.cmsg = unsafe { self.hdr.cmsg_nxt_hdr(current).as_ref() };
+
+        #[cfg(apple_fast)]
+        {
+            // On MacOS < 14 CMSG_NXTHDR might continuously return a zeroed cmsg. In
+            // such case, return `None` instead, thus indicating the end of
+            // the cmsghdr chain.
+            if current.len() < std::mem::size_of::<Self::Item>() {
```
Ralith, line 121: `Self::Item` is a reference. Should this take the size of `M::ControlMessage` instead?

quinn-udp/tests/tests.rs
```diff
@@ -31,6 +31,29 @@ fn basic() {
     );
 }
 
+#[test]
+fn basic_src_ip_ipv6() {
+    let send = UdpSocket::bind((Ipv6Addr::LOCALHOST, 0))
+        .or_else(|_| UdpSocket::bind((Ipv4Addr::LOCALHOST, 0)))
```
Ralith, line 37: If this test is IPv6 specific, do these `or_else` clauses make sense?

thomaseizinger: Ah that was a left over, it is not IPv6 speciifc. As soon as we add more than the initial cmsg, this will break.

thomaseizinger:
Should I squash the last two commits? The test is purposely failing on the 2nd commit and only fixed in the 3rd so if someone were to do a bisect across that, they may run into that.

djc:
> Should I squash the last two commits? The test is purposely failing on the 2nd commit and only fixed in the 3rd so if someone were to do a bisect across that, they may run into that.

Yes, please. Or, add #[ignore] in the former commit and remove it in the latter.

thomaseizinger:
> > Should I squash the last two commits? The test is purposely failing on the 2nd commit and only fixed in the 3rd so if someone were to do a bisect across that, they may run into that.
> 
> Yes, please. Or, add #[ignore] in the former commit and remove it in the latter.

Squashed them together.

.github/workflows/rust.yml
```diff
@@ -122,6 +122,8 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo build --locked --all-targets
       - run: cargo test --locked
+      - run: cargo test --locked -p quinn-udp --features fast-apple-datapath
+        if: ${{ runner.os }} == "macOS"
```
mxinden, line 126: Further below we use `${{ matrix.os }} == "macos-latest"`. Should we be consistent, either changing here or below?

quinn-udp/src/cmsg/mod.rs
```diff
@@ -112,6 +112,17 @@ impl<'a, M: MsgHdr> Iterator for Iter<'a, M> {
     fn next(&mut self) -> Option<Self::Item> {
         let current = self.cmsg.take()?;
         self.cmsg = unsafe { self.hdr.cmsg_nxt_hdr(current).as_ref() };
+
+        #[cfg(apple_fast)]
```
mxinden, line 116: On non `apple_fast`, is there value in reading a `< std::mem::size_of::<M::ControlMessage>()` control message?

mxinden (approved):
Sorry for the regression. Thanks for the thorough debugging work!

quinn-udp/src/cmsg/mod.rs
thomaseizinger: Probably not. But there is also no bug (that we know of) on other systems ([as you argued earlier](https://github.com/quinn-rs/quinn/pull/2154#issuecomment-2658548664)). I am fine either way, I just thought keeping the semantics the same is good!

.github/workflows/rust.yml
mxinden: I suggest changing the entry below, i.e. running the benchmark across all MacOS versions. Happy to do a follow-up pull request in case you prefer not doing it here.

quinn-udp/src/cmsg/mod.rs
mxinden: :wave: old Max.

I am fine either way. Thanks.

.github/workflows/rust.yml
thomaseizinger: Updated!

quinn-udp/src/cmsg/mod.rs
```diff
@@ -112,6 +112,17 @@ impl<'a, M: MsgHdr> Iterator for Iter<'a, M> {
...
+        {
+            // On MacOS < 14 CMSG_NXTHDR might continuously return a zeroed cmsg. In
+            // such case, return `None` instead, thus indicating the end of
+            // the cmsghdr chain.
+            if current.len() < std::mem::size_of::<M::ControlMessage>() {
```
djc, line 121: Looks like you can drop the `std::` prefix?

djc (approved):
Nice!

thomaseizinger:
@mxinden @djc Is there anything left to do here to be able to merge this? The `apple-fast-datapath` is currently broken as soon as you start adding more cmsg (like ECN information and source IPs).

quinn-udp/src/cmsg/mod.rs
thomaseizinger: Ah, it would have been nice to reproduce this in CI or have some other form of regression test.

Do you think there is much value in adding a regression test that attemps to read a cmsg buffer that ends up returning zerod cmsgs?

gretchenfrage:
This has been approved, it looks like conversations have been resolved to satisfaction, the commit history is clean, it passes CI, there's no merge conflicts, and the diff since mxinden's approval is negligible and agreeable. So I'm putting this in the merge queue. I resolved [this conversation](https://github.com/quinn-rs/quinn/pull/2208#discussion_r2051230220) so I can do this but feel free to make a new issue and/or PR for that if you decide that makes sense.

PR #2209, opened 2025-04-16
Bump the quinn-proto version to 0.11.11 for release

faern:
Bumping the version to be able to get a patch release of `quinn-proto`, in order to have a version with the fix for #2167 on crates.io. I think we only need `quinn-proto` for this, not `quinn` itself, right?

I'm not sure if there is any other ceremony around releases? I saw nothing interesting in the history, and no contributions-file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1581,7 +1581,7 @@ dependencies = [
 
 [[package]]
 name = "quinn-proto"
-version = "0.11.10"
+version = "0.11.11"
 dependencies = [
  "arbitrary",
  "assert_matches",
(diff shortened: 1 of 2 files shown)


djc (approved):

djc:
* Published quinn-proto v0.11.11 at registry `crates-io`
* [new tag]           quinn-proto-0.11.11 -> quinn-proto-0.11.11
* [Release notes](https://github.com/quinn-rs/quinn/releases/tag/quinn-proto-0.11.11)

No extra ceremony, thanks!

Issue #2214, opened 2025-04-28
quinn-udp: panic in `CMsgHdr` `decode` with `fast-apple-datapath` on MacOS 10.15


mxinden:
_Documenting early findings thus far. Still debugging._

On Firefox's CI we see MacOS 10.15 failing the following `debug_assert_eq`.

https://github.com/quinn-rs/quinn/blob/8936cc059c66c6010aeee873a996f82d3fbbc389/quinn-udp/src/cmsg/mod.rs#L84-L88

```
[task 2025-04-23T18:01:03.715Z] 18:01:03     INFO -  PID 11647 | [11649] Hit MOZ_CRASH(assertion `left == right` failed
[task 2025-04-23T18:01:03.715Z] 18:01:03     INFO -  PID 11647 |   left: 1626934053
[task 2025-04-23T18:01:03.715Z] 18:01:03     INFO -  PID 11647 |  right: 16) at /builds/worker/checkouts/gecko/third_party/rust/quinn-udp/src/cmsg/mod.rs:86
```

https://treeherder.mozilla.org/logviewer?job_id=505190482&repo=autoland&lineNumber=7087

//CC @larseggert

mxinden:
I believe I found the root of this issue.

We currently don't initialize (zeroize) the control message array before passing it to `recvmsg_x`:

https://github.com/quinn-rs/quinn/blob/8936cc059c66c6010aeee873a996f82d3fbbc389/quinn-udp/src/unix.rs#L509

While `recvmsg_x` overwrites the first couple of entries with valid control messages, it does not zero the ones it doesn't need.

We then use `CMSG_NXTHDR` to iterate through the control message array:

https://github.com/quinn-rs/quinn/blob/8936cc059c66c6010aeee873a996f82d3fbbc389/quinn-udp/src/cmsg/unix.rs#L44-L47

`CMSG_NXTHDR` _should_ discard the unset control messages based on `msg_controllen`.

https://docs.rs/libc/latest/src/libc/unix/linux_like/linux/mod.rs.html#6021-6025

But in our case on MacOS 10.15 it does not seem to do it.

Thus on MacOS 10.15 we are seeing garbage values like the following:

```
control_len() = 88
cmsg.cmsg_len = 2638540346
cmsg.cmsg_level = -1270110489
cmsg.cmsg_type = 171422048
```

When we zero the control message array instead:

``` diff
[...]

PR #2216, opened 2025-04-29
fix(udp): zero control message array on fast-apple-datapath

mxinden:
`quinn-udp` with fast-apple-datapath previously did not initialize the control message array memory before passing it to `recvmsg_x`.

On MacOS 10.15 `recvmsg_x` does not seem to set `msg_controllen`. Thus `CMSG_NXTHDR` reads beyond the control messages written by `recvmsg_x`, into the unitinialized memory region.

With this commit, the control message array is initialized (with zeroes) before passing it to `recvmsg_x`, thus no longer reading unset control messages.

See https://github.com/quinn-rs/quinn/issues/2214 for details.

//CC @larseggert

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -506,7 +506,13 @@ fn recv(io: SockRef<'_>, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) -> 
 #[cfg(apple_fast)]
 fn recv(io: SockRef<'_>, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) -> io::Result<usize> {
     let mut names = [MaybeUninit::<libc::sockaddr_storage>::uninit(); BATCH_SIZE];
-    let mut ctrls = [cmsg::Aligned(MaybeUninit::<[u8; CMSG_LEN]>::uninit()); BATCH_SIZE];
+    // MacOS 10.15 `recvmsg_x` does not override the `msghdr_x`
+    // `msg_controllen`. Thus, after the call to `recvmsg_x`, one does not know
+    // which control messages have been written to. To prevent reading
+    // uninitialized memory, do not use `MaybeUninit` for `ctrls`, instead
+    // initialize `ctrls` with `0`s. A control message of all `0`s is
+    // automatically skipped by `libc::CMSG_NXTHDR`.
+    let mut ctrls = [cmsg::Aligned([0u8; CMSG_LEN]); BATCH_SIZE];
     let mut hdrs = unsafe { mem::zeroed::<[msghdr_x; BATCH_SIZE]>() };
     let max_msg_count = bufs.len().min(BATCH_SIZE);
     for i in 0..max_msg_count {
@@ -674,7 +680,7 @@ fn prepare_recv(
 fn prepare_recv(
     buf: &mut IoSliceMut,
     name: &mut MaybeUninit<libc::sockaddr_storage>,
-    ctrl: &mut cmsg::Aligned<MaybeUninit<[u8; CMSG_LEN]>>,
+    ctrl: &mut cmsg::Aligned<[u8; CMSG_LEN]>,
     hdr: &mut msghdr_x,
 ) {
     hdr.msg_name = name.as_mut_ptr() as _;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1608,7 +1608,7 @@ dependencies = [
 
 [[package]]
 name = "quinn-udp"
-version = "0.5.11"
+version = "0.5.12"
 dependencies = [
  "cfg_aliases",
  "criterion",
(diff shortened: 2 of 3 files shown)


quinn-udp/src/unix.rs
```diff
@@ -506,7 +506,7 @@ fn recv(io: SockRef<'_>, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) ->
 #[cfg(apple_fast)]
 fn recv(io: SockRef<'_>, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) -> io::Result<usize> {
     let mut names = [MaybeUninit::<libc::sockaddr_storage>::uninit(); BATCH_SIZE];
-    let mut ctrls = [cmsg::Aligned(MaybeUninit::<[u8; CMSG_LEN]>::uninit()); BATCH_SIZE];
+    let mut ctrls = [cmsg::Aligned([0u8; CMSG_LEN]); BATCH_SIZE];
```
mxinden, line 509: With `CMSG_LEN` being `88` and `BATCH_SIZE` being `32`, this is not a small allocation. I will try to run a benchmark on a recent Mac to see whether it has a performance impact.

thomaseizinger: `msghdr_x` below on line 510 is already 56 bytes and also an array of `BATCH_SIZE`. Not quite the same size but same ballpark.

Can we use `mem::zeroed` for the `ctrl`s array as well? Would that be faster?

Maybe we can leave it at `MaybeUninit` and only initialise up to `bufs.len()`?

thomaseizinger: If I understand this code correctly, then `ctrls` already takes up 2kb of stack space, even with the use of `MaybeUninit`. So the only additional cost here is zeroing out the 2kb which I'd think isn't going to be that expensive.

djc: Would be good to do some basic benchmarking either way.

larseggert: `msghdr_x` has an extra field `msg_datalen` (compared to `msghdr`): https://github.com/apple/darwin-xnu/blob/2ff845c2e033bd0ff64b5b6aa6063a1f8f65aa32/bsd/sys/socket.h#L849

Pretty sure that isn't useful here, but wanted to mention it just in case.

mxinden:
Below is the output of the `quinn-udp` benchmarks, comparing `main` with this pull request (`zeroed`), on an Apple M2:

```
critcmp main zeroed                                                 
group                                            main                                   zeroed
-----                                            ----                                   ------
gso_false_gro_false_recvmmsg_false/throughput    1.00     81.0±0.54ms   123.4 MB/sec    1.00     81.3±0.63ms   123.0 MB/sec
gso_false_gro_false_recvmmsg_true/throughput     1.01     81.8±0.45ms   122.3 MB/sec    1.00     81.2±0.54ms   123.2 MB/sec
gso_false_gro_true_recvmmsg_false/throughput     1.00     81.6±1.77ms   122.6 MB/sec    1.00     81.6±0.65ms   122.6 MB/sec
gso_false_gro_true_recvmmsg_true/throughput      1.00     81.8±0.51ms   122.2 MB/sec    1.01     82.6±0.67ms   121.1 MB/sec
gso_true_gro_false_recvmmsg_false/throughput     1.00     21.8±0.13ms   458.2 MB/sec    1.02     22.2±0.12ms   450.2 MB/sec
gso_true_gro_false_recvmmsg_true/throughput      1.00     17.1±0.06ms   584.8 MB/sec    1.00     17.1±0.08ms   583.7 MB/sec
gso_true_gro_true_recvmmsg_false/throughput      1.00     21.8±0.11ms   458.0 MB/sec    1.02     22.3±0.12ms   449.1 MB/sec
gso_true_gro_true_recvmmsg_true/throughput       1.00     17.1±0.07ms   585.7 MB/sec    1.00     17.2±0.07ms   583.0 MB/sec
```

Consecutive runs of the benchmark show noise within < 5% both on `main` and this pull request.

With that, I believe it is safe to assume that this pull request does not introduce a performance regression.

quinn-udp/src/unix.rs
Ralith, line 509: Please leave a comment explaining why this is necessary.

Ralith (approved):
Nice catch!

djc (approved):

quinn-udp/src/unix.rs
mxinden: Of course. Sorry about that. Done.

mxinden:
Thanks for all the help everyone. Much appreciated!

This is ready to merge from my end.

thomaseizinger:
It would be great if we could include #2217 in the upcoming release.

djc:
> It would be great if we could include #2217 in the upcoming release.

Yup, was thinking the same thing.

PR #2220, opened 2025-05-08
feat: Make the future returned by `SendStream::stopped` static

Frando:
Currently, the future returned by `SendStream::stopped` is bound to `&mut self`, and therefore cannot be stored and polled while also sending on the stream. This makes the future returned by `SendStream::stopped` be `'static`, so that it can be stored in futures combinators while also still sending on the stream.

This is useful when you want to clean up local state once the corresponding receiver to a send stream is closed or dropped, independently of actually using the stream. Especially if sends occur infrequently, tracking the stopped notification allows to clean up state at the earliest time and not only once you try to send again.

There's a complication though: Once `SendStream::stopped` returns a `'static` future, we'd have to account for the fact that there may be multiple futures waiting for a wakeup once the stream is stopped. Therefore, we'd need to store a *list* of wakers instead of a single waker per stream. I used a `Slab`, open to other ideas how to achieve this.

For the intended functionality I don't actually need the possibility to create multiple futures waiting for the stopped wakeup. However, I didn't yet come up with an API that encodes that only the last future returned from `SendStream::stopped` would be woken up reliably - if you have ideas here, please share.

I added two tests with various combinations of using stopped while stopping and/or dropping the send stream.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -12,6 +12,7 @@ use std::{
 use bytes::Bytes;
 use pin_project_lite::pin_project;
 use rustc_hash::FxHashMap;
+use slab::Slab;
 use thiserror::Error;
 use tokio::sync::{Notify, futures::Notified, mpsc, oneshot};
 use tracing::{Instrument, Span, debug_span};
@@ -963,7 +964,7 @@ pub(crate) struct State {
     endpoint_events: mpsc::UnboundedSender<(ConnectionHandle, EndpointEvent)>,
     pub(crate) blocked_writers: FxHashMap<StreamId, Waker>,
     pub(crate) blocked_readers: FxHashMap<StreamId, Waker>,
-    pub(crate) stopped: FxHashMap<StreamId, Waker>,
+    pub(crate) stopped: FxHashMap<StreamId, Slab<Waker>>,
     /// Always set to Some before the connection becomes drained
     pub(crate) error: Option<ConnectionError>,
     /// Number of live handles that can be used to initiate or handle I/O; excludes the driver
@@ -1105,7 +1106,7 @@ impl State {
                         // `ZeroRttRejected` errors.
                         wake_all(&mut self.blocked_writers);
                         wake_all(&mut self.blocked_readers);
-                        wake_all(&mut self.stopped);
+                        wake_all_slab(&mut self.stopped);
                     }
                 }
                 ConnectionLost { reason } => {
@@ -1129,9 +1130,9 @@ impl State {
                     // Might mean any number of streams are ready, so we wake up everyone
                     shared.stream_budget_available[dir as usize].notify_waiters();
                 }
-                Stream(StreamEvent::Finished { id }) => wake_stream(id, &mut self.stopped),
+                Stream(StreamEvent::Finished { id }) => wake_stream_slab(id, &mut self.stopped),
                 Stream(StreamEvent::Stopped { id, .. }) => {
-                    wake_stream(id, &mut self.stopped);
+                    wake_stream_slab(id, &mut self.stopped);
                     wake_stream(id, &mut self.blocked_writers);
                 }
             }
@@ -1212,7 +1213,7 @@ impl State {
         if let Some(x) = self.on_connected.take() {
             let _ = x.send(false);
         }
-        wake_all(&mut self.stopped);
+        wake_all_slab(&mut self.stopped);
         shared.closed.notify_waiters();
     }
 
@@ -1264,6 +1265,18 @@ fn wake_stream(stream_id: StreamId, wakers: &mut FxHashMap<StreamId, Waker>) {
 
 fn wake_all(wakers: &mut FxHashMap<StreamId, Waker>) {
     wakers.drain().for_each(|(_, waker)| waker.wake())
+}
+
+fn wake_stream_slab(stream_id: StreamId, wakers: &mut FxHashMap<StreamId, Slab<Waker>>) {
+    if let Some(mut slab) = wakers.remove(&stream_id) {
+        slab.drain().for_each(|waker| waker.wake())
+    }
+}
+
+fn wake_all_slab(wakers: &mut FxHashMap<StreamId, Slab<Waker>>) {
+    wakers
+        .drain()
+        .for_each(|(_, mut slab)| slab.drain().for_each(|waker| waker.wake()))
 }
 
 /// Errors that can arise when sending a datagram
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -199,28 +199,12 @@ impl SendStream {
     /// For a variety of reasons, the peer may not send acknowledgements immediately upon receiving
     /// data. As such, relying on `stopped` to know when the peer has read a stream to completion
     /// may introduce more latency than using an application-level response of some sort.
-    pub async fn stopped(&mut self) -> Result<Option<VarInt>, StoppedError> {
-        Stopped { stream: self }.await
-    }
-
-    fn poll_stopped(&mut self, cx: &mut Context) -> Poll<Result<Option<VarInt>, StoppedError>> {
-        let mut conn = self.conn.state.lock("SendStream::poll_stopped");
-
-        if self.is_0rtt {
-            conn.check_0rtt()
-                .map_err(|()| StoppedError::ZeroRttRejected)?;
-        }
-
-        match conn.inner.send_stream(self.stream).stopped() {
-            Err(_) => Poll::Ready(Ok(None)),
-            Ok(Some(error_code)) => Poll::Ready(Ok(Some(error_code))),
-            Ok(None) => {
-                if let Some(e) = &conn.error {
-                    return Poll::Ready(Err(e.clone().into()));
-                }
-                conn.stopped.insert(self.stream, cx.waker().clone());
-                Poll::Pending
-            }
+    pub fn stopped(&self) -> Stopped {
+        Stopped {
+            conn: self.conn.clone(),
+            stream: self.stream,
+            is_0rtt: self.is_0rtt,
+            waker_key: None,
         }
     }
 
@@ -303,15 +287,46 @@ impl Drop for SendStream {
 }
 
 /// Future produced by `SendStream::stopped`
-struct Stopped<'a> {
-    stream: &'a mut SendStream,
+pub struct Stopped {
+    conn: ConnectionRef,
+    stream: StreamId,
+    is_0rtt: bool,
+    waker_key: Option<usize>,
 }
 
-impl Future for Stopped<'_> {
+impl Stopped {
+    fn poll_unpin(&mut self, cx: &mut Context) -> Poll<<Self as Future>::Output> {
+        let mut conn = self.conn.state.lock("SendStream::poll_stopped");
+
+        if self.is_0rtt {
+            conn.check_0rtt()
+                .map_err(|()| StoppedError::ZeroRttRejected)?;
+        }
+
+        match conn.inner.send_stream(self.stream).stopped() {
+            Err(_) => Poll::Ready(Ok(None)),
+            Ok(Some(error_code)) => Poll::Ready(Ok(Some(error_code))),
+            Ok(None) => {
+                if let Some(e) = &conn.error {
+                    return Poll::Ready(Err(e.clone().into()));
+                }
+                let slab = conn.stopped.entry(self.stream).or_default();
+                if let Some(key) = self.waker_key.take() {
+                    let _ = slab.try_remove(key);
+                }
+                let key = slab.insert(cx.waker().clone());
+                self.waker_key = Some(key);
+                Poll::Pending
+            }
+        }
+    }
+}
+
+impl Future for Stopped {
     type Output = Result<Option<VarInt>, StoppedError>;
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
-        self.get_mut().stream.poll_stopped(cx)
+        self.get_mut().poll_unpin(cx)
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1592,6 +1592,7 @@ dependencies = [
  "rustc-hash 2.1.1",
  "rustls",
  "rustls-pemfile",
+ "slab",
  "smol",
  "socket2",
  "thiserror 2.0.12",
--- a/quinn/Cargo.toml
+++ b/quinn/Cargo.toml
@@ -54,6 +54,7 @@ rustc-hash = { workspace = true }
 pin-project-lite = { workspace = true }
 proto = { package = "quinn-proto", path = "../quinn-proto", version = "0.11.10", default-features = false }
 rustls = { workspace = true, optional = true }
+slab = { workspace = true }
 smol = { workspace = true, optional = true }
 thiserror = { workspace = true }
 tracing =  { workspace = true }
--- a/quinn/src/tests.rs
+++ b/quinn/src/tests.rs
@@ -861,3 +861,87 @@ async fn multiple_conns_with_zero_length_cids() {
     .instrument(error_span!("server"));
     tokio::join!(client1, client2, server);
 }
+
+#[tokio::test]
+async fn stream_stopped() {
+    let _guard = subscribe();
+    let factory = EndpointFactory::new();
+    let server = {
+        let _guard = error_span!("server").entered();
+        factory.endpoint()
+    };
+    let server_addr = server.local_addr().unwrap();
+
+    let client = {
+        let _guard = error_span!("client1").entered();
+        factory.endpoint()
+    };
+
+    let client = async move {
+        let conn = client
+            .connect(server_addr, "localhost")
+            .unwrap()
+            .await
+            .unwrap();
+        let mut stream = conn.open_uni().await.unwrap();
+        let stopped1 = stream.stopped();
+        let stopped2 = stream.stopped();
+        let stopped3 = stream.stopped();
+
+        stream.write_all(b"hi").await.unwrap();
+        // spawn one of the futures into a task
+        let stopped1 = tokio::task::spawn(stopped1);
+        // verify that both futures resolved
+        let (stopped1, stopped2) = tokio::join!(stopped1, stopped2);
+        assert!(matches!(stopped1, Ok(Ok(Some(val))) if val == 42u32.into()));
+        assert!(matches!(stopped2, Ok(Some(val)) if val == 42u32.into()));
+        // drop the stream
+        drop(stream);
+        // verify that a future also resolves after dropping the stream
+        let stopped3 = stopped3.await;
+        assert_eq!(stopped3, Ok(Some(42u32.into())));
+    };
+    let client =
+        tokio::time::timeout(Duration::from_millis(100), client).instrument(error_span!("client"));
+    let server = async move {
+        let conn = server.accept().await.unwrap().await.unwrap();
+        let mut stream = conn.accept_uni().await.unwrap();
+        let mut buf = [0u8; 2];
+        stream.read_exact(&mut buf).await.unwrap();
+        stream.stop(42u32.into()).unwrap();
+        conn
+    }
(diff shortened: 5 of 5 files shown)


quinn/src/send_stream.rs
```diff
@@ -283,7 +299,6 @@ impl Drop for SendStream {
         let mut conn = self.conn.state.lock("SendStream::drop");
 
         // clean up any previously registered wakers
-        conn.stopped.remove(&self.stream);
```
Ralith, line 286: Why is this removed?

Frando: If we'd remove the `Notify` from `conn.stopped` when dropping the send stream, a future previously returned from `SendStream::stopped` would remain pending forever.
i.e.
```rust
let stopped = send_stream.stopped();
drop(send_stream);
stopped.await; // would never yield
```

The `Notify` in `conn.stopped` should be cleared when `Event::Stream(StreamEvent::Finished)` is handled in `quinn::connection::State::forward_app_events` (or at least that is my understanding, I'm not deeply invested in these codepaths yet).

Frando: We could remove the notify here and trigger the wakeup though. I think it should be equal to notifiying from `forward_app_events` on the `StreamEvent::Finished`?

Ralith: No, I think the way you've got it now is correct -- the event should still work just fine after the stream handle is dropped.

quinn/src/send_stream.rs
```diff
@@ -245,6 +237,30 @@ impl SendStream {
...
+            if let Some(e) = &conn.error {
+                Break(Err(e.clone().into()))
+            } else {
+                let notify = conn.stopped.entry(stream).or_default().clone();
+                Continue(notify)
+            }
+        }
+    }
+}
+
```
gretchenfrage, line 263: Is a loop necessary?

I don't see any documentation on `Notify` that indicates it can experience spurious wakeups. This makes it simpler than using, say, a condvar. Shouldn't you be able to just do something like:

```pseudocode
async move {
    let notify_closed = {
        let lock = state.lock()
        if lock.closed() { return } else { Arc::clone(&lock.notify_closed) }
    }
    notify_closed.notified().await
}
```

gretchenfrage (changes requested):

Ralith (approved):

quinn/src/send_stream.rs
```diff
@@ -199,27 +202,16 @@ impl SendStream {
...
-                    return Poll::Ready(Err(e.clone().into()));
+    pub fn stopped(
+        &self,
+    ) -> impl Future<Output = Result<Option<VarInt>, StoppedError>> + Send + Sync + 'static {
+        let (conn, stream, is_0rtt) = (self.conn.clone(), self.stream, self.is_0rtt);
+        async move {
+            loop {
+                match stopped_or_notify(&conn, stream, is_0rtt) {
+                    ControlFlow::Break(res) => return res,
+                    ControlFlow::Continue(notify) => notify.notified().await,
```
Ralith, line 213: On review, this is racy. `notify_waiters` only wakes currently registered tasks, but might be called in between the lock being released and `notified` being called. The `Notified` future *must* be constructed (but not polled!) before the connection lock is released.

Ralith (changes requested):

quinn/src/send_stream.rs
Ralith: We need to inspect the connection state to determine what to return, and with a loop we can easily do that with the same code that runs before the future is constructed. You're probably right that we'll never complete a second iteration, but that's fine.

gretchenfrage: I was thinking something like this?

```diff
diff --git a/quinn/src/mutex.rs b/quinn/src/mutex.rs
index 7c24df46..df44b20f 100644
--- a/quinn/src/mutex.rs
+++ b/quinn/src/mutex.rs
@@ -110,7 +110,7 @@ mod tracking {
 }
 
 #[cfg(feature = "lock_tracking")]
-pub(crate) use tracking::Mutex;
+pub(crate) use tracking::{Mutex, MutexGuard};
 
 #[cfg(not(feature = "lock_tracking"))]
 mod non_tracking {
@@ -160,4 +160,4 @@ mod non_tracking {
 }
 
 #[cfg(not(feature = "lock_tracking"))]
-pub(crate) use non_tracking::Mutex;
+pub(crate) use non_tracking::{Mutex, MutexGuard};
diff --git a/quinn/src/send_stream.rs b/quinn/src/send_stream.rs
index 16125cbc..607f1b3a 100644
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -1,7 +1,6 @@
 use std::{
     future::Future,
     io,
-    ops::ControlFlow,
     pin::Pin,
     sync::Arc,
     task::{Context, Poll, ready},
@@ -10,9 +9,12 @@ use std::{
 use bytes::Bytes;
 use proto::{ClosedStream, ConnectionError, FinishError, StreamId, Written};
 use thiserror::Error;
-use tokio::sync::Notify;
 
-use crate::{VarInt, connection::ConnectionRef};
+use crate::{
+    VarInt,
+    connection::{ConnectionRef, State},
+    mutex::MutexGuard,
+};
 
 /// A stream that can only be used to send data
 ///
@@ -205,15 +207,7 @@ impl SendStream {
     pub fn stopped(
         &self,
     ) -> impl Future<Output = Result<Option<VarInt>, StoppedError>> + Send + Sync + 'static {
-        let (conn, stream, is_0rtt) = (self.conn.clone(), self.stream, self.is_0rtt);
-        async move {
-            loop {
-                match stopped_or_notify(&conn, stream, is_0rtt) {
-                    ControlFlow::Break(res) => return res,
-                    ControlFlow::Continue(notify) => notify.notified().await,
-                }
-            }
-        }
+        bikeshed1(self.conn.clone(), self.stream, self.is_0rtt)
     }
 
     /// Get the identity of this stream
@@ -237,28 +231,37 @@ impl SendStream {
     }
 }
 
-fn stopped_or_notify(
-    conn: &ConnectionRef,
+async fn bikeshed1(
+    conn: ConnectionRef,
     stream: StreamId,
     is_0rtt: bool,
-) -> ControlFlow<Result<Option<VarInt>, StoppedError>, Arc<Notify>> {
-    use ControlFlow::*;
-    let mut conn = conn.state.lock("SendStream::stopped");
-    if is_0rtt && conn.check_0rtt().is_err() {
-        return Break(Err(StoppedError::ZeroRttRejected));
-    }
-    match conn.inner.send_stream(stream).stopped() {
-        Err(ClosedStream { .. }) => Break(Ok(None)),
-        Ok(Some(error_code)) => Break(Ok(Some(error_code))),
-        Ok(None) => {
-            if let Some(e) = &conn.error {
-                Break(Err(e.clone().into()))
-            } else {
-                let notify = conn.stopped.entry(stream).or_default().clone();
-                Continue(notify)
-            }
+) -> Result<Option<VarInt>, StoppedError> {
+    {
+        let mut guard = conn.state.lock("SendStream::stopped");
+        if let Some(result) = bikeshed2(&mut guard, stream, is_0rtt) {
+            return result;
+        } else {
+            Arc::clone(&guard.stopped.entry(stream).or_default())
         }
     }
+    .notified()
+    .await;
+    bikeshed2(&mut conn.state.lock("SendStream::stopped"), stream, is_0rtt).unwrap()
+}
+
+fn bikeshed2(
+    guard: &mut MutexGuard<State>,
+    stream: StreamId,
+    is_0rtt: bool,
+) -> Option<Result<Option<VarInt>, StoppedError>> {
+    if is_0rtt && guard.check_0rtt().is_err() {
+        return Some(Err(StoppedError::ZeroRttRejected));
+    }
+    match guard.inner.send_stream(stream).stopped() {
+        Err(ClosedStream { .. }) => Some(Ok(None)),
+        Ok(Some(error_code)) => Some(Ok(Some(error_code))),
+        Ok(None) => guard.error.clone().map(|e| Err(e.into())),
+    }
 }
 
 #[cfg(feature = "futures-io")]

```

quinn/src/send_stream.rs
Frando: Good catch. I refactored it so that the `Notified` future is created while the lock is still being held. Due to `Notified` being lifetime-bound to `Notify`, this needed a manual `super let` construction, see the comment in the code.

quinn/src/send_stream.rs
gretchenfrage: (Sorry if you already read the original version of my above comment, I tweaked it a bit). That diff reduces lines of code, removes the loop than will only loop either 1 or 2 times, and removes the need to import ControlFlow.

Frando: Your version features, I think, the same issue Ralith mentioned below: The `Notified` future needs to be created *before* dropping the lock. In your code it is created after the lock is dropped.

Here's my version without a loop:
https://github.com/quinn-rs/quinn/commit/2cb4eba62ebaba8a892680771bb80f24baeba684

I don't have a strong preference between the `loop` (as in the PR atm) and the `expect` (as in the linked commit) versions. Feel free to decide together with @Ralith - if you prefer the non-loop version, I'll push the above commit to this branch.

Ralith (approved):

Ralith:
This is technically a breaking change since we're going from `&mut self` to `&self`, though it's the sort of esoteric breakage we could theoretically sneak through.

quinn/src/send_stream.rs
gretchenfrage: Upon closer inspection of the surrounding code, I think the loop is necessary after all (e.g. `forward_app_events` blindly wakes this future for all streams when 0-RTT is rejected, in case any of them are 0-RTT streams). However if it weren't for that I think it would have been better to go with the option that is simpler and smaller.

gretchenfrage: Upon closer inspection of the surrounding code, I think the loop is necessary after all (e.g. forward_app_events blindly wakes this future for all streams when 0-RTT is rejected, in case any of them are 0-RTT streams).

quinn/src/send_stream.rs
```diff
@@ -199,27 +205,27 @@ impl SendStream {
...
-        match conn.inner.send_stream(self.stream).stopped() {
-            Err(_) => Poll::Ready(Ok(None)),
-            Ok(Some(error_code)) => Poll::Ready(Ok(Some(error_code))),
-            Ok(None) => {
-                if let Some(e) = &conn.error {
-                    return Poll::Ready(Err(e.clone().into()));
+    pub fn stopped(
+        &self,
+    ) -> impl Future<Output = Result<Option<VarInt>, StoppedError>> + Send + Sync + 'static {
+        let (conn, stream, is_0rtt) = (self.conn.clone(), self.stream, self.is_0rtt);
```
gretchenfrage, line 211: I'd prefer this be split onto 3 lines

quinn/src/send_stream.rs
```diff
@@ -199,27 +205,27 @@ impl SendStream {
...
+        async move {
+            loop {
+                // The `Notify::notified` future needs to be created while the lock is being
+                // held, otherwise a wakeup could be missed if triggered inbetween releasing the lock
+                // and creating the future. The lock may only be held in a block without `await`s,
+                // otherwise the future becomes `!Send`. `Notify::notified` is lifetime-bound to `Notify`,
```
gretchenfrage, line 217: Would you mind limiting to 100 columns

gretchenfrage (changes requested):
I understand now that the loop is necessary. Btw, I really like how you included tests for this.

I still think it can simplify things to have the helper function only try to return the `Poll::Ready` value, and not also have it saddled with trying to clone the `Notify`: 

```diff
diff --git a/quinn/src/mutex.rs b/quinn/src/mutex.rs
index 7c24df46..df44b20f 100644
--- a/quinn/src/mutex.rs
+++ b/quinn/src/mutex.rs
@@ -110,7 +110,7 @@ mod tracking {
 }
 
 #[cfg(feature = "lock_tracking")]
-pub(crate) use tracking::Mutex;
+pub(crate) use tracking::{Mutex, MutexGuard};
 
 #[cfg(not(feature = "lock_tracking"))]
 mod non_tracking {
@@ -160,4 +160,4 @@ mod non_tracking {
 }
 
 #[cfg(not(feature = "lock_tracking"))]
-pub(crate) use non_tracking::Mutex;
+pub(crate) use non_tracking::{Mutex, MutexGuard};
diff --git a/quinn/src/send_stream.rs b/quinn/src/send_stream.rs
index 759734ea..87581dc8 100644
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -1,7 +1,6 @@
 use std::{
     future::Future,
     io,
-    ops::ControlFlow,
     pin::Pin,
     sync::Arc,
     task::{Context, Poll, ready},
@@ -10,11 +9,11 @@ use std::{
 use bytes::Bytes;
 use proto::{ClosedStream, ConnectionError, FinishError, StreamId, Written};
 use thiserror::Error;
-use tokio::sync::Notify;
 
 use crate::{
     VarInt,
     connection::{ConnectionRef, State},
+    mutex::MutexGuard,
 };
 
 /// A stream that can only be used to send data
@@ -219,10 +218,10 @@ impl SendStream {
                 let notify;
                 {
                     let mut conn = conn.state.lock("SendStream::stopped");
-                    notify = match stopped_or_notify(&mut conn, stream, is_0rtt) {
-                        ControlFlow::Break(res) => return res,
-                        ControlFlow::Continue(notify) => notify,
-                    };
+                    if let Some(result) = stopped_output(&mut conn, stream, is_0rtt) {
+                        break result;
+                    }
+                    notify = Arc::clone(&conn.stopped.entry(stream).or_default());
                     notify.notified()
                 }
                 .await
@@ -251,26 +250,18 @@ impl SendStream {
     }
 }
 
-fn stopped_or_notify(
-    conn: &mut State,
+fn stopped_output(
+    conn: &mut MutexGuard<State>,
     stream: StreamId,
     is_0rtt: bool,
-) -> ControlFlow<Result<Option<VarInt>, StoppedError>, Arc<Notify>> {
-    use ControlFlow::*;
+) -> Option<Result<Option<VarInt>, StoppedError>> {
     if is_0rtt && conn.check_0rtt().is_err() {
-        return Break(Err(StoppedError::ZeroRttRejected));
+        return Some(Err(StoppedError::ZeroRttRejected));
     }
     match conn.inner.send_stream(stream).stopped() {
-        Err(ClosedStream { .. }) => Break(Ok(None)),
-        Ok(Some(error_code)) => Break(Ok(Some(error_code))),
-        Ok(None) => {
-            if let Some(e) = &conn.error {
-                Break(Err(e.clone().into()))
-            } else {
-                let notify = conn.stopped.entry(stream).or_default().clone();
-                Continue(notify)
-            }
-        }
+        Err(ClosedStream { .. }) => Some(Ok(None)),
+        Ok(Some(error_code)) => Some(Ok(Some(error_code))),
+        Ok(None) => conn.error.clone().map(|e| Err(e.into())),
     }
 }
 
```

</details>

(Consider that this is net-negative 9 LOC, avoids needing to import control flow, and allows an if-let-else chain to be replaced a simple option map.)

quinn/src/send_stream.rs
Frando: Done

quinn/src/send_stream.rs
Frando: Done

Frando:
I pushed another commit that removes the `ControlFlow` and instead returns `Option` as suggested by @gretchenfrage.

Frando:
I also rebased onto `main` (had conflicts in the import section), combined the commits into one and added a proper commit message.

quinn/src/connection.rs
```diff
@@ -963,7 +963,7 @@ pub(crate) struct State {
     endpoint_events: mpsc::UnboundedSender<(ConnectionHandle, EndpointEvent)>,
     pub(crate) blocked_writers: FxHashMap<StreamId, Waker>,
     pub(crate) blocked_readers: FxHashMap<StreamId, Waker>,
-    pub(crate) stopped: FxHashMap<StreamId, Waker>,
+    pub(crate) stopped: FxHashMap<StreamId, Arc<Notify>>,
```
djc, line 966: Nit: can you extract this and the related changes into a separate, earlier commit (that still compiles and passes tests at least locally)?

quinn/src/send_stream.rs
```diff
@@ -245,6 +252,32 @@ impl SendStream {
...
+///
+/// Returns `Some` if the stream is stopped or the connection is closed.
+/// Returns `None` if the stream is not stopped.
+fn send_stream_stopped(
+    conn: &mut State,
+    stream: StreamId,
+    is_0rtt: bool,
+) -> Option<Result<Option<VarInt>, StoppedError>> {
+    if is_0rtt && conn.check_0rtt().is_err() {
+        Some(Err(StoppedError::ZeroRttRejected))
```
djc, line 265: Nit: use an early `return` for this so we can reduce rightward drift for the other branch.

quinn/src/send_stream.rs
```diff
@@ -245,6 +252,32 @@ impl SendStream {
...
+    } else {
+        match conn.inner.send_stream(stream).stopped() {
+            Err(ClosedStream { .. }) => Some(Ok(None)),
+            Ok(Some(error_code)) => Some(Ok(Some(error_code))),
+            Ok(None) => {
+                if let Some(error) = &conn.error {
```
djc, line 271: I think this could be a `map()` call? `conn.error.cloned().map(Into::into)` maybe?

djc:
> This is technically a breaking change since we're going from `&mut self` to `&self`, though it's the sort of esoteric breakage we could theoretically sneak through.

Not sure "esoteric" is the right word here, but I think rustc will gladly allow reborrowing `&mut` as `&` so I guess it will be mostly fine?

quinn/src/connection.rs
Frando: Not sure I understand, what part of the overall commit should *not* be included in that commit then?

quinn/src/send_stream.rs
Frando: I can do, but I actually just changed it to that form 30min ago because I preferred the if/else branching to be visible. But your call ofc.

quinn/src/send_stream.rs
Frando: It could, again I preferred this form because for me it reads much clearer this way with the quite complex return type. For me, the `if let` is very straightforward to read, where as the `cloned().map(Into::into)` I have to take two or three looks to be sure what exactly is going on. But again, your call ofc.

quinn/src/send_stream.rs
djc: Yes, I think we generally prefer early returns.

quinn/src/send_stream.rs
Frando: Ah well, clippy complains about my manual if-let, so changing it to the map.

quinn/src/connection.rs
djc: I think the changes to the `stopped()` API can be in a separate commit from the internal API changes?

Frando: Not sure how exactly that would work. If I change `State::stopped` to `Arc<Notify>`, I have to change `SendStream::poll_stopped`  as well, which is possible of course but not done in a minute because the `Notified` future has to be stored and pinned and polled in a loop. I can do but is this really worth it if the correct solution is already implemented?

djc: Okay, let's leave it in a single commit.

quinn/src/send_stream.rs
```diff
@@ -245,6 +252,25 @@ impl SendStream {
     }
 }
 
+/// Check if a send stream is stopped.
```
gretchenfrage, line 255: Project convention is for the first line to not have a trailing period. This is private though so that's even more of a nitpick than usual.

gretchenfrage (commented):
Thanks! Haven't re-looked in detail, but leaving a comment review which should clear my "changes requested" state. The fact that the other maintainers approved is sufficient.

gretchenfrage:
And it seems like we all agree that we're going to consider this not to be semver breaking in a way that matters

Issue #2221, opened 2025-05-08
`MAX_TRANSMIT_SEGMENTS` is ignored in the latest `quinn` version


dlon:
https://github.com/quinn-rs/quinn/commit/f8165c339483a09204514377c430579ceb6509e5 moved out the limit on number of datagrams per transmit to `quinn` (instead of `quinn-proto`). `0.11.11` was the first version of `quinn-proto` to include this change, but `quinn` itself still doesn't have it. So if you're on the latest published version of all crates, `MAX_TRANSMIT_SEGMENTS` is never used.

This seems to have caused a performance regression for us when we updated `quinn-proto` recently. Could a new version of `quinn` be released to fix this?

djc:
Ahh, sorry about that. Can you submit a version bump PR for quinn?

dlon:
Thanks!

PR #2222, opened 2025-05-08
Bump quinn version to 0.11.8

dlon:
Bump the version of `quinn` so that #2221 may be fixed. Feel free to ping me if you need me to do anything else.


djc (approved):

djc:
Thanks! Will handle the release in #2223.

PR #2226, opened 2025-05-10
quinn: Remove explicit write future structs

gretchenfrage:
We maintain explicit send stream future implementation structs for `Write`, `WriteAll`, `WriteChunk`, and `WriteAllChunks`. However, these are private. We can significantly simplify this code by instead leaning more directly on functionality such as async functions and `std::future::poll_fn`.

### DIFF
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -1,8 +1,8 @@
 use std::{
-    future::Future,
+    future::{Future, poll_fn},
     io,
     pin::Pin,
-    task::{Context, Poll, ready},
+    task::{Context, Poll},
 };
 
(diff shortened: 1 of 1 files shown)


Ralith (approved):
Great cleanup! As you might guess, the deleted code here is very old, but I didn't realize how much more verbose it was. I'm glad we took the step of hiding all the named futures a while back to make this seamless.

djc:
Pretty sure this needs to be tagged as breaking?

gretchenfrage:
> Pretty sure this needs to be tagged as breaking?

Why?

djc (approved):
Ah no, I was wrong, sorry!

PR #2227, opened 2025-05-10
docs(quinn): Revise SendStream::write docs

gretchenfrage:
Aligns with 100 columns, attempts to improve clarity, takes inspiration from tokio::io::AsyncWriteExt docs.

### DIFF
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -43,53 +43,80 @@ impl SendStream {
         }
     }
 
-    /// Write bytes to the stream
+    /// Attempts to write a buffer into this stream, returning how many bytes were written
     ///
-    /// Yields the number of bytes written on success. Congestion and flow control may cause this to
-    /// be shorter than `buf.len()`, indicating that only a prefix of `buf` was written.
+    /// Unless this method errors, it waits until some amount of `buf` can be written into this
+    /// stream, and then writes as much as it can without waiting again. Due to congestion and flow
+    /// control, this may be shorter than `buf.len()`. On success this resolves to the length of
+    /// the prefix that was written.
     ///
-    /// This operation is cancel-safe.
+    /// This method is cancellation safe: If it does not resolve, no bytes were written.
     pub async fn write(&mut self, buf: &[u8]) -> Result<usize, WriteError> {
         Write { stream: self, buf }.await
     }
 
-    /// Convenience method to write an entire buffer to the stream
+    /// Attempts to write an buffer into this stream in its entirety
     ///
-    /// This operation is *not* cancel-safe.
+    /// This method repeatedly calls [`write`](Self::write) until all bytes are written, or an
+    /// error occurs.
+    ///
+    /// This method is *not* cancellation safe: Even if this does not resolve, some prefix of `buf`
+    /// may have been written when previously polled.
     pub async fn write_all(&mut self, buf: &[u8]) -> Result<(), WriteError> {
         WriteAll { stream: self, buf }.await
     }
 
-    /// Write chunks to the stream
+    /// Like [`write`](Self::write), except that it writes from a slice of [`Bytes`]
     ///
-    /// Yields the number of bytes and chunks written on success.
-    /// Congestion and flow control may cause this to be shorter than `buf.len()`,
-    /// indicating that only a prefix of `bufs` was written
+    /// Bytes to try to write are provided to this method as an array of cheaply cloneable chunks.
+    /// Unless this method errors, it waits until some amount of those bytes can be written into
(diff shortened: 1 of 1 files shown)


Ralith:
Initial bikeshedding:
- Why introduce "resolves to" in addition to "yields"?
- Why replace "cancel-safe" with "cancellation safe"?

quinn/src/send_stream.rs
```diff
@@ -43,53 +43,80 @@ impl SendStream {
...
     ///
-    /// Yields the number of bytes written on success. Congestion and flow control may cause this to
-    /// be shorter than `buf.len()`, indicating that only a prefix of `buf` was written.
+    /// Unless this method errors, it waits until some amount of `buf` can be written into this
+    /// stream, and then writes as much as it can without waiting again. Due to congestion and flow
+    /// control, this may be shorter than `buf.len()`. On success this resolves to the length of
+    /// the prefix that was written.
     ///
-    /// This operation is cancel-safe.
+    /// This method is cancellation safe: If it does not resolve, no bytes were written.
```
djc, line 53: Nit: I don't think we should capitalize after a colon?

quinn/src/send_stream.rs
```diff
@@ -43,53 +43,80 @@ impl SendStream {
...
+    /// error occurs.
+    ///
+    /// This method is *not* cancellation safe: Even if this does not resolve, some prefix of `buf`
+    /// may have been written when previously polled.
     pub async fn write_all(&mut self, buf: &[u8]) -> Result<(), WriteError> {
         WriteAll { stream: self, buf }.await
     }
 
-    /// Write chunks to the stream
+    /// Like [`write`](Self::write), except that it writes from a slice of [`Bytes`]
```
djc, line 69: I don't like these formulations for the one-line description; I think they're relying too much on indirection. It could be useful to explain the functionality in relation to `Self::write()`, but not in the first line.

quinn/src/send_stream.rs
```diff
@@ -43,53 +43,80 @@ impl SendStream {
...
-    pub async fn write_chunk(&mut self, buf: Bytes) -> Result<(), WriteError> {
-        WriteChunk {
+    /// Bytes to write are provided to this method as an array of cheaply cloneable chunks. This
+    /// method repeatedly calls [`write_chunks`](Self::write_chunks) until all bytes are written,
+    /// or an error occurs. This method mutates `bufs` by mutating all chunks to be
+    /// [empty](Bytes::is_empty).
+    ///
+    /// This method is *not* cancellation safe: Even if this does not resolve, some prefix of `buf`
+    /// may have been written when previously polled.
+    pub async fn write_all_chunks(&mut self, bufs: &mut [Bytes]) -> Result<(), WriteError> {
```
djc, line 99: Nit: don't reorder items in the same commit as you're rewriting the docs.

quinn/src/send_stream.rs
flub, line 53: nit: tokio uses a `# Cancel safety` heading for these.  Which makes it stand out still a bit more.  I'd be +1 to adopt that convention.

flub:
>     * Why replace "cancel-safe" with "cancellation safe"?

This mirrors how tokio talks about it.  They use a section header named `Cancel safety` and in prose use "This method is cancellation safe".

djc:
@gretchenfrage want to polish this up so we can merge it?

gretchenfrage:
@djc Yeah, I'd like to prioritize #2168 personally because I don't like how long I've been blocking it. I haven't lost track of anything, I just haven't had capacity for a while.

djc:
> @djc Yeah, I'd like to prioritize #2168 personally because I don't like how long I've been blocking it. I haven't lost track of anything, I just haven't had capacity for a while.

Would be nice to get this over the finish line.

gretchenfrage:
My revision of this PR is in-progress, I'm hoping I can put it back into review tomorrow. I appreciate your patience with me as my capacity has been reduced a lot recently.

gretchenfrage:
- Rebased over main
- Changed the first lines of comments to be imperative, for stylistic consistency with rest of project
- No longer starting first line of docs with "Attempt to" (Tokio sometimes does that but sometimes not. My
  reasoning for removing it is that we don't need to explicitly say that a method which returns a `Result`
  might fail any more than we need to explicitly say that a method which is `async` returns a future, at
  least within the first line of the doc comment.)
- Take flub's suggestion to use Tokio-inspired `# Cancel safety` headers
- Slight tweaks / corrections to my own grammar
- Take Ralith's suggestion to replace phrase "resolves to" with "yields"
- No longer rearranging order of methods

PR #2230, opened 2025-05-10
Remove `BytesSource` and refactor `write_source`

### LINKED ISSUE #2228: Unclear why `BytesSource` is public
It's unclear to me why [`BytesSource`](https://docs.rs/quinn-proto/latest/quinn_proto/trait.BytesSource.html) is public. I see any public APIs it can be passed to.

gretchenfrage:

Closes #2228.

Breaking because it removes a publicly exposed item.

### DIFF
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -19,9 +19,8 @@ use recv::Recv;
 pub use recv::{Chunks, ReadError, ReadableError};
 
 mod send;
-pub(crate) use send::{ByteSlice, BytesArray};
-use send::{BytesSource, Send, SendState};
 pub use send::{FinishError, WriteError, Written};
+use send::{Send, SendState};
 
 mod state;
 #[allow(unreachable_pub)] // fuzzing only
@@ -221,7 +220,11 @@ impl<'a> SendStream<'a> {
     ///
     /// Returns the number of bytes successfully written.
     pub fn write(&mut self, data: &[u8]) -> Result<usize, WriteError> {
-        Ok(self.write_source(&mut ByteSlice::from_slice(data))?.bytes)
+        self.write_source(|limit, chunks| {
+            let prefix = &data[..limit.min(data.len())];
+            chunks.push(prefix.to_vec().into());
+            prefix.len()
+        })
     }
 
     /// Send data on the given stream
@@ -231,10 +234,38 @@ impl<'a> SendStream<'a> {
     /// [`Written::chunks`] will not count this chunk as fully written. However
     /// the chunk will be advanced and contain only non-written data after the call.
     pub fn write_chunks(&mut self, data: &mut [Bytes]) -> Result<Written, WriteError> {
-        self.write_source(&mut BytesArray::from_chunks(data))
+        self.write_source(|limit, chunks| {
+            let mut written = Written::default();
+            for chunk in data {
+                let prefix = chunk.split_to(chunk.len().min(limit - written.bytes));
+                written.bytes += prefix.len();
+                chunks.push(prefix);
+
+                if chunk.is_empty() {
+                    written.chunks += 1;
+                }
+
+                debug_assert!(written.bytes <= limit);
+                if written.bytes == limit {
+                    break;
+                }
+            }
+            written
+        })
     }
 
-    fn write_source<B: BytesSource>(&mut self, source: &mut B) -> Result<Written, WriteError> {
+    /// Send data on the given stream
+    ///
+    /// The `source` callback is invoked with the number of bytes that can be written immediately,
+    /// as well as an initially empty `&mut Vec<Bytes>` to which it can push bytes to write. If the
+    /// callback pushes a total number of bytes less than or equal to the provided limit, it is
+    /// guaranteed they will all be written. If it provides more bytes than this, it is guaranteed
+    /// that a prefix of the provided cumulative bytes will be written equal in length to the
+    /// provided limit.
+    fn write_source<T>(
+        &mut self,
+        source: impl FnOnce(usize, &mut Vec<Bytes>) -> T,
+    ) -> Result<T, WriteError> {
         if self.conn_state.is_closed() {
             trace!(%self.id, "write blocked; connection draining");
             return Err(WriteError::Blocked);
@@ -264,14 +295,14 @@ impl<'a> SendStream<'a> {
         }
 
         let was_pending = stream.is_pending();
-        let written = stream.write(source, limit)?;
-        self.state.data_sent += written.bytes as u64;
-        self.state.unacked_data += written.bytes as u64;
-        trace!(stream = %self.id, "wrote {} bytes", written.bytes);
+        let (written, source_output) = stream.write(source, limit)?;
+        self.state.data_sent += written as u64;
+        self.state.unacked_data += written as u64;
+        trace!(stream = %self.id, "wrote {} bytes", written);
         if !was_pending {
             self.state.pending.push_pending(self.id, stream.priority);
         }
-        Ok(written)
+        Ok(source_output)
     }
 
     /// Check if this stream was stopped, get the reason if it was
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -20,8 +20,8 @@ pub use recv::{Chunks, ReadError, ReadableError};
 
 mod send;
 pub(crate) use send::{ByteSlice, BytesArray};
-pub use send::{BytesSource, FinishError, WriteError, Written};
-use send::{Send, SendState};
+use send::{BytesSource, Send, SendState};
+pub use send::{FinishError, WriteError, Written};
 
 mod state;
 #[allow(unreachable_pub)] // fuzzing only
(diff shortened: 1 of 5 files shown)


quinn-proto/src/connection/streams/mod.rs
```diff
@@ -231,10 +234,38 @@ impl<'a> SendStream<'a> {
...
 
-    fn write_source<B: BytesSource>(&mut self, source: &mut B) -> Result<Written, WriteError> {
+    /// Send data on the given stream
+    ///
+    /// The `source` callback is invoked with the number of bytes that can be written immediately,
+    /// as well as an initially empty `&mut Vec<Bytes>` to which it can push bytes to write. If the
+    /// callback pushes a total number of bytes less than or equal to the provided limit, it is
+    /// guaranteed they will all be written. If it provides more bytes than this, it is guaranteed
+    /// that a prefix of the provided cumulative bytes will be written equal in length to the
+    /// provided limit.
```
Ralith, line 264: What happens to the excess bytes? Would it be simpler to require (and even `assert`) that the limit is respected?

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -231,10 +234,38 @@ impl<'a> SendStream<'a> {
...
+    fn write_source<T>(
+        &mut self,
+        source: impl FnOnce(usize, &mut Vec<Bytes>) -> T,
```
Ralith, line 267: Straw alternative proposal: `impl FnOnce(usize) -> I` where `I: IntoIter<Item=Bytes>`. Let callers worry about their own return values via side-effects. This makes for a less leaky abstraction, saves a `Vec`, and relieves us from having to plumb a `T` back to the caller.

Ralith (commented):
I love this simplification! The prospect of eventually having a `pub`-worthy generic `write` method is also very appealing; we're picking up too many variants.

Ralith:
I think we should consider this non-breaking, since `BytesSource` wasn't intentionally exposed and wasn't actually useful. I doubt anyone's actually touched it.

quinn-proto/src/connection/streams/send.rs
```diff
@@ -66,12 +69,12 @@ impl Send {
...
-        let mut chunks = Vec::new();
-        let source_output = source(limit, &mut chunks);
+        debug_assert!(self.chunks.is_empty());
+        let source_output = source(limit, &mut self.chunks);
 
         let mut written = 0;
 
-        for mut chunk in chunks.drain(..) {
+        for mut chunk in self.chunks.drain(..) {
             let prefix = chunk.split_to(chunk.len().min(limit - written));
```
Ralith, line 78: It might be nice to skip the `split_to` when we can consume the whole chunk. It's at least an indirect call, and might conceivably fiddle with refcounts.

gretchenfrage: It looks like `split_to` already has an optimization for this, so I don't think we need to :)

https://docs.rs/bytes/latest/src/bytes/bytes.rs.html#548-571

```rs
        if at == self.len() {
            let end_ptr = self.ptr.wrapping_add(at);
            return mem::replace(self, Bytes::new_empty_with_ptr(end_ptr));
        }
```

quinn-proto/src/connection/streams/mod.rs
gretchenfrage: **Re: `I: IntoIter<Item=Bytes>`:**

I'm contemplating that, and I may experiment with it. You've mentioned the pros, the cons include:

- More complex generics. Harder to read, adds compile time(?).
- Probably trickier to use (I imagine I would change `write_chunks` to use `std::iter::from_fn` or something).
- Application code able to panic in the middle of proto looping through the chunks, whereas currently it can only panic before that loop. Might be trickier to analyze.
- May be in tension with your above suggestion to panic if the user tries to write too many bytes, because it makes it impossible to confirm up-front how many bytes the user is submitting (unless we have the user return an `IntoIterator` and then just collect it into a `Vec`, which seems like the worst of both worlds in many ways).

I'm kind of leaning against.

gretchenfrage: **Re: "Let callers worry about their own return values via side-effects":**

Here's the diff I would make to do that. It does not remove any lines of code from `write_source`/`Send::write`, it adds 2 lines of code in net, it forces me to think about the subtleties of closure variable capturing stuff, and it makes the callers do this awkward `.map(|()| written)` thing. Conversely, plumbing it through acts as a nice type-level proof of when the callback will or won't be called, making the code easier to reason about. Especially if in the future we expose it publicly, as in #2231.

```diff
diff --git a/quinn-proto/src/connection/streams/mod.rs b/quinn-proto/src/connection/streams/mod.rs
index 70f51e39..137cf370 100644
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -220,11 +220,13 @@ impl<'a> SendStream<'a> {
     ///
     /// Returns the number of bytes successfully written.
     pub fn write(&mut self, data: &[u8]) -> Result<usize, WriteError> {
+        let mut written = 0;
         self.write_source(|limit, chunks| {
             let prefix = &data[..limit.min(data.len())];
             chunks.push(prefix.to_vec().into());
-            prefix.len()
+            written = prefix.len();
         })
+        .map(|()| written)
     }
 
     /// Send data on the given stream
@@ -234,8 +236,8 @@ impl<'a> SendStream<'a> {
     /// [`Written::chunks`] will not count this chunk as fully written. However
     /// the chunk will be advanced and contain only non-written data after the call.
     pub fn write_chunks(&mut self, data: &mut [Bytes]) -> Result<Written, WriteError> {
+        let mut written = Written::default();
         self.write_source(|limit, chunks| {
-            let mut written = Written::default();
             for chunk in data {
                 let prefix = chunk.split_to(chunk.len().min(limit - written.bytes));
                 written.bytes += prefix.len();
@@ -250,8 +252,8 @@ impl<'a> SendStream<'a> {
                     break;
                 }
             }
-            written
         })
+        .map(|()| written)
     }
 
     /// Send data on the given stream
@@ -262,10 +264,10 @@ impl<'a> SendStream<'a> {
     /// guaranteed they will all be written. If it provides more bytes than this, it is guaranteed
     /// that a prefix of the provided cumulative bytes will be written equal in length to the
     /// provided limit.
-    fn write_source<T>(
+    fn write_source(
         &mut self,
-        source: impl FnOnce(usize, &mut Vec<Bytes>) -> T,
-    ) -> Result<T, WriteError> {
+        source: impl FnOnce(usize, &mut Vec<Bytes>),
+    ) -> Result<(), WriteError> {
         if self.conn_state.is_closed() {
             trace!(%self.id, "write blocked; connection draining");
             return Err(WriteError::Blocked);
@@ -295,14 +297,14 @@ impl<'a> SendStream<'a> {
         }
 
         let was_pending = stream.is_pending();
-        let (written, source_output) = stream.write(source, limit)?;
+        let written = stream.write(source, limit)?;
         self.state.data_sent += written as u64;
         self.state.unacked_data += written as u64;
         trace!(stream = %self.id, "wrote {} bytes", written);
         if !was_pending {
             self.state.pending.push_pending(self.id, stream.priority);
         }
-        Ok(source_output)
+        Ok(())
     }
 
     /// Check if this stream was stopped, get the reason if it was
diff --git a/quinn-proto/src/connection/streams/send.rs b/quinn-proto/src/connection/streams/send.rs
index 52a9b714..2217b988 100644
--- a/quinn-proto/src/connection/streams/send.rs
+++ b/quinn-proto/src/connection/streams/send.rs
@@ -52,11 +52,11 @@ impl Send {
         }
     }
 
-    pub(super) fn write<T>(
+    pub(super) fn write(
         &mut self,
-        source: impl FnOnce(usize, &mut Vec<Bytes>) -> T,
+        source: impl FnOnce(usize, &mut Vec<Bytes>),
         limit: u64,
-    ) -> Result<(usize, T), WriteError> {
+    ) -> Result<usize, WriteError> {
         if !self.is_writable() {
             return Err(WriteError::ClosedStream);
         }
@@ -70,7 +70,7 @@ impl Send {
         let limit = limit.min(budget) as usize;
 
         debug_assert!(self.chunks.is_empty());
-        let source_output = source(limit, &mut self.chunks);
+        source(limit, &mut self.chunks);
 
         let mut written = 0;
 
@@ -85,7 +85,7 @@ impl Send {
             }
         }
 
-        Ok((written, source_output))
+        Ok(written)
     }
 
     /// Update stream state due to a reset sent by the local application

```

</details>

djc: I think I prefer it with the diff. Avoiding the generic is worth the 2 extra lines to me and IMO distributes responsibility in a better way.

gretchenfrage:
Removing the breaking label; rebased over #2233 which took care of that complication.

quinn-proto/src/connection/streams/mod.rs
Ralith: > it makes the callers do this awkward .map(|()| written) thing

I think this will look much less weird if you phrase it as `?; Ok(written)`.

gretchenfrage:
Closing in favor of #2260 or successor.

gretchenfrage:
Closing in favor of https://github.com/quinn-rs/quinn/pull/2260 or successor.

PR #2231, opened 2025-05-10
(Draft) Public `write_source` API

gretchenfrage:
This draft PR currently targets a temporary base branch formed by merging #2230 and #2226 into main.

---

Overall, this PR makes the refactored `write_source` API proposed in #2230 public and threads it through to a public API in Quinn as well. This has value both for simplicity and utility. From the simplicity side, it allows us to supersede and delete the `quinn::SendStream::execute_poll` function.

From the utility side, this API, for example, should be considered to close #2229, as it could be used to implement a `write_atomic` method purely on top of existing public methods as such:

```diff
diff --git a/quinn/src/send_stream.rs b/quinn/src/send_stream.rs
index fcea07cf..b5a4240c 100644
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -1,4 +1,5 @@
 use std::{
+    convert::identity,
     fmt,
     future::{Future, poll_fn},
     io,
@@ -101,6 +102,20 @@ impl SendStream {
         Ok(())
     }
 
+    pub async fn write_atomic(&mut self, buf: &[u8]) -> Result<(), WriteAtomicError> {
+        self.write_source(|limit, chunks| {
+            if limit >= buf.len() {
+                chunks.push(buf.to_vec().into());
+                Ok(())
+            } else {
+                Err(WriteAtomicError::TooLarge)
+            }
+        })
+        .await
+        .map_err(|e| WriteAtomicError::WriteError(e.error))
+        .and_then(identity)
+    }
+
     /// Attempts to write bytes into this stream from a byte-providing callback
     ///
     /// This is a low-level writing API that can be used to perform writes in a way that is atomic
@@ -465,3 +480,14 @@ impl<F> From<WriteSourceError<F>> for io::Error {
         e.error.into()
     }
 }
+
+/// Errors that arise from writing to a stream atomically
+#[derive(Debug, Error, Clone, PartialEq, Eq)]
+pub enum WriteAtomicError {
+    /// The write was too large to send atomically at this time
+    #[error("too large for atomic write")]
+    TooLarge,
+    /// Some other write error occurred
+    #[error("stream write error")]
+    WriteError(WriteError),
+}
```

PR #2232, opened 2025-05-11
refactor(quinn): Remove some usage of execute_poll

gretchenfrage:
This commit makes there be fewer places where SendStream::execute_poll is called directly. AsyncWrite implementations now go through poll_write, and poll_write now goes through the write future. This means that execute_poll is now only called directly from write and write_chunks.


djc:
Why is this better?

gretchenfrage:
> Why is this better?

It's simpler. Consider that the previous versions of these lines involve a closure, whereas the updated versions don't. The previous versions of the lines involve calling two of our custom methods (`execute_poll` and `write`), whereas the updated versions only call one (`poll_write` _or_ `write`).

Previously, the logic of integrating `execute_poll` and `write` in an appropriate way existed at 4 duplicate sites, whereas with this PR it exists only in one site (`write`) and the other places becomes derivative of that (for example, in #2231 this makes it simpler to replace `execute_poll` entirely. But I'm only bringing up that draft PR for illustratory purposes--this refactor should stand on its own.)

djc (approved):

Ralith (approved):
Nice to know that `poll_write` isn't actually primitive, though it may make sense to keep as a discoverable convenience.

PR #2242, opened 2025-05-13
proto: Replace write_source with WriteGuard<'_>

gretchenfrage:
This PR is **competing** and mutually exclusive with my #2230 PR. In the course of developing 2230 I had the idea that this approach could be tried instead. The impact on LOC is comparable, and this is simpler in some ways (avoids inversion of control). I'm currently creating this as a draft because I haven't thought about it for very long but you're welcome to look at it and see if you think this is better than 2230?

---

This commit removes the ByteSource trait and the write_source method. Instead, a new method is introduced as such:

fn SendStream::write_guard(&mut self) -> Result<WriteGuard<'_>, WriteError>

The resultant WriteGuard contains the limit of how many bytes could be written now, as well as borrows of the fields necessary to perform those writes. The write and write_chunks methods are written to more simply use this.

### DIFF
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -19,9 +19,8 @@ use recv::Recv;
 pub use recv::{Chunks, ReadError, ReadableError};
 
 mod send;
-pub(crate) use send::{ByteSlice, BytesArray};
-use send::{BytesSource, Send, SendState};
 pub use send::{FinishError, WriteError, Written};
+use send::{Send, SendState};
 
 mod state;
 #[allow(unreachable_pub)] // fuzzing only
@@ -221,7 +220,10 @@ impl<'a> SendStream<'a> {
     ///
     /// Returns the number of bytes successfully written.
     pub fn write(&mut self, data: &[u8]) -> Result<usize, WriteError> {
-        Ok(self.write_source(&mut ByteSlice::from_slice(data))?.bytes)
+        let mut guard = self.write_guard()?;
+        let written = data.len().min(guard.limit);
+        guard.write(data[..written].to_vec().into());
+        Ok(written)
     }
 
     /// Send data on the given stream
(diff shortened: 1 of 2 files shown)


djc (commented):
I like this direction -- I think guard types generally make for more idiomatic API than callbacks (whether direct or via a custom trait definition).

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -367,6 +385,28 @@ impl<'a> SendStream<'a> {
     }
 }
 
+struct WriteGuard<'a> {
+    limit: usize,
+    id: StreamId,
+    stream: &'a mut Send,
+    data_sent: &'a mut u64,
+    unacked_data: &'a mut u64,
+    pending: &'a mut PendingStreamsQueue,
```
Ralith, line 394: Should these be a single reference to a helper struct? That might also be easier to adapt to the `quinn` layer.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -367,6 +385,28 @@ impl<'a> SendStream<'a> {
     }
 }
 
+struct WriteGuard<'a> {
```
Ralith, line 388: bikeshed: `WriteBuilder`? "Guard" usually means "cleans something up on Drop".

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -231,10 +233,25 @@ impl<'a> SendStream<'a> {
...
+
+            if guard.limit == 0 {
+                break;
+            }
+        }
+        Ok(written)
     }
 
-    fn write_source<B: BytesSource>(&mut self, source: &mut B) -> Result<Written, WriteError> {
+    fn write_guard(&mut self) -> Result<WriteGuard, WriteError> {
```
Ralith, line 254: bikeshed: `build_write`, or maybe `begin_write`.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -221,7 +220,10 @@ impl<'a> SendStream<'a> {
     ///
     /// Returns the number of bytes successfully written.
     pub fn write(&mut self, data: &[u8]) -> Result<usize, WriteError> {
-        Ok(self.write_source(&mut ByteSlice::from_slice(data))?.bytes)
+        let mut guard = self.write_guard()?;
+        let written = data.len().min(guard.limit);
+        guard.write(data[..written].to_vec().into());
```
Ralith, line 225:
```suggestion
        guard.write(Bytes::copy_from_slice(&data[..written]));
```

Ralith (commented):
+1 to this direction.

quinn-proto/src/connection/streams/mod.rs
djc: Maybe `StreamWriter`? `Builder` invokes builder pattern for construction to me.

gretchenfrage:
Closing in favor of #2260 or successor.

gretchenfrage:
Closing in favor of https://github.com/quinn-rs/quinn/pull/2260 or successor.

PR #2248, opened 2025-05-17
Fix false positive GSO detection

inetic:
On some devices with kernel <4.18 the existing checks for GSO not being enabled on the system don't trigger. Quinn then continues to use GSO which results in eventual disconnection as discussed in issue https://github.com/quinn-rs/quinn/issues/2246.

This PR adds a high level test to showcase the disconnection and a runtime check based on parsing Linux kernel version string to disable GSO in `quinn-udp` on kernels where it hasn't yet been implemented (4.18 https://github.com/torvalds/linux/commit/cb586c63e3fc5b227c51fd8c4cb40b34d3750645).


quinn/tests/send_receive.rs
thomaseizinger, line 1: We already have the GSO test failing on that device right? Is there much value in an additional test?

quinn-udp/src/unix.rs
```diff
@@ -804,11 +854,18 @@ mod gso {
     // TODO: Add this to libc
     const UDP_SEGMENT: libc::c_int = 103;
 
+    // Global mutex to avoid calling `libc::uname` for each socket.
+    static SUPPORTED_BY_OS: Mutex<Option<bool>> = Mutex::new(None);
```
thomaseizinger, line 858: Perhaps a `OnceLock` is more convenient here?

thomaseizinger: We can then get rid of the `KernelVersion` struct and have a stateless function that checks if the version is < 4.18

quinn-udp/src/unix.rs
```diff
@@ -830,6 +887,37 @@ mod gso {
...
+                return false;
+            }
+        };
+
+        let supported = match *lock {
+            Some(value) => return value,
+            None => match linux::KernelVersion::from_uname() {
+                Some(kernel_version) => {
+                    if kernel_version.lower_than(4, 18) {
+                        crate::log::trace!(
```
thomaseizinger, line 904: I think this is worth an INFO log.

quinn-udp/src/unix.rs
```diff
@@ -804,11 +854,18 @@ mod gso {
     // TODO: Add this to libc
     const UDP_SEGMENT: libc::c_int = 103;
 
+    // Global mutex to avoid calling `libc::uname` for each socket.
+    static SUPPORTED_BY_OS: Mutex<Option<bool>> = Mutex::new(None);
+
     /// Checks whether GSO support is available by setting the UDP_SEGMENT
```
thomaseizinger, line 860: Stale docs

quinn-udp/src/unix.rs
```diff
@@ -794,6 +794,56 @@ pub(crate) const BATCH_SIZE: usize = 32;
...
+        pub(crate) string: String,
+    }
+
+    impl KernelVersion {
+        pub(crate) fn from_uname() -> Option<Self> {
+            let mut n = unsafe { std::mem::zeroed() };
+            let r = unsafe { libc::uname(&mut n) };
+            if r != 0 {
+                crate::log::trace!("Failed to run libc::uname ({})", r);
+                return None;
```
thomaseizinger, line 812: You can create `std::io::Error::other` here and for all the other failure cases too. Then, you only need to log in a single place as to why GSO is not supported:

- kernel version too low
- some IO / parsing error in the check

I'd almost say the IO error should be a WARN or an ERROR?

thomaseizinger (commented):
Left a few comments with food for thought!

quinn-udp/src/unix.rs
inetic: > Then, you only need to log in a single place as to why GSO is not supported:

If it's only for this reason, I could also return `Result<Self, String>`. Or do you prefer `Result<Self, std::io::Error>`?

quinn-udp/src/unix.rs
inetic: > We can then get rid of the KernelVersion struct and have a stateless function that checks if the version is < 4.18

I had the opposite plan to expand the `KernelVersion` struct by further spliting the parsing code into `KernelVersion::from_str` which would be called from `KernelVersion::from_uname` and write small tests with sting samples. The test could be a way of tracking if some kernels have inconsistent version strings as you noted here https://github.com/quinn-rs/quinn/issues/2246#issuecomment-2887779815.

This wouldn't be possible with a single function returning a boolean.

Let me know which way you prefer.

quinn/tests/send_receive.rs
inetic: Not a lot of added value I think. I'm happy to remove it unless someone objects in the mean time.

quinn-udp/src/unix.rs
thomaseizinger: No strong preference.

quinn-udp/src/unix.rs
thomaseizinger: > This wouldn't be possible with a single function returning a boolean.

You can split it into multiple functions and test the pure ones separately:

- 1 function that calls `libc::uname` and returns a string
- 1 function that checks if a given string is a kernel version > 4.18
- 1 function that combines the above too or perhaps just do that inline in the closure passed to `once_cell`

The 2nd one can then be easily unit-tested with multiple inputs.

I don't have a strong preference but if we can avoid intermediary structs, I think it is easier to understand.

quinn/tests/send_receive.rs
djc: Yes, and the test you added is pretty extensive. Would prefer to have it removed.

quinn-udp/src/unix.rs
inetic: > I don't think we can use OnceLock for MSRV reasons, 

Running [`cargo msrv find`](https://github.com/foresterre/cargo-msrv) I get the result 1.74.1 (log: [msrv-find.log](https://github.com/user-attachments/files/20279627/msrv-find.log)). `OnceLock` seems to be [supported since 1.70.0](https://doc.rust-lang.org/std/sync/struct.OnceLock.html). So unless there is a plan to decrease the current MSRV I think `OnceLock` should be OK to use.

djc: Ah, okay, good.

inetic: I did modify the code a bit, though I left the `KernelVersion` struct there. The reason was that I felt that the tests would be more to the point if they tested the parsing alone instead of parsing and checking >=4.18. It might be a matter of taste/style, if so, then perhaps it'd be better to leave that part to you (the code owners)?

inetic:
One outstanding question to resolve is whether to enable GSO if `uname` or the parsing fails, as mentioned here https://github.com/quinn-rs/quinn/issues/2246#issuecomment-2887779815. There are pros and cons:

Fallback is to enable:

* pros: fast bandwidth and low cycles on modern kernels
* cons: broken connections on old kernels

Fallback is to disable

* pros: connections work on old kernels
* cons: slow bandwidth and high cycles on modern kernels

quinn-udp/src/unix.rs
```diff
@@ -794,6 +796,91 @@ pub(crate) const BATCH_SIZE: usize = 32;
...
+            let mut split = release
+                .split_once('-')
+                .map(|pair| pair.0)
+                .unwrap_or(release)
+                .split('.');
+
+            let Some(version) = split.next().and_then(|s| s.parse().ok()) else {
+                return Err(format!("Failed to parse kernel version from {release:?}"));
+            };
+            let Some(major_revision) = split.next().and_then(|s| s.parse().ok()) else {
```
thomaseizinger, line 839: This could be a `.ok_or_else()?`.

quinn-udp/src/unix.rs
```diff
@@ -830,6 +926,36 @@ mod gso {
...
+                return false;
+            }
+        };
+
+        if kernel_version < SUPPORTED_SINCE {
+            crate::log::info!(
+                "GSO supported on Linux kernels 4.18+, current is {:?}",
+                kernel_version_string
+            );
+            false
```
thomaseizinger, line 953:
```suggestion
            return false
```

Then you can remove the `else`.

quinn-udp/src/unix.rs
```diff
@@ -830,6 +926,36 @@ mod gso {
         }
     }
 
+    fn supported_by_current_kernel() -> bool {
+        let kernel_version_string = match linux::kernel_version_string() {
+            Ok(kernel_version_string) => kernel_version_string,
+            Err(_errno) => {
+                crate::log::warn!(
+                    "Failed to retrieve kernel version string, GSO not enabled ({_errno})"
```
thomaseizinger, line 934:
```suggestion
                    "GSO disabled: uname returned {_errno}"
```

quinn-udp/src/unix.rs
```diff
@@ -830,6 +926,36 @@ mod gso {
...
+                );
+                return false;
+            }
+        };
+
+        let kernel_version = match linux::KernelVersion::from_str(&kernel_version_string) {
+            Ok(kernel_version) => kernel_version,
+            Err(_reason) => {
+                crate::log::warn!("GSO not enabled: {}", _reason);
```
thomaseizinger, line 943:
```suggestion
                crate::log::warn!("GSO disabled: {}", _reason);
```

quinn-udp/src/unix.rs
```diff
@@ -830,6 +926,36 @@ mod gso {
...
+                return false;
+            }
+        };
+
+        if kernel_version < SUPPORTED_SINCE {
+            crate::log::info!(
+                "GSO supported on Linux kernels 4.18+, current is {:?}",
```
thomaseizinger, line 950:
```suggestion
                "GSO disabled: kernel too old ({kernel_version_string}); need 4.18+",
```

thomaseizinger (approved):
Nice work, looks great! I left a few comments, nothing blocking :)

Ralith:
> whether to enable GSO if uname or the parsing fails

Are we aware of any cases where these could fail with a modern kernel? Especially uname failing suggests to me that we'd be in a very exotic environment where it might be better to be conservative, although maybe syscall filtering could cause that?

thomaseizinger:
> > whether to enable GSO if uname or the parsing fails
> 
> Are we aware of any cases where these could fail with a modern kernel? Especially uname failing suggests to me that we'd be in a very exotic environment where it might be better to be conservative, although maybe syscall filtering could cause that?

I am not aware of any cases. The idea of falling back to GSO _enabled_ is that in the rare(?) case of an error, this doesn't cause a performance regression by disabling GSO.

My guess would be that `quinn-udp` is predominantely run on newer Linux kernels so defaulting to `true` seems like the better option. If it does indeed fail on an older kernel, the user will see it in the logs and can report a bug.

quinn-udp/src/unix.rs
```diff
@@ -794,21 +794,120 @@ pub(crate) const BATCH_SIZE: usize = 32;
 #[cfg(apple_slow)]
 pub(crate) const BATCH_SIZE: usize = 1;
 
+#[cfg(any(target_os = "linux", target_os = "android"))]
+mod linux {
```
djc, line 798: Nit: there's no point in making a separate module for this, just move the contents of this into the existing Linux `gso` module. Try to keep the additions in top-down order, so callers come before callees.

quinn-udp/src/unix.rs
```diff
@@ -794,21 +794,120 @@ pub(crate) const BATCH_SIZE: usize = 32;
...
+        pub(crate) fn from_str(release: &str) -> Result<Self, String> {
+            let mut split = release
+                .split_once('-')
+                .map(|pair| pair.0)
+                .unwrap_or(release)
+                .split('.');
+
+            let version = split
+                .next()
+                .and_then(|s| s.parse().ok())
```
djc, line 836: Nit: please use `u8::from_str()` instead of `.parse()`, and avoid capitalizing error messages.

djc:
Thanks for working on this!

quinn-udp/src/unix.rs
inetic: > use u8::from_str() instead of .parse()

This will require `use std::str::FromStr;`, are you ok with it?

djc: Yes; that's the same stuff that `parse()` uses.

inetic: > avoid capitalizing error messages

Also for "GSO" in "GSO disabled:..."?

inetic:
> Thanks for working on this!

Thanks for all the feedback :)

quinn-udp/src/unix.rs
djc: Sorry, no -- just the first letter.

quinn-udp/src/unix.rs
```diff
@@ -833,6 +844,118 @@ mod gso {
...
+    impl KernelVersion {
+        const fn new(version: u8, major_revision: u8) -> Self {
+            Self {
+                version,
+                major_revision,
+            }
+        }
+
+        fn from_str(release: &str) -> Result<Self, String> {
+            use std::str::FromStr;
```
djc, line 902: Nit: import this at the top of the module.

quinn-udp/src/unix.rs
```diff
@@ -833,6 +844,118 @@ mod gso {
...
+
+    impl KernelVersion {
+        const fn new(version: u8, major_revision: u8) -> Self {
+            Self {
+                version,
+                major_revision,
+            }
+        }
+
+        fn from_str(release: &str) -> Result<Self, String> {
```
djc, line 901: Nit: suggest returning `Option<Self>` from this.

quinn-udp/src/unix.rs
```diff
@@ -833,6 +844,118 @@ mod gso {
...
+            use std::str::FromStr;
+
+            let mut split = release
+                .split_once('-')
+                .map(|pair| pair.0)
+                .unwrap_or(release)
+                .split('.');
```
djc, line 908: ... then this can be `release.split_once('-')?.0.split('.')`.

quinn-udp/src/unix.rs
```diff
@@ -833,6 +844,118 @@ mod gso {
...
+
+            let version = split
```
djc, line 910: And this can be `u8::from_str(split.next()?).ok()?`.

quinn-udp/src/unix.rs
```diff
@@ -833,6 +844,118 @@ mod gso {
...
+                .next()
+                .and_then(|s| u8::from_str(s).ok())
+                .ok_or_else(|| format!("failed to parse kernel version from {release:?}"))?;
+
+            let major_revision = split
+                .next()
+                .and_then(|s| u8::from_str(s).ok())
+                .ok_or_else(|| format!("failed to parse kernel major revision from {release:?}"))?;
```
djc, line 918: And similar here.

quinn-udp/src/unix.rs
```diff
@@ -833,6 +844,118 @@ mod gso {
...
+    }
+
+    fn kernel_version_string() -> Result<String, libc::c_int> {
+        let mut n = unsafe { std::mem::zeroed() };
+        let r = unsafe { libc::uname(&mut n) };
+        if r != 0 {
+            return Err(r);
+        }
+        Ok(unsafe {
+            std::ffi::CStr::from_ptr(n.release[..].as_ptr())
```
djc, line 880: Nit: import `CStr` at the top of the module, and `mem` too.

quinn-udp/src/unix.rs
```diff
@@ -833,6 +844,118 @@ mod gso {
     pub(crate) fn set_segment_size(encoder: &mut cmsg::Encoder<libc::msghdr>, segment_size: u16) {
         encoder.push(libc::SOL_UDP, UDP_SEGMENT, segment_size);
     }
+
+    fn supported_by_current_kernel() -> bool {
+        let kernel_version_string = match kernel_version_string() {
+            Ok(kernel_version_string) => kernel_version_string,
+            Err(_errno) => {
```
djc, line 851: Why do these need `_` prefixes?

quinn-udp/src/unix.rs
```diff
@@ -797,18 +797,29 @@ pub(crate) const BATCH_SIZE: usize = 1;
...
     // TODO: Add this to libc
     const UDP_SEGMENT: libc::c_int = 103;
 
-    /// Checks whether GSO support is available by setting the UDP_SEGMENT
-    /// option on a socket
+    // Support for UDP GSO has been added to linux kernel in version 4.18
+    // https://github.com/torvalds/linux/commit/cb586c63e3fc5b227c51fd8c4cb40b34d3750645
+    const SUPPORTED_SINCE: KernelVersion = KernelVersion::new(4, 18);
+    // Avoid calling `supported_by_current_kernel` for each socket by using `OnceLock`.
+    static SUPPORTED_BY_CURRENT_KERNEL: OnceLock<bool> = OnceLock::new();
```
djc, line 812: Nit: move `SUPPORTED_BY_CURRENT_KERNEL` below `set_segment_size()` (it should be below `max_gso_segments()` and above `supported_by_current_kernel()`).

djc (commented):
A bunch of style feedback, I think we're close!

quinn-udp/src/unix.rs
thomaseizinger: That is what we originally had and I suggested returning an error to consolidate the places where we log what went wrong :)

quinn-udp/src/unix.rs
thomaseizinger: When the log macros are configured out (no features active), they'd otherwise show up as unused variables.

quinn-udp/src/unix.rs
thomaseizinger: I guess as long as we print the original string returned from `uname` as part of the WARN log, we can always reproduce it if this gets hit in the wild.

quinn-udp/src/unix.rs
inetic: The `.unwrap_or(release)` part is there to avoid failing on strings which don't have "-", like "4.18" for example. I don't know whether such release strings exist in real life, but might be good to keep it just in case?

On the other hand if we go with the "let's not worry about what we don't know" approach, then we could get rid of the `.split_once('-')` part all together because the formats that I've seen so far were always "X.Y.Z-...", for which `let mut split = release.split('.');` would suffice.

quinn-udp/src/unix.rs
```diff
@@ -833,6 +845,116 @@ mod gso {
...
+            Ok(kernel_version_string) => kernel_version_string,
+            Err(_errno) => {
+                crate::log::warn!("GSO disabled: uname returned {_errno}");
+                return false;
+            }
+        };
+
+        let Some(kernel_version) = KernelVersion::from_str(&kernel_version_string) else {
+            crate::log::warn!(
+                "GSO disabled: failed to parse kernel version ({kernel_version_string:?})"
```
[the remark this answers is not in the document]
inetic: Sorry, I thought you might have made a mistake when you suggested the string to use `Display` instead of `Debug`, I'll change it but for the sake of knowing next time, may I ask why the preference for `Display`? My preference for `Debug` was that it adds quotes around the string so problems with empty strings or white spaces are easier to spot.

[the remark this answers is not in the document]
thomaseizinger: Just personal preference really, we are already putting it in brackets so I don't see a need for quoting it again. I actually thought it would also print `String("")`! Just tried that though and it doesn't :)

Don't really mind, happy to leave on `Debug` if you prefer it!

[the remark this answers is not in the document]
inetic: I completely missed the brackets 🤦🙂

mxinden (commented):
One comment, otherwise looks good to me. 

Thank you for the work!

quinn-udp/src/unix.rs
```diff
@@ -833,6 +845,116 @@ mod gso {
...
+
+        true
+    }
+
+    fn kernel_version_string() -> Result<String, libc::c_int> {
+        let mut n = unsafe { mem::zeroed() };
+        let r = unsafe { libc::uname(&mut n) };
+        if r != 0 {
+            return Err(r);
+        }
```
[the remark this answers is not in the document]
thomaseizinger: See also: https://doc.rust-lang.org/std/io/struct.Error.html#method.last_os_error

djc (approved):
Okay, going to merge this and do some follow up. Thanks for all the iterations!

PR #2249, opened 2025-05-18
Fix new clippy warnings about large errors

gretchenfrage:
It seems that clippy added some new warnings about `Result::Err` variants with large byte sizes. This is interfering with attempts to merge #2245 (and presumably, any other PR). This PR:

- Suppresses the warning for `AcceptError`, since it is impossible to fix that one without semver breakage. In the next breaking release, we could consider fixing it somehow, such as by changing error type for that method to `Box<AcceptError>`.
- Fixes the warning for `RetryError` by boxing the contents.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -516,6 +516,8 @@ impl Endpoint {
     }
 
     /// Attempt to accept this incoming connection (an error may still occur)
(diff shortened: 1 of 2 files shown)


flub (approved):

djc (approved):
Can you file an issue with appropriate label to follow up on this once we start work on 0.12?

PR #2257, opened 2025-05-30
feat: add event and function to wait for resumption tickets

Frando:
This adds a new connection event to quinn-proto: `ResumptionEnabled`. This event is emitted if TLS13 resumption tickets were received. The event is emitted at most once and only on the client side.

In `quinn`, this adds a new function `resumption_tickets_received`, which returns a future that completes once the connection received TLS resumption tickets. The future will be pending forever on the server side. On the client side, it will complete immediately if tickets were already received, or complete once tickets are received, or stay pending forever if the server does not send us tickets.

This new feature is useful when doing many short-lived 0rtt connections: It may happen easily that the application-level data exchange is done in a single roundtrip (eg when doing a single, small request-response) but the resumption tickets from the server are only sent in the next packet. If closing the connection after the application data exchange completes, we did not yet receive the new tickets and thus cannot issue further 0rtt connections when our initial tickets are consumed. With this new API, a client-side 0rtt connection can wait for `conn.resumption_tickets_received()` before closing the connection (should also apply a timeout in case the server does not send any tickets). Thus, it can issue further 0rtt connections without having to go back to 1rtt inbetween.

The PR uses the functionality added in [this rustls PR](https://github.com/rustls/rustls/pull/2476) to expose the number of TLS13 resumption tickets that were received so far.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -238,6 +238,8 @@ pub struct Connection {
     stats: ConnectionStats,
     /// QUIC version used for the connection.
     version: u32,
+    /// True if we emitted the event for received resumption tickets
+    resumption_tickets_received: bool,
 }
 
 impl Connection {
@@ -355,6 +357,7 @@ impl Connection {
             rng,
             stats: ConnectionStats::default(),
             version,
+            resumption_tickets_received: false,
         };
         if path_validated {
             this.on_path_validated();
@@ -2111,6 +2114,12 @@ impl Connection {
             if self.crypto.read_handshake(&chunk.bytes)? {
                 self.events.push_back(Event::HandshakeDataReady);
             }
+        }
+
+        if !self.resumption_tickets_received && self.crypto.resumption_tickets_received() > Some(0)
+        {
+            self.resumption_tickets_received = true;
+            self.events.push_back(Event::ResumptionEnabled)
         }
 
         Ok(())
@@ -4000,6 +4009,13 @@ pub enum Event {
     DatagramReceived,
     /// One or more application datagrams have been sent after blocking
     DatagramsUnblocked,
+    /// Resumption of the cryptographic session is now possible.
+    ///
+    /// When using the rustls TLS session provider, this event is emitted when one or more
+    /// TLS session resumption tickets have been received.
+    ///
+    /// It is only emitted on the client, and is emitted at most once per connection.
+    ResumptionEnabled,
 }
 
 fn instant_saturating_sub(x: Instant, y: Instant) -> Duration {
--- a/quinn-proto/src/crypto.rs
+++ b/quinn-proto/src/crypto.rs
@@ -91,6 +91,11 @@ pub trait Session: Send + Sync + 'static {
         label: &[u8],
         context: &[u8],
     ) -> Result<(), ExportKeyingMaterialError>;
+
+    /// Returns the number of TLS1.3 session resumption tickets that were received
+    ///
+    /// Returns `None` on the server side.
+    fn resumption_tickets_received(&self) -> Option<u32>;
 }
 
 /// A pair of keys for bidirectional communication
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -651,6 +651,37 @@ impl Connection {
         // May need to send MAX_STREAMS to make progress
         conn.wake();
     }
+
+    /// Waits until the connection received TLS resumption tickets
+    ///
+    /// Yields `true` once resumption tickets were received. Resolves immediately
+    /// if tickets were already received, otherwise it resolves once tickets arrive.
+    /// If the server does not send any tickets, the returned future will remain pending forever.
+    ///
+    /// This should only be used on the client side. On the server side, it will
+    /// always resolve immediately and yield `false`.
+    pub fn resumption_tickets_received(&self) -> impl Future<Output = bool> + Send + 'static {
+        let conn = self.0.clone();
+        async move {
+            let notify;
+            let (mut notified, out) = {
+                let conn = conn.state.lock("resumption_tickets_received");
+                let (notified, out) = match conn.resumption_tickets.as_ref() {
+                    Some(ResumptionTicketState::Received) => (None, true),
+                    Some(ResumptionTicketState::Pending(n)) => {
+                        notify = n.clone();
+                        (Some(notify.notified()), true)
+                    }
+                    None => (None, false),
+                };
+                (notified, out)
+            };
+            if let Some(notified) = notified.take() {
+                notified.await;
+            }
+            out
+        }
+    }
 }
 
 pin_project! {
@@ -885,6 +916,10 @@ impl ConnectionRef {
         socket: Arc<dyn AsyncUdpSocket>,
         runtime: Arc<dyn Runtime>,
     ) -> Self {
+        let resumption_tickets = match conn.side() {
+            Side::Client => Some(ResumptionTicketState::Pending(Default::default())),
+            Side::Server => None,
+        };
         Self(Arc::new(ConnectionInner {
             state: Mutex::new(State {
                 inner: conn,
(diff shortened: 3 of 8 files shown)


Frando:
I am pretty sure the failure is unrelated. Are android tests known to be flaky?

quinn-proto/src/connection/mod.rs
```diff
@@ -3920,6 +3930,10 @@ pub enum Event {
     DatagramReceived,
     /// One or more application datagrams have been sent after blocking
     DatagramsUnblocked,
+    /// One or more TLS13 resumption tickets have been received.
```
flub:
```suggestion
    /// One or more TLS13 resumption tickets have been received
```

This repo's style guide deviates from the standard rust one here and does not use full stops on the header sentence.

quinn-proto/src/crypto.rs
```diff
@@ -91,6 +91,11 @@ pub trait Session: Send + Sync + 'static {
         label: &[u8],
         context: &[u8],
     ) -> Result<(), ExportKeyingMaterialError>;
+
+    /// Returns the number of TLS13 session tickets that were received.
```
flub:
```suggestion
    /// Returns the number of TLS13 session tickets that were received
```

quinn/src/connection.rs
```diff
@@ -636,6 +636,30 @@ impl Connection {
         // May need to send MAX_STREAMS to make progress
         conn.wake();
     }
+
+    /// Waits until the connection received TLS resumption tickets.
```
flub:
```suggestion
    /// Waits until the connection received TLS resumption tickets
```

quinn/src/connection.rs
```diff
@@ -636,6 +636,30 @@ impl Connection {
...
+    ///
+    /// Returns `true` once resumption tickets were received. Resolves immediately
```
flub: Perhaps "Yields" instead of "Returns"?  That seems to be the common terminology used here for futures.

quinn/src/connection.rs
```diff
@@ -636,6 +636,30 @@ impl Connection {
...
+    /// if tickets were already received, otherwise it resolves once tickets arrive.
+    /// If the server does not send any tickets, the returned future will remain pending forever.
+    ///
+    /// This should only be used on the client side. On the server side, it will
+    /// always resolve immediately and return `false`.
```
flub:
```suggestion
    /// always resolve immediately and yield `false`.
```

quinn-proto/src/connection/mod.rs
Frando: changed

quinn-proto/src/connection/mod.rs
```diff
@@ -3920,6 +3929,10 @@ pub enum Event {
     DatagramReceived,
     /// One or more application datagrams have been sent after blocking
     DatagramsUnblocked,
+    /// One or more TLS session resumption tickets have been received
+    ///
+    /// This event is only emitted on the client, and is emitted at most once per connection.
+    ResumptionTicketsReceived,
```
djc: This seems sort of specific to the TLS implementation -- our crypto traits are intended to be more general than that. Maybe this should be renamed to `ResumptionEnabled`?

Also, note that `Event` is not `non_exhaustive`, so this constitutes a semver-incompatible change. In general, this feels fairly invasive for what seems like a somewhat niche feature. Maybe there is some other way to structure this?

quinn-proto/src/crypto.rs
```diff
@@ -91,6 +91,11 @@ pub trait Session: Send + Sync + 'static {
         label: &[u8],
         context: &[u8],
     ) -> Result<(), ExportKeyingMaterialError>;
+
+    /// Returns the number of TLS1.3 session resumption tickets that were received
+    ///
+    /// Returns `None` on the server side.
+    fn resumption_tickets_received(&self) -> Option<u32>;
```
djc, line 98: Same here.

quinn-proto/src/connection/mod.rs
Frando: I'll rename the event :+1:

As to other ways to forward the info: I see these options:
- Keep the event, wait until next major release
- Add a public getter on proto::Connection for `resumption_received` or such, and check that in quinn's connection driver `poll`

Frando: WDTY - Should I keep the new event, or add a getter that is checked on each poll? Or did you have something else in mind for how to forward the event from `quinn-proto` to `quinn`?

djc: Would like to hear what @Ralith thinks. Personally I still think an event does not make sense for this.

djc:
@Frando want to continue with this?

Frando:
> @Frando want to continue with this?

Sure. It's ready from my side. There were open question and a request for further review which I can't seem to address myself.

I'm on vacation for the next two weeks, will rebase afterwards.

Frando:
I rebased, removed the patch for `rustls` (the upstream change was released) and renamed the event to `ResumptionEnabled` as suggested by @djc.

From my side the PR is ready. @djc wanted @Ralith's opinion on  the question whether this should be an event at all.  To me an event seems fine for this. The only alternative would be, I think, to check for a field on `quinn_proto::Connection` on each poll of the `quinn::ConnectionDriver::poll`. IMO an event is the cleaner way than a check on each poll. However, if the latter is preferred I can change it of course.

quinn-proto/src/connection/mod.rs
```diff
@@ -2113,6 +2116,12 @@ impl Connection {
             }
         }
 
+        if !self.resumption_tickets_received && self.crypto.resumption_tickets_received() > Some(0)
```
Ralith, line 2119: Is the number of tickets received guaranteed to change at most once in the lifetime of a connection? If not, should we raise an event every time it changes?

quinn-proto/src/crypto.rs
Ralith, line 98: Providing a default impl returning `None` would reduce churn for this niche use case.

quinn/src/tests.rs
```diff
@@ -384,6 +384,71 @@ async fn zero_rtt() {
     endpoint.wait_idle().await;
 }
 
+#[tokio::test]
+async fn zero_rtt_resumption() {
+    let _guard = subscribe();
+    let endpoint = endpoint();
+
+    let endpoint2 = endpoint.clone();
+    tokio::spawn(async move {
```
Ralith, line 393: Do we need to join this at the end of the test to ensure any panics are surfaced?

quinn/src/connection.rs
```diff
@@ -651,6 +651,37 @@ impl Connection {
...
+
+    /// Waits until the connection received TLS resumption tickets
+    ///
+    /// Yields `true` once resumption tickets were received. Resolves immediately
+    /// if tickets were already received, otherwise it resolves once tickets arrive.
+    /// If the server does not send any tickets, the returned future will remain pending forever.
+    ///
+    /// This should only be used on the client side. On the server side, it will
+    /// always resolve immediately and yield `false`.
+    pub fn resumption_tickets_received(&self) -> impl Future<Output = bool> + Send + 'static {
```
Ralith, line 663: Is anyone going to ever inspect this `bool`? Seems redundant to checking `side()`.

Ralith, line 663: How is the user meant to actually get the ticket? If the expectation is that they'll still be going through rustls's `ClientSessionStore`, would it be simpler to bypass quinn entirely and use a `ClientSessionStore` that notifies you directly on insert?

Ralith (commented):
No strong feeling on event vs. polling, and it's an implementation detail we can easily change.

quinn-proto/src/connection/mod.rs
Ralith: For Quinn to provide an async API, polling would need to happen on the critical path to trigger a `Notify` or similar, which strikes me as almost as invasive. An event is more plumbing, but doesn't add a branch that gets checked on every poll forever. If we're going to have this, I lean slightly towards an event, but I don't feel very strongly.

PR #2259, opened 2025-06-03
feat(quinn): Pass in `&mut self` for `AsyncUdpSocket`'s `poll_recv` and re-introduce `poll_send`

matheus23:
This PR contains some changes to the `AsyncUdpSocket` trait. In short:
- `poll_recv` now takes `&mut self` instead of `&self`
- `try_send` is removed from `AsyncUdpSocket`
- `create_io_poller` and the `UdpPoller` trait are now replaced with `create_sender` and the `UdpSender` trait
- the `UdpSender` trait has a `try_send` as well as a `poll_send` function (that takes `self: Pin<&mut Self>`) instead of a `poll_writable` function

The final traits look like this:
```rs
pub trait AsyncUdpSocket: Send + Sync + Debug + 'static {
    fn create_sender(&self) -> Pin<Box<dyn UdpSender>>;

    fn poll_recv(
        &mut self,
        cx: &mut Context,
        bufs: &mut [IoSliceMut<'_>],
        meta: &mut [RecvMeta],
    ) -> Poll<io::Result<usize>>;

    fn local_addr(&self) -> io::Result<SocketAddr>;

    fn max_receive_segments(&self) -> usize;

    fn may_fragment(&self) -> bool;
}

pub trait UdpSender: Send + Sync + Debug + 'static {
    fn poll_send(
        self: Pin<&mut Self>,
        transmit: &Transmit,
        cx: &mut Context,
    ) -> Poll<io::Result<()>>;

    fn max_transmit_segments(&self) -> usize;
    
    fn try_send(self: Pin<&mut Self>, transmit: &Transmit) -> io::Result<()>;
}
```

## Motivation

The general idea is to encode the invariants that the runtime and socket implementations need to adhere to into rust's type system.
One of the harder things for me was to understand initially when I was working with quinn's `AsyncUdpSocket` was when you're allowed to overwrite the waker you're getting from `poll_recv` and when you can't, and similarly for the old `UdpPoller`.
Also, with iroh we had annoying issues with passing state from `poll_writable` to `try_send`. With a single `poll_send` that gives you (essentially) `&mut self` access, this gets *much* easier.

With all of these changes it's also a lot clearer when it's fine to overwrite wakers in your custom implementation. If `&mut self` is passed in from your `poll` function, then you *know* you can only be called from a single task that owns you at a time, thus you can safely overwrite old wakers (and even do so without any use of mutexes or fancy linked lists or anything like that).

## Performance

I've tried testing the performance of this on various systems (my laptop, dedicated bare metal servers, etc.). I wasn't able to determine any differences, but at the same time, I always got very inconsistent results (both on quinn `main` and this PR).
I'd really appreciate any help from someone who knows how to benchmark quinn effectively. There's some more things I can try, such as fiddling with the constants mentioned in https://github.com/quinn-rs/quinn/issues/1126, but other than that I'm out of ideas.
Looking at the code, this *should not* have a negative performance impact. I think there's technically 1 more Arc clone for initiating connections, but other than that it should result in pretty much the same amount of work/same sequence of syscalls.

### DIFF
--- a/quinn/src/runtime.rs
+++ b/quinn/src/runtime.rs
@@ -21,7 +21,7 @@ pub trait Runtime: Send + Sync + Debug + 'static {
     fn spawn(&self, future: Pin<Box<dyn Future<Output = ()> + Send>>);
     /// Convert `t` into the socket type used by this runtime
     #[cfg(not(wasm_browser))]
-    fn wrap_udp_socket(&self, t: std::net::UdpSocket) -> io::Result<Arc<dyn AsyncUdpSocket>>;
+    fn wrap_udp_socket(&self, t: std::net::UdpSocket) -> io::Result<Box<dyn AsyncUdpSocket>>;
     /// Look up the current time
     ///
     /// Allows simulating the flow of time for testing.
@@ -50,11 +50,11 @@ pub trait AsyncUdpSocket: Send + Sync + Debug + 'static {
     /// [`Waker`].
     ///
     /// [`Waker`]: std::task::Waker
-    fn create_sender(self: Arc<Self>) -> Pin<Box<dyn UdpSender>>;
+    fn create_sender(&self) -> Pin<Box<dyn UdpSender>>;
 
     /// Receive UDP datagrams, or register to be woken if receiving may succeed in the future
     fn poll_recv(
-        &self,
+        &mut self,
         cx: &mut Context,
         bufs: &mut [IoSliceMut<'_>],
         meta: &mut [RecvMeta],
--- a/quinn/src/runtime.rs
+++ b/quinn/src/runtime.rs
@@ -1,5 +1,5 @@
 use std::{
-    fmt::Debug,
+    fmt::{self, Debug},
     future::Future,
     io::{self, IoSliceMut},
     net::SocketAddr,
@@ -40,23 +40,17 @@ pub trait AsyncTimer: Send + Debug + 'static {
 
 /// Abstract implementation of a UDP socket for runtime independence
 pub trait AsyncUdpSocket: Send + Sync + Debug + 'static {
-    /// Create a [`UdpPoller`] that can register a single task for write-readiness notifications
+    /// Create a [`UdpSender`] that can register a single task for write-readiness notifications
+    /// and send a transmit, if ready.
     ///
     /// A `poll_send` method on a single object can usually store only one [`Waker`] at a time,
     /// i.e. allow at most one caller to wait for an event. This method allows any number of
-    /// interested tasks to construct their own [`UdpPoller`] object. They can all then wait for the
-    /// same event and be notified concurrently, because each [`UdpPoller`] can store a separate
+    /// interested tasks to construct their own [`UdpSender`] object. They can all then wait for the
+    /// same event and be notified concurrently, because each [`UdpSender`] can store a separate
     /// [`Waker`].
     ///
     /// [`Waker`]: std::task::Waker
-    fn create_io_poller(self: Arc<Self>) -> Pin<Box<dyn UdpPoller>>;
-
-    /// Send UDP datagrams from `transmits`, or return `WouldBlock` and clear the underlying
-    /// socket's readiness, or return an I/O error
-    ///
-    /// If this returns [`io::ErrorKind::WouldBlock`], [`UdpPoller::poll_writable`] must be called
-    /// to register the calling task to be woken when a send should be attempted again.
-    fn try_send(&self, transmit: &Transmit) -> io::Result<()>;
+    fn create_sender(self: Arc<Self>) -> Pin<Box<dyn UdpSender>>;
 
     /// Receive UDP datagrams, or register to be woken if receiving may succeed in the future
     fn poll_recv(
@@ -68,11 +62,6 @@ pub trait AsyncUdpSocket: Send + Sync + Debug + 'static {
 
     /// Look up the local IP address and port used by this socket
     fn local_addr(&self) -> io::Result<SocketAddr>;
-
-    /// Maximum number of datagrams that a [`Transmit`] may encode
-    fn max_transmit_segments(&self) -> usize {
-        1
-    }
 
     /// Maximum number of datagrams that might be described by a single [`RecvMeta`]
     fn max_receive_segments(&self) -> usize {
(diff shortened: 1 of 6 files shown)


matheus23:
When manually setting `const MAX_TRANSMIT_DATAGRAMS: usize = 100`, I get more consistent benchmarking results.
Here's what the criterion part of `cargo bench` says:
```
gso_false_gro_false_recvmmsg_false/throughput
                        Change within noise threshold.
gso_false_gro_false_recvmmsg_true/throughput
                        No change in performance detected.
gso_false_gro_true_recvmmsg_false/throughput
                        Change within noise threshold.
gso_false_gro_true_recvmmsg_true/throughput
                        No change in performance detected.
gso_true_gro_false_recvmmsg_false/throughput
                        Change within noise threshold.
gso_true_gro_false_recvmmsg_true/throughput
                        No change in performance detected.
gso_true_gro_true_recvmmsg_false/throughput
                        Change within noise threshold.
gso_true_gro_true_recvmmsg_true/throughput
                        Change within noise threshold.
```
(this is a comparison between `main` and this branch with the max transmit datagrams patch. I've removed lines in between.)

The other benchmark is very similar as well (but much noisier!).
```
# main:
test large_data_10_streams  ... bench:  13,008,394 ns/iter (+/- 6,523,637) = 806 MB/s
test large_data_1_stream    ... bench:   1,271,029 ns/iter (+/- 5,250,573) = 824 MB/s
test small_data_100_streams ... bench:  11,856,811 ns/iter (+/- 19,145,692)
test small_data_1_stream    ... bench:  23,830,707 ns/iter (+/- 1,572,245)
# matheus23/mut-self-upstream:
test large_data_10_streams  ... bench:  12,480,406 ns/iter (+/- 1,986,101) = 840 MB/s
test large_data_1_stream    ... bench:   1,294,919 ns/iter (+/- 5,450,900) = 809 MB/s
test small_data_100_streams ... bench:  11,770,875 ns/iter (+/- 21,641,366)
test small_data_1_stream    ... bench:  24,368,665 ns/iter (+/- 1,775,232)
```

quinn/src/runtime.rs
```diff
@@ -87,78 +76,161 @@ pub trait AsyncUdpSocket: Send + Sync + Debug + 'static {
...
+    /// The quinn endpoint uses this function when sending
+    ///
+    /// - A version negotiation response due to an unknown version
+    /// - A `CLOSE` due to a malformed or unwanted connection attempt
+    /// - A stateless reset due to an unrecognized connection
+    /// - A `Retry` packet due to a connection attempt when `use_retry` is set
+    ///
+    /// If sending in these cases fails, a well-behaved peer will re-try. Thus it's fine
+    /// if we drop datagrams sometimes with this function.
+    fn try_send(self: Pin<&mut Self>, transmit: &Transmit) -> io::Result<()>;
```
Ralith: Would it be simpler to omit this and call `poll_send` with a no-op waker?

quinn/src/runtime.rs
```diff
@@ -87,78 +76,161 @@ pub trait AsyncUdpSocket: Send + Sync + Debug + 'static {
...
+    ///
+    /// This struct implements [`UdpSender`] if `MakeWritableFn` produces a `WritableFut`.
+    ///
+    /// Also serves as a trick, since `WritableFut` doesn't need to be a named future,
+    /// it can be an anonymous async block, as long as `MakeWritableFn` produces that
+    /// anonymous async block type.
+    ///
+    /// The `UdpSenderHelper` generic type parameters don't need to named, as it will be
+    /// used in its dyn-compatible form as a `Pin<Box<dyn UdpSender>>`.
+    pub struct UdpSenderHelper<Socket, MakeWritableFutFn, WritableFut> {
```
Ralith: I don't love committing to this (now even more) complex and awkward utility as part of our public API. Can we split that out for separate discussion?

quinn/src/runtime.rs
```diff
@@ -266,7 +266,7 @@ mod tokio;
 #[cfg(feature = "runtime-tokio")]
 pub use self::tokio::TokioRuntime;
 
-#[cfg(feature = "async-io")]
```
Ralith: Might be better to cfg out `async_io::UdpSocket` specifically? The `AsyncTimer` impl could, theoretically, be useful on its own, and so long as we have a freestanding async-io feature we might as well use it for that.

Might also be best to place this fix in a separate PR so we can merge it independently.

Ralith (changes requested):
Thanks! I think this is a good direction overall, though we may want to delay merging somewhat as it's breaking.

Ralith:
> Performance

I see no reason to expect that these changes present any performance hazard, and I'm satisfied with the smoke-testing you've done to check that assumption.

quinn/src/runtime.rs
matheus23: Fair! I know `UdpPollHelper` used to be private, and I flipped this to public since I've found it quite useful, but I don't mind making it private - if I really need it I can copy it out for my own use cases.

quinn/src/runtime.rs
matheus23: I had to change this because for some reason keeping [this causes a CI failure](https://github.com/quinn-rs/quinn/actions/runs/15419699076/job/43391217488), while it didn't before?

Given it causes a CI failure, I'm not 100% sure we can merge this independently.

I had a version that cfg-ed out `UdpSocket` and all the imports needed for it, but it's quite a bit noisier, and I don't really understand why it's needed for this PR, but not for `main` :/ I'll take another look.

matheus23: In addition to that, even though the `async_io` module exists with only the `async-io` feature, it won't be exported, because the `pub use self::async_io::*` line below is feature-gated to the smol and async-std feature.
I don't think this makes sense, these two cfgs should just agree.

So. I can't *not* do changes here, as this breaks CI. But any other change doesn't seem sensical to me (and would be much bigger diffs). I'd argue this is the best fix?

quinn/src/runtime.rs
matheus23: Unfortunately, you don't have a dependency that provides a no-op waker yet. There's a `std::task::Waker::noop()`, but it's only stable since rust 1.85, which is far younger than your MSRV 1.71, so I've inlined the implementation.

quinn/src/runtime.rs
Ralith: > I don't really understand why it's needed for this PR, but not for main

That does seem strange. Maybe the feature testing is randomized? Maybe something in the toolchain updated since the last run on main?

> Given it causes a CI failure, I'm not 100% sure we can merge this independently.

I figured we'd merge it *first*, since with the proposed tweak it'd be non-breaking, and then this PR can be rebased and merged whenever we're ready to break.

> the pub use self::async_io::* line below is feature-gated to the smol and async-std feature.

We don't actually need to re-export anything for the trait impl on `Timer` to be available downstream. While it's a pretty niche hypothetical case where that's useful, masking it out is technically avoidable breakage.

matheus23: I have now rebased onto main after #2264

matheus23:
> Thanks! I think this is a good direction overall, though we may want to delay merging somewhat as it's breaking.

Just to spell it out: I think I've addressed all reviewing concerns, and I think we're waiting only for the right moment to merge these breaking changes?

I realized I might be wrong though, given you're only saying "good direction" and requesting changes, so I'm double-checking now.

djc:
> Just to spell it out: I think I've addressed all reviewing concerns, and I think we're waiting only for the right moment to merge these breaking changes?

That matches my understanding. We want to get a release out before we start merging incompatible changes.

quinn/src/runtime.rs
```diff
@@ -87,77 +76,149 @@ pub trait AsyncUdpSocket: Send + Sync + Debug + 'static {
...
 }
 
-impl<MakeFut, Fut> UdpPollHelper<MakeFut, Fut> {
-    /// Construct a [`UdpPoller`] that calls `make_fut` to get the future to poll, storing it until
-    /// it yields [`Poll::Ready`], then creating a new one on the next
-    /// [`poll_writable`](UdpPoller::poll_writable)
+impl<Socket, MakeWritableFutFn, WritableFut> Debug
+    for UdpSenderHelper<Socket, MakeWritableFutFn, WritableFut>
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
djc: Nit: import `fmt` please.

quinn/src/runtime.rs
```diff
@@ -87,77 +76,149 @@ pub trait AsyncUdpSocket: Send + Sync + Debug + 'static {
...
             // a new `Future` to be created on the next call.
-            this.fut.set(None);
+            this.writable_fut.set(None);
+
+            // If .writable() fails, propagate the error
+            result?;
+
+            let result = this.socket.try_send(transmit);
+
+            match result {
```
djc: Nit: inline the `result` binding?

quinn/src/runtime.rs
```diff
@@ -87,77 +76,149 @@ pub trait AsyncUdpSocket: Send + Sync + Debug + 'static {
...
+                // We thought the socket was writable, but it wasn't, then retry so that either another
+                // `writable().await` call determines that the socket is indeed not writable and
+                // registers us for a wakeup, or the send succeeds if this really was just a
+                // transient failure.
+                Err(ref e) if e.kind() == io::ErrorKind::WouldBlock => continue,
```
djc: Nit: prefer to avoid `ref` in favor of match ergonomics.

quinn/src/runtime/async_io.rs
```diff
@@ -103,17 +103,27 @@ impl UdpSocket {
 }
 
 #[cfg(any(feature = "runtime-smol", feature = "runtime-async-std"))]
-impl AsyncUdpSocket for UdpSocket {
-    fn create_io_poller(self: Arc<Self>) -> Pin<Box<dyn super::UdpPoller>> {
-        Box::pin(UdpPollHelper::new(move || {
-            let socket = self.clone();
-            async move { socket.io.writable().await }
-        }))
+impl super::UdpSenderHelperSocket for Arc<UdpSocket> {
```
djc: Nit: import `UdpSenderHelperSocket` directly.

matheus23:
Addressed nits and rebased.

matheus23:
> I think there should be a bunch more comments explaining the design here. The previous design was already pretty complex and these changes only make it more so.

I can try adding more comments. Any place in particular you'd see these comments in?

> I don't think a combination of `UdpSender`, `UdpSenderHelper` (with pretty inscrutable generic arguments) and `UdpSenderHelperSocket` make for code that will be easy to maintain going forward.

... It's very disappointing to hear this. I want to point out that this can read as you voicing your opinion and being entirely unactionable from my side. I'll try to instead read this as a prompt for improving things.

It's admittedly a lot of work for possibly little benefit: The whole point of the `UdpSenderHelper` struct and the `UdpSenderHelperSocket` trait is to reduce code duplication between the smol and tokio runtime implementations.
I merely extended the previous abstraction so it works with the changed design, but perhaps it'd be better to `impl UdpSender` twice, onec for smol and onec for tokio. Would you prefer that?

Otherwise I don't see a way of improving the abstraction beyond further renames and documentation (which I've [discussed with colleagues](https://github.com/n0-computer/quinn/pull/67#discussion_r2109116213) in the past), if you have any suggestions please let me know.

djc:
> It's admittedly a lot of work for possibly little benefit: The whole point of the `UdpSenderHelper` struct and the `UdpSenderHelperSocket` trait is to reduce code duplication between the smol and tokio runtime implementations. I merely extended the previous abstraction so it works with the changed design, but perhaps it'd be better to `impl UdpSender` twice, onec for smol and onec for tokio. Would you prefer that?

I think it would be good to review the diff for that on top of the current state to compare the two approaches. Would be great if you can work on that.

> ... It's very disappointing to hear this. I want to point out that this can read as you voicing your opinion and being entirely unactionable from my side. I'll try to instead read this as a prompt for improving things.

Yes, I do realize this and I'm sorry it's not very actionable advice. I'm not necessarily opposed to merging this in its current state but I would like to explore if there are ways to make this part of the code more approachable.

matheus23:
> I think it would be good to review the diff for that on top of the current state to compare the two approaches. Would be great if you can work on that.

I looked into this and sadly I think I was wrong earlier. This is impossible without performance implications, actually.
Even if the `UdpSenderHelper` were only used in one place, it would literally be impossible to change the abstract implementation into a concrete one, because you *need* the `WriteableFut` generic parameter, as you can't "name the future" that's returned from `tokio::net::UdpSocket::writable`. (The same goes for the `UdpPollHelper` on `main`)
The only option for naming that future is to box it, but that might mean worse performance.
The reason we don't need to box the future is that the generic argument is inferred to be an unnamed type in the compiler before we're forced to "name it" in syntax, by boxing the `UdpSenderHelper`/`UdpPollHelper` in its entirety in `AsyncUdpSocket::create_sender`/`AsyncUdpSocket::create_io_poller`.

Ralith:
Something like the `UdpPollHelper` is absolutely necessary if we don't want boxing on every operation.

I do suspect that, with enough effort, we might be able to get rid of `UdpSenderHelperSocket`. For example, maybe `UdpSenderHelper` could own the `UdpSocketState` directly?

That said, given that these are implementation details, that the public API change here is a clear win, and that this has already been open for a few months, I lean towards merging and iterating further in follow-up work.

matheus23:
> For example, maybe `UdpSenderHelper` could own the `UdpSocketState` directly?

Unfortunately the implementation of `UdpSenderHelperSocket` references both the `UdpSocket` and the `UdpSocketState` in its `try_send` implementation, and `UdpSocket` is a different type between `runtime-smol` and `runtime-tokio`.

djc (approved):
Okay, let's get this in as is for now.

Ralith (approved):

PR #2261, opened 2025-06-04
expose `Connection::side()` in `quinn`

abonander:
[`Side` is exported in `quinn`](https://docs.rs/quinn/latest/quinn/enum.Side.html) but there's no actual way to get it from `Connection`. Seems like a simple oversight. 

Otherwise, this means it must be passed around separately to do any kind of client- or server-specific logic, which gets quite annoying.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -25,8 +25,8 @@ use crate::{
     udp_transmit,
 };
 use proto::{
-    ConnectionError, ConnectionHandle, ConnectionStats, Dir, EndpointEvent, StreamEvent, StreamId,
-    congestion::Controller,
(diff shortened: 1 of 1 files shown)


flub (approved):

djc (approved):
Seems okay to me.

Ralith (approved):
Funnily enough, we already reexport the enum, but nothing that yields it.

Ralith:
> this means it must be passed around separately to do any kind of client- or server-specific logic

I think normally people are already in client- or server-specific logic when they get a `Connection` in the first place, which is probably why this hasn't come up yet.

abonander:
This is for a peer-to-peer application where connections are happening both ways and it uses largely the same code paths. Checking the side of a connection is mainly for logging and deciding which connection to drop and which to keep when both sides open one at the same time.

PR #2264, opened 2025-06-06
refactor(quinn): Configure out `async_io::UdpSocket` when unused

matheus23:
For some reason I'm not getting a "`UdpSocket` is unused" when I run `cargo check -p quinn --no-default-features --features async-io", even though cfg-ing it out totally works and it fails in conjunction with other changes (see #2259).

Previous discussion: https://github.com/quinn-rs/quinn/pull/2259#discussion_r2127726193


djc:
Thanks!

PR #2273, opened 2025-06-19
Catch EOPNOTSUPP for opportunistic socket flags

gongy:
The [OS error raised by gvisor](https://github.com/google/gvisor/blob/master/pkg/sentry/socket/netstack/netstack.go#L1638) when setting `SOL_IP` options on UDP sockets is `EOPNOTSUPP`.

Since the flags are labeled as opportunistic on Linux, this patch allows the connection to be made by considering `EOPNOTSUPP` an acceptable code.


djc (approved):

mxinden (approved):
Function docs need updating. Otherwise looks good to me.

djc:
@gongy would you be able to finish this with some documentation updates?

djc:
@gongy ping?

PR #2274, opened 2025-06-20
proto: Add option to pad all application data packets to MTU

FlorianUekermann:
The goal is to pad all outgoing packets that contain application data to MTU.


djc:
Why?

FlorianUekermann:
I updated the PR description with our motivation.

Ralith:
Can't roughly the same analysis still be performed based on timing?

FlorianUekermann:
> Can't roughly the same analysis still be performed based on timing?

For our usecase:

Yes, both timing and size are relevant features. Removing size makes association of individual packets within a time window a lot harder. And the application layer is free to delay delay packets or even generate fake traffic. Packet size is just the one traffic feature where the transport implementation (e.g. quinn) needs to cooperate.
Note that with many users sharing an outgoing relay IP, association has to happen per connection (basically the pair of outgoing port + dest), so for some connections the amount of observable information is very limited and attribution of relay <-> dest traffic to a tunnel has to be done per connection.

In general (affects our usecase and others):

Packet sizes patterns can can be used as client or traffic type fingerprints [1].

[1] https://www.usenix.org/conference/usenixsecurity24/presentation/xue-fingerprinting

quinn-proto/src/connection/mod.rs
```diff
@@ -904,6 +910,9 @@ impl Connection {
             if pad_datagram {
                 builder.pad_to(MIN_INITIAL_SIZE);
             }
+            if pad_datagram_to_mtu {
+                builder.pad_to(self.path.current_mtu());
```
Ralith: I think this is incorrect for tail loss probes, which must never exceed `INITIAL_MTU` so that losses due to an MTU black hole can be detected and handled. Instead, we should pad to whatever the size limit is for the current datagram (see e.g. `PacketBuilder::max_size`, but note this is an absolute offset in the GSO batch).

Ralith (changes requested):
Ah, yeah, makes sense that this would be pretty effective when there's many users.

quinn-proto/src/connection/mod.rs
FlorianUekermann: Yes, you are right. This was incorrect for tail loss probes. Same issue in the GSO batching code.

It took me some time to fully understand the whole `poll_transmit` logic, but it's fixed now. I've update the PR to only introduce one additional `builder.pad_to()`, which is protected by the loss probe check. The regular GSO batch padding can take care of the rest with a small modification.

The CI i686-linux-android failure is new and odd. I don't see how any change in behavior is possible with `self.config.pad_to_mtu == false` with this PR.

djc: > The CI i686-linux-android failure is new and odd. I don't see how any change in behavior is possible with `self.config.pad_to_mtu == false` with this PR.

I think that's a known intermittent:

- https://github.com/quinn-rs/quinn/issues/1695

quinn-proto/src/config/transport.rs
```diff
@@ -203,6 +204,12 @@ impl TransportConfig {
         self
     }
 
+    /// Specifies if all application data packets should be padded to current maximum UDP payload size.
```
Ralith: Let's add a few words explaining when this might/might not be desirable. Also, nit: no trailing period in the first line.

Ralith: Do we want to pad all packets in the application data space, or just all packets which carry application data (e.g. not ACK-only)? The former is implemented, but the wording here is a little ambiguous as to what the intent is; please clarify.

Ralith (changes requested):
I still need to take a careful look at the interaction with GSO batching.

quinn-proto/src/config/transport.rs
FlorianUekermann: done

FlorianUekermann: Either works for for us, so whichever you prefer. For now I described the weaker variant to allow for implementation flexibility going forward.

I'm not sure how to best exclude ack-only UDP datagrams from padding if you would prefer that.

Ralith: I'm happy with the current behavior because it's simpler. Padding ACK-only datagrams will of course increase bandwidth consumption, but whether that matters is an application/deployment-specific question.

Ralith (approved):
LGTM. It's a little non-obvious how `segment_size` is guaranteed to always be the MTU when this option is set, but I don't immediately see a way to make that clearer, the change in logic is basically simple, and I am satisfied that it's correct.

djc:
Thanks!

djc:
Hmm, not sure why CI is failing in the merge queue:

```rust
error[E0283]: type annotations needed
    --> quinn-proto/src/tests/mod.rs:3256:52
     |
3256 |     assert_eq!(pair.client.outbound[0].0.size, MTU.into());
     |                                                    ^^^^
     |
     = note: cannot satisfy `_: From<u16>`
     = note: required for `u16` to implement `Into<_>`
help: try using a fully qualified path to specify the expected types
     |
3256 -     assert_eq!(pair.client.outbound[0].0.size, MTU.into());
3256 +     assert_eq!(pair.client.outbound[0].0.size, <u16 as Into<T>>::into(MTU));
```

FlorianUekermann:
Strange... Maybe a different Rust version? I'll change `MTU.into()` to `usize::from(MTU)`. The error looks like that would fix it.

PR #2278, opened 2025-06-25
Collect qlog recovery metrics

stormshield-fabs:
This PR implements the initial support for emitting qlog events, using the [`qlog`](https://docs.rs/qlog/latest/qlog/) crate from Cloudflare's [`quiche`](https://docs.rs/quiche/latest/quiche/). It focuses on the [`recovery_metrics_updated`](https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-quic-events.html#name-recovery_metrics_updated) event and is largely inspired from the `quiche` implementation.

It introduces a `qlog` feature in both `quinn_proto` and `quinn`, and adds a `set_qlog` method on `quinn::Connection` that forwards to `quinn_proto::Connection::set_qlog` to register a writer for the JSON-SEQ output. 

The entrypoints for the metrics updates are inspired from the `quiche` implementation:
- when a packet is emitted (`poll_transmit` for `quinn`)
- when a packet is received (`handle_packet`, `handle_first_packet`)
- when a packet loss is detected (`handle_timeout`)

According to the draft RFC, the metrics are only emitted when their value has been updated.

The resulting file can be imported into [`qvis`](https://qvis.quictools.info/#/files) but sadly the x-axis does not scale correctly (I suspect it uses the timestamps of the other event kinds that we don't emit yet, but I did not look into it). For now, I'm using `matplotlib` to plot the values.

~~I'm slightly unhappy with the `Controller::ssthresh` and `Controller::pacing_rate` getter, since they leak implementation details. Maybe an API like `Controller::update_metrics(&self, &mut CongestionMetrics)` would be more appropriate?~~


djc:
> I'm slightly unhappy with the `Controller::ssthresh` and `Controller::pacing_rate` getter, since they leak implementation details. Maybe an API like `Controller::update_metrics(&self, &mut CongestionMetrics)` would be more appropriate?

Yeah, that sounds right.

stormshield-fabs:
I've added a commit for the proposed `update_metrics` pattern.

quinn-proto/src/congestion.rs
```diff
@@ -65,15 +67,8 @@ pub trait Controller: Send + Sync {
...
-    fn ssthresh(&self) -> Option<u64> {
-        None
-    }
-
-    /// Pacing rate in bytes per second.
-    fn pacing_rate(&self) -> Option<u64> {
-        None
-    }
+    /// Update implementation-specific metrics.
+    fn update_metrics(&self, metrics: &mut CongestionMetrics);
```
djc: I think this needs a noop default implementation to be semver-compatible.

stormshield-fabs: Sure, that makes sense! Fixed this and the associated feature gates.

quinn-proto/src/connection/mod.rs
```diff
@@ -362,6 +435,85 @@ impl Connection {
         this
     }
 
+    /// Set up qlog for this connection.
+    #[cfg(feature = "qlog")]
+    pub fn set_qlog(
+        &mut self,
+        writer: Box<dyn std::io::Write + Send + Sync>,
+        title: String,
+        description: String,
```
Ralith: Should these be `Option`? They don't seem to be required by the crate, and `None` is probably more convenient for a lot of quick and dirty stuff.

quinn-proto/src/connection/mod.rs
```diff
@@ -362,6 +435,85 @@ impl Connection {
...
+        let level = qlog::events::EventImportance::Core;
+
+        let trace = qlog::TraceSeq::new(
+            qlog::VantagePoint {
+                name: None,
+                ty: vp,
+                flow: None,
+            },
+            Some(title.to_string()),
+            Some(description.to_string()),
```
Ralith: Prefer `clone()` over `to_string()` where applicable for clarity of intent.

quinn-proto/src/connection/mod.rs
```diff
@@ -362,6 +435,85 @@ impl Connection {
...
+            }),
+            None,
+        );
+
+        let mut streamer = qlog::streamer::QlogStreamer::new(
+            qlog::QLOG_VERSION.to_string(),
+            Some(title),
+            Some(description),
+            None,
+            std::time::Instant::now(),
```
Ralith: To aid deterministic testing, we prefer not to call `Instant::now` in `quinn-proto`, instead letting the caller pass an `Instant` in. Should we preserve that pattern here? Might not matter since this is just a diagnostic, but deterministic qlog traces could conceivably be a useful tool.

quinn-proto/src/connection/mod.rs
```diff
@@ -362,6 +435,85 @@ impl Connection {
...
+
+        streamer.start_log().ok();
+
+        self.qlog_streamer = Some(streamer);
+    }
+
+    /// Emit a `MetricsUpdated` event to the qlog streamer containing only the updated values.
+    #[cfg(feature = "qlog")]
+    fn emit_qlog_recovery_metrics(&mut self) {
+        if self.qlog_streamer.is_none() {
```
Ralith: style: could we use let-else here rather than matching on the pattern redundantly below?

quinn-proto/src/connection/mod.rs
```diff
@@ -362,6 +435,85 @@ impl Connection {
         this
     }
 
+    /// Set up qlog for this connection.
+    #[cfg(feature = "qlog")]
+    pub fn set_qlog(
```
Ralith, line 380: Would it make sense for this to be part of `EndpointConfig`? IIUC qlogs may span multiple connections, and include events not associated with a connection.

quinn-proto/src/connection/mod.rs
```diff
@@ -362,6 +435,86 @@ impl Connection {
...
+    /// Emit a `MetricsUpdated` event to the qlog streamer containing only the updated values.
+    #[cfg(feature = "qlog")]
+    fn emit_qlog_recovery_metrics(&mut self) {
+        if self.qlog_streamer.is_none() {
+            return;
+        }
+
+        let rtt = &self.path.rtt;
+
+        let mut metrics = CongestionMetrics {
```
Ralith: All but one of the fields here are come from `self.path`. Should this be lifted out into a helper method on it? The `connection` module is already huge so it's nice to keep new logic outside it when possible.

quinn-proto/src/connection/mod.rs
```diff
@@ -233,6 +233,74 @@ pub struct Connection {
...
+    qlog_streamer: Option<qlog::streamer::QlogStreamer>,
+
+    /// Snapshot of the qlog recovery metrics
+    #[cfg(feature = "qlog")]
+    recovery_metrics: CongestionMetrics,
+}
+
+#[cfg(feature = "qlog")]
+#[derive(Default, PartialEq)]
+pub struct CongestionMetrics {
```
Ralith: Should this be `#[non_exhaustive]`? That would allow us to add new fields in the future without breaking congestion controller impls.

quinn-proto/src/congestion.rs
```diff
@@ -65,6 +67,10 @@ pub trait Controller: Send + Sync {
     /// Number of ack-eliciting bytes that may be in flight
     fn window(&self) -> u64;
 
+    /// Update implementation-specific metrics.
+    #[cfg(feature = "qlog")]
```
Ralith: I don't think trait methods should be feature-gated: an external congestion controller implementation would break if the feature was enabled on `quinn`. Because `CongestionMetrics` is a simple type we control, there shouldn't be much drawback to exposing this method unconditionally.

quinn-proto/src/connection/mod.rs
```diff
@@ -233,6 +233,74 @@ pub struct Connection {
...
+    pub congestion_window: Option<u64>,
+    pub min_rtt: Option<Duration>,
+    pub smoothed_rtt: Option<Duration>,
+    pub latest_rtt: Option<Duration>,
+    pub rtt_variance: Option<Duration>,
+    pub pto_count: Option<u32>,
+    pub bytes_in_flight: Option<u64>,
+    pub packets_in_flight: Option<u64>,
+    pub ssthresh: Option<u64>,
+    pub pacing_rate: Option<u64>,
```
Ralith: A lot of these metrics don't seem to be the business of the congestion controller to track or report. Should we split out a separate struct that is exclusively congestion-controller-related for the benefit of the trait method? That would help keep the public interface narrower.

Ralith (changes requested):
Thanks for getting this kicked off!

quinn-proto/src/connection/mod.rs
stormshield-fabs: yes, I moved this logic to `PathData`:

```rust
fn congestion_metrics(&self, pto_count: u32) -> CongestionMetrics
```

stormshield-fabs:
It looks like this requires an MSRV bump to `1.74.1` (according to `cargo-msrv`). Would that be acceptable?

Ralith:
That was more than a year ago; should be fine!

quinn-proto/src/connection/mod.rs
stormshield-fabs: I believe the streaming version of qlog does not support multiple connections (https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-main-schema#name-qlogfileseq-schema).

I like the idea of configuring qlog earlier but I'm not sure what the API should look like since we'd need to provide a new writer for each connection.

djc:
(Should this be marked ready for review?)

quinn-proto/src/connection/mod.rs
Ralith: > I believe the streaming version of qlog does not support multiple connections

I don't immediately see where this is specified -- can you cite specific text?

> I'm not sure what the API should look like since we'd need to provide a new writer for each connection.

If they do need to be separate files, a config could take a `Box<dyn Fn() -> Box<dyn Write + Send + Sync>>` or similar. Maybe `Fn(&Connection)`?

quinn-proto/src/congestion.rs
```diff
@@ -65,6 +65,15 @@ pub trait Controller: Send + Sync {
     /// Number of ack-eliciting bytes that may be in flight
     fn window(&self) -> u64;
 
+    /// Update implementation-specific metrics.
```
Ralith: Style nit: strip trailing period

Ralith: Maybe mention that this is used to populate qlog traces when they're enabled? Might be pretty mysterious otherwise.

quinn-proto/src/congestion.rs
```diff
@@ -82,3 +91,14 @@ pub trait ControllerFactory {
 }
 
 const BASE_DATAGRAM_SIZE: u64 = 1200;
+
+/// Common congestion controller metrics
+#[non_exhaustive]
+pub struct ControllerMetrics {
```
Ralith, line 90: This needs `#[derive(Default)]` or `#[non_exhaustive]` will make it impossible to construct externally.

quinn-proto/src/connection/mod.rs
```diff
@@ -2268,7 +2345,13 @@ impl Connection {
                             packet.header.is_1rtt(),
                         );
                     }
-                    self.process_decrypted_packet(now, remote, number, packet)
+
+                    let result = self.process_decrypted_packet(now, remote, number, packet);
+
+                    #[cfg(feature = "qlog")]
+                    self.emit_qlog_recovery_metrics(now);
```
Ralith: If the intent is to call this only after a datagram is fully processed, we should probably lift it into `handle_event`, after the call to `handle_coalesced`, just like how the other callsite added in this PR follows the `handle_coalesced` call in `handle_first_packet`. As is, this might get called multiple times for one datagram, although that probably isn't harmful.

quinn-proto/src/connection/paths.rs
```diff
@@ -185,6 +287,16 @@ impl RttEstimator {
...
+    /// The latest RTT estimation.
+    pub fn latest(&self) -> Duration {
+        self.latest
+    }
+
+    /// RTT variance as described in RFC6298.
+    pub fn variance(&self) -> Duration {
+        self.var
+    }
+
```
Ralith: Do we still need these now that metrics are gathered in this module?

quinn-proto/src/connection/mod.rs
```diff
@@ -362,6 +369,66 @@ impl Connection {
...
+    }
+
+    /// Emit a `MetricsUpdated` event to the qlog streamer containing only the updated values.
+    #[cfg(feature = "qlog")]
+    fn emit_qlog_recovery_metrics(&mut self, now: Instant) {
+        let Some(qlog_streamer) = &mut self.qlog_streamer else {
+            return;
+        };
+
+        let metrics = self.path.qlog_congestion_metrics(self.pto_count);
```
Ralith: Should we handle the case where `metrics` is all `None` specially? Probably a bit error-prone, but might save a lot of space in the output. Maybe the qlog crate is already filtering those for us? Might merit a comment noting that if so.

Ralith (changes requested):
Looking good!

quinn-proto/src/congestion.rs
djc: Also, let's move this above the `ControllerFactory` definition.

quinn-proto/src/connection/mod.rs
```diff
@@ -233,6 +233,10 @@ pub struct Connection {
     stats: ConnectionStats,
     /// QUIC version used for the connection.
     version: u32,
+
+    /// qlog streamer to ouput events as JSON text sequences
+    #[cfg(feature = "qlog")]
+    qlog_streamer: Option<qlog::streamer::QlogStreamer>,
```
djc: Nit: prefer to import the `QlogStreamer` (especially since it has a crate-specific name anyway), with a conditional import at the top of the module.

quinn-proto/src/connection/mod.rs
```diff
@@ -362,6 +369,66 @@ impl Connection {
...
+            title,
+            description,
+            None,
+            now,
+            trace,
+            level,
+            writer,
+        );
+
+        streamer.start_log().ok();
```
djc: Why not propagate any errors here?

quinn-proto/src/connection/mod.rs
```diff
@@ -362,6 +369,66 @@ impl Connection {
...
+    /// Set up qlog for this connection.
+    #[cfg(feature = "qlog")]
+    pub fn set_qlog(
+        &mut self,
+        writer: Box<dyn std::io::Write + Send + Sync>,
+        title: Option<String>,
+        description: Option<String>,
+        now: Instant,
+    ) {
+        let vp = if self.side.is_server() {
```
djc: Nit: let's import all of these types. Call the `vp` binding `ty` instead to match the upstream name?

quinn-proto/src/connection/mod.rs
```diff
@@ -362,6 +369,66 @@ impl Connection {
         this
     }
 
+    /// Set up qlog for this connection.
+    #[cfg(feature = "qlog")]
+    pub fn set_qlog(
+        &mut self,
+        writer: Box<dyn std::io::Write + Send + Sync>,
```
djc: Nit: import `std::io` at the top of the module.

quinn-proto/src/connection/mod.rs
```diff
@@ -362,6 +369,66 @@ impl Connection {
...
+    /// Emit a `MetricsUpdated` event to the qlog streamer containing only the updated values.
+    #[cfg(feature = "qlog")]
+    fn emit_qlog_recovery_metrics(&mut self, now: Instant) {
+        let Some(qlog_streamer) = &mut self.qlog_streamer else {
+            return;
+        };
+
+        let metrics = self.path.qlog_congestion_metrics(self.pto_count);
+        let event = qlog::events::EventData::MetricsUpdated(metrics);
+        let _ = qlog_streamer.add_event_data_with_instant(event, now);
```
djc: What's the rationale for ignoring any errors from this?

perf/src/bin/perf_client.rs
```diff
@@ -170,11 +174,17 @@ async fn run(opt: Opt) -> Result<()> {
...
 
-    let connection = endpoint
+    let mut connection = endpoint
         .connect_with(config, addr, host_name)?
         .await
         .context("connecting")?;
 
+    #[cfg(feature = "qlog")]
+    if let Some(qlog_file) = &opt.qlog_file {
+        let writer = std::fs::File::create(qlog_file)?;
```
djc: Nit: import `File` directly.

quinn-proto/src/connection/mod.rs
stormshield-fabs: Yes, `qlog` directly serializes the event so we might end up with empty events in the output. 

This situation is now handled.

quinn-proto/src/connection/mod.rs
stormshield-fabs: As I understand it, there are two ways to collect qlog traces: 
1. a [`QlogFile` that "can contain several individual traces and logs from multiple vantage points"](https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-main-schema#section-4)
2. a [`QlogFileSeq` that uses the streamable JSON format](https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-main-schema#section-5)

A `QlogFileSeq` contains a single `TraceSeq` whereas a `QlogFile` holds an array of `Trace`s. The latter requires some kind of buffering but traces can get quite big (we reach 70MB for a minute-long trace with only the recovery metrics) so I decided to use the streaming version (this is also the choice made by `quiche`).

A writer factory pattern would work indeed.

quinn-proto/src/connection/mod.rs
stormshield-fabs: Some callsites are not fallible so I wasn't sure how to propagate the error. Also it sounds wrong that the instrumentation might interfere with the connection.

Maybe we could emit a log on the first error we encounter and drop the `streamer`?

djc: Yeah, not sure propagation is the right solution, but I think we definitely want to at least log something.

Could also `unwrap()` potentially, depending on what might cause the errors here?

stormshield-fabs: The errors would raise from a Writer error (for example a full disk). I'll add log statements and discard the streamer accordingly (the same applies for `start_log`).

quinn-proto/src/connection/mod.rs
stormshield-fabs: I won't have the time to implement this before going on vacation for a month on Friday. 

Could we say that this interface in unstable while the `qlog` feature settles? Otherwise I'm fine with someone else amending this PR.

djc: > Could we say that this interface in unstable while the `qlog` feature settles? Otherwise I'm fine with someone else amending this PR.

Seems reasonable to me, maybe change the Cargo feature to be `__qlog`?

stormshield-fabs: renamed the feature from `qlog` to `__qlog`

quinn-proto/src/connection/mod.rs
Ralith, line 1: These changes got amended into the wrong commit.

Ralith (approved):
LGTM, thanks! We can iterate further on configuration after merging.

quinn-proto/src/connection/mod.rs
stormshield-fabs: ah yes, fixed

djc (approved):

djc:
Should we rename `sstresh` to `ss_threshold`? Feels more idiomatic to me, but maybe using the term of art should win?

Ralith:
`ssthresh` is the exact name used in the RFC, which is nice insofar as that the RFC is relevant to a particular implementation. Otherwise, maybe `slow_start_threshold`?

Ralith:
For visibility: a qlog `TraceSeq` is *not* scoped to one connection: [§ Grouping](https://quicwg.org/qlog/draft-ietf-quic-qlog-main-schema.html#name-grouping):
> For example, a server implementation might choose to log events for all incoming connections in a single large (streamed) qlog file.

This is expressed by setting `group_id` per-event.

PR #2279, opened 2025-06-26
Various enhancement to quinn perf

stormshield-damiend:
Allow :
- selecting initial rtt
- enabling ack frequency
- selecting congestion control algorithm

Leave some async tasks early if they are not needed.

### DIFF
--- a/perf/src/bin/perf_client.rs
+++ b/perf/src/bin/perf_client.rs
@@ -244,6 +244,10 @@ async fn drain_stream(
     download: u64,
     stream_stats: OpenStreamStats,
 ) -> Result<()> {
+    if download == 0 {
+        return Ok(());
+    }
+
     #[rustfmt::skip]
     let mut bufs = [
         Bytes::new(), Bytes::new(), Bytes::new(), Bytes::new(),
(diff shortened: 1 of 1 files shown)


perf/src/bin/perf_client.rs
```diff
@@ -70,6 +70,9 @@ struct Opt {
     /// Disable packet encryption/decryption (for debugging purpose)
     #[clap(long = "no-protection")]
     no_protection: bool,
+    /// The initial round-trip-time (in msecs)
+    #[clap(long, default_value = "333")]
+    initial_rtt: u64,
```
djc: Maybe make this an `Option` so we can avoid restating the default value here?

(Also for the server.)

Ralith:
CI failures look unrelated.

perf/src/bin/perf_client.rs
stormshield-fabs: this also applies to the congestion algorithm parameter

djc: Yup!

stormshield-damiend: i am changing both

djc (approved):

djc:
Thanks!

PR #2282, opened 2025-06-29
Don't violate client CID limits when advertising a preferred address

Ralith:
Fixes https://github.com/quinn-rs/quinn/issues/2280.


quinn-proto/src/connection/mod.rs
```diff
@@ -3046,7 +3046,13 @@ impl Connection {
         }
 
         // Subtract 1 to account for the CID we supplied while handshaking
-        let n = self.peer_params.issue_cids_limit() - 1;
+        let mut n = self.peer_params.issue_cids_limit() - 1;
+        if let ConnectionSide::Server { ref server_config } = self.side {
```
djc: Nit: IMO `&self.side` instead of `ref` is more idiomatic these days.

PR #2284, opened 2025-06-30
Updated quinn::Connection::send_datagram() docs

gry8t:
Comments on behaviour when calling said function on a full datagram queue


djc (approved):
Thanks!

PR #2289, opened 2025-07-02
Implement spurious loss recovery

stormshield-fabs:
Benchmarks in the cloud between two Linux machines on a 10ms/1Gbps link revealed that `quinn` performs consistently worse than `msquic`. Plotting the congestion window for both implementations shows that `msquic` reaches higher congestion window sizes, in part because [it implements spurious loss recovery](https://github.com/microsoft/msquic/blob/2623c07df62b4bd171f469fb29c2714b6735b676/src/core/loss_detection.c#L1400) (as described in [rfc8312bis](https://datatracker.ietf.org/doc/html/draft-ietf-tcpm-rfc8312bis-15#section-4.9) which is now [rfc9438](https://datatracker.ietf.org/doc/html/rfc9438#section-4.9)).

This PR implements a similar spurious loss recovery mechanism (`quinn` implements `rfc8312`)

Lost packets are stored in the `PacketSpace` and used on ACK reception to determine if all the packets we deemed lost were in fact acknowledged. This list is regularly purged from packets that were sent more than 2 PTO ago (this is the value used in `msquic`: https://github.com/microsoft/msquic/blob/2623c07df62b4bd171f469fb29c2714b6735b676/src/core/loss_detection.c#L925).

Overall, this appears to effectively improve throughput on the benchmark I described above: we've seen up to a 30% increase in upload bandwidth.

|         | main     | patch    | gain |
|---------|----------|----------|------|
| average | 314 Mbps | 391 Mbps | 25%  |
| P50     | 302 Mbps | 402 Mbps | 33%  |

### DIFF
--- a/quinn-proto/src/congestion.rs
+++ b/quinn-proto/src/congestion.rs
@@ -56,8 +56,15 @@ pub trait Controller: Send + Sync {
         now: Instant,
         sent: Instant,
         is_persistent_congestion: bool,
+        is_ecn: bool,
         lost_bytes: u64,
     );
+
+    /// Packets were incorrectly deemed lost
+    ///
+    /// This function is called when all packets that were deemed lost (for instance because
+    /// of packet reordering) are acknowledged after the congestion event was raised.
+    fn on_spurious_congestion_event(&mut self) {}
 
     /// The known MTU for the current network path has been updated
     fn on_mtu_update(&mut self, new_mtu: u16);
--- a/quinn-proto/src/congestion/cubic.rs
+++ b/quinn-proto/src/congestion/cubic.rs
@@ -75,8 +75,10 @@ impl State {
 #[derive(Debug, Clone)]
 pub struct Cubic {
     config: Arc<CubicConfig>,
-    state: State,
     current_mtu: u64,
+    state: State,
+    /// Copy of the controller state to restore when a spurious congestion event is detected.
+    pre_congestion_state: Option<State>,
 }
 
 impl Cubic {
@@ -89,6 +91,7 @@ impl Cubic {
                 ..Default::default()
             },
             current_mtu: current_mtu as u64,
+            pre_congestion_state: None,
             config,
         }
     }
@@ -176,6 +179,7 @@ impl Controller for Cubic {
         now: Instant,
         sent: Instant,
         is_persistent_congestion: bool,
+        is_ecn: bool,
         _lost_bytes: u64,
     ) {
         if self
@@ -185,6 +189,11 @@ impl Controller for Cubic {
             .unwrap_or(false)
         {
             return;
+        }
+
+        // Save state in case this event ends up being spurious
+        if !is_ecn {
+            self.pre_congestion_state = Some(self.state.clone());
         }
 
         self.state.recovery_start_time = Some(now);
@@ -218,6 +227,14 @@ impl Controller for Cubic {
             self.state.cwnd_inc = 0;
 
             self.state.window = self.minimum_window();
+        }
+    }
+
+    fn on_spurious_congestion_event(&mut self) {
+        if let Some(prior_state) = self.pre_congestion_state.take() {
+            if self.state.window < prior_state.window {
+                self.state = prior_state;
+            }
         }
     }
 
--- a/quinn-proto/src/congestion/cubic.rs
+++ b/quinn-proto/src/congestion/cubic.rs
@@ -19,12 +19,27 @@ const C: f64 = 0.4;
 /// k, w_max are described in the RFC.
 #[derive(Debug, Default, Clone)]
 pub(super) struct State {
+    /// Time period that the cubic function takes to increase the window size to W_max.
     k: f64,
 
+    /// Congestion window size when the last congestion event occurred.
     w_max: f64,
 
-    // Store cwnd increment during congestion avoidance.
+    /// Congestion window increment stored during congestion avoidance.
     cwnd_inc: u64,
+
+    /// Maximum number of bytes in flight that may be sent.
+    window: u64,
+
+    /// Slow start threshold in bytes.
+    ///
+    /// When the congestion window is below ssthresh, the mode is slow start
+    /// and the window grows by the number of bytes acknowledged.
+    ssthresh: u64,
+
+    /// The time when QUIC first detects a loss, causing it to enter recovery. When a packet sent
+    /// after this time is acknowledged, QUIC exits recovery.
+    recovery_start_time: Option<Instant>,
 }
 
 /// CUBIC Functions.
@@ -39,7 +54,7 @@ impl State {
         (w_max * (1.0 - BETA_CUBIC) / C).cbrt()
     }
 
-    // W_cubic(t) = C * (t - K)^3 - w_max (Eq. 1)
+    // W_cubic(t) = C * (t - K)^3 + w_max (Eq. 1)
     fn w_cubic(&self, t: Duration, max_datagram_size: u64) -> f64 {
         let w_max = self.w_max / max_datagram_size as f64;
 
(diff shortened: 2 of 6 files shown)


stormshield-fabs:
The `msquic` congestion window plot that helped find this difference between both implementations
image

Before/after plots for `quinn` (the orange curve is the average of all samples)
image

quinn-proto/src/congestion/cubic.rs
```diff
@@ -70,6 +70,8 @@ pub struct Cubic {
     recovery_start_time: Option<Instant>,
     cubic_state: State,
     current_mtu: u64,
+    /// Copy of the controller state to restore on when a spurious congestion event is detected.
+    snapshot: Option<Box<Self>>,
```
Ralith: Let's refactor this into something like:
```rs
pub struct Cubic {
    config: Arc<CubicConfig>,
    current_mtu: u64,
    current_state: State,
    pre_congestion_state: Option<State>,
}
```
either by moving fields into the current `State` structure or by introducing another intermediate struct.

I'm not sure it's actually useful to avoid the boxing, but this way we avoid accidentally reverting `current_mtu`, and storing the config pointer twice is a bit funny besides.

quinn-proto/src/connection/mod.rs
```diff
@@ -1573,6 +1577,37 @@ impl Connection {
         Ok(())
     }
 
+    fn detect_spurious_loss(&mut self, ack: &frame::Ack, space: SpaceId) -> bool {
+        let lost_packets = &mut self.spaces[space].lost_packets;
+
+        if !lost_packets.is_empty() {
```
Ralith: style: reduce rightward drift by inverting this condition and returning early.

quinn-proto/src/connection/mod.rs
```diff
@@ -1573,6 +1577,37 @@ impl Connection {
...
+                    .map(|(pn, _info)| pn)
+                    .copied()
+                    .collect::<Vec<_>>();
+
+                for pn in spurious_losses {
+                    lost_packets.remove(&pn);
+                }
+            }
+
+            lost_packets.is_empty()
```
Ralith: Add comments discussing the behavior when we remove some, but not all elements of `lost_packets`. Intuitively that does seem like a case of spurious loss, so we should explain why we don't treat it as one.

quinn-proto/src/connection/mod.rs
```diff
@@ -1767,12 +1804,15 @@ impl Connection {
 
             for &packet in &lost_packets {
                 let info = self.spaces[pn_space].take(packet).unwrap(); // safe: lost_packets is populated just above
+
                 self.remove_in_flight(packet, &info);
-                for frame in info.stream_frames {
+                for frame in info.stream_frames.clone() {
```
Ralith: `.iter().cloned()` to save a possible alloc

Or this can be left as-is if `lost_packets` doesn't need the whole `info` at all.

quinn-proto/src/connection/spaces.rs
```diff
@@ -39,6 +39,8 @@ pub(super) struct PacketSpace {
     /// Transmitted but not acked
     // We use a BTreeMap here so we can efficiently query by range on ACK and for loss detection
     pub(super) sent_packets: BTreeMap<u64, SentPacket>,
+    /// Packets that were deemed lost
+    pub(super) lost_packets: BTreeMap<u64, SentPacket>,
```
Ralith: Do we need the full `SentPacket` structure here? It looks like the `time_sent` field is the only one we look at. `SentPacket` isn't always trivial to clone, so storing only that field could save some overhead.

quinn-proto/src/congestion.rs
```diff
@@ -56,9 +56,13 @@ pub trait Controller: Send + Sync {
         now: Instant,
         sent: Instant,
         is_persistent_congestion: bool,
+        is_ecn: bool,
         lost_bytes: u64,
     );
 
+    /// Packets were incorrectly deemed lost.
```
Ralith: Let's expand on exactly when this is called, and discuss some corner cases. In particular: what happens if two separate congestion events occur, and then the packet loss from the *first* event is found to be spurious?

Perhaps we don't want to commit to a specific behavior here, but we should be clear about what can be relied on.

Ralith: Style nit: omit trailing period on the first line.

Ralith (changes requested):
Thanks, this is great work! It's much simpler than I feared, and the improved throughput is really compelling. I wonder if this will also account for our slightly lagging throughput numbers in the interop test...

I'd like to see test coverage for the new behavior.

quinn-proto/src/connection/spaces.rs
```diff
@@ -39,6 +39,8 @@ pub(super) struct PacketSpace {
     /// Transmitted but not acked
     // We use a BTreeMap here so we can efficiently query by range on ACK and for loss detection
     pub(super) sent_packets: BTreeMap<u64, SentPacket>,
+    /// Packets that were deemed lost
```
Ralith, line 42: Document that this has limited retention due to `drain_lost_packets`.

quinn-proto/src/connection/mod.rs
```diff
@@ -1500,6 +1577,37 @@ impl Connection {
...
+
+            lost_packets.is_empty()
+        } else {
+            false
+        }
+    }
+
+    /// Drain lost packets that we think will reasonably never arrive
+    fn drain_lost_packets(&mut self, now: Instant, space: SpaceId) {
+        let two_pto = 2 * self.path.rtt.pto_base();
```
Ralith, line 1594: We should explain why this value was chosen, even if that explanation is just "blindly cribbed from msquic". Best of all would be to run some experiments and see where returns start to diminish in one or more real-world conditions, of course, but we can leave that to future work.

quinn-proto/src/connection/spaces.rs
stormshield-fabs: used a `LostPacket` structure that only stores `time_sent`

quinn-proto/src/connection/mod.rs
stormshield-fabs: I don't think a partial acknowledgment can be considered for spurious loss detection: we still have more packets that were not acknowledged and might be lost for real.

In the best case scenario, the next ACK will clear `lost_packets` and we'll correctly detect the spurious loss.

quinn-proto/src/congestion.rs
stormshield-fabs: If two congestion events occur back-to-back, then we can never find the first one alone to be spurious since `lost_packets` will also contain packets from the second event. If we find the both events to be spurious (before the packets expire), then we'll restore the state to the new values computed after the first event.

[RFC9438](https://www.rfc-editor.org/rfc/rfc9438.html#name-spurious-congestion-events) talks about smarter algorithms that we might want to try in the future.

quinn-proto/src/congestion/cubic.rs
```diff
@@ -221,6 +230,12 @@ impl Controller for Cubic {
         }
     }
 
+    fn on_spurious_congestion_event(&mut self) {
+        if let Some(state) = self.pre_congestion_state.take() {
+            self.state = state;
```
Ralith: The RFC recommends restoring only if the old cwnd is better than the current. Should we include that check? It seems reasonable to expect that it will be redundant in practice because we don't keep track of lost packets for very long, but we should at least document that assumption, and it's not obvious to me that there are no weird corner cases.

Ralith:
Note this is a breaking change due to the tweak to the congestion controller interface. We could adjust the design to be non-breaking, but I think it's fine; the current form is fairly natural and it's been a while.

quinn-proto/src/congestion/cubic.rs
stormshield-fabs: Done. Might be redundant but I agree on the corner cases.

djc:
> Note this is a breaking change due to the tweak to the congestion controller interface. We could adjust the design to be non-breaking, but I think it's fine; the current form is fairly natural and it's been a while.

Should we get another release out with non-breaking improvements before we merge this?

Ralith:
No objection from me.

stormshield-fabs:
We are working on implementing RFC 9438 which will likely require more changes to the `Controller` trait. 

Can we wait a bit more before merging this breaking change?

Ralith:
I think we should merge this as soon as we get a non-breaking release out the door, but we can hold the breaking release.

PR #2300, opened 2025-07-09
perf: display stats when drive tasks are cancelled by ctrl-c or timeout

stormshield-damiend:
This is still a work in progress and is placed here for discussion.

This allow uploading or downloading an infinite amount of bytes for a given duration and still have throughput statistics at the end of the duration.

Under the hood, this is a big rework of async tasks to spawn them. It uses two tokio CancellationTokens to first stop the ```drive_uni``` and  ```drive_bi``` tasks, then stop the statistics tasks after having collected the statistics.

It work fine with upload only and download only but if you do upload and download at the same time, the download statistics are broken.

I did not tested the ```drive_uni``` part for now.

Example of call :
```
./perf_client localhost:4433 --duration 120 --upload-size 99999999999 --download-size 0
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -885,6 +885,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "futures-sink"
+version = "0.3.31"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


stormshield-damiend:
For reference, download only mode is broken

PR #2305, opened 2025-07-16
Configure qlog in TransportConfig

Ralith:
Follow-up for https://github.com/quinn-rs/quinn/pull/2278.

This reworks qlog configuration to be consistent with other connection configuration, allows it to be specified when connections are first created, and allows users to control whether a qlog file is exclusive to a connection or arbitrarily shared.

Future work could include adding a `QlogStream` option to `EndpointConfig` for non-connection-related events, subject to the same optional sharing.

### DIFF
--- a/quinn-proto/src/config/transport.rs
+++ b/quinn-proto/src/config/transport.rs
@@ -5,7 +5,12 @@ use std::{io, sync::Mutex, time::Instant};
 #[cfg(feature = "qlog")]
 use qlog::streamer::QlogStreamer;
 
-use crate::{Duration, INITIAL_MTU, MAX_UDP_PAYLOAD, VarInt, VarIntBoundsExceeded, congestion};
+#[cfg(feature = "qlog")]
+use crate::QlogStream;
+use crate::{
+    Duration, INITIAL_MTU, MAX_UDP_PAYLOAD, VarInt, VarIntBoundsExceeded, congestion,
+    connection::qlog::QlogSink,
+};
 
 /// Parameters governing the core QUIC state machine
 ///
@@ -50,8 +55,7 @@ pub struct TransportConfig {
 
     pub(crate) enable_segmentation_offload: bool,
 
-    #[cfg(feature = "qlog")]
-    pub(crate) qlog_stream: Option<QlogStream>,
+    pub(crate) qlog_sink: QlogSink,
 }
 
 impl TransportConfig {
@@ -341,7 +345,7 @@ impl TransportConfig {
     /// qlog capture configuration to use for a particular connection
     #[cfg(feature = "qlog")]
     pub fn qlog_stream(&mut self, stream: Option<QlogStream>) -> &mut Self {
-        self.qlog_stream = stream;
+        self.qlog_sink = stream.into();
         self
     }
 }
@@ -386,8 +390,7 @@ impl Default for TransportConfig {
 
             enable_segmentation_offload: true,
 
-            #[cfg(feature = "qlog")]
-            qlog_stream: None,
+            qlog_sink: QlogSink::default(),
         }
     }
 }
@@ -420,8 +423,7 @@ impl fmt::Debug for TransportConfig {
                 deterministic_packet_numbers: _,
             congestion_controller_factory: _,
             enable_segmentation_offload,
-            #[cfg(feature = "qlog")]
-            qlog_stream,
+            qlog_sink,
         } = self;
         let mut s = fmt.debug_struct("TransportConfig");
 
@@ -451,9 +453,8 @@ impl fmt::Debug for TransportConfig {
             .field("datagram_send_buffer_size", datagram_send_buffer_size)
             // congestion_controller_factory not debug
             .field("enable_segmentation_offload", enable_segmentation_offload);
-        #[cfg(feature = "qlog")]
-        {
-            s.field("qlog_stream", &qlog_stream.is_some());
+        if cfg!(feature = "qlog") {
+            s.field("qlog_stream", &qlog_sink.is_enabled());
         }
 
         s.finish_non_exhaustive()
@@ -537,11 +538,6 @@ impl Default for AckFrequencyConfig {
         }
     }
 }
-
-/// Shareable handle to a single qlog output stream
-#[cfg(feature = "qlog")]
-#[derive(Clone)]
-pub struct QlogStream(pub(crate) Arc<Mutex<QlogStreamer>>);
 
 /// Configuration for qlog trace logging
 #[cfg(feature = "qlog")]
--- a/quinn-proto/src/config/transport.rs
+++ b/quinn-proto/src/config/transport.rs
@@ -1,8 +1,8 @@
 use std::{fmt, sync::Arc};
-#[cfg(feature = "__qlog")]
+#[cfg(feature = "qlog")]
 use std::{io, sync::Mutex, time::Instant};
 
-#[cfg(feature = "__qlog")]
+#[cfg(feature = "qlog")]
 use qlog::streamer::QlogStreamer;
 
 use crate::{Duration, INITIAL_MTU, MAX_UDP_PAYLOAD, VarInt, VarIntBoundsExceeded, congestion};
@@ -50,7 +50,7 @@ pub struct TransportConfig {
 
     pub(crate) enable_segmentation_offload: bool,
 
-    #[cfg(feature = "__qlog")]
+    #[cfg(feature = "qlog")]
     pub(crate) qlog_stream: Option<QlogStream>,
 }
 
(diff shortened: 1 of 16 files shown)


quinn-proto/src/config/transport.rs
```diff
@@ -329,6 +337,13 @@ impl TransportConfig {
         self.enable_segmentation_offload = enabled;
         self
     }
+
+    #[cfg(feature = "__qlog")]
```
djc: Nit: docstring before attribute?

quinn-proto/src/config/transport.rs
```diff
@@ -668,3 +694,103 @@ impl fmt::Debug for IdleTimeout {
         self.0.fmt(f)
     }
 }
+
+/// Shareable handle to a single qlog output stream
+#[cfg(feature = "__qlog")]
```
djc: Maybe this deserves to be above `IdleTimeout` and/or `MtuDiscoveryConfig`?

quinn-proto/src/config/transport.rs
```diff
@@ -668,3 +694,103 @@ impl fmt::Debug for IdleTimeout {
...
+            qlog::events::EventImportance::Core,
+            writer,
+        );
+
+        if let Err(e) = streamer.start_log() {
+            warn!("could not initialize endpoint qlog streamer: {e}");
+            return None;
+        }
+
+        Some(Self(Arc::new(Mutex::new(streamer))))
```
djc: Maybe make this a `match`?

```rust
match streamer.start_log() {
    Ok(streamer) => Some(Self(Arc::new(Mutex::new(streamer)))),
    Err(e) => {
        warn!("could not initialize endpoint qlog streamer: {e}");
        return None;
    }
}
```

quinn-proto/src/config/transport.rs
```diff
@@ -668,3 +694,103 @@ impl fmt::Debug for IdleTimeout {
...
+
+/// Shareable handle to a single qlog output stream
+#[cfg(feature = "__qlog")]
+#[derive(Clone)]
+pub struct QlogStream(pub(crate) Arc<Mutex<QlogStreamer>>);
+
+#[cfg(feature = "__qlog")]
+impl QlogStream {
+    /// Create the qlog output stream described by a [`QlogConfig`]
+    pub fn new(config: QlogConfig) -> Option<Self> {
```
djc: This interface feels a little pointless, can just inline this in `QlogConfig::build()`?

quinn-proto/src/config/transport.rs
```diff
@@ -668,3 +694,103 @@ impl fmt::Debug for IdleTimeout {
...
+            warn!("could not initialize endpoint qlog streamer: {e}");
+            return None;
+        }
+
+        Some(Self(Arc::new(Mutex::new(streamer))))
+    }
+}
+
+/// Configuration for qlog trace logging
+#[cfg(feature = "__qlog")]
```
djc: In other crates we've been using `struct` `non_exhaustive` to avoid builder stuff. Maybe we can do that here? Would be a little inconsistent, but also slightly more ergonomic?

djc (approved):
Nice work!

quinn-proto/src/connection/mod.rs
```diff
@@ -375,73 +365,24 @@ impl Connection {
...
-            warn!("could not initialize qlog streamer: {e}");
-            return;
-        }
-
-        self.qlog_streamer = Some(streamer);
-    }
-
     /// Emit a `MetricsUpdated` event to the qlog streamer containing only updated values
-    #[cfg(feature = "__qlog")]
+    #[cfg(feature = "qlog")]
```
flub: Would it not be nicer to establish a convention of doing the `cfg` *inside* this function, so that all the callsites do not have to worry about this?  I imagine there will be a lot of these functions eventually, and having all of them take 2 lines to call instead of a single function call could become a bit noisy?  I imagine the compiler would strip the unused function and there would be no runtime overhead.

quinn-proto/src/config/transport.rs
Ralith: We originally did that for config structs in quinn, but migrated to builders to decouple interface from representation and for better stability (e.g. we can refactor a setter and keep the old one around marked deprecated). While it's not immediately obvious that we'll need that flexibility in this particular case, consistency is nice.

quinn-proto/src/connection/mod.rs
Ralith: I think this is absolutely the right way to go internally.

I'm waffling on whether we should extend that pattern to the public API. Normally I'd say yes, but I worry that it would be a footgun -- users (particularly unfamiliar users in a support situation) might struggle to understand why configuring and enabling qlogs in the API is doing nothing, but if that API is feature-gated than rustc will explain the oversight at compile time.

quinn-proto/src/config/transport.rs
```diff
@@ -512,6 +538,99 @@ impl Default for AckFrequencyConfig {
...
+    }
+
+    /// Epoch qlog event times are recorded relative to
+    pub fn start_time(&mut self, start_time: Instant) -> &mut Self {
+        self.start_time = start_time;
+        self
+    }
+
+    /// Construct the [`QlogStream`] described by this configuration
+    pub fn build(self) -> Option<QlogStream> {
```
Ralith: I'm a bit unhappy with a method named `build` on a struct not named `...Builder`. I'm tempted to rename this struct to `QlogStreamBuilder`, but that would be less consistent with other config structs. Thoughts?

Ralith:
Addressed feedback, but also added significant new code.

Ralith:
This now covers all the events of "core" importance that recur within a connection. Next step is the remaining core-importance events, around connection establishment, but I don't feel a pressing need to cover that.

quinn-proto/src/config/transport.rs
djc: Ignore your unhappiness, seems fine? Or could rename to `into_stream()`.

quinn-proto/src/connection/qlog.rs
```diff
@@ -0,0 +1,79 @@
...
+impl QlogSink {
+    pub(crate) fn is_enabled(&self) -> bool {
+        #[cfg(feature = "qlog")]
+        {
+            self.stream.is_some()
+        }
+        #[cfg(not(feature = "qlog"))]
+        {
+            false
+        }
```
djc, line 61: I don't love the extra indentation level incurred by having the `cfg`s inside the functions like this. Maybe having duplicate noop functions is worth it?

quinn-proto/src/connection/paths.rs
```diff
@@ -171,7 +171,7 @@ impl PathData {
     pub(super) fn qlog_congestion_metrics(&mut self, pto_count: u32) -> Option<MetricsUpdated> {
         let controller_metrics = self.congestion.metrics();
 
-        let metrics = CongestionMetrics {
+        let metrics = RecoveryMetrics {
```
djc, line 174: Should the containing method be renamed, too?

quinn-proto/src/connection/qlog.rs
```diff
@@ -69,6 +73,44 @@ impl QlogSink {
...
+        now: Instant,
+        orig_rem_cid: ConnectionId,
+    ) {
+        #[cfg(feature = "qlog")]
+        {
+            let Some(stream) = self.stream.as_ref() else {
+                return;
+            };
+
+            let event = qlog::events::quic::PacketLost {
```
djc: Now that this is all in a `qlog` module, I'd suggest importing these, potentially aliasing like `QlogPacketLost`, rather than having full paths.

quinn-proto/src/connection/qlog.rs
```diff
@@ -145,6 +145,38 @@ impl QlogSink {
...
+
+            let event = qlog::events::quic::PacketReceived {
+                header: qlog::events::quic::PacketHeader {
+                    packet_number: Some(pn),
+                    packet_type: match space {
+                        SpaceId::Initial => qlog::events::quic::PacketType::Initial,
+                        SpaceId::Handshake => qlog::events::quic::PacketType::Handshake,
+                        SpaceId::Data if is_0rtt => qlog::events::quic::PacketType::ZeroRtt,
+                        SpaceId::Data => qlog::events::quic::PacketType::OneRtt,
+                    },
```
djc: Worth a helper function/trait impl?

quinn-proto/src/connection/qlog.rs
Ralith: While I don't love the extra indentation either, I really appreciate not having to worry about synchronizing two separate function prototypes.

djc: I guess...

djc (approved):

quinn-proto/src/config/transport.rs
djc: @Ralith please resolve this when you're ready.

Ralith: Renamed to `into_stream`.

PR #2306, opened 2025-07-16
Tiny clean up of a comment that was in the wrong place

flub:
This comment being in the wrong place makes reading this code a bit confusing.


djc (approved):
Thanks!

Issue #1413, opened 2022-09-22
Assertion failure when sending a lot of datagrams


jfro:
This is same assert as #1196, reproducing in debug config & a setup that sends 1000 datagrams 1ms apart.

https://github.com/jfro/quictest

just `cargo run` should demonstrate it. currently happens every time on my MBP M1, haven't checked it machine matters though.

Ralith:
This bug is not triggered by sending large numbers of datagrams, but rather by sending datagrams at or near the maximum size while there are ACKs pending. We include all unacknowledged ACK frames at the start of every outgoing packet, which consumes a portion of the MTU not accounted for by the datagram size limit. This leads to an unexpected failure to dequeue the application datagram, triggering an assert due to an unexpected composition of an ACK-only packet in debug builds, and leading to unreasonably high rates of datagram loss in release builds.

We should solve this by including ACK frames less frequently, ensuring packets with capacity for a maximum-size datagram can regularly be sent. However, this must be balanced against RFC9000's requirements for the timely transmission of ACKs. Tentatively, I think a reasonable strategy would be to include an ACK iff we have received an ACK-eliciting packet since we last transmitted, and perhaps also if our most recently sent ACK-bearing packet is deemed lost.

Ralith:
The logic outlined above was already implemented in https://github.com/quinn-rs/quinn/commit/a2fe103f4b5f08e8c7a623ab7da1c4b6d17e4749, which is in HEAD but not the 0.8 branch. This prevents the unreasonable loss rates observed in 0.8. The assert can still be triggered whenever both a fresh ACK and a full-size datagram is pending, because a single packet cannot contain both an ACK frame and a full-sized datagram, and the current behavior gives precedence to the ACK frame.

If an ACK-only packet is not currently permitted, we must instead give precedence to the datagrams, else we are in violation of RFC 9000 §13.2.1. Conversely, when an ACK-only packet is permitted, to guarantee that a timely ACK is transmitted at all we must send one even if we have full-sized datagrams queued that could be sent instead.

A simple solution would be to take the above commit a step further, and only send ACKs on the receipt of ACK-eliciting packets (i.e. when an ACK-only packet is permissible), rather than after receipt of any packet. This should also simplify our logic slightly.

Matthias247:
There seem to be multiple strategies to avoid this:
1. Make sure large sized datagram packets don't include ACKs, and send ACKs separately if necessary
[...]

PR #2307, opened 2025-07-18
Send ACK frames with any outgoing packet

flub:
Without this change ACK frames would only be sent if something triggered an immediate ack or when the max ack delay timer fires. Essentially these are times when an ACK frame SHOULD be sent.

However there are many times when it is possible to acknowledge new packet numbers, but is is not yet required to do so.  These should not trigger the sending of a packet, but when a packet is anyway being sent including them will speed up the knowledge of the acknowledgements by the peer.  In turn this has a positive effect on the congestion and flow control windows.

Benchmarks from the Quinn perf tool show a small, but mostly consistently noticible improvement.

Before:

Overall stats: RPS: 369.09 (22149 requests in 60.01s)
Stream metrics:
|      | Upload Duration | Download Duration | FBL | Upload Throughput | Download Throughput |
| ---- | --------------- | ----------------- | --- | ----------------- | ------------------- |
| AVG  |          1.45ms |        1.25ms | 30.00µs |      6904.11 Mb/s |        7490.94 Mb/s |
| P0   |        523.00µs |       94.00µs |  0.00ns |       257.43 Mb/s |         264.77 Mb/s |
| P10  |          1.05ms |      972.00µs |  5.00µs |      4869.59 Mb/s |        5284.82 Mb/s |
| P50  |          1.20ms |        1.16ms | 28.00µs |      6975.13 Mb/s |        7226.79 Mb/s |
| P90  |          1.72ms |        1.59ms | 50.00µs |      8023.70 Mb/s |        8631.88 Mb/s |
| P100 |         32.59ms |       31.68ms |  1.05ms |     16047.41 Mb/s |       89254.79 Mb/s |

After:

Overall stats: RPS: 457.24 (27439 requests in 60.01s)
Stream metrics:

|      | Upload Duration | Download Duration | FBL | Upload Throughput | Download Throughput |
| ---- | --------------- | ----------------- | --- | ----------------- | ------------------- |
| AVG  |          1.11ms |        1.07ms | 29.00µs |      7905.00 Mb/s |        8176.26 Mb/s |
| P0   |        516.00µs |      393.00µs |  0.00ns |       282.07 Mb/s |         295.17 Mb/s |
| P10  |        751.00µs |      735.00µs | 12.00µs |      6866.08 Mb/s |        7117.73 Mb/s |
| P50  |          1.11ms |        1.07ms | 27.00µs |      7566.52 Mb/s |        7826.57 Mb/s |
| P90  |          1.22ms |        1.18ms | 49.00µs |     11173.63 Mb/s |       11416.90 Mb/s |
| P100 |         29.74ms |       28.42ms |  1.07ms |     16257.12 Mb/s |       21357.40 Mb/s |

 
I'm aware I might be missing a lot of history, so there might have been a reason that Quinn was not doing this.

One issue I can think of is the interaction with large DATAGRAM frames.  ACK frames are populated first, and they don't currently have a size estimate even.  If the application is sending datagrams at the limit, adding these ACK frames make make the datagrams no longer fit and new packets have to be built.  It would even break a GSO batch due to the size difference between the packets.  Though the fact no other test fails because of this change is a good sign I think.


Ralith:
Overall, this is a tricky one. For history, see 346f2e93a7e7b1ecc514fa852308959d58f9fe0c and https://github.com/quinn-rs/quinn/issues/1413. I also understand from IETF discussions that ACK processing is traditionally understood to be a cost center, so batching them may substantially improve sender efficiency. While that hasn't been yet shown to be a major concern in Quinn, it was reported empirically by e.g. [fastly](https://www.fastly.com/blog/measuring-quic-vs-tcp-computational-efficiency). And, of course, it adds some direct packet space overhead.

Note also that Quinn implements the ACK frequency extension, which allows senders to influence the rate at which ACKs are generated. Can that be tuned to achieve similar results without requiring changes to the implementation? Note that such tuning may be condition and workload dependent.

What network conditions (in particular round trip) did you test with? If this data was on loopback, it may not be representative.

flub:
Oh yes, good point about loopback testing.  Re-doing the tests over a "normal" internet connection gives me:

Before (AVG):
RPS: 7.78  (467 requests in 60s)
|     | Upload Duration | Download Duration | FBL | Upload Throughput | Download Throughput |
| --- | --------------- | ----------------- | --- | ----------------- | ------------------- |
| AVG | 59.42ms         | 68.73ms      | 618.00µs |       146.10 Mb/s |         151.50 Mb/s |

cpu time: 4.48 usr, 4.87 sys

After:
RPS: 8.00 (480 requests in 60s)
|     | Upload Duration | Download Duration | FBL | Upload Throughput | Download Throughput |
| --- | --------------- | ----------------- | --- | ----------------- | ------------------- |
| AVG | 60.61ms         | 64.33ms        | 1.23ms |       146.14 Mb/s |         144.97 Mb/s |

cpu time: 5.58 user, 4.92 sys

flub:
(apologies, posted the above too soon)

This is still a very rough measurement and not a proper statistical analysis (which I expect to be needed for ACK changes at this point to be fair): perf wise doesn't seem to much.  But CPU does indeed increase.

Thanks also for finding the earlier commits, this also confirms my fear of one the biggest dangers.  Without also addressing that I don't think this would be a good idea.

For now I'm going to assume this isn't really worth chasing further.  At least not until there's more motivation.  Thanks also for the pointer so requesting immediate acks.  I kind of overlooked that possibility.  I'll see if that can address why I first ran into this (in itself that wasn't a serious issue, but enough to let me explore this a little).

Ralith:
Thanks for studying this!

PR #2316, opened 2025-07-22
Inline StepRng

djc:
This has been [deprecated](https://github.com/rust-random/rand/pull/1634) upstream. Inlining a simplified copy of the upstream code seems like the easiest solution to the problem we're solving here (testing the reserved transport parameter ID generation code for edge cases).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1684,9 +1684,9 @@ checksum = "74765f6d916ee2faa39bc8e68e4f3ed8949b48cccdac59983d287a7cb71ce9c5"
 
 [[package]]
 name = "rand"
-version = "0.9.1"
+version = "0.9.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


Ralith:
The deprecation PR makes a credible argument that code, such as ours, which isn't implementing a primitive (i.e. `next_u64`-based) distribution shouldn't be using this at all. IIUC the intent of this test is just to run the generator a bunch and sanity-check its outputs. Should we switch to a single instance of a stable real RNG instead?

djc:
> Should we switch to a single instance of a stable real RNG instead?

The test code here seems carefully constructed to test a bunch of edge cases that might occur. We wouldn't get that level of coverage from switching to a real RNG, right?

Ralith:
My concern is that the proposition that those edge cases are, in fact, being exercised relies on assumptions about the implementation of `random_range` that may not be true today, much less forever.

Issue #2286, opened 2025-06-30
Low bandwidth compatible pacer


vitocln:
We are experimenting with using Quinn over slow internet connections, such as 400 Kbps. We've observed unexpected and continuous packet losses in this context. After some investigation, we've identified the pacer as the primary cause.

The current pacer is designed for typical internet connections and enforces a hard lower bound of 10 packets per 2 ms. This rate is excessive for low-bandwidth networks.

To address this, we've implemented a "dirty fix" that allows for reduced pacing capacity and slower token refill rates. With this patch, the unexpected packet losses were eliminated:

```
diff --git a/quinn-proto/src/connection/pacing.rs b/quinn-proto/src/connection/pacing.rs
index 2e469948..da2dbcf0 100644
--- a/quinn-proto/src/connection/pacing.rs
+++ b/quinn-proto/src/connection/pacing.rs
@@ -28,7 +28,7 @@ impl Pacer {
             capacity,
             last_window: window,
             last_mtu: mtu,
-            tokens: capacity,
+            tokens: capacity.max(mtu as u64), // build up the first token directly
             prev: now,
         }
     }
@@ -89,11 +89,11 @@ impl Pacer {
         }
 
         let elapsed_rtts = time_elapsed.as_secs_f64() / smoothed_rtt.as_secs_f64();
-        let new_tokens = window as f64 * 1.25 * elapsed_rtts;
+        let new_tokens = (window as f64 * 1.25 * elapsed_rtts).min(self.capacity as f64);
+
         self.tokens = self
             .tokens
-            .saturating_add(new_tokens as _)
-            .min(self.capacity);
+            .saturating_add(new_tokens as _);
 
         self.prev = now;
 
@@ -103,7 +103,7 @@ impl Pacer {
         }
 
         let unscaled_delay = smoothed_rtt
-            .checked_mul((bytes_to_send.max(self.capacity) - self.tokens) as _)
+            .checked_mul((bytes_to_send.max(self.capacity) - self.tokens).min(self.capacity) as _)
             .unwrap_or(Duration::MAX)
             / window;
 
@@ -131,9 +131,7 @@ fn optimal_capacity(smoothed_rtt: Duration, window: u64, mtu: u16) -> u64 {
 
     let capacity = ((window as u128 * BURST_INTERVAL_NANOS) / rtt) as u64;
 
-    // Small bursts are less efficient (no GSO), could increase latency and don't effectively
[...]

PR #2323, opened 2025-08-07
proto: pacer slow mode

vitocln:
Linked to this [issue](https://github.com/quinn-rs/quinn/issues/2286).

The pacer enters dynamically in slow mode when the inter-packet pacing exceed 10 ms. In slow mode, the pacer sends only single packet per interval and disables the default burst sending mode.

### DIFF
--- a/quinn-proto/src/connection/pacing.rs
+++ b/quinn-proto/src/connection/pacing.rs
@@ -210,6 +210,7 @@ const MAX_INTERVAL_MS: u64 = 10;
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::ops::{Mul, Sub};
 
     #[test]
     fn does_not_panic_on_bad_instant() {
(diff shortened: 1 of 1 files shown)


quinn-proto/src/connection/pacing.rs
```diff
@@ -18,18 +18,20 @@ pub(super) struct Pacer {
     last_mtu: u16,
     tokens: u64,
     prev: Instant,
+    burst_mode: bool,
```
Ralith, line 21: Does this need to be stateful? Would it be sufficient to instead just clamp the maximum delay returned?

Ralith (changes requested):
Please include unit tests for the new behavior, and resolve the failing integration tests.

djc:
Could we make this more dynamic instead of using a boolean burst mode?

quinn-proto/src/connection/pacing.rs
vitocln: That was the first approach I tried, but it didn’t resolve the issue. If the pacer gets stalled for an extended period—e.g., due to the congestion window—it accumulates tokens up to the cap (minimum 10 packets), which then results in a burst when it resumes sending.

Do you have a better idea?

Ralith: Hmm, I see how that would be a problem. I'll need to find time to think on this more. In the mean time, please add unit tests that validate the new behavior, both in steady state and after an idle period as you outlined above.

quinn-proto/src/connection/pacing.rs
```diff
@@ -90,10 +98,16 @@ impl Pacer {
...
         let elapsed_rtts = time_elapsed.as_secs_f64() / smoothed_rtt.as_secs_f64();
         let new_tokens = window as f64 * 1.25 * elapsed_rtts;
-        self.tokens = self
-            .tokens
-            .saturating_add(new_tokens as _)
-            .min(self.capacity);
+
+        self.tokens = self.tokens.saturating_add(new_tokens as _);
+
+        self.tokens = self.tokens.min(if self.burst_mode {
```
djc, line 104: Nit: suggest folding the `saturating_add()` call into the second expression, as in `self.tokens = self.tokens.saturating_add(new_tokens as _).min(..);`.

quinn-proto/src/connection/pacing.rs
```diff
@@ -18,18 +18,21 @@ pub(super) struct Pacer {
...
     tokens: u64,
     prev: Instant,
+    burst_mode: bool,
 }
 
 impl Pacer {
     /// Obtains a new [`Pacer`].
     pub(super) fn new(smoothed_rtt: Duration, window: u64, mtu: u16, now: Instant) -> Self {
-        let capacity = optimal_capacity(smoothed_rtt, window, mtu);
+        // burst mode is the default for quick handshake
```
djc, line 27: Nit: this comment isn't very clear to me. Do you mean QUIC handshake? Or, to improve handshake latency?

Ralith:
Alright, I've had a good think about this. It seems to me that reducing the pacer's capacity alone should be sufficient to improve behavior here. @vitocln, in the original issue you reduced the minimum burst size to 1, but also made several other changes. Was reducing the minimum burst size sufficient? If so, I think that's the minimum change we could make here. There should be no drawbacks for users on faster connections, which will continue to use larger bursts.

@kixelated points out that BBR [recommends](https://www.ietf.org/archive/id/draft-ietf-ccwg-bbr-03.html#section-5.6.3) a minimum burst size of 2 packets, so our choice of 10 was probably over-ambitious to begin with. I'd be happy to merge a change which just drops the minimum burst size to 2. Alternatively, I've drafted the smarter latency-limiting version at https://github.com/quinn-rs/quinn/pull/2372.

vitocln:
> Alright, I've had a good think about this. It seems to me that reducing the pacer's capacity alone should be sufficient to improve behavior here. @vitocln, in the original issue you reduced the minimum burst size to 1, but also made several other changes. Was reducing the minimum burst size sufficient? If so, I think that's the minimum change we could make here. There should be no drawbacks for users on faster connections, which will continue to use larger bursts.
> 
> @kixelated points out that BBR [recommends](https://www.ietf.org/archive/id/draft-ietf-ccwg-bbr-03.html#section-5.6.3) a minimum burst size of 2 packets, so our choice of 10 was probably over-ambitious to begin with. I'd be happy to merge a change which just drops the minimum burst size to 2. Alternatively, I've drafted the smarter latency-limiting version at #2372.

Reducing the burst size to 1 was my initial attempt, but I do not recall why I decided to proceed further. In any case, setting the burst size to 1 appears to resolve the issue. I have reviewed your pull request and it looks good. Thank you for your effort; it is appreciated.

Ralith:
Great! Thanks for helping us work through this.

PR #2326, opened 2025-08-08
Check for in-flight data on-demand

Ralith:
Simpler and more robust than maintaining a separate running total.

As proposed in Discord by @flub. No functional change,


djc (approved):

flub:
Oh, nice.  Thanks for adding this simplification!

PR #2339, opened 2025-08-12
Add interface_index to RecvMeta

sinhpham:
I wanna use `quinn-udp`, but I also need the interface index information.

Currently `quinn-udp` reads the `IP_PKTINFO` header and extracts only `destination_ip`, discarding `interface_index`.

This PR adds `interface_index` to `RecvMeta`

### DIFF
--- a/quinn-udp/src/fallback.rs
+++ b/quinn-udp/src/fallback.rs
@@ -72,6 +72,7 @@ impl UdpSocketState {
             addr: addr.as_socket().unwrap(),
             ecn: None,
             dst_ip: None,
+            interface_index: None,
         };
         Ok(1)
     }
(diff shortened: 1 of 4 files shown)


Ralith:
Can you elaborate on the use case? I'm hesitant to support stuff that's not QUIC-relevant, though I admit this seems particularly low-cost.

sinhpham:
In my case, I need to differentiate between LAN and Wi-Fi interfaces. Unfortunately, I can't provide more specific details, as the code is proprietary.

djc:
Please squash your commits.

Your organization should [sponsor](https://opencollective.com/quinn-rs) Quinn.

sinhpham:
I squashed the commits 👍

PR #2340, opened 2025-08-13
Improve quinn::ConnectionStats docs

gry8t:


quinn-proto/src/connection/stats.rs
```diff
@@ -3,6 +3,9 @@
 use crate::{Dir, Duration, frame::Frame};
 
 /// Statistics about UDP datagrams transmitted or received on a connection
+///
+/// All information is regarding global UDP activity and is not specific to any higher level QUIC
+/// operations
```
Ralith: I think this is a net loss of clarity. The existing text already specifies UDP datagrams, and this information is not global; it's specific to the connection.

Would something like this have helped you?
```suggestion
///
/// All QUIC packets are carried by UDP datagrams. Hence, these statistics cover all traffic on a connection.
```

gry8t: Yes, I found confusing the original description, was unsure if they were statistics about all traffic or just the datagram channel. "These statistics cover all traffic on a connection" solves the doubt.

djc: Okay, please incorporate the suggestion in your commit.

djc (approved):

djc:
(Needs squashing during merge.)

Ralith (approved):
Thanks!

PR #2342, opened 2025-08-14
Various  enhancement for perf binaries

stormshield-damiend:
Factorize common options on clap side.

Factorize init code for TransportConfig.

Add option to disable ansi color codes in logs.

Use a bufered writer for qlogs, here are bench results on a internet link :
- qlog off -> 295 Mb/s
- qlog on (client) no buffered writer -> 240 Mbps
- qlog on (client) with buffered writer -> 284 Mbps

### DIFF
--- a/perf/src/bin/perf_client.rs
+++ b/perf/src/bin/perf_client.rs
@@ -15,7 +15,7 @@ use tokio::sync::Semaphore;
 use tracing::{debug, error, info};
 
 use perf::{
-    CommonOpt, bind_socket, build_transport_config,
+    CommonOpt, bind_socket, build_transport_config, init_tracing,
     noprotection::NoProtectionClientConfig,
     stats::{OpenStreamStats, Stats},
 };
@@ -64,7 +64,7 @@ struct Opt {
 async fn main() {
     let opt = Opt::parse();
 
-    tracing_subscriber::fmt::init();
+    init_tracing(&opt.common);
 
     if let Err(e) = run(opt).await {
         error!("{:#}", e);
--- a/perf/src/bin/perf_client.rs
+++ b/perf/src/bin/perf_client.rs
@@ -1,5 +1,5 @@
 #[cfg(feature = "json-output")]
-use std::{fs::File, path::PathBuf};
+use std::path::PathBuf;
 use std::{
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
     sync::Arc,
@@ -15,7 +15,7 @@ use tokio::sync::Semaphore;
 use tracing::{debug, error, info};
 
 use perf::{
-    CongestionAlgorithm, bind_socket,
+    CommonOpt, bind_socket, build_transport_config,
     noprotection::NoProtectionClientConfig,
     stats::{OpenStreamStats, Stats},
 };
@@ -30,6 +30,9 @@ struct Opt {
     /// Override DNS resolution for host
     #[clap(long)]
     ip: Option<IpAddr>,
+    /// Specify the local socket address
+    #[clap(long)]
+    local_addr: Option<SocketAddr>,
     /// Number of unidirectional requests to maintain concurrently
     #[clap(long, default_value = "0")]
     uni_requests: u64,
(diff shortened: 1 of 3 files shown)


perf/src/lib.rs
```diff
@@ -32,6 +71,36 @@ impl CongestionAlgorithm {
     }
 }
 
+pub fn build_transport_config(
```
djc: Suggest attaching this to `CommonOpt` as a method.

perf/src/lib.rs
```diff
@@ -32,6 +71,36 @@ impl CongestionAlgorithm {
...
+    }
+
+    if let Some(cong_alg) = common.cong_alg {
+        transport.congestion_controller_factory(cong_alg.build());
+    }
+
+    #[cfg(feature = "qlog")]
+    if let Some(qlog_file) = &common.qlog_file {
+        let mut qlog = quinn::QlogConfig::default();
+        qlog.writer(Box::new(std::fs::File::create(qlog_file)?))
```
djc: Nit: please import `File` as before, and also all the other symbols here (`TransportConfig`, `AckFrequencyConfig`, `QlogConfig`, etc).

perf/src/bin/perf_client.rs
```diff
@@ -128,7 +100,11 @@ async fn run(opt: Opt) -> Result<()> {
 
     info!("local addr {:?}", bind_addr);
 
-    let socket = bind_socket(bind_addr, opt.send_buffer_size, opt.recv_buffer_size)?;
+    let socket = bind_socket(
```
djc: Maybe this becomes `CommonOpts::bind_socket()`?

perf/src/lib.rs
```diff
@@ -93,8 +93,9 @@ pub fn build_transport_config(
     #[cfg(feature = "qlog")]
     if let Some(qlog_file) = &common.qlog_file {
         let mut qlog = quinn::QlogConfig::default();
-        qlog.writer(Box::new(std::fs::File::create(qlog_file)?))
-            .title(Some(name.into()));
+        let file = std::fs::File::create(qlog_file)?;
```
djc: Nit: import `BufWriter` as well.

perf/src/lib.rs
```diff
@@ -52,6 +53,9 @@ pub struct CommonOpt {
     #[cfg(feature = "qlog")]
     #[clap(long = "qlog")]
     pub qlog_file: Option<PathBuf>,
+    /// disable ansi colors in logs
+    #[clap(long)]
+    pub no_color: bool,
```
djc: I wouldn't be surprised if there are environment variables that tracing-subscriber already knows about? Would be nicer than reinventing stuff here.

perf/src/lib.rs
```diff
@@ -102,6 +106,18 @@ pub fn build_transport_config(
     Ok(transport)
 }
 
+pub fn init_tracing(common: &CommonOpt) {
+    let fmt_layer = tracing_subscriber::fmt::layer().with_ansi(!common.no_color);
+    let filter_layer = tracing_subscriber::EnvFilter::try_from_default_env()
```
djc: Nit: import `EnvFilter`.

djc (commented):
Makes sense, nice!

perf/src/lib.rs
stormshield-damiend: Indeed it support NO_COLOR env var if call to with_ansi is not made https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Layer.html#method.with_ansi, i will remove this modification.

stormshield-damiend: i will keep the init method as it factorize code between client and server

perf/src/bin/perf_client.rs
stormshield-damiend: Can also be a ```let socket = opt.bind_socket(addr)``` ?

perf/src/lib.rs
stormshield-damiend: done with fmt and some needed traits

perf/src/lib.rs
stormshield-damiend: done

perf/src/lib.rs
stormshield-damiend: done

perf/src/bin/perf_client.rs
djc: I think that's the same thing I was suggesting?

stormshield-damiend: Yes, i will do it

perf/src/lib.rs
```diff
@@ -15,6 +20,117 @@ pub mod stats;
...
+            let mut qlog = QlogConfig::default();
+            qlog.writer(Box::new(File::create(qlog_file)?))
+                .title(Some(name.into()));
+            transport.qlog_stream(qlog.into_stream());
+        }
+
+        Ok(transport)
+    }
+
+    pub fn bind_socket(&self, addr: SocketAddr) -> Result<std::net::UdpSocket> {
```
djc, line 92: This seems to be duplicating rather than replacing the free `bind_socket()` function?

Also, would prefer to keep this in a separate commit from the introduction of `CommonOpts` and `build_transport_config()`.

perf/src/lib.rs
```diff
@@ -149,6 +150,18 @@ impl CongestionAlgorithm {
...
 }
 
+pub fn init_tracing() {
+    let fmt_layer = fmt::layer();
+    let filter_layer = EnvFilter::try_from_default_env()
+        .or_else(|_| EnvFilter::try_new("warn"))
+        .unwrap();
+
+    tracing_subscriber::registry()
+        .with(filter_layer)
```
djc: Nit: suggest inlining the `fmt_layer` and `filter_layer` bindings.

perf/src/lib.rs
stormshield-damiend: oops during my rebase i kept the original code, i will fix it, i can also do it in a separate commit, want one separate commit for ```build_transport_config``` ?

djc: I think it's fine to keep the introduction of `build_transport_config()` in the same commit as the extraction of `CommonOpts` would prefer the move of `bind_socket()` to be in another commit.

perf/src/lib.rs
stormshield-damiend: done

perf/src/lib.rs
stormshield-damiend: ok i will do that

djc:
Thanks!

PR #2343, opened 2025-08-18
perf: add max udp payload size option

Dreamacro:
Sometimes the actual MTU of the interface is not 1500, so I would like perf to customize this.

I noticed that #2342 also makes changes to perf, so I can wait until it's merged and adjust the code.

### DIFF
--- a/perf/src/bin/perf_client.rs
+++ b/perf/src/bin/perf_client.rs
@@ -109,7 +109,10 @@ async fn run(opt: Opt) -> Result<()> {
 
     let socket = opt.common.bind_socket(bind_addr)?;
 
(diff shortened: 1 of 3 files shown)


Dreamacro:
@djc already rebase main

djc:
(I'd like @Ralith to take a look before merging.)

Ralith (approved):
There's a case for `Option<u16>` so that we only have to write the default value in one spot, but I like how, as written, the default shows up gracefully in the auto-generated CLI flag documentation.

Ralith:
Using the interface MTU by default would also be a nice idea, but the question of *which* interface MTU becomes very hairy.

PR #2352, opened 2025-08-22
Perf enhance2

stormshield-damiend:
Borrow code from bench binary to support human SI size for various params.

Allow setting stream or connection received / send windows from command line.


perf/src/bin/perf_client.rs
```diff
@@ -40,10 +41,16 @@ struct Opt {
     #[clap(long, default_value = "1")]
     bi_requests: u64,
     /// Number of bytes to request
-    #[clap(long, default_value = "1048576")]
+    ///
+    /// This can use SI prefixes for sizes. E.g. 1M will transfer 1MiB, 10G
```
djc: "prefixes" -> "suffixes" (also, below). Try to avoid "e.g.", maybe use "For example" instead?

perf/src/lib.rs
```diff
@@ -133,6 +139,28 @@ impl CommonOpt {
...
+        _ => 1,
+    };
+
+    let s = if multiplier != 1 {
+        &s[..s.len() - 1]
+    } else {
+        s
+    };
+
+    let base: u64 = u64::from_str(s)?;
```
djc: Nit: no need to annotate this type, probably okay to inline this into `Ok(u64::from_str(s)? * multiplier)`?

perf/src/lib.rs
```diff
@@ -133,6 +139,28 @@ impl CommonOpt {
...
+        Some('M') => 1024 * 1024,
+        Some('k') => 1024,
+        _ => 1,
+    };
+
+    let s = if multiplier != 1 {
+        &s[..s.len() - 1]
+    } else {
+        s
+    };
```
djc: Nit: this is a little bit nicer IMO as

```rust
let s = match multiplier {
     1 => s,
     _ => &s[..s.len() - 1],
};
```

djc (approved):
LGTM modulo some nits. Thanks!

perf/src/bin/perf_client.rs
stormshield-damiend: This was copy-paste from https://github.com/quinn-rs/quinn/blob/main/bench/src/lib.rs#L180, i will fix both

perf/src/lib.rs
stormshield-damiend: Same, copy paste from https://github.com/quinn-rs/quinn/blob/main/bench/src/lib.rs#L206, i will fix it

perf/src/lib.rs
stormshield-damiend: Copy paste from https://github.com/quinn-rs/quinn/blob/main/bench/src/lib.rs#L206, i will fix it

PR #2354, opened 2025-08-25
perf: move client and server on one binary

Dreamacro:
I've been using `perf_client` and `perf_server` lately, but the split-binary design feels cumbersome.

It'd be much more convenient—like `iperf3` or msquic's perf—to have a single binary that supports both client and server modes via command-line flags.

Converted the previous top-level options into Perf subcommands. All other behavior remains consistent with the prior implementation.

old version:
```bash
$ ./perf_server --no-protection
$ ./perf_client ip:4433 --no-protection
```

new version:
```bash
$ ./perf server --no-protection
$ ./perf client ip:4433 --no-protection
```

### DIFF
--- a/perf/src/bin/perf.rs
+++ b/perf/src/bin/perf.rs
@@ -1,0 +1,42 @@
+use clap::{Parser, Subcommand};
+use tracing::error;
+use tracing_subscriber::{EnvFilter, fmt, layer::SubscriberExt, util::SubscriberInitExt};
+
+use perf::{client, server};
+
+#[derive(Parser)]
+#[clap(long_about = None)]
+struct Cli {
+    #[command(subcommand)]
+    command: Commands,
+}
+
+#[derive(Subcommand)]
+enum Commands {
+    /// Run as a perf server
+    Server(server::Opt),
+    /// Run as a perf client
+    Client(client::Opt),
+}
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() {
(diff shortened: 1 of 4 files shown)


perf/src/bin/perf.rs
```diff
@@ -0,0 +1,34 @@
+use clap::{Parser, Subcommand};
+use tracing::error;
+
+use perf::{client, init_tracing, server};
+
+#[derive(Subcommand)]
+enum Commands {
```
djc, line 15: Nit: item ordering, `main()`, then `Cli`, then `Commands`.

perf/src/bin/perf.rs
```diff
@@ -0,0 +1,34 @@
...
+struct Cli {
+    #[command(subcommand)]
+    command: Commands,
+}
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() {
+    let opt = Cli::parse();
+
+    init_tracing();
```
djc: I guess we can now inline `init_tracing()` again?

Ralith (approved):

perf/src/bin/perf.rs
Dreamacro: done

Dreamacro: Previously, `main` was placed under the clap structure, and I adjusted the order of `Cli` and `Commands`.

Issue #2351, opened 2025-08-22
Feature Request: `datagram_readable` &`datagram_writeable` & `try_send/recv_datagram`


Aderinom:
Hello Team! 

At the moment quinn, on datagrams, offers:

- `send_datagram`      - which overwrites datagrams in the send buffer
- `send_datagram_wait` - which waits until the datagram can be enqueued
- `read_datagram`      - which waits until a datagram has been received

These functions all require: 
1. The buffer which should possibly be sent/received on at that moment to already exist -> Possibly wasted work
2. The caller to wait for them/ or accepting data to be lost -> Caller has no control

With this, the User lacks some low level control over the Connection.

I would like to propose adding:

- `poll_send_datagram` - checking if send is blocked, and if so registering the poller with the datagrams_unblocked notifier. 
- `poll_read_datagram` - checking if receive is blocked, and if so registering the poller with the datagrams_received notifier.

For higher level integration and for the caller to attempt operation without side effect on fail:

- `try_send_datagram` \ `try_receive_datagram` - returning a `WouldBlock` if operation can not be handled without waiting

This would allow a user to check the state of the socket, and decide what to do based on this.
Especially latency sensitive applications or ones which have do internal congestion control could profit from this.

Ralith:
If you need to register for wakeups, create and poll a future using the existing APIs you've already identified. You can't register to be woken without a future, because that's how multiple tasks waiting for the same event are tracked.

If you don't need to register for wakeups, you can poll the same futures with a no-op waker, but this is usually a design error.

Aderinom:
Ah, understood, I missed that we need to keep a pinned Notified, so a pure poll won't work.

For what we are trying to do, we would require:

- a way to wait until the transport is readable/writable without side effects
- a way to try to read/write without side effects if it currently would block

Is there a way to do this with the current api?

As a reference - the API we need to implement for our project roughly looks like this:

```rs
trait DatagramTransport {
    /// Completes when the transport may have data available.
    async fn readable(&self) -> io::Result<()>;

    /// Completes when the transport may be able to accept a datagram.
    async fn writable(&self) -> io::Result<()>;

    /// Attempts to receive a datagram into the provided buffer.
    /// 
    /// Returns the number of bytes read on success.
    /// Returns WouldBlock if no datagram is currently available.
    fn try_recv(&self, buf: &mut [u8]) -> Result<usize, io::Error>;

    /// Attempts to send a datagram from the provided buffer.
    /// 
[...]

PR #2366, opened 2025-08-28
feat(datagrams): add sendable/readable futures and try_send/try_recv methods

Aderinom:
Related Issues:
- #2351 
- #1423 (introduces notification when send buffer space reaches a threshold via `datagram_sendable`)

Additions:

- `Connection::try_send_datagram` and `Connection::try_read_datagram`, returning `WouldBlock` when the operation cannot complete immediately
- `Connection::datagram_readable`, resolving as soon as any datagram is readable
- `Connection::datagram_sendable`, resolving as soon as a specified number of bytes are writable

### DIFF
--- a/quinn-proto/src/connection/datagrams.rs
+++ b/quinn-proto/src/connection/datagrams.rs
@@ -57,7 +57,7 @@ impl Datagrams<'_> {
         Ok(())
     }
 
-    /// By calling this function, next time a datagram is sent a DatagramsUnblocked event will be created
+    /// Next time a Datagram is sent, datagram_unblocked will be notified
     pub fn request_datagram_unblocked_notification(&mut self) {
         self.conn.datagrams.send_blocked = true;
     }
--- a/quinn-proto/src/connection/datagrams.rs
+++ b/quinn-proto/src/connection/datagrams.rs
@@ -57,6 +57,11 @@ impl Datagrams<'_> {
         Ok(())
     }
 
+    /// By calling this function, next time a datagram is sent a DatagramsUnblocked event will be created
+    pub fn request_datagram_unblocked_notification(&mut self) {
+        self.conn.datagrams.send_blocked = true;
+    }
+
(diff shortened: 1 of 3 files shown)


quinn/src/connection.rs
```diff
@@ -353,6 +361,23 @@ impl Connection {
         }
     }
 
+    /// Attempts to receive an application datagram
+    ///
+    /// If there are no readable datagrams, this will return [TryReceiveDatagramError::WouldBlock]
+    pub fn try_read_datagram(&self) -> Result<Bytes, TryReceiveDatagramError> {
```
Ralith: `Result<Option<Bytes>, ConnectionError>` seems much less error-prone, since it separates the exceptional case.

quinn-proto/src/connection/datagrams.rs
```diff
@@ -57,6 +57,13 @@ impl Datagrams<'_> {
         Ok(())
     }
 
+    /// Sets the send_blocked flag to true
```
Ralith: The send_blocked flag is not part of the public API, so this is not helpful documentation. Maybe try an imperative phrasing of what you wrote just below?

quinn/src/connection.rs
```diff
@@ -345,6 +345,14 @@ impl Connection {
         }
     }
 
+    /// Creates a future, resolving as soon as a readable datagram is buffered
```
Ralith: Don't prefix async function documentation with "Creates a future"; it doesn't add information.

Ralith (changes requested):
Per discussion in #2351, it's not clear what the `try_` methods add. I don't want to maintain two copies of every single async function when it's straightforward enough to poll the futures we already have.

If you're not comfortable relying on that without a documented guarantee, feel free to propose some documentation, as @djc suggested. It would be very strange for any future that *could* complete immediately to return `Pending` instead, but I don't mind explicitly committing not to.

PR #2369, opened 2025-08-29
Update tracing-subscriber to 0.3.20

djc:
To address https://github.com/quinn-rs/quinn/security/dependabot/1.


Ralith (approved):
I don't think we refer to tracing-subscriber in any of our libs, but still nice to have.

PR #2372, opened 2025-08-31
 Limit pacing batch size to 10ms of traffic

Ralith:
Fixes https://github.com/quinn-rs/quinn/issues/2286.

~Draft because this breaks assumptions implicit in some of our tests.~

### DIFF
--- a/quinn-proto/src/connection/pacing.rs
+++ b/quinn-proto/src/connection/pacing.rs
@@ -89,13 +89,14 @@ impl Pacer {
         }
 
         let elapsed_rtts = time_elapsed.as_secs_f64() / smoothed_rtt.as_secs_f64();
-        let new_tokens = window as f64 * 1.25 * elapsed_rtts;
-        self.tokens = self
-            .tokens
-            .saturating_add(new_tokens as _)
-            .min(self.capacity);
+        let new_tokens = (window as f64 * 1.25 * elapsed_rtts).round() as u64;
+        self.tokens = self.tokens.saturating_add(new_tokens).min(self.capacity);
 
-        self.prev = now;
+        // In the unlikely event that we're getting polled faster than tokens are generated, ensure
+        // that `elapsed_rtts` can grow until we make progress.
+        if new_tokens > 0 {
+            self.prev = now;
+        }
 
         // if we can already send a packet, there is no need for delay
         if self.tokens >= bytes_to_send {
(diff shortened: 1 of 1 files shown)


quinn-proto/src/connection/pacing.rs
```diff
@@ -128,23 +130,34 @@ impl Pacer {
...
-    let capacity = ((window as u128 * BURST_INTERVAL.as_nanos()) / rtt) as u64;
-
-    // Small bursts are less efficient (no GSO), could increase latency and don't effectively
-    // use the channel's buffer capacity. Large bursts might block the connection on sending.
-    capacity.clamp(MIN_BURST_SIZE * mtu as u64, MAX_BURST_SIZE * mtu as u64)
+    let mtu = u64::from(mtu);
+
+    let target_capacity = ((window as u128 * TARGET_BURST_INTERVAL.as_nanos()) / rtt) as u64;
+    // Never restrict capacity below one MTU.
+    let max_capacity = cmp::max(
```
djc: Nit: suggest using `Ord::max()` and `Ord::min()` instead -- `Ord` is part of the prelude.

Since `max_capacity` is being used in a `cmp::min()` call below, I'd guess `min_capacity` is a better name? (It also better matches the comment.)

quinn-proto/src/connection/pacing.rs
```diff
@@ -128,23 +130,34 @@ impl Pacer {
...
+    // worth of traffic.
+    cmp::min(
+        max_capacity,
+        target_capacity.clamp(MIN_BURST_SIZE * mtu, MAX_BURST_SIZE * mtu),
+    )
 }
 
-/// The burst interval
+/// Period of traffic to batch together on a reasonably fast connection
+const TARGET_BURST_INTERVAL: Duration = Duration::from_millis(2);
```
djc, line 151: Nit: maybe isolate this rename in a separate commit?

quinn-proto/src/connection/pacing.rs
Ralith: > Nit: suggest using Ord::max() and Ord::min() instead -- Ord is part of the prelude.

I *knew* I was forgetting something about that pattern.

> Since max_capacity is being used in a cmp::min() call below, I'd guess min_capacity is a better name? (It also better matches the comment.)

No, you're getting bitten by the usual confusion, which is exactly why I gave it a detailed name :D

Ralith:
> This looks semver-compatible, should we backport to 0.11.x?

If it addresses @vitocln's issue and they'd benefit from such a release, sure.

quinn-proto/src/connection/pacing.rs
djc: Ugh, that stuff hurts my head!

PR #2374, opened 2025-09-02
Use newer rustls-pki-types PEM parser API

djc:
Could potentially be backported to `0.11.x`.


Ralith:
Always nice to save a dep!

PR #2375, opened 2025-09-04
docs: add guarantees to the read and send_datagram function

Aderinom:
Related Issue: #2351 

Adds documentation for the required Gurantees to allow a noop poll to send/read_datagram function to act as a try_write/try_recv similar to e.g. [Tokio UDP Socket](https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.try_recv)

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -346,6 +346,10 @@ impl Connection {
     }
 
     /// Receive an application datagram
+    ///
+    /// Guarantees:
+    /// - Cancel-safe future
+    /// - First poll resolves immediately if a datagram is available
     pub fn read_datagram(&self) -> ReadDatagram<'_> {
         ReadDatagram {
             conn: &self.0,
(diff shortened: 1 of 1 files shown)


gretchenfrage:
Thank you! Always love a docs enhancement.

Now, I'm not opposed to this per se, but I do wanna take a moment to step back and question whether this is the right path to take regardless:

In principle, it shouldn't be that hard for a caller to perform the action of "poll a future, and if it returns `Pending` but also notifies its `Waker` before even returning, then poll it again, and repeat." It seems like the main thing holding this back is a slightly weird lack of any established library for it (see e.g. [my own code](https://github.com/gretchenfrage/aqueduct/blob/b2d51ba2d1a25929d3108efffcc7ef82aa045f0e/packages/aqueduct/src/channel/polling.rs) where I modified the `pollster` crate to be able to do that, but that's a bad example because that code is also doing other weirder things). We _could_ simply not commit to doing this, and expect our users to do it instead. In contrast to cancel safety, this doesn't substantively what functionality we are providing.

Thoughts @djc @ralith?

PR #2377, opened 2025-09-06
refactor: Pass ConnectionId by value

### LINKED ISSUE #2140: Pass `ConnectionId` by value
Per discussion in #2109, whenever we release 0.12, we want to change API-surfaced usages of `ConnectionId` so that they're always passed by value rather than shared reference, since it is a small copy type.

gretchenfrage:
Closes #2140


djc (approved):
Nice, thanks for following up on these!

PR #2382, opened 2025-09-08
Update dependencies

djc:
cc @mxinden


mxinden:
Thank you for the ping. This is fine for Firefox.

PR #2400, opened 2025-09-29
Black hole detection fixes

zdave-parity:
Relax an MTU discovery state assertion: while we will never start probing before the connection has been established, it is possible for black hole detection to trigger, which will put MTU discovery into the `Phase::Complete` state.

Fix some comparisons in the black hole detector:

- Lost packets with size exactly matching the minimum MTU should not be treated as suspicious; by definition the MTU will never be reduced below this.
- Similarly, lost packets with size exactly matching a more recent successfully transmitted packet should not be treated as suspicious.

### DIFF
--- a/quinn-proto/src/connection/mtud.rs
+++ b/quinn-proto/src/connection/mtud.rs
@@ -83,9 +83,14 @@ impl MtuDiscovery {
         self.current_mtu = self.current_mtu.min(peer_max_udp_payload_size);
 
         if let Some(state) = self.state.as_mut() {
-            // MTUD is only active after the connection has been fully established, so it is
-            // guaranteed we will receive the peer's transport parameters before we start probing
-            debug_assert!(matches!(state.phase, Phase::Initial));
+            // It is possible for black hole detection to trigger before the connection has been
+            // fully established, if the initial MTU is greater the minimum MTU. We should never
+            // send probes before the connection has been fully established and we have received
+            // the peer's transport parameters though.
+            debug_assert!(
+                !matches!(state.phase, Phase::Searching(_)),
+                "Transport parameters received after MTU probing started"
+            );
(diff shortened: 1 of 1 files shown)


zdave-parity:
There is a test which picks up this issue [here](https://github.com/zdave-parity/quinn/commit/3b46dc5aab72643ab0f167c56a0ade6eda21e0c0), but it's not really suitable for including in the test suite. It iterates over many patterns of handshake packet loss and so takes quite a while to run. Even with this bug fix the test doesn't actually pass because the static `CLIENT_PORTS` range in `tests/util.rs` gets exhausted :). It would be possible to check just the pattern which currently fails, but it's unclear to me how useful that would be. I think this is the sort of thing best caught by fuzz testing.

djc (approved):
LGTM, thanks!

Can you squash the test commit into the commit that broke it, and the formatting fix into the commit that originated the formatting deviation?

zdave-parity:
> Can you squash the test commit into the commit that broke it, and the formatting fix into the commit that originated the formatting deviation?

Sure, done.

djc:
@Ralith would be nice if you could take a look at this.

@zdave-parity `main` is currently aiming for a 0.12.0 release. Would you want/need this backported to 0.11.x?

zdave-parity:
> @zdave-parity `main` is currently aiming for a 0.12.0 release. Would you want/need this backported to 0.11.x?

I think we'd be happy to just upgrade to 0.12 when it is released.

Ralith:
> There is a test which picks up this issue [here](https://github.com/zdave-parity/quinn/commit/3b46dc5aab72643ab0f167c56a0ade6eda21e0c0), but it's not really suitable for including in the test suite

I'm not sure we should exclude such tests -- exhaustive searches and fuzzing are important, as this PR illustrates! Might want a separate test binary that gets built in release mode, or something, though...

PR #2407, opened 2025-10-01
Use `checked_mul` when calculating `congestion_period` to not panic on overflow

pablosichert:
When using a large enough value for `persistent_congestion_threshold`, the calculation for `congestion_period` panics.

In my case, I was using `u32::MAX` since I did not want the connection to ever be marked as persistently congested.

With this PR, the `congestion_period` is calculated using `checked_mul`, so that no panic on overflow is possible.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1679,8 +1679,10 @@ impl Connection {
         // InPersistentCongestion: Determine if all packets in the time period before the newest
         // lost packet, including the edges, are marked lost. PTO computation must always
         // include max ACK delay, i.e. operate as if in Data space (see RFC9001 §7.6.1).
-        let congestion_period =
-            self.pto(SpaceId::Data) * self.config.persistent_congestion_threshold;
+        let congestion_period = self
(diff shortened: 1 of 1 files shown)


pablosichert:
> though I'm not sure if your use case is a good idea...

I'm not disagreeing at all 😆 We're still experimenting with a lot of things.

gretchenfrage (changes requested):
Very nice! However, simplify by using `saturating_mul` instead please?

gretchenfrage:
@pablosichert no pressure, but if you [allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests) on this PR I can just do that for you.

djc:
@gretchenfrage might be faster to just merge this and have you send a separate PR after.

pablosichert:
Hol'up, I just saw since I declared bankruptcy on my GitHub notifications 😄

Going to make the edit! Didn't find a way to allow edits from maintainers 🤔

PR #2411, opened 2025-10-05
feat(quinn): Add `SendStream::set_reset_on_drop`

gretchenfrage:
The existing default behavior is preserved that, if a `SendStream` is dropped and it had not previously been finished or reset, then it automatically finishes. However, a new `SendStream::set_reset_on_drop` method is added which can be used to alter this behavior, so that if the stream is dropped without having been finished or reset it instead automatically resets with a provided error code.

The general concept of this has been [discussed in discord][1], wherein it was believed that there is value to such a behavior but that tradeoffs do not favor making it the default. As such, this commit leaves the existing default as-is, but allows it to be overridden.

[1]: https://discord.com/channels/976380008299917365/1063547094863978677/1370477346758328441

### DIFF
--- a/quinn/src/send_stream.rs
+++ b/quinn/src/send_stream.rs
@@ -35,6 +35,7 @@ pub struct SendStream {
     conn: ConnectionRef,
     stream: StreamId,
     is_0rtt: bool,
+    reset_on_drop: Option<VarInt>,
 }
 
 impl SendStream {
@@ -43,6 +44,7 @@ impl SendStream {
             conn,
             stream,
             is_0rtt,
+            reset_on_drop: None,
         }
     }
 
@@ -287,6 +289,15 @@ impl SendStream {
     ) -> Poll<Result<usize, WriteError>> {
         pin!(self.get_mut().write(buf)).as_mut().poll(cx)
     }
+
+    /// Set whether the stream automatically resets rather than finishes when dropped
+    ///
+    /// If this stream is dropped before [`finish()`](Self::finish) or [`reset()`](Self::reset) is
+    /// called on it, it automatically finishes if `error_code` is set to `None`, or automatically
+    /// resets with `error_code` if `error_code` is set to `Some`. Defaults to `None`.
+    pub fn set_reset_on_drop(&mut self, error_code: Option<VarInt>) {
+        self.reset_on_drop = error_code;
+    }
 }
 
 /// Check if a send stream is stopped.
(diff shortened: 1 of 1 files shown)


gretchenfrage:
@ralith @djc PR is still draft because I must write tests but thoughts on whether you would approve of this direction?

quinn/src/send_stream.rs
```diff
@@ -287,6 +289,15 @@ impl SendStream {
     ) -> Poll<Result<usize, WriteError>> {
         pin!(self.get_mut().write(buf)).as_mut().poll(cx)
     }
+
+    /// Set whether the stream automatically resets rather than finishes when dropped
+    ///
+    /// If this stream is dropped before [`finish()`](Self::finish) or [`reset()`](Self::reset) is
+    /// called on it, it automatically finishes if `error_code` is set to `None`, or automatically
+    /// resets with `error_code` if `error_code` is set to `Some`. Defaults to `None`.
+    pub fn set_reset_on_drop(&mut self, error_code: Option<VarInt>) {
```
djc, line 298: Would it make more sense to drop the Option wrapper from the argument here?

djc (commented):
Sure, this makes sense to me.

quinn/src/send_stream.rs
```diff
@@ -287,6 +289,15 @@ impl SendStream {
     ) -> Poll<Result<usize, WriteError>> {
         pin!(self.get_mut().write(buf)).as_mut().poll(cx)
     }
+
+    /// Set whether the stream automatically resets rather than finishes when dropped
```
djc, line 293: Nit: don't love "automatically" here, maybe "implicitly"?

quinn/src/send_stream.rs
gretchenfrage: It seems conceivably useful that it can be both set and "unset"

djc: Fair enough -- maybe it should not be `Option<VarInt>` but `StreamDrop::Reset(VarInt)`?

flub:
while I love the idea of providing this type of behaviour, I wonder if having this as a mutation property on the same type is a great API.  It is now very hard to know how it will behave, since you need to trace all the calls that may have been made on the object.  A manual and error-prone task.  Would a wrapping type be a better API?

An example is `tokio_util::task::AbortOnDropHandle`, you would clearly indicate its behaviour and also allows switching back to the wrapped type.

Ralith:
I have the same reservations as @flub. Alternative ideas:
- A wrapper `struct ResetOnDropStream(Option<SendStream>, VarInt)` that implements `Deref` and shadows `finish` and `reset`. I don't love cheesy `Deref` impls, but I also don't love mass-duplicating API...
- Just document a custom wrapper type as a good downstream practice

djc:
On reflection, I broadly agree with @flub's point.

(I also think `Deref` is a perfectly fine tool and should not be shunned when it fits the use case.)

gretchenfrage:
Alright, closing since people aren't in favor of this, and I realized it wouldn't have been as useful to my project as I originally thought when I prototyped this.

flub:
> I have the same reservations as @flub. Alternative ideas:
> 
>     * A wrapper `struct ResetOnDropStream(Option<SendStream>, VarInt)` that implements `Deref` and shadows `finish` and `reset`. I don't love cheesy `Deref` impls, but I also don't love mass-duplicating API...
> 
>     * Just document a custom wrapper type as a good downstream practice

I do like the idea of providing this out of the box.  Simply it existing also draws attention to the default behaviour.

Also agree that `Deref` is probably suitable here.

PR #2421, opened 2025-10-21
feat(ip_recverr): ADD IP_RECVERR and IPV6_RECVERR  support for linux 

### LINKED ISSUE #2052: Add support for `IP_RECVERR` and `IPV6_RECVERR`
Would be useful, for example, to receive `fragmentation needed and DF set` ICMP messages.

RaghavendraRQ:
On Linux, enables `IP_RECVERR` and `IPV6_RECVERR` socket options

This allows the socket to receive ICMP errors via the error **queue,** enabling faster detection of network unreachability instead of waiting for connection timeouts.

Features Implemented 

- Enable IP_RECVERR and IPV6_RECVERR socket options on Linux
- Add error queue polling in recv path
- Add IcmpError type for ICMP error reporting
- Add tests for ICMP error handling (Not Completed)

Closes #2052

### DIFF
--- a/quinn-udp/src/lib.rs
+++ b/quinn-udp/src/lib.rs
@@ -238,3 +238,18 @@ impl EcnCodepoint {
         })
     }
 }
+
+#[derive(Clone, Debug, Copy, PartialEq, Eq)]
+pub struct IcmpError {
+    pub dst: SocketAddr,
+    pub kind: IcmpErrorKind,
+}
+
+#[derive(Clone, Debug, Copy, PartialEq, Eq)]
+pub enum IcmpErrorKind {
+    NetworkUnreachable,
+    HostUnreachable,
+    PortUnreachable,
+    PacketTooBig,
+    Other { icmp_type: u8, icmp_code: u8 },
+}
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -14,6 +14,11 @@ use std::{
 
 use socket2::SockRef;
 
+use crate::IcmpError;
+
+#[cfg(target_os = "linux")]
+use crate::IcmpErrorKind;
+
 use super::{
     EcnCodepoint, IO_ERROR_LOG_INTERVAL, RecvMeta, Transmit, UdpSockRef, cmsg, log_sendmsg_error,
 };
@@ -31,6 +36,66 @@ pub(crate) struct msghdr_x {
     pub msg_controllen: libc::socklen_t,
     pub msg_flags: libc::c_int,
     pub msg_datalen: usize,
+}
+
+#[cfg(target_os = "linux")]
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct SockExtendedError {
+    pub errno: u32,
+    pub origin: u8,
+    pub r#type: u8,
+    pub code: u8,
+    pub pad: u8,
+    pub info: u32,
+    pub data: u32,
+}
+
+#[cfg(target_os = "linux")]
+impl SockExtendedError {
+    fn kind(&self) -> IcmpErrorKind {
+        const ICMP_DEST_UNREACH: u8 = 3; // Type 3: Destination Unreachable
+        const ICMP_NET_UNREACH: u8 = 0;
+        const ICMP_HOST_UNREACH: u8 = 1;
+        const ICMP_PORT_UNREACH: u8 = 3;
+        const ICMP_FRAG_NEEDED: u8 = 4;
+
+        const ICMPV6_DEST_UNREACH: u8 = 1; // Type 1: Destination Unreachable for IPv6
+        const ICMPV6_NO_ROUTE: u8 = 0;
+        const ICMPV6_ADDR_UNREACH: u8 = 1;
+        const ICMPV6_PORT_UNREACH: u8 = 4;
+
+        const ICMPV6_PACKET_TOO_BIG: u8 = 2;
+
+        match (self.origin, self.r#type, self.code) {
+            (libc::SO_EE_ORIGIN_ICMP, ICMP_DEST_UNREACH, ICMP_NET_UNREACH) => {
+                IcmpErrorKind::NetworkUnreachable
+            }
+            (libc::SO_EE_ORIGIN_ICMP, ICMP_DEST_UNREACH, ICMP_HOST_UNREACH) => {
+                IcmpErrorKind::HostUnreachable
+            }
+            (libc::SO_EE_ORIGIN_ICMP, ICMP_DEST_UNREACH, ICMP_PORT_UNREACH) => {
+                IcmpErrorKind::PortUnreachable
+            }
+            (libc::SO_EE_ORIGIN_ICMP, ICMP_DEST_UNREACH, ICMP_FRAG_NEEDED) => {
+                IcmpErrorKind::PacketTooBig
+            }
+            (libc::SO_EE_ORIGIN_ICMP6, ICMPV6_DEST_UNREACH, ICMPV6_NO_ROUTE) => {
+                IcmpErrorKind::NetworkUnreachable
+            }
+            (libc::SO_EE_ORIGIN_ICMP6, ICMPV6_DEST_UNREACH, ICMPV6_ADDR_UNREACH) => {
+                IcmpErrorKind::HostUnreachable
+            }
+            (libc::SO_EE_ORIGIN_ICMP6, ICMPV6_DEST_UNREACH, ICMPV6_PORT_UNREACH) => {
+                IcmpErrorKind::PortUnreachable
+            }
+            (libc::SO_EE_ORIGIN_ICMP6, ICMPV6_PACKET_TOO_BIG, _) => IcmpErrorKind::PacketTooBig,
+            _ => IcmpErrorKind::Other {
+                icmp_type: self.r#type,
+                icmp_code: self.code,
+            },
+        }
+    }
 }
 
 #[cfg(apple_fast)]
@@ -115,8 +180,21 @@ impl UdpSocketState {
             if let Err(_err) =
                 set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_RECVTOS, OPTION_ON)
             {
-                crate::log::debug!("Ignoring error setting IP_RECVTOS on socket: {_err:?}");
+                crate::log::debug!("Ignoring error setting IP_RECVTOS on socket: {}", _err);
             }
+        }
+
+        // Enable IP_RECVERR and IPV6_RECVERR for ICMP Errors
+        #[cfg(target_os = "linux")]
+        if is_ipv4 {
+            if let Err(err) = set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_RECVERR, OPTION_ON)
+            {
+                crate::log::debug!("Failed to enable IP_RECVERR: {}", err);
+            }
+        } else if let Err(err) =
+            set_socket_option(&*io, libc::IPPROTO_IPV6, libc::IPV6_RECVERR, OPTION_ON)
+        {
+            crate::log::debug!("Failed to enable IPV6_RECVERR: {}", err);
         }
 
         let mut may_fragment = false;
@@ -232,6 +310,10 @@ impl UdpSocketState {
         meta: &mut [RecvMeta],
     ) -> io::Result<usize> {
         recv(socket.0, bufs, meta)
+    }
+
+    pub fn recv_icmp_err(&self, socket: UdpSockRef<'_>) -> io::Result<Option<IcmpError>> {
+        recv_err(socket.0)
     }
 
     /// The maximum amount of segments which can be transmitted if a platform
@@ -507,6 +589,7 @@ fn recv(io: SockRef<'_>, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) -> 
     for i in 0..(msg_count as usize) {
         meta[i] = decode_recv(&names[i], &hdrs[i].msg_hdr, hdrs[i].msg_len as usize)?;
     }
+
     Ok(msg_count as usize)
 }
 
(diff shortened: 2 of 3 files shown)


quinn-udp/src/unix.rs
```diff
@@ -795,6 +841,96 @@ fn decode_recv(
...
+    // Have followed the previous declarations
+    let mut hdr: libc::msghdr = unsafe { mem::zeroed() };
+    hdr.msg_name = &mut addr_storage as *mut _ as *mut _;
+    hdr.msg_namelen = mem::size_of::<libc::sockaddr_storage>() as libc::socklen_t;
+    hdr.msg_iov = &mut iovec;
+    hdr.msg_iovlen = 1;
+    hdr.msg_control = control.0.as_mut_ptr() as *mut _;
+    hdr.msg_controllen = control.0.len() as _;
+
+    let ret = unsafe { libc::recvmsg(fd, &mut hdr, libc::MSG_ERRQUEUE) };
```
mxinden, line 920: I wonder what performance impact this additional sys call has. Would you mind running the benchmark suite with and without and post the results here?

https://github.com/quinn-rs/quinn/blob/main/quinn-udp/benches/throughput.rs

mxinden (commented):
Good work. I will take a more in-depth look.

quinn-udp/src/unix.rs
thomaseizinger: Even if it has low overhead, I think it might be better to add a new function to the socket, called something like: `recv_icmp_errors`.

That means users who don't care about this functionality don't need to pay its cost, plus it will allow us to return a strongly-typed struct like `IcmpError` from this function without having to change existing APIs.

Returning ICMP messages as errors from `recv` seems a bit odd to me because it is not the `recv` call that is failing.

RaghavendraRQ: Thanks for the feedback. Calling it automatically in `recv` was a bit heavy.
I will change it to separate `recv_icmp_err` as you have suggested, so people can opt-in when they need it.

I have ran the benchmark suite and there is noticeable performance regression as you said.  here is the result: 

**diff.txt:** 
[diff.txt](https://github.com/user-attachments/files/23047519/diff.txt)

After the refactor, the benchmarks should show zero performance impact since we're removing the syscall from recv entirely

quinn-udp/src/lib.rs
```diff
@@ -238,3 +238,52 @@ impl EcnCodepoint {
...
+    ee_type: u8,
+    ee_code: u8,
+    ee_pad: u8,
+    ee_info: u32,
+    ee_data: u32,
+}
+
+#[cfg(target_os = "linux")]
+#[derive(Clone, Debug, Copy)]
+pub struct ICMPError {
```
thomaseizinger:
```suggestion
pub struct IcmpError {
```

I believe that is the convention followed here, i.e. we also don't say `UDPSocketState` but `UdpSocketState`.

quinn-udp/src/lib.rs
```diff
@@ -238,3 +238,52 @@ impl EcnCodepoint {
...
+    pub addr: SocketAddr,
```
thomaseizinger: Is this the socket we tried to send to? Maybe we can name this `dst` or something?

quinn-udp/src/unix.rs
```diff
@@ -122,6 +125,20 @@ impl UdpSocketState {
             }
         }
 
+        // Enable IP_RECVERR and IPV6_RECVERR for ICMP Errors
+        #[cfg(target_os = "linux")]
+        if is_ipv4 {
+            if let Err(_err) =
+                set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_RECVERR, OPTION_ON)
+            {
+                crate::log::warn!("Failed to enable IP_RECVERR: {}", _err);
```
thomaseizinger: IMO, warn is a bit much for this, I think logging this on debug would suffice.

quinn-udp/src/unix.rs
```diff
@@ -233,6 +250,16 @@ impl UdpSocketState {
...
     }
 
+    #[cfg(target_os = "linux")]
+    pub fn recv_icmp_err(&self, socket: UdpSockRef<'_>) -> io::Result<Option<ICMPError>> {
+        recv_err(socket.0)
+    }
+
+    // #[cfg(not(target_os = "linux"))]
+    // pub fn recv_icmp_err(&self, _socket: UdpSockRef<'_>) -> io::Result<Option<ICMPError>> {
+    //     Ok(None)
```
thomaseizinger: I think this fallback makes sense.

quinn-udp/src/unix.rs
```diff
@@ -795,6 +824,92 @@ fn decode_recv(
...
 
+#[cfg(target_os = "linux")]
+fn recv_err(io: SockRef<'_>) -> io::Result<Option<ICMPError>> {
+    use std::mem;
+
+    let fd = io.as_raw_fd();
+
+    let mut control = cmsg::Aligned([0u8; CMSG_LEN]);
+
+    // SEnding error info for now no data is recieved
```
thomaseizinger: Is this a left-over from iterating on the implementation?

quinn-udp/src/unix.rs
```diff
@@ -795,6 +824,92 @@ fn decode_recv(
...
+    let cmsg_iter = unsafe { cmsg::Iter::new(&hdr) };
+
+    for cmsg in cmsg_iter {
+        const IP_RECVERR: libc::c_int = 11;
+        const IPV6_RECVERR: libc::c_int = 25;
+
+        let is_ip_err = cmsg.cmsg_level == libc::IPPROTO_IP && cmsg.cmsg_type == IP_RECVERR;
+        let is_ipv6_err = cmsg.cmsg_level == libc::IPPROTO_IPV6 && cmsg.cmsg_type == IPV6_RECVERR;
+
+        if is_ip_err || is_ipv6_err {
```
thomaseizinger: I'd suggest we flip this around: Negate it and use `continue`. That avoids right-wards drift of the code below.

quinn-udp/src/unix.rs
```diff
@@ -795,6 +824,92 @@ fn decode_recv(
...
+        return Err(err);
+    }
+
+    let cmsg_iter = unsafe { cmsg::Iter::new(&hdr) };
+
+    for cmsg in cmsg_iter {
+        const IP_RECVERR: libc::c_int = 11;
+        const IPV6_RECVERR: libc::c_int = 25;
+
+        let is_ip_err = cmsg.cmsg_level == libc::IPPROTO_IP && cmsg.cmsg_type == IP_RECVERR;
```
thomaseizinger:
```suggestion
        let is_ipv4_err = cmsg.cmsg_level == libc::IPPROTO_IP && cmsg.cmsg_type == IP_RECVERR;
```

For consistency.

quinn-udp/src/unix.rs
```diff
@@ -795,6 +824,92 @@ fn decode_recv(
...
+        let is_ipv6_err = cmsg.cmsg_level == libc::IPPROTO_IPV6 && cmsg.cmsg_type == IPV6_RECVERR;
+
+        if is_ip_err || is_ipv6_err {
+            let err_data = unsafe { cmsg::decode::<SockExtendedErr, libc::cmsghdr>(cmsg) };
+
+            let addr = unsafe {
```
thomaseizinger: Do we not have this code somewhere already in the codebase? Perhaps extracting a small utility function could make this a bit more concise.

quinn-udp/src/unix.rs
```diff
@@ -795,6 +824,92 @@ fn decode_recv(
...
+                    libc::AF_INET6 => {
+                        let addr_in6 = &*(addr_ptr as *const libc::sockaddr_in6);
+                        SocketAddr::V6(std::net::SocketAddrV6::new(
+                            std::net::Ipv6Addr::from(addr_in6.sin6_addr.s6_addr),
+                            u16::from_be(addr_in6.sin6_port),
+                            addr_in6.sin6_flowinfo,
+                            addr_in6.sin6_scope_id,
+                        ))
+                    }
+                    _ => return Ok(None), // Unknown address family
```
thomaseizinger: Should we log a warning instead and continue to the next cmsg?

quinn-udp/src/unix.rs
```diff
@@ -795,6 +824,92 @@ fn decode_recv(
...
+            return Ok(Some(ICMPError {
+                addr,
+                kind: crate::ICMPErrorKind::from_extended_err(&err_data),
+            }));
+        }
+    }
+    Ok(None)
+}
+
+// I don't know about how other platforms handle this.
```
thomaseizinger: I think the proposed fallback makes sense for now.

quinn-udp/tests/tests.rs
```diff
@@ -369,3 +369,158 @@ fn ip_to_v6_mapped(x: IpAddr) -> IpAddr {
...
+        destination: unreachable_addr,
+        ecn: None,
+        contents: b"test packet to unreachable destination",
+        segment_size: None,
+        src_ip: None,
+    };
+
+    // Send the packet
+    let state_result = state.try_send((&socket).into(), &transmit);
+    assert!(state_result.is_err(), "Expected to fail to transmit");
```
thomaseizinger, line 396: What is the error we are getting here? Do we get the ICMP error twice now? I imagine locally generated ICMP errors are directly delivered to the socket whereas the ones from intermediary routers would not show up here.

quinn-udp/tests/tests.rs
```diff
@@ -369,3 +369,158 @@ fn ip_to_v6_mapped(x: IpAddr) -> IpAddr {
...
+    //     std::thread::sleep(Duration::from_millis(10));
+    // }
+
+    // if received_icmp_error {
+    //     println!("IP_RECVERR is working! Received ICMP error.");
+    // } else {
+    //     println!("No ICMP error received (may be normal depending on network config)");
+    // }
+
+    match state.recv_icmp_err((&socket).into()) {
```
thomaseizinger, line 400: Instead of `match`, should we just unwrap?

thomaseizinger (commented):
Gave this a first pass.

quinn-udp/src/unix.rs
RaghavendraRQ: Nice! That's much cleaner - never thought of inverting the condition to reduce nesting.

quinn-udp/tests/tests.rs
RaghavendraRQ: Agreed! Switched to `expect()`. Saw the tests above. I think it will be better.

quinn-udp/src/unix.rs
RaghavendraRQ: Changed to log a warning and continue to the next cmsg instead of returning early

quinn-udp/tests/tests.rs
RaghavendraRQ: The EINVAL error was caused by using socket2::Socket, which doesn’t produce a valid UdpSockRef. I’ve switched to using std::net::UdpSocket, which aligns with how `try_send` expects sockets to be passed. After this change, the send call produces a proper network error  and recv_icmp_err receives the ICMP error as expected. We won’t get duplicate ICMP events.

quinn-udp/src/unix.rs
RaghavendraRQ: I considered factoring it out, but this logic is fairly localized and only used here. so I’ll keep it inline for clarity.

RaghavendraRQ:
This is my first open-source contribution, so I understand there may be several mistakes. Thank you for taking the time to review my work and provide guidance—I really appreciate it.

quinn-udp/src/unix.rs
```diff
@@ -233,6 +250,16 @@ impl UdpSocketState {
         recv(socket.0, bufs, meta)
     }
 
+    #[cfg(target_os = "linux")]
+    pub fn recv_icmp_err(&self, socket: UdpSockRef<'_>) -> io::Result<Option<IcmpError>> {
+        recv_err(socket.0)
+    }
+
+    #[cfg(not(target_os = "linux"))]
```
thomaseizinger: I think we should aim for a unified API across all platforms, even if they currently don't return a value. Otherwise code that uses this also needs to use a `cfg` to compile on multiple platforms.

quinn-udp/src/unix.rs
```diff
@@ -475,6 +502,7 @@ fn recv(io: SockRef<'_>, bufs: &mut [IoSliceMut<'_>], meta: &mut [RecvMeta]) ->
             &mut hdrs[i].msg_hdr,
         );
     }
+
```
thomaseizinger: Nit: Let's avoid unnecessary formatting changes.

quinn-udp/src/unix.rs
```diff
@@ -795,6 +824,99 @@ fn decode_recv(
...
 
+#[cfg(target_os = "linux")]
+fn recv_err(io: SockRef<'_>) -> io::Result<Option<IcmpError>> {
+    use std::mem;
+
+    let fd = io.as_raw_fd();
+
+    let mut control = cmsg::Aligned([0u8; CMSG_LEN]);
+
+    // We don't need actual data, just the error info
```
thomaseizinger, line 904: Oh! I understand now. So to receive error info, we pass a flag to `recvmsg` but otherwise it is the same syscall. An alternative design would be to do this as part of the other `recv` flow and store the received ICMP messages in a (bounded) queue. That would be more efficient in terms of syscall usage, especially because most of the time, this syscall here is not going to return anything. Yet, an application wanting to handle this still needs to call this in a loop with the other `recv`, taking up CPU usage of the recv hot path.

Curious to hear what other people think about this.

thomaseizinger (commented):
Great progress. One big item to discuss.

quinn-udp/src/unix.rs
djc: Your suggestion makes sense to me.

quinn-udp/src/lib.rs
```diff
@@ -238,3 +238,52 @@ impl EcnCodepoint {
...
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+struct SockExtendedErr {
+    ee_errno: u32,
+    ee_origin: u8,
+    ee_type: u8,
+    ee_code: u8,
+    ee_pad: u8,
+    ee_info: u32,
+    ee_data: u32,
```
djc: Nit: without any compelling reason to the contrary (like shared standardization with other APIs), suggest naming this `SocketExtendedError` and dropping the common `ee_` prefix on fields (which is a kind of stuttering).

Also, since these only seem relevant on Unix, suggest that these should live in the `unix` module?

quinn-udp/src/lib.rs
```diff
@@ -238,3 +238,52 @@ impl EcnCodepoint {
...
+}
+
+#[cfg(target_os = "linux")]
+#[derive(Clone, Debug, Copy)]
+pub struct IcmpError {
+    pub dst: SocketAddr,
+    pub kind: IcmpErrorKind,
+}
+#[cfg(target_os = "linux")]
```
djc: Nit: empty line between items.

quinn-udp/src/lib.rs
```diff
@@ -238,3 +238,52 @@ impl EcnCodepoint {
...
+#[derive(Clone, Debug, Copy, PartialEq, Eq)]
+pub enum IcmpErrorKind {
+    NetworkUnreachable,
+    HostUnreachable,
+    PortUnreachable,
+    PacketTooBig,
+    Other { icmp_type: u8, icmp_code: u8 },
```
djc, line 254: Nit: drop the `icmp_` prefix, which is pretty obvious for a type named `IcmpErrorKind`.

quinn-udp/src/lib.rs
```diff
@@ -238,3 +238,52 @@ impl EcnCodepoint {
...
+}
+
+#[cfg(target_os = "linux")]
+impl IcmpErrorKind {
+    fn from_extended_err(err: &SockExtendedErr) -> Self {
```
djc: It feels like this should be `SocketExtendedError::kind()`.

Can we name the magic constants for `type` and `code` here with their common libc or whatever names?

quinn-udp/src/lib.rs
RaghavendraRQ: Since `type` is a reserved keyword in Rust, it can’t be used directly as a field name, We can use raw identifier instead `r#type`. Can i do that?

djc: Seems okay to me.

djc:
Please squash all your changes before pushing.

quinn-udp/src/unix.rs
mxinden: Interesting. In case someone here knows, what is the intended use, or asked differently, how do other applications use it?

quinn-udp/src/lib.rs
djc: > Also, since these only seem relevant on Unix, suggest that these should live in the `unix` module?

I don't think you addressed this.

Also, please keep it these types near the bottom of the module to maintain the top-down ordering of items.

djc:
> Please squash all your changes before pushing.

Including merges.

quinn-udp/src/lib.rs
```diff
@@ -238,3 +238,23 @@ impl EcnCodepoint {
...
+
+#[cfg(target_os = "linux")]
+#[derive(Clone, Debug, Copy)]
+pub struct IcmpError {
+    pub dst: SocketAddr,
+    pub kind: IcmpErrorKind,
+}
+
+#[cfg(not(target_os = "linux"))]
+pub struct IcmpError;
```
gretchenfrage: Having the struct be unit-like on non-linux doesn't seem necessary, and requires consumers of the library to do platform-specific conditional compilation. Unless I'm misinterpreting, there should just be a single non-conditional `IcmpError` struct and `IcmpErrorKind` struct, and conditional compilation can be used to make functions which return them always return `Err` or `None` or something on non-linux.

quinn-udp/src/unix.rs
```diff
@@ -233,6 +312,16 @@ impl UdpSocketState {
         recv(socket.0, bufs, meta)
     }
 
+    #[cfg(target_os = "linux")]
+    pub fn recv_icmp_err(&self, socket: UdpSockRef<'_>) -> io::Result<Option<IcmpError>> {
+        recv_err(socket.0)
+    }
```
gretchenfrage: This `#[cfg(target_os = "linux")]` function and the `#[cfg(not(target_os = "linux"))]` function below it are identical and so you should be able to reduplicate them and remove the conditional compilation.

quinn-udp/src/unix.rs
```diff
@@ -795,6 +885,99 @@ fn decode_recv(
     })
 }
 
+#[cfg(target_os = "linux")]
+fn recv_err(io: SockRef<'_>) -> io::Result<Option<IcmpError>> {
+    use std::mem;
```
gretchenfrage: Prefer this to be imported on the top of the module.

quinn-udp/src/unix.rs
```diff
@@ -795,6 +885,99 @@ fn decode_recv(
...
+
+    // We don't need actual data, just the error info
+    let mut iovec = libc::iovec {
+        iov_base: std::ptr::null_mut(),
+        iov_len: 0,
+    };
+
+    let mut addr_storage: libc::sockaddr_storage = unsafe { mem::zeroed() };
+
+    // Have followed the previous declarations
```
gretchenfrage: Would you mind changing this comment to something more clear, I'm not sure what it means

quinn-udp/src/unix.rs
```diff
@@ -122,6 +187,20 @@ impl UdpSocketState {
...
+        if is_ipv4 {
+            if let Err(_err) =
+                set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_RECVERR, OPTION_ON)
+            {
+                crate::log::debug!("Failed to enable IP_RECVERR: {}", _err);
+            }
+        } else if let Err(_err) =
+            set_socket_option(&*io, libc::IPPROTO_IPV6, libc::IPV6_RECVERR, OPTION_ON)
+        {
+            crate::log::debug!("Failed to enable IPV6_RECVERR: {}", _err);
```
gretchenfrage: `_err` can be renamed to `err`, as you _are_ using it (in the call to the `debug!` macro)

gretchenfrage (changes requested):
Thank you, looks like a valuable feature! Some minor code change requests

syszery:
Hi @RaghavendraRQ,

I noticed this PR has been quiet for a while. I’ve been working on a fresh implementation locally that follows the “separate opt-in function” approach suggested by @thomaseizinger, while leveraging the existing `ControlMetadata::decode` method. It’s Linux/Android only.

Would you be okay if I take over and submit a new PR based on this design? I would be happy to credit your initial work as a co-author.

RaghavendraRQ:
Hi @syszery 

I’ve been working on a few other things and completely forgot about this. I’d be totally okay with you taking it over - feel free to submit a new PR.

PR #2423, opened 2025-10-24
fix(pacing): allow ±1ns tolerance in computes_pause_correctly test on i386

ice-user:
### Summary
This PR fixes a test failure on 32-bit (i386) architectures in `connection::pacing::tests::computes_pause_correctly`.

### Context
On i386, the test occasionally fails due to a one-nanosecond rounding difference:

This appears to be caused by floating-point rounding during the conversion between `f64` and integer types in `Duration` arithmetic.

### Changes
- Updated the test to allow a ±1ns tolerance when comparing durations.
- This prevents spurious failures on 32-bit architectures while keeping test precision intact.

### Notes
- Verified locally that all tests pass on x86_64.
- Should not affect runtime behavior, only the test assertion.
- Related: Debian CI report where this issue was originally observed on i386.

---

**Closes:** N/A  
**Type:**  Test Fix

### DIFF
--- a/quinn-proto/src/connection/pacing.rs
+++ b/quinn-proto/src/connection/pacing.rs
@@ -273,14 +273,18 @@ mod tests {
 
         let pace_duration = Duration::from_nanos((TARGET_BURST_INTERVAL.as_nanos() * 4 / 5) as u64);
 
-        assert_eq!(
-            pacer
-                .delay(rtt, mtu as u64, mtu, window, old_instant)
-                .expect("Send must be delayed")
-                .duration_since(old_instant),
-            pace_duration
-        );
+        let actual_delay = pacer
+            .delay(rtt, mtu as u64, mtu, window, old_instant)
+            .expect("Send must be delayed")
+            .duration_since(old_instant);
 
+        let diff = actual_delay.abs_diff(pace_duration);
+
+        // Allow up to 2ns difference due to rounding
+        assert!(
+            diff < Duration::from_nanos(2),
+            "expected ≈ {pace_duration:?}, got {actual_delay:?} (diff {diff:?})"
+        );
         // Refill half of the tokens
         assert_eq!(
             pacer.delay(


Ralith (approved):
Thanks! I'm not really satisfied by the "rounding" explanation (Rust should be using the same IEEE754-compliant ops regardless), but maybe there's some nuance to how `Duration` is represented, and in the end all that matters is that the behavior is reasonable.

quinn-proto/src/connection/pacing.rs
```diff
@@ -273,12 +273,24 @@ mod tests {
...
-                .delay(rtt, mtu as u64, mtu, window, old_instant)
-                .expect("Send must be delayed")
-                .duration_since(old_instant),
-            pace_duration
+        let actual_delay = pacer
+            .delay(rtt, mtu as u64, mtu, window, old_instant)
+            .expect("Send must be delayed")
+            .duration_since(old_instant);
+
+        let diff = if actual_delay > pace_duration {
```
djc: Nit: suggest using `Duration::abs_diff()`?

quinn-proto/src/connection/pacing.rs
```diff
@@ -273,12 +273,24 @@ mod tests {
...
+            pace_duration - actual_delay
+        };
+
+        // Allow up to 2ns difference due to rounding
+        assert!(
+            diff < Duration::from_nanos(2),
+            "expected ≈ {:?}, got {:?} (diff {:?})",
+            pace_duration,
+            actual_delay,
+            diff
```
djc: Nit: inline format arguments, please.

djc (approved):
Thanks! LGTM, modulo nits.

quinn-proto/src/connection/pacing.rs
ice-user: Helloo ! ,I’ve inlined the format arguments as suggested, but this version isn’t cargo fmt–compatible — it gets reformatted automatically when I run the formatter.
I can revert to the cargo fmt version to make CI pass, or add #[rustfmt::skip] if inline formatting is preferred.
Please let me know which you’d prefer so I can update the PR accordingly.

quinn-proto/src/connection/pacing.rs
ice-user: hey . I also  updated the code to use Duration::abs_diff() instead of the manual if/else difference calculation as suggested.

quinn-proto/src/connection/pacing.rs
```diff
@@ -273,13 +273,15 @@ mod tests {
...
-        );
+        let actual_delay = pacer
+            .delay(rtt, mtu as u64, mtu, window, old_instant)
+            .expect("Send must be delayed")
+            .duration_since(old_instant);
+
+        let diff = actual_delay.abs_diff(pace_duration);
+
+        // Allow up to 2ns difference due to rounding
+        assert!( diff < Duration::from_nanos(2),"expected ≈ {:?}, got {:?} (diff {:?})", pace_duration,actual_delay,diff);
```
djc: I meant this (and then run `cargo fmt`).

```suggestion
        assert!(diff < Duration::from_nanos(2), "expected ≈ {pace_duration:?}, got {actual_delay:?} (diff {diff:?})");
```

ice-user: formatted!

djc: You didn't actually apply my suggestion...

ice-user: my bad  , there was some issue with the IDE i will fix it .

PR #2432, opened 2025-10-30
feat(quinn): add serde support

cathaysia:
Add serde support for status structs


quinn-proto/src/connection/stats.rs
```diff
@@ -6,6 +6,7 @@ use crate::{Dir, Duration, frame::Frame};
 ///
 /// All QUIC packets are carried by UDP datagrams. Hence, these statistics cover all traffic on a connection.
 #[derive(Default, Debug, Copy, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
```
djc: Nit: would prefer to (conditionally) import `Serialize` at the top of the module.

djc (approved):
This makes sense to me, thanks!

`main` is currently targeting 0.12 -- if you need this sooner, consider backporting to `0.11.x` after merge.

cathaysia:
Thanks, but the main branch is enough for me for now. ;)

cathaysia:
This might not be my problem. Do I need to fix this in this PR?

djc:
> This might not be my problem. Do I need to fix this in this PR?

Nope.

PR #2434, opened 2025-10-30
Apply suggestions from clippy 1.91

djc:


Ralith (approved):

djc:
@matheus23 do you know what might be up with the WASM breakage here? I suppose it is related to the 1.91 upgrade somehow?

matheus23:
Taking a look :eyes:

matheus23:
Yeah indeed, I can reproduce this failure in 1.91, but not in 1.89.
And it's `Instant::checked_sub` which fails in Wasm.
Not sure why the rust version has an effect on this, given it should run the same `web_time::Instant` code in both cases.
I'll keep looking.

matheus23:
This now also fails for me in 1.89. I'm debugging the values: `now: Instant(15.746652507s) loss_delay: 17.246250153s`.
So this is running into basically a version of https://github.com/rust-lang/rust/issues/100141

I think this *can* be an issue in other platforms as well (not only Wasm, e.g. Mac, as the issue above suggests), as the `Instant` isn't guaranteed to have sufficient precision for the subtraction.

Anyway, we need to avoid `.unwrap()`ing here. I've opened #2436 to fix this issue

PR #2435, opened 2025-10-31
feat(quinn-proto): export min_initial_size

cathaysia:
Export the `min_initial_size` parameter as an unsafe parameter. This should be helpful for environments where interoperability is not a concern and the MTU is less than 1200.

If this function is not exported, users in this situation can only modify this parameter by fork this repo.

### DIFF
--- a/quinn-proto/src/config/mod.rs
+++ b/quinn-proto/src/config/mod.rs
@@ -18,8 +18,9 @@ use crate::NoneTokenLog;
 #[cfg(any(feature = "rustls-aws-lc-rs", feature = "rustls-ring"))]
 use crate::crypto::rustls::{QuicServerConfig, configured_provider};
 use crate::{
-    DEFAULT_SUPPORTED_VERSIONS, Duration, MAX_CID_SIZE, RandomConnectionIdGenerator, SystemTime,
-    TokenLog, TokenMemoryCache, TokenStore, VarInt, VarIntBoundsExceeded,
+    DEFAULT_SUPPORTED_VERSIONS, Duration, MAX_CID_SIZE, MIN_INITIAL_SIZE,
+    RandomConnectionIdGenerator, SystemTime, TokenLog, TokenMemoryCache, TokenStore, VarInt,
+    VarIntBoundsExceeded,
     cid_generator::{ConnectionIdGenerator, HashedConnectionIdGenerator},
     crypto::{self, HandshakeTokenKey, HmacKey},
     shared::ConnectionId,
@@ -48,6 +49,7 @@ pub struct EndpointConfig {
     pub(crate) min_reset_interval: Duration,
     /// Optional seed to be used internally for random number generation
     pub(crate) rng_seed: Option<[u8; 32]>,
+    pub(crate) min_initial_size: u16,
 }
 
 impl EndpointConfig {
@@ -63,6 +65,7 @@ impl EndpointConfig {
             grease_quic_bit: true,
             min_reset_interval: Duration::from_millis(20),
             rng_seed: None,
+            min_initial_size: MIN_INITIAL_SIZE,
         }
     }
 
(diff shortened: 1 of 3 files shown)


cathaysia:
@djc What do you think of this?

djc:
There's no need to mention me. I watch this repo and will get notified for new issues and PRs.

quinn-proto/src/config/mod.rs
```diff
@@ -159,6 +162,25 @@ impl EndpointConfig {
...
+    ///
+    /// According to the quinn specification, this value MUST BE 1200. Modifying this value may have
+    /// unexpected consequences.
+    ///
+    /// </div>
+    ///
+    /// # Safety
+    ///
+    /// This function should only be called when you do not need to consider interoperability and know
+    /// exactly what you are doing.
```
Ralith: That's not what `unsafe` means in Rust. Calling this function cannot introduce undefined behavior, so it is safe. Safe doesn't mean correct.

quinn-proto/src/config/mod.rs
```diff
@@ -159,6 +162,25 @@ impl EndpointConfig {
...
         self
     }
+
+    /// Minimum size of the first Initial packet. You should NEVER consider calling this function
+    /// under any circumstances.
+    ///
+    /// <div class="warning">
+    ///
+    /// According to the quinn specification, this value MUST BE 1200. Modifying this value may have
+    /// unexpected consequences.
```
Ralith: Quinn doesn't have a specification. Quinn is an implementation of the QUIC specification. Changing this value makes Quinn's behavior non-conformant with the QUIC specification. For this documentation to be actionable, it should outline *what* the consequences are, not just handwave them as "unexpected".

quinn-proto/src/config/mod.rs
```diff
@@ -159,6 +162,25 @@ impl EndpointConfig {
         self.rng_seed = seed;
         self
     }
+
+    /// Minimum size of the first Initial packet. You should NEVER consider calling this function
+    /// under any circumstances.
```
Ralith: The strength of this wording contradicts your intent in submitting this PR. If it's "NEVER" appropriate to call this, then we shouldn't merge the change. If the interface is worth having, then its documentation should explain when it makes sense to use.

quinn-proto/src/config/mod.rs
```diff
@@ -159,6 +162,25 @@ impl EndpointConfig {
...
+    /// According to the quinn specification, this value MUST BE 1200. Modifying this value may have
+    /// unexpected consequences.
+    ///
+    /// </div>
+    ///
+    /// # Safety
+    ///
+    /// This function should only be called when you do not need to consider interoperability and know
+    /// exactly what you are doing.
+    pub unsafe fn min_initial_size(&mut self, value: u16) -> &mut Self {
```
Ralith, line 206: How does this interact with/compare to `TransportConfig::min_mtu`? Document.

Ralith (commented):
Can you elaborate on your use case? Forking Quinn to introduce non-conformant behavior for a niche application isn't necessarily the wrong answer.

quinn-proto/src/config/mod.rs
cathaysia: From an implementation standpoint, this is a safe approach. However, changing it to below 1200 violates the QUIC specification. This might introduce interoperability issues, so I've marked it as unsafe.

cathaysia:
> Can you elaborate on your use case? Forking Quinn to introduce non-conformant behavior for a niche application isn't necessarily the wrong answer.

In my case, I'm using Quinn as a reliable UDP-based transport and multiplexer. I don't need it to interoperate with other protocols. Furthermore, the MTU size is variable and has a high probability of falling below 1200 bytes.

For me, BBR is the congestion control I care about most because my network is a long fat pipeline.

quinn-proto/src/config/mod.rs
djc: Violating the QUIC specification does not make it "unsafe" as usually understood by Rust developers. For this PR to be acceptable, you should drop the safety wording and instead talk about protocol/spec compliance.

cathaysia: I would rather not merge this pr if it wasn't marked as unsafe. :(

PR #2436, opened 2025-11-03
fix(quinn-proto): Avoid underflow panic in packet loss `Instant` calculations

matheus23:
Addresses the failure in #2434 

Subtractions from `Instant` can underflow (and thus panic), because the `u64` stored internally might be relative to some startup time that might be too recent.

This is what caused the failure in Wasm: The instant was relative to the process startup time, in one case ~15s, but the `loss_delay` was set to ~17s in that case.

Instead, this PR aims to compare the saturated difference of `now` and `info.time_sent`, and see if that's bigger than `loss_delay` or not.

Unfortunately I don't know of a different way of resolving this. There's no such thing as `Instant::saturating_sub`, and there's no `Instant::MIN` or similar with which it could be implemented.

Even though this addresses a problem in Wasm, this case can also happen in MacOS and other platforms, but requires  different, less common edge-cases to trigger.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1712,8 +1712,6 @@ impl Connection {
         let rtt = self.path.rtt.conservative();
         let loss_delay = cmp::max(rtt.mul_f32(self.config.time_threshold), TIMER_GRANULARITY);
 
-        // Packets sent before this time are deemed lost.
-        let lost_send_time = now.checked_sub(loss_delay).unwrap();
         let largest_acked_packet = self.spaces[pn_space].largest_acked_packet.unwrap();
         let packet_threshold = self.config.packet_threshold as u64;
         let mut size_of_lost_packets = 0u64;
@@ -1737,8 +1735,11 @@ impl Connection {
                 persistent_congestion_start = None;
             }
 
-            if info.time_sent <= lost_send_time || largest_acked_packet >= packet + packet_threshold
-            {
+            // Packets sent before now - loss_delay are deemed lost.
+            // However, we avoid this subtraction as it can panic and there's no
+            // saturating equivalent of this substraction operation with a Duration.
+            let packet_too_old = now.saturating_duration_since(info.time_sent) >= loss_delay;
+            if packet_too_old || largest_acked_packet >= packet + packet_threshold {
                 if Some(packet) == in_flight_mtu_probe {
                     // Lost MTU probes are not included in `lost_packets`, because they should not
                     // trigger a congestion control response


quinn-proto/src/connection/mod.rs
```diff
@@ -1737,8 +1735,10 @@ impl Connection {
                 persistent_congestion_start = None;
             }
 
-            if info.time_sent <= lost_send_time || largest_acked_packet >= packet + packet_threshold
```
thomaseizinger, line 1740: Disclaimer: I am not familiar with this part of `quinn`.

Couldn't you use `Instant::duration_since` instead and compare that to `loss_delay`?

matheus23: `Instant::duration_since` can panic, which I guess is why most code in quinn used to use `instant_saturating_sub`, I wanted to follow that.

Seems like there is `Instant::saturating_duration_since` now though, which would make `instant_saturating_sub` obsolete.

Quinn maintainers please let me know what you prefer. Should I migrate all `instant_saturating_sub` use to `Instant::saturating_duration_since`?

matheus23: But to answer your question most directly: This is essentially what I'm doing with this PR...

matheus23: FWIW, I've removed the `instant_saturating_sub` custom implementation and replaced all uses with `Instant::saturating_duration_since`.

matheus23:
I messed up the computation ever so slightly with a `>` instead of a `>=`. I fixed it and now tests shouldn't hang.

quinn-proto/src/connection/mod.rs
djc: Can you do that first in a separate commit?

thomaseizinger: > `Instant::duration_since` can panic

What I am suggesting is conceptually different (from the original code):

If you subtract a `Duration` from an `Instant` you get another `Instant` and that point in time needs to be representable internally which is platform dependent. From memory, Linux is pretty flexible here but Windows for example references the boot time I think. We've had bugs in other sans IO code that had the same issue (https://github.com/algesten/str0m/pull/600).

If you instead compute the difference between two `Instant`s and compare that to the time period, this problem cannot happen because you are computing a `Duration` and not another `Instant`.

Right now `duration_since` also saturates btw but it is probably safer to be explicit and use `saturating_duration_since`: https://doc.rust-lang.org/std/time/struct.Instant.html#method.duration_since

matheus23: > Right now duration_since also saturates btw

Huh, well, that's interesting. The documentation specifically mentions a panic, so I'm avoiding it.

> What I am suggesting is conceptually different (from the original code)

Yes, I'm aware :) Thank you for your suggestion. What I'm trying to say here is that the *changed* code does exactly what you're suggesting :)

quinn-proto/src/connection/mod.rs
```diff
@@ -1791,6 +1792,17 @@ impl Connection {
...
 
+            // Packets sent before this time are deemed lost.
+            // We avoid computing this value above, since it's possible for this to panic
+            // if the `loss_delay` value internally stores a bigger `Duration` than the
+            // `Duration` that's stored inside the `Instant`, because some platforms may
+            // implement the `Instant` with a counter relative to system or even process
+            // startup (Wasm is one such case).
+            // If we're at this point, then it must be possible to have instants that are
+            // longer ago than `loss_delay` (see the `packet_too_old` computation
+            // above).
```
djc: This invariant isn't at all obvious to me. Can you explain more how/why this is the case?

Couldn't we do `now.checked_duration_since(info.time_sent)` and key off the returning `Option` instead, to avoid the `unwrap()` here?

matheus23: > Couldn't we do `now.checked_duration_since(info.time_sent)` and key off the returning `Option` instead, to avoid the `unwrap()` here?

I thought so at first too, but then realized that `now - info.time_sent` != `now - loss_delay`.
So this would change the qlog events emitted.

---

> This invariant isn't at all obvious to me. Can you explain more how/why this is the case?

I checked this again, which made me notice this invariant isn't quite correct.

Should we rather skip qlog events (and ideally log a warning?) if `now.checked_sub(loss_delay)` ends up being `None`?

djc: > I checked this again, which made me notice this invariant isn't quite correct.
> 
> Should we rather skip qlog events (and ideally log a warning?) if `now.checked_sub(loss_delay)` ends up being `None`?

Yes, that sounds better.

matheus23: Found a way of refactoring `QlogSink::emit_packet_lost` in a way that avoids computing `lost_send_time` at all. So this avoids the `.unwrap()` altogether.

djc: Very nice! Can you re-order the commits so that more of the temporary state falls away? Maybe move the `emit_packet_lost()` commit before the second one, or just squash the latter two?

matheus23: Done that :v:

Ralith (approved):
Nice, thanks!

Issue #2439, opened 2025-11-06
Feature Request -  Expose server-side 0-RTT acceptance status (early_data_accepted) on Connection or Incoming


iadev09:
### Motivation
When using **QUIC** in authenticated or side-effectful services (e.g., **HTTP POST**, chat, or payment APIs), developers often need to know whether the first packet(s) may have been **0-RTT**, and therefore possibly subject to **replay attacks**.

Without this information, the only safe option is to **disable 0-RTT entirely**, or to **treat the first request specially** (for example, by enforcing idempotency or requiring a handshake confirmation before sensitive actions).

---

### Proposal
Expose a method on `Connection` (or `Incoming`, whichever fits best), such as:

```rust
pub fn early_data_accepted(&self) -> Option<bool>
```

---

### Example Usage
Here’s an example where early data is allowed **only for idempotent requests** (GET, HEAD, OPTIONS), while non-idempotent ones (POST, PUT, DELETE) are rejected with **HTTP 425 Too Early**:

```rust
if is_early_data_conn {
    // Reject only non-idempotent methods for security
    if req.method() != http::Method::GET &&
       req.method() != http::Method::HEAD &&
       req.method() != http::Method::OPTIONS
    {
        warn!("⚠️  0-RTT POST/Non-idempotent request rejected for security.");
        let mut stream = resolver.response_stream().await.unwrap();
        let _ = stream.send_response(
            http::Response::builder()
                .status(http::status::StatusCode::TOO_EARLY)
                .body(())
                .unwrap()
        ).await;
        continue; // Proceed to next stream
    }
}

djc:
Sounds okay -- are you able to submit a PR for this?

flub:
I think all the information is already available in a more usable way from `RecvStream::is_0rtt`.

PR #2440, opened 2025-11-07
Add server-side 0-RTT detection via early_data_accepted()

iadev09:
Expose a public API to detect whether 0-RTT (early data) was used on both client and server sides. This addresses issue #2439 by allowing servers to identify  possible replay-vulnerable requests.

Changes:
- Add Connection::early_data_accepted() public API
- Track 0-RTT packet reception on server side in quinn-proto
- Add test coverage in existing zero_rtt test

For clients: Some(true) if server accepted 0-RTT, Some(false) if rejected For servers: Some(true) if client sent 0-RTT data, Some(false) otherwise Returns None while handshake is in progress

This is useful for servers to implement replay attack mitigation, such as rejecting non-idempotent operations (POST/PUT/DELETE) that arrived via 0-RTT with HTTP 425 "Too Early" status.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -2689,6 +2689,10 @@ impl Connection {
                 ..
             } => {
                 self.process_payload(now, remote, number.unwrap(), packet)?;
+                // Track that 0-RTT data was received and accepted (server-side)
+                if self.side.is_server() {
+                    self.accepted_0rtt = true;
+                }
                 Ok(())
             }
             Header::VersionNegotiate { .. } => {
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -561,6 +561,23 @@ impl Connection {
             .handshake_data()
     }
 
+    /// Whether 0-RTT was accepted
+    ///
+    /// Returns `None` while the handshake is in progress. Once the handshake completes:
+    /// - For clients: `Some(true)` if the server accepted our 0-RTT data, `Some(false)` if rejected
+    /// - For servers: `Some(true)` if the client sent 0-RTT data that we accepted, `Some(false)` otherwise
+    ///
+    /// This is useful on the server side to detect replay attacks - data sent via 0-RTT
+    /// is vulnerable to replay and should be treated carefully (e.g., reject non-idempotent operations).
+    pub fn early_data_accepted(&self) -> Option<bool> {
+        let state = self.0.state.lock("early_data_accepted");
+        if state.inner.is_handshaking() {
+            return None;
+        }
+
+        Some(state.inner.accepted_0rtt())
+    }
+
     /// Cryptographic identity of the peer
     ///
     /// The dynamic type returned is determined by the configured
--- a/quinn/src/tests.rs
+++ b/quinn/src/tests.rs
@@ -334,6 +334,12 @@ async fn zero_rtt() {
             s.write_all(MSG1).await.expect("write");
             // The peer might close the connection before ACKing
             let _ = s.finish();
+
+            // Test early_data_accepted() on server side
+            let early_data = connection.early_data_accepted();
+            info!("server early_data_accepted: {:?}", early_data);
+            // First connection should show Some(false) - no 0-RTT
+            // Second connection should show Some(true) - 0-RTT received
         }
     });
 
@@ -378,6 +384,15 @@ async fn zero_rtt() {
     let msg = stream.read_to_end(usize::MAX).await.expect("read_to_end");
     assert_eq!(msg, MSG0);
     assert!(zero_rtt.await);
+
+    // Test early_data_accepted() on client side after handshake completes
+    let early_data = connection.early_data_accepted();
+    info!("client early_data_accepted: {:?}", early_data);
+    assert_eq!(
+        early_data,
+        Some(true),
+        "Client should see that 0-RTT was accepted by server"
+    );
 
     drop((stream, connection));
 


djc (approved):
This seems alright to me. Thanks!

iadev09:
Thanks for the quick review!

iadev09:
I have implemented  early_data_accepted() in my forked repo and  noticed  my implementation  is too aggressive 


```rust

// At connection start (after handshake completes)
let early_data_accepted = conn.early_data_accepted().unwrap_or(false);

// Store in connection context (frozen for entire connection lifetime)
let conn_ctx = Arc::new(ConnCtx {
    used_early_data: early_data_accepted,
    // ... other fields
});

// Later, for EVERY request on this connection
// when EarlyDataPolicy::AllowIdempotent 
if conn_ctx.used_early_data {
    match method {
        GET | HEAD | OPTIONS => accept(),
        POST | PUT | DELETE => reject_425_too_early(),
    }
}
```

**Problem:** This rejects non-idempotent methods for the ENTIRE connection lifetime if ANY request arrived via 0-RTT, even for requests that arrive after handshake completion (which are 1-RTT and safe).

This is overly conservative  for what RFC 8470 requires, but still  usefull for **server scoped** or aggrsive policies and also logging, monitoring and telemetry.

Also, I noticed ( not-proved)  in prod.  that  to disable Early-Data and propogate h3 precendence with Alt-svc, may result availibility problems for h3 aggresive browsers like  safari  (esp. mobile )

Also I couldn't come to a decision whether a connection that started with 0-RTT should still be considered replay-vulnerable after the handshake completes.

Once the handshake is done, the early data phase is over and 0-RTT keys are discarded — so the connection itself can no longer be replayed. 

However. **it might have been exposed  replayable data before handshake is completed ?** 


Therefore, as i understand,  we should check **per-request** (stream-level), not **per-connection**.


```rust
impl quinn::RecvStream {
    /// Check if this stream has been opened during 0-RTT.
    /// Any non-idempotent request should be considered dangerous
    /// because read data is subject to replay attacks.
    pub fn is_0rtt(&self) -> bool
}
```


### h3-quinn Doesn't Expose This

h3-quinn's `RecvStream` wraps Quinn's `RecvStream` but doesn't expose the `is_0rtt()` method:

```rust
// h3-quinn/src/lib.rs
pub struct RecvStream {
    stream: Option<quinn::RecvStream>,  // Private
    read_chunk_fut: ReadChunkFuture,
}
```

am I on right path and should i send a PR for h3-quinn as well ?

djc:
> am I on right path and should i send a PR for h3-quinn as well ?

It seems plausible to me that you want to check whether 0-RTT was used at the stream level, not the connection level. Does that mean you no longer need the API implemented in this PR?

iadev09:
I’m currently relying on **early_data_accepted()** for telemetry, logging, and especially debugging. Without this signal it becomes quite difficult to understand the real cause of failures: on the browser side everything simply shows up as a generic “CORS error” or “connection error”, while the underlying issue was actually early data being **disabled** or **rejected**.

Modern browsers make heavy use of 0-RTT, and I’m trying to better understand their behaviour by logging connections that start with early data. Right now the only way I can reliably reproduce these cases is by introducing client-side network changes (switching Wi-Fi, toggling VPN, etc.).

For the original replay-mitigation use-case I mentioned earlier, I eventually had to relax my policy: I no longer reject requests and I allow early data. In my environment, JavaScript running inside the browser container cannot explicitly open a fresh connection, so once a connection is established with early data the server has no other way to tell.

Because of this, early_data_accepted() is still a very lightweight and useful indicator for me, and I would prefer to keep this API.

I think the method name **early_data_accepted()** is fine as is.
To avoid misunderstandings, it may help to add a note in the docs clarifying that this reflects the **early-data** status only at handshake completion.
It's a connection-level, point-in-time signal, not a per-stream replay-safety indicator.
This should help users avoid over-applying the value for later requests on the same connection.

Stream-level helpers can be added  to h3-quinn, but the source-of-truth necessarily lives on the connection.

connection.early_data_accepted()
+ 
stream.opened_before_handshake_completed()

iadev09:
**Without proper early-data handling, real browsers cannot reliably use H3 with the server,**
because 0-RTT behaviour depends on each client’s session-ticket state, leading to failures that affect **only some users** and are extremely hard to reproduce or diagnose.

This is especially problematic for SPAs whose first request goes to a different subdomain and therefore triggers a new H3 connection, where early-request failures can block the application even though the HTML has already loaded. 

Browser falls back to H2 for HTML load, but  does not always infer protocol the same way for JS-opened connections.
And this mismatch between navigation and JavaScript-initiated connections is precisely why diagnosing these failures is so difficult.

And we also have no visibility into why the browser switched protocols, which makes these issues even harder to understand and debug.

So using this API for debugging is not something that can be dismissed — it is one of the few reliable ways to surface these otherwise invisible H3/0-RTT failures.

flub:
> Also I couldn't come to a decision whether a connection that started with 0-RTT should still be considered replay-vulnerable after the handshake completes.

No, once a handshake completes all data sent in all streams uses the normal connection's encryption and can not be replayed.  Even the streams that started as 0-RTT: any new data on those streams can no longer be replayed.

You are correct that `RecvStream::is_0rtt` already provides the necessary API for this in Quinn.  I don't think the API suggested in this PR is helpful.  If anything it only adds confusion.

iadev09:
> > Also I couldn't come to a decision whether a connection that started with 0-RTT should still be considered replay-vulnerable after the handshake completes.
> 
> No, once a handshake completes all data sent in all streams uses the normal connection's encryption and can not be replayed. Even the streams that started as 0-RTT: any new data on those streams can no longer be replayed.
> 
> You are correct that `RecvStream::is_0rtt` already provides the necessary API for this in Quinn. I don't think the API suggested in this PR is helpful. If anything it only adds confusion.

Thank you for clarifying a blind spot in my understanding. I agree that the terminology can be confusing. I don’t foresee a functional reason to expose this right now. On the other hand, since early data is an instantaneous state, this remains the most natural place to observe it. I’ll focus on stream-level handling for my main purpose.

PR #2441, opened 2025-11-08
Track handshake confirmation

Magnap:
Adds a new variant to `quinn_proto::Event` for when the TLS handshake is confirmed. For the server, this is irrelevant, as the TLS handshake is confirmed simultaneously with being completed. However, for the client the handshake is only confirmed one RTT after the handshake is completed. If mTLS is in use, being able to tell when the handshake is confirmed enables the client to wait for client authentication to complete. For this purpose, an async function is added on `quin::Connection` to await the handshake confirmation.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -2637,6 +2637,8 @@ impl Connection {
                     // Server-only
                     self.spaces[SpaceId::Data].pending.handshake_done = true;
                     self.discard_space(now, SpaceId::Handshake);
+                    self.events.push_back(Event::HandshakeConfirmed);
+                    trace!("handshake confirmed");
                 }
 
                 self.events.push_back(Event::Connected);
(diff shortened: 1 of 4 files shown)


flub (approved):
Nice to have this API!

djc:
I'm not opposed to adding public API for this but not sure adding an `Event` variant is the best way to go about it. Couldn't we tie this into observing `proto::Connection` changes in response to any of the existing events?

Magnap:
> Couldn't we tie this into observing `proto::Connection` changes in response to any of the existing events?

As far as I can tell, no. On the server side, handshake completion and confirmation are simultaneous, so that's fine, but on the client side, handshake confirmation happens 1 roundtrip after handshake completion, and there is no `Event` that is guaranteed to happen at this time.
We already have the ability to wait for the `StreamEvent`s from the server opening streams (if it sends on them), but if the client needs to "go first" and wants to wait until it's authenticated, it won't see any events past `Event::Connected`, which is one roundtrip too early.

Ralith:
See also a previous draft of this feature in https://github.com/quinn-rs/quinn/pull/1944 (which we reverted because the new event broke semver, but we can do that now).

PR #2443, opened 2025-11-11
quinn: simplify smol runtime setup

djc:


Ralith:
I think the original intent here was to support async-io independently of smol. This seems to drop that, though I'm not sure if the async-io feature was usable in isolation anyway...

djc:
It looked to me like it was mostly about sharing some code between async-std and smol, which had some things in common. Don't see much point in retaining async-io support at this point -- can always regeneralize this if someone comes up with a use case for it...

Ralith (approved):
Fair enough.

Issue #2424, opened 2025-10-25
Using quinn without a runtime feature enabled, gives the runtime error: "Error: no async runtime found"


paulcdejean:
Reproduction code:

```rust
use rustls_pki_types::pem::PemObject;
use std::sync::Arc;

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    // <Hardcoded example config>
    let pem_keyfile_path: &std::path::Path = std::path::Path::new("cert/server.key");
    let pem_certfile_path: &std::path::Path = std::path::Path::new("cert/server.cert");
    let listen_addr: std::net::SocketAddr = "127.0.0.1:4433".parse()?;
    // </Hardcoded example config>

    let tls_key: rustls_pki_types::PrivateKeyDer<'static> =
        match rustls_pki_types::PrivateKeyDer::from_pem_file(pem_keyfile_path) {
            Ok(k) => k,
            Err(e) => {
                panic!(
                    "Could not load keyfile {}, error {e}",
                    pem_keyfile_path.display()
                );
            }
        };

    let tls_cert_chain: Vec<rustls_pki_types::CertificateDer<'static>> =
        match rustls_pki_types::CertificateDer::pem_file_iter(pem_certfile_path) {
            Ok(iter) => iter.collect::<Result<Vec<_>, _>>()?,
            Err(e) => {
                panic!(
                    "Coul not load certfile {}, error {e}",
                    pem_certfile_path.display()
                );
            }
        };

    let rustls_config: rustls::server::ServerConfig = rustls::ServerConfig::builder()
        .with_no_client_auth()
        .with_single_cert(tls_cert_chain, tls_key)?;
    let quinn_crypto_config: quinn::crypto::rustls::QuicServerConfig =
        quinn::crypto::rustls::QuicServerConfig::try_from(rustls_config)?;
    let quinn_crypto_config_arc: Arc<quinn::crypto::rustls::QuicServerConfig> =
        Arc::new(quinn_crypto_config);
    let quinn_server_config: quinn::ServerConfig =
        quinn::ServerConfig::with_crypto(quinn_crypto_config_arc);
    dbg!(tokio::runtime::Handle::current());
    let quinn_server_endpoint: quinn::Endpoint = quinn::Endpoint::server(quinn_server_config, listen_addr)?;
    std::hint::black_box(quinn_server_endpoint);
    Ok(())
}
```

[...]

PR #2444, opened 2025-11-11
Check if a runtime is enabled at compile time

naveensrinivas282:
This PR references the issue #2424 and sets a new default that checks if a runtime has been enabled and then throws an error `no async runtime found` if it hasn't been. It also creates a new feature, 'allow-missing-runtime', which, when enabled, bypasses the check

### DIFF
--- a/quinn/Cargo.toml
+++ b/quinn/Cargo.toml
@@ -36,7 +36,7 @@ rustls-ring = ["dep:rustls", "ring", "proto/rustls-ring", "proto/ring"]
 # Outside wasm*-unknown-unknown targets, this enables `Endpoint::client` and `Endpoint::server` conveniences.
 ring = ["proto/ring"]
 runtime-tokio = ["tokio/time", "tokio/rt", "tokio/net"]
-runtime-smol = ["dep:async-io", "smol"]
+runtime-smol = ["dep:async-io", "dep:smol"]
 
 # Configure `tracing` to log events via `log` if no `tracing` subscriber exists.
 tracing-log = ["tracing/log", "proto/tracing-log", "udp/tracing-log"]


naveensrinivas282:
The failing test runs with no default features and the feature is meant to throw an error if no runtime or `allow-missing-runtime` is enabled
@djc Could you please guide me on this ?

quinn/src/lib.rs
```diff
@@ -57,6 +57,9 @@ pub(crate) use std::time::{Duration, Instant};
 #[cfg(wasm_browser)]
 pub(crate) use web_time::{Duration, Instant};
 
+#[cfg(all(not(feature = "runtime-tokio"), not(feature = "runtime-smol"), not(feature = "allow-missing-runtime")))]
```
Ralith: This would be more terse phrased as `not(any(...))`.

quinn/src/lib.rs
```diff
@@ -57,6 +57,9 @@ pub(crate) use std::time::{Duration, Instant};
...
+compile_error!("no async runtime found");
```
Ralith: The error should explain why and suggest solutions (i.e. enabling one of the runtime features)

quinn/Cargo.toml
```diff
@@ -38,6 +38,9 @@ ring = ["proto/ring"]
 runtime-tokio = ["tokio/time", "tokio/rt", "tokio/net"]
 runtime-smol = ["async-io", "smol"]
 
+# Allows users to run code without a runtime
+allow-missing-runtime = []
```
Ralith: This should be named something like `runtime-custom` to reflect its intended use.

quinn/Cargo.toml
```diff
@@ -38,6 +38,9 @@ ring = ["proto/ring"]
 runtime-tokio = ["tokio/time", "tokio/rt", "tokio/net"]
 runtime-smol = ["async-io", "smol"]
 
+# Allows users to run code without a runtime
```
Ralith: This comment should explain the intended use: using Quinn with an explicitly supplied custom `Runtime` implementation. Quinn is not very useful with no runtime at all.

quinn/src/lib.rs
Ralith: This should be written in the `runtime` module, adjacent to `default_runtime` for ease of reference, since the conditional compilation in each must be kept in sync.

Ralith:
> The failing test runs with no default features

You will need to modify the test scripts to enable the newly added feature in this case.

quinn/src/lib.rs
naveensrinivas282: I have fixed this in the latest commit

Issue #2422, opened 2025-10-23
proto: Leverage `BTreeMap::extract_if` in `RangeSet` for better performance


Ralith:
`BTreeMap::extract_if` was stabilized in Rust 1.91. It could be used to avoid repeatedly traversing the `BTreeMap` from the root in `insert` and perhaps elsewhere.

Perelyn-sama:
I'll make a PR if no one is working on this

Ralith:
That would be great, thanks! Note 1.91 isn't actually out for another few days, so it won't work in stable rust *quite* yet and there may be an MSRV concern here -- but this code is very stable so I'm not too worried about merge conflicts if we have to shuffle the change around.

Perelyn-sama:
1.91 is out now, I'll make a PR within the week. Also, i think it'll be more efficient to use extract_if in the remove method since other methods depend on it. what do you think?

Ralith:
I expect it's probably beneficial anywhere that it can be used without unreasonable contortions; `insert` was just an example.

DSharifi:
Hey @Ralith , I created a PR #2474 that implements this.

I see one issue with this PR, which is increasing the `msrv` to 1.91 if we want to use `BtreeMap::extract_if`. Is this something we want to do?

djc:
I definitely think it's too early to bump the MSRV to 1.91, but I think we discussed whether this could be an opt-in feature?

DSharifi:
Otherwise, can we just leave the PR open until `msrv` is bumpted to 1.91 or greater?

djc:
That's definitely an option.

PR #2445, opened 2025-11-13
refactor(quinn-proto): Use `ArrayRangeSet` in favor of `RangeSet`

matheus23:
This seems to improve performance by ~8-10% in local benchmarks for me.

### DIFF
--- a/quinn-proto/src/connection/send_buffer.rs
+++ b/quinn-proto/src/connection/send_buffer.rs
@@ -2,7 +2,7 @@ use std::{collections::VecDeque, ops::Range};
 
 use bytes::{Buf, Bytes};
 
-use crate::{VarInt, range_set::RangeSet};
+use crate::{VarInt, range_set::ArrayRangeSet};
 
 /// Buffer of outgoing retransmittable stream data
 #[derive(Default, Debug)]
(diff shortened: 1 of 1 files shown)


djc (approved):
Nice and simple, and seems plausible?

matheus23:
I would hold off on merging this too eagerly. For me this improves some benchmarks by quite a bit, but others also regress.

Mostly this tells me that overall CPU use seems to depend on the range set performance characteristics.

Ralith:
Another concern that bears some thought is the hazard of a CPU DoS attack. A peer can exert substantial control over the number of ranges in these fields, and the B tree structure is less sensitive to total size, though how much this matters depends on the exact accesses we're performing.

Ralith:
https://github.com/quinn-rs/quinn/issues/2422 will also improve some B-tree operations, which might tilt the balance.

gretchenfrage:
> Another concern that bears some thought is the hazard of a CPU DoS attack. A peer can exert substantial control over the number of ranges in these fields, and the B tree structure is less sensitive to total size, though how much this matters depends on the exact accesses we're performing.

Could consider an enum that tries to make it an array range set, but switches over to a btree range set if the number of ranges grows beyond some small constant?

Ralith:
That should work, yeah, at the cost of a little more complexity and boilerplate.

PR #2446, opened 2025-11-14
Bump MSRV to 1.80, remove `once_cell` and `lazy_static` deps

CathalMullan:
Based on the conversation in #2422, sounds like a MSRV bump is acceptable (though maybe not to 1.91 just yet).

Replaces `once_cell` and `lazy_static` with std equivalents.
Includes a few other MSRV related changes too.


quinn-proto/src/token.rs
```diff
@@ -247,9 +247,8 @@ impl Token {
     fn decode(key: &dyn HandshakeTokenKey, raw_token_bytes: &[u8]) -> Option<Self> {
         // Decrypt
 
-        // MSRV: split_at_checked requires 1.80.0
-        let nonce_slice_start = raw_token_bytes.len().checked_sub(size_of::<u128>())?;
-        let (sealed_token, nonce_bytes) = raw_token_bytes.split_at(nonce_slice_start);
+        let nonce_slice_start = raw_token_bytes.len() - size_of::<u128>();
+        let (sealed_token, nonce_bytes) = raw_token_bytes.split_at_checked(nonce_slice_start)?;
```
CathalMullan: Wasn't quite sure what the original comment here wanted. Assuming 2 *_checked calls would be redundant?

djc: I think we'll still want `checked_sub()` for the `nonce_slice_start()` calculation?

PR #2449, opened 2025-11-21
Add an option to retrieve TLS errors

matszczygiel:
Hello,

I'm using `quinn` create together with TLS Encrypted Client Hello (ECH) extension.
However, I encountered a problem where the client uses an invalid server ECH config. There is a mechanism described in the ECH RFC that the server can respond with the correct config, and the client can reconnect using it instead.

And theoretically, the `rustls` crate offers a way to retrieve the server ECH config from the TLS error, as discussed in https://github.com/rustls/rustls/issues/2572.

Sadly, when using the `quinn` crate, there is no way to retrieve the underlying TLS error and extract the ECH config.

This PR aims to provide a way to get the underlying `rustls` error and effectively allow for ECH client retry.

### DIFF
--- a/quinn-proto/src/transport_error.rs
+++ b/quinn-proto/src/transport_error.rs
@@ -1,4 +1,4 @@
-use std::fmt;
+use std::{fmt, sync::Arc};
 
 use bytes::{Buf, BufMut};
 
@@ -8,7 +8,12 @@ use crate::{
 };
 
 /// Transport-level errors occur when a peer violates the protocol specification
-#[derive(Debug, Clone, Eq, PartialEq)]
+///
+/// # Note
+///
+/// The `PartialEq` implementation for this type performs comparison on the `code` field only
+#[derive(Debug, Clone)]
+#[non_exhaustive]
 pub struct Error {
     /// Type of error
     pub code: Code,
@@ -16,7 +21,29 @@ pub struct Error {
     pub frame: Option<frame::FrameType>,
     /// Human-readable explanation of the reason
     pub reason: String,
+    /// An underlying crypto (e.g. TLS) layer error
+    pub crypto: Option<Arc<dyn std::error::Error + Send + Sync>>,
 }
+
+impl Error {
+    /// Construct an error with a code and a reason
+    pub fn new(code: Code, reason: String) -> Self {
+        Self {
+            code,
+            frame: None,
+            reason,
+            crypto: None,
+        }
+    }
+}
+
+impl PartialEq for Error {
+    fn eq(&self, other: &Self) -> bool {
+        self.code == other.code
+    }
+}
+
+impl Eq for Error {}
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
@@ -39,6 +66,7 @@ impl From<Code> for Error {
             code: x,
             frame: None,
             reason: "".to_string(),
+            crypto: None,
         }
     }
 }
@@ -79,6 +107,7 @@ macro_rules! errors {
                     code: Code::$name,
                     frame: None,
                     reason: reason.into(),
+                    crypto: None,
                 }
             }
             )*
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -2605,6 +2605,7 @@ impl Connection {
                                 code: TransportErrorCode::crypto(0x6d),
                                 frame: None,
                                 reason: "transport parameters missing".into(),
+                                crypto: None,
                             })?;
 
                     if self.has_0rtt() {
@@ -2678,6 +2679,7 @@ impl Connection {
                                 code: TransportErrorCode::crypto(0x6d),
                                 frame: None,
                                 reason: "transport parameters missing".into(),
+                                crypto: None,
                             })?;
                     self.handle_peer_params(params)?;
                     self.issue_first_cids(now);
--- a/quinn-proto/src/crypto/rustls.rs
+++ b/quinn-proto/src/crypto/rustls.rs
@@ -110,6 +110,7 @@ impl crypto::Session for TlsSession {
                     code: TransportErrorCode::crypto(alert.into()),
                     frame: None,
                     reason: e.to_string(),
+                    crypto: Some(Arc::new(e)),
                 }
             } else {
                 TransportError::PROTOCOL_VIOLATION(format!("TLS error: {e}"))
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -1074,11 +1074,10 @@ impl State {
                     self.close(error_code, reason, shared);
                 }
                 Poll::Ready(None) => {
-                    return Err(ConnectionError::TransportError(proto::TransportError {
-                        code: proto::TransportErrorCode::INTERNAL_ERROR,
-                        frame: None,
-                        reason: "endpoint driver future was dropped".to_string(),
-                    }));
+                    return Err(ConnectionError::TransportError(proto::TransportError::new(
+                        proto::TransportErrorCode::INTERNAL_ERROR,
+                        "endpoint driver future was dropped".to_string(),
+                    )));
                 }
                 Poll::Pending => {
                     return Ok(());


quinn-proto/src/transport_error.rs
```diff
@@ -8,16 +8,20 @@ use crate::{
...
     pub code: Code,
     /// Frame type that triggered the error
     pub frame: Option<frame::FrameType>,
     /// Human-readable explanation of the reason
     pub reason: String,
+    /// An underlying TLS layer error
+    pub tls_error: Option<rustls::Error>,
 }
 
+impl Eq for Error {}
```
djc, line 46: Presumably `rustls::Error` doesn't implement `Eq` for a reason, so this feels potentially wrong... Maybe our `Eq` and `PartialEq` implementations should completely ignore the `tls_error` field instead?

quinn-proto/src/transport_error.rs
```diff
@@ -8,16 +8,20 @@ use crate::{
...
+#[derive(Debug, Clone, PartialEq)]
 pub struct Error {
     /// Type of error
     pub code: Code,
     /// Frame type that triggered the error
     pub frame: Option<frame::FrameType>,
     /// Human-readable explanation of the reason
     pub reason: String,
+    /// An underlying TLS layer error
+    pub tls_error: Option<rustls::Error>,
```
djc: This is technically semver-breaking, which is fine for us since `main` is targeting 0.12 anyway, but will mean you won't be able to use this on 0.11.x.

We should probably make this type `#[non_exhaustive]`, do you want to do that here?

Not sure we need the `_error` suffix on the field, I guess `tls` might be sufficient given that it is in a type called `Error` and the field's type is `Option<rustls::Error>`?

matszczygiel:
Oh, I didn't notice that `rustls` is an optional dependency. Then perhaps we should store it under `#[cfg(any(rustls-ring, rustls-aws-lc-rs, rustls-aws-lc-rs-fips))]`?
Alternatively, we can use `io::Error` as an error representation, as downstream apps can always cast it to a concrete error type.

I'm not sure which option would be better in this case.

quinn-proto/src/transport_error.rs
matszczygiel: Sure, makes sense. Let me fix it

quinn-proto/src/transport_error.rs
matszczygiel: Yes, I can mark it in this PR

matszczygiel: @djc regarding `#[non_exhaustive]` so there is one case where `quinn` crate constructs the `Error` type directly https://github.com/quinn-rs/quinn/blob/main/quinn/src/connection.rs#L1077
And since it is a downstream crate of `quinn-proto`, it fails to compile.

quinn-proto/src/transport_error.rs
matszczygiel: Done

quinn-proto/src/transport_error.rs
matszczygiel: I've renamed the field to `tls`

djc: > @djc regarding `#[non_exhaustive]` so there is one case where `quinn` crate constructs the `Error` type directly https://github.com/quinn-rs/quinn/blob/main/quinn/src/connection.rs#L1077 And since it is a downstream crate of `quinn-proto`, it fails to compile.

Add a constructor that takes a code and a reason?

djc:
> Oh, I didn't notice that `rustls` is an optional dependency. Then perhaps we should store it under `#[cfg(any(rustls-ring, rustls-aws-lc-rs, rustls-aws-lc-rs-fips))]`? Alternatively, we can use `io::Error` as an error representation, as downstream apps can always cast it to a concrete error type.

Definitely don't think `cfg` guards are the right solution here, since it becomes confusing for non-rustls `crypto::Session` implementers.

quinn-proto/src/transport_error.rs
matszczygiel: There are already constructors here https://github.com/quinn-rs/quinn/blob/main/quinn-proto/src/transport_error.rs#L77. Though marked `pub(crate)`.
Do you think it makes sense to make them `pub`?

matszczygiel:
> > Oh, I didn't notice that `rustls` is an optional dependency. Then perhaps we should store it under `#[cfg(any(rustls-ring, rustls-aws-lc-rs, rustls-aws-lc-rs-fips))]`? Alternatively, we can use `io::Error` as an error representation, as downstream apps can always cast it to a concrete error type.
> 
> Definitely don't think `cfg` guards are the right solution here, since it becomes confusing for non-rustls `crypto::Session` implementers.

Yeah, I was thinking of using `Option<Arc<io::Error>>` because of the `Clone` requirement, which `io::Error` does not provide, but it feels a bit off. Maybe `Option<Box<dyn (Any + Clone)>>` would work in this case?

matszczygiel:
I've ended up using `Option<Arc<dyn std::error::Error + Send + Sync>>` to not tie it to any particular error type. And still provide the downcasting method https://doc.rust-lang.org/stable/std/error/trait.Error.html#method.downcast_ref

quinn-proto/src/transport_error.rs
```diff
@@ -8,16 +8,26 @@ use crate::{
...
+    pub tls: Option<Arc<dyn std::error::Error + Send + Sync>>,
 }
 
+impl PartialEq for Error {
+    fn eq(&self, other: &Self) -> bool {
+        self.code == other.code && self.frame == other.frame && self.reason == other.reason
+    }
+}
+
+impl Eq for Error {}
```
Ralith, line 46: Maybe we should ditch this impl outright? Users should probably only be performing comparisons on `code` anyway.

quinn-proto/src/transport_error.rs
```diff
@@ -8,16 +8,26 @@ use crate::{
...
+#[derive(Debug, Clone)]
 pub struct Error {
     /// Type of error
     pub code: Code,
     /// Frame type that triggered the error
     pub frame: Option<frame::FrameType>,
     /// Human-readable explanation of the reason
     pub reason: String,
+    /// An underlying TLS layer error
+    pub tls: Option<Arc<dyn std::error::Error + Send + Sync>>,
```
Ralith: Should we name/document this in a more generic way? While QUIC requires TLS, Quinn technically doesn't, and we have some non-TLS downstream users. We refer to the "crypto" layer elsewhere...

quinn-proto/src/transport_error.rs
Ralith: No, we should only publicly expose a single constructor that takes a code and a reason.

quinn-proto/src/transport_error.rs
matszczygiel: That would mean dropping `PartialEq, Eq` from the `ConnectionError` here https://github.com/quinn-rs/quinn/blob/main/quinn-proto/src/connection/mod.rs#L3917, or converting it to an ugly manual impl. Perhaps we should keep the comparison here, but only for the `code` field?

quinn-proto/src/transport_error.rs
matszczygiel: I've renamed the field to `crypto` and updated the doc comment.

quinn-proto/src/transport_error.rs
matszczygiel: I've added `fn new(code, reason)` constructor.

matszczygiel:
I've updated the `PartialEq` manual implementation so that it compares only the `code` field. I hope this covers all the use cases.

quinn-proto/src/transport_error.rs
djc: This comment seems to be referencing `Eq` but not `PartialEq`. I think it'd be okay to drop `Eq` from `ConnectionError`, but we should probably try to preserve `PartialEq`.

matszczygiel: Oh, I assumed differently, as in that case, there's little difference between `Eq` and `PartialEq`. @Ralith can you clarify what you meant?

>  Users should probably only be performing comparisons on code anyway.

Regarding this, I rewrote the `PartialEq` comparison to include only `code`. Is that OK?

Ralith: I intended to refer to both, but I agree that losing comparison for `ConnectionError`s sucks. I think your compromise makes sense.

quinn-proto/src/transport_error.rs
```diff
@@ -8,16 +8,27 @@ use crate::{
 };
 
 /// Transport-level errors occur when a peer violates the protocol specification
-#[derive(Debug, Clone, Eq, PartialEq)]
+#[derive(Debug, Clone)]
+#[non_exhaustive]
 pub struct Error {
     /// Type of error
     pub code: Code,
```
Ralith, line 19: We should document that this is the only field that `PartialEq` considers.

Ralith (approved):

quinn-proto/src/transport_error.rs
matszczygiel: I've added the note

quinn-proto/src/transport_error.rs
```diff
@@ -39,6 +54,19 @@ impl From<Code> for Error {
             code: x,
             frame: None,
             reason: "".to_string(),
+            crypto: None,
+        }
+    }
+}
+
+impl Error {
```
djc: Nit: please keep the inherent `impl` before any trait `impl`s.

matszczygiel: Done

djc:
Thanks!

PR #2450, opened 2025-11-22
Throw compile time error for missing runtime

naveensrinivas282:
This PR allows to
- throw a compile-time error if none of `runtime-tokio`, `runtime-smol`, `custom-runtime` are enabled
- create the custom-runtime feature which allows user to declare their own custom runtime

### DIFF
--- a/quinn/src/runtime/mod.rs
+++ b/quinn/src/runtime/mod.rs
@@ -1,10 +1,11 @@
+#[cfg(any(feature = "runtime-tokio", feature = "runtime-smol"))]
+use std::sync::Arc;
 use std::{
     fmt::{self, Debug},
     future::Future,
     io::{self, IoSliceMut},
     net::SocketAddr,
     pin::Pin,
-    sync::Arc,
     task::{Context, Poll},
 };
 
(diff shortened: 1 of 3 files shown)


quinn/src/runtime/mod.rs
```diff
@@ -12,6 +12,9 @@ use udp::{RecvMeta, Transmit};
 
 use crate::Instant;
 
+#[cfg(not(any(feature = "runtime-tokio", feature = "runtime-smol", feature = "custom-runtime")))]
+compile_error!("no async runtime found. Please enable a runtimse such as runtime-tokio, runtime-smol or custom-runtime");
```
Ralith: typo: "runtimse". Let's also word this as "a runtime feature" to provide a clearer breadcrumb to users who might not know what we're talking about.

Also please be consistent about capitalization and punctuation: if the text contains multiple sentences, then the first letter should be capitalized and there should be a trailing period.

djc:
Hi @naveen28204280 -- sorry, I didn't like the direction of this PR much so implemented an alternative in #2451. That means we no longer need this PR -- we still appreciate your efforts, though!

PR #2455, opened 2025-11-29
Fix quinn-udp on DragonFly BSD

mneumann:
- There is no libc::IP_RECVTOST defined for DragonFly. We do support it in the kernel, but it's not yet defined in libc.

- There is no recvmmsg(2) support in DragonFly kernel.

With these changes, "cargo build" for quinn succeeds on DragonFly 6.5.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -113,7 +113,12 @@ impl UdpSocketState {
 
         // mac and ios do not support IP_RECVTOS on dual-stack sockets :(
         // older macos versions also don't have the flag and will error out if we don't ignore it
-        #[cfg(not(any(target_os = "openbsd", target_os = "netbsd", solarish)))]
+        #[cfg(not(any(
+            target_os = "openbsd",
+            target_os = "netbsd",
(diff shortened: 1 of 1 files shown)


djc:
Thanks!

Ralith:
> There is no libc::IP_RECVTOST defined for DragonFly. We do support it in the kernel, but it's not yet defined in libc.

We could just define the constant ourselves, though this might be a liability if it's very new and quinn is likely to run on older kernels. We've done similar in the past for stuff that's mature but not yet exposed in the libc crate.

mneumann:
@Ralith IP_RECVTOS was added to DragonFly pretty much 5 years ago (4th December 2020). See commit https://github.com/DragonFlyBSD/DragonFlyBSD/commit/544d23f43f4a40b396d411d5ba2fbe3da17a10d2. So we can expect it to be present. It's just that I ported rust libc to DragonFly much much earlier :). I'll quickly do a follow up PR. I think it's easier to define the value in quinn-rs vs doing it in rust libc (it takes longer to settle).

PR #2460, opened 2025-12-02
draft: switch to n0-qlog with support for latest drafts & multipath

Frando:
Switches from `qlog`  to `n0-qlog`, which is a fork of `qlog` with these two PRs applied:

* https://github.com/n0-computer/n0-qlog/pull/1 (i.e. https://github.com/cloudflare/quiche/pull/2168): update to the latest qlog and qlog-quic drafts 
* https://github.com/n0-computer/n0-qlog/pull/2: add support for quic multipath

### DIFF
--- a/quinn-proto/src/connection/qlog.rs
+++ b/quinn-proto/src/connection/qlog.rs
@@ -271,13 +271,26 @@ impl QlogSentPacket {
     /// This is a no-op if the `qlog` feature is not enabled.
     pub(crate) fn frame_path_ack(
         &mut self,
-        _path_id: PathId,
-        _delay: u64,
-        _ranges: &ArrayRangeSet,
-        _ecn: Option<&EcnCounts>,
+        path_id: PathId,
+        delay: u64,
+        ranges: &ArrayRangeSet,
+        ecn: Option<&EcnCounts>,
     ) {
         #[cfg(feature = "qlog")]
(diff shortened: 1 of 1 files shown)


djc:
Any particular reason folks keep opening misdirected PRs today?

mxinden:
I assume because they are maintaining a fork:

https://github.com/n0-computer/quinn

djc:
> I assume because they are maintaining a fork:
> 
> 
> 
> https://github.com/n0-computer/quinn

Yes, but they've been doing that for years and it happened twice in the same week that they opened a PR here...

thomaseizinger:
> Yes, but they've been doing that for years and it happened twice in the same week that they opened a PR here...

I blame GitHub's UX. It happened to me a couple of times with some forks. When you push a branch to your fork and click on the link returned from git's output, GitHub's default is to open a PR against upstream. You need to then manually select your own fork from a small drop-down instead.

As far as I know, there is no way in GitHub to change that default which is annoying.

Frando:
Sorry for that.

GitHub's UI defaults to a fork's base repository when opening a PR, and when accidentally clicking OK before changing that, GitHub gives no chance to undo this. That it happened twice in the past week is likely caused by more members of the n0 team submitted PRs than in the months before.

I will post a reminder to avoid this nuisance better in the future.

PR #2462, opened 2025-12-03
Fix dropping oversized unblocks datagram senders

### LINKED ISSUE #2456: call to send_datagram_wait can block indefinitely
I came across an edgecase where `send_datagram_wait` blocks indefinitely during a benchmark.

It happens when the path MTU changes and [oversized packets are dropped](https://github.com/quinn-rs/quinn/blob/8bf6825da7876dc233434baaea546d3addc4b0ed/quinn-proto/src/connection/datagrams.rs#L147).

Dropping packets in the outgoing queue should unblock callers of `send_datagram_wait` that are waiting for `Event::DatagramsUnblocked`.

If no datagram can be sent e.g. because all packets in the outgoing queue were too large for `max_payload` then the callers will wait forever.

uniquefine:
When dropping oversized datagrams after an MTU change, notify
callers of `send_datagram_wait` that are waiting for
space in the send buffer.

Fixes #2456

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1821,7 +1821,12 @@ impl Connection {
                     .congestion
                     .on_mtu_update(self.path.mtud.current_mtu());
                 if let Some(max_datagram_size) = self.datagrams().max_size() {
-                    self.datagrams.drop_oversized(max_datagram_size);
+                    if self.datagrams.drop_oversized(max_datagram_size)
+                        && self.datagrams.send_blocked
+                    {
+                        self.datagrams.send_blocked = false;
+                        self.events.push_back(Event::DatagramsUnblocked);
+                    }
                 }
             }
 
--- a/quinn-proto/src/tests/mod.rs
+++ b/quinn-proto/src/tests/mod.rs
@@ -3379,6 +3379,87 @@ fn voluntary_ack_with_large_datagrams() {
     );
 }
 
+/// Verify that dropping oversized datagrams will trigger a DatagramsUnblocked event.
+#[test]
+fn oversized_datagrams_trigger_unblock() {
+    let _guard = subscribe();
+    let mut pair = Pair::default();
+    // Start the connection with a large MTU.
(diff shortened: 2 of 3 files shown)


quinn-proto/src/connection/mod.rs
```diff
@@ -1824,7 +1824,10 @@ impl Connection {
                     .congestion
                     .on_mtu_update(self.path.mtud.current_mtu());
                 if let Some(max_datagram_size) = self.datagrams().max_size() {
-                    self.datagrams.drop_oversized(max_datagram_size);
+                    if self.datagrams.drop_oversized(max_datagram_size) {
+                        self.datagrams.send_blocked = false;
+                        self.events.push_back(Event::DatagramsUnblocked);
```
Ralith, line 1828: Should we also gate this on `send_blocked` having previously been true?

uniquefine: Yes, good point. Updated.

quinn-proto/src/tests/mod.rs
```diff
@@ -3379,6 +3379,53 @@ fn voluntary_ack_with_large_datagrams() {
...
+                break;
+            }
+            Err(e) => panic!("unexpected error: {e}"),
+        }
+    }
+    // Set the MTU to a smaller value so the queued datagrams cannot be sent.
+    // This should trigger a black hole detection and cause the queued datagrams to be dropped.
+    pair.mtu = 1200;
+    pair.drive();
+    // We expect all datagrams to be dropped (they all exceed the new MTU).
```
Ralith: How do we know they were dropped due to the MTU change, and not merely sent (and lost in flight) before the MTU change was detected?

uniquefine: True, we couldn't know.
Changed the test to keep the send buffer filled after each step.

uniquefine:
@Ralith Can you please restart the android tests, I think that was a flake.

uniquefine:
@Ralith happy new year :partying_face: when you have time, could you please take another look at this?

uniquefine:
Done

PR #2463, opened 2025-12-04
feat(quinn-udp): add opt-in Apple fast UDP datapath

larseggert:
Add runtime dispatch so callers explicitly opt in to the `sendmsg_x`/`recvmsg_x` fast datapath instead of using it unconditionally.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -72,6 +72,13 @@ pub struct UdpSocketState {
     /// In particular, we do not use IP_TOS cmsg_type in this case,
     /// which is not supported on Linux <3.13 and results in not sending the UDP packet at all.
     sendmsg_einval: AtomicBool,
+
+    /// Whether to use Apple's fast `sendmsg_x`/`recvmsg_x` APIs.
+    ///
+    /// These private APIs provide better performance but may not be available on all
+    /// Apple OS versions. Callers must verify availability before enabling.
+    #[cfg(apple_fast)]
+    apple_fast_path: AtomicBool,
 }
 
 impl UdpSocketState {
@@ -191,6 +198,8 @@ impl UdpSocketState {
             gro_segments: gro::gro_segments(),
             may_fragment,
             sendmsg_einval: AtomicBool::new(false),
+            #[cfg(apple_fast)]
+            apple_fast_path: AtomicBool::new(false),
         })
     }
 
@@ -225,13 +234,50 @@ impl UdpSocketState {
         send(self, socket.0, transmit)
     }
 
+    #[cfg(not(any(
+        apple,
+        target_os = "openbsd",
+        target_os = "netbsd",
+        target_os = "dragonfly",
+        solarish
+    )))]
     pub fn recv(
         &self,
         socket: UdpSockRef<'_>,
         bufs: &mut [IoSliceMut<'_>],
         meta: &mut [RecvMeta],
     ) -> io::Result<usize> {
-        recv(socket.0, bufs, meta)
+        recv_via_recvmmsg(socket.0, bufs, meta)
+    }
+
+    #[cfg(apple_fast)]
+    pub fn recv(
+        &self,
+        socket: UdpSockRef<'_>,
+        bufs: &mut [IoSliceMut<'_>],
+        meta: &mut [RecvMeta],
+    ) -> io::Result<usize> {
+        if self.is_apple_fast_path_enabled() {
+            recv_via_recvmsg_x(socket.0, bufs, meta)
+        } else {
+            recv_single(socket.0, bufs, meta)
+        }
+    }
+
+    #[cfg(any(
+        target_os = "openbsd",
+        target_os = "netbsd",
+        target_os = "dragonfly",
+        solarish,
+        apple_slow
+    ))]
+    pub fn recv(
+        &self,
+        socket: UdpSockRef<'_>,
+        bufs: &mut [IoSliceMut<'_>],
+        meta: &mut [RecvMeta],
+    ) -> io::Result<usize> {
+        recv_single(socket.0, bufs, meta)
     }
 
     /// The maximum amount of segments which can be transmitted if a platform
@@ -294,6 +340,27 @@ impl UdpSocketState {
     #[cfg(not(any(apple, target_os = "openbsd", target_os = "netbsd")))]
     fn set_sendmsg_einval(&self) {
         self.sendmsg_einval.store(true, Ordering::Relaxed)
+    }
+
+    /// Enables Apple's fast UDP datapath using private `sendmsg_x`/`recvmsg_x` APIs.
+    /// Once enabled, this also updates [`max_gso_segments`] to allow batched sends.
+    ///
+    /// # Safety
+    ///
+    /// These APIs may crash on unsupported OS versions, so callers must verify
+    /// availability before enabling.
+    ///
+    /// [`max_gso_segments`]: Self::max_gso_segments
+    #[cfg(apple_fast)]
+    pub unsafe fn set_apple_fast_path(&self) {
+        self.apple_fast_path.store(true, Ordering::Relaxed);
+        self.max_gso_segments.store(BATCH_SIZE, Ordering::Relaxed);
+    }
+
+    /// Returns whether Apple's fast UDP datapath is enabled for this socket.
+    #[cfg(apple_fast)]
+    pub fn is_apple_fast_path_enabled(&self) -> bool {
+        self.apple_fast_path.load(Ordering::Relaxed)
     }
 }
 
@@ -384,6 +451,20 @@ fn send(
 
 #[cfg(apple_fast)]
 fn send(state: &UdpSocketState, io: SockRef<'_>, transmit: &Transmit<'_>) -> io::Result<()> {
+    if state.is_apple_fast_path_enabled() {
+        send_via_sendmsg_x(state, io, transmit)
+    } else {
+        send_single(state, io, transmit)
+    }
+}
+
+/// Send using the fast `sendmsg_x` API.
+#[cfg(apple_fast)]
+fn send_via_sendmsg_x(
+    state: &UdpSocketState,
+    io: SockRef<'_>,
+    transmit: &Transmit<'_>,
+) -> io::Result<()> {
     let mut hdrs = unsafe { mem::zeroed::<[msghdr_x; BATCH_SIZE]>() };
     let mut iovs = unsafe { mem::zeroed::<[libc::iovec; BATCH_SIZE]>() };
     let mut ctrls = [cmsg::Aligned([0u8; CMSG_LEN]); BATCH_SIZE];
@@ -468,6 +549,7 @@ fn send_single(state: &UdpSocketState, io: SockRef<'_>, transmit: &Transmit<'_>)
     }
 }
 
+/// Receive using the batched `recvmmsg` syscall.
 #[cfg(not(any(
     apple,
     target_os = "openbsd",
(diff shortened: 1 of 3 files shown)


larseggert:
@mxinden thinking about this some more, I wonder if we need to pass in the sockets from Gecko (due to Rust code limitations around socket creation), or if we should even do this probe entirely in Gecko and simply runtime-configure which Apple datapath `quinn-udp` should use?

djc:
> (due to Rust code limitations around socket creation)

What do you mean?

djc:
I think each of this should be a separate commit:

> * Extract `decode_cmsg` helper to reduce duplication in recv decoding
> * Make `decode_recv` generic over `MsgHdr` trait

It probably also makes sense to split some of this out:

> * Split fast path (`send_via_sendmsg_x`/`recv_via_recvmsg_x`) from slow path (`send_single`/`recv_single`) for runtime dispatch

larseggert:
> > (due to Rust code limitations around socket creation)
> 
> 
> 
> What do you mean?

Gecko is sandboxing all Rust code (and other third-party code) so it cannot open sockets.

larseggert:
@djc I'll do some refactoring on this next week per your suggestions.

larseggert:
> Gecko is sandboxing all Rust code (and other third-party code) so it cannot open sockets.

FWIW, I just build Firefox with this PR and it does run (and detect the fast Apple datapath after the probe), so something has changed related to the sandbox 🎉

mxinden:
Concerning the usage of quinn-udp in Firefox only:

> > Gecko is sandboxing all Rust code (and other third-party code) so it cannot open sockets.
> 
> FWIW, I just build Firefox with this PR and it does run (and detect the fast Apple datapath after the probe), so something has changed related to the sandbox 🎉

Referencing a discussion out-of-band. The "sandboxing" is happening at compile time.

https://github.com/mozilla-firefox/firefox/blob/43c97a062976252bce5968555060136810099ac2/python/mozbuild/mozbuild/action/check_binary.py#L209-L278

There is also the runtime sandboxing of the process. Though as far as I am aware, that isn't enabled by default.

https://github.com/mozilla-firefox/firefox/blob/main/security/sandbox/linux/SandboxFilter.cpp

larseggert:
ACK. I will revamp this once back from vacation.

Ralith:
> These APIs may crash on unsupported OS versions

What sort of crash, specifically? I hope we're not invoking UB or risking a collision with application signal handlers or something.

larseggert:
The concern is that Apple would eliminate or change the signature of the syscalls.

Ralith:
That would make a call UB, right? Can we rely on the exit code we get from the subprocess in that case?

larseggert:
I'll investigate if we can move the probe function into neqo/Gecko, and instead have some sort of flag in quinn for whether the fast datapath should be used. Marking this PR as draft until then.

larseggert:
OK, I revamped this and moved the entire logic for deciding whether the fast datapath should be used out of `quinn-udp`.

quinn-udp/src/unix.rs
```diff
@@ -1087,3 +1197,36 @@ mod gro {
...
     }
 }
+
+/// Fast path availability for Apple's private `sendmsg_x`/`recvmsg_x` APIs.
+#[cfg(apple_fast)]
+mod probe {
+    use std::sync::OnceLock;
+
+    /// Fast path availability flag. Defaults to `false` (disabled).
+    static FAST_PATH_AVAILABLE: OnceLock<bool> = OnceLock::new();
```
mxinden: Whereever possible, I think `quinn-udp` should not use global variables. Say there are two isolated `quinn-udp` users in one binary, one might affect the other by setting this to `true`.

Instead of a global variable, why not make this a property on `UdpSocketState` like we do for all other settings? Users of `quinn-udp` can first do the probing, then construct all future sockets with the option enabled.

larseggert: Yeah, that makes way more sense.

larseggert:
@mxinden please check the current revision.

quinn-udp/src/lib.rs
```diff
@@ -259,6 +259,64 @@ impl EcnCodepoint {
...
+/// extern "C" void quinn_udp_socket_state_enable_apple_fast_path(void* state);
+///
+/// // After probing for fast datapath availability:
+/// if (apple_fast_datapath_available) {
+///     quinn_udp_socket_state_enable_apple_fast_path(socket_state);
+/// }
+/// ```
+#[cfg(apple_fast)]
+#[no_mangle]
+pub unsafe extern "C" fn quinn_udp_socket_state_enable_apple_fast_path(
```
mxinden: Does the quinn project define any C FFIs? Do I understand correctly that one can define these C APIs anywhere?

If both are true, I suggest not introducing the very first C FFI to quinn here. Instead, in the case of Firefox, I suggest defining this function in `neqo_glue`.

quinn-udp/src/unix.rs
```diff
@@ -231,7 +240,7 @@ impl UdpSocketState {
         bufs: &mut [IoSliceMut<'_>],
         meta: &mut [RecvMeta],
     ) -> io::Result<usize> {
-        recv(socket.0, bufs, meta)
+        recv(self, socket.0, bufs, meta)
```
mxinden: Given that `UdpSocketState::recv` now simply delegates to the free function `recv`, including `self`, why not move the latter into the `UdpSocketState` `impl` block, replacing the current `UdpSocketState::recv`?

quinn-udp/src/lib.rs
larseggert: I wasn't sure if one can define a C FFI for something in a crate outside of it. If yes, doing this in Gecko would be much nicer.

larseggert: I think this should work. Will remove it.

larseggert:
Is this still waiting for anything before it can be merged? (And once it is, a new release at some point soon would be nice, so I can vendor it in.)

mxinden:
> Is this still waiting for anything before it can be merged?

Ready to merge from my end. Any objections by other maintainers?

> (And once it is, a new release at some point soon would be nice, so I can vendor it in.)

Why are you blocked on vendoring it in for testing in Firefox @larseggert? Can you not simply point to the Git reference? I assume we won't merge to Firefox Nightly any time soon anyways?

larseggert:
I did test it, I want to land it

quinn-udp/src/unix.rs
```diff
@@ -295,6 +379,25 @@ impl UdpSocketState {
...
+
+    /// Enables Apple's fast UDP datapath using private `sendmsg_x`/`recvmsg_x` APIs.
+    ///
+    /// These APIs may crash on unsupported OS versions, so callers must verify
+    /// availability before enabling. Once enabled, this also updates [`max_gso_segments`]
+    /// to allow batched sends.
+    ///
+    /// [`max_gso_segments`]: Self::max_gso_segments
+    #[cfg(apple_fast)]
+    pub fn enable_apple_fast_path(&self) {
```
Ralith: `unsafe fn`? The comment sounds like it's describing a safety invariant, unless we know for sure "crash" means something like "graceful abort".

larseggert: Just so I understand, you would like me to mark this function as `unsafe`?

quinn-udp/src/unix.rs
```diff
@@ -225,13 +234,88 @@ impl UdpSocketState {
...
 
+    #[cfg(not(any(
+        apple,
+        target_os = "openbsd",
+        target_os = "netbsd",
+        target_os = "dragonfly",
+        solarish
+    )))]
+    pub fn recv(
+        &self,
```
djc, line 245: This doesn't actually seem to use the `self` argument, so why are we moving it into the `UdpSocketState` impl? I'd prefer to keep it outside as before, and have a short wrapper/stub that forwards, so that all the platform-conditional `recv()` methods are (a) together (b) similarly short (hopefully making it easy/easier to determine which gets called when).

quinn-udp/src/unix.rs
mxinden:
```suggestion
    pub fn set_apple_fast_path(&self, enable: bool) {
```

That would be consistent with Window's `set_gro`. If you don't mind, I think we will appreciate the consistency in the future.

https://github.com/quinn-rs/quinn/blob/e5b30f51341c323ca3fbaa4006912b105efc1018/quinn-udp/src/windows.rs#L168

larseggert:
I think I made all the requested changes, please re-check?

djc:
As for releasing this, not sure whether we can release this from `main` or whether it would make more sense to backport it to 0.11.x before release. @mxinden can you look into what the situation is with quinn-udp on the `main` branch?

larseggert:
Ping?

mxinden:
`quinn-udp` on `main` is currently at `v0.6.0`. Given the minor version bump (`v0.5` -> `v0.6`) we can and have been making breaking changes.

https://github.com/quinn-rs/quinn/blob/43a1789995be72a44f320f9cd9e636d3c9963f8e/quinn-udp/Cargo.toml#L3

Latest release of `quinn-udp` is `v0.5.6`:

https://github.com/quinn-rs/quinn/releases/tag/quinn-udp-0.5.14



> As for releasing this, not sure whether we can release this from `main` or whether it would make more sense to backport it to 0.11.x before release. @mxinden can you look into what the situation is with quinn-udp on the `main` branch?

@djc why can we not release `quinn-udp` `v0.6.0` off of `main`? Do you want to be able to continue merging breaking changes into the `v0.6.0` release? Couldn't we instead merge those into a future `v0.7.0`?

djc:
> @djc why can we not release `quinn-udp` `v0.6.0` off of `main`? Do you want to be able to continue merging breaking changes into the `v0.6.0` release? Couldn't we instead merge those into a future `v0.7.0`?

I'm not saying we can't. It would be a little annoying to have an extra branch for quinn-udp 0.6.x if we end up making more breaking changes before quinn 0.12.0 is released, but I guess we'll live?

@mxinden can you execute on publishing 0.6.0?

Ralith:
I don't think we should be reluctant to have intermediate breaking releases of leaf crates. Branches and tags are cheap!

quinn-udp/src/unix.rs
```diff
@@ -295,6 +341,27 @@ impl UdpSocketState {
...
+    pub unsafe fn set_apple_fast_path(&self) {
+        self.apple_fast_path.store(true, Ordering::Relaxed);
+        self.max_gso_segments.store(BATCH_SIZE, Ordering::Relaxed);
+    }
+
+    /// Returns whether Apple's fast UDP datapath is enabled for this socket.
+    #[cfg(apple_fast)]
+    pub fn is_apple_fast_path_enabled(&self) -> bool {
+        self.apple_fast_path.load(Ordering::Relaxed)
+    }
```
thomaseizinger, line 364: These public APIs are only available on the UNIX `UdpSocketState`. Should we mirror these on the Windows one to avoid needing `cfg`s in downstream code?

mxinden: I am in favor of mirroring the public API across `unix.rs` and `windows.rs`. That said, I suggest deviating in this case because:

1. These two functions are very OS specific, likely wrapped in a `#[cfg(target_os = "macos")]` on the caller side, potentially along with some code determining whether they are safe to execute on the OS version.
2. These two functions are very experimental and thus I am not sure they will stay forever.

For the record, we also have `set_gro` on `windows.rs` only:

https://github.com/quinn-rs/quinn/blob/c67e9d8930c87270c37f142ffab2a8f50b21d50b/quinn-udp/src/windows.rs#L161-L168

@thomaseizinger do you feel strongly about mirroring these to `windows.rs`? If yes, what would `set_apple_fast_path` do on Windows?

thomaseizinger: We are currently using this feature and had it enabled for a while without any problems. Our socket code does not contain any platform-specific probing (yet) so this would be the first case of us having to add a `cfg` block to being able to call this function and continue to use the Apple fast data path.

> If yes, what would `set_apple_fast_path` do on Windows?

I would suggest that it just does nothing.

mxinden: I suggest moving the discussion to https://github.com/quinn-rs/quinn/pull/2566.

mxinden:
> @mxinden can you execute on publishing 0.6.0?

I plan to publish the release tomorrow (Wednesday), thus giving https://github.com/quinn-rs/quinn/pull/2564 and https://github.com/quinn-rs/quinn/pull/2565 time to merge and make it into `quinn-udp` `v0.6.0`.

@larseggert is that good enough for you?

mxinden:
@larseggert released as `quinn-udp` `v0.6.0`:

https://github.com/quinn-rs/quinn/releases/tag/quinn-udp-0.6.0

I will create a patch against mozilla/neqo next to update `quinn-udp`.

PR #2464, opened 2025-12-06
Use IP_RECVTOS on DragonFly

mneumann:
And fix compiler warnings.

For some reason cfg(bsd) does not include target_os="dragonfly" on Rust, so explicitly add it.

Related-Issue: PR #2455

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -1,4 +1,4 @@
-#[cfg(not(any(apple, target_os = "openbsd", solarish)))]
+#[cfg(not(any(apple, target_os = "openbsd", target_os = "dragonfly", solarish)))]
 use std::ptr;
 use std::{
     io::{self, IoSliceMut},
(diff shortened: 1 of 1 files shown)


djc:
Do you also have a PR out for libc? We generally prefer to keep these temporary.

Ralith (approved):

quinn-udp/src/unix.rs
```diff
@@ -58,6 +58,18 @@ const IPV6_DONTFRAG: libc::c_int = 62;
 #[cfg(not(any(target_os = "openbsd", target_os = "netbsd")))]
 const IPV6_DONTFRAG: libc::c_int = libc::IPV6_DONTFRAG;
 
+// Defined in netinet/in.h on DragonFly (since 4th December 2020) but not exported
+// via Rust's libc crate.
```
mxinden, line 62: In case there is a pull request for libc already, I suggest linking it here.

Issue #1978, opened 2024-09-02
Clamp effective maximum MTU using the underlying interface limits


Ralith:
This is made straightforward by Mozilla's shiny new [`mtu`](https://crates.io/crates/mtu) crate. As with MTUD support, this should be layered on top of, rather than taking the place of, the user's configuration, passed down as an additional constraint on the side from the `quinn` layer.

nadongjun:
@Ralith I’d like to take this on—please assign it to me if possible. Thanks!

Ralith:
Thanks for having a look! Feel free to reach out if you have any questions.

Ralith:
Some complexities to be wary of:
- `quinn-proto` may generate transmits to multiple addresses, even within a single connection. Transmits to different addresses may be interleaved. There isn't any mechanism to know, in advance, where the next transmit will go.
- We probably shouldn't poll this information from the OS every time we transmit.
- The value reported to the OS may change over time. It may or may not be important to react to that automatically.

nadongjun:
@Ralith Thank you for the detailed explanation. I’ll reach out if I have any questions as I progress.

larseggert:
> * We probably shouldn't poll this information from the OS every time we transmit.

Definitely not. Getting this info via routing sockets is not cheap (many syscalls).

PR #2465, opened 2025-12-06
Integrate Mozilla's mtu crate to constrain MTU discovery

ali-molajani:
This change layers interface MTU detection on top of user configuration rather than replacing it. When creating client connections, we query the interface MTU for the destination IP using Mozilla's mtu crate and apply it as an additional constraint on the MTU discovery upper bound.

Key changes:
- Add mtu crate dependency to quinn
- Query interface MTU when creating client connections in quinn layer
- Pass interface MTU constraint through to quinn-proto layer
- Apply constraint in MTU discovery SearchState, respecting user config and peer limits (effective_upper_bound = min(user_bound, interface_mtu, peer_limit))
- Add tests for interface MTU constraint behavior

The constraint is applied as an additional limit, so user configuration takes precedence if it's more restrictive, maintaining backward compatibility.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloca"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e5a7d05ea6aea7e9e64d25b9156ba2fee3fdd659e34e41063cd2fc7cd020d7f4"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "anes"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -86,6 +95,45 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c3d036a3c4ab069c7b410a2ce876bd74808d2d0888a82667669f8e783a898bf1"
 dependencies = [
  "derive_arbitrary",
+]
+
+[[package]]
+name = "asn1-rs"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "56624a96882bb8c26d61312ae18cb45868e5a9992ea73c58e45c3101e56a1e60"
+dependencies = [
+ "asn1-rs-derive",
+ "asn1-rs-impl",
+ "displaydoc",
+ "nom",
+ "num-traits",
+ "rusticata-macros",
+ "thiserror 2.0.17",
+ "time",
+]
+
+[[package]]
+name = "asn1-rs-derive"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3109e49b1e4909e9db6515a30c633684d68cdeaa252f215214cb4fa1a5bfee2c"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+ "synstructure",
+]
+
+[[package]]
+name = "asn1-rs-impl"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7b18050c2cd6fe86c3a76584ef5e0baf286d038cda203eb6223df2cc413565f7"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
 ]
 
 [[package]]
@@ -678,10 +726,11 @@ dependencies = [
 
 [[package]]
 name = "criterion"
-version = "0.7.0"
+version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e1c047a62b0cc3e145fa84415a3191f628e980b194c2755aa12300a4e6cbd928"
+checksum = "4d883447757bb0ee46f233e9dc22eb84d93a9508c9b868687b274fc431d886bf"
 dependencies = [
+ "alloca",
  "anes",
  "cast",
  "ciborium",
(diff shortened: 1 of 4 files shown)


Ralith:
See some previous discussion at https://github.com/quinn-rs/quinn/issues/1978 also. One other limitation in the current implementation is incorrect behavior when the client involuntarily migrates (e.g. a phone moving between wifi and cellular), though the worst case there is that the MTU remains clamped lower than it should be (if it's too large, that's just the status quo again), so probably not a blocker. We might want to explore monitoring the source address of outgoing packets for changes in follow-up work.

mxinden:
//CC @larseggert as the author of the mtu crate.

quinn/src/endpoint.rs
```diff
@@ -236,9 +236,22 @@ impl Endpoint {
...
+        // Query interface MTU for the destination IP as an additional constraint
+        let interface_mtu_constraint = mtu::interface_and_mtu(addr.ip())
+            .ok()
+            .and_then(|(_interface, interface_mtu)| {
+                // Convert interface MTU to UDP payload size (subtract IP + UDP headers)
+                // IPv4: 20 bytes IP header + 8 bytes UDP header = 28 bytes
+                // IPv6: 40 bytes IP header + 8 bytes UDP header = 48 bytes
+                let header_overhead = if addr.is_ipv6() { 48 } else { 28 };
+                let udp_payload_size = interface_mtu.saturating_sub(header_overhead);
+                // Convert usize to u16, clamping to u16::MAX if too large
```
mxinden, line 248: Seems like you are returning `None` instead of clamping here, no?

quinn-proto/src/connection/mod.rs
```diff
@@ -283,7 +284,7 @@ impl Connection {
                 now,
                 if pref_addr_cid.is_some() { 2 } else { 1 },
             ),
-            path: PathData::new(remote, allow_mtud, None, 0, now, &config),
+            path: PathData::new(remote, allow_mtud, None, 0, now, &config, interface_mtu_constraint),
```
djc, line 287: How about we drop the `interface_` prefix in names? It seems pretty verbose and it's not like we have a bunch of other MTU constraints right now (can expand more on it in documentation).

Cargo.lock
```diff
@@ -350,6 +350,26 @@ version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7dfdb4953a096c551ce9ace855a604d702e6e62d77fac690575ae347571717f5"
 
+[[package]]
+name = "bindgen"
```
djc, line 402: This seems ugly. Why does mtu need bindgen? Can it be updated to something newer?

Ralith: Oof, yeah, I missed that. I would rather not have a bindgen build dependency.

larseggert: We need `bindgen`, because `libc` and similar don't have the necessary bindings for talking to routing sockets.
https://github.com/mozilla/neqo/blob/042e684f33df3127f485deeb3e122a8963168158/mtu/build.rs#L51-L65

Ralith: Could we commit the bindgen output, or vendor handwritten bindings for the necessary interfaces?

larseggert: I think we originally tried that. It's a pain, because of course all the BSDs are slightly different.

djc: Sure, but they're not going to change very often, right? That feels like worth it for a small dependency crate like this (over dragging bindgen into everyone's  `Cargo.lock`).

larseggert: Well, we depend on it already anyway 🤷

djc: Sure, but that's not going to be true for most projects that want to use quinn.

djc: Alternatively if you don't want to do the legwork to make the mtu crate more broadly usable we could design API to be able to inject it somehow.

larseggert: I'm not opposed to making this more lightweight, but it won't be a priority for us, given that we already have that `bindgen` dependency for other reasons. Whatever is done would need to be something that doesn't cause breakage when the underlying OS constants and structs change.

Ralith: > when the underlying OS constants and structs change

That would break every existing binary using those calls on the OS in question, so I would be surprised if it ever happens. E.g. Linux explicitly guarantees syscall ABI perma-stability, and even Windows provides perma-stable wrappers. Surely BSDs make similar promises?

dignifiedquire: could you use one of the https://docs.rs/rtnetlink crates?

dignifiedquire: Fyi we are currently using https://docs.rs/netdev/ for these things, which is quite nice and includes support for parsing MTUs from the network interfaces

larseggert: Isn't for the first one Linux-only?

For the second, I _think_ I looked at it in the past, and it didn't do MTU. If it does, maybe we can switch to it for `neqo`, too.

larseggert: Ah, I think the issue with netdev was that it doesn't do routing table lookups.

dignifiedquire: > Ah, I think the issue with netdev was that it doesn't do routing table lookups.

I think it would be great to add that to netdev, the developer has been super responsive to any requests and improvements and it is becoming a really great crate to collect these kinds of informations across OSes

quinn/Cargo.toml
```diff
@@ -57,6 +57,7 @@ bytes = { workspace = true }
 futures-io = { workspace = true, optional = true }
 rustc-hash = { workspace = true }
 pin-project-lite = { workspace = true }
+mtu = "0.2"
```
mxinden, line 60: I just published `mtu` `v0.3.0`: https://crates.io/crates/mtu/0.3.0

```suggestion
mtu = "0.3"
```

PR #2472, opened 2025-12-09
refactor(quinn-udp): split fast&slow send/recv paths

larseggert:
Separate the fast path (`msghdr_x`-based) and slow path (`msghdr`-based) implementations to prepare for runtime dispatch between them.

Broken out of #2463 as suggested by @djc.

### DIFF
--- a/quinn-udp/src/lib.rs
+++ b/quinn-udp/src/lib.rs
@@ -161,6 +161,7 @@ impl Transmit<'_> {
     /// This case is actually quite common when splitting up a prepared GSO batch
     /// again after GSO has been disabled because the last datagram in a GSO
     /// batch is allowed to be smaller than the segment size.
+    #[cfg_attr(apple_fast, allow(dead_code))] // Used by prepare_msg, which is unused when apple_fast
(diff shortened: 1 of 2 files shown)


djc:
Can you rebase this and get CI passing?

djc (approved):

thomaseizinger:
I haven't tested the feasibility of this idea but I am wondering at what point it will be simpler to not just split the code in `quinn-udp` by Windows & UNIX but actually have different files for e.g. different operating systems / kernels. We can still share code between them through other modules in case duplication is a concern.

larseggert:
Any chance we can and this, so I can rebase #2463 to land that after?

Ralith (approved):

Ralith:
This had fallen through the cracks; thanks for the ping!

> I am wondering at what point it will be simpler to not just split the code in quinn-udp by Windows & UNIX but actually have different files for e.g. different operating systems / kernels

I don't see anything wrong with exploring this; it's definitely grown organically into a bit of a messy blob at this point.

PR #2474, opened 2025-12-12
perf(proto): use `extract_if` in `RangeSet` for better performance

### LINKED ISSUE #2422: proto: Leverage `BTreeMap::extract_if` in `RangeSet` for better performance

DSharifi:
closes #2422

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1948,6 +1948,7 @@ dependencies = [
  "rustls",
  "rustls-pki-types",
  "rustls-platform-verifier",
+ "rustversion",
  "slab",
  "thiserror 2.0.17",
  "tinyvec",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ rand = "0.9"
 rcgen = "0.14"
 ring = "0.17"
 rustc-hash = "2"
+rustversion = "1"
 rustls = { version = "0.23.5", default-features = false, features = ["std"] }
 rustls-platform-verifier = "0.6"
 rustls-pki-types = "1.7"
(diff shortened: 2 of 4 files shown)


gretchenfrage:
Thank you!

Our MSRV policy is not to bump the MSRV to a version less than 6 months old. `extract_if` was stabilized in Rust 1.91.0, which was released Oct/30/2025. So if this PR is merged before Apr/30/2026, it will have to be put behind a feature gate.

DSharifi:
> Thank you!
> 
> Our MSRV policy is not to bump the MSRV to a version less than 6 months old. `extract_if` was stabilized in Rust 1.91.0, which was released Oct/30/2025. So if this PR is merged before Apr/30/2026, it will have to be put behind a feature gate.


Yes, discussing the MSRV also came up in the issue it closes https://github.com/quinn-rs/quinn/issues/2422#issuecomment-3670650834

I'm happy with keeping it as a draft until May next year, but I can also create a feature gated version if you prefer that :)

djc:
The policy is not to bump until the MSRV is at least 6 months old, but we've mostly been more conservative, sort of matching up with tokio's MSRV. So if you want this anytime soon, feature gating is probably the better option.

gretchenfrage:
Unless you or someone really wants this soon, I think we can leave it a draft for now :)

DSharifi:
Personally, I don't mind keeping it in draft.

Let me know what you guys decide. For now I added a change to feature gate the changes with the `rustversion` crate. If we keep it feature gated, we'd need to also update CI to run these unit tests with +1.91

JTMax1:
It's July 12, 2026 now.

djc:
We should re-evaluate our MSRV stance when we're closer to releasing 0.12. I don't think I'd take the 1.91 bump today.

PR #2475, opened 2025-12-15
Re-export `InvalidCid` from `quinn-proto` in `quinn`

CircuitCoder:
The interface of `ConnectionIdGenerator` refers to `InvalidCid`, which is not re-exported in `quinn` right now. So implementers have to depend on `quinn-proto` directly.

Re-exporting `InvalidCid` alongside `ConnectionIdGenerator` makes it more convenient to provide custom CID generators.


djc:
Thanks!

PR #2479, opened 2025-12-17
Workspace lints

djc:
Have been doing similar things recently in [rustls](https://github.com/rustls/rustls/pull/2782) and HickoryDNS. In particular, I think `unnameable_types` will help us get out in front of stuff like

- #2475

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,4 +70,5 @@ unreachable_pub = "warn"
 
 [workspace.lints.clippy]
 manual_let_else = "warn"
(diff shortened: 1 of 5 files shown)


mxinden (approved):
Approving `quinn-udp` changes.

For what it is worth, here are the lints Neqo enables. Credit goes to @larseggert.

https://github.com/mozilla/neqo/blob/819f23bdfaefcce4f4d580e812dd0ad343879e87/Cargo.toml#L50-L136

PR #2486, opened 2025-12-23
feat: add try_read_datagram

Sherlock-Holo:
this allow user try to acquire as many datagram after call read_datagram().await, to reduce latency

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -357,6 +357,21 @@ impl Connection {
         }
     }
 
+    /// Try to receive an application datagram, if application datagram is unavailable, will return
+    /// `Ok(None)`
+    pub fn try_read_datagram(&self) -> Result<Option<Bytes>, ConnectionError> {
+        let mut state = self.0.state.lock("Connection::try_read_datagram");
+        // Check for buffered datagrams before checking `state.error` so that already-received
+        // datagrams, which are necessarily finite, can be drained from a closed connection.
+        if let Some(x) = state.inner.datagrams().recv() {
+            Ok(Some(x))
+        } else if let Some(e) = state.error.clone() {
+            Err(e)
+        } else {
+            Ok(None)
+        }
+    }
+
     /// Wait for the connection to be closed for any reason
     ///
     /// Despite the return type's name, closed connections are often not an error condition at the


gretchenfrage:
Is it possible to achieve this by calling `read_datagram` and then polling it once?

gretchenfrage:
Is it possible to achieve this by calling `read_datagram` and then polling it once?

Sherlock-Holo:
@gretchenfrage it is ok, but call `read_datagram` then poll once, it require user use a noop waker to poll it, or use something like https://docs.rs/futures-util/latest/futures_util/future/fn.poll_immediate.html

no matter which way, when there is no available datagram, it still need to create a `ReadDatagram` struct, call `self.0.shared.datagram_received.notified()` and poll the `Notified` once, this may be a bit too much

with this new method, user just do the necessary job to check Connection received queue, and return directly

Ralith:
Please share the latency measurements that motivate this change.

Sherlock-Holo:
> Please share the latency measurements that motivate this change.

i haven't record the latency data in my service, maybe it just reduce a little latency, but with this new method, we actually avoid to crate a `Notificed`. Every time when create a `Notificed`, it have to atomic load the `Notify` inner state (https://docs.rs/tokio/latest/src/tokio/sync/notify.rs.html#568)

gretchenfrage:
> > Please share the latency measurements that motivate this change.
> 
> i haven't record the latency data in my service, maybe it just reduce a little latency, but with this new method, we actually avoid to crate a `Notificed`. Every time when create a `Notificed`, it have to atomic load the `Notify` inner state (https://docs.rs/tokio/latest/src/tokio/sync/notify.rs.html#568)

I am sympathetic to the ergonomics argument here more-so than the performance argument.

- **On the ergonomics argument:** It is true that the user having to pull in `futures_util::future::poll_immediate` to have to do something basic like this is unfortunate. However, I'm not sure where the line is here, as in the extreme case, we would end up adding boilerplate "try now" functionality to every single future we provide to make it slightly more convenient.
- **On the performance argument:** An un-contended atomic load can be insanely cheap. Moreover, even if the `Notify` usage were removed, this would still involve locking the internal state, which presumably _also involves an atomic load_. Unless data indicates otherwise, I assume there wouldn't be a meaningful performance difference compared to somehow polling the future once.

Ralith:
Thank you for your interest in optimizing Quinn, but next time please start by benchmarking and profiling.

PR #2489, opened 2025-12-29
fix(windows): detect GSO compatibility using `WinSock::WSASendMsg`

jamilbk:


quinn-udp/src/windows.rs
```diff
@@ -434,21 +423,67 @@ static WSARECVMSG_PTR: Lazy<WinSock::LPFN_WSARECVMSG> = Lazy::new(|| {
...
-        GSO_SIZE,
-    ) {
-        // Empirically found on Windows 11 x64
-        Ok(()) => 512,
+
+    socket.set_nonblocking(true).ok();
+
+    // Test GSO support by actually trying to send with the UDP_SEND_MSG_SIZE control message.
+    // Using setsockopt() alone is not sufficient as it may succeed even when WSASendMsg()
+    // with the control message fails (e.g., due to network adapter/driver limitations).
```
mxinden, line 435: I would have assumed sending on localhost does not actually test driver support. Would you mind describing the bug you have been able to fix with this patch?

jamilbk:
> As you have closed this pull request right away, I assume you intended to open it against a different repository.

@mxinden Correct - this was meant to be opened on the firezone/quinn repository. It was very much a work in progress intended to get some early feedback on from @thomaseizinger.

The PR that ended up alleviating this particular issue in our case is: https://github.com/firezone/firezone/pull/11584.

PR #2490, opened 2025-12-29
fix(windows): do not set `UDP_SEND_MSG_SIZE` unless necessary

thomaseizinger:
A similar fix like this has landed for the UNIX codepath in #2050. This is doing the same thing for Windows by extracting a common `effective_segment_size` function on `Transmit`.

### DIFF
--- a/quinn-udp/src/windows.rs
+++ b/quinn-udp/src/windows.rs
@@ -391,7 +391,7 @@ fn send(socket: UdpSockRef<'_>, transmit: &Transmit<'_>) -> io::Result<()> {
     }
 
     // Segment size is a u32 https://learn.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-wsasetudpsendmessagesize
-    if let Some(segment_size) = transmit.segment_size {
+    if let Some(segment_size) = transmit.effective_segment_size() {
         encoder.push(
             WinSock::IPPROTO_UDP,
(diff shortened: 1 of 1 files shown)


quinn-udp/src/lib.rs
```diff
@@ -150,6 +150,28 @@ pub struct Transmit<'a> {
...
+    /// again after GSO has been disabled because the last datagram in a GSO
+    /// batch is allowed to be smaller than the segment size.
+    fn effective_segment_size(&self) -> Option<usize> {
+        let segment_size = self.segment_size?;
+
+        if segment_size >= self.contents.len() {
+            return None;
+        }
+
+        Some(segment_size)
```
djc: How about this:

```rust
match self.segment_size? {
     size if size >= self.contents.len() => None,
     size => Some(size),
}
```

djc (approved):
Makes sense to me.

thomaseizinger:
> Is this fixing a concrete issue you are seeing on Windows?

I can't 100% confirm that. What I can say is that we are seeing a lot telemetry logs where sending single packets (like DNS queries) doesn't work. We are fixing this now by not using `quinn-udp` sockets for everything.

We did see and fix a similar issue like this for Android versions with 5.x kernels.

I don't have the hardware to confirm this fix unfortunately so it is more of an educated guess that this is causing issues.

thomaseizinger:
The concrete error we are seeing from Windows is:

> An invalid argument was specified (os error 10022)

PR #2495, opened 2026-01-09
Add early return in RecvStream::drop when all data read is true

Perelyn-sama:
Solana validators ingest hundreds of thousands of streams per second during leader rounds (1 stream = 1 transaction). The receive loop looks something like this (pseudo code):

```rust
loop {
  let mut stream = conn.accept_uni().await;
  while stream_has_data {
    stream.read_chunks(&mut chunks).await;        
  }
  // implicit drop(stream)
}
```

So we accept, read a stream until completion, and then drop it.

RecvStream::drop currently acquires the connection state mutex which is quite contended (see profile below).

The solution I'm proposing for this problem is an early return in cases where all data read is true and it would be safe to ignore the remove operation on `conn.blocked_readers` since you cannot have a blocked reader when the all data read is true as tested in https://github.com/Perelyn-sama/quinn/blob/be95075542a318a19d05461c80669e3f12ee0c66/quinn/src/tests.rs#L949 and the error case also does an early drop, so it is safe to ignore as well.

Futhermore, the optimization extends beyond recvstream::drop, since `ConnectionRef` is also part of `RecvStream` and it has a mutex lock in its own drop function, I was still facing contention. My solution is to move the connection reference counter from `connection::State` to `connection::Shared` and wrap it around an `AtomicUsize` so it could be safely operated on in multiple threads. I'm doing this because `ref_count` is the main variable that is needed in `ConnectionRef::clone` and `ConnectionRef::drop` and placing a mutex lock just to access a usize seemed inefficient.

With the changes in this PR, I was able to improve `RecvStream::drop` by 14% in this simple client/server project that reproduces the issue(https://github.com/Perelyn-sama/ales-interlude/blob/master/src/bin/server.rs)

Without patch:
photo_2026-01-08_03-40-47

with patch
photo_2026-01-08_03-43-00

### DIFF
--- a/quinn/src/tests.rs
+++ b/quinn/src/tests.rs
@@ -7,10 +7,15 @@ use rustls::crypto::ring::default_provider;
 
 use std::{
     convert::TryInto,
+    future::Future,
     io,
     net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr, UdpSocket},
     str,
-    sync::Arc,
+    sync::{
+        Arc,
+        atomic::{AtomicUsize, Ordering},
+    },
+    task::{Context, Poll, RawWaker, RawWakerVTable, Waker},
 };
 
 use crate::runtime::TokioRuntime;
@@ -23,6 +28,7 @@ use rustls::{
     pki_types::{CertificateDer, PrivateKeyDer, PrivatePkcs8KeyDer},
 };
 use tokio::runtime::{Builder, Runtime};
+use tokio::time::{sleep, timeout};
 use tracing::{error_span, info};
 use tracing_futures::Instrument as _;
 use tracing_subscriber::EnvFilter;
@@ -157,7 +163,7 @@ fn read_after_close() {
             .unwrap()
             .await
             .expect("connect");
-        tokio::time::sleep(Duration::from_millis(100)).await;
+        sleep(Duration::from_millis(100)).await;
         let mut stream = new_conn.accept_uni().await.expect("incoming streams");
         let msg = stream.read_to_end(usize::MAX).await.expect("read_to_end");
         assert_eq!(msg, MSG);
@@ -900,8 +906,7 @@ async fn stream_stopped() {
         let stopped3 = stopped3.await;
         assert_eq!(stopped3, Ok(Some(42u32.into())));
     };
-    let client =
-        tokio::time::timeout(Duration::from_millis(100), client).instrument(error_span!("client"));
+    let client = timeout(Duration::from_millis(100), client).instrument(error_span!("client"));
     let server = async move {
         let conn = server.accept().await.unwrap().await.unwrap();
         let mut stream = conn.accept_uni().await.unwrap();
@@ -929,7 +934,7 @@ async fn stream_stopped_2() {
     )
     .unwrap();
     let send_stream = conn.open_uni().await.unwrap();
-    let stopped = tokio::time::timeout(Duration::from_millis(100), send_stream.stopped())
+    let stopped = timeout(Duration::from_millis(100), send_stream.stopped())
         .instrument(error_span!("stopped"));
     tokio::pin!(stopped);
     // poll the future once so that the waker is registered.
@@ -943,4 +948,113 @@ async fn stream_stopped_2() {
     // make sure the stopped future still resolves
     let res = stopped.await;
     assert_eq!(res, Ok(Ok(None)));
+}
+
+#[tokio::test]
+async fn stream_drop_removes_blocked_reader() {
+    let _guard = subscribe();
+
+    for drop_stream in [false, true] {
+        let endpoint_factory = EndpointFactory::new();
+        let server = endpoint_factory.endpoint();
+        let server_address = server.local_addr().unwrap();
+        let client = endpoint_factory.endpoint();
+
+        let server_task = tokio::spawn(async move {
+            let conn = server.accept().await.unwrap().await.unwrap();
+            let mut stream = conn.accept_uni().await.unwrap();
+
+            // read "hello"
+            let mut buf = [0u8; 5];
+            stream.read_exact(&mut buf).await.unwrap();
+
+            let (waker, wake_counter) = new_count_waker();
+            let mut cx = Context::from_waker(&waker);
+            // do a blocking read which will add the stream in conn.blocked_readers
+            {
+                let mut buf = [0u8; 64];
+                let read_fut = stream.read(&mut buf);
+                tokio::pin!(read_fut);
+                assert!(matches!(read_fut.as_mut().poll(&mut cx), Poll::Pending));
+            }
+
+            if !drop_stream {
+                assert_eq!(wake_counter.wakes(), 0);
+                // We have a blocked reader, closing the connection should wake it. We use this as
+                // a proxy to assert that the stream is in conn.blocked_readers.
+                conn.close(0u32.into(), b"done");
+                assert_eq!(wake_counter.wakes(), 1);
+            } else {
+                // dropping the stream should remove it from conn.blocked_readers, so we don't
+                // expect any wakeups
+                drop(stream);
+                assert_eq!(wake_counter.wakes(), 0, "no wakeups should have occurred");
+                conn.close(0u32.into(), b"done");
+                assert_eq!(wake_counter.wakes(), 0, "no wakeups should have occurred");
+            }
+        });
+
+        let conn = client
+            .connect(server_address, "localhost")
+            .unwrap()
+            .await
+            .unwrap();
+        let mut stream = conn.open_uni().await.unwrap();
+        // need to send some data to actually start the stream
+        stream.write_all(b"hello").await.unwrap();
+
+        server_task.await.unwrap();
+    }
+}
+
+#[derive(Default)]
+struct WakeCounter {
+    wakes: AtomicUsize,
+}
+
+impl WakeCounter {
+    fn wakes(&self) -> usize {
+        self.wakes.load(Ordering::SeqCst)
+    }
+}
+
+fn new_count_waker() -> (Waker, Arc<WakeCounter>) {
+    // instance of WakeCounter
+    let counter = Arc::new(WakeCounter::default());
+
+    // convert
+    let waker = unsafe { Waker::from_raw(raw_waker(counter.clone())) };
+    (waker, counter)
+}
+
+fn raw_waker(counter: Arc<WakeCounter>) -> RawWaker {
+    // Store an Arc<WakeCounter> behind the raw pointer.
+    let ptr = Arc::into_raw(counter) as *const ();
+    RawWaker::new(ptr, &VTABLE)
+}
+
+static VTABLE: RawWakerVTable =
+    RawWakerVTable::new(clone_waker, wake_waker, wake_by_ref_waker, drop_waker);
+
+unsafe fn clone_waker(data: *const ()) -> RawWaker {
+    let arc = Arc::<WakeCounter>::from_raw(data as *const WakeCounter);
+    let cloned = arc.clone();
+    std::mem::forget(arc);
+    raw_waker(cloned)
+}
+
+unsafe fn wake_waker(data: *const ()) {
+    let arc = Arc::<WakeCounter>::from_raw(data as *const WakeCounter);
+    arc.wakes.fetch_add(1, Ordering::SeqCst);
+    // arc drops here
+}
+
+unsafe fn wake_by_ref_waker(data: *const ()) {
+    let arc = Arc::<WakeCounter>::from_raw(data as *const WakeCounter);
+    arc.wakes.fetch_add(1, Ordering::SeqCst);
+    std::mem::forget(arc);
+}
+
+unsafe fn drop_waker(data: *const ()) {
+    drop(Arc::<WakeCounter>::from_raw(data as *const WakeCounter));
 }
--- a/quinn/src/recv_stream.rs
+++ b/quinn/src/recv_stream.rs
@@ -505,6 +505,19 @@ impl tokio::io::AsyncRead for RecvStream {
 
 impl Drop for RecvStream {
     fn drop(&mut self) {
+        if self.all_data_read {
+            debug_assert!(
+                !self
+                    .conn
+                    .state
+                    .lock("RecvStream:drop")
+                    .blocked_readers
+                    .contains_key(&self.stream),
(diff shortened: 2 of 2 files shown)


quinn/src/tests.rs
```diff
@@ -944,3 +944,40 @@ async fn stream_stopped_2() {
...
+    let new_conn = client
+        .connect(server_address.unwrap(), "localhost")
+        .unwrap()
+        .await
+        .unwrap();
+
+    let mut stream = new_conn.open_uni().await.unwrap();
+    stream.write_all(b"some data").await.unwrap();
+
+    tokio::time::sleep(Duration::from_millis(5000)).await;
```
Ralith: Relying on timing like this is fragile and degrades the development experience. Can we instead send an explicit signal from the server task?

quinn/src/tests.rs
```diff
@@ -944,3 +944,40 @@ async fn stream_stopped_2() {
     let res = stopped.await;
     assert_eq!(res, Ok(Ok(None)));
 }
+
+#[tokio::test]
+async fn test_blocked_readers() {
```
Ralith: Please document the intent of this test -- reverse-engineering it from the name and code is not trivial.

quinn/src/tests.rs
```diff
@@ -944,3 +944,40 @@ async fn stream_stopped_2() {
...
+    let client = endpoint_factory.endpoint();
+
+    let server_task = tokio::spawn(async move {
+        let new_conn = server.accept().await.unwrap().await.unwrap();
+        let mut s = new_conn.accept_uni().await.unwrap();
+
+        let mut buf = [0u8; 64];
+        s.read(&mut buf).await.unwrap();
+
+        let r = tokio::time::timeout(Duration::from_millis(1), s.read(&mut buf)).await;
```
Ralith: Is this timeout necessary? It's not obvious to me if a 1ms timeout is even guaranteed to do anything. Could we just poll the future and assert that it returns `Pending`?

quinn/src/connection.rs
```diff
@@ -917,17 +917,24 @@ impl ConnectionRef {
...
 impl Drop for ConnectionRef {
     fn drop(&mut self) {
-        let conn = &mut *self.state.lock("drop");
-        if let Some(x) = conn.ref_count.checked_sub(1) {
-            conn.ref_count = x;
-            if x == 0 && !conn.inner.is_closed() {
+        let ref_count = self
+            .shared
+            .ref_count
+            .fetch_sub(1, std::sync::atomic::Ordering::Relaxed);
```
Ralith: A similar refactor of `EndpointRef` would be very welcome.

quinn/src/connection.rs
```diff
@@ -917,19 +920,14 @@ impl ConnectionRef {
 
 impl Clone for ConnectionRef {
     fn clone(&self) -> Self {
-        self.shared
-            .ref_count
-            .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+        self.shared.ref_count.fetch_add(1, Ordering::Relaxed);
```
Ralith, line 923: nit: please squash into the prior commit

Perelyn-sama: done 

[move ref_count from state to shared](https://github.com/quinn-rs/quinn/pull/2495/commits/b941ee0474bdb9af03920b30641124102fa4184f)

quinn/src/connection.rs
Perelyn-sama: thank you for pointing that out - https://github.com/quinn-rs/quinn/pull/2495/commits/6256a450e3b23e473e394d717595b4c355917856

quinn/src/tests.rs
```diff
@@ -961,8 +963,14 @@ async fn test_blocked_readers() {
         let mut buf = [0u8; 64];
         s.read(&mut buf).await.unwrap();
 
-        let r = tokio::time::timeout(Duration::from_millis(1), s.read(&mut buf)).await;
-        assert!(r.is_err());
+        // Verify that another read would block (returns Poll::Pending)
+        let read_fut = s.read(&mut buf);
+        tokio::pin!(read_fut);
+        tokio::select! {
```
Ralith: `select` is a *really* big hammer for this. Can we just call `Future::poll`? Note [`Waker::noop`](https://doc.rust-lang.org/std/task/struct.Waker.html#method.noop) will help.

quinn/src/tests.rs
```diff
@@ -945,6 +946,7 @@ async fn stream_stopped_2() {
     assert_eq!(res, Ok(Ok(None)));
 }
 
+/// Verify that dropping a stream after all reads complete leaves no blocked readers
```
Ralith: How does the test verify that there were no blocked readers?

quinn/src/tests.rs
Perelyn-sama: removed the sleep and moved `server_task.await` before sendstream drop, so the the server completes all its reads before the drop - https://github.com/Perelyn-sama/quinn/blob/574b55c66dec52b24cdb4246b15d8936aaf837a5/quinn/src/tests.rs#L986

quinn/src/tests.rs
Perelyn-sama: I added some comments - https://github.com/Perelyn-sama/quinn/blob/574b55c66dec52b24cdb4246b15d8936aaf837a5/quinn/src/tests.rs#L949 

let me know if you need me to add more

quinn/src/tests.rs
Perelyn-sama: yes - https://github.com/Perelyn-sama/quinn/blob/574b55c66dec52b24cdb4246b15d8936aaf837a5/quinn/src/tests.rs#L966

quinn/src/tests.rs
Ralith: No receive stream in this test ever enters the "all reads complete" state, because the corresponding send stream is never finished.

quinn/src/tests.rs
Perelyn-sama: done - https://github.com/Perelyn-sama/quinn/blob/fee20346373815145f41541ebf00c12dee20f709/quinn/src/tests.rs#L976

thanks for the `Waker::noop` hint, it saved me time

quinn/src/tests.rs
```diff
@@ -944,3 +944,40 @@ async fn stream_stopped_2() {
     let res = stopped.await;
     assert_eq!(res, Ok(Ok(None)));
 }
+
+#[tokio::test]
+async fn test_blocked_readers() {
+    let __guard = subscribe();
```
djc: Style nit: I think convention in this repo is single underscore prefix.

quinn/src/tests.rs
djc: Style nit: please import `timeout()` and `sleep()` at the module level.

quinn/src/connection.rs
```diff
@@ -917,17 +920,19 @@ impl ConnectionRef {
 
 impl Clone for ConnectionRef {
     fn clone(&self) -> Self {
-        self.state.lock("clone").ref_count += 1;
+        self.shared.ref_count.fetch_add(1, Ordering::Relaxed);
```
djc, line 923: Can we add comments for these ref count manipulations justifying why `Ordering::Relaxed` is sufficient to get the semantics we need?

quinn/src/connection.rs
```diff
@@ -917,17 +920,19 @@ impl ConnectionRef {
...
     }
 }
 
 impl Drop for ConnectionRef {
     fn drop(&mut self) {
-        let conn = &mut *self.state.lock("drop");
-        if let Some(x) = conn.ref_count.checked_sub(1) {
-            conn.ref_count = x;
-            if x == 0 && !conn.inner.is_closed() {
+        let ref_count = self.shared.ref_count.fetch_sub(1, Ordering::Relaxed);
```
djc: Pre-existing, but it would be nice to use an early return here and in the other similar case in the next commit, reducing rightward drift.

quinn/src/tests.rs
```diff
@@ -961,8 +964,16 @@ async fn test_blocked_readers() {
         let mut buf = [0u8; 64];
         s.read(&mut buf).await.unwrap();
 
-        let r = tokio::time::timeout(Duration::from_millis(1), s.read(&mut buf)).await;
-        assert!(r.is_err());
+        // Verify that another read would block (returns Poll::Pending)
```
djc: This last commit should be squashed into the commit adding the test.

quinn/src/connection.rs
Ralith: FWIW I feel this is fairly self-explanatory (we aren't using the atomics to guard any shared memory access), though a comment wouldn't do any harm.

djc: Fair enough.

quinn/src/tests.rs
Perelyn-sama: you're right, it doesn't. Rewrote the test to verify that my changes don't break the `blocked_readers.remove()` behavior instead - https://github.com/Perelyn-sama/quinn/blob/6556952ff2e121cb81d56bea698cd4c02263f9b7/quinn/src/tests.rs#L954

quinn/src/tests.rs
Perelyn-sama: thanks for pointing that out. Fixed - https://github.com/Perelyn-sama/quinn/blob/6556952ff2e121cb81d56bea698cd4c02263f9b7/quinn/src/tests.rs#L955

quinn/src/tests.rs
Perelyn-sama: I removed `timeout()` and `sleep()` from my test but I noticed some other tests do this, so I fixed it - https://github.com/Perelyn-sama/quinn/commit/ba037c552ede0d1482146a997436d406c2ec8ad0

quinn/src/tests.rs
Perelyn-sama: I'll squash it all to one commit when this gets approved

quinn/src/connection.rs
Perelyn-sama: added early return to `EndpointRef::drop` and `ConnectionRef::drop` - https://github.com/Perelyn-sama/quinn/commit/70f88ece46f789975688a442df7ba074601c3aed

quinn/src/tests.rs
djc: I'd like to see three commits:

- Fix the (pre-existing) rightward drift by inverting conditions
- Move the ref counts out
- Early return in `RecvStream::drop()`

Perelyn-sama: done

quinn/src/tests.rs
```diff
@@ -944,3 +949,112 @@ async fn stream_stopped_2() {
...
+        let server_address = server.local_addr().unwrap();
+        let client = endpoint_factory.endpoint();
+
+        let server_task = tokio::spawn(async move {
+            let conn = server.accept().await.unwrap().await.unwrap();
+            let mut stream = conn.accept_uni().await.unwrap();
+
+            // read "hello"
+            let mut buf = [0u8; 64];
+            stream.read(&mut buf).await.unwrap();
```
Ralith: `read` isn't guaranteed to read everything that was sent, only some nonempty prefix. Consider `read_exact` with a slice exactly matching the length of the expected message. Otherwise, this might e.g. read a single byte, leaving the next `read` operation to succeed reading some amount of the remaining data.

quinn/src/tests.rs
```diff
@@ -944,3 +949,112 @@ async fn stream_stopped_2() {
...
+        });
+
+        let conn = client
+            .connect(server_address, "localhost")
+            .unwrap()
+            .await
+            .unwrap();
+        let mut stream = conn.open_uni().await.unwrap();
+        // need to send some data to actually start the stream
+        stream.write_all(b"hello").await.unwrap();
```
Ralith, line 1004: Is this test intended to exercise the new logic? It still won't, because it's still leaving the stream open, and hence possibly eligible for future writes. `write_all` doesn't finish the stream, it just ensures all the supplied data gets enqueued. 

To trigger the new path in `RecvStream::drop`, you need to first `finish` (or equivalently, drop) the `SendStream` on the send side, then on the receive side, wait for `read` to return `None`, or equivalently or for `read_to_end` to complete. Only then will `RecvStream::all_data_read` become true.

Ralith (changes requested):

quinn/src/tests.rs
Perelyn-sama: The `stream_drop_removes_blocked_reader` test  is intended to be a regression test that confirms blocked readers are properly removed even with the new changes made to `RecvStream::drop`

Ralith: Ah, I think I see -- we're making sure that the new code *doesn't* run in the cases where that's the intention.

quinn/src/tests.rs
Perelyn-sama: fixed - https://github.com/Perelyn-sama/quinn/blob/c0fa4e1c9d65f4ff7ae68dd4ee219cefac2d2709/quinn/src/tests.rs#L969

Ralith (approved):
Thanks!

Itsusinn:
Using `RecvStream::stop` before `drop` will casue [panic](https://github.com/quinn-rs/quinn/pull/2495/changes#diff-100957b148fa9608d9193aa3afbece8d75f4e28947405663f9e6c382e8260c31R508-R520). Is this expected?

Ralith:
No, that is not expected. Have you observed this panic? Can you share a minimal test which reproduces it?

Itsusinn:
@Ralith Seems it's my problem. I considered RecvStream as a `AsyncRead + AsyncWrite` and used `tokio::io::AsyncReadExt`. 

Using [`AsyncReadExt::read`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#method.read) within a cancelled `tokio::select!` branch might created a `blocked_readers`. (`Even after exceeding the scope of select, it remains`)

Anyway, there is my changes to my own code

https://github.com/Itsusinn/tuic/commit/78dc13295445ef075f3d77cc26597d8db32413eb#diff-733033a70a4174a15bb0da2648bb2934a96773e6599a11cedb65496fb2035f0b

Ralith:
I'm glad you've found a workaround, but if that panic is reachable by *any* means through our public API, then it's our bug. The clue about a `select!` branch was the key -- https://github.com/quinn-rs/quinn/pull/2541 should fix this, and your original code should work again.

Issue #2504, opened 2026-01-15
tokio agnosticism


parazyd:
Hi.

I've noticed that quinn has a mandatory dependency on tokio's sync primitives.
For example here: https://github.com/quinn-rs/quinn/blob/main/quinn/src/connection.rs#L16

For users of `smol`/`async-std` it should be possible to avoid this dependency. It would require
an abstraction over these primitives so that when `runtime-tokio` is requested, the existing
code would be used, or - when `runtime-smol` is requested - it should be possible to implement
these using `async-channel`, `event-listener`, and `futures-lite`.

This would greatly help projects that use `smol` and do not want to introduce a `tokio` dependency,
regardless if it's just the sync primitives.

Is this something that the developers would consider?

djc:
If someone wants to do the work and it can be wrapped up in a way to avoid complexity in the code libraries, I think we'd be open to that.

Ralith:
> This would greatly help projects that use smol and do not want to introduce a tokio dependency,
regardless if it's just the sync primitives.

Can you elaborate on why that's a useful goal? Is there something wrong with tokio's sync primitives?

parazyd:
For users who deliberately choose `smol` for its light footprint, having a mandatory dependency on tokio undermines that choice. Silently pulling in tokio feels like a leaky abstraction and gives off the vibe that the runtime choice isn't fully respected.

Users choosing smol do so to minimize their dependency tree and by forcing tokio this isn't really working out.

smol already has its ecosystem of crates that does everything necessary and when opting into it, you don't really want to use tokio or have to implement shims that make things compatible with tokio's stuff.

Ralith:
Using tokio's sync primitives with runtime features disabled is very different from using the tokio runtime, and no shims are required to interact with quinn. We don't expose our choice of sync primitives, and even if we did, they don't require shims regardless.

Have you measured an adverse impact on footprint? How large is it?

parazyd:
I haven't made precise measurements, but on a baseline Tokio's sync module is ~9000 SLOC, while event-listener and async-channel together are ~3000 SLOC.

What's more important is that smol ecosystem users prefer to use smol's ecosystem libraries. And forcing tokio on users clashes with this.

djc:
[...]

PR #2505, opened 2026-01-15
quinn: Implement abstractions over hard tokio dependencies.

parazyd:
This PR implements abstractions over the hard `tokio` dependency allowing
users of `runtime-smol` to avoid having to introduce a dep on `tokio` in their
projects.

I was unable to extrapolate the tests to not use tokio since that's heavily integrated,
but I have had success using `runtime-smol` in an external project and it did not have
to pull in tokio.

This is related to https://github.com/quinn-rs/quinn/issues/2504

It's an initial suggestion and I'm open to feedback and review.

This change should also require one of either `runtime-tokio` or `runtime-smol`
to be enabled, and they should likely be exclusive.

Cc: @djc

### DIFF
--- a/quinn/src/abstraction/sync.rs
+++ b/quinn/src/abstraction/sync.rs
@@ -1,0 +1,228 @@
+#[cfg(feature = "runtime-tokio")]
+pub(crate) use tokio::sync::{Notify, futures::Notified};
+
+#[cfg(all(feature = "runtime-smol", not(feature = "runtime-tokio")))]
+pub(crate) use smol_impl::{Notified, Notify};
+
+#[cfg(all(feature = "runtime-smol", not(feature = "runtime-tokio")))]
+mod smol_impl {
+    use event_listener::{Event, EventListener};
+    use std::fmt;
+    use std::future::Future;
+    use std::pin::Pin;
+    use std::sync::atomic::{AtomicUsize, Ordering};
+    use std::task::{Context, Poll};
+
+    /// A notification primitive for waking tasks.
+    pub(crate) struct Notify {
+        /// The underlying event for broadcasting notifications.
+        event: Event,
+        /// Counter to track stored notifications for `notify_one` semantics.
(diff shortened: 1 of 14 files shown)


quinn/src/abstraction/sync.rs
```diff
@@ -0,0 +1,314 @@
...
+use std::task::{Context, Poll};
+
+#[cfg(feature = "runtime-tokio")]
+pub(crate) use tokio_impl::{Notified, Notify};
+
+#[cfg(all(feature = "runtime-smol", not(feature = "runtime-tokio")))]
+pub(crate) use smol_impl::{Notified, Notify};
+
+#[cfg(feature = "runtime-tokio")]
+mod tokio_impl {
```
djc: Why do we need wrappers for the existing tokio stuff?

parazyd: They are zero-cost, and this makes it so that no changes are needed in the existing code.
By having this, the changes are trivial: just changing the imports.

djc: Yes, that's what I meant. As a maintainer, I don't want to have to pour over all that boilerplate.

parazyd: Okay, thanks. Now it should be better :)

gretchenfrage:
Hmm, I'm not sure I see the motivation for this.

For the async runtime itself, I see the reason we make it modular. It leaks through the abstraction, determining what runtime your futures have to be spawned into. Also, it does system calls and spawns threads in opinionated ways, etc.

However, for just using tokio's `sync` feature, that seems equivalent to any other utility crate we're using. I feel that if the tokio project separated it out into some `tokio_sync` crate rather than a `sync` feature of the `tokio` crate it would be substantively the same, but less controversial.

Am I missing something?

Ralith:
See ongoing discussion on that issue in https://github.com/quinn-rs/quinn/issues/2504.

PR #2532, opened 2026-02-13
quinn-udp: make ECN best-effort on Windows (Wine/Proton)

ogyrec-o:
## What changed
Some Winsock providers (notably Wine/Proton) don't implement ECN socket options/control messages
(`IP_RECVECN`/`IPV6_RECVECN` and `IP_ECN`/`IPV6_ECN`). When ECN is enabled unconditionally, endpoint
creation can fail with `WSAEOPNOTSUPP (10045)` / `WSAENOPROTOOPT (10042)`.

This PR makes ECN best-effort on Windows:

- During socket initialization, attempt to enable `IP_RECVECN` / `IPV6_RECVECN`.
  - If the provider reports `WSAENOPROTOOPT` or `WSAEOPNOTSUPP`, ECN is disabled for that address family
    and we continue without failing.
  - Other errors are still treated as fatal (preserves existing strictness for real failures).
- When sending, we only attach ECN control messages if ECN is supported for the destination's address family.
- Existing behavior for `PKTINFO`, `DONTFRAG`, and GSO is preserved.

## Why
ECN is an optimization; transport should remain functional even if ECN is unavailable. Under Wine/Proton,
unconditional ECN setup can prevent QUIC endpoints from being created at all.

## Repro / symptoms (before)
Running a Quinn-based client under Wine can fail during endpoint creation with something like:

```text
fixme:winsock:setsockopt Unknown IPPROTO_IP optname 0x00000032
fixme:winsock:setsockopt Unknown IPPROTO_IPV6 optname 0x00000032
... failed to create QUIC endpoint: OS Error 10045 ...
```

## Result (after)
QUIC endpoint creation succeeds under Wine/Proton, with ECN automatically disabled when unsupported.

### DIFF
--- a/quinn-udp/src/windows.rs
+++ b/quinn-udp/src/windows.rs
@@ -24,6 +24,15 @@ use crate::{
 #[derive(Debug)]
 pub struct UdpSocketState {
     last_send_error: Mutex<Instant>,
+
+    /// Whether the underlying Winsock provider supports IPv4 ECN socket options/control messages.
+    ///
+    /// Some environments (notably Wine/Proton) don't implement IP_RECVECN/IP_ECN.
+    /// ECN is best-effort: when unsupported we continue without it.
+    ecn_v4_supported: bool,
+
+    /// Whether the underlying Winsock provider supports IPv6 ECN socket options/control messages.
+    ecn_v6_supported: bool,
 }
 
 impl UdpSocketState {
@@ -67,6 +76,20 @@ impl UdpSocketState {
             ));
         }
 
+        // ECN is best-effort on Windows: if the Winsock provider doesn't support these options
+        // (common under Wine/Proton), we disable ECN and keep working.
+        let is_ecn_unsupported = |e: &io::Error| {
+            matches!(
+                e.raw_os_error(),
+                Some(code)
+                    if code == WinSock::WSAENOPROTOOPT as i32
+                    || code == WinSock::WSAEOPNOTSUPP as i32
+            )
+        };
+
+        let mut ecn_v4_supported = true;
+        let mut ecn_v6_supported = true;
+
         if is_ipv4 {
             set_socket_option(
                 &*socket.0,
@@ -81,12 +104,20 @@ impl UdpSocketState {
                 WinSock::IP_PKTINFO,
                 OPTION_ON,
             )?;
-            set_socket_option(
+
+            if let Err(e) = set_socket_option(
                 &*socket.0,
                 WinSock::IPPROTO_IP,
                 WinSock::IP_RECVECN,
                 OPTION_ON,
-            )?;
+            ) {
+                if is_ecn_unsupported(&e) {
+                    ecn_v4_supported = false;
+                    debug!("quinn-udp: ECN disabled for IPv4 (IP_RECVECN unsupported): {e}");
+                } else {
+                    return Err(e);
+                }
+            }
         }
 
         if is_ipv6 {
@@ -104,17 +135,26 @@ impl UdpSocketState {
                 OPTION_ON,
             )?;
 
-            set_socket_option(
+            if let Err(e) = set_socket_option(
                 &*socket.0,
                 WinSock::IPPROTO_IPV6,
                 WinSock::IPV6_RECVECN,
                 OPTION_ON,
-            )?;
+            ) {
+                if is_ecn_unsupported(&e) {
+                    ecn_v6_supported = false;
+                    debug!("quinn-udp: ECN disabled for IPv6 (IPV6_RECVECN unsupported): {e}");
+                } else {
+                    return Err(e);
+                }
+            }
         }
 
         let now = Instant::now();
         Ok(Self {
             last_send_error: Mutex::new(now.checked_sub(2 * IO_ERROR_LOG_INTERVAL).unwrap_or(now)),
+            ecn_v4_supported,
+            ecn_v6_supported,
         })
     }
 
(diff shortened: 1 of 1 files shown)


Ralith:
The general approach seems reasonable. Do we ever actually expect the support state for IPv4 and IPv6 to be different?

ogyrec-o:
> The general approach seems reasonable. Do we ever actually expect the support state for IPv4 and IPv6 to be different?

Good question.

In practice they'll often match on "real" Windows, but i don't think we can rely on that as an invariant:
- IPv4 and IPv6 use different socket options (IP_RECVECN vs IPV6_RECVECN) and different control messages (IP_ECN vs IPV6_ECN), i.e. distinct code paths in the Winsock provider.
- In Wine/Proton (the motivation for this change) it's plausible for one family to be implemented while the other isn't, or to fail with different error codes.

Keeping two booleans is a tiny cost and lets us preserve ECN on whichever family works instead of disabling it globally. if you strongly prefer simplification, i can collapse to a single `ecn_supported` that disables ECN entirely when either option is unsupported, but that would be a slightly more conservative behavior

quinn-udp/src/windows.rs
```diff
@@ -24,6 +24,15 @@ use crate::{
 #[derive(Debug)]
 pub struct UdpSocketState {
     last_send_error: Mutex<Instant>,
+
+    /// Whether the underlying Winsock provider supports IPv4 ECN socket options/control messages.
+    ///
+    /// Some environments (notably Wine/Proton) don't implement IP_RECVECN/IP_ECN.
```
mxinden, line 30: Is there an authoritative upstream reference to a bug or similar that we can link here?

quinn-udp/src/windows.rs
```diff
@@ -67,6 +76,24 @@ impl UdpSocketState {
             ));
         }
 
+        // ECN is best-effort on Windows: if the Winsock provider doesn't support these options
+        // (common under Wine/Proton), we disable ECN and keep working.
+        //
+        // Typical errors:
+        // - WSAENOPROTOOPT (10042): option not supported
+        // - WSAEOPNOTSUPP (10045): operation not supported
```
mxinden:
```suggestion
```

Feel free to ignore. In my eyes, clear already through the implementation itself.

quinn-udp/src/windows.rs
```diff
@@ -379,15 +438,24 @@ fn send(socket: UdpSockRef<'_>, transmit: &Transmit<'_>) -> io::Result<()> {
...
-        encoder.push(WinSock::IPPROTO_IP, WinSock::IP_ECN, ecn);
+
+    let ecn_supported = if is_ipv4 {
+        ecn_v4_supported
     } else {
-        encoder.push(WinSock::IPPROTO_IPV6, WinSock::IPV6_ECN, ecn);
+        ecn_v6_supported
+    };
+
+    if ecn_supported {
```
mxinden: Wouldn't this be easier?


```suggestion
    if (is_ipv4 && ecn_v4_supported) || (!is_ipv4 && ecn_v6_supported) {
```

mxinden:
> Keeping two booleans is a tiny cost and lets us preserve ECN on whichever family works instead of disabling it globally.

:+1:

ogyrec-o:
> Do I understand correctly that Wine and Proton return these errors on `set_socket_option` only and not when actually sending?

In my repro (Wine/Proton), the failure happens during socket initialization when enabling IP_RECVECN / IPV6_RECVECN via setsockopt (WSAENOPROTOOPT 10042 / WSAEOPNOTSUPP 10045), which prevents endpoint creation entirely

I didn't get to observe send-time behavior before this change because endpoint creation failed early. The patch avoids attaching ECN control messages unless the corresponding IP*_RECVECN option was successfully enabled for that address family, so we shouldn't hit send-time errors for ECN in the unsupported case

If you think it's valuable, I can add a follow-up fallback that retries WSASendMsg without ECN on the same error codes, but I wanted to keep this PR minimal and focused on the initialization failure

quinn-udp/src/windows.rs
ogyrec-o: I couldn't find a public Wine/Proton tracker issue specifically for IP_RECVECN/IP_ECN yet

What I can link as an authoritative reference is the official Winsock ECN documentation describing these options/control messages:
https://learn.microsoft.com/en-us/windows/win32/winsock/winsock-ecn

If you'd prefer a Wine-specific reference, I can file a WineHQ bug with a minimal repro + the observed logs (Unknown IPPROTO_* optname 0x32 / WSAEOPNOTSUPP) and then update this comment with that link

quinn-udp/src/windows.rs
ogyrec-o: Yep, that's fine too

quinn-udp/src/windows.rs
mxinden: I am fine without. Thank you.

mxinden:
> If you think it's valuable, I can add a follow-up fallback that retries WSASendMsg without ECN on the same error codes, but I wanted to keep this PR minimal and focused on the initialization failure

I suggest only doing so in case we see these failures on `send` in the wild.

quinn-udp/src/windows.rs
ogyrec-o: > I am fine without. Thank you.

Thanks, understood

FYI I went ahead and filed a WineHQ bug with a minimal repro + Wine winsock trace:
https://bugs.winehq.org/show_bug.cgi?id=59427

No need to block this PR on it. sharing the reference in case it's useful later

ogyrec-o:
> > If you think it's valuable, I can add a follow-up fallback that retries WSASendMsg without ECN on the same error codes, but I wanted to keep this PR minimal and focused on the initialization failure
> 
> I suggest only doing so in case we see these failures on `send` in the wild.

Sounds good - agreed

ogyrec-o:
> LGTM -- can you squash all commits into one?

Done. squashed to a single commit and force-pushed. thanks!

PR #2533, opened 2026-02-13
quinn-proto: fix CID exhaustion check overflow on 32-bit targets

KevinRoebert:
## Summary

Fixes an overflow in `Endpoint::cids_exhausted()` on 32-bit targets when CID length is 4.

The previous check used `usize` exponentiation:
- `2usize.pow(cid_len * 8)` overflows for `cid_len == 4` on 32-bit platforms (`2^32` does not fit in `usize`).

## Changes

- Replaced the exhaustion arithmetic with an overflow-safe implementation using `u64`.

## Notes

- Kept the same exhaustion threshold semantics (>3/4 used).
- Preserved the early return behavior for `cid_len == 0` and `cid_len > 4`.
- No behavior change is intended on 64-bit targets.
- Arithmetic is now architecture-independent.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -913,11 +913,18 @@ impl Endpoint {
     /// We leave some space unused so that `new_cid` can be relied upon to finish quickly. We don't
     /// bother to check when CID longer than 4 bytes are used because 2^40 connections is a lot.
     fn cids_exhausted(&self) -> bool {
-        self.local_cid_generator.cid_len() <= 4
-            && self.local_cid_generator.cid_len() != 0
-            && (2usize.pow(self.local_cid_generator.cid_len() as u32 * 8)
-                - self.index.connection_ids.len())
-                < 2usize.pow(self.local_cid_generator.cid_len() as u32 * 8 - 2)
+        let cid_len = self.local_cid_generator.cid_len();
+        if cid_len == 0 || cid_len > 4 {
+            return false;
+        }
+
+        // Keep this architecture-independent: on 32-bit targets, 2usize.pow(32) overflows.
+        let bits = (cid_len * 8) as u32;
+        let space = 1u64 << bits;
+        let reserve = 1u64 << (bits - 2);
+        let len = self.index.connection_ids.len() as u64;
+
+        len > (space - reserve)
     }
 }
 


quinn-proto/src/endpoint.rs
```diff
@@ -913,11 +913,17 @@ impl Endpoint {
...
-                - self.index.connection_ids.len())
-                < 2usize.pow(self.local_cid_generator.cid_len() as u32 * 8 - 2)
+        let cid_len = self.local_cid_generator.cid_len();
+        if cid_len == 0 || cid_len > 4 {
+            return false;
+        }
+
+        // Keep this architecture-independent: on 32-bit targets, 2usize.pow(32) overflows.
+        let bits = (cid_len * 8) as u32;
+        let space = 1u128 << bits;
```
Ralith: Why `u128`? If we know `cid_len` is at most 4, then `u64` should be plenty.

KevinRoebert: I had only considered that `cid_len` might be greater than 4 in some update. However, `u64` is currently sufficient. I have changed the code accordingly.

quinn-proto/src/endpoint.rs
```diff
@@ -913,11 +913,17 @@ impl Endpoint {
...
-                < 2usize.pow(self.local_cid_generator.cid_len() as u32 * 8 - 2)
+        let cid_len = self.local_cid_generator.cid_len();
+        if cid_len == 0 || cid_len > 4 {
+            return false;
+        }
+
+        // Keep this architecture-independent: on 32-bit targets, 2usize.pow(32) overflows.
+        let bits = (cid_len * 8) as u32;
+        let space = 1u64 << bits;
+        let free = space.saturating_sub(self.index.connection_ids.len() as u64);
```
Ralith: Surely this can never saturate? `connection_ids.len()` is at most 2^bits - 1, even if we ignore the limit this code enforces, and `space` is 2^bits.

Ralith (approved):

quinn-proto/src/endpoint.rs
KevinRoebert: Thanks for the hint. That comment also prompted me to simplify and tighten the code a bit.

I rewrote the check to avoid computing `space - len` and instead compare directly against a threshold:

$$
\text{free} = \text{space} - \text{len}
$$

$$
\text{free} < \text{reserve}
$$

$$
\text{space} - \text{len} < \text{reserve}
$$

$$
\text{space} < \text{reserve} + \text{len}
$$

$$
\text{space} - \text{reserve} < \text{len}
$$

$$
\text{len} > \text{space} - \text{reserve}
$$

So we now use `len > (space - reserve)` directly. This is mathematically equivalent, avoids a subtraction with a runtime-dependent value, and makes it easier for the compiler to constant-fold and optimize the expression.

djc:
If you want this on 0.11.x, you'll want to do a backport to the release branch.

PR #2541, opened 2026-02-16
Remove `RecvStream`s from `blocked_readers` on `stop`

Ralith:
Fixes the issue reported in https://github.com/quinn-rs/quinn/pull/2495#issuecomment-3904293203.

### DIFF
--- a/quinn/src/recv_stream.rs
+++ b/quinn/src/recv_stream.rs
@@ -280,6 +280,9 @@ impl RecvStream {
         conn.inner.recv_stream(self.stream).stop(error_code)?;
         conn.wake();
         self.all_data_read = true;
+        // Clean up shared state that might be left over from a cancalled read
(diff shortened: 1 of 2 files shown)


quinn/src/recv_stream.rs
```diff
@@ -280,6 +280,9 @@ impl RecvStream {
         conn.inner.recv_stream(self.stream).stop(error_code)?;
         conn.wake();
         self.all_data_read = true;
+        // Clean up shared state that might be left over from a cancalled read
+        // operation, so `drop` doesn't have to
+        conn.blocked_readers.remove(&self.stream);
```
Ralith, line 285: An alternative strategy would be to ensure every single read future cleans this state up in its own drop, but there's fewer places that set `all_data_read` than there are read futures, so this seems less error prone.

PR #2549, opened 2026-02-27
add metric for spurious congestion events

stormshield-fabs:
This is inspired by recent work seen on `neqo`: we would like to track spurious congestion events through telemetry to understand the congestion controller behavior at scale.


stormshield-fabs:
We're already using the unreleased `0.12` so I don't need a backport, but I can open a similar PR targeting 0.11 if you want me to.

djc:
> We're already using the unreleased `0.12` so I don't need a backport, but I can open a similar PR targeting 0.11 if you want me to.

Nope, no need to for me.

Issue #2544, opened 2026-02-21
quinn-perf table output not suppressed in json mode


Zotyamester:
## Problem
While neither the command-line help nor the comments in the code elaborate on the expected behavior when `--json` is specified in a quinn-perf client's arguments, based on intuition and experience using iperf3, one would assume that enabling it automatically suppresses all text- or table-based output (at least when the JSON output target is stdout, e.g., when specifying `--json -`).

Currently, this is not the case.

## Possible Resolution
Since json output is implemented as an optional compile-time feature, the easiest solution I can imagine for disabling is to add a conditionally included if expression before invoking `stats.print()` and any additional `println!` macros present in [client.rs](https://github.com/quinn-rs/quinn/tree/main/perf/src/client.rs). This feels a bit quirky though.

Ralith:
It looks like the design intent is to write it to a file, rather than stdout? You could use a named pipe if you want to stream it somewhere.

I wouldn't be opposed to a change to make the stdout behavior friendlier. You could avoid too much conditional compilation by having a flag that's syntactically valid in either configuration but is always `false` when json support is disabled, maybe.

zoltanszatmary:
Fair point, thanks!
I was thinking if it ever made sense to have both the table output written to stdout and the JSON dump written to the specified file at the same time, but I guess changing on this behavior would break compatibility for those relying on this tool already.

Ralith:
It's fine to break the CLI interface for quinn-perf.

PR #2550, opened 2026-02-27
Suppress table output in quinn-perf when JSON output targets stdout (#2544)

Zotyamester:
Suppress table output meant to provide human-friendly visualization when using the `--json` CLI option and its argument is `-`, that is, when the JSON output targets stdout.


djc (approved):
Thanks!

PR #2556, opened 2026-03-06
Introduce `max_outgoing_bytes_per_second` option

aochagavia:
cc @Ralith 

This is something we discussed a while ago. It came up in the context of deep space QUIC networking, which requires being able to rate-limit outgoing traffic (traditional congestion control algorithms don't work when an ACK takes minutes to arrive 😅).

### DIFF
--- a/quinn-proto/src/connection/pacing.rs
+++ b/quinn-proto/src/connection/pacing.rs
@@ -17,20 +17,34 @@ pub(super) struct Pacer {
     last_window: u64,
     last_mtu: u16,
     tokens: u64,
+    max_bytes_per_second: Option<u64>,
     prev: Instant,
 }
 
 impl Pacer {
     /// Obtains a new [`Pacer`].
-    pub(super) fn new(smoothed_rtt: Duration, window: u64, mtu: u16, now: Instant) -> Self {
+    pub(super) fn new(
+        smoothed_rtt: Duration,
+        window: u64,
+        mtu: u16,
+        max_bytes_per_second: Option<u64>,
+        now: Instant,
+    ) -> Self {
+        let window = rate_limited_window(smoothed_rtt, window, max_bytes_per_second);
         let capacity = optimal_capacity(smoothed_rtt, window, mtu);
         Self {
             capacity,
             last_window: window,
             last_mtu: mtu,
             tokens: capacity,
+            max_bytes_per_second,
             prev: now,
         }
+    }
+
+    /// Obtains the `max_bytes_per_second` used when this [`Pacer`] was constructed.
+    pub(crate) fn max_bytes_per_second(&self) -> Option<u64> {
+        self.max_bytes_per_second
     }
 
     /// Record that a packet has been transmitted.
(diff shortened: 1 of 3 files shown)


marcblanchet:
fyi. requirement discussed in draft-many-tiptop-quic-profile

quinn-proto/src/connection/pacing.rs
```diff
@@ -147,6 +162,21 @@ fn optimal_capacity(smoothed_rtt: Duration, window: u64, mtu: u16) -> u64 {
     )
 }
 
+/// Clamps the window to achieve rate-limiting, if `max_bytes_per_second` is provided
+fn rate_limited_window(
+    smoothed_rtt: Duration,
+    window: u64,
+    max_bytes_per_second: Option<u64>,
+) -> u64 {
+    if let Some(max_bytes_per_second) = max_bytes_per_second {
```
Ralith: style nit: `let else` to reduce rightward drift?

quinn-proto/src/connection/pacing.rs
```diff
@@ -147,6 +162,21 @@ fn optimal_capacity(smoothed_rtt: Duration, window: u64, mtu: u16) -> u64 {
...
+        // `rate_window` is the window that would yield a throughput of `max_bytes_per_second`
+        let rate_window = max_bytes_per_second as f64 / 1.25 * smoothed_rtt.as_secs_f64();
```
Ralith: Document where the 1.25 factor comes from?

aochagavia:
Looking at the code, it seems like it could also live inside a congestion controller because all we do is manipulate the congestion window. In all honesty, that approach seems cleaner to me than what I've implemented in this PR. What do you think? The only caveat I can think of right now is that the initial congestion window needs to be calculated based on the initial RTT (which is configurable), and the congestion controller doesn't have access to that information.

Ralith:
I like that, as written, this feature will Just Work with any congestion controller. I'm not absolutely certain how broadly useful it is, but there's definitely a number of internet applications where a manual total bandwidth limit is handy.

aochagavia:
In that case, let's keep things as-is :)

quinn-proto/src/connection/pacing.rs
```diff
@@ -17,22 +17,36 @@ pub(super) struct Pacer {
...
             last_window: window,
             last_mtu: mtu,
             tokens: capacity,
+            max_bytes_per_second,
             prev: now,
         }
     }
 
+    /// Obtains the `max_bytes_per_second` used when this [`Pacer`] was constructed.
+    pub(crate) fn get_max_bytes_per_second(&self) -> Option<u64> {
```
djc: Nit: avoid `get_()` prefixes per the API guidelines, please.

djc:
Thanks for your patience!

quinn-proto/src/connection/pacing.rs
aochagavia: Oops, sorry about that

aochagavia:
Thanks for the review ;)

PR #2558, opened 2026-03-09
Avoid unwrapping VarInt decoding during parameter parsing

djc:
Fixes https://github.com/quinn-rs/quinn/security/advisories/GHSA-6xvm-j4wr-6v98.


djc:
Carrying forward earlier private approval from @Ralith.

Issue #2491, opened 2025-12-29
RFC: Tie `MAX_GSO_SEGMENTS` on Windows to socket state


thomaseizinger:
Currently, `MAX_GSO_SEGMENTS` on Windows is wrapped in a `LazyLock` and therefore only evaluates once for the entire lifetime of the program in memory.

In our experience, the support for things like URO/GRO/GSO is tied to the current network driver and thus can change at runtime when the user e.g. plugs their laptop into a dock.

Hence, I believe it would be good to tie this state to the `UdpSocketState` instead to allow applications to re-evaluate this on their own terms. In Firezone for example, we detect when the user roams networks and reset all kinds of network state in order to make things work again.

Thoughts?

djc:
Seems sensible to me.

mxinden:
Would we determine `MAX_GSO_SEGMENTS` on each new socket? If so, isn't that quite expensive?

https://github.com/quinn-rs/quinn/blob/0dcf9d9db774a11e3a30c7cb2b044ed0375bf354/quinn-udp/src/windows.rs#L497-L515

thomaseizinger:
> Would we determine `MAX_GSO_SEGMENTS` on each new socket? If so, isn't that quite expensive?

I think if we do it per socket, there is no need to open an additional socket to check that it works? So it should only be a single syscall which isn't that bad I guess? At least in Firezone, we don't use `quinn-udp` for all our UDP needs, only those that need good UDP IO. DNS queries for example are sent from "regular" sockets.

PR #2564, opened 2026-03-10
fix: evaluate `max_gso_segments` for every socket on Windows

thomaseizinger:
This follows the behaviour we have for UNIX-based systems. Support for GSO varies depending on the network driver which may change at runtime when e.g. a device switches between Ethernet and WiFi or gets plugged into a docking station.

Resolves: https://github.com/quinn-rs/quinn/issues/2491

### DIFF
--- a/quinn-udp/src/windows.rs
+++ b/quinn-udp/src/windows.rs
@@ -4,7 +4,10 @@ use std::{
     net::{IpAddr, Ipv4Addr},
     os::windows::io::AsRawSocket,
     ptr,
-    sync::{LazyLock, Mutex},
+    sync::{
+        LazyLock, Mutex,
+        atomic::{AtomicUsize, Ordering},
+    },
     time::Instant,
 };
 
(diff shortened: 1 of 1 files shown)


quinn-udp/src/windows.rs
```diff
@@ -153,6 +157,7 @@ impl UdpSocketState {
         let now = Instant::now();
         Ok(Self {
             last_send_error: Mutex::new(now.checked_sub(2 * IO_ERROR_LOG_INTERVAL).unwrap_or(now)),
+            max_gso_segments: AtomicUsize::new(max_gso_segments()),
```
mxinden: `max_gso_segments` still opens an additional socket. In #2491 you suggest using the current socket.

> I think if we do it per socket, there is no need to open an additional socket to check that it works?

https://github.com/quinn-rs/quinn/issues/2491#issuecomment-3703041301

In addition to reducing the overhead of socket creation, using the current socket might give us a more accurate GSO segment count, as, I would assume, the count could depend e.g. other socket options that we set.

Thoughts @thomaseizinger?

thomaseizinger: Yeah makes sense! Will update to use the current socket.

thomaseizinger: We should probably do the same thing on UNIX then. There we also open an additional socket.

thomaseizinger: > We should probably do the same thing on UNIX then. There we also open an additional socket.

See #2565.

PR #2565, opened 2026-03-10
fix: reuse existing socket for probing GRO/GSO support

thomaseizinger:
Instead of creating two throw-away sockets for probing GRO & GSO, use the existing socket that gets passed into `UdpSocketState::new`.

Related: #2564

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -135,9 +135,6 @@ impl UdpSocketState {
 
         #[cfg(any(target_os = "linux", target_os = "android"))]
         {
-            // opportunistically try to enable GRO. See gro::gro_segments().
-            let _ = set_socket_option(&*io, libc::SOL_UDP, libc::UDP_GRO, OPTION_ON);
-
             // Forbid IPv4 fragmentation. Set even for IPv6 to account for IPv6 mapped IPv4 addresses.
             // Set `may_fragment` to `true` if this option is not supported on the platform.
             may_fragment |= !set_socket_option_supported(


mxinden:
Do you know why we used a new socket in the past? Are there any scenarios where GSO & GRO probing might move the socket into an inconsistent state?

I could not find a reason in git history, nor GitHub history.

thomaseizinger:
> Do you know why we used a new socket in the past? Are there any scenarios where GSO & GRO probing might move the socket into an inconsistent state?
> 
> I could not find a reason in git history, nor GitHub history.

Nope, that predates my involvement in this project.

djc (approved):
LGTM, let's get a review from @Ralith too.

quinn-udp/src/unix.rs
```diff
@@ -1181,21 +1175,15 @@ mod gso {
...
+    pub(crate) fn gro_segments(socket: &impl AsRawFd) -> usize {
         // As defined in net/ipv4/udp_offload.c
         // #define UDP_GRO_CNT_MAX 64
         //
         // NOTE: this MUST be set to UDP_GRO_CNT_MAX to ensure that the receive buffer size
         // (get_max_udp_payload_size() * gro_segments()) is large enough to hold the largest GRO
         // list the kernel might potentially produce. See
         // https://github.com/quinn-rs/quinn/pull/1354.
-        match set_socket_option(&socket, libc::SOL_UDP, libc::UDP_GRO, OPTION_ON) {
+        match set_socket_option(socket, libc::SOL_UDP, libc::UDP_GRO, OPTION_ON) {
```
Ralith: We're also calling this in `new`. Can we deduplicate? This helper function might not be needed at all.

mxinden:
I would like to release `quinn-udp` `v0.6.0` to unblock https://github.com/quinn-rs/quinn/pull/2463. @thomaseizinger @djc are you fine if I release this pull request in a follow-up release (e.g. `v0.6.1`)? As far as I can tell, this pull request does not introduce a breaking change.

djc:
Seems fine to me.

quinn-udp/src/unix.rs
```diff
@@ -152,6 +154,17 @@ impl UdpSocketState {
                     libc::IPV6_PMTUDISC_PROBE,
                 )?;
             }
+
+            if set_socket_option(&*io, libc::SOL_UDP, libc::UDP_GRO, OPTION_ON).is_ok() {
```
Ralith, line 159: This is also called on line 135. Please deduplicate.

thomaseizinger: Oh! Now I see it, sorry. I guess we can just remove the above invocation then?

Ralith: Yep, that should be fine.

Ralith (approved):

Ralith:
CI issue is legitimate (if trivial).

thomaseizinger:
Should be fixed now!

djc:
Do you want another release for this? If so, please submit a version bump PR.

thomaseizinger:
> Do you want another release for this? If so, please submit a version bump PR.

No rush, happy to depend on it via Git for now.

PR #2566, opened 2026-03-10
fix(quinn-udp): add Apple fast path API stubs for windows and fallback

mxinden:
Mirror the Apple fast datapath APIs to windows.rs and fallback.rs.

See discussion in https://github.com/quinn-rs/quinn/pull/2463#discussion_r2909250172.

### DIFF
--- a/quinn-udp/src/fallback.rs
+++ b/quinn-udp/src/fallback.rs
@@ -115,6 +115,20 @@ impl UdpSocketState {
     pub fn may_fragment(&self) -> bool {
         true
     }
+
+    /// Stub: Apple's fast UDP datapath is not available on this platform.
+    ///
+    /// # Safety
+    ///
+    /// No-op on this platform. Present for API compatibility with the Unix implementation.
+    pub unsafe fn set_apple_fast_path(&self) {}
+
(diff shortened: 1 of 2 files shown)


mxinden:
Turns out, this is a bit more complex than I thought:

In `unix.rs` they are currently gated on `apple_fast`:

https://github.com/quinn-rs/quinn/blob/c67e9d8930c87270c37f142ffab2a8f50b21d50b/quinn-udp/src/unix.rs#L345-L364

`apple_fast` is only `true` on Apple platforms:

https://github.com/quinn-rs/quinn/blob/c67e9d8930c87270c37f142ffab2a8f50b21d50b/quinn-udp/build.rs#L30

Thus e.g. on Linux, the APIs are currently not available.

@thomaseizinger how should we proceed?

1. Make them available on all platforms, including Linux?
2. Make them available on MacOS only (i.e. status quo).

thomaseizinger:
> Make them available on MacOS only (i.e. status quo).

I guess this is the better option then!

mxinden:
:+1: will keep as is then, i.e. closing here.

PR #2570, opened 2026-03-17
docs(quinn): improve `Connection::close_reason()` documentation

zphrs:
Follow-up to https://github.com/quinn-rs/quinn/pull/1781#issuecomment-4072173953.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -386,9 +386,14 @@ impl Connection {
             .clone()
     }
 
-    /// If the connection is closed, the reason why.
+    /// Whether the connection is closed, and why.
     ///
-    /// Returns `None` if the connection is still open.
+    /// The close_reason is always set to `Some(ConnectionError)` when a socket is
+    /// closed; whether it was closed manually by calling [`Connection::close()`] or due to
+    /// an internal error (such as an idle timeout or the peer closing the
+    /// connection).
+    ///
+    /// Note: when the connection is closed, `connection.close_reason().is_some()` will always be true.
     pub fn close_reason(&self) -> Option<ConnectionError> {
         self.0.state.lock("close_reason").error.clone()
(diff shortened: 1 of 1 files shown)


quinn/src/connection.rs
```diff
@@ -386,9 +386,24 @@ impl Connection {
             .clone()
     }
 
-    /// If the connection is closed, the reason why.
+    /// If the connection is still open then it will return `None`.
```
djc: I think the first line here would be more informatively phrased as "Whether the connection is closed, and why".

quinn/src/connection.rs
```diff
@@ -386,9 +386,24 @@ impl Connection {
...
+    /// connection).
+    ///
+    /// Note: this function can be used to synchronously check
+    /// whether the connection has been closed like so:
+    ///
+    /// ```no_run
+    /// # use quinn::Connection;
+    /// # let connection: Connection = unimplemented!();
+    /// let is_closed = connection.close_reason().is_some();
+    /// ```
```
djc: This seems superfluous to me relative to the documentation above.

zphrs: I spent a bit of time looking for an is_closed method myself so I wanted to throw a bone with a snippet to anyone similarly searching through the docs for a way to synchronously tell if a method is closed.

djc: Okay, but I'd prefer for that to be done more concisely. Maybe something like "Detection of an `is_closed` condition can be built on this method" or something? The example feels like it's hitting me over the head with obviousness in the context of this method.

zphrs: How's this?

> Note: `connection.close_reason().is_some()` will always be true when the connection is closed.

zphrs:
Hopefully now it's concise enough!

PR #2571, opened 2026-03-19
fix: Resolve `sendmsg_x`/`recvmsg_x` via `dlsym`

larseggert:
The `fast-apple-datapath` feature previously declared `sendmsg_x` and `recvmsg_x` as hard `extern "C"` symbols. On Apple OS versions where these private APIs don't exist, the dynamic linker would reject the binary at load time.

Replace the `extern "C"` block with lazy `OnceLock`-based resolvers using `dlsym`. Both functions now return `Option<Fn>`, and the call sites in `send_via_sendmsg_x` and `recv_via_recvmsg_x` propagate an `Unsupported` error if the symbol is absent. This allows callers to probe availability at runtime before enabling the fast path.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -33,23 +33,6 @@ pub(crate) struct msghdr_x {
     pub msg_datalen: usize,
 }
 
-#[cfg(apple_fast)]
-extern "C" {
-    fn recvmsg_x(
-        s: libc::c_int,
-        msgp: *const msghdr_x,
-        cnt: libc::c_uint,
-        flags: libc::c_int,
-    ) -> isize;
-
-    fn sendmsg_x(
-        s: libc::c_int,
-        msgp: *const msghdr_x,
-        cnt: libc::c_uint,
-        flags: libc::c_int,
-    ) -> isize;
-}
-
 #[cfg(target_os = "freebsd")]
 type IpTosTy = libc::c_uchar;
 #[cfg(not(any(target_os = "freebsd", target_os = "netbsd")))]
@@ -272,7 +255,7 @@ impl UdpSocketState {
         meta: &mut [RecvMeta],
     ) -> io::Result<usize> {
         if self.is_apple_fast_path_enabled() {
-            recv_via_recvmsg_x(socket.0, bufs, meta)
+            recv_via_recvmsg_x(self, socket.0, bufs, meta)
         } else {
             recv_single(socket.0, bufs, meta)
         }
@@ -375,6 +358,24 @@ impl UdpSocketState {
     #[cfg(apple_fast)]
     pub fn is_apple_fast_path_enabled(&self) -> bool {
         self.apple_fast_path.load(Ordering::Relaxed)
+    }
+
+    /// Disables Apple's fast UDP datapath, reverting to `sendmsg`/`recvmsg`.
+    #[cfg(apple_fast)]
+    fn disable_apple_fast_path(&self) {
+        self.apple_fast_path.store(false, Ordering::Relaxed);
+        self.max_gso_segments.store(1, Ordering::Relaxed);
+    }
+
+    /// Resolves an Apple fast-path function pointer via `resolver`, disabling the fast path if
+    /// the symbol is absent so that future calls use the slow path directly.
+    #[cfg(apple_fast)]
+    fn resolve_apple_fast_fn<T>(&self, resolver: fn() -> Option<T>) -> Option<T> {
+        let f = resolver();
+        if f.is_none() {
+            self.disable_apple_fast_path();
+        }
+        f
     }
 }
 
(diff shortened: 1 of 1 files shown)


quinn-udp/src/unix.rs
```diff
@@ -34,20 +34,36 @@ pub(crate) struct msghdr_x {
...
+    unsafe extern "C" fn(libc::c_int, *mut msghdr_x, libc::c_uint, libc::c_int) -> isize;
+
+/// Returns the `sendmsg_x` function pointer, resolving it via `dlsym` on first call.
+///
+/// Returns `None` if the symbol is not available on the current OS version.
+#[cfg(apple_fast)]
+fn sendmsg_x_fn() -> Option<SendmsgXFn> {
+    static FN: std::sync::OnceLock<Option<SendmsgXFn>> = std::sync::OnceLock::new();
+    *FN.get_or_init(|| {
+        let ptr = unsafe { libc::dlsym(libc::RTLD_DEFAULT, b"sendmsg_x\0".as_ptr().cast()) };
```
Ralith: nit: use `c"..."` literals here and below, which don't need an explicit null terminator and may not need a cast.

quinn-udp/src/unix.rs
```diff
@@ -34,20 +34,36 @@ pub(crate) struct msghdr_x {
...
+        (!ptr.is_null())
+            .then(|| unsafe { std::mem::transmute::<*mut libc::c_void, SendmsgXFn>(ptr) })
```
thomaseizinger: Wouldn't [`as_ref`](https://doc.rust-lang.org/std/primitive.pointer.html#method.as_ref) be a bit cleaner here?

quinn-udp/src/unix.rs
```diff
@@ -510,6 +526,8 @@ fn send_via_sendmsg_x(
         hdrs[i].msg_datalen = chunk.len();
         cnt += 1;
     }
+    let sendmsg_x = sendmsg_x_fn()
+        .ok_or_else(|| io::Error::new(io::ErrorKind::Unsupported, "sendmsg_x not available"))?;
```
thomaseizinger: Similarly to how we reset GSO on EINVAL in Linux, wouldn't it make sense to change the state variable such that the next call will succeed?

larseggert:
Thanks for the feedback, I tried to address it. I also introduced some helpers and applied them where possible (esp. `call_with_eintr_retry`; let me know if that should be factored out?

quinn-udp/src/unix.rs
```diff
@@ -376,6 +400,39 @@ impl UdpSocketState {
...
+    fn resolve_apple_fast_fn<T>(&self, resolver: fn() -> Option<T>) -> Option<T> {
+        let f = resolver();
+        if f.is_none() {
+            self.disable_apple_fast_path();
+        }
+        f
+    }
+}
+
+/// Calls `f` in a loop, retrying on `EINTR`, and returns the non-negative result or the first
```
djc: I would like to see this extraction in a separate commit.

quinn-udp/src/unix.rs
```diff
@@ -376,6 +400,39 @@ impl UdpSocketState {
...
+/// non-`EINTR` error.
+fn call_with_eintr_retry(mut f: impl FnMut() -> isize) -> io::Result<isize> {
```
thomaseizinger: What do you think of this name?

```suggestion
fn retry_if_interrupted(mut f: impl FnMut() -> isize) -> io::Result<isize> {
```

larseggert:
I'll rebase and squash this once #2583 has landed.

quinn-udp/src/unix.rs
```diff
@@ -608,6 +616,47 @@ fn recv_via_recvmsg_x(
     Ok(msg_count as usize)
 }
 
+#[cfg(apple_fast)]
+type SendmsgXFn =
+    unsafe extern "C" fn(libc::c_int, *const msghdr_x, libc::c_uint, libc::c_int) -> isize;
+#[cfg(apple_fast)]
+type RecvmsgXFn =
+    unsafe extern "C" fn(libc::c_int, *mut msghdr_x, libc::c_uint, libc::c_int) -> isize;
```
djc: Nit: these items are used as return values in the functions below, so they should be ordered below them.

djc:
> How does a caller actually probe these experimentally? By sending a test packet? I wonder if we should have something more direct.

Is there anything left to do here?

@mxinden do you want to do a final review?

thomaseizinger:
So with this change, do we even need the feature-flag? Wouldn't it be safe to always try and use these functions when they are available?

larseggert:
We could remove the feature flag. But even if the private functions are present, their ABI may change with different OS versions, so an app does need to handle that (e.g., try to use them in a child to verify they work and deal with that child crashing; or allow-list specific OS versions, etc.)

larseggert:
Could we merge this soon-ish and cut a new point release?

djc:
Want to submit a version bump PR? Also feel free to include version bump commits in PRs where useful.

larseggert:
> Want to submit a version bump PR? Also feel free to include version bump commits in PRs where useful.

For all of quinn or just quinn-udp?

djc:
> > Want to submit a version bump PR? Also feel free to include version bump commits in PRs where useful.
> 
> For all of quinn or just quinn-udp?

Just quinn-udp.

PR #2572, opened 2026-03-19
quinn: Make `Endpoint::server` dual-stack V6 by default

rofferom:
#1913 made `Endpoint::client` dual-stack IPv6 by default, but `Endpoint::server` wasn't updated.

This PR does the same thing as #1913, but for the server side, making both sides consistent.


Ralith (approved):
Thanks!

PR #2573, opened 2026-03-19
docs: clarify that Event::ConnectionLost is not emitted on local close

### LINKED ISSUE #1495: Clarify if `Event::ConnectionLost` is always emitted
The docs for `Event::ConnectionLost` and L113 in the lifecycle test suggest that no such `Event` is emitted on the side closing the connection. However, from the documentation on `is_closed()` and the existence of `ConnectionError::LocallyClosed` I expected otherwise.
Am I confusing something here, or is the documentation a bit misleading?

As a user I think emitting `Event::ConnectionLost` on the closing side would be nice, because it results in nicer code in the driver future (and adding a `Event::ConnectionDrained` would be even nicer).
https://github.com/quinn-rs/quinn/blob/390f2ebeba910ef4c88ed98711f7271a4a530b5d/quinn-proto/src/connection/mod.rs#L1035-L1036
https://github.com/quinn-rs/quinn/blob/390f2ebeba910ef4c88ed98711f7271a4a530b5d/quinn-proto/src/connection/mod.rs#L3199-L3202

https://github.com/quinn-rs/quinn/blob/390f2ebeba910ef4c88ed98711f7271a4a530b5d/quinn-proto/src/connection/mod.rs#L3338-L3339
https://github.com/quinn-rs/quinn/blob/390f2ebeba910ef4c88ed98711f7271a4a530b5d/quinn-proto/src/tests/mod.rs#L92-L118

alexchenai:
## Summary

Fixes #1495.

The documentation for `Event::ConnectionLost`, `Connection::is_closed()`, and `Connection::is_handshaking()` was misleading about when the `ConnectionLost` event is emitted. Specifically:

- `Event::ConnectionLost` doc said "Emitted if the peer closes the connection or an error is encountered" without clarifying that locally-initiated closes do **not** emit this event.
- `is_closed()` doc said "A `ConnectionLost` event is emitted with details when the connection becomes closed" which implied it is always emitted, including for local closes.
- `is_handshaking()` doc referenced `ConnectionLost` without noting the local-close exception.

This PR updates all three doc comments to clearly state that `ConnectionLost` is only emitted for peer-initiated closes and errors/timeouts, and that local closes via `Connection::close()` instead cause pending operations to fail with `ConnectionError::LocallyClosed`.

Also adds proper intra-doc links replacing plain-text references.

## Changes

- `Event::ConnectionLost`: clarified not emitted on local close
- `Connection::is_closed()`: clarified `ConnectionLost` scope, added note about `LocallyClosed`
- `Connection::is_handshaking()`: added note about local-close exception, added intra-doc links

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1306,7 +1306,9 @@ impl Connection {
     /// Whether the connection is in the process of being established
     ///
     /// If this returns `false`, the connection may be either established or closed, signaled by the
-    /// emission of a `Connected` or `ConnectionLost` message respectively.
+    /// emission of a [`Connected`](Event::Connected) or [`ConnectionLost`](Event::ConnectionLost)
+    /// event respectively. Note that locally-initiated closes via [`close()`](Self::close) do not
+    /// emit a `ConnectionLost` event.
     pub fn is_handshaking(&self) -> bool {
         self.state.is_handshake()
     }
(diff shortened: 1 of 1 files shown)


djc:
Thanks!

PR #2579, opened 2026-03-21
proto: send STREAMS_BLOCKED when stream limit is hit  

cong-or:
Addresses a long-standing [TODO](https://github.com/quinn-rs/quinn/blob/b8e4d3b9acf205c5c3304fe901ac40d80408e199/quinn-proto/src/connection/streams/mod.rs#L51) — we now send a STREAMS_BLOCKED frame when we can't open a stream because the peer's concurrency limit is reached ([RFC 9000 §4.6](https://www.rfc-editor.org/rfc/rfc9000#section-4.6)). The frame type already existed but was never actually sent.                                                                                                                                                
                                                                                                                                                                                              
Uses the same staging/retransmit pattern as MAX_STREAMS.


Ralith:
Thanks! Please rebase/squash, and also include a negative test: that we don't send such a frame when streams were opened without becoming blocked.

cong-or:
> This looks good, thanks! Could you add a test?

@Ralith thanks for the review, I added a test that verifies the frame gets sent and received when the stream limit is hit.

Ralith:
Please see my feedback above.

cong-or:
> Please see my feedback above.

@Ralith negative test added.

Issue #2350, opened 2025-08-21
change more transport config after connection is established


FlorianUekermann:
We would like to change some transport config details after the connection is already established. In particular we are interested in:
- `max_idle_timeout`
- `keepalive_interval`
- `pad_to_mtu`

The reason for the first two is that different stages of the connection lifecycle make different settings reasonable (small values make for a responsive experience, but can have a large impact on power consumption on mobile). This can be worked around in most cases on the application level, but only if the user controls the application protocol.

The reason for the latter is that it is mostly a privacy option, which protects to an extent against fingerprinting. This has a significant performance cost in some network environments (Wifi), even if the total transmission size doesn't actually change much. Therefore it would be good to be able to configure this on an already running connection, without adding more entropy for fingerprinting to the handshake (e.g. configuration hints via SNI).

I wanted to ask what your thoughts are on these before working on PRs.

djc:
Seems reasonable to me. I wonder if any of these result in state changes that propagate outward from just the configuration value in ways that we'd have to correct for?

Ralith:
The idle timeout can't be reliably changed on an existing connection because it's part of the transport parameters sent to the peer, the peer is responsible for respecting the value it receives, and no mechanism exists to notify the peer of a change. I don't see any immediate problems with the others, though of course adjusting the keep-alive interval to a value that exceeds the negotiated idle timeout isn't very useful.

FlorianUekermann:
Oh. Interesting. I thought keep-alive / idle-timeout mechanism was just based on emitting something that makes the remote send an ack and tracking time since the last datagram was received. Seems like I need to do some reading.

KirillLykov:
I would like to change keep alive interval on connection level. The motivation is the following -- we have p2p network and each host might have it's own custom idle timeout. So client/server negotiates this parameter to be the minimal of two. Hence, it might happen that `negotiated_idle_timeout < client_transport_config.keep_alive_interval`. I would like to set keep alive to something like 1/2 of negotiated idle timeout. Any ideas if it is possible to do or requires patching quinn?

flub:
On Thu 18 Dec 2025 at 06:05 -0800, kirill lykov wrote:

> KirillLykov left a comment (quinn-rs/quinn#2350)
>
[...]

Issue #2156, opened 2025-02-17
quinn_proto::connection: blocked by congestion control/blocked by pacing


tubzby:
I built a UDP proxy to emulate a network of 20ms delay and no packet lost, sitting between a local QUIC client and QUIC server.
I'm trying to simulate a 4Mbps video stream, using `quinn datagram`, and the data was generated:

1. around 30 times a second, each with a 33ms gap
2. the first frame of each round is 30% * 4Mbps = 1.2Mbps = 150 Kbytes
3. the other frame is (4-1.2)Mbps / (30 -1 ) = 12 Kbytes

I'm using the default `quinn::TransportConfig`, and I record the bytes sent from the client and received on the server, expecting to see it matches.

The process was running normally for about 30 seconds, then I saw only 80% of packets were received on the server, so I'm trying to debug, and found some tracing logs:

``` console
2025-02-16T12:02:21.203843Z TRACE drive{id=0}: quinn_proto::connection: blocked by congestion control
...
2025-02-16T12:02:41.226979Z TRACE drive{id=0}: quinn_proto::connection: blocked by pacing

```

I have investigated a little bit, it seems the window scale of `TransportConfig` should be increased, but I have no idea where the `congestion control` coming from.

Is there anything wrong with my simulation? 4Mbs and 20ms delay is relatively low.

Ralith:
> I saw only 80% of packets were received on the server

Are you losing IP packets or application-layer datagrams?

If there's no IP packet loss, but you are losing application-layer datagrams, then you're likely trying to send data at a higher rate than the congestion controller thinks is possible, and exhausting buffer space as a result. This should automatically stabilize over time, but you could adjust initial conditions by raising [the congestion controller's initial window](https://docs.rs/quinn/0.11.5/quinn/congestion/struct.CubicConfig.html#method.initial_window). Note that this may lead to catastrophic IP packet loss if you overestimate the path's true capacity.

Streaming media might benefit from a slightly different congestion control strategy than cubic, but I think this is an open research area. I think @kixelated has been exploring this as part of their MoQ implementation effort.

kixelated:
Yeah, live media is what is known as "application-limited". It means the congestion controller has trouble increasing the bitrate because there's limited data to send. TCP congestion control algorithms (ex. Reno, CUBIC, BBR) won't increase the congestion window size unless it's being fully utilized. They just were not designed for frame-by-frame delivery.

[...]

PR #2582, opened 2026-03-22
proto: add runtime set_stream_receive_window and FlowControlStats

JavaPerformance:
## Summary

Two related additions:

### 1. `Connection::set_stream_receive_window(VarInt)`

Runtime adjustment of the per-stream receive window. This completes the runtime window adjustment API — `set_receive_window()` (connection-level) and `set_send_window()` already exist.

### 2. `FlowControlStats` on `ConnectionStats`

```rust
pub struct FlowControlStats {
    pub send_credit_remaining: u64,
    pub recv_credit_remaining: u64,
    pub streams_blocked_by_conn_fc: u64,
}
```

### Stream window setter

On high-BDP paths (e.g., 250ms RTT cross-region), the default stream receive window can become a throughput bottleneck. Applications that monitor path characteristics and adapt windows need to adjust both connection-level and stream-level windows at runtime. Connection-level is already possible; stream-level is not.

The implementation is minimal because `stream_receive_window` is stored as a single field on `StreamsState` and passed by value to `max_stream_data()` on each flow control decision. Changing it immediately affects the next `MAX_STREAM_DATA` frame for any stream without iterating per-stream state.

Only expansion is safe: `MAX_STREAM_DATA` is a one-way ratchet per RFC 9000 — previously advertised limits cannot be revoked. The doc comment makes this clear.

### Flow control stats

`FrameStats` already exposes `data_blocked` and `stream_data_blocked` frame counts, which tell you *how many times* the peer was blocked. `FlowControlStats` complements this with real-time utilization:

- `send_credit_remaining` — how much headroom before we hit the peer's `MAX_DATA`
- `recv_credit_remaining` — how much headroom before the peer exhausts our `MAX_DATA`
- `streams_blocked_by_conn_fc` — number of streams currently waiting on connection-level flow control

This enables applications to detect flow control pressure *before* it becomes a bottleneck, rather than only after `DATA_BLOCKED` frames have been sent.

## Changes

- `quinn-proto/src/connection/streams/state.rs` — `FlowControlStats` struct, `set_stream_receive_window()`, `flow_control_stats()`
- `quinn-proto/src/connection/streams/mod.rs` — re-export
- `quinn-proto/src/connection/mod.rs` — expose on `Connection`, populate in `stats()`
- `quinn-proto/src/connection/stats.rs` — add `flow_control` field to `ConnectionStats`
- `quinn-proto/src/lib.rs` — crate-level re-export
- `quinn/src/connection.rs` — public `set_stream_receive_window()`

All new types use `#[non_exhaustive]`. Backwards-compatible.

## Related

- #2350 (change transport config after connection established)
- #2156 (blocked by congestion control vs pacing)

### DIFF
--- a/quinn-proto/src/connection/streams/state.rs
+++ b/quinn-proto/src/connection/streams/state.rs
@@ -890,10 +890,10 @@ impl StreamsState {
     /// Affects the `MAX_STREAM_DATA` limit advertised to the peer for all streams opened after
     /// this call, and for already-open streams on their next flow control update.
     ///
-    /// No explicit `MAX_STREAM_DATA` transmission is triggered because `max_stream_data()` reads
-    /// `stream_receive_window` on every flow control evaluation. The next time data is consumed
-    /// and Quinn checks whether to send a `MAX_STREAM_DATA` update, it will use the new window
-    /// value and advertise accordingly.
+    /// No explicit `MAX_STREAM_DATA` transmission is triggered. The new window takes effect the
+    /// next time data is consumed on a stream and Quinn evaluates whether to send a
+    /// `MAX_STREAM_DATA` update. Idle streams will not see the updated window until they become
+    /// active again.
     ///
     /// Only expansion is safe at the QUIC protocol level. Shrinking does not revoke previously
     /// advertised limits; it only reduces what is advertised on future updates.


quinn-proto/src/connection/streams/state.rs
```diff
@@ -20,6 +20,29 @@ use crate::{
...
 
+/// Snapshot of connection-level flow control utilization
+#[derive(Debug, Default, Copy, Clone)]
+#[non_exhaustive]
+pub struct FlowControlStats {
+    /// Remaining send credit on the connection
+    ///
+    /// This is the peer's advertised `MAX_DATA` limit minus the sum of data offsets across all
+    /// send streams. When this reaches zero, the connection is send-blocked at the flow control
+    /// level and a `DATA_BLOCKED` frame will be emitted.
```
Ralith: > and a `DATA_BLOCKED` frame will be emitted.

This is false. Quinn does not currently implement sending `DATA_BLOCKED`.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -20,6 +20,29 @@ use crate::{
...
+    pub send_credit_remaining: u64,
+    /// Remaining receive credit on the connection
+    ///
+    /// This is our advertised `MAX_DATA` limit minus the sum of data received across all receive
```
Ralith: s/data/max offsets/ to match the precision above, perhaps?

quinn-proto/src/connection/streams/state.rs
```diff
@@ -862,6 +885,30 @@ impl StreamsState {
...
+    /// this call, and for already-open streams on their next flow control update.
+    ///
+    /// The value is not copied into per-stream state. Instead, `max_stream_data()` reads
+    /// `self.stream_receive_window` each time it is called, so updating it here takes effect on
+    /// the next flow control decision for any stream.
+    ///
+    /// Only expansion is safe at the QUIC protocol level. Shrinking does not revoke previously
+    /// advertised limits; it only reduces what is advertised on future updates.
+    pub(crate) fn set_stream_receive_window(&mut self, window: u64) {
+        self.stream_receive_window = window;
```
Ralith, line 901: Should this also trigger transmission of new `MAX_STREAM_DATA` frames where appropriate? If not, document why.

JavaPerformance:
Good question. Two reasons:

  1. Memory budget across many connections. A relay proxy may handle hundreds of connections to backends with very different path characteristics (LAN vs cross-continent). Setting the upper bound to what the worst-case WAN path needs (e.g., 32 MiB) for every connection wastes memory on
  LAN connections that only need 2 MiB. We start conservative and grow per-connection based on observed BDP, so the aggregate memory stays proportional to actual need rather than worst-case.
  2. The initial window advertisement affects the peer immediately. If we configure a 32 MiB stream window up front, the receiver advertises that full credit to the sender at connection start. On paths where the actual BDP is 2 MB, this over-advertises by 16x — the sender can have 32 MiB
   in flight before we can signal backpressure. Starting small and growing based on observed throughput keeps the advertised credit proportional to what the path can actually sustain.

  set_receive_window already exists for the connection-level case and the same reasoning applies at the stream level.

  On the doc comments:

  - Will fix the DATA_BLOCKED claim — thanks for the correction.
  - Will use "max offsets" for precision on the receive side.
  - On whether set_stream_receive_window should trigger MAX_STREAM_DATA transmission: it doesn't need to because max_stream_data() is called on every flow control evaluation and reads the current value of stream_receive_window at that point. The next time the receiver consumes data and
  Quinn checks whether to send a MAX_STREAM_DATA update, it will use the new (larger) window and advertise accordingly. No explicit trigger is needed — it takes effect on the next natural flow control cycle. Will add a note.

Ralith:
> it takes effect on the next natural flow control cycle

That was clear, but why is that necessarily okay? There's no guarantee that flow control for a specific stream will be refreshed on any particular time scale.

Ralith:
> Setting the upper bound to what the worst-case WAN path needs (e.g., 32 MiB) for every connection wastes memory on LAN connections that only need 2 MiB

Where is that memory wasted? These are upper bounds; we shouldn't actually commit the memory if we don't have data to store.

If you're worried about badly behaved peers forcing excess memory consumption with synthetic workloads, such peers could just as well spoof your heuristic by faking high latency.

JavaPerformance:
> There's no guarantee that flow control for a specific stream will be refreshed on any particular time scale.

You're right — if a stream is idle, the new window won't be advertised until data flows again. In our use case this is acceptable because we only grow the window on connections that are actively transferring (we measure BDP from observed goodput). An idle stream doesn't need a larger window advertised.

That said, if the expectation is that `set_stream_receive_window` should take effect promptly for all open streams, it would need to iterate `recv` state and mark streams for `MAX_STREAM_DATA` retransmission — similar to how `set_receive_window` queues `pending.max_data`. Happy to add that if you think it's the right behavior. The implementation would be straightforward since the `Recv` instances don't store the window — they'd just need to be marked for a flow control update on their next check.

> Where is that memory wasted? These are upper bounds; we shouldn't actually commit the memory if we don't have data to store.

Fair point — the receiver doesn't pre-allocate. The cost is on the sender side: a large advertised window lets the sender have more data in flight, which consumes sender-side send buffer memory. But you're right that this is the sender's problem, not the receiver's.

The stronger motivation is the second one: over-advertising credit on a low-BDP path means the sender can burst far more data than the path can sustain in one RTT. That creates large queues at intermediate routers and inflates latency. Starting conservative and growing based on measured path capacity keeps the advertised credit proportional to what the network can absorb without bufferbloat.

That said, for many applications, configuring the maximum tolerable window up front and letting the peer figure out pacing is perfectly fine. The runtime setter is most useful for long-lived connections (hours/days) where path characteristics change, or for relay proxies that handle heterogeneous paths and want to minimize aggregate in-flight data across many connections.

Ralith:
> An idle stream doesn't need a larger window advertised

I'm happy relying on this assumption for simplicity so long as we document it, so there's no uncertainty if we need to revisit it in the future.

> The stronger motivation is the second one: ...

This sounds a lot like trying to implement congestion control at the application layer. Would it make more sense to use the congestion controller for this, and let senders set their own internal buffer sizes according to their preference?

JavaPerformance:
> I'm happy relying on this assumption for simplicity so long as we document it

Will do — I'll add a doc note that the new window only takes effect on the next natural flow control update, and idle streams won't see it until they become active again.

> This sounds a lot like trying to implement congestion control at the application layer.

Fair criticism — that argument was weaker than I made it sound. The congestion controller should handle pacing, you're right.

The simpler and honest motivation: `set_receive_window` already exists for the connection level. This adds the stream-level equivalent for API completeness. Our use case is long-lived relay connections where we want to grow stream windows after observing path characteristics, the same way we already grow connection windows at runtime. That's it.

I'll update the PR description to focus on API completeness rather than the bufferbloat rationale.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -862,6 +885,31 @@ impl StreamsState {
...
+    /// Affects the `MAX_STREAM_DATA` limit advertised to the peer for all streams opened after
+    /// this call, and for already-open streams on their next flow control update.
+    ///
+    /// No explicit `MAX_STREAM_DATA` transmission is triggered because `max_stream_data()` reads
+    /// `stream_receive_window` on every flow control evaluation. The next time data is consumed
+    /// and Quinn checks whether to send a `MAX_STREAM_DATA` update, it will use the new window
+    /// value and advertise accordingly.
+    ///
+    /// Only expansion is safe at the QUIC protocol level. Shrinking does not revoke previously
+    /// advertised limits; it only reduces what is advertised on future updates.
```
Ralith, line 899: This comment is a good description of what the behavior *should* be, but I'm not convinced it's accurate to what the behavior is as written. For example, if a flow control credit update is scheduled for a stream whose receive window has been shrunk, the new value will be computed based on the latest `stream_receive_window` value, when instead it should be the high water mark.

PR #2583, opened 2026-03-23
refactor(quinn-udp): add `retry_if_interrupted` helper

larseggert:
Replaces the repeated EINTR-retry loop pattern in `send_single`, `send_via_sendmsg_x`, `recv_via_recvmmsg`, and `recv_via_recvmsg_x`.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -510,20 +510,8 @@ fn send_via_sendmsg_x(
         hdrs[i].msg_datalen = chunk.len();
         cnt += 1;
     }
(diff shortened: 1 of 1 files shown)


quinn-udp/src/unix.rs
```diff
@@ -378,6 +378,21 @@ impl UdpSocketState {
     }
 }
 
+/// Calls `f` in a loop, retrying on `EINTR`, and returns the non-negative result or the first
```
djc: Nit: should go at the bottom, below all its callers.

Issue #2293, opened 2025-07-06
Should we map ENOBUFS on macOS to `WouldBlock`?


thomaseizinger:
We are seeing the following errors from the UDP socket on some installations on Mac OS:

> Tunnel error: UDP socket failed: Failed to send datagram-batch with segment_size 1312 and total length 7872 to X.X.X.X:57689: No buffer space available (os error 55)

This appears to be the `ENOBUFS` error mentioned in the [man page](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/sendmsg.2.html):

>     [ENOBUFS]         The system is unable to allocate an internal buffer.
>                       The operation may succeed when buffers become avail-able. available.
>                       able.

`libc` for bsd-like systems maps the error code 55 to this error: https://github.com/rust-lang/libc/blob/8e6f36c6ba5c91c150f2d79a06ffdc3d0b83a0f6/src/unix/bsd/freebsdlike/mod.rs#L693

According to the [Apple developer forum](https://developer.apple.com/forums/thread/42334), this is a form of back-pressure and we should retry sending the packet when there are buffers available again.

I figured, we could do this in `quinn-udp` directly and map this error code on macOS to `WouldBlock`? Thoughts?

Ralith:
Will a socket that previously returned `ENOBUFS` be marked as writable for kqueue purposes once buffer space becomes available?

thomaseizinger:
I've found this NetBSD discussion that suggests that this is an acceptable pattern: https://mail-index.netbsd.org/netbsd-users/2012/09/20/msg011462.html

> UDP packets are sent using sendto() directly
> in loops similar to: 
>
>     tries = 0;
>     do {
>       if tries > 0
>          sleep
>       rc = sendto(sock, ...);
>       tries++;
>     } while((rc is EAGAIN || EWOULDBLOCK || ENOBUFS) && tries < 8)

thomaseizinger:
Actually, just noticed the `sleep` in there. That is not what we want!

thomaseizinger:
`libuv` appears to be treating `ENOBUFS` identical to `EWOULDBLOCK`: https://github.com/libuv/libuv/blob/aac866f39923670d9c136851f866531f5f6cd889/src/unix/stream.c#L820

thomaseizinger:
The [manpage](https://man.freebsd.org/cgi/man.cgi?query=kqueue&sektion=2) for kqueue says:

[...]

PR #2594, opened 2026-03-31
Test write readiness after ENOBUFS

mxinden:
Trying to reproduce missing wake-up on `ENOBUFS` discussed in https://github.com/quinn-rs/quinn/issues/2293.

### DIFF
--- a/quinn-udp/Cargo.toml
+++ b/quinn-udp/Cargo.toml
@@ -33,7 +33,8 @@ windows-sys = { workspace = true }
 
 [dev-dependencies]
 criterion = { version = "0.8", default-features = false, features = ["async_tokio"] }
-tokio = { workspace = true, features = ["rt", "rt-multi-thread", "net"] }
+libc = "0.2.175"
+tokio = { workspace = true, features = ["rt", "rt-multi-thread", "net", "time", "macros"] }
 
 [build-dependencies]
 cfg_aliases = { workspace = true }
--- a/quinn-udp/tests/tests.rs
+++ b/quinn-udp/tests/tests.rs
@@ -370,6 +370,111 @@ fn ip_to_v6_mapped(x: IpAddr) -> IpAddr {
     }
 }
 
+/// Reproducer for <https://github.com/quinn-rs/quinn/issues/2293>.
+///
+/// On macOS, `ENOBUFS` ("No buffer space available", os error 55) is returned by
+/// `sendmsg` when the kernel's network buffer pool (mbufs) is exhausted. Unlike
+/// `EAGAIN`/`EWOULDBLOCK` — which signals that the per-socket send buffer is full and
+/// for which the I/O driver reliably fires a WRITE event once buffer space returns —
+/// `ENOBUFS` is a system-wide condition the I/O driver has no direct visibility into.
+///
+/// Mapping `ENOBUFS` to `WouldBlock` is therefore incorrect: after `ENOBUFS`, the
+/// I/O driver may not re-signal writability, so a task waiting for a WRITE event can
+/// suspend indefinitely (confirmed by Firezone in production).
+///
+/// This test verifies the expected behaviour: after `ENOBUFS` is returned and the
+/// receiver has drained its buffer (freeing kernel memory), the tokio reactor IS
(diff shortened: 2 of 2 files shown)


mxinden:
@larseggert I am having a hard time triggering an `ENOBUFS` on a MacOS CI runner. Am I doing something wrong? Does this test trigger on your Mac?

larseggert:
You might not be able to trigger this over loopback.

thomaseizinger:
I was able to easily trigger this by sending packets to a random public IP (pick one from a dark block in https://map.bgp.tools/). The program will simply hang if you map `ENOBUFS` to `WouldBlock` inside `UdpSocket::async_io`. That is a pretty clear reproducer IMO that wakers are not correctly registered in that case.

mxinden:
@thomaseizinger were you able to reproduce with the test provided in this pull request? If no, can you try this test?

I am unable to trigger a `ENOBUFS` in the first place. Both here on CI and on a MacOS M1.

> You might not be able to trigger this over loopback.

I tried loopback, some random IP as well as a Google STUN server. @larseggert do you have any luck?

thomaseizinger:
> @thomaseizinger were you able to reproduce with the test provided in this pull request? If no, can you try this test?

I haven't tried that yet. Currently travelling without my Macbook but can run it next week when I am back. I did reproduce it on a Macbook Air M1 by sending buffers of > 1000 bytes to `1.1.1.1` if I remember correctly.

I didn't use a test at the time but wrote a little binary. Have you tried compiling the test in release mode?

mxinden:
>  I did reproduce it on a Macbook Air M1 by sending buffers of > 1000 bytes to 1.1.1.1 if I remember correctly.

I just tried `1.1.1.1:53` without success.

> Have you tried compiling the test in release mode?

Yes. Can't trigger an `ENOBUFS` in debug nor release mode.

> Currently travelling without my Macbook but can run it next week when I am back.

That would be greatly appreciated :pray:

larseggert:
> I just tried `1.1.1.1:53` without success.

Same here :-( I do get a ton of `WouldBlockdBlock` though.

thomaseizinger:
@mxinden @larseggert I managed to reproduce it again. I changed the test such that it fails for me. Importantly though, this only fails for the WiFi chip in my Macbook Air. If I plug in a Thunderbolt adapter to use Ethernet, it never triggers ENOBUFS.

larseggert:
@thomaseizinger could you share your changes?

djc:
@larseggert see the latest commit?

mxinden:
Thank you @thomaseizinger for the changes! This is helpful.

@larseggert can you either propose a fix, or revert https://phabricator.services.mozilla.com/D288120?

larseggert:
@mxinden see https://github.com/mxinden/quinn/pull/1

thomaseizinger:
FWIW: We also perform retries with a backoff: https://github.com/firezone/firezone/blob/4e608f13acd4866e57fa61b8876af7eb4d3cf185/rust/libs/connlib/socket-factory/src/lib.rs#L327

`backoff` is defined further down.

larseggert:
@thomaseizinger that's where I got the idea from :-)

mxinden:
Regarding https://github.com/mxinden/quinn/pull/1, I don't think `sleep`ing a thread is a suitable option as a UDP IO library, especially a high fixed value like 1ms. I think it is up to the user to make that decision.

In the case of Firefox, we could schedule an async wake up. I don't think sleeping the whole (single threaded) network thread for 1ms is an option.

Backing up a bit, how often do Firefox users actually see `ENOBUFS`? Do we have any data? @larseggert is any of this worth the complexity it introduces?

larseggert:
Well, _you_ wanted the fix in quinn-udp :-) Without changing the API contract on `send`, I don't see a better approach. And it only sleeps for 1ms max.

For neqo, I think treating the error as `WouldBlock` is actually fine, since we will retry.

mxinden:
Closing here. I don't think a solution in quinn-udp is worth the complexity. Summary in https://github.com/quinn-rs/quinn/issues/2293#issuecomment-4480831487.

Thanks everyone for the help!

PR #2601, opened 2026-04-11
Lazy initialize remote streams

### LINKED ISSUE #2600: Eager memory allocation based on MAX_STREAMS
Hey guys,

MoQ uses a stream for each subscription/request similar to HTTP/3, but it also uses any number of streams to actually deliver subscriptions. You could feasibly create/close 60 streams a second for an audio track, one for each frame so they can be delivered independently. I have relay nodes that proxy arbitrary subscriptions between themselves and to handle this, I set MAX_STREAMS to 10k.

I had a memory leak on my server preventing `Connection` objects from getting dropped, even after the `Connection` was closed. While that was a separate issue, is raised the fact that Quinn immediately allocates a HashMap containing all 10k streams allowed. A new connection involves a MB of allocations, even before any stream data is sent.

Basically, each Stream ID can be in one of four states:
- Closed
- Active
- Allowed
- Not Allowed

Currently, the HashMap contains all Active/Allowed streams. If a stream is not present in the HashMap, it means it's been closed, or it's not allowed based on MAX_STREAMS.

I'd like to change it so we avoid inserting `Allowed` streams into the HashMap. Instead we keep track of a `max_active_id` or something that is used to determine if a missing hash map entry means Closed, or if it means Allowed. I'm working on a PR.

kixelated:
Fixes #2600.

Previously, we would insert into the send/recv HashMaps based on our advertised MAX_STREAMS amount. So if we establish a connection with MAX_STREAMS_UNI=10k and MAX_STREAMS_BI=10k, we'll end up allocating 30k HashMap entries. Obviously this takes up a lot of RAM, even if the connection only uses a handful of streams.

Now we only insert into the HashMap based on the maximum stream ID ever received. If the maximum we've seen is 3 and we get a stream for 6, then we'll insert 4..=6 into the HashMap. Senders typically open stream IDs in ascending order (although it's not a requirement) so the usual case means we insert once. An attacker could open the maximum stream ID allowed to cause the old behavior... but again that's the current behavior, and the whole point of MAX_STREAMS is meant to limit this resource exhaustion attack.

The reason why we can't insert only 6 into the HashMap is because we need some way of tracking what streams have ever been closed. With this change, we're using HashMap removal as the Closed state, but only for streams below next_remote (7 in this example). We would need some separate HashSet to store closed (or not closed) streams which just kind of defeats the entire purpose of inserting into the HashMap in the first place? IDK unless somebody can think of an alternative.

### DIFF
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -55,7 +55,7 @@ impl<'a> Streams<'a> {
 
         self.state.next[dir as usize] += 1;
         let id = StreamId::new(self.state.side, dir, self.state.next[dir as usize] - 1);
-        self.state.insert(false, id);
+        self.state.insert_local(id);
         self.state.send_streams += 1;
         Some(id)
     }
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -457,7 +457,9 @@ struct PendingStream {
 /// Application events about streams
 #[derive(Debug, PartialEq, Eq)]
 pub enum StreamEvent {
-    /// One or more new streams has been opened and might be readable
+    /// One or more new streams has been opened and can be accepted
+    ///
+    /// A separate [`StreamEvent::Readable`] is emitted once data arrives.
     Opened {
         /// Directionality for which streams have been opened
         dir: Dir,
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -457,7 +457,10 @@ struct PendingStream {
 /// Application events about streams
 #[derive(Debug, PartialEq, Eq)]
 pub enum StreamEvent {
-    /// One or more new streams has been opened and might be readable
+    /// One or more new streams has been opened and can be accepted
+    ///
+    /// This event does not imply that the new streams have data available; a separate
+    /// [`StreamEvent::Readable`] is emitted when data (or a reset) actually arrives.
     Opened {
         /// Directionality for which streams have been opened
         dir: Dir,
--- a/quinn-proto/src/connection/streams/state.rs
+++ b/quinn-proto/src/connection/streams/state.rs
@@ -151,7 +151,7 @@ impl StreamsState {
         receive_window: VarInt,
         stream_receive_window: VarInt,
     ) -> Self {
-        let mut this = Self {
+        Self {
             side,
             send: FxHashMap::default(),
             recv: FxHashMap::default(),
@@ -184,15 +184,7 @@ impl StreamsState {
             initial_max_stream_data_bidi_remote: 0u32.into(),
             receive_window_shrink_debt: 0,
             streams_blocked: [false, false],
-        };
-
-        for dir in Dir::iter() {
-            for i in 0..this.max_remote[dir as usize] {
-                this.insert(true, StreamId::new(!side, dir, i));
-            }
         }
-
-        this
     }
 
     pub(crate) fn set_params(&mut self, params: &TransportParameters) {
@@ -202,10 +194,11 @@ impl StreamsState {
         self.max[Dir::Bi as usize] = params.initial_max_streams_bidi.into();
         self.max[Dir::Uni as usize] = params.initial_max_streams_uni.into();
         self.received_max_data(params.initial_max_data);
-        for i in 0..self.max_remote[Dir::Bi as usize] {
-            let id = StreamId::new(!self.side, Dir::Bi, i);
-            if let Some(s) = self.send.get_mut(&id).and_then(|s| s.as_mut()) {
-                s.max_data = params.initial_max_stream_data_bidi_local.into();
+        for (&id, slot) in self.send.iter_mut() {
+            if id.initiator() != self.side && id.dir() == Dir::Bi {
+                if let Some(s) = slot.as_mut() {
+                    s.max_data = params.initial_max_stream_data_bidi_local.into();
+                }
             }
         }
     }
@@ -215,10 +208,6 @@ impl StreamsState {
     fn ensure_remote_streams(&mut self, dir: Dir) {
         let new_count = self.max_concurrent_remote_count[dir as usize]
             .saturating_sub(self.allocated_remote_count[dir as usize]);
-        for i in 0..new_count {
-            let id = StreamId::new(!self.side, dir, self.max_remote[dir as usize] + i);
-            self.insert(true, id);
-        }
         self.allocated_remote_count[dir as usize] += new_count;
         self.max_remote[dir as usize] += new_count;
     }
@@ -263,6 +252,9 @@ impl StreamsState {
             debug!("received illegal STREAM frame");
         })?;
 
+        // Create state for this stream if the remote peer created it.
+        let newly_created = self.ensure_remote(id);
+
         let Some(rs) = self
             .recv
             .get_mut(&id)
@@ -282,7 +274,11 @@ impl StreamsState {
         self.data_recvd = self.data_recvd.saturating_add(new_bytes);
 
         if !rs.stopped {
-            self.on_stream_frame(true, id);
+            // Newly opened streams are inherently readable; the app discovers them via
+            // `StreamEvent::Opened` and a separate `Readable` would be redundant.
+            if !newly_created {
+                self.events.push_back(StreamEvent::Readable { id });
+            }
             return Ok(ShouldTransmit(false));
         }
 
@@ -313,6 +309,9 @@ impl StreamsState {
             debug!("received illegal RESET_STREAM frame");
         })?;
 
+        // Create state for this stream if the remote peer created it.
+        let newly_created = self.ensure_remote(id);
+
         let Some(rs) = self
             .recv
             .get_mut(&id)
@@ -339,8 +338,11 @@ impl StreamsState {
             // Stopped streams should be disposed immediately on reset
             let rs = self.recv.remove(&id).flatten().unwrap();
             self.stream_recv_freed(id, rs);
+        } else if !newly_created {
+            // Newly opened streams are inherently readable; the app discovers them via
+            // `StreamEvent::Opened` and a separate `Readable` would be redundant.
+            self.events.push_back(StreamEvent::Readable { id });
         }
-        self.on_stream_frame(!stopped, id);
 
         // Update connection-level flow control
         Ok(if bytes_read != final_offset.into_inner() {
@@ -357,6 +359,9 @@ impl StreamsState {
     /// Process incoming `STOP_SENDING` frame
     #[allow(unreachable_pub)] // fuzzing only
     pub fn received_stop_sending(&mut self, id: StreamId, error_code: VarInt) {
+        // Create state for this stream if the remote peer created it.
+        self.ensure_remote(id);
+
         let max_send_data = self.max_send_data(id);
         let Some(stream) = self
             .send
@@ -369,7 +374,6 @@ impl StreamsState {
         if stream.try_stop(error_code) {
             self.events
                 .push_back(StreamEvent::Stopped { id, error_code });
-            self.on_stream_frame(false, id);
         }
     }
 
@@ -632,24 +636,6 @@ impl StreamsState {
         stream_frames
     }
 
-    /// Notify the application that new streams were opened or a stream became readable.
-    fn on_stream_frame(&mut self, notify_readable: bool, stream: StreamId) {
-        if stream.initiator() == self.side {
-            // Notifying about the opening of locally-initiated streams would be redundant.
-            if notify_readable {
-                self.events.push_back(StreamEvent::Readable { id: stream });
-            }
-            return;
-        }
-        let next = &mut self.next_remote[stream.dir() as usize];
-        if stream.index() >= *next {
-            *next = stream.index() + 1;
-            self.opened[stream.dir() as usize] = true;
-        } else if notify_readable {
-            self.events.push_back(StreamEvent::Readable { id: stream });
-        }
-    }
-
     pub(crate) fn received_ack_of(&mut self, frame: frame::StreamMeta) {
         let mut entry = match self.send.entry(frame.id) {
             hash_map::Entry::Vacant(_) => return,
@@ -750,6 +736,9 @@ impl StreamsState {
             ));
         }
 
+        // Create state for this stream if the remote peer created it.
+        self.ensure_remote(id);
+
         let write_limit = self.write_limit();
         let max_send_data = self.max_send_data(id);
         if let Some(ss) = self
@@ -775,7 +764,6 @@ impl StreamsState {
             ));
         }
 
-        self.on_stream_frame(false, id);
         Ok(())
     }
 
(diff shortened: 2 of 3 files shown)


quinn-proto/src/connection/streams/mod.rs
```diff
@@ -457,7 +457,10 @@ struct PendingStream {
 /// Application events about streams
 #[derive(Debug, PartialEq, Eq)]
 pub enum StreamEvent {
-    /// One or more new streams has been opened and might be readable
+    /// One or more new streams has been opened and can be accepted
+    ///
+    /// This event does not imply that the new streams have data available; a separate
+    /// [`StreamEvent::Readable`] is emitted when data (or a reset) actually arrives.
```
kixelated: This change isn't needed, but simplifies the logic a little bit. I can revert it if desired.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -184,15 +184,7 @@ impl StreamsState {
             initial_max_stream_data_bidi_remote: 0u32.into(),
             receive_window_shrink_debt: 0,
             streams_blocked: [false, false],
-        };
-
-        for dir in Dir::iter() {
-            for i in 0..this.max_remote[dir as usize] {
-                this.insert(true, StreamId::new(!side, dir, i));
```
kixelated, line 191: This is the big memory allocation btw.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -632,24 +630,6 @@ impl StreamsState {
...
-            if notify_readable {
-                self.events.push_back(StreamEvent::Readable { id: stream });
-            }
-            return;
-        }
-        let next = &mut self.next_remote[stream.dir() as usize];
-        if stream.index() >= *next {
-            *next = stream.index() + 1;
-            self.opened[stream.dir() as usize] = true;
-        } else if notify_readable {
```
kixelated, line 648: This `else if` is why I changed it so `StreamEvent::Readable` follows `StreamEvent::Opened`. It's a little easier to always push a Readable event if `notify_readable`, instead of making it conditional.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -632,24 +630,6 @@ impl StreamsState {
...
-    /// Notify the application that new streams were opened or a stream became readable.
-    fn on_stream_frame(&mut self, notify_readable: bool, stream: StreamId) {
-        if stream.initiator() == self.side {
-            // Notifying about the opening of locally-initiated streams would be redundant.
-            if notify_readable {
-                self.events.push_back(StreamEvent::Readable { id: stream });
-            }
-            return;
-        }
-        let next = &mut self.next_remote[stream.dir() as usize];
```
kixelated, line 644: `insert_remote` does this logic now. Callers were inlined to push a Readable event if `notify_readable`, otherwise it's a no-op.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -892,19 +874,51 @@ impl StreamsState {
...
+    /// for index N implicitly opens indices `0..N` of the same type, so we must materialize
+    /// tombstones for the skipped indices — otherwise a later out-of-order frame for one of
+    /// them would land on "absent from map" and get dropped as closed.
+    fn insert_remote(&mut self, id: StreamId) {
+        let dir_idx = id.dir() as usize;
+
+        // If we initiated this stream, nothing to do
+        if id.initiator() == self.side
+            // If this stream is larger than the max allowed, return.
+            // NOTE: STREAM/RESET_STREAM already enforces this, however STOP_SENDING/MAX_STREAM_DATA do not
```
kixelated, line 908: Quinn currently ignores `STOP_SENDING` and `MAX_STREAM_DATA` frames referencing a stream ID above MAX_STREAMS, instead of returning a protocol violation. Technically a bug but who cares.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1880,10 +1894,163 @@ mod tests {
...
+            (1024 * 1024u32).into(),
+        );
+        assert!(client.recv.is_empty());
+        assert!(client.send.is_empty());
+        assert_eq!(client.recv.capacity(), 0);
+        assert_eq!(client.send.capacity(), 0);
+    }
+
+    #[test]
+    fn out_of_order_implicit_open() {
```
kixelated, line 1933: Note Claude wrote these tests but they seem fine?

quinn-proto/src/tests/mod.rs
```diff
@@ -670,10 +674,11 @@ fn zero_rtt_rejection() {
...
     assert_eq!(s, s2);
 
+    // The server never successfully decrypted the 0-RTT STREAM frame (its 0-RTT keys are
+    // discarded on rejection) and the client does not retransmit it, so the server has no
+    // state for `s2`. Under lazy allocation, absence from the map means `ClosedStream`.
     let mut recv = pair.server_recv(server_ch, s2);
-    let mut chunks = recv.read(false).unwrap();
-    assert_eq!(chunks.next(usize::MAX), Err(ReadError::Blocked));
-    let _ = chunks.finalize();
+    assert_eq!(recv.read(false).err(), Some(ReadableError::ClosedStream));
```
kixelated, line 677: Hmm doesn't seem ideal? Lemme investigate if we can return Blocked again?

kixelated: Okay so I could use some guidance.

New behavior:
- stream_id < next_remote: Blocked
- stream_id < max_remote: ClosedStream
- stream_id >= max_remote: ClosedStream

Old behavior:
- stream_id < next_remote: Blocked
- stream_id < max_remote: Blocked
- stream_id >= max_remote: ClosedStream

Both are kind of wrong. I think Quinn should return an error if we ever try to operate on a stream that has never been opened by the remote?
Before my PR, quinn-proto allows (?) sending, reseting, stopping streams that are supposed to be remotely initiated provided they're below MAX_STREAMS. But I imagine most QUIC implementations would choke on this... they haven't even opened the stream?

For example, if the peer is Quinn, it will return a STREAM_STATE_ERROR and close the connection apparently. There are is_local_unopened checks for this.

After my PR, you'll get ClosedStream instead which is technically wrong because the stream was never opened, but at least nothing gets sent over the network.

Ralith: The documentation for `ReadableError::ClosedStream` says:
> The stream **has not been opened** or was already stopped, finished, or reset

So, the new behavior seems correct to me.

quinn-proto/src/connection/streams/mod.rs
Ralith: I don't particularly mind the change, but I'd slightly prefer for it to be made in a separate commit. Feel free to leave as-is if that would take significantly more effort.

quinn-proto/src/connection/streams/state.rs
Ralith: Overlooking that kind of a thing tends to make it easier for more serious bugs in the peer (Quinn or otherwise) to go unnoticed, so IMO it's worth having, though irrelevant to this PR. Open an issue?

Ralith:
Thanks! The design intent of this change makes sense to me, though I haven't yet completed a detailed review.

> We would need some separate HashSet to store closed (or not closed) streams which just kind of defeats the entire purpose of inserting into the HashMap in the first place?

It might be slightly smaller, I guess? Pretty marginal benefit in return for code complexity though.

kixelated:
> Thanks! The design intent of this change makes sense to me, though I haven't yet completed a detailed review.
> 
> > We would need some separate HashSet to store closed (or not closed) streams which just kind of defeats the entire purpose of inserting into the HashMap in the first place?
> 
> It might be slightly smaller, I guess? Pretty marginal benefit in return for code complexity though.

I reverted it, it's just an extra conditional in a few places to avoid queuing the event.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -892,19 +880,53 @@ impl StreamsState {
...
 
-    pub(super) fn insert(&mut self, remote: bool, id: StreamId) {
-        let bi = id.dir() == Dir::Bi;
-        // bidirectional OR (unidirectional AND NOT remote)
-        if bi || !remote {
-            assert!(self.send.insert(id, None).is_none());
-        }
-        // bidirectional OR (unidirectional AND remote)
-        if bi || remote {
+    /// Insert `(id, None)` tombstones for a locally-initiated stream into `send` (and `recv`
```
Ralith, line 883: "tombstone" to me means "marker to indicate something was destroyed", which doesn't seem right here. Maybe "placeholder"? Arises in a few other places below as well.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -892,19 +880,53 @@ impl StreamsState {
...
+        debug_assert_eq!(id.initiator(), self.side);
+        assert!(self.send.insert(id, None).is_none());
+        if id.dir() == Dir::Bi {
             let recv = self.free_recv.pop();
             assert!(self.recv.insert(id, recv).is_none());
         }
     }
 
+    /// Advance the `next_remote` frontier to cover `id` and insert `(tid, None)` tombstones
+    /// for every index skipped along the way.
```
Ralith, line 895: We seem to frequently call this with local stream IDs as well; let's document the behavior for that case explicitly.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -892,19 +880,53 @@ impl StreamsState {
...
+    ///
+    /// Called at the top of each remote receive path. RFC 9000 §3.2 says receiving a frame
+    /// for index N implicitly opens indices `0..N` of the same type, so we must materialize
+    /// tombstones for the skipped indices — otherwise a later out-of-order frame for one of
+    /// them would land on "absent from map" and get dropped as closed.
+    /// Returns `true` if stream `id` was newly created (i.e. not previously opened).
+    fn insert_remote(&mut self, id: StreamId) -> bool {
```
Ralith, line 902: Since we frequently call this with already-known stream IDs, I think "ensure" might be clearer than "insert".

quinn-proto/src/connection/streams/state.rs
```diff
@@ -892,19 +880,53 @@ impl StreamsState {
...
+            || id.index() >= self.max_remote[dir_idx]
+            // If this stream has already been opened, nothing to do
+            || id.index() < self.next_remote[dir_idx]
+        {
+            return false;
+        }
+
+        // Create all of the streams between the largest opened and this stream.
+        for i in self.next_remote[dir_idx]..=id.index() {
+            let tid = StreamId::new(!self.side, id.dir(), i);
```
Ralith, line 918: What does the "t" stand for?

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1880,10 +1902,163 @@ mod tests {
...
+            10_000u32.into(),
+            10_000u32.into(),
+            1024 * 1024,
+            (1024 * 1024u32).into(),
+            (1024 * 1024u32).into(),
+        );
+        assert!(client.recv.is_empty());
+        assert!(client.send.is_empty());
+        assert_eq!(client.recv.capacity(), 0);
+        assert_eq!(client.send.capacity(), 0);
```
Ralith, line 1929: Document why we're specifically interested in capacity, in addition to `is_empty`.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1880,10 +1902,163 @@ mod tests {
...
+        // Verify the data actually landed on stream 3.
+        let id = StreamId::new(Side::Server, Dir::Uni, 3);
+        let mut pending = Retransmits::default();
+        let mut recv = RecvStream {
+            id,
+            state: &mut client,
+            pending: &mut pending,
+        };
+        let mut chunks = recv.read(true).unwrap();
+        assert_eq!(chunks.next(4).unwrap().unwrap().bytes.len(), 4);
```
Ralith, line 1973: Does the 4 here need to match the 4 in the received frame? If so, use a shared constant to make that intention clear.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1880,10 +1902,163 @@ mod tests {
...
+                    id: StreamId::new(Side::Server, Dir::Uni, 3),
+                    offset: 0,
+                    fin: true,
+                    data: Bytes::from_static(&[0; 4]),
+                },
+                4,
+            ),
+            Ok(ShouldTransmit(false))
+        );
+        // Verify the data actually landed on stream 3.
```
Ralith, line 1964: How do we know we're not getting the data from stream 5 here?

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1880,10 +1902,163 @@ mod tests {
...
+        );
+        // Drain the stream so it's fully freed.
+        let mut pending = Retransmits::default();
+        let mut recv = RecvStream {
+            id,
+            state: &mut client,
+            pending: &mut pending,
+        };
+        let mut chunks = recv.read(true).unwrap();
+        assert_eq!(chunks.next(4).unwrap().unwrap().bytes.len(), 4);
```
Ralith, line 2004: Avoid magic numbers, as above.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1880,10 +1902,163 @@ mod tests {
...
+                    id,
+                    offset: 0,
+                    fin: true,
+                    data: Bytes::from_static(&[0; 4]),
+                },
+                4,
+            ),
+            Ok(ShouldTransmit(false))
+        );
+        // Drain the stream so it's fully freed.
```
Ralith, line 1996: Would it be clearer/simpler to stop the stream rather than draining it?

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1880,10 +1902,163 @@ mod tests {
...
+    #[test]
+    fn churn_keeps_maps_bounded() {
+        // Rapidly open + fully close a long sequence of remote streams. The maps must stay
+        // bounded (only active streams are materialized) even though thousands of ids have
+        // been used over the connection's lifetime.
+        let mut client = make(Side::Client);
+        const N: u64 = 5_000;
+        for i in 0..N {
+            // Give ourselves room as we churn through ids.
+            client.set_max_concurrent(Dir::Uni, (i + 200).try_into().unwrap());
```
Ralith, line 2034: Why are we calling this on every iteration, rather than once up front? Why do we need this at all, when the loop appears to only ever have a single stream open at a time?

quinn-proto/src/tests/mod.rs
```diff
@@ -670,10 +670,11 @@ fn zero_rtt_rejection() {
     let s2 = pair.client_streams(client_ch).open(Dir::Uni).unwrap();
     assert_eq!(s, s2);
 
+    // The server never successfully decrypted the 0-RTT STREAM frame (its 0-RTT keys are
+    // discarded on rejection) and the client does not retransmit it, so the server has no
+    // state for `s2`. Under lazy allocation, absence from the map means `ClosedStream`.
```
Ralith, line 675: This behavior is semantically correct, and indeed publicly documented, so we shouldn't justify it with the lazy allocation implementation detail.

Ralith (commented):
Implementation logic LGTM on first pass; mostly cosmetic issues. Would appreciate a careful second set of eyes on this though.

Ralith:
As usual, please squash commits that revise earlier commits in the same PR.

quinn-proto/src/connection/streams/state.rs
```diff
@@ -1880,10 +1900,169 @@ mod tests {
...
+    #[test]
+    fn churn_keeps_maps_bounded() {
+        // Rapidly open + fully close a long sequence of remote streams. The maps must stay
+        // bounded (only active streams are materialized) even though thousands of ids have
+        // been used over the connection's lifetime.
+        const N: u64 = 5_000;
+
+        let mut client = make(Side::Client);
+        // Give ourselves enough room up front to accept every id we're about to churn through.
+        client.set_max_concurrent(Dir::Uni, N.try_into().unwrap());
```
Ralith, line 2037: Why is this needed, given that the loop below seems to drain each stream as soon as it's opened?

Ralith (approved):

kixelated:
I think this is ready, although I haven't tested outside of running the unit tests.

quinn-proto/src/connection/streams/state.rs
Ralith: Opened https://github.com/quinn-rs/quinn/issues/2650.

Ralith (approved):
LGTM (though please rebase rather than merging when updating)

kixelated:
> LGTM (though please rebase rather than merging when updating)

heh sorry not used to force pushing

djc (commented):
I think the direction is good. It feels like the single commit is too large too confidently review for me -- I think it could be split into a few smaller changes that make it more obvious where/how previous processing steps are moved.

Like, maybe start with a commit that splits `insert()` into `insert_local()` and `ensure_remote()`, have a commit that disperses the logic from `on_stream_frame()`, etc.

kixelated:
> At 500 connections, edge reached 548 MiB RSS, 126 MiB swap, and a 597 MiB high-water mark.
>
> Jemalloc measured roughly 697 MiB of live heap growth. About 88% came from Quinn stream-state hash tables. The relay advertises 10,000 concurrent streams, and Quinn eagerly creates 10,000 send plus 20,000 receive entries per connection: roughly 15 million entries at 500 connections. That explains the memory alert.

😢 working on splitting up this PR now @djc

Issue #1772, opened 2024-03-04
Connection Migration Questions


levaitamas:
@nemethf and I have been trying to migrate a client connection. It seems, based on the [example code](https://github.com/quinn-rs/quinn/blob/main/quinn/examples/client.rs#L118), one just need to rebind the client endpoint to a new UDP socket. Is this the recommended approach? 

Furthermore,  [RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#section-9.1) allows the client to validate a path even before migrating the connection. In `quinn` is it possible to validate a path before rewriting the endpoint address? This would allow us to migrate only when the server can be reached from the new address.

Thanks!

Ralith:
Migration is typically involuntary, e.g. due to a user switching networks. We don't currently have an API to spontaneously probe a new path, but it could probably be implemented. Can you elaborate on your use case for programmatic migration?

Ralith:
Elaborating a little: typically you'd bind your client endpoint to a wildcard IP, which will cause involuntary migration whenever the host's default route changes, e.g. due to moving between wifi networks and/or a cellular modem.

nemethf:
There is time for pre-validation when the user switches to a preferred network.  For example, a wifi connection might be cheaper than cellular one, or a wired connection might be faster and more reliable than a wifi connection.  Even switching from wifi to cellular can be scheduled if the application has access to the wifi signal strength.

We are experimenting with real-time traffic where delay jitter should be kept low.  Hopefully, pre-validation results in a less noticeable connection migration.

Thanks again.

Ralith:
> This feature seems quite complicated for us since we are new to quinn (and to some extent to rust as well.)

No worries, and no pressure; it's just likely to be the fastest path to seeing it happen. We're always happy to answer questions and provide feedback.

> a new `paths: Vec<PathData>` field to `Connection`.

I'm wary of extending `Connection` to track multiple concurrently valid paths in the scope of this task. That starts looking like #224, which is likely a much larger undertaking, albeit a desirable one. Off the cuff, I think a simpler approach might be to have a table of paths that we're actively probing (maybe just a `Slab<ConnectionId>`?), with entries being added on `probe_path` calls and removed on timeout or successful receipt. The application could then decide to attempt actual migration, or not, based on the results. A further refinement might be to pass through the probe results to the migration attempt (e.g. marking the path as already-validated, providing a better initial RTT guess, choice of CID) but that isn't required for a MVP.

[...]

PR #2603, opened 2026-04-12
Don't use initial server connection ID longer than necessary.

Kazik24:
I was analyzing what connection ids are transmitted when running `bench` and saw that the initial server connection ID (one assigned after the client sends us first packet with random cid) is used during the whole connection, which makes it easier for a passive observer to identify a specific connection.

Tested with default bench params (just `cargo run` in bench dir)

The connection proceeds as follows:
1. Client sends first initial packet with `src_cid: <client_cid>` and dst_cid: <some_random_value>
2. Server continues handshake but replaces `<some_random_value>` with `<server_cid>`, server sends packet with long header and `src_cid: <server_cid>` and `dst_id: <client_cid>` (observer can see both ids and link them together)
3. Handshake continues ..
4. Server sends new identifiers to client with `<new_server_id>`
5. Client sends new_identifiers to server with `<new_client_id>`
6. Client sends short header packets with `dst_id: <server_cid>` and Server sends short header packets with `dst_id: <client_cid>`
7. Server starts sending packets with `dst_id: <new_client_id>` and `<client_cid>` is retired.
---
8. <server_cid> is never retired during this connection. Passive observer knows that `<server_cid>` is linked to `<client_cid>` and `<some_random_value>` so they know when client sends data for this connection to server (probably not forever cause later ids may change).

The fix is pretty simple one line change to update remote CID not only in server code.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -2981,9 +2981,9 @@ impl Connection {
                         }
                     };
 
-                    if self.side.is_server() && self.rem_cids.active_seq() == 0 {
-                        // We're a server still using the initial remote CID for the client, so
-                        // let's switch immediately to enable clientside stateless resets.
+                    if self.rem_cids.active_seq() == 0 {
+                        // We're a server still using the initial remote CID for the client or
+                        // server, so let's switch immediately to enable clientside stateless resets.
                         self.update_rem_cid();
                     }
                 }


Kazik24:
Looks like tests are failing, I'll check it.

Ralith:
What's the threat model for which this makes a difference? You're still using a fixed ID, just one assigned very slightly later. Even if you changed ID for every packet, an observer can identify the flow using the 4-tuple unless a migration happens, and per the RFC Quinn already changes CIDs when migrating.

Kazik24:
IMO it's slightly inconsistent that `client_cid` is rotated just after the handshake, but `server_cid` is not. I wanted to make sure that all CIDs after handshake were exchanged in encrypted portion of a packet. It's ok that Im using fixed CID for latter packets if those CIDs were exchanged undercover. By using 4-tuple you can identify peers but I don't think you should be able to identify different connections between the same peers.

My use case (maybe a little unrelated) is that I'm trying to see if it's possible to build a relay for QUIC without wrapping it in any other protocol and also making it harder for any eavesdropper to fingerprint connections.

Ralith:
> it's slightly inconsistent

The reason for this is given in the comment. It has nothing to do with traceability.

> make sure that all CIDs after handshake were exchanged in encrypted portion of a packet

Active CIDs are always transmitted in the clear -- that's their entire purpose. If the observer you're worried about could see the server's initial CID, they can just as well see any other one you start using on the same path. Changing CIDs without changing paths has no impact on traceability whatsoever.

> I don't think you should be able to identify different connections between the same peers.

That is not achieved by this change, because a connection will still use a fixed CID on any given path once established. An observer who, for some strange reason, cares about the initial CID specifically could trivially link them by looking at packet timing.

> making it harder for any eavesdropper to fingerprint connections

It sounds like what you actually want is to change CIDs regularly, not just once. As noted above, this is only useful if you do so when packet timing does not make the effort moot. RFC9000 § 9.5 notes:

> A client might wish to reduce linkability by switching to a new connection ID, source UDP port, or IP address (see [[RFC8981](https://www.rfc-editor.org/rfc/rfc9000.html#RFC8981)]) when sending traffic **after a period of inactivity**.

(emphasis added)
I'd be open to that behavior, but as written this PR does not appear to solve any problem.

Kazik24:
> It sounds like what you actually want is to change CIDs regularly, not just once.

Yeah, I agree that would be better proposal than the one mentioned in this PR. 
Perhaps public API to force CID rotation can be added. Then application could decide if it wants to rotate first server_cid or any other CID. Config param to change CID after some configured inactivity time could also be good.
I'll toy with it on my spare time.

Ralith:
Sounds good!

I agree that some sort of public API could make sense, since that would make changing IP address/UDP port possible as well. Maybe as part of an explicit migration API (see e.g. discussion in https://github.com/quinn-rs/quinn/issues/1772). That wouldn't handle the "after inactivity" part, though. I wonder if we could somehow expose inactivity periods and delegate that whole concern to the application, rather than baking the complexity in?

In any case, closing this PR since we're agreed that another direction makes more sense, but please feel free to open a new one for one of those directions, and/or a broader issue for further discussion.

PR #2609, opened 2026-04-16
quinn: fix ref count logic for ConnectionRef and EndpointRef

stormshield-damiend:
PR #2495 broke the ref count logic when moving from non atomic usize to atomic usize.

We found it in one of our test where we did the following usage pattern on an Endpoint create / use / drop / create another one / use / drop.

When the first Endpoint is dropped, its driver continued to run for some seconds and accepted(network) and refused(proto) some connections.

After the fix, the driver is stopped as soon as the last ref is dropped. (Validated by us)

The ConnectionRef drop also has the same logic, we have no test to validate it.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -932,7 +932,7 @@ impl Clone for ConnectionRef {
 
 impl Drop for ConnectionRef {
     fn drop(&mut self) {
-        if self.shared.ref_count.fetch_sub(1, Ordering::Relaxed) > 0 {
+        if self.shared.ref_count.fetch_sub(1, Ordering::Relaxed) > 1 {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


djc:
Sorry, looking at the diff it's not clear to me why this fixes anything? Can you add comments to clarify what invariants this is maintaining that the previous approach did not?

stormshield-damiend:
Before the Atomic rework, the code was using a check_sub(1) and check that rc == 0 after the sub(1) to wake up the driver.

After the Atomic rework, the code is now using the value before the sub(1) (as it uses fetch_sub(1)), thus when dropping the last reference the driver is never woken up. In our test tokio wakes it because of spurious wake up and the driver stops.

djc:
Okay, can you keep the early return structure so that that change is more obvious and we can still avoid the rightward drift?

djc (approved):
I forget, are you tracking `main` or would you like a 0.11.x release with this?

stormshield-damiend:
We are tracking main, we will just rebase our internal fork after merge. Question is more if #2495 is in a release ? but this is your call.

djc:
See

- #2610 

for CI fixes.

Ralith (approved):
Great catch, thanks!

PR #2612, opened 2026-04-17
Test cleanup of connections/endpoints when all references are dropped

Ralith:
Some attempt to prevent further regressions per #2609. Spurious wakeups mean these aren't totally reliable, but at least `dropped_connection_cleans_up` seems to detect the bug in practice.

Depends on #2611


djc (approved):

djc:
(You know you can stack PRs by setting the base to your parent branch, right?)

Ralith:
Yeah, but that gets really confusing if the child PR gets merged first.

PR #2616, opened 2026-04-21
Improve 0-RTT usability

Ralith:
Some breaking API tweaks inspired by #2613.

IMO this is a more discoverable and simpler API. As detailed in commit messages, a stream-independent `ZeroRttAccepted` signal was difficult to use correctly beyond trivial cases.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -16,7 +16,11 @@ use bytes::Bytes;
 use pin_project_lite::pin_project;
 use rustc_hash::FxHashMap;
 use thiserror::Error;
-use tokio::sync::{Notify, futures::Notified, mpsc, oneshot};
+use tokio::sync::{
+    Notify,
+    futures::{Notified, OwnedNotified},
+    mpsc, oneshot,
+};
 use tracing::{Instrument, Span, debug_span};
 
 use crate::{
@@ -36,7 +40,7 @@ use proto::{
 #[derive(Debug)]
 pub struct Connecting {
     conn: Option<ConnectionRef>,
-    connected: oneshot::Receiver<bool>,
+    connected: Pin<Box<OwnedNotified>>,
     handshake_data_ready: Option<oneshot::Receiver<()>>,
 }
 
@@ -50,7 +54,6 @@ impl Connecting {
         runtime: Arc<dyn Runtime>,
     ) -> Self {
         let (on_handshake_data_send, on_handshake_data_recv) = oneshot::channel();
-        let (on_connected_send, on_connected_recv) = oneshot::channel();
 
         let conn = ConnectionRef(Arc::new(ConnectionInner {
             state: Mutex::new(State::new(
@@ -59,12 +62,12 @@ impl Connecting {
                 endpoint_events,
                 conn_events,
                 on_handshake_data_send,
-                on_connected_send,
                 sender,
                 runtime.clone(),
             )),
             shared: Shared::default(),
         }));
+        let connected = Box::pin(conn.shared.connected.clone().notified_owned());
 
         let driver = ConnectionDriver(conn.clone());
         runtime.spawn(Box::pin(
@@ -78,7 +81,7 @@ impl Connecting {
 
         Self {
             conn: Some(conn),
-            connected: on_connected_recv,
+            connected,
             handshake_data_ready: Some(on_handshake_data_recv),
         }
     }
@@ -88,7 +91,7 @@ impl Connecting {
     /// Returns `Ok` immediately if the local endpoint is able to attempt sending 0/0.5-RTT data.
     /// If so, the returned [`Connection`] can be used to send application data without waiting for
     /// the rest of the handshake to complete, at the cost of weakened cryptographic security
-    /// guarantees. The returned [`ZeroRttAccepted`] future resolves when the handshake does
+    /// guarantees. The [`Connection::authenticated`] future resolves when the handshake does
     /// complete, at which point subsequently opened streams and written data will have full
     /// cryptographic protection.
     ///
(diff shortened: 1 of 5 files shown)


quinn/src/recv_stream.rs
```diff
@@ -286,10 +286,15 @@ impl RecvStream {
...
+    /// Check if this stream predates completion of the handshake on an incoming connection.
     ///
-    /// In which case any non-idempotent request should be considered dangerous at the application
-    /// level. Because read data is subject to replay attacks.
+    /// True only if the stream was accepted before the handshake completed, which is only possible
+    /// if you successfully called [`Connecting::into_0rtt`](crate::Connecting::into_0rtt) and the
+    /// client chose to send 0-RTT data.
+    ///
+    /// Under those conditions, depending on cryptographic layer configuration, received stream data
+    /// may be a malicious replay attack. Data vulnerable to replays must not trigger non-idempotent
```
djc, line 296: The previous wording wasn't great, but I don't love "may be a malicious replay attack" either. Maybe something like "might have been replayed from an earlier connection by an unauthenticated network peer"?

Maybe also somehow switch the order of the sentences in this paragraph to bring the "must not trigger non-idempotent operations" upfront?

quinn/src/connection.rs
```diff
@@ -695,6 +695,31 @@ impl Connection {
...
+    /// Completes when the connection becomes fully established or lost.
+    ///
+    /// Only interesting for `Connection`s obtained from [`Connecting::into_0rtt`]. On 1-RTT
+    /// connections, always completes immediately.
+    ///
+    /// For incoming connections, reads from [`RecvStream`]s are guaranteed not to arise from replay
+    /// attacks after this succeeds, even for streams accepted or read during 0-RTT. For outgoing
+    /// connections, streams opened after this succeeds will never be discarded by the server due to
+    /// 0-RTT rejection.
+    pub async fn connected(&self) -> Result<(), ConnectionError> {
```
djc, line 708: Maybe `connected()` is not the right name for this? In many ways, once you have a `Connection` you're connected... how about `handshake_complete()` or `handshake_completed()`?

Really it feels like a 0-RTT stage `Connection` should be a different type with different Rust types for 0-RTT-originated streams, but maybe that would need too much boilerplate?

djc:
Nice work!

quinn/src/recv_stream.rs
Ralith: > Maybe something like "might have been replayed from an earlier connection by an unauthenticated network peer"?

This seems like equivocation about the peer's malice. Is that what you had in mind? I think it's useful to highlight the security hazard here.

quinn/src/connection.rs
Ralith: > Maybe connected() is not the right name for this?

I guess that comes down to what "connected" really means, and how we reconcile that meaning with API terms. A 0-RTT capable client will give you a `Connection` before ever hearing back from the server, which does not intuitively (or per RFC) seem like a connected state to me; the server may not even exist any more. However, I'm open to the argument that defining "connected" to mean anything other than "in possession of a `Connection` object" is confusing, although error states already undermine that somewhat.

I also worry that "handshake" is a less obvious term, although there's a fair argument that we shouldn't be trying to funnel non-experts into 0-RTT. Another possibility is "established".

> it feels like a 0-RTT stage Connection should be a different type with different Rust types for 0-RTT-originated streams

While it would be nice to get `ZeroRttRejected` out of the error enums in 1-RTT, I suspect it's useful for stream types to erase this distinction: if you have a "do a request" function, it allows you to kick it off whether or not you're in 0-RTT. A similar principle holds for logic that might use a `Connection` to send datagrams or use future 0-RTT-friendly extensions.

I agree there's a footgun here in that it makes it easier to carelessly trigger non-idempotent work, but even that's permissible for services which reliably prevent token reuse. We could support both cases by allowing a 0-RTT resource to be upcast to an RTT-generic resource, but that indeed seems like a lot of effort and complexity for very marginal benefit.

quinn/src/recv_stream.rs
djc: I think my objection to "malicious" is that it's superfluous when you're talking about an "attack"? I agree we should be highlighting the security hazard here and I actually think "replay attack" is the right wording since it brings up useful references when searched, but "received stream data may be a malicious replay attack" feels awkward.

quinn/src/connection.rs
djc: > I guess that comes down to what "connected" really means, and how we reconcile that meaning with API terms. A 0-RTT capable client will give you a `Connection` before ever hearing back from the server, which does not intuitively (or per RFC) seem like a connected state to me; the server may not even exist any more. However, I'm open to the argument that defining "connected" to mean anything other than "in possession of a `Connection` object" is confusing, although error states already undermine that somewhat.
> 
> I also worry that "handshake" is a less obvious term, although there's a fair argument that we shouldn't be trying to funnel non-experts into 0-RTT. Another possibility is "established".

I feel like "established" is pretty similar to "connected". What about something more specific to what we're trying to say here? Like `fresh_keys()` or `authenticated()` or `keys_exchanged()` or `key_exchange_complete()`?

We should put pointers on 0-RTT enablement API that point to this method.

> I agree there's a footgun here in that it makes it easier to carelessly trigger non-idempotent work, but even that's permissible for services which reliably prevent token reuse. We could support both cases by allowing a 0-RTT resource to be upcast to an RTT-generic resource, but that indeed seems like a lot of effort and complexity for very marginal benefit.

Yup.

quinn/src/recv_stream.rs
Ralith: Fair point; reworded.

quinn/src/connection.rs
Ralith: Went with `authenticated`. I worry that might be confused with `handshake_confirmed`, but I made an effort to guard against that with docs.

> We should put pointers on 0-RTT enablement API that point to this method.

I did update `Connecting::into_0rtt`. Any others?

quinn/src/connection.rs
```diff
@@ -589,6 +589,33 @@ impl Connection {
...
 
+    /// Completes when the connection becomes fully established or lost.
+    ///
+    /// Only interesting for `Connection`s obtained from [`Connecting::into_0rtt`]. On 1-RTT
+    /// connections, always completes immediately. Contrast
+    /// [`handshake_confirmed`](Self::handshake_confirmed), which waits longer on clients e.g. to
+    /// confirm client authentication.
+    ///
+    /// For incoming connections, reads from [`RecvStream`]s are guaranteed not to arise from replay
+    /// attacks after this succeeds, even for streams accepted or read during 0-RTT. For outgoing
```
djc, line 591: Is this technically entirely true? Isn't this a possible scenario?

- 0-RTT stream frames arrive
- connection establishment completes
- `authenticates()` becomes `Ready`
- pre-establishment completion stream `read()` yields frames from before establishment

quinn/src/connection.rs
```diff
@@ -589,6 +589,33 @@ impl Connection {
             .clone_box()
     }
 
+    /// Completes when the connection becomes fully established or lost.
```
djc: Nit: maybe this should be more aligned with the current name?

djc (approved):

quinn/src/connection.rs
djc: > I did update `Connecting::into_0rtt`. Any others?

I think that's probably sufficient.

djc:
(audit failure should fall away if you rebase on `main`.)

quinn/src/connection.rs
Ralith: In fact, that's probably the most common case, because Quinn's default configuration gives you servers that will accept 0-RTT but not yield it to the application until the handshake is authenticated. That's why the "even for ..." clause calls it out specifically.

Unless I'm very confused, the key point here is that once the handshake is authenticated, you know *retroactively* that any 0-RTT data was legitimate. I believe our current logic won't even mark a stream as 0-RTT if you accept it after handshake authentication, even if it was created by 0-RTT packets.

quinn/src/connection.rs
Ralith: Reworded.

quinn/src/connection.rs
djc: Ah yes, I think that makes sense.

Issue #2617, opened 2026-04-22
I'd like to propose refactoring ProtectedHeader::decode to be generic over the inner buffer type


cooladj:
Hello, I'd like to propose refactoring ProtectedHeader::decode to be generic over the inner buffer type:

    pub fn decode<B: AsRef<[u8]>>(
    buf: &mut io::Cursor<B>,
    cid_parser: &(impl ConnectionIdParser + ?Sized),
    supported_versions: &[u32],
    grease_quic_bit: bool,
) -> Result<Self, PacketDecodeError>

Currently the signature requires &mut io::Cursor<BytesMut>, which forces external callers who have a borrowed &[u8] (e.g., from a UDP recv buffer) to allocate and copy into a BytesMut just to parse the header.
The body of decode doesn't rely on any BytesMut-specific behavior every method it uses (get, get_var, position, remaining, advance) is available on Cursor<B> whenever B: AsRef<[u8]>. BytesMut, &[u8], Vec<u8>, etc. all satisfy this bound, so existing callers would continue to work unchanged.

Before I put up a PR, I wanted to check:

Is there a reason the signature is concrete on BytesMut that I'm missing? I couldn't find anywhere in the body that actually needs it.

see PR: https://github.com/quinn-rs/quinn/pull/2618

Ralith:
It's concrete because we haven't previously needed it to be generic -- as a general rule, we try not to abstract things unless we need to. Do you plan to add callsites?

PR #2618, opened 2026-04-22
refactored ProtectedHeader::decode to be generic over the inner buffer type

cooladj:
https://github.com/quinn-rs/quinn/issues/2617

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2233,9 +2233,9 @@ checksum = "f87165f0995f63a9fbeea62b64d10b4d9d8e78ec6d7d51fb2125fda7bb36788f"
 
 [[package]]
 name = "rustls-webpki"
-version = "0.103.12"
+version = "0.103.13"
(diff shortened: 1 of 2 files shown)


djc (approved):
Makes sense to me. I think we could use impl Trait as the generic, which would be a little nicer IMO?

Ralith (approved):
Change is fine with me, but this PR shouldn't update Cargo.lock.

cooladj:
> Change is fine with me, but this PR shouldn't update Cargo.lock.

The Cargo.lock change was made because rustls-webpki was failing the security audit. I updated the package so CI would pass should that be in a separate PR?

djc:
Please make two commits with clear commit messages: one for your code change, and one for the lockfile update to upgrade past the security issue.

cooladj:
> Makes sense to me. I think we could use impl Trait as the generic, which would be a little nicer IMO?

Yeah, you’re right going to make that change right now.

djc:
Thanks!

PR #2620, opened 2026-04-23
Lazily allocate remote stream state

stablebits:
Previously, every stream the peer was permitted to open was eagerly inserted into the send/recv maps. Allocate entries only when streams are actually created, and track how far state has been allocated via a new `initialized_remote_count`.

This optimisation matters when peers are granted permissive MAX_STREAMS quotas that many connections never fully utilise. Eager insertion imposes the worst-case allocation cost on every new connection, which is especially costly when connections arrive under high throughput load.


djc:
Did you see #2601?

stablebits:
> Did you see #2601?

ups, no, I haven't seen it. So it's already been reviewed, nice! I'll take a look at it and test it in our setup.

PR #2625, opened 2026-04-27
Fix DATAGRAM send buffer accounting

zerox80:
`Datagrams::send` did not account for the size of the datagram being queued when `drop` was true.

The first commit only moves the existing queue pruning into a helper. The second commit changes the helper to make room for the datagram being queued and keeps the capacity check overflow-safe.

Tested with:

```text
cargo fmt --check
cargo test -p quinn-proto --lib
cargo test -p quinn-proto datagram
```

### DIFF
--- a/quinn-proto/src/connection/datagrams.rs
+++ b/quinn-proto/src/connection/datagrams.rs
@@ -38,9 +38,11 @@ impl Datagrams<'_> {
         if drop {
             self.conn
                 .datagrams
-                .make_space_for(self.conn.config.datagram_send_buffer_size);
-        } else if self.conn.datagrams.outgoing_total + data.len()
-            > self.conn.config.datagram_send_buffer_size
+                .make_space_for(data.len(), self.conn.config.datagram_send_buffer_size);
+        } else if !self
+            .conn
+            .datagrams
+            .has_send_buffer_space(data.len(), self.conn.config.datagram_send_buffer_size)
         {
             self.conn.datagrams.send_blocked = true;
             return Err(SendDatagramError::Blocked(data));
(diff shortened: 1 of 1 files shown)


quinn-proto/src/connection/datagrams.rs
```diff
@@ -140,6 +134,16 @@ impl DatagramState {
         Ok(was_empty)
     }
 
+    fn drop_until_send_buffer_has_space(&mut self, incoming_len: usize, send_buffer_size: usize) {
```
Ralith: This name is very wordy. Maybe something like `make_space_for`?

Ralith: "incoming" is a confusing term to describe an outgoing datagram. Maybe `datagram_len`?

quinn-proto/src/connection/datagrams.rs
```diff
@@ -140,6 +134,16 @@ impl DatagramState {
...
+        while self.outgoing_total + incoming_len > send_buffer_size {
```
Ralith: Account for overflow.

quinn-proto/src/connection/datagrams.rs
Ralith: Please use separate commits for changing behavior and moving code to a new location. That makes it much easier to review the behavior changes.

Ralith (commented):
The current behavior wasn't an accident (it over-allocates by at most a constant amount), but I don't mind dialing it in.

Ralith:
> I have not run the full test suite locally.

Why not?

zerox80:
Will improve the commit soon, thanks for feedback

zerox80:
Updated this now. The helper is make_space_for, the length argument is datagram_len, and the capacity check uses checked addition so it does not overflow. I also split the branch into one refactor commit and one behavior commit, and updated the test notes in the PR body.

djc:
> I also split the branch into one refactor commit and one behavior commit, and updated the test notes in the PR body.

Thanks, but in that case please rebase your branch instead of merging.

djc:
I don't think there is much point in keeping this separate from

- #2626 

so let's handle all review there.

PR #2626, opened 2026-04-27
Reject DATAGRAMs larger than the send buffer

zerox80:
Follow-up to #2625.

A DATAGRAM larger than `datagram_send_buffer_size` cannot fit in the outgoing send buffer by itself. This returns `TooLarge` before trying either send mode, so the behavior is the same for `drop = true` and `drop = false`.

Tested with:

```text
cargo fmt --check
cargo test -p quinn-proto --lib
cargo test -p quinn-proto datagram
```

### DIFF
--- a/quinn-proto/src/connection/datagrams.rs
+++ b/quinn-proto/src/connection/datagrams.rs
@@ -38,7 +38,10 @@ impl Datagrams<'_> {
         }
         if drop {
             self.conn.datagrams.make_space_for(send_buffer_size);
-        } else if self.conn.datagrams.outgoing_total + data.len() > send_buffer_size
+        } else if !self
+            .conn
+            .datagrams
+            .has_send_buffer_space(data.len(), send_buffer_size)
         {
             self.conn.datagrams.send_blocked = true;
             return Err(SendDatagramError::Blocked(data));
@@ -139,6 +142,10 @@ impl DatagramState {
             trace!(len = prev.data.len(), "dropping outgoing datagram");
             self.outgoing_total -= prev.data.len();
         }
+    }
+
+    fn has_send_buffer_space(&self, datagram_len: usize, send_buffer_size: usize) -> bool {
+        self.outgoing_total + datagram_len <= send_buffer_size
     }
 
     /// Discard outgoing datagrams with a payload larger than `max_payload` bytes
--- a/quinn-proto/src/connection/datagrams.rs
+++ b/quinn-proto/src/connection/datagrams.rs
@@ -33,7 +33,7 @@ impl Datagrams<'_> {
             .max_size()
             .ok_or(SendDatagramError::UnsupportedByPeer)?;
         let send_buffer_size = self.conn.config.datagram_send_buffer_size;
-        if data.len() > max {
+        if data.len() > Ord::min(max, send_buffer_size) {
             return Err(SendDatagramError::TooLarge);
         }
         if drop {
--- a/quinn-proto/src/connection/datagrams.rs
+++ b/quinn-proto/src/connection/datagrams.rs
@@ -32,11 +32,12 @@ impl Datagrams<'_> {
         let max = self
             .max_size()
             .ok_or(SendDatagramError::UnsupportedByPeer)?;
+        let send_buffer_size = self.conn.config.datagram_send_buffer_size;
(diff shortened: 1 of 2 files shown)


quinn-proto/src/connection/datagrams.rs
```diff
@@ -36,16 +36,13 @@ impl Datagrams<'_> {
...
-            while self.conn.datagrams.outgoing_total > self.conn.config.datagram_send_buffer_size {
-                let prev = self
-                    .conn
-                    .datagrams
-                    .outgoing
-                    .pop_front()
-                    .expect("datagrams.outgoing_total desynchronized");
-                trace!(len = prev.data.len(), "dropping outgoing datagram");
-                self.conn.datagrams.outgoing_total -= prev.data.len();
+            if data.len() > self.conn.config.datagram_send_buffer_size {
```
Ralith: Is the behavior correct when `!drop`?

zerox80:
Updated this too. The size check now happens before the drop branch, so drop = false returns TooLarge for a datagram that cannot fit in the send buffer. I added a test for both send modes.

zerox80:
What is the current state?

quinn-proto/src/connection/datagrams.rs
```diff
@@ -143,6 +145,12 @@ impl DatagramState {
         }
     }
 
+    fn has_send_buffer_space(&self, datagram_len: usize, send_buffer_size: usize) -> bool {
+        self.outgoing_total
+            .checked_add(datagram_len)
+            .is_some_and(|total| total <= send_buffer_size)
```
djc: Nit: in this case, I think it would be easier to follow something like this:

```
let Some(total) = self.outgoing_total.checked_add(datagram_len) else {
     return false;
};

total <= send_buffer_size
```

quinn-proto/src/connection/datagrams.rs
```diff
@@ -35,6 +35,9 @@ impl Datagrams<'_> {
         if data.len() > max {
             return Err(SendDatagramError::TooLarge);
         }
+        if data.len() > self.conn.config.datagram_send_buffer_size {
```
djc: Suggest changing the previous comparison to `> Ord::min(max, self.conn.config.datagram_send_buffer_size)`.

djc (commented):
This looks okay to me, thanks.

Perhaps you'll want to backport it to the 0.11.x branch, too?

zerox80:
> This looks okay to me, thanks.
> 
> Perhaps you'll want to backport it to the 0.11.x branch, too?

Good point, yes, this should also be backported to 0.11.x. I checked that the same send buffer logic is still present there. I’ll address the small nits here first, and once this lands on main I’ll open a separate backport PR against 0.11.x

quinn-proto/src/connection/datagrams.rs
```diff
@@ -143,6 +145,12 @@ impl DatagramState {
         }
     }
 
+    fn has_send_buffer_space(&self, datagram_len: usize, send_buffer_size: usize) -> bool {
```
djc, line 149: I think this extraction should also be in a separate commit.

quinn-proto/src/connection/datagrams.rs
```diff
@@ -38,9 +38,11 @@ impl Datagrams<'_> {
         if drop {
```
djc, line 39: Let's have a local binding `let send_buffer_size = self.conn.config.datagram_send_buffer_size;` to make things a bit more concise, please introduce it in a separate (first) commit.

djc (commented):
Please squash the changes from review into their originating commits.

zerox80:
Done, thanks. I rewrote the stack so the review changes are squashed into their originating commits, added the `send_buffer_size` local binding as the first commit, and kept the `has_send_buffer_space` extraction as its own separate commit.

Tested with:
cargo fmt --check
cargo test -p quinn-proto --lib
cargo test -p quinn-proto datagram

djc (approved):
LGTM, thanks for the iterations!

Ralith (approved):
LGTM, thanks for the highly reviewable commit history!

zerox80:
I forgot the backport @djc

it's coming soon

PR #2631, opened 2026-05-04
qlog: emit RTT values in milliseconds

marcblanchet:
currently, fn to_qlog_event emit RTT values within the qlog data structure as secs, while draft-ietf-quic-qlog-main-schema-13.txt says:
"   All timestamps and time-related values (e.g., offsets) in qlog are
   logged as float64 in the millisecond resolution."

The current qlog event timestamp is correctly emitted in ms.


djc (approved):

djc:
Thanks!

Ralith (approved):

PR #2632, opened 2026-05-05
feat: Add OpenSSL support and update crypto provider configuration

qzhuyan:
Add system OpenSSL (and rustls-openssl) as a pluggable crypto backend

- Introduced OpenSSL as a new cryptography backend.
- Refactored crypto provider logic to support OpenSSL alongside existing backends.
- Modified various modules to utilize the new OpenSSL-based implementations.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -19,15 +19,21 @@ jobs:
         # they have complex build requirements:
         #     https://github.com/aws/aws-lc/blob/3263ce2a553e4e917217fb487f8c6f488fcb1866/BUILDING.md#build-prerequisites
         #
+        # On Windows, OpenSSL is installed but the Rust `openssl` crate needs OPENSSL_DIR to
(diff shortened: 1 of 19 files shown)


djc:
Why do you want this?

qzhuyan:
@djc  thanks for asking. I am trying to port quinn no/IO to resource constrained devices that flash space matters. 

I need to shrink the size of build binary, by using these compile flag that links to the openssl , the `ring` could be decoupled from `quinn` which reduce the size from 2.4MB to 1.5MB, saved around 900KB (in noio setup).

Another reason is we have requirement to link to the openssl for future extension. 

My goal is to keep the default unchanged, just intro the cargo feature flag to customise it.  WDYT?  Is it interesting?

djc:
There has been some effort by @ctz to port our non-TLS crypto code to the rustls crypto provider API directly. I'd definitely prefer to get that in first so we wouldn't have to deal with OpenSSL API directly in this crate.

PR #2634, opened 2026-05-07
feat: implement kernel receive timestamps on Linux/Android

### LINKED ISSUE #2574: `SO_TIMESTAMPNS`
It would be neat to enable `SO_TIMESTAMPNS` for platforms that support it. See https://docs.kernel.org/networking/timestamping.html

syszery:
Closes #2574.

Add Linux/Android support for kernel receive timestamps via `SO_TIMESTAMPNS`.

- Enable `SO_TIMESTAMPNS` on socket setup
- Parse `SCM_TIMESTAMPNS` ancillary messages
- Expose timestamp via `RecvMeta::timestamp` (`Option<u64>`, nanoseconds since Unix epoch)
- Increase `CMSG_LEN` from 88 to 96 bytes on Linux/Android

Tested on Linux x86_64. All tests pass locally.

### DIFF
--- a/quinn-udp/src/lib.rs
+++ b/quinn-udp/src/lib.rs
@@ -27,16 +27,14 @@
 #![warn(unreachable_pub)]
 #![warn(clippy::use_self)]
 
+use core::time::Duration;
 use std::net::{IpAddr, Ipv6Addr, SocketAddr};
 #[cfg(unix)]
 use std::os::unix::io::AsFd;
 #[cfg(windows)]
 use std::os::windows::io::AsSocket;
 #[cfg(not(wasm_browser))]
-use std::{
-    sync::Mutex,
-    time::{Duration, Instant},
-};
+use std::{sync::Mutex, time::Instant};
 
 #[cfg(apple_fast)]
 mod apple_fast;
@@ -124,6 +122,10 @@ pub struct RecvMeta {
     pub dst_ip: Option<IpAddr>,
     /// The interface index of the interface on which the datagram was received
     pub interface_index: Option<u32>,
+    /// Kernel receive timestamp as Unix epoch
+    ///
+    /// Populated on platforms: Linux, Android.
+    pub timestamp: Option<Duration>,
 }
 
 impl Default for RecvMeta {
@@ -136,6 +138,7 @@ impl Default for RecvMeta {
             ecn: None,
             dst_ip: None,
             interface_index: None,
+            timestamp: None,
         }
     }
 }
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -9,7 +9,7 @@ use std::{
         Mutex,
         atomic::{AtomicBool, AtomicUsize, Ordering},
     },
-    time::Instant,
+    time::{Duration, Instant},
 };
 
 use socket2::SockRef;
@@ -135,6 +135,12 @@ impl UdpSocketState {
                 // list the kernel might potentially produce. See
                 // https://github.com/quinn-rs/quinn/pull/1354.
                 gro_segments = 64
+            }
+
+            if let Err(_err) =
+                set_socket_option(&*io, libc::SOL_SOCKET, libc::SO_TIMESTAMPNS, OPTION_ON)
+            {
+                crate::log::debug!("Ignoring error setting SO_TIMESTAMPNS on socket: {_err:?}");
             }
 
             if is_ipv4 || !io.only_v6()? {
@@ -725,6 +731,7 @@ pub(crate) fn decode_recv<M: cmsg::MsgHdr<ControlMessage = libc::cmsghdr>>(
         dst_ip: None,
         interface_index: None,
         stride: len,
+        timestamp: None,
     };
 
     let cmsg_iter = unsafe { cmsg::Iter::new(hdr) };
@@ -739,6 +746,7 @@ pub(crate) fn decode_recv<M: cmsg::MsgHdr<ControlMessage = libc::cmsghdr>>(
         ecn: EcnCodepoint::from_bits(ctrl.ecn_bits),
         dst_ip: ctrl.dst_ip,
         interface_index: ctrl.interface_index,
+        timestamp: ctrl.timestamp,
     })
 }
 
@@ -748,6 +756,7 @@ struct ControlMetadata {
     dst_ip: Option<IpAddr>,
     interface_index: Option<u32>,
     stride: usize,
+    timestamp: Option<Duration>,
 }
 
 impl ControlMetadata {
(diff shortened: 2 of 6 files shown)


quinn-udp/src/lib.rs
```diff
@@ -118,6 +118,10 @@ pub struct RecvMeta {
     pub dst_ip: Option<IpAddr>,
     /// The interface index of the interface on which the datagram was received
     pub interface_index: Option<u32>,
+    /// Kernel receive timestamp in nanoseconds since the Unix epoch
+    ///
+    /// Populated on platforms: Linux, Android.
+    pub timestamp: Option<u64>,
```
thomaseizinger: Should we represent it as `Duration`?

quinn-udp/src/unix.rs
```diff
@@ -696,6 +702,12 @@ fn recv_single(
     Ok(1)
 }
 
+// Space for packet info, ECN, GRO, and SCM_TIMESTAMPNS ancillary data.
+// 96 bytes covers the worst-case currently supported Unix control messages.
+#[cfg(any(target_os = "linux", target_os = "android"))]
+const CMSG_LEN: usize = 96;
+
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
 const CMSG_LEN: usize = 88;
```
thomaseizinger: Is this worth splitting? We already have a lot of conditional code in here. All this constant does is define the max length of a few arrays. I'd lean towards not making this conditional and just accept that on a few platforms, we'll use more stack space than strictly necessary.

thomaseizinger (commented):
Thanks! Looks clean.

Two comments.

quinn-udp/src/lib.rs
syszery: I initially considered `Duration`, too, but noticed that `std::time::Duration` is currently opted out for `wasm_browser` (see lines 35-39 in `lib.rs`). I decided using `u64` instead to keep the `RecvMeta` struct uniform across all platforms.

quinn-udp/src/unix.rs
syszery: I don't have a strong preference here and am happy to unify it if it’s more in line with the project’s style.

quinn-udp/src/unix.rs
```diff
@@ -149,6 +149,12 @@ impl UdpSocketState {
                 // https://github.com/quinn-rs/quinn/pull/1354.
                 gro_segments = 64
             }
+
+            if let Err(_err) =
+                set_socket_option(&*io, libc::SOL_SOCKET, libc::SO_TIMESTAMPNS, OPTION_ON)
+            {
+                crate::log::debug!("Ignoring error setting SO_TIMESTAMPNS on socket: {_err:?}");
```
djc, line 143: Unless this is the prevailing style, prefer to import `debug` directly.

quinn-udp/src/unix.rs
syszery: I've just removed the conditional `CMSG_LEN`.

quinn-udp/src/lib.rs
thomaseizinger: `Duration` lives in `core::time::Duration` so we can just change that import and it should work fine?

thomaseizinger: Just checked and doing that compiles fine with `wasm32-unknown-unknown`.

thomaseizinger (approved):
LGTM

@djc let me know if you are happy with the change to `core::time::Duration`.

Commits should be squashed into one before merging. Can you do that please @syszery ? Don't worry about attributing to me for the last one.

syszery:
Hi @thomaseizinger, thank you for the feedback and the improvements! I reviewed the changes, and I noticed that we now convert with `Duration::from_secs`. But I think we would need `from_nanos` here because the current implementation in the decoder uses nanoseconds.

However, to make the intent clearer and the code safer, I suggest to make the following change to decode directly into a  `Duration`. This way, we avoid this entirely.

```Rust
#[cfg(any(target_os = "linux", target_os = "android"))]
(libc::SOL_SOCKET, libc::SCM_TIMESTAMPNS) => {
    let ts = unsafe { cmsg::decode::<libc::timespec, libc::cmsghdr>(cmsg) };
    
    // Handle potential negative values on 32-bit/64-bit systems
    let secs = u64::try_from(ts.tv_sec).unwrap_or(0);
    let nsecs = u32::try_from(ts.tv_nsec).unwrap_or(0);

    self.timestamp = Some(core::time::Duration::new(secs, nsecs));
}

```

quinn-udp/src/unix.rs
```diff
@@ -908,6 +915,7 @@ fn decode_recv<M: cmsg::MsgHdr<ControlMessage = libc::cmsghdr>>(
         ecn: EcnCodepoint::from_bits(ctrl.ecn_bits),
         dst_ip: ctrl.dst_ip,
         interface_index: ctrl.interface_index,
+        timestamp: ctrl.timestamp.map(core::time::Duration::from_secs),
```
djc: Nit: prefer to import `Duration`.

quinn-udp/tests/tests.rs
```diff
@@ -359,6 +359,28 @@ fn test_send_recv(send: &Socket, recv: &Socket, transmit: Transmit<'_>) {
         } else {
             assert_eq!(meta.ecn, transmit.ecn);
         }
+
+        // On Linux and Android, we expect the kernel to provide a receive timestamp
+        // since we explicitly enabled `SO_TIMESTAMPNS`.
+        #[cfg(any(target_os = "linux", target_os = "android"))]
+        {
+            use std::time::Duration;
```
djc: Nit: prefer to import this at the module level.

djc (approved):
Looks okay to me modulo a few nits. Thanks!

quinn-udp/src/unix.rs
syszery: Moved import to the top of the module.

quinn-udp/tests/tests.rs
syszery: Moved import to the top of the module.

quinn-udp/src/unix.rs
syszery: I checked the rest of `unix.rs` and this seems to be the prevailing style within this module.

syszery:
I've addressed all the feedback and squashed the history.

- Moved the `Duration` imports to the top level of the corresponding modules (`unix.rs` and `tests.rs`).
- Converted `libc::timespec` directly into a `Duration` in `ControlMetadata::decode`.
- Kept the `crate::log::` prefix in `unix.rs` as it is the prevailing style across the rest of this file.

quinn-udp/src/lib.rs
```diff
@@ -118,6 +116,10 @@ pub struct RecvMeta {
     pub dst_ip: Option<IpAddr>,
     /// The interface index of the interface on which the datagram was received
     pub interface_index: Option<u32>,
+    /// Kernel receive timestamp in nanoseconds since the Unix epoch
```
thomaseizinger:
```suggestion
    /// Kernel receive timestamp as Unix epoch
```

thomaseizinger (approved):
One more nit, otherwise LGTM

quinn-udp/src/lib.rs
syszery: Done, amended the doc comment.

Ralith:
It looks like this just needs a rebase, and then we can merge it.

syszery:
Rebase is done. I also increased the buffer size for socket error queue from 128 to 256 because enabling `SO_TIMESTAMPNS` increases ancillary message size and as a result ICMP error tests failed due to truncation.

PR #2481, opened 2025-12-21
feat(bbr3) Implemented bbr3's latest version based on the RFC

Tipuch:
solves #1254
Based on https://www.ietf.org/archive/id/draft-ietf-ccwg-bbr-04.txt

tested with sudo tc qdisc add dev lo root netem delay 50ms 10ms distribution normal with a big file, this is a single connection 
and it's BBR3 vs Cubic (the default) for quinn

1206KiB/s for BB3 vs 1049 KiB/s for Cubic so approximately 13% faster transfer speed (this is with a big file so over a long period of time).

Screenshot_20251221_093603
Screenshot_20251221_093639

Let me know if you need any additional information or if you want me to change the implementation in some ways.

PR #2640, opened 2026-05-13
congestion: avoid double-reducing CUBIC fast convergence

0xdeafbeef:
Quinn mistakenly applied the beta multiplier to the newly adjusted `W_max` instead of the current window. This double-reduced the window to 59.5% instead of the intended 70%.  This fixes the math by snapshotting the original window for the multiplicative decrease.

### DIFF
--- a/quinn-proto/src/congestion/cubic.rs
+++ b/quinn-proto/src/congestion/cubic.rs
@@ -197,18 +197,19 @@ impl Controller for Cubic {
         }
 
         self.state.recovery_start_time = Some(now);
+        let window = self.state.window as f64;
 
-        // Fast convergence
-        if (self.state.window as f64) < self.state.w_max {
-            self.state.w_max = self.state.window as f64 * (1.0 + BETA_CUBIC) / 2.0;
+        // Fast convergence lowers W_max first; the 0.7 loss reduction still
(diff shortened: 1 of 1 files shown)


djc (approved):
Great find, thanks!

Do you want to backport this to the `0.11.x` branch for a release?

You might also be interested in

- https://github.com/quinn-rs/quinn/pull/2481

0xdeafbeef:
> Great find, thanks!
> 
> Do you want to backport this to the `0.11.x` branch for a release?
> 
> You might also be interested in
> 
>     * #2481

I've tried to backport it

Ralith (approved):
Great catch, thank you! Our congestion control performance has always been a little worse than it should; I wonder if this is why... any chance you have before/after data for some interesting case?

0xdeafbeef:
> Great catch, thank you! Our congestion control performance has always been a little worse than it should; I wonder if this is why... any chance you have before/after data for some interesting case?

Unfortunately no :( I had issues with cubic previously with LFN network(500mbit, 200 ms latency) slowing connection down to several kib/s. 

These findings are just comparisons of sources with rfcs and quiche by llm.

I can check other congestion controllers later

PR #2642, opened 2026-05-13
congestion: preserve excess CUBIC cwnd increment

0xdeafbeef:
CUBIC congestion avoidance can accumulate more than one MTU of cwnd_inc credit.  After growing the window by one MTU, Quinn reset cwnd_inc to zero, discarding the excess credit.  RFC 9002 section 7.3.3 limits a single congestion-avoidance increase to one max_datagram_size per congestion window acknowledged; it does not require throwing away the remainder.

Subtract one MTU from cwnd_inc instead, and saturate the accumulator so retained credit cannot wrap on large CUBIC targets.

### DIFF
--- a/quinn-proto/src/congestion/cubic.rs
+++ b/quinn-proto/src/congestion/cubic.rs
@@ -162,14 +162,20 @@ impl Controller for Cubic {
             }
 
             // Update the increment and increase cwnd by MSS.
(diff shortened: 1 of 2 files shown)


djc (approved):

Ralith (approved):
Great work, thanks!

djc:
Do you want to backport this one as well?

PR #2643, opened 2026-05-13
Upgrade to rand 0.10

### LINKED ISSUE #2636: Bump the `rand` crate to 0.10?
Hi there! 

I'm working on dedupling some of uv's resolutions, and one of our last resolutions of `rand@0.9` is via `quinn-proto`. Would you all be open to bumping your bound to `rand@0.10`?

If so, I'm happy to send a PR with that bump (including any internal API changes needed) 🙂

ww-oai:
This is atop #2531. The main things I've done are switch to PCG as the RNG impl, plus make a previous implicit (rand-side) panic explicit (on the quinn side).

Closes #2531. Closes #2636.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -515,6 +515,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
+name = "chacha20"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6f8d983286843e49675a4b7a2d174efe136dc93a18d69130dd18198a6c167601"
+dependencies = [
+ "cfg-if",
+ "cpufeatures",
+ "rand_core",
+]
+
+[[package]]
 name = "ciborium"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -682,6 +693,15 @@ name = "core-foundation-sys"
 version = "0.8.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
+
+[[package]]
+name = "cpufeatures"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b2a41393f66f16b0823bb79094d54ac5fbd34ab292ddafb9a0456ac9f87d201"
+dependencies = [
+ "libc",
+]
 
 [[package]]
 name = "crc"
@@ -924,10 +944,9 @@ version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ef975e30683b2d965054bb0a836f8973857c4ebf6acf274fe46617cd285060d8"
 dependencies = [
- "foldhash",
+ "foldhash 0.2.0",
  "libm",
  "portable-atomic",
- "rand",
  "siphasher",
 ]
 
@@ -958,6 +977,12 @@ name = "fnv"
 version = "1.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
+
+[[package]]
+name = "foldhash"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
 
 [[package]]
 name = "foldhash"
@@ -1067,10 +1092,24 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "899def5c37c4fd7b2664648c28120ecec138e4d395b459e5ca34f9cce2dd77fd"
 dependencies = [
  "cfg-if",
- "js-sys",
  "libc",
  "r-efi",
  "wasip2",
+]
+
+[[package]]
+name = "getrandom"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "139ef39800118c7683f2fd3c98c1b23c09ae076556b435f8e9064ae108aaeeec"
+dependencies = [
+ "cfg-if",
+ "js-sys",
+ "libc",
+ "r-efi",
+ "rand_core",
+ "wasip2",
+ "wasip3",
  "wasm-bindgen",
 ]
 
@@ -1108,6 +1147,15 @@ dependencies = [
  "cfg-if",
  "crunchy",
  "zerocopy",
+]
+
+[[package]]
+name = "hashbrown"
+version = "0.15.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9229cfe53dfd69f0609a49f65461bd93001ea1ef889cd5529dd176593f5338a1"
+dependencies = [
+ "foldhash 0.1.5",
 ]
 
 [[package]]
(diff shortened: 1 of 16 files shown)


Cargo.lock
```diff
@@ -924,10 +944,10 @@ version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ef975e30683b2d965054bb0a836f8973857c4ebf6acf274fe46617cd285060d8"
 dependencies = [
- "foldhash",
+ "foldhash 0.2.0",
  "libm",
  "portable-atomic",
- "rand",
+ "rand 0.9.4",
```
ww-oai: 🤔

djc: They've generally been responsive to my PRs...

woodruffw: Looks like they have this open, but it's held back by MSRV concerns: https://github.com/tomtomwombat/fastbloom/pull/27

I think in the meantime we _could_ disable the `rand` feature; I confirmed that the build/tests work fine without it. But doing so might have performance implications.

djc:
> This is atop #2531. The main things I've done are switch to PCG as the RNG impl, plus make a previous implicit (rand-side) panic explicit (on the quinn side).

I think we'd prefer to squash all these changes into a single commit (which we'll end up rebasing onto `main`). But, do you think it could make sense to isolate the switch to PCG in a separate commit?

Cargo.lock
woodruffw: (I'll push up a revertible commit demonstrating that.)

woodruffw:
> I think we'd prefer to squash all these changes into a single commit (which we'll end up rebasing onto `main`). But, do you think it could make sense to isolate the switch to PCG in a separate commit?

Yeah, can do!

Cargo.lock
ww-oai: [94cd243](https://github.com/quinn-rs/quinn/pull/2643/commits/94cd243e96be8fe6553a3a083364a1be550f396a) demonstrates that.

djc: That's because of me:

- https://github.com/tomtomwombat/fastbloom/pull/16

but we've since bumped our MSRV:

- https://github.com/quinn-rs/quinn/pull/2602

ww-oai:
Okay, rewrote the branch's history; the base commit is now the rand bump, and the next commit adds PCG via `rand_pcg`.

(Also, so sorry for all the account switches. Still trying to develop muscle memory around that...)

djc:
> (Also, so sorry for all the account switches. Still trying to develop muscle memory around that...)

No worries, I caught on to ww-oai pretty quickly... was a little surprised that I have to approve workflows for you even though we only restrict workflows for people new to GitHub. 😄

quinn-proto/src/endpoint.rs
```diff
@@ -67,9 +70,11 @@ impl Endpoint {
         allow_mtud: bool,
     ) -> Self {
         Self {
-            rng: config
-                .rng_seed
-                .map_or_else(StdRng::from_os_rng, StdRng::from_seed),
+            rng: match config.rng_seed {
+                Some(seed) => StdRng::from_seed(seed),
+                None => StdRng::try_from_rng(&mut SysRng)
+                    .expect("failed to seed random number generator from system"),
```
djc, line 78: So this is externalizing previous behavior? I'd still feel more comfortable yielding this out as an error.

Cargo.toml
```diff
@@ -22,16 +22,16 @@ bytes = "1"
 clap = { version = "4", features = ["derive"] }
 crc = "3"
 directories-next = "2"
-fastbloom = "0.17"
+fastbloom = { version = "0.17", default-features = false }
```
djc, line 25: This is probably good anyway, if we don't need it!

djc (approved):

Cargo.toml
```diff
@@ -22,16 +22,16 @@ bytes = "1"
...
-getrandom = { version = "0.3", default-features = false }
+getrandom = { version = "0.4", default-features = false }
 hdrhistogram = { version = "7.2", default-features = false }
 hex-literal = "1"
 lru-slab = "0.1.2"
 log = "0.4"
 pin-project-lite = "0.2"
 qlog = "0.17"
-rand = "0.9"
+rand = "0.10"
```
djc: Make this 0.10.1 to avoid the audit issue?

djc:
Might be nice to retain me as co-author on the first commit?

quinn-proto/src/endpoint.rs
ww-oai: Yep, the previous behavior was an internal panic. To my understanding this can only fail under abnormal/early initialization system states which is why it was deemed acceptable-for-panic in earlier versions of rand.

I can make it an error though, if you prefer! The other option would be to switch to an infallible thread-local RNG here, but I don't enough about your policy around forking/user responsibility for reseeding across threads to make that change confidently 😅

ww-oai:
> Might be nice to retain me as co-author on the first commit?

Oops, yes. Fixing that...

quinn-proto/src/endpoint.rs
djc: Seems like yielding `Result<Self, NoRandomBytes>` from here wouldn't be too painful, and leaves it up to the caller to deal with this.

ww-oai: Sounds good, added a `NoRandomBytes` unit type for the error 🙂. With that there are some new `unwrap`s in the unit tests, but I _think_ that's okay.

djc: Sorry, I think there was an existing `NoRandomBytes` type somewhere?

ww-oai: > Sorry, I think there was an existing `NoRandomBytes` type somewhere?

There was in #2531 but it got squashed out with the original panic change, so I re-added it!

quinn-proto/src/congestion/bbr/mod.rs
```diff
@@ -99,8 +98,8 @@ impl Bbr {
             bw_at_last_round: 0,
             round_wo_bw_gain: 0,
             ack_aggregation: AckAggregationState::default(),
-            random_number_generator: StdRng::try_from_rng(&mut SysRng)?,
-        })
+            random_number_generator: Pcg32::from_rng(&mut rand::rng()),
```
Ralith, line 101: While I think `Pcg32` makes more sense here than `StdRng`, and this module is likely not long for the world anyway (see #2481), I think as far as on-topic effects go, this change is just smuggling in an unwrap: `rand::rng()` [is documented as panicking](https://docs.rs/rand/latest/rand/fn.rng.html#panics) if `SysRng` can't be used for initialization.

It's also not obvious to me that propagating the `Result` as in the prior state is actually worth going out of our way to avoid, unless we're making a blanket judgement that we don't care about the case where `StdRng` is unavailable, which is at odds with the choice to propagate the error in `Endpoint::new`. I don't feel strongly either way, but we should probably be consistent, and I'd be happy to merge the PR with the error propagation here intact.

Conversely, if you intend to backport this to stable, we'd need a version of the PR with unwraps everywhere to prevent breakage regardless of what happens on main.

ww-oai: > It's also not obvious to me that propagating the `Result` as in the prior state is actually worth going out of our way to avoid, unless we're making a blanket judgement that we don't care about the case where `StdRng` is unavailable, which is at odds with the choice to propagate the error in `Endpoint::new`. I don't feel strongly either way, but we should probably be consistent, and I'd be happy to merge the PR with the error propagation here intact.

My 0.02c here is consistency in favor of panics, but I'm by far less familiar with this codebase than either of you  -- my experience with OS CSPRNGs is that they're only fallible under abnormal system conditions _or_ early initialization, and in both of those cases panicking rather than propagating is _probably_ the right thing to do (since it's generally not a recoverable state).

So, my proposal is to keep this change as it is (smuggling a panic, as you note) and revert the `Endpoint::new` change so that it's an explicit panic again (which would be consistent with the implicit behavior on `rand@0.9`. If @djc is okay with that as well then I'll make that change 🙂

djc: Sure, I can live with that.

ww-oai: Thanks both, I reverted back to the panicking behavior.

Ralith (approved):
Thanks!

Issue #2076, opened 2024-12-02
How I can use difference cert/pkey for different TLS sni?


taikulawo:
On rustls, We have [LazyConfigAcceptor](https://docs.rs/tokio-rustls/latest/tokio_rustls/struct.LazyConfigAcceptor.html.
Does quinn also has Accetpor let user decide which cert/pkey to use?

djc:
See #2024.

Issue #2024, opened 2024-10-29
Access Client Hello before accepting connection


FlorianUekermann:
I need to access the `rustls::server::ClientHello` or at least the content of `server_name()` before accepting a connection with a rustls config. `alpn()` would also be very useful. Could this be exposed via `Incoming` or is that too early?

Ralith:
`Incoming` is too early: it is yielded before the endpoint even decrypts first packet, and a `ClientHello` might span multiple packets. I think your use case is worth supporting, but it will require an additional intermediate stage between `Incoming` and `Connection`, and changes in the proto layer to suspend work once the `ClientHello` is received.

FlorianUekermann:
I see. Sounds like a non-trivial amount if work. That's unfortunate. Maybe it would be easier to add an accept method, which takes a closure mapping Client Hello to a rustls config as a first step.

djc:
We already have a `Connecting` stage, so this might fit in there? I wonder if we could do this by adding a (provided) method to quinn-proto's `ServerConfig` trait.

FlorianUekermann:
> We already have a `Connecting` stage, so this might fit in there?

`Connecting` exposes the rustls handshake data (including the negotiated protocol), so I don't think we can get to hat stage without a rustls config. So even changing the semantics of the `Connecting` stage slightly would be tricky for compatibility reasons.

But the fact that the handshake data is available in `Connecting`, which you get by calling `Incoming::accept()` (which is not `async`, so presumably not waiting for network traffic) is a little hard to reconcile with @Ralith statement. Doesn't that mean the `Client Hello` would need to be received before we reach the `Incoming` stage? I'll try to take a look.

Ralith:
`Connection::handshake_data` is async. If your Client Hello has not been fully received, it won't be ready yet.

Ralith:
> We already have a Connecting stage, so this might fit in there?

I'm not sure if this would be feasible. The motivation here is to make the connection's configuration depending on the Client Hello, so we have to defer the connection logic until after the application has had its opportunity to weigh in, similar to how `Incoming` defers all frame processing.

FlorianUekermann:
Adding an async `Incoming::client_hello_data`, which waits until the data is available similar to `Connecting::handshake_data` would be best from a user perspective imo. I don't think the public API would benefit from an intermediate stage, because I think it would just duplicate the methods `Incoming` already has.

Ralith:
[...]

PR #2644, opened 2026-05-13
Expose TLS handshake bytes from `Incoming`

aiguy110:
## Summary

Adds `Incoming::handshake_bytes()` (and a forwarding method on the high-level `quinn::Incoming` wrapper), which decrypts the Initial packet using the publicly-derivable Initial keys, walks the QUIC frames, and returns the concatenated `CRYPTO` frame contents in offset order.

Per RFC 9001 §4.1.3, TLS handshake messages travel inside `CRYPTO` frames with no intervening TLS record layer, so the returned bytes start directly with the TLS handshake message header (`0x01` for `ClientHello`, then a 3-byte length, then the body). Callers can feed this to a TLS parser to inspect extensions such as SNI or ALPN before deciding whether to `accept()`, `retry()`, `refuse()`, or `ignore()` the connection.

## Motivation

`Incoming` (from #2076) gave users the ability to make accept/refuse/retry/**ignore** decisions before any QUIC response goes on the wire — a very nice primitive for hardening. The decision can today be informed by remote address, address-validation status, original DCID, and local IP. What's missing is any ability to look at the ClientHello itself.

Concrete use case I bumped into: port-knocking style authentication on a UDP service. The goal is for the port to appear closed to scanners that don't carry a pre-shared token (an HMAC, say, embedded in a custom ALPN entry). Without ClientHello inspection, the closest one can do today is gate at the rustls layer via a custom `ResolvesServerCert` — but by that point a TLS alert has already left the box, so the service is observable. With `handshake_bytes()` + `Incoming::ignore()`, the port becomes truly silent for invalid clients (WireGuard-grade posture), while still being a normal QUIC service for valid ones. I've verified this end-to-end with packet capture: zero response packets to clients failing the check.

Other plausible uses: SNI-keyed rate limiting before TLS handshake, ClientHello-fingerprint denylists, GREASE/ECH-aware admission, eBPF-style port-knocking with cleaner state handling than out-of-band knock daemons.

## API shape

```rust
impl Incoming {
    pub fn handshake_bytes(&self) -> Result<Vec<u8>, HandshakeBytesError>;
}

pub enum HandshakeBytesError {
    Decrypt,            // AEAD authentication failed
    MalformedFrames,    // frame parsing failed
    NonContiguous,      // CRYPTO frames don't run from offset 0
}
```

Re-exported through `quinn-proto::Incoming` and `quinn::Incoming`.

## Design choices / alternatives considered

- **Why return bytes instead of a parsed `ClientHello`?** Keeps `quinn-proto` firmly in QUIC framing and out of the TLS layer. Callers already have rustls (or webpki-types, or whatever) available for the TLS-side parsing. Avoids growing `quinn-proto`'s API surface with a TLS data model.
- **Why a method on `Incoming` rather than exposing internal frame parsing?** `Iter`/`Frame`/`Crypto` are currently `pub(crate)`; making them public is a much larger commitment. A single inspection method is the smallest user-facing surface for the same capability.
- **Why `Vec<u8>` and not `Bytes` / `&[u8]`?** The CRYPTO frames need to be concatenated in offset order, so the result is inherently owned. Happy to switch to `Bytes` if that's preferred.
- **Non-`CRYPTO` frames** (`PADDING`, `PING`, `ACK`, `CONNECTION_CLOSE`) are silently skipped — they're legal in Initials per RFC 9000 §17.2.2.
- **Non-contiguous CRYPTO frames** return an explicit error rather than being silently reassembled with gaps. Real ClientHellos in Initials are typically a single frame at offset 0; a gap suggests either a fragmented handshake (rare for ClientHello) or something malformed.
- **Cost.** The method clones the (encrypted) payload buffer and does one AEAD decrypt + a frame walk. Cheap enough that callers can invoke it unconditionally; not so cheap that it should be called many times — the doc could note "cache the result if you need it more than once," happy to add that.

## What's not in this PR

- **No tests.** Most happy to add one — an integration test that captures an Initial via the existing simulated-IO test harness and asserts the returned bytes start with `0x01` and parse as a ClientHello would be straightforward. Held off pending design feedback on the API shape.
- **No changes to `accept()`/`ignore()`/etc.** The method is purely an additional read-only inspection point.

## Compatibility

Additive only. No existing methods change behaviour or signatures.

---

Happy to convert this to an issue / discussion if you'd prefer to align on the API shape before reviewing code. Also happy to iterate on naming (`handshake_bytes` vs `client_hello_bytes` vs `initial_crypto` etc.), error type granularity, or return type.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -1198,6 +1198,75 @@ impl Incoming {
     pub fn orig_dst_cid(&self) -> ConnectionId {
         self.token.orig_dst_cid
     }
+
+    /// The TLS handshake bytes carried by the Initial packet, suitable for
+    /// inspecting the ClientHello before deciding whether to
+    /// [`accept`](Endpoint::accept) or [`ignore`](Endpoint::ignore) this
+    /// connection attempt.
+    ///
+    /// This decrypts the Initial packet's payload using the (publicly
+    /// derivable) Initial keys, walks the QUIC frames, and concatenates the
+    /// data carried by `CRYPTO` frames in offset order. Per RFC 9001 §4.1.3,
+    /// QUIC carries TLS handshake messages directly inside `CRYPTO` frames
+    /// with no intervening TLS record layer, so the returned bytes begin with
+    /// the TLS handshake message header (`0x01` = ClientHello, then a 3-byte
+    /// length, then the body).
+    ///
+    /// Returns an error if the packet fails AEAD authentication, if its
+    /// frames cannot be parsed, or if the `CRYPTO` frames are non-contiguous
+    /// from offset 0. Non-`CRYPTO` frames (e.g. `PADDING`, `PING`) are
+    /// ignored.
+    pub fn handshake_bytes(&self) -> Result<Vec<u8>, HandshakeBytesError> {
+        let packet_number = self.packet.header.number.expand(0);
+        let mut payload = self.packet.payload.clone();
+        self.crypto
+            .packet
+            .remote
+            .decrypt(packet_number, &self.packet.header_data, &mut payload)
+            .map_err(|_| HandshakeBytesError::Decrypt)?;
+
+        let mut chunks: Vec<(u64, Bytes)> = Vec::new();
+        let iter =
+            frame::Iter::new(payload.freeze()).map_err(|_| HandshakeBytesError::MalformedFrames)?;
+        for frame in iter {
+            // PADDING, PING, ACK, CONNECTION_CLOSE are legal in Initials —
+            // skip them. Anything else would be a protocol violation, but we
+            // leave that judgement to the post-accept handshake path.
+            if let frame::Frame::Crypto(c) =
+                frame.map_err(|_| HandshakeBytesError::MalformedFrames)?
+            {
+                chunks.push((c.offset, c.data));
+            }
+        }
+
+        chunks.sort_by_key(|(off, _)| *off);
+        let mut out = Vec::new();
+        let mut expected = 0u64;
+        for (off, data) in chunks {
+            if off != expected {
+                return Err(HandshakeBytesError::NonContiguous);
+            }
+            expected += data.len() as u64;
+            out.extend_from_slice(&data);
+        }
+
+        Ok(out)
+    }
+}
+
+/// Error returned by [`Incoming::handshake_bytes`].
+#[derive(Debug, Error)]
+pub enum HandshakeBytesError {
+    /// The Initial packet failed AEAD authentication.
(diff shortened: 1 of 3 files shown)


Ralith:
This is a really cool idea! The use case makes perfect sense to me, and is something I'm happy to support. I particularly appreciate the design to yield unprocessed plaintext bytes; that's good layering.

One alternative strategy might be to have the crypto layer parse them into something that's then passed back with type erasure, like we've done for `Connection::peer_identity`. This could be an improvement over requiring callers to all reimplement TLS decoding, *if* there's a stronger type still broad enough not to exclude reasonable use cases, which isn't immediately obvious to me. I'm inclined to accept the raw bytes as a starting point, and we can always iterate later.

I do have some reservations about the frame processing strategy:
- We should handle out-of-order frames. We already have the tools in `connection::Assembler`, so it's strange to reject easily intelligible packets.
- I'm concerned that, for example, post-quantum handshakes might make multi-packet `ClientHello`s routine. I'd prefer to expose a `RecvStream` or something very similar here, so callers can decide for themselves how much it's useful to read. Later packets are currently queued per connection in `quinn_proto::Endpoint::incoming_buffers`.

djc:
In my mind this is (mostly) a poor man's version of

- https://github.com/quinn-rs/quinn/issues/2024

I'd prefer to get that done holistically, including making sure that callers don't have to commit to a `ServerConfig` up front but can only select one on `accept()`.

iadev09:
One scope note: #2671 is aimed at #2024: selecting a Quinn `ServerConfig` after rustls has produced a typed `ClientHello`.

It does not try to solve the broader raw early-policy/fingerprinting shape discussed here. Separately, rustls/rustls#3070 moves the relevant ClientHello acceptor primitive to the rustls side, which may be the better boundary for that.

If this is still valuable, I think the next experiment should probably be on the rustls side: a separate borrowed view with a lifetime tied to the accepted ClientHello state, similar to the existing typed `ClientHello` shape.

For example:

```rust
let hello = accepted.client_hello();      // semantic typed view
let raw = accepted.raw_client_hello();    // wire-exact/raw view
```

Both views could stay borrowed from the same accepted ClientHello state, so this does not need a second parse or an owned copy on the normal path. It also keeps the boundary narrow for rustls while giving downstream users enough information to build their own policy. 

Having this on the rustls side could also let downstreams share the same policy logic across TCP/rustls and QUIC/rustls paths.

djc:
The use cases listed in the PR description also seem to be serviceable from a rustls `ClientHello`, such as the one exposed via the `Acceptor`. I think exposing a `rustls::quic::Acceptor` via `Incoming` makes sense, and avoids having multiple avenues to do roughly the same thing.

PR #2646, opened 2026-05-17
feat(udp): add support for `IP_RECVERR` and `IPV6_RECVERR` (Linux/Android)

### LINKED ISSUE #2052: Add support for `IP_RECVERR` and `IPV6_RECVERR`

syszery:
### Description

Closes #2052

This PR adds support for `IP_RECVERR` and `IPV6_RECVERR` on Linux and Android, enabling reception of asynchronous transport-layer errors via the socket error queue (`MSG_ERRQUEUE`).

It follows the “opt-in” approach (`recv_transport_error`) suggested in the original review of #2421. This keeps error handling out of the normal receive path and avoids any overhead for users who don’t need it.

### Changes

- Enable `IP_RECVERR` / `IPV6_RECVERR` on supported sockets
- Add `recv_transport_error` API for polling kernel error queue entries
- Decode Linux `sock_extended_err` messages into a `TransportError`
- On unsupported platforms, the API returns `Ok(None)`

### Notes

This builds on the work from #2421. Co-authorship has been preserved with the original contributor.

Co-authored-by: @RaghavendraRQ

### DIFF
--- a/quinn-udp/src/lib.rs
+++ b/quinn-udp/src/lib.rs
@@ -170,6 +170,51 @@ impl Transmit<'_> {
     }
 }
 
+/// Asynchronous transport-layer errors reported by the operating system
+///
+/// On Linux and Android these are delivered via the socket error queue
+/// (`MSG_ERRQUEUE`) and originate from ICMP messages.
+///
+/// These errors are out-of-band and do not correspond to a received packet.
+#[derive(Debug, Copy, Clone)]
+#[non_exhaustive]
+pub struct TransportError {
+    /// Address associated with the error
+    ///
+    /// This is the remote peer or an intermediate network device that triggered the error.
+    /// Returns `None` if the kernel cannot determine the source (e.g. `AF_UNSPEC`).
+    pub addr: Option<SocketAddr>,
+    /// Transport-layer error details
+    pub payload: TransportErrorPayload,
+    /// The raw error code from the underlying operating system
+    pub raw_errno: i32,
+}
+
+impl TransportError {
+    /// Returns the recommended MTU for packet-too-big errors
+    pub fn mtu(&self) -> Option<u32> {
+        match self.payload {
+            TransportErrorPayload::TooBig { mtu } => Some(mtu),
+            _ => None,
+        }
+    }
+}
+
+/// Transport-layer error details reported by the kernel
+#[derive(Debug, Copy, Clone)]
+#[non_exhaustive]
+pub enum TransportErrorPayload {
+    /// Destination host or port is unreachable
+    Unreachable,
+    /// Packet exceeds path MTU
+    TooBig {
+        /// Recommended Maximum Transmission Unit
+        mtu: u32,
+    },
+    /// Other transport-layer or kernel-reported error
+    Other,
+}
+
 /// Log at most 1 IO error per minute
 #[cfg(not(wasm_browser))]
 const IO_ERROR_LOG_INTERVAL: Duration = std::time::Duration::from_secs(60);
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -15,8 +15,12 @@ use std::{
 use socket2::SockRef;
 
 use super::{
-    EcnCodepoint, IO_ERROR_LOG_INTERVAL, RecvMeta, Transmit, UdpSockRef, cmsg, log_sendmsg_error,
+    EcnCodepoint, IO_ERROR_LOG_INTERVAL, RecvMeta, Transmit, TransportError, UdpSockRef, cmsg,
+    log_sendmsg_error,
 };
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+use super::TransportErrorPayload;
 
 // Adapted from https://github.com/apple-oss-distributions/xnu/blob/8d741a5de7ff4191bf97d57b9f54c2f6d4a15585/bsd/sys/socket_private.h
 #[cfg(apple_fast)]
@@ -149,6 +153,21 @@ impl UdpSocketState {
                 // https://github.com/quinn-rs/quinn/pull/1354.
                 gro_segments = 64
             }
+
+            if is_ipv4 || !io.only_v6()? {
+                if let Err(_err) =
+                    set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_RECVERR, OPTION_ON)
+                {
+                    crate::log::debug!("ignoring error setting IP_RECVERR on socket: {_err:?}");
+                }
+            }
+            if !is_ipv4 {
+                if let Err(_err) =
+                    set_socket_option(&*io, libc::IPPROTO_IPV6, libc::IPV6_RECVERR, OPTION_ON)
+                {
+                    crate::log::debug!("ignoring error setting IPV6_RECVERR on socket: {_err:?}");
+                }
+            }
         }
         #[cfg(any(target_os = "freebsd", apple))]
         {
@@ -277,6 +296,29 @@ impl UdpSocketState {
         recv_single(socket.0, bufs, meta)
     }
 
+    /// Receives a pending, asynchronous transport-layer error from this socket.
+    ///
+    /// On Linux and Android this pops one entry from the socket error queue
+    /// (`MSG_ERRQUEUE`). Returns `None` if the queue is empty or if the
+    /// underlying platform is unsupported.
+    ///
+    /// Returns an error if the underlying system call fails unexpectedly.
+    pub fn recv_transport_error(
+        &self,
+        socket: UdpSockRef<'_>,
+    ) -> io::Result<Option<TransportError>> {
+        #[cfg(any(target_os = "linux", target_os = "android"))]
+        {
+            Ok(LinuxError::recv(socket.0)?.map(TransportError::from))
+        }
+
+        #[cfg(not(any(target_os = "linux", target_os = "android")))]
+        {
+            let _ = socket;
+            Ok(None)
+        }
+    }
+
     /// The maximum amount of segments which can be transmitted if a platform
     /// supports Generic Send Offload (GSO).
     ///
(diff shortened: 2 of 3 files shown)


quinn-udp/src/unix.rs
```diff
@@ -277,6 +296,20 @@ impl UdpSocketState {
         recv_single(socket.0, bufs, meta)
     }
 
+    /// Receives a pending asynchronous transport-layer error from this socket
```
thomaseizinger:
```suggestion
    /// Receives a pending, asynchronous transport-layer error from this socket.
```

quinn-udp/src/unix.rs
```diff
@@ -696,6 +729,162 @@ fn recv_single(
...
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
+fn recv_transport_error(io: SockRef<'_>) -> io::Result<Option<TransportError>> {
+    let _ = io;
+    Ok(None)
+}
+
+/// Converts a Linux socket error queue entry into a transport-level error
+#[cfg(any(target_os = "linux", target_os = "android"))]
+fn decode_transport_err(raw: LinuxError) -> TransportError {
+    crate::log::debug!(
```
thomaseizinger: I think debug might be a bit noisy for this? Anything unconditional should IMO be trace.

quinn-udp/src/unix.rs
```diff
@@ -696,6 +729,162 @@ fn recv_single(
...
+        return Err(io::Error::new(
+            io::ErrorKind::InvalidData,
+            "control message truncated",
+        ));
+    }
+
+    let cmsg_iter = unsafe { cmsg::Iter::new(&hdr) };
+
+    for cmsg in cmsg_iter {
+        if (cmsg.cmsg_level == libc::IPPROTO_IP && cmsg.cmsg_type == libc::IP_RECVERR)
```
thomaseizinger: Would it make sense to move these checks into `decode`? Then the decoder doesn't have to say "assumes layout of ...".

thomaseizinger (commented):
Great stuff, looks really good. Left a few comments.

quinn-udp/src/unix.rs
syszery: Done, changed this to `trace!`

quinn-udp/src/unix.rs
syszery: Good point, I've moved this into `LinuxError::decode`.

quinn-udp/src/unix.rs
```diff
@@ -696,6 +729,162 @@ fn recv_single(
...
+struct LinuxError {
+    ee: libc::sock_extended_err,
+    offender: Option<SocketAddr>,
+}
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+impl LinuxError {
+    /// Attempts to decode a Linux `sock_extended_err` from a MSG_ERRQUEUE control message.
+    fn decode(cmsg: &libc::cmsghdr) -> Option<Self> {
+        let is_recverr = (cmsg.cmsg_level == libc::IPPROTO_IP
```
thomaseizinger: Mostly personal preference but this could also be expressed as two `if`s that check for "is not ip | ipv6" and "is not recv error".

That might make the boolean chain a bit easier to understand at the expense of a few lines.

thomaseizinger: Not a strong opinion though.

thomaseizinger (approved):
LGTM

quinn-udp/src/unix.rs
syszery: Makes sense 👍

quinn-udp/src/lib.rs
```diff
@@ -170,6 +170,42 @@ impl Transmit<'_> {
...
+    ///
+    /// This is the remote peer or an intermediate network device that triggered the error.
+    /// Returns `None` if the kernel cannot determine the source (e.g. `AF_UNSPEC`).
+    pub addr: Option<SocketAddr>,
+    /// The general category of the error
+    pub category: TransportErrorCategory,
+    /// The recommended Maximum Transmission Unit (MTU)
+    ///
+    /// Present when `category` is [`TransportErrorCategory::TooBig`].
+    pub mtu: Option<u32>,
```
mxinden: If this is only present on `TooBig`, why not make the `TooBig` variant carry the mtu?

```
pub enum TransportErrorCategory {
    /// Destination host or port is unreachable
    Unreachable,
    /// Packet exceeds path MTU
    TooBig(u32),
    /// Other transport-layer or kernel-reported error
    Other,
}
```

mxinden (commented):
Good work. Thank you.

We might leverage this in Firefox. //CC @larseggert

quinn-udp/src/lib.rs
thomaseizinger: I had this thought as well. I am unsure as to whether this enum should be called `Category` then as that name to me implies that it doesn't hold extra data, similar to `io::ErrorKind`. Perhaps all of `TransportError` should be an enum with accessors for the shared fields?

syszery: Thanks for the feedback and the suggestions. I don‘t have a strong preference and would be happy to implement whichever approach is preferred.

If I understand @thomaseizinger‘s suggestion correctly, it would look something like this?

```rust
enum TransportError {
    Unreachable {
        from: SocketAddr,
    },
    TooBig {
        from: SocketAddr,
        mtu: u32,
    },
    Other {
        from: SocketAddr,
    },
}

impl TransportError {
    pub fn from(&self) -> SocketAddr {
        match self {
            TransportError::Unreachable { from } => *from,
            …
        }
    }
}

```

thomaseizinger: Yeah that is what I meant although the duplication of `from` is a bit meh. Or maybe we just rename the enum to `TransportErrorPayload`? That would be in line with the format of ICMP itself. There, data like the MTU is part of the ICMP payload.

syszery: I replaced `TransportErrorCategory` with `TransportErrorPayload`, moving MTU into the `TooBig` variant and added a mock test for `LinuxError::decode` to validate CMSG parsing.

thomaseizinger (approved):

quinn-udp/src/unix.rs
```diff
@@ -149,6 +153,21 @@ impl UdpSocketState {
                 // https://github.com/quinn-rs/quinn/pull/1354.
                 gro_segments = 64
             }
+
+            if is_ipv4 || !io.only_v6()? {
+                if let Err(_err) =
+                    set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_RECVERR, OPTION_ON)
+                {
+                    crate::log::debug!("Ignoring error setting IP_RECVERR on socket: {_err:?}");
```
djc: Nit: prevailing style is to have all-lowercase log messages.

quinn-udp/src/unix.rs
```diff
@@ -696,6 +729,166 @@ fn recv_single(
...
+#[cfg(any(target_os = "linux", target_os = "android"))]
+fn recv_transport_error(io: SockRef<'_>) -> io::Result<Option<TransportError>> {
+    Ok(recv_linux_error(io)?.map(decode_transport_err))
+}
+
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
+fn recv_transport_error(io: SockRef<'_>) -> io::Result<Option<TransportError>> {
+    let _ = io;
+    Ok(None)
+}
```
djc: Suggest inlining these into their single caller. (For the non-Linux path, consider that the argument can be named `_io` directly, saving syntactic overhead on renaming boilerplate.

quinn-udp/src/unix.rs
```diff
@@ -696,6 +729,166 @@ fn recv_single(
     Ok(1)
 }
 
+// Control message buffer size for socket error queue (MSG_ERRQUEUE)
+#[cfg(any(target_os = "linux", target_os = "android"))]
+const ERR_CMSG_LEN: usize = 128;
+
+// Reads one entry from the Linux socket error queue (MSG_ERRQUEUE)
+#[cfg(any(target_os = "linux", target_os = "android"))]
+fn recv_linux_error(io: SockRef<'_>) -> io::Result<Option<LinuxError>> {
```
djc: What's the value of the keeping `LinuxError` as a separate type from `TransportError`? IMO this would be clearer with a single method that merges `recv_linux_error()` and `decode_transport_error()`, potentially attached to `TransportError` as a constructor.

(Please make sure to order items in top down order, that is with callers before callees, including constants like `ERR_CMSG_LEN`.)

djc (commented):
Thanks, this is mostly looking good save for some organizational nits.

Please squash all commits.

quinn-udp/src/unix.rs
syszery: `TransportError` lives in `lib.rs` and is currently platform-agnostic. Moving the Linux decoding logic into a `TransportError` constructor would either require introducing `libc` and `sock_extended_err` into `lib.rs`, or attaching Linux-specific construction logic to a cross-platform type. `LinuxError` avoids that by acting as a private adapter that contains the Linux kernel ABI entirely within `unix.rs`.

The split also reflects two different responsibilities. `LinuxError::decode` owns the unsafe kernel ABI boundary, while the conversion to `TransportError` is purely safe interpretation. Keeping those stages separate makes the unsafe code easier to audit independently from the transport-error mapping.

That said, I agree the standalone `decode_transport_err` function doesn't add much value. I'm happy to fold that into `LinuxError::into_transport_error()`.

```rust
struct LinuxError {
    ee: libc::sock_extended_err,
    offender: Option<SocketAddr>,
}

impl LinuxError {
    const CMSG_BUF_LEN: usize = 128;

    fn recv(io: SockRef<'_>) -> io::Result<Option<Self>> { ... }
    fn decode(cmsg: &libc::cmsghdr) -> Option<Self> { ... }
    fn into_transport_error(self) -> TransportError { ... }
}
```

djc: > That said, I agree the standalone `decode_transport_err` function doesn't add much value. I'm happy to fold that into `LinuxError::into_transport_error()`.

Suggest making it an `impl From<LinuxError> for TransportError`.

syszery:
Thanks for the feedback. I've addressed all points:

* lower case log messages
* inlined `recv_transport_error`​
* consolidated all `LinuxError​` methods
* implemented `from`​ for `TransportError`​
* reordered items as suggested

djc:
Thanks!

PR #2651, opened 2026-05-20
proto: reject stream flow-control frames above limits

GHX5T-SOL:
Fixes #2650.

## Summary
- validate send-side stream IDs before handling `STOP_SENDING` and `MAX_STREAM_DATA`
- return `STREAM_LIMIT_ERROR` for remote-initiated bidirectional stream IDs above the advertised limit
- add regression coverage for both frame types

## Validation
- `cargo test --locked -p quinn-proto stream_limit` passed: 5 passed, 0 failed
- `cargo test --locked -p quinn-proto` passed: 273 unit tests and 3 doctests passed
- `cargo fmt --all -- --check` passed
- `cargo clippy --locked -p quinn-proto --all-targets -- -D warnings` passed
- `cargo test --locked --manifest-path fuzz/Cargo.toml` passed
- `RUSTFLAGS='--cfg fuzzing' cargo build --locked --manifest-path fuzz/Cargo.toml --bin streams` passed
- `git diff --check` passed
- `gitleaks protect --staged --redact --no-banner` passed: no leaks found

Note: #2601 also touches remote stream allocation state; this PR is against current `main` and may need a rebase if #2601 lands first.

PR #2652, opened 2026-05-20
streams: reject STOP_SENDING and MAX_STREAM_DATA beyond the stream limit

### LINKED ISSUE #2650: Raise a transport error on out-of-bounds STOP_SENDING and MAX_STREAM_DATA
@kixelated notes:
> Quinn currently ignores STOP_SENDING and MAX_STREAM_DATA frames referencing a stream ID above MAX_STREAMS, instead of returning a protocol violation

c-tonneslan:
Closes #2650.

`received_reset` runs `validate_receive_id`, so a RESET_STREAM that references a stream ID past the limit we advertised gets caught as a `STREAM_LIMIT_ERROR`. `received_stop_sending` and `received_max_stream_data` never had the equivalent check, so a peer could send STOP_SENDING or MAX_STREAM_DATA for a bidi stream it was never allowed to open and we'd silently ignore the frame instead of raising a protocol violation.

I pulled the check into `validate_remote_stream_limit` and call it from both handlers. Locally initiated streams stay bounded by the existing `is_local_unopened` checks, so this only tightens handling of peer-initiated stream IDs. Added unit tests for both frames covering the within-limit and over-limit cases.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -2913,7 +2913,7 @@ impl Connection {
                             "STOP_SENDING on unopened stream",
                         ));
                     }
-                    self.streams.received_stop_sending(id, error_code);
+                    self.streams.received_stop_sending(id, error_code)?;
                 }
                 Frame::RetireConnectionId { sequence } => {
                     let allow_more_cids = self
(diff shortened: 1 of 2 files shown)


Ralith (commented):
Thanks! Unfortunately it looks like you've raced with https://github.com/quinn-rs/quinn/pull/2651, which came in earlier. Code looks good, aside from the missed update to the fuzzer.

Ralith (approved):
The other submitter was unresponsive, so I'm inclined to move forwards with this. Can you fix up the fuzzing path?

Issue #2201, opened 2025-04-12
Should `quinn-udp` automatically chunk `Transmit::contents`?


thomaseizinger:
As part of implementing GSO in our application, we ran into a bug where we submitted `Transmit`s that had a way too large `contents` slice. I went about the implementation a bit too naive unfortunately: https://github.com/firezone/firezone/blob/649c03e2908c797664ded06ef93234312e694cb7/rust/socket-factory/src/lib.rs#L308-L326

Can you spot the bug? Depending on what `segment_size` is, we may end up with a slice that is much larger than the allowed 65535 bytes by the kernel. A typical MTU of 1200-1500 can easily hit this in a bandwidth heavy application. `max_gso_segments` defaults to 64 so any `segment_size` over 1023 will result in a `Transmit` that will actually be rejected by the kernel.

Here is the fix: https://github.com/firezone/firezone/pull/8754

In order to surface this correctly to the user, I opened https://github.com/quinn-rs/quinn/pull/2199 but I think we should go further. In fact, I think I shouldn't need to write the above linked code at all. `quinn-udp` already has the knowledge of how many gso segments are currently allowed. My suggestion would be that `quinn-udp` should implement the linked code fragment and slice the passed `Transmit::contents` into chunks that are acceptable for the kernel.

From an API design perspective, I think this is much better as it reduces the "chatter" we need to have with `UdpSocketState`: Instead of having to ask it for the `max_gso_segments`, we can just pass it a possibly massive slice and `quinn-udp` takes care of:

- Ensuring that we don't do GSO if we shouldn't, i.e. effectively making many `Transmit`s out of the one that was passed
- Ensuring that we only pass a slice of at most 65535 bytes if the kernel supports GSO.

Thoughts?

Ralith:
We also want to make it easy for downstream code to allocate resources efficiently. Typically batch size is somewhat arbitrary, and very large batches can waste memory, potentially to a harmful extent on resource-constrained devices. I don't immediately see a problem with exposing a more DWIM-oriented API, but doing so wouldn't remove the need to expose constants describing the platform's size constraints so that conscientious downstreams can ensure they don't perform more buffering than is useful.

mxinden:
> Ensuring that we only pass a slice of at most 65535 bytes if the kernel supports GSO.

If I recall correctly, the maximum UDP datagram size (here 65535 bytes) might differ across operating systems and might even be configurable on some. How does one know when to slice?

[...]

PR #2653, opened 2026-05-24
fix(unix): split oversized GSO sends into smaller chunks

### LINKED ISSUE #2399: udp: retry failed GSO sendmsg on EIO without GSO

Reports:

kixelated:
(summary is AI generated)

When a `Transmit` exceeds what one `sendmsg` syscall can accept — either because it has more segments than `max_gso_segments` allows, or because its total size exceeds the kernel's per-call limit (`u16::MAX` bytes) — split it upfront into smaller batches dispatched via separate `sendmsg` calls. This catches three cases that today silently drop datagrams or rely on the QUIC layer's loss timers to recover:

* Stale in-flight GSO transmits left over from a GSO halt (#2399). After `EIO` halts GSO via `max_gso_segments.store(1)`, any transmits queued while GSO was still enabled now violate the new limit and need to be re-batched as individual sends rather than dropped.

* The first GSO send that triggers the halt itself (#2399). The `EIO` arm now re-dispatches the failing transmit through `send` after the halt, where the upfront chunking check splits it into individual datagrams without waiting for a PTO.

* Callers that pass oversized `Transmit::contents` (partial #2201) — e.g. >65535 bytes total, which the kernel rejects with `EMSGSIZE`. These are now broken into kernel-acceptable batches, and each chunk still uses GSO when its segment count is > 1.

The public contract of `max_gso_segments` is unchanged — callers should still query it and size their `Transmit`s accordingly. This path is a defensive net so that callers which don't (or which raced a GSO halt) don't silently drop datagrams.

Cellular Android, where the kernel reports GSO support but the radio driver returns `EIO` on the first real send, is the original motivation: recovery now completes inside the `sendmsg` call instead of waiting for a ~1s handshake PTO. Firefox is hot-fixing the same case in D265812 and Firezone hit it independently (firezone/firezone#10279).

Fixes #2399.
Refs #2201.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -416,11 +416,110 @@ impl UdpSocketState {
     }
 }
 
+/// Maximum UDP payload bytes packable into one `sendmsg` call.
+///
+/// Two upper bounds:
+///   * `u16::MAX - ip_udp_header_overhead`: the kernel builds one IP packet per
+///     `sendmsg` whose 16-bit total-length field caps header + payload at
+///     `u16::MAX`. IPv4 headers can include up to 40 bytes of options (60 total);
+///     IPv6's fixed header is 40; UDP is 8.
+///   * `max_gso_segments * segment_size`: the kernel rejects GSO sends with more
+///     than `UDP_MAX_SEGMENTS` segments; we cap our own per-batch segment count
+///     at the runtime budget.
+///
+/// The batch may end with a partial segment, so a transmit whose total byte
+/// length is at most `max_batch_bytes` ships in a single `sendmsg` even when
+/// `contents.len() / segment_size` exceeds the segment budget by one partial
+/// segment.
+#[cfg(any(not(any(apple, target_os = "openbsd", target_os = "netbsd")), test))]
+fn max_batch_bytes(max_gso_segments: usize, segment_size: usize, destination: SocketAddr) -> usize {
+    let header_overhead = match destination {
+        SocketAddr::V4(_) => 60 + 8,
+        SocketAddr::V6(_) => 40 + 8,
+    };
+    let max_payload = (u16::MAX as usize) - header_overhead;
+    max_payload.min(max_gso_segments * segment_size)
+}
+
+/// Number of bytes to send in the next `sendmsg`, given `remaining` bytes left
+/// in the transmit and a per-call budget of `max_bytes`.
+///
+/// `max_batch_bytes` can land mid-segment (the `u16::MAX` byte cap is rarely a
+/// multiple of `segment_size`). But the kernel segments a GSO buffer purely by
+/// byte offset — every `segment_size` bytes — with no knowledge of intended
+/// datagram boundaries. So an *intermediate* batch must stop on a segment
+/// boundary; otherwise the split emits a short datagram in the middle of the
+/// stream and shifts every segment after it. Only the genuine final batch (the
+/// remainder of the whole transmit) may carry a partial last segment.
+#[cfg(any(not(any(apple, target_os = "openbsd", target_os = "netbsd")), test))]
+fn next_batch_bytes(remaining: usize, max_bytes: usize, segment_size: usize) -> usize {
+    if remaining <= max_bytes {
+        // Final batch: send what's left, partial last segment and all.
+        remaining
+    } else {
+        // Intermediate batch: round the byte budget down to whole segments, but
+        // always make progress even if a single segment exceeds the budget.
+        (max_bytes / segment_size).max(1) * segment_size
+    }
+}
+
+/// Send a `Transmit` as one or more `sendmsg` calls.
+///
+/// Each iteration packs up to `max_bytes` bytes (chosen so that both the
+/// kernel byte limit and the runtime GSO segment budget are respected) into
+/// one syscall, rounded down to a segment boundary by [`next_batch_bytes`] so
+/// intermediate batches never split a segment. `prepare_msg` re-checks
+/// `effective_segment_size` per batch and elides `UDP_SEGMENT` when the batch
+/// holds one segment or less, so the partial-last-batch and no-GSO cases fall
+/// out automatically.
+///
+/// If a multi-segment batch fails with `EIO`/`EINVAL` (common on Android
+/// cellular: the kernel reports GSO support but the radio driver rejects the
+/// actual send), the loop halts GSO globally, drops `max_bytes` to one
+/// segment, and retries the failed batch as individual datagrams.
 #[cfg(not(any(apple, target_os = "openbsd", target_os = "netbsd")))]
-fn send(
-    #[allow(unused_variables)] // only used on Linux
+fn send(state: &UdpSocketState, io: SockRef<'_>, transmit: &Transmit<'_>) -> io::Result<()> {
+    let segment_size = transmit.segment_size.unwrap_or(transmit.contents.len());
+    let mut max_bytes =
+        max_batch_bytes(state.max_gso_segments(), segment_size, transmit.destination);
+
+    let mut pos = 0;
+    while pos < transmit.contents.len() {
+        let remaining = transmit.contents.len() - pos;
+        let batch_end = pos + next_batch_bytes(remaining, max_bytes, segment_size);
+        let batch = Transmit {
+            destination: transmit.destination,
+            ecn: transmit.ecn,
+            contents: &transmit.contents[pos..batch_end],
+            // Keep `segment_size` set: `prepare_msg` consults `effective_segment_size`
+            // and elides the GSO cmsg automatically when a batch only holds one segment.
+            segment_size: transmit.segment_size,
+            src_ip: transmit.src_ip,
+        };
+        match send_packet(state, &io, &batch) {
+            Ok(()) => pos = batch_end,
+            // Multi-segment `EIO`/`EINVAL` likely means GSO isn't actually supported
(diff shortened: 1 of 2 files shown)


thomaseizinger:
> (untested, and mostly AI generated)

FTR: I don't mind this in principle and I see you've marked it as draft for now. Let us know when you've reviewed it yourself and done some testing on it.

I am glad this is being worked on. We do this on top of `quinn-udp` right now. I suggest you have a look at https://github.com/firezone/firezone/blob/445f46da6273dcce917a197f0747f91af782624b/rust/libs/connlib/socket-factory/src/lib.rs#L412 too.

quinn-udp/src/unix.rs
```diff
@@ -464,6 +488,71 @@ fn send(
...
+/// kernel's per-`sendmsg` byte limit are unchanged: callers should still
+/// query `max_gso_segments` and size their `Transmit`s accordingly. This
+/// path exists so that callers which don't (or which raced a GSO halt)
+/// don't silently drop datagrams.
+#[cfg(any(target_os = "linux", target_os = "android"))]
+fn send_chunked(
+    state: &UdpSocketState,
+    io: SockRef<'_>,
+    transmit: &Transmit<'_>,
+    max_per_chunk: usize,
```
flub: is there any practical value to making this dynamic rather than simply hardcoding this to 1?

quinn-udp/src/unix.rs
```diff
@@ -464,6 +488,71 @@ fn send(
...
+) -> io::Result<()> {
+    let segment_size = transmit
+        .effective_segment_size()
+        .expect("caller verified segment_size is set");
```
flub: If there isn't, can this not safely create a single chunk and send it instead of having to panic? The invariants being held so far from each other often ends up in tears sometimes. I feel like we need to try harder to avoid these things.

quinn-udp/src/unix.rs
kixelated: Yeah it's hard to follow, working on renaming stuff and avoiding some recursion now.

`max_per_chunk` should be called `gso_segments` or something. It's the maximum number of segments that can be contained in a single packet, initially starting at >1 (depending on the OS) and shrinking down to 1 on an error.

If it shrinks down to 1, we call send again to repeat this logic, but `gso_segments` will be 1 now.

quinn-udp/src/unix.rs
kixelated: This should be passed in as an argument instead of recomputing it anyway.

kixelated:
kk I simplified the logic and removed the recursion. It's still untested (on my Mac and the customer has the failing Android anyway).

Some of the comments are still AI generated, but I've closely reviewed everything.

quinn-udp/src/unix.rs
```diff
@@ -418,53 +498,46 @@ fn send(
...
+        // disabling `IP_TOS` is a one-way door (it stays off for the life of
+        // the socket), and we don't want a GSO-related `EINVAL` to permanently
+        // strip ECN.
+        //
+        // Note: Checking the local `transmit` (rather than the socket's
+        // `max_gso_segments`) keeps this race-free against a concurrent
+        // `send_batched` halting GSO mid-syscall.
+        if e.raw_os_error() == Some(libc::EINVAL)
+            && !state.sendmsg_einval()
+            && transmit.effective_segment_size().is_none()
```
kixelated, line 579: This fixes an existing bug, where a GSO failure could also disable ECN. Can split into a separate PR if you'd prefer.

quinn-udp/src/unix.rs
```diff
@@ -389,6 +385,89 @@ fn send(
...
+    // This is more efficient than sendmmsg because the NIC splits a single buffer at MTU boundaries.
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    if let Some(segment_size) = transmit.effective_segment_size() {
+        // Get the (GSO) segment size, and compute the number of segments we could fit into a UDP payload.
+        let segment_count = max_segments_per_send(state.max_gso_segments(), segment_size);
+        return send_batched(state, io, transmit, segment_size, segment_count);
+    }
+
+    // Fall back to sending each packet individually.
+    send_packet(state, &io, transmit)
```
thomaseizinger: I don't really understand this. There should not be a need for branching here because a single packet is a batch of size one with segment_size == packet_len, right?

quinn-udp/src/unix.rs
```diff
@@ -389,6 +385,89 @@ fn send(
...
+}
+
+/// Maximum total payload accepted by a single `sendmsg` syscall.
+///
+/// The kernel rejects any GSO send whose total `contents` length exceeds
+/// `u16::MAX`, regardless of `max_gso_segments`.
+#[cfg(any(target_os = "linux", target_os = "android", test))]
+const MAX_SEND_BYTES: usize = u16::MAX as usize;
```
thomaseizinger: As I've already pointed out, it is not that easy. You need to take into account IP and UDP headers here. See the link I've left in the previous comment.

thomaseizinger (changes requested):
Left two initial comments.

I also think we should have integration-level test coverage for this.

quinn-udp/src/unix.rs
kixelated: It's mostly a micro-optimization to avoid running the `send_batched` loop/logic if GSO isn't enabled/supported. 

You're right that we could `effective_segment_size` instead of `transmit.segment_size`. It would end up being a batch of 1 packet.

Lemme see if I can combine the code paths to avoid the branching.

kixelated:
@thomaseizinger better?

thomaseizinger (commented):
Overall better yes but I am not sure we are going into the right direction in general. Some questions to think about:

- Shouldn't this logic apply to all code-paths, i.e. Windows + Apple as well? The way I implemented this in Firezone, it is running above `quinn-udp`, regardless of whether GSO / apple-fast path is active.
- What about integration-level tests? Whether or not we are doing the slicing correctly is only surfaced if we actually hit a kernel and pass it a giant `Transmit`, i.e. one with a content that is larger than `u16::MAX`.

If we don't apply this loop globally, then I think we should inline this into the current `send` function. We already have a loop there and we just need to keep looping and call `prepare_msg`, right?

kixelated:
> Shouldn't this logic apply to all code-paths, i.e. Windows + Apple as well? The way I implemented this in Firezone, it is running above quinn-udp, regardless of whether GSO / apple-fast path is active.

I had Claude whip up [something](https://github.com/kixelated/quinn/pull/1) based on what I think you're suggested. But I don't really understand the suggestion and I don't see the purpose, except to increase the scope?

AFAIK `sendmsg_x` is sendmmsg, not sendmsg + GSO. The batching logic is completely different?

thomaseizinger:
> AFAIK `sendmsg_x` is sendmmsg, not sendmsg + GSO. The batching logic is completely different?

The mechanism is different but the API exposed by `quinn-udp` is the same: Take in a giant `Transmit` and send it with as little syscall overhead as possible.

For connected UDP sockets, `sendmsg_x` has a fast-path in the XNU kernel where the buffers get fast tracked through the layers so it is not quite the same as `sendmmsg` on Linux.

In both cases though, the provided `Transmit` already needs to be appropriately sized according to the current GSO configuration (i.e. batch-size). Passing to big of a buffer, i.e. not doing the slicing, will result in IO errors.

kixelated:
@thomaseizinger so I tested on Windows and it doesn't have the same u16::MAX limitation. IDK about Apple.

My main goal is to retry failed packets when disabling GSO. Right now WebSocket is winning the race with QUIC because it takes 500ms+ (or whatever) to retransmit the dropped Initial packets. The current behavior really hurts connection time for Android.

IDK I can make a smaller PR that doesn't touch the batch size if you want. I don't really want to get sucked into scope creep if it makes it harder to land a fix.

thomaseizinger:
> Right now WebSocket is winning the race with QUIC because it takes 500ms+ (or whatever) to retransmit the dropped Initial packets. The current behavior really hurts connection time for Android.

I am only a maintainer of `quinn-udp` so I can't speak for `quinn` but an easy, targeted fix should be possible inside `quinn` itself the same way we currently immediately retry this in Firezone without waiting for a retransmit.

kixelated:
> > Right now WebSocket is winning the race with QUIC because it takes 500ms+ (or whatever) to retransmit the dropped Initial packets. The current behavior really hurts connection time for Android.
> 
> I am only a maintainer of `quinn-udp` so I can't speak for `quinn` but an easy, targeted fix should be possible inside `quinn` itself the same way we currently immediately retry this in Firezone without waiting for a retransmit.

I don't think so, because you have to know if the EIO error occurred while GSO was enabled? Otherwise you would retry on EIO errors after GSO was disabled?

And I mean the layering just seems gross. Like you'd have to know you're supposed to retry specific errors... the library itself should just retry.

kixelated:
I made https://github.com/quinn-rs/quinn/pull/2724 as the "simpler and actually fixes my bug" PR. It's still more correct to split oversized GSO sends but it's a separate issue.

thomaseizinger:
Thanks for building this out. I don't think we need it actually. If we end up merging #2748, then the retry loop that every user has to build automatically handles oversized chunks.

kixelated:
> Thanks for building this out. I don't think we need it actually. If we end up merging #2748, then the retry loop that every user has to build automatically handles oversized chunks.

So long as something like #2724 merges please. QUIC doesn't work on (some?) Android devices without it, as TCP wins the race.

I would highly recommend merging #2724 first but I really don't care. Whatever is easier to review.

PR #2659, opened 2026-05-27
fix: Patches for Redox targets

willnode:
These are patches to support compiling for Redox targets. It is already been used to port Servo and Firefox to the platform.

The lack of IPv6 support is intentional, there are plans to it but it will be months or years.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -782,7 +782,7 @@ fn prepare_msg(
                 }
             }
             #[cfg(target_os = "redox")]
-            IpAddr::V6(v6) => {}
+            IpAddr::V6(_) => {}
             #[cfg(not(target_os = "redox"))]
             IpAddr::V6(v6) => {
                 let pktinfo = libc::in6_pktinfo {
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -68,6 +68,7 @@ impl UdpSocketState {
     pub fn new(sock: UdpSockRef<'_>) -> io::Result<Self> {
         let io = sock.0;
         let mut cmsg_platform_space = 0;
+        #[cfg(not(target_os = "redox"))]
         if cfg!(target_os = "linux")
             || cfg!(bsd)
             || cfg!(apple)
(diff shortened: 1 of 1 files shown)


quinn-udp/src/unix.rs
```diff
@@ -774,6 +781,9 @@ fn prepare_msg(
                     }
                 }
             }
+            #[cfg(target_os = "redox")]
+            IpAddr::V6(v6) => {}
```
mxinden:
```suggestion
            IpAddr::V6(_) => {}
```

`v6` is unused otherwise, no?

quinn-udp/src/unix.rs
```diff
@@ -173,6 +174,7 @@ impl UdpSocketState {
         }
 
         // Options standardized in RFC 3542
+        #[cfg(not(target_os = "redox"))]
```
mxinden, line 177: If the user requests an IPv6 socket (i.e. `is_ipv4` is false), this would simply ignore the users choice on redox, correct?

mxinden (commented):
Thanks for the work.

Does servo use quinn as its QUIC stack?

quinn-udp/src/unix.rs
willnode: Correct, but redox doesn't have `libc::IPPROTO_IPV6` and other ipv6 constants at this moment

willnode:
Sorry, I looked at Servo's Cargo.lock and there isn't anything mentioned quinn. I must have mistaken it with other ports as this patch been on my stash for three months and I repurpose it again for [firefox](https://github.com/mozilla/neqo/pull/3652).

Nonetheless there have been a number of rust programs unable to compile due to `quinn-udp` in Redox OS.

quinn-udp/src/unix.rs
mxinden: To make sure I understand correctly, you are saying that a user could not request an IPv6 socket in the first place, as the symbols aren't even available? I wonder whether we can be more careful here, in case Redox every adds support, e.g. via an error or a `debug_assert`. Will defer to you and others. Not a blocker.

mxinden:
> Sorry, I looked at Servo's Cargo.lock and there isn't anything mentioned quinn. I must have mistaken it with other ports as this patch been on my stash for three months and I repurpose it again for [firefox](https://github.com/mozilla/neqo/pull/3652).

Thanks for clarifying. Yes, Firefox uses quinn-udp. (:wave: Firefox maintainer here).

mxinden (approved):

quinn-udp/src/unix.rs
willnode: Yes, the symbols aren't even available. If I remove this, the crate wouldn't compile. All software ported in Redox OS is having IPv6 disabled at this moment.

I could came back removing this later when IPv6 is available.

djc (approved):

PR #2660, opened 2026-05-28
fix: set loss detection timer on path validation failure

### LINKED ISSUE #2658: Invariant violation on loss detection timeout: peer should not have completed address validation
We've been experimenting with connection migration and triggered a debug assertion in [`pto_time_and_space`](https://github.com/quinn-rs/quinn/blob/b2f4d932a21bc7b2ac1d142bd2320b1d0a222bb9/quinn-proto/src/connection/mod.rs#L1860).

https://github.com/quinn-rs/quinn/blob/b2f4d932a21bc7b2ac1d142bd2320b1d0a222bb9/quinn-proto/src/connection/mod.rs#L1855-L1860

<details>
<summary>Partial backtrace</summary>

```
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
   8: core[a4b43c77cd16befc]::panicking::panic<unknown>
   9: <quinn_proto[162b9cdc5a2f85b2]::connection::Connection>::pto_time_and_space<unknown>
  10: <quinn_proto[162b9cdc5a2f85b2]::connection::Connection>::on_loss_detection_timeout<unknown>
  11: <quinn_proto[162b9cdc5a2f85b2]::connection::Connection>::handle_timeout<unknown>
  12: <quinn[28efa520a44e82ae]::connection::State>::drive_timer<unknown>
  13: <quinn[28efa520a44e82ae]::connection::ConnectionDriver as core[a4b43c77cd16befc]::future::future::Future>::poll<unknown>
```

</details>

Our current reproduction is as follows:
- 2 peers talking on an IPv6 link
- the address used by peerA expires and is not replaced by a new preferred address
- the kernel on peerA uses another address to emit the UDP packets, triggering a path migration on peerB
- but peerB cannot reach the new address of peerA (so all packets to the new address of peerA are lost)

I've tried to come up with a test like [`migration()`](https://github.com/quinn-rs/quinn/blob/b2f4d932a21bc7b2ac1d142bd2320b1d0a222bb9/quinn-proto/src/tests/mod.rs#L1340) but did not succeed so far.

The assertion triggers about a minute after the initial migration attempt. Adding logs showed that we reach `on_loss_detection_timeout` with `self.loss_time_and_space() == None` so we end up calling `pto_time_and_space()` that contains the assertion. At this point I'm confused: peerB is a server so `peer_completed_address_validation()` is always `true`, and the assertion is always bound to trigger when we not longer have ACK-eliciting packets in flight? 

https://github.com/quinn-rs/quinn/blob/b2f4d932a21bc7b2ac1d142bd2320b1d0a222bb9/quinn-proto/src/connection/mod.rs#L1663-L1673

stormshield-fabs:
Reset the loss detection timer on path validation failure.

Fixes https://github.com/quinn-rs/quinn/issues/2658

In our tests, the previous path no longer has in-flight ack-eliciting packets when restored, so the invariant that the loss timeout handler should not be called when not needed no longer holds.

I have not been able to write a unit test that reaches the debug assertion but the failure rate on our IPv6 migration scenario that initially triggered this investigation went from ~35% to 0% with this patch.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1199,6 +1199,7 @@ impl Connection {
                     debug!("path validation failed");
                     if let Some((_, prev)) = self.prev_path.take() {
                         self.path = prev;
+                        self.set_loss_detection_timer(now);
                     }
                     self.path.challenge = None;
                     self.path.challenge_pending = false;


Ralith (approved):
Your analysis makes sense to me, thanks!

djc (approved):
Nice work, thanks!

PR #2661, opened 2026-05-31
Add ClientHello server config selection

iadev09:
## Summary

This is the 0.11-compatible callback bridge for selecting a rustls server config
from ClientHello data.

Related to quinn-rs/quinn#2024.

It targets the current rustls 0.23 integration.

The intended shape is:

```rust
let crypto = ClientHelloServerConfig::new(default_server_config, move |hello| {
    match hello.server_name() {
        Some("host-b.test") => Some(host_b_server_config.clone()),
        _ => None,
    }
})?;

let server_config = quinn::ServerConfig::with_crypto(Arc::new(crypto));
```

Returning `Some(config)` continues with that selected rustls server config.
Returning `None` falls back to the default config.

## What This Solves

This allows one Quinn server endpoint to serve multiple prebuilt rustls server
configs selected from ClientHello data.

That can be used for different certificate stores, or for different TLS policy
shapes that need separate `rustls::ServerConfig` values, while still using the
same Quinn endpoint/socket.

## Scope

- opt-in: only used when the caller wraps its crypto config in
  `ClientHelloServerConfig`
- keeps Quinn's existing `Incoming` API shape
- synchronous and snapshot/in-memory oriented
- parses only enough ClientHello data to select a config
- intended as a compatibility bridge

## Limitations

Because this bridge does not use a rustls `Acceptor`, the opt-in path parses the
small ClientHello subset needed for selection, and rustls later parses
ClientHello again during the normal handshake.

That duplication is intentional for this compatibility path. The longer-term
rustls-0.24 `Acceptor` direction avoids this by keeping ClientHello parsing
inside rustls.

## Test

Adds regression coverage:

```text
select_server_config_from_client_hello
```

The test verifies that the callback sees SNI and ALPN from ClientHello, selects
the matching server config, and the resulting connection reports the selected
ALPN/SNI in handshake data.

### DIFF
--- a/quinn-proto/src/crypto/rustls.rs
+++ b/quinn-proto/src/crypto/rustls.rs
@@ -1,4 +1,4 @@
-use std::{any::Any, io, str, sync::Arc};
+use std::{any::Any, io, mem, str, sync::Arc};
 
 #[cfg(all(feature = "aws-lc-rs", not(feature = "ring")))]
 use aws_lc_rs::aead;
@@ -11,7 +11,7 @@ use rustls::NamedGroup;
 use rustls::{
     self, CipherSuite,
     client::danger::ServerCertVerifier,
-    pki_types::{CertificateDer, PrivateKeyDer, ServerName},
+    pki_types::{CertificateDer, DnsName, PrivateKeyDer, ServerName},
     quic::{Connection, HeaderProtectionKey, KeyChange, PacketKey, Secrets, Suite, Version},
 };
 #[cfg(feature = "platform-verifier")]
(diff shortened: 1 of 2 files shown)


iadev09:
This can technically live outside Quinn via `ServerConfig::with_crypto()` and
the public crypto traits.

```rust
impl quinn::crypto::ServerConfig for DeferredServerConfig
```

```rust
impl quinn::crypto::Session for DeferredSession
```

But the implementation has to duplicate the
Quinn/rustls QUIC glue, so keeping a small bridge in Quinn may be more
consistent with the existing rustls integration.

djc:
Suggest making smaller commits that separate refactoring from adding new code. I'm not convinced we'll accept something like this for 0.11.x since it seems relatively complex and I'm not sure it will integrate well.

(As requested previously, do not use LLMs to generate PR descriptions and comments.)

iadev09:
Quinn Api exports enough to implement this externally (as a wrapper) without changes to Quinn, without need to merge  Quinn changes to the external crate also. 

(noted)

PR #2670, opened 2026-06-04
Add type-erased TLS metadata to rustls handshake data

### LINKED ISSUE #2662: Expose negotiated TLS metadata for QUIC connections
## Summary

Quinn's rustls `HandshakeData` already exposes ALPN (`protocol`) and SNI
(`server_name`), and `Connection::peer_identity()` exposes peer certificates.
However, there does not seem to be a public way to read the negotiated TLS
protocol version or cipher suite for a QUIC connection.

For HTTP/1.1 and HTTP/2 over TCP, server integrations using rustls can expose
request metadata such as:

- `SSL_PROTOCOL`
- `SSL_CIPHER`
- client certificate subject/issuer

For HTTP/3 over Quinn, the same server can expose client certificates through
`peer_identity()`, and can infer TLS 1.3 because QUIC requires it, but cannot
expose the negotiated cipher suite.

## Use Case

Edge/server frameworks often want H1/H2/H3 request-context parity. For example,
CGI/FastCGI/PHP-style environments commonly expose TLS metadata to application
code. With Quinn today, the HTTP/3 path cannot fill `SSL_CIPHER` without
reaching into Quinn/rustls internals.

The use case is connection observability and H1/H2/H3 server-framework metadata parity.

The cipher suite is the practical missing piece. Exposing the TLS protocol
version as well would make the TLS metadata complete and avoid special-casing
QUIC as "always TLS 1.3" in server integrations.

## Possible Shape

Would Quinn be open to exposing negotiated TLS metadata in rustls
`HandshakeData`, for example:

```rust
pub struct HandshakeData {
    pub protocol: Option<Vec<u8>>, // ALPN
    pub server_name: Option<String>,  
    pub protocol_version: Option<rustls::ProtocolVersion>,
    pub negotiated_cipher_suite: Option<rustls::CipherSuite>,
}

### PR DESCRIPTION
**Pull request by iadev09**
Closes  #2662.

This adds the negotiated TLS protocol version and cipher suite to rustls `HandshakeData` as type-erased values (`Box<dyn Any>`).

I also extended the existing `alpn_success` test to cover the new metadata.

### DIFF
--- a/quinn-proto/src/crypto/rustls.rs
+++ b/quinn-proto/src/crypto/rustls.rs
@@ -63,10 +63,20 @@ impl crypto::Session for TlsSession {
         }
         Some(Box::new(HandshakeData {
             protocol: self.inner.alpn_protocol().map(|x| x.into()),
-            server_name: match self.inner {
+            server_name: match &self.inner {
                 Connection::Client(_) => None,
-                Connection::Server(ref session) => session.server_name().map(|x| x.into()),
+                Connection::Server(session) => session.server_name().map(|x| x.into()),
             },
+            protocol_version: match &self.inner {
+                Connection::Client(session) => session.protocol_version(),
+                Connection::Server(session) => session.protocol_version(),
+            }
+            .map(|x| -> Box<dyn Any> { Box::new(x) }),
+            cipher_suite: match &self.inner {
+                Connection::Client(session) => session.negotiated_cipher_suite(),
+                Connection::Server(session) => session.negotiated_cipher_suite(),
+            }
+            .map(|suite| -> Box<dyn Any> { Box::new(suite.suite()) }),
             #[cfg(feature = "__rustls-post-quantum-test")]
             negotiated_key_exchange_group: self
                 .inner
@@ -265,6 +275,10 @@ pub struct HandshakeData {
     ///
     /// Always `None` for outgoing connections
     pub server_name: Option<String>,
+    /// The protocol version negotiated with the peer, if any
+    pub protocol_version: Option<Box<dyn Any>>,
+    /// The cipher suite negotiated with the peer, if any
+    pub cipher_suite: Option<Box<dyn Any>>,
     /// The key exchange group negotiated with the peer
     #[cfg(feature = "__rustls-post-quantum-test")]
     pub negotiated_key_exchange_group: NamedGroup,
--- a/quinn-proto/src/crypto/rustls.rs
+++ b/quinn-proto/src/crypto/rustls.rs
@@ -266,6 +266,7 @@ impl HeaderKey for Box<dyn HeaderProtectionKey> {
 }
 
 /// Authentication data for (rustls) TLS session
+#[non_exhaustive]
 pub struct HandshakeData {
     /// The negotiated application protocol, if ALPN is in use
     ///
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: quinn-proto/src/crypto/rustls.rs [hunk 1]
```diff
@@ -50,6 +50,21 @@ impl TlsSession {
...
+        }
+    }
+
+    fn negotiated_cipher_suite(&self) -> Option<CipherSuite> {
+        match self.inner {
+            Connection::Client(ref session) => session.negotiated_cipher_suite(),
+            Connection::Server(ref session) => session.negotiated_cipher_suite(),
+        }
+        .map(|suite| suite.suite())
+    }
```
djc: Let's inline these into `handshake_data()`? Please use match ergonomics (match `&self.inner` so you don't need the `ref`).

quinn-proto/src/crypto/rustls.rs
djc, line 270: Can you add an additional commit that makes this `#[non_exhaustive]`?

iadev09: [Make rustls HandshakeData non-exhaustive](https://github.com/quinn-rs/quinn/commit/9bad3a21b68f09b2980945a1f673821bd937f95c) added

quinn-proto/src/crypto/rustls.rs
iadev09: inlined to `handshake_data()`

quinn-proto/src/crypto/rustls.rs
```diff
@@ -63,10 +63,21 @@ impl crypto::Session for TlsSession {
...
+            server_name: match &self.inner {
                 Connection::Client(_) => None,
-                Connection::Server(ref session) => session.server_name().map(|x| x.into()),
+                Connection::Server(session) => session.server_name().map(|x| x.into()),
             },
+            protocol_version: match &self.inner {
+                Connection::Client(session) => session.protocol_version(),
+                Connection::Server(session) => session.protocol_version(),
+            }
+            .map(|x| Box::new(x) as Box<dyn Any>),
```
djc: Do we need the as cast here?

quinn-proto/src/crypto/rustls.rs
```diff
@@ -63,10 +63,21 @@ impl crypto::Session for TlsSession {
...
+            cipher_suite: match &self.inner {
+                Connection::Client(session) => session.negotiated_cipher_suite(),
+                Connection::Server(session) => session.negotiated_cipher_suite(),
+            }
+            .map(|suite| suite.suite())
```
djc: Can collapse the two map() calls, supposedly?

djc (approved):

quinn-proto/src/crypto/rustls.rs
iadev09: Removed the `as` cast

quinn-proto/src/crypto/rustls.rs
iadev09: collapsed this into a single `map()`.

djc:
Please rebase instead of adding merge commits.

iadev09:
Rebased.

PR #2671, opened 2026-06-05
Add staged ClientHello acceptor for ServerConfig selection

### LINKED ISSUE #2024: Access Client Hello before accepting connection

iadev09:
Closes #2024.

Adds a staged ClientHello accept path so a server can inspect the rustls `ClientHello` before finalizing the Quinn `ServerConfig`.

### Commit Structure

**1. rustls 0.24 upgrade path**

Kept as one compatibility commit. This is where the temporary rustls git dependencies live, including the rustls-platform-verifier 0.8 branch.

- [Update rustls dependencies to upstream 0.24](https://github.com/quinn-rs/quinn/pull/2671/changes/04e1efcde3fbdcaea9be71829243da8e3f8d8ece)

Related rustls 0.24 PR outside Quinn: rustls/rustls-platform-verifier#233.

**2. staged ClientHello acceptor primitive**

- [Refactor rustls transport error mapping](https://github.com/quinn-rs/quinn/pull/2671/changes/8e0e5429637e71ba1f8693b349530daaf1c9ad50)
- [Start server sessions from rustls Accepted](https://github.com/quinn-rs/quinn/pull/2671/changes/b62040a8bdc441d8b04a5aca6eb8b436cc1e9d44)
- [Extract shared server accept finalization](https://github.com/quinn-rs/quinn/pull/2671/changes/bd22484d7c85488e7695953f784cd58982b314cc)
- [Make connection assembler available to endpoint](https://github.com/quinn-rs/quinn/pull/2671/changes/334c9e88f3c07765b7dca9ee4193f1a683c38dad)
- [Add staged rustls ClientHello acceptor](https://github.com/quinn-rs/quinn/pull/2671/changes/d5a33747d4c37d1a9d101221f4884ee520458745)
- [Support multi-packet staged ClientHello](https://github.com/quinn-rs/quinn/pull/2671/changes/b5a7edd499c44614789f78d839faf8e7cb48d9b7)
- [Create connection from rustls Accepted](https://github.com/quinn-rs/quinn/pull/2671/changes/64a2ad9ec29128649e4829caa00de815555949a7)
- [Expose staged acceptor on Incoming](https://github.com/quinn-rs/quinn/pull/2671/changes/bf5d395f9e0b0aa3deed202022d4944a10797b5a)
- [Test staged rustls acceptor paths](https://github.com/quinn-rs/quinn/pull/2671/changes/e105ce88df3872ec2236fcdeb5f31dab0d0f0d44)

**3. selector helper**

Adds a `ServerConfigSelector` helper trait on top of the staged acceptor API.

- [Add ServerConfigSelector helper](https://github.com/quinn-rs/quinn/pull/2671/changes/e42adb1afc99c7e9bba94b55147ec48c3aa5deb3)

```rust
pub trait ServerConfigSelector {
    /// Error returned by the selector.
    type Error;

    /// Select a server config for the provided ClientHello.
    fn select_server_config<'a>(
        &'a self,
        client_hello: rustls::server::ClientHello<'a>,
    ) -> impl Future<Output = Result<Option<Arc<ServerConfig>>, Self::Error>> + 'a;
}
```

### Tests

```sh
cargo test -p quinn incoming_acceptor \
  --features runtime-tokio,rustls-aws-lc-rs \
  --no-default-features \
  --locked \
  -- --nocapture

cargo test -p quinn incoming_acceptor \
  --features runtime-tokio,rustls-ring \
  --no-default-features \
  --locked \
  -- --nocapture
```

### Bench

https://github.com/iadev09/quinn-acceptor-bench 

In local 1000-sample smoke runs, the selected path stayed in the same range as continuing with the default `ServerConfig`.

### Comparison

Compared with the TCP/rustls shape, this exposes the staged primitive directly.

For TCP, config selection usually needs an intermediate acceptor/future shape:

```rust
let acceptor =
    LazyConfigAcceptor::new(rustls::server::Acceptor::default(), tcp_stream);
tokio::pin!(acceptor);
let start = acceptor.as_mut().await?;
```

For Quinn, the proposed primitive lets the caller inspect the `ClientHello` and choose the `quinn::ServerConfig` directly.

Because the `Endpoint` already owns the Initial/CRYPTO handling, it does not need a `LazyConfigAcceptor`-style stream wrapper.

```rust
let accepted = incoming.acceptor()?.await?;
let hello = accepted.client_hello();
let config = app_state.select_server_config(hello).await?;
let connection = accepted.accept_with(config)?.await?;
```

Same scenario on TCP/rustls `LazyConfigAcceptor`:

https://github.com/iadev09/lazy-config-acceptor-bench

> This only measures the selector/no-selector difference; TCP connect and kernel transport work are not included.

### DIFF
--- a/quinn/Cargo.toml
+++ b/quinn/Cargo.toml
@@ -28,10 +28,10 @@ platform-verifier = ["proto/platform-verifier"]
 # For backwards compatibility, `rustls` forwards to `rustls-ring`
 rustls = ["rustls-ring"]
 # Enable rustls with the `aws-lc-rs` crypto provider
-rustls-aws-lc-rs = ["dep:rustls", "aws-lc-rs", "proto/rustls-aws-lc-rs", "proto/aws-lc-rs"]
-rustls-aws-lc-rs-fips = ["dep:rustls", "aws-lc-rs-fips", "proto/rustls-aws-lc-rs-fips", "proto/aws-lc-rs-fips"]
+rustls-aws-lc-rs = ["__rustls", "dep:rustls-aws-lc-rs", "aws-lc-rs", "proto/rustls-aws-lc-rs", "proto/aws-lc-rs"]
+rustls-aws-lc-rs-fips = ["rustls-aws-lc-rs", "aws-lc-rs-fips", "proto/rustls-aws-lc-rs-fips", "proto/aws-lc-rs-fips"]
 # Enable rustls with the `ring` crypto provider
-rustls-ring = ["dep:rustls", "ring", "proto/rustls-ring", "proto/ring"]
+rustls-ring = ["__rustls", "dep:rustls-ring", "ring", "proto/rustls-ring", "proto/ring"]
 # Enable the `ring` crypto provider.
 # Outside wasm*-unknown-unknown targets, this enables `Endpoint::client` and `Endpoint::server` conveniences.
 ring = ["proto/ring"]
@@ -41,14 +41,15 @@ runtime-smol = ["dep:async-io", "dep:smol"]
 # Configure `tracing` to log events via `log` if no `tracing` subscriber exists.
 tracing-log = ["tracing/log", "proto/tracing-log", "udp/tracing-log"]
 # Enable rustls logging
-rustls-log = ["rustls?/logging"]
+rustls-log = ["rustls?/log"]
 # Enable qlog support
 qlog = ["proto/qlog"]
 
 # Internal (PRIVATE!) features used to aid testing.
 # Don't rely on these whatsoever. They may disappear at any time.
 
-__rustls-post-quantum-test =  ["rustls/prefer-post-quantum", "rustls-aws-lc-rs", "proto/__rustls-post-quantum-test"]
+__rustls-post-quantum-test =  ["rustls-aws-lc-rs", "proto/__rustls-post-quantum-test"]
+__rustls = ["dep:rustls"]
 
 [dependencies]
 async-io = { workspace = true, optional = true }
@@ -59,6 +60,8 @@ rustc-hash = { workspace = true }
 pin-project-lite = { workspace = true }
 proto = { package = "quinn-proto", path = "../quinn-proto", version = "0.12.0", default-features = false }
 rustls = { workspace = true, optional = true }
+rustls-aws-lc-rs = { workspace = true, optional = true }
+rustls-ring = { workspace = true, optional = true }
 smol = { workspace = true, optional = true }
 thiserror = { workspace = true }
 tracing =  { workspace = true }
@@ -78,6 +81,7 @@ bencher = { workspace = true }
 directories-next = { workspace = true }
 rand = { workspace = true }
 rcgen = { workspace = true }
+rustls-util = { workspace = true }
 clap = { workspace = true }
 tokio = { workspace = true, features = ["rt", "rt-multi-thread", "time", "macros", "test-util"] }
 tracing-subscriber = { workspace = true }
@@ -92,23 +96,23 @@ workspace = true
 
 [[example]]
 name = "server"
-required-features = ["rustls-ring"]
+required-features = ["__rustls"]
 
 [[example]]
 name = "client"
-required-features = ["rustls-ring"]
+required-features = ["__rustls"]
 
 [[example]]
 name = "insecure_connection"
-required-features = ["rustls-ring"]
+required-features = ["__rustls"]
 
 [[example]]
 name = "single_socket"
-required-features = ["rustls-ring"]
+required-features = ["__rustls"]
 
 [[example]]
 name = "connection"
-required-features = ["rustls-ring"]
+required-features = ["__rustls"]
 
 [[test]]
 name = "post_quantum"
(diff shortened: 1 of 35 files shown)


quinn/Cargo.toml
```diff
@@ -15,23 +15,21 @@ rust-version.workspace = true
 [features]
 # NOTE: Please keep this in sync with the feature list in `.github/workflows/codecov.yml`, see
 # comment in that file for more information.
-default = ["tracing-log", "platform-verifier", "runtime-tokio", "rustls-ring", "bloom"]
```
djc, line 18: I think we'll want to retain this. Would you mind also doing a rustls 0.24 branch for rustls-platform-verifier? I think it should be straightforward.

quinn/Cargo.toml
```diff
@@ -15,23 +15,21 @@ rust-version.workspace = true
...
 # Records how long locks are held, and warns if they are held >= 1ms
 lock_tracking = []
-# Provides `ClientConfig::with_platform_verifier()` convenience method
-platform-verifier = ["proto/platform-verifier"]
 # For backwards compatibility, `rustls` forwards to `rustls-ring`
 rustls = ["rustls-ring"]
 # Enable rustls with the `aws-lc-rs` crypto provider
-rustls-aws-lc-rs = ["dep:rustls", "aws-lc-rs", "proto/rustls-aws-lc-rs", "proto/aws-lc-rs"]
-rustls-aws-lc-rs-fips = ["dep:rustls", "aws-lc-rs-fips", "proto/rustls-aws-lc-rs-fips", "proto/aws-lc-rs-fips"]
+rustls-aws-lc-rs = ["__rustls", "dep:rustls-aws-lc-rs", "aws-lc-rs", "proto/rustls-aws-lc-rs", "proto/aws-lc-rs"]
```
djc, line 31: I think this feature refactoring can be done separately, earlier (suggest in a separate PR). rustls 0.24 requires that callers provide a crypto provider explicitly. Since rustls is a public dependency for Quinn anyway, I suggest that we drop the provider-specific features in favor of requiring an explicit provider, too.

@Ralith do you agree?

If so, I'd suggest that the explicit provider passing can also become a separate commit that can be merged ahead of this one. (May obviate the need for a `__rustls` feature?)

quinn/Cargo.toml
iadev09: While exercising this path, we had already prepared rustls 0.24 branches for the half-sibling crates involved here: `tokio-rustls`, `hyper-rustls`, and `rustls-platform-verifier`.

I can open the others too if that split is useful.

iadev09: https://github.com/rustls/rustls-platform-verifier/pull/233

.github/workflows/codecov.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         #       | paste -sd ',' -
         run: |
           cargo llvm-cov \
-            --features="arbitrary,aws-lc-rs,bloom,log,fast-apple-datapath,futures-io,json-output,lock_tracking,tracing-log,qlog,ring,runtime-smol,runtime-tokio,rustls,rustls-aws-lc-rs,rustls-log,rustls-ring,serde,serde_json,smol,tracing" \
+            --features="arbitrary,aws-lc-rs,bloom,log,fast-apple-datapath,futures-io,json-output,lock_tracking,tracing-log,platform-verifier,qlog,ring,runtime-smol,runtime-tokio,rustls,rustls-aws-lc-rs,rustls-log,rustls-ring,serde,serde_json,smol,tracing" \
```
djc: Please squash this back into the original commit that removed it.

quinn/Cargo.toml
Ralith: > I suggest that we drop the provider-specific features in favor of requiring an explicit provider, too.

Yep, that should be much more composable. We may still want to spend some effort having a reasonable default, but we should try to avoid needing to know about other providers.

djc: I guess the challenge here is that we still have the raw primitives in use for packet protection. Maybe the first objective is finishing https://github.com/quinn-rs/quinn/compare/main...ctz:quinn:jbp-retry-calculation-uses-rustls...

.github/workflows/codecov.yml
iadev09: Squashed back and removed the unrelated codecov diff.

djc:
So I think if you want to make progress on getting this merged, this needs to be done as a separate PR:

> I guess the challenge here is that we still have the raw primitives in use for packet protection. Maybe the first objective is finishing [main...ctz:quinn:jbp-retry-calculation-uses-rustls](https://github.com/quinn-rs/quinn/compare/main...ctz:quinn:jbp-retry-calculation-uses-rustls)...

I think the upgrade to 0.24 should also be a separate PR -- not sure which of these two makes sense to merge first.

iadev09:
I split the rustls 0.24 upgrade out into a separate branch first.  #2701 

After that, I can look at the retry packet protection cleanup as another small PR before rebasing the staged ClientHello acceptor work on top.

PR #2674, opened 2026-06-08
Backport of #2495 to 0.11.x

stablebits:
Backport of "Add early return in RecvStream::drop when all data read is true" https://github.com/quinn-rs/quinn/pull/2495 to 0.11.x.

What's included: the original 3 commits:
  f6113170  Move the ref counts out  [ conflicts due to 'State struct' divergence were resolved manually ]
  298a7c18  Fix the (pre-existing) rightward drift by inverting conditions
  80914039  Early return in `RecvStream::drop()`
    
 and 2 related fixes:
  5a26dc36  Remove `RecvStream`s from `blocked_readers` on `stop`
  dfe72254  quinn: fix ref count logic for ConnectionRef and EndpointRef

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -928,7 +928,7 @@ impl Clone for ConnectionRef {
 
 impl Drop for ConnectionRef {
     fn drop(&mut self) {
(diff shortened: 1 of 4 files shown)


djc (approved):
Please also submit a PR with version bumps for the relevant crate(s).

Issue #753, opened 2020-05-07
Prevent starvation by tokio's operation budget


Ralith:
Tokio's [automatic cooperative task yielding](https://tokio.rs/blog/2020-04-preemption/) feature can cause abruptly prevent a task from making further progress. A very busy connection driver might as a result end up failing to fairly serve transmits or timeouts, resulting in bad behavior (e.g. spurious timeouts or congestion window reduction) under heavy load.

Ideally tokio will expose some tools to arrange for budget fairness across subtasks (e.g. annotations to designate subtasks that should have independent budgets), but if none are forthcoming, we could address this internally by e.g. varying the order we perform operations in by round robin. The current order was chosen for efficiency, however, so an upstream solution is preferred.

Ralith:
A minimal API to address this appears to have been added in tokio in https://github.com/tokio-rs/tokio/pull/3547.

syszery:
Hi! I'd like to take a look into this. 

I‘ve checked and I think this hasn‘t been addressed, yet. The way the connection driver handles Tokio's budget might also be a contributor to the problem described in #1126.

Ralith:
Please feel free!

syszery:
I think besides the fairness issue there is also the part that Tokio's coop layer seems to overload `Poll::Pending` with two meanings:

* "there is currently no work and a wakeup has been registered"
* "the coop budget was exhausted and the task has already been scheduled for another poll"

`ConnectionDriver::poll` may not distinguish between these two meanings of `Pending` when deciding:

```rust
let mut keep_going = conn.drive_transmit(cx)?;
keep_going |= conn.drive_timer(cx);
...
if !conn.inner.is_drained() {
    if keep_going {
        cx.waker().wake_by_ref();
    } else {
        conn.driver = Some(cx.waker().clone());
    }
    return Poll::Pending;
}
```

The transmit path seems to exhibit this ambiguity: `drive_transmit` returns `Ok(true)` only when it reaches `MAX_TRANSMIT_DATAGRAMS`, otherwise it returns `Ok(false)`. That means `Ok(false)` covers both:

* `poll_transmit()` returning `None`, and
* `poll_send()` returning coop-induced `Pending`.

The second case doesn't require the budget to be exhausted before `drive_transmit` starts. Since `process_conn_events` and `drive_transmit` share the same budget, it is enough for `process_conn_events` to leave less budget than `drive_transmit` needs to reach its own transmit cap.

[...]

PR #2677, opened 2026-06-09
fix(conn): handle overdue timers under Tokio budget exhaustion

syszery:
This fixes a correctness issue where QUIC protocol timers could fail to fire within a scheduling round under load when Tokio’s cooperative task budget is exhausted.

`drive_timer()` previously relied on `AsyncTimer::poll()` to determine whether a deadline had elapsed. However, `Sleep::poll()` may return `Poll::Pending` for already-expired deadlines once the task has exhausted its cooperative budget. In that case, `handle_timeout()` would not be called until a later scheduling round, despite the deadline having already passed.

This addresses the timer starvation aspect of #753.

### Fix

The timer logic is changed to:

* Always check `runtime.now() >= deadline` before polling the async timer
* Use the async timer only for future wakeups (waker registration)

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -1207,6 +1207,16 @@ impl State {
             return false;
         };
 
+        // Use the clock rather than the async timer to detect expiry: Sleep::poll
+        // respects Tokio's cooperative budget and can return Pending for elapsed
+        // deadlines.
+        let now = self.runtime.now();
+        if now >= deadline {
+            self.inner.handle_timeout(now);
+            self.timer_deadline = None;
+            return true;
+        }
+
         match &mut self.timer {
             // Avoid resetting the timer when the deadline is unchanged.
             Some(delay) if self.timer_deadline != Some(deadline) => {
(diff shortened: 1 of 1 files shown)


djc:
(Not sure why this is draft? Will hold off on reviewing for now.)

syszery:
Thanks. I just double checked something. The PR is ready for review now.

Ralith (approved):

djc:
I'd like to have this split in two commits:

- One to use let-else instead of `match`
- Another to change the logic

A bunch of comments were removed -- were those removals motivated or just LLM collateral?

syszery:
The PR is split into two commits now.

The comment changes were intentional:

* The first two described the old tangled `match` structure, which I think is easier to follow after the `let...else` extraction.
* The "store expiration time" comment was just restating the line below it.
* The final two became slightly inaccurate after the fix.

djc (approved):

djc:
Thanks! Happy to take a backport for this one, too.

PR #2678, opened 2026-06-10
Add rustls-no-provider

### LINKED ISSUE #2341: Support rustls feature without using ring or aws-lc-rs
As the title says, support rustls without specifically specifying the underlying provider.

A lot of libraries expose some way of picking the provider but also allow for no provider to be installed.

notpeter:
- Add: new crate feature `rustls-no-provider` in `quinn` and `quinn-proto`
- Closes: https://github.com/quinn-rs/quinn/issues/2341
- See also: https://github.com/seanmonstar/reqwest/issues/2935

This is my first quinn contribution, so I'm very open to feedback. I did my best to preserve  the existing structure and make this change as minimal as possible.

CC: @TroyKomodo is this something you can test for your use case?

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -27,7 +27,7 @@ jobs:
         #       | paste -sd ',' -
         run: |
           cargo llvm-cov \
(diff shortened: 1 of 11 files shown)


djc:
I think you might want to do this against the `0.11.x` branch?

For `main`, I think this will go away via a different route:

https://github.com/quinn-rs/quinn/pull/2671#discussion_r3372535699

notpeter:
> I think you might want to do this against the `0.11.x` branch?

I've rebased my branch and retargeted this PR.
 
> For `main`, I think this will go away via a different route:
> 
> [#2671 (comment)](https://github.com/quinn-rs/quinn/pull/2671#discussion_r3372535699)

Oh, that makes sense.  That might make this PR superfluous if that's coming in the near future.

djc:
Okay, I'm not willing to take on a dependency, even optional, on any RustCrypto crates.

PR #2679, opened 2026-06-11
Abstract `Instant` for server-side WASM support

itpetey:
When building for `wasm(n)` targets, Quinn defaults to `web_time::Instant`, which only works when run inside the browser. For server-side runtime environments, e.g. `wasmtime`, `wasmer` etc. we need a different abstraction.

This PR allows implementors to provide that abstraction.

### DIFF
--- a/quinn/src/runtime/mod.rs
+++ b/quinn/src/runtime/mod.rs
@@ -11,12 +11,48 @@ use std::{
 
 use udp::{RecvMeta, Transmit};
 
-use crate::Instant;
+/// Abstracts the `Instant` type for server-side WASM environments that require their own
+/// timer implementations.
+pub trait RuntimeInstant: Clone {
+    /// The `Duration` type used by the abstraction
+    type Duration;
+
+    /// Returns an instant corresponding to "now".
+    fn now() -> Self;
+
+    /// Returns the amount of time elapsed from another instant to this one,
+    /// or zero duration if that instant is later than this one.
+    fn duration_since(&self, earlier: Self) -> Self::Duration;
+
+    /// Returns the amount of time elapsed from another instant to this one,
+    /// or None if that instant is later than this one.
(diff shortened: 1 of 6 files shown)


djc:
> we need a different abstraction.

Who's we and are you going to help fund our maintenance of your niche requirements? 😛

quinn/src/runtime/mod.rs
```diff
@@ -11,12 +11,48 @@ use std::{
...
+/// Abstracts the `Instant` type for server-side WASM environments that require their own
+/// timer implementations.
+pub trait RuntimeInstant: Clone {
+    /// The `Duration` type used by the abstraction
+    type Duration;
+
+    /// Returns an instant corresponding to "now".
+    fn now() -> Self;
+
+    /// Returns the amount of time elapsed from another instant to this one,
```
djc, line 23: All docstrings should be start with a single-line sentence fragment that does not end in a period.

djc (commented):
I'm a little confused what you think this achieves, given that, at the quinn-proto layer, we're still using `std::time::Instant` or `web_time::Instant` exclusively.

(And, if you were to fix that, I'd like to see a benchmark report on the impact of virtualizing all those method calls.)

itpetey:
> I'm a little confused what you think this achieves, given that, at the quinn-proto layer, we're still using std::time::Instant or web_time::Instant exclusively.

I think that quinn-proto doesn't call `Instant::now()` outside of tests and the "qlog" feature, which is the only problematic method that needed addressing.

For anyone else with "niche requirements", I came up with an easier workaround:

```rust
impl quinn::Runtime for MyNicheWasmRuntime {
    // ...

    fn now(&self) -> web_time::Instant {
        let now = get_time_from_host_call(); // <-- Your host call abstraction
        let d = Duration::from_nanos(now);
        Instant::from(d)
    }
}
```

Unfortunately this relies on a separate PR: https://github.com/daxpedda/web-time/pull/47, but is a far smaller change.

PR #2683, opened 2026-06-12
quinn-proto: drop Initials silently when saturated

stablebits:
When the `max_incoming` queue is full (or CIDs are exhausted), quinn replied to each Initial with CONNECTION_REFUSED. Building that reply derives the packet's initial keys, which is computationally expensive.
A flood of Initials then forces the endpoint's packet-processing task to do per-Initial crypto work, leaving it less time for legitimate packets and degrading already-established connections.

So drop Initials silently when saturated.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -430,6 +430,16 @@ impl Endpoint {
             return None;
         }
 
+        // Saturation only happens under heavy load, where deriving initial keys per Initial just to
+        // reply with CONNECTION_REFUSED would starve packet processing for existing connections.
+        if self.cids_exhausted() || self.incoming_buffers.len() >= server_config.max_incoming {
+            debug!(
+                "ignoring initial for connection {} due to saturation",
+                dst_cid
+            );
+            return None;
+        }
+
         let crypto = match server_config.crypto.initial_keys(header.version, dst_cid) {
             Ok(keys) => keys,
             Err(UnsupportedVersion) => {
@@ -669,11 +679,6 @@ impl Endpoint {
         &mut self,
         header: &ProtectedInitialHeader,
     ) -> Result<(), TransportError> {
-        let config = &self.server_config.as_ref().unwrap();
-        if self.cids_exhausted() || self.incoming_buffers.len() >= config.max_incoming {
-            return Err(TransportError::CONNECTION_REFUSED(""));
-        }
-
         // RFC9000 §7.2 dictates that initial (client-chosen) destination CIDs must be at least 8
         // bytes. If this is a Retry packet, then the length must instead match our usual CID
         // length. If we ever issue non-Retry address validation tokens via `NEW_TOKEN`, then we'll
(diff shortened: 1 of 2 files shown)


djc:
Don't we already have `ignore()` for this?

alexpyattaev:
> Don't we already have `ignore()` for this?

`.ignore` operates after Endpoint has spent the CPU time processing the initial_keys in Incoming packets.

stablebits:
> > Don't we already have `ignore()` for this?
> 
> `.ignore` operates after Endpoint has spent the CPU time processing the initial_keys in Incoming packets.

A user app would need to get `incoming` from `endpoint.accept()` in order to call `ignore()`, wouldn't it? But this won't happen if the endpoint's incomings queue is full...

Ralith:
Perhaps silently dropping unrecognized packets when `max_incoming` is saturated should be the default/only behavior? That condition should only arise under heavy load, after all.

djc:
> Perhaps silently dropping unrecognized packets when `max_incoming` is saturated should be the default/only behavior? That condition should only arise under heavy load, after all.

Seems reasonable to me.

stablebits:
> > Perhaps silently dropping unrecognized packets when `max_incoming` is saturated should be the default/only behavior? That condition should only arise under heavy load, after all.
> 
> Seems reasonable to me.

Made it default/only-behaviour under saturation

quinn-proto/src/endpoint.rs
```diff
@@ -425,18 +418,27 @@ impl Endpoint {
...
                 .map(DatagramEvent::Response);
-        };
+        }
 
         if datagram_len < MIN_INITIAL_SIZE as usize {
             debug!("ignoring short initial for connection {}", dst_cid);
             return None;
         }
 
+        let server_config = self.server_config.as_ref().unwrap();
```
Ralith: Why was this change made? It seems purely cosmetic. Cosmetic changes should be in dedicated commits, and in any case I prefer the previous version that doesn't involve an `unwrap`.

quinn-proto/src/endpoint.rs
Ralith, line 667: What's the reasoning behind this wording change? It is still true that this method does not consider the packet's contents.

quinn-proto/src/endpoint.rs
```diff
@@ -425,18 +418,27 @@ impl Endpoint {
...
+        if self.cids_exhausted() || self.incoming_buffers.len() >= server_config.max_incoming {
```
Ralith, line 435: It'd be nice to include a comment justifying this behavior by capturing the conclusion of the discussion in the PR.

Ralith (approved):
LGTM, some nits.

quinn-proto/src/endpoint.rs
stablebits: restored the original

quinn-proto/src/endpoint.rs
stablebits: added a comment

djc (approved):
Nice! Feel free to backport.

PR #2684, opened 2026-06-14
Connection: Make poll_timeout immutable

ThomasHabets:
Looks like `poll_timeout` is needlessly `mut`. At least I can't find the reason.


djc (approved):

djc:
Thanks!

Ralith (approved):

PR #2688, opened 2026-06-15
0.11.x backport | quinn-proto: drop Initials silently when saturated

stablebits:
When the `max_incoming` queue is full (or CIDs are exhausted), quinn replied to each Initial with CONNECTION_REFUSED. Building that reply derives the packet's initial keys, which is computationally expensive. A flood of Initials then forces the endpoint's packet-processing task to do per-Initial crypto work, leaving it less time for legitimate packets and degrading already-established connections.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -435,6 +435,16 @@ impl Endpoint {
             return None;
         }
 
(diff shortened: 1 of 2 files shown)


djc (approved):

stablebits:
Could you please release new 0.11.x quinn and quinn-proto releases?

PR #2691, opened 2026-06-18
Fix Windows IPv6 bind option selection

### LINKED ISSUE #2682: Panic when binding to non-wildcard ipv6 on Windows 11
In latest commit, when binding to `[::1]:4443` on windows11, quinn-udp panics reporting `os error 10022`

This command could reproduce:
```
 cargo run --example server -p quinn -- /
```

This only appears on windows. Linux/MacOS works as usual

nkgotcode:
Fixes #2682.

This changes the Windows UDP socket setup so IPv4 packet-info options are only
enabled for IPv4 sockets and wildcard dual-stack IPv6 sockets. Non-wildcard
IPv6 binds such as `[::1]:4433` no longer take the IPv4 option path.

Testing:

- `cargo test -p quinn-udp --no-default-features`

### DIFF
--- a/quinn-udp/src/lib.rs
+++ b/quinn-udp/src/lib.rs
@@ -84,6 +84,14 @@ mod log {
 #[cfg(not(wasm_browser))]
 pub use imp::UdpSocketState;
 
+#[cfg_attr(not(windows), allow(dead_code))]
+fn should_set_ipv4_options(addr: SocketAddr, v6only: bool) -> bool {
+    match addr {
+        SocketAddr::V4(_) => true,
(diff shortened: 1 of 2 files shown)


Ralith:
Thanks for looking into this! The issue claims that the bug was introduced in https://github.com/quinn-rs/quinn/commit/b8e4d3b9acf205c5c3304fe901ac40d80408e199, which doesn't involve quinn-udp, so I don't think these changes are sufficient, and I'm not sure if they're necessary either, though they do make sense intuitively.

Please add an end-to-end test at the quinn level to be sure we capture (and don't regress) the behavior of the whole stack.

PR #2700, opened 2026-06-22
Implement `Connection::watch_remote_address`

stormshield-fabs:
As discussed on [Discord](https://discord.com/channels/976380008299917365/1063547094863978677/1507029204083150869), this PR adds a `watch_remote_address` method on `quinn::Connection` that allows the user to watch for remote address changes without having to periodically call the existing `remote_address` method.

The implementation uses a `tokio::sync::watch` channel since it has good semantics. I wrapped it in a new type that does not leak this implementation detail.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -1201,6 +1201,7 @@ impl Connection {
                     if let Some((_, prev)) = self.prev_path.take() {
                         self.path = prev;
                         self.set_loss_detection_timer(now);
+                        self.events.push_back(Event::PathUpdated);
                     }
                     self.path.challenge = None;
                     self.path.challenge_pending = false;
@@ -3129,6 +3130,8 @@ impl Connection {
         let prev_pto = self.pto(SpaceId::Data);
 
         let mut prev = mem::replace(&mut self.path, new_path);
+        self.events.push_back(Event::PathUpdated);
+
         // Don't clobber the original path if the previous one hasn't been validated yet
         if prev.challenge.is_none() {
             prev.challenge = Some(self.rng.random());
@@ -4055,6 +4058,8 @@ pub enum Event {
     DatagramReceived,
     /// One or more application datagrams have been sent after blocking
     DatagramsUnblocked,
+    /// The currently active path was updated
+    PathUpdated,
 }
 
 fn get_max_ack_delay(params: &TransportParameters) -> Duration {
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -19,7 +19,7 @@ use thiserror::Error;
 use tokio::sync::{
     Notify,
     futures::{Notified, OwnedNotified},
-    mpsc, oneshot,
+    mpsc, oneshot, watch,
 };
 use tracing::{Instrument, Span, debug_span};
 
@@ -559,6 +559,21 @@ impl Connection {
         self.0.state.lock("remote_address").inner.remote_address()
     }
 
+    /// Watch the peer's UDP address.
+    ///
+    /// If `ServerConfig::migration` is `true`, clients may change addresses at will, e.g. when
+    /// switching to a cellular internet connection.
+    pub fn watch_remote_address(&self) -> RemoteAddressWatcher {
+        let receiver = self
+            .0
+            .state
+            .lock("watch_remote_address")
+            .remote_address
+            .subscribe();
+
+        RemoteAddressWatcher { receiver }
+    }
+
     /// The local IP address which was used when the peer established
     /// the connection
     ///
@@ -944,6 +959,25 @@ impl Future for SendDatagram<'_> {
     }
 }
 
+/// Watcher produced by [`Connection::watch_remote_address`]
+pub struct RemoteAddressWatcher {
+    receiver: watch::Receiver<SocketAddr>,
+}
+
+impl RemoteAddressWatcher {
+    /// Wait for an address change, mark it as seen and return the address.
+    ///
+    /// If multiple migrations occur between calls to `wait` or polls of the resulting future,
+    /// the last observed address will be returned, skipping intermediate updates.
+    pub async fn wait(&mut self) -> Option<SocketAddr> {
+        if self.receiver.changed().await.is_err() {
+            return None;
+        }
+
+        Some(*self.receiver.borrow())
+    }
+}
+
 #[derive(Debug)]
 pub(crate) struct ConnectionRef(Arc<ConnectionInner>);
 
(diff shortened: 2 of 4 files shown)


quinn/src/connection.rs
```diff
@@ -547,6 +547,21 @@ impl Connection {
         self.0.state.lock("remote_address").inner.remote_address()
     }
 
+    /// Watch the peer's UDP address.
+    ///
+    /// If `ServerConfig::migration` is `true`, clients may change addresses at will, e.g. when
+    /// switching to a cellular internet connection.
```
Ralith, line 565: Let's also document that this is never expected for outgoing connections.

quinn-proto/src/connection/mod.rs
Ralith, line 1: Test coverage for the new event?

quinn/src/connection.rs
```diff
@@ -547,6 +547,21 @@ impl Connection {
...
+    pub fn watch_remote_address(&self) -> WatchRemoteAddress {
```
Ralith: From a user's perspective, it's not clear to me why we return a helper object here, rather than simply having a `pub async fn remote_address_changed(&self)` or similar. Can we do that instead for simplicity? Perhaps the receiver could go in `Shared`.

Ralith (changes requested):

quinn/src/connection.rs
stormshield-fabs: If my understanding is correct, you're proposing to have a single `watch::Receiver`?

I think the semantics would be a bit strange: since `Connection` implements `Clone`, do we want the notifications to be "spread" across all instances? I'm thinking about a situation where the user calls `remote_address_changed` from two different places and thus only sees half of the events on each side.

To me, it's less surprising to have this helper object that clearly holds the state.

Ralith: The correct behavior is for notifications to be broadcast against any callers interested in them. That should match the behavior you get in that scenario in practice. The state is in the future itself.

stormshield-fabs: I'm sorry I don't understand where you're trying to go.

I think storing a `watch::Receiver` in `Shared`, cloning it in `remote_address_changed` and waiting for a change will not work since new calls will clone the same initial state and trigger immediately without waiting for a change.

Subscribing to the `watch::Sender` on each call has slightly better semantics (that's what the current implementation does, and we could indeed remove the helper object). However, updates sent before the subscription are ignored, so if the endpoint sees a migration before the user subscribes to updates, they will have to wait for the next migration, which sounds unsatisfying as well?

Maybe `watch` is not the correct tool to reach for in our case? Or maybe I'm missing something obvious.

Ralith: This fell off my stack, sorry to leave you hanging!

> However, updates sent before the subscription are ignored, so if the endpoint sees a migration before the user subscribes to updates, they will have to wait for the next migration, which sounds unsatisfying as well?

This hazard exists regardless. Either whatever mechanism we expose always yields the current address first, or there's a race between a migration happening and the user registering interest in migrations. The solution is to create the future, then poll the current state and compare it to what they last observed, and *only then* wait on the future.

However, on review I think I agree that this is too difficult for a user who wants to continuously monitor migrations. Just as Tokio undoubtedly realized in the first place, keeping track of the last seen value automatically is a valuable convenience, so I'm satisfied that the currently written approach is the better option.

quinn/src/connection.rs
```diff
@@ -932,6 +947,27 @@ impl Future for SendDatagram<'_> {
     }
 }
 
+/// Watcher produced by [`Connection::watch_remote_address`]
+pub struct WatchRemoteAddress {
+    receiver: watch::Receiver<SocketAddr>,
+}
+
+/// Error    that can arise when watching the remote address
```
Ralith: Stray whitespace?

Ralith: Document the significance of the error. The reader should be left with no doubt as to what they should do about this condition.

quinn/src/connection.rs
```diff
@@ -932,6 +947,27 @@ impl Future for SendDatagram<'_> {
...
+#[derive(Debug, Error, Clone, Eq, PartialEq)]
+#[error("wait() called on a dropped connection")]
+pub struct WatchRemoteAddressError;
+
+impl WatchRemoteAddress {
+    /// Wait for an address change, mark it as seen and return it.
+    pub async fn wait(&mut self) -> Result<SocketAddr, WatchRemoteAddressError> {
```
Ralith: Please document the behavior when multiple migrations occur in between calls to wait or polls of a wait future.

Ralith: Would it be simpler/more obvious to return `Option` here? It's basically an async iterator, after all.

Ralith (changes requested):

quinn/src/connection.rs
stormshield-fabs: Sure

quinn-proto/src/connection/mod.rs
stormshield-fabs: done

stormshield-fabs:
Thanks for the review @Ralith! Sorry for my own delay, I was working at a summer camp.

I've addressed your comments and added a test for the new API.

quinn/src/connection.rs
```diff
@@ -944,6 +959,25 @@ impl Future for SendDatagram<'_> {
     }
 }
 
+/// Watcher produced by [`Connection::watch_remote_address`]
+pub struct WatchRemoteAddress {
```
Ralith: bikeshed: I think we should prefer to name types with noun phrases, e.g. maybe `RemoteAddressWatcher`?

Ralith (approved):
LGTM, thanks!

quinn/src/connection.rs
stormshield-fabs: Other type names in the file (`ReadDatagram`, `AcceptBi`, `OpenUni`) don't appear to follow this convention but I don't have strong feelings about it.

Rebased and updated to follow your suggestion.

Ralith: Verb phrases do make sense for futures, but this is something else 👍

djc (approved):

quinn-proto/src/connection/mod.rs
```diff
@@ -4055,6 +4058,8 @@ pub enum Event {
     DatagramReceived,
     /// One or more application datagrams have been sent after blocking
     DatagramsUnblocked,
+    /// The currently active path was updated
+    PathUpdated,
```
djc, line 4062: Should we make `Event` `non_exhaustive`?

Ralith: Probably, yeah. Historically we've only ever added stuff to it that would be fine for existing code to ignore.

stormshield-fabs: I've opened https://github.com/quinn-rs/quinn/pull/2780 to address this.

PR #2705, opened 2026-06-26
quinn-proto: fix unresolved intra-doc link for platform verifier

MerlijnW70:
`QuicClientConfig`'s docs link to
`ClientConfig::try_with_platform_verifier()`, which is gated behind
`#[cfg(feature = "platform-verifier")]`. When docs are built without that
feature, the intra-doc link is unresolved and `cargo doc` fails under
`-D warnings` (the same bar CI enforces).

This gates the doc sentence and its link reference to the same feature, so
the link is only present when its target is. Verified `cargo doc --no-deps`
is clean both with and without `--features platform-verifier`.

No code or API changes.

### DIFF
--- a/quinn-proto/src/crypto/rustls.rs
+++ b/quinn-proto/src/crypto/rustls.rs
@@ -288,7 +288,12 @@ pub struct HandshakeData {
 /// A QUIC-compatible TLS client configuration
 ///
 /// Quinn implicitly constructs a `QuicClientConfig` with reasonable defaults within
-/// [`ClientConfig::with_root_certificates()`][root_certs] and [`ClientConfig::try_with_platform_verifier()`][platform].
+/// [`ClientConfig::with_root_certificates()`][root_certs].
+#[cfg_attr(
+    feature = "platform-verifier",
+    doc = "It is also constructed implicitly within [`ClientConfig::try_with_platform_verifier()`][platform]."
+)]
+///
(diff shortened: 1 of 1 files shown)


MerlijnW70:
The `test on illumos` job failed before the build started — it couldn't fetch a dependency from crates.io:

```
error: failed to get `rustix` as a dependency of package `async-io v2.6.0`
  download of ru/st/rustix failed
  curl failed
  [55] Failed sending data to the peer (OpenSSL SSL_read: SSL_ERROR_SYSCALL, errno 0)
```

That's a transient network error on the runner, unrelated to this change (which only adjusts doc comments). All other jobs are green, including `lint`/docs and the freebsd and solaris VM jobs. A re-run of the illumos job should clear it — I don't have permission to trigger it on this repo.

quinn-proto/src/crypto/rustls.rs
```diff
@@ -288,7 +288,12 @@ pub struct HandshakeData {
 /// A QUIC-compatible TLS client configuration
 ///
 /// Quinn implicitly constructs a `QuicClientConfig` with reasonable defaults within
-/// [`ClientConfig::with_root_certificates()`][root_certs] and [`ClientConfig::try_with_platform_verifier()`][platform].
+/// [`ClientConfig::with_root_certificates()`][root_certs].
```
djc, line 291: This seems pretty messy. How about we just change this to "with reasonable defaults in [`ClientConfig`] constructors".

Issue #2702, opened 2026-06-23
Additional `quinn-proto` robustness panic candidates


xc01:
I found several generated `quinn-proto` panic sites. The strongest ones are unchecked arithmetic/state handling in token logging and congestion accounting. Some of the other generated cases depend on custom crypto implementations or invalid enum values, so I would keep those as lower-priority notes rather than lead with them.

I checked nearby comments/docs around the main sites below and did not find `# Panics` coverage for these conditions.

Version checked: `quinn-proto 0.11.14`

## Main examples

| Area | Panic site | Trigger |
| --- | --- | --- |
| Token replay log | `src/bloom_token_log.rs:74`, then `src/bloom_token_log.rs:70` | Extreme `issued + lifetime` overflows while the mutex is held, poisoning the token log; the next validation panics on `lock().unwrap()`. |
| NewReno congestion control | `src/congestion/new_reno.rs:58` | A very large ACK byte count overflows `self.window += bytes`. |
| BBR congestion control | `src/congestion/bbr/bw_estimation.rs:57`, `src/congestion/bbr/mod.rs:409`, `src/congestion/bbr/mod.rs:578` | Extreme ACK accounting values can overflow internal counters. |

## Example 1: `BloomTokenLog` can poison itself on extreme token lifetime

Relevant code:

```rust
let mut guard = self.0.lock().unwrap();
let state = &mut *guard;

// calculate how many periods past period 1 the token expires
let expires_at = issued + lifetime;
```

Reproducer shape:

```rust
use std::panic::{catch_unwind, AssertUnwindSafe};
use std::time::{Duration, SystemTime};
use quinn_proto::{BloomTokenLog, TokenLog};

#[test]
#[should_panic]
fn bloom_token_log_extreme_lifetime_poison_then_panics() {
    let log = BloomTokenLog::new_expected_items(1024, 16);

    let first = catch_unwind(AssertUnwindSafe(|| {
        let _ = log.check_and_insert(1, SystemTime::now(), Duration::from_secs(u64::MAX));
    }));
    assert!(first.is_err());

    let _ = log.check_and_insert(2, SystemTime::now(), Duration::from_secs(1));
}
```

Observed behavior:

The first call panics while adding a huge duration to `SystemTime`. Because this happens while the mutex guard is held, the mutex becomes poisoned. A later normal call then panics on `self.0.lock().unwrap()`.

Expected behavior:

[...]

PR #2706, opened 2026-06-26
quinn-proto: harden congestion arithmetic and resolve token cache panic surface

MerlijnW70:
While reviewing the congestion control and token management logic, I
identified several scenarios — particularly under extreme ACK-aggregation
and token-replay conditions — where `unwrap()` calls and arithmetic
operations could trigger panics.

This PR implements hardening measures so the protocol degrades gracefully
rather than crashing under these edge cases:

**Congestion control** — replaced potentially overflowing arithmetic in
NewReno and BBR with saturating operations (`saturating_add`,
`saturating_mul`) on the ACK-driven window, byte counters, and
ack-aggregation state.

**Token cache** — improved safety in `token_memory_cache.rs` and
`bloom_token_log.rs` by recovering from a poisoned lock
(`PoisonError::into_inner`) instead of `lock().unwrap()`, so one panicking
caller can no longer poison the cache for the whole system.
`BloomTokenLog` additionally computes token expiry with
`SystemTime::checked_add`, rejecting an overflowing lifetime as token reuse
instead of panicking while the lock is held.

**Regression tests** — added 3 unit tests driving these paths with extreme
inputs (`u64::MAX` byte counts, an overflowing token lifetime) to verify
they are now contained and panic-free. Each test panicked prior to the fix.

This addresses the stability concerns outlined in #2702. The changes are
regression-tested and introduce no public API signature changes.

### DIFF
--- a/quinn-proto/src/bloom_token_log.rs
+++ b/quinn-proto/src/bloom_token_log.rs
@@ -3,7 +3,7 @@ use std::{
     f64::consts::LN_2,
     hash::{BuildHasher, Hasher},
     mem::{size_of, take},
-    sync::Mutex,
+    sync::{Mutex, PoisonError},
 };
 
 use fastbloom::BloomFilter;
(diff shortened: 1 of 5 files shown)


djc:
Please use three separate commits for this.

MerlijnW70:
Split into three commits as requested — one per subsystem (congestion arithmetic, token-cache lock recovery, BloomTokenLog lifetime), each carrying its own regression test.

djc (approved):
Looks okay to me, thanks.

djc:
Are you interested in backporting these to the `0.11.x` branch?

PR #2707, opened 2026-06-29
0.11.x: upgrade dependencies

djc:
I think all of these changes are semver-compatible?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,9 +75,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.102"
+version = "1.0.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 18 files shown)


zh-jq:
can you also bump quinn-udp to a new version？

djc:
> can you also bump quinn-udp to a new version？

I don't think so, because quinn re-exports quinn-udp, so that would be a semver-incompatible change. Anything in particular you're looking for?

zh-jq:
I would like to see a new quinn-udp version that can use new windows-sys https://github.com/quinn-rs/quinn/pull/2689.

djc:
Ah, sorry, I misunderstood -- added a version bump commit for quinn-udp too.

In the future, feel free to submit a version bump PR if you think a release would be beneficial!

Ralith (approved):

Issue #2711, opened 2026-07-05
Fixed packet_threshold spuriously fast-retransmits under rate-scaled reordering displacement — consider RACK-style adaptation


length967:
## Summary

Loss detection's packet-count threshold (`TransportConfig::packet_threshold`, default 3 per RFC 9002) compares packet-number displacement, but reordering displacement is a *time* phenomenon: at a fixed reordering delay, displacement in packet numbers scales linearly with send rate (~130 PN per ms at 1.6 Gbit/s). So for any fixed threshold there is a rate above which benign reordering is declared loss, triggering spurious fast-retransmit storms and congestion response. Linux TCP handles this with adaptive reordering detection (RACK-TLP, `tcp_reordering` adapts 3→300); RFC 9002 §6.1.1 explicitly permits adapting the threshold.

## Measured dose-response

Same setup as https://github.com/quinn-rs/quinn/issues/2710 (197 ms real path, BBR, 512 MiB transfers, netem `delay 5ms reorder 25% 50%` — with the dedup-window issue from that report patched out so this effect is isolated):

| packet_threshold | goodput | stats().path.lost_packets |
|---|---|---|
| 3 (default) | 125 Mbit/s | ~65% of sent |
| 300 | 306 Mbit/s | ~64% |
| 1000 | 563 Mbit/s | ~52% |
| effectively ∞ | **632–710 Mbit/s** | **~14%** |
| (clean path reference) | 828 Mbit/s | ~1% |

With the threshold effectively disabled, recovery is purely time-based (~1.125×RTT) and remained healthy under genuine loss in our testing (1% and 10% netem loss rows unaffected) — for bulk transfer the packet-count path adds little except reorder-fragility.

## Secondary effect worth knowing

With defaults, the spurious-retransmit storm can escalate to a connection-fatal abort: the stream reassembler's anti-DoS cap (`assembler.rs`, >1024 buffered chunks after defragment) returns `TransportError::INTERNAL_ERROR("too many gaps in stream buffer")` and kills the connection. We observed this repeatedly under ±10 ms jitter; it disappears once the spurious losses do.

## Suggestion

[...]

PR #2712, opened 2026-07-05
proto: widen packet dedup window to 2048 bits

### LINKED ISSUE #2710: Dedup window (129 packets) silently discards delivered reordered packets — throughput collapse on reordering paths at Gbit rates
## Summary

`quinn-proto`'s packet dedup window is a `u128` bitmask (`connection/spaces.rs`: `type Window = u128`, currently main line ~568), i.e. 129 packets. Any packet arriving more than 129 packet numbers behind the highest authenticated one returns "might be duplicate" from `Dedup::insert` and the connection silently discards it (`debug!("discarding possible duplicate packet")`).

At WAN rates, milliseconds of path reordering span far more than 129 packet numbers (~130 PN per ms at 1.6 Gbit/s with 1452-byte packets). Result: delivered, decryptable data is discarded above the kernel, never ACKed, spuriously retransmitted — and the retransmits inflate packet-number spread, feeding back into more discards. Throughput collapses several-fold below TCP on the same path.

## Environment

- quinn 0.11.11 / quinn-proto 0.11.15 (defect present on current `main`)
- Two cloud VMs (8 vCPU), Sydney↔Chicago, 197 ms RTT, BBR congestion controller, 256 MiB windows, bulk transfer over uni streams (8 MiB per stream)
- Reordering injected with netem on the sender egress: `delay 5ms reorder 25% 50%` (note: netem sits above GSO segmentation, so it reorders ~44-packet super-packets — *milder* than per-packet reordering at the same percentage)

## Measured impact

| path condition | goodput |
|---|---|
| clean | 610–828 Mbit/s |
| 25% reorder / 5 ms displacement | **122–127 Mbit/s** (kernel TCP-BBR on the same path: 696) |

## Evidence that the "loss" is manufactured by the receiver

- netem qdisc stats after a 512 MiB transfer: `dropped 0` (all 1.06 M packets left the sender)
- receiver NIC `rx_packets` delta ≈ packets sent (1.044 M vs 1.046 M)
- receiver `UdpInErrors` / `UdpRcvbufErrors` / softnet drop deltas: 0
- yet `Connection::stats().path.lost_packets` ≈ **56% of sent**
- the loss is insensitive to `packet_threshold` (tested up to ∞) *combined with* `time_threshold(8.0)` — because loss detection isn't the cause; the receiver discarded the data before ACKing it
- GSO off (per-packet reordering, finer-grained displacement) makes it *worse*: 67% "lost"

## Fix tested

Drop-in widening of the window to a 2048-bit fixed bitset (`[u64; 32]` wrapper with the same shift/test/set semantics; `WINDOW_SIZE`/`BITFIELD_SIZE` formulas unchanged since they derive from `size_of::<Window>()`): +240 bytes per connection, replay-protection semantics identical, `Dedup` unit tests pass with width-dependent constants parametrized on `WINDOW_SIZE`.

Result on the same path/profile: 205 → 306 Mbit/s with a raised packet threshold, and 705–719 Mbit/s (≈ clean, ≥ TCP-BBR's 696) once combined with addressing the fixed reordering threshold (separate issue, linked below). Full investigation write-up with the elimination chain and numbers: https://github.com/length967/haul/blob/main/haul/wan-test/results-realworld-2026-07-05.md

Happy to open a PR with the widened window (possibly const-generic or config-gated if you'd rather not grow the default). A related, smaller observation from the same investigation: `MAX_ACK_BLOCKS = 64` evicts the lowest pending-ACK range via `pop_min()` under many-hole reordering, which discards acknowledgment state for delivered packets (we left it alone since the ACK frame encoder assumes the cap fits one min-MTU packet).

length967:
Fixes #2710.

## Problem

The `Dedup` window is a `u128` bitmask, so any packet arriving more than 129 packet numbers behind the highest authenticated one is treated as a possible duplicate and silently discarded — even though it was delivered and decrypts fine. Reordering displacement in packet-number space scales with send rate (~130 PN per ms of reordering at 1.6 Gbit/s with 1452-byte packets), so at WAN rates a few milliseconds of path reordering silently discards a large fraction of delivered packets. The peer never sees ACKs for them, spuriously retransmits, and the retransmissions widen the packet-number spread further — a feedback loop that collapsed measured throughput several-fold below TCP on the same path (details and elimination chain in #2710).

## Change

`type Window = u128` → a fixed 2048-bit window (`BigWindow([u64; 32])`) exposing the same operations `Dedup` used on the `u128` (`shl` saturating to zero, bit test/set, and the highest-zero-in-range scan used by `smallest_missing_in_interval`). `WINDOW_SIZE` and `BITFIELD_SIZE` are derived from `size_of::<Window>()` / `Window::BITS` as before, so every bound scales automatically. Replay-protection semantics are unchanged — the window is just wider. Cost: 240 additional bytes per connection.

Width-dependent constants in the `Dedup` tests are parametrized on `WINDOW_SIZE` so they express the same scenarios at any width; the remaining tests are untouched and pass unchanged (275/275 `quinn-proto --lib`).

## Measurements

Two 8-vCPU VMs, Sydney↔Chicago (197 ms RTT), BBR, 512 MiB bulk transfers, `netem delay 5ms reorder 25% 50%` on the sender egress:

| configuration | goodput |
|---|---|
| u128 window (this crate today), packet_threshold=300 | 205 Mbit/s |
| **2048-bit window (this PR)**, packet_threshold=300 | **306 Mbit/s** |
| 2048-bit window + large packet_threshold (see #2711) | 632–710 Mbit/s (clean-path: ~730–830) |

Also verified no regression on clean, 1%-loss, and 10%-loss profiles.

Happy to reshape (e.g. const-generic width or a config knob) if you'd prefer not to grow the default.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/quinn-proto/src/connection/spaces.rs
+++ b/quinn-proto/src/connection/spaces.rs
@@ -1,5 +1,4 @@
 use std::{
-    cmp,
     collections::{BTreeMap, VecDeque},
     mem,
     ops::{Bound, Index, IndexMut},
@@ -462,7 +461,10 @@ pub(super) struct Dedup {
 impl Dedup {
     /// Construct an empty window positioned at the start.
     pub(super) fn new() -> Self {
-        Self { window: 0, next: 0 }
+        Self {
+            window: Window::ZERO,
+            next: 0,
+        }
     }
 
     /// Highest packet number authenticated.
@@ -476,18 +478,17 @@ impl Dedup {
     pub(super) fn insert(&mut self, packet: u64) -> bool {
         if let Some(diff) = packet.checked_sub(self.next) {
             // Right of window
-            self.window = ((self.window << 1) | 1)
-                .checked_shl(cmp::min(diff, u64::from(u32::MAX)) as u32)
-                .unwrap_or(0);
+            let mut shifted = self.window.shl(1);
+            shifted.set(0);
+            self.window = shifted.shl(diff); // saturates to zero past the window width
             self.next = packet + 1;
             false
         } else if self.highest() - packet < WINDOW_SIZE {
             // Within window
             if let Some(bit) = (self.highest() - packet).checked_sub(1) {
                 // < highest
-                let mask = 1 << bit;
-                let duplicate = self.window & mask != 0;
-                self.window |= mask;
+                let duplicate = self.window.get(bit);
+                self.window.set(bit);
                 duplicate
             } else {
                 // == highest
@@ -533,16 +534,13 @@ impl Dedup {
             return None;
         }
 
-        // Ensure the shift is within bounds (we already know start_offset < BITFIELD_SIZE,
-        // because of the early return)
-        let mask = if range_len == BITFIELD_SIZE {
-            u128::MAX
-        } else {
-            ((1u128 << range_len) - 1) << start_offset
-        };
-        let gaps = !self.window & mask;
-
-        let smallest_missing_offset = 128 - gaps.leading_zeros() as u64;
+        // Highest un-received offset within the window slice (the bitwise
+        // `(!window & mask).leading_zeros()` scan, expressed on the wide window); earlier
+        // packets are considered received.
+        let smallest_missing_offset = self
+            .window
+            .highest_zero_in(start_offset, start_offset + range_len)?
+            + 1;
         let smallest_missing_packet = self.highest() - smallest_missing_offset;
 
         if smallest_missing_packet <= upper_bound {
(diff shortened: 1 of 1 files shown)


length967:
Withdrawing this PR — we'll carry the change in our fork for now. The analysis in #2710 stands if maintainers want to pick the approach up.

djc:
> Withdrawing this PR — we'll carry the change in our fork for now. The analysis in #2710 stands if maintainers want to pick the approach up.

Why?

PR #2719, opened 2026-07-07
quinn-udp: handle EIO from sendmsg in the cmsg fallback (not just EINVAL)

### LINKED ISSUE #2718: quinn-udp: sendmsg EIO from ECN control message not handled (cmsg fallback only triggers on EINVAL), no datagrams sent on some Android kernels
Fix: #2719

## Summary

On some Android kernels, `sendmsg` returns **`EIO`** (not `EINVAL`) when a control message it doesn't accept is present (observed with the **ECN** `IP_TOS` cmsg, alongside GSO). `quinn-udp`'s cmsg fallback (which sets `sendmsg_einval` to stop emitting the ECN cmsg and retries) is gated on **`EINVAL` only**. On `EIO` it disables GSO but keeps emitting the ECN cmsg, so every subsequent `sendmsg` keeps failing and **no datagrams are ever sent**, and the QUIC handshake times out.

## Environment

- `quinn-udp` 0.5.15 (via `quinn` 0.11.11, `quinn-proto` 0.11.16)
- Android **17**, Pixel 9a (`arm64-v8a`), app targeting `target_os = "android"`
- App has `INTERNET`; a raw `nc -u` (shell UID) to the same host:port works, so the network path is fine.

## Symptom (logcat)

```
quinn_udp::imp: `libc::sendmsg` failed with I/O error (os error 5); halting segmentation offload
quinn_udp: sendmsg error: Os { code: 5, kind: Uncategorized, message: "I/O error" },
  Transmit { destination: 192.168.1.32:4557, src_ip: None, ecn: Some(Ect0), len: 2400, segment_size: Some(1200) }
... (subsequent sends keep failing, rate-limited so only logged once) ...
QUIC connection failed: timed out
```

A UDP witness server on the destination receives **zero** packets from the app.

## Root cause

`src/unix.rs` (Linux/Android send error handling):

```rust
// halts GSO on EIO OR EINVAL:
if let Some(libc::EIO) | Some(libc::EINVAL) = e.raw_os_error() {
    if state.max_gso_segments() > 1 { /* store 1 */ }
}

// but the cmsg fallback (which drops the ECN IP_TOS cmsg via `sendmsg_einval` + retry)
// only triggers on EINVAL:
if e.raw_os_error() == Some(libc::EINVAL) && !state.sendmsg_einval() {
    state.set_sendmsg_einval();
    prepare_msg(...); // sendmsg_einval=true => no ECN cmsg
    continue;         // retry
}
```

So when the kernel reports **`EIO`** for the ECN cmsg, `sendmsg_einval` is never set, `prepare_msg` keeps pushing `IP_TOS`/ECN, and every send keeps failing.

## Suggested fix

Extend the `sendmsg_einval` fallback to also trigger on `EIO`:

```rust
if matches!(e.raw_os_error(), Some(libc::EINVAL) | Some(libc::EIO)) && !state.sendmsg_einval() {
    state.set_sendmsg_einval();
    // ...retry...
}
```

Locally, extending the fallback this way (plus disabling GSO/ECN at socket init on Android as a workaround) removes the `EIO`. On my specific device sends then hit a separate **`EPERM`** that is an Android per-UID socket-to-network binding issue (`android_setsocknetwork`), unrelated to quinn-udp. I mention it only so it isn't mistaken for the same bug.

Happy to open a PR for the `EIO` fallback if it looks right.

SkeletonGamer:
The GSO-disable path just above already handles both `EIO` and `EINVAL`, but the cmsg fallback (`sendmsg_einval`, which stops setting the ECN `IP_TOS` cmsg) only triggers on `EINVAL`. Some Android kernels return `EIO` for the ECN cmsg, so the fallback never kicks in and every `sendmsg` keeps failing.

Closes #2718.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -479,7 +479,9 @@ fn send(
 
                 // Some arguments to `sendmsg` are not supported. Switch to
                 // fallback mode and retry if we haven't already.
-                if e.raw_os_error() == Some(libc::EINVAL) && !state.sendmsg_einval() {
+                if matches!(e.raw_os_error(), Some(libc::EINVAL) | Some(libc::EIO))
+                    && !state.sendmsg_einval()
+                {
                     state.set_sendmsg_einval();
                     prepare_msg(
                         transmit,


djc (approved):
Seems okay to me...

Ralith (approved):

PR #2722, opened 2026-07-09
quinn-proto: add transport error Code::is_crypto() predicate

mjansson:
## Motivation
  
TLS alerts are mapped into transport error codes in the reserved `0x0100..0x0200` range by `Code::crypto()`. Applications that need to react to a connection being closed by a failed cryptographic handshake (e.g. a certificate or client-authentication failure) currently have to open-code that range check themselves. Internally, `Code`'s own `Debug` and `Display` impls already do the same open-coded check in two slightly different ways.
  
We hit this downstream in [Lore](https://github.com/EpicGames/Lore): our transport layer distinguishes a crypto-class connection close from an ordinary read/transport error so it can surface an authentication failure to the caller instead of an opaque error and reconnect loop.
  
## Summary  
- Add a public `Code::is_crypto()` predicate (a `const fn`).
- Use it in the `Debug` and `Display` impls, replacing the two duplicated range checks.
    
No behavioral change; this only exposes an existing internal notion as public API and de-duplicates it.

### DIFF
--- a/quinn-proto/src/transport_error.rs
+++ b/quinn-proto/src/transport_error.rs
@@ -69,6 +69,15 @@ impl Code {
     pub fn crypto(code: u8) -> Self {
         Self(0x100 | u64::from(code))
     }
+
+    /// Whether this is a crypto error code
+    ///
+    /// Crypto errors occupy the reserved `0x0100..0x0200` range, each encoding a TLS alert (see
+    /// [`Code::crypto`]). This makes it possible to distinguish a failed cryptographic handshake
+    /// (e.g. certificate or authentication failure) from other transport errors.
+    pub const fn is_crypto(self) -> bool {
+        0x100 <= self.0 && self.0 < 0x200
+    }
 }
 
 impl coding::Codec for Code {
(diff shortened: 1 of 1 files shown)


mjansson:
Linter failure is preexisting and not related to this change as far as I can tell

Ralith (approved):
Seems fine to me, though I suspect you might not want to handle all crypto errors identically downstream.

djc (approved):
Thanks!

Cool to see that you're using Quinn for Lore! As a former Mercurial maintainer, always interesting to see new VCS projects.

Quinn is maintained on a volunteer basis -- do you think Epic would be able to support maintenance? We have some options on [OpenCollective](https://opencollective.com/quinn-rs) and we also offer support contracts which include things like coordinated disclosure participation.

Happy to merge this, can you rebase on top of current main to get CI passing?

quinn-proto/src/transport_error.rs
```diff
@@ -69,6 +69,15 @@ impl Code {
     pub fn crypto(code: u8) -> Self {
         Self(0x100 | u64::from(code))
     }
+
+    /// Whether this is a crypto error code
+    ///
+    /// Crypto errors occupy the reserved `0x0100..0x0200` range, each encoding a TLS alert (see
+    /// [`Code::crypto`]). This makes it possible to distinguish a failed cryptographic handshake
+    /// (e.g. certificate or authentication failure) from other transport errors.
+    pub const fn is_crypto(self) -> bool {
```
Ralith, line 78: Random thought: I wonder if `fn as_crypto(self) -> Option<u8>` would be a little more convenient...

PR #2723, opened 2026-07-09
quinn-proto: add TransportConfig::max_rtt to bound RTT samples

mjansson:
## Motivation
  
Every RTT sample feeds straight into the smoothed RTT estimator with no upper bound. When the host is momentarily unable to process incoming packets in a timely fashion — e.g. an async runtime starved of CPU, or packets sitting unread in the socket receive buffer — an ack can be processed long after it arrived, producing a spuriously large RTT sample. That single sample poisons the smoothed estimate, which inflates the PTO, the time-based loss-detection threshold and the idle timeout, and can leave a connection unable to recover.
  
We observed exactly this under runtime starvation and now configure a 2s bound on both client and server connections. 
  
## Summary
- Add a `TransportConfig::max_rtt` option that clamps each raw RTT sample before it enters the estimator (`RttEstimator` gains a `max` bound).
- Defaults to **2 seconds**, comfortably above any legitimate internet RTT, so existing behavior is unchanged in practice.
- Covered by unit tests (`rtt_estimator_clamps_samples_above_max`, `rtt_estimator_keeps_samples_below_max`).

### DIFF
--- a/quinn-proto/src/config/transport.rs
+++ b/quinn-proto/src/config/transport.rs
@@ -36,6 +36,7 @@ pub struct TransportConfig {
     pub(crate) packet_threshold: u32,
     pub(crate) time_threshold: f32,
     pub(crate) initial_rtt: Duration,
+    pub(crate) max_rtt: Duration,
     pub(crate) initial_mtu: u16,
     pub(crate) min_mtu: u16,
     pub(crate) mtu_discovery_config: Option<MtuDiscoveryConfig>,
@@ -171,6 +172,20 @@ impl TransportConfig {
     /// The RTT used before an RTT sample is taken
     pub fn initial_rtt(&mut self, value: Duration) -> &mut Self {
         self.initial_rtt = value;
+        self
+    }
+
+    /// Upper bound applied to individual RTT samples
+    ///
+    /// Each RTT sample is clamped to this value before it is fed into the smoothed RTT estimator.
+    /// Without a bound, a single spuriously large sample -- for example caused by the async
+    /// runtime being starved, or by packets sitting unread in the socket buffer -- poisons the
+    /// estimate. That in turn inflates the PTO, time-based loss detection threshold and idle
+    /// timeout, which can leave a connection unable to recover.
+    ///
+    /// Defaults to 2 seconds, comfortably above any legitimate internet RTT.
+    pub fn max_rtt(&mut self, value: Duration) -> &mut Self {
+        self.max_rtt = value;
         self
     }
 
@@ -380,6 +395,7 @@ impl Default for TransportConfig {
             packet_threshold: 3,
             time_threshold: 9.0 / 8.0,
             initial_rtt: Duration::from_millis(333), // per spec, intentionally distinct from EXPECTED_RTT
+            max_rtt: Duration::from_secs(2),
             initial_mtu: INITIAL_MTU,
             min_mtu: INITIAL_MTU,
             mtu_discovery_config: Some(MtuDiscoveryConfig::default()),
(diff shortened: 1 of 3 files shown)


mjansson:
Linter failure is preexisting and not related to this change as far as I can tell

Ralith:
Can you elaborate on "unable to recover"? It's not obvious to me why the spurious estimate wouldn't get averaged right back down with following traffic. Even if the connection is lost immediately after such an event, surely even an inflated PTO should still resolve in a reasonable amount of time? There should be a hard upper bound at which the connection becomes lost before the delayed ACK is processed anyway.

mjansson:
I'm digging up my old notes here, we had this issue ~4 months ago but due to being in crunch to get Lore released we never go around to putting up the upstream PR.

The problem we saw is when the spurious sample is not followed by clean samples. No returning ACKs → no new RTT samples → nothing that average the estimate back down. The estimator is pinned at the poisoned value precisely during the window where recovery has to happen.

The timers that are supposed to bound and repair the connection are all keyed off that poisoned estimate, including the idle timeout. In `connection/mod.rs:2014  (reset_idle_timeout)`

`let dt = cmp::max(timeout, 3 * self.pto(space));`

The effective idle timeout is `max(negotiated_timeout, 3*PTO)`. And PTO is `smoothed + max(4*rttvar, granularity)`. So poisoning rttvar inflates PTO, which inflates the idle-timeout floor, so the hard upper bound expected to fire moves with the poisoned sample.

Loss detection is coupled the same way with `loss_delay = conservative_rtt * time_threshold` (mod.rs:1703, where `conservative() = max(smoothed, latest)`) so genuine losses aren't declared lost for far longer, and PTO probes back off geometrically on top of the inflated base (`pto_base * 2^backoff`, mod.rs:1857).

Concretely, baseline RTT `~50 ms` + one `20 s` sample:
  - `smoothed = (7*50 ms + 20 s)/8 = 20350/8 ≈ **2.54 s**`
  - `rttvar   = (3*~20 ms + |50 ms − 20 s|)/4 = (60 + 19950)/4 ≈ **5.0 s**`
  - `PTO ≈ smoothed + 4*rttvar = 2.54 + 20.0 ≈ **22.5 s**`
  - `idle floor = 3*PTO ≈ **68 s**`
  - first loss interval (uses `conservative = max(smoothed, latest) → latest` = 20 s): `loss_delay ≈ 20 * 9/8 ≈ **22.5 s**`

And since receiving any packet resets the idle timer to `now + 3*PTO`, a peer that keeps retransmitting can keep a barely-progressing connection pinned open well past the configured timeout.

Ralith:
> the hard upper bound expected to fire moves with the poisoned sample.

The hard upper bound I was referring to was the duration of delay after which the connection times out before the delayed ACK can be processed at all.

> a peer that keeps retransmitting can keep a barely-progressing connection pinned open well past the configured timeout.

I don't follow. If the peer is still communicating, why shouldn't the connection remain open?

HeikoBornholdt:
Jumping in because I ran into the same unexpected behavior.

Regarding the hard upper bound: it bounds the _sample_, not what gets derived from it.

Example:
1. Baseline 50 ms
2. server degrades, one 20 s ACK lands before it goes fully unavailable
3. smoothed RTT jumps to ~2.5 s, rttvar to ~5 s, PTO ~22.5 s, idle floor 3x that = ~68 s.
4. Client asked for 30 s timeout, but connection is kept for +68 s.

So a single delayed ACK can "poison" the PTO. This is problematic when you have multiple backend servers and clients therefore stick too long on the unhealthy one instead of re-initiating a connection that would bring them to another healthy backend server.

Being able to set a hard upper RTT bound would cap what the derived timers can grow to, so a configured idle timeout actually holds.

Ralith:
Could we alternatively solve the problem by handling packets which have sat in the receive queue for too long differently? That's attractive, as it doesn't require the user to know in advance what a reasonable upper bound might be, which in many cases may be downright impossible.

PR #2724, opened 2026-07-09
quinn-udp: retry GSO batch after disabling offload

### LINKED ISSUE #2399: udp: retry failed GSO sendmsg on EIO without GSO

kixelated:
Fixes https://github.com/quinn-rs/quinn/issues/2399

The first batch that causes a GSO failure is often the initial packet, which takes forever to retry sending at the QUIC layer because of the default RTT. This should improves connection time dramatically on some Linux devices (apparently Android).

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -474,6 +474,14 @@ fn send(
                             "`libc::sendmsg` failed with {e}; halting segmentation offload"
                         );
                         state.max_gso_segments.store(1, Ordering::Relaxed);
+
+                        // Immediately retry this batch as individual datagrams rather than waiting
+                        // for the upper layer to time out and retransmit. These sends run with GSO
+                        // now disabled, so a repeated failure falls through here instead of
+                        // recursing.
+                        if let Some(segment_size) = transmit.segment_size {
+                            return send_individual_datagrams(state, io, transmit, segment_size);
+                        }
                     }
                 }
 
@@ -499,6 +507,33 @@ fn send(
             }
         }
     }
+}
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+fn send_individual_datagrams(
+    state: &UdpSocketState,
+    io: SockRef<'_>,
+    transmit: &Transmit<'_>,
+    segment_size: usize,
+) -> io::Result<()> {
+    debug_assert!(segment_size > 0, "GSO batch with a zero segment size");
+
+    for contents in transmit.contents.chunks(segment_size) {
+        let io = SockRef::from(&*io);
+        send(
+            state,
+            io,
+            &Transmit {
+                destination: transmit.destination,
+                ecn: transmit.ecn,
+                contents,
+                segment_size: None,
+                src_ip: transmit.src_ip,
+            },
+        )?;
+    }
+
+    Ok(())
 }
 
 #[cfg(any(target_os = "openbsd", target_os = "netbsd", apple_slow))]


quinn-proto/src/connection/packet_crypto.rs
```diff
@@ -88,16 +88,10 @@ pub(super) fn decrypt_packet_body(
         &zero_rtt_crypto.unwrap().packet
     } else if packet_key_phase == conn_key_phase || space != SpaceId::Data {
         &spaces[space].crypto.as_ref().unwrap().packet.remote
-    } else if let Some(prev) = prev_crypto.and_then(|crypto| {
+    } else if let Some(prev) = prev_crypto.filter(|&crypto| {
```
Ralith: Very welcome cleanup, but split it into a separate commit?

quinn-udp/src/unix.rs
```diff
@@ -467,6 +467,18 @@ fn send(
...
+                            // Prevent new transmits from being scheduled using GSO, then
+                            // immediately retry this batch as individual datagrams.
+                            crate::log::info!(
+                                "`libc::sendmsg` failed with {e}; halting segmentation offload"
+                            );
+                            state.max_gso_segments.store(1, Ordering::Relaxed);
+                            return send_individual_datagrams(state, io, transmit, segment_size);
+                        }
+                    }
+
```
Ralith: This seems to substantially duplicate the logic just below. Why do we need both?

quinn-udp/src/unix.rs
```diff
@@ -467,6 +467,18 @@ fn send(
                 // when we try to actually send datagrams using it.
                 #[cfg(any(target_os = "linux", target_os = "android"))]
                 if let Some(libc::EIO) | Some(libc::EINVAL) = e.raw_os_error() {
+                    if let Some(segment_size) = transmit.effective_segment_size() {
+                        if segment_size > 0 && state.max_gso_segments() > 1 {
```
Ralith: Why do we need to guard against segment_size = 0? That doesn't seem like a reasonable value for it to have, especially as we use `Option` rather than sentinel values.

quinn-udp/tests/tests.rs
```diff
@@ -1,10 +1,11 @@
 #[cfg(not(any(target_os = "openbsd", target_os = "netbsd", solarish)))]
 use std::net::{SocketAddr, SocketAddrV6};
+#[cfg(any(target_os = "linux", target_os = "android"))]
+use std::time::Duration;
```
Ralith, line 4: Unrelated change?

Ralith (changes requested):

mxinden:
Very much in favor. Thanks for working on it.

We (Firefox) ran into the same issue. We currently re-send within Firefox. Instead re-sending in quinn-udp would be much cleanre.

https://bugzilla.mozilla.org/show_bug.cgi?id=2049334

quinn-proto/src/connection/packet_crypto.rs
kixelated: This is an automatic clippy fix, sure.

quinn-udp/src/unix.rs
kixelated: we don't

quinn-proto/src/connection/packet_crypto.rs
kixelated: which is already on main anyway

quinn-udp/src/unix.rs
```diff
@@ -474,6 +474,12 @@ fn send(
                             "`libc::sendmsg` failed with {e}; halting segmentation offload"
                         );
                         state.max_gso_segments.store(1, Ordering::Relaxed);
+
+                        // Immediately retry this batch as individual datagrams rather than waiting
+                        // for the QUIC layer to time out and retransmit. These sends run with GSO
```
thomaseizinger: Nit: Refer to "upper" layer here.

quinn-udp/src/unix.rs
```diff
@@ -501,16 +507,43 @@ fn send(
...
+    state: &UdpSocketState,
+    io: &SockRef<'_>,
+    transmit: &Transmit<'_>,
+) -> io::Result<()> {
+    // Split on the segment size if this was a GSO batch, otherwise send the payload as-is.
+    let segment_size = transmit
+        .segment_size
+        .unwrap_or(transmit.contents.len())
+        .max(1);
+    for contents in transmit.contents.chunks(segment_size) {
```
thomaseizinger, line 521: Style nit: Would prefer a newline here.

quinn-udp/src/unix.rs
```diff
@@ -501,16 +507,43 @@ fn send(
...
+            &Transmit {
+                destination: transmit.destination,
+                ecn: transmit.ecn,
+                contents,
+                segment_size: None,
+                src_ip: transmit.src_ip,
+            },
+        )?;
+    }
+    Ok(())
```
thomaseizinger, line 536: And here

quinn-udp/src/unix.rs
```diff
@@ -420,7 +420,7 @@ impl UdpSocketState {
 fn send(
     #[allow(unused_variables)] // only used on Linux
     state: &UdpSocketState,
-    io: SockRef<'_>,
+    io: &SockRef<'_>,
```
thomaseizinger: We can avoid this by re-borrowing the ref instead:

```rust
let io = SockRef::from(&*io);
```

thomaseizinger (changes requested):
Concept ACK, a few nits

quinn-udp/src/unix.rs
kixelated: Yeah I opted for `&` because it's feels less magic, but IDK.

thomaseizinger (approved):

quinn-udp/src/unix.rs
```diff
@@ -556,6 +507,36 @@ fn send(
...
+fn send_individual_datagrams(
+    state: &UdpSocketState,
+    io: SockRef<'_>,
+    transmit: &Transmit<'_>,
+) -> io::Result<()> {
+    // Split on the segment size if this was a GSO batch, otherwise send the payload as-is.
+    let segment_size = transmit
+        .segment_size
+        .unwrap_or(transmit.contents.len())
+        .max(1);
```
thomaseizinger: What is that for?

If `contents.len` is 0, we can just return `Ok(())`, right? And if `segment_size` is 0, we should probably fail?

thomaseizinger (changes requested):

Ralith (approved):
Thanks!

thomaseizinger:
Note that this also gets fixed in #2748 which may be preferable to merge because it also fixes another issue.

PR #2726, opened 2026-07-10
Contribution guidelines

djc:
This is based on rustls' [guidelines](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md), with an AI policy adapted from [ripgrep's](https://github.com/BurntSushi/ripgrep/blob/master/AI_POLICY.md).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,396 @@
+# Contributing
+
+Thanks for considering helping this project. There are many ways you can help: using the library
+and reporting bugs, reporting usability issues, making additions and improvements to the library,
+documentation and finding security bugs.
+
+## Reporting bugs
+
+Please file a GitHub issue. Include as much information as possible. Suspected protocol bugs are
+easier debugged with a pcap or reproduction steps.
+
+Feel free to file GitHub issues to get help, or ask a question.
+
+If you believe you've found a security bug please
+[open a draft security advisory](https://github.com/quinn-rs/quinn/security/advisories/new)
+in GitHub, and not as a regular repository issue.
+
+## AI policy
+
+Using AI (LLMs) as tools for coding is welcome. A high bar is held for all contributions to this
+project. Moreover, the project maintainers remain responsible for any code that is published as
+part of a release. Contributors are expected to be responsible for any code they publish.
+
+AI should not be used to generate comments when communicating with maintainers. Comments are
+expected to be written by humans. Comments that are believed to be written by AI may be hidden
+without notice.
+
+If you are opening an issue, you should be able to describe the problem in your own words.
+
+If you are opening a pull request, you are expected to be able to explain the proposed changes in
+your own words. This includes the pull request body and responses to questions. Make sure you have
+reviewed the PR yourself before submitting it for review to the maintainers. Do not copy responses
+from the AI when replying to questions from maintainers.
+
+If you wish to include context from an interaction with AI in your comments, it must be in a
+quote block (using `>`) and disclosed as such. It must be accompanied by human commentary
+explaining the relevance and implications of the context. Do not share long snippets.
+
+AI is useful when communicating as a non-native English speaker. If you are using AI to edit your
+comments for this purpose, please take the time to ensure it reflects your own voice and ideas.
+When using AI for translation, we recommend writing in your native language and including the AI
+translation in a quote block.
+
+## Code changes
+
+Some ideas and guidelines for contributions:
+
+- For large features, file an issue prior to starting work.
+  This means everyone can see what is in progress prior to a PR.
+- Feel free to submit a PR even if the work is not totally finished, for early feedback.
+- Prefer not to reference GitHub issue or PR numbers in commit messages.
+- Try to keep code formatting commits separate from functional commits.
+
+## Commit history
+
+We prefer to keep the commit history clean and easy to follow. As such, we prefer small commits
+that do one thing. In particular:
+
+* Avoid mixing refactoring and functional changes in the same commit if possible
+* Make mechanical changes (like renaming or moving code around) in a separate commit
+* Isolate updates to `Cargo.lock` in their own commits
+
+Our default workflow is to rebase clean commit history from a PR to `main`. As a result,
+we ask that review feedback is addressed in the originating commit rather than in new commits,
+by amending existing commits or squashing later commits back into the original commit(s).
+
+## Testing
+
+PRs which cause test failures or a significant coverage decrease are unlikely to be accepted.
+
+## Style guide
+
+### Ordering
+
+#### Top-down ordering within modules
+
+Within a module, we prefer to order items top-down. This means that items within
+a module will depend on items defined below them, but not (usually) above them.
+The idea here is that the public API, with more internal dependencies, will be
+read (and changed) more often, and putting it closer to the top of the module
+makes it more accessible.
+
+Usually `const` values will thus go on the bottom of the module (least complex,
+usually no dependencies of their own), although in larger modules it can make
+sense to place a `const` directly below the user (especially if there is a
+single user, or just a few co-located users).
+
+The `#[cfg(test)] mod tests {}` module goes on the very bottom, if present.
+Other module definitions (like `mod foo { .. }`) can be ordered among other
+items as it makes sense in the context of the items imported from them.
+Module declarations (like `mod foo;`) should be ordered before other items
+but after imports. Imports from local modules (both declared and defined)
+should be kept close to the module declaration/definition.
+
+Files that have substantial amounts of code inside inline modules should
+probably avoid also having much code outside of these modules.
+
+#### Ordering for a given type
+
+For a given type, we prefer to order items as follows:
+
+1. The type definition (`struct` or `enum`)
+2. The inherent `impl` block (that is, not a trait implementation)
+3. `impl` blocks for traits, from most specific to least specific.
+   The least specific would be something like a `Debug` or `Clone` impl.
+
+#### Ordering associated functions within an inherent `impl` block
+
+Here's a guide to how we like to order associated functions:
+
+0. Associated functions (that is, `fn foo() {}` instead of `fn foo(&self) {}`)
+1. Constructors, starting with the constructor that takes the least arguments
+2. Public API that takes a `&mut self`
+3. Public API that takes a `&self`
+4. Private API that takes a `&mut self`
+5. Private API that takes a `&self`
+6. `const` values
+
+Note that we usually also practice top-down ordering here; where these are in
+conflict, make a choice that you think makes sense. For getters and setters, the
+order should typically mirror the order of the fields in the type definition.
+
+#### Attribute ordering
+
+Order attributes so that documentation appears first, and the attributes with the
+most effect on the meaning and function of the type appear last.  For example:
+
+```rust
+/// Doc comment always first
+#[cfg(feature-gates)]
+#[allow(lint-configuration)]
+#[non_exhaustive]
+#[derive(Clone, Debug)]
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,485 @@
...
+violation of this will be closed, perhaps without notice.
+
+If you wish to include context from an interaction with AI in your comments, it must be in a
+quote block (using `>`) and disclosed as such. It must be accompanied by human commentary
+explaining the relevance and implications of the context. Do not share long snippets.
+
+AI is useful when communicating as a non-native English speaker. If you are using AI to edit your
+comments for this purpose, please take the time to ensure it reflects your own voice and ideas.
+When using AI for translation, we recommend writing in your native language and including the AI
+translation in a quote block.
```
flub, line 42: This is very nice, might be worth putting higher up in this section.

CONTRIBUTING.md
```diff
@@ -0,0 +1,485 @@
...
+
+## Code changes
+
+Some ideas and guidelines for contributions:
+
+- For large features, file an issue prior to starting work.
+  This means everyone can see what is in progress prior to a PR.
+- Feel free to submit a PR even if the work is not totally finished, for early feedback.
+- Prefer not to reference GitHub issue or PR numbers in commit messages.
```
flub, line 51: Why is this? I've noticed consuming Quinn commits downstream that it is often a bit hard to find out the reasoning and motivation behind commits. Information that would be more readily available if issues and PRs were mentioned. Usually commit messages are not really part of the review process so they tend to be terse and rely on the PR context.

I'm not asking to change how you work, only curious how this practice came to be. Feel free to ignore :)

flub (approved):

CONTRIBUTING.md
djc: The problem is that GitHub generates a lot of spam from these, so for example each rebase (as well as downstream merges) would generate a link in the linked issue, or, worse, a notification for the mentioned user.

CONTRIBUTING.md
djc: Hmm, not sure -- I think at the bottom is probably okay for this since the other points seem more relevant to most consumers.

CONTRIBUTING.md
```diff
@@ -0,0 +1,485 @@
...
+
+- Linker friendliness
+- Small mandatory API
+- Safe and sensible defaults
+- Separation of mechanism and policy
+
+### Linker friendliness
+
+When a program incorporates rustls, we should try to ensure that parts
+that are not used can be discarded by the linker.
```
mxinden: Either state that this is a copy of the CONTRIBUTING of rustls, or replace with quinn here?

CONTRIBUTING.md
```diff
@@ -0,0 +1,485 @@
...
+without notice.
+
+If you are opening an issue, you should be able to describe the problem in your own words.
+
+If you are opening a pull request, you are expected to be able to explain the proposed changes in
+your own words. This includes the pull request body and responses to questions. Make sure you have
+reviewed the PR yourself before submitting it for review to the maintainers. Do not copy responses
+from the AI when replying to questions from maintainers.
+
+This project requires a human in the loop who understands the work produced by AI. Autonomous
```
djc: Someone I work with on other projects had questions about this part, in particular the second sentence, which seems fairly broad. Given that the constraints are pretty clearly enumerated in other paragraphs, perhaps this paragraph/sentence should be removed?

CONTRIBUTING.md
djc: Dropped this section and made sure there are no references to rustls anymore.

CONTRIBUTING.md
Ralith: That sounds good to me. As written, this paragraph might cause confusion about people using agentic workflows to draft code locally, which we don't intrinsically object to.

CONTRIBUTING.md
```diff
@@ -0,0 +1,414 @@
...
+- Try to keep code formatting commits separate from functional commits.
+
+## Commit history
+
+We prefer to keep the commit history clean and easy to follow. As such, we prefer small commits
+that do one thing. In particular:
+
+* Avoid mixing refactoring and functional changes in the same commit if possible
+* Make mechanical changes (like renaming or moving code around) in a separate commit
+* Isolate updates to `Cargo.lock` in their own commits
```
Ralith, line 61: Should we also mention amending/squashing/rebasing PR commits in response to feedback?

CONTRIBUTING.md
```diff
@@ -0,0 +1,414 @@
...
+
+Within a module, we prefer to order items top-down. This means that items within
+a module will depend on items defined below them, but not (usually) above them.
+The idea here is that the public API, with more internal dependencies, will be
+read (and changed) more often, and putting it closer to the top of the module
+makes it more accessible.
+
+This can be surprising to many engineers who are used to the bottom-up ordering
+used in languages like Python, where items can have a run-time dependency on
+other items defined in the same module.
```
Ralith: Is this paragraph carrying any weight? It reads like it's just throwing shade at Python.

CONTRIBUTING.md
```diff
@@ -0,0 +1,414 @@
...
+use alloc::vec::Vec;
+```
+
+We prefer to reference types and traits by an imported symbol name instead of
+using qualified references. Qualification paths generally add noise and are
+unnecessary. The one exception to this is when the symbol name is overly
+generic, or easily confused between different crates. In this case we prefer to
+import the symbol name under an alias, or if the parent module name is short,
+using a one-level qualified path. E.g. for a crate with a local `Error` type,
+prefer to `import std::error::Error as StdError`.
```
Ralith: Should this say `use`?

CONTRIBUTING.md
```diff
@@ -0,0 +1,414 @@
...
+```rust
+// Incorrect
+use alloc::{format, vec::Vec};
+
+// Correct
+use alloc::format;
+use alloc::vec::Vec;
+```
+
+We prefer to reference types and traits by an imported symbol name instead of
```
Ralith: editorial:
```suggestion
We prefer to reference items by an imported symbol name instead of
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,414 @@
...
+import the symbol name under an alias, or if the parent module name is short,
+using a one-level qualified path. E.g. for a crate with a local `Error` type,
+prefer to `import std::error::Error as StdError`.
+
+### Exports
+
+We prefer to export types under a single name, avoiding re-exporting types from
+the top-level `lib.rs`. The exception to this are "paved path" exports that we
+expect every user will need. The canonical example of such types are
+`client::ClientConfig` and `server::ServerConfig`. In general this sort of type
```
Ralith: This example is confusing to a quinn user.

CONTRIBUTING.md
```diff
@@ -0,0 +1,414 @@
...
+unnecessary. The one exception to this is when the symbol name is overly
+generic, or easily confused between different crates. In this case we prefer to
+import the symbol name under an alias, or if the parent module name is short,
+using a one-level qualified path. E.g. for a crate with a local `Error` type,
+prefer to `import std::error::Error as StdError`.
+
+### Exports
+
+We prefer to export types under a single name, avoiding re-exporting types from
+the top-level `lib.rs`. The exception to this are "paved path" exports that we
```
Ralith: I don't think this is at all true for Quinn, and I'm pretty happy with that the way it is. We go to some effort exposing a flatter namespace than our internal code organization would naturally yield, regardless of how likely a typical downstream user is to need to name something. We do have some few public modules, but those are mostly limited to cases where all the types are very closely associated and might need a common prefix otherwise. The result is, IMO, substantially more ergonomic for users.

CONTRIBUTING.md
```diff
@@ -0,0 +1,414 @@
...
+expect every user will need. The canonical example of such types are
+`client::ClientConfig` and `server::ServerConfig`. In general this sort of type
+is rare and most new types should be exported only from the module in which they
+are defined.
+
+### Comments
+
+Crate items should have descriptive doccomments on them. These are required for
```
Ralith: editorial: I'm skeptical of "doccomments" as one word.

CONTRIBUTING.md
```diff
@@ -0,0 +1,414 @@
...
+### Exports
+
+We prefer to export types under a single name, avoiding re-exporting types from
+the top-level `lib.rs`. The exception to this are "paved path" exports that we
+expect every user will need. The canonical example of such types are
+`client::ClientConfig` and `server::ServerConfig`. In general this sort of type
+is rare and most new types should be exported only from the module in which they
+are defined.
+
+### Comments
```
Ralith, line 331: Should we talk about non-doc comments?

CONTRIBUTING.md
```diff
@@ -0,0 +1,414 @@
...
+#### Avoid type aliases
+
+We prefer to avoid type aliases as they obfuscate the underlying type and
+don't provide additional type safety. Using the
+[newtype idiom](https://doc.rust-lang.org/rust-by-example/generics/new_types.html)
+is one alternative when an abstraction boundary is worth the added complexity.
+
+#### Type exhaustiveness
+
+Public enums should be marked as _either_ `#[non_exhaustive]` or `#[allow(clippy::exhaustive_enums)]`.
```
Ralith: Do we need to turn this lint on in CI?

CONTRIBUTING.md
```diff
@@ -0,0 +1,414 @@
...
+- Safe and sensible defaults
+- Separation of mechanism and policy
+
+### Safe and sensible defaults
+
+We should be confident to make decisions on behalf of our users and
+express that in the [library's defaults][defaults].  It is good to
+encode our experience of TLS in this way, and we should not require that
+users are experts in TLS to end up with a result that is working,
+secure, stable, and performant.
```
Ralith, line 380: Slightly odd elaboration for Quinn: while TLS is certainly a large part of what we must be wary of overwhelming the user with, there's much more beyond it. Maybe prefix "For example, it is good to ..."?

Ralith (changes requested):

CONTRIBUTING.md
djc: I've dropped the section on linker friendliness, which I don't think is important for us.

CONTRIBUTING.md
djc: I've dropped this section for now. Something we could look into in the future.

CONTRIBUTING.md
djc: I don't think we really need to? They're not visible to the outside, so not as important.

CONTRIBUTING.md
djc: Expanded it to documentation comments.

CONTRIBUTING.md
djc: Fair, cut this section on exports.

CONTRIBUTING.md
djc: Cut this section.

CONTRIBUTING.md
djc: Yup, nice catch!

CONTRIBUTING.md
djc: Fair enough, dropped it.

CONTRIBUTING.md
djc: Added a sentence about this.

CONTRIBUTING.md
djc: Dropped the paragraph.

thomaseizinger (approved):
Very good write-up.

Ralith (approved):

PR #2727, opened 2026-07-10
fix(quinn-udp): avoid permanent `EWOULDBLOCK` from macOS `sendmsg` bug

larseggert:
Non-blocking `sendmsg`/`sendmsg_x` on Apple platforms can permanently return with `EWOULDBLOCK` when a datagram plus its `cmsg` lands just under `SO_SNDBUF`. Enforce a safe `SO_SNDBUF` floor on socket creation and resize, with a per-send `EMSGSIZE` safety net.

See https://feedbackassistant.apple.com/feedback/23671230

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -197,6 +197,15 @@ impl UdpSocketState {
             )?;
         }
 
+        // Enlarge SO_SNDBUF to a safe minimum.
+        #[cfg(apple)]
+        if io
+            .send_buffer_size()
+            .is_ok_and(|cur| cur < Self::MIN_SAFE_SNDBUF)
+        {
+            let _ = io.set_send_buffer_size(Self::MIN_SAFE_SNDBUF);
+        }
+
         let now = Instant::now();
         Ok(Self {
             last_send_error: Mutex::new(now.checked_sub(2 * IO_ERROR_LOG_INTERVAL).unwrap_or(now)),
@@ -331,8 +340,20 @@ impl UdpSocketState {
     }
 
     /// Resize the send buffer of `socket` to `bytes`
+    ///
+    /// On Apple platforms, `bytes` is silently raised to a safe minimum if smaller.
     #[inline]
     pub fn set_send_buffer_size(&self, socket: UdpSockRef<'_>, bytes: usize) -> io::Result<()> {
+        #[cfg(apple)]
+        let bytes = if bytes < Self::MIN_SAFE_SNDBUF {
+            crate::log::debug!(
+                "raising requested SO_SNDBUF from {bytes} to {}",
+                Self::MIN_SAFE_SNDBUF
+            );
+            Self::MIN_SAFE_SNDBUF
+        } else {
+            bytes
+        };
         socket.0.set_send_buffer_size(bytes)
     }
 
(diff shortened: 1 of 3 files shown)


djc:
(Needs a rebase both to get CI to pass and to fix some conflicts.)

thomaseizinger (approved):
Concept ACK

Can we split those two orthogonal things into separate commits? The send buffer size minimum and the runtime check returning EMSGSIZE

quinn-udp/src/unix.rs
```diff
@@ -50,6 +50,12 @@ pub struct UdpSocketState {
     apple_fast_path: AtomicBool,
 }
 
+/// Smallest `SO_SNDBUF` that mitigates https://feedbackassistant.apple.com/feedback/23671230:
+/// On macOS, a non-blocking SOCK_DGRAM `sendmsg`/`sendmsg_x` call with ancillary data returns
+/// `EWOULDBLOCK` when the payload length is at or just under `SO_SNDBUF`.
+#[cfg(apple)]
+const MIN_SAFE_SNDBUF: usize = 65535 + cmsg::LEN;
```
djc: Nit: let's move this to the bottom of the `UdpSocketState` `impl` block.

quinn-udp/src/unix.rs
```diff
@@ -197,6 +203,12 @@ impl UdpSocketState {
             )?;
         }
 
+        #[cfg(apple)]
+        // Enlarge SO_SNDBUF to a safe minimum.
```
djc, line 204: Nit: comment before attributes, please.

quinn-udp/tests/tests.rs
```diff
@@ -484,6 +488,120 @@ fn apple_fast_datapath() {
     assert_eq!(total_received, segments, "should receive all segments");
 }
 
+#[cfg(apple)]
```
djc: Nits: callers before callees.

quinn-udp/src/unix.rs
```diff
@@ -205,9 +209,12 @@ impl UdpSocketState {
 
         #[cfg(apple)]
         // Enlarge SO_SNDBUF to a safe minimum.
-        if io.send_buffer_size().is_ok_and(|cur| cur < MIN_SAFE_SNDBUF) {
-            let _ = io.set_send_buffer_size(MIN_SAFE_SNDBUF);
-        }
+        let send_buffer_size = {
```
djc: Nit: the nested scope doesn't seem very useful here?

quinn-udp/src/unix.rs
```diff
@@ -547,10 +562,33 @@ pub(crate) fn send_single(
...
 
+#[cfg(apple)]
+fn query_send_buffer_size(io: &SockRef<'_>) -> usize {
+    io.send_buffer_size().unwrap_or(usize::MAX)
+}
+
+/// Safety net returning `EMSGSIZE` for payload sizes in the bug's region.
+#[cfg(apple)]
+pub(crate) fn check_send_buffer_limit(
+    state: &UdpSocketState,
```
djc: Why aren't these `UdpSocketState` methods?

quinn-udp/src/unix.rs
```diff
@@ -354,7 +363,13 @@ impl UdpSocketState {
         } else {
             bytes
         };
-        socket.0.set_send_buffer_size(bytes)
+        socket.0.set_send_buffer_size(bytes)?;
```
djc, line 363: Should the `send_buffer_size.store()` call happen inside `set_send_buffer_size()`?

djc (approved):

djc:
Thanks!

Issue #2487, opened 2025-12-27
Expose ECN information in Controller::on_ack


B1gG:
## Summary
Add ECN (Explicit Congestion Notification) visibility to the `Controller::on_ack` callback to enable custom congestion control algorithms to react to per-packet ECN CE marks.

## Motivation
Currently, Quinn's `Controller` trait receives ECN CE information only through `on_congestion_event`, which is triggered after ECN-based congestion detection. However, custom congestion control algorithms may need finer-grained access to ECN information on a per-ACK basis to:

1. **Implement ECN-aware algorithms**: Custom controllers could implement sophisticated ECN-based congestion control (e.g., DCTCP-like algorithms, L4S, or experimental protocols)
2. **Early congestion detection**: React to ECN CE marks before they accumulate into a congestion event
3. **Telemetry and monitoring**: Track ECN mark rates for debugging and performance analysis
4. **Research and experimentation**: Enable development of novel congestion control mechanisms that leverage ECN

## Proposed API Change

Modify the `Controller::on_ack` method signature to include ECN information:

```rust
fn on_ack(
    &mut self,
    now: Instant,
    sent: Instant,
    bytes: u64,
    app_limited: bool,
    rtt: &RttEstimator,
    ecn: Option<EcnCodepoint>,
) {
    // Default implementation remains unchanged for backward compatibility
}
```

### Parameters
- `ecn`: An `Option<EcnCodepoint>` indicating the ECN marking on the acknowledged packet
  - `Some(EcnCodepoint::Ce)`: Packet was marked with Congestion Experienced
  - `Some(EcnCodepoint::Ect0)` or `Some(EcnCodepoint::Ect1)`: ECN-capable but not congested
  - `None`: No ECN information available (e.g., peer doesn't support ECN, or network stripped ECN bits)

## Example Use Case

```rust
impl Controller for CustomController {
    fn on_ack(
        &mut self,
        now: Instant,
        sent: Instant,
        bytes: u64,
        app_limited: bool,
        rtt: &RttEstimator,
        ecn: Option<EcnCodepoint>,
    ) {
        // Track ECN CE marks for telemetry
        if let Some(EcnCodepoint::Ce) = ecn {
            self.ce_marks_received += 1;
            // React immediately to congestion signal
            self.reduce_cwnd_early();
        }
        
        // Continue with normal ACK processing
        self.handle_ack(bytes, rtt);
    }
}
```

djc:
[...]

PR #2729, opened 2026-07-10
Implement L4S support and the Prague CCA

### LINKED ISSUE #1498: Consider Pargue (L4S-compliant) congestion control support
[draft-briscoe-iccrg-prague-congestion-control-01](https://datatracker.ietf.org/doc/html/draft-briscoe-iccrg-prague-congestion-control-01) proposes a new congestion controller that performs well on paths bottlenecked on ECN-capable links. I believe these are rare in most networks, but it might still be interesting to support. Another avenue to explore might be extending Cubic or BBR to satisfy L4S requirements.

zoltanszatmary:
## Summary

This change set introduces support for the **Low Latency, Low Loss, and Scalable Throughput (L4S)** architecture (RFC 9330) by exposing ECN configuration and implementing the Prague scalable congestion control algorithm.

## Changes

The changes made are three-fold:

1. **ECN Configuration**: Adds configurable ECN modes (`Classic`, `L4S`, or `Disabled`) to `TransportConfig` and enables `ECT(1)` markings.
2. **TCP Prague**: Implements the Prague congestion control algorithm with EWMA-based $\alpha$ (CE-fraction) estimation, virtual RTT scaling to reduce RTT dependence, and loss/ECN event double-reduction cooldowns.
3. **`quinn-perf` Support**: Updates `quinn-perf` to support Prague and configured ECN, tracks stream RTT/RTTvar, and optimizes client shutdown logic.

## Testing & Verification

To verify the implementation, I created a Mininet-based testbed (available [here](https://github.com/Zotyamester/l4s-bench), though unfortunately, not everything is documented as of yet), the topology of which is seen in the figure below.

env

I measured six metrics in time (four endpoint-based, and one network-based).
Namely, one-way delay, RTT, CWND, network queue length, Prague's alpha value,
and the estimated throughput. The endpoint-based metrics were gathered and/or
derived from qlog entries (e.g., for calculating the one-way delay estimates,
the script paired up corresponding 1RTT packets -- assuming most packets were
not dropped). On the other hand, the queue length was monitored by a short
script running on the host taking the role of the router (configured with a
DualPI2 queue on its outbound interface facing the quinn-perf server host).

During the measurement scenario here, the quinn-perf client (on host h1) ran
for 60 seconds, continuously sending data to the quinn-perf server (on host h2)
through the network, with each round-trip experiencing a total delay of 24 ms,
and the network bandwidth being limited to 32 Mbps (with HTB).

The results of this run are depicted in the figure below.

plot

## Referenced Issues

Fixes #1498 #2487.

### DIFF
--- a/perf/src/client.rs
+++ b/perf/src/client.rs
@@ -165,6 +165,7 @@ pub async fn run(opt: Opt) -> Result<()> {
     let mut stats = Stats::default();
 
     let stats_fut = async {
+        let total_duration = Duration::from_secs(opt.duration);
         let interval_duration = Duration::from_secs(opt.interval);
 
         #[cfg(feature = "json-output")]
@@ -172,9 +173,16 @@ pub async fn run(opt: Opt) -> Result<()> {
         #[cfg(not(feature = "json-output"))]
         let allow_table_output = true;
 
+        // Periodically report statistics.
         loop {
+            let elapsed = stats.elapsed_since_start();
+            let remaining = total_duration.saturating_sub(elapsed);
+            if remaining.is_zero() {
+                break;
+            }
+
             let start = Instant::now();
-            tokio::time::sleep(interval_duration).await;
+            tokio::time::sleep(interval_duration.min(remaining)).await;
             {
                 stats.on_interval(start, &stream_stats);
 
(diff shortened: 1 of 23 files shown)


djc:
Thanks for working on this! My review queue is a bit full so it might take me some time to give this a serious look. Also, I will be reviewing

- https://github.com/quinn-rs/quinn/pull/2481

before getting to this, which also makes some changes to the congestion controller interface. Would be good if you can have a look at whether any changes you've made to the trait are incompatible (if so, maybe give feedback on that PR).

zoltanszatmary:
Well, aside from that few parts of the code base where congestion control was touched by both branches (e.g., quinn-perf's `--congestion` selector, which could cause a minor merge conflicts), I don't see any major incompatibilities. Although, I certainly see that the `Controller` interface is getting a bit out of hand.

I mean I extended it with 7 more methods in addition to the 3 new methods added in his patch.

I've been thinking of better ways to achieve the same thing with less addition to the interface, but I couldn't come up with any so far. Or maybe one...?

So currently, Prague relies on an inner `Controller` composed inside of it for increasing the window until it receives increased ECN counts (e.g., regular old AIMD with `NewReno`). This however, could be changed by integrating the logic from `NewReno` directly into `Prague` at the cost of some code duplication.

`Pacer::delay` has also been modified by both of us, so there's that, but I find it one of the less important changes from this PR.

Issue #2720, opened 2026-07-09
BTreeMap for ACK takes too much CPU time for memery allocate


EileenGalvin:
As you can see, 75% of memery allocate caused by BtreeMap-operate witch gives a lot of pressure on edge divces like router
someone mention this at #1107

there are two Node for VacantEntry<K,V,A>::insert_entry maybe one for SentPacket watting for ACK, another for ACK respons

```
▼ PP 1/1 (4 children) {
    Total:     27,831,006,883 bytes (100%, 255,915,532.68/s) in 210,452 blocks (100%, 1,935.18/s), avg size 132,243.96 bytes, avg lifetime 443,670.8 µs (0.41% of program duration)
    At t-gmax: 1,685,097,549 bytes (100%) in 2,306 blocks (100%), avg size 730,744.82 bytes
    At t-end:  378,974,633 bytes (100%) in 587 blocks (100%), avg size 645,612.66 bytes
    Allocated at {
      #0: [root]
    }
  }
  ├─▼ PP 1.1/4 (5 children) {   <-----BtreeMap-operate all under this Node 
  │     Total:     122,324,604 bytes (0.44%, 1,124,816.16/s) in 158,075 blocks **(75.11%, 1,453.55/s)**, avg size 773.84 bytes, avg lifetime 126,883.43 µs (0.12% of program duration)
  │     At t-gmax: 100,800 bytes (0.01%) in 48 blocks (2.08%), avg size 2,100 bytes
  │     At t-end:  100,008 bytes (0.03%) in 59 blocks (10.05%), avg size 1,695.05 bytes
  │     Allocated at {
  │       #1: 0x563dde224e89: <alloc::alloc::Global as core::alloc::Allocator>::allocate (alloc/src/alloc.rs:449:14)
  │     }
  │   }
  │   ├─▼ PP 1.1.1/5 (2 children) {
  │   │     Total:     91,395,240 bytes (0.33%, 840,410.18/s) in 78,789 blocks (37.44%, 724.49/s), avg size 1,160 bytes, avg lifetime 161,407.41 µs (0.15% of program duration)
  │   │     At t-gmax: 5,800 bytes (0%) in 5 blocks (0.22%), avg size 1,160 bytes
  │   │     At t-end:  2,320 bytes (0%) in 2 blocks (0.34%), avg size 1,160 bytes
  │   │     Allocated at {
[...]

PR #2735, opened 2026-07-15
Reduce receive-path allocations by reusing one buffer per batch

### LINKED ISSUE #2734: Per-datagram allocation and copy in the receive path
## Summary

On the receive path in `quinn/src/endpoint.rs`, every datagram is copied out of the reusable `recv_buf` into a freshly allocated `BytesMut`, and the following split allocates a shared control block. Both happen once per datagram. On a 200 MiB loopback transfer this is ~91% of remaining heap allocations and ~95% of allocated bytes, including a full 214 MiB extra copy of the received data.

This is the receive-side counterpart to #2730 (which proposes to fix the send-side churn in #2720). With `sent_packets` no longer allocating per packet, `RecvState::poll_socket` is now the dominant allocation site. It's independent of #2730; the receive code is unchanged by it.

## The hot loop

```rust
for (meta, buf) in metas.iter().zip(iovs.iter()).take(msgs) {
    let mut data: BytesMut = buf[0..meta.len].into();          // malloc + full copy per datagram
    while !data.is_empty() {
        let buf = data.split_to(meta.stride.min(data.len()));  // promote_to_shared: Box<Shared> per datagram
        // ... endpoint.handle(.., buf, ..)
    }
}
```

`recv_buf` is reused across polls, so the bytes are copied into an owned `BytesMut` before the next poll overwrites the buffer. `Endpoint::handle` takes `BytesMut` by value and retains split-off regions, so it does need *owned* bytes, but not a fresh per-datagram allocation.

## Profile (200 MiB loopback, dhat, `bench` bulk)

Total: 40,157 blocks / 225.2 MB. Top sites, all in `RecvState::poll_socket`:

| Site | Blocks | Bytes | What |
|---|--:|--:|---|
| `endpoint.rs:866` | 15,154 | 214.5 MB | `BytesMut::from(&[u8])` — malloc + copy per datagram |
| `endpoint.rs:868` | 15,154 | 0.6 MB | `split_to` → `promote_to_shared` per datagram |

(Both sites also appear a second time at ~3,159 blocks — `bench` runs client and server in one process, so the ACK receiver hits the same code. Combined, these four points are 36,626 of 40,157 blocks and 215 of 225 MB.)

## Proposed fix

Receive into a reusable `BytesMut` arena and split owned slices out of it instead of copying:

- Replace `recv_buf: Box<[u8]>` with a `BytesMut` carrying spare capacity; point the `poll_recv` iovecs at its uninitialized tail.
- After the recv, `split_to` each datagram's region as an owned `BytesMut` — shares the arena's allocation by refcount, no copy, no per-datagram malloc.
- When spare capacity drops below one batch, reserve/allocate a fresh arena; outstanding slices keep the old one alive until dropped.

Contained in `RecvState`; the `Endpoint::handle` signature is unchanged.

## Expected impact

- Removes ~91% of remaining per-datagram allocations (~36k blocks → a handful of arena refills).
- Eliminates ~214 MiB of memcpy per 200 MiB transferred — one full pass over all received bytes, so unlike the send-side churn this should help throughput and not just allocator time, most at high packet rates and on constrained devices.

I haven't measured a throughput delta yet; happy to run the same loopback A/B used for #2730 if you think work is warranted (which I'd gladly pick up).

ojhermann:
RecvState::poll_socket allocated a fresh BytesMut for every received datagram, copying it out of recv_buf, and the following split_to promoted it to shared. This keeps one reusable BytesMut, copies each batch into it, and splits every datagram off as an owned slice sharing that allocation. reserve reclaims the buffer once the previous batch's slices have been dropped, so it only reallocates while some datagrams are still in flight, and empty polls don't allocate at all. The rest of the receive path stays zero-copy.

This uses the reserve approach @EileenGalvin suggested on #2734.

Allocations:
- dhat, 200 MiB loopback bulk transfer: receive-path allocations drop from ~38.6k blocks to ~6.7k.
- Counting allocator in the perf client, 4 GiB loopback download: whole-process allocations drop from ~207k to ~57k per GiB (-72%).

Throughput and CPU:
- No throughput change within noise, on loopback and across a netem veth pair (no delay, 10 and 40 ms RTT, 0.1% loss). No regression anywhere.
- Client CPU per GiB is unchanged on x86: an alloc+free of these small buffers is about 25 ns here, so cutting ~65k allocs/s at 450 MiB/s is under 1% of CPU. The benefit grows where allocation is a larger share of the budget: @EileenGalvin reported that this plus #2720 took an edge device from 100% to 60% peak CPU at the same throughput.

Fixes #2734

### DIFF
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -863,6 +863,14 @@ impl RecvState {
             match socket.poll_recv(cx, &mut iovs, &mut metas) {
                 Poll::Ready(Ok(msgs)) => {
                     self.recv_limiter.record_work(msgs);
+                    // Copy the received batch into `datagrams` and split each datagram off as
+                    // an owned slice. We can't read directly into `datagrams` and drop
+                    // `recv_buf`: once a slice has been handed to a connection the allocation
+                    // is shared, so the next `reserve` must allocate fresh backing memory.
+                    // Reading in place would therefore mean reserving the worst-case iovec
+                    // envelope before every poll and reallocating it whenever a prior datagram
+                    // is still in flight. `recv_buf` stays an allocate-once scratch buffer, and
+                    // we reserve only the bytes actually received.
                     let batch_len = metas.iter().take(msgs).map(|meta| meta.len).sum();
                     self.datagrams.reserve(batch_len);
                     for (meta, buf) in metas.iter().zip(iovs.iter()).take(msgs) {
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -809,6 +809,8 @@ struct RecvState {
     incoming: VecDeque<proto::Incoming>,
     connections: ConnectionSet,
     recv_buf: Box<[u8]>,
+    /// Reused buffer that received datagrams are copied into and split into owned slices.
+    datagrams: BytesMut,
     recv_limiter: WorkLimiter,
 }
 
@@ -818,12 +820,10 @@ impl RecvState {
         max_receive_segments: usize,
         endpoint: &proto::Endpoint,
     ) -> Self {
-        let recv_buf = vec![
-            0;
-            endpoint.config().get_max_udp_payload_size().min(64 * 1024) as usize
-                * max_receive_segments
-                * BATCH_SIZE
-        ];
+        let datagram_capacity = endpoint.config().get_max_udp_payload_size().min(64 * 1024)
+            as usize
+            * max_receive_segments;
+        let recv_buf = vec![0; datagram_capacity * BATCH_SIZE];
         Self {
             connections: ConnectionSet {
                 senders: FxHashMap::default(),
@@ -832,6 +832,7 @@ impl RecvState {
             },
             incoming: VecDeque::new(),
             recv_buf: recv_buf.into(),
+            datagrams: BytesMut::with_capacity(datagram_capacity),
             recv_limiter: WorkLimiter::new(RECV_TIME_BOUND),
         }
     }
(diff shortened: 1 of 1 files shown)


djc (approved):

Ralith (changes requested):
Nice improvement, thanks! However, wouldn't it be better to read directly into the `BytesMut` to begin with, replacing `recv_buf` entirely? If not, please document why the extra copy is justified.

ojhermann:
The copy isn't new here — `main` already copies each datagram out of `recv_buf` with `buf[..meta.len].into()`, allocating a fresh `BytesMut` per datagram. This PR just aims those copies at one reused buffer.

Reading directly into the `BytesMut` and dropping `recv_buf` doesn't quite work out, because the two buffers play different roles:

- `recv_buf` is scratch we never hand out, so it's allocated once and reused for the life of the endpoint. It has to stay worst-case-sized (`datagram_capacity * BATCH_SIZE`) because `recvmmsg` needs every iovec slot pre-sized before the call.
- `datagrams` *is* handed out — every `split_to` gives a connection a refcounted slice. As soon as one slice is still live, the next `reserve` has to grab fresh backing memory; we can't reuse an allocation whose contents we've already given away.

So reading straight into the `BytesMut` would mean reserving the full worst-case iovec envelope before every poll, and reallocating that whole envelope whenever a datagram from the previous batch is still in flight — which is most polls under load. That wouldn't obviously allocate any less than what's here, and it needs `unsafe` to build the iovecs over uninitialized capacity. The copy is what lets us reserve only the bytes actually received and keep the scratch buffer allocation-free.

I'll add a comment at the copy site spelling this out.

Ralith (approved):
Thanks!

ojhermann:
> Thanks!

Happy to contribute

PR #2736, opened 2026-07-15
Ignore EMSGSIZE error triggered by ICMP Frag Needed (Path To Big)

### LINKED ISSUE #2733: [quinn] I/O error: Message too long
## Bug Description

When running a simple `quinn-perf` benchmark on my testbed (consisting of two VMs interconnected through a 5G test network), `sendmsg` fails on both ends of the connection right before the first packets are sent (see the logs below).

*Client:*
> $ ./quinn-perf client --ip <SERVER_IP> --duration 10
> 2026-07-14T08:48:46.714515Z  WARN perf: Unable to set desired send buffer size. Desired: 2097152, Actual: 425984
> 2026-07-14T08:48:46.714559Z  WARN perf: Unable to set desired recv buffer size. Desired: 2097152, Actual: 425984
> 2026-07-14T08:48:46.800090Z  WARN quinn_udp: sendmsg error: Os { code: 111, kind: ConnectionRefused, message: "Connection refused" }, Transmit: { destination: <SERVER_IP>:4433, src_ip: None, ecn: None, len: 2652, segment_size: Some(1326) }
> ^C2026-07-14T08:48:49.103102Z ERROR perf::client: request failed: connection lost: closed

*Server:*
> $ ./quinn-perf server
> 2026-07-14T08:48:44.068569Z  WARN perf: Unable to set desired send buffer size. Desired: 2097152, Actual: 425984
> 2026-07-14T08:48:44.068664Z  WARN perf: Unable to set desired recv buffer size. Desired: 2097152, Actual: 425984
> 2026-07-14T08:48:46.760799Z ERROR quinn::endpoint: I/O error: Message too long (os error 90)
> 2026-07-14T08:48:46.761083Z ERROR perf::server: request failed: connection lost: the endpoint encountered an internal error and cannot continue with the connection: endpoint driver future was dropped

I'm inclined to believe that this is not a problem with the testbed, as the connection between the two hosts is perfectly stable, and I could verify this by performing tests with `ping`, `iperf3`, and even with earlier builds of `quinn`/`quinn-perf`. Moreover, I've also tried disabling segmentation offloading (which were on by default), but it didn't show any effect on the end result.

## Reproduction

Currently, I couldn't find an exact way to reproduce it, as I have not experienced it neither when running directly on a host (and letting the client and the server connect through loopback), nor in a Mininet environment.

That said, the issue persists on my testbed, and is 100% reproducible, what's more, I was able to trace the bug's first occurrence back to commit [`57ea8a5`](https://github.com/quinn-rs/quinn/commit/57ea8a5124d67f7f2e7f13941330cd2fb05793fb) (#2721).

Zotyamester:
We're currently not handling ICMPv4 Fragmentation Needed and ICMPv6 Packet Too Big (PTB) messages since they cannot be authenticated, and Datagram Packetization Layer Path MTU Discovery (DPLPMTUD) works without it anyways.

A dedicated check for this type became necessary due to a recent change in how internal errors originating from recvmsg are exposed to the caller as errors (rather than letting the runtime silently retry the read).

Fixes #2733.

### DIFF
--- a/quinn-udp/src/lib.rs
+++ b/quinn-udp/src/lib.rs
@@ -28,13 +28,18 @@
 #![warn(clippy::use_self)]
 
 use core::time::Duration;
-use std::net::{IpAddr, Ipv6Addr, SocketAddr};
 #[cfg(unix)]
 use std::os::unix::io::AsFd;
 #[cfg(windows)]
 use std::os::windows::io::AsSocket;
+use std::{
+    io,
+    net::{IpAddr, Ipv6Addr, SocketAddr},
+};
 #[cfg(not(wasm_browser))]
 use std::{sync::Mutex, time::Instant};
+#[cfg(windows)]
+use windows_sys::Win32::Networking::WinSock;
 
 #[cfg(apple_fast)]
 mod apple_fast;
@@ -222,6 +227,32 @@ pub enum TransportErrorPayload {
     },
     /// Other transport-layer or kernel-reported error
     Other,
+}
+
+/// Returns true if the given I/O error corresponds to a message size error
+///
+/// Useful, for example, after invoking [`io::Error::last_os_error()`]
+/// to check if the last OS error was a message size error
+/// (EMSGSIZE on Unix, WSAEMSGSIZE on Windows).
+///
+/// Note: EMSGSIZE's value is not standardized across OSes (90 on Linux,
+/// 40 on macOS/iOS/BSD; on Windows, `io::Error::raw_os_error()` returns
+/// the Winsock error WSAEMSGSIZE (10040) via `GetLastError()`, which is
+/// distinct from the MSVCRT `errno.h` EMSGSIZE value of 115, which is
+/// never actually populated by socket operations).
+pub fn is_msg_size_err(err: &io::Error) -> bool {
+    #[cfg(unix)]
+    {
+        err.raw_os_error() == Some(libc::EMSGSIZE)
+    }
+    #[cfg(windows)]
+    {
+        err.raw_os_error() == Some(WinSock::WSAEMSGSIZE)
+    }
+    #[cfg(not(any(unix, windows)))]
+    {
+        false
+    }
 }
 
 /// Log at most 1 IO error per minute
--- a/quinn/src/endpoint.rs
+++ b/quinn/src/endpoint.rs
@@ -40,7 +40,7 @@ use rustc_hash::FxHashMap;
 use socket2::{Domain, Protocol, Socket, Type};
 use tokio::sync::{Notify, futures::Notified, mpsc};
 use tracing::{Instrument, Span};
-use udp::{BATCH_SIZE, RecvMeta};
+use udp::{BATCH_SIZE, RecvMeta, is_msg_size_err};
 
 use crate::{
     ConnectionEvent, EndpointConfig, IO_LOOP_BOUND, RECV_TIME_BOUND, VarInt,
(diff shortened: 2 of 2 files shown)


quinn-udp/src/lib.rs
```diff
@@ -143,6 +145,30 @@ impl Default for RecvMeta {
...
+
+/// Returns true if the given error corresponds to a message size error returned by,
+/// for example, [`io::Error::raw_os_error`] (EMSGSIZE on Unix, WSAEMSGSIZE on Windows)
+///
+/// Note: EMSGSIZE's value is not standardized across OSes (90 on Linux,
+/// 40 on macOS/iOS/BSD; on Windows, `io::Error::raw_os_error()` returns
+/// the Winsock error WSAEMSGSIZE (=10040) via `GetLastError()`, which is
+/// distinct from the MSVCRT `errno.h` EMSGSIZE value of 115, which is
+/// never actually populated by socket operations).
+pub fn is_msg_size_err(err: &io::Error) -> bool {
```
djc: Nit: let's move this after `TransportErrorPayload`.

quinn-udp/src/lib.rs
```diff
@@ -143,6 +145,30 @@ impl Default for RecvMeta {
     }
 }
 
+
+/// Returns true if the given error corresponds to a message size error returned by,
+/// for example, [`io::Error::raw_os_error`] (EMSGSIZE on Unix, WSAEMSGSIZE on Windows)
```
djc: Nit: the initial sentence of the docstring should fit on a single line.

quinn-udp/src/lib.rs
```diff
@@ -143,6 +145,30 @@ impl Default for RecvMeta {
...
+/// distinct from the MSVCRT `errno.h` EMSGSIZE value of 115, which is
+/// never actually populated by socket operations).
+pub fn is_msg_size_err(err: &io::Error) -> bool {
+    #[cfg(unix)]
+    {
+        err.raw_os_error() == Some(libc::EMSGSIZE)
+    }
+    #[cfg(windows)]
+    {
+        err.raw_os_error() == Some(windows_sys::Win32::Networking::WinSock::WSAEMSGSIZE)
```
djc: Nit: import `WinSock` or `WSAEMSGSIZE` directly at the top of the module, please.

djc (approved):
LGTM modulo some nits, thanks!

djc:
Don't forget to run `cargo fmt`.

zoltanszatmary:
> Don't forget to run cargo fmt.

Fair point, sorry about that. Except, the only thing I forgot to run was that. Build, check, fix, clippy were all run before pushing.

mxinden (approved):
In favor. Allows me to remove the exact same code on the Firefox side.

https://github.com/mozilla/neqo/blob/8a04d065c2d35c8e8fd804f91c7081ab6bb60b89/neqo-udp/src/lib.rs#L108-L123

zoltanszatmary:
> Allows me to remove the exact same code on the Firefox side.

Perhaps we can just relocate the function here? It appears to be somewhat more robust. For example, it also handles the edge case of Windows USO (UDP Segmentation Offload) that I didn't cover here.

Ralith (approved):

djc:
> > Allows me to remove the exact same code on the Firefox side.
> 
> Perhaps we can just relocate the function here? It appears to be somewhat more robust. For example, it also handles the edge case of Windows USO (UDP Segmentation Offload) that I didn't cover here.

@mxinden do you think that would make sense? Open to having this as a followup.

The other thing I was wondering is, should we feed this back into PMTUD as a signal?

Issue #1899, opened 2024-06-15
Bundle flow control and ACK frames opportunistically


Ralith:
`MAX_DATA`, `MAX_STREAM_DATA`, `MAX_STREAMS`, and `ACK` frames are all sent automatically by endpoints receiving application data, and usually do not need to be sent at the first opportunity. We currently reduce traffic by relying on a significance threshold (e.g. 1/8 of a flow control window, or a certain time or packet count for ACKs) to judge whether they must be sent. This fails to reliably include these automatically generated frames in packets bearing application data, even if such packets are being sent regularly (e.g. responses to requests).

We could capture this opportunity by introducing an additional, separate, lower significance threshold at which flow control and/or ACK frames will be included in an outgoing packet. If packets are being sent regularly, then this threshold will typically be crossed and a frame included before a free-standing packet would become necessary.

We should be careful not to generate too many ACK frames even when bundled in otherwise-scheduled packets, as they are thought to be relatively expensive for the peer to process. Flow control frames have no such constraint and need only consider the cost of the space to encode them.

PR #2747, opened 2026-07-21
Bundle ACK in packets with DATAGRAM or STREAM frames

anatawa12:
Implements #1899 but only for ACK frames.
This implements bundling ACK for DATAGRAM and STREAEM packet when there are room in packet.


quinn-proto/src/connection/mod.rs
```diff
@@ -3446,20 +3442,37 @@ impl Connection {
             self.stats.frame_tx.stream += sent.stream_frames.len() as u64;
         }
 
+        // Bundle ACK for data frames when there is room for them
+        if (sent_datagrams || !sent.stream_frames.is_empty())
```
Ralith: Why restrict to these cases only?

quinn-proto/src/connection/spaces.rs
```diff
@@ -659,11 +662,16 @@ impl PendingAcks {
             .map(|earliest_unacked| earliest_unacked + max_ack_delay)
     }
 
-    /// Whether any ACK frames can be sent
+    /// Whether any ACK frames can be sent in ack initiated packet
```
Ralith: "ack initiated packet" is unconventional terminology. Maybe something like "even if doing so would require a dedicated packet"?

quinn-proto/src/connection/mod.rs
```diff
@@ -3063,6 +3067,7 @@ impl Connection {
         {
             self.timers
                 .set(Timer::MaxAckDelay, now + self.ack_frequency.max_ack_delay);
+            self.next_bundled_ack_time = Some(now);
```
Ralith, line 3092: Why short-circuit here? Seems like it could result in far more ACKs being sent than we might want.

quinn-proto/src/connection/mod.rs
```diff
@@ -3446,20 +3442,37 @@ impl Connection {
             self.stats.frame_tx.stream += sent.stream_frames.len() as u64;
         }
 
+        // Bundle ACK for data frames when there is room for them
```
Ralith: Document in more detail why we want to do this.

quinn-proto/src/tests/mod.rs
```diff
@@ -3490,6 +3490,29 @@ fn voluntary_ack_with_large_datagrams() {
...
+    pair.drive_client();
+    pair.drive_server();
+    pair.server_datagrams(server_ch).send(vec![0; 1].into(), false).unwrap();
+    pair.drive_server();
+
+    let server_tx_after_datagrams = pair.server_conn_mut(client_ch).stats().udp_tx.bytes;
+
+    pair.drive();
+
+    // No more packets should be sent from server since ACK to the first packet has been sent
```
Ralith: Would it be simpler and more robust to check the number of ACK frames and packets received?

quinn-proto/src/tests/mod.rs
Ralith, line 1: When fixing code in the same PR, please squash the fix into the affected commit.

Ralith (changes requested):

quinn-proto/src/connection/mod.rs
anatawa12: I want to ensure some other frames are actually included in current packet.
in my understanding, in cases like trying to send datagram frames with data larger than mtu, no frames can be created and we must not create packet only with ACK in this case.
it's mostly prevented by Datagrams::send method but in my understanding MTU might be smaller than the time Datagrams::send was called when if path have changed.

There is no reason for not checking other frame types.

It's cleaner to add flag to track if we can send bundled ack and set them true in most frames so I'll do that
EDIT: it's easier to compare buffer size to check if there are any frames added to the packet so I'll do that instead.

quinn-proto/src/connection/mod.rs
anatawa12: I added short-circuit because 2nd ack-eliciting packet would trigger immediate ACK and I want to reduce underlying traffic overhead as possible.
If we small delay on sending bundled ack frames. there might be some outgoing packets and 2nd incoming ack-eliciting packet waiting for the delay and would not reuse packet.

We might add shorter timer to reuse other packet before sending ACK triggered by 2nd ack-eliciting packet, but it's more complex than allowing sending bundled ACK any time.

`bundled_ack_interval` is interval of retransmiting ACKs so not using here is intended.

anatawa12:
Is there anything I need to do to get the updated changes reviewed, or is it just delayed because the reviewers are busy?

As far as I know, some projects prefer PR authors to resolve conversations, while others have reviewers do it, so I haven't resolved the discussions myself. Should I resolve a conversation if I've applied the suggested changes?

Ralith:
I'm not waiting for anything from you, just swamped; thanks for your patience.

I don't think we currently have a written policy for resolving feedback. I don't have a strong preference myself, except that we need *some* signal that you believe you've addressed them. That could be post in the PR, replies to individual comments, or marking the threads as resolved. Maybe @djc has a preference?

Merely updating the branch can be ambiguous, especially when we're too busy to make optimistic passes on code that might be WIP.

djc:
> I don't think we currently have a written policy for resolving feedback. I don't have a strong preference myself, except that we need _some_ signal that you believe you've addressed them. That could be post in the PR, replies to individual comments, or marking the threads as resolved. Maybe @djc has a preference?

I prefer for the author to resolve comment threads when they're reasonably sure they've resolved the feedback. That makes it easier for the reviewer to figure out what still needs to be addressed/discussed.

quinn-proto/src/connection/mod.rs
```diff
@@ -3446,20 +3450,41 @@ impl Connection {
...
+
         sent
     }
 
     /// Write pending ACKs into a buffer
     ///
     /// This method assumes ACKs are pending, and should only be called if
     /// `!PendingAcks::ranges().is_empty()` returns `true`.
-    fn populate_acks(
+    fn try_populate_acks(
```
Ralith, line 3507: Update the doc comment to describe the new, non-obvious behavior.

quinn-proto/src/connection/mod.rs
```diff
@@ -3446,20 +3450,41 @@ impl Connection {
             self.stats.frame_tx.stream += sent.stream_frames.len() as u64;
         }
 
+        // Bundle ACK with other frames when there is room for them.
+        // We want to reuse encryption and unnderlying protocol overhead,
```
Ralith: nit: "unnderlying" is a typo

quinn-proto/src/config/transport.rs
```diff
@@ -385,6 +386,7 @@ impl Default for TransportConfig {
             mtu_discovery_config: Some(MtuDiscoveryConfig::default()),
             pad_to_mtu: false,
             ack_frequency_config: None,
+            bundled_ack_interval: Duration::from_millis(100),
```
Ralith: Did you intend for this to be user-configurable? If so, you'll also need to add a setter. If not, it would be simpler to define a constant.

quinn-proto/src/connection/mod.rs
```diff
@@ -192,6 +192,8 @@ pub struct Connection {
     permit_idle_reset: bool,
     /// Negotiated idle timeout
     idle_timeout: Option<Duration>,
+    /// The time we send next bundled ACK
+    next_bundled_ack_time: Option<Instant>,
```
Ralith, line 200: Why do we choose a time threshold rather than a packet threshold? Document.

quinn-proto/src/config/transport.rs
Ralith: Ideally we should find a way to automatically select reasonable behavior rather than hard-coding a default that only makes sense for certain cases and relying on users to know what to override when. Users will typically not fully understand every option we expose, and quinn has diverse applications, including e.g. deep space networking experiments with round-trip latency many orders of magnitude higher than this value.

Is there a way we could avoid relying on a manually chosen threshold here? Maybe we could use some fraction of the effective ack delay interval, rather than a completely independent value?

Ralith (approved):
This looks very clean, thank you! I'm wary of adding knobs that users are unlikely to set correctly, but it does seem likely that this will improve behavior in most scenarios as written. Thanks also for your patience with our review backlog.

Any chance of some benchmarks demonstrating the impact? I imagine it should be pretty substantial for the right bidirectional (e.g. request/response oriented) workload.

quinn-proto/src/config/transport.rs
anatawa12: I was intended to make it configurable, but I forgot adding setter.
I chose 100ms for a random not small value and expected to be similar to or less than RTT, but I can understand deep space networking will have larger RTT .

Therefore, I replaced `bundled_ack_interval` with RTT-based automatically computed delay in 26ce2e1bef80014efbe01e6222b5121c7c0ac597. It's similar to pto but transmitting bundled ACK is much cheaper than transmitting STREAM frame so I choose simpler estimation that does eager transmittion.

anatawa12:
Benchmark is not ready but I'll try writing benchmark later.

Ralith:
I'd expect you could demonstrate the benefit with the existing perf client/server without specific new code, FWIW.

anatawa12:
I was trying to include an additional benchmark for bidi, non-request-response-style data streaming and datagram, so I wrote "I'll try writing a benchmark", but yes, request-response-style performance is much important for QUIC, especially since it is similar to H3.

Here is a comparison between c8b02d1a (current base) and ce155d87 (current head). The tests were run on Windows 11 with almost all applications closed, using the bench profile.

I ran each combination three times and selected the run with the best RPS. The full benchmark results are available [here][gist].

It looks like applying this patch on the server side improves both RPS and download throughput.
The client-side change does not seem to have much impact on performance in this benchmark. The best result below shows a small performance regression, but I think it is within the margin of error when compared with the other runs in the [gist].


**head server, head client**
```
Overall stats:
RPS: 246.13 (14769 requests in 60.00s)

Stream metrics:

 AVG  │          2.21ms │            1.84ms │    14.00µs │      3836.70 Mb/s │      10065.91 Mb/s
 P0   │          1.99ms │           24.00µs │     0.00ns │      1284.51 Mb/s │       2225.08 Mb/s
 P10  │          2.10ms │            1.82ms │     0.00ns │      3657.43 Mb/s │       4274.00 Mb/s
 P50  │          2.14ms │            1.86ms │     0.00ns │      3921.67 Mb/s │       4521.46 Mb/s
 P90  │          2.30ms │            1.96ms │    54.00µs │      3990.88 Mb/s │       4617.93 Mb/s
 P100 │          6.53ms │            3.77ms │   271.00µs │      4225.76 Mb/s │     349771.40 Mb/s
```

**head server, base client**

```
Overall stats:
RPS: 246.64 (14799 requests in 60.00s)

Stream metrics:

 AVG  │          2.20ms │            1.84ms │    14.00µs │      3852.69 Mb/s │      10493.44 Mb/s
 P0   │          1.94ms │           24.00µs │     0.00ns │      1382.02 Mb/s │       2009.07 Mb/s
 P10  │          2.09ms │            1.82ms │     0.00ns │      3670.02 Mb/s │       4250.93 Mb/s
 P50  │          2.13ms │            1.86ms │     0.00ns │      3938.45 Mb/s │       4525.65 Mb/s
 P90  │          2.29ms │            1.97ms │    53.00µs │      4009.75 Mb/s │       4609.54 Mb/s
 P100 │          6.07ms │            4.18ms │   619.00µs │      4315.94 Mb/s │     349771.40 Mb/s
```

**base server, head client**

```
Overall stats:
RPS: 214.14 (12849 requests in 60.00s)

Stream metrics:

 AVG  │          2.80ms │            1.86ms │    16.00µs │      3790.56 Mb/s │       7689.47 Mb/s
 P0   │          1.98ms │           27.00µs │     0.00ns │       182.58 Mb/s │       2107.64 Mb/s
 P10  │          2.10ms │            1.82ms │     0.00ns │      3619.68 Mb/s │       4206.89 Mb/s
 P50  │          2.14ms │            1.86ms │     1.00µs │      3930.06 Mb/s │       4508.88 Mb/s
 P90  │          2.32ms │            2.00ms │    54.00µs │      4003.46 Mb/s │       4617.93 Mb/s
 P100 │         45.95ms │            3.98ms │   975.00µs │      4242.54 Mb/s │     310848.26 Mb/s
 ```
 
 **base server, base client**
 
 ```
Overall stats:
RPS: 219.24 (13155 requests in 60.00s)

Stream metrics:

─────┼─────────────────┼───────────────────┼────────────┼───────────────────┼────────────────────
 AVG  │          2.69ms │            1.86ms │    13.00µs │      3809.60 Mb/s │       7033.54 Mb/s
 P0   │          1.98ms │           31.00µs │     0.00ns │       182.58 Mb/s │       2239.76 Mb/s
 P10  │          2.10ms │            1.82ms │     0.00ns │      3628.07 Mb/s │       4225.76 Mb/s
 P50  │          2.13ms │            1.86ms │     0.00ns │      3936.35 Mb/s │       4525.65 Mb/s
 P90  │          2.31ms │            1.99ms │    53.00µs │      4005.56 Mb/s │       4617.93 Mb/s
 P100 │         45.95ms │            3.74ms │   249.00µs │      4236.25 Mb/s │     270717.16 Mb/s
 ```

[gist]: https://gist.github.com/anatawa12/ff7ce87e56ee47138021f956527a978a

**Review by Ralith — APPROVED** [review]
LGTM, thanks! Please squash the reworked delay computation and test fix down into the prior commits.

**Comment by anatawa12** [comment]
I have rebased onto latest main branch, with squashing to three commits: basic implementation, frequency improvement, and adding test.

File: quinn-proto/src/tests/mod.rs [hunk 10]
```diff
@@ -3343,7 +3343,12 @@ fn stream_gso() {
 fn datagram_gso() {
     let _guard = subscribe();
     let mut pair = Pair::default();
-    let (client_ch, _) = pair.connect();
+    let (client_ch, server_ch) = pair.connect();
+
+    // Sending ack-eliciting packet from server let client send ACK, which prevents
+    // sending bundled ACK for a while.
+    pair.server_datagrams(server_ch).send(Bytes::new(), false).unwrap();
+    pair.drive();
```
anatawa12, line 3476: The previous (b025aadf24cc39c49d28559761479ca20016338e) test was flaky because `PacketNumberFilter` might skip a related packet number and would make ACK frame larger than normal.
This is the root cause of https://github.com/quinn-rs/quinn/actions/runs/32320541444/job/96281705013?pr=2747 failure. I confirmed locally by fixing initial `next_skipped_packet_number` to 7.

Sending ack-eliciting packet before sending packets from client prevents sending bundled ACK frames while testing gso so I have changed the test to do that.

Ralith: Works for me.

Ralith (approved):

quinn-proto/src/connection/mod.rs
```diff
@@ -3191,6 +3192,8 @@ impl Connection {
         let is_0rtt = space_id == SpaceId::Data && space.crypto.is_none();
         space.pending_acks.maybe_ack_non_eliciting();
 
+        let buf_initial_len = buf.len();
```
djc: What does `initial` mean here? Can we be more specific, like maybe `pre_payload_len` or something?

quinn-proto/src/connection/mod.rs
```diff
@@ -3471,7 +3525,14 @@ impl Connection {
             delay_micros
         );
 
+        let original_len = buf.len();
```
djc: Nit: let's call this something like `no_acks_len`?

quinn-proto/src/connection/mod.rs
```diff
@@ -3433,20 +3465,43 @@ impl Connection {
...
+                max_size,
+            );
+        }
+
         sent
     }
 
-    /// Write pending ACKs into a buffer
+    /// Tries to write pending ACKs into a buffer if there is enough space. If the ACK frame does
+    /// not fit into the buffer, the ACK frame will not be sent at all.
```
djc: Nit: the first line of a documentation comment should be a single sentence (fragment) by itself, so please keep this shorter and move elaboration into further on.

quinn-proto/src/connection/mod.rs
```diff
@@ -790,13 +798,14 @@ impl Connection {
...
                 if !self.spaces[space_id].pending_acks.ranges().is_empty() {
-                    Self::populate_acks(
+                    Self::try_populate_acks(
                         now,
                         self.receiving_ecn,
                         &mut SentFrames::default(),
                         &mut self.spaces[space_id],
                         buf,
                         &mut self.stats,
+                        usize::MAX,
```
djc: This seems kind of crazy, don't we have a more realistic value we can use here like maybe `buf_capacity`?

djc (approved):
Thanks for working on this, LGTM modulo a few nits.

quinn-proto/src/connection/mod.rs
anatawa12: I agree with that buf_capacity is a reasonable value. changed in a99b20a

quinn-proto/src/connection/mod.rs
anatawa12: addressed in a99b20a6 and e3f4ac2a (for another method)

quinn-proto/src/connection/mod.rs
anatawa12: addressed in a99b20a6

quinn-proto/src/connection/mod.rs
anatawa12: addressed in a99b20a6

djc:
Thanks!

If you're interested, might be nice to backport this to `0.11.x` to make it available sooner.

PR #2748, opened 2026-07-22
feat(udp): support partial sends

### LINKED ISSUE #2399: udp: retry failed GSO sendmsg on EIO without GSO

thomaseizinger:
This PR offers an alternative to #2672 that doesn't require buffering. For a detailed description, see the commit message.

Primarily, this fixes silent packet drops on the apple fast data path but it is also useful for other platforms to report only partial progress of the passed in `Transmit`.

### DIFF
--- a/quinn-udp/src/lib.rs
+++ b/quinn-udp/src/lib.rs
@@ -35,6 +35,7 @@ use std::os::windows::io::AsSocket;
 use std::{
     io,
     net::{IpAddr, Ipv6Addr, SocketAddr},
+    num::NonZeroUsize,
 };
 #[cfg(not(wasm_browser))]
 use std::{sync::Mutex, time::Instant};
@@ -158,13 +159,80 @@ pub struct Transmit<'a> {
     /// Contents of the datagram
     pub contents: &'a [u8],
     /// The segment size if this transmission contains multiple datagrams.
-    /// This is `None` if the transmit only contains a single datagram
+    ///
+    /// This is `None` if the transmit only contains a single datagram.
+    /// A value of zero is treated like `None`.
     pub segment_size: Option<usize>,
     /// Optional source IP address for the datagram
     pub src_ip: Option<IpAddr>,
 }
 
-impl Transmit<'_> {
+impl<'a> Transmit<'a> {
+    /// Returns the number of datagrams encoded by this transmit.
+    ///
+    /// A transmit without a `segment_size` always represents one datagram, including when its
+    /// contents are empty.
+    pub fn datagram_count(&self) -> usize {
+        match self.segment_size {
+            Some(size) if size != 0 && size < self.contents.len() => {
+                self.contents.len().div_ceil(size)
+            }
+            Some(_) | None => 1,
+        }
+    }
+
+    /// Advances this transmit by `sent` datagrams.
+    ///
+    /// Returns the unsent remainder, or `None` if the entire transmit was consumed. The returned
+    /// transmit borrows a suffix of the same payload; packet data is never copied.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// use std::{io, net::UdpSocket};
+    ///
+    /// use quinn_udp::{Transmit, UdpSocketState};
+    ///
+    /// fn send_pending<'a>(
+    ///     state: &UdpSocketState,
+    ///     socket: &UdpSocket,
+    ///     pending: &mut Option<Transmit<'a>>,
+    /// ) -> io::Result<()> {
+    ///     while let Some(transmit) = pending.take() {
+    ///         match state.send(socket.into(), &transmit) {
+    ///             Ok(sent) => *pending = transmit.advance(sent),
+    ///             Err(error) => {
+    ///                 *pending = Some(transmit);
+    ///                 return Err(error);
+    ///             }
+    ///         }
+    ///     }
+    ///
+    ///     Ok(())
+    /// }
+    /// ```
+    pub fn advance(mut self, sent: SendCount) -> Option<Self> {
+        let sent = sent.get();
+        let segment_size = match self.segment_size {
+            Some(size) if size != 0 && size < self.contents.len() => size,
+            Some(_) | None => {
+                debug_assert_eq!(sent, 1, "a single-datagram transmit must consume once");
+                return None;
+            }
+        };
+
+        debug_assert!(sent <= self.contents.len().div_ceil(segment_size));
+
+        let offset = sent.saturating_mul(segment_size);
+        if offset >= self.contents.len() {
+            return None;
+        }
+
+        self.contents = &self.contents[offset..];
+
+        Some(self)
+    }
+
     /// Computes the effective segment-size of the packet.
     ///
     /// Some (older) network drivers don't like being told to do GSO even if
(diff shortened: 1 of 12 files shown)


thomaseizinger:
cc @larseggert

larseggert:
I like it!

quinn-udp/src/lib.rs
```diff
@@ -178,10 +180,104 @@ impl Transmit<'_> {
...
         }
     }
+
+    /// Returns the number of datagrams encoded by this transmit.
+    ///
+    /// A transmit without a `segment_size` always represents one datagram, including when its
+    /// contents are empty.
+    pub fn datagram_count(&self) -> usize {
+        match self.segment_size {
+            Some(0) => panic!("segment size must be non-zero"),
```
thomaseizinger: This is a bit of an annoying case.

Has it been discussed before to change segment size to a `NonZeroUsize`?

quinn-udp/src/unix.rs
```diff
@@ -235,24 +235,39 @@ impl UdpSocketState {
...
+        let attempted_datagrams = attempted.datagram_count();
+
+        match send(self, socket.0, attempted) {
+            Ok(sent) => Ok(SendCount::from_datagram_count(sent)),
             Err(e) if e.kind() == io::ErrorKind::WouldBlock => Err(e),
             // - EMSGSIZE is expected for MTU probes. Future work might be able to avoid
             //   these by automatically clamping the MTUD upper bound to the interface MTU.
-            Err(e) if e.raw_os_error() == Some(libc::EMSGSIZE) => Ok(()),
+            Err(e) if e.raw_os_error() == Some(libc::EMSGSIZE) => {
+                Ok(SendCount::from_datagram_count(attempted_datagrams))
```
thomaseizinger: This is also somewhat unfortunate. If it were after me, we should remove /deprecate the `send` functions and only keep `try_send`.

quinn-udp/src/lib.rs
thomaseizinger: I've removed this in favor of treating `Some(0)` like an unsegmented datagram.

thomaseizinger:
@mxinden Would appreciate a review of this!

quinn-udp/src/unix.rs
```diff
@@ -235,24 +235,39 @@ impl UdpSocketState {
...
-                Ok(())
+                Ok(SendCount::from_datagram_count(attempted_datagrams))
             }
         }
     }
 
-    /// Sends a [`Transmit`] on the given socket without any additional error handling
-    pub fn try_send(&self, socket: UdpSockRef<'_>, transmit: &Transmit<'_>) -> io::Result<()> {
-        send(self, socket.0, transmit)
+    /// Sends a prefix of a [`Transmit`] without any additional error handling.
```
mxinden, line 204: "without any additional error handling" isn't quite right, as e.g. we handle `EIO`, right?

quinn-udp/src/unix.rs
```diff
@@ -471,13 +486,59 @@ impl UdpSocketState {
...
+            {
+                if state.max_gso_segments.swap(1, Ordering::Relaxed) > 1 {
+                    crate::log::info!(
+                        "`libc::sendmsg` failed with {e}; halting segmentation offload"
+                    );
+                }
+
+                transmit = transmit.limit(1);
+                continue;
+            }
```
mxinden, line 512: Do I understand correctly that on `EIO` quinn-udp would only send the first segment in the batch and then require the caller to retry with the remaining?

Theoretically we might as well send all of them, up until e.g. `WouldBlock`.

Why did you choose to send a single one only? Granted, the caller needs to implement the retry logic anyways.

mxinden (commented):
Thank you for doing this work.

This PR is no longer `sendmsg_x` specific. How about `feat(udp): support partial sends` instead?

Calling out that this is a breaking change. Unfortunately I assume this will delay merging here quite a bit.

Implementation looks good to me. Very much in favor of the strong integration into the type system through new helper types.

With this change the caller is required to implement retry logic themselves. What do you think of a Rust doc comment, either on crate level, or attached to the function showing a sample implementation?

quinn-udp/src/unix.rs
thomaseizinger: Right now, `SendCount` is implemented using a `NonZeroUsize`. This is modelled after the system APIs: If a `sendmmsg` or `sendmsg_x` returns with success, it will at least have sent 1 message. We can therefore not return `SendCount(0)` and have the caller retry the entire `Transmit`.

What is implemented as part of this PR as well is that we honor the `max_gso_segments` value (see the first line in `try_send`). Here in this particular code-path, we are setting `max_gso_segments` to 1 and thus we only retry the first packet. This is the same behaviour that a user will get if they continue to submit datagrams that are larger than the `max_gso_segments` value: Only one of them will be sent at a time and their outer retry loop needs to handle the remaining ones.

I think limiting to 1 packet here makes this behaviour more consistent. It would also be more difficult to implement because we'd have to store both the intermediary `Transmit` and the original one. I don't see a benefit in the complexity because the API forces the caller to retry anyway.

quinn-udp/src/unix.rs
thomaseizinger: These docs are in reference to the `send` function. The error handling we do is IMO an implementation detail that is necessary to uphold the API contract.

thomaseizinger: If it were after me, we should completely deprecate the `send` functions. I think they are foot-guns and especially with this API being added, they add another layer where packets gets silently dropped which will leave someone debugging for hours before they find out.

thomaseizinger:
> With this change the caller is required to implement retry logic themselves. What do you think of a Rust doc comment, either on crate level, or attached to the function showing a sample implementation?

I added one, what do you think?

`quinn-udp`s nature of being runtime agnostic makes this a bit tricky but I think the example does explain quite well how to use this correctly. Plus, the compiler should guide users towards correct usage.

quinn-udp/src/unix.rs
mxinden: I am fine with deprecating the `send` function.

quinn-udp/src/unix.rs
mxinden: Thanks for the details. This makes sense. 

Intuitively, as a user, I would expect `quinn-udp` to send as much as it can. That said, I am fine either way.

mxinden (approved):

quinn-udp/src/unix.rs
thomaseizinger: > Intuitively, as a user, I would expect `quinn-udp` to send as much as it can.

Yeah I somewhat agree with that. I think it is okay to claim that we basically do that as long as you are using the API correctly. And the type-system and docs steer you into that direction.

quinn-udp/src/unix.rs
thomaseizinger: Should we do that as another PR onto main and then rebase this work onto it? I don't think it is really in-scope for this PR.

djc:
> Calling out that this is a breaking change. Unfortunately I assume this will delay merging here quite a bit.

I don't think it has to? We can just push out a semver-incompatible release for this while `main` is targeting Quinn 0.12 development.

> With this change the caller is required to implement retry logic themselves. What do you think of a Rust doc comment, either on crate level, or attached to the function showing a sample implementation?

Could/should we have an implementation for use in quinn exposed for other users?

thomaseizinger:
> Could/should we have an implementation for use in quinn exposed for other users?

Do you mean a runtime integration? Or something more modular where users have to plug in a future that can suspend on `WouldBlock`?

I don't think it is worth it. It is really not that much code and looking at how we integrate `quinn-udp` (https://github.com/firezone/firezone/blob/f840f4eeded0884f027823b935112f305597c73a/rust/libs/connlib/socket-factory/src/lib.rs#L406-L722) to handle things like connected sockets on Apple to get proper backpressure on ENOSPC etc it seems to me like this loop is really something you want to have control over and any provided integration is likely to fall short in several ways (logging, error handling policy, metrics, etc)

PR #2749, opened 2026-07-23
congestion: saturate CUBIC window increment to avoid overflow

LesnyRumcajs:
Resolves a panic:
```
`thread 'tokio-rt-worker' (2693719) panicked at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/quinn-proto-0.11.16/src/congestion/cubic.rs:149:17:
  attempt to add with overflow
```

`Cubic::on_ack` computes `cubic_cwnd += cubic_inc as u64` (`congestion/cubic.rs:161`). `w_cubic` grows cubically with the time since the last congestion event, so on a long-lived connection with no loss for roughly a week, `cubic_inc` exceeds `u64::MAX` and the addition overflows; a panic in builds with `overflow-checks = true`, a silent wrap otherwise.

Kind of along the lines of the ones below: https://github.com/quinn-rs/quinn/blob/fec2f8960df489767bfccef5b08b72013f1b992a/quinn-proto/src/congestion/cubic.rs#L164-L169

### DIFF
--- a/quinn-proto/src/congestion/cubic.rs
+++ b/quinn-proto/src/congestion/cubic.rs
@@ -158,7 +158,9 @@ impl Controller for Cubic {
                 let cubic_inc =
                     (w_cubic - cubic_cwnd as f64) / cubic_cwnd as f64 * self.current_mtu as f64;
 
-                cubic_cwnd += cubic_inc as u64;
+                // w_cubic grows cubically with the time since the last congestion
+                // event and can exceed `u64::MAX` after a long lossless period.
+                cubic_cwnd = cubic_cwnd.saturating_add(cubic_inc as u64);
             }
 
             // Update the increment and increase cwnd by MSS.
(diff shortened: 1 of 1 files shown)


Ralith:
Hmm, we shouldn't be increasing the congestion except when sending at congestion-limited rate, which should be pretty difficult to do without packet loss when your window gets big enough for this to be a threat. Still, this change is clearly reasonable.

Ralith (approved):

LesnyRumcajs:
Just for reference, this crash happened on a node running for around seven days, the entire thing running on a Digital Ocean VPS.

djc (approved):
Thanks!

Would you be able to submit a backport to the `0.11.x` branch for this?

LesnyRumcajs:
> Would you be able to submit a backport to the 0.11.x branch for this?

Will do!

PR #2750, opened 2026-07-23
Fix quinn-udp compile errors and assertions on musl

### LINKED ISSUE #2703: Failure to build on Linux musl targets
Recent changes relative to the support of `IP_RECVERR` (https://github.com/quinn-rs/quinn/pull/2646) broke the compilation of the `main` branch on Linux musl targets.

https://github.com/quinn-rs/quinn/blob/fed0321a9a672819662caab37f5662f1ad91308e/quinn-udp/src/linux.rs#L90

On [musl targets, the `libc::cmsghdr` structure has a different layout](https://docs.rs/libc/latest/x86_64-unknown-linux-musl/libc/struct.cmsghdr.html#structfield.cmsg_len) and the `cmsg_len` has a different type (`u32` when it's [`usize` on GNU targets](https://docs.rs/libc/latest/libc/struct.cmsghdr.html#structfield.cmsg_len)).

I'd be happy to provide a patch but would appreciate some guidance before:

1. should we be using the helpers defined in `quinn-udp/src/cmsg` instead of accessing fields directly?
2. is it acceptable to add a cast and silence the `unnecessary_cast` Clippy lint?
3. are you interested in adding a CI job to check this platform?

thomaseizinger:
This PR adds a new CI check to compile `quinn-udp` on musl. Certain alignments and types are different on musl and `quinn-udp` actually fails to compile and run on musl currently.

The two commits following the CI commit each fix a dedicated bug.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -82,6 +82,21 @@ jobs:
             . "$HOME/.cargo/env"
             cargo build --locked --all-targets && cargo test --locked && cargo test --locked -- --ignored stress && cargo test --locked --manifest-path fuzz/Cargo.toml && cargo test --locked -p quinn-udp --benches
 
+  test-musl:
+    name: test on musl
+    runs-on: ubuntu-latest
+    env:
+      # `cc` looks for a target-prefixed compiler; point it at the one `musl-tools` installs.
+      CC_x86_64_unknown_linux_musl: musl-gcc
+    steps:
+      - uses: actions/checkout@v7
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          target: x86_64-unknown-linux-musl
+      - run: sudo apt-get update && sudo apt-get install -y musl-tools
+      - uses: Swatinem/rust-cache@v2
+      - run: cargo test --locked -p quinn-udp --target x86_64-unknown-linux-musl
+
   test:
     strategy:
       matrix:
--- a/quinn-udp/src/cmsg/mod.rs
+++ b/quinn-udp/src/cmsg/mod.rs
@@ -78,9 +78,10 @@ impl<M: MsgHdr> Drop for Encoder<'_, M> {
 ///
 /// `cmsg` must refer to a native cmsg containing a payload of type `T`
 pub(crate) unsafe fn decode<T: Copy, C: CMsgHdr>(cmsg: &impl CMsgHdr) -> T {
-    assert!(align_of::<T>() <= align_of::<C>());
     debug_assert_eq!(cmsg.len(), C::cmsg_len(size_of::<T>()));
-    ptr::read(cmsg.cmsg_data() as *const T)
+    // The payload is only aligned for `C`, which on musl is less strict than payloads such as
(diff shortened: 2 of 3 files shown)


stormshield-damiend:
Thanks for taking time to fix this issue @thomaseizinger 

There is also this one (linter) #2703

quinn-udp/src/cmsg/mod.rs
```diff
@@ -78,9 +78,10 @@ impl<M: MsgHdr> Drop for Encoder<'_, M> {
 ///
 /// `cmsg` must refer to a native cmsg containing a payload of type `T`
 pub(crate) unsafe fn decode<T: Copy, C: CMsgHdr>(cmsg: &impl CMsgHdr) -> T {
-    assert!(align_of::<T>() <= align_of::<C>());
     debug_assert_eq!(cmsg.len(), C::cmsg_len(size_of::<T>()));
-    ptr::read(cmsg.cmsg_data() as *const T)
+    // The payload is only aligned for `C`, which on musl is less strict than payloads such as
+    // `libc::timespec`, so it cannot be read through an aligned `ptr::read`.
+    ptr::read_unaligned(cmsg.cmsg_data() as *const T)
```
djc, line 84: I'm guessing this would be slower? Should we just use `#[cfg(..)]` so switch these for musl targets?

.github/workflows/rust.yml
```diff
@@ -82,6 +82,21 @@ jobs:
             . "$HOME/.cargo/env"
             cargo build --locked --all-targets && cargo test --locked && cargo test --locked -- --ignored stress && cargo test --locked --manifest-path fuzz/Cargo.toml && cargo test --locked -p quinn-udp --benches
 
+  test-musl:
```
djc, line 85: Are you taking on responsibility to fix this if it starts failing?

stormshield-damiend: As we are planning to use musl, we (stormshield) can be involved in this.

djc: Thanks!

Ralith (approved):

quinn-udp/src/cmsg/mod.rs
Ralith: Are we certain this is necessarily a musl-specific hazard? I suspect any overhead will be pretty trivial.

thomaseizinger: I don't think it is worth the complexity but I haven't done any benchmarking.

.github/workflows/rust.yml
thomaseizinger: Yep, happy to keep this working. We use `quinn-udp` heavily on musl.

thomaseizinger:
> There is also this one (linter) #2703

That is fixed in the second commit, right?

stormshield-damiend:
> > There is also this one (linter) #2703
> 
> That is fixed in the second commit, right?

yes it is (was to reference it in commit if not done)

djc (approved):

PR #2755, opened 2026-07-26
quinn-udp: support wasm32-wasip2 and remove the fallback backend

### LINKED ISSUE #2754: quinn-udp: the fallback backend is not built by any CI target and has stopped compiling
Apologies, I have read the AI guide and I believe the below should be compliant, please let me know if you need any edits.

`fallback.rs` is selected by `cfg(not(any(wasm_browser, unix, windows)))`, and no CI target matches that, so it is never built. 

Three compile errors on current main: `recv_from_vectored not found on SockRef` (fallback.rs:68), `missing field timestamp in the RecvMeta initializer` (:69), `mismatched types` (:121).

**Repro**:
``` bash
rustup target add wasm32-wasip2
cargo check -p quinn-udp --target wasm32-wasip2
```

- send returns Socket::send_to's io::Result<usize> from a function declared io::Result<()>, since d442b504 landed Oct 2024.
- RecvMeta gained timestamp in 9849790f, Jun 2026, and every backend except this one was updated.

So the file is broken for any target that selects it. Nothing about wasi is required to consider it a bug.

It has happened before, same failure mode, repaired in 07222809, Dec 2023.

That's the argument that fixing the three errors isn't the fix. One CI target that selects the module would stop it recurring.

Compile-checking is the minimum, wasip2 happens to be the only such target with real sockets, so it could run the crate's tests rather than just build them.

I am happy to contribute this code, I just want Quinn working on WASM, I think people will build cool stuff with it.

In a little under a month there will be the stable release of <I can't remember exactly which package of the top of my head but you can google it> which is the blocker for `wasm32-wasip3` target going stable. We then have full async too. This is very timely.

temporaryfix:
fix the compile errors, route wasi into the module, make the tests build there, run them in CI

Fixes #2754

I attest that I have read and understand this code.

### DIFF
--- a/quinn-udp/src/wasi.rs
+++ b/quinn-udp/src/wasi.rs
@@ -1,15 +1,15 @@
 use std::{
     io::{self, IoSliceMut},
+    mem::MaybeUninit,
     sync::Mutex,
     time::Instant,
 };
 
 use super::{IO_ERROR_LOG_INTERVAL, RecvMeta, Transmit, UdpSockRef, log_sendmsg_error};
 
-/// Fallback UDP socket interface that stubs out all special functionality
+/// UDP socket interface for WASI
 ///
-/// Used when a better implementation is not available for a particular target, at the cost of
-/// reduced performance compared to that enabled by some target-specific interfaces.
+/// `wasi:sockets` exposes none of this crate's advanced features, so they are stubbed out.
 #[derive(Debug)]
 pub struct UdpSocketState {
     last_send_error: Mutex<Instant>,
@@ -17,7 +17,8 @@ pub struct UdpSocketState {
 
 impl UdpSocketState {
     pub fn new(socket: UdpSockRef<'_>) -> io::Result<Self> {
-        socket.0.set_nonblocking(true)?;
+        socket.set_nonblocking(true)?;
+
         let now = Instant::now();
         Ok(Self {
             last_send_error: Mutex::new(now.checked_sub(2 * IO_ERROR_LOG_INTERVAL).unwrap_or(now)),
@@ -59,14 +60,11 @@ impl UdpSocketState {
         bufs: &mut [IoSliceMut<'_>],
         meta: &mut [RecvMeta],
     ) -> io::Result<usize> {
-        // Safety: both `IoSliceMut` and `MaybeUninitSlice` promise to have the
-        // same layout, that of `iovec`/`WSABUF`. Furthermore `recv_vectored`
-        // promises to not write unitialised bytes to the `bufs` and pass it
-        // directly to the `recvmsg` system call, so this is safe.
-        let bufs = unsafe {
-            &mut *(bufs as *mut [IoSliceMut<'_>] as *mut [socket2::MaybeUninitSlice<'_>])
-        };
-        let (len, _flags, addr) = socket.0.recv_from_vectored(bufs)?;
+        // `wasi:sockets` has no vectored receive, and `BATCH_SIZE` is 1 regardless.
+        // Safety: `MaybeUninit<u8>` shares the layout of `u8`, and `recv_from` writes only
+        // the prefix it reports as filled.
+        let buf = unsafe { &mut *(&mut *bufs[0] as *mut [u8] as *mut [MaybeUninit<u8>]) };
+        let (len, addr) = socket.0.recv_from(buf)?;
         meta[0] = RecvMeta {
             len,
             stride: len,
@@ -74,6 +72,7 @@ impl UdpSocketState {
             ecn: None,
             dst_ip: None,
             interface_index: None,
+            timestamp: None,
         };
         Ok(1)
     }
@@ -122,7 +121,8 @@ fn send(socket: UdpSockRef<'_>, transmit: &Transmit<'_>) -> io::Result<()> {
     socket.0.send_to(
         transmit.contents,
         &socket2::SockAddr::from(transmit.destination),
-    )
+    )?;
+    Ok(())
 }
 
 pub(crate) const BATCH_SIZE: usize = 1;
--- a/quinn-udp/tests/tests.rs
+++ b/quinn-udp/tests/tests.rs
@@ -1,6 +1,11 @@
 #[cfg(apple)]
 use std::io::{self, IoSlice};
-#[cfg(not(any(target_os = "openbsd", target_os = "netbsd", solarish)))]
+#[cfg(not(any(
+    target_os = "openbsd",
+    target_os = "netbsd",
+    target_os = "wasi",
+    solarish
+)))]
 use std::net::{SocketAddr, SocketAddrV6};
 #[cfg(apple)]
 use std::os::fd::AsRawFd;
@@ -12,7 +17,9 @@ use std::{
     slice,
 };
 
-use quinn_udp::{EcnCodepoint, RecvMeta, Transmit, UdpSocketState};
+#[cfg(not(target_os = "wasi"))]
+use quinn_udp::EcnCodepoint;
+use quinn_udp::{RecvMeta, Transmit, UdpSockRef, UdpSocketState};
 #[cfg(apple)]
 use socket2::MsgHdr;
 use socket2::Socket;
@@ -62,7 +69,9 @@ fn basic_src_ip() {
     );
 }
 
+// The WASI backend sets no ECN codepoint and always reports `None`.
 #[test]
+#[cfg(not(target_os = "wasi"))]
 fn ecn_v6() {
     let send = Socket::from(UdpSocket::bind((Ipv6Addr::LOCALHOST, 0)).unwrap());
     let recv = Socket::from(UdpSocket::bind((Ipv6Addr::LOCALHOST, 0)).unwrap());
@@ -82,7 +91,12 @@ fn ecn_v6() {
 }
 
 #[test]
-#[cfg(not(any(target_os = "openbsd", target_os = "netbsd", solarish)))]
+#[cfg(not(any(
+    target_os = "openbsd",
+    target_os = "netbsd",
+    target_os = "wasi",
+    solarish
+)))]
 fn ecn_v4() {
     let send = Socket::from(UdpSocket::bind((Ipv4Addr::LOCALHOST, 0)).unwrap());
     let recv = Socket::from(UdpSocket::bind((Ipv4Addr::LOCALHOST, 0)).unwrap());
@@ -102,7 +116,12 @@ fn ecn_v4() {
 }
 
 #[test]
-#[cfg(not(any(target_os = "openbsd", target_os = "netbsd", solarish)))]
+#[cfg(not(any(
+    target_os = "openbsd",
+    target_os = "netbsd",
+    target_os = "wasi",
+    solarish
+)))]
 fn ecn_v6_dualstack() {
     let recv = Socket::new(
         socket2::Domain::IPV6,
@@ -148,7 +167,12 @@ fn ecn_v6_dualstack() {
 }
 
 #[test]
-#[cfg(not(any(target_os = "openbsd", target_os = "netbsd", solarish)))]
+#[cfg(not(any(
+    target_os = "openbsd",
+    target_os = "netbsd",
+    target_os = "wasi",
+    solarish
+)))]
 fn ecn_v4_mapped_v6() {
     let send = Socket::new(
         socket2::Domain::IPV6,
@@ -295,7 +319,8 @@ fn test_send_recv(send: &Socket, recv: &Socket, transmit: Transmit<'_>) {
     let recv_state = UdpSocketState::new(recv.into()).unwrap();
 
     // Reverse non-blocking flag set by `UdpSocketState` to make the test non-racy
-    recv.set_nonblocking(false).unwrap();
+    UdpSockRef::from(send).set_nonblocking(false).unwrap();
+    UdpSockRef::from(recv).set_nonblocking(false).unwrap();
 
     send_state.try_send(send.into(), &transmit).unwrap();
 
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -148,15 +148,20 @@ jobs:
     steps:
       - uses: actions/checkout@v7
       - uses: dtolnay/rust-toolchain@stable
-      - run: rustup target add wasm32-unknown-unknown
+      - run: rustup target add wasm32-unknown-unknown wasm32-wasip2
       - uses: actions/setup-node@v7
         with:
           node-version: 20
       - uses: bytecodealliance/actions/wasm-tools/setup@v1
+      - uses: bytecodealliance/actions/wasmtime/setup@v1
       - uses: cargo-bins/cargo-binstall@main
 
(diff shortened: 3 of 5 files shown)


quinn-udp/src/fallback.rs
```diff
@@ -1,3 +1,7 @@
+#[cfg(target_os = "wasi")]
+use std::mem::{ManuallyDrop, MaybeUninit};
+#[cfg(target_os = "wasi")]
+use std::os::fd::{AsRawFd, FromRawFd};
```
djc: Nit: these could/should be grouped into a single statement.

djc (commented):
Can you edit/trim back the comments and commit messages to remove superfluous detail/things that are pretty obvious from reading the code?

temporaryfix:
Is this better or do you want me to scrap the comments entirely:

https://github.com/quinn-rs/quinn/compare/733e444c725073be9e485d59a40aee70932cbe54..e3b2b662c1379ba43d6b7acc24dd25cbded7b4f5

djc:
> Is this better or do you want me to scrap the comments entirely:
> 
> https://github.com/quinn-rs/quinn/compare/733e444c725073be9e485d59a40aee70932cbe54..e3b2b662c1379ba43d6b7acc24dd25cbded7b4f5

They're much better. I think the Actions workflow and Cargo.toml comments could be tighter still -- they only need to provide a clue as to why the context is what it is.

temporaryfix:
One of the tests seems to be flakey, investigating now, please don't merge until I have fixed it.

temporaryfix:
The retry is unconditional; it could be wasi-only, and the 10s deadline is arbitrary. Do you have a preference?

> The red job isn't from the comment trim — same wasmtime as the run that passed yesterday, and the only non-comment change was grouping two use lines. The test is just racy here: 2 failures in 40 local runs, always WouldBlock at the recv.
> 
> wasi:sockets has no blocking mode, so the set_blocking call in the test succeeds but does nothing, and UdpSocketState::recv goes straight to socket2's recv_from rather than std's poll emulation. The recv races the send and usually wins on loopback.
> 
> Fix: drop the wasi half of set_blocking (removes an unsafe block) and retry WouldBlock with a deadline — on the send too, since with only the recv fixed it still failed 1 in 100 at try_send. 200/200 after.

- Claude Code v2.1.220 using Opus 5 (1M context) with xhigh effort

Would you like me to squash any of this?

quinn-udp/tests/tests.rs
```diff
@@ -294,24 +317,46 @@ fn test_send_recv(send: &Socket, recv: &Socket, transmit: Transmit<'_>) {
     let send_state = UdpSocketState::new(send.into()).unwrap();
     let recv_state = UdpSocketState::new(recv.into()).unwrap();
 
-    // Reverse non-blocking flag set by `UdpSocketState` to make the test non-racy
+    // Reverse non-blocking flag set by `UdpSocketState` to make the test non-racy.
+    // `wasi:sockets` has no blocking mode, so there the loop below polls instead.
```
thomaseizinger: Perhaps we should have a separate test for wasi then, if it needs different handling?

Or we build a test harness where we use non-blocking mode? That sounds better IMO because it feels like the current blocking mode has been done out of simplicity and that not longer works now.

Either way, the current modifications to the tests make it hard to understand.

quinn-udp/src/fallback.rs
thomaseizinger, line 1: This file is not very big, so duplication isn't much of a concern. Should we introduce a dedicated `wasi` file with its own implementation instead of adding conditionals to `fallback.rs`?

thomaseizinger (commented):
Thank you for extending the platform support. I have two questions.

quinn-udp/tests/tests.rs
temporaryfix: The non-blocking mode sounds like the better solution but is it overkill and is there another good reason for blocking?

quinn-udp/src/fallback.rs
temporaryfix: What would the benefit of doing so be? Does the duplication serve enough of a purpose?

thomaseizinger: We organise the different backends into different files to minimize the conditional compilation instructions within a file. It makes them easier to understand.

It does meantl that we'll need another CI job to actually compile the fallback module I think but a `cargo check` for ab obscure target should suffice for that.

quinn-udp/tests/tests.rs
thomaseizinger: I think simplicity of the test suite was the only reason for the blocking mode.

I don't think it is overkill but the question is whether `#[tokio::test]` works on wasi? Probably not.

Perhaps it is better to duplicate the test for wasi and rewrite it to do its own polling-thing.

Thoughts @djc ?

quinn-udp/src/fallback.rs
temporaryfix: Thanks for explaining. Whilst WASI is kinda obscure now I expect that to be less and less the case. Especially when wasm32-wasip3 goes stable in a little over a month or so.

quinn-udp/tests/tests.rs
temporaryfix: Macros (including #[tokio::test]) should work fine without tokio_unstable but when it comes to WASI stuff sometimes it doesn't work how it is supposed to, especially with wasip2. I will find out and get back to you.

quinn-udp/src/fallback.rs
thomaseizinger: I agree. I wasn't meant to imply that WASI is obsure! But what led us here was `fallback.rs` not being compiled and if WASI gets its own file, we are again in a situation where `fallback.rs` does not get compiled. So we'll need to add another check for something that isn't covered by any of the specific backends :)

temporaryfix:
Pushed the split. One thing I need a decision on before this is done.

Moving WASI into its own file leaves `fallback.rs` compiled by nothing, and I couldn't find a target to check it with. socket2 supports unix, windows and wasi-other-than-p1 only. Of the 33 rustup targets that are none of those, `wasm32-wasip2` is the one that compiles, and it now selects `wasi.rs`.

I checked rather than assumed: with `fallback.rs` replaced by invalid Rust, the native tests, clippy, the `wasm32-unknown-unknown` check and the wasip2 test all still pass. Force-selecting it on unix compiles clean, so it's unreachable rather than broken.

Three ways I can see: drop it, keep WASI in it as before so the wasip2 job keeps compiling it, or force-select it in a check job with a bespoke `--cfg`. I lean to dropping it, but that deletes a file so I've left it in place. Which would you prefer?

temporaryfix:
On tokio — it does work, I shouldn't have doubted it. `rt`, `macros` and `time` all build for wasm and `#[tokio::test]` runs fine under wasmtime. What doesn't build is `net` and `rt-multi-thread`; tokio stops with `Only features sync,macros,io-util,rt,time are supported on wasm`. So the macro is there but the socket isn't, and there's nothing to be async over. That's also why the dev-dependencies here are scoped off wasm — it's `net` that breaks, not tokio itself.

The polling turned out to be unnecessary anyway, and my earlier explanation of the flake was wrong, sorry. `wasi:sockets` does have a working blocking mode. What fails is socket2's `set_nonblocking`, in both directions, because it goes through `fcntl(F_SETFL)` and WASI rejects that — `std` uses `ioctl(FIONBIO)` and works. And it was the send that was racing, not the receive. So the fix is just to reverse the flag through `std` on both sockets: no retry loop, no arbitrary deadline, and the native path calls exactly what it called before. 200/200 on wasip2, where the recv-only version failed 20/200 and 53/200 across two runs.

On wasip3 I don't think anything needs doing here. `std` puts p2 and p3 on the same socket backend with p1 off on its own, socket2's guard is `not(p1)` so it already accepts p3, and 0.3's async lives in the component model rather than `std::net`. No p3 `rust-std` ships yet so I couldn't build it to check — when one does, it should just be another target on that CI line.

thomaseizinger:
Thanks for investigating, in this case, I think dropping `fallback.rs` makes sense. `quinn-udp` already can't function on whatever target that is already anyway.

quinn-udp/tests/tests.rs
```diff
@@ -295,7 +323,10 @@ fn test_send_recv(send: &Socket, recv: &Socket, transmit: Transmit<'_>) {
     let recv_state = UdpSocketState::new(recv.into()).unwrap();
 
     // Reverse non-blocking flag set by `UdpSocketState` to make the test non-racy
-    recv.set_nonblocking(false).unwrap();
+    set_blocking(recv);
+    // WASI can report the send socket as not ready too, so `try_send` needs it blocking.
+    #[cfg(target_os = "wasi")]
```
thomaseizinger: Do the other tests work if we set the send socket to blocking? Perhaps we don't need to make it conditional here.

thomaseizinger (commented):
Very good progress. One question and I think if it is genuinely unreachable, we can delete `fallback.rs` with this.

quinn-udp/tests/tests.rs
temporaryfix: Could you please request any edits you need in one go or when it is something small edit it yourself?

Alternatively getting some AI tooling around here would obviate all of this.

thomaseizinger: Sorry, I am not sure if I am following? I think it is pretty normal that a PR goes through different rounds of review. It is not always possible nor does it make sense to flag everything in one pass. More often than not, the first review will be one that is more of a directional nature and only once that is clear, future review rounds will be more detailed and e.g. point out things like documentation nits or more trivial things like function ordering or other stylistic elements.

thomaseizinger:
> if it is genuinely unreachable, we can delete `fallback.rs` with this.

If we are deleting this, then I think the order of commits should be:

- Split out wasi module
- Adapt tests to cover wasi
- Add CI for wasi
- Delete fallback

quinn-udp/tests/tests.rs
temporaryfix: Lets just get this over the line so we can all move on. I'm making the fixes you have requested. I know not everybody agrees regarding ways of working / AI etc. That's fine. Let me know if you need more changes.

quinn-udp/src/fallback.rs
temporaryfix: Done — `fallback.rs` is deleted in the last commit, so there is no longer a module needing a CI target of its own. `wasi.rs` is the only backend added, and wasip2 builds and runs its tests.

quinn-udp/tests/tests.rs
temporaryfix: They do. Both sockets are set blocking unconditionally now and the `cfg` is gone.

macOS is 8 passed / 1 ignored, identical to before the change; wasip2 is 3 passed / 1 ignored, and 60/60 clean runs of the test binary.

temporaryfix:
Rebuilt in the order you asked for:

- `quinn-udp: support wasm32-wasip2` — `wasi.rs` and the routing
- `quinn-udp: run the test suite on wasm32-wasip2` — tests and the dev-dependency scoping
- `ci: test quinn-udp on wasm32-wasip2`
- `quinn-udp: remove the fallback backend`

The fallback repair commit is dropped, since the file goes away at the end. Both test sockets are set blocking unconditionally, so the `cfg` in `test_send_recv` is gone.

Every commit builds on its own, native and wasip2.

thomaseizinger (approved):
Thank you for working on all the feedback!

This looks good from my end.

djc (commented):
Can we get the first commit to move `fallback` instead of introducing a new file? Makes it easier to review the changes that were made in the process.

temporaryfix:
I switched to Noq anyway so I no longer need this. I won't be jumping through hoops with your nits any more. I asked for all feedback to be given grouped not trickled out like this to respect all of our time. I'll leave it there.

thomaseizinger:
> Makes it easier to review the changes that were made in the process.

The effective diff is actually shown quite well in the overall PR diff: https://github.com/quinn-rs/quinn/pull/2755/changes#diff-aad31c23033f5ed0b577741ec2335f55f63fafacf9ed0e19471264a2684c6e9b

FWIW, the reason I asked for this order is because I think the deletion of the fallback backend is something that becomes possible as a result of introducing the WASI backend and that intent is better described in separate commits IMO.

YMMV though and I don't feel strongly about this particular order.

thomaseizinger:
I can push this over the finish line next week.

djc:
> I switched to Noq anyway so I no longer need this. I won't be jumping through hoops with your nits any more. I asked for all feedback to be given grouped not trickled out like this to respect all of our time. I'll leave it there.

I'm sorry contributing has been a bad experience for you. Unfortunately our review bandwidth has been limited (for example, I have been on vacation for most of the past weeks) and of all the organizations leveraging our project only a single one is helping to sustain maintenance.

> I can push this over the finish line next week.

Thank you!

temporaryfix:
If your review bandwidth is limited, dragging this out over 3 separate PRs feels like an unfortunate use of time.

thomaseizinger:
@djc I squashed the last commit into the first one and re-wrote the commit message. Should be ready now!

quinn-udp/src/wasi.rs
```diff
@@ -122,7 +126,8 @@ fn send(socket: UdpSockRef<'_>, transmit: &Transmit<'_>) -> io::Result<()> {
     socket.0.send_to(
         transmit.contents,
         &socket2::SockAddr::from(transmit.destination),
-    )
+    )?;
+    Ok(())
```
djc, line 125: Why do we need this?

quinn-udp/tests/tests.rs
```diff
@@ -391,6 +420,19 @@ fn test_send_recv(send: &Socket, recv: &Socket, transmit: Transmit<'_>) {
...
+    socket.set_nonblocking(false).unwrap();
+}
+
+#[cfg(target_os = "wasi")]
+fn set_blocking(socket: &Socket) {
+    // socket2 flips the flag with `fcntl`, which WASI rejects; `std` uses `ioctl(FIONBIO)`.
+    // Safety: `socket` outlives the borrow, and `ManuallyDrop` prevents a second close.
+    let borrowed = ManuallyDrop::new(unsafe { UdpSocket::from_raw_fd(socket.as_raw_fd()) });
+    borrowed.set_nonblocking(false).unwrap();
+}
```
djc: Should we extract these for use in `UdpSocketState::new()` (in the previous commit)?

djc (approved):
Thanks, LGTM modulo some minor stuff.

quinn-udp/src/wasi.rs
thomaseizinger: Do you mean the explicit `Ok(())`? It just a style question. Personally, I prefer an explicit `Ok` at the end of functions that can fail and `?` at every call-site that can return an error. With a `Result` expression as the final call, it is not obvious that there is another fallible call here unless you remember / check the function signature.

Don't feel strongly though if you'd rather keep the diff small.

djc: Okay, seems fine.

djc (approved):

temporaryfix:
Thankyou for merging this, it is appreciated. I have a few other PRs that I will be submitting. Hopefully we can learn to work well together without all this needless friction. I hope you enjoyed your vacation.

PR #2763, opened 2026-07-29
udp: preserve socket-level DSCP in the per-packet ECN cmsg

### LINKED ISSUE #1749: Suggestions for quinn-udp

Baughn:
While working on a different application (github.com/baughn/dessplay) I found a need to use per-stream DSCP tagging. Since Quinn has no support (#1749), I attempted to work around this by creating the socket up-front, setting DSCP, then handing it off. This also did not work, because quinn-udp clears any DSCP bits while attempting to set the ECN bit.

Since there's no intent to handle DSCP internally, this PR sets the ECN setting by read-modify-write so other bits survive unmodified. Tested on OSX + Linux; I do not have a BSD machine of any kind, which makes me suspicious of the BSD-specific code, but I expect the worst it can do is recreate the original bug.

Patch and regression tests coded by Fable, reviewed by me. And because it insists on being helpful:

> On Unix platforms every datagram carries an IP_TOS / IPV6_TCLASS cmsg whose value is derived solely from the transmit's ECN codepoint. A per-packet cmsg overrides the socket-level option, so any DSCP marking an application configured with setsockopt before handing the socket to quinn is silently rewritten to 0 on the wire. Windows is unaffected: its IP_ECN / IPV6_ECN options touch only the ECN bits.
>
> Capture the socket's TOS byte (ECN bits masked off) once at UdpSocketState creation and OR it into every outgoing TOS cmsg, so the DSCP bits survive while ECN continues to work exactly as before. The capture is best-effort: on an unmarked socket the base is 0 and the cmsg value is unchanged from today.
>
> Closes #1749 for the setsockopt use case without adding any API surface: applications mark the socket before constructing the endpoint, as they already must for other socket options.

### DIFF
--- a/quinn-udp/src/apple_fast.rs
+++ b/quinn-udp/src/apple_fast.rs
@@ -57,7 +57,7 @@ fn send_via_sendmsg_x(
             &mut iovs[i],
             &mut ctrls[i],
             true,
-            state.sendmsg_einval(),
+            state,
         );
         hdrs[i].msg_datalen = chunk.len();
         state.check_send_buffer_limit(chunk.len(), &hdrs[i])?;
@@ -78,7 +78,7 @@ fn prepare_msg_x(
     iov: &mut libc::iovec,
     ctrl: &mut cmsg::Aligned<[u8; cmsg::LEN]>,
     #[allow(unused_variables)] encode_src_ip: bool,
-    sendmsg_einval: bool,
+    state: &UdpSocketState,
 ) {
     iov.iov_base = transmit.contents.as_ptr() as *const _ as *mut _;
     iov.iov_len = transmit.contents.len();
@@ -93,15 +93,15 @@ fn prepare_msg_x(
     hdr.msg_control = ctrl.0.as_mut_ptr() as _;
     hdr.msg_controllen = cmsg::LEN as _;
     let mut encoder = unsafe { cmsg::Encoder::new(hdr) };
-    let ecn = transmit.ecn.map_or(0, |x| x as libc::c_int);
+    let tos = state.tos_base() as libc::c_int | transmit.ecn.map_or(0, |x| x as libc::c_int);
     let is_ipv4 = transmit.destination.is_ipv4()
         || matches!(transmit.destination.ip(), IpAddr::V6(addr) if addr.to_ipv4_mapped().is_some());
     if is_ipv4 {
-        if !sendmsg_einval {
-            encoder.push(libc::IPPROTO_IP, libc::IP_TOS, ecn as IpTosTy);
+        if !state.sendmsg_einval() {
+            encoder.push(libc::IPPROTO_IP, libc::IP_TOS, tos as IpTosTy);
         }
     } else {
-        encoder.push(libc::IPPROTO_IPV6, libc::IPV6_TCLASS, ecn);
+        encoder.push(libc::IPPROTO_IPV6, libc::IPV6_TCLASS, tos);
     }
 
     if let Some(ip) = &transmit.src_ip {
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -42,6 +42,14 @@ pub struct UdpSocketState {
     /// which is not supported on Linux <3.13 and results in not sending the UDP packet at all.
     sendmsg_einval: AtomicBool,
 
+    /// The socket-level TOS / traffic-class byte at creation, ECN bits masked off.
+    ///
+    /// The per-packet `IP_TOS` / `IPV6_TCLASS` cmsg overrides the socket-level
+    /// option, so a cmsg valued solely from the ECN codepoint would zero any
+    /// DSCP marking the application configured via `setsockopt` before handing
+    /// the socket over. ORing this base into the cmsg preserves it.
+    tos_base: u8,
+
     /// Whether to use Apple's fast `sendmsg_x`/`recvmsg_x` APIs.
     ///
     /// These private APIs provide better performance but may not be available on all
@@ -83,6 +91,8 @@ impl UdpSocketState {
 
         let addr = io.local_addr()?;
         let is_ipv4 = addr.family() == libc::AF_INET as libc::sa_family_t;
+
+        let tos_base = socket_tos_base(&*io, is_ipv4);
 
         // mac and ios do not support IP_RECVTOS on dual-stack sockets :(
         // older macos versions also don't have the flag and will error out if we don't ignore it
@@ -217,6 +227,7 @@ impl UdpSocketState {
             gro_segments,
             may_fragment,
             sendmsg_einval: AtomicBool::new(false),
+            tos_base,
             #[cfg(apple_fast)]
             apple_fast_path: AtomicBool::new(false),
             #[cfg(apple)]
@@ -400,6 +411,11 @@ impl UdpSocketState {
         self.sendmsg_einval.load(Ordering::Relaxed)
     }
 
+    /// Returns the socket-level TOS / traffic-class byte captured at creation.
+    pub(crate) fn tos_base(&self) -> u8 {
+        self.tos_base
+    }
+
     /// Sets the flag indicating we got EINVAL error from `sendmsg` syscall.
     #[cfg(not(any(apple, target_os = "openbsd", target_os = "netbsd")))]
     fn set_sendmsg_einval(&self) {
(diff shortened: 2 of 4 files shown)


Baughn:
...furthermore, I _absolutely_ do not have illumos, solaris or android handy to test on. Would you like me to write exclusions regardless?

thomaseizinger:
> ...furthermore, I _absolutely_ do not have illumos, solaris or android handy to test on. Would you like me to write exclusions regardless?

That is what we have CI for :)

quinn-udp/src/apple_fast.rs
```diff
@@ -93,15 +93,15 @@ fn prepare_msg_x(
     hdr.msg_control = ctrl.0.as_mut_ptr() as _;
     hdr.msg_controllen = cmsg::LEN as _;
     let mut encoder = unsafe { cmsg::Encoder::new(hdr) };
-    let ecn = transmit.ecn.map_or(0, |x| x as libc::c_int);
+    let tos = state.tos_base() as libc::c_int | transmit.ecn.map_or(0, |x| x as libc::c_int);
```
thomaseizinger, line 96: What if we define a new-type here and move the encoding of the bits onto that type? That would remove the duplication between the different paths and would allow us to independently unit test this, plus you get some compile-time guarantees that the value is not just discarded.

Baughn:
> > ...furthermore, I _absolutely_ do not have illumos, solaris or android handy to test on. Would you like me to write exclusions regardless?
> 
> That is what we have CI for :)

Not as useful as you'd hope, when the commit involves editing the tests. It _looks_ right, I guess? The logic makes sense? But I remain unfamiliar with android.

quinn-udp/src/unix.rs
```diff
@@ -967,6 +983,43 @@ fn set_socket_option_supported(
...
+        true => (libc::IPPROTO_IP, libc::IP_TOS),
+        #[cfg(not(target_os = "redox"))]
+        false => (libc::IPPROTO_IPV6, libc::IPV6_TCLASS),
+        // Redox lacks `IPV6_TCLASS` (and no TCLASS cmsg is sent there)
+        #[cfg(target_os = "redox")]
+        false => return 0,
+    };
+    let mut val = [0u8; size_of::<libc::c_int>()];
+    let mut len = val.len() as libc::socklen_t;
+    let rc = unsafe {
```
thomaseizinger, line 1003: Nit: I'd appreciate a newline here.

quinn-udp/src/unix.rs
```diff
@@ -967,6 +983,43 @@ fn set_socket_option_supported(
...
+        libc::getsockopt(
+            socket.as_raw_fd(),
+            level,
+            opt,
+            val.as_mut_ptr() as *mut _,
+            &mut len,
+        )
+    };
+    if rc != 0 {
```
thomaseizinger, line 1012: And here.

quinn-udp/src/unix.rs
```diff
@@ -967,6 +983,43 @@ fn set_socket_option_supported(
...
+        return 0;
+    }
+    // Some BSD-derived systems yield a single byte, others a full `c_int`
```
thomaseizinger, line 1015: And here.

quinn-udp/src/unix.rs
```diff
@@ -967,6 +983,43 @@ fn set_socket_option_supported(
...
+    let tos = match len {
+        1 => val[0],
+        _ => libc::c_int::from_ne_bytes(val) as u8,
+    };
+    tos & !0b11
```
thomaseizinger, line 1020: And here too.

quinn-udp/tests/tests.rs
```diff
@@ -716,3 +716,182 @@ fn recv_transport_error_ipv6() {
         assert_eq!(addr.ip(), dst.ip());
     }
 }
+
+/// The DSCP marking under test: AF41 (RFC 2597)
+#[cfg(unix)]
+const DSCP: u8 = 34;
```
thomaseizinger, line 722: Should this const move into the test?

quinn-udp/tests/tests.rs
```diff
@@ -716,3 +716,182 @@ fn recv_transport_error_ipv6() {
...
+/// The `IP_TOS` / `IPV6_TCLASS` cmsg attached to every datagram overrides the
+/// socket-level option, so a cmsg valued solely from the ECN codepoint would
+/// rewrite the wire TOS byte to `ecn` — zeroing DSCP markings the application
+/// configured before handing the socket over. (Windows is unaffected: its
+/// `IP_ECN` / `IPV6_ECN` options touch only the ECN bits.)
+///
+/// Sends one datagram over loopback through `UdpSocketState` on a DSCP-tagged
+/// socket and receives it with a raw `recvmsg` requesting the TOS byte.
+#[cfg(unix)]
+fn test_dscp_preserved(ipv4: bool) {
```
thomaseizinger, line 748: I am not a huge fan of the boolean argument here, what if we pass the loopback address instead?

quinn-udp/tests/tests.rs
```diff
@@ -716,3 +716,182 @@ fn recv_transport_error_ipv6() {
         assert_eq!(addr.ip(), dst.ip());
     }
 }
+
+/// The DSCP marking under test: AF41 (RFC 2597)
+#[cfg(unix)]
+const DSCP: u8 = 34;
+
+#[test]
+#[cfg(unix)]
```
thomaseizinger, line 725: I might be worth writing these tests in such a way that they also assert the behaviour on Windows, even if we don't have to do anything there to support it.

thomaseizinger (commented):
I can't speak much for the actual feature addition itself but I left some comments that I'd like to see implement if this is accepted.

I'll defer the actual ACK to @djc and @Ralith .

quinn-udp/src/apple_fast.rs
Baughn: Makes total sense, but it took me ages to find the thirty minutes to do this. Anyhow, how's this?

I might be focusing a bit too much on my specific problem. My _preferred_ solution would be to add DSCP support as a first-class option, which is what was already rejected, so I'm unsure which of the alternatives is ideal.

quinn-udp/src/unix.rs
Baughn: Around the unsafe block, I assume you mean.

quinn-udp/src/unix.rs
Baughn: Done

quinn-udp/src/unix.rs
Baughn: Done

quinn-udp/tests/tests.rs
Baughn: I don't think that's viable. https://stackoverflow.com/questions/59747990/java-13-socketoption-ip-tos-settrafficclass suggests Windows ignores DSCP flags set through IP_TOS.

I don't have a Windows machine, so can't test this myself. I _could_ try to implement it anyway and lean on the presubmit checks, but it'd be blind coding.

quinn-udp/tests/tests.rs
Baughn: Done.

quinn-udp/tests/tests.rs
Baughn: Done.

Baughn:
Further testing found that Linux at least tracks separate DSCP settings for separate address families, and an IP6 socket can connect to an IP4 destination. That would have been a highly inconvenient bug to track down.

Socket clients are supposed to set _both_. I added some code to _preserve_ both, and documented it, but it's still a fragile interface IMO.

quinn-udp/src/apple_fast.rs
djc: > My _preferred_ solution would be to add DSCP support as a first-class option, which is what was already rejected, so I'm unsure which of the alternatives is ideal.

Where, in your understanding, did that get rejected? Things may have moved since then.

Baughn: In https://github.com/quinn-rs/quinn/issues/1749, mentioned in the PR description.

PR #2764, opened 2026-07-30
docs(quinn): document cancellation safety for accept_uni, accept_bi and read_datagram

### LINKED ISSUE #2247: Document cancellation safety of `read_datagram`, `accept_bi`, and `accept_uni`
I was looking at a server being able to read/support any connection type and was thinking about putting `accept_bi`, `accept_uni`, and `read_datagram` in a `tokio::select` but I noticed none of those methods have docs regarding whether they're cancellation safe.

cnsky1103:
Closes #2247 

For these 3 methods, if they consume a stream or datagram (in poll_accept or ReadDatagram::poll), they will definitely return Poll::Ready in this poll; if a Poll::Pending is returned, no intermediate states are stored so it's safe to drop it. As a result they are cancellation safe.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -319,6 +319,10 @@ impl Connection {
     }
 
     /// Accept the next incoming uni-directional stream
+    ///
+    /// # Cancel safety
+    ///
+    /// This method is cancellation safe. If this does not resolve, no streams were consumed.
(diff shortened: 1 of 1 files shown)


djc (approved):

Ralith (approved):
Thanks!

Issue #2704, opened 2026-06-25
IP_RECVERR can prevent graceful connection termination


stormshield-fabs:
The recently added support of `IP_RECVERR` (https://github.com/quinn-rs/quinn/pull/2646) has an unintended impact on the normal operation of the socket: when an error is received on the socket, the next syscall (`recvmmsg` or `sendmsg` for `quinn`) will ignore the data and return this error (once for each enqueued error).

This is described (although quite discreetly I think) in the [`recv(2)` manpage](https://man7.org/linux/man-pages/man2/recv.2.html#:~:text=MSG_ERRQUEUE):

> After an error has been passed, the pending socket error is regenerated based on the next queued error and will be passed on the next socket operation.

In practice, this does not have a huge functional impact since reads and writes are retried in `quinn`.

However, we stumbled upon this when chasing a lost `CONNECTION_CLOSE` frame that should have been sent when closing the `Connection`. It turned out the packet containing this frame was absorbed by [`sendmsg`](https://github.com/quinn-rs/quinn/blob/fed0321a9a672819662caab37f5662f1ad91308e/quinn-udp/src/unix.rs#L453), since an ICMP error had occured right before it was called. This error is only meant to be a notification that an error was enqueued and is not related to the packet we're trying to sent (the syscall returns an error even if the error was for a different remote address). Because close notifications are [not retried](https://github.com/quinn-rs/quinn/blob/fed0321a9a672819662caab37f5662f1ad91308e/quinn-proto/src/connection/mod.rs#L491), the remote peer never gets it.

I believe this was understood in a previous implementation, since this comment hints at a refactor from a design that pulled errors immediately to the external getter design : https://github.com/quinn-rs/quinn/pull/2421/changes#r2450585268

To move forward, we would like to suggest adding an option to let users choose if the `IP_RECVERR` flag is passed on the socket. 

Additionally, we think both `recvmmsg` and `sendmsg` calls should be wrapped to capture such errors and emit a notification that the user could receive to know if they should call the new `recv_transport_error` method. Then, the syscall should be retried (and that would fix the silently dropped `CONNECTION_CLOSE`).

djc:
@thomaseizinger @mxinden thoughts?

syszery:
I was able to reproduce the `sendmsg` absorption locally with a minimal test: send to a closed port, then send to a valid receiver without draining the queue -> the second packet never arrives.

[...]

PR #2765, opened 2026-07-30
feat(udp): make transport error reception opt-in on Linux/Android

syszery:
Addresses #2704

#2646 enabled `IP_RECVERR`/`IPV6_RECVERR` unconditionally on Linux and Android. This has two side effects discussed in #2704:
* queued errors can cause subsequent syscalls (e.g. `recvmmsg` or `sendmsg`) to return the error instead of completing the operation;
* enabling this feature has a non-zero cost since queued errors use the same receive buffer as incoming packets.

This PR implements the proposed resolution from #2704 and makes transport error reception opt-in via a new `enable_transport_errors` method on `UdpSocketState`. Users that need asynchronous transport errors can enable the feature explicitly and retrieve them through the existing `recv_transport_error`.

When enabled, callers are responsible for draining pending transport errors before sending or receiving, as queued errors may otherwise be returned by subsequent socket operations.

Existing transport error tests are updated to enable the feature explicitly, and a regression test verifies the documented usage pattern.

### DIFF
--- a/quinn-udp/src/unix.rs
+++ b/quinn-udp/src/unix.rs
@@ -146,21 +146,6 @@ impl UdpSocketState {
             {
                 crate::log::debug!("Ignoring error setting SO_TIMESTAMPNS on socket: {_err:?}");
             }
-
-            if is_ipv4 || !io.only_v6()? {
-                if let Err(_err) =
-                    set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_RECVERR, OPTION_ON)
-                {
-                    crate::log::debug!("ignoring error setting IP_RECVERR on socket: {_err:?}");
-                }
-            }
-            if !is_ipv4 {
-                if let Err(_err) =
-                    set_socket_option(&*io, libc::IPPROTO_IPV6, libc::IPV6_RECVERR, OPTION_ON)
-                {
-                    crate::log::debug!("ignoring error setting IPV6_RECVERR on socket: {_err:?}");
-                }
-            }
         }
         #[cfg(any(target_os = "freebsd", apple))]
         {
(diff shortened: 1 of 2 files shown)


quinn-udp/src/unix.rs
```diff
@@ -303,13 +288,52 @@ impl UdpSocketState {
...
+    ///
+    /// When this feature is enabled, the kernel may return a pending error on the
+    /// next `sendmsg` or `recvmsg` call, preventing that operation from completing.
+    /// To avoid this, callers should drain the error queue via
+    /// [`recv_transport_error`] before sending. Failing to do so can prevent packets
+    /// from being transmitted, including frames such as `CONNECTION_CLOSE` that are
+    /// not retried by the QUIC implementation.
+    ///
+    /// [`recv_transport_error`]: Self::recv_transport_error
+    #[cfg(any(target_os = "linux", target_os = "android"))]
```
thomaseizinger: I think so far the consensus has been to expose the same public API across all platforms so that users of `quinn-udp` don't have to use conditional compilation to call its APIs.

quinn-udp/src/unix.rs
```diff
@@ -303,13 +288,52 @@ impl UdpSocketState {
...
+    /// dropped by the kernel. There is no overhead when the error queue is empty.
+    ///
+    /// # Usage
+    ///
+    /// When this feature is enabled, the kernel may return a pending error on the
+    /// next `sendmsg` or `recvmsg` call, preventing that operation from completing.
+    /// To avoid this, callers should drain the error queue via
+    /// [`recv_transport_error`] before sending. Failing to do so can prevent packets
+    /// from being transmitted, including frames such as `CONNECTION_CLOSE` that are
+    /// not retried by the QUIC implementation.
```
thomaseizinger: I think this is a rather specific case that we don't need to mention here? We are already documenting that the next send operation may fail. That is enough IMO.

quinn-udp/src/unix.rs
```diff
@@ -303,13 +288,52 @@ impl UdpSocketState {
...
+        let io = socket.0;
+        let addr = io.local_addr()?;
+        let is_ipv4 = addr.family() == libc::AF_INET as libc::sa_family_t;
+        if is_ipv4 || !io.only_v6()? {
+            set_socket_option(&*io, libc::IPPROTO_IP, libc::IP_RECVERR, OPTION_ON)?;
+        }
+        if !is_ipv4 {
+            set_socket_option(&*io, libc::IPPROTO_IPV6, libc::IPV6_RECVERR, OPTION_ON)?;
+        }
+        Ok(())
```
thomaseizinger, line 329: Nit: would prefer a newline above this.

quinn-udp/src/unix.rs
```diff
@@ -303,13 +288,52 @@ impl UdpSocketState {
...
+    /// from being transmitted, including frames such as `CONNECTION_CLOSE` that are
+    /// not retried by the QUIC implementation.
+    ///
+    /// [`recv_transport_error`]: Self::recv_transport_error
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    pub fn enable_transport_errors(&self, socket: UdpSockRef<'_>) -> io::Result<()> {
+        let io = socket.0;
+        let addr = io.local_addr()?;
+        let is_ipv4 = addr.family() == libc::AF_INET as libc::sa_family_t;
+        if is_ipv4 || !io.only_v6()? {
```
thomaseizinger: Nit: Would prefer a newline above both of the `if`s too.

thomaseizinger (commented):
Thanks for pushing this forward! Left some comments.

quinn-udp/src/unix.rs
syszery: Updated the method to mirror the style of `recv_transport_error`.

quinn-udp/src/unix.rs
syszery: I removed the obsolete sentence and updated the documentation to reflect the change above.

djc:
(Squash the review feedback back into your original commit.)

thomaseizinger (approved):

quinn-udp/src/unix.rs
```diff
@@ -303,13 +288,55 @@ impl UdpSocketState {
         recv_single(socket.0, bufs, meta)
     }
 
-    /// Receives a pending, asynchronous transport-layer error from this socket
+    /// Enables asynchronous transport-layer error reception for this socket.
```
djc: Documentation comment nits: please start each documentation (including the test function) comment with a sentence (fragment) that fits on a single line, and doesn't end in a ".".

djc (approved):
Thanks!

quinn-udp/src/unix.rs
syszery: I've updated the docs as requested.

djc:
Thanks!

PR #2771, opened 2026-08-04
Reducing EndpointDriver state lock contention (part 1): preparations

stablebits:
The first 3 commits of https://github.com/quinn-rs/quinn/pull/2633, submitted separately as requested in [this review comment](https://github.com/quinn-rs/quinn/pull/2633#discussion_r3691218602).

Preparatory refactors with no functional changes.

2 more parts follow: further behavior-preserving refactors around accept, then the actual split of accept around the endpoint lock.


quinn-udp/src/unix.rs
```diff
@@ -865,7 +865,7 @@ impl ControlMetadata {
             (libc::IPPROTO_IPV6, libc::IPV6_PKTINFO) => {
                 let pktinfo = unsafe { cmsg::decode::<libc::in6_pktinfo, libc::cmsghdr>(cmsg) };
                 self.dst_ip = Some(IpAddr::V6(Ipv6Addr::from(pktinfo.ipi6_addr.s6_addr)));
-                #[cfg_attr(not(target_os = "android"), expect(clippy::unnecessary_cast))]
```
thomaseizinger, line 868: If possible, we should stick to `expect` so the compiler can flag when it is no longer needed.

stablebits: My bad, this commit isn't necessary at all. It (partially) undid https://github.com/quinn-rs/quinn/commit/7f7d9ea90137d701c859bbf86adbf51536d55941. My local build was using a pre-1.95 rust :-/ I discarded this commit.

djc (approved):

alessandrod:
can we merge pls? we have quite a few more patches coming based on this split

Ralith (approved):

PR #2773, opened 2026-08-05
Congestion window should be allowed to be usable in full

bastienvty:
RFC 9002 only forbids sending a packet that would grow `bytes_in_flight` beyond the window, so a sender that stops one datagram short whenever the window is an exact multiple of the datagram size (as the default initial window is) is too conservative.

[RFC 9002 §7](https://datatracker.ietf.org/doc/html/rfc9002#section-7-7):

> An endpoint MUST NOT send a packet if it would cause bytes_in_flight (see Appendix B.2) to be larger than the congestion window, unless the packet is sent on a PTO timer expiration (see Section 6.2) or when entering recovery (see Section 7.3.2).


Ralith (approved):
Thanks!

djc (approved):

PR #2776, opened 2026-08-08
fix(quinn-udp): repair fallback backend compile errors

teddytennant:
## Summary

The `fallback` UDP backend (`cfg(not(any(wasm_browser, unix, windows)))`) no longer
compiled: `send_to` returns `usize`, `RecvMeta` gained a `timestamp` field, and
`recv_from_vectored` is gone from `socket2::Socket`.

- Map `send_to` into `io::Result<()>`
- Fill `timestamp: None` on `RecvMeta`
- Receive with `recv_from` over a `MaybeUninit<u8>` view of the first buffer

Verified by temporarily selecting the fallback module on this host and
`cargo check -p quinn-udp`.

Related to #2754 (orthogonal to #2755, which removes the backend for wasi).

### DIFF
--- a/quinn-udp/src/fallback.rs
+++ b/quinn-udp/src/fallback.rs
@@ -1,5 +1,6 @@
 use std::{
     io::{self, IoSliceMut},
+    mem::MaybeUninit,
     sync::Mutex,
     time::Instant,
 };
@@ -61,10 +62,21 @@ impl UdpSocketState {
         if bufs.is_empty() || meta.is_empty() {
             return Ok(0);
         }
-        // socket2::recv_from requires &mut [MaybeUninit<u8>]; IoSliceMut is [u8].
+
+        // `recv_from_vectored` is not available on every target that uses this backend: socket2
+        // gates it out on redox, wasi and horizon, since they have no `recvmsg`. We only ever
+        // receive one datagram at a time here (`BATCH_SIZE` is 1), so read into the first buffer
+        // with plain `recv_from` instead, which is available everywhere.
+        //
+        // Safety: `recv_from` takes `&mut [MaybeUninit<u8>]`, and `MaybeUninit<u8>` has the same
+        // size and alignment as `u8`, so the pointer and length taken from `bufs[0]` describe a
+        // valid slice of the same region. Treating initialised memory as possibly-uninitialised is
+        // always sound, and `recv_from` promises not to write uninitialised bytes, so `bufs[0]`
+        // stays fully initialised. The mutable borrow of `bufs` lasts for the whole call, so
+        // nothing else can access the region while `buf` is alive.
         let buf = unsafe {
             std::slice::from_raw_parts_mut(
-                bufs[0].as_mut_ptr() as *mut std::mem::MaybeUninit<u8>,
+                bufs[0].as_mut_ptr().cast::<MaybeUninit<u8>>(),
                 bufs[0].len(),
             )
         };
@@ -72,9 +84,7 @@ impl UdpSocketState {
         meta[0] = RecvMeta {
             len,
             stride: len,
-            addr: addr
-                .as_socket()
-                .unwrap_or_else(|| std::net::SocketAddr::from(([0, 0, 0, 0], 0))),
+            addr: addr.as_socket().unwrap(),
             ecn: None,
             dst_ip: None,
             interface_index: None,
--- a/quinn-udp/src/fallback.rs
+++ b/quinn-udp/src/fallback.rs
@@ -58,21 +58,27 @@ impl UdpSocketState {
         bufs: &mut [IoSliceMut<'_>],
         meta: &mut [RecvMeta],
     ) -> io::Result<usize> {
-        // Safety: both `IoSliceMut` and `MaybeUninitSlice` promise to have the
-        // same layout, that of `iovec`/`WSABUF`. Furthermore `recv_vectored`
-        // promises to not write unitialised bytes to the `bufs` and pass it
-        // directly to the `recvmsg` system call, so this is safe.
-        let bufs = unsafe {
-            &mut *(bufs as *mut [IoSliceMut<'_>] as *mut [socket2::MaybeUninitSlice<'_>])
+        if bufs.is_empty() || meta.is_empty() {
+            return Ok(0);
+        }
+        // socket2::recv_from requires &mut [MaybeUninit<u8>]; IoSliceMut is [u8].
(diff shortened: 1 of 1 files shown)


quinn-udp/src/fallback.rs
```diff
@@ -58,21 +58,27 @@ impl UdpSocketState {
...
-        // Safety: both `IoSliceMut` and `MaybeUninitSlice` promise to have the
-        // same layout, that of `iovec`/`WSABUF`. Furthermore `recv_vectored`
-        // promises to not write unitialised bytes to the `bufs` and pass it
-        // directly to the `recvmsg` system call, so this is safe.
-        let bufs = unsafe {
-            &mut *(bufs as *mut [IoSliceMut<'_>] as *mut [socket2::MaybeUninitSlice<'_>])
+        if bufs.is_empty() || meta.is_empty() {
+            return Ok(0);
+        }
+        // socket2::recv_from requires &mut [MaybeUninit<u8>]; IoSliceMut is [u8].
```
Ralith: Please include safety comments on `unsafe` blocks documenting how the unsafe invariants are satisfied, similar to the one you removed.

quinn-udp/src/fallback.rs
```diff
@@ -58,21 +58,27 @@ impl UdpSocketState {
...
+        let buf = unsafe {
+            std::slice::from_raw_parts_mut(
+                bufs[0].as_mut_ptr() as *mut std::mem::MaybeUninit<u8>,
```
Ralith: style nit: prefer `.cast::<*mut std::mem::MaybeUninit<u8>>()`

quinn-udp/src/fallback.rs
```diff
@@ -58,21 +58,27 @@ impl UdpSocketState {
...
         };
-        let (len, _flags, addr) = socket.0.recv_from_vectored(bufs)?;
+        let (len, addr) = socket.0.recv_from(buf)?;
         meta[0] = RecvMeta {
             len,
             stride: len,
-            addr: addr.as_socket().unwrap(),
+            addr: addr
+                .as_socket()
+                .unwrap_or_else(|| std::net::SocketAddr::from(([0, 0, 0, 0], 0))),
```
Ralith: What's the motivation for this change? Why is this particular default value correct? When will it be used?

quinn-udp/src/fallback.rs
```diff
@@ -58,21 +58,27 @@ impl UdpSocketState {
...
+        }
+        // socket2::recv_from requires &mut [MaybeUninit<u8>]; IoSliceMut is [u8].
+        let buf = unsafe {
+            std::slice::from_raw_parts_mut(
+                bufs[0].as_mut_ptr() as *mut std::mem::MaybeUninit<u8>,
+                bufs[0].len(),
+            )
         };
-        let (len, _flags, addr) = socket.0.recv_from_vectored(bufs)?;
+        let (len, addr) = socket.0.recv_from(buf)?;
```
Ralith, line 83: Document why we're not using the vectored version: because it's unavailable on some platforms (not outright gone).

Ralith (changes requested):

quinn-udp/src/fallback.rs
djc: (Just import `MaybeUninit` directly at the top of the module.)

quinn-udp/src/fallback.rs
teddytennant: Added one back, covering the layout equivalence of MaybeUninit<u8> and u8, why going from init to maybe-uninit is fine, and that recv_from does not write uninitialised bytes.

quinn-udp/src/fallback.rs
teddytennant: Done, MaybeUninit is imported at the top now and the cast is `.cast::<MaybeUninit<u8>>()`.

quinn-udp/src/fallback.rs
teddytennant: No good motivation, I was just being twitchy about the unwrap. It is never actually used, recv_from on a UDP socket always gives back an AF_INET or AF_INET6 address, so as_socket is always Some. And if it somehow were not, handing quinn 0.0.0.0:0 as the peer address is worse than panicking. Reverted to unwrap(), which is also what the windows backend does.

quinn-udp/src/fallback.rs
teddytennant: You are right, it is not gone. socket2 cfg-gates it off redox, wasi and horizon since there is no recvmsg there, and those are exactly targets this backend covers. Noted that in a comment.

Ralith (approved):
Thanks!

thomaseizinger:
@Ralith Would you mind reviewing #2755 so we can decide which one of these PRs we should merge? I don't think it makes sense to merge this and #2755 as it will just create merge conflicts in a file that is going to get deleted.

djc:
> @Ralith Would you mind reviewing #2755 so we can decide which one of these PRs we should merge? I don't think it makes sense to merge this and #2755 as it will just create merge conflicts in a file that is going to get deleted.

Do you have a preference?

thomaseizinger:
> Do you have a preference?

I'd prefer to merge #2755 instead because it brings support for WASI and the deletion of the fallback backend in there is essentially just a fallout because we realised that there is no other target that it would actually be active on. Hence, I think it is the better fix.

temporaryfix:
I cannot believe this many of us have sat around for this long discussing this whilst the code on the repo is actively broken, the fix is so simple and sat there ready to merge.

There is clearly a need for it too as someone else now has offered a solution.

Why? I believe it is because I used AI to write a PR description in https://github.com/quinn-rs/quinn/pull/2753 and that has upset one of the product owners. This is not a good reason to not fix broken code and this is not the way to treat people trying to help you fix your code.

thomaseizinger:
> Why? I believe it is because I used AI to write a PR description in #2753 and that has upset one of the product owners.

As far as I know, nobody is upset here. We are simply all volunteers when it comes to working on this codebase and respectful co-maintainership of it also means that we need to discuss and agree as to how we want to fix things. That sometimes takes a while. The easiest interim solution is to depend on a fork in our own projects.

djc:
Going to close this in favor of the other PR. Thanks for contributing to our project!

teddytennant:
No worries. If you have any issues that need fixes or want me to help with anything else just mention me on an issue or PR and I would be happy to help

PR #2779, opened 2026-08-10
chore(udp): deprecate `UdpSocketState::send`

thomaseizinger:
In preparation for https://github.com/quinn-rs/quinn/pull/2748, this PR deprecates the `send` API and encourages users to use `try_send` instead.

https://github.com/quinn-rs/quinn/pull/2748 is a breaking change by itself. I'd therefore suggest we make a patch-release that deprecates `send` and then remove it as part of https://github.com/quinn-rs/quinn/pull/2748.

### DIFF
--- a/quinn-udp/src/fallback.rs
+++ b/quinn-udp/src/fallback.rs
@@ -35,6 +35,7 @@ impl UdpSocketState {
     ///
     /// If you would like to handle these errors yourself, use [`UdpSocketState::try_send`]
     /// instead.
+    #[deprecated(note = "silences I/O errors; use `UdpSocketState::try_send() instead")]
     pub fn send(&self, socket: UdpSockRef<'_>, transmit: &Transmit<'_>) -> io::Result<()> {
         match send(socket, transmit) {
             Ok(()) => Ok(()),
--- a/quinn/src/runtime/mod.rs
+++ b/quinn/src/runtime/mod.rs
@@ -7,6 +7,7 @@ use std::{
     net::SocketAddr,
     pin::Pin,
     task::{self, Context, Poll},
+    time::Duration,
 };
 
 use udp::{RecvMeta, Transmit};
@@ -118,6 +119,7 @@ pin_project_lite::pin_project! {
     /// used in its dyn-compatible form as a `Pin<Box<dyn UdpSender>>`.
     struct UdpSenderHelper<Socket, MakeWritableFutFn, WritableFut> {
         socket: Socket,
+        last_send_error: Option<Instant>,
         make_writable_fut_fn: MakeWritableFutFn,
         #[pin]
         writable_fut: Option<WritableFut>,
@@ -145,6 +147,7 @@ impl<Socket, MakeWritableFutFn, WriteableFut>
     fn new(inner: Socket, make_fut: MakeWritableFutFn) -> Self {
         Self {
             socket: inner,
+            last_send_error: None,
             make_writable_fut_fn: make_fut,
             writable_fut: None,
         }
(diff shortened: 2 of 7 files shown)


quinn/src/runtime/mod.rs
```diff
@@ -145,6 +147,7 @@ impl<Socket, MakeWritableFutFn, WriteableFut>
     fn new(inner: Socket, make_fut: MakeWritableFutFn) -> Self {
         Self {
             socket: inner,
+            last_send_error: None,
```
thomaseizinger, line 150: As documented in the commit message, this presents a slight change in behaviour. We are now rate-limiting these errors per `UdpSenderHelper` instead of per socket, meaning users may see these errors more often if multiple `UdpSenderHelper`s share the same socket.

I think this is an acceptable change given that it allows us to remove a `Mutex` from the hot-path of sending packets.

quinn-udp/src/fallback.rs
```diff
@@ -35,6 +35,7 @@ impl UdpSocketState {
     ///
     /// If you would like to handle these errors yourself, use [`UdpSocketState::try_send`]
     /// instead.
+    #[deprecated(note = "Silences IO errors. Use `UdpSocketState::try_send` instead.")]
```
djc: Bunch of nits, I'd spell the note as:

"silences I/O errors; use `UdpSocketState::try_send() instead"

quinn/src/runtime/mod.rs
```diff
@@ -199,6 +205,39 @@ where
...
+    let now = Instant::now();
+    if last_send_error
+        .is_some_and(|last| now.saturating_duration_since(last) <= IO_ERROR_LOG_INTERVAL)
+    {
+        return;
+    }
+    *last_send_error = Some(now);
+
+    tracing::warn!(
+        "sendmsg error: {:?}, Transmit: {{ destination: {:?}, src_ip: {:?}, ecn: {:?}, len: {:?}, segment_size: {:?} }}",
```
djc, line 231: Use structured logging here instead of inlining all this stuff into the message?

djc (approved):
Looks directionally correct to me, but would like to get a review from @Ralith as well.

mxinden (approved):
quinn-udp changes look good to me.

quinn/src/runtime/mod.rs
thomaseizinger: Right now, this is exactly what we do in `quinn-udp`: https://github.com/quinn-rs/quinn/blob/d3fe39198e5540db93f75755f9229a50feb06132/quinn-udp/src/lib.rs#L276-L284

Would you like me to change both then?

djc: Let's keep it for this PR and fix both in another one? If you could take a look at other tracing events, that would be great too.

thomaseizinger: Structured logging is tricky to support here because we have different logging backends (`log` and `tracing`) and they use different notations for embedding fields. We'd have to build a small parser to normalise this based on which backend is active.

I prototyped this in https://github.com/thomaseizinger/quinn/tree/structured-tracing-events. What do you think?

djc: Uhhh, I forget why we support both? IMO we should only support tracing going forward, probably.

Ralith: Motivation was described by @mxinden in https://github.com/quinn-rs/quinn/pull/1921#issue-2415703512.

djc: Okay, then I think doing structured logging is probably not worth it.

PR #2780, opened 2026-08-11
Mark quinn_proto::Event as non-exhaustive

stormshield-fabs:
As suggested in https://github.com/quinn-rs/quinn/pull/2700#discussion_r3759273482


djc (approved):

djc:
Thanks!

PR #2781, opened 2026-08-12
Update AI policy to recognise assistive technology use (Closes #2756)

### LINKED ISSUE #2756: AI is often used as an assistive tool by disabled people, the current AI policy doesn't take that into account.
There is guidance given for non-native English speakers which is helpful:

> AI is useful when communicating as a non-native English speaker. If you are using AI to edit your comments for this purpose, please take the time to ensure it reflects your own voice and ideas. When using AI for translation, we recommend writing in your native language and including the AI translation in a quote block.

However no guidance is given for those using AI as an assistive aid e.g. due to disabilities such as memory loss, RSI or a myriad of other issues. Unfortunately that means that it is not currently an inclusive policy, likely quite inadvertently.

temporaryfix:
The previous AI policy required comments to be "written by humans" and allowed hiding anything that looked AI-generated. This created a barrier for disabled contributors who use AI as assistive technology (e.g. due to dyslexia, RSI, aphasia, memory issues, or similar).

This change:

- Allows AI (and other assistive tools) for writing comments and descriptions
- Explicitly states that disability-related assistive use is permitted
- Removes the rule that comments may be hidden simply for appearing AI-assisted
- Keeps the ability to hide clear low-effort spam or content the author cannot discuss
- Clarifies that ordinary assistive use does not need to be disclosed

The goal remains the same: contributors must understand and take responsibility for *what* they post. The process of *how* the text is produced is no longer policed in a way that excludes disabled people.

Closes #2756

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,25 +21,25 @@ Using AI (LLMs) as tools for coding is welcome. A high bar is held for all contr
 project. Moreover, the project maintainers remain responsible for any code that is published as
 part of a release. Contributors are expected to be responsible for any code they publish.
 
-AI should not be used to generate comments when communicating with maintainers. Comments are
-expected to be written by humans. Comments that are believed to be written by AI may be hidden
-without notice.
+Comments and issue/PR descriptions should reflect your own understanding of the topic. You are
+responsible for everything you post and should be able to discuss and defend it.
+
+Using AI (or any other assistive technology) to help write is allowed. This includes use by
+disabled people for whom AI is an accessibility tool. We will not hide comments simply because
+they appear to have been written with AI assistance.
+
+We may still hide comments that are clearly low-effort spam, bulk-generated noise, or that the
+author cannot meaningfully discuss when asked.
 
 If you are opening an issue, you should be able to describe the problem in your own words.
 
 If you are opening a pull request, you are expected to be able to explain the proposed changes in
 your own words. This includes the pull request body and responses to questions. Make sure you have
-reviewed the PR yourself before submitting it for review to the maintainers. Do not copy responses
-from the AI when replying to questions from maintainers.
+reviewed the PR yourself before submitting it for review to the maintainers.
 
-If you wish to include context from an interaction with AI in your comments, it must be in a
-quote block (using `>`) and disclosed as such. It must be accompanied by human commentary
-explaining the relevance and implications of the context. Do not share long snippets.
-
-AI is useful when communicating as a non-native English speaker. If you are using AI to edit your
-comments for this purpose, please take the time to ensure it reflects your own voice and ideas.
-When using AI for translation, we recommend writing in your native language and including the AI
-translation in a quote block.
+If you want to include raw AI output as supporting context (for example, to show what a model said
+about a technical point), put it in a quote block and say that it came from AI. You do not need to
+disclose ordinary use of AI as an assistive writing tool.
 
 ## Code changes
 


CONTRIBUTING.md
```diff
@@ -21,25 +21,25 @@ Using AI (LLMs) as tools for coding is welcome. A high bar is held for all contr
...
 part of a release. Contributors are expected to be responsible for any code they publish.
 
-AI should not be used to generate comments when communicating with maintainers. Comments are
-expected to be written by humans. Comments that are believed to be written by AI may be hidden
-without notice.
+Comments and issue/PR descriptions should reflect your own understanding of the topic. You are
+responsible for everything you post and should be able to discuss and defend it.
+
+Using AI (or any other assistive technology) to help write is allowed. This includes use by
+disabled people for whom AI is an accessibility tool. We will not hide comments simply because
```
djc, line 28: I don't think I want to promise this here, as there are way too many people using AI to simply move burden from themselves to the maintainers.

I'd rather take a smaller patch that takes the existing exception for non-native English speakers and explicitly expands it to assistive usage.

temporaryfix: understand the concern about people using AI to offload work onto maintainers.

The reason I went with the broader change is that the “smaller patch” approach (just expanding the non-native speaker exception) leaves the core problem in place.

As long as the policy still says comments are expected to be “written by humans” and that anything that looks AI-generated can be hidden, disabled people who rely on AI as assistive technology remain at risk of having their comments removed or dismissed on appearance alone.

Making disability a special-case exception next to the language one still treats our use as an outlier that needs justifying, rather than a normal and legitimate way of writing.

The version in this PR keeps the important guardrails:

You must understand and be able to discuss what you post
Clear spam / bulk-generated noise can still be hidden
People who just paste AI output and can’t engage can still be shut down

It just stops treating the method of writing as the thing being policed. That’s the part that was creating the accessibility barrier.

Happy to adjust wording if there’s a way to address the burden-shifting concern without reintroducing the “must look human” test, but I don’t think bolting disability onto the existing exception fully solves it.

djc: I understand what you're trying to do. As an example of something this does not cover, LLMs like to enumerate all the things they "Verified" in a PR description. This is usually extremely low value because all of those things are also checked by CI. Just saying "must understand and be able to discuss" does not address this category of verbosity/slop.

I don't care that it is "written by humans" per se, but I have become somewhat allergic to the verbose, elaborate text generated by LLMs.

temporaryfix: Ok so how do we modify the text to preclude that without bringing back the ableism? I understand the exact AI tell you are talking about. We could also potentially look at including instructions for agents to try and avoid that style of AI prose from our end.

djc: This is where it feels like it's not unreasonable to say that everyone who is able to do so without undue burden should express themselves directly -- how does that come off to you?

temporaryfix: I wouldn't know how to define undue burden personally, I still think the answer is to focus on the outcome not on the process. Ultimately you say yourself you don't care if it is human written per se.

djc: Just a note -- want to continue this conversation but I'm going on vacation for a while; will pick this back up after.

temporaryfix: It feels like you might be using "--" and other AI tells / tropes in a mocking way and dragging your feet on this. I'm probably mistaken because I'm sure you wouldn't be publicly mocking disabled people and their use of assistive technologies.

djc: All of my comments here are 100% written by hand. I am of a generation that actually uses "--", I guess.

PR #2775, opened 2026-08-07
Reducing EndpointDriver state lock contention (part 2): behavior-preserving refactors

stablebits:
Part 2 of the series split out of https://github.com/quinn-rs/quinn/pull/2633.

No functional changes: consolidates route registration into register_connection and reshapes Endpoint::accept so the split around the endpoint lock (part 3) falls along clean seams. Details in the individual commit messages.

PR #2782, opened 2026-08-12
Reducing EndpointDriver state lock contention (part 3): core changes

stablebits:
The changes are on top of https://github.com/quinn-rs/quinn/pull/2775.

Accept API currently hold the endpoint lock across TLS session setup, Connection::new(), and Connection::handle_first_packet(). Under a high rate of new connections, that stalls the endpoint driver whenever it encounters a new Initial, which in turn cuts throughput for already-established traffic.

This change splits server-side accept into three phases:

  1. Reserve endpoint and routing state under the lock.
  2. Perform TLS session creation and first-packet handling without the lock.
  3. Reacquire the lock to activate the connection or clean up after failure.

The transient state is made explicit in quinn-proto with an Accepting connection state. Packets that arrive during the split window are routed to Accepting and buffered until finalization, so retransmitted Initials and 0-RTT packets are preserved.

  ### Performance testing

Note that before this change, the endpoint lock implicitly serialized Incoming::accept(). With this change in place, applications have to resolve the desired trade-off between established-connection throughput and new-connection acceptance by controlling handshake concurrency.
Without application-level admission control, more handshakes can run concurrently and compete with established connections for CPU. Applications accepting many connections concurrently should therefore bound handshake concurrency according to their workload. This can be done either by using a dedicated semaphore or tokio runtime for accept work.

In a synthetic test with 20K new connections per second offered, explicit admission control allowed approximately 13.8K connections per second to be accepted, compared with approximately 4.3K/s before this change, while preserving the same upload throughput.

In other words, this change provided roughly three times the new-connection capacity without reducing established connection throughput when concurrency was appropriately bounded.

The benchmark used ECDSA-P256 handshakes, 16 connection-generating clients, one continuously uploading client, MAX_STREAMS=64, an eight-thread server runtime, and a 10-second measurement window on a Linux development machine.

In our production environment with many concurrent uploads and ~2K new connection attempts per second, the aggragate throughput of the concurrent uploads is about 50% higher with this change.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -543,11 +543,29 @@ impl Endpoint {
     // box err to avoid clippy::result_large_err
     pub fn accept(
         &mut self,
-        mut incoming: Incoming,
+        incoming: Incoming,
         now: Instant,
         buf: &mut Vec<u8>,
         server_config: Option<Arc<ServerConfig>>,
     ) -> Result<(ConnectionHandle, Connection), Box<AcceptError>> {
+        let accepting = self.start_accept(incoming, now, buf, server_config)?;
+        match accepting.finish_without_endpoint() {
+            Ok(accepted) => Ok(self.finish_accept(accepted)),
+            Err(error) => Err(self.finish_accept_error(error, buf)),
+        }
+    }
+
+    /// Reserve CIDs and routing state before constructing a connection.
+    ///
+    /// The returned state must be completed without endpoint access and then passed back to
+    /// this endpoint through `finish_accept` or `finish_accept_error`.
+    fn start_accept(
+        &mut self,
+        mut incoming: Incoming,
+        now: Instant,
+        buf: &mut Vec<u8>,
+        server_config: Option<Arc<ServerConfig>>,
+    ) -> Result<Accepting, Box<AcceptError>> {
         let remote_address_validated = incoming.remote_address_validated();
 
         let packet_number = incoming.packet.header.number.expand(0);
@@ -614,7 +632,8 @@ impl Endpoint {
             }));
         };
 
-        let loc_cid = self.new_cid(RouteDatagramTo::Incoming(incoming.incoming_idx));
+        let accepting_idx = incoming.incoming_idx;
+        let loc_cid = self.new_cid(RouteDatagramTo::Incoming(accepting_idx));
         let mut params = TransportParameters::new(
             &server_config.transport,
             &self.config,
@@ -628,7 +647,7 @@ impl Endpoint {
         params.retry_src_cid = incoming.token.retry_src_cid;
         let mut pref_addr_cid = None;
         if server_config.has_preferred_address() {
-            let cid = self.new_cid(RouteDatagramTo::Incoming(incoming.incoming_idx));
+            let cid = self.new_cid(RouteDatagramTo::Incoming(accepting_idx));
             pref_addr_cid = Some(cid);
             params.preferred_address = Some(PreferredAddress {
                 address_v4: server_config.preferred_address_v4,
@@ -638,83 +657,92 @@ impl Endpoint {
             });
         }
 
-        incoming.improper_drop_warner.dismiss();
-
-        let tls = server_config.crypto.clone().start_session(version, &params);
-        let transport_config = server_config.transport.clone();
+        // Gather everything needed to create the Connection outside the lock.
         let mut rng_seed = [0; 32];
         self.rng.fill_bytes(&mut rng_seed);
-        let mut conn = Connection::new(
-            self.config.clone(),
-            transport_config,
-            dst_cid,
+        let endpoint_config = self.config.clone();
+        let cid_len = self.local_cid_generator.cid_len();
+        let cid_lifetime = self.local_cid_generator.cid_lifetime();
+        let allow_mtud = self.allow_mtud;
+
+        let reservation = AcceptReservation {
+            incoming_idx: accepting_idx,
+            init_cid: dst_cid,
+            addresses: incoming.addresses,
             loc_cid,
-            src_cid,
-            incoming.addresses.remote,
-            incoming.addresses.local_ip,
-            tls,
-            self.local_cid_generator.cid_len(),
-            self.local_cid_generator.cid_lifetime(),
-            incoming.received_at,
+            pref_addr_cid,
+        };
+
+        Ok(Accepting {
+            reservation,
             version,
-            self.allow_mtud,
-            rng_seed,
-            SideArgs::Server {
-                server_config,
-                pref_addr_cid,
-                path_validated: remote_address_validated,
-            },
-        );
-
-        match conn.handle_first_packet(
-            incoming.received_at,
-            incoming.addresses.remote,
-            incoming.ecn,
+            src_cid,
             packet_number,
-            incoming.packet,
-            incoming.rest,
-        ) {
-            Ok(()) => {
-                let incoming_buffer = self.remove_incoming_buffer(incoming.incoming_idx);
-                let ch = ConnectionHandle(self.connections.vacant_key());
-                self.register_connection(
-                    ch,
-                    dst_cid,
-                    loc_cid,
-                    pref_addr_cid,
-                    incoming.addresses,
-                    Side::Server,
-                );
-                trace!(id = ch.0, icid = %dst_cid, "new connection");
+            incoming,
+            // Deferred connection creation state
+            server_config,
+            params,
+            remote_address_validated,
+            rng_seed,
+            endpoint_config,
+            cid_len,
+            cid_lifetime,
+            allow_mtud,
+        })
+    }
 
-                for event in incoming_buffer.datagrams {
-                    conn.handle_event(ConnectionEvent(ConnectionEventInner::Datagram(event)))
-                }
+    fn finish_accept(&mut self, accepted: Accepted) -> (ConnectionHandle, Connection) {
+        let Accepted {
+            reservation,
+            mut conn,
+        } = accepted;
+        let accepting_buffer = self.remove_accept_reservation(&reservation);
+        let ch = ConnectionHandle(self.connections.vacant_key());
+        self.register_connection(
+            ch,
+            reservation.init_cid,
+            reservation.loc_cid,
+            reservation.pref_addr_cid,
+            reservation.addresses,
+            Side::Server,
+        );
+        trace!(id = ch.0, icid = %reservation.init_cid, "new connection");
 
-                Ok((ch, conn))
-            }
-            Err(e) => {
-                debug!("handshake failed: {}", e);
-                let response = match &e {
-                    ConnectionError::TransportError(e) => Some(self.initial_close(
-                        version,
-                        incoming.addresses,
-                        &incoming.crypto,
-                        src_cid,
-                        e.clone(),
-                        buf,
-                    )),
-                    _ => None,
-                };
-                self.index.remove_initial(dst_cid);
-                self.index.retire(loc_cid);
-                if let Some(cid) = pref_addr_cid {
-                    self.index.retire(cid);
-                }
-                self.remove_incoming_buffer(incoming.incoming_idx);
-                Err(Box::new(AcceptError { cause: e, response }))
-            }
+        for event in accepting_buffer.datagrams {
+            conn.handle_event(ConnectionEvent(ConnectionEventInner::Datagram(event)))
         }
+
+        (ch, conn)
+    }
+
+    /// Clean up after a failed [`Accepting::finish_without_endpoint`] and optionally generate a
+    /// close response.
+    fn finish_accept_error(
+        &mut self,
+        error: Box<AcceptingError>,
+        buf: &mut Vec<u8>,
+    ) -> Box<AcceptError> {
+        let AcceptingError {
+            cause,
+            reservation,
+            version,
+            src_cid,
+            crypto,
+        } = *error;
+        debug!("handshake failed: {}", cause);
+        let response = match &cause {
+            ConnectionError::TransportError(e) => Some(self.initial_close(
+                version,
+                reservation.addresses,
+                &crypto,
+                src_cid,
+                e.clone(),
+                buf,
+            )),
+            _ => None,
+        };
+        self.remove_accept_reservation(&reservation);
+        Box::new(AcceptError { cause, response })
     }
 
     /// Check if we should refuse a connection attempt regardless of the packet's contents
@@ -823,6 +851,15 @@ impl Endpoint {
     fn remove_incoming_state(&mut self, incoming: &Incoming) {
         self.index.remove_initial(incoming.packet.header.dst_cid);
         self.remove_incoming_buffer(incoming.incoming_idx);
+    }
+
+    fn remove_accept_reservation(&mut self, reservation: &AcceptReservation) -> IncomingBuffer {
+        self.index.remove_initial(reservation.init_cid);
+        self.index.retire(reservation.loc_cid);
+        if let Some(cid) = reservation.pref_addr_cid {
+            self.index.retire(cid);
+        }
+        self.remove_incoming_buffer(reservation.incoming_idx)
     }
 
     fn remove_incoming_buffer(&mut self, incoming_idx: usize) -> IncomingBuffer {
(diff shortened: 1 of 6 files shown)


stablebits:
Doing additional refactoring... converted to Draft. ETA: 1-2 days.

stablebits:
> Doing additional refactoring... converted to Draft. ETA: 1-2 days.

Ready; @djc @Ralith please review when you have time

quinn-proto/src/endpoint.rs
```diff
@@ -543,11 +543,29 @@ impl Endpoint {
...
+        let accepting = self.start_accept(incoming, now, buf, server_config)?;
+        match accepting.finish_without_endpoint() {
+            Ok(accepted) => Ok(self.finish_accept(accepted)),
+            Err(error) => Err(self.finish_accept_error(error, buf)),
+        }
+    }
+
+    /// Reserve CIDs and routing state before constructing a connection.
+    ///
+    /// The returned state must be completed without endpoint access and then passed back to
```
Ralith, line 560: nit: "must be completed without endpoint access" implies that the caller reasonably could attempt to complete it *with* endpoint access otherwise. Maybe just reference the method that must be called on it?

quinn-proto/src/endpoint.rs
```diff
@@ -543,11 +543,29 @@ impl Endpoint {
...
     pub fn accept(
         &mut self,
-        mut incoming: Incoming,
+        incoming: Incoming,
         now: Instant,
         buf: &mut Vec<u8>,
         server_config: Option<Arc<ServerConfig>>,
     ) -> Result<(ConnectionHandle, Connection), Box<AcceptError>> {
+        let accepting = self.start_accept(incoming, now, buf, server_config)?;
+        match accepting.finish_without_endpoint() {
```
Ralith: Is it useful for the `Ok`/`Err` distinction to be part of the API surface, or would it be simpler to bundle both cases inside `Accepted` and combine the two `finish_accept` methods together? It seems like every caller will need to match like this, so we might as well push the match down inside the abstraction and have simpler API types.

quinn-proto/src/endpoint.rs
```diff
@@ -614,7 +632,8 @@ impl Endpoint {
             }));
         };
 
-        let loc_cid = self.new_cid(RouteDatagramTo::Incoming(incoming.incoming_idx));
+        let accepting_idx = incoming.incoming_idx;
```
Ralith, line 635: nit: not sure why we're referring to this as `accepting_idx` rather than `incoming_idx`, as used even in the other new code?

quinn-proto/src/endpoint.rs
```diff
@@ -1338,6 +1376,107 @@ pub struct AcceptError {
...
+
+impl Accepting {
+    /// Complete computationally expensive connection setup steps without holding the endpoint lock.
+    ///
+    /// Creates the `Connection` and processes the first packet.
+    /// None of this requires `&mut Endpoint`.
+    ///
+    /// On success, returns the connection plus the reservation that still needs to be activated
+    /// under the endpoint lock.
+    fn finish_without_endpoint(self) -> Result<Accepted, Box<AcceptingError>> {
```
Ralith, line 1418: bikedshed: This name is verbose and not all that helpful. Maybe `accept()`? That matches the return type, and reduces the number of different operations named "finish". Or maybe `eval()`, or `process()`?

Ralith, line 1418: Is it important for the `Ok`/`Err` distinction to be part of the API surface, or would it be simpler to bundle both cases inside `Accepted` and combine the two `finish_accept` methods together into a single one that itself returns `Result`? It seems like every caller will need to match like this, so we might as well push the match down inside the abstraction and have simpler API types.

quinn-proto/src/endpoint.rs
```diff
@@ -1338,6 +1376,107 @@ pub struct AcceptError {
...
+        self.incoming.improper_drop_warner.dismiss();
+
+        let transport_config = self.server_config.transport.clone();
+        let tls = self
+            .server_config
+            .crypto
+            .clone()
+            .start_session(self.version, &self.params);
+        let mut conn = Connection::new(
```
Ralith, line 1427: I think this call and the `Accepting` type would both be much clearer if we introduced (in a prior commit) something like a `ConnectionOptions` struct that `Connection::new` takes, which `Accepting` can just own directly, rather than having to duplicate much of `Connection::new`'s signature.

quinn-proto/src/endpoint.rs
```diff
@@ -559,7 +559,8 @@ impl Endpoint {
     ///
     /// The returned state must be completed without endpoint access and then passed back to
     /// this endpoint through `finish_accept` or `finish_accept_error`.
-    fn start_accept(
+    #[doc(hidden)]
```
Ralith, line 562: Why hide these? You seem to have clearly established that they're worth using, and I don't see why we should restrict that to the `quinn` layer.

quinn-proto/src/endpoint.rs
```diff
@@ -673,6 +675,7 @@ impl Endpoint {
             loc_cid,
             pref_addr_cid,
         };
+        self.pending_accepts += 1;
```
Ralith, line 678: This seems like a counter that could be maintained just as easily by the caller. The commit description indicates it's for the caller's benefit; would it be simpler to move it there outright? Compare `EndpointStats`.

Ralith (changes requested):
Thanks, overall this looks great! Some notes on style and API design, but the shape makes sense and logic looks correct.

I'm not sure I entirely follow how this was able to speed up your workload. What other threads were contending with the endpoint lock?

Regardless, this is well aligned with our long term goal of parallelizing receive, so very happy to have the architectural improvement.

stablebits:
> Thanks, overall this looks great! Some notes on style and API design, but the shape makes sense and logic looks correct.
> 
> I'm not sure I entirely follow how this was able to speed up your workload. What other threads were contending with the endpoint lock?
> 

Thanks for the review!

The `EndpointDriver`'s `poll()` (quinn/src/endpoint.rs) takes the same `EndpointInner::State` lock. It drives IO on an endpoint, routing incoming datagrams to connections that own them or `Initial`s to the app code that accepts them. So while the state lock is being held inside `Incoming::accept()` (in the app-specific tokio task), the endpoint driver can't route datagrams for existing connections... so their throughput may go down.

`Endpoint::accept()`, which runs under the lock, can take some time. For example, here is the old example that I posted a while ago:
https://github.com/quinn-rs/quinn/pull/2633#issuecomment-4379994561

Another potentially heavy part is 
https://github.com/quinn-rs/quinn/blob/main/quinn-proto/src/connection/streams/state.rs#L194
`Endpoint::accept() -> Connection::new() -> StreamState::new()` for connections with very larger MAX_STREAMS (say, 10K, in our experimental code).

Ralith:
Oh, duh, that makes sense. Thanks!

quinn-proto/src/endpoint.rs
stablebits: Rewritten; it now describes the 3-phase accept approach and mentiones explicitly that `Accepting::accept` and `finish_accept` must be called next

quinn-proto/src/endpoint.rs
stablebits: Removed this alias; `self.new_cid()` now accept `incoming.incoming_idx` directly.

quinn-proto/src/endpoint.rs
stablebits: It's now `Accepting::accept()`, where `Accepting` is returned by `start_accept()`.

stablebits: This is a better approach indeed. So, `Accepted` now holds `Result<Connection, ConnectionError>`. `finish_accept()` returns the same Result as `Endpoint::accept()`. Deleted `AcceptingError` and `finish_accept_error()`.

quinn-proto/src/endpoint.rs
stablebits: `ConnectionOptions` are introduced in the 3rd commit now. The call is `Connection::new(crypto, options)`, and `Accepting` has the `options` field.

quinn-proto/src/endpoint.rs
stablebits: removed

quinn-proto/src/endpoint.rs
stablebits: It's part of `State` now: read by `State::is_idle()` and incremented/decremented by `EndpointInner::accept()`. Everything is local to quinn/src/endpoint.rs. Much better indeed!

Ralith (approved):
LGTM, thanks!

quinn-proto/src/connection/mod.rs
```diff
@@ -3956,6 +3956,30 @@ impl From<SideArgs> for ConnectionSide {
     }
 }
 
+/// Parameters to `Connection::new` other than the TLS session
+///
+/// The TLS session is passed separately because creating it is comparatively expensive, and the
+/// endpoint gathers these parameters before that work is done.
+pub(crate) struct ConnectionOptions {
```
djc: Nit: don't love the `Options` suffix (these really aren't "options"), maybe `Input` or `Seed` or something else?

quinn-proto/src/endpoint.rs
```diff
@@ -1299,6 +1377,30 @@ impl Drop for IncomingImproperDropWarner {
...
+        mem::forget(self);
+    }
+}
+
+impl Drop for AcceptDropGuard {
+    fn drop(&mut self) {
+        warn!(
+            "quinn_proto::Accepting or Accepted dropped before reaching Endpoint::finish_accept \
+             (leaks connection IDs and buffered datagrams, and may cause eventual inability to \
+             accept new connections)"
```
djc, line 1394: I don't think this needs to be this long? Suggest we keep it to one line.

quinn-proto/src/endpoint.rs
```diff
@@ -540,14 +540,46 @@ impl Endpoint {
     }
 
     /// Attempt to accept this incoming connection (an error may still occur)
+    ///
+    /// Equivalent to [`start_accept`](Self::start_accept) followed by [`Accepting::accept`] and
```
djc: Nit: suggest using a `()` suffix in the link text when linking to functions. (Also for other documentation comments in this commit.)

quinn/src/endpoint.rs
```diff
@@ -511,6 +549,11 @@ pub(crate) struct State {
     runtime: Arc<dyn Runtime>,
     stats: EndpointStats,
     default_client_config: Option<ClientConfig>,
+    /// Accepts that have reserved state in `inner` but are still constructing their connection
```
djc: Nit: use a sentence that fits on one line for the start of a documentation comment, please.

quinn-proto/src/endpoint.rs
```diff
@@ -849,21 +850,23 @@ impl Endpoint {
...
         let pref_addr_cid = side_args.pref_addr_cid();
         let conn = Connection::new(
-            self.config.clone(),
-            transport_config,
-            init_cid,
-            loc_cid,
-            rem_cid,
-            addresses.remote,
-            addresses.local_ip,
             tls,
```
djc, line 930: (With rustls maintainer hat on)

"The TLS session stays separate because creating it is the expensive
part of connection setup" -- what about this is expensive? Just the crypto operations?

djc (approved):
Thanks, this is good stuff!

And sorry for all the delays in getting this merged.

quinn-proto/src/endpoint.rs
stablebits: The comment is a bit misleading (I'll fix it). I didn't mean `ServerConfig::start_session()` (as in "the TLS _session_") but rather any crypto/TLS work that was done under the lock as part of `Incoming::accept()`, which is basically ClientHello/ServerHello processing invoked by `handle_first_packet()`.

I don't have that profile anymore, but you can see "crypto" and "tls" calls on top of `handle_first_packet()` here:
https://github.com/quinn-rs/quinn/pull/2633#issuecomment-4379994561

Below is a quick (AI-assisted) local simulation. The overhead of `Connection::new()` scales with `MAX_STREAMS`. At 10K for both bidi and uni here (we use uni in prod), it far outweighs the crypto overhead for P-256.

```
Setup. Apple M4 Max, cargo test --release, rustls 0.23 with the ring provider, X25519 key exchange, no client auth. Medians over 300 iterations, two runs agreeing within a few percent.

Per-accept cost by phase. The lock-held parts after the split are tiny and constant. Everything that varies with the cert is in phase 2.

┌──────────────────────────────────────────┬──────────────────────┬───────┬──────────────────────────────────┬───────┬─────────┬──────────┐
│                   Step                   │ Under lock after PR? │ P-256 │ P-256 + 10k bidi/10k uni streams │ P-384 │ Ed25519 │ RSA-2048 │
├──────────────────────────────────────────┼──────────────────────┼───────┼──────────────────────────────────┼───────┼─────────┼──────────┤
│ Endpoint::handle of the Initial (driver) │ yes                  │ 1.7us │ 1.7us                            │ 1.8us │ 1.7us   │ 1.7us    │
├──────────────────────────────────────────┼──────────────────────┼───────┼──────────────────────────────────┼───────┼─────────┼──────────┤
│ Phase 1 start_accept                     │ yes                  │ 0.6us │ 0.6us                            │ 0.6us │ 0.5us   │ 0.6us    │
├──────────────────────────────────────────┼──────────────────────┼───────┼──────────────────────────────────┼───────┼─────────┼──────────┤
│ start_session                            │ no                   │ 0.6us │ 0.6us                            │ 0.6us │ 0.6us   │ 0.6us    │
├──────────────────────────────────────────┼──────────────────────┼───────┼──────────────────────────────────┼───────┼─────────┼──────────┤
│ Connection::new                          │ no                   │ 3.8us │ 215us                            │ 4us   │ 3.8us   │ 3.8us    │
├──────────────────────────────────────────┼──────────────────────┼───────┼──────────────────────────────────┼───────┼─────────┼──────────┤
│ handle_first_packet                      │ no                   │ 53us  │ 62us                             │ 253us │ 50us    │ 350us    │
├──────────────────────────────────────────┼──────────────────────┼───────┼──────────────────────────────────┼───────┼─────────┼──────────┤
│ Phase 2 total (moved off the lock)       │ no                   │ 58us  │ 280us                            │ 258us │ 55us    │ 355us    │
├──────────────────────────────────────────┼──────────────────────┼───────┼──────────────────────────────────┼───────┼─────────┼──────────┤
│ Phase 3 finish_accept                    │ yes                  │ 0.7us │ 0.7us                            │ 0.7us │ 0.6us   │ 0.7us    │
└──────────────────────────────────────────┴──────────────────────┴───────┴──────────────────────────────────┴───────┴─────────┴──────────┘

So the split removes over 98% of the per-accept lock hold time. What remains under the lock is about 1.3us in the accept call plus 1.7us in the driver for the Initial.

Where phase 2 goes. Direct rustls without quinn: ServerConnection::new is 0.1us, read_hs on the ClientHello is 41us for P-256 and 342us for RSA-2048, and draining the flight with write_hs is 4.5us. quinn's own framing inside handle_first_packet adds roughly 7 to 10us on top.
```

quinn-proto/src/connection/mod.rs
stablebits: How about `ConnectionArgs`? similar to the existing `SideArgs` below it?

djc: Yup, makes sense!

quinn-proto/src/endpoint.rs
Ralith: I think it's widely understood that cryptographic handshake processing is the most costly part of any incoming connection setup.

> The overhead of Connection::new() scales with MAX_STREAMS. At 10K for both bidi and uni here (we use uni in prod), it far outweighs the crypto overhead for P-256.

I think this is resolved by https://github.com/quinn-rs/quinn/pull/2601, currently waiting on @djc's review.

quinn-proto/src/endpoint.rs
stablebits: simplified:
```
/// Warns when an [`Accepting`] or [`Accepted`] is dropped before [`Endpoint::finish_accept()`]
```

quinn/src/endpoint.rs
stablebits: changed as follows:
```
    /// Connections in the process of being accepted
```

PR #2787, opened 2026-08-17
Send CONNECTION_CLOSE even when congestion blocked

### LINKED ISSUE #2785: CONNECTION_CLOSE is never sent when close() is called while congestion-blocked; peer hangs until idle timeout
A connection that is closed while its congestion window is saturated never sends CONNECTION_CLOSE. The peer receives nothing and only finds out via its own idle timeout.

Observed on quinn 0.11.11 / quinn-proto 0.11.16; verified still present on master (148d760d8845dc6315d554a39a1f1f83a4134bfe) with the repro test below.

## Mechanism

1. `Connection::close()` → `close_inner()` calls `close_common()`, which stops **all** timers (including loss detection), then re-arms only `Timer::Close` ([connection/mod.rs#L1251-L1259](https://github.com/quinn-rs/quinn/blob/148d760d8845dc6315d554a39a1f1f83a4134bfe/quinn-proto/src/connection/mod.rs#L1251-L1259), [#L3494-L3499](https://github.com/quinn-rs/quinn/blob/148d760d8845dc6315d554a39a1f1f83a4134bfe/quinn-proto/src/connection/mod.rs#L3494-L3499)).

2. The CONNECTION_CLOSE frame can only be emitted from `poll_transmit`'s per-space loop. That loop computes `ack_eliciting` from `pending` — which still contains the stream retransmit data queued before the close — and then congestion-gates the packet: `ack_eliciting && loss_probes == 0 && in_flight.bytes + bytes_to_send > congestion.window()` → skip the space ([#L595-L616](https://github.com/quinn-rs/quinn/blob/148d760d8845dc6315d554a39a1f1f83a4134bfe/quinn-proto/src/connection/mod.rs#L595-L616)). The packet that would actually be built contains **only ACKs + CONNECTION_CLOSE** ([#L786-L848](https://github.com/quinn-rs/quinn/blob/148d760d8845dc6315d554a39a1f1f83a4134bfe/quinn-proto/src/connection/mod.rs#L786-L848)) — neither is ack-eliciting nor counts as in-flight — but the gate classifies it by the pending frames a closing connection will never send.

3. Nothing can ever unblock the gate:
   - In `State::Closed`, incoming packets are scanned only for `Frame::Close`; **ACK frames are not processed**, so `in_flight` never drains ([#L2475-L2497](https://github.com/quinn-rs/quinn/blob/148d760d8845dc6315d554a39a1f1f83a4134bfe/quinn-proto/src/connection/mod.rs#L2475-L2497)).
   - Loss-detection timers were stopped in step 1, so nothing is declared lost and no loss probes are queued — the `loss_probes != 0` bypass never fires.
   - Each incoming peer packet re-arms `self.close = true` ([#L2451-L2454](https://github.com/quinn-rs/quinn/blob/148d760d8845dc6315d554a39a1f1f83a4134bfe/quinn-proto/src/connection/mod.rs#L2451-L2454)), which just repeats the blocked attempt.

So if `close()` lands while the connection holds ≥ cwnd of unacked ack-eliciting data — routine for a bulk sender that cancels mid-transfer — the CONNECTION_CLOSE datagram is never sent. The closing side drains locally after 3×PTO and the peer sees silence until its idle timeout.

## Real-world impact

We hit this in a file-transfer engine: cancelling a sender mid-transfer left the receiver as a zombie for its full 134 s idle timeout, on every cancel that landed while the window was saturated. Packet captures show zero CONNECTION_CLOSE datagrams leaving the sender.

## Repro

Test against master's `quinn-proto` internal harness. The identical unsaturated flow (the existing `lifecycle` test) passes; this fails:

```rust
/// A connection closed while its congestion window is saturated must still
/// deliver CONNECTION_CLOSE to the peer promptly, rather than leaving the peer
/// to discover the close via its idle timeout.
#[test]
fn connection_close_sent_while_congestion_blocked() {
    let _guard = subscribe();
    let (client_ch, server_ch) = pair.connect();
    pair.drive();

    // Saturate the congestion window with unacknowledged stream data by
    // transmitting from the client without ever driving the server, so no
    // ACKs come back and `in_flight` stays pinned at the window.
    let s = pair.client_streams(client_ch).open(Dir::Uni).unwrap();
    let data = vec![42u8; 1024 * 1024];
    pair.client_send(client_ch, s).write(&data).unwrap();
    pair.drive_client();

    // Close while the window is full and stream data is still pending.
    const REASON: &[u8] = b"whee";
    let close_time = pair.time;
    pair.client.connections.get_mut(&client_ch).unwrap().close(
        pair.time,
        VarInt(42),
        REASON.into(),
    );
    pair.drive();

    // Drain stream events; the server should have learned of the close via a
    // CONNECTION_CLOSE datagram within a few PTOs, not by idle-timing out.
    let mut reason = None;
    while let Some(event) = pair.server_conn_mut(server_ch).poll() {
        if let Event::ConnectionLost { reason: r } = event {
            reason = Some(r);
        }
    }
    assert!(
        pair.time - close_time < Duration::from_secs(5),
        "server only noticed the close after {:?} (reason: {:?})",
        pair.time - close_time,
        reason,
    );
    assert_matches!(reason, Some(ConnectionError::ApplicationClosed(_)));
}
```

```
assertion failed: `None` does not match `Some(ConnectionError::ApplicationClosed(_))`
```

The trace shows the mechanism directly — immediately after `close()`, every transmit attempt is congestion-blocked and only the local close timer ever fires:

```
TRACE client: quinn_proto::connection: got Data packet (34 bytes) from [::1]:4433 ...   <- server ACK, ignored in Closed state
TRACE client: quinn_proto::connection: blocked by congestion control
TRACE quinn_proto::tests::util: advancing to 104.64ms for client
TRACE client: quinn_proto::connection: timeout timer=Close
```

## Expected behavior

RFC 9000 §10.2.1/§10.2.2: an endpoint entering the closing state SHOULD send a packet containing CONNECTION_CLOSE immediately. RFC 9002 exempts packets containing only ACK/CONNECTION_CLOSE from congestion accounting (they are not ack-eliciting and not in-flight), so sending the close-only packet while congestion-blocked is permitted.

A minimal fix direction: when `self.close` is pending (or the state is closing/draining), the packet to be built is the close-only packet, so treat it as non-ack-eliciting for the congestion/pacing gate (or skip the gate entirely for that space).

## Workaround

For anyone else hitting this: resetting outstanding send streams and then waiting ~1 RTT before calling `close()` (so returning ACKs can drain `in_flight` below the window while ACK processing still runs) lets the close frame out reliably in practice.

christofferbing:
Fixes #2785.

Calling `close()` while the congestion window is full never gets CONNECTION_CLOSE onto
the wire. The gate in `poll_transmit` classifies the packet by the stream frames still
pending in the space, but a closing connection will never send those — the packet it
actually builds is just ACKs + CONNECTION_CLOSE. And once closed nothing can reopen the
gate: incoming ACKs aren't processed anymore so in-flight bytes never drain, and
`close_common()` stopped the loss timers so the `loss_probes` bypass never fires. The
peer just sits there until its idle timeout.

We hit this cancelling a bulk sender mid-transfer — captures showed zero close datagrams
leaving, every time the cancel landed on a saturated window.

Fix is to exempt close packets from the gate, same idea as the existing tail-loss-probe
exemption on the line above: the close packet isn't congestion controlled, and blocking
it can deadlock.

The new test saturates the window with unacked data, closes, and checks the peer sees
the right close code/reason within a few PTOs instead of idle-timing out. Fails on main,
passes with the change.

### DIFF
--- a/quinn-proto/src/tests/mod.rs
+++ b/quinn-proto/src/tests/mod.rs
@@ -1362,6 +1362,57 @@ fn connection_close_sends_acks() {
     );
 }
 
+/// A connection closed while its congestion window is saturated must still deliver
+/// CONNECTION_CLOSE to the peer promptly, rather than leaving the peer to discover the close via
+/// its idle timeout (see https://github.com/quinn-rs/quinn/issues/2785)
+#[test]
+fn connection_close_while_congestion_blocked() {
+    let _guard = subscribe();
+    let mut pair = Pair::default();
+    let (client_ch, server_ch) = pair.connect();
+
+    // Saturate the congestion window with unacknowledged stream data by transmitting from the
+    // client without driving the server, so no ACKs come back and in-flight bytes stay pinned at
+    // the window
+    let s = pair.client_streams(client_ch).open(Dir::Uni).unwrap();
+    pair.client_send(client_ch, s)
+        .write(&[42; 1024 * 1024])
+        .unwrap();
+    pair.drive_client();
+
+    // Close while the window is full and stream data is still pending
+    const REASON: &[u8] = b"whee";
+    let close_time = pair.time;
+    pair.client.connections.get_mut(&client_ch).unwrap().close(
+        pair.time,
+        VarInt(42),
+        REASON.into(),
+    );
+
+    // Step the simulation by hand so we can catch the exact moment the server hears about the
+    // close: check for the event after each packet exchange, before the clock jumps ahead
+    let mut result = None;
+    loop {
+        pair.drive_client();
+        pair.drive_server();
+        while let Some(event) = pair.server_conn_mut(server_ch).poll() {
+            if let Event::ConnectionLost { reason } = event {
+                result = Some((reason, pair.time));
+            }
+        }
+        if result.is_some() || !pair.step() {
+            break;
+        }
+    }
+    let (reason, delivered_at) = result.expect("server never learned of the close");
+    assert_matches!(reason, ConnectionError::ApplicationClosed(
+        ApplicationClose { error_code: VarInt(42), ref reason }
+    ) if reason == REASON);
+    // Close packets aren't congestion controlled and the test link has no latency, so the close
+    // should arrive the moment it was issued — any delay means a timer had to rescue it
+    assert_eq!(delivered_at, close_time);
+}
+
 #[test]
 fn server_hs_retransmit() {
     let _guard = subscribe();
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -594,8 +594,13 @@ impl Connection {
                 }
 
                 // Congestion control and pacing checks
-                // Tail loss probes must not be blocked by congestion, or a deadlock could arise
-                if ack_eliciting && self.spaces[space_id].loss_probes == 0 {
+                // Tail loss probes must not be blocked by congestion, or a deadlock could arise.
+                // Close packets contain only ACKs and CONNECTION_CLOSE, neither of which is
+                // congestion controlled, and must not be blocked either: `ack_eliciting` reflects
+                // pending frames that will never be sent once closing, and a closed connection no
+                // longer processes ACKs, so the window could never drain
+                // (see https://github.com/quinn-rs/quinn/issues/2785)
+                if ack_eliciting && self.spaces[space_id].loss_probes == 0 && !close {
                     // Assume the current packet will get padded to fill the segment
                     let untracked_bytes = if let Some(builder) = &builder_storage {
                         buf_capacity - builder.partial_encode.start


quinn-proto/src/tests/mod.rs
```diff
@@ -1362,6 +1362,50 @@ fn connection_close_sends_acks() {
...
+    while let Some(event) = pair.server_conn_mut(server_ch).poll() {
+        if let Event::ConnectionLost { reason: r } = event {
+            reason = Some(r);
+        }
+    }
+    assert_matches!(reason, Some(ConnectionError::ApplicationClosed(
+        ApplicationClose { error_code: VarInt(42), ref reason }
+    )) if reason == REASON);
+    assert!(
+        pair.time - close_time < Duration::from_secs(5),
```
Ralith: Hard-coding 5 seconds here is opaque and potentially fragile. Maybe configure an explicit idle timeout in the test, and assert that the delay is less than that? Or better yet, perhaps we could exit the `while` loop above upon receiving `ApplicationClosed` and assert that the time difference is zero?

Ralith (approved):
LGTM aside from a test nit, thanks!

djc (approved):

quinn-proto/src/tests/mod.rs
christofferbing: Good call, went with the second option. The loop now observes the delivery instant directly and asserts zero delay.

djc:
Would you be able to backport this to `0.11.x`? Would be nice.

Issue #1855, opened 2024-05-08
Black hole detection false-positives


Ralith:
Black hole detection triggers when the number of "suspicious loss bursts" since the last MTU-sized packet exceeds a threshold:

https://github.com/quinn-rs/quinn/blob/a35ad73f97d7c4087c7c62b1650c589d2c0637ea/quinn-proto/src/connection/mtud.rs#L398-L409

Due to approximations in packet assembly, we (almost?) never actually produce an exactly MTU-sized packet. As a result, any loss rate of packets larger than the minimum threshold to be suspicious will eventually trigger black hole detection, resulting in a spurious MTU reduction and loss of throughput.

We should relax the conditions for resetting the counter. For example, if we've confirmed delivery of a packet of a certain size, loss bursts containing packets of equal or smaller size shouldn't be considered suspicious any more.

Ralith:
Design proposal:

A loss burst is said to be suspicious (might be caused by an MTU reduction) if its smallest packet is larger than any more recently acknowledged packet. Packets are deemed lost when at most one packet threshold (typically set to 3) worth of later packets have been acknowledged. We therefore need to store only at most one packet threshold worth of packet sizes, those of the most recently acknowledged packets not preceding a loss burst, to judge that a future loss burst might be caused by an MTU reduction. Past loss bursts are forgotten immediately upon acknowledgement of a larger packet than their smallest. If the number of past loss bursts exceeds a threshold, then we report a black hole.

djc:
Sounds okay. Should we run this by quicdev or something?

Ralith:
Some thoughts from that discussion:

- Some paths don't have a stable MTU, e.g. ones that internally load-balance over heterogeneous paths at a lower level. These are very difficult to detect, perhaps best dealt with manually (complain to your network administrator and/or configure a maximum MTU).
- Clamping MTU to the largest packet successfully delivered in some fixed window is safe, but prone to ratcheting downwards with application datagrams.
- Probing regularly and/or when suspicious loss occurs helps provide a source of truth when application traffic isn't so obliging, but isn't free.

Still seems to me like the approach I sketched above is an attractive sweet spot.

djc:
> A loss burst is said to be suspicious (possibly caused by an MTU reduction) if its smallest packet is larger than any more recently acknowledged packet.

[...]

PR #2792, opened 2026-08-18
proto: treat an equal-size delivery as evidence against a preceding loss burst

### LINKED ISSUE #2791: Black hole detector ignores deliveries of the same size as `acked_mtu`, so ordinary congestion loss in a bulk transfer is judged an MTU black hole
## What we see

On a production QUIC server (quinn-proto 0.11.16, BBR, ~80–95 ms RTT path with a little non-congestive loss), every connection that carries a sustained bulk transfer has its MTU knocked from the discovered 1452 down to `min_mtu` (1200) shortly after the transfer starts, and it stays there for the rest of the transfer. Idle connections on the same path probe to 1452 and stay there, so the path itself is fine.

`PathStats::black_holes_detected` after we exposed it in our close log (a 5 minute window, one speed test in the middle):

| connection | lifetime | final `current_mtu` | `black_holes_detected` |
|---|---|---|---|
| carried the ~150–200 Mbit/s download | 4m50s | **1200** | **3316** |
| light traffic | 8m03s | 1452 | 183 |
| light traffic | 5m28s | 1452 | 10 |
| light traffic | 12m32s | 1452 | 2 |
| light traffic | 6m04s | 1452 | 1 |
| idle (6 connections) | 2–5 min | 1452 | 0 |

The peer acknowledges roughly every 4th packet with `max_ack_delay = 100 ms`. A packet capture confirmed the server's datagrams drop from 1452 to 1200 bytes a few seconds into the download and never go back up while it lasts.

I'm aware of #2212, where this symptom was called expected. I think the specific reason it happens is a precision bug in the detector rather than the heuristic working as designed, so I'm filing it separately with the analysis below. The design statement in #1855 says a burst is suspicious only if its smallest packet is *"larger than any more recently acknowledged packet"*; the implementation doesn't quite honor the *any*.

## Why it happens

Two things combine, both in `BlackHoleDetector` (`quinn-proto/src/connection/mtud.rs`).

**1. Deliveries of the same size as `acked_mtu` are not recorded as evidence.** `on_non_probe_acked` returns early when `len <= acked_mtu`, so once a full-size packet has been acknowledged, `largest_post_loss_packet` never advances again while every packet is full-size — which is exactly the steady state of a bulk transfer. `on_ack_received` registers newly acknowledged packets *before* it runs loss detection (`mtud.on_acked` precedes `detect_lost_packets`), so a single ACK that both confirms newer full-size packets and reveals ≥ `BLACK_HOLE_THRESHOLD + 1` older, non-adjacent holes has those holes judged as newer than `largest_post_loss_packet` → all suspicious → black hole, even though the ACK itself just proved that full-size packets are still getting through. With ~13k packets/s and one ACK per ~4 packets, four non-adjacent holes revealed by one ACK is routine under congestion loss. This is still the case on `main`.

**2. On 0.11.x, once at `min_mtu` the detector keeps re-triggering.** `finish_loss_burst` uses `smallest_packet_size < self.min_mtu`, so a burst of packets that are exactly `min_mtu` bytes is still suspicious. After the first false positive every packet is 1200 bytes, every congestion-loss burst counts, the detector fires again every four bursts, and each firing resets `Phase::Complete(now + black_hole_cooldown)`, so probing never resumes while the transfer lasts. That is what the 3316 in the table is: it is not 3316 round trips through discovery, it is one false positive followed by ~3300 re-triggers at the floor. #2400 fixed this comparison on `main` (`<=`), but it was never backported — the author didn't need it — and 0.12 hasn't been released, so 0.11.17 still has it.

## Fix

Part 1 is a one-branch change: when `len == acked_mtu`, keep `acked_mtu` as is but advance `largest_post_loss_packet` to the newer packet number. Loss bursts that precede a delivery of that size are then non-suspicious under the existing rule (with #2400's `<=`), and losses transmitted after the last delivery are still suspicious, so a real black hole (large packets lost while only small ones get through) is still detected. The existing `blackhole_after_mtu_change_repairs_itself` end-to-end test keeps passing.

PR against `main`: #2792. A backport branch for 0.11.x that carries the two #2400 commits plus this fix is ready as well (https://github.com/vcarus/quinn/tree/bhd-0.11.x, `cargo test -p quinn-proto` green); happy to open it if you'd take a 0.11.18 with it.

Cost of the current behavior for us is small (about 20% more packets, ~1% goodput, and a smaller `max_datagram_size` while a download is in flight); the reason to fix it is that the detector reaches the wrong verdict on the most common traffic pattern there is.

vcarus:
Fixes #2791.

`BlackHoleDetector::on_non_probe_acked` returned early for any packet no larger than `acked_mtu`, so once a full-size packet had been acknowledged, `largest_post_loss_packet` never advanced again as long as every packet was full-size — the steady state of a bulk transfer. Because `on_ack_received` registers acknowledgments before running loss detection, one ACK that confirms newer full-size packets and reveals ≥ `BLACK_HOLE_THRESHOLD + 1` older non-adjacent holes then judged all of those holes suspicious, even though the same ACK had just shown that full-size packets still get through. On a lossy path this turns ordinary congestion loss during a download into a black hole verdict and a drop to `min_mtu` (details and production numbers in #2791).

The change: when `len == acked_mtu`, keep `acked_mtu` but advance `largest_post_loss_packet` to the newer packet number, so loss bursts that precede that delivery are non-suspicious under the existing rule while losses transmitted after the last delivery remain suspicious. This matches the design in #1855 ("larger than any more recently acknowledged packet").

The new unit test fails on `main` before the change and passes after; the existing black hole tests, including `blackhole_after_mtu_change_repairs_itself`, still pass.

On 0.11.x this fix alone is not sufficient because `finish_loss_burst` there still uses strict `<` (fixed on `main` by #2400 but not backported); I have a branch with #2400's two commits cherry-picked plus this one, all tests green (https://github.com/vcarus/quinn/tree/bhd-0.11.x), if you'd like a backport PR.

### DIFF
--- a/quinn-proto/src/connection/mtud.rs
+++ b/quinn-proto/src/connection/mtud.rs
@@ -403,9 +403,16 @@ impl BlackHoleDetector {
     }
 
     fn on_non_probe_acked(&mut self, pn: u64, len: u16) {
-        if len <= self.acked_mtu {
+        if len < self.acked_mtu {
             // We've already seen a larger packet since the most recent suspicious loss burst;
             // nothing to do.
+            return;
+        }
+        if len == self.acked_mtu {
+            // Another delivery of the largest size seen since the most recent suspicious loss
+            // burst. It doesn't raise `acked_mtu`, but loss bursts that precede it cannot be
+            // explained by an MTU reduction either, so remember it as the newest such delivery.
+            self.largest_post_loss_packet = self.largest_post_loss_packet.max(pn);
             return;
         }
         self.acked_mtu = len;
@@ -961,6 +968,36 @@ mod tests {
         assert!(
             bhd.black_hole_detected(),
             "1300 byte losses following a 1400 byte delivery are suspicious"
+        );
+    }
+
+    // Loss bursts that precede the delivery of another packet of the largest size seen so far
+    // are not suspicious: that delivery proves the path still carries packets of that size, even
+    // though it does not raise `acked_mtu`. This is the steady state of a bulk transfer, where
+    // every packet is full-size and one ACK typically both confirms new packets and reveals
+    // older losses.
+    #[test]
+    fn equal_size_delivery_clears_preceding_bursts() {
+        let mut bhd = BlackHoleDetector::new(1200);
+        // A full-size packet was delivered long ago...
+        bhd.on_non_probe_acked(0, 1400);
+        // ...and a newer full-size packet is delivered now, before loss detection runs
+        bhd.on_non_probe_acked((BLACK_HOLE_THRESHOLD + 1) as u64 * 2, 1400);
+        // Loss detection then reveals bursts that were transmitted before that delivery
+        for i in 0..(BLACK_HOLE_THRESHOLD + 1) {
+            bhd.on_non_probe_lost(i as u64 * 2 + 1, 1400);
+        }
+        assert!(
+            !bhd.black_hole_detected(),
+            "full-size losses preceding a full-size delivery are not suspicious"
+        );
+        // Full-size losses transmitted after the last delivery are still suspicious
+        for i in 0..(BLACK_HOLE_THRESHOLD + 1) {
+            bhd.on_non_probe_lost((BLACK_HOLE_THRESHOLD as u64 + 1 + i as u64) * 2 + 1, 1400);
+        }
+        assert!(
+            bhd.black_hole_detected(),
+            "full-size losses following the last full-size delivery are suspicious"
         );
     }
 


Ralith (approved):
LGTM, thanks! A backport would be welcome.

djc (approved):

vcarus:
Heads-up on the merge queue run: the only red job out of 20 was `audit`, and it is unrelated to this PR.

RUSTSEC-2026-0258 (h2 < 0.4.16, empty DATA frames) was published on 2026-08-17 and cargo-deny now errors on it: https://github.com/quinn-rs/quinn/actions/runs/32289479766/job/96186717814

h2 is in the lock file only through `perf`'s non-default `tokio-console` feature (console-subscriber -> tonic -> hyper), so no published crate is affected, but the lock file entry fails the check for every PR that enters the queue. I opened #2797 with the two-line lock file bump; re-enqueuing this one should be green once that lands.

Issue #2783, opened 2026-08-13
Cannot configure ack_eliciting_threshold.


nwin:
Currently, `quinn-proto` follows the recommendation in the QUIC specification and only acknowledges every other packet (`PendingAcks::ack_eliciting_threshold == 1` in `PendingAcks::new()`).

It is possible to configure the ACK frequency in the sending direction via `AckFrequencyConfig`, and Quinn also obeys when the peer sends an `ACK_FREQUENCY ` frame:

https://github.com/quinn-rs/quinn/blob/148d760d8845dc6315d554a39a1f1f83a4134bfe/quinn-proto/src/connection/spaces.rs#L644

https://github.com/quinn-rs/quinn/blob/148d760d8845dc6315d554a39a1f1f83a4134bfe/quinn-proto/src/connection/ack_frequency.rs#L141

However, there does not appear to be a way to configure `quinn-protos`'s own receive-side ACK frequency via code. The default threshold remains hard-coded until the peer requests a different value through ACK_FREQUENCY.

This can be harmful in cases when the peer does not support or send ACK_FREQUENCY. For example, if the peer sends a single ACK-eliciting packet and waits for its acknowledgement before sending more data, `quinn` has to wait for the delayed-ACK timeout (typically 25 ms), even when the application would prefer to acknowledge every packet immediately.

In particular, this seems to be the behavior of the SMB over QUIC client of Windows.

I would be happy to open a PR if the change is desired and someone would advise where to change the API.

Ralith:
It is possible for Quinn to support customizing this behavior. However, taking the scenario you describe at face value, it sounds like this is only needed as a workaround for a poorly behaved peer: applications should not rely on transport-layer ACKs to drive application-layer behavior such as sending new application data. Have you reported this behavior to Microsoft?

A fix on their end would improve performance for everyone, whereas configuring Quinn this way would likely be more costly for both peers compared to a system which does not improperly rely on transport ACKs.

nwin:
I would not say this would be a workaround. RFC 9000 §13.2.2 says:
[...]

PR #2793, opened 2026-08-18
Add option to configure ACK frequency. Fixes 2783

nwin:
Proposed fix for #2783.

The idea is to reuse the existing `quinn::TransportConfig`.

### DIFF
--- a/quinn-proto/src/config/transport.rs
+++ b/quinn-proto/src/config/transport.rs
@@ -233,8 +233,9 @@ impl TransportConfig {
 
     /// Specifies the ACK frequency config (see [`AckFrequencyConfig`] for details)
     ///
-    /// The provided configuration will be ignored if the peer does not support the acknowledgement
-    /// frequency QUIC extension.
+    /// The configuration is also used as the local endpoint's initial acknowledgement policy for
+    /// application-data packets. A valid `ACK_FREQUENCY` frame received from the peer may
+    /// subsequently replace it.
     ///
     /// Defaults to `None`, which disables controlling the peer's acknowledgement frequency. Even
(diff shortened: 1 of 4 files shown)


quinn-proto/src/config/transport.rs
```diff
@@ -233,8 +233,9 @@ impl TransportConfig {
 
     /// Specifies the ACK frequency config (see [`AckFrequencyConfig`] for details)
     ///
-    /// The provided configuration will be ignored if the peer does not support the acknowledgement
-    /// frequency QUIC extension.
+    /// The configuration is also used as the local endpoint's initial acknowledgement policy for
+    /// application-data packets. A valid `ACK_FREQUENCY` frame received from the peer may
+    /// subsequently replace it.
```
Ralith, line 238: This conflates customizing the default ACK behavior of the local endpoint with requesting specific behavior from the peer. This is confusing and not necessarily desirable, since communication is frequently asymmetrical.

In your issue you mentioned that other implementations expose a similar setting. What granularity does that setting have, and how does it interact with the ACK frequency extension?

nwin: ngtcp2 is actually the most straightforward

```
#include <ngtcp2/ngtcp2.h>
ngtcp2_settings settings;
ngtcp2_settings_default(&settings);
// Note 1 means after every packet
settings.ack_thresh = 1;
```
So going from this I could just not re-use the ACK frequency setting struct but add a new setting.

Ralith: That excerpt doesn't clearly answer the questions I asked above. A new setting does seem like the least surprising thing to do, intuitively.

PR #2794, opened 2026-08-19
quinn-proto: avoid unnecessary PING when a DATAGRAM fits into a tail-loss probe

### LINKED ISSUE #2173: When sending a tail-loss probe a PING frame can be added to a DATAGRAM frame
When a tail-loss probe needs to be sent the current `queue_probe` algorithm adds a PING frame even if a DATAGRAM frame could also be sent.  This wastes a single byte, potentially amplified by being the start of a GSO batch.

This isn't that bad, there are many cases in which the user datagram would not fit into the probe packet anyway.  Still, it could be a nice small improvement.

See #2169 which tried to solve this before, but probably broke some edge cases: https://github.com/quinn-rs/quinn/pull/2169/files#r1987103861

See #2172 which reverts this part of #2169 again.

hunterinvariants:
When the peer does not support ACK Frequency, the loss-probe logic falls back to a PING frame to make the probe ack-eliciting. If a DATAGRAM is already queued and actually fits into that probe, the extra PING is redundant – DATAGRAM frames are ack-eliciting on their own.

The catch is that loss probes are always capped at INITIAL_MTU (1200 bytes), which can be smaller than the current path MTU. A DATAGRAM that comfortably fits a normal-sized packet might not fit into the smaller probe packet. So we cannot just check whether a DATAGRAM is pending; we have to check whether it physically fits into that constrained probe space.

The change does exactly that: for the 1-RTT space it computes the available frame space inside an INITIAL_MTU-sized probe and only avoids queueing the fallback PING if can_send_1rtt() confirms that the pending ACK-eliciting 1-RTT data can actually be placed there.

The new regression test (tail_loss_probe_keeps_ping_when_datagram_does_not_fit) sets the path MTU to 1452, queues a DATAGRAM that fits 1452 but not 1200, and verifies that the fallback PING is queued. The existing tail_loss_small_segment_size test was adjusted to confirm that the fallback PING is not queued when the DATAGRAM does fit. All local checks – formatting, clippy, and the full quinn-proto test suite with default and all features enabled – are green.

Closes #2173

Related context: #2169 and #2172

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -475,9 +475,25 @@ impl Connection {
 
         // If we need to send a probe, make sure we have something to send.
         for space in SpaceId::iter() {
-            let request_immediate_ack =
-                space == SpaceId::Data && self.peer_supports_ack_frequency();
-            self.spaces[space].maybe_queue_probe(request_immediate_ack, &self.streams);
+            if space != SpaceId::Data {
+                self.spaces[space].maybe_queue_probe(false, false, &self.streams);
+                continue;
+            }
+
+            let has_ack_eliciting_data = self.can_send_1rtt(
+                Ord::min(segment_size, usize::from(INITIAL_MTU)).saturating_sub(
+                    self.predict_1rtt_overhead(Some(
+                        self.packet_number_filter.peek(&self.spaces[SpaceId::Data]),
+                    )),
+                ),
+            );
+            let request_immediate_ack = self.peer_supports_ack_frequency();
+
+            self.spaces[space].maybe_queue_probe(
+                request_immediate_ack,
+                has_ack_eliciting_data,
+                &self.streams,
+            );
         }
 
         // Check whether we need to send a close message
@@ -3590,6 +3606,11 @@ impl Connection {
 
     fn peer_supports_ack_frequency(&self) -> bool {
         self.peer_params.min_ack_delay.is_some()
+    }
+
+    #[cfg(test)]
+    pub(crate) fn disable_peer_ack_frequency(&mut self) {
+        self.peer_params.min_ack_delay = None;
     }
 
     /// Send an IMMEDIATE_ACK frame to the remote endpoint
--- a/quinn-proto/src/connection/spaces.rs
+++ b/quinn-proto/src/connection/spaces.rs
@@ -119,6 +119,7 @@ impl PacketSpace {
     pub(super) fn maybe_queue_probe(
         &mut self,
         request_immediate_ack: bool,
+        has_ack_eliciting_data: bool,
         streams: &StreamsState,
     ) {
         if self.loss_probes == 0 {
@@ -133,6 +134,12 @@ impl PacketSpace {
 
         if !self.pending.is_empty(streams) {
             // There's real data to send here, no need to make something up
+            return;
+        }
+
+        if has_ack_eliciting_data {
+            // A fitting DATAGRAM is held outside `pending` until packet assembly, but it
+            // still makes the probe ack-eliciting without a fallback frame.
             return;
         }
 
--- a/quinn-proto/src/tests/mod.rs
+++ b/quinn-proto/src/tests/mod.rs
@@ -1917,6 +1917,9 @@ fn tail_loss_small_segment_size() {
     let _guard = subscribe();
     let mut pair = Pair::default();
     let (client_ch, server_ch) = pair.connect();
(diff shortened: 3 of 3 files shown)


hunterinvariants:
Quick check on the remaining audit failure: the DATAGRAM tail-loss-probe change itself is ready, and the other checks pass. Is the audit result something this branch is expected to resolve through a dependency refresh, or is it currently a repository-wide condition? I can update the branch if a PR-local change is required.

djc:
I've submitted

- https://github.com/quinn-rs/quinn/pull/2811

to deal with the audit failure.

quinn-proto/src/connection/mod.rs
```diff
@@ -475,9 +475,19 @@ impl Connection {
...
+                self.can_send_1rtt(frame_space)
+            };
             let request_immediate_ack =
                 space == SpaceId::Data && self.peer_supports_ack_frequency();
-            self.spaces[space].maybe_queue_probe(request_immediate_ack, &self.streams);
+            self.spaces[space].maybe_queue_probe(
+                request_immediate_ack,
+                has_ack_eliciting_data,
+                &self.streams,
+            );
```
djc: This looks a little complicated. How about we do an early return for `space != `SpaceId::Data`?

hunterinvariants: Thanks, that is clearer. I changed the non-Data path to queue its probe and continue early, leaving the Data-specific sizing and ACK-frequency logic linear. `cargo test --locked --package quinn-proto --all-features` passes with 290 unit tests and 3 doc tests.

quinn-proto/src/connection/spaces.rs
```diff
@@ -136,6 +137,11 @@ impl PacketSpace {
             return;
         }
 
+        if has_ack_eliciting_data {
```
Ralith, line 140: In context, it's not obvious why this is different from `self.pending.is_empty`. Document.

quinn-proto/src/tests/mod.rs
```diff
@@ -1917,6 +1917,7 @@ fn tail_loss_small_segment_size() {
     let _guard = subscribe();
     let mut pair = Pair::default();
     let (client_ch, server_ch) = pair.connect();
+    pair.client_conn_mut(client_ch).disable_peer_ack_frequency();
```
Ralith, line 1922: Document why this is needed.

Ralith (changes requested):

quinn-proto/src/connection/spaces.rs
hunterinvariants: Documented in 7646f4ab: fitting DATAGRAMs remain outside `pending` until packet assembly.

quinn-proto/src/tests/mod.rs
hunterinvariants: Documented in 7646f4ab: disabling ACK frequency prevents IMMEDIATE_ACK from masking the DATAGRAM-versus-PING assertion.

quinn-proto/src/connection/mod.rs
```diff
@@ -475,9 +475,21 @@ impl Connection {
...
+            if space != SpaceId::Data {
+                self.spaces[space].maybe_queue_probe(false, false, &self.streams);
+                continue;
+            }
+
+            let pn = self.packet_number_filter.peek(&self.spaces[SpaceId::Data]);
+            let probe_size = segment_size.min(usize::from(INITIAL_MTU));
+            let frame_space = probe_size.saturating_sub(self.predict_1rtt_overhead(Some(pn)));
+            let has_ack_eliciting_data = self.can_send_1rtt(frame_space);
+            let request_immediate_ack = self.peer_supports_ack_frequency();
```
djc: Nits: instead of single-use bindings, suggest inlining most of these values.

Please use a two-sided `min()` like `Ord::min()` instead of using the asymmetric `min()` method.

(And squash all your commits into a single one.)

hunterinvariants: Addressed in 162005d0: switched to `Ord::min()`, inlined the size calculation, and squashed the PR to one commit. The two boolean results remain precomputed to avoid overlapping the mutable `spaces` borrow. All `quinn-proto` tests pass.

djc (approved):

hunterinvariants:
@Ralith The requested documentation changes are included in the current squashed commit `162005d`. All 20 checks pass, and djc has approved the current commit. Could you re-review and update the previous changes-requested review?

Ralith (approved):

PR #2795, opened 2026-08-19
Release endpoint lock during incoming TLS setup

kkroo:
We hit this running a server that resolves certificates on demand rather than from a static keypair. Throughput would collapse whenever a new client connected, and it took a while to work out why.

`EndpointInner::accept` takes the endpoint state lock and holds it across `state.inner.accept(...)`, which is where rustls processes the ClientHello — and rustls calls `ResolvesServerCert` synchronously from there. So if your resolver does anything slow, you are holding the endpoint lock for the duration. Not just that handshake: UDP intake and every other connection on the endpoint wait behind it.

The test I added sets up a resolver that blocks, then checks that a connection which was already established keeps making progress while a new handshake is stuck in it. On main it fails:

```
panicked at quinn/src/tests.rs:445:
  existing connection stalled during certificate resolution: Elapsed(())
```

The fix splits accept into three parts so the lock only covers the endpoint state mutations. `prepare_accept` allocates the connection under the lock and points its packets at the bounded incoming buffer. `process_first_packet` drives the ClientHello with the lock released, which is where the resolver runs. `finish_accept` takes the lock again to reinstall routing and register the connection.

Packets that arrive while the resolver is working keep going to the incoming buffer that already exists for pending connections, so nothing gets dropped and there is no new queue to bound. The single-call `accept` still works exactly as before, so this is invisible to existing callers; the two halves are `#[doc(hidden)]` in quinn-proto since quinn is the only intended caller.

Worth mentioning #2024, since it is adjacent. That asks for ClientHello access before accepting, which would let people use the rustls `Acceptor` pattern and do resolution before a connection exists — a better answer in general. This is narrower: it does not expose anything new, it just stops a synchronous resolver from blocking the whole endpoint. The two are compatible, and this still helps anyone who wants to keep a sync resolver.

If you would rather `prepare_accept`/`finish_accept` stay private and the split live entirely inside quinn, I am happy to redo it that way — I put them in quinn-proto because that is where the routing state lives, but I do not feel strongly about it.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -343,7 +343,8 @@ impl Endpoint {
         trace!(initial_dcid = %remote_id);
 
         let ch = ConnectionHandle(self.connections.vacant_key());
-        let loc_cid = self.new_cid(RouteDatagramTo::Connection(ch));
+        let route = RouteDatagramTo::Connection(ch);
+        let loc_cid = self.new_cid(route);
         let params = TransportParameters::new(
             &config.transport,
             &self.config,
@@ -373,6 +374,7 @@ impl Endpoint {
                 token_store: config.token_store,
                 server_name: server_name.into(),
             },
+            route,
         );
         Ok((ch, conn))
     }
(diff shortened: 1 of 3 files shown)


Ralith:
I believe this is already covered by https://github.com/quinn-rs/quinn/pull/2782, but please let me know if I'm overlooking something.

PR #2798, opened 2026-08-20
Bound BBR startup cwnd growth on app-limited connections

poka-IT:
BBR's STARTUP path can grow cwnd without bound on connections that are app-limited from birth (a tunnel trickling below link rate, never saturating). Two causes:

- `cwnd_gain < target_window as f32` compares a gain (about 2.9) against a byte count, so the guard is effectively always true and cwnd tracks cumulative acked bytes. Changed to `cwnd < target_window`.
- App-limited bandwidth samples are never admitted to the max filter. On a connection app-limited from its first round the filter stays empty, `expected_bytes_acked` stays 0, the ack-aggregation epoch never resets, and `excess_acked` grows with every ack. Samples are now admitted when they raise the estimate, mirroring Chromium/quiche's admission rule.

Three tests cover the bounded window in both app-limited scenarios and the normal STARTUP ramp. Observed and fixed while running quinn as a VPN data plane; the fix has been in production in our fork.

### DIFF
--- a/quinn-proto/src/congestion/bbr/bw_estimation.rs
+++ b/quinn-proto/src/congestion/bbr/bw_estimation.rs
@@ -61,7 +61,18 @@ impl BandwidthEstimation {
         };
 
         let bandwidth = send_rate.min(ack_rate);
-        if !app_limited && self.max_filter.get() < bandwidth {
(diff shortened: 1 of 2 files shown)


Ralith:
I'm open to merging this, but we're planning to replace this entire module soon in #2481.

PR #2800, opened 2026-08-20
proto: tell ConnectionIdGenerator which peer a CID is for

efagerho:
Connection ID generators are constructed once per endpoint and are handed no context, so a generator cannot encode anything about the peer into the CIDs it issues. That rules out address-aware load balancing, and it rules out packet-level authentication schemes that bind the peer's address into the connection ID and validate it before the QUIC layer parses a packet.

Add a defaulted generate_cid_for(remote, local_ip) alongside generate_cid and call it everywhere a local CID is minted. Every call site already had the addresses to hand, except CID re-issuance, which now uses the address the connection was last observed at. Existing implementations, including both built-in generators, are unaffected.

ConnectionMeta::addresses was deliberately stale because zero-length CIDs cannot migrate. Refresh its remote address when the peer switches CIDs, which RFC 9000 requires it to do when migrating, so that re-issued CIDs follow a peer that moves.


Ralith:
> That rules out address-aware load balancing

Can you elaborate on the use case for this?

> it rules out packet-level authentication schemes that bind the peer's address

Are you sure you want this? It seems incompatible with connection migration. Under ideal conditions a migrating client should use a previously-issued but unused CID (which cannot know about the new address) on the very first outgoing packet to reduce linkability with the previous path, and in practice clients typically won't realize they're migrating and will continue to use their current CID.

Under what scenario is binding CIDs to addresses a useful improvement in security?

efagerho:
> > That rules out address-aware load balancing
> 
> Can you elaborate on the use case for this?
> 
> > it rules out packet-level authentication schemes that bind the peer's address
> 
> Are you sure you want this? It seems incompatible with connection migration. Under ideal conditions a migrating client should use a previously-issued but unused CID (which cannot know about the new address) on the very first outgoing packet to reduce linkability with the previous path, and in practice clients typically won't realize they're migrating and will continue to use their current CID.
> 
> Under what scenario is binding CIDs to addresses a useful improvement in security?

Line rate packet filtering between Solana validators. They do not move, so connection migration never happens and I’m aware that doing this prevents it.

Using this I can detect invalid packets in less than 200 clock cycles per packet.

Ralith:
Do you often deal with invalid packets that contain a valid CID from a recently live connection, but have the wrong source address?

efagerho:
> Do you often deal with invalid packets that contain a valid CID from a recently live connection, but have the wrong source address?

No. Not for the endpoints where this would be relevant. This patch is for a design, which has a proof of concept implementation here:

https://github.com/efagerho/pktauth

Ralith:
What's the reason for involving the source address at all, then? Superficially, it seems like just having a nonce + expiry time would solve the problem without these drawbacks.

Ralith:
To be clear, I'm not categorically opposed to this change, but it's a bit of a footgun so I want to be sure there isn't a simpler way.

efagerho:
> What's the reason for involving the source address at all, then? Superficially, it seems like just having a nonce + expiry time would solve the problem without these drawbacks.

I guess you could argue about the threat model. Let's say someone can listen to the traffic, then they could just copy the CIDs to their botnet, which could flood QUIC Initials with the captured CIDs that pass the filter. If you bind the IP, then your bots would need to spoof the source IP and that's getting progressively more difficult as ISPs are starting to take that problem more seriously.

Leaving out the source IP is probably beneficial in some settings, but not so in others.

quinn-proto/src/cid_generator.rs
```diff
@@ -17,6 +18,25 @@ pub trait ConnectionIdGenerator: Send + Sync {
...
+    /// issues, as address-aware load balancers and packet-level authentication
+    /// schemes need to. Note that a peer that migrates will keep using CIDs
+    /// issued for its previous address until it is issued new ones, and that
+    /// `local_ip` is only populated on platforms that report the destination
+    /// address of a received datagram.
+    ///
+    /// The requirements documented on [`generate_cid`](Self::generate_cid)
+    /// apply here too: whatever is encoded must still be opaque and
+    /// high-entropy to an observer who does not hold the issuer's secrets.
+    fn generate_cid_for(&mut self, _remote: SocketAddr, _local_ip: Option<IpAddr>) -> ConnectionId {
```
Ralith, line 36: Is there a risk that we'll come up with other context that might be desired in the future? I wonder if we want an extensible context object here.

PR #2802, opened 2026-08-20
proto: expose BBR bandwidth estimate in path stats

kixelated:
Expose the bandwidth estimate from the congestion controller.

With Reno/CUBIC we can compute this from window / RTT, but for BBR we need the actual estimate from the CC algorithm.


Ralith (approved):

djc (approved):

djc:
Thanks! Can you fix up the lint issue?

Ralith:
Squash the last two?

Did you work out why rustdoc doesn't like that link? Seems pretty strange.

PR #2803, opened 2026-08-20
proto: replace the assembler BinaryHeap with a sorted VecDeque

### LINKED ISSUE #2739: Stream assembler uses a BinaryHeap even when data arrives in order
**Summary**
The stream reassembly buffer in `quinn-proto/src/connection/assembler.rs` keeps chunks in a `BinaryHeap<Buffer>` and goes through it even when the stream arrives in order, which is the common case for a bulk transfer.

Avoiding the heap on the in-order path seems worth 2 to 3% of client CPU on the receive side. It will not show as throughput on x86, where the receive path is dominated by AES-GCM; as with #2735, the value is CPU on constrained hardware. Two options for your consideration, neither of which I have written yet:

  1. Hold contiguous chunks in a `VecDeque` and fall back to the heap when a gap appears. This removes the per-chunk heap work on the in order path. Cost: there would be two containers, and `buffered`, `allocated`, `end` and defragmentation all touch the heap today.
  2. Something smaller. A partial read in `read()` advances the front chunk's offset through `PeekMut`. Mutating through a `PeekMut` makes the heap sift the root back down when the `PeekMut` is dropped, in case the mutation broke the ordering. Usually it did not: the offset moves forward only within the chunk that was already at the front, so that chunk still sorts first and the sift changes nothing. Avoiding it on that path would be a much smaller change than option 1.

  **Measurements**
  I added a temporary counter to `Assembler::insert` to see what the heap holds during a bulk download. Over 900k inserts on a link with a 10ms delay:

  - average heap length at insert: 25
  - maximum heap length: 127
  - chunks starting beyond the highest offset seen so far, ie a gap: 5, or 0.001%

  The heap sits around 25 deep while the data is essentially always in order. The depth looks like the receive batch filling the assembler faster than the application drains it, rather than reordering. Every insert does `data.push()` and every read does `data.peek_mut()`, so each one sifts a 56 byte `Buffer` through about five levels to keep an order the data already has.

  `BinaryHeap<Buffer>::push` plus `PeekMut<Buffer>::pop` are 2.1% of client CPU on a 3 GiB loopback download, and about 3% on the 10ms link. Setup: `quinn-perf`, main plus #2730 and #2735, release build with debuginfo and frame pointers, `perf record -F 999`.

  One caveat. I could not get `tc netem`'s `reorder` to actually reorder in my test environment: at 25%, 50% and 99% it moved about one packet in fifty, and the assembler counters agree. So everything above is an in-order path. If you want the reordered case measured first, I can try again with two delay bands.

ojhermann:
Closes #2739 via option 1.

Chunks now live in a `VecDeque` ordered by offset instead of a `BinaryHeap`:
- `read` takes from the front
- an in-order frame appends to the back
- a frame below everything buffered pushes to the front; anything else is placed by binary search

Two orderings are now explicit:
- a partial read can advance the front chunk past a neighbour it overlaps
- `defragment` has to emit the accumulated chunk before a later defragmented one

Tie order stays unspecified, as it was under the heap.

Measurements
- nanoseconds per frame
- 1200 byte frames
- best-of-N with N sized to at least 300ms per case
- median of five processes, or three for the bounded-read and 32 chunk rows
- binaries interleaved

Note: the figures I posted in the issue were best-of-40 and too noisy to use, so please ignore them.

| case | BinaryHeap | VecDeque |
| --- | --- | --- |
| in-order, 1 chunk buffered | 52.6 | 31.5 |
| in-order, 25 chunks buffered | 64.1 | 31.5 |
| in-order, 127 chunks buffered | 84.7 | 31.8 |
| in-order, 1000 chunks buffered | 108.4 | 31.9 |
| in-order, 127 buffered, 1024 byte reads | 106.6 | 55.8 |
| descending offsets | 36.1 | 18.8 |
| gapped inserts mid-range, 32 buffered | 37.1 | 68.9 |
| gapped inserts mid-range, 500 buffered | 38.5 | 187.2 |
| gapped inserts mid-range, 1500 buffered | 37.1 | 423.9 |
| one `defragment` call at 1900 chunks | 179 us | 39 us |

Summary
- about 19ns of each in-order row (in both columns) is `Bytes::clone` plus the `Chunk` drop
- the heap slows down as the buffer deepens and the deque does not
- the bounded-read row exercises the partial-read path, where the new code restores ordering by hand; it costs nothing measurable, since the guard returns immediately on in-order data

The gapped rows are what it costs: the deque is already the slower of the two by about 32 buffered chunks; `COMPACT_THRESHOLD` holds the buffer to 2048 chunks of 56 bytes, so the shift should stay under ~57KB; in unordered mode the dedup loop can do several of these inserts for one frame, which I measured at parity with the heap.

`defragment` gets cheaper because its input is already ordered, so it no longer sorts.

### DIFF
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -1,8 +1,4 @@
-use std::{
-    cmp::Ordering,
-    collections::{BinaryHeap, binary_heap::PeekMut},
-    mem,
-};
+use std::{cmp::Ordering, collections::VecDeque, mem};
 
 use bytes::{Buf, Bytes, BytesMut};
 
@@ -12,7 +8,8 @@ use crate::range_set::RangeSet;
 #[derive(Debug, Default)]
 pub(super) struct Assembler {
     state: State,
-    data: BinaryHeap<Buffer>,
+    /// Buffered chunks, in `Buffer::cmp` order
+    data: VecDeque<Buffer>,
     /// Total number of buffered bytes, including duplicates in ordered mode.
     buffered: usize,
     /// Estimated number of allocated bytes, will never be less than `buffered`.
@@ -59,7 +56,7 @@ impl Assembler {
     /// Get the the next chunk
     pub(super) fn read(&mut self, max_length: usize, ordered: bool) -> Option<Chunk> {
         loop {
-            let mut chunk = self.data.peek_mut()?;
+            let chunk = self.data.front_mut()?;
 
             if ordered {
                 if chunk.offset > self.bytes_read {
@@ -69,12 +66,13 @@ impl Assembler {
                     // Next chunk is useless as the read index is beyond its end
                     self.buffered -= chunk.bytes.len();
                     self.allocated -= chunk.allocation_size;
-                    PeekMut::pop(chunk);
+                    self.data.pop_front();
                     continue;
                 }
 
                 // Determine `start` and `len` of the slice of useful data in chunk
                 let start = (self.bytes_read - chunk.offset) as usize;
+                // Advancing the offset can push the front past data[1]; both exits below fix that
                 if start > 0 {
                     chunk.bytes.advance(start);
                     chunk.offset += start as u64;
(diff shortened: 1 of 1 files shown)


quinn-proto/src/connection/assembler.rs
```diff
@@ -69,47 +66,78 @@ impl Assembler {
...
+            let bytes = mem::take(&mut chunk.bytes);
+            self.data.pop_front();
+            return Some(Chunk::new(offset, bytes));
+        }
+    }
+
+    /// O(1) for a frame in order or below everything buffered, otherwise an insert that
+    /// shifts up to half of `data` (see `COMPACT_THRESHOLD`)
+    // Not a method: `insert` holds a borrow of `self.state`.
+    fn insert_ordered(data: &mut VecDeque<Buffer>, buffer: Buffer) {
```
gretchenfrage: Thanks for this work! I don't believe there's any other functions in this project that are in an `impl` block but don't use a `self` receiver, the convention for these helper functions is to put it near the end of the file.

ojhermann: OK; change has been made

quinn-proto/src/connection/assembler.rs
```diff
@@ -745,6 +781,125 @@ mod test {
...
+
+    #[test]
+    fn assemble_random_unordered() {
+        random_stream(false);
+    }
+
+    /// Fragments arriving duplicated and shuffled must still read back as the exact stream
+    fn random_stream(ordered: bool) {
+        const LEN: usize = 2048;
+        let stream: Vec<u8> = (0..LEN).map(|i| (i % 251) as u8).collect();
```
Ralith: - Why 251?
- Would a purely random block of data be more robust?

quinn-proto/src/connection/assembler.rs
```diff
@@ -745,6 +781,125 @@ mod test {
...
+        let mut rng = new_rng();
+        for _ in 0..64 {
+            let mut frags = Vec::new();
+            let mut at = 0;
+            while at < LEN {
+                let size = rng.random_range(1..=32).min(LEN - at);
+                frags.push((at as u64, size));
+                at += size;
+            }
+            let dupes: Vec<_> = frags
```
Ralith: This will only simulate duplicates of exactly the same size as another frame. Is that intended? In reality we might have arbitrary overlaps, since QUIC doesn't require retransmission to use the same frames (and e.g. we don't in practice).

Ralith (approved):
LGTM, thanks! Please squash feedback-related changes to code introduced in the same PR.

quinn-proto/src/connection/assembler.rs
ojhermann: Agreed and updated. The stream is now random bytes. 251 was a prime, so the pattern wouldn't line up with frame boundaries, but there's no reason to prefer it over random data.

quinn-proto/src/connection/assembler.rs
ojhermann: Agreed and updated. Not intended, and you're right that it left partial overlap untested. Retransmits are now refragmented from scratch over a random range, so they cross the original frame boundaries. To check that this actually tests more, I dropped the non-duplicate prefix in the unordered dedup path in `insert`: the new test fails and the old one passes.

gretchenfrage (approved):

PR #2804, opened 2026-08-28
proto: apply RFC 9000 §12.4 packet-type rules to CONNECTION_CLOSE

jaideeppyne:
RFC 9000 §12.4 Table 3 marks CONNECTION_CLOSE `ih01`, where `ih` means "Only a CONNECTION_CLOSE frame of type 0x1c can appear in Initial or Handshake packets", and §12.4 requires PROTOCOL_VIOLATION on receipt of a frame in a packet type that is not permitted.

quinn applies this on the send path: `poll_transmit` rewrites an application close into a 0x1c close with APPLICATION_ERROR for non-Data spaces, per §10.2.3. The receive path does not.

Two changes, one per direction:

1. `process_early_payload` accepted both close types. Initial packets are protected with keys derived from the cleartext original DCID, so anyone observing the handshake can forge one and hand the application an arbitrary application error code and reason phrase. Now only 0x1c is accepted; 0x1d falls through to the existing "illegal frame type in handshake" arm.

2. The 0-RTT filter rejected a 0x1d close. §12.5 lists the frames that cannot appear in 0-RTT as ACK, CRYPTO, HANDSHAKE_DONE, NEW_TOKEN, PATH_RESPONSE and RETIRE_CONNECTION_ID. CONNECTION_CLOSE is not in that list, and §19.19 says 0x1d "can only be sent using 0-RTT or 1-RTT packets". This turned a peer's clean application close into a transport error.

Test: `application_close_in_initial_is_rejected` builds an Initial-key-protected packet containing a 0x1d close and checks the connection fails with PROTOCOL_VIOLATION rather than surfacing `ApplicationClosed`. It fails before this change. There is no test for the 0-RTT half because quinn never emits such a packet itself: `poll_transmit` skips the Data space for a close unless 1-RTT keys exist, so it cannot be reproduced quinn to quinn.

`cargo test -p quinn-proto` passes (294), `cargo test -p quinn` passes (23), fmt and clippy are clean.

### DIFF
--- a/quinn-proto/src/connection/mod.rs
+++ b/quinn-proto/src/connection/mod.rs
@@ -2750,7 +2750,10 @@ impl Connection {
                 Frame::Ack(ack) => {
                     self.on_ack_received(now, packet.header.space(), ack)?;
                 }
-                Frame::Close(reason) => {
+                // Per RFC 9000 §12.4 Table 3, only a CONNECTION_CLOSE frame of type 0x1c may
+                // appear in Initial or Handshake packets. An application close (0x1d) falls
+                // through to the catch-all arm below.
+                Frame::Close(reason @ Close::Connection(_)) => {
                     self.error = Some(reason.into());
                     self.state = State::Draining;
                     return Ok(());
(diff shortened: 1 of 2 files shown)


Ralith (approved):
Nice catch, thanks! Spec checks out.

djc:
Would you be able to backport this to `0.11.x`? Might be useful.

PR #2806, opened 2026-08-29
0.11.x: fix payload_bytes double subtraction when evicting outgoing datagrams

### LINKED ISSUE #2805: [0.11.x] Datagrams::send(drop=true) subtracts payload_bytes twice per evicted datagram (fixed on main)
On the `0.11.x` branch, the `send(drop = true)` eviction loop in `quinn-proto/src/connection/datagrams.rs` subtracts an evicted datagram's size from `payload_bytes` twice.

The loop calls `DatagramBuffer::pop_front`, which already does `self.payload_bytes -= datagram.data.len()` internally, and then the call site subtracts it again:

```rust
if drop {
    while self.conn.datagrams.outgoing.memory_used()
        > self.conn.config.datagram_send_buffer_size
    {
        let prev = self
            .conn
            .datagrams
            .outgoing
            .pop_front()
            .expect("datagrams.outgoing.payload_bytes desynchronized");
        trace!(len = prev.data.len(), "dropping outgoing datagram");
        self.conn.datagrams.outgoing.payload_bytes -= prev.data.len(); // second subtraction
    }
}
```

Each evicted datagram is therefore accounted twice, so under sustained eviction — an outgoing datagram buffer kept over `datagram_send_buffer_size` because the peer is not draining it — `payload_bytes` drifts below its true value and eventually underflows. A debug build hits `attempt to subtract with overflow`; a release build wraps the counter, so `memory_used()` becomes huge, the loop drains the whole queue, and the next `pop_front()` returns `None` and panics on `.expect("datagrams.outgoing.payload_bytes desynchronized")` — inside `send_datagram`, while the connection lock is held, poisoning it for every other task on that connection.

This looks like a leftover from factoring `payload_bytes` bookkeeping into `DatagramBuffer::pop_front`: the manual subtraction at the call site was not removed. `drop_oversized` is fine — it removes via `retain`, not `pop_front`, so its manual subtraction is correct.

`main` already carries the fix: the loop there is `let Some(prev) = self.outgoing.pop_front() else { break };` with no second subtraction (and a graceful break instead of the `.expect`). It does not appear to be on `0.11.x`. Could the fix be backported before the next 0.11.x release?

Deterministic repro (release or debug): a current-thread tokio runtime, then `connection.send_datagram(vec![0u8; 1100].into())` in a loop about 100,000 times with no `.await` in between, so the outgoing buffer never drains and the eviction path runs on every send.

Happy to open a backport PR if that is the more useful form.

vcarus:
Fixes #2805.

`Datagrams::send(drop = true)`'s eviction loop subtracts an evicted datagram's size from `payload_bytes` at the call site even though `DatagramBuffer::pop_front` already does so internally. Under sustained eviction the counter drifts below its true value and underflows: a debug build panics with `attempt to subtract with overflow`, and a release build wraps the counter, drains the whole queue, and then panics on the `.expect("datagrams.outgoing.payload_bytes desynchronized")` inside `send_datagram`, while the connection lock is held.

The fix is deleting the redundant subtraction; eviction semantics are otherwise unchanged. `main` is not affected — its version of the outgoing memory limit (`DatagramState::make_space_for`) never had the manual subtraction, so there is no upstream commit to cherry-pick and this is written directly against `0.11.x`.

The added regression test keeps a small send buffer in sustained eviction and asserts the surviving datagrams are delivered. On current `0.11.x` it fails at the removed line with `attempt to subtract with overflow` (debug) before reaching any assertion.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/quinn-proto/src/connection/datagrams.rs
+++ b/quinn-proto/src/connection/datagrams.rs
@@ -46,7 +46,6 @@ impl Datagrams<'_> {
                     .pop_front()
                     .expect("datagrams.outgoing.payload_bytes desynchronized");
                 trace!(len = prev.data.len(), "dropping outgoing datagram");
-                self.conn.datagrams.outgoing.payload_bytes -= prev.data.len();
             }
         } else if self.conn.datagrams.outgoing.payload_bytes + data.len() + size_of::<Datagram>()
             > self.conn.config.datagram_send_buffer_size
--- a/quinn-proto/src/tests/mod.rs
+++ b/quinn-proto/src/tests/mod.rs
@@ -1947,6 +1947,46 @@ fn datagram_recv_buffer_overflow() {
 }
 
 #[test]
+fn datagram_send_buffer_overflow() {
+    let _guard = subscribe();
+    const PAYLOAD_WINDOW: usize = 100;
+    const METADATA_WINDOW: usize = 2 * size_of::<Datagram>();
+    const WINDOW: usize = PAYLOAD_WINDOW + METADATA_WINDOW;
+    let client_config = {
+        let mut config = client_config();
+        let mut transport = TransportConfig::default();
+        transport.datagram_send_buffer_size(WINDOW);
+        config.transport_config(transport.into());
+        config
+    };
+    let mut pair = Pair::default();
+    let (client_ch, server_ch) = pair.connect_with(client_config);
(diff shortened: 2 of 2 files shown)


Ralith (approved):
Thanks! I knew I should have just backported the intervening changes...

PR #2808, opened 2026-08-30
refactor(proto): Refactor `Assembler::read`

gretchenfrage:
Builds on #2803

See individual commit messages.

### DIFF
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -84,23 +84,23 @@ impl Assembler {
                 self.buffered -= skip;
             }
 
-            if max_length < front.bytes.len() {
-                self.bytes_read += max_length as u64;
+            let chunk = if max_length < front.bytes.len() {
                 let offset = front.offset;
                 front.offset += max_length as u64;
-                self.buffered -= max_length;
                 let bytes = front.bytes.split_to(max_length);
                 self.restore_front_order();
-                return Some(Chunk::new(offset, bytes));
-            }
+                Chunk::new(offset, bytes)
+            } else {
+                self.allocated -= front.allocation_size;
+                let offset = front.offset;
+                let bytes = mem::take(&mut front.bytes);
+                self.data.pop_front();
+                Chunk::new(offset, bytes)
+            };
 
-            self.bytes_read += front.bytes.len() as u64;
-            self.buffered -= front.bytes.len();
-            self.allocated -= front.allocation_size;
-            let offset = front.offset;
-            let bytes = mem::take(&mut front.bytes);
-            self.data.pop_front();
-            return Some(Chunk::new(offset, bytes));
+            self.bytes_read += chunk.bytes.len() as u64;
+            self.buffered -= chunk.bytes.len();
+            return Some(chunk);
         }
     }
 
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -75,9 +75,9 @@ impl Assembler {
                 continue;
             }
 
+            let mut front = self.data.pop_front().unwrap();
             if ordered && self.bytes_read > front.offset {
                 // Advance front to the slice of useful data
-                // Advancing the offset can push the front past data[1]; both exits below fix that
                 let skip = (self.bytes_read - front.offset) as usize;
                 front.bytes.advance(skip);
                 front.offset = self.bytes_read;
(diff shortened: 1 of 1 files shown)


djc (commented):
Overall this looks like a good change, but I don't love the MSRV bump and I don't think the `pop_front_if()` call spanning multiple lines is very readable.

Would you be able to split this into a series of smaller commits, maybe roughly one commit per bullet in the commit message? And maybe just stick with `front()`/`front_mut()` + `pop().unwrap()` for now to avoid the MSRV bump?

gretchenfrage:
> Would you be able to split this into a series of smaller commits, maybe roughly one commit per bullet in the commit message? And maybe just stick with `front()`/`front_mut()` + `pop().unwrap()` for now to avoid the MSRV bump?

Done, ready for review again.

quinn-proto/src/connection/assembler.rs
```diff
@@ -66,7 +66,9 @@ impl Assembler {
                 if front.offset > self.bytes_read {
                     // Next buffer starts after the current read index
                     return None;
-                } else if front.end() <= self.bytes_read {
+                }
+
+                if front.end() <= self.bytes_read {
```
djc, line 71: Personally I prefer the `else if`, because it visually expresses that the second cannot happen if the first happens, and it's more compact.

@Ralith thoughts?

quinn-proto/src/connection/assembler.rs
```diff
@@ -76,10 +76,10 @@ impl Assembler {
                     continue;
                 }
 
-                // Advance front to the slice of useful data
-                let skip = (self.bytes_read - front.offset) as usize;
-                // Advancing the offset can push the front past data[1]; both exits below fix that
-                if skip > 0 {
+                if self.bytes_read > front.offset {
```
djc, line 79: Alternatively, maybe we could do some sort of `checked_sub()` thing above to let the type system carry the invariant? As in:

```rust
let Some(skip) = self.bytes_read.checked_sub(self.bytes_read) {
     return None;
};
```

(Not sure if this makes your subsequent changes harder.)

djc (approved):
Very nice!

quinn-proto/src/connection/assembler.rs
gretchenfrage: That logic isn't quite the same, as we want the condition to be a difference of at least 1, not a difference of at least 0. It also becomes tricky because I eventually combine in an `ordered && ` clause to the front, which doesn't play suuuper nicely with if-let. It also doesn't play super nicely with the `as usize` part.

For comparison, this PR currently makes this code end up like this:

```rs
        if ordered && self.bytes_read > front.offset {
            // Advance front to the slice of useful data
            let skip = (self.bytes_read - front.offset) as usize;
            front.bytes.advance(skip);
            front.offset = self.bytes_read;
            self.buffered -= skip;
        }
```

Trying to use if-let and `checked_sub` turns out like this:

```rs
        if let Some(skip) = self
            .bytes_read
            .checked_sub(front.offset)
            .filter(|&skip| ordered && skip > 0)
        {
            // Advance front to the slice of useful data
            front.bytes.advance(skip as usize);
            front.offset = self.bytes_read;
            self.buffered -= skip as usize;
        }
```

I guess we could utilize the fact that it's fine to run this logic even if `skip` does equal 0 (all the operations it would perform are trivially cheap and logically valid), and turn it into this:

```rs
        if ordered {
            if let Some(skip) = self.bytes_read.checked_sub(front.offset) {
                // Advance front to the slice of useful data
                front.bytes.advance(skip as usize);
                front.offset = self.bytes_read;
                self.buffered -= skip as usize;
            }
        }
```

Which isn't as bad. I'm hesitant, thought. Thoughts between those options?

quinn-proto/src/connection/assembler.rs
gretchenfrage: I can see that benefits of that. However, in this case I think this point is moot, because this PR ultimately transforms the `else`(?)-`if` branch into a `while` loop in a subsequent commit.

Ralith: Reducing the diff in a commit that follows in the same PR is, IMO, sufficient justification for a slightly funny but basically fine structure in an earlier commit.

Ralith (approved):
Great stuff!

quinn-proto/src/connection/assembler.rs
Ralith: I think I *slightly* prefer the final form for the reason @djc gave. Note also that when we move to the 2024 edition we can reduce the indentation using let-chains.

djc: > I think I _slightly_ prefer the final form for the reason @djc gave. Note also that when we move to the 2024 edition we can reduce the indentation using let-chains.

This sounds good to me.

quinn-proto/src/connection/assembler.rs
```diff
@@ -59,56 +59,43 @@ impl Assembler {
...
-                // Advancing the offset can push the front past data[1]; both exits below fix that
-                if start > 0 {
-                    chunk.bytes.advance(start);
-                    chunk.offset += start as u64;
-                    self.buffered -= start;
-                }
-            }
+        while ordered && self.data.front()?.end() <= self.bytes_read {
+            // Next buffer ends before the current read index
+            let front = self.data.pop_front().unwrap();
```
banocean, line 64: I think it could be `let Some(front) = self.data.pop_front()` inside of while condition so it doesn't have unwrap inside of the code

banocean (changes requested):

quinn-proto/src/connection/assembler.rs
Ralith: IMO `unwrap` can sometimes be clearer because it conveys that we always expect `Some`, and ensures someone notices if that expectations is not satisfied.

quinn-proto/src/connection/assembler.rs
gretchenfrage: Since djc already set the edition to 2024 since this review, I simply changed this PR to use this logic:

```rs
        if ordered && let Some(skip) = self.bytes_read.checked_sub(front.offset) {
            // Advance front to the slice of useful data
            front.bytes.advance(skip as usize);
            front.offset = self.bytes_read;
            self.buffered -= skip as usize;
        }
```

PR #2813, opened 2026-09-01
feat(proto): Use linear instead of binary search in `Assembler::restore_front_order`

gretchenfrage:
Builds on #2803.

Factored out from #2808.

Changes `Assembler::restore_front_order` to use a linear search instead of a binary search. This may be faster if the index of the front does not change by much, which is probably the common case, and this remains O(N) either way.

This allows us to remove the hard-coded short-circuit for if its position did not change, which no longer makes sense as an optimization due to it becoming just an implicit degenerate case of the linear search.

### DIFF
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -106,11 +106,8 @@ impl Assembler {
 
     /// Restore ordering after `read` advanced the front's offset
     fn restore_front_order(&mut self) {
-        if self.data.len() < 2 || self.data[0] <= self.data[1] {
-            return;
-        }
         let chunk = self.data.pop_front().unwrap();
-        let idx = self.data.partition_point(|other| *other <= chunk);
+        let idx = self.data.iter().take_while(|other| **other < chunk).count();
         self.data.insert(idx, chunk);
     }
 


quinn-proto/src/connection/assembler.rs
```diff
@@ -102,11 +102,8 @@ impl Assembler {
 
     /// Restore ordering after `read` advanced the front's offset
     fn restore_front_order(&mut self) {
-        if self.data.len() < 2 || self.data[0] <= self.data[1] {
-            return;
-        }
         let chunk = self.data.pop_front().unwrap();
-        let idx = self.data.partition_point(|other| *other <= chunk);
+        let idx = self.data.iter().filter(|other| **other < chunk).count();
```
Ralith: `Iterator::position`?

djc (approved):
LGTM modulo @Ralith's nit.

quinn-proto/src/connection/assembler.rs
gretchenfrage: Since `position` returns an `Option`, that requires an `unwrap_or` added to the end to make it work. If I do that it turns to this:

```rs
        let idx = self
            .data
            .iter()
            .position(|other| *other < chunk)
            .unwrap_or(self.data.len());
```

In addition to it being longer, I think it makes it appear like more of an edge case than it is. Thoughts?

Ralith: I think the intent is still clearer? I don't feel that strongly though, feel free to resolve w/o action if you prefer.

gretchenfrage:
Sorry for the confusion, but I 100% meant for this to use `take_while` instead of `filter`. My comment about "This allows us to remove the hard-coded short-circuit for if its position did not change, which no longer makes sense as an optimization due to it becoming just an implicit degenerate case of the linear search" only makes sense with it being `take_while`, and I think this makes the intent clearer too.

I changed this branch to use `take_while`, could one of you please re-affirm your approval with that in mind?

Ralith (approved):

Ralith:
That does make more sense and read better besides. Embarassing that both of us missed the wasted work under review!

PR #2814, opened 2026-09-01
proto: coalesce contiguous chunks during defragment

lyons-e:
## Summary

Previously, `defragment()` would ignore high utilization chunks, while still capping `MAX_CHUNKS=1024` (Triggered at `COMPACT_THRESHOLD=2048`)
This places a practical limit on `stream_receive_window` of `frame_len * 2048`, or 2.44 MB with an MTU of 1200.

Any peer sending lossless contiguous data with a window larger than this can exceed the limit and get a transport error if the reader lags behind:
```
INTERNAL_ERROR("too many gaps in stream buffer").
```

## Fix

Contiguous chunks are coalesced together during `defragment()` if they are smaller than `min_chunk_size = buffered / MAX_CHUNKS`
This allows us to fit the buffered data while still protecting from floods of out of order data.

We don't want to unconditionally merge chunks, as that could result in many copy operations which would be wasteful.
By only coalescing chunks smaller than `min_chunk_size` we avoid copying chunks which are already a good size for the current buffer. This also protects us from compacting the same chunk multiple times.

### DIFF
--- a/quinn-proto/src/connection/assembler.rs
+++ b/quinn-proto/src/connection/assembler.rs
@@ -1,4 +1,8 @@
-use std::{cmp::Ordering, collections::VecDeque, mem};
+use std::{
+    cmp::{Ordering, max},
+    collections::VecDeque,
+    mem,
+};
 
 use bytes::{Buf, Bytes, BytesMut};
 
@@ -115,6 +119,8 @@ impl Assembler {
     /// This makes sure we're not unnecessarily holding on to many larger allocations.
     /// We merge contiguous chunks in the process of doing so.
     fn defragment(&mut self) {
+        // Chunks smaller than min_chunk_size are merged regardless of fragmentation
+        let min_chunk_size = max(self.buffered.div_ceil(MAX_CHUNKS), MIN_RETAINED_CHUNK_SIZE);
         let mut buffers = mem::take(&mut self.data);
         self.buffered = 0;
         let mut fragmented_buffered = 0;
@@ -124,7 +130,7 @@ impl Assembler {
             let size = chunk.bytes.len();
             offset = chunk.offset + size as u64;
             self.buffered += size;
-            if !chunk.defragmented {
+            if !chunk.defragmented || size < min_chunk_size {
                 fragmented_buffered += size;
             }
         }
@@ -133,11 +139,11 @@ impl Assembler {
         let mut buffer = BytesMut::with_capacity(fragmented_buffered);
         let mut offset = 0;
         for chunk in buffers {
-            if chunk.defragmented {
-                // bytes might be empty after try_mark_defragment
-                if chunk.bytes.is_empty() {
-                    continue;
-                }
+            // bytes might be empty after try_mark_defragment
+            if chunk.bytes.is_empty() {
+                continue;
+            }
+            if chunk.defragmented && chunk.bytes.len() >= min_chunk_size {
                 // The accumulated chunk starts before this one, so it goes first
                 if !buffer.is_empty() {
                     self.data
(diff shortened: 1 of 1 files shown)


quinn-proto/src/connection/assembler.rs
```diff
@@ -117,27 +117,30 @@ impl Assembler {
...
             let size = chunk.bytes.len();
             offset = chunk.offset + size as u64;
             self.buffered += size;
-            if !chunk.defragmented {
-                fragmented_buffered += size;
-            }
         }
         self.allocated = self.buffered;
+        // Coalesce contiguous runs. Chunks smaller than min_chunk_size get merged together
+        let min_chunk_size = (self.buffered / MAX_CHUNKS).max(1);
```
Ralith: nit: a higher minimum might make sense; maybe 128 or something? If you do raise this, lift it into a named constant and document the choice of value with a few words (even if it's just "chosen based on vibes").

quinn-proto/src/connection/assembler.rs
```diff
@@ -117,27 +117,30 @@ impl Assembler {
...
+        let coalesced: usize = buffers
+            .iter()
+            .filter(|chunk| !(chunk.defragmented && chunk.bytes.len() >= min_chunk_size))
+            .map(|chunk| chunk.bytes.len())
+            .sum();
```
Ralith: nit: this change would be smaller if this summation stayed in the loop above (and we'd save a pass over `buffers`, for whatever that's worth)

quinn-proto/src/connection/assembler.rs
```diff
@@ -712,6 +715,20 @@ mod test {
         assert_eq!(x.read(3, false), None);
     }
 
+    #[test]
+    fn ordered_lossless_stream_is_not_rejected() {
+        const FRAME_LEN: usize = crate::INITIAL_MTU as usize;
```
Ralith: Would smaller frames be a more interesting test?

quinn-proto/src/connection/assembler.rs
Ralith: Integer division rounds down. For example, if `self.buffered` is 101 and `MAX_CHUNKS` were 2, and the set of buffers had sizes 50, 50, and 1, then we might still fail. We should round up, and ensure that corner case is covered by a test.

Ralith (changes requested):

quinn-proto/src/connection/assembler.rs
djc: Nit: use a symmetric `max()` like `Ord::max()` instead of the method formulation.

quinn-proto/src/connection/assembler.rs
djc: Nit: turbofish the `sum()` call instead of adding a type annotation.

djc (commented):
Some more nits.

quinn-proto/src/connection/assembler.rs
lyons-e: Added as `MIN_RETAINED_CHUNK_SIZE`
I did a very small benchmark sweep comparing `1, 64, 128, 256` and 128 was the best candidate.

quinn-proto/src/connection/assembler.rs
lyons-e: Removed and implemented in the existing `for` loop.
One consequence of this is now `min_chunk_size` is calculated from the pre defragmented buffer size. This means `min_chunk_size` over estimates if many chunks overlap or contain duplicates, which can result in unnecessary copying.

The trade off is acceptable though, as chunks will generally not get copied again on subsequent compactions.

quinn-proto/src/connection/assembler.rs
lyons-e: Reduced to: 
```rs
const FRAME_LEN: usize = 32
```
This has the added benefit of exercising chunks smaller than `MIN_RETAINED_CHUNK_SIZE`

quinn-proto/src/connection/assembler.rs
lyons-e: Good catch.
Fixed and added a test exercising the boundary

lyons-e: Replaced with symmetric form:
```rs
let min_chunk_size = max(self.buffered.div_ceil(MAX_CHUNKS), MIN_RETAINED_CHUNK_SIZE);
```

quinn-proto/src/connection/assembler.rs
lyons-e: This section is removed entirely, [at the suggestion of another comment](https://github.com/quinn-rs/quinn/pull/2814#discussion_r3910100152)

The check is now done in the original `for` loop

Ralith: Oh yeah, didn't think of that. Seems fine, anyway.

Ralith (approved):
Thanks!

djc (approved):

Ralith:
Backporting in https://github.com/quinn-rs/quinn/pull/2826.

PR #2816, opened 2026-09-01
Add minimum rtt to path stats for connection

### LINKED ISSUE #2815: Expose min_rtt in `PathStats`
Hello 👋🏼, 

The use case here is that sometimes you just need a stable estimate of a peer’s RTT. latest_rtt is the last ACK sample, so it can swing a lot; the  `Path.rtt` still tracks those transients, just more slowly. min_rtt tends to give a stable floor.

somtochiama:
This pull request add  minimum rtt to `PathStats`, useful for users that need a stable floor estimate of rtt.

Closes: #2815


djc (approved):
Seems okay, thanks.

Please rebase instead of adding a merge commit.

Ralith (approved):

PR #2819, opened 2026-09-03
0.11.x: Bundle ACK in packets with DATAGRAM or STREAM frames

anatawa12:
Implements #1899 but only for ACK frames.
This implements bundling ACK for DATAGRAM and STREAEM packet when there are room in packet.

This backports #2747 to 0.11.x.


Ralith (approved):

anatawa12:
I found that `ack_bundled_with_datagrams` test is a little flaky.
The case is client might creates gaps in packet numbers and the server may respond with immediate ACK in case of missing_in_interval.

Simply changing `Pair::default` with `Pair::default_with_deterministic_pns` does not fix the problem since `default_with_deterministic_pns` only makes deterministic pns for server endpoint, but current test expects deterministic client pns. Therefore I think we need to swap server and client role in test code, and make test use `default_with_deterministic_pns`.

Should I update the test on this branch now and then forward port test fix? or should we fix on main branch first?

Ralith:
Either is fine, but perhaps it would be better to revise `default_with_deterministic_pns` to make both endpoints deterministic? I think that was the original intention.

anatawa12:
> but perhaps it would be better to revise default_with_deterministic_pns to make both endpoints deterministic?

I found that it's relatively hard since `TransportConfig` cannot be specified on Endpoint initialization for clients.
`Pair::default_with_deterministic_pns()` with `pair.connect_with(client_config_with_deterministic_pns())` would make both endpoints have deterministic pns. It might be better to rename `default_with_deterministic_pns` and mention `connect_with(client_config_with_deterministic_pns())` in doc comments of `default_with_deterministic_pns`, but I hope this is out of scope of this PR.

I'll update the test to use `connect_with(client_config_with_deterministic_pns())`

Ralith:
That works. There's also `Endpoint::set_default_client_config`.

Ralith (approved):

PR #2821, opened 2026-09-03
fix(udp): don't panic in `UdpSocketState::new` on unbound socket

thomaseizinger:
Currently, `UdpSocketState::new` on Windows learns the socket's address family via `local_addr` (which uses `getsockname` internally). This fails on an unbound socket.

Why would we want to do this on an unbound socket? On Windows (and Apple for that matter), the destination IP in `RecvMeta` gets attached based on the socket setting at the time the datagram arrives. When a socket gets discarded and rebound on the same port, datagrams that are still in-flight and arrive between us calling `bind` and initializing the `UdpSocketState` don't have a destination IP attached.

In Firezone, we have to drop such packets because we use this information to learn about our local interfaces. Constructing the `UdpSocketState` prior to calling `UdpSocket::bind` fixes this race condition but requires us to not use `getsockname` on Windows. On Unix, this is already possible and therefore requires no changes.

Code has been written by Claude. Description and commit messages have been written by me.

### DIFF
--- a/quinn-udp/src/windows.rs
+++ b/quinn-udp/src/windows.rs
@@ -53,12 +53,19 @@ impl UdpSocketState {
         );
 
         socket.0.set_nonblocking(true)?;
-        let addr = socket.0.local_addr()?;
-        let is_ipv6 = addr.as_socket_ipv6().is_some();
+        let info = unsafe {
+            get_socket_option::<WinSock::WSAPROTOCOL_INFOW>(
+                &*socket.0,
+                WinSock::SOL_SOCKET,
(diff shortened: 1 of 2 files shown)


quinn-udp/src/windows.rs
```diff
@@ -484,6 +472,32 @@ fn send(
     }
 }
 
+/// Reads a socket option into a zero-initialised `T`.
```
djc: Nit: US English spelling in code/comments, please.

djc (approved):

Ralith (approved):

PR #2822, opened 2026-09-03
quinn-proto: drop short packets with unsupported versions

vcarus:
`Endpoint::handle` sends a Version Negotiation packet for any long-header packet whose version is not in `supported_versions`, regardless of datagram size.

RFC 9000 §5.2.2: "Servers MUST drop smaller packets that specify unsupported versions." A 15-byte datagram currently elicits a larger VN response, so a spoofed source address can use the server as a small amplifier.

This adds the same `datagram_len < MIN_INITIAL_SIZE` check that already guards Initial packets (§14.1) to the `UnsupportedVersion` branch. `version_negotiate_server` now asserts that the short packet is dropped and that a 1200-byte one still gets a VN response.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -172,6 +172,13 @@ impl Endpoint {
                     debug!("dropping packet with unsupported version");
                     return None;
                 }
+                // RFC 9000 §5.2.2: "Servers MUST drop smaller packets that specify unsupported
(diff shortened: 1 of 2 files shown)


djc (approved):

djc:
Thanks!

Do you want to backport this to `0.11.x`, too?

vcarus:
Sure. I'll do it now.

Ralith (approved):
Thanks!

PR #2584, opened 2026-03-24
fix(unix): disable GSO after probing

thomaseizinger:
Resolves: #2575

PR #2824, opened 2026-09-03
fix(windows): disable USO after probing for GSO support

### LINKED ISSUE #2818: quinn-udp: Windows GSO probe leaves UDP_SEND_MSG_SIZE set on the caller's socket
### Description

#2564 changed the Windows GSO probe to run on the
socket passed to `UdpSocketState::new` instead of on a throwaway socket.
`max_gso_segments` in `quinn-udp/src/windows.rs` sets the
`UDP_SEND_MSG_SIZE` socket option to 1500 and returns 512 when that
succeeds. It does not clear the option again.

The same change on Linux caused #2575, and #2584 fixed it by resetting
`UDP_SEGMENT` to 0 after the probe. The Windows code did not get the
same fix.

After `UdpSocketState::new`, every socket on Windows has segmentation
enabled at the socket level with a segment size of 1500 bytes. A
`Transmit` without a segment size, or with one that
`effective_segment_size` reduces to `None`, does not add a
`UDP_SEND_MSG_SIZE` control message. Winsock then applies the socket
level value and splits the buffer into 1500 byte datagrams.

### Reproduction

On Windows, with quinn-udp 0.6.1:

1. Bind two UDP sockets on 127.0.0.1 and create a `UdpSocketState` for
   the sender.
2. Send a `Transmit` with a 2048 byte payload and `segment_size: None`
   through `UdpSocketState::try_send`.
3. Receive on the other socket with a 4096 byte buffer.

Expected: one datagram of 2048 bytes.
Actual: one datagram of 1500 bytes, followed by one of 548 bytes.

The same steps with quinn-udp 0.5.15 deliver one 2048 byte datagram.

### Proposed fix

Mirror #2584 in `quinn-udp/src/windows.rs`. After the probe succeeds,
reset the option so that segmentation is only requested per message
through the control message.

I can open a pull request with this change if you agree with the
approach.

shitz:
Since #2564, the Windows GSO probe sets `UDP_SEND_MSG_SIZE` to 1500 on
the caller's socket and never clears it. Winsock splits every send
without a `UDP_SEND_MSG_SIZE` control message into 1500 byte datagrams.
Reset the option to 0 after a successful probe, as #2584 did for Linux,
and add a test that sends a 2048 byte datagram and checks that it
arrives in one piece.

Fixes #2818

### DIFF
--- a/quinn-udp/src/windows.rs
+++ b/quinn-udp/src/windows.rs
@@ -567,8 +567,16 @@ fn max_gso_segments(socket: &impl AsRawSocket) -> usize {
         WinSock::UDP_SEND_MSG_SIZE,
         GSO_SIZE,
     ) {
-        // Empirically found on Windows 11 x64
-        Ok(()) => 512,
+        Ok(()) => {
+            // Disable USO again at the socket level so that it is only enabled per message through
+            // the `UDP_SEND_MSG_SIZE` control message. This mirrors the Linux behaviour, see:
+            // - https://github.com/quinn-rs/quinn/issues/2818
+            // - https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-udp-socket-options
+            let _ = set_socket_option(socket, WinSock::IPPROTO_UDP, WinSock::UDP_SEND_MSG_SIZE, 0);
+
+            // Empirically found on Windows 11 x64
+            512
+        }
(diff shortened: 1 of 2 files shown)


djc (approved):

Ralith (approved):
Thanks!

PR #2826, opened 2026-09-03
0.11.x: backport #2814 (proto: coalesce contiguous chunks during defragment)

Ralith:


Ralith:
Nontrivial conflict resolution due to the `VecDeque`/`BinaryHeap` change, but I think it makes sense.

djc (approved):

djc:
> Nontrivial conflict resolution due to the `VecDeque`/`BinaryHeap` change, but I think it makes sense.

Maybe we should also backport that?

Ralith:
> Maybe we should also backport that?

I think we shouldn't, unless the backports become extremely painful. It's a relatively risky logic change with no security implications and marginal performance benefits; it can wait, and shouldn't be tightly coupled with important reliability/security fixes.

PR #2828, opened 2026-09-04
Avoid re-instantiating debug span in ConnectionDriver::poll

stormshield-fabs:
Instantiating and dropping the span in `ConnectionDriver::poll` accounts for around 7% of CPU consumption on a machine with a lot of remote peers and handling heavy traffic.

This MR stores the span in the `ConnectionDriver`, creating it only once.

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -69,7 +69,7 @@ impl Connecting {
         }));
         let connected = Box::pin(conn.shared.connected.clone().notified_owned());
 
-        let driver = ConnectionDriver(conn.clone());
+        let driver = ConnectionDriver::new(conn.clone());
         runtime.spawn(Box::pin(
             async {
                 if let Err(e) = driver.await {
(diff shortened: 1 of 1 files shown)


djc (approved):
Nice, thanks!

I guess you don't care, but we might want to backport this.

stormshield-fabs:
Opened https://github.com/quinn-rs/quinn/pull/2829 with the suggested backport.

djc:
Did you figure out what part of building the span is expensive? Is it locking the connection?

djc:
> Opened #2829 with the suggested backport.

Thanks!

stormshield-fabs:
> Did you figure out what part of building the span is expensive?

We have a bunch of `tracing` layers and it seems that creating and dropping a span requires iterating over all them.

> Is it locking the connection?

No, the lock is necessary to do anything useful in `poll` anyway.

Ralith (approved):

PR #2830, opened 2026-09-04
refactor: Inline `get_or_insert_send` and de-curry `get_or_insert_recv`

gretchenfrage:
See individual commits.

`get_or_insert_send` and `get_or_insert_recv` are two functions which return closures out of an attempt to work well with `map` calls, but I think this just makes things pretty confusing, and the former is also trivially small. Inlines and removes the former, removes the closuring from the second.


Ralith (approved):
That does seem more straightforward; nice!

PR #2833, opened 2026-09-06
refactor(proto): Factor out get_or_init_stream!

gretchenfrage:
Leaving this as a draft for now, because I'm rather uncertain about this and would like to ask @djc and @Ralith what you think of this sort of eager usage of macros.

Four different sites in `SendStream` have logic that looks like:

```rs
        let stream = self
            .state
            .send
            .get_mut(&self.id)
            .ok_or(WriteError::ClosedStream)?
            .get_or_insert_with(|| Send::new(max_send_data));
```

This is repetitive and takes up a lot of lines, but it's tricky to factor out into a function because it borrows from several different fields of `self`, but ultimately constructs a value that only borrows `self.state.send` specifically, and then that value gets used alongside other fields of `self`.

Factoring it out into a macro instead of a function avoids these obstacles, but macros may be considered something to avoid using frivolously. It does shave off 11 lines of code, though, and looks nicer in my opinion.

I'm wondering your thoughts?

### DIFF
--- a/quinn-proto/src/connection/streams/mod.rs
+++ b/quinn-proto/src/connection/streams/mod.rs
@@ -196,6 +196,20 @@ pub struct SendStream<'a> {
     pub(super) conn_state: &'a super::State,
 }
 
+/// Acts like a method on `SendStream`, but is a macro for borrowing reasons. Gets
+/// `self.state.send[self.id]` as an `Option<&mut Send>`, initializing the lazy-initializable
+/// `Option<Box<Send>>` value if present in the `HashMap` but set to `None`.
+macro_rules! get_or_init_stream {
+    ($sendstream:ident) => {{
+        let max_send_data = $sendstream.state.max_send_data($sendstream.id);
+        $sendstream
+            .state
+            .send
+            .get_mut(&$sendstream.id)
+            .map(|opt| &mut **opt.get_or_insert_with(|| Send::new(max_send_data)))
+    }};
+}
+
 #[allow(clippy::needless_lifetimes)] // Needed for cfg(fuzzing)
 impl<'a> SendStream<'a> {
     #[cfg(fuzzing)]
@@ -238,14 +252,7 @@ impl<'a> SendStream<'a> {
 
         let limit = self.state.write_limit();
 
-        let max_send_data = self.state.max_send_data(self.id);
-
-        let stream = self
-            .state
-            .send
-            .get_mut(&self.id)
-            .ok_or(WriteError::ClosedStream)?
-            .get_or_insert_with(|| Send::new(max_send_data));
+        let stream = get_or_init_stream!(self).ok_or(WriteError::ClosedStream)?;
 
         if limit == 0 {
             trace!(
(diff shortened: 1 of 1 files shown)


djc:
What is the minimum you could get using a helper function instead of a macro? It looks like there'd still be a way to reduce it a bit?

gretchenfrage:
> What is the minimum you could get using a helper function instead of a macro? It looks like there'd still be a way to reduce it a bit?

The annoying part is that `max_send_data` borrows the entirety of  `SendStream.state`, but then the expressed `Option<&mut Send>` needs to live contemporaneously with borrows of other fields of `SendStream.state`. So for example, this would not work:

```
fn get_or_init_stream(id: StreamId, state: &mut StreamsState) -> Option<&mut Send> {
    let max_send_data = state.max_send_data(id);
    state
        .send
        .get_mut(&id)
        .map(|opt| &mut **opt.get_or_insert_with(|| Send::new(max_send_data)))
}
```

We _could_ do this:

```rs
fn get_or_init_stream(
    max_send_data: VarInt,
    id: StreamId,
    map: &mut FxHashMap<StreamId, Option<Box<Send>>>,
) -> Option<&mut Send> {
    map.get_mut(&id)
        .map(|opt| &mut **opt.get_or_insert_with(|| Send::new(max_send_data)))
}
```

This fails to factor out the `let max_send_data = ...` line, so the call sites end up like this:

```rs
        let stream = get_or_init_stream(max_send_data, self.id, &mut self.state.send)
            .ok_or(WriteError::ClosedStream)?;
```

For comparison, in this current macro-based version of the PR, the call sites end up like this (one line instead of 3):

```rs
        let stream = get_or_init_stream!(self).ok_or(FinishError::ClosedStream)?;
```

I guess another way we could do it would be using this currying-based approach, which solves the borrowing issues and also factors out the `max_send_data` call without using macros:

```rs
    fn get_or_init_stream(
        &self,
    ) -> impl FnOnce(&mut FxHashMap<StreamId, Option<Box<Send>>>) -> Option<&mut Send> {
        let id = self.id;
        let max_send_data = self.state.max_send_data(id);
        move |map| {
            map.get_mut(&id)
                .map(|opt| &mut **opt.get_or_insert_with(|| Send::new(max_send_data)))
        }
    }
```

It makes its call sites end up like this:

```rs
        let stream =
            self.get_or_init_stream()(&mut self.state.send).ok_or(FinishError::ClosedStream)?;
```

I'm not too enthused about that one, I think it's complex.

Ralith:
I don't feel super strongly about any of these options, including the status quo, TBH. Repetition is mainly a hazard when it degrades readability (only mild impact here) and represents a maintenance hazard from the risk of some sites not getting updated in a refactor (no obvious hazard here).

The usual refrain when there's borrow-checker problems like this is "your data is organized wrong." Would it make sense to move the fields accessed by `max_send_data` together with the `send` and `recv` tables into a helper struct with methods to get-or-init send/recv, so that we only hold a borrow to that helper struct rather than all of `StreamsState`?

PR #2834, opened 2026-09-06
refactor: Remove `write_source` (`write_limit`/`mark_blocked`/`write_immediate`)

gretchenfrage:
This PR is intended to supersede my previous attempt to achieve these goals in #2242, #2230, and #2260. It stems from [this comment](https://github.com/quinn-rs/quinn/pull/2260#issuecomment-2952787731) in that PR where me and Ralith were discussing the idea of deleting the `WriteSource` abstraction within proto and replacing it with these methods:

```rs
    /// Get how many bytes could be written immediately
    ///
    /// Always returns `Ok(0)` if blocked, never returns `WriteError::Blocked`.
    pub fn write_limit(&self) -> Result<usize, WriteError> { .. }

    /// Ensure that a [`StreamEvent::Writable`][1] event is emitted once [`write_limit`][2]
    /// transitions from `Ok(0)` to a non-zero value
    ///
    /// Panics if `self.write_limit() != Ok(0)`.
    ///
    /// [1]: crate::StreamEvent::Writable
    /// [2]: Self::write_limit
    pub fn mark_blocked(&mut self) { .. }

    /// Immediately write the entirety of `chunk` on the given stream, or panic if unable
    ///
    /// Requires that `self.write_limit().unwrap() >= chunk.len()`, panics otherwise.
    pub fn write_immediate(&mut self, chunk: Bytes) { .. }
```

This refactor shaves off over 100 lines of code, and results in an architecture with less indirection and more up-front guarantees. Moreover, in my 54cf579df59d57bf446979e28fb7b706de70cb7f commit (not included in this PR) I prototype how this functionality could be usefully exposed on the proto layer and even also on the Quinn layer with a lock-based API, which was also discussed in the linked thread.

This PR implements this refactor as an actual pure refactor, without introducing any new external API. It's done as a series of smaller commits, each of which leaves the code in a fully functional state. Unfortunately, most of the commits make the code slightly larger, until we get to the final one, at which point we can delete hundreds of lines of code. I'm not sure if I can do better than that, as this PR sort of tunnels from which architectural local minima to another. However, each individual commit is designed to be as small and mechanical as possible so that they can be successfully reviewed.


Ralith:
Thanks for picking this back up!

gretchenfrage:
> Thanks for picking this back up!

Thanks for not removing me from the maintainer list when I disappeared for a year 😅

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -281,10 +287,29 @@ impl<'a> SendStream<'a> {
...
+            self.mark_blocked();
             return Err(WriteError::Blocked);
         }
 
+        let max_send_data = self.state.max_send_data(self.id);
+        let stream = self
+            .state
+            .send
+            .get_mut(&self.id)
+            .ok_or(WriteError::ClosedStream)?
```
Ralith: Should we `unwrap` here for the same reason that we can in `mark_blocked`? I'm fine with being defensive, but maybe leave a brief comment that explains that this *shouldn't* arise.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -308,40 +325,32 @@ impl<'a> SendStream<'a> {
         Ok(limit)
     }
 
-    fn write_source<B: BytesSource>(&mut self, source: &mut B) -> Result<Written, WriteError> {
-        let mut write_limit = self.write_limit_or_mark_blocked()?;
+    /// Immediately write the entirety of `chunk` on the given stream, or panic if unable
```
Ralith, line 328: "unable" is vague, and it's not immediately obvious whether the remaining text fully defines it. Maybe reword "on the given stream, or panic if unable" to "which must fit within `write_limit`"?

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -227,7 +228,23 @@ impl<'a> SendStream<'a> {
...
+            self.write_immediate(prefix);
+
+            if chunk.is_empty() {
+                written.chunks += 1;
+            }
+
+            debug_assert!(written.bytes <= write_limit);
+            if written.bytes == write_limit {
+                break;
+            }
```
Ralith, line 245: nit: I think the logic here would be a little easier to follow if there was a single if/else, rather than two seemingly but not actually independent branches.

Ralith (approved):
This looks great, thank you! Super happy to see this new, exceptionally composable API concept land.

Some nits:
- Multiple commit descriptions mention "WriteSource"; I think they meant `BytesSource`.
- You also don't need to justify the motivation separately in every commit -- the PR text is a good place for that discussion.

quinn-proto/src/connection/streams/send.rs
```diff
@@ -49,18 +49,26 @@ impl Send {
...
-    pub(super) fn write<S: BytesSource>(
-        &mut self,
-        source: &mut S,
-        limit: u64,
-    ) -> Result<Written, WriteError> {
+    /// Get how many bytes could be written immediately while respecting stream-level limits
+    ///
+    /// Always returns `Ok(0)` if blocked, never returns `WriteError::Blocked`. Returning `Err`
+    /// indicates that this stream cannot ever be written on again.
+    fn write_limit(&self) -> Result<u64, WriteError> {
```
djc, line 56: Nit: `write_limit()` should be defined below `write()`.

quinn-proto/src/connection/streams/send.rs
```diff
@@ -49,18 +49,26 @@ impl Send {
...
-        let budget = self.max_data - self.pending.offset();
+        Ok(self.max_data - self.pending.offset())
+    }
+
+    pub(super) fn write<S: BytesSource>(
+        &mut self,
+        source: &mut S,
+        limit: u64,
+    ) -> Result<Written, WriteError> {
+        let budget = self.write_limit()?;
```
djc, line 71: Nit, maybe something like

```rust
let budget = match self.write_limit()? {
     0 => return Err(WriteError::Blocked),
     budget => budget,
};
```

or maybe the conversion of 0 to `WriteError::Blocked` should happen inside `write_limit()`?

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -236,7 +236,7 @@ impl<'a> SendStream<'a> {
             return Err(WriteError::Blocked);
         }
 
-        let limit = self.state.write_limit();
+        let connection_write_limit = self.state.write_limit();
```
djc, line 239: Since we're inside `write()` (and this commit also changes code in the `send` module), how about we drop the `_write` stanza from these variable names?

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -239,6 +239,16 @@ impl<'a> SendStream<'a> {
...
         let max_send_data = self.state.max_send_data(self.id);
+        let stream = self
+            .state
+            .send
+            .get(&self.id)
+            .ok_or(WriteError::ClosedStream)?;
+        let stream_write_limit = stream
+            .as_ref()
+            .map(|stream| stream.write_limit())
+            .unwrap_or_else(|| Ok(max_send_data.into()))?;
```
djc, line 250: Nit: not a fan of `map().unwrap_or_else()` constructions -- IME `match` expressions are usually better for these cases.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -230,11 +230,13 @@ impl<'a> SendStream<'a> {
...
-        if self.conn_state.is_closed() {
-            trace!(%self.id, "write blocked; connection draining");
-            return Err(WriteError::Blocked);
-        }
+    /// Get how many bytes could be written immediately
+    ///
+    /// It is invalid to call this if `self.conn_state.is_closed()`.
+    ///
+    /// Always returns `Ok(0)` if blocked, never returns `WriteError::Blocked`.
+    fn write_limit(&self) -> Result<usize, WriteError> {
```
djc, line 238: Nit: `write_limit()` should be moved below `write_source()`.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -254,21 +254,27 @@ impl<'a> SendStream<'a> {
...
-            trace!(%self.id, "write blocked; connection draining");
-            return Err(WriteError::Blocked);
-        }
+    /// Ensure that a [`StreamEvent::Writable`] event is emitted once [`write_limit`][2]
+    /// transitions from `Ok(0)` to a non-zero value
+    ///
+    /// It is invalid to call this if `self.write_limit() != Ok(0)`.
+    ///
+    /// [2]: Self::write_limit
+    fn mark_blocked(&mut self) {
```
djc, line 263: Nit: `mark_blocked()` should be moved below `write_source()`.

quinn-proto/src/connection/streams/mod.rs
```diff
@@ -254,21 +254,27 @@ impl<'a> SendStream<'a> {
...
         let connection_write_limit = self.state.write_limit();
-        let mut write_limit = self.write_limit()?;
 
         let max_send_data = self.state.max_send_data(self.id);
         let stream = self
             .state
             .send
             .get_mut(&self.id)
-            .ok_or(WriteError::ClosedStream)?
+            .unwrap() // Unwrap safety: write_limit() would have returned Err(ClosedStream)
```
djc, line 277: So can we express this differently so we can keep the borrow from `write_limit()`?

djc (commented):
Great direction -- +103/-238 is always welcome!

You factored out a bunch of methods which end up above their caller. These should end up being below their callers in the final commit (if both still exist), though ideally if that makes the diffs smaller moving them should be in a separate commit.

I'm a little wary of the `unwrap()` calls -- it would be nice to do some work to figure out if there's a different organization of the data that would allow borrowing to work without the unwrapping.

quinn-proto/src/connection/streams/send.rs
gretchenfrage: > or maybe the conversion of 0 to `WriteError::Blocked` should happen inside `write_limit()`?

Perhaps, although I am hesitant because `write_limit`'s note about "Returning `Err` indicates that this stream cannot ever be written on again" is an elegant seeming property that would be lost by doing so.

PR #2835, opened 2026-09-07
proto: release the send window when rejecting 0-RTT

zerox80:
Rejecting 0-RTT discards the outgoing streams and packets but leaves their bytes in `StreamsState::unacked_data`. Those bytes keep consuming the connection's send window, so 1-RTT writes can remain blocked even after the handshake succeeds.

Reset the outstanding byte count alongside the other outgoing stream state. The existing `zero_rtt_rejection` case now fills a small send window with early data and checks that the same capacity is available for 1-RTT traffic after rejection.

### DIFF
--- a/quinn-proto/src/connection/streams/state.rs
+++ b/quinn-proto/src/connection/streams/state.rs
@@ -247,6 +247,7 @@ impl StreamsState {
         self.pending.clear();
         self.send_streams = 0;
         self.data_sent = 0;
+        self.unacked_data = 0;
         self.connection_blocked.clear();
     }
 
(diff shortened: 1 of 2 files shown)


djc (approved):
LGTM, thanks!

djc:
Would you be able to backport this to the `0.11.x` branch?

Curious to hear more about what you're using Quinn for!

zerox80:
> Would you be able to backport this to the `0.11.x` branch?
> 
> Curious to hear more about what you're using Quinn for!

Sure, I can do a backport.

I’ve made quite a few patches in my own fork for a VPN protocol based on HTTP/3/QUIC. I also used Codex to put together a few PRs, and the results were surprisingly good: …

I closed them for now because I haven’t properly reviewed them yet, and they were opened automatically without me explicitly confirming it. I’ll probably reopen them soon.

Ralith (approved):
Nice catch!

PR #2836, opened 2026-09-07
quinn: refresh the endpoint sender after rebinding

zerox80:
`Endpoint::rebind_abstract` replaces the receive socket and updates existing connections, but leaves the endpoint's response sender attached to the previous socket. A Retry response therefore leaves from the old address and the retried handshake cannot complete through the rebound endpoint.

Create a new endpoint sender from the replacement socket. The included regression case rebinds a server, requests Retry and checks completion of the retried handshake.


djc (approved):

djc:
Makes sense, thanks!

(Want to backport this, too?)

Ralith (approved):

PR #2837, opened 2026-09-07
proto: avoid duplicate flow-control credit for stopped streams

zerox80:
Stopping a receive stream already returns connection-level read credit for the unread data up to its highest offset. A later `RESET_STREAM` currently returns credit for those bytes again because it subtracts the assembler's read offset.

For stopped streams, calculate the remaining credit from the highest offset instead. The existing `stopped_reset` case covers final offsets equal to and greater than the received offset, checking both `data_recvd` and the advertised connection limit.

### DIFF
--- a/quinn-proto/src/connection/streams/state.rs
+++ b/quinn-proto/src/connection/streams/state.rs
@@ -342,6 +342,8 @@ impl StreamsState {
         let bytes_read = rs.assembler.bytes_read();
         let stopped = rs.stopped;
         let end = rs.end;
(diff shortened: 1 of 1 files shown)


Ralith (approved):
Nice catch, thanks! Please rebase rather than merging.

zerox80:
> Nice catch, thanks! Please rebase rather than merging.

sure will do next time. 0.11.x has this patch already, so i won't open a PR for backport

djc (approved):

PR #2838, opened 2026-09-07
proto: preserve the server config for pending incoming connections

zerox80:
A pending `Incoming` still depends on the endpoint's current server configuration for buffering, default acceptance and Retry. Calling `set_server_config(None)` can therefore panic when that attempt is subsequently handled; replacing the configuration can also change the TLS configuration of an attempt that already arrived.

Retain the original configuration in `IncomingBuffer` and use it for those operations. An explicit configuration passed to `accept` still takes precedence. The included regression cases cover buffering and accepting after removal or replacement of the configuration, and issuing Retry after disabling new incoming connections.

This touches some of the incoming-buffer code also changed by #2782. It additionally preserves the default acceptance and Retry configuration for attempts that are still pending.

### DIFF
--- a/quinn-proto/src/endpoint.rs
+++ b/quinn-proto/src/endpoint.rs
@@ -90,6 +90,8 @@ impl Endpoint {
     }
 
     /// Replace the server configuration, affecting new incoming connections only
+    ///
+    /// Pending incoming connections retain the configuration active when they first arrived.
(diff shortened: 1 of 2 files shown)


djc (approved):
This makes sense to me, thanks.

Ralith (approved):

PR #2839, opened 2026-09-07
proto: prune queued DATAGRAMs after path MTU changes

zerox80:
Queued DATAGRAMs can stop fitting after `path_changed` or migration resets the path MTU. Cleanup currently runs only after MTU black-hole detection, leaving those payloads at the front of the queue and senders waiting for buffer space.

This PR also runs cleanup after an explicit path reset or migration, wakes blocked senders when queued datagrams are dropped, and preserves payloads that exactly match the maximum allowed size.

The three commits separate the cleanup refactor, the payload-limit correction, and the path-change handling. Regression tests cover the exact limit, queue capacity, sender notification, and delivery of a smaller DATAGRAM after both a path reset and migration. The migration test delivers packets inline before polling server transmits, so its assertions observe the cleanup before sending can free buffer space.

Validation:

- `cargo test --locked -p quinn-proto`: 306 unit tests and 3 documentation tests passed.
- `cargo clippy --locked -p quinn-proto --all-targets -- -D warnings`: passed.
- `cargo fmt --all -- --check` and `git diff --check`: passed.

Prepared with AI assistance.

### DIFF
--- a/quinn-proto/src/connection/datagrams.rs
+++ b/quinn-proto/src/connection/datagrams.rs
@@ -4,7 +4,7 @@ use bytes::Bytes;
 use thiserror::Error;
 use tracing::{debug, trace};
 
-use super::Connection;
+use super::{Connection, Event};
 use crate::{
     TransportError,
     frame::{Datagram, FrameStruct},
@@ -50,6 +50,21 @@ impl Datagrams<'_> {
         }
         self.conn.datagrams.outgoing.push_back(Datagram { data });
         Ok(())
+    }
+
+    /// Discard queued datagrams that no longer fit the active path, waking blocked senders.
+    pub(super) fn drop_oversized(&mut self) {
+        let Some(max_datagram_size) = self.max_size() else {
+            return;
+        };
+        if !self.conn.datagrams.drop_oversized(max_datagram_size)
+            || !self.conn.datagrams.send_blocked
+        {
+            return;
+        }
+
+        self.conn.datagrams.send_blocked = false;
+        self.conn.events.push_back(Event::DatagramsUnblocked);
     }
 
     /// Compute the maximum size of datagrams that may passed to `send_datagram`
--- a/quinn-proto/src/connection/datagrams.rs
+++ b/quinn-proto/src/connection/datagrams.rs
@@ -174,7 +174,7 @@ impl DatagramState {
     pub(super) fn drop_oversized(&mut self, max_payload: usize) -> bool {
         let mut dropped_any = false;
         self.outgoing.queue.retain(|datagram| {
-            let result = datagram.data.len() < max_payload;
+            let result = datagram.data.len() <= max_payload;
             if !result {
                 trace!(
                     "dropping {} byte datagram violating {} byte limit",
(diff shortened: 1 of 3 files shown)


quinn-proto/src/connection/datagrams.rs
```diff
@@ -4,12 +4,24 @@ use bytes::Bytes;
...
 use tracing::{debug, trace};
 
-use super::Connection;
+use super::{Connection, Event};
 use crate::{
     TransportError,
     frame::{Datagram, FrameStruct},
 };
 
+impl Connection {
```
djc: I don't think we want a separate `Connection` impl? block here. Can this just be moved into the `Datagrams` impl block?

Also, suggest to use let-else to linearize this:

```rust
let Some(max_datagram_size) = self.datagrams.max_size() else {
     return;
};

if !self.datagrams.drop_oversized(max_datagram_size) || !self.datagrams.send_blocked {
     return;
}
```

quinn-proto/src/tests/mod.rs
```diff
@@ -3842,6 +3842,64 @@ fn ack_bundled_with_datagrams() {
...
+            pair.server_conn_mut(server_ch).path_changed(now);
+        }
+
+        assert!(pair.server_datagrams(server_ch).max_size().unwrap() < old_max);
+        assert_eq!(
+            pair.server_datagrams(server_ch).send_buffer_space(),
+            empty_space
+        );
+        assert!(
+            std::iter::from_fn(|| pair.server_conn_mut(server_ch).poll())
```
djc: Nit: import `iter` at the top of the module.

quinn-proto/src/tests/util.rs
```diff
@@ -406,6 +406,16 @@ impl TestEndpoint {
         }
     }
 
+    /// Deliver queued connection events without transmitting any packets.
+    pub(super) fn drive_conn_events(&mut self) {
```
djc: This should be added only if/when there are some (preferably more than one) callers for it.

zerox80: There was already one caller in the migration branch of `path_changes_unblock_oversized_datagrams`. The test needs to deliver the connection event before polling server transmits, so the buffer-space and unblock assertions check the MTU cleanup before sending can free any space.

A separate helper isn't needed for that single use, though. Packet delivery is now inline in the test, and the helper commit has been removed.

quinn-proto/src/connection/datagrams.rs
```diff
@@ -93,6 +93,21 @@ impl Datagrams<'_> {
             .saturating_sub(self.conn.datagrams.outgoing.payload_bytes)
             .saturating_sub(size_of::<Datagram>())
     }
+
+    /// Discard queued datagrams that no longer fit the active path, waking blocked senders.
+    pub(super) fn drop_oversized(&mut self) {
```
djc: Nit: move this directly below `send()`, please.

djc (approved):
Thanks!

(Looks like this could also be backported...)

Ralith (approved):

PR #2849, opened 2026-09-08
[0.11.x] proto: prune queued DATAGRAMs after path MTU changes

zerox80:
Queued DATAGRAMs can stop fitting after `path_changed()` or peer migration lowers the path MTU, leaving the send buffer occupied and `send_datagram_wait()` callers blocked. Prune oversized payloads after these path changes, notify blocked senders when space is freed, and retain payloads exactly at the current size limit.

Backports #2839 at 0888165de743a589a53a5945f562ef84aa2a8a14 together with prerequisite commit c09ff210125b0ebecbe9d74bba2ae401322e32ef, which adds the missing sender notification after MTU black-hole cleanup. The prerequisite is adapted to the `0.11.x` queue and memory accounting.

The four commits separate the prerequisite, cleanup refactor, payload-limit correction, and path-change handling. Cleanup is a private method on `Datagrams`. The migration test delivers packets inline before polling server transmits, so it checks queue cleanup before sending can free buffer space.

Regression coverage includes explicit path reset, peer migration, exact-limit retention, and sender notification after MTU black-hole detection.

Validation on the updated backport branch:

- `cargo test --locked -p quinn-proto`: 285 unit tests and 3 documentation tests passed.
- `cargo clippy --locked -p quinn-proto --all-targets -- -D warnings`: passed.
- `cargo fmt --all -- --check` and `git diff --check`: passed.
- Rust 1.85.0: `cargo check --locked --lib -p quinn-udp -p quinn-proto -p quinn` passed.

Local validation limits: the Rust 1.85.0 check with `--all-features` could not complete because the AWS-LC FIPS build requires Go, which is not installed. Running tests with Rust 1.85.0 is blocked by the existing lockfile's `time` and `time-core` test dependencies, which require Rust 1.88; the tests above passed with Rust 1.98.0. The PR CI includes the full MSRV library check with all features.

Prepared with AI assistance.

### DIFF
--- a/quinn-proto/src/connection/datagrams.rs
+++ b/quinn-proto/src/connection/datagrams.rs
@@ -182,7 +182,7 @@ impl DatagramState {
     pub(super) fn drop_oversized(&mut self, max_payload: usize) -> bool {
         let mut dropped_any = false;
         self.outgoing.queue.retain(|datagram| {
-            let result = datagram.data.len() < max_payload;
+            let result = datagram.data.len() <= max_payload;
             if !result {
                 trace!(
                     "dropping {} byte datagram violating {} byte limit",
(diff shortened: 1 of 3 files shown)


djc (approved):
LGTM modulo review feedback from the original PR.

Ralith (approved):

PR #2853, opened 2026-09-08
fix: make `handshake_data` cancellation-safe

robertkiel:
### Problem description

`quinn` exposes a method to read the negotiated parameters of a connection, [`handshake_data`](https://docs.rs/quinn/latest/quinn/struct.Connecting.html#method.handshake_data). Due to an inconsistent state change, the method is not cancellation-safe. 

The method first takes out the boxed future and then awaits it. Polling again the same future may lead to a panic.

```rust
if let Some(x) = self.handshake_data_ready.take() {
    let _ = x.await;
}
```

This can be easily fixed by taking out the value _after_ awaiting the future.

```rust
if let Some(x) = self.handshake_data_ready.as_mut() {
    let _ = x.await;
    // Once the data is ready, apply state changes. This prevents a panic due to
    // inconsistent state when retrying a call to `handshake_data`.
    self.handshake_data_ready = None;
}
```

### DIFF
--- a/quinn/src/connection.rs
+++ b/quinn/src/connection.rs
@@ -153,12 +153,17 @@ impl Connecting {
     /// [`Session`](proto::crypto::Session). For the default `rustls` session, the return value can
     /// be [`downcast`](Box::downcast) to a
     /// [`crypto::rustls::HandshakeData`](crate::crypto::rustls::HandshakeData).
+    ///
+    /// This operation is cancel-safe.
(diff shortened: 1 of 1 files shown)


Ralith (approved):
Nice catch, thanks!

djc (approved):

djc:
Would you be able to backport this to the `0.11.x` branch?

PR #2847, opened 2026-09-07
proto: send DATA_BLOCKED and STREAM_DATA_BLOCKED when blocked by flow control

heitzlki:
We parse DATA_BLOCKED and STREAM_DATA_BLOCKED but never send them. Now a write refused by the peer's connection or stream limit queues the matching frame ([RFC 9000 4.1](https://www.rfc-editor.org/rfc/rfc9000#section-4.1)). Our own send_window doesn't count. Same pattern as STREAMS_BLOCKED in #2579.

The periodic re-send from 4.1 is left for a follow-up.

Closes #658

Issue #658, opened 2020-03-05
Send (STREAM_)DATA_BLOCKED periodically while blocked by flow control


Ralith:
In addition to diagnostic convenience, this ensures a blocked connection won't time out.

samuel-baron:
I would like to try this issue, is anyone working on this?

djc:
@samuel-baron feel free to try, as long as you're able to do so in a way that complies with the AI policy laid out in `CONTRIBUTING.md`.

heitzlki:
Had a go at this, PR is #2847, same approach as STREAMS_BLOCKED in #2579. 
@samuel-baron sorry if you had something in progress, happy to compare details.

PR #2855, opened 2026-09-08
Repeat blocked frames

### LINKED ISSUE #2854: Repeat blocked frames periodically while flow control limited
#2847 sends DATA_BLOCKED / STREAM_DATA_BLOCKED once and on loss, but not periodically as [RFC 9000 4.1](https://www.rfc-editor.org/info/rfc9000/#section-4.1) suggests to keep a stalled connection from idle-timing out. #658 was about this but got closed by #2847.

I have a patch: a timer armed once the frame is acked and nothing is in flight, firing at half the idle timeout. Nothing when there's no idle timeout. s2n-quic uses PTO plus backoff instead, happy to do that if you prefer. Will open a PR unless someone objects.

heitzlki:
Follow-up to #2847, adds the periodic part of [RFC 9000 4.1](https://www.rfc-editor.org/info/rfc9000/#section-4.1) that #658 was about. When a blocked frame is acked and nothing is in flight, a timer repeats it at half the idle timeout until the limit is raised. No timer when there's no idle timeout. 

Closes #2854


heitzlki:
Rebased on main and squashed two fixes into the second commit: the blocked timer is re-armed after a loss detection timeout too, and the limit-raised test no longer samples the count while a repetition is due, which made it flaky.

quinn-proto/src/connection/mod.rs
```diff
@@ -1610,6 +1616,7 @@ impl Connection {
         }
 
         self.set_loss_detection_timer(now);
+        self.set_blocked_timer(now);
```
Ralith, line 1625: Please document why this is the main entry point for arming the blocked timer.

quinn-proto/src/connection/mod.rs
```diff
@@ -1233,8 +1233,14 @@ impl Connection {
...
                     self.ping();
                 }
+                Timer::Blocked => {
+                    trace!("repeating blocked frames");
+                    self.streams
+                        .queue_blocked(&mut self.spaces[SpaceId::Data].pending);
+                }
                 Timer::LossDetection => {
                     self.on_loss_detection_timeout(now);
+                    self.set_blocked_timer(now);
```
Ralith, line 1246: Document the importance of this. Won't a tail loss probe trigger an ACK which rearms the timer anyway?

quinn-proto/src/connection/streams/state.rs
Ralith, line 810: Should we prune blocked_streams here too, perhaps using the `connection_blocked` list?

quinn-proto/src/connection/streams/state.rs
```diff
@@ -421,10 +425,24 @@ impl StreamsState {
...
+    /// Queue a blocked frame again for every limit that is still holding back a write
+    pub(crate) fn queue_blocked(&mut self, pending: &mut Retransmits) {
+        pending.data_blocked |= self.can_send_data_blocked();
+        let send = &self.send;
+        self.blocked_streams.retain(|id| {
+            send.get(id)
+                .and_then(|s| s.as_ref())
+                .is_some_and(|s| s.blocked_limit().is_some())
+        });
+        pending.stream_data_blocked.extend(&self.blocked_streams);
```
Ralith, line 447: Traversing all elements in a `HashMap` is O(capacity). Is that appropriate here?

Ralith (changes requested):

quinn-proto/src/connection/streams/state.rs
heitzlki: I don't think so: blocked_streams only holds streams stuck at their own stream limit, and MAX_DATA doesn't move those, so no entry becomes stale here. They're removed when MAX_STREAM_DATA raises the stream's limit, when the stream is freed, and lazily by the retain. Streams stuck on the connection limit are the connection_blocked list's case and repeat via data_blocked_limit instead.

quinn-proto/src/connection/streams/state.rs
heitzlki: It only runs when the blocked timer fires, at most once per idle_timeout/2, and only while blocked with nothing in flight, and entries are removed eagerly on MAX_STREAM_DATA and stream_freed, so the set stays small. 
Added a comment. 
Happy to switch to a Vec plus a flag like connection_blocked if you'd rather avoid the scan

quinn-proto/src/connection/mod.rs
heitzlki: A tail loss probe does. The gap is the time-threshold branch, which sends no probe: if the packet it writes off carried nothing retransmittable (e.g. a lone PING), nothing goes out and nothing is left in flight, so no ACK arrives to arm it. Added a comment.

PR #2856, opened 2026-09-09
fix: stale waker in `blocked_readers`

robertkiel:
# Problem description

`Connection`s maintain a list of blocked readers in which they store `Waker` instances so they can get notified once the streams get unblocked or closed and the `Future` can make progress.

If a stream is considered done, i.e. the property `all_data_read` is set, the stream may be dropped. The `Drop` implementation distinguishes streams with `all_data_read` property being set and returns early.

```rust
if self.all_data_read {
    debug_assert!(
        !self
            .conn
            .state
            .lock("RecvStream:drop")
            .blocked_readers
            .contains_key(&self.stream),
        "Stream {} should not have a blocked reader when all data read is true",
        self.stream
    );
    return;
}
```

However, `poll_read_generic` sets `all_data_read` when streams have `ReadStatus::Finished` and `ReadStatus::Failed(None, Reset(_))`. This leaves stale entries in `blocked_readers` and drains memory unnecessarily.

Even if the `Waker` received a `wake` call, the `poll` function will return early.

```rust
if self.all_data_read {
    return Poll::Ready(Ok(None));
}
```

# Code changes

- Remove stale waker from `blocked_readers` after the `all_data_read` property got set.
- Use in-memory manipulation with [`HashMap::entry(_)`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry) and [`Waker::clone_from(_)`](https://doc.rust-lang.org/std/task/struct.Waker.html#method.clone_from) when updating wakers in `blocked_readers`

### DIFF
--- a/quinn/src/recv_stream.rs
+++ b/quinn/src/recv_stream.rs
@@ -364,6 +364,8 @@ impl RecvStream {
         T: FnMut(&mut Chunks<'_>) -> ReadStatus<U>,
     {
         use proto::ReadError::*;
+        use std::collections::hash_map::Entry;
+
         if self.all_data_read {
             return Poll::Ready(Ok(None));
         }
@@ -392,6 +394,8 @@ impl RecvStream {
             ReadStatus::Readable(read) => Poll::Ready(Ok(Some(read))),
             ReadStatus::Finished(read) => {
                 self.all_data_read = true;
+                // Stream is dead, cleanup stale waker
+                conn.blocked_readers.remove(&self.stream);
                 Poll::Ready(Ok(read))
             }
             ReadStatus::Failed(read, Blocked) => match read {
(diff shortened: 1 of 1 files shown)


quinn/src/recv_stream.rs
```diff
@@ -392,22 +394,33 @@ impl RecvStream {
...
+                        Entry::Occupied(mut entry) => {
+                            entry.get_mut().clone_from(cx.waker());
+                        }
+                        Entry::Vacant(entry) => {
+                            entry.insert(cx.waker().clone());
+                        }
+                    }
                     Poll::Pending
                 }
+                val => Poll::Ready(Ok(val)),
```
djc: This seems like an unrelated/random change?

djc (commented):
Mostly LGTM -- please rebase and squash all your commits.

quinn/src/recv_stream.rs
robertkiel: the idea was to use the same style in the whole code block. another arm had this pattern, this is why I applied it here as well. should be no functional change.

djc: Please keep it as is to minimize the diffs for review.

robertkiel:
squashed the commits and revoked the unrelated change

djc (approved):
Thanks!

Would you be able to backport these changes to `0.11.x`, as well?

Ralith (approved):
Thanks!

PR #2857, opened 2026-09-09
fix: cleanup state if transmit failed

robertkiel:
# Problem description

When polling a `Connecting` struct in order to yield `Connection`, transmitting data via UDP may fail. Hence, within `<ConnectionDriver as Future>::poll(_)` the call to `drive_transmit` can fail. Due to the `?` operator, the function immediately returns the error without cleaning up the state, which in this case means closing the connection.

```rust
fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
    // ...
    let mut keep_going = conn.drive_transmit(cx)?; // exits without cleanup
    // ...
}
```

# Fix

In the case no data can be transmitted, close the connection.


quinn/src/connection.rs
```diff
@@ -264,7 +259,17 @@ impl Future for ConnectionDriver {
             conn.terminate(e, &self.conn.shared);
             return Poll::Ready(Ok(()));
         }
-        let mut keep_going = conn.drive_transmit(cx)?;
+        let mut keep_going = match conn.drive_transmit(cx) {
```
Ralith, line 267: nit: since the `Err` case diverges, this would be more concise as a let-else.

Ralith (approved):
Not sure this is actually reachable, but if it were, closing the connection would certainly be better than turning it into a zombie.

quinn/src/connection.rs
```diff
@@ -264,7 +259,17 @@ impl Future for ConnectionDriver {
...
+            Ok(keep_going) => keep_going,
+            Err(e) => {
+                // Transmit failed, so close the connection and clean state before returning
+                // the error.
+                if !conn.inner.is_closed() {
+                    conn.implicit_close(&self.conn.shared);
+                }
```
thomaseizinger, line 274: Perhaps we should be doing this in `Drop` and therefore it would happen automatically on error too (unless the user leaks the future which is unlikely to happen).

Ralith: That makes sense to me. I see we have some existing machinery around `driver_lost`, but it doesn't reach very far; perhaps the thinking was that the connection would wind down gracefully after the idle timeout?

quinn/src/connection.rs
```diff
@@ -264,7 +259,17 @@ impl Future for ConnectionDriver {
             conn.terminate(e, &self.conn.shared);
             return Poll::Ready(Ok(()));
         }
-        let mut keep_going = conn.drive_transmit(cx)?;
+        let mut keep_going = match conn.drive_transmit(cx) {
+            Ok(keep_going) => keep_going,
+            Err(e) => {
```
djc, line 269: Should we treat some kinds of errors as transient here?

Ralith: As far as I can tell, the only path for an error to get here is from `UdpSender::poll_send`, which forwards errors from `tokio::net::udp::UdpSocket::writable` but does not forward I/O errors.

djc:
@robertkiel can you change this to use let-else?

robertkiel:
> @robertkiel can you change this to use let-else?

sure, wanted to look through everything after finishing other tasks. Also planning for the backport PRs for the other fixes.

quinn/src/connection.rs
robertkiel: Revisited the code: I don't think `Drop` catches this. The `ConnectionDriver` [drives a clone of the connection](https://github.com/quinn-rs/quinn/blob/621e38abbccdfe8d664d50b517fccec3316fe658/quinn/src/connection.rs#L58-L80) - hence the error decreases the reference count but `Drop` will run later. So treating the error early, as done in this PR, ends the connection much earlier than relying on `Drop` and timeouts. This should free resources much earlier and prevent "zombie" connections.

Also checked the errors that the underlying [`socket.writable()`](https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.writable) can produce. Seems that this is pretty much unrecoverable.

robertkiel:
revisited the code and made the code more concise

djc (approved):

quinn/src/connection.rs
thomaseizinger: > The `ConnectionDriver` [drives a clone of the connection](https://github.com/quinn-rs/quinn/blob/621e38abbccdfe8d664d50b517fccec3316fe658/quinn/src/connection.rs#L58-L80) - hence the error decreases the reference count but `Drop` will run later.

It is likely a too large refactoring for this change but I think this could be solved by introducing a `WeakConnectionRef` and handing that out? That breaks the reference cycle and allows the driver to fully own the connection's lifecycle.

djc (approved):

PR #2858, opened 2026-09-09
proto: bound sent-packet storage by live entries

zerox80:
An old unacknowledged packet can keep the sent-packet ring buffer's offset fixed while later ACK-only entries are evicted. The live-entry limit then leaves a growing span of empty slots, increasing memory use and the cost of finding the next retained packet.

Use a `BTreeMap` behind the existing `SentPackets` API so storage follows live entries. Ordered iteration, skipped packet numbers, and the constant-time in-flight counter are preserved. This trades the ring buffer's constant-time lookup for bounded sparse storage.

The regression keeps the first packet alive through 4,095 insert/remove cycles and checks the retained storage and packet state. It fails before the fix and passes afterward.

Validation:

- `cargo test --locked -p quinn-proto`: 313 unit tests and 3 documentation tests passed on Windows.

The Windows Rust 1.95 Clippy run stopped on an existing `nonminimal_bool` warning in unchanged `connection/spaces.rs`; the Linux Rust 1.98.1 run passed without suppressions.

Assisted by GPT-6 Astra.

### DIFF
--- a/quinn-proto/src/connection/sent_packets.rs
+++ b/quinn-proto/src/connection/sent_packets.rs
@@ -1,18 +1,15 @@
-use std::collections::VecDeque;
-use std::ops::{Bound, RangeBounds};
+use std::collections::BTreeMap;
+use std::ops::RangeBounds;
 
 use super::spaces::SentPacket;
 
-/// A sparse map from packet number to [`SentPacket`], backed by a ring buffer.
+/// A sparse map from packet number to [`SentPacket`].
 ///
-/// Indexed by `packet number - offset`, giving O(1) insert and lookup without
-/// the per-entry allocation of the `BTreeMap` it replaces (#2720).
+/// Storage must depend on the number of tracked packets, not the distance between
+/// packet numbers: an old unacknowledged packet can outlive many removed entries.
 #[derive(Default)]
 pub(super) struct SentPackets {
-    /// Packet number of `slots.front()` when non-empty.
-    offset: u64,
-    /// `slots[i]` holds packet number `offset + i`, or `None` if removed or skipped.
-    slots: VecDeque<Option<SentPacket>>,
+    packets: BTreeMap<u64, SentPacket>,
     /// Count of present entries with `size != 0`, for O(1) `has_in_flight`.
     in_flight: usize,
 }
@@ -20,42 +17,29 @@ pub(super) struct SentPackets {
 impl SentPackets {
     /// Insert `value` at `pn`, which must exceed every previously inserted packet number.
     pub(super) fn insert(&mut self, pn: u64, value: SentPacket) {
-        if self.slots.is_empty() {
-            self.offset = pn;
-        } else {
-            debug_assert!(
-                pn >= self.offset + self.slots.len() as u64,
-                "packet numbers must be inserted in increasing order"
-            );
-        }
-        let index = (pn - self.offset) as usize;
-        // Pad skipped packet numbers.
-        self.slots.resize(index, None);
+        debug_assert!(
+            self.packets
+                .last_key_value()
+                .is_none_or(|(&last, _)| pn > last)
+        );
         if value.size != 0 {
             self.in_flight += 1;
         }
-        self.slots.push_back(Some(value));
+        self.packets.insert(pn, value);
     }
 
     /// Remove and return the entry for `pn`.
     pub(super) fn remove(&mut self, pn: u64) -> Option<SentPacket> {
-        let index = usize::try_from(pn.checked_sub(self.offset)?).ok()?;
-        let value = self.slots.get_mut(index)?.take()?;
+        let value = self.packets.remove(&pn)?;
         if value.size != 0 {
             self.in_flight -= 1;
-        }
-        // Reclaim leading vacant slots so the buffer tracks the live window.
-        while let Some(None) = self.slots.front() {
-            self.slots.pop_front();
-            self.offset += 1;
         }
         Some(value)
     }
 
     /// Return the entry for `pn`.
     pub(super) fn get(&self, pn: u64) -> Option<&SentPacket> {
-        let index = usize::try_from(pn.checked_sub(self.offset)?).ok()?;
-        self.slots.get(index)?.as_ref()
+        self.packets.get(&pn)
     }
 
     /// Whether any present entry has `size != 0`.
@@ -68,37 +52,17 @@ impl SentPackets {
         &self,
         range: impl RangeBounds<u64>,
     ) -> impl Iterator<Item = (u64, &SentPacket)> + '_ {
-        let end = self.offset + self.slots.len() as u64;
-        let lo = Ord::max(
-            match range.start_bound() {
-                Bound::Included(&n) => n,
-                Bound::Excluded(&n) => n.saturating_add(1),
-                Bound::Unbounded => self.offset,
-            },
-            self.offset,
-        );
-        let hi = Ord::min(
-            match range.end_bound() {
-                Bound::Included(&n) => n.saturating_add(1),
-                Bound::Excluded(&n) => n,
-                Bound::Unbounded => end,
-            },
-            end,
-        );
-        let start = (lo - self.offset) as usize;
-        let stop = Ord::max(hi.saturating_sub(self.offset) as usize, start);
-        (start..stop)
-            .filter_map(move |i| self.slots[i].as_ref().map(|v| (self.offset + i as u64, v)))
+        self.packets.range(range).map(|(&pn, packet)| (pn, packet))
     }
 
     /// Mutably iterate present entries in increasing packet-number order.
     pub(super) fn values_mut(&mut self) -> impl Iterator<Item = &mut SentPacket> + '_ {
-        self.slots.iter_mut().filter_map(Option::as_mut)
+        self.packets.values_mut()
     }
 
     /// Consume the map, yielding present entries in increasing packet-number order.
     pub(super) fn into_values(self) -> impl Iterator<Item = SentPacket> {
-        self.slots.into_iter().flatten()
+        self.packets.into_values()
     }
 }
 
@@ -106,6 +70,21 @@ impl SentPackets {
 mod tests {
     use super::*;
     use crate::Instant;
+    use std::ops::Bound;
+
+    #[test]
+    fn storage_is_bounded_by_live_packets() {
+        let mut packets = SentPackets::default();
+        packets.insert(0, packet(1200));
+        for pn in 1..4096 {
+            packets.insert(pn, packet(0));
+            packets.remove(pn);
+        }
+        assert_eq!(packets.range(..).count(), 1);
+        assert!(packets.packets.len() == 1);
+        assert!(packets.has_in_flight());
+        assert_eq!(packets.get(0).unwrap().size, 1200);
+    }
 
     #[test]
     fn insert_get_and_order() {


zerox80:
I ran a local performance comparison of this PR (`d11d6155`) against its base (`65a87d40`).

Setup: Windows 11, Intel Core Ultra 7 155U, Rust 1.95.0, optimized builds, identical dependencies, fixed process affinity, and seven alternating runs per version after warmup.

Sent-packet bookkeeping, median results:

| Scenario | Before | After |
|---|---:|---:|
| 1,024 live packets, ACK batches of 32 | 26.2 ns/packet | 124.1 ns/packet |
| Successor lookup across a packet-number gap of ~65K | 147.4 µs/lookup | 13.9 ns/lookup |

The normal bookkeeping path costs approximately 98 ns more per packet in that scenario. The sparse successor lookup improves substantially. These microbenchmarks use the real `SentPacket` type with empty stream/retransmit metadata.

I also measured actual QUIC transfers over IPv4 loopback, using existing connections and two-second measurement intervals after warmup:

| Workload | Before | After |
|---|---:|---:|
| 1 × 1 MiB | 10.81 ms | 11.18 ms |
| 10 × 1 MiB concurrently | 169.41 ms | 166.90 ms |
| 1 × 1 byte | 32.47 ms | 33.55 ms |
| 100 × 1 byte concurrently | 24.44 ms | 18.89 ms |

These are medians of per-run average completion times, including acknowledgment of all streams.

There is a clear CPU trade-off in packet bookkeeping, but no consistent end-to-end slowdown in this setup. The Windows loopback measurements have substantial variability, so the apparent improvements and small regressions should not be treated as established effects. A follow-up on a quiet Linux machine would help quantify the throughput impact.

I’ll repeat these benchmarks on Fedora to check whether the results hold on Linux and better quantify the end-to-end performance impact.

zerox80:
The lookup benchmark with the ~65K packet-number gap models the condition addressed by this PR: an old packet remains retained while later entries are removed, leaving a large gap that the ring buffer must scan to find the next retained packet.

Ralith:
Would a sparse sorted `VecDeque` like we're moving to in `Assembler` get us the best of both worlds?

zerox80:
I tried a sorted VecDeque with the real SentPacket type. On Fedora, the 1,024-packet / 32-packet ACK-batch benchmark went from about 121 ns per packet with BTreeMap to 36 ns with the deque, and the sparse-gap case stayed cheap.

The caveat is removal from the middle: it shifts entries, and a synthetic test retaining an older prefix showed a large regression at 16K live entries. Those are container-level measurements, so they don't establish an end-to-end regression. The deque looks promising, but I think we should check selective ACK batches and loss handling before switching; batched removal or bounded holes with occasional compaction might help if needed.

Ralith (approved):
I think we may ultimately want something like an ordered intrusive singly linked skip list here, but given the quoted end-to-end performance numbers I'm willing to accept that a BTreeMap is a step in the right direction. Doesn't hurt that it makes the implementation simpler.

zerox80:
**AI disclosure:** The benchmark harnesses and experimental implementations, benchmark execution, analysis, and this report were produced by an AI assistant (OpenAI Codex) at my request. The measurements were run locally on my Fedora laptop; I reviewed and approved posting this comment.

Additional benchmarks compared base `65a87d40` with PR head `d11d6155` on Fedora 44 / Intel Core Ultra 7 155U.

**The sparse-storage problem is reproducible, and BTreeMap fixes it. There is also a measurable throughput trade-off on this machine.**

- Retaining one old packet across a 1,048,576-packet-number span used **100,667,536 bytes with the original ring versus 1,168 bytes with BTreeMap**, measured as fresh-process allocator deltas, not RSS.
- In the original AC-powered study, encrypted **64 MiB single-stream loopback transfers were 9.2% slower** with BTreeMap (paired 95% bootstrap interval: **−11.3% to −7.0%**), with **6.8% more process CPU time per byte**. This used twelve matched pairs; BTreeMap was slower in all twelve.
- A separate AC-powered control reproduced the direction: **−7.3% throughput [−12.1%, −3.1%]**, with **+5.5% CPU time per byte**. Both comparisons used enlarged socket buffers and reported zero packet loss.

The existing package tests passed, and these checks found no correctness issue requiring a change to the BTreeMap implementation. I would treat this as an implementation trade-off for the merge decision: the straightforward sparse-memory fix comes with a throughput cost in this workload.

These numbers are specific to one thermally stressed laptop and are not a claim that all Quinn workloads regress by the same amount. The full report below includes methodology, temperature/power limitations, and exploratory queue alternatives. The broader alternative study ran on battery and is kept separate from the AC results above; those prototypes are not presented as required changes to this PR.

<details>
<summary>Full benchmark report: methods, results, alternatives, and limitations</summary>

## Fedora benchmark report — 14 September 2026

### Scope and environment

Compared base `65a87d40c22e0818e50adfa277611610d792a43d` with PR head `d11d61556d41395aedbde7ff52e0b9e1af25aa42`. These results apply to those revisions.

- Fedora 44 Workstation; Intel Core Ultra 7 155U.
- Kernel reported by `uname`: `7.3.0-rc2.vanilla.fc44.20260908221956518389`.
- Rust 1.98.1 / LLVM 22.1.8; repository release/bench optimization settings, identical Cargo.lock, no additional RUSTFLAGS.
- Separate worktrees and Cargo target directories. Final executables were checked for distinct hashes and the expected sparse-memory behavior. Base and PR production implementations were unchanged; instrumentation was added in test/benchmark code.
- Real encrypted IPv4 loopback transfers on established connections. Client and server pinned to separate physical P-cores, CPUs 0 and 2. Every measured batch waits for FIN acknowledgment; received application byte totals are checked.
- Large socket-buffer cases request 4 MiB per buffer; Linux reports 8 MiB including kernel accounting. No system-wide network settings were changed.

Three datasets are kept separate: the original 350-process AC-powered study, a 500-process alternative study on battery, and the final 18-process AC-powered control. They are not pooled.

### Original AC-powered study: real transfers

Twelve matched base/PR pairs per scenario, alternating AB/BA order, with scenario order shuffled using a fixed seed. One second warmup and at least three seconds measurement per process. No simultaneous benchmark processes, builds, or tests.

Changes below are geometric means of within-pair ratios. Brackets are descriptive 95% percentile bootstrap intervals over the twelve pairs, using 20,000 resamples. They are not corrected for multiple comparisons; inner iterations are not independent replicates.

| Workload | PR throughput change [95% interval] | PR process CPU time/byte change [95% interval] |
|---|---:|---:|
| 1 MiB × 1 stream, default buffers | -7.4% [-11.0%, -3.3%] | +5.2% [+2.6%, +7.5%] |
| 1 MiB × 10 streams, default buffers | -8.2% [-10.9%, -5.6%] | +6.7% [+4.9%, +8.6%] |
| 1 byte × 1 stream, default buffers | -0.2% [-0.5%, +0.1%] | +2.7% [-1.2%, +6.8%] |
| 1 byte × 100 streams, default buffers | -0.6% [-5.6%, +4.4%] | +5.8% [+3.5%, +7.9%] |
| 1 MiB × 1 stream, large buffers | -5.5% [-8.1%, -2.9%] | +4.1% [+1.9%, +6.6%] |
| 64 MiB × 1 stream, large buffers | -9.2% [-11.3%, -7.0%] | +6.8% [+4.8%, +8.7%] |

The large-buffer cases reported zero packet loss. Default-buffer throughput cases had declared packet losses, making them less clean CPU comparisons. The one-byte cases are dominated by delayed ACK timing and are weak evidence about small latency changes.

In the 64 MiB large-buffer case the PR was slower in all twelve pairs. The paired throughput changes were −9.4% with base first and −8.9% with PR first; the first and second halves were −8.0% and −10.3%. Process CPU time includes both endpoints and their system calls. Declared loss is Quinn's `PathStats.lost_packets`, not a direct kernel-drop measurement.

### Original AC-powered study: packet storage

Microbenchmarks use the actual internal production types. Dense cycles look up the largest ACKed entry, collect an ACK range, remove a batch, insert replacements, and inspect the in-flight count. `meta` adds real inline STREAM metadata; `reorder` reverses deletion order within a batch and is not a network-reordering simulation. Successor lookup is timed after setup of a sparse span. Values are medians in ns per harness operation; they must not be interpreted as whole-transfer slowdowns.

| Case | Base ns/operation | BTreeMap ns/operation |
|---|---:|---:|
| `dense_1024_1` | 50.53 | 221.11 |
| `dense_1024_32` | 38.84 | 191.71 |
| `dense_1024_32_meta` | 38.96 | 191.11 |
| `dense_16384_32` | 38.57 | 232.09 |
| `dense_64_32` | 37.26 | 147.18 |
| `reorder_1024_32` | 39.45 | 200.00 |
| `successor_4096` | 6,565.44 | 12.48 |
| `successor_65536` | 155,334.20 | 12.19 |

Memory measurements use fresh-process glibc allocator deltas, `mallinfo2().uordblks + hblkhd`, including mmap allocations. These are not RSS or total connection memory; allocator caches can retain small freed blocks.

| Case | Live entries | Base allocated bytes | BTreeMap allocated bytes |
|---|---:|---:|---:|
| Dense window of 1,024 | 1,024 | 98,320 | 202,128 |
| Dense window of 16,384 | 16,384 | 1,577,008 | 3,223,552 |
| Sparse span of 65,536 packet numbers | 1 | 6,295,632 | 1,168 |
| Sparse span of 1,048,576 packet numbers | 1 | 100,667,536 | 1,168 |
| Actual ACK-tail caller, 65,536 packet numbers | 1,002 | 6,295,632 | 197,360 |

The actual `PacketSpace::sent()` ACK-tail reproduction took 4.518 s with base and 0.007574 s with BTreeMap, retaining the same 1,002 entries. This is one deterministic reproduction per version, not a repeated statistical timing experiment or a network exploit test.

### Final AC-powered control

Six matched rounds covering all six permutations of base, BTreeMap, and an experimental bounded-tombstone queue: 18 successful processes. Same 64 MiB single-stream workload and P-core placement, one second warmup, at least four seconds measurement, two seconds pause between processes. AC was checked every 250 ms while each process ran; losing AC would abort and exclude the sample. All recorded runs remained on AC and the performance profile, passed transfer checks, and reported zero packet loss.

| Comparison | Throughput change [95% interval] | Process CPU time/byte change [95% interval] |
|---|---:|---:|
| BTreeMap / base | -7.3% [-12.1%, -3.1%] | +5.5% [+1.9%, +9.3%] |
| Bounded queue / base | -1.7% [-4.9%, +2.3%] | +1.5% [-1.6%, +4.1%] |
| Bounded queue / BTreeMap | +6.1% [+0.8%, +12.5%] | -3.8% [-6.9%, -0.8%] |

Intervals use 10,000 bootstrap resamples of the six matched rounds. Median throughput was 2.281 Gbit/s for base, 2.095 for BTreeMap, and 2.172 for the bounded queue; percentage changes use paired ratios, not ratios of these medians. Absolute throughput differs considerably from the earlier study, which is another reason not to pool them. These six rounds provide a directional check, not a cross-platform guarantee.

### Additional alternatives: separate battery-powered study

All 500 processes in this study ran on battery: AC was 0 at all 1,000 recorded process boundaries. AC had been present during preparation, but the runner did not enforce it at the start. The performance profile remained enabled. These results are not described as an AC replication.

Four versions were tested in eight matched rounds per timed scenario using balanced Williams orders: base, BTreeMap, a sorted `VecDeque<(u64, SentPacket)>`, and a sorted `VecDeque<(u64, Option<SentPacket>)>` with bounded tombstones. Both prototypes use safe Rust and preserve the existing API. The latter marks removals, drops front tombstones, compacts above `max(128, 2 × live)` slots, and shrinks excess capacity with hysteresis. Packet-number gaps do not allocate padding.

The simple queue shifts elements on middle removal. The bounded queue avoids that routine shifting but adds eight bytes per slot on this platform, can scan tombstones during iteration/successor lookup, and has O(n) compaction/shrinking operations. Individual compaction pauses have not been separately measured.

Microbenchmarks below ran on CPU 4. `selective_N` retains the older half while deleting the newer half; `random_N` deletes in a fixed random order. Both time complete repeated fill/drain cycles with STREAM metadata, normalized by packet count. Values are median ns per harness operation.

| Case | Base | BTreeMap | Simple queue | Bounded queue |
|---|---:|---:|---:|---:|
| `dense_1024_32` | 60.47 | 194.16 | 62.71 | 61.51 |
| `dense_1024_32_meta` | 69.31 | 212.47 | 64.03 | 63.35 |
| `dense_16384_32` | 61.88 | 257.91 | 72.52 | 80.74 |
| `dense_64_32` | 53.47 | 172.71 | 57.02 | 54.62 |
| `random_1024` | 68.30 | 410.87 | 829.82 | 165.69 |
| `random_16384` | 63.81 | 632.48 | 13,362.31 | 200.06 |
| `selective_1024` | 69.85 | 261.78 | 811.79 | 120.37 |
| `selective_16384` | 63.95 | 349.73 | 13,300.97 | 115.71 |
| `successor_65536` | 251,185.16 | 15.04 | 17.40 | 153.16 |

The simple queue is particularly expensive under selective or random removal at 16K entries. The bounded queue is promising but does not match the base ring in every microbenchmark. Its sparse allocation was 26,640 bytes in both the 65,536- and 1,048,576-span cases. In the actual ACK-tail case it used 217,088 bytes versus BTreeMap's 197,360 bytes; bounded memory does not mean universally lower memory.

Real 64 MiB transfers were also measured separately on E-cores 4/5 and P-cores 0/2, with one second warmup and at least three seconds measurement. Eight paired rounds per phase, 10,000 bootstrap resamples:

| Phase | BTreeMap / base throughput | Bounded queue / base throughput | Bounded queue / BTreeMap throughput |
|---|---:|---:|---:|
| E-cores | -5.3% [-10.6%, +1.1%] | +0.8% [-2.4%, +4.1%] | +6.5% [-1.3%, +14.4%] |
| P-cores | -8.7% [-14.4%, -3.5%] | -0.7% [-5.4%, +4.9%] | +8.8% [+3.9%, +13.9%] |

Protocol simulations transferred and checked 8 MiB of encrypted data with a 4 MiB initial congestion window and large flow-control windows. Handshake and full payload checks were outside the processing timer. Impairments began after handshake, with a shared deterministic datagram counter affecting both directions. Measured wall time includes simulator overhead. No link-bandwidth or queue-capacity model was included, so these are processing-cost comparisons, not WAN throughput predictions.

| Simulation | Bounded queue / BTreeMap processing time per byte [95% interval] |
|---|---:|
| No added RTT | -6.1% [-18.2%, +7.9%] |
| 50 ms RTT | -8.0% [-17.7%, +2.9%] |
| 50 ms RTT, every 100th datagram dropped | -12.1% [-18.0%, -6.3%] |
| 50 ms RTT, every 16th datagram delayed another 3 ms | -10.0% [-19.9%, -1.3%] |

The zero-RTT simulator can produce an undefined virtual rate. Its `inf` output was normalized to JSON `null`; the affected initial unrecorded attempt was repeated, and valid prior samples were retained. The measured processing time was unaffected. All 128 recorded simulation runs and 64 real-transfer runs passed their payload/completion checks.

### Correctness and limits

The unmodified PR passed 354 unit/integration tests and four doc-tests for `quinn-proto`, `quinn`, and `quinn-udp`; four pre-existing tests were ignored. Separate release `quinn-proto` runs passed on base and PR. Both prototypes passed 356 unit/integration tests and four doc-tests each. Additional coverage includes large packet-number gaps, shrinking after a burst, retention of an old packet, and twelve deterministic 8,000-operation traces per version against a reference model. These checks found no correctness issue in BTreeMap but cannot prove the absence of all bugs.

The prototypes are experimental: formatting and Clippy checks remain outstanding because the components are not installed locally, and worst individual compaction pauses still need evaluation. They are not presented as required changes to this PR.

This is one thermally stressed desktop laptop, not an isolated fixed-frequency benchmark host. Package temperatures were 95–101 °C in the original AC study, 96–103 °C in the battery study, and 96–101 °C in the AC control. Thermal-throttling counters increased during the original study. CPU affinity is not exclusive isolation; GNOME and background processes remained active. Pairing and balanced order help but do not remove thermal/frequency effects. No claim is made about other CPUs, real WAN performance, energy consumption, or universal superiority of a data structure.

The evidence supports treating BTreeMap as a straightforward fix for the sparse-storage problem with a measurable throughput cost on this machine. Keeping BTreeMap is a reasonable implementation choice if that trade-off is acceptable; these benchmarks do not establish a correctness change that must be made before merging.

</details>

zerox80:
I followed up with an experimental alternative in a draft PR in my fork:

https://github.com/zerox80/quinn/pull/11

It now contains the implementation, regression tests, reproducible Fedora benchmark harnesses, raw observations and the full report. Disclosure: OpenAI Codex implemented the prototypes and harnesses, performed the measurements under my direction, and prepared the report and PR text.

The design uses mutable edge blocks and packed, boxed interior blocks of at most 64 slots. It avoids full-window compaction; sparse fragmented storage with 1,024 survivors measured 182,976 bytes, versus 198,528 for BTreeMap and 327,536 for the earlier unpacked block prototype.

I also addressed the previously unfavorable successor lookup directly. The real `PacketSpace::sent` call now uses a specialized `first_after(pn)` path. In the original two-live-entry / 65,536-number-gap microbenchmark, the current candidate measured 8.46 ns per query, versus 11.13 ns for BTreeMap and 31.35 ns for the previously published packed version. This is specific to the successor query; it does not make every general range scan faster.

In twelve matched triplets of encrypted 64 MiB loopback transfers, the current candidate was +4.5% throughput [+2.1%, +6.6%] versus BTreeMap. Its difference from the previous packed version was -1.0% [-4.1%, +1.7%], so that comparison remains unresolved. The report also retains a measured cost: a short 32-entry general range scan was 10.5% slower than the previous packed version.

Across 18 constructed cleanup cases, the largest current observed cleanup operation was 1.377 us. These are operation-level diagnostics, not network p99 measurements or hard latency guarantees.

The design still adds complexity over BTreeMap, so keeping the simpler BTreeMap fix as the immediate upstream outcome remains reasonable. I would appreciate feedback on whether this bounded packed-block direction seems worth pursuing further.

zerox80:
I think with a bit more fine-tuning, this direction could become quite interesting. I'd really appreciate your thoughts on whether the overall approach seems worth pursuing further.

It's clearly not perfect yet, but the current results suggest there may still be meaningful room for improvement.

djc (approved):

djc:
Always interested in meaningful latency/throughput/memory-use improvements. For me personally, I'd need a better understanding of the benchmark impact across throughput, memory use and complexity (maybe expressed in lines of code) -- ideally in a table or chart (not just a prose summary) contrasting the approach before this PR with the `BTreeMap` with your new attempt.

PR #2859, opened 2026-09-09
streams: bound retained send storage by the send window

zerox80:
Selective acknowledgments can restore `send_window` credit while the acknowledged suffix remains allocated behind an earlier gap. A continuing writer can therefore retain more data than the configured send window permits.

The commits separate two related changes:

1. Release abandoned buffers on reset, preserving the final offset and ignoring later loss notifications for reset streams.
2. Charge the send window for retained storage and refund it only when storage is released. This includes acknowledged suffixes and a trimmed prefix that still shares the first segment's allocation.

This deliberately applies backpressure while acknowledged data is still retained. It preserves stream offsets, retransmission, flow control, and existing zero-copy application buffers. The accounting covers the bytes supplied in each application slice; it cannot infer a larger allocation an application chose to slice before writing.

Regression tests fail before the changes and pass afterward. They cover suffix acknowledgments, a partially released prefix, restored write credit after full acknowledgment, reset storage release, final-offset preservation, and late ACK/loss notifications.

Validation:

- `cargo test --locked -p quinn-proto`: 314 unit tests and 3 documentation tests passed on Windows.

Assisted by GPT-6 Astra.

### DIFF
--- a/quinn-proto/src/connection/send_buffer.rs
+++ b/quinn-proto/src/connection/send_buffer.rs
@@ -155,6 +155,15 @@ impl SendBuffer {
         &[]
     }
 
+    /// Release abandoned data while preserving the final offset for RESET_STREAM
+    pub(super) fn discard(&mut self) {
+        *self = Self {
+            offset: self.offset,
+            unsent: self.offset,
+            ..Self::default()
+        };
+    }
+
     /// Queue a range of sent but unacknowledged data to be retransmitted
     pub(super) fn retransmit(&mut self, range: Range<u64>) {
         debug_assert!(range.end <= self.unsent, "unsent data can't be lost");
--- a/quinn-proto/src/connection/send_buffer.rs
+++ b/quinn-proto/src/connection/send_buffer.rs
@@ -11,6 +11,8 @@ pub(super) struct SendBuffer {
     unacked_segments: VecDeque<Bytes>,
     /// Total size of `unacked_segments`
     unacked_len: usize,
+    /// Acknowledged bytes removed from the first segment's view but still held by its allocation
+    front_trimmed: usize,
     /// The first offset that hasn't been written by the application, i.e. the offset past the end of `unacked`
     offset: u64,
     /// The first offset that hasn't been sent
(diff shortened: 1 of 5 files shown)


Ralith (approved):
I'm a bit concerned this might surprise/confuse folks, but I suppose it's the safer default.

Alternatively, we could implement a defragmentation scheme similar to Assembler, where we realloc and copy into a single buffer if the proportion of used storage falls below a threshold.

quinn-proto/src/connection/send_buffer.rs
```diff
@@ -188,6 +188,15 @@ impl SendBuffer {
...
     }
+
+    /// Release abandoned data while preserving the final offset for RESET_STREAM.
+    pub(super) fn discard(&mut self) {
+        *self = Self {
+            offset: self.offset,
+            unsent: self.offset,
+            ..Self::default()
+        };
+    }
```
djc: Nits: move this above `retransmit()`; the documentation comment should not end in a period.

quinn-proto/src/connection/send_buffer.rs
```diff
@@ -11,6 +11,8 @@ pub(super) struct SendBuffer {
     unacked_segments: VecDeque<Bytes>,
     /// Total size of `unacked_segments`
     unacked_len: usize,
+    /// Acknowledged bytes removed from the first segment's view but still held by its allocation.
```
djc: Nit: match surrounding style by omitting period on the documentation comment.

quinn-proto/src/connection/send_buffer.rs
```diff
@@ -184,9 +188,9 @@ impl SendBuffer {
         self.unsent != self.offset || !self.retransmits.is_empty()
     }
 
-    /// Compute the amount of data that hasn't been acknowledged
-    pub(super) fn unacked(&self) -> u64 {
-        self.unacked_len as u64 - self.acks.iter().map(|x| x.end - x.start).sum::<u64>()
+    /// Bytes still retained from application writes, including acknowledged data.
```
djc: Same here.

djc (approved):

zerox80:
@djc Moved `discard()` above `retransmit()` and removed the trailing periods from all three doc comments.

@Ralith i have clarified the `send_window` docs to explain why acknowledged data can keep writes blocked until its storage is released. Defragmentation sounds worth exploring as a follow-up.

djc:
We're planning to release 0.11.18 soon, do you want to backport this?

PR #2860, opened 2026-09-09
proto: enforce packet capacity for large Initial tokens

zerox80:
An Initial header containing a large Retry or cached token can exceed the outgoing packet capacity, or leave too little space to consume pending CRYPTO data. Packet construction then violates its size or progress assumptions.

At the shared Initial packet builder, require enough capacity for header protection and frame construction. If the token cannot fit, discard the partial header and terminate the connection with an observable transport error. This covers both Retry and token-store inputs and respects a configured larger initial MTU.

The cached-token regression panics before the fix and passes afterward. Tests also cover an integrity-valid Retry, headers that leave insufficient frame space, ordinary tokens, an empty token, a larger configured MTU, and subsequent transmit polls after rejection. A separate regression covers a large token with pending ACK ranges during close; ACK encoding reserves space for CONNECTION_CLOSE. This regression also fails before the close-path adjustment and passes afterward.

Validation:

- `cargo test --locked -p quinn-proto`: 315 unit tests and 3 documentation tests passed on Windows.
- `cargo test --locked -p quinn-proto --release token`: all 23 token-related tests passed.

Assisted by GPT-6 Astra.


quinn-proto/src/tests/token.rs
```diff
@@ -2,6 +2,100 @@
...
+        let (_, mut conn) = endpoint
+            .connect(now, client_config(), address, "localhost")
+            .unwrap();
+        let mut buf = Vec::new();
+        let _ = conn.poll_transmit(now, 1, &mut buf).unwrap();
+        let version = u32::from_be_bytes(buf[1..5].try_into().unwrap());
+        let dcid_len = buf[5] as usize;
+        let original_cid = ConnectionId::new(&buf[6..6 + dcid_len]);
+        let scid_len = buf[6 + dcid_len] as usize;
+        let client_cid = ConnectionId::new(&buf[7 + dcid_len..7 + dcid_len + scid_len]);
```
Ralith: Can we reuse `ProtectedHeader::decode` or similar rather than open-coding this?

quinn-proto/src/connection/packet_builder.rs
```diff
@@ -152,7 +152,21 @@ impl PacketBuilder {
...
-        let max_size = buffer_capacity - tag_len;
+        let max_size = buffer_capacity.saturating_sub(tag_len);
+        if space_id == SpaceId::Initial
+            && (max_size < min_size
+                || max_size.saturating_sub(buffer.len())
+                    <= frame::Crypto::SIZE_BOUND.max(frame::ConnectionClose::SIZE_BOUND))
+        {
+            buffer.truncate(partial_encode.start);
+            conn.kill(
+                TransportError::PROTOCOL_VIOLATION(
```
Ralith: A protocol violation is an error made by the peer. This is an error in the configuration of the local endpoint, so we should probably use `INTERNAL_ERROR`.

quinn-proto/src/connection/mod.rs
```diff
@@ -826,13 +826,12 @@ impl Connection {
                         &mut self.spaces[space_id],
                         buf,
                         &mut self.stats,
-                        buf_capacity,
+                        builder.max_size - frame::ConnectionClose::SIZE_BOUND - 1,
```
Ralith, line 832: Why - 1? Document the reasoning behind this arithmetic.

quinn-proto/src/connection/mod.rs
```diff
@@ -826,13 +826,12 @@ impl Connection {
...
                     );
                 }
 
-                // Since there only 64 ACK frames there will always be enough space
-                // to encode the ConnectionClose frame too. However we still have the
-                // check here to prevent crashes if something changes.
+                // Reserve room for CONNECTION_CLOSE even when an Initial token
+                // leaves little space after the header.
```
Ralith: This comment is confusing. `debug_assert` doesn't reserve anything. Did you mean for it to be on `try_populate_acks`?

Ralith (changes requested):

quinn-proto/src/connection/mod.rs
zerox80: the explanation can be found in the surrounding code. There is both an assertion and an actual `if` condition using a strict `<` comparison:

```rust
if buf.len() + frame::ConnectionClose::SIZE_BOUND < builder.max_size {
    // The close frame is written here.
}
```

so with the existing code, it is not enough for the reserved space to fit exactly. the sum must remain strictly smaller than `builder.max_size`, which is why the extra `- 1` is needed.

quinn-proto/src/connection/packet_builder.rs
zerox80: Ur right here.

zerox80:
is it fine now?

Ralith:
Will re-review when able; we're under a lot of load right now.

quinn-proto/src/connection/packet_builder.rs
```diff
@@ -152,7 +152,19 @@ impl PacketBuilder {
...
-        let max_size = buffer_capacity - tag_len;
+        let max_size = buffer_capacity.saturating_sub(tag_len);
+        if space_id == SpaceId::Initial
+            && (max_size < min_size
+                || max_size.saturating_sub(buffer.len())
+                    <= frame::Crypto::SIZE_BOUND.max(frame::ConnectionClose::SIZE_BOUND))
+        {
+            buffer.truncate(partial_encode.start);
+            conn.kill(
+                TransportError::INTERNAL_ERROR("Initial token leaves insufficient packet space")
```
Ralith: Actually, on second thought, isn't this the peer's responsibility after all? Address validation tokens are supplied by the remote endpoint in a previous connection.

quinn-proto/src/connection/mod.rs
```diff
@@ -819,20 +819,20 @@ impl Connection {
                 // especially important with ack delay, since the peer might not
                 // have gotten any other ACK for the data earlier on.
                 if !self.spaces[space_id].pending_acks.ranges().is_empty() {
+                    // Reserve room for CONNECTION_CLOSE even when an Initial token leaves
+                    // little space after the header. The extra byte satisfies the strict
+                    // `< builder.max_size` check below; try_populate_acks permits an exact fit.
```
Ralith: Should the check below be <=?

quinn-proto/src/connection/packet_builder.rs
zerox80: Agreed. Both Retry tokens and cached address validation tokens originate from the peer; the token store only retains them. I've changed the reported error to PROTOCOL_VIOLATION and updated both regression tests accordingly.

quinn-proto/src/connection/mod.rs
zerox80: Yes, <= is sufficient here. I've updated both checks and removed the extra - 1 from the ACK budget. The Initial packet guard now also accepts an exact fit for CONNECTION_CLOSE while still requiring room for CRYPTO data. The regression test decodes the emitted packet and verifies that an exactly fitting ACK is retained, an oversized ACK is omitted, and the close frame is present in both cases.

quinn-proto/src/tests/token.rs
zerox80: This is already updated: the Retry regression test uses ProtectedHeader::decode to obtain the version and connection IDs from the Initial header.

Ralith (approved):

quinn-proto/src/connection/packet_builder.rs
```diff
@@ -152,7 +152,22 @@ impl PacketBuilder {
             buffer.len() + (sample_size + 4).saturating_sub(number.len() + tag_len),
             partial_encode.start + dst_cid.len() + 6,
         );
-        let max_size = buffer_capacity - tag_len;
+        let max_size = buffer_capacity.saturating_sub(tag_len);
+        if space_id == SpaceId::Initial
+            && (max_size < min_size
+                || max_size.saturating_sub(buffer.len())
+                    <= frame::Crypto::SIZE_BOUND.max(frame::ConnectionClose::SIZE_BOUND))
```
djc, line 159: I think this needs some comments on what this is trying to solve (and, potentially, how) -- seems like very dense condition. Also, please use a symmetric `max()` like `Ord::max(frame::Crypto::SIZE_BOUND, frame::ConnectionClose::SIZE_BOUND)`.

quinn-proto/src/connection/mod.rs
```diff
@@ -819,25 +819,24 @@ impl Connection {
...
-                // Since there only 64 ACK frames there will always be enough space
-                // to encode the ConnectionClose frame too. However we still have the
-                // check here to prevent crashes if something changes.
                 debug_assert!(
-                    buf.len() + frame::ConnectionClose::SIZE_BOUND < builder.max_size,
+                    buf.len() + frame::ConnectionClose::SIZE_BOUND <= builder.max_size,
                     "ACKs should leave space for ConnectionClose"
                 );
-                if buf.len() + frame::ConnectionClose::SIZE_BOUND < builder.max_size {
+                if buf.len() + frame::ConnectionClose::SIZE_BOUND <= builder.max_size {
```
djc, line 839: If we're asserting the condition, why do we have a normal test for it directly after? Seems strange.

PR #2862, opened 2026-09-09
proto: detach received DATAGRAM payloads from packet storage

zerox80:
A received DATAGRAM payload is a `Bytes` slice that can keep an entire packet or receive batch allocated. The receive queue only charges the visible payload and entry overhead, so small queued payloads can retain substantially more memory than the configured budget.

Copy accepted receive payloads into independent storage at `DatagramState::received`. This leaves outgoing application-owned datagrams unchanged and preserves receive ordering and eviction behavior. It adds one copy for each accepted incoming DATAGRAM.

Two regressions verify that a large backing owner is released while queued payloads remain correct, including shared storage and eviction. Both fail before the fix and pass afterward. The existing empty-frame budget regression also passes.

Validation:

- `cargo test --locked -p quinn-proto`: 314 unit tests and 3 documentation tests passed on Windows.

Assisted by GPT-6 Astra.

### DIFF
--- a/quinn-proto/src/connection/datagrams.rs
+++ b/quinn-proto/src/connection/datagrams.rs
@@ -129,7 +129,11 @@ impl DatagramState {
             self.recv();
         }
 
-        self.incoming.push_back(datagram);
+        // A small payload can otherwise retain an entire receive batch, which is not
+        // included in the queue's memory budget.
+        self.incoming.push_back(Datagram {
+            data: Bytes::copy_from_slice(&datagram.data),
+        });
         Ok(was_empty)
     }
 
@@ -241,6 +245,43 @@ impl DatagramBuffer {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::sync::Arc;
+
+    #[test]
+    fn received_datagram_releases_packet_storage() {
+        let storage: Arc<[u8]> = vec![42; 64 * 1024].into();
+        let data = Bytes::from_owner(storage.clone()).slice(100..101);
+        let mut state = DatagramState::default();
+
+        assert!(state.received(Datagram { data }, &Some(100)).unwrap());
+        assert_eq!(Arc::strong_count(&storage), 1);
+        assert_eq!(state.recv().unwrap().as_ref(), &[42]);
+        assert!(state.recv().is_none());
+    }
+
+    #[test]
+    fn received_datagrams_release_shared_storage_after_eviction() {
+        let storage: Arc<[u8]> = (0..=255).collect::<Vec<u8>>().into();
+        let packet = Bytes::from_owner(storage.clone());
+        let mut state = DatagramState::default();
+        let window = Some(2 * (size_of::<Datagram>() + 1));
+        for index in 0..3 {
+            state
+                .received(
+                    Datagram {
+                        data: packet.slice(index..index + 1),
+                    },
+                    &window,
+                )
+                .unwrap();
+        }
+        drop(packet);
+
+        assert_eq!(Arc::strong_count(&storage), 1);
+        assert_eq!(state.recv().unwrap().as_ref(), &[1]);
+        assert_eq!(state.recv().unwrap().as_ref(), &[2]);
+        assert!(state.recv().is_none());
+    }
 
     #[test]
     fn make_space_for_accounts_for_new_datagram() {
--- a/quinn-proto/src/connection/datagrams.rs
+++ b/quinn-proto/src/connection/datagrams.rs
@@ -79,7 +79,7 @@ impl Datagrams<'_> {
 
     /// Receive an unreliable, unordered datagram
     pub fn recv(&mut self) -> Option<Bytes> {
-        self.conn.datagrams.recv()
+        self.conn.datagrams.recv().map(Bytes::from)
     }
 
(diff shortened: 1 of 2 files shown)


quinn-proto/src/connection/datagrams.rs
```diff
@@ -129,7 +129,11 @@ impl DatagramState {
             self.recv();
         }
 
-        self.incoming.push_back(datagram);
+        // A small payload can otherwise retain an entire receive batch, which is not
+        // included in the queue's memory budget.
+        self.incoming.push_back(Datagram {
+            data: Bytes::copy_from_slice(&datagram.data),
```
Ralith: Does it make sense to still use `Bytes` at all, if we're copying unconditionally? We'll still need to in 0.11.x for API compatibility, but maybe it'd be clearer not to internally even there.

Ralith (commented):
I'm on board with this in principle (I've been thinking of doing similar for streams), but I'd like to see some performance data to understand what we're giving up (if anything).

zerox80:
> I'm on board with this in principle (I've been thinking of doing similar for streams), but I'd like to see some performance data to understand what we're giving up (if anything).

I measured the cost and changed the incoming queue to `Box<[u8]>`. The public receive API still returns `Bytes`, transferring the allocation without another payload copy. Queue-entry metadata drops from 32 to 16 bytes on x86-64; accounting and eviction tests now reflect that.

I compared the PR parent, the original PR, and the revised version. Median ns/datagram over nine runs on an Intel Core Ultra 7 155U, Linux, Rust 1.98.1, release builds, pinned to one logical CPU with rotating variant order:

| Payload | Before fix | Original PR | Revised | Revised vs before |
| ---: | ---: | ---: | ---: | ---: |
| 64 B | 271.9 | 303.3 | 297.1 | +9.3% |
| 256 B | 468.4 | 511.3 | 510.5 | +9.0% |
| 1200 B | 1469.5 | 1548.5 | 1517.7 | +3.3% |

This uses the in-memory protocol test pair with real crypto, batches of 32, and payload/count verification. Handshake and network I/O are excluded. There is run-to-run variation, particularly at 1200 B, so these percentages aren't universal throughput estimates.

For the isolated receive queue, batching and draining 32 entries costs 19.2 → 41.5 ns/datagram at 64 B and 19.2 → 78.7 ns at 1200 B (before fix → revised). The memory fix does have a measurable allocation/copy cost; boxes reduce some of the bookkeeping overhead.

The storage-release regressions, updated eviction tests, and new receive-accounting test pass, as do the quinn-proto/quinn suites, formatting, and Clippy.

quinn-proto/src/connection/datagrams.rs
zerox80: I agree,  the incoming queue now stores Box<[u8]>, converting to Bytes only at the public API boundary without another payload copy.

djc:
Doesn't seem all that attractive... maybe it should be configurable?

zerox80:
> AI-assisted measurements; benchmark harness and runs prepared with Codex.
>
> Real QUIC/UDP over loopback, with a 64 KiB DATAGRAM receive budget. The receiver queues 512 datagrams of 64 B each while continuously consuming 32 MiB of stream data.
>
> Median additional live Rust heap across five runs per revision:
>
> - Before: 32.62 MiB
> - Revised PR: 43.6 KiB
>
> Draining those 512 datagrams releases 32.61 MiB before the fix, versus 32 KiB with the PR. This confirms that the queued datagrams are retaining the larger receive allocations.
>
> There is also a counterexample: with the same 512 datagrams but no accompanying stream traffic, additional heap increases from 24.3 KiB to 48.3 KiB with the PR.
>
> These are heap deltas relative to the post-handshake baseline, not total RSS; allocator overhead and kernel buffers are excluded.

the benefit is clearly workload-dependent, but the retention issue can be substantial when datagrams remain queued alongside otherwise-consumed traffic.

making copying configurable sounds reasonable to me. From an API usability perspective, I'd still like a sensible default, with clear documentation explaining when application developers should choose the other behavior. What default would you favor given this trade-off? @Ralith what do u think?

Ralith:
> the retention issue can be substantial when datagrams remain queued alongside otherwise-consumed traffic.

Is this a realistic use case? I'd expect applications to consume all forms of application data in a timely manner.

> I'd still like a sensible default, with clear documentation explaining when application developers should choose the other behavior.

Strongly agree that this should be a priority. A knob that applications are unlikely to set correctly does more harm than good, even before considering our maintenance costs

> What default would you favor given this trade-off? @Ralith what do u think?

I'm not yet convinced that a 10% increase in CPU overhead on a hot path is justified by reduced memory use for applications which accept datagrams without consuming them in a timely manner, which does not seam realistic. However, I wonder if we might get the time back if the receive path is modified to avoid copying all UDP datagrams up front?

https://github.com/quinn-rs/quinn/blob/621e38abbccdfe8d664d50b517fccec3316fe658/quinn/src/endpoint.rs#L881

When incoming application data is retained, this will often realloc in practice, and is undoubtedly imposing some cost. Trying that experiment would also impact streams, but if we can demonstrate end-to-end nonregression of application datagram receive performance, that would help motivate any necessary work to further revise the stream receive path as well.

PR #2872, opened 2026-09-15
Take semver-compatible dependency updates

djc:
Notably, a rustls fix for https://rustsec.org/advisories/RUSTSEC-2026-0285.html.


Ralith (approved):

djc:
Going to deal with the test-android failures in

- https://github.com/quinn-rs/quinn/pull/2873

PR #2874, opened 2026-09-16
quinn: refactor Connection::drive_transmit to use fewer mutations

robertkiel:
### Overview

This PR slightly refactors `Connection::drive_transmit` to use fewer writes to `self.buffered_transmit` by removing the  take-and-put-back pattern. Pseudocode:

```
if existing_work {
    finish_work
}

if new_work_available {
    finish work
    if pending {
        buffer work
    }
}
```

The version is slightly more robust because it removes the windows where a previous attempt is owned locally.


djc:
Does this solve an actually observed problem? I find the changed code substantially harder to follow.

robertkiel:
> Does this solve an actually observed problem? I find the changed code substantially harder to follow.

The code originates from an analysis of the `drive_transmit` function. It turned out that it led to more writes than necessary. 

In the previous version, `drive_transmit` writes unconditionally. It first `take()`s out the last attempt and - if transmitting fails - the value of `self.buffered_transmit` is gone.

djc:
> The code originates from an analysis of the `drive_transmit` function. It turned out that it led to more writes than necessary.

Why are "more writes than necessary" a problem, though? What is the observed effect on your processes?

robertkiel:
Did some benchmarks and it shows that this implementation is about 30% faster than the previous one in an isolated test. It gives 5 to 7 ns. Shall I add the benchmark code?

djc:
Does it show up in macrobenchmarks like in the perf crate?

robertkiel:
> Does it show up in macrobenchmarks like in the perf crate?

It is about 0.2%

djc:
Doesn't seem worth it to me... @Ralith thoughts?

robertkiel:
Will recheck how to increase readability. Probably worth to delay this a bit.

robertkiel:
refactored the code, should be more readable now

PR #2877, opened 2026-09-21
fix: potential CID leak in connect()

robertkiel:
When connecting to an endpoint, `connect()` samples a local connection id and [stores](https://github.com/quinn-rs/quinn/blob/d10c95bdc51a54989113e0e106b2fe0a1f613e6b/quinn-proto/src/endpoint.rs#L421) it into its `connection_ids` map. 

However, if `crypto.start_session` fails, the just created CID leaks since `self.add_connection` is called later.

[code](https://github.com/quinn-rs/quinn/blob/d10c95bdc51a54989113e0e106b2fe0a1f613e6b/quinn-proto/src/endpoint.rs#L355-L366)
```rust
let loc_cid = self.new_cid(RouteDatagramTo::Connection(ch));
let params = TransportParameters::new(
    &config.transport,
    &self.config,
    self.local_cid_generator.as_ref(),
    loc_cid,
    None,
    &mut self.rng,
);
let tls = config
    .crypto
    .start_session(config.version, server_name, &params))?; // if errors, the CID leaks
```

This can be easily fixed by removing the `loc_cid` in case `start_session` fails.

```rust
let tls = config
    .crypto
    .start_session(config.version, server_name, &params)
    .inspect_err(|_| {
        // Remove just issued connection id
        self.index.connection_ids.remove(&loc_cid);
    })?;
```


djc (approved):
Thanks, this makes sense.

Should we backport this to `0.11.x`, too?

Ralith (approved):

